PR #875, opened 2011-12-23
Draw spiky stars for faraway cities and orbital stations as in Frontier.

WaveMotion:
Primarily, this makes it easier for manual fliers to determine whether a station is on your side of a planet or on the other side, even if you are far away or the station is on the dark side of a planet.

To make it work correctly, I had to make Camera update TerrainBodies' camDist based on distance to the horizon instead of to the center, which is unfortunate but shouldn't be too terrible computationally.

Some things that might need to be worked on:
- Adjust the size of the spikes relative to those of planets and asteroids. In Frontier, spikes for stations and ships were tiny five-pixel blips.
- As you approach a station, there is a noticeable gap between when the spike disappears and when the station starts becoming visible. I'm not sure how much can be adjusted to fix that, since there are many factors that can thwart the visibility of a city (e.g. antialiasing, terrain updating, atmosphere, FOV settings).
- Rewrite DrawSpike such that it doesn't have to resort to shenanigans like disabling depth testing.

### DIFF
--- a/src/Camera.cpp
+++ b/src/Camera.cpp
@@ -120,7 +120,16 @@ void Camera::Update()
 		attrs.body = b;
 		Frame::GetFrameRenderTransform(b->GetFrame(), m_camFrame, attrs.viewTransform);
 		attrs.viewCoords = attrs.viewTransform * b->GetInterpolatedPosition();
-		attrs.camDist = attrs.viewCoords.Length();
+		// Large spheres should be drawn after objects past their horizons.
+		// This is just to make sure spikes are drawn properly.
+		if (b->IsType(Object::TERRAINBODY)) {
+			double rad = b->GetSBody()->GetRadius();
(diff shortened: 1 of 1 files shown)


impaktor:
This was never merged, and I know @robn has rewritten some of the code since then when doing the billboard-stuff, but I wonder if this is something that might still catch someones interest.

Seems to me like we still don't have this functionality in Pioneer.

corundscale:
We should have some generic way to display objects below screen resolution (just like you can see the stars at night despite them being well below your eyes' resolution) - not just system's bodies and surface ports, but also ships, etc.

fluffyfreak:
I did start on this about 8 months ago if someone wants to pick it up : https://github.com/fluffyfreak/pioneer/compare/view_distance

fluffyfreak:
You'd want to rip out the spiky code, tidy and test but otherwise it was going ok.

PR #1365, opened 2012-07-16
New module Bargain

GizmoR13:
New module similar to - 'commodity' needed -  from frontier.
Sorry for the english translation, if anyone can look at it, it would be nice.


impaktor:
For posterity: this is in master, since #5059, as the SoldOut module.

PR #1808, opened 2012-12-14
Hyperdrive malfunction repair ( very much WIP)

Chronothan:
This is supposed to allow the player to repair their hyperdrive after it has been destroyed by a malfunction. Right now it is only replaced with Class 1 Drive. It works on my version of Pioneer. I want to activate a timer to make the player wait after they land for a couple hours to simulate time spent working on it. How to do that?

This is my first functioning function. So it is most likely clumsily written and in need of a lot of optimization. Any pointers would be extremely helpful.

### DIFF
--- a/data/modules/BreakdownServicing/BreakdownServicing.lua
+++ b/data/modules/BreakdownServicing/BreakdownServicing.lua
@@ -83,15 +83,15 @@ GetDriveCondition = function ()
 		end
 	end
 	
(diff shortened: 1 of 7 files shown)


laarmen:
Having a short look at the diff, this would at least need to be rebased and updated to use modern API.

fluffyfreak:
I am closing this, it might be nice to have but has anyone ever had a hyperdrive fail on them?
Also I'm not sure that @Chronothan is still interested in working on Pioneer since his repo appears to be unknown know.
Closing.

PR #2054, opened 2013-02-11
Add 'Set Hyperspace Target' button on mission screens.

Chronothan:
As the title says, I added a button on the mission screens that will set the players hyperspace target to the mission location StarSystem. I attempted to add it in InfoView.lua, but couldn't make room on the screen for it. Although, I am running at 1024x768 resolution so I'm sure it's not as bad on higher resolutions.

I'm not very happy with how it looks, however. It seems kind of out of place. 
image

Ideas on how to improve this are very welcome. 

Included in this PR is a commit adding a font argument to the SmallLabeledButton, I found it useful for making the text smaller; as far as I know, you can't call :SetFont() on it. But I'm not sure if it needs to be in here.

### DIFF
--- a/data/libs/ui/SmallLabeledButton.lua
+++ b/data/libs/ui/SmallLabeledButton.lua
@@ -2,11 +2,12 @@ local ui = Engine.ui
 
 UI.SmallLabeledButton = {
 
-New = function (text)
+New = function (text,font)
 	local self = {
 		button = ui:SmallButton(),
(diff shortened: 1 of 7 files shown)


fluffyfreak:
This would be nice to have but if this PR is dead then can we just open it as an issue and close this?

impaktor:
#3767 closes this.

PR #2107, opened 2013-02-23
Adding buoy model with textures

jmf:
I have made a navigation buoy. The Idea was inspired by the buoys in oolite, the model is completely new.
Screenshot:
- Buoy in the red blinking phase
  
- Buoy in the green blinking phase


ecraven:
Could we use this in front of station entrances, to help with aligning correctly?

fluffyfreak:
That might be a good idea, or to identify queuing locations outside stations.

clausimu:
We'll have to make sure the position of the queue (or buoy) is outside of the current launch path. That is the biggest problem right now with increasing traffic - ships landing and starting are crashing into each other (as well as ships lining up to land crash into other ships waiting to land).

I think nozmajner also made a nice buoy model a few months back when I wanted to include in the SAR missions (got voted down though - with good reason).

nozmajner:
Shouldn't these kind of designators be UI only? I mean it feels strange to put a bouy anywhere, could easily become a navigational hazard. Like if balloons would designate parts of an airport.

nozmajner:
Don't know if this is needed right now. Will repoen, if/when so.

PR #2789, opened 2014-02-25
Flight following mouse movement

Darkdesire67:
High,

this is a new flight control. You can enable/disable it with space key (fire is now with left mouse button). When in mouse following mode, you vessel will follow your mouse cursor.

I've got only problem with the description item in the config screen of the game, it appears as it is defined in langstrings.inc.h (FLIGHT_FOLLOW_MOUSE). Don't know how to make a good items.

### DIFF
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -135,6 +135,7 @@ bool Pi::hudTrailsDisplayed = false;
 Gui::Fixed *Pi::menu;
 bool Pi::DrawGUI = true;
 int Pi::FlightModel=0;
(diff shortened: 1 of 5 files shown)


impaktor:
Possibly, with the big changes in UI with imgui being merged to the left and to the right, this is the time to discuss this PR? @nozmajner @ecraven ?

(it's obviously stale branch, but is this functionality something we want to incorporate?)

nozmajner:
So it's basically a toggle instead of a button hold?
I think it would be nice to have, but I'd also keep the current way.

PR #2990, opened 2014-06-30
WIP for review: Adding ShipManufacturer constants

impaktor:
## What this does

Introduces ship manufacturer constant, which are then used by the ship definitions, and gives warning if entered incorrectly when loading the ship models. Also lets faction scripts set weight of a manufacturers presence in said faction. There's no handling for default behaviour here yet. That is to come.
## Why

Instead of doing one giant pull request, I'm splitting this into many small ones, and this is (hopefully) the first in a series, to bring in faction specific repair prices and availability of ships in ship market. (as suggested on [dev-forum:post](http://pioneerspacesim.net/forum/viewtopic.php?p=2803#p2803) by @nozmajner)
## Review

I'm looking to get my a$$ handed to me, because I know this can be improved, so please point me in the right direction from here. I put the constants in `Polit.cpp`, feeling very dirty as I did so. Please grunt if they should be moved. I also had to convert the manufacturer constant to lowercase when using them on the ship textures, which feels a bit "ugly". I guess this aught to be in Lua but I was more confident on how to do it in c++, since we already have existing code for that. 

I've also added a commit just to add some manufacturers to SolFed, and tests error message printout.
## next step

Will be to expose this to Lua. But that's another PR, if this isn't completely killed.


fluffyfreak:
So what's happening with this? Do I close it or will it be updated one fine day?

impaktor:
I actually started talking with @laarmen about it just one week ago. (I scream at him when I need help understanding lua-C++ interaction)

> The only reason you want to have them in core currently seems to be that you want to check that define_ship refers to an existing manufacturer. Is that right?

Not quite. The reason I put the manufacturer constants in core / C++ is that parameters added to a faction definition are read into C++ from where I can export them back into lua. I didn't know a better way to do this, umm, +1 year ago.  Still don't actually.

But ideally, I just want to have access on the lua side to parameters defined in the faction-lua files. I agree that it is a bit strange to take the detour through C++ core.

So once I know how to do that (suggestions welcome), I'll either rebase this whole thing, I guess most commits go out the window, and will be replaced by new ones, or do a new PR.

fluffyfreak:
Closing as it will need a new PR/epic-rebasing/rewrite(?).

Either way we can come at it clean another time.

PR #3249, opened 2014-11-19
Post Processing with Bloom filter (Paragon)

fluffyfreak:
[Paragon](https://github.com/MeteoricGames/pioneer) added a post-processing framework and being GPL3 it means we can have it too :)

So this is the back-port and cleanup of their code.

It includes the Bloom filter but I have added the option of enabling or disabling it to the options screen, it is **off** by default.

There were a number of things which I didn't like about their code, _mostly due to time pressures they face that I don't I must add_, so I have fixed them up for Pioneer.

It adds a new string, 3 new shaders in a subfolder, a new checkbox option, and the new framework itself.

Of note is that this will **break the anti-aliasing again**. Even if the actual Bloom filter is disabled the whole screen is still rendered to a texture and it is that step which disables anti-aliasing.

I have another branch where I am hoping to implement [smaa](https://github.com/fluffyfreak/pioneer/tree/smaa) in the coming weeks but have decided to split these into two separate commits and the work was becoming too large and convoluted for review.

Any questions feel free to ask.

Andy


HybridDog:
l dislike bloom, but l would like http://docs.gimp.org/en/plug-in-retinex.html this effect.

fluffyfreak:
Closing. This is old and should be done another way.

HybridDog:
>l dislike bloom

Actually l like it if it's used in fog, mist or dust.

PR #3272, opened 2014-12-10
Describe translation strings better

impaktor:
One of our translators (@dreadk2002) has asked to get better description on some of the strings we have. Possibly I'll add more to this commit as he/she works through all our strings.

(That "HAT" string is insane for instance)

Also, is `FRONT` and `REAR` not used anymore by core?

```
git grep "Lang::FRONT" $(git rev-list --all) 
```

Suggests it was used in `ShipType.cpp` but all commit hashes are wrong. Or I am? Remove them? Unused strings aught to be removed.


dreadk2002:
@impaktor Sorry, for the holidays I was completely disconnected from 
everything. I will resume translations asap. Work first, of course.

PR #3274, opened 2014-12-20
New mission: CargoRun

robothauler:
This is a new mission type similar DeliveryPackage but with real cargo.


impaktor:
Sorry for forgetting this PR. I blame the fact that there is no notification when new commits are pushed to a repo. I'm looking at it now, and commenting as I go. 

The code is very clean, and a joy to read, so thanks for that.

data/lang/module-cargorun/en.json
```diff
@@ -0,0 +1,610 @@
...
+   },
+   "HOWMUCH_5" : {
+      "description" : "",
+      "message" : "There are only {cargo}t."
+   },
+   "HOWMUCH_6" : {
+      "description" : "",
+      "message" : "Only {cargo}t."
+   },
+   "HOWMUCH_WHOLESALER_1" : {
```
impaktor: Could you please add to the translation description of `HOWMUCH_5` and `HOWMUCH_6` that these strings are only used for 1-2 items/tonnes of goods.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1148 @@
...
+			form:SetMessage(l["ACCEPTED_PICKUP_" .. Engine.rand:Integer(1, getNumberOf("ACCEPTED_PICKUP"))])
+		else
+			form:SetMessage(l["ACCEPTED_" .. Engine.rand:Integer(1, getNumberOf("ACCEPTED"))])
+		end
+		return
+
+	elseif option == 4 then
+		local num = math.floor(ad.urgency * (getNumberOf("URGENCY") - 1)) + 1
+		local urgency = l["URGENCY__" .. ad.branch .. "__" .. ad.cargotype] or l["URGENCY__" .. ad.branch] or l["URGENCY_" .. num]
+		form:SetMessage(urgency .. Format.Date(ad.due))
```
impaktor: If possible, I think you should use the `{date}` construct so that other languages can put the time string anywhere in the string

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1148 @@
...
+			dom_system	= ad.domicile:GetStarSystem().name,
+			dom_sectorx	= ad.domicile.sectorX,
+			dom_sectory	= ad.domicile.sectorY,
+			dom_sectorz	= ad.domicile.sectorZ,
+			dist	  = string.format("%.2f", ad.dist),
+		})
+		form:SetMessage(introtext)
+
+	elseif option == 1 then
+		local whysomuchtext = l["WHYSOMUCHTEXT__" .. ad.branch .. "__" .. ad.cargotype] or l["WHYSOMUCHTEXT__" .. ad.branch]
```
impaktor: If I'm to nitpick the slightly ambiguous `whysomuchtext / WHYSOMUCHTEXT_` variable and string names are not consistent with other string variable names like `howmuch`, I'd opt for just `whysomuch`.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1148 @@
...
+		if #nearbysystems == 0 then return nil end
+		nearbysystem = nearbysystems[Engine.rand:Integer(1,#nearbysystems)]
+		dist = nearbysystem:DistanceTo(Game.system)
+		nearbystations = nearbysystem:GetStationPaths()
+		location = nearbystations[Engine.rand:Integer(1,#nearbystations)]
+		if freeCargoSpace ~= nil then
+			local approxFuelForJump = math.floor(dist / 10)
+			cargo = (freeCargoSpace - approxFuelForJump) <= 1 and 1 or Engine.rand:Integer(1, (freeCargoSpace - approxFuelForJump))
+			introtext = Engine.rand:Integer(1, getNumberOf("INTROTEXT"))
+			wholesaler = false
```
impaktor: Nitpicking: `pickup=false` is set explicitly at several places (which is nice, makes things clear) but not for this condition, which relies on its default behaviour for an unassigned variable.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1148 @@
...
+		isEnabled   = isEnabled })
+	ads[ref] = ad
+
+	return ad
+end
+
+local onCreateBB = function (station)
+	local num = Engine.rand:Integer(0, math.ceil(Game.system.population))
+	local numAchievableJobs = 0
+	local reputation = Character.persistent.player.reputation
```
impaktor: Looks like `reputation` isn't needed in this function?

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1148 @@
...
+local onShipDocked = function (player, station)
+	if not player:IsPlayer() then return end
+
+	for ref,mission in pairs(missions) do
+		if (mission.location == station.path and not mission.pickup) or (mission.domicile == station.path and mission.pickup and mission.cargo_picked_up) then
+			local reward
+			local oldReputation = Character.persistent.player.reputation
+			local cargo = unloadCargo(mission)
+
+			if mission.localdelivery then reputation = 1 else reputation = 1.5 end
```
impaktor: Does this work? Feels like there is a `local reputation` missing.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1148 @@
...
+			if wholesaler then
+				cargo = Engine.rand:Integer(max_cargo, max_cargo_wholesaler)
+				introtext = Engine.rand:Integer(1, getNumberOf("INTROTEXT_WHOLESALER"))
+				pickup = false
+			else
+				cargo = Engine.rand:Integer(1, max_cargo)
+				pickup = Engine.rand:Number(0, 1) > 0.75 and true or false
+			end
+		end
+		risk = custom_cargo[branch][cargotype].price / 400 + Engine.rand:Number(0, 0.25) -- goods with price 300 have a risk of 0.75 to 1
```
impaktor: Perhaps store the `400` as a variable: `max_price` defined in the top of the file, so when someone later adds more custom cargo to this module, like: "the galactic emperors dirty undies" they're more likely to realize they need to modify `max_price = 1000`.

Also, currently, the most expensive cargo looks to be Diamonds, which are 300 and not 400, so I'd do something like `risk = constant * price / max_price + rand`

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1148 @@
...
+	local cargotype = custom_cargo[mission.branch][mission.cargotype]
+	local uc = Game.player:RemoveEquip(cargotype, mission.cargo, "cargo")
+	if uc ~= mission.cargo then -- this is ugly
+		local list = Game.player:GetEquip("cargo")
+		for _, i in pairs(list) do
+			if i.l10n_key == cargotype.l10n_key then
+				Game.player:RemoveEquip(i, 1, "cargo")
+				uc = uc + 1
+				if uc == mission.cargo then break end
+			end
```
impaktor: Why do you need to do this? Does it have something to do with the case of the player jettisoning the cargo? Or is `Game.player:RemoveEquip` not behaving the way it should?

robothauler: If the player loads a saved game, the Game.player:RemoveEquip() will fail. At least at that time I tested it. And, after loading the saved game, the player could accept a mission with the same cargo type. In that case, the Game.player:RemoveEquip() couldn't unload enough cargo.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1148 @@
...
+	local localdelivery = Engine.rand:Number(0, 1) > 0.5 and true or false
+	local branch = Engine.rand:Integer(1, #custom_cargo)
+	local cargotype = Engine.rand:Integer(1, #custom_cargo[branch])
+
+	if localdelivery then
+		nearbysystem = Game.system
+		nearbystations = findNearbyStations(station, 1000)
+		if #nearbystations == 0 then return nil end
+		if freeCargoSpace ~= nil then
+			cargo = freeCargoSpace == 1 and 1 or Engine.rand:Integer(1, freeCargoSpace)
```
impaktor: You don't need to tree the specail case of `freeCargoSpace==1`, since `Engine.rand:Integer(1, 1)` will return 1.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1155 @@
...
+local onShipDocked = function (player, station)
+	if not player:IsPlayer() then return end
+
+	for ref,mission in pairs(missions) do
+		if (mission.location == station.path and not mission.pickup) or (mission.domicile == station.path and mission.pickup and mission.cargo_picked_up) then
+			local reputation
+			local oldReputation = Character.persistent.player.reputation
+			local cargo = unloadCargo(mission)
+
+			if mission.localdelivery then reputation = 1 else reputation = 1.5 end
```
impaktor: Some more nitpicking, but I'm pointing out everything I see: L790 and L794 can be combined.

`local reputation = mission.localdelivery and 1 or 1.5`

impaktor:
First, this is very nice, and it was very interesting to read through. Thanks for a great job! I do have a bunch of comments and thoughts, please don't consider them written on stone tablets or anything, just me thinking out loud. This post became a bit lengthy, so you might want to work through this a few times, I know I have.


## Nomenclature

For a general audience reading this, I should define what is meant by:

* __flavour:__ Same as usual in pioneer, but to anyone unfamiliar, we refer to there being many _flavours_ of strings with variation for the same thing being said. E.g. `N` different ways to say _"take this package to system X"_.

* __branch:__ Basically what I suggested in my lengthy post above. A group of missions (read: cargo types) that fit with a group of _flavours_. So if Plutonium belongs to both the "Radioactive" branch and the "Expensive" branch, all flavours of type "gosh this is dangerous" and "oh, my how expensive" can be used. Where as e.g. diamonds are only using the different flavours in the "Expensive" branch.


## Potential pitfalls

The code tries to prevent player from accepting missions where the ship lacks capacity, like picking up more cargo than he can carry. Danger with this is that there will always be special cases you can not check for, without a lot of extra code. I'd just opt for a basic check that the ship has large enough cargo hold, and skip estimates of fuel required.

Here are some potential pitfalls, I'm OK with all of them, just wanted to point them out.

1. When the player lost the cargo, and the value of it is deducted from the player, what happens if the player has negative money? Player stranded for all time on the planet? This is a general question that we (pioneer community) might want to think about how to handle. Perhaps buying hydrogen should always be allowed even when negative balance?

2. Nothing prevents the player from accepting two separate missions of type "go to `X` and pick up `N`t of cargo", where location `X` is the same for both but player only has space for (`N`t) one of them. Actually, this is fine, player can go two times to `X` and do each mission after another. Never mind.

3. Player accepts mission for N tonnes to be picked up at X, _before_ he refuels. There is some check for how much fuel is needed but very approximate.

I certainly don't consider 2 or 3 to be anything we need "fixing", but number 1 might need some thinking, but it doesn't affect this module, it's more this module triggers us having to consider that case.


## In makeAdvert, store the full keys, rather than just the index?

Perhaps I'm missing something, but couldn't your `introtext` variable store the _entire_ translation key (when set in `makeAdvert`), instead of just the index number? That would remove all the repeated switch-like if-else statements in `onClick` and `onChat` I think. E.g. instead of `introtext=2` have it (in current code) `introtext=INTROTEXT__7__2`, or perhaps I haven't through this through?


## Branched mission flavours

Example of what it is (to anyone interested): Neptunium is both "expensive" and "hardware", thus all flavours of type _"move expensive cargo to X"_ as well as all of type _"shipment needed at construction site at X"_ are possible. I see two things with the implementation I'd like to comment on.


### 1. Index on name instead of number

First of all, I suspect indexing the different branches on their actual names instead of numbers might be more readable and be more stable against bugs. E.g. changing from

```lua
local custom_cargo = {}
table.insert(custom_cargo, chemical)            -- 1
table.insert(custom_cargo, mining)                      -- 2
table.insert(custom_cargo, hardware)            -- 3
table.insert(custom_cargo, infrastructure)      -- 4
table.insert(custom_cargo, consumer_goods)      -- 5
table.insert(custom_cargo, expensive)           -- 6
table.insert(custom_cargo, crazy)                       -- 7
...

local branch = Engine.rand:Integer(1, #custom_cargo)
```

to (note: code untested)

```lua
local custom_cargo = {
         "CHEMICAL" = chemical,
         "MINING" = mining,
         "HARDWARE" = hardware,
         "INFRASTRUCTURE" = infrastructure,
         "CONSUMER_GOODS" = consumer_goods,
         "EXPENSIVE" = expensive,
         "CRAZY" = crazy}
...
--- in `makeAdvert`:
local RandomKey = function (T)
        local keyset = {}
        for k,v in pairs(T) do table.insert(keyset, k) end
        return Engine.rand:Integer(1, #keyset)
end

local branch = RandomKey(custom_cargo)
```

plus the keys for the translation strings, and I think the rest should work unchanged.


### 2. Index numbers aught to not be tied into a specific cargo in a branch

If I've sold you on that idea, it means you now only have one layer of numbers, e.g. `URGENCY__3__5` is now `URGENCY_HARDWARE_5`. I would now expect that all `URGENCY_HARDWARE_x` (or `URGENCY__3__x` in current code) would work with all strings in the hardware branch, but this is not so.

I think the (last) indexing number in a key should not matter which exact cargo within a branch flavour it is, as that would defeat the purpose of the implementation. Once a branch is decided upon, I think all flavours belonging to that branch should work with all cargo for that branch. It's mostly like that now, but not to a 100%. For example `URGENCY__3__5`, where `3` means "hardware branch", and `5` means neptunium, thus the "5" doesn't mean it is the fifth random acceptable flavour of any cargo in the "hardware" branch; only works with neptunium:

```json
"URGENCY__3__5" : {
   "description" : "",
   "message" : "The cargo consists of neptunium-235. So the delivery must be completed until {date}."
},
```

There are some strings like this that are very specific, but only a few. So for the above example of neptunium I'd suggest removing it or making it more general. For other specific ones like the Qibble mission I put you up to, that is best had in a separate "quibble"-branch, I think.

So, right now you have
```lua
l["RISK__" .. ad.branch .. "__" .. ad.cargotype] or l["RISK__" .. ad.branch] or l["RISK_" .. num]
```

And I'm thinking, that unless `ad.cargotype` is a string (e.g. "Neptunium") which would remove the ambiguity in some numbers being random flavour index and others being specific to a certain cargo within a branch, then this would be nicer:

```lua
l["RISK__" .. ad.branch .. "__" .. a_random_number] or l["RISK_" .. num]
```



## Probability theory (yay!)

In `makeAdvert` you first select `branch` and then `cargotype`, but since the branches hold different number of possible cargo types the probability for them will be skewed. Thus the two custom cargo in the "crazy" branch have a much higher probability than the many in "hardware" branch. Possible ways to fix this is to have some probability weight to each branch (i.e. the length of the table). (If you do what I suggested with changing `custom_cargo={}` from array to a mapping dictionary you need to grab the `tableLength` function defined in the top of NewsEventCommodity.lua.)

This issue is actually the only thing I actually consider to be an "issue" that needs fixing. I could try to hack something up for it next weekend, if you're busy.


## Minor random comments

* I haven't tried the mission, but `onShipFiring` doesn't do much and could be removed, (It makes the escort ship say "Guns, guns, guns." every other minute when firing). No doubt useful for debugging and testing though.

* I prefer to name my tables in plural, like: `pirates={}`, and `escorts={}`.

* Also, I like to have my Boolean variables be questions: `ispickup`, `is_cargo_picked_up`, `iswholesaler`, which of course isn't consistent with the other modules, which all have the variable name `localdelivery`. (I guess the "is" is implied, I'm just a sucker for [Scheme](https://en.wikipedia.org/wiki/Scheme_(programming_language)) where all Boolean variables end with a "?", e.g. `integer?`.) Feel free to ignore.

* I don't know how much having the "HABER" flavour gives you. It simply picks another ship for the escort, and has another greeting, while the "cost" for this is a bunch of extra code and if-statements, that looks a bit "patchy". I don't think players will notice this much.

* For the string: `YOU_DO_NOT_HAVE_ENOUGH_CARGO_SPACE_ON_YOUR_SHIP`, perhaps "_come back_ when you have enough space to hold the cargo" is better, indicating the player needs to _land again_, with lighter load to trigger the check again.

* `approxFuelForJump = dist / 10`. It will depend on mass and ship size as well. I think stopping the player from taking a mission he might not have the capabilities to do (`cargo` + `fuel needed` `>` `free space`) is bound to fail at least sometimes, so question is how much effort to spend here? I'd be fine with not checking fuel at all, I think, since taking all potential pitfalls into account will just complicate things and lead to lazy players.

data/modules/CargoRun/CargoRun.lua
impaktor: @laarmen thoughts on this?

robothauler:
Thank you for your commitment. Here are a few quick easy answers.

onShipFiring
This is related to the Fox code. http://en.wikipedia.org/wiki/Fox_(code_word)
If I remember correctly, the game Falcon 3.0 had this too. This is quite fun and should not be removed!

HABER
I was inspired because of the Haber story ;-) The intention was to expand this for other fractions at some time to have different escort ships. But I have no plans to do so. So, if the vote is to remove this flavour, I will not cry too much.

approxFuelForJump
I agree, I should remove this. The mission should be balanced only with the max_cargo and max_cargo_wholesaler constants.

impaktor:
> onShipFiring
> This is related to the Fox code. http://en.wikipedia.org/wiki/Fox_(code_word)
> If I remember correctly, the game Falcon 3.0 had this too. This is quite fun and should not be removed!

OK. I don't have any further opinion since I haven't tested it. I'd like to point out that the "GUNS_GUNS_GUNS" key needs a description for the translators e.g. "said by escort in the heat of battle when firing at enemy" or something similar, and there are probably other strings that need description. When they are ambiguous, or not clear in what context they will be used.

> HABER
> I was inspired because of the Haber story ;-) The intention was to expand this for other fractions at some time to have different escort ships. But I have no plans to do so. So, if the vote is to remove this flavour, I will not cry too much.

I was thinking along the same lines, that one can have more faction specific stuff, but I think that can be done in the future. There's un-merged code that gives each faction unique police ships, so that can be used then.

I have not yet actually tried any mission, perhaps others are interested in testing it (@walterar, @jmf), and I haven't looked at balance of deadline, distance, and pay but we should write some general library for this in the future anyway. The code can always be tweaked and fixed after merge if something is strange or unbalanced, but the language strings should preferably be done, since any change there pokes all our translators.

So, anyone wanting to test this, might want to set `risk=1` in `makeAdvert` around L495 to make life more interesting, or just play many cargo run missions.

walterar:
I'm doing a distressing task: Change OS version. In a few hours (or days?), Surely this PR will be my first test. I like the idea. I think it will be welcomed by the players.

impaktor:
This code should, for instance, make hardware branch 11 times more likely to be randomly selected than the wedding branch (I hope), but I haven't tested it and I'm notorious for writing buggy code, but it should be approximately what I/you want to do. If it runs. 

(One could make the accumulator into an array, to not have to do the sum every time in `randomBranch` but then I didn't, because this is less code I think.)

```lua
-- Each branch should have a probability weight proportional to its size
local weights = {}
local weight_sum = 0
for bracnh_key,branch_array in pairs(custom_cargo) do
	weights[key] = #branch_array
	weight_sum = weight_sum + #branch_array
end

local randomBranch = function()
	local accumulator = 0
	local r = Engine.rand:Integer(0,weight_sum)

	for k,w in pairs(weights) do
		accumulator = w + accumulator
		if r < accumulator then
			return k
		end
	end
	error("Oh, dear! This should not happen.")
end
```

laarmen:
Well, first off there is a typo line 4, an unknown symbol "key" line 5.
Also, I'd chop your head off if I stumbled across this error message when
using your software.  Just sayin'.

impaktor:
> Well, first off there is a typo line 4, an unknown symbol "key" line 5.

Yeah, I never ran the code, it was more a proof of concept / suggestion on my part. @robothauler implemented some other version of it without my typo.

> Also, I'd chop your head off if I stumbled across this error message when using your software. 

I suppose you want it to say something like: "[Guru Meditation](https://en.wikipedia.org/wiki/Guru_Meditation)" instead? :wink: 

Actually, @robothauler, you might want to remove (or replace) the error message in your (/my) code, because I hope my head will remain firmly on top of my body for a while longer.

On another note: my main problem now is that the module is not actually working for me. It always crashes when attempting to reach one index too much in  `getNumberOf` when creating the adverts. Does this not happen for you? Thus any attempt at creating an advert will cause a crash for me.

robothauler:
>Actually, @robothauler, you might want to remove (or replace) the error message in your (/my) code, because I hope my head will remain firmly on top of my body for a while longer.

:-D To not lose my head too. Will a error message like "Random number out of range" do?

>On another note: my main problem now is that the module is not actually working for me. It always crashes when attempting to reach one index too much in getNumberOf when creating the adverts. Does this not happen for you? Thus any attempt at creating an advert will cause a crash for me. 

I never stumbled over this issue. As long as you have TEXT_1 to TEXT_x this should work.

impaktor:
> I never stumbled over this issue. As long as you have TEXT_1 to TEXT_x this should work.

Never mind, my binary build was old, from back when not finding a language string caused a crash, instead of a `nil`.

Thinking out loud: Do we want the player to always have to explicitly ask how much cargo the contract is on? I guess if it was said in the main text of the contract you would loose the flavour of them saying _"just X tonnes"_ when it's under 3t, which is a downside if it was removed. I have no direct opinion on this, perhaps others (@jmf, @fluffyfreak, @nozmajner, @laarmen) have thoughts?

fluffyfreak:
In this case [I have no strong feelings one way or another](https://www.youtube.com/watch?v=ussCHoQttyQ) :)

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+	l10n_resource="module-cargorun"
+})
+wedding_dresses = Equipment.EquipType.New({
+	l10n_key = 'WEDDING_DRESSES', slots="cargo", price=15,
+	capabilities={mass=1},
+	purchasable=false, icon_name="Default",
+	l10n_resource="module-cargorun"
+})
+
+local chemical = {}
```
laarmen: Why do you declare the equipments above as globals (baaaaaaaaad!), just to stash them in a local table afterwards ?

You could directly declare the table with the equipment inline, thus sparing the ~50 lines of table.insert

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+table.insert(radioactive, plutonium)
+table.insert(radioactive, uranium)
+
+local centrifuges = {}
+table.insert(centrifuges, aluminium_tubes)
+table.insert(centrifuges, machine_tools)
+table.insert(centrifuges, titanium)
+
+local custom_cargo = {
+	{ bkey = "CHEMICAL", goods = chemical, weight },
```
laarmen: Unless I am missing something, weight doesn't appear to be declared anywhere.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+local ads = {}
+local missions = {}
+
+local isQualifiedFor = function(reputation, ad)
+	return
+		reputation >= 12 or
+		ad.localdelivery or
+		(ad.risk <  0.1 and ad.urgency <= 0.1) or
+		(ad.risk <  0.5 and ad.urgency <= 0.5 and reputation >= 4) or
+		false
```
laarmen:
```
return something or false
```
is semantically equivalent to 
```
return something
```

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+end
+
+local getNumberOf = function (str)
+	local num = 0
+
+	while l[str .. "_" .. num + 1] do
+		num = num + 1
+	end
+	return num
```
laarmen: There is a number of things that are wrong with this function.

First, instead of starting at 0 and always use the +1 value, you could start at 1, and just decrement at the end.

Second: What's that business with the translation table ? I am supposed to understand what is the purpose of the function when reading it, and yet I have no idea what is the intended use so far. My educated guess would be that you are trying to figure out how many flavours there are of some string. If that is the case, well... I find that whole business terribly hackish, but I don't think we have any other way to figure it out with the current state of affairs.

However, I'd like us to introduce an API within our translation module to do just that, as it seems to me that this thing might be a recurrent problem. That shouldn't be very hard to do, but that is not the scope of this PR.

In the mean time, would you be so kind as to document this function and/or rename it to a more explicit moniker?

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+	if not qualified then
+		form:SetMessage(l["DENY_" .. Engine.rand:Integer(1, getNumberOf("DENY"))])
+		return
+	end
+
+	if option == 0 then
+		local introtext = string.interp(ad.introtext, {
+			name		= ad.client.name,
+			cash		= Format.Money(ad.reward),
+			cargoname	= l[ad.cargotype.l10n_key],
```
laarmen: Please use add.cargotype.GetName() here.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+			dom_sectory	= ad.domicile.sectorY,
+			dom_sectorz	= ad.domicile.sectorZ,
+			dist		= string.format("%.2f", ad.dist),
+		})
+		form:SetMessage(introtext)
+
+	elseif option == 1 then
+		local n = getNumberOf("WHYSOMUCH_" .. ad.branch)
+		if n >= 1 then
+			form:SetMessage(string.interp(l["WHYSOMUCH_" .. ad.branch .. "_" .. Engine.rand:Integer(1, n)], { cargoname = l[ad.cargotype.l10n_key] }))
```
laarmen: See previous remark.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+		elseif ad.urgency >= 0.8 then
+			form:SetMessage(string.interp(l["WHYSOMUCH_URGENT_" .. Engine.rand:Integer( 1, getNumberOf("WHYSOMUCH_URGENT"))], { cargoname = l[ad.cargotype.l10n_key] }))
```
laarmen: And again. For the next ones, I'll just assume you'll have them all fixed ;-)

(The rationale for that is that I consider the l10n_key to name relationship part of the equipment system internals)

data/modules/CargoRun/CargoRun.lua
impaktor: A bit further down:

```lua
-- Each branch should have a probability weight proportional to its size
local custom_cargo_weight_sum = 0
for branch,branch_array in pairs(custom_cargo) do
    custom_cargo[branch].weight = #branch_array.goods
    custom_cargo_weight_sum = custom_cargo_weight_sum + #branch_array.goods
end
```

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+				and def.hyperdriveClass > 0 and def.hullMass <= 400 end, pairs(ShipDef)))
+			if #shipdefs == 0 then return end
+
+			local pirate
+
+			while pirates > 0 do
+				pirates = pirates - 1
+
+				if Engine.rand:Number(1) <= risk then
+					local shipdef = shipdefs[Engine.rand:Integer(1,#shipdefs)]
```
laarmen: Mpfff... We really should introduce a standard way to spawn random ships. Being the one that broke the hyperdriveClass API and had to go and fix all these snippets, I can tell you there is a *lot* of cargo-cult coding going on :-)

data/modules/CargoRun/CargoRun.lua
impaktor: I found the function clear when seeing how it is used. Possibly `getNumberOfFlavours`? I also had the thought that this is useful enough to be used in the other modules.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+		escort_ships = {}
+	end
+end
+
+local unloadCargo = function (mission)
+	local uc = Game.player:RemoveEquip(mission.cargotype, mission.cargo, "cargo")
+	if uc ~= mission.cargo then -- this is ugly
+		local list = Game.player:GetEquip("cargo")
+		for _, i in pairs(list) do
+			if i.l10n_key == mission.cargotype.l10n_key then
```
laarmen: This is *wrong*. Really wrong.

Having the same l10n_key doesn't necessarily mean that we are dealing with the same piece of equipment. I can imagine a module introducing lasers as cargo equipment, which means they could very well have LASER_1MW as l10n_key, but refer to its own l10n module.

Last time I had a similar problem, it was because of a serializing/deserializing issue. Was it also for you ?

impaktor: I was almost going to suggest storing the actual cargo data structure instead of just the cargo name/string in the `cargotype` variable, but then I didn't. But doing that would fix this, no?

@robothauler Also, if we're nitpicking again: The variable `cargo` should probably be `mass` or `cargomass`, or something similar.

laarmen: I'd say "amount", since mass is intrinsics to the cargo type.

(so far it is 1t for everyone, but there's a reason I've set it explicitly).

@impaktor is right on the fact that you should store a reference to the actual table instead of its name. And/or (I'd advise the "and" here), you should store all the equipment tables you generated into the savegame, and restore them on load (see data/libs/Equipment.lua for an example), so that you are sure there are no duplicates of the same equip type.

(Yeah, equipment was my pet project for a while, I know it inside and out, hence the nitpicks :-) )

data/modules/CargoRun/CargoRun.lua
laarmen: Still. In this case, ", weight }" means "Add the value of the symbol weight to the array part of the table", where weight is nil, so basically it is a NOP. I'd suggest "weight = 1" or whatever the default value should be.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1138 @@
...
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local Engine = import("Engine")
+local Lang = import("Lang")
+local Game = import("Game")
+local Space = import("Space")
+local Comms = import("Comms")
+local Event = import("Event")
+local Mission = import("Mission")
+local NameGen = import("NameGen")
```
impaktor: `NameGen` isn't used, so you can just remove it. (An inheritance from the copy-paste from DeliverPackage.)

impaktor:
I feel better now that you got rid of the `unloadCargo` which I never understood, since it should not have been needed.

I should have spoken sooner, but I don't think I agree with @laarmen's comment on not using `table.insert`, because (if I'm not mistaken) the code relies on the tables to be arrays and not dictionaries/maps. However, if I am mistaken then the code should work as it is now, if not, all tables will have 0 length, and `custom_cargo_weight_sum == 0`, since any call to `#custom_cargo[*].goods` will be `0`. At least that is what happens when I run the code (at a quick glance) in my mind.

OK, so I should read through `en.json`, also but I kind of hope I don't do it today, because I really should be working on other stuff. I hope I can merge this by next weekend. This is a very big contribution, but thanks for being patient, and staying with us.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -0,0 +1,1136 @@
...
+	elseif option == 2 then
+		local howmuch
+		if ad.wholesaler then
+			howmuch = string.interp(l["HOWMUCH_WHOLESALER_" .. Engine.rand:Integer(1, getNumberOfFlavours("HOWMUCH_WHOLESALER"))], {
+				amount    = ad.amount,
+				cargoname = ad.cargotype:GetName(),
+			})
+		else
+			local num = ad.amount <= 3 and getNumberOfFlavours("HOWMUCH") or (getNumberOfFlavours("HOWMUCH") - 2) -- The last 2 strings contain "only"
+			howmuch = string.interp(l["HOWMUCH_" .. Engine.rand:Integer(1, num)], {
```
impaktor: I think the comparison should be the other way around, `ad.amount >= 3`. I've rewritten it now, so don't worry about it, I've made some changes I merge into your commits, and then I merge this into master shortly, so I've got this from now.

Also, I've changed this to be `HOWMUCH_SINGULAR_` and `HOWMUCH_` since I think the distinction between is/are is more importnat than adding an "only" if it's below 3t. So now if it's just 1t it says "is only..." or uses the plural strings.

impaktor:
__Merged!__ Rebasing this was painful, but I squashed it down to 2-3 commits. @robothauler is author of CargoRun.lua and en.json and @nozmajner for the two .png files.

@robothauler would you like your real name to be used in the AUTHORS.txt file? Let me know here or on irc, and I'll add you.

Big thanks for contributing, and sticking by us through all our comments.

PR #3295, opened 2014-12-30
Cleanup equip strings

impaktor:
This is the second part of #3294, so it must be merged first.

### First commit

If the flight mode is called "Manual Control", I guess it should be "Autopilot Control" when in Autopilot, either way: don't reuse the equipment string, for indicating the flight mode.

### Second commit

I think pioneer cheated by reusing equipment strings for showing equipment mounts. Now we can have "Atmospheric __s__ hielding yes/no" and "Atmospheric __S__ hielding" for the equipment, which is consistent with how we write equipment, and their respective slots.

### Third commit

This might be the controversial one, and I'm only arguing for it if, as I remember @robn (ping!) saying, that transifex allows moving of strings, so no re-translation should be needed. I hope my memory of him saying that is right.

So, my arguments for moving equipment strings from core to its own module are:

1. `core` is for C++ / `src/` strings.

2. It is easier for translators when each module/file to translate is used in a specific context. Like equipment market, and mission `lang/module-`s.

3. Conversely, translating `core` strings is very difficult since the strings are just piled on top of each other and often you have no idea where the strings are used. Example: `HAT` (is not worn on the head), `JOY` (is not an emotion), `X`, (is not an in game coordinate nor location, but a `JOY`stick axis).

4. `core` is a behemoth and overwhelmingly long, possibly being intimidating for a translator to start working on.

__Question:__ I don't know if the name of the new module should be `equipment-core`? Once the rest of the UI is in lua, also the commodities aught to be moved out from core, but into the same `equipment-core` module, or into another one?


fluffyfreak:
Yeah this seems like a good idea, as an English speaker you forget that people won't automatically know the abbreviations for things, your `HAT` one is a perfect example.

Also `core` is overwhelming, and a pain to locate things in for changing.

impaktor:
I don't dare merge this until @robn gives thumbs up that transifex is smart enough to see when strings have moved from one module to a new one, since I'm moving quite a bunch of them in the last commit, and I don't want our translators to be forced to re-translate the same strings one more time, at least not if it can be avoided in some way.

_Also a reminder to me/rob/jpab: a new module needs to be added on the transifex side._

robn:
The last time I did it the Transifex people told me they should carry over, but there was still a problem. Unfortunately I don't remember what the problem was!

If they don't carry over we can reorganise the generated files and push those to Transifex, which will get the translations back into their correct places. That's all a bit fiddly though. Would you like me to take care of this merge? I'm willing to put aside time this evening to look at it.

impaktor:
> Would you like me to take care of this merge? I'm willing to put aside time this evening to look at it.

That would be fine by me, and I assume @fluffyfreak is fine with it as well. Thanks!

robn:
Ok cool. Won't be tonight (got it too late), and I'm out a lot over the next few days but I should get a couple of hours somewhere to look at it.

fluffyfreak:
Is this one still waiting on anything? Transifex?

robn:
Still on me to make sure Transifex doesn't eat things. *checks schedule* I think I'll have some time on Friday night to look at it.

fluffyfreak:
OK cool :)

robn:
Ok working on this now. Hopefully I can get it in the next hour.

PR #3099, opened 2014-07-30
Orbital station

bszlrd:
## Please test

(details bellow)
I'm putting this up early, so it's up for @Fluffyfreak to work on the new way of dock setup (and to motivate myself a bit).

My aim is to make it's textures passable as soon as I can, then incrementally add more detail and polish to the textures and the model eventually, but that's a separate issue.

Right now it looks like this:
screenshot-20140730-232645
screenshot-20140730-232703
screenshot-20140730-232805

PR #3297, opened 2014-12-31
Bundle Collision with SGM file

fluffyfreak:
# Description:
We have an optimised & preprocessed file format called SGM that was written by Luomu but, for some reason, we don't ship SGM files in with the main distribution. Part of that reason was that it helped loading times a lot but not as much as it could have due to the collision mesh generation being excluded from it.

With the merging of #3099 giving us 3 very complex new space stations and very complex collision meshes I've tried to add support for serialising and deserialising the collision meshes into our SGM format.

## Distribution:
With this change it would make sense to build the `.SGM` files in-place, delete the `.DAE` from the build and only have them in Git and `pioneer-assets`. 
This way players don't have to worry about having the correct files or building `.SGM` versions.

# Loading times:
A quick comparison, please note that this is done on my laptop with a 512GiB SSD so there's little variation due to disk access, but possibly some due to thermal throttling/behaviour.

## SGM Load times:
I did 3 runs and the times were
* Loading took: 23865.460385 milliseconds
* Loading took: 23486.056590 milliseconds
* Loading took: 23832.089322 milliseconds
For an average of **~23727.87** milliseconds.

## Standard Load times:
Once again but with no SGM files.
* Loading took: 128687.466982 milliseconds
* Loading took: 126967.270705 milliseconds
* Loading took: 126556.816863 milliseconds
For an average of **~127403.85** milliseconds

## Result:
This means that using SGM files with the collision mesh data results in the game loading to the intro screen **~5.37** times faster.

# At the Command line:
**WARNING**: The SGM files this creates will be chosen in preference to the original data in future so you might miss updates to files. Best to delete all of these SGM files after experimenting with this **WARNING**

To recreate the SGM files you just need to build the ModelCompiler, this should be shipped with the binaries versions, or you can build it yourself. It's built alongside Pioneer.
Simply go to the Pioneer folder where ModelCompiler & Pioneer reside, open a prompt / terminal and enter something like: `modelcompiler -b inplace`
That will create an SGM file for every model in the game and place it next to it's `.model` file.

# Status:
Worth testing, as I've just fixed the bug I had caused and collision now works again. Loads fast and works!

I've chosen to **compress** all of the SGM files the modelcompilier produces as this results in a reduction from ~132MB down to just ~39.5MB which is a significant reduction. It does have a negative impact on the loading times, raising it from an average 20 seconds to 24 seconds. However I think this is more than worth it.

Andy


fluffyfreak:
> In place = player's computer, or on the build machine? If the latter, then I assume something is needed for the scripts/nightly-build.pl?

On the build machine, although if we had an installer I guess we could do it locally. No, build machine definitely.

Build the SGM files using `./src/modelcompiler -batch inplace`, delete the `.dae` files, then package as normal.

robn:
Noted. I'll need to cook it into the configure script as well.

impaktor:
I noticed there's at least one more recent version of miniz since 1.14. From [minz-homepage](https://code.google.com/p/miniz/)

> - v1.15 r4 STABLE - Oct 13, 2013: Merged over a few very minor bug fixes that I fixed in the zip64 branch. This is downloadable from [here](http://code.google.com/p/miniz/downloads/list) and also in SVN head (as of 10/19/13).

> - v1.15 - Oct. 13, 2013: Interim bugfix release while I work on the next major release with zip64 and streaming compression/decompression support. Fixed the MZ_ZIP_FLAG_DO_NOT_SORT_CENTRAL_DIRECTORY bug (thanks kahmyong.moon@hp.com), which could cause the locate files func to not find files when this flag was specified. Also fixed a bug in mz_zip_reader_extract_to_mem_no_alloc() with user provided read buffers (thanks kymoon). I also merged lots of compiler fixes from various github repo branches and Google Code issue reports. I finally added cmake support (only tested under for Linux so far), compiled and tested with clang v3.3 and gcc 4.6 (under Linux), added defl_write_image_to_png_file_in_memory_ex() (supports Y flipping for OpenGL use, real-time compression), added a new PNG example (example6.c - Mandelbrot), and I added 64-bit file I/O support (stat64(), etc.) for glibc.

fluffyfreak:
Alright I would like to merge this today/tomorrow so that people can enjoy 24 seconds startup times, instead of 2 minutes, as soon as possible.

@robn lets do those changes to the build script separately once this PR is merged.
@impaktor you got any time to give this a quick test, once without the SGM files and once with?

impaktor:
I don't have access to a Pioneer capable computer today, but tomorrow I'm all yours, except when I'm not whoring for my boss that is.

fluffyfreak:
Tested with Linux/Windows with and without sgm files, it's no worse without and them and much better with them so have merged.

impaktor:
Without SGM:

    Loading took: 70323.744892 milliseconds

with:

    Loading took: 15531.321882 milliseconds

fluffyfreak:
That's fairly speedy computer you have there! Good improvement all the same.

robn:
I'm just looking at hooking this into the build system. Question: do the compiled files work cross-platform? Or do I have to compile them for windows with the windows binary?

fluffyfreak:
@robn it should work cross platform I believe.

robn:
Yeah, tested Linux files under WINE, all looked good. That's excellent because now I can do a single build and copy it into the build archives before upload.

robn:
20150201 includes SGM files. Phew.

impaktor:
@robn I've noticed that the builds on the pioneer-page are still on Jan 20 build in spite of the tags for 20150121 and 20150201 that you just mentioned, so I think there's something failing.

I see the 20150121 version on sourceforge, but on the pioneer page we're still on 20150120. And no sign of the 20150201 version yet.
http://sourceforge.net/p/pioneerspacesim/activity/

robn:
Weird. I thought maybe we'd hit a limit, so I've deleted everything before December (stuff right back to 2013, wow).

robn:
They're visible, but not downloadable, if I login. So yeah,. I guess something is stuck somewhere.

screenshot - 020215 - 10 11 24

I'll get onto support, let you know how I go.

robn:
Ahh: https://twitter.com/sfnet_ops/status/561228107630129153

I checked in with their support, and they reckon file uploads is part of it. I'll keep an eye on it. Hopefully it will all just come back to life.

robn:
It came to life :)

impaktor:
> so I've deleted everything before December

We sometimes* refer players to older builds, in particular to the last OpenGL 2.1 build:


http://sourceforge.net/projects/pioneerspacesim/files/pioneer-20141118-win32.7z/download
http://sourceforge.net/projects/pioneerspacesim/files/pioneer-20141118-linux64.tar.bz2/download
http://sourceforge.net/projects/pioneerspacesim/files/pioneer-20141118-linux32.tar.bz2/download
http://sourceforge.net/projects/pioneerspacesim/files/pioneer-20140919-osx.zip/download

_*In IRC, on the FAQ, and from Andy's blog_

robn:
Oh. Does anyone have backups then? Or I can go back in time and do some older builds.

(Personally my care factor is below zero).

PR #3304, opened 2015-01-06
Faster text buffer rendering

fluffyfreak:
# Description:
Create and reuse VertexBuffer objects wherever we can for text rendering.

The idea being that this should be faster than the existing code which creates, renders then destroys a great many `VertexBuffer` objects per-frame.

# WIP:
This actually seems slower at the moment, it is a work-in-progress.

Andy

### DIFF
--- a/src/GameLog.cpp
+++ b/src/GameLog.cpp
@@ -1,6 +1,7 @@
 #include "GameLog.h"
 #include "StringF.h"
 #include "graphics/Renderer.h"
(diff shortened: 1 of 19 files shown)


fluffyfreak:
@robn would it be possible to have a new build please? With this and the previous PR #3309 in it?

robn:
Ding! Build starting, and now :sleeping:

PR #3306, opened 2015-01-06
Dummy renderer

robn:
Adds a dummy renderer, which does nothing but returns true for pretty much every call. The motivation for this is to allow things that require a renderer (eg modelcompiler) to run in environments where OpenGL is not available.

This also moves remaining GL calls from core & graphics into the GL renderer itself. This is various error checks, screendump and glDepthRange. I do not claim that these are good-quality APIs and I'm happy for them to be changed or removed in the future. Right now they're just decoupling.

There's also a basic registration system that allows the program entry points (Pi, ModelViewer and modelcompiler) to prepare the renderers they want to make available. Its whole purpose it to make it so that graphics doesn't link directly to the individual renderer constructors. Sort of a static-build plugin system, egads.

Right now modelcompiler still has to link the GL renderer to satisfy indirect dependencies (scenegraph -> Color -> LuaRef -> Pi (for Pi::serializer) -> RendererOGL. That's a fairly simple chain to break, but its not really part of this PR and I'm almost certain I won't need it on the build machine anyway. I have OpenGL there, so link succeeds, I just can't run OpenGL, which I now won't have to.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -396,6 +396,8 @@ AC_CONFIG_FILES([
 	src/gameui/Makefile
 	src/gui/Makefile
 	src/graphics/Makefile
+	src/graphics/opengl/Makefile
+	src/graphics/dummy/Makefile
 	src/scenegraph/Makefile
 	src/galaxy/Makefile
 	src/terrain/Makefile
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -286,6 +286,7 @@ pioneer_LDADD = \
 	collider/libcollider.a \
 	gui/libgui.a \
 	graphics/libgraphics.a \
+	graphics/opengl/libgraphicsopengl.a \
 	galaxy/libgalaxy.a \
 	scenegraph/libscenegraph.a \
 	text/libtext.a \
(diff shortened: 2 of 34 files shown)


fluffyfreak:
This is a good move, it would be nicer to fix the issues caused by using glDepthRange etc but that's not this PR's scope as far as I'm concerned. I like that it finally just hides it away within the Renderer implementations.

I'll try and properly review it today and then hopefully merge it some time this afternoon (workload permitting). Then it would be good if we could have another release.

fluffyfreak:
Ok I've got some really minor fixes on [a branch](https://github.com/fluffyfreak/pioneer/tree/dummy-renderer) which is really just a single commit to fix the usual visual studio problems https://github.com/fluffyfreak/pioneer/commit/dd195f228a33061ed75be2f8fcc5e14e7b3a1757

Can merge this then it or it into this, I don't mind but otherwise I've tested it and it works fine.

robn:
Great. I'll merge it all in the morning and do a build, with compiled models bundled in. Wheee!

PR #3318, opened 2015-01-12
Adding some letters to fonts (for the Romanian translation)

### LINKED ISSUE #3314: Text has missing pixels and some special 2 bit characters dissapear
Text is looking kind of strange in build 20150109. Also some (only some) of the characters are dissapearing. The sistem I run Pioneer on is win7, it has GeForce GTX 760 2Gb video card. Also the characters "ș" and "ț" are only sometimes showing. In "Ieșire" is the same character as in "Pornește" (see 2nd and 3rd pictures).

loading
menu
strings

nozmajner:
Adding the letters ș,ț,Ș and Ț to the three fonts: Titillium, Orbiteer and Inconsolata which were missing it.
These are characters of the Romanian alphabet, but they were substituted to similar looking but different characters in the translation.
The OSL licence mostly seems to be OK with this, with the exception that we can't present the original name (not the filename) to the user. I think we don't do that anyway, soshould be in compliance with the license.

>"3) No Modified Version of the Font Software may use the Reserved Font
>Name(s) unless explicit written permission is granted by the corresponding
>Copyright Holder. This restriction only applies to the primary font name as
>presented to the users."

url: http://scripts.sil.org/cms/scripts/page.php?item_id=OFL_web

Should fix  #3314 partially


johnbartholomew:
Regarding naming, the original font is called Orbitron. When we added
characters previously we renamed our modified version to Orbiteer to comply
with the licence. I'm pretty sure we don't need to rename it again.

johnbartholomew:
Oh... but of course if we're modifying other fonts maybe we need to rename
those... I'll investigate a bit.

nozmajner:
The license says that it only matters if we present the font name to the player somehow. At least that's my understanding.

johnbartholomew:
@laarmen, do you have any opinion about whether we should rename Titillium and Inconsolata for our modified versions?

I'm not sure what the best choice is here... I'm not a lawyer, but my gut feeling is that renaming them is the legally "safe" option. Renaming them also matches what we decided when we modified Orbitron the first time. I'm not sure there's any significant disadvantage of renaming them except for the extra work it creates for us, which I hope would be small.

For Orbitron, when we first modified it we also contacted the original creator of the font to see if he wanted to incorporate our changes, which would have avoided the renaming problem, but unfortunately that never actually happened. We could attempt that for Inconsolata (good chance of success here... the guy that made Inconsolata is familiar with open source development), and Titillium (unfortunately I'm not even certain who exactly we should contact for this).

laarmen:
I'd play on the safe side, but still, we should upstream our changes
whenever possible. I'm guessing adding extra characters to a font as
widely used as Inconsolata will be useful to others :-)

nozmajner:
It seems that the newer Titillium already has the S characters, but not the T ones.

johnbartholomew:
I've sent an email to the guy that made Inconsolata.

johnbartholomew:
I haven't had any response from the Inconsolata creator. I suggest we rename our copies so we can get the required characters in-game, and then we can continue to try to push the changes upstream afterwards.

impaktor:
> I haven't had any response from the Inconsolata creator. I suggest we rename our copies

To: Incontinentia?

johnbartholomew:
Well, we could do the boring thing (assuming that these names are considered "different enough"):
Inconsolata -> PioneerInconsolata
Titillium -> PioneerTitillium

Or a silly thing:
Inconsolata -> Incontinentia / Inspaceolata
Titillium -> Titeerium (this name sucks, btw, but I can't think of any non-sucky alternatives)

nozmajner:
Tilillioneer  and Inkrumpliata? (krumpli = potato, my favourite word :) )
Do I need to change the name embedded into the font, or the filename change is enough?

johnbartholomew:
> Tilillioneer and Inkrumpliata? (krumpli = potato, my favourite word :) )
Sounds good to me!

> Do I need to change the name embedded into the font, or the filename change is enough?
Please change the names embedded in the font.

PR #3319, opened 2015-01-13
Faster deduplication of collision data

fluffyfreak:
What took the longer when generating the collision data was the deduplication of vertices from the collision mesh.

For example:
> CreateCollisionMesh for : (orbital_station_2-10k)
>  -- GeomTree::GeomTree took: 101175.531197 milliseconds
> Compiling "orbital_station_2-10k" took: 125375.918620

This is a simple perfect weld operation so I went looking for faster options and have used the `public domain` welding code from [https://code.google.com/p/nvidia-mesh-tools/](https://code.google.com/p/nvidia-mesh-tools/) with some modifications to only use `std::vector` etc.

The result is:
> CreateCollisionMesh for : (orbital_station_2-10k)
> ---   43370 vertices welded
>  -- GeomTree::GeomTree took: 1881.504192 milliseconds
> Compiling "orbital_station_2-10k" took: 24707.535703

The actual welding operation is therefore (101175.53 / 1881.50) = **53.77** times faster.

There's still overhead from other operations, compressing the sgm files for example but with the largest files like `orbital_station_2-10k` it does make it about 5 times faster overall.

Andy


laarmen:
Well, technically it's MIT licensed, not public domain (which is good, because explicit public domain licensing isn't valid in most continental Europe countries). I'll have a closer look at the algorithm later, to satisfy my curiosity more than anything else.

fluffyfreak:
RE: license, good point, I was just going by what's at the top of the `Weld.h` file which states `// This code is in the public domain (PD) -- castanyo@yahoo.es` as it was developed with a few others on a newsgroup :)

PD is the original Open Source as far as I'm concerned ;) back when everything was shared via bbs there were no licenses used for any of it with any onerous GPL clauses.

Anyway the output seems to be identical to the original algorithm, I wrote a bunch of checking code and then stupidly deleted it when cleaning up - it wasn't committed :(

It's quite simple, it hashes each vector3f (`but can be any data`), checks to see if it's already in there and enters it into a map if it's not, whatever index it gets - either a new one or an existing entry - gets pushed into an index buffer.
Afterwards you just remap the old indexes to the new ones.

Voila, all done.

Much faster because it's not having to compare every entry against every other entry.

The only potential issue is the triangle skip flag, but nothing in the code uses them anymore. In fact I was thinking of removing them entirely.

impaktor:
Compiled it.
Tried it.
Runs.
Nice.

PR #3320, opened 2015-01-14
Many ship optimise

fluffyfreak:
After a [forum discussion](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=300) about running the game with more traffic, i.e. More ships, I decided to look into the performance problems it causes.

The main areas were the ships constant checking to see what's nearby in `Ship::UpdateAlertState` and the `Animation::Interpolate` which was being run even though the animation time hadn't been changed.

I also changed the container types used for `geom` and `body` pointers in `CollisionSpace` and `Space` after recently [reading an article](http://baptiste-wicht.com/posts/2012/12/cpp-benchmark-vector-list-deque.html) about the conditions and situations where `std::list` performs better than `std::vector` or `std::deque`. Although we do some addition and removal from the containers it's relatively rare compared to our constant iteration and accessing of them which vastly dominates their usage. 
Therefore I've switched us from `std::list` to `std::vector` for them both.

Andy

### DIFF
--- a/src/Body.h
+++ b/src/Body.h
@@ -119,4 +119,7 @@ private:
 	double m_physRadius;
 };
 
+
+typedef std::vector<Body*> BodyList;
+
 #endif /* _BODY_H */
--- a/src/Missile.cpp
+++ b/src/Missile.cpp
@@ -64,7 +64,7 @@ void Missile::TimeStepUpdate(const float timeStep)
 	if (!m_owner) {
 		Explode();
 	} else if (m_armed) {
-		Space::BodyNearList nearby;
+		BodyList nearby;
 		Pi::game->GetSpace()->GetBodiesMaybeNear(this, MISSILE_DETECTION_RADIUS, nearby);
 		for (Space::BodyNearIterator i = nearby.begin(); i != nearby.end(); ++i) {
 			if (*i == this) continue;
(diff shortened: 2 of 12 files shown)


walterar:
The biggest problem I see, long ago, are the ships that remain "trapped" in docking. This produces a huge drop in performance. This problem, is not new, is produced in the last stage of docking with accelerated time.
However, this may help if it works. I will test now.

walterar:
Testing. Compiles and runs fine.
In game, seems to have no problems.

walterar:
There seems to be a problem with FiringAlert. Check it.

fluffyfreak:
What is the problem with FireAlert? I might not spot the problem unless I know what it is!

walterar:
http://youtu.be/FOcdcGUC_DI

walterar:
Also you can see some overlap of messages. Apparently this is not completely solved.

fluffyfreak:
Ah yes, I see what I've done there. Thanks @walterar

walterar:
No disables alert (red or yellow to green) after nn seconds (60) of cessation of the cause.

walterar:
There is something strange; the AI seems to be no working properly. But I'm not sure it's because of this PR. I will continue testing to confirm.

fluffyfreak:
@walterar to make this cleaner and the issue separate I'm closing this PR and have opened separate PR's for #3327 and #3328

PR #3321, opened 2015-01-14
Many point stars on starless galaxy

fluffyfreak:
The title may be confusing.

Some time ago we switched to having the stars baked into the skybox texture. This is great, it allows for a lot more artist control.

On the downside however is that you can get smearing, stretching and deformation of the skybox image which means that stars being point sources don't always have a constant size.

This is a proposed change to use the skybox for just the galaxy image and have the stars drawn using points. To get a similar look to the skybox only approach we must increase the number of stars by **at least** 10 times. From 10,000 to 100,000 points specifically.

Please test, compare, contrast and let me know what you think as this is more of an artistic / design issue. Do mention however if you suffer from poor performance when using it compared to the usual version.

Thanks,

Andy

### DIFF
--- a/src/Background.h
+++ b/src/Background.h
@@ -60,7 +60,7 @@ namespace Background
 
 	private:
 		void Init();
-		static const int BG_STAR_MAX = 10000;
+		static const int BG_STAR_MAX = 100000;
 		std::unique_ptr<Graphics::VertexBuffer> m_vertexBuffer;
 
 		//hyperspace animation vertex data


walterar:
This is a side effect from #3219. The first thing I tried is to increase the amount of 10000, but it does not work. From what I've seen, the problem is not the quantity but the size of the stars, which is much smaller.

walterar:
If you can not increase the size, what I liked most is around 40000.

impaktor:
Looks OK to me.

For some `10^5` stars from the Hipparcos catalogue, can be found here:
http://cdsarc.u-strasbg.fr/viz-bin/nph-Cat/tar.gz?I%2F311

impaktor:
I shouldn't high jack this PR with speculation on position of stars, so let's take that on the [forum](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=302)

fluffyfreak:
Note to everyone, try hyperspacing as well, tell me if it significantly slows down, or looks terrible to you or anything.

impaktor:
> ... if significantly slows down, or looks terrible to you or anything.

Works OK for me. Same FPS (30) for me.

Issue #2749, opened 2014-02-11
some translated strings are stored in the save game (and used after load game)


radius75:
example:
-start new game (EN)
-create save
-select other game lang.
-load game

some previously translated strings are used and visible, despite changes game language
![screenshot-20140211-114757](https://f.cloud.github.com/assets/1929111/2136053/09df1e80-930a-11e3-90f1-caa721e0291e.png)

This is not a good thing

impaktor:
> FWIW, I'm going to provide an exemple of how to proceed to NOT have those strings stored in the save file in my next PR, as I have solved (well, I had already, but this time I'll do it differently) the problem for the equipment types.

...and that was done in #3279, and works very well

PR #3324, opened 2015-01-17
Player Character title translateable

radius75:
see https://github.com/pioneerspacesim/pioneer/commit/cfb5d6c2848debadaf79404370d8ce8f3d991069#diff-4

positive for https://github.com/pioneerspacesim/pioneer/issues/2749

### DIFF
--- a/data/libs/Character.lua
+++ b/data/libs/Character.lua
@@ -1179,7 +1179,7 @@ local onGameStart = function ()
 		-- the average values.  We'll find some way to ask the
 		-- player for a new name in the future.
 		local PlayerCharacter = Character.New()
-		PlayerCharacter.title = 'Commander'
+		PlayerCharacter.title = 'COMMANDER'
 		PlayerCharacter.player = true
 		-- Gave the player a missions table (for Misssions.lua)
 		PlayerCharacter.missions = {}
--- a/data/libs/ui/InfoFace.lua
+++ b/data/libs/ui/InfoFace.lua
@@ -21,7 +21,7 @@ local setFaceInfo = function (face, character)
 				ui:Gradient({r=0.1,g=0.1,b=0.1,a=0.8}, {r=0.0,g=0.0,b=0.1,a=0.0}, "HORIZONTAL"):SetInnerWidget(
 					ui:Margin(20):SetInnerWidget(ui:VBox():PackEnd({
 						ui:Label(character.name):SetFont("HEADING_NORMAL"),
-						ui:Label(character.title or ''):SetFont("HEADING_SMALL"),
+						ui:Label(character.titleTranslate or ''):SetFont("HEADING_SMALL"),
 					}))
 				)
 			)
--- a/data/modules/CrewContracts/CrewContracts.lua
+++ b/data/modules/CrewContracts/CrewContracts.lua
@@ -396,11 +396,12 @@ local onCreateBB = function (station)
 									hopefulCrew.piloting,
 									hopefulCrew.navigation,
 									hopefulCrew.sensors)
+		hopefulCrew.title = 'GENERAL_CREW'
 		if maxScore > 45 then
-			if hopefulCrew.engineering == maxScore then hopefulCrew.title = l.SHIPS_ENGINEER end
-			if hopefulCrew.piloting == maxScore then hopefulCrew.title = l.PILOT end
-			if hopefulCrew.navigation == maxScore then hopefulCrew.title = l.NAVIGATOR end
-			if hopefulCrew.sensors == maxScore then hopefulCrew.title = l.SENSORS_AND_DEFENCE end
+			if hopefulCrew.engineering == maxScore then hopefulCrew.title = 'SHIPS_ENGINEER' end
+			if hopefulCrew.piloting == maxScore then hopefulCrew.title = 'PILOT' end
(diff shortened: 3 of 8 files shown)


impaktor:
> see cfb5d6c#diff-4

I don't think I follow, where did (the removal of) `libs/Translate.lua` come from? That's not in current master, and not in the (sum) of your commits either?

I'd poke @laarmen to review this, if it weren't for the fact that he recently announced that he currently doesn't have much time for pioneer

radius75:
commit cfb5d6c is a part of https://github.com/pioneerspacesim/pioneer/pull/2521/commits
there player title no longer be translated

laarmen:
"all in one" -> Please don't.

Reviewing a patchset is painful enough with correct commit logs, but reading a diff without any idea what it is all about is just a special brand of masochism. b3c85768 is an ideal commit message :-)

fluffyfreak:
Ok @radius75 sorry for only just getting around to this one :(

The code itself looks ok, maybe the Git commit history could be cleaned up but that's not my personal priority so long as the feature does what it's supposed to :)

What else needs doing to get this merged? I see changes to `ui-core/en.json` so does this need something doing on (@robn) Transifex too?

impaktor:
> What else needs doing to get this merged? I see changes to ui-core/en.json so does this need something doing on (@robn) Transifex too?

Nope. But if it had added new files then yes.

Generally it is a bad idea to move strings from modules to (ui-)core, right? Since it "de-modulerizes" pioneer.

fluffyfreak:
Maybe but is there another way to make those strings available for translation?

impaktor:
Could we hold on this, I'd like to give it a test first. Give me a day or two.

fluffyfreak:
Yeah of course :) I just realised that I hadn't really looked at it before and it had been here for a while.

radius75:
Currently "crew member title"/"station crew" strings are scattered:

in ui-core: GENERAL_CREW, COMMANDER, STATION_MANAGER, CONSTABLE,CHIEF_MECHANIC

in module-crewcontract: NAVIGATOR, PILOT, SENSORS_AND_DEFENCE, SHIPS_ENGINEER

To be able to easily translate "title" in the text and in the images (InfoFace.lua
), I moved them in the same place ```ui-core```.

impaktor:
>What else needs doing to get this merged? I see changes to ui-core/en.json so does this need something doing on (@robn) Transifex too? 

I change my mind, to __yes__, since this is _moving_ already translated strings, there is some @robn magic needed I think. I didn't think that far in my first iteration of reading the question.

impaktor:
OK, I've looked at this some now. So the actual _bug_ being fixed by this is that "Commander" wasn't being translated. All other strings worked but were inconsistent in which language module they were stored.

I don't have any strong objections to this.. I'll harass @robn this weekend to fix the auto-copy of all the translated files, so this can be merged.

radius75:
It is worth mentioning:
If you use the older save than this FIX, then we will have "empty title" ```''``` instead of the previously translated string (crew and player title stored in old Save).
In New Game work as it should.

fluffyfreak:
What's happening with this PR? Is it active or can it be closed?

impaktor:
I think it's active, and waiting for @robn magic on transifex side, since it moves already translated strings.

But now tthere's also merge-conflicts I see. Would be good if those were solved.

radius75:
I resolve conflicts this weekend

radius75:
Done

robn:
I've move the language keys from `module-crewcontracts` to `ui-core`. I haven't touched the rest of this PR; I'll leave that for someone else.

radius75:
I cleaned up this PR. Updated and now one commit in branch.

radius75:
I think the translation on transifex has been prepared by @robn in https://github.com/pioneerspacesim/pioneer/commit/61d75a60460af21d9694a4ba5e61e9fd377f2695 and  PR is ready to merge:)

impaktor:
> I think the translation on transifex has been prepared

Yes. Now the question is if the code is "sane" enough to go in. I understand what the aim is (not have translated strings "hardcoded" in the save), but rob expressed some reluctance to merging it, as it might not be the optimal way to do it. I don't know myself. Maybe @johnbartholomew @laarmen or @fluffyfreak have opinions?

fluffyfreak:
This just translates the job title yes? So `Commander` becomes `Kommondant` in German for example? Then I'm ok with it.

impaktor:
> This just translates the job title yes? 

Not quite, it splits the `.title` into `.title` and `.titletranslate`, apparently to aviod including translated strings in the save file, making it language dependent. I'm not entierly sure why/if `.titletranslate` is needed for this? Why not simply use `.title` in the translation system when we want the title? `l[character.title]`? I'll admitt I have not gone in-depth here, but it does bother me slightly to have both a `title` variable _and_ a `titletranslate`, the latter "feeling" slightly like a sub-optimal solution. If they're really needed then a `character.get_title()` might feel more natural?

radius75:
Translation in the ```data/libs/ui/InfoFace.lua``` is possible?

via: https://github.com/pioneerspacesim/pioneer/commit/cfb5d6c2848debadaf79404370d8ce8f3d991069

I applied the ```titletranslate``` that does not translate into libs. I had no better idea.

PR #3328, opened 2015-01-20
Ship nearby optimise

fluffyfreak:
Only update the list of nearby ships for every ship once per second.

Split out of #3320

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -1315,6 +1315,9 @@ void Ship::EnterHyperspace() {
 		return;
 	}
 
+	// Clear ships cached list of nearby bodies so we don't try to access them.
+	m_nearbyBodies.clear();
+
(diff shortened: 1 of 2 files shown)


laarmen:
By definition, a Bite-sized tag applies to issues, not PRs, since it is supposed to qualify issues that a newcomer could tackle on his own :-)

Cf OpenHatch (the Pioneer page is still buggy, but that's another story)

laarmen:
What happens if one tries to access a ship that was nearby and has been destroyed ? Glancing at the diff, I don't see this taken into account.

fluffyfreak:
It would probably crash, I'll need to handle that.

fluffyfreak:
@laarmen nah actually it's ok.

That list is only accessed within the loop where it is also updated.

laarmen:
No way to access it from Lua ?

fluffyfreak:
Nope :)

PR #3332, opened 2015-01-22
Fix the UI strange rendering.

### LINKED ISSUE #2779: Invisible ink
someone has the same? vs2013 error?
screenshot-20140221-191754
screenshot-20140221-191758
screenshot-20140221-191802

mostly when minimize/maximize game window.

### LINKED ISSUE #3308: Strange rendering
After the latest merges yesterday/today: #3303 (and #3306) I'm seeing some pretty funky stuff.

When flying it seems like the ground, or parts of it, follows me up into the air, and it looks like the city buildings are sunk into the earth when starting a game. It's a bit difficult to take a screenshot of, since a video would be needed to show it properly, but this is also pretty strange:

screenshot-20150108-110454

screenshot-20150108-111201

Interestingly, Barnard's Star seems to work fine, but not starting from a planet side space port.

fluffyfreak:
Clear the depth buffer so that it draws the UI fresh.

Trying to fix #3308 and fix #2779

Slight problem is that I can't test it, Linux users (@impaktor) please give it a try!

### DIFF
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -1224,6 +1224,8 @@ void Pi::MainLoop()
 		SetMouseGrab(Pi::MouseButtonState(SDL_BUTTON_RIGHT));
 
 		Pi::renderer->EndFrame();
+
+		Pi::renderer->ClearDepthBuffer();
 		if( DrawGUI ) {
 			Gui::Draw();
 			if (game)
--- a/src/ui/Context.cpp
+++ b/src/ui/Context.cpp
@@ -144,6 +144,7 @@ void Context::Update()
 void Context::Draw()
 {
 	Graphics::Renderer *r = GetRenderer();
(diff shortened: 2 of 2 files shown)


impaktor:
Will give it a try tomorrow. 
Thanks.

impaktor:
OK, tested this just now.
BBS is fixed by this but not world view. The ground is too high up, and "follows me around" so to say, so that part is still as in #3308.

fluffyfreak:
This is a bit brute force of an approach but should definitely work.

radius75:
Wow, it also fixes  #2779 for me.

impaktor:
> Wow, it also fixes #2779 for me.

Awesome.

I'll test this ASAP, which for me will be on Monday (if this isn't merged before then).

PR #3349, opened 2015-02-06
LuaSerializer refactor

robn:
This is fully decoupling the Lua serialization setup from the game proper. The big problem is that LuaSerializer has special knowledge of Lua class wrappers. These should have been hidden away inside the wrappers proper.

So now any class wrapper that wants to be serializable registers a serializer and deserializer with LuaObject. LuaSerializer calls into LuaObject for those. Its all very nice.

And then, if you need to take LuaSerializer as a dependency (eg for simple programs like uitest), you don't need eg Body unless you take LuaBody. And since that depends on Game, you don't need that either :)

I've been careful to keep the savefile format the same. There's a single line we can clean up at next savefile bump. Its noted in SAVEBUMP.txt.

After this, I've got a single commit which will get uitest building again, and then I can get back to UI work!

### DIFF
--- a/SAVEBUMP.txt
+++ b/SAVEBUMP.txt
@@ -13,5 +13,5 @@ Regular tasks
 - data/libs/NameGen.lua: Add new authors to name generator
 
 One-time tasks
-
-[ There is nothing we know we need to do at the next savebump. ]
+- Remove serializer registration for ModelSkin (see XXX comment at bottom for
+  scengraph/LuaModelSkin.cpp)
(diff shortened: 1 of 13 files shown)


laarmen:
I hereby commit myself to reviewing this PR this afternoon (Europe time).

laarmen:
All right, I started the review but since I saw methods from LuaRef which I
didn't recognize (which was weird, since I was thus far almost the only
author of the module as far as I knew) it took more time than expected, and
I had to pay some attention to my lecture so... I'll take it up again
tomorrow.

robn:
Thanks @laarmen, I appreciate the effort.

fluffyfreak:
I haven't looked at the code but I can confirm that it builds just fine under VS2013 so no worries there :)

PR #3350, opened 2015-02-11
This is the branch with the JSON save/load system.

nick-the-ninja:
It is written, but still contains at least one bug (currently crashes when
loading the views). All the code I've written has a comment starting with
"npw -" for easy identification.
The JSON save/load functions are located side-by-side with the existing
serialisation. The pioneer code is currently set to use the JSON version.
Existing Serialize and Unserialize functions have ToJson and FromJson
functions next to them. Existing Save and Load functions have SaveToJson
and LoadFromJson functions next to them. The JSON system has the SAME
structure as the existing system, but implemented as nested JSON data.

With help from nozmajner on IRC, I have committed this branch, but I
haven't done a MERGE. Truth is, I've always been the only person working
on any code, so never had to do a proper merge. I was planning on doing a
manual merge using notepad++, but this is what has been done:
I've copied all files containing the JSON save/load code to my local git
and committed it. If none of the files have been modified by others since
about 3rd January then this will be fine, but I suspect this is not the
case! I'll reiterate this: any files listed below that have changed since
3rd January will not contain any of those changes - only my JSON code!
Hopefully there are only a few files that will need to have the
appropriate code added/changed.

I've probably put several days worth of coding time into this, and I need
to focus on other things now. Aside from the remaining bug(s), I believe
the code to be well written and robust, but will need further testing
(e.g. get a criminal record and then save and load the game, or put game
in other states and then save/load). My frame rate is 3 frames per second,
and the game takes 20 minutes to start in debug mode, which makes testing
difficult! I hope that recent optimisations will have fixed frame rate.

Im apologise for my lack of knowledge of git and non-manual merging, but
I'm a bit out of my depth with this. I'm happy to continue working on the
remaining bug(s) and testing, but thought I ought to start using git and
do a merge before upstream moves on too far.

All "save" code is prefixed with "// npw - new code" and all "load" code
is prefixed with "npw - new code (under construction)". I've not removed
the "under construction" yet, but all the code should work, so I'm
expecting that the load views bug is something silly I've done.

The JSON save/load code is easy to follow. It mirrors the existing
save/load code - just scroll up or down a bit to see the existing code.
There are only two new files: JsonUtils.h and JsonUtils.cpp
Please enter the commit message for your changes. Lines starting

### DIFF
--- a/contrib/json/JsonUtils.cpp
+++ b/contrib/json/JsonUtils.cpp
@@ -1,0 +1,337 @@
+// npw - new code
+
+#include "JsonUtils.h"
+#include "utils.h"
+#include "Serializer.h" // Need this for the exceptions (probably should move the exceptions out of Serializer.h)
+
+void VectorToJson(Json::Value &jsonObj, const vector3f &vec, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value vecArray(Json::arrayValue); // Create JSON array to contain vector data.
+	vecArray[0] = FloatToStr(vec.x);
+	vecArray[1] = FloatToStr(vec.y);
+	vecArray[2] = FloatToStr(vec.z);
+	jsonObj[name] = vecArray; // Add vector array to supplied object.
+}
+
+void VectorToJson(Json::Value &jsonObj, const vector3d &vec, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value vecArray(Json::arrayValue); // Create JSON array to contain vector data.
+	vecArray[0] = DoubleToStr(vec.x);
+	vecArray[1] = DoubleToStr(vec.y);
+	vecArray[2] = DoubleToStr(vec.z);
+	jsonObj[name] = vecArray; // Add vector array to supplied object.
+}
+
+void QuaternionToJson(Json::Value &jsonObj, const Quaternionf &quat, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value quatArray(Json::arrayValue); // Create JSON array to contain quaternion data.
+	quatArray[0] = FloatToStr(quat.w);
+	quatArray[1] = FloatToStr(quat.x);
+	quatArray[2] = FloatToStr(quat.y);
+	quatArray[3] = FloatToStr(quat.z);
+	jsonObj[name] = quatArray; // Add quaternion array to supplied object.
+}
+
+void QuaternionToJson(Json::Value &jsonObj, const Quaterniond &quat, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value quatArray(Json::arrayValue); // Create JSON array to contain quaternion data.
+	quatArray[0] = DoubleToStr(quat.w);
+	quatArray[1] = DoubleToStr(quat.x);
+	quatArray[2] = DoubleToStr(quat.y);
+	quatArray[3] = DoubleToStr(quat.z);
+	jsonObj[name] = quatArray; // Add quaternion array to supplied object.
+}
+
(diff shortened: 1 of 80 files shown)


laarmen:
Hi!

Thanks for all the work!

Before actually looking at the code, a few remarks:

  - You don't need to prefix your code with comments saying it is yours. Git knows it already, so the comments are redundant and add noise to the PR.
  - I wouldn't have kept around the old code. Once again, git can still dig it up if need be, and this way we make sure you don't call the old code by mistake.
  - A patch adding with +3200/-300 makes me a bit nervous. Glancing at the diff I see lines removed which don't seem related with serializing at first sight. Would it be possible to have those modifications in separate commits with proper explanations as to why those changes ?
  - Finally, the Travis build failed. I'm guessing you're using Windows ? It would be nice to fix it, but I guess the first Linux dev that'll try out your branch will provide you with the patch.

nick-the-ninja:
Thanks for your feedback.
I'll remove the comments at some point. I put them there in case I do a manual merge (so I can easily identify my code) because my working copy is not managed by git. In fact, I just copied the files I modified to an up-to-date local git copy - so all changes since 3rd of January (when I downloaded the source and begain work) will not be in the files - hence the +3200/-300 and removed code you spotted.
I'm not familiar with git, nor merging, but I'm learning. My plan is to do a proper merge and update the branch.
There is also a bug I need to fix. Apart from that, the JSON save/load code is written.

fluffyfreak:
I think we should write up a how-to guide for Git & Pioneer. Typically your branch *is* your working copy. You make [many small commits](https://github.com/fluffyfreak/pioneer/compare/asteroid) which might only be a few lines or might be hundreds but they represent each chunk of work done.

Then when you do a PR like this there is a clear trail of development to follow so others can see why you have done things, rather than a single large commit showing only the end result.

impaktor:
> I think we should write up a how-to guide for Git & Pioneer. 

Other than http://pioneerwiki.com/wiki/Using_git_and_GitHub?

nick-the-ninja:
This branch and pull request is really just a practice for me (I'm unfamiliar with git) and for others who are interested to have a look at what I've been doing.

I'm working on pioneer source that I downloaded on 3rd January which isn't managed by git (I'm gradually learning it). So when it comes to merging - it'll all be added in one go! Sorry about this, but I'd never have got started if I had to learn git first. My skills are C++, maths & physics, and object oriented programming, but I'm not so experienced with software management and merging tools (I've always been the only programmer working on any code, so I've never had to do a real merge).

The move from binary flat file to JSON is kind of all-or-nothing - in the sense that a mixed save system will be a mess that won't work. So adding it all in one go isn't so bad :)

Also note:
I'm not changing the actual data saved - just moving from binary flat to JSON, and there is a direct one-to-one correspondance between the existing and new JSON save system (which is why I've left the existing code in, for now). Therefore, the new system is easy to follow - by comparing it to the existing code.

fluffyfreak:
Ok, what we're going to have to do is fix this up properly but this PR can't be merged either as-is or with tidying up because it seems you have done the work in another folder and then just copied your changed files into an up-to-date copy of the repository.

That's brought in old and broken code and this PR won't compile and would need changes just to allow that to happen.

@nick-the-ninja I've posted on the forums about how I would proceed to create a clean commit history and make a new PR. It's not too demanding, I'll often do it myself to bring in changes from work that was just too experimental and messy for a PR.

For now though I am going to close this particular PR, please don't be put off. I think this could be good work, but I'm having to fix it up and undo a lot of things because of the file copying locally which means that it cannot be merged in it's current state.

fluffyfreak:
Ok, I got this to build and run locally but to do so I had to undo many of the changes to the following files:
* GalacticView.h/cpp
* Game.cpp
* LuaRef.h/cpp
* LuaSerializer.h/cpp
* PropertyMap.h/cpp
* SectorView.h/cpp
* ShipCPanelMultiFuncDisplay.h/cpp
* WorldView.h/cpp
* utils.h/cpp

then update the json project with the utils files.
The problem is that you copied in the FILES, not just the code changes and so you overwrote newer files with older ones and then committed those older files over the new ones.

Issue #3345, opened 2015-02-05
Red alert in recovery saved games.


walterar:
This happens sometimes and can be seen at the beginning of the video. You need to load it twice.
http://youtu.be/X_XE2xdpO5g
The video is from Pioneer Scout+, but this also happens in the Pioneer clean motor. I was lucky to capture it.

fluffyfreak:
Did the comms window not work the first time? it's hard to tell.

walterar:
When loading the first time: Red Alert. Exit and reload: Green Alert.

walterar:
When the game was recorded, was in Green Alert (ALERT_NONE).
When he recovers, he is on Red Alert (ALERT_SHIP_FIRING). (incorrect)
I close and I reload, is on Green Alert (ALERT_NONE). (correct)

It is understood better now?

nick-the-ninja:
This has happened for JSONSerialise branch too (not surprising - as JSONSerialise serialises exactly the same data).
The cause of this (and similar issues) will be much easier to track down with a JSON savegame file.

walterar:
Yes, that's what I think. Use your weapons! :))

impaktor:
@walterar was this bug fixed by #3355?

walterar:
@Impaktor Fixed in Linux, not in Windows

walterar:
Verified in latest version for Windows compiled by me with MXE and VS2013. The latest version for Windows published in http://sourceforge.net/projects/pioneerspacesim/files/pioneer-20150319-win32.7z/download not incorporate #3355

fluffyfreak:
So @robn we need a new windows build doing then is what you're trying to say?

nozmajner:
I mentioned this on IRC before, but let me note this here next to a build call. :)
The latest two windows builds (0314, 0319) are still showing 20150201(50c25a6) version number in the main menu, and and it seems that the new features (headlook for example) aren't showing up in them.

robn:
Noted, I'll check as soon as I can.

joonicks:
bug still present in 20150912 build (windows)
specifically I got it after exiting hyperspace, "laser fire detected" with nothing but me and my hyperspace cloud around.

notinlist:
[...]

PR #3355, opened 2015-02-24
JSONSerialise: migrating the save-game serialisation from a binary flat file to JSON

nick-the-ninja:
The branch JSONSerialise contains the JSON serialisation code.

The JSON serialisation saves precisely the same data as the original serialisation. There is a one-to-one correspondance between the original and JSON serialisation code. Part of the review should be a cross-check between the two.

There are two new files: JsonUtils.h and JsonUtils.cpp

Essentially, the relevant "Serialize" and "Unserialize" functions have been replaced by "ToJson" and "FromJson" functions. The relevant "Save" and "Load" functions have been replaced by "SaveToJson" and "LoadFromJson" functions. However, "ModelSkin::Save" and "ModelSkin::Load" are called by code other than the savegame process, so they remain (next to their corresponding "SaveToJson" and "LoadFromJson" functions).

The structure of the original serialisation code has been preserved. Relevant comments have also been carried over to the JSON serialisation code.

All JSON serialisation functions pass a reference to a JSON Value that is a JSON object. The only exception to this is "Model::SaveToJson" and "Model::LoadFromJson" where the SaveVisitor/LoadVisitor (now SaveVisitorJson/LoadVisitorJson) is passed a JSON Value that is a JSON array.

The save code typically passes a reference to the current JSON object into a function where other JSON objects and/or primitives are created and attached to the supplied object. Some serialisation functions are called recursively (as does the original serialisation).

The load code navigates the JSON tree as expected. All load functions begin by checking the contents of their supplied JSON object for appropriately named sub-objects/values. In the future, when the serialisation code has settled into a future-proof state, it is here that checks on missing objects can be put, and appropriate action taken for backwards compatibility.

JSON arrays of objects are used as follows:
When saving, the array is created and a loop entered, where for each iteration, an element object is created and passed into the JSON function which then populates it. The object is then appended to the array. Finally, when the loop is done, the array is added to the current object. The process can be easily followed.

Issues to be addressed/considered:

(1) Hexfloats
I couldn't get them to work (see forum topic "Hexfloats"), so floats and doubles are saved by memcopying them to an int of the same size, then the int is saved. The reverse is done to load the float/double. Although the data suffers no loss of precision, this is not ideal because the data in the JSON save file is not human readable (although float 0.000000 is written as integer "0").

(2) Lua data
This large string is not preserved when saved and loaded due to the non-printable characters at or near the end. The current solution is to save as a JSON array of integers (each integer is a char of the string). This works, but increases the size of the save file. The data is also not human readable.

(3) #include "Serialiser.h"
These remain, because the JSON serialisation code uses the exceptions. Perhaps the exceptions should be moved? Other code may also require functions declared in "Serialiser.h".

I have come across issues that (for me) also occur in pioneer from upstream. Please check this is not the case before flagging the issue as a JSONSerialise bug :)

Thanks :)
Nick

### DIFF
--- a/contrib/json/JsonUtils.cpp
+++ b/contrib/json/JsonUtils.cpp
@@ -1,0 +1,362 @@
+// Copyright © 2008-2015 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+#include "JsonUtils.h"
+#include "utils.h"
+#include "Serializer.h" // Need this for the exceptions
+
+void VectorToJson(Json::Value &jsonObj, const vector3f &vec, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value vecArray(Json::arrayValue); // Create JSON array to contain vector data.
+	vecArray[0] = FloatToStr(vec.x);
+	vecArray[1] = FloatToStr(vec.y);
+	vecArray[2] = FloatToStr(vec.z);
+	jsonObj[name] = vecArray; // Add vector array to supplied object.
+}
+
+void VectorToJson(Json::Value &jsonObj, const vector3d &vec, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value vecArray(Json::arrayValue); // Create JSON array to contain vector data.
+	vecArray[0] = DoubleToStr(vec.x);
+	vecArray[1] = DoubleToStr(vec.y);
+	vecArray[2] = DoubleToStr(vec.z);
+	jsonObj[name] = vecArray; // Add vector array to supplied object.
+}
+
+void QuaternionToJson(Json::Value &jsonObj, const Quaternionf &quat, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value quatArray(Json::arrayValue); // Create JSON array to contain quaternion data.
+	quatArray[0] = FloatToStr(quat.w);
+	quatArray[1] = FloatToStr(quat.x);
+	quatArray[2] = FloatToStr(quat.y);
+	quatArray[3] = FloatToStr(quat.z);
+	jsonObj[name] = quatArray; // Add quaternion array to supplied object.
+}
+
+void QuaternionToJson(Json::Value &jsonObj, const Quaterniond &quat, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value quatArray(Json::arrayValue); // Create JSON array to contain quaternion data.
+	quatArray[0] = DoubleToStr(quat.w);
+	quatArray[1] = DoubleToStr(quat.x);
+	quatArray[2] = DoubleToStr(quat.y);
+	quatArray[3] = DoubleToStr(quat.z);
+	jsonObj[name] = quatArray; // Add quaternion array to supplied object.
+}
+
+void MatrixToJson(Json::Value &jsonObj, const matrix3x3f &mat, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value matArray(Json::arrayValue); // Create JSON array to contain matrix data.
+	matArray[0] = FloatToStr(mat[0]);
+	matArray[1] = FloatToStr(mat[1]);
+	matArray[2] = FloatToStr(mat[2]);
+	matArray[3] = FloatToStr(mat[3]);
+	matArray[4] = FloatToStr(mat[4]);
+	matArray[5] = FloatToStr(mat[5]);
+	matArray[6] = FloatToStr(mat[6]);
+	matArray[7] = FloatToStr(mat[7]);
+	matArray[8] = FloatToStr(mat[8]);
+	jsonObj[name] = matArray; // Add matrix array to supplied object.
+}
+
+void MatrixToJson(Json::Value &jsonObj, const matrix3x3d &mat, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value matArray(Json::arrayValue); // Create JSON array to contain matrix data.
+	matArray[0] = DoubleToStr(mat[0]);
+	matArray[1] = DoubleToStr(mat[1]);
+	matArray[2] = DoubleToStr(mat[2]);
+	matArray[3] = DoubleToStr(mat[3]);
+	matArray[4] = DoubleToStr(mat[4]);
+	matArray[5] = DoubleToStr(mat[5]);
+	matArray[6] = DoubleToStr(mat[6]);
+	matArray[7] = DoubleToStr(mat[7]);
+	matArray[8] = DoubleToStr(mat[8]);
+	jsonObj[name] = matArray; // Add matrix array to supplied object.
+}
+
+void MatrixToJson(Json::Value &jsonObj, const matrix4x4f &mat, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value matArray(Json::arrayValue); // Create JSON array to contain matrix data.
+	matArray[0] = FloatToStr(mat[0]);
+	matArray[1] = FloatToStr(mat[1]);
+	matArray[2] = FloatToStr(mat[2]);
+	matArray[3] = FloatToStr(mat[3]);
+	matArray[4] = FloatToStr(mat[4]);
+	matArray[5] = FloatToStr(mat[5]);
+	matArray[6] = FloatToStr(mat[6]);
+	matArray[7] = FloatToStr(mat[7]);
+	matArray[8] = FloatToStr(mat[8]);
+	matArray[9] = FloatToStr(mat[9]);
+	matArray[10] = FloatToStr(mat[10]);
+	matArray[11] = FloatToStr(mat[11]);
+	matArray[12] = FloatToStr(mat[12]);
+	matArray[13] = FloatToStr(mat[13]);
+	matArray[14] = FloatToStr(mat[14]);
+	matArray[15] = FloatToStr(mat[15]);
+	jsonObj[name] = matArray; // Add matrix array to supplied object.
+}
+
+void MatrixToJson(Json::Value &jsonObj, const matrix4x4d &mat, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value matArray(Json::arrayValue); // Create JSON array to contain matrix data.
+	matArray[0] = DoubleToStr(mat[0]);
+	matArray[1] = DoubleToStr(mat[1]);
+	matArray[2] = DoubleToStr(mat[2]);
+	matArray[3] = DoubleToStr(mat[3]);
+	matArray[4] = DoubleToStr(mat[4]);
+	matArray[5] = DoubleToStr(mat[5]);
+	matArray[6] = DoubleToStr(mat[6]);
+	matArray[7] = DoubleToStr(mat[7]);
+	matArray[8] = DoubleToStr(mat[8]);
+	matArray[9] = DoubleToStr(mat[9]);
+	matArray[10] = DoubleToStr(mat[10]);
+	matArray[11] = DoubleToStr(mat[11]);
+	matArray[12] = DoubleToStr(mat[12]);
+	matArray[13] = DoubleToStr(mat[13]);
+	matArray[14] = DoubleToStr(mat[14]);
+	matArray[15] = DoubleToStr(mat[15]);
+	jsonObj[name] = matArray; // Add matrix array to supplied object.
+}
+
+void ColorToJson(Json::Value &jsonObj, const Color3ub &col, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value colArray(Json::arrayValue); // Create JSON array to contain color data.
+	colArray[0] = col.r;
+	colArray[1] = col.g;
+	colArray[2] = col.b;
+	jsonObj[name] = colArray; // Add color array to supplied object.
+}
+
+void ColorToJson(Json::Value &jsonObj, const Color4ub &col, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value colArray(Json::arrayValue); // Create JSON array to contain color data.
+	colArray[0] = col.r;
+	colArray[1] = col.g;
+	colArray[2] = col.b;
+	colArray[3] = col.a;
+	jsonObj[name] = colArray; // Add color array to supplied object.
+}
+
+void BinStrToJson(Json::Value &jsonObj, const std::string &binStr, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	Json::Value binStrArray(Json::arrayValue); // Create JSON array to contain binary string data.
+	for (unsigned int charIndex = 0; charIndex < binStr.size(); ++charIndex)
+		binStrArray[charIndex] = int(binStr[charIndex]);
+	jsonObj[name] = binStrArray; // Add binary string array to supplied object.
+}
+
+void JsonToVector(vector3f *pVec, const Json::Value &jsonObj, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+	
+	if (!jsonObj.isMember(name.c_str())) throw SavedGameCorruptException();
+	Json::Value vecArray = jsonObj[name.c_str()];
+	if (!vecArray.isArray()) throw SavedGameCorruptException();
+	if (vecArray.size() != 3) throw SavedGameCorruptException();
+
+	pVec->x = StrToFloat(vecArray[0].asString());
+	pVec->y = StrToFloat(vecArray[1].asString());
+	pVec->z = StrToFloat(vecArray[2].asString());
+}
+
+void JsonToVector(vector3d *pVec, const Json::Value &jsonObj, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	if (!jsonObj.isMember(name.c_str())) throw SavedGameCorruptException();
+	Json::Value vecArray = jsonObj[name.c_str()];
+	if (!vecArray.isArray()) throw SavedGameCorruptException();
+	if (vecArray.size() != 3) throw SavedGameCorruptException();
+
+	pVec->x = StrToDouble(vecArray[0].asString());
+	pVec->y = StrToDouble(vecArray[1].asString());
+	pVec->z = StrToDouble(vecArray[2].asString());
+}
+
+void JsonToQuaternion(Quaternionf *pQuat, const Json::Value &jsonObj, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	if (!jsonObj.isMember(name.c_str())) throw SavedGameCorruptException();
+	Json::Value quatArray = jsonObj[name.c_str()];
+	if (!quatArray.isArray()) throw SavedGameCorruptException();
+	if (quatArray.size() != 4) throw SavedGameCorruptException();
+
+	pQuat->w = StrToFloat(quatArray[0].asString());
+	pQuat->x = StrToFloat(quatArray[1].asString());
+	pQuat->y = StrToFloat(quatArray[2].asString());
+	pQuat->z = StrToFloat(quatArray[3].asString());
+}
+
+void JsonToQuaternion(Quaterniond *pQuat, const Json::Value &jsonObj, const std::string &name)
+{
+	assert(!name.empty()); // Can't do anything if no name supplied.
+
+	if (!jsonObj.isMember(name.c_str())) throw SavedGameCorruptException();
+	Json::Value quatArray = jsonObj[name.c_str()];
+	if (!quatArray.isArray()) throw SavedGameCorruptException();
+	if (quatArray.size() != 4) throw SavedGameCorruptException();
+
+	pQuat->w = StrToDouble(quatArray[0].asString());
+	pQuat->x = StrToDouble(quatArray[1].asString());
+	pQuat->y = StrToDouble(quatArray[2].asString());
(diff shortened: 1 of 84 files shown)


fluffyfreak:
I will try and find time to look at this as soon as possible :)

nick-the-ninja:
Much appreciated :)
This may provide some useful information: http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=263&start=70

nick-the-ninja:
Have found some surface base nav-lights that don't work when in normal time (frozen). E.g. base on Ganymede. Will investigate.

walterar:
Hello nick-the-ninja
I'm anxious to try this, but I have some errors when compiling in Linux. I can spend error messages if you wish.

walterar:
Maybe this will serve:

[...]
Shields.cpp: In member function ‘virtual void Shields::SaveToJson(Json::Value&)’:
Shields.cpp:225:28: error: conversion from ‘std::vector<Shields::Shield>::size_type {aka long unsigned int}’ to ‘const Json::Value’ is ambiguous
  shieldsObj["num_shields"] = m_shields.size();
                            ^
Shields.cpp:225:45: note: candidates are:
  shieldsObj["num_shields"] = m_shields.size();
                                             ^

fluffyfreak:
Changing
```
shieldsObj["num_shields"] = m_shields.size();
```
to
```
shieldsObj["num_shields"] = Json::Value::UInt(m_shields.size());
```
Will fix those kinds of errors.

Sorry I haven't had time to check this out yet.

walterar:
We are closer.
[...]
  CXXLD  modelcompiler
scenegraph/libscenegraph.a(Model.o): In function `SceneGraph::LoadVisitorJson::ApplyMatrixTransform(SceneGraph::MatrixTransform&)':
/media/auxiliar/Building/APIONEER64/pioneergit2/src/scenegraph/Model.cpp:476: undefined reference to `JsonToMatrix(matrix4x4<float>*, Json::Value const&, std::string const&)'
scenegraph/libscenegraph.a(Model.o): In function `SceneGraph::SaveVisitorJson::ApplyMatrixTransform(SceneGraph::MatrixTransform&)':
/media/auxiliar/Building/APIONEER64/pioneergit2/src/scenegraph/Model.cpp:442: undefined reference to `MatrixToJson(Json::Value&, matrix4x4<float> const&, std::string const&)'
collect2: error: ld returned 1 exit status
make[2]: *** [modelcompiler] Error 1

nick-the-ninja:
Thanks for the feedback. Will fix the linux errors when I can.
Anyone please feel free to go ahead and make JSONSerialise linux friendly and update the branch :)

nick-the-ninja:
Hi walterar,
I was writing post before yours came in - just seen it.
Thanks for your help :)
Have you made the code changes suggested by fluffyfreak?

nick-the-ninja:
I've merged in the changes to upstream. Will push to origin.

walterar:
Linux complains that #include "json/JsonUtils.h" missing in many xx.h eg. Shields.h and others. You surely know better than I, where to locate them. :)

nick-the-ninja:
OK will do. And thanks for the feedback :)
But why does linux require that Shields.h include json/JsonUtils.h ?
Shields.h doesn't need to know anything about the functions in JsonUtils.h
Will have a look this evening if I can.
Thank you

walterar:
This may help: https://travis-ci.org/pioneerspacesim/pioneer/builds/52980587

nick-the-ninja:
FYI: It's json.h that needs including in some header files (not JsonUtils.h which contains JSON helper functions for serialisation). MSVC somehow managed to build it :)

nick-the-ninja:
Travis CI build now runs for 21.5 mins.
Now fails on "undefined reference" to functions in JsonUtils - but the files all #include "json/JsonUtils.h" so I'm not sure what is happening here?
Will try to figure it out this evening.

nick-the-ninja:
Looks like linux might not be compiling JsonUtils.cpp
Any idea why?

fluffyfreak:
The MSVC version will have included it because it uses Pre-Compiled Headers to speedup compilation. So the PCH.h file will have pulled in the "Json.h" header file. Linux & OSX don't use PCH files.

You will need to update the makefile at `contrib/json/makefile` to include the new `JsonUtils.h/cpp` files you've added as it doesn't know about them right now.

You're doing good by the way! This is the awkward cross-platform part I always have to go through too :)

nick-the-ninja:
Thanks :)
We should have something that builds on all platforms this evening.

nick-the-ninja:
contrib\jsonMakefile.am now contains:

#  Process this file with automake to produce Makefile.in
include $(top_srcdir)/Makefile.common

AM_CPPFLAGS += -I @top_srcdir@/contrib

noinst_LIBRARIES = libjson.a
libjson_a_SOURCES = jsoncpp.cpp JsonUtils.cpp
noinst_HEADERS = json.h JsonUtils.h

and Travis CI build fails as follows:
JsonUtils.h:8:21: fatal error: vector3.h: No such file or directory

So... I need to give the exact paths of include files in JsonUtils.h ?

#include "vector3.h"
should become:
#include "src/vector3.h" ?
or "../src/vector3.h"

I'm busy doing other things for the next 2 hours, but will occasionally check for replies :)

nick-the-ninja:
I've no idea how to fix this:

 CXX JsonUtils.o
In file included from /usr/include/c++/4.8/atomic:38:0,
from ../../src/RefCounted.h:7,
from ../../src/Random.h:17,
from ../../src/libs.h:67,
from ../../src/utils.h:11,
from JsonUtils.cpp:5:
/usr/include/c++/4.8/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.

nick-the-ninja:
Adding -std=c++0x to compiler flags might fix this?

nick-the-ninja:
Will try adding -std=c++11
AM_CPPFLAGS += -I @top_srcdir@/contrib -std=c++11
and try the Travis CI again.

walterar:
Congratulations! Compiles without errors. Start testing.

nick-the-ninja:
Yay! :D

walterar:
Load fine, but trying to save:
[...]
  *** Buffer overflow detected ***: src /./ pioneer terminated
[...]
:((

We are closer. Aspirin? :)

nick-the-ninja:
Hmmm. Works fine on MSVC++.
Maybe I've used a char array that isn't big enough. But I've been careful not to do that.
Any other info from the crash? Filenames?

walterar:
Here is the output, maybe help:
https://dl.dropboxusercontent.com/u/15391580/Pioneer/output.txt

nick-the-ninja:
Thanks!
sprintf seems to be the culpret.
SInt64ToStr might be the problem:
I've written an ascii 64 bit int into a 32 byte char array. I figured it was enough.

nick-the-ninja:
...also an ascii 32 bit int into a 16 byte char array.
I'll double the char array sizes and try again...

nick-the-ninja:
I've quadrupled the sizes of the char arrays and updated the PR.
Your help testing linux is much appreciated :)
Gonna have to call it a night.

nick-the-ninja:
All the instances of sprintf in the JSON serialisation code:

sprintf(str, "%I64d", val);
sprintf(str, "%I64u", val);
std::sprintf(str, "%I32u", intVal);
std::sprintf(str, "%I64u", intVal);
std::sprintf(str, "%I32u", intVal);
std::sprintf(str, "%I64u", intVal);

so the problem must be here somewhere.
I was experimenting with sprintf when trying to get Hexfloats to work, so some are printf and some are std::sprintf. I see both of them used elsewhere in the source.

The format specifiers might be to blame. I could try %d, %u, %ld and %lu (for 32 and 64 bit ints)?

walterar:
[...]
Loading took: 17804.469244 milliseconds
Game::LoadGame('Gates')
savefile version: 80
Creating new galaxy generator 'legacy' version 1
Clearing and re-using previous Galaxy object
StarSystemCache: misses: 0, slave hits: 0, master hits: 0
SectorCache: misses: 101, slave hits: 0, master hits: 1
Sol:
[...]

Everything seems to work fine. Time save / load is faster than I expected, "almost" no difference.

When retrieving a saved game, the game starts paused. Everything seems to be there, but to remove the pause have a system output.

There seems to be a problem with the clock recovered. Considers that may be a previous problem, which is now manifest.

nick-the-ninja:
That's great! It was either the format specifiers or the size of the buffer. Won't hurt to use larger sprintf buffers though.
I'll have a look at the clock issue tomorrow. Could you explain what you see please? (clock works for me)
Thanks for all your help!
Really must get to bed now :)

walterar:
Yes, it is time to sleep around here too. GMT-3 at the end of the world :))

The work has just begun, we continue tomorrow. :+1:

impaktor:
> When retrieving a saved game, the game starts paused. Everything seems to be there, but to remove the pause have a system output.

I could not reproduce this on GNU/Linux. Works fine for me (did a very fast testing at work).

nick-the-ninja:
Doesn't the game always load in a paused state?
Also, when in ESC menu:
press ESC - game returns as it was
click "Return to Game" - game returns in a paused state

"but to remove the pause have a system output" - unpause the game and get a crash / errors in log?

impaktor:
> Doesn't the game always load in a paused state?

Yes.

> "but to remove the pause have a system output" - unpause the game and get a crash / errors in log?

That is also how I interpreted it, and this is what I could not reproduce.

walterar:
The good news is that today, turning on my machine and try again, this error does not exist. Now I'm going to merge it in my master work to continue testing in depth. Any strange thing to see, I tell you.

nick-the-ninja:
Much appreciated :)
Good luck!

nick-the-ninja:
The navlights issue I mentioned above (frozen for surface bases) also occurs for the original serialisation. So to date, I have not found any issues in JSONSerialise that do not also occur for the original serialisation :)

walterar:
After many hours of play, I did not find any problems after applying this PR.
Nor was repeated #3345. I wonder if this fixed it, or is it a coincidence

nick-the-ninja:
That's fantastic walterar! Many thanks for your help testing.
JSONSerialise is looking good for windows and linux :)

fluffyfreak:
Ok, testing this and doing some timings by adding things like `MsgTimer timey;` and `timey.Mark("Game::SaveGame");`

Gives me these results for JSON save game:
> (52.81) seconds in Game::SaveGame
> (37.21) seconds in Game::SaveGame
> (3.33) seconds in Game::LoadGame
> (73.72) seconds in Game::SaveGame
> (42.61) seconds in Game::SaveGame

And this for the original non-JSON timings:
> (1.52) seconds in Game::SaveGame
> (1.90) seconds in Game::SaveGame
> (0.002) seconds in Game::LoadGame
> (1.43) seconds in Game::SaveGame
> (1.43) seconds in Game::SaveGame

This is a problem as a 1 and a half second pause when taking off is acceptable, but waiting for 1 minute 13 seconds each time you takeoff or dock, i.e: when the game autosaves.

@robn you interested in this still?

fluffyfreak:
I'm doing some profiling to see if there's anything that can be done to optimise the saving. Loading is perfectly acceptable.

nick-the-ninja:
I was getting terrible save and load times when running in debug mode (over 2 minutes for a load). Running the release version is very fast for me (a save in less than a second, a load in 1 second).
I am running pioneer by executing "pioneer.exe" itself though (I had crashes when running from MSVC++, even in release mode).
If for some people the save and load times are long, then maybe switch auto-save off by default, with the user able to switch it on if they want? The JSON serialisation is quite intensive since the tree is deep in places. I don't see how the process could be further optimised (other than refactoring and optimising the savegame data itself, which should be done at some point, but probably not right now). A binary flat file will always be much quicker to save and load.

nick-the-ninja:
Interesting... saves are faster for me now. But in debug mode, I think the loads were slower.

fluffyfreak:
Yeah I'll have to check some more another night it's gotten late now, but I do have some profiling setup and results.

nick-the-ninja:
Sorry about that comment (now deleted) - I got my logic wrong!
Saves are faster than loads for me.
Now: save < 0.5 sec, load = 1 sec
in debug, save was at least a minute, load was at least 2 minutes

nick-the-ninja:
OK. Thanks for taking a look :)

nick-the-ninja:
My saves and loads are never very far into the game. I don't know if that's why they are fast. At best, I've made say 5 hyperspace jumps, traded 2 or 3 items, taken on and delivered on a couple of contracts, bought a missile, explored several planets.
Thanks for setting up the profiling :)

walterar:
I'm trying to build for win32 on my cross compiling environment (MXE) and I got this error:
[...]
Dist/contrib/json»
  CXX    jsoncpp.o
  CXX    JsonUtils.o
In file included from ../../src/utils.h:11:0,
                 from JsonUtils.cpp:5:
../../src/libs.h: In function 'double DEG2RAD(double)':
../../src/libs.h:83:45: error: 'M_PI' was not declared in this scope
 inline double DEG2RAD(double x) { return x*(M_PI/180.); }
                                             ^
../../src/libs.h: In function 'float DEG2RAD(float)':
../../src/libs.h:84:51: error: 'M_PI' was not declared in this scope
 inline float  DEG2RAD(float  x) { return x*(float(M_PI)/180.f); }
                                                   ^
../../src/libs.h: In function 'double RAD2DEG(double)':
../../src/libs.h:85:50: error: 'M_PI' was not declared in this scope
 inline double RAD2DEG(double x) { return x*(180./M_PI); }
                                                  ^
../../src/libs.h: In function 'float RAD2DEG(float)':
../../src/libs.h:86:57: error: 'M_PI' was not declared in this scope
 inline float  RAD2DEG(float  x) { return x*(180.f/float(M_PI)); }
                                                         ^
make[2]: *** [JsonUtils.o] Error 1
[...]

In everything else works great. On my machine I have no differences load / save with the previous system "no json". I have not measured, but the "feel" is no difference.

fluffyfreak:
I wonder if this is actually a false issue caused by the [Visual Studio Debug Heap](http://preshing.com/20110717/the-windows-heap-is-slow-when-launched-from-the-debugger/), I will have to check it out tomorrow night.

walterar:
nick-the-ninja, if you did not know, Pioneer releases for windows are made by @robn, in an environment of cross compilation on Linux, using MXE. It is therefore very important that any work, compile seamlessly in this environment.

Andy, to "simple view", you can think how to solve this error?

fluffyfreak:
@walterar not right now, might just be a simple build/configuration mistake. Will only take a short while to understand and fix I think.

fluffyfreak:
That is immensely frustrating but good to have discovered. It runs a tiny bit slower than the old code but  more than fast enough, you just have to make sure that `_NO_DEBUG_HEAP=1` is set in the VS2013 Debugging Environment option :(

I've been using this thing for years and didn't know about that, or had forgotten!

fluffyfreak:
Oh wait, `M_PI` is a Microsoft thing. We have our own definition of PI for Pioneer.

walterar:
M_PI is everywhere. Where to you suggest applying this exactly?

walterar:
Given that this error only happens in cross-compilation environment.

fluffyfreak:
As I said, directly above that code which won't compile :)
Only as a temporary fix since I can't run MXE right now.

nick-the-ninja:
Will have a look later this evening, and add the definition of M_PI to JsonUtils.cpp if that's what is needed.

walterar:
Applying the definition, compile, but with -fpermissive. The times to save/load are gone to hell.

walterar:
nick-the-ninja, now we have a conflict with the newly merged #3361 :(

nick-the-ninja:
I'm curious to see what happens when I pull upstream master to JSONSerialise :)
I expect the conflict(s) will open a VIM editor console to resolve them.

A couple of questions:
(1) Suppose I mess up the conflict resolution and exit VIM - will the merge be partially done? Or will the whole merge be cancelled?
(2) Should I set up kdiff3 as merging tool? (executed by the command below)

git config --global merge.tool kdiff3
git config --global mergetool.kdiff3.cmd '"C:\\Program Files (x86)\\KDiff3\\kdiff3" $BASE $LOCAL $REMOTE -o $MERGED'

impaktor:
@nick-the-ninja you can always abort a merge. I think it might be `git merge --abort` or something like that. It will tell you, or at least my git does so.

impaktor:
...and if not, just do a git reset to the latest commit of the branch, like HEAD or 673a950

nick-the-ninja:
Done "git pull upstream master" and it appears to be successful. Doesn't say anything about a conflict.
"git status" reports nothing to commit, working directory clean.
Should I execute "git mergetool"?

nick-the-ninja:
I'll have a look at the modified files and see if the automerge has caused any corruption.

nick-the-ninja:
I don't believe there was a conflict. Will update the PR.

walterar:
These are the files that have changed in master and require apply your new job.
https://github.com/pioneerspacesim/pioneer/pull/3361/files

nick-the-ninja:
Automerge has handled it :) None of the save and load functions have been modified.

walterar:
That's magic! I like to do it manually, I'm a masochist. :))

Compiles fine in Linux

nick-the-ninja:
Excellent!
I admit, I was a tiny bit disappointed that there was no conflict to resolve :)

walterar:
This solves PR # 3345
If you're looking for excitement, answers this question: why? :))

nick-the-ninja:
Got a lot on at work at the mo, need to conserve my energy ;)

fluffyfreak:
Ok so I didn't get to look at this last weekend.
I've tested it out in full debug under Visual Studio 2013 loading and saving without any asserts or anything obviously wrong.

@impaktor, @robn, @johnbartholomew any objections to me merging this? I don't want it to become a many-month PR that sits with no-one reviewing it.

johnbartholomew:
I have no opinion. I am very unlikely to have time to look at this closely (unfortunately).

robn:
This isn't the way that I wanted to do JSON savefiles, but I haven't moved on that branch for almost two years, and I'm not going to get chance to review this any time soon. If I feel suitably motivated I can always rip it all out later ;)

So yeah, no opinion from me.

impaktor:
I have no opinion on this.

laarmen:
I would love to have an opinion, but I'd need time for it.

Glancing at the diff and commit logs, something bothers me but I can't
quite put my finger on it so I'll just shut up and say "I told you so"
whenever this blows up in our face :-)

(well, there is the whole github account mixup that's a bit weird, but
that's no deal breaker)

(@nick-the-ninja or @infosecgeek or whatever, don't worry about breaking
stuff, I'm probably still the title holder for the "Let's merge and
break the world" award).

Fluffy, go for it!

fluffyfreak:
@laarmen I thought we were tied for project wrecking ;) challenge accepted!

fluffyfreak:
@nick-the-ninja ok merged and congrats on getting this done!
Sorry for taking so long to get it reviewed :)

laarmen:
Quoting Andrew Copland (2015-03-16 19:45:05)
> @laarmen I thought we were tied for project wrecking ;) challenge accepted!

Well, depends on how you calculate the score: I'm fairly sure #1719 is
the one that brought the most havoc, but if we add all the (more or less)
small game wrecking changes you blessfully made to this game over the
years, you sir are, without a doubt, the clear winner.

/me tips his hat.

laarmen:
By the way, you and I have a very different notion of "long time" WRT to
reviews. It hasn't been a month yet ! ;-)

nick-the-ninja:
Hey that's great! Thanks guys :)
BTW, I have no idea who infosecgeek is. I figured it was some sort of default non-human user account that 'does' the commits.
This project has been interesting and a challenge. I've learned JSON, some GIT, and had my first experience of open-source programming.
I'm confident in the code I've written. I would like to see PersistSystemData save an array of "dict" items though (see pioneerspacesim.net/forum/viewtopic.php?f=3&t=263&start=71). It should work fine. But if there is a problem - it will be an easy fix to put right :)
Two things that ought to be addressed at some point:
(1) Lua serialisation would be much nicer if factored into JSON chunks, and the mystery of the non-printable character solved so that it can be written as a human readable string.
(2) Hexfloats - would be great to get these working.
Anyone familiar with the old serialisation will be right at home with the new code. In fact, precisely the same data is saved and loaded - so what could possibly go wrong? :D
Happy serialisation!

robn:
> (1) Lua serialisation would be much nicer if factored into JSON chunks, and the mystery of the non-printable character solved so that it can be written as a human readable string.
> (2) Hexfloats - would be great to get these working.

Both solved problems in stuff I have in branches. But we'll let this settle for a while first :)

This is going to break existing saves, of course. Anyone care about timing on a build or shall I just do it now.

nick-the-ninja:
Now is fine with me.

impaktor:
I'de like to point out that we have stuff in `SAVEBUMP.txt` file, that should be added if this bumps save game.

@nick-the-ninja the pioneer dev-philosophy is: "if you breake it - you fix it". Thus, keep us in mind over the comming weeks if bug reports start comming in.

@nick-the-ninja, please let us know what name/alias you would you like to use in the AUTHORS.txt, and I (or someone else) will add you.

> I would like to see PersistSystemData save an array of "dict" items though

I think @lwho was the last person to do stuff on that, but I might be wrong. Perhaps he has input on the matter.

nick-the-ninja:
No problem - I'll be keeping an eye on things.
Nick Wanstall in AUTHORS.TXT please.

fluffyfreak:
@robn I would give it the usual couple of days for issues to be noticed with a PR this big.

fluffyfreak:
@nick-the-ninja I have [just the challenge](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=318&p=3831#p3831) for your next project if you like :+1:

impaktor:
@nick-the-ninja added. 5df0e8733f22571c3c0ec17ffa843766eca2f1d4

fluffyfreak:
@robn @impaktor I am doing the SAVEBUMP stuff right now, will have a PR to merge for it asap.

laarmen:
@impaktor: Really? Here I thought our philosophy was "If you break it,
don't worry, @impaktor will either fix it or find someone who will" :-)

nick-the-ninja:
@fluffyfreak I've been thinking about this too! (next project)

PR #3361, opened 2015-03-08
Mouselook for internal camera

dvolk:
This add mouselook for the internal camera. Intended to be used with the cockpit, but works without too.

Doesn't have any rotational bounds so the cockpit model will probably need to be updated.

Could use more testing too.

### DIFF
--- a/src/CameraController.cpp
+++ b/src/CameraController.cpp
@@ -34,8 +34,14 @@ void CameraController::Update()
 
 
 InternalCameraController::InternalCameraController(RefCountedPtr<CameraContext> camera, const Ship *ship) :
-	CameraController(camera, ship),
-	m_mode(MODE_FRONT)
+	MoveableCameraController(camera, ship),
+	m_mode(MODE_FRONT),
+	m_rotX(0),
+	m_rotY(0),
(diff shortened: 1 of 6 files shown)


fluffyfreak:
@dvolk excellent! Thanks :)

fluffyfreak:
@dvolk this works great so it can be merged but one thing that might improve it is a reset function so you can press a key like `5` on the numpad and it will reset to the default position.

dvolk:
Good point. 

I removed an if that prevented the camera mode resetting if the mode is already set (since that previously did nothing). So now you can be in the front camera, look around, and then 'switch' to the front camera to reset the facing.

nozmajner:
Forgot to say: yay! and thanks :)  I'll update the default cockpit as soon as I can.
The no limit you said means you can turn around all the way, right? Maybe a limit would be useful, especially if it's exposed in the config file or somewhere, to allow some adjusting, so I can experiment out a limit that feels natural.

fluffyfreak:
I think this can be merged as it is right now. An improved cockpit model and limits could be added.

This makes a nicer interface to attach VR support too as well for the future.

Issue #3364, opened 2015-03-17
Game freeze


radius75:
VSE2013 Release build 7901020258ca17ec8ee4fcce2d67da7535d5d7c3
- Run Pioneer
- Start at Earth
- F5 = freeze for long time (hear the ship engine)

If the game will move, it again freezes (after EXIT or SAVE)

Edit
Build 4becd2e871773c0196311e205cfe909bc0677710 works fine for me

Edit2
the problem starts here bf39e997590239d16612cdaf1e6a9ad503ee435d

impaktor:
@nick-the-ninja just updated the save format. Are you sure it is a 100% freeze and not just a longer (auto)save time?

radius75:
Perhaps it autosave. But it takes about 80 seconds.

fluffyfreak:
Relax, there's nothing wrong with the code, it's running the game via Visual Studios Debugger [that is the problem](http://preshing.com/20110717/the-windows-heap-is-slow-when-launched-from-the-debugger/).

It's a little known feature of the windows memory allocator that they're actually fixing for VS2015 because it makes debugging slower and no-one knows how to use it anyway.

Go into the project settings for pioneer, open the 'Debugging' panel, and in the 'Environment' options add '_NO_DEBUG_HEAP=1' for all configurations.

Now it will run normally again.

I hope to add this somehow by default but visual studio makes it a per-user setting currently :(

laarmen:
Mmmmh, that would explain a few performance issues I've had at work
today, actually ! Thanks for the tip, @fluffyfreak.

fluffyfreak:
@laarmen yeah it can really hammer STL containers when debugging in vs2013 :(

fluffyfreak:
Actually integrating dlmalloc into pioneer with vs2013... has proven to be a screaming pit of madness.
Once I got it as far as trying to link, and that took an hour, things went really wrong.

I might try again at another time.

impaktor:
So this "Game freeze" issue is solved, by disabling a debug flag in Microsoft visual studio, right?
Close?

fluffyfreak:
It kind of is, but I want to keep this open a while longer, I'll assign it to me as well.
There shouldn't be a freeze but there's no way to distribute a fix for it because of the way MS have implemented it so I've looked into alternative memory allocators but they have issues with visual studio, linking, dlls etc...

So it's ongoing.

fluffyfreak:
[...]

PR #3372, opened 2015-03-24
Instancing

fluffyfreak:
# Overview - aka: "oh goodness what has Andy done now!?!?!"
I add some more information gathering recently (#3371) which revealed that we have a very large number of draw calls. Now this isn't as bad for an OpenGL game like ours as for a DirectX 9 game but there are situations where we are pushing over 5000 per frame.

This is especially crazy given that in that particular case we have a city in view which is made up of just 14 types of building repeated a hundred or so times each. 

Much better would be if we could just traverse the scenegraph once and render each building type with a single draw call!

## Instancing:
Welcome to [instancing](http://www.learnopengl.com/#!Advanced-OpenGL/Instancing)!

This PR makes use of some OpenGL 3.3 features and has required me to get a bit hacky with defining the features available from glLoadGen.

It still requests an OpenGL 3.1 context, I have tested it on my Linux where it works but there's still a chance that something will trip it up despite these being pretty common extensions so if everyone can test on their home machine that would be great.

# Results:
Well... it's a bit quicker, with Vsync off I get ~108fps versus ~90fps, which is hardly worth it. However the only thing that I have instanced so far are the city buildings. On Linux this seemed to make a bigger difference but I wasn't actively trying to get a benchmark on that.

Mostly this is just a cleaner, less CPU intense, way of doing things and I'll extend it to other areas in the future.

Andy


laarmen:
I saw the title, and thought "Oh, goodies, some more fluffy awsomeness".

As usual, I have no clue about all that stuff, but still, :+1:

impaktor:
what @laarmen said. 

> This PR makes use of some OpenGL 3.3 feature

So this moves us from 3.1 to 3.3? Or was the "hacky" part that it will still run on 3.1?

laarmen:
As I understand it, OpenGL compatibility thingies from the vendors are pretty conservatives. Not allowing a 3.2 context doesn't mean that you don't have most 3.2 features, only that there is a least one that you don't support. And I'm guessing @fluffyfreak is using features that are common, even though implemented as extensions.

I'll give it a try tonight on my 3.1-only laptop.

fluffyfreak:
Pioneer is currently using 3.1, but to implement this I had to use 3.3 whilst I developed it, then I brought the number back down to 3.1 again... so that things would still work I had to get a little hacky.

So the glLoadGen created wrapper is for OpenGL 3.3 + a bunch of extensions, and it will hook up any and all functions that it can find from the 3.3 spec. The actual context requested from the driver is 3.1 CORE and will depend on what we're given.

As I say I've tested this on a few things and it seems ok. The actual hardware requirements for these extensions are really low, like, an old nVidia 8800 kind of low! So they might even work on Intel... ;)

laarmen:
Damn you... Now I'm reading the Instancing thingy instead of being productive at work. Damn you!

impaktor:
> Damn you... Now I'm reading the Instancing thingy instead of being productive at work. Damn you!

pota-damn-you-all-to-hell

src/CityOnPlanet.cpp
```diff
@@ -23,41 +22,21 @@ static const unsigned int DEFAULT_NUM_BUILDINGS = 1000;
...
-	RefCountedPtr<CollMesh> collMesh;
-};
-
-struct citybuildinglist_t {
-	const char *modelTagName;
-	double minRadius, maxRadius;
-	unsigned int numBuildings;
-	citybuilding_t *buildings;
-};
+bool CityOnPlanet::s_cityBuildingsInitted = false;
```
laarmen: Initted, seriously? oO

fluffyfreak: Don't blame me I just copied n' pasted that part :)

fluffyfreak:
Tested on Windows Intel GPU and it works too.

fluffyfreak:
Well I am impressed, it seems to work on OSX (`Mac Mini with Mavericks`) too with an Intel HD3000 :)

laarmen:
And on my Intel laptop as well.

However, I've just noticed that Mesa has been upgraded, and now it adverts itself as 3.3-compliant in glxinfo, so...

fluffyfreak:
Off-topic: It does!!! Holy cow at long last! Hey can you also try with the eclipse rendering enabled?

laarmen:
It was never disabled AFAICT (in my config.ini, at least)...

laarmen:
Yep, I checked on Epsilon Eridani, eclipse works.

nozmajner:
Nice.
Do you imagine making this work for the orbitals in the future too?
I was thinking about putting buildings in the residential area for example, but that was quite tri-heavy. In the area of millions. :D 
And the station internals use a lot of reusable stuff, so it might be useful there too.

fluffyfreak:
@nozmajner there's probably a bunch of things that it could be used for, anything where a lot of duplicates are present.

fluffyfreak:
@eloquentmess it would be good to get another pair of eyes on this for OSX, and also if you could confirm that OSX suffers from the #3364 stall when taking off and saving a game etc.

walterar:
This does not work with MMX. Compiles fine but the buildings simply disappear. The problem reported in 3364, continues.

fluffyfreak:
@walterar MMX?

walterar:
MXE

fluffyfreak:
@walterar managed to build an MXE version myself, it works fine.

walterar:
@fluffyfreak  I get the same result with this: make -j 4 'CXXFLAGS + = -O2 -std=gnu++11'

I think -std=c++11 should not be in json/Makefile.am Something wrong with configure.ac

The buildings remain invisible.

fluffyfreak:
@walterar I'm not doing any of that I just do:
> make clean
> ./configure --with-mxe=/opt/mxe
> make -j4

and it works fine. @robn you got any opinion on this?

fluffyfreak:
@walterar can you try building MXE Pioneer in the way that I do? 
What happens if you do that?
Also can you see any error messages in the `output.txt`? 
Can you upload the `opengl.txt` to somewhere, forum perhaps, so that I can take a look at it?

I would like to get this merged today if at all possible or it will be another week before I really get the chance too again.

walterar:
@fluffyfreak  My English is null, and this that will not change. It is desirable that you make an effort to understand me, as I do make, to  understand to you.

I built this in every possible way, even with different libraries, with the same result.

I can send you everything you want, even different executables.

I am working with Ubuntu 12.04, and the latest version of mxe with this configuration:

--mxe--
gcc 4.9.2
freetype 2.5.5
2.0.3 sdl2
2.0.0 sdl2_image
vorbis 1.3.5
libsigc ++ 2.4.0
curl 7.41.0
assimp 3.0.1270
--
assimp 3.1.1 has some problems on other things, but in this case, worked with the same results.

fluffyfreak:
@walterar I need the `opengl.txt` because it will tell me if your GPU supports the OpenGL extensions that I am using for this patch. I would expect it to crash if it didn't but since it runs it is doing something just not what I expect.

The `output.txt` is useful just to see if it's put error messages or warning in there whilst running.

Also are you doing this on a fork of `Pioneer` or on your own `Pioneer Scout +` game? Because I need to know if the problem is in `Pioneer` or if something has gone wrong when merging the code into your own game.

How did you get that listing of the libraries used by MXE? I don't use LInux regularly and I can't see how you'd get it.

walterar:
https://dl.dropboxusercontent.com/u/15391580/Pioneer/opengl.txt
https://dl.dropboxusercontent.com/u/15391580/Pioneer/output.txt
https://dl.dropboxusercontent.com/u/15391580/Pioneer/config.ini

screenshot-20150329-160305

https://dl.dropboxusercontent.com/u/15391580/Pioneer/pioneer.exe
https://dl.dropboxusercontent.com/u/15391580/Pioneer/modelcompiler.exe

Replace the libraries is somewhat complicated, but here I command you need.
https://dl.dropboxusercontent.com/u/15391580/Pioneer/assimp.mk
https://dl.dropboxusercontent.com/u/15391580/Pioneer/assimp--3.0.1270-source-only.zip

 Here are more info http://mxe.cc/#creating-packages

fluffyfreak:
@walterar ok thankyou, I will hold off merging this until we've figured out why it doesn't work for you.

walterar:
git clone git://github.com/mxe/mxe.git
cd mxe
make gcc freetype sdl2 sdl2_image vorbis libsigc++ curl assimp

Should suffice. If you want to change the version of assimp need him I sent you before, but that's more difficult for me to explain in English.

walterar:
@fluffyfreak 
I understood that you wanted to know how to change the versions of libraries mxe. I misunderstood. I hope this helps someone to compile win32 versions on Linux. Because until now, it works only for you. I would like to know why.

To find library versions you only have to look in the folder mxe/pkg

walterar:
I just made a new building with a new master cloned and now it works.
screenshot-20150330-001339

Obviously there was something that you will see here: https://dl.dropboxusercontent.com/u/15391580/Pioneer/output2.txt

Compare to the other:
https://dl.dropboxusercontent.com/u/15391580/Pioneer/output.txt

I replaced the executables if anyone wants to try.
https://dl.dropboxusercontent.com/u/15391580/Pioneer/pioneer.exe
https://dl.dropboxusercontent.com/u/15391580/Pioneer/modelcompiler.exe

fluffyfreak:
So now it works?

fluffyfreak:
@walterar that new exe does not work for me. You're still building with your own custom command line and not the one used by the build machine correct?

@robn I could really do with knowing if the standard MXE build will actually produce a workable build. This works on every other platform tested so far, several flavours of Linux, 2 OSX machines and Windows vs2013 & the MXE that I built myself. However it doesn't work for @walterar when he builds.

robn:
I'm doing a build now. I'll report back shortly.

fluffyfreak:
@robn cheers dude :) I can't find the commit `ff5b0c5` anywhere in any commit history either, I'm starting to feel like I'm going crazy.

robn:
First mention of `ff5b0c5` in this PR. Is that significant?

fluffyfreak:
It's the build that @walterar seems to have done. 
His log has it like in the `output2.log`
> ver git (ff5b0c5) on: Windows

But I can't find the branch / commit that it refers too so I have no idea what he has built against :(

fluffyfreak:
Also his MXE exe is 15.5MB whereas mine was 135MB, so I presume that he is building with different parameters than the default.

robn:
Builds and runs in MXE (under Wine, but that's always been fine).

```
robn@starmoth:~/code/pioneer-win32$ ls -lh src/pioneer.exe 
-rwxr-xr-x 1 robn robn 132M Mar 31 07:35 src/pioneer.exe
```

snap

Assuming that the presence of many buildings means its working, then I'm satisfied that there's no problem here with stock Pioneer and stock MXE - it will build a viable Windows release.

:+1: for merge.

fluffyfreak:
Thanks goodness. MERGING!!! FIRE IN THE HOLE!!! ;)

walterar:
@fluffyfreak What I was trying to you understand, is that .sgm files not work with this PR

walterar:
VertexBufferGL.cpp:378: Graphics::OGL::InstanceBuffer::InstanceBuffer(Uint32, Graphics::BufferUsage): Assertion `size > 0' failed.

This often occurs when loading a saved game

fluffyfreak:
@walterar tested this and cannot reproduce it. Also this thread is closed so no-ones sees your posts in here! Better to open a new issue since it is merged.

PR #3373, opened 2015-03-26
Fix stderr redirection

AMDmi3:
You cannot modify dereferenced FILE, this is UB. FILE may also be
opaque type so this code may not compile (and in fact it doesn't
compile on DragonFlyBSD). freopen is quite enough to do the thing.

### DIFF
--- a/src/posix/OSPosix.cpp
+++ b/src/posix/OSPosix.cpp
@@ -41,13 +41,9 @@ void RedirectStdio()
 
 	f = freopen(output_path.c_str(), "w", stderr);
 	if (!f)
-		f = fopen(output_path.c_str(), "w");
-	if (!f)
 		Output("ERROR: Couldn't redirect output to '%s': %s\n", output_path.c_str(), strerror(errno));
-	else {
+	else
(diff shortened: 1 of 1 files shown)


robn:
freopen has been seen to fail (though I don't remember which platform), which is why the fallback us there. Unfortunately I don't remember the details and can't find them anywhere, so I don't remember why.

I'm happy to switch to just freopen for now, and if it causes problems somewhere then we can bring the fallback path back wrapped in some defines.

Does freopen preserve buffering modes? I can't find anything conclusive. If not, then we should probably still set line-buffering explicitly.

AMDmi3:
> Does freopen preserve buffering modes?

Can't find explicit statements regarding this in the standard, but my guess is that it doesn't (at least unless first argument is NULL) - it closes the old stream and opens the new one just like fopen will do.

I've actually missed that setvbuf is called even if freopen succeeds, redoing the patch.

johnbartholomew:
I'm not sure why this wasn't merged back in March, but it sounds from robn's comment like he intended to merge it, so I've done that.

PR #3374, opened 2015-03-30
Add confirmation for quitting the game

impaktor:
Add confirmation for quitting the game, in 17 flavours.

A little homage to great game devs of the 90s like ID Software and Apogee (Doom 1-2, Spear of Destiny, Wolfenstein 3D), by having random confirmation questions. It can be disabled, which is perhaps useful when testing and developing.

2015-03-30-210141_859x528_scrot

This probably isn't the right place to put this option, please advice:

2015-03-30-210229_377x128_scrot

### DIFF
--- a/data/lang/quitconfirmation-core/en.json
+++ b/data/lang/quitconfirmation-core/en.json
@@ -1,0 +1,82 @@
+{
+   "NO" : {
+      "description" : "No, don't quit program",
+      "message" : "No"
+   },
+   "YES" : {
+      "description" : "Yes, quit program",
+      "message" : "Yes"
+   },
+   "QUIT" : {
+      "description" : "Quit game, as a question",
+      "message" : "Quit?"
+   },
+   "MSG_1" : {
+      "description" : "",
+      "message" : "Say {yes} to get out of here and go back to your boring programs?"
+   },
+   "MSG_2" : {
+      "description" : "",
+      "message" : "Don't go now, there's a pirate waiting for you in the other program."
+   },
+   "MSG_3" : {
+      "description" : "",
+      "message" : "Are you sure you want to quit this great game that cost you nothing?"
+   },
+   "MSG_4" : {
+      "description" : "",
+      "message" : "Doest thou wish to leave with such hasty abandon?"
+   },
+   "MSG_5" : {
+      "description" : "(Inspired by Wolfenstein 3D)",
+      "message" : "Choose {no} for guns and glory, {yes} for work and worry."
+   },
+   "MSG_6" : {
+      "description" : "",
+      "message" : "Click {no} to be brave, click {yes} to cower in shame."
+   },
+   "MSG_7" : {
+      "description" : "Doom 1 anyone? Although it said DOS instead of Command line",
+      "message" : "I wouldn't leave if I were you. The command line is much worse."
+   },
+   "MSG_8" : {
+      "description" : "",
+      "message" : "I'm thinking you want to click {no} to play more. You do it."
+   },
+   "MSG_9" : {
+      "description" : "",
+      "message" : "Clicking {yes} means you must return to the humdrum workday world."
+   },
+   "MSG_10" : {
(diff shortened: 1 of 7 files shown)


fluffyfreak:
Yes, I want this in the game!

robn:
Cute :shipit:

If the option is only useful for testing and dev, then make it a config setting only. Otherwise, don't worry about the placement. The settings need a bit of information-architecture brains thrown at it, until then chucking everything in one basket is probably fine.

data/lang/quitconfirmation-core/en.json
```diff
@@ -0,0 +1,82 @@
...
+      "description" : "No, don't quit program",
+      "message" : "No"
+   },
+   "MSG_1" : {
+      "description" : "",
+      "message" : "Get out of here and go back to your boring programs?"
+   },
+   "MSG_2" : {
+      "description" : "",
+      "message" : "Don't go now, there's a Pirate waiting for you in the other program."
```
robn: Lowercase P on pirate.

data/lang/quitconfirmation-core/en.json
```diff
@@ -0,0 +1,82 @@
...
+      "description" : "",
+      "message" : "Are you sure you want to quit this great game that cost you nothing?"
+   },
+   "MSG_4" : {
+      "description" : "",
+      "message" : "Doest thou wish to leave with such hasty abandon?"
+   },
+   "MSG_5" : {
+      "description" : "(Inspired by Wolfenstein 3D)",
+      "message" : "Choose {no} for guns and glory, {yes} for work and worry."
```
robn: NO and YES better?

data/lang/quitconfirmation-core/en.json
```diff
@@ -0,0 +1,82 @@
...
+      "description" : "",
+      "message" : "Please choose {no} to waste a few more productive hours. Please?"
+   },
+   "MSG_11" : {
+      "description" : "",
+      "message" : "Explorers don't quit, but go ahead and click {yes} if you aren't one."
+   },
+   "MSG_12" : {
+      "description" : "",
+      "message" : "Choose {no} If you are a though bounty hunter, if not, click {yes} daintily."
```
robn: thorough?

robn:
Doesn't work for if you quit by closing the window, of course. I think that's fine - its quite awkward to do. Just saying :)

data/lang/quitconfirmation-core/en.json
impaktor: Fixed.
Thanks.
Will push later.

data/lang/quitconfirmation-core/en.json
impaktor: Ironically, I'm thinking both yes and no.

Uppercase does have its merits, but in other missions we have `{price}`, not `{PRICE}`, so just trying to be consistent.

If that was what you meant?

data/lang/quitconfirmation-core/en.json
impaktor: Doh! should be "Tough". 
Thanks. 
Will fix.
(Hacked it all up today in "the between time" at work.)

data/lang/quitconfirmation-core/en.json
robn: Oh, these are interpolated values. I withdraw my stupid comment.

impaktor:
Updated and pushed the fixes.

> If the option is only useful for testing and dev, then make it a config setting only.

I was just speculating as to who/why people might object to this, so I made it optional. I'll leave it in for now, as you say, it is easy to move or change it later.

I do point out that this continues the (new) tradition of not pouring everything into `ui-core` or `core`, but put it in a new `quitconfirmation-core` module. Perhaps not the best name? Perhaps have `ui-` somewhere in the name as well?

Anyway I'll add the module to transifex when merging.

walterar:
I like this :+1:

PR #3375, opened 2015-03-31
Update the sgm file version

fluffyfreak:
Due to recent changes SGM files will need to be rebuilt.
Also a single small optimisation to avoid updating the transforms if a local matrix is the identity matrix.


fluffyfreak:
Build timed out rather than failing, built ok on my own Linux box. Merging.

PR #3381, opened 2015-04-17
Fix #3377 - broken cockpit rotations using headlook.

### LINKED ISSUE #3377: Headlook rotates the cockpit model oddly
I noticed that the headlook rotates the cockpit model oddly in the opposing direction. I was able to do an almost 180 turn to it by rotating the view around about three times. Happens on both horizontal and vertical rotation.
@dvolk might be interested in this. :)

A gif about it:
https://lh6.googleusercontent.com/-zNEH0s3S48A/VR1LRcmvR9I/AAAAAAAADxY/mB8AQ4UuALI/s2048/mouselook_cockpitrot.gif)

fluffyfreak:
This fixes #3377 by removing the "cached" values and just converting the actual values using 'DEG2RAD' that are correctly calculated.

I also noticed that some bad formatting had crept in so just blanket reformatted the files involved.

### DIFF
--- a/src/CameraController.cpp
+++ b/src/CameraController.cpp
@@ -9,10 +9,10 @@
 #include "json/JsonUtils.h"
 
 CameraController::CameraController(RefCountedPtr<CameraContext> camera, const Ship *ship) :
-	m_camera(camera),
-	m_ship(ship),
-	m_pos(0.0),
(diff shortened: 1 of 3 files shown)


nozmajner:
Neat! :+1:

PR #3382, opened 2015-04-22
Fix typo in language strings (possibly WIP)

impaktor:
Turkish translator found this typo.

I open this PR to collect any/all typos in the already translated strings. I don't want to merge this at the moment, since it might not be of enough severity to require re-translating all existing versions of this string, which are probably correct to begin with.

Also, maybe rob can do some magic to alleviate that, once he attacks #3324

### DIFF
--- a/data/lang/module-deliverpackage/en.json
+++ b/data/lang/module-deliverpackage/en.json
@@ -281,7 +281,7 @@
    },
    "PIRATE_TAUNTS_7" : {
       "description" : "",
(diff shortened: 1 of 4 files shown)


impaktor:
Technically, I this PR was waiting for robn-transifex magic to not need re-translating these strings, but nevermind.

fluffyfreak:
It had waited since April 22nd, magic be damned.

PR #3383, opened 2015-04-22
Lua sanity and bug fixes

impaktor:
Some minor bugs fixed. 
Some sanity instilled.


walterar:
You're using an old version of Equipment.lua?

impaktor:
I did this from my old master (20141118), so yes, files are old, but only the changes will be merged, so it should not be a problem. I just did a test merge into current master, to make sure, and it worked fine.

walterar:
:+1:

walterar:
I think Game.player: SetMoney (100) needs local Player = import ("Player") in MainMenu.lua

laarmen:
I haven't looked at the code, but syntactically from what you said, Game.player only needs the global "Game", all other symbols are then defined from within the imported module.

Technically, you don't need to import the module where a method is defined to use it. It is just the most straightforward way to access it, but access may be leaked from other objects (for instance, Game leaks a bit of the Player module via its "player" object). Usually, the leak is either intended, or harmless.

impaktor:
@laarmen You're absolutely right, what was I thinking! 
"Cure gender bugs" is much better!

walterar:
Theoretically it should not happen, but it happens.

error: ui/MainMenu.lua:30: attempt to call method 'SetMoney' (a nil value)

impaktor:
I just tried running it, and yes, there are some things happening that should not, so I'll need to fix them first.

impaktor:
I reverted the changes that didn't work, thus logic suggests that it should work now.
Merged.

PR #3384, opened 2015-04-23
Cppcheck changes

fluffyfreak:
Ran CppCheck against the current master and came up with a few easy changes. Removal of a little dead code, some cleanup for 'operator=' and '++/--' prefix vs postfix operators etc.

The biggest, most numerous, changes are in 'fixed.h' which was passing a lot of things by value instead of by reference. 

I don't think anything will have much impact so this is more an attempt to get the number of messages reported by CppCheck reduced.

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -1,6 +1,9 @@
 April 2015
    * New Features
      * Confirm quit question added (#3374)
+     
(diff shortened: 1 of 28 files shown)


impaktor:
Spring cleaning, yay!

PR #3385, opened 2015-04-25
Enable vsync

fluffyfreak:
A little while back I fixed up the VSync code and set it to check the 'VSync' config option. However I didn't fix the UI handling of it so the only way to change it was by editing the options file manually.

This just puts the checkbox back into the options screen.

I haven't done a translation for it though, brains a bit too tired today to work it out so if anyone wants to tell me what I need to do then I'll do it :)

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -823,6 +823,10 @@
       "description" : "",
       "message" : "Quit"
    },
+   "QUIT_CONFIRMATION" : {
+      "description" : "For the checkbox in settings window",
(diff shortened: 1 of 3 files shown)


impaktor:
You mean: https://github.com/pioneerspacesim/pioneer/commit/b3d7363c2c7d5839cf11d91fd3dff454317e2ea0

(from https://github.com/impaktor/pioneer/tree/trans_checkbox )

fluffyfreak:
@impaktor yep that'll do it :) thanks!

PR #3389, opened 2015-05-31
Calculate the distribution correctly for placing orbitals.

### LINKED ISSUE #3388: Too many stations packed into one orbit.
There are too many space stations being packed into an orbit and it looks a bit rubbish.

They should be better spaced out and if possible in a variety of orbits.

The problem was caused by #2973 and reported here [596](http://spacesimcentral.com/ssc/topic/2116-wanted-new-screenshots/?p=53674) by Marcel.

fluffyfreak:
It seems that I got my maths a little wrong when placing multiple orbitals originally.

This now distributes them evenly around a whole orbit which is still very artificial (hah!) however it does look better than only doing it over half of the available orbit.

Should fix #3388 and I will create a new feature request for the other items listed in it.

### DIFF
--- a/src/galaxy/StarSystemGenerator.cpp
+++ b/src/galaxy/StarSystemGenerator.cpp
@@ -1573,7 +1573,7 @@ void PopulateStarSystemGenerator::PopulateAddStations(SystemBody* sbody, StarSys
 
 			sp->m_orbit.SetShapeAroundPrimary(sp->GetSemiMajorAxisAsFixed().ToDouble()*AU, sbody->GetMassAsFixed().ToDouble() * EARTH_MASS, 0.0);
 			if (NumToMake > 1) {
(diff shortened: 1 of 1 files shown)


fluffyfreak:
@impaktor and @laarmen you reckon this deserves a save game bump?
I think save games "might" still just work in most cases since any docked ships will just be reinserted in any relocated stations.

Might confuse any save games where the player was on their way using autopilot but they should be able to recover with just some maneuvering etc.

impaktor:
I'd vote no save bump. Better to have it ruin saves for some rather than all.

laarmen:
Hey, welcome back Andrew!

laarmen:
And I'm with @impaktor. We should maybe mention the potential weird
autopilot behaviour on upgrade in the changelog, but I wouldn't do any save
bump.

fluffyfreak:
@laarmen yep back from holiday :)

Ok I'll merge this now and make a note in the changelog.

PR #3394, opened 2015-06-02
please ignore I need to clean up the history

### LINKED ISSUE #3287: Planet Scanner equipment type (ideal for newcomers)

Goal of the document ?

Usage:

Proposal:
* New slot "sensors"

  * method "PauseAcquisition"

  * method "ClearAcquisition"

Bonus points:

cccpxx78:


laarmen:
You know, you can always just rebase and force-push a branch that is in a PR, github will pick up and that and update the PR. #1719 is the first example that comes to my mind. (But then again, 1719 is always on my mind when it comes to pioneer)

PR #3395, opened 2015-06-02
New bodyscanner and SensorType. fixes #3287

### LINKED ISSUE #3287: Planet Scanner equipment type (ideal for newcomers)

cccpxx78:
This implements the bodyscanner sensor for #3287

### DIFF
--- a/data/lang/equipment-core/en.json
+++ b/data/lang/equipment-core/en.json
@@ -254,5 +254,29 @@
    "UNOCCUPIED_CABIN_DESCRIPTION" : {
       "description" : "",
       "message" : "Required for transport of a single passenger"
+   },
+   "BODYSCANNER" : {
+      "description" : "",
+      "message" : "Body Scanner"
+   },
+   "BODYSCANNER_DESCRIPTION" : {
+      "description" : "",
+      "message" : "Scan nearest body for resources."
+   },
+   "STARTING_SCAN" : {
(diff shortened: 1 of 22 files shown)


cccpxx78:
after connecting it to the UI locally I noticed there are still several issues with the scanner starting and stopping

laarmen:
I guess I would have split it in two commits, one for the general SensorType and one for the specific bodyscanner instanciation, but that's mostly personal preferences.

I'll have a glance whenever I have time (which will probably not be before Friday)

cccpxx78:
Hi @laarmen I amended some fixes that I did after the ui integration (on another branch).
I think this is ready for a first review. 
There are still a few problems I know off which I marked with TODO in the code
One of them is surviving a reload, the function pointers are missing in the table for the bodyscanner.do you have an idea on how to reconstruct them?

laarmen:
Yeah, that's a tough one. Turns out my idea of using a single class for all
sensors might have come back to bite us in the rear.

You could implement this as a singleton class on top of the SensorType
class, though, and declare this new class to the serializer.
This way the class reloading would take care of your problem.

cccpxx78:
this fix is included in #3396

PR #3396, opened 2015-06-02
generic sensor array and two bodyscanners with ui. fixes #3287

### LINKED ISSUE #3287: Planet Scanner equipment type (ideal for newcomers)

cccpxx78:
still going in the right direction here?
there is also some test code in there to see it working.
this contains the other bodyscanner code too (different commit), but I will spilt out these commits later.
bodyscanner


impaktor:
Just judging from the screenshot, the progressbar looks good, I don't know about the exact placement of the scanner icon, let's ask @nozmajner about that.

nozmajner:
Nice
I think the button should be consistent with the ECM button, being "hidden" behind the Scanner, revealed by the little Rocket and Wrench button on the right.

impaktor:
Nice progress so far. I just took it for a spin.

Left to do:

- Scanner on/off button should probably be under the missile/wrench button and not the (ironically) scanner/radar button

- Scanner progress should maybe not reset if you go out of range. I imagine it resets when inactivating it, i.e. turn it on/off, but if you dip out of range the progress just stops accumulating, but maintains it's past progress made.

- @nozmajner when should the scanner progress bar be shown? I'd recon it is only visible when scanner is active? At the moment it is visible all the time.

- So if out of range, you constantly get the "out of scanrange" message:
screenshot-20150605-181445
Perhaps just send the message to the message/log window? I don't know. Or maybe this is good? It goes away if deactivating the scanner, as it should, so I'll pass this Q over to @nozmajner.


- I haven't looked at the code yet, but basically you should not have the scanner bar/button/capability if you haven't actually bought a (body)scanner, and I can't see it installed on the ship. So the body-scanner function should only be there if the ship is actually equipped with a bodyscanner.

As for the UI, it's all in C++ at the moment, so I wouldn't expect any Lua implementation, although there should be some way of doing it.

Hmm, also, is it possible now, from Lua to create a new scanner-thingy and specify the position of the bar and the button? That would be nice, I think, although I'm not 100% on how @laarmen had envisioned it.

nozmajner:
I agree, the bar should only be visible when the scanner is active in my opinion.
Also agree with the message point, it would be more consistent if any message from the scanner would go there. Although it makes sense to show the out of range text all the time when it's relevant, but then it should be positioned under, or next to the bar in my opinion. Or even better, if possible to update it on the fly, then in the text inside the bar.

impaktor:
Showing text directly on the screen is not something we've done so far, so just sending a message to the log/message field aught to be enough, I think.

Luxury idea: ...and change the color of the progress bar to red, or grey it out, or something (or blinking scanner button, or something), but that's for another day and another pull request, perhaps. I think we should focus on getting this into the game as soon as possible, so just send a message "in rage" / "out of range" or something similar to the message field should be enough for now.

cccpxx78:
@impaktor : I reworked a lot and fixed these issues:
+ fixed button position and visibility now toggles with the radar
+ fixed scanner bar reset when out of range
+ scanner gauge is now only visible when a scanner is installed, but also visible when inactive
+ used Comms.Message for the messages. but there is no out of range message yet
+ a lot of code cleanup and rework

not fixed:
- support to add more scanners from lua
- doesn't survive a reload. (functions of bodyscanner type not recreated?)
- several todos in the code

impaktor:
Quick test, yes, works like I want, with button placement, and buying of equipment. Except I think there must be a bug, since scanner progress isn't accumulating any progress anymore. last time I tested it that part worked. It just seems to be frosen. Or maybe I'm doing something wrong?

Also, is there any point in showing the scanner bar when it's turned off? I'm thinking this could be how it should work:

- Activate scanner:
 - Show scanner bar
 - scanner progress starts at 0.
 - scanner button highlighted

- If out of range, message, and scan progress halted, but not reset.

- if scanner turned off 
 - reset scanner progress to 0
 - hide scanner bar
 - scanner button normal

At least I think that is a sensible way for it to work, @nozmajner might have ideas on UI.

Also, sorry if we're giving you conflicting descriptions, I think you're very close to the way we/I want it, and thanks for staying with us.

nozmajner:
I think what @impaktor says makes sense. What would happen if the scanning is finished? Would that turn the scanner off along with a "completed" message? Or only message, full bar, and the player has to turn the scanner off?

impaktor:
> What would happen if the scanning is finished? 

I also thought about this today. I think the bar could stay, (the player would manually deactivate the scanner), and a message sent "scan comlete" or something similar.

cccpxx78:
bodyscanner_messages
@impaktor I don't know why the scanner was broke when you tested. This latest version is working for me at least, but it doesn't survive a reload yet.

I did some UI tweaks:
* scanner bar not visible when inactive
* moved icons and text defines for scanner to lua
* @nozmajner I fired all messages in the screenshot above (normally these wouldn't all be visible at once).  you would get them in that order when:
  * you activate the scanner
  * go out of range
  * come back in range
  * finish the scan
  
  I think I should add number formatting and a unit "m" maybe?

impaktor:
How about "Max scan range exceeded: xx km" instead?
For scan completed I think it just could say "Scan Compleated".
Also, we have Formating function for formating distances. in src/utils I think.

Anyway, feels like we're just polishing now, and we really should review your code, which means it's our turn to actually do something, besides just having a bunch of opinions. So I wonder how #3395 relates to this? I.e. should we (that's you @laarmen :wink: ) first review that pull request, or do you have changes here not transferred to there?

> This latest version is working for me at least, but it doesn't survive a reload yet.

(I'll give this a test tomorrow.)

> but it doesn't survive a reload yet.

I'm sure we'll get to that, there are some really clever programmers here who maybe can have a look. I'll have a look as well.

cccpxx78:
if I do this the bodyscanner survives a reload a560d344 but this is more to show what I think is missing/going wrong: the function pointers of the derived class were missing after reload.

impaktor:
I'm still experiencing the previously mentioned strangeness of progressbar not increasing.
1. Start new game at New Hope.
2. Buy Body Scanner for $1.
3. Take off, and set manual control, and speed 0 (F5), 
4. Raise landing gear F6.
5. Turn scanner ON, while probably having a low altitude

Scanner starts, but then stops after a while (possibly because I went out of range, so all OK so far). But I notice that when I go from low altitude to higher, I get the two simultaneous messages "Scanner within range", "Scan range exceeded: xxx..." (in that order). Same thing when I went from high altitude (500 m) to low (< 300 m).

Does the scanner assume it is turned on when it is out of range?

OK, tested one more time, scan worked as long as I was low enough, but going up, out of range, and then down again, it didn't restart. I think there must be some state after going out of range that it doesn't restart itself propperly once it is within range again.

This was on: 6d53bbd47555ee63c589f831ab78c625264b4000
(GNU/Linux)

cccpxx78:
> Anyway, feels like we're just polishing now, and we really should review your code, which means it's our turn to actually do something, besides just having a bunch of opinions. So I wonder how #3395 relates to this? I.e. should we (that's you @laarmen :wink: ) first review that pull request, or do you have changes here not transferred to there?

@impaktor I cleaned up the commits and split them off, but this one is based on and dependent of  #3395. So commit 68f402e8 is the one with only the changes for the UI.
The code is functional but is missing support to define more scanners from lua only, if you have the time I would appreciate the feedback.

impaktor:
My conclusion:

- It has some TODO's in the code. I don't know if those should be sorted out before merge?

- Should `l_ship_set_scanner_percent` be documented? Do we want people from outside to make use of this function? @laarmen 

- The code is not generic in the sense that after this merge, anyone can add a second, (or third) scanner-type by only using lua code. there still have to be member variables added on the c++ side for the progress of each scanner, etc. I'm fine with this for this PR:

In an ideal world pioneer is modder friendly enough to allow a mod that basically copy pastes the lua code of the scanner and makes it do something different, and on lua side specify png-icon (if not using default), and position for progress bar. But that's the "elephant gun". If we can get the current code to master, then hopefully I can make use of this code for the scanner I know I want to have.

Talk about other scanners is just hypothetical at this point, although I do see us using a mineral scanner when planets have fun mineral deposits, etc. And adding more scanners is as easy as duplicating this PR.

Anyway, this concludes my review of the GUI part.

cccpxx78:
bodyscanner_multi

@impaktor 
* I made the code more generic to support adding more scanners from lua only without any C++ changes required. To demonstrate this I added another scanner 'Fast body scanner'.
  * Something to check here is that now the MeterBar is re-created in the Update function, possible performance issue? But this way it saved the effort of bookkeeping the meterbars.
  * I also need to check who deletes this, is it the container in the ClearAll?
* l_ship_set_scanner_percent is not longer required because of the rework to make more generic.

@laarmen 
* I fixed the loading issue like you said with a BodyScannerType, this now used for two types of bodyscanners (so no singleton).
But when the scanner is active during save, it still tries to serialize a function value. I fixed this with type check for now in EquipmentType.Serialize but this could also be done in a BodyScannerType.Serialize, but I thought this would be ok to check in the baseclass?

cccpxx78:
bodyscanner_multi2

@impaktor , @laarmen I think the code is ready for review (and merge), the two known bugs are not critical in my opinion and have a workaround/fix in game (see known bugs).
This commit contains SensorType, and subtype BodyScannerType. Of BodyScannerType there are two instances with different stats: a body scanner and a fast body scanner to demonstrate how to add more sensors from lua only.
The buttons and MeterBars are in included in C++ in the WorldView as a generic loop over the equipped sensors.

last changes I did:
* I solved the scanner name conflict by making this a sensor and the body scanner is a type of sensor.
* I fixed the UI alignment

known bugs:
* stop and start the sensor in quick succession results in multiple timers running and a faster scan. But the timers all get stopped when the scan is completed and the next scan is normal again.
* a started scanner is defunct after reload, the timer is not restarted. But it can be restarted: stop-start and it works again but with progress reset.

impaktor:
Looks good, from what I can tell (I haven't looked at it line-by-line, just general glance), but @laarmen, or other person of competence will need to have a look before final merge.

I have the following thoughts:

1. What happens if a player fits two scanners of the same type? Do you predict any bugs or issues?

2. The "Bodyscanner fast" is good for testing, but I'm not sure we want it in master? Maybe have a separate commit for that one so that we can weed it out at merge time? (you don't _have_ to change it, just me thinking out loud)

3. Adding it to all the `<ships>.json` isn't necessary in my opinion, one could omit it, and rely on a default value of `1` or something. @nozmajner likes to think about ship stats so he could add them in a separate commit later. (again, I'm just bringing it up for discussion, so maybe hold on changing anything)

4. Is there some lua function that returns something (like a bool) if a scan is completed? My intention is to use this scanner in a mission script, so I would have a bool `scan_compleate_for_mission_planet`, that is switched to true if the scanner completed on the correct planet and altitude. Hmm, I'm not sure I've thought this through in detail,  how it  will all work, but here's my stab at it: I write a mission script that tells the player which planet (and what min/max?) altitude to scan at. It's pretty good if the player, when accepting a mission, knows what span of altitudes are OK. The mission script writer (me) could hack it and pull the altitude tolerance from the equipment script where it is defined, or he/I could set it from my mission script and change the default values. Thus an easier mission would have a bigger span of acceptable min/max altitude. If so, the script writer (me) needs to change the scanner's "scanner altitude tolerance" from the default values. Hm, so then the script writer (I) can check that the player started the scanner on the correct altitude and that the scan was completed. Is that the way you imagined it to be used?

There is something weird here that's bothering me. What is a logical way for the scanner to work when scanning the wrong planet vs. scanning the correct one?

Say my mission is to scan Saturn at altitude X +/- epsilon km. I start by going to Jupiter and start the scanner at some altitude. The tolerance will be som default value. It will show progress and complete, maybe it has generated say a "planet info packet" but no one is interested in Jupiter (i.e. no mission complete nor money for it), so the player goes to Saturn and starts the scanner at the correct altitude and now the tolerance is something else, because the mission difficulty was easy? Hm. what happens if I scan 50% of Jupiter, then go to Saturn and starts the scanner at altitude X? Who's job is it to zero out the scanner when leaving a frame?

Hm, quite possible number 4 here was the result of my brain crashing, it's late, and I need to think.

## End note
          
Also, I know I speak for all of us when I say that we hope you continue contributing, if you don't mind working on another pull request while this is open. Maybe you have an idea of your own, or something from the issue tracker that tickles your interest. This has been a very detailed review process since we (for the first time?!) had detailed description of what it should do. Normally things go in a lot easier, especially after a few pull requests when we know you and you know the code.

So please don't take our lengthy periods of silence as a disinterest on our part, as it is more due to us having too little time.

impaktor:
@laarmen, do you have time to look through the code, and check that it is in compliance with your grand lua equipment philosophy?

data/libs/Equipment.lua
```diff
@@ -52,7 +55,9 @@ function EquipType:Serialize()
 	local tmp = EquipType.Super().Serialize(self)
 	local ret = {}
 	for k,v in pairs(tmp) do
-		ret[k] = v
+		if type(v) ~= "function" then
```
laarmen: I'm guessing this is for the callback ? If so, I'd be a bit happier if you'd implement it in a SensorType:Serialize override, but I'll probably do it myself, your suffering has gone on long enough.

data/libs/Ship.lua
```diff
@@ -533,6 +533,48 @@ function Ship:FireMissileAt(which_missile, target)
...
+-- Availability:
+--
+--   2015 June
+--
+-- Status:
+--
+--   experimental
+--
+
+function Ship:StartSensor(idx)
```
laarmen: You could at least provide a callback ;-)

laarmen:
OK, we're going to merge this thing. However, two things:

1/ I don't know how you want to figure in AUTHORS, so I'll write you down as cccpxx78, we can always change that later on.
2/ Your code makes a lot of assumptions about the sensors being used by the player. This is fine for now, but it would be nice if a subsequent PR would make this usable by any ship (which brings the question of what happens when multiple sensors are used at the same time...)

impaktor:
@cccpxx78 :+1: 
@laarmen :+1: 
Thanks!

impaktor:
@cccpxx78 please let us know (here or on IRC) how you want to be listed in [AUTHORS.txt](https://raw.githubusercontent.com/pioneerspacesim/pioneer/master/AUTHORS.txt). You're currently just listed by your github handle in it.

cccpxx78:
Hi @impaktor, @laarmen thanks for all the patience and support! 
For the name AUTHERS.txt please put in Pieter Thielens.

> 2/ Your code makes a lot of assumptions about the sensors being used by the player. This is fine for now, but it would be nice if a subsequent PR would make this usable by any ship (which brings the question of what happens when multiple sensors are used at the same time...)

I've been looking at this issue and I see what you mean when you say these are connected. I'm thinking on how to improve this. When I have a proposal I will open a new pull request so we can discuss the solution there, let me know if I should approach this otherwise.

@laarmen I'll also include the move of the != 'function' check to a SensorType.Serialize function.

Another thing that has been bothering me is that I 'new' the MeterBar in the Update function in WorldView.cpp. I could rewrite it a bit so that that only the value is set on the MeterBar, which feels a bit more logical in an update loop (maybe also performance wise?). 
What is your opinion on this?

impaktor:
> assumptions about the sensors being used by the player.

It sure is nice to have generic code code that works for both player and NPCs, so that the player isn't treated "special" in the universe, but at least I can't see any application anytime soon on why/how NPC would use a scanner, and the player notice it.

> I 'new' the MeterBar in the Update function in `WorldView.cpp`. I could rewrite it a bit so that that only the value is set on the MeterBar, which feels a bit more logical in an update loop (maybe also performance wise?). 

I have no strong feelings one way or another, maybe @fluffyfreak has thoughts on that.

Have you looked at laarmen's code-comment regarding callback?

(Updated Authors: @a131939aceca05b54bb571d646bd4ed9bff1f5a2)

PR #3403, opened 2015-06-06
SpawnPirate() Function

prestotron:
I was chatting with impaktor on the IRC to see what could be done for missions, so s/he recommended that I make a function to spawn pirates rather than continue to copy and paste it, so I made a file in /data/libs/ called PublicFunctions.lua which has a SpawnPirate() function to spawn pirates rather than copying and pasting everytime.

(The missions Taxi, DeliverPackage and CargoRun use this function currently, so if you need an example on how to use the function, check it out)

Seems like a good idea, eh?

Also, my IRC name is prestotron, I plan on changing my GitHub name to prestotron soon as well. Just letting you all know :) (My GitHub name may have changed already)


data/libs/PublicFunctions.lua
```diff
@@ -0,0 +1,20 @@
...
+		function (k,l) return l:IsValidSlot('laser_front') and l.capabilities.mass <= max_laser_size and l.l10n_key:find("PULSECANNON") end,
+			pairs(Equipment.laser)
+		))
+	local laserdef = laserdefs[Engine.rand:Integer(1,#laserdefs)]
+
+	name = Space.SpawnShipNear(shipdef.id, Game.player, 50, 100)
+	name:SetLabel(Ship.MakeRandomLabel())
+	name:AddEquip(default_drive)
+	name:AddEquip(laserdef)
+
```
impaktor: This makes use of a number of functions that needs to be imported. Basically anything to the right of the `=` not previously defined, like Engine, Space, and utils.

Also, `shipdefs` was defined somewhere from where you copied the code.

And as I mentioned, the function needs to return the ship it has spawned (what you call `name` here). It aught to be `local` and maybe called `ship`?

There's the [API documentation](http://eatenbyagrue.org/f/pioneer/codedoc/files/LuaBody-cpp.html) for support.

data/modules/CargoRun/CargoRun.lua
```diff
@@ -20,6 +20,8 @@ local InfoFace = import("ui/InfoFace")
 
 local l = Lang.GetResource("module-cargorun")
 
+import ("libs/PublicFunctions.lua")
+
```
impaktor: `local Publicfunctions = import (PublicFunctions.lua)`
should be enough.

impaktor:
Yes, @prestotron wants to work on pioneer as a summer project, and you are most welcome! Ideas discussed were C++, Lua-mission scripting and 3D modelling of ships and buildings.

@prestotron regarding this PR, please see my comments above, and test the code. I.e. set spawning rate to some fixed number of pirates and test DeliverPackage, Taxi, etc. by jumping to the mission system. I don't think the current code would work. At least not when I run it in my mind.

prestotron:
@impaktor I'll work on it a bit tomorrow. I'll probably be in IRC for questions haha.

prestotron:
@impaktor Since you're calling the function  where something is defined, should it just work anyways? I don't see it needed to be imported inside the function, unless they're not public

data/libs/PublicFunctions.lua
prestotron: Well, I called it name because in some cases it was "pirate" and most other cases it was "ship", so I left it flexible.

laarmen:
Err, 17 commits seems a bit overkill to me. Would it be possible to squash those down to one or two commits?

laarmen:
Also, don't we already have a "utils" module in the libs folder? You could add the function there instead of creating a new file.

prestotron:
@laarmen 

1.) I have no idea how I could possibly squash commits, unless I reupload the pull. So, not sure what to do there.

2.) @impaktor said a public functions file would be nice. But if you do not like that, it can easily be moved.

impaktor:
> @impaktor said a public functions file would be nice.

Well, I just said it should go into `data/libs` somewhere:

```
<devmarth> Where would I put the functions file though? Just in /data/modules?
<impaktor> laarmen: would data/libs/<some new file>.lua be the place for new
           general purpose functions? like "SetUpAndSpawnPirate()"?
<devmarth> libs sounds actually like a good place imo                   [20:10]
```

I wasn't sure if `libs/utils.lua` was the right one, since that is mostly some low level table functions and such, but do as laarmen suggested.

> I have no idea how I could possibly squash commit

For squashing commits, either I'll do it when/if merge time comes, or you can follow my write up on the [wiki](http://pioneerwiki.com/wiki/Using_git_and_GitHub#Fixing.2FUpdating_your_pull_request) basically a `git rebase -i <sha>`, followed by a `git push -f`

laarmen:
Well, as I see it utils is where things go when they're too small and/or
too specific to deserve their own file, but we don't know where else to put
them. After a while, utils will become too big and then we'll figure out
how split off some of it. I think we've already gone through this cycle one
or twice since I've been around.

So, for the lib code, either you can figure out a good module name that
says what the content is about, or you put it in utils. No need for two
files with nondescript names :-)

data/libs/PublicFunctions.lua
impaktor: OK, so this actually works (apparently) because of indirect imports of variables, assuming they're named to what they are in the calling script, and because tables and user data are passed by reference in lua functions (unlike values, booleans, and strings, which are passed by copy).

I'd feel a lot calmer if you would import and define the `Engine`, `shipdefs`, etc. variables.

Also, I'd prefer the variable name change: `name` -> `ship`

thanks.

laarmen: For what it's worth, I wholeheartedly agree

prestotron: @laarmen Yeah, what he just said is currently being done as we speak, I'm just working stuff out right now in the IRC before I start doing it.

After that, we should be all set to go. :)

prestotron: Fixed.

Issue #3387, opened 2015-05-13
Can't load saved games in Arch-based distros


Raiz1337:
When i try to load a game in pioneer the game crashes immediately, i asked people @ the IRC they told me to post this on the issue tracker (( Note: this issue is only for Arch-based distros ))

Tichy:
I can confirm this.

Just started from Earth, heading to Uranus. Saved. Loaded. Switched to max time acceleration. In a couple of seconds the game crashed leaving these messages on console.

```
cleanTSTable:total:48,active:40,removed:0
deneb jump status is not OK
WJ-3158 docked with Enki Catena ship:deneb
pioneer: LuaTable.h:100: LuaTable::LuaTable(lua_State*, int): Assertion `(lua_type(m_lua, (m_index)) == 5)' failed.
[1]    14389 abort (core dumped)  pioneer
```

This is my savefile.

https://app.box.com/s/135nhm2kga41ek5crmjqilov95m1y2zs

Pioneer is compiled on my machine using the AUR package.
https://aur.archlinux.org/packages/pioneer-git/

impaktor:
Are you both on Arch "Stable"?
I know Arch Testing just got GCC 5.0. If that can have something to do with it?

Tichy:
On 2015-05-14 10:50, Karl F wrote:
> Are you both on Arch "Stable"?
>  I know Arch Testing just got GCC 5.0. If that can have something to
> do with it?

I'm on stable.

laarmen:
Do you compile the game or use the pre-built thingy?

Tichy:
On 2015-05-14 22:24, Simon Chopin wrote:
> Do you compile the game or use the pre-built thingy?

Pioneer is compiled on my machine using the AUR package.
https://aur.archlinux.org/packages/pioneer-git/

Which is an automated way to clone the git repo, compile it and build a 
binary package.

jmf:
I can confirm this on my arch, too.
I am compiiling from git repo manually.
Stable archlinux, Linux 4.0.1-1-ARCH #1 SMP PREEMPT x86_64
gcc 4.9.2
Lua 5.2.4

`pioneer: LuaTable.h:100: LuaTable::LuaTable(lua_State*, int): Assertion '(lua_type(m_lua, (m_index)) == 5)' failed.`
`Aborted (core dumped)`

laarmen:
[...]

PR #3407, opened 2015-06-08
Fix inconsistent types for 'imports' and 'exports' variables in TradeShips.lua

johnbartholomew:
For the problem reported in #3387.


fluffyfreak:
I was seeing this crash but it's gone now with this patch.

PR #3410, opened 2015-06-14
Allow custom stations to have realistic orbit heights

mr555ru:
Also, a Low Earth Orbit station.

### DIFF
--- a/data/systems/00_sol.lua
+++ b/data/systems/00_sol.lua
@@ -87,6 +87,9 @@ local earth_starports = {
 	CustomSystemBody:new('Los Angeles', 'STARPORT_SURFACE')
 		:latitude(math.deg2rad(34))
 		:longitude(math.deg2rad(118)),
+        CustomSystemBody:new('Low Earth Orbit', 'STARPORT_ORBITAL')
+                :semi_major_axis(f(458,10000000))
+                :rotation_period(f(2,24)),
 	CustomSystemBody:new('Gates Spaceport', 'STARPORT_ORBITAL')
 		:semi_major_axis(f(100,100000))
 		:rotation_period(f(1,24*60*3)),
--- a/data/systems/00_sol.lua
+++ b/data/systems/00_sol.lua
@@ -87,9 +87,6 @@ local earth_starports = {
 	CustomSystemBody:new('Los Angeles', 'STARPORT_SURFACE')
 		:latitude(math.deg2rad(34))
 		:longitude(math.deg2rad(118)),
-        CustomSystemBody:new('Low Earth Orbit', 'STARPORT_ORBITAL')
-                :semi_major_axis(f(458,10000000))
-                :rotation_period(f(2,24)),
 	CustomSystemBody:new('Gates Spaceport', 'STARPORT_ORBITAL')
 		:semi_major_axis(f(100,100000))
 		:rotation_period(f(1,24*60*3)),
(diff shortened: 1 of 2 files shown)


mr555ru:
I like how it looks, but there is an issue: Low orbits have small orbital periods, measuring in hours. People tried to avoid this before in code, so is be there a particular reason on that? For my edits to apply, I need to change 5b5b9d1df3c72cae5c93f59b01854feb548ac9b6 where there is "limit inner orbits to 1 <earth> day".

fluffyfreak:
Yes, orbit times measured in hours are difficult - often impossible - to reach! 

Also you are talking about hour long orbit times around big planets, but there are many small and dense planets that have orbital times measured in minutes!

The other reason is the atmosphere. Although you can get an orbit that low around a world with no atmosphere no-one would choose to put a spacestation into that orbit. Around worlds with an atmosphere the orbit might *just* be above the officially recognised limit of the atmosphere however in reality it fades slowly away for hundreds of miles. Therefore a station in an orbit that low would slow down due to drag and crash into the planet very rapidly, perhaps within just weeks of being put there.

mr555ru:
I'm playtesting it - have no significant difficulties reaching low-orbit stations with autopilot. Theres one little bug with offset at exiting station, but I couldnt reproduce it at the same savefile. It wasn't anything critical.
About flight without autopilot, I think it should be okay as long as you can see relative speed and distance at the screen. Just get close enough, make your relative speed low and your orbits would be close enough to navigate without too much of a problem. Haven't playtested that, though.

On orbits around small planets, your code at the mentioned commit is easily adjustable to limit inner orbits to at least several hours. 

About the atmosphere, I may either assume stations correct their orbits as real ones do, or if it doesn't sound reasonable, tweak the limits a little bit - say, 500-600 km for earth radius?..

fluffyfreak:
I think that limiting to at least a few hours orbit time and at least 500km above a planets atmosphere - which will vary in thickness - is a good combination of tests.

If you make those changes then I don't see why we can't have this change merged :)

impaktor:
If we imagine the ship with the slowest acceleration when fully loaded, it can easily catch up with the station?

Will this require save bump?

nozmajner:
I think catching up should be possible with all ships. The weakest ship has about 2 to 4 G forward acceleration, and if you achieved orbit, then it should be fidgeting anyway. Although in would need actual orbital rendezvous with too low thrust.

fluffyfreak:
Ok are you happy that this is ready for review?

Things are rather busy for me lately but I will try and find some time to test and give feedback.

mr555ru:
I understand that this proposal isn't on project's roadmap, so please feel free to do what you like.
I was able to reach low orbital station from surface without autopilot on Natrix by simple ballistic trajectory following relative speed markers. (As soon as I entered station frame, ballistic trajectory became station's orbit, and this left me wondering about how weird is orbital mechanics.)

impaktor:
> I understand that this proposal isn't on project's roadmap

The roadmap is old/outdated, and no one follows it anyway.
If a contribution is an improvement it's merged (when someone has the time).

fluffyfreak:
@mr555ru I think this is ok to be merged so will do so soon, what name would you like to be known as in the `AUTHORS.txt` file?

mr555ru:
Anton Korzun will be fine.

Issue #3344, opened 2015-02-05
Design error in docking sequence, orbital_station


walterar:
http://youtu.be/X_XE2xdpO5g

fluffyfreak:
Does this happen in regular Pioneer as well or is it unique to Scout+?

walterar:
It happens in my PC, not in yours?

walterar:
I'll give you a hint: The ships enter of aft in the new orbital_station.

This also happens, of course, in Scout+

Question, this happens only in my pc? Or nobody cares?

laarmen:
Wait, did you just say it only happens in Scout+? The question of whether is happend on vanilla Pioneer is crucial: we shouldn't need to install a third-party mod to do some troubleshooting. You need to come up with a way to reproduce this on plain old Pioneer.

walterar:
This ALSO happens, of course, in Scout+

walterar:
ALSO == ONLY ?

walterar:
Make an effort and look carefully.  What you see? A ship incoming? or ship leaving?
![bad-entrance](https://cloud.githubusercontent.com/assets/1334989/8184917/83064ca4-1417-11e5-9368-f3c91e0b26a8.png)

laarmen:
I meant no offense, man. It's just, you know, better safe than sorry (in the end, Andrew seem to have taken it upon himself)

walterar:
There are none so blind as those who will not see.

walterar:
To more than a year of broken, the ships keep entering backwards at orbital port. Also it removed the final sequence to docking at ports of surface.  This looks very ugly. Should we take this as new features?

fluffyfreak:
Yes, and i've been seeing it in Pioneer.
It's on my to-do list that I have to fix it.

walterar:
Thanks Andy. It had not been for your work, orbitals of nozmajner would still be waiting. But need finish, is ugly this.

fluffyfreak:
I've had a go at this over the last week and it is a nightmare to figure out.
I will keep at it but I wanted you to know that I am looking into it.

walterar:
Need help? I can recite an old poem Vogon, if necessary...

fluffyfreak:
@walterar yeah I am sorry, what was meant to take 2 or 3 days I've now spent 5 weeks working on...

...a bit longer than planned!

walterar:
[...]

PR #3414, opened 2015-06-28
Smaller orbital station

nozmajner:
I played around with the existing station parts and made a smaller orbital station using them.
It has one section of docks, with 18 bays: 1 large, 8 medium and 9 small.
The diameter is about 4km, and it spins at 0.6 rpm, providing 0.8G at the outter rim.
I made a new, smaller habitat ring for it, using the already existing textures. Same goes for the whole model, I just shifted around parts, to avoid making any new textures.
On my copy Eagle's Nest in Epsilon Eridani and the Barnard's starting point is this station.
screenshot-20150628-163546
screenshot-20150628-163648
screenshot-20150628-163751
screenshot-20150628-163805
screenshot-20150628-163904
screenshot-20150628-164006


impaktor:
nozmajner reported earlier today that the station was going crazy over the collision mesh.

@jmf's interpretation:
dxdo7qd

walterar:
Looks good. But the station entrance is backwards. Static ships will must be parked at the entrance. Besides, #3344 is still there.

nozmajner:
The above commit rotates the docking entrances so they are lined up with the static ships. For consistency's sake, because I'm not that convinced that static ships have their place only at the entrance. As in: they never enter the station, but clutter up the approach area.

Can't do anything about #3344, the autopilot still goes nuts if I rotate those markers, and tries to enter backwards from the other direction.

fluffyfreak:
Yeah don't try to correct #3344 in the model I just need to find time to fix the code I guess.

walterar:
@nozmajner It's OK now. :+1: 

big-pos-ok

"I'm not that convinced that static ships have their place only at the entrance"

For that, in the future, it would be good to change SpawnShipParked

ship = Space.SpawnShipParked (type, station)
to
ship = Space.SpawnShipParked (type, station, position)

@fluffyfreak 
"I just need to find time to fix the code I guess."

May you find time. This model will look much better then. :+1:

walterar:
@nozmajner 
You have achieved ships correctly positioned in this station. With the nose pointing toward the exit. See difference with the other stations.

impaktor:
Merged!
(rebased the last two commits, thus closing through commit message instead)

PR #3419, opened 2015-07-12
Numerous optimisations

fluffyfreak:
A collection of optimisations done to fix #3417 aka "More Text = Less Frames".

- NavLights were taking a ridiculously long time to update due to stupidity.
- Don't update the time text unless the time has changed.
- Reserve space for text before filling out the buffers.
- If the text and colour hasn't changed then don't update the buffers.
- The big one: Cache the vertex buffers created by the TextureFont and reuse them.
- Avoid rebuilding the line rendering multiple times per-frame in SectorView!

On my laptop this improves the SectorView (F2) screen from approximately 17fps with all text on show to a steady 60fps.

Andy

PR #3420, opened 2015-07-13
Sort terrain gen

fluffyfreak:
To follow #3419 (Numerous Optimisations)...

This delays submissions of the terrain generation requests in order to gather them up, sort them by distance from the camera, then submit them.

This is so that they are hopefully processed by those nearest to the camera first.


walterar:
go! go! go! :) :+1:

PR #3421, opened 2015-07-14
Remove output spamming messages.

fluffyfreak:
Spam glorious spam, it comes in a tin can AAAAAAND our debug output log...

...but not anymore.


impaktor:
For something trivial like this, I think it is OK to just push directly to master, without opening a PR. I think @robn said that was OK.

fluffyfreak:
Yeah but it's the way I'm used to working.

laarmen:
Plus, for those of us watching from afar, we can see what's going on
without having to read the git log or logging into github.

PR #3422, opened 2015-07-14
Rename bodyscanner, and set it as not purchasable for now,

impaktor:
I'll make use of #3396 soon enough*, but for now, we don't need to see it in the equipment market.

The code that was there was for testing purposes.

__EDIT:__ (*) https://github.com/impaktor/pioneer/tree/scout-mission


impaktor:
Mereged, as I wanted to beat buildbot, and transifexbot, and robn-bot (robbot?).

PR #3425, opened 2015-07-18
Fixed color of labels starting with empty strings

dodomorandi:
Labels with some empty words set their colour to black. This avoid this
behaviour, just checking if the word is empty. Only if not, the color
is updated.


fluffyfreak:
@dodomorandi nice and simple fix, if I merge it today/tomorrow what name would you like to be known as in the `AUTHORS.txt` file?

dodomorandi:
Edoardo Morandi. And thank you!

fluffyfreak:
Thank you for the patch :)

PR #3428, opened 2015-07-22
Render far away

fluffyfreak:
Keep rendering stars and planets even though they are very very far away.

I was going through some old branches to clear them out and found this one. Thought it worth fixing up and submitting for review.

Andy


impaktor:
> Keep rendering stars and planets even though they are very very far away.

Didn't @robn do some billboard light stuff, to not render things less than 4 pixels big or something?

fluffyfreak:
Yes but as people have pointed out you can still see planets and stars at hilarious distances, since they're only being drawn as billboards it's very cheap. Might as well just show them.

fluffyfreak:
Also this doesn't change @robn billboard rendering optimisations. It just keeps rendering planets & stars as billboards for longer. All other objects are culled once they get small enough.

impaktor:
OK. I have no further input.

PR #3431, opened 2015-07-25
Control how many stars in the sky

### LINKED ISSUE #3343: Too many stars in sky background
Something changed with the sky background recently. There are many more stars and they seem much smaller. The disadvantage is that it is getting too noisy and distracting (especially in dogfights, when you are trying to follow the crosshairs, etc.). Can the number of stars be reduced - or even better, the brightness and "crispness" be reduced to make it more of a background feature?

fluffyfreak:
After many complaints to fix #3343 I've implemented a slider to adjust from 0% to 100% the number of stars in the star field.

Andy

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -16,6 +16,7 @@ July 2015
      * Inactivate BodyScanner equipment testing code (#3422)
      * Drawing optimisations (#3424)
      * Render far away (#3428)
+     * Modelviewer refactor (#3429)
 
 June 2015
    * New Features
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -951,6 +951,10 @@
       "description" : "",
       "message" : "Sound"
    },
+   "STAR_FIELD_DENSITY" : {
+      "description" : "",
+      "message" : "Density of Star field"
+   },
    "START_AT_BARNARDS_STAR" : {
       "description" : "",
       "message" : "Start at Barnard's Star"
--- a/data/ui/Settings.lua
+++ b/data/ui/Settings.lua
@@ -364,19 +364,12 @@ ui.templates.Settings = function (args)
 	setTabs:AddTab({ id = "Language", title = l.LANGUAGE, icon = "Globe1",      template = wrapWithScroller(languageTemplate) })
 	setTabs:AddTab({ id = "Controls", title = l.CONTROLS, icon = "Gamepad",     template = wrapWithScroller(controlsTemplate) })
 
-	local function onCloseSettings()
-		Engine.RefreshBackground()
-	end
-
 	local close_buttons = {}
 	do
 		local items = args.closeButtons
 		for i = 1, #items do
 			local btn = ui:Button():SetInnerWidget(ui:Label(items[i].text))
-			btn.onClick:Connect(function()
-				onCloseSettings()
-				items[i].onClick()
-			end)
+			btn.onClick:Connect(items[i].onClick)
 			close_buttons[i] = btn
 			if (items[i].toDisable and items[i].toDisable()) then
 				btn:Disable()
--- a/data/ui/Settings.lua
+++ b/data/ui/Settings.lua
@@ -365,7 +365,7 @@ ui.templates.Settings = function (args)
 	setTabs:AddTab({ id = "Controls", title = l.CONTROLS, icon = "Gamepad",     template = wrapWithScroller(controlsTemplate) })
 
 	local function onCloseSettings()
-		print('Refreshing game settings now...')
+		Engine.RefreshBackground()
 	end
 
 	local close_buttons = {}
@@ -375,7 +375,7 @@ ui.templates.Settings = function (args)
 			local btn = ui:Button():SetInnerWidget(ui:Label(items[i].text))
 			btn.onClick:Connect(function()
 				onCloseSettings()
-				items[i].onClick
+				items[i].onClick()
 			end)
 			close_buttons[i] = btn
 			if (items[i].toDisable and items[i].toDisable()) then
--- a/data/ui/Settings.lua
+++ b/data/ui/Settings.lua
@@ -123,6 +123,18 @@ ui.templates.Settings = function (args)
 		local fullScreenCheckBox = optionCheckBox(
 			Engine.GetFullscreen, Engine.SetFullscreen,
 			l.FULL_SCREEN)
+			
+		local starDensity = function (caption, getter, setter)
+			local initial_value = getter()
+			local slider = ui:HSlider()
+			local label = ui:Label(caption .. " " .. math.floor(initial_value * 100) .. "%")
+			slider:SetValue(initial_value)
+			slider.onValueChanged:Connect(function (new_value)
+					label:SetText(caption .. " " .. math.floor(new_value * 100) .. "%")
+					setter(new_value)
+				end)
+			return ui:VBox():PackEnd({label, slider})
+		end
 
 		return ui:Grid({1,1}, 1)
 			:SetCell(0,0, ui:Margin(5, 'ALL', ui:VBox(5):PackEnd({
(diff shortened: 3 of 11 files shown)


walterar:
error: [string "[T] @ui/Settings.lua"]:130: attempt to concatenate local 'caption' (a nil value)
stack traceback:
	[string "[T] @ui/Settings.lua"]:130: in function 'starDensity'
	[string "[T] @ui/Settings.lua"]:158: in function 'template'
	[string "[T] @ui/Settings.lua"]:355: in function 'template'
	[string "[T] @libs/ui/TabView.lua"]:130: in function 'SwitchToNum'
	[string "[T] @libs/ui/TabView.lua"]:96: in function 'AddTab'
	[string "[T] @ui/Settings.lua"]:362: in function 'Settings'
	[string "[T] @ui/MainMenu.lua"]:78: in function <[string "[T] @ui/MainMenu.lua"]:76>

walterar:
This happens when another non-English language is used. No problem of this PR. But it is a problem.

walterar:
run Pioneer
In config; set "Density of Star field" 20%
leave
run Pioneer
In config; "Density of Star field" 100%

Furthermore, I see no difference in density in any point between 0 and 100

Linux

impaktor:
> This happens when another non-English language is used. No problem of this PR. But it is a problem.

You can't chose any other language than English in a PR, since this isn't merged, that new string hasn't been copied to any other language file by the bot yet, so a crash is expected for any un-merged code that tries to access a new language string.

fluffyfreak:
@walterar 
> Furthermore, I see no difference in density in any point between 0 and 100

I said in the description:
> If you do this ingame you will need to jump to another system, or quit and start/load for it to take effect.

fluffyfreak:
I had omitted a call to save the config file after changing the settings :(

Is there a way to call a function on closing the options screen? @impaktor / @laarmen

impaktor:
> Is there a way to call a function on closing the options screen? 

Not sure. I'll pass that question on to @johnbartholomew. 
Hack: add a button next to the slider to save it?

So you have a slider now? Would having a text (number) input field make it simpler?

walterar:
"You can't chose any other language than English in a PR"

If you do not admit there is a problem, you'll never see the solution

This is one of the solutions that come to mind at this time, can be improved.

--- /pioneergit2/data/ui/Settings.lua
+++ /pioneergit2/data/ui/Settings.lua~
@@ -123,14 +123,14 @@
   local fullScreenCheckBox = optionCheckBox(
    Engine.GetFullscreen, Engine.SetFullscreen,
    l.FULL_SCREEN)
-
+  local capt = l.STAR_FIELD_DENSITY or "Density of Star field"
   local starDensity = function (caption, getter, setter)
    local initial_value = getter()
    local slider = ui:HSlider()
    local label = ui:Label(caption .. " " .. math.floor(initial_value * 100) .. "%")
    slider:SetValue(initial_value)
    slider.onValueChanged:Connect(function (new_value)
-     label:SetText(caption .. " " .. math.floor(new_value * 100) .. "%")
+     label:SetText(capt .. " " .. math.floor(new_value * 100) .. "%")
      setter(new_value)
     end)
    return ui:VBox():PackEnd({label, slider})
@@ -155,7 +155,7 @@
     cockpitCheckBox,
     compactScannerCheckBox,
     confirmQuit,
-    starDensity(l.STAR_FIELD_DENSITY, Engine.GetAmountStars, Engine.SetAmountStars),
+    starDensity(capt, Engine.GetAmountStars, Engine.SetAmountStars),
    })))
  end

Andy

manual changing config.ini

AmountOfBackgroundStars = 0.05
or
AmountOfBackgroundStars = 0.80
or
AmountOfBackgroundStars = 1.0

no difference

fluffyfreak:
@impaktor it's saving now, I fixed that up earlier. 
About @walterar problem I have no idea, it works fine for me - but you do have to quit the game, and start/load one before you see the difference.

If I could call something when the options screen closed then I could refresh the background so that would not be necessary.

fluffyfreak:
I should also note that I'm just trying to do a build on my Linux box to test it.

fluffyfreak:
Ok I've tested it on Linux and it built and works perfectly. I'm not sure why you're having a problem @walterar, please not it won't affect any stars baked into a skybox texture, only the point stars.

walterar:
No 3431 With  "static const int BG_STAR_MAX = 5000;"  in Background.h
no_3431

With 3431. With AmountOfBackgroundStars = 0.05 in config.ini
with_3431

fluffyfreak:
Ok, thanks @walterar but on Linux and Windows it works fine for me and we're going to need a 3rd party (@impaktor / @johnbartholomew / @laarmen) to help investigate it further.

Some questions to help:
 - Did you pull the latest change I added earlier today?
 - Did you do a clean build?

walterar:
Did you pull the latest change I added earlier today? yes  eec4e6e

Did you do a clean build? yes

fluffyfreak:
@johnbartholomew or @laarmen how would I go about calling a function when closing the options screen?
What I'd like to do is refresh the background when the options screen closed, rather than wait for the player to jump systems / reload.
Also how can I get the slider and title to appear on the same line instead of one after the other?

shadmar:
Here is a nice wormhole effect (all glsl)
http://glslsandbox.com/e#27050.0

impaktor:
@shadmar looks cool. What I like about our current one is that it has a natural transition from normal space to hyperspace.

impaktor:
Setting the slider to lowest:

    pioneer: VertexBufferGL.cpp:61: Graphics::OGL::VertexBuffer::VertexBuffer(const Graphics::VertexBufferDesc&): Assertion `m_desc.numVertices > 0' failed.

fluffyfreak:
I should probably run that range from 1 (or another sensible minimum) upwards instead of from 0, thanks :)

johnbartholomew:
@fluffyfreak Something like this: https://github.com/fluffyfreak/pioneer/compare/howmanystars...johnbartholomew:howmanystars

Completely untested. Won't do quite the right thing for the Save button on the in-game settings menu (it'll trigger the refresh code when the user presses the "Save Game" button, even though the settings menu hasn't been closed), but probably good enough for a first version.

fluffyfreak:
Ah-ha so that's how you do it, thanks @johnbartholomew :)

walterar:
Now it works! Tested on Linux and Windows (MXE)

:+1:  FIRE IN THE HOLE !!! ;)

walterar:
I do not understand why BG_STAR_MAX value is different from AmountOfBackgroundStars

And if you try K.I.S.S. method? :)

fluffyfreak:
Because `BG_STAR_MAX` is the upper limit of the value, whereas `AmountOfBackgroundStars` is the value itself.

Which is why you see it being clamped between `BG_STAR_MIN` (the lowest value) and `BG_STAR_MAX` (the maximum value).

The sliders values have to be limited in a different way because they run from 0.0 to 1.0, therefore I limit them to `0.01` to mean 1% and `1.0` to mean 100%

walterar:
K.I.S.S.

fluffyfreak:
@walterar that couldn't be any simpler. It is simple. 
The one complication with it is because I didn't want to refresh the background each time the slider moved so it has to be done via a lua registered callback.

fluffyfreak:
There, that should always work by effectively setting a dirty flag when the amount of stars is changed.
It's a bit ugly, but it's almost infallible.

walterar:
With AmountOfBackgroundStars=0.05 always starts with 100,000 stars

fluffyfreak:
@walterar clarify "starts with", do you mean in the start screen?

walterar:
In the start screen and into the game

fluffyfreak:
Ah ok, thanks @walterar

walterar:
This is what I see https://dl.dropboxusercontent.com/u/15391580/Pioneer/4a4c1e8.avi

Marcel-T:
This discussion gave me the idea of a dynamic slider. When the player is in the outer reaches of a solar system the slider would be set to maximum, and when close to a star the slider would be set to minimum, with a function based on distance or lumens to determine the slider setting dynamically. It could be toggled on the option screen.

fluffyfreak:
That would be one hacky way to do it, really though we could do with HDR rendering and then everything just, sorts itself out.

That and a way of generating the background galaxy-scape.

PR #3432, opened 2015-07-25
fixes for bodyscanner

cccpxx78:
Fixes for #3396 as requested by @laarmen 
I moved the 'not function' check for serialization to a new SensorType.Serialize function.
It is now possible to pass a callback function to the StartSensor function of Ship.

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -55,9 +55,7 @@ function EquipType:Serialize()
 	local tmp = EquipType.Super().Serialize(self)
 	local ret = {}
 	for k,v in pairs(tmp) do
-		if type(v) ~= "function" then
-			ret[k] = v
-		end
+		ret[k] = v
 	end
 	ret.volatile = nil
 	return ret
@@ -369,6 +367,18 @@ end
 
 function SensorType:DoCallBack()
 	if self.callback then self.callback(self.progress, self.state) end
+end
+
+function SensorType:Serialize()
+	local tmp = SensorType.Super().Serialize(self)
+	local ret = {}
+	for k,v in pairs(tmp) do
+		if type(v) ~= "function" then
+			ret[k] = v
+		end
+	end
+	ret.volatile = nil
+	return ret
 end
 
 local BodyScannerType = utils.inherits(SensorType, "BodyScannerType")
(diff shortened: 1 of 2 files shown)


cccpxx78:
Hi @laarmen I talked to @impaktor about how he wants to use this scanner in a script and we think there is still something missing: even with this fix it is only possible to pass a callback to, i.e. to monitor, a scanner that is started from a script, it is not possible to monitor a scan started by the player.
GetLastResult could be used in a script to check if a particular 'mission'-planet  is scanned but this would only work if that scan is the last scan.
We discussed polling as a solution but that doesn't seem to be a good a solution and we were thinking in the direction of either
* a more detailed history of scanrecords 
* or some sort event when a scan is completed to which a script can subscribe

I'm currently checking the event direction, how and if this will fit. Do you have any thoughts on this?

laarmen:
I don't get it. Since you give a callback that is called regularly (every time the situation changes, in fact), how can you not monitor a scan started by the player ? You're the one writing the code behind the "start the scan" button or whatever, just write the callback that updates the gauge.

fluffyfreak:
So does this fix #3396 and is there still something wrong with it that prevent merging?

I had a look at the Lua but it made my head spin.

cccpxx78:
The thing was where scripts can register a callback to get notified when a scan is done and get the results. In this fix the callback is passed to begin function of the lua scanner, but that function is normally called from the c++ button. This means scripters can only listen to the result of a scanner started by a script and not by user interaction, scripts can also overwrite a previously registered callback. 
This does not fit the first scenario for the scanner. Either a separate function on ship to register a callback or an event fired from lua ship would maybe be a better fit for this? 
@laarmen sees this problem differently, or not as a problem. To be honest don't understand what he suggests. 
A.t.m. I am on vacation without computer, I'll be back end of September. I can maybe write up a prototype and we can discuss on irc?

fluffyfreak:
Is this PR Active or dead?

impaktor:
Active.
Me or @laarmen just need to sit down with it.
I've been feeling guilty a long time for not doing it sooner.

impaktor:
@fluffyfreak 😄 
Do note that was _December_ 2015 so almost 2016, and we're still _almost_ in 2017... with a little strech of the imagination.

PR #3435, opened 2015-07-30
Remove the animation Interpolate optimisation.

fluffyfreak:
This is try and fix a [reported issue](http://spacesimcentral.com/ssc/topic/4639-landing-gear-behave/) by Gernot.

I'm not able to repeat the problem locally but I do remember this giving us some trouble at the time.

@walterar if you would like to test this change, perhaps you can supply Gernot with executables for him to try out?

Andy


walterar:
Here are the executables for this PR
https://dl.dropboxusercontent.com/u/15391580/Pioneer/pioneer-3435.zip
You will need http://sourceforge.net/projects/pioneerspacesim/files/pioneer-20150715-win32.7z/download

PR #3437, opened 2015-08-10
Fixed nan on burn indicator

dodomorandi:
When the ship is docked, the orbit is null, even if the docking object has an orbit. This avoids showing a nan value in first person view.

### DIFF
--- a/src/WorldView.cpp
+++ b/src/WorldView.cpp
@@ -1587,20 +1587,22 @@ void WorldView::UpdateProjectedObjects()
 		HideIndicator(m_burnIndicator);
 	} else if(systemBody) {
 		Orbit playerOrbit = Pi::player->ComputeOrbit();
-		double mass = systemBody->GetMass();
-		// XXX The best solution would be to store the mass(es) on Orbit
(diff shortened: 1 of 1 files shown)


impaktor:
Hmmm, now I can't even see my ship when docked at Gates for instance. No red marker for "me" and no red orbit. Is that correct?

dodomorandi:
Correct, but the two issues are not related. The problem is that the SystemBody for the player is different from the one passed to PutBody. If you try to perform the same test on the master branch, you will notice the same issue.
Probably, we should change this behaviour, because, as you can see, it is not possible to design a maneuver until undocked.

impaktor:
I couldn't see my own current orbit, nor the planned, after undocking. Tried flying a bit from the station, and accelerating some away from the station.

dodomorandi:
Unfortunately, the issue exists until the ship moves "relative" to the station. As the orbit changes to "System", everything starts working correctly. The issue was present before the experimental branch merge (I just compiled 26e67f6fdd4f704968c2bab42cd0db87a862099d, and the issue exists).
I will study the problem during next days, I see if I am able to solve it! :)

impaktor:
Code looks straight forward enough.

PR #3438, opened 2015-08-10
Two state hyperspace button

impaktor:
This is a work-done-in-2-minutes-PR:

I just whipped this together based on an icon @nozmajner gave me.

The `OnHyperspace` code now takes the button state as input, but doesn't use it, since it just checks if countdown is in progress (as current master does it), so I might clean that up.

Also, I did not change the mouse over text, since if you abort, you wont have time to read it anyway. Might fix that as well.

I'm not sure about the icon, part of me is wondering about having it just highlight the background like when the landing gear is down.

My editor cleaned up some auto-detected wrong indentation so the diff shows a bit of unnecessary edits.


impaktor:
screenshot-20150811-100243

nozmajner:
How about these?
hyperspace_abort_g
hyperspace_abort_h
hyperspace_abort_i
hyperspace_abort_j
The plugs are from the wpzoom pach we use for the other icons, but I don't like it that much.

impaktor:
Nice.

The first and last one don't work well with translation system, and they would mark a first in having text on them, so they have that against them.

The second one is what I had in mind. It's discreet and shows that a button has two states, and that it's currently "active". Perhaps not all players will instantly understand that you can abort the count down, but I think this has a nice balance and looking good, since all other options look a bit strange to me.

So my vote goes to number 2.

nozmajner:
I agree.

impaktor:
Sorry for pretty unclear and unstructured first post in this PR. (I've removed the indentation clean up, to not clutter up this PR, and updated the commits.)


## What have I done

- New hyperspace button by @nozmajner, for when hyperdrive is active
- New mouse over text for both inactive and active hyperdrive button
- After hyperspace exit, reset hyperdrive button to inactive drive state


## Current problem with this two-state button

- If a hyperspace target is selected while docked, and the player starts clicking the button, it will toggle between the two states, without either engaging hyperdrive nor printing error message.


## Suggested solution

- When the player tries to stert the hyperdrive, check the return statement of the start hyperdrive function in `WorldView::OnClickHyperspace()`, and...

- ...print an error/warning message if the player tries to start the hyperdrive while docked, (and reset the button back to inactive state)


## Problem implementing said solution

I (think I) need to get the return enum `Ship::HyperjumpStatus` from the function:

1. `Ship::InitiateHyperjumpTo()` (in src/Ship.cpp) which is returned by the function that calls it which is:
2. `Player::InitiateHyperjumpTo()` (in src/Player.cpp) which is called from
3. `HyperspaceTo()` in libs/Equipment.lua which returns a tuple on the lua side (which is seen as void in c++?) when it is called from
4. `WorldView::OnClickHyperspace()`

So, the problem here (for my brain) I think is in step 3, where `Ship::HyperjumpStatus` seems to be lost? Ping @laarmen

fluffyfreak:
Should the button to hyperspace be visible at all when the player is docked? 
Avoid the problems entirely by disabling the button when docked.

impaktor:
That would work, but then the player wouldn't have feedback that he's "good to go" as soon as he takes off. I think it's nice see that the hyperdrive is "loaded". I solved it another way, in the "whatever works" mantra.

fluffyfreak:
@impaktor is this good to merge now?

impaktor:
@fluffyfreak It now works, but I'm not 100% happy with the code, ideally, I'd like to catch the state of the hyperdrive, and still show the button when landed (if a hypertarget is selected), but this will do for now, unless @johnbartholomew wants to suggest/object to the code.

johnbartholomew:
I have no objection. Unfortunately I also have no time to come up with any useful suggestions.

fluffyfreak:
At the moment I'm trying to get a few pending PR's merged at the start of the month rather than the end! Then we can identify and resolve any issues and do code cleaning as we go.

PR #3444, opened 2015-08-30
Changed/removed some non-portable C++11 usage.

fluffyfreak:
Just a minor fix for the vs2013 build.

Cannot use `constexpr`, just use `const` also usage of `or`, `and`, `not` etc doesn't work in vs2013.


impaktor:
OK, this is difficult to search, and I'm too lazy to read the docs, but what's the added functionality with using `and` instead of `&&`? Just feels very strange to see that in C++ code to me.

robn:
No added functionality, they're just a relic from a time when keyboards might not have had the needed symbols. Don't use them.

PR #3445, opened 2015-09-03
Fix forced time acceleration

clausimu:
This allows forced time acceleration (Ctrl + click on respective time acceleration) in cases where ships are close but are not firing. Before, forced time acceleration was only allowed if close to bodies other than ships. With this enabled we can work on missions where ships travel together (wingmen, etc.).

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -27,6 +27,7 @@
 #include "ui/Context.h"
 #include "galaxy/GalaxyGenerator.h"
 
+
 static const int  s_saveVersion   = 81;
 static const char s_saveStart[]   = "PIONEER";
 static const char s_saveEnd[]     = "END";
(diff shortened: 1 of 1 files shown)


impaktor:
Good to see you got github working, and thanks for contributing.

Without looking at the code, I guess a wingman functionality should not need a forced time acceleration at all but just "work". I haven't thought about what one might do to achieve that, but I guess perhaps somehow adding a ship to enter formation with your own (F4?), creating some kind of pseudo "group"?

In extension, if we had such "groups" then those could be used for AI tradeship convoys or pirates. But this sounds like a very big project, I'd imagine, but I'm not at all familiar with the code and what would go into it.

clausimu:
Long term you are right. But my C++ skills are not good enough yet to pull something like this off. I still have a hard time finding my way around the pioneer source... :)
I figured this is easy for me to fix and at least allows me to test convoys or missions involving other ships.

fluffyfreak:
Ok, so the question here is: Do we want to allow players to force time acceleration when another ship is close by?

This looks like it gets around the under-attack problem where you could be destroyed rapidly by another ship so it shouldn't be harmful.

walterar:
With active acceleration time, all NPC ships are faster than the player. If you have an NPC behind you, he will catch you.

No problem with this PR, it may be the beginning of much more interesting things, but someone, someday, will be happy to investigate the many things that go wrong during the acceleration time. :)

My opinion is: :+1:

fluffyfreak:
Ok I have no strong feelings about this, worst case we can always revert it if people find it to be a problem.

Any dissenting voices before I merge it?

impaktor:
I have no opinion either, beyond what's been said.

fluffyfreak:
Then @clausimu we'll revert it should the need arise but with no dissenting voices it's gone in!
I hope you're wingman stuff is going well.

Issue #3443, opened 2015-08-27
OS X Homebrew formula


tomyun:
I've just made a formula for the OS X package manager, [Homebrew](http://brew.sh/):

Homebrew/homebrew-games#329

It's not building an app bundle for you, but still easily installable by just typing `brew install pioneer`. If interested, please have a look and leave any issues or comments.

nozmajner:
Oh, nobody commented on this ever since? I guess because none of us have any apples around to properly test it, but let's shout them up anyway. :)
@impaktor @fluffyfreak @johnbartholomew

johnbartholomew:
I don't have a system I can test this on, but we should probably mention it in COMPILING.txt

https://github.com/pioneerspacesim/pioneer/blob/master/COMPILING.txt

impaktor:
I have no clue what any of this means. Is it like a build script for pioneer on Mac?
Is there anything we should do?

@tomyun could you suggest text to be added to COMPILING:txt?

tomyun:
Just opened a PR for COMPILING.txt: #3447

impaktor:
OK, so the new Hombrew PR is: https://github.com/Homebrew/homebrew-games/pull/411

Given that in the past month they have 52 open PRs and 0 merged, I'm sceptical to how long it will take for them to merge it.

tomyun:
Homebrew doesn't rely on GitHub's merge feature, instead let maintainers commit themselves after sign-off each PR, so their merge count could be quite misleading. But, it's also true that it's taking longer than usual. I guess they've been pretty busy getting ready for El Captian released this week. I'll ping them again anyways.

tomyun:
`pioneer` is now available via `homebrew-games`.

PR #3447, opened 2015-09-05
Add description for Homebrew on OS X

tomyun:
Per request of #3443, description for building and installing pioneer using [Homebrew](http://brew.sh/) package manager on Mac OS X is added.

By the way, `pioneer` package is not merged to the public yet, waiting for CI test. Also a new formula for the latest version (20150904) is on waiting for submission. I'll ping this thread once all done.

### DIFF
--- a/COMPILING.txt
+++ b/COMPILING.txt
@@ -5,8 +5,9 @@ Table of Contents
   1.2 Windows - MSVC
   1.3 Windows - Autotools (Linux cross-compile)
   1.4 OS X - Autotools
-  1.5 OSX - XCode 4
-  1.6 Nix - Autotools
+  1.5 OS X - XCode 4
+  1.6 OS X - Homebrew
+  1.7 Nix - Autotools
 2 pioneer-thirdparty
   2.1 Linux - Autotools
   2.2 Windows - MSVC
@@ -149,8 +150,8 @@ Note: Compiling from source this way isn't recommended as it doesn't allow you
       bundle). It also isn't the 'Apple way', To do that you need to use XCode.
 
 
-1.5 OSX - XCode 4
------------------
+1.5 OS X - XCode 4
+------------------
 
 * The XCode project isn't as up to date as the 'autotools' method and may
   be broken.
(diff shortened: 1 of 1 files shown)


impaktor:
Thanks!

I guess we'll wait with merging this until your pioneer-hombrew pull request has been merged.

tomyun:
While we are still waiting for the formula get merged in `homebrew-games`, I just got a concern on our current versioning scheme. After we set the current version to `20150927`, we might never get back to `1.0` once it's ever released, simply because `20150927` higher than `1.0` both numerically and lexically.

One good solution could be a [semantic versioning](http://semver.org/), but if it's too much for now, maybe at least we need to leave some room for the future by versioning more carefully, i.e., `0.0.20150927` or `0.1+20150927`, etc.

I'm about to open a PR for the new version in `homebrew-games` anyways, do you have any suggestion on this? This issue should not be unique with homebrew, but probably universal with all other sensible packaging systems.

impaktor:
I believe pioneer abandoned the old version scheme, and will never release a 1.0 version, and will keep to the 20YYMMDD system.

In part, since we no longer strive to clone Frontier, there is no clear goal for a "1.0" version, which I guess, would be a fully featured clone on said game. (the road map on our wiki that mentions v. 1.0 is outdated.)

tomyun:
Good to hear about that! Maybe I was referring to [an old wiki](http://pioneerwiki.com/wiki/Roadmap). I'll open a PR for `20150927` with no worries then.

impaktor:
I've updated the wiki now.

Yeah, version naming convention is an interesting topic, I've found out. Not only do we have the "normal" [Semantic](http://semver.org/) system with: `v. major.minor.patch`, but also there's [Sentimental](http://sentimentalversioning.org/) system which is more eccentric, like Donald Knuth's two projects __TeX__, (Each version approaches `pi` asymptotically, currently on [v. 3.14159265](https://en.wikipedia.org/wiki/TeX)), or __Metapost__ which uses `e` instead.

impaktor:
@tomyun thanks for contribution!

Mack-Mack:
Does this mean there might be regular OSX releases again?

impaktor:
@Mack-Mack this means that it is now easier for OS X users to build pioneer, since dependencies are resolved automatically. However, at the moment there are still no Mac builds being done. @fluffyfreak is trying to get pioneer to compile for him using Xcode, with the ambition of providing one build per month.

Mack-Mack:
Thanks for the info Impaktor and thanks fluffyfreak for the work on the OSX version.

PR #3448, opened 2015-09-05
Terrain detail textures.

fluffyfreak:
# Description:
Detail texturing is a cheap graphics programming trick. It doesn't really add more detail it just uses some textures to make it look like you have!

There are downsides of course, this uses the same two textures absolutely everywhere, so a barren moon will receive the same detail texturing as lush verdant forests, or deserts, or lava. That's the downside too it being a cheap trick.

This particular implementation started out as an experiment by @Luomu at some point in the dim and distant past. I liked the idea and brought it up to date with the codebase about a year ago. Then I got distracted by a shiny thing, maybe a laser pointer, or my foot, who knows.

Today I have updated it to the latest master, converted the textures to DDS, fixed a couple of issues, added per-patch texture repeat scaling to stop the texture changing size and cleaned things up.

# Issues:
None of which I am aware, however it would be good to have some other eyeballs on this reporting what they like / dislike.

A good way of testing is to fly to a planet, land and take a look around, then takeoff again and as you fly away (this works best at daytime folks!) look back to see the planet recede into the distance.

Those are the most likely times you will see issues causes by too much or not enough texture detail.

Comments welcome, and @Luomu I hope you don't mind me pinching this old experiment of yours, I just needed something to hack on and get over an unproductive patch! :)

# Notes:
They can be disabled in the config file by adding the line `DisableDetailMaps=1`.

Andy


impaktor:
Looking forward to giving this a try after the weekend.

walterar:
I can test? No risk of explosion?

fluffyfreak:
@walterar give it a go :) it's a nice simple trick that any computer should be able to run.

I expect there to be some issues with this. 
There is some difficulty in working out the correct relationship between terrain patch and texture repetition that might go wrong in some cases.

I didn't find any yet where it does, but that just means I haven't explored enough.

Is it worth me adding an option to disable it in the config screen as well?

impaktor:
> Is it worth me adding an option to disable it in the config screen as well?

I suggest waiting with that to see what players think of this feature. (Also, one can have options just be in the config file, and not be exposed on the settings screen.)

walterar:
MXE compiled without problems too.

I merged this in my daily work master of Scout, and not notice differences. But perhaps I have not looked in the right place?

fluffyfreak:
@walterar when testing something like this just do it directly not merged into something else, there's so many things that can go wrong doing that it makes testing pointless.

To see the effect you just need to get close to the surface of the planet with the detail set to medium and the planetary textures off. You will see it most clearly that way.

fluffyfreak:
As for what it should look like:
Detail ON:
Detail on
Detail OFF:
Detail off

nozmajner:
Looks neat. Can't wait to play with the textures. :).

impaktor:
Looks nice. I don't see any reason why anyone would not want to have it on, if it's cheap.

Also:
```
GeoSphereMaterial.cpp: In member function ‘void Graphics::OGL::GeoSphereSurfaceMaterial::SetGSUniforms()’:
GeoSphereMaterial.cpp:109:105: warning: use of old-style cast [-Wold-style-cast]

TextureFont.cpp: In member function ‘Text::TextureFont::Glyph Text::TextureFont::BakeGlyph(Uint32)’:
TextureFont.cpp:467:75: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
TextureFont.cpp:473:47: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
TextureFont.cpp:531:70: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
TextureFont.cpp:537:41: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
```

fluffyfreak:
Merged, if anything blows up I'm sure to hear of it quite quickly :)

fluffrabbit:
@fluffyfreak KABOOM! On Windows 8 w/ Intel integrated laptop graphics I get this right after Earth loads before the terrain shows up.

pioneererror

PR #3452, opened 2015-09-12
Resolve Detail Texturing abort

### LINKED ISSUE #3451: Detail Texturing abort.
@fluffrabbit reported the following

> @fluffyfreak KABOOM! On Windows 8 w/ Intel integrated laptop graphics I get this right after Earth loads before the terrain shows up.
> 
> pioneererror

In the #3448 commit thread.

Seems that I didn't check it on enough GPU's and platforms and as usual nVidia happily let me get away with out-of-spec GLSL usage.

Fixing it now, this issue is just to track what's going on.

fluffyfreak:
Should fix #3451 by using `texture` keyword instead of `texture2D` which has been deprecated and which Intel appears to enforce, on Windows, not on Linux or OSX though... hooray for consistency.

### DIFF
--- a/data/shaders/opengl/geosphere_terrain.frag
+++ b/data/shaders/opengl/geosphere_terrain.frag
@@ -79,8 +79,8 @@ float discCovered(const in float dist, const in float rad) {
 void main(void)
 {
 #ifdef DETAIL_MAPS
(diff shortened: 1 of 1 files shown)


fluffyfreak:
Tested this on my Windows 8 laptop with Intel HD 4000. Worked fine so merging now.

PR #3453, opened 2015-09-16
WIP - Terrain texturing

fluffyfreak:
Well here's some updated and [Textured Terrain](http://quantum-thoughts.net/Pioneer/terrain/screenshot-20150916-213921.png) to wet your appetites.

Things to do... many:
- [x] Determine the max terrain height, approximately anyway (currently mul'd 277.0f).
- [ ] Fix the water and lava rendering somehow.
- [ ] Generate the texture array for a planet.
- [ ] Improve the input data used to lookup the texture.
- [ ] Use a different Lookup texture for each planet type (generated?).
- [ ] Get some better textures!
- [ ] Make a better lookup texture!
- [ ] Document it all.

This destroys the terrain colour noise stuff and replaces it with terrain texturing based on the slope and height of the terrain.

It currently uses a choice of 16 textures, which are actually sourced from just 4 originals and duplicated whilst I'm developing it.

Since we'll be texturing whole planets it seems silly to keep the detail texturing optional, so I'll also be removing that option and it will always be there.

Andy


impaktor:
Neat. Looking forward to this.

Regarding your question on IRC, JohnJ (creator of JJFFE) replied:

```
<JohnJ_> <FluffyFreak> I have a question, how can I guess at the maximum 
         height of the terrain for a planet?			        [01:58] 
<JohnJ_> If it's noise-based then you can do a probability maximum.
```

Also, for the RSS issue, I'd contact github support. Could it have something to do with you using windows? I'm not sure what else differentiates your setup with the rest of the devs.

fluffyfreak:
@impaktor if it were just noise that might be possible, but there's a lot of logic and some really stupidly bizarre other maths in there tied into it. 
I think sampling of some form is going to be the most reliable way and hopefully won't be too expensive to process.

fluffyfreak:
Going to close this since i have plans for terrain stuff, might as well leaved it alone until they're done and supercede this.

PR #3460, opened 2015-09-20
Orbit fading

dodomorandi:
Orbit fading have been designed to show the direction of the orbit. This should help the user to better understand the layout of a system without using the time controls.

schermata da 2015-09-20 14-06-14

Here a simple screenshot. For now, I hardcoded when the fading starts, but it could be put inside the options. Now... we have got a problem:
schermata da 2015-09-20 14-06-46

As you can see, when two orbits overlap, things start going crazy. Ideas? Probably, the best one is the check for the same orbits before calling `SystemView::PutOrbit`, but, for now, `Orbit` instances of two bodies that occupy the same orbit but not the same space **are** different.

I was working in my spare time on a completely different system to handle orbits, using the *patched conic approximation* model, but it is taking much more time that I expected, and I have been quite busy in the last month. Honestly, I do not know how much hard it could be to perform the check I was talking before, and if things are going to change, I do not know if it is worth doing.

Opinions? Ideas?

### DIFF
--- a/src/SystemView.cpp
+++ b/src/SystemView.cpp
@@ -443,18 +443,37 @@ void SystemView::PutOrbit(const Orbit *orbit, const vector3d &offset, const Colo
 		}
 	}
 
+	static const float startTrailPercent = 0.75;
+	static const float fadedColorParameter = 0.4;
+
 	vector3f vts[n_vertices_max];
+	Uint16 fadingColors = 0;
+	double t0 = m_game->GetTime();
 	for (unsigned short i = 0; i < n_vertices_max; ++i) {
 		const double t = double(i) / double(n_vertices_max) * maxT;
-		const vector3d pos = orbit->EvenSpacedPosTrajectory(t);
+		if(fadingColors == 0 && t >= startTrailPercent * maxT)
+			fadingColors = i;
+		const vector3d pos = orbit->EvenSpacedPosTrajectory(t, m_time - t0);
 		vts[i] = vector3f(offset + pos * double(m_zoom));
 		++num_vertices;
 		if (pos.Length() < planetRadius)
 			break;
 	}
 
+	std::unique_ptr<Color[]> colors(new Color[num_vertices]);
+	Color fadedColor = color * fadedColorParameter;
+	std::fill_n(colors.get(), num_vertices, fadedColor);
+	Uint16 trailLength = num_vertices - fadingColors;
+
+	for (Uint16 currentColor = 0; currentColor < trailLength; ++currentColor)	
+	{
+		float scalingParameter = fadedColorParameter + static_cast<float>(currentColor) / trailLength * (1.f - fadedColorParameter);
+		colors[currentColor + fadingColors] = color * scalingParameter;
+	}
+
 	if (num_vertices > 1) {
-		m_orbits.SetData(num_vertices, vts, color);
+		m_orbits.SetData(num_vertices, vts, colors.get());
+
 		// don't close the loop for hyperbolas and parabolas and crashed ellipses
 		if (maxT < 1. || (orbit->GetEccentricity() > 1.0)) {
 			m_orbits.Draw(m_renderer, m_lineState, LINE_STRIP);
@@ -462,6 +481,7 @@ void SystemView::PutOrbit(const Orbit *orbit, const vector3d &offset, const Colo
 			m_orbits.Draw(m_renderer, m_lineState, LINE_LOOP);
 		}
 	}
+	colors.reset();
 
 	Gui::Screen::EnterOrtho();
 	vector3d pos;
--- a/src/Orbit.cpp
+++ b/src/Orbit.cpp
@@ -202,10 +202,10 @@ vector3d Orbit::OrbitalVelocityAtTime(double totalMass, double t) const
 // used for stepping through the orbit in small fractions
 // mean anomaly <-> true anomaly conversion doesn't have
 // to be taken into account
-vector3d Orbit::EvenSpacedPosTrajectory(double t) const
+vector3d Orbit::EvenSpacedPosTrajectory(double t, double timeOffset) const
 {
 	const double e = m_eccentricity;
-	double v = 2*M_PI*t + TrueAnomalyFromMeanAnomaly(m_orbitalPhaseAtStart);
+	double v = 2*M_PI*t + TrueAnomalyFromMeanAnomaly(MeanAnomalyAtTime(timeOffset));
 	double r;
 
 	if (e < 1.0) {
--- a/src/Orbit.h
+++ b/src/Orbit.h
@@ -40,7 +40,7 @@ public:
 	vector3d OrbitalVelocityAtTime(double totalMass, double t) const;
 
 	// 0.0 <= t <= 1.0. Not for finding orbital pos
-	vector3d EvenSpacedPosTrajectory(double t) const;
+	vector3d EvenSpacedPosTrajectory(double t, double timeOffset = 0) const;
 
 	double Period() const;
 	vector3d Apogeum() const;


impaktor:
Looks very nice! :+1: 

Regarding Bernard's star, I wouldn't worry too much about it, if you can't find a solution. It's sort of a special system anyway, and having two orbits on top of each other is a strange thing to have anyway.

dodomorandi:
Maybe trying to solve these *speciale* issues when I (hopefully) finished to work with the new orbital system would be the best idea.

nozmajner:
Yeah, it looks nice. 
I was going to say the same as @impaktor about Barnard's, and we can always change the system a bit (I was thinking about it, adding a couple of small desolate planets, where inmates could break stones or whatever, and they could add some opportunities to learn orbital maneuvering)
But the system generator sometimes do something similar, even though rarely as far as I know.

impaktor:
Looking at the planets, and accelerating time, I would, based on the "phosphorous fade out"-effect, expect the planets to move in the opposite direction.

I think of it like the green orbit trails behind the planet, and after some time fades out. Thus the dark gap would be in front of the planet, relative its direction.

I.e. this is how I interpret the direction of the planets when looking at the fade out (but they move in the opposite direction):

ex

To add a bit of luxury, it would be cool if the orbit also rotates around the sun, i.e. have the fade out move with the planet. But that's just a thought, I don't know how much time you want to spend on that.

dodomorandi:
Gosh, you are right: when I wrote the code, I was just imaging the fading part of the trail as... the "end" of the orbit! The problem is that this representation can be interpreted in both ways, and I think we need a better system. Ideas? One could be an animated rotating orbit, but it would need a 3d torus with a minimal texture instead of a simple polyline... What is your opinion?

Regarding movement with planets, I have to give it a look. It should not be too hard, but I need to check...

fluffyfreak:
@dodomorandi I would suggest reversing the current direction so that the line is brightest where the planet has just been. So it fades out with the "age" of the orbit. 

Then to emphasise the effect have it fade out to the darkest colour within 1/4 of the orbit, so that it's more obvious that the orbit itself is dark green, and the bright green fade shows where the planet is coming from.

dodomorandi:
I like it!

This is how it works now:
schermata da 2015-09-22 11-34-00

...ahem... Yes! The direction of the fading is the same of the planet! I just realized that it is impossible to perceive it from the screenshot!

I still need to check how to "move" the fading along the planet. Just a minor question: now the fading scales down the 4 RGBA channels, but is it a good idea to change the alpha? I am not sure, so I am asking. ;)

impaktor:
> The direction of the fading is the same of the planet! I just realized that it is impossible to perceive it from the screenshot!

Not sure I'm understanding you, but in your screen shot above, the planets move counter-clockwise, so I think it's still the opposite to what I expect: i.e. in front of the planet the orbit is dark/faded/old, behind it it is bright.

dodomorandi:
Ok, sorry, you are completely right. It is because my mind *wants* to see it in the opposite way. Fixed!

schermata da 2015-09-22 14-16-28

impaktor:
Yes, looks good now. I would not be opposed to making the "standard" dark green orbit just a nudge darker. I like how the image looks in the github scaled version in your post (in my browser), but clicking it (or trying it in-game), the dark green becomes brighter. Just my own very subjective initial thoughts. Maybe @nozmajner has some thoughts on this?

Either way, nicely done!

dodomorandi:
I have the same feeling, but I was not sure. I set the `fadedColorParameter` to control the fading. The less the value, the *darker* the color. Do you think that it could be better to leave as it is (alpha is scaled with RGB) or always leave alpha to the original value? In any case, we could try to adjust the value to find a good one... or I could add an option inside the menu. But I need your help to find easily how to do it without hacking the whole project ;)

src/SystemView.cpp
```diff
@@ -443,9 +443,15 @@ void SystemView::PutOrbit(const Orbit *orbit, const vector3d &offset, const Colo
 		}
 	}
 
+	static constexpr float startTrailPercent = 0.75;
```
fluffyfreak: These need to be `const` instead of `constexpr` for vs2013  compatibilty

src/SystemView.cpp
```diff
@@ -443,9 +443,15 @@ void SystemView::PutOrbit(const Orbit *orbit, const vector3d &offset, const Colo
...
+	static constexpr float fadedColorParameter = 0.4;
+
 	vector3f vts[n_vertices_max];
+	unsigned fadingColors = 0;
```
fluffyfreak: This should be a `Uint32` instead of a bare `unsigned`.

src/SystemView.cpp
```diff
@@ -453,8 +459,20 @@ void SystemView::PutOrbit(const Orbit *orbit, const vector3d &offset, const Colo
 			break;
 	}
 
+	Color* colors = new Color[num_vertices];
+	Color fadedColor = color * fadedColorParameter;
+	std::fill_n(colors, num_vertices, fadedColor);
+	unsigned trailLength = num_vertices - fadingColors;
+
+	for (unsigned currentColor = 0; currentColor < trailLength; ++currentColor)
```
fluffyfreak: `Uint32` here too please.

src/SystemView.cpp
```diff
@@ -453,8 +459,20 @@ void SystemView::PutOrbit(const Orbit *orbit, const vector3d &offset, const Colo
 			break;
 	}
 
+	Color* colors = new Color[num_vertices];
```
fluffyfreak: Maybe use a `std::unique_ptr` or a `std::vector` here instead. Just for convention.

fluffyfreak:
I've made a few comments but there's nothing serious, and I realise that you're updating existing code so some of them will already have been old, bad and crusty before you started.

This looks good and I think improves the screen.

dodomorandi:
Sure, but... I want to ask you the reason to change `unsigned`to `uin32_t`. Should it be the same for 32 and 64 architectures, isn't it? On the other hand, `fadingColor` will be always less then `n_vertices_max`, which is a short int -- 65535 possible values, obviously enough to store a value of `1000`.

fluffyfreak:
Well `fadingColor` could be `uint16_t` if that's the case. 

The other part is just that we use our aliases defined for all of these: `Sint32` `Uint32` `Sint16` `Uint16`.
I've just checked and a few uses of `uint32_t` have crept in that should be `Uint32` but in most cases I don't mind.

The reason for using our aliases was cross platform compatibility with other compilers. The Visual Studio compiler has lagged behind badly for a long time so even `uint32_t` for example wasn't always standard with it.

dodomorandi:
Ok, *good* to know that Visual Studio always tries to make a mess. Thank you, because I generally use the idea of *"100% standard = cross-compatible"*, and, as you said, it is not always true.

You also made me discover one simple thing: std::make_unique can be used for arrays only since C++14, but the respective std::unique_ptr constructor works flawlessly since C++11, and I was not aware about that.

dodomorandi:
It took me some time to find the easiest way to move orbital trails, but it was easier than I expected! Give it a look, but the modification is simple and it should work flawlessly.

In theory, this modification is subject to double overflow, like else anything related to time control. When the `double time` will be replaced by `Time time`, the `Orbit::EvenSpacedPosTrajectory` method will have to be updated as well.

impaktor:
For some bizarre reason, pioneer stopped working for me now (even master), I think I've encountered this problem a year ago or similar, might require a reboot on my part, but can't restart my computer for a few more days, so it will have to wait. Either way, just wanted to check if you consider it done, and if you'll squash/rebase the branch after review/before merge, or you want us to do it?

dodomorandi:
No problem for me! For now, I kept the commits separated so you could analyse them easily. I think it is finished, and if/when you consider the code as reviewed, I can squash everything in one single commit.

impaktor:
> No problem for me! For now, I kept the commits separated so you could analyse them easily. I think it is finished, and if/when you consider the code as reviewed, I can squash everything in one single commit.

That's the perfect way to do things, in my opinion. :+1:

dodomorandi:
Commits squashed!

impaktor:
I am unable to test this before Thursday, just letting you know. Maybe someone else can test and merge before then.

fluffyfreak:
Testing is for wimps! :P MERGING!!!

fluffyfreak:
@dodomorandi thanks for the improvement!

PR #3461, opened 2015-09-21
boot screen visuals

joonicks:
center-aligned the and shrank it to not cover the whole width of the
screen


joonicks:
bootscreen

fluffyfreak:
Ok this looks good, I haven't had a chance to check the code yet but does any else like our resident artist @nozmajner have any comment about the placement / size?

impaktor:
Comments from noz and others in IRC have been positive. I think it serves as a good first commit, to get into the development process.

I thought you would like to look the code over. That `(xsize > 700)` threw me off a bit, I'd have expected it to, say, always have the bar be "2/3" of the width or something. Haven't looked at it in any more detail.

impaktor:
Here's just a quick idea (mockup) of how to spice up the boot screen further, using our `data/galaxy_colour.bmp`

boot_suggestion

(Yes, I'm posting in this pull request, since this is where we had the main discussion, and both #3462 and this tracks your master-branch, so they're identical in all aspects).

Issue #3456, opened 2015-09-18
Crew for hire: empty


joonicks:
Intended, or?
I dont think I would post on a BBS if I had nothing to offer,
Crew for hire: empty offer on Gates BBS right after starting a new game on earth.

impaktor:
> Crew for hire: empty offer on Gates BBS right after starting a new game on earth.

Well, it's random, so not sure if that will help reproduce it, but you could have a look at the code (data/modules/Crew) to see if you can think of why that could happen

joonicks:
Imho, it shouldnt post unless there is atleast one crew to hire, so, I see it as a bug/flaw

demolitions:
I have looked into this, been able to reproduce, and want to contribute a fix, it is a very simple matter (currently the number of crew generated can be 0), but I wanted both the devs and your advice:
Is it better to:
1) have always at least one crew to hire at every station
2) not post the advert if no crew to hire is generated

In the meanwhile, I'm going to look out how to correctly make a contribution via GIT.

impaktor:
@demolitions Hi! I definitely vote for number 2. And we love to have you on board as a contributor.

joonicks:
maybe like this, generate 2 random numbers from -5 to 15 (example numbers), average them: to make extremes more unlikely, and if the result is below 1, dont post on the bbs, otherwise post the number of crew.

maybe it could take the population of the system into account?

impaktor:
The correct and "full nerd" way to do it is how it is done in second hand module, although it is quite overkill: [SecondHand.lua#L161](https://github.com/pioneerspacesim/pioneer/blob/master/data/modules/SecondHand/SecondHand.lua#L161)

But a simple scheme could be done as well. Yes, take population / station size is preferable. Without having looked at the code recently, I suspect it does the right thing, meaning, I see many crew contracts in London, but none in Bernard's Star, or other small stations.

My suggestion would be to just fix the bug and leave the code as is. If the existing code is replaced I want to see graph of how probability would depend on station size, like in #3220, and how it improves/changes the result of the current code.

demolitions:
Actually, I didn't plan to replace the current code, it already takes the system population in account, its only flaw is that it can generate 0 crew to add (ceil(0, random)), and the BBS post is added before the crew generation, this occasionally results in an empty post.
The only thing I thought worth doing was simply moving the AddAdvert call **after** the generation, and only executing it if enough crew is generated (at least 1).

[...]

PR #3465, opened 2015-09-25
Fix for #3456: Crew for hire empty

demolitions:
Fixes https://github.com/pioneerspacesim/pioneer/issues/3456:
If there aren't enough candidates to hire, the BBS entry is disabled and a message saying so is displayed.
~~Plus, italian translation.~~

### DIFF
--- a/data/modules/CrewContracts/CrewContracts.lua
+++ b/data/modules/CrewContracts/CrewContracts.lua
@@ -208,9 +208,17 @@ local onChat = function (form,ref,option)
 		form:ClearFace()
 		form:Clear()
 		form:SetTitle(l.CREW_FOR_HIRE)
-		form:SetMessage("\n"..l.POTENTIAL_CREW_MEMBERS:interp({station=station.label}))
+		local numCrewInThisStation = 0
 		for k,c in ipairs(crewInThisStation) do
-			form:AddOption(l.CREWMEMBER_WAGE_PER_WEEK:interp({potentialCrewMember = c.name,wage = Format.Money(c.estimatedWage)}),k)
+			numCrewInThisStation = numCrewInThisStation + 1
+		end
+		if numCrewInThisStation > 0 then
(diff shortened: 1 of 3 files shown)


impaktor:
Could you please remove the italian translation, since this will at best be overwritten by our bot, and at worst we'll get a conflict. 

The way new strings are added works like so: When this is merged, the new string in en.json is copied to all other json files before our transifex translation system is updated. Then if the string changes in any of the json files in transifex those will be pushed into our git repository every 12 h I think.

Also, I took the liberty to edit your post: "Fix for" --> "Fixes", since then this issue will auto-close the bug report upon merge:  https://help.github.com/articles/closing-issues-via-commit-messages/

data/modules/CrewContracts/CrewContracts.lua
```diff
@@ -208,9 +208,17 @@ local onChat = function (form,ref,option)
 		form:ClearFace()
 		form:Clear()
 		form:SetTitle(l.CREW_FOR_HIRE)
-		form:SetMessage("\n"..l.POTENTIAL_CREW_MEMBERS:interp({station=station.label}))
+		local numCrewInThisStation = 0
 		for k,c in ipairs(crewInThisStation) do
-			form:AddOption(l.CREWMEMBER_WAGE_PER_WEEK:interp({potentialCrewMember = c.name,wage = Format.Money(c.estimatedWage)}),k)
+			numCrewInThisStation = numCrewInThisStation + 1
+		end
```
impaktor: Of what I can see, the `crewInThisStation` is an array, since it's only using `insert`, so `#crewInThisStation` should give you the array lenght. Just for future reference.

impaktor:
Squashed commits, and merged them manually. Thus the "close", rather than "merge".

impaktor:
...and thanks for contributing!
you're in the AUTHORS file with name as agreed in irc.

PR #3467, opened 2015-09-26
commodityMarket rewrite

joonicks:
now supports buying more than one unit at a time,
with checks and warnings. will also display descriptions if they are
present (right now they are not)

### DIFF
--- a/data/ui/StationView/CommodityMarket.lua
+++ b/data/ui/StationView/CommodityMarket.lua
@@ -6,7 +6,6 @@ local Lang = import("Lang")
 local Game = import("Game")
 local Format = import("Format")
 local Equipment = import("Equipment")
-local SmallLabeledButton = import("ui/SmallLabeledButton")
 
 local EquipmentTableWidgets = import("EquipmentTableWidgets")
 local ui = Engine.ui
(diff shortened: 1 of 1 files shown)


joonicks:
screenshot-20150926-193509

nozmajner:
The mockups I posted on IRC a couple of days ago. Only some rearranging. @impaktor thought it might be handy to have them here.
Buy "tab"
Sell "tab"

fluffyfreak:
To repeat what I said in IRC:

I think this is good and I am tempted to merge it as-is.

However it's a little bit confusing to use right now because I kept thinking that by pressing the +/- keys I could buy *OR* sell instead of having to switch to the "sell" page. 
Others in IRC also found this to be how they were trying to use it.

@joonicks would it be possible to change it so that it worked that way?
If not then let us know and I will merge it as-is for us to refine in the future because I think that it is definitely better than our current market screen.

fluffyfreak:
@joonicks are you going to do any other changes to this?

joonicks:
I could comment it so others stand a better chance of altering it, but its a big change so right now id like to merge it so people can have opinions on what to tweak and build consensus on whats the best

impaktor:
Maybe @radius75 would be interested in having a go at this.

PR #3469, opened 2015-09-28
Fade orbits faster and darker

fluffyfreak:
This is really 3 changes. 2 minor and one a little bigger.

The bigger change is avoiding the allocation and deletion of the former 'colors' buffer per-orbit / per-frame.
This is now a member variable that is initialised to the maximum number of vertices during construction. I also replaced the 'vts' array with a similar arrangement. The buffer is allocated once, fixed in size, and the code just uses as much of it as it needs.

The other 2 changes are that I've shortened the distance before the orbit fades out from 25% of the orbit to 15% and then I've made it fade to a much darker colour.

Personally I think that this looks much nicer and shows the direction thanks to the fading much better.

### DIFF
--- a/src/SystemView.cpp
+++ b/src/SystemView.cpp
@@ -22,6 +22,7 @@
 using namespace Graphics;
 
 const double SystemView::PICK_OBJECT_RECT_SIZE = 12.0;
+const Uint16 SystemView::N_VERTICES_MAX = 100;
 static const float MIN_ZOOM = 1e-30f;		// Just to avoid having 0
 static const float MAX_ZOOM = 1e30f;
 static const float ZOOM_IN_SPEED = 2;
@@ -385,6 +386,9 @@ SystemView::SystemView(Game* game) : UIView(), m_game(game)
 	RefreshShips();
 	m_shipDrawing = OFF;
 	m_planner = Pi::planner;
+
+	m_orbitVts.reset( new vector3f[N_VERTICES_MAX] );
+	m_orbitColors.reset( new Color[N_VERTICES_MAX] );
 }
 
 SystemView::~SystemView()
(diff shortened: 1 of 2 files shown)


impaktor:
Since this is a patch to #3460, I guess @dodomorandi wants to be pinged. 
Perhaps @dvolk as well?

dodomorandi:
Thank you for pinging me!
Nice done! I am still a bit afraid of doing these sort of things by myself, because I could not be aware of eventual multithread access to some instances (I am still a noob with Pioneer hacking).
Now it is harder, better... ok, you already know that! :+1:

impaktor:
I was just trying this out. It's gotten pretty dark in here (systemview) or is it just me? :smile: 

screenshot-20151002-164029

I can't see any orbit at all other than the trail.

fluffyfreak:
I like it, I'M A GOTH! :P

impaktor:
> I'M A GOTH!

I'll counter with Immortal's "Blacker than darkness".

Seriously though, I think it looks a bit weird with the (peri)apsis icons floating in the darkness, lost forever (or until the next planetary revolution).

I should experiment some, and post my suggestion.

impaktor:
I find I like `fadedColorParameter = 0.2;`, I also tried `0.15` in which I found the orbits look good, but the periapsis icons are quite "contrasty" to the very dark orbit.

We could await the next build, and see what others think.

fluffyfreak:
See, I told you I should have made it configurable! :P

There's a few things, like #3467, that I've merged because I think we need community feedback on and to try and agitate people into getting involved to change ;)

impaktor:
> See, I told you I should have made it configurable! :P

I prefer a good default instead of 10000 settings in the optoins menu.

Long term, it would be nice if all hardcoded colours, alphas, in c++ were exposed to some theme file. Step one would be to check which hardcoded values @walterar changed in scout+. Then we could have a scout+ theme.

PR #3470, opened 2015-09-30
Visual Studio 2015 project files

fluffyfreak:
Build the game using Visual Studio 2015 Community edition.

Requires some new libs and dlls in thirdparty.


impaktor:
There are icons in here as well. Didn't expect that.

fluffyfreak:
Yes, for the profiler output html formatting.
When you run the profiler it dumps a html file which when you open it uses them to format and layout the results.

PR #3474, opened 2015-10-04
Make trade analyzer module into a non-removable module

joonicks:
step 2 would be to alter the ship mechanic to actually being able to
purge it from the ship if the player so wished...

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -736,7 +736,7 @@ misc.hull_autorepair = EquipType.New({
 })
 misc.trade_analyzer = EquipType.New({
 	l10n_key="TRADE_ANALYZER", slots="trade_analyzer", price=400,
-	capabilities={mass=0, trade_analyzer=1}, purchasable=true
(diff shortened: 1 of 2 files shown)


joonicks:
screenshot-20151004-132332

impaktor:
Just posted some general thoughts, slightly related to this PR: http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=364

fluffyfreak:
I have no strong feelings about this one way of the other but I do question whether something needs to be a piece of "equipment" if it cannot be removed and everyone always has one?

Isn't it then just a function of the machine and thus we should remove the check to see if you have one and just assume that you do?

impaktor:
I kind of feel this doesn't serve much function on its own. Maybe if there was more "features". As it stands, it feels like it's removing functionality. 

I can think of the trade analyzer (Maybe it should have been called "trade computer" in the first place?) as beeing a physical cartridge, or expansion board to a computer. Maybe the player wants to sell it to get the cash?

fluffyfreak:
Ok now that I understand what this is for I think it makes sense especially if it's the foundations for future work. Merging.

PR #3479, opened 2015-10-05
Expose ship::Blastoff() to Lua API

impaktor:
@clausimu requested this on the [forum](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=363), since he's placing ships with

    ship = Space.SpawnShipLanded(shipdef.id, Space.GetBody(planet_systembody.index), ad.lat, ad.long)

But we have no method to make them move (!), since hyperjump will not work in
landed state, and `AIFlyTo()` doesn't work (maybe we should look into that instead/also?)

This method:
- [x] Blasts off ship in normal direction
- [x] Retract landing wheels
- [ ] Set speed to 0 relative planet
- [x] Ponder on expected behaviour of `AIFlyTo`

### DIFF
--- a/src/LuaShip.cpp
+++ b/src/LuaShip.cpp
@@ -416,6 +416,31 @@ static int l_ship_undock(lua_State *l)
 	return 1;
 }
 
+/*
+ * Method: Blastoff
+ *
+ * Blast off, in normal direction to the ground, and retract landingear
+ *
+ * > ship:Blastoff()
+ *
+ * Availability:
+ *
+ *  20151005
+ *
+ * Status:
+ *
+ *  experimental
+ */
+static int l_ship_blastoff(lua_State *l)
+{
+	Ship *s = LuaObject<Ship>::CheckFromLua(1);
+	if (s->GetFlightState() != Ship::LANDED)
+		luaL_error(l, "Can't blastoff if not landed");
+	s->Blastoff();
+	s->SetWheelState(false);
+	return 1;
+}
+
 /* Method: SpawnMissile
  *
  * Spawn a missile near the ship.
@@ -943,6 +968,7 @@ template <> void LuaObject<Ship>::RegisterClass()
 
 		{ "GetDockedWith", l_ship_get_docked_with },
 		{ "Undock",        l_ship_undock          },
+		{ "Blastoff",      l_ship_blastoff        },
 
 		{ "Explode", l_ship_explode },
 


impaktor:
Digging into the code, `AIFlyTo` does call `ship->blastoff()` if flightstate is `LANDED`, so maybe this PR isn't needed after all. 

For @clausimu to test:

1. test `AIFlyTo()` on the player ship `LANDED`
2. test `AIFlyTo()` on the player ship `DOCKED`

clausimu:
Just to have a record: 

AI ship LANDED => AIFlyTo(orbital_station) => ship takes off
AI ship LANDED => AIFlyTo(ground_station) => ship stays on the ground (status: LANDED)
player ship LANDED => AIFlyTo(orbital_station) => ship takes off
player ship LANDED => AIFlyTo(ground_station) => ship stays on the ground (status: LANDED)

impaktor:
Great find!

Now, it would be interesting to see if FlyTo a ground station (from LANDED state) works if:

1. ground_station is on _another planet_? If so then test:
2. ground_station is on _same_ planet, (but distant enough, London -> Tokyo)?

clausimu:
I just tried with AIDockWith() using the player ship - and it does not work with any base if my ship is landed (ground base, orbital base, any distance)...

impaktor:
@clausimu reported that the issue with `ground_station` not working is due to it being within 15 km, which `FlyTo` considers to be "here", thus no where to fly to. I intend to add a bit more text to the API description for the `FyTo` command, to make it clear what it does and doesn't do. 

It is thus clear that the code in this PR is not needed, but I'd like to investigate how `AIDockWith()` is supposed to work, so this PR is morphing into some semi-bug-report.

@clausimu, I guess, to get ships off the ground just `FlyTo` the system star, then spool up jump drive, or issue `AIDockWith`.

impaktor:
OK, looking fast at the code, `AIDockWith()` assumes the player is in FLYING state. If not, it assumes the player has laned (at its destination), I think, which the comment also hints at.
https://github.com/pioneerspacesim/pioneer/blob/master/src/ShipAICmd.cpp#L912
```c++
	if (m_ship->GetFlightState() != Ship::FLYING) {		// todo: should probably launch if docked with something else
		m_ship->ClearThrusterState();
		return true; // docked, hopefully
	}
```

I'll add that to the API documentation as well.

clausimu:
Thank you. I think I can now accomplish what is needed to get ships to go where they are supposed to to.

However, if I had any say in this I would still vote for having that AIBlastOff() method. This does two things:

a) I can move ships that are within 15 km from their target without having to give them some random target that is further away to get them off the ground and then switch targets back to the original target.

b) It clarifies the steps necessary to go from LANDED to DOCKED (1. AIBlastOff(), 2. AIDockWith()) without having to explain (and remember) that AIFlyTo() did work when ships are LANDED, while AIDockWith() did not, while both work when ships are FLYING.

Just my 2 cents.

impaktor:
> if I had any say in this I would still vote for having that AIBlastOff() method. This does two things

If no one else objects, I have no objections. Then this PR is "on" again.

walterar:
It is difficult to give an opinion without seeing the actual code that is not working.
There may be many ways to do it. Or just this.

walterar:
AIFlyTo() has changed long ago. Not working properly today, in TradeShips.lua. He is to blame for the ships into hyperspace jump on the platform of the space station surface.

clausimu:
@walterar I don't think I understand your last comment?

walterar:
@clausimu Show me the code does not work, there may be another way to do it.

clausimu:
@walterar Sorry - if you mean the actual C++ code I didn't work on that, @impaktor did.

But to reproduce just fly your ship within 15 km of a ground station and land on the planet. Then use the lua console to AIFlyTo(that_ground_station). Your ship won't lift off. Once you pass the 15 km mark the ship will lift off and fly to the station.
If you do the same with AIDockWith(that_ground_station) your ship won't lift off at any distance.

I need this functionality for the search&rescue mission script I'm working on. I can get ships to fly where I need them - but it is clumsy once they have landed (not docked!). I would have to first choose a target that is > 15 km away and use AIFlyTo() to lift off the ship. Then I can AIDockWith() any station.

Maybe I should open a merge request for this script - even though it is not ready for prime time yet. That way smart people (like you) can look at the code and provide critique.

walterar:
Try something like this:

shipx:AIEnterLowOrbit(Game.player:FindNearestTo("PLANET") or Game.player:FindNearestTo("STAR"))
  Timer:CallAt(Game.time + 5, function ()
    shipx:AIFlyTo(target)

clausimu:
@walterar I see what you are doing - nice idea with the timer function. But I still think a more "natural" way would be 1) blast off, 2) fly to. One thing we will have to work at in the future is to make AIBlastOff() aware of surrounding ships. I have noticed that with highly increased traffic I get accidents between departing ships crashing into landing ships during launch. Separating this blast off phase into a separate function might make adding "safety" functionality later easier. 

But I'm a total newbie here so I am very willing to be overruled. :)

walterar:
In this case, the decision is Impaktor. If you go ahead, I suggest, to avoid confusion, the name TakeOff, instead of Blastoff, because we have an event onShipTakeOff

walterar:
Although, maybe it easier and coherent with C++, change onShipTakeOff by onShipBlastOff. No one is using this event, today.

fluffyfreak:
Is this PR active or dead? 
If it's active then what needs doing so that it can be merged, or is it ready now?
If it's dead then can I close it and we'll address it's issues another way?

clausimu:
I guess that depends on @impaktor. I think it is nice to have and makes AI launching more natural and easier to script. Things work as they are now - but you need smart tricks like what @walterar is doing (1. send ship to low orbit with a timer active, 2. once timer runs out change target to where you really want to go).

johnbartholomew:
I'm sorry to come in with a dissenting opinion after all the discussion has
happened, but adding an AIBlastoff method doesn't seem necessary. If you
tell the autopilot to fly to somewhere other than your current location,
then it should take off automatically. There have been some clear test
results reported in this thread showing that it doesn't always do that. It
sounds to me like we should just fix that bug, rather than trying to work
around it.

The other reason I'm somewhat against this is that Blastoff doesn't really
fit with the rest of the autopilot's commands. The other methods give the
autopilot some goal, but Blastoff isn't a goal, it's just an action.

walterar:
" It sounds to me like we should just fix that bug, rather than trying to work
around it."

@johnbartholomew  That bug, years ago is among us and nobody was interested in solving it.

fluffyfreak:
No-one know about that bug, if you know about a bug, add it to the issue page.
No-one here is psychic or knows about all of the bugs, which is why it was such a surprise when it was discovered in the message above!

impaktor:
What bug?

The code behaves as expected. FlyTo flies to the vicinity (15 km) of whatever you give it. If you land next to a station (< 15 km) and thell it to fly to the station, then there's nothing to do.

FlyTo _is not_ the same as Dock.

I don't see a bug here, or am I missng something?

walterar:
I said in every possible way for me, but your rules are not mine.

PR #3463, opened 2015-09-23
Systeminfoview target box

joonicks:
Only show target box for
a) current system targets
b) remote hyperspace targets if more than one

Does not show anymore target around a remote systems single star.

PR #3483, opened 2015-10-07
systeminfoview target box oct7

joonicks:
removes the green target box in systeminfoview if there is only 1 valid target to set

### DIFF
--- a/src/SystemInfoView.cpp
+++ b/src/SystemInfoView.cpp
@@ -19,6 +19,10 @@
 #include "Factions.h"
 #include <functional>
 
+#define CANSELECT_ANY		2
+#define CANSELECT_STARS		3
+#define CANSELECT_NOTHING	4
+
 SystemInfoView::SystemInfoView(Game* game) : UIView(), m_game(game)
 {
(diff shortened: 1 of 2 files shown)


fluffyfreak:
This is the replacement for #3463? I'll close that one.

fluffyfreak:
Ok I can see why you might think this is a good idea but don't think of it from a coder perspective, think of it from the players.

They pick one star system and can click on things and a green box appears around them, they go to another system and click on things and... oh wait now there's no green box? Why can't they target this star? Does that mean then can't go there? Maybe they should just go to that place where they can target the star? Oh no wait they can still go there it just won't let them click it... that seems like a bug they'd better report it...

It lead to an inconsistent user experience.
I've just done this in fact, checked-out, built and run your code, it works but it's confusing even though I knew what to expect.

You don't seem to like hearing alternatives but here goes:
What might make sense is to always automatically highlight the first star in any system you're looking to jump too, perhaps with a different coloured highlight over the other possible jump targets for multiple stars in a system.

Either way I'm closing this particular PR.

PR #3484, opened 2015-10-07
expose table mouseover hover color to lua interface

joonicks:


fluffyfreak:
I like you're commitment to doing things but could you start posting descriptive message about what your doing and why you're doing them?

Exposing the mouseover hover colour to lua? I ... just have no idea why you would do that so I can review the code, maybe test it, but I don't know what I'm really looking for or why you would ever want it so I'd be inclined to close the request without knowing more.

If you look at other PRs like #3453 we tend to provide a description of what something does, and why we'd want to do that. Actually #3453 isn't that great an example but that's because it has a link to an entire forum thread about the why and with more technical detail... which is an acceptable substitute.

joonicks:
I noticed the mouseover hover color is pretty much hardcoded into the theme (black, 60% alpha), I tried to change it once while working on some lua gui things and found there was no way. so I found this "fix" pretty easy to do and it worked without much fuss. yes exposing the theme mouseover color might be a bit more productive but this change might still be useful down the line.

go to ui/StationView/CommodityMarket.lua, line 75, and add
:SetHoverColor({ r = 1.0, g = 1.0, b = 1.0, a = 0.5 })

then start the game, go to the commodity market and move the mouse over the list of commodities and you'll see the color has changed.

robn:
The Lua UI is entirely controlled by Lua, including it's colours. For UI elements, that's done in the widgets file. For most colours it's currently hardcoded in places (eg the top of `ui/TabView.lua`), but that should be moved out to the skin config (easy to write, annoying to change) or to a palette swatch in the widgets image (better for designers, harder to implement, but totally cool).

Any aspects of the old UI should never be exposed to Lua (unless, for some bizarre reason, you need it in Lua-driven game logic), Entangling the two UI systems is possibly the thing that would kill off any remaining slender hope that the old UI will go away.

robn:
I had a chat with @joonicks in IRC just now. I misunderstood what this was doing from the PR comments here (I initially didn't read the code) - I thought this was bringing out old UI color values to the new UI ("expose" and "hardcoded" were the main words that threw me).

I apologise for jumping to conclusions without checking properly.

To the actual feature, the main driver for this was to make it possible to change the hover colour in the future because someone might want that. There's no actual feature in development that requires this. So this change is introducing a maintenance burden.

It's also not the right way to do it. Internally, `DrawRectHover` is intended to draw a rectangle "suitable for hover effects". The details, such as colour choices, shapes, etc, are not supposed to be important. The idea is that you will always get something that matches the rest of the UI - that's what the skin setup is intended for. Yes, you don't get fine control, but that's not what you want most of the time.

If it's desirable in the future to modify the normal/hover/select colours globally, that should be done by extending those values in the skin to a full RGBA value and using them in `DrawRectNormal`/`DrawRectHover`/`DrawRectSelect`. If a particular feature wants to be able to modify them individually, that should be done with mouse event handlers, custom widgets (Lua or C++) or some other mechanism.

That said, you always want to think really carefully before diverting from the design provided from the skin. Use of colour is always a great example; changing them can make things hard on visually-impaired users. Not that I'm claiming for a moment that our current skin is great for that, but the UI system was designed to make it easy to uniformly apply a particular style to the entire UI.

(Above is actual design stuff that I don't think I ever wrote down anywhere. Sorry.)

Issue #2108, opened 2013-02-24
Check random number generator


robn:
```
<jpab> I'm a bit concerned by the new random number generator
<robn> jpab: anything specific, or just gut feeling?
<jpab> specifically, I'm concerned by it only having 32 bits of state
<robn> so we need to stress it or something?
<robn> I'm no good on rng theory
<robn> I could revert it
<robn> it just makes startup so damn fast though :)
<robn> factions really hurt
<jpab> nah, don't revert, I'm just putting a marker in that I think we'll have to revisit it
```

impaktor:
## Revisit, anno 2021
Just doing some necro posting, to drop a link and summarize current state of random number generation issues in pioneer by: 
1. we've merged "Use PCG for random number generation" (#3485, which is where we have the main discussion on RGN), 
2. We might also have this: "Investigate the modulo bias in Pioneer" #4036
3. New link: [Random number generators for C++ performance tested](https://thompsonsed.co.uk/random-number-generators-for-c-performance-tested)]. I dunno about  Xoshiro256+, but looks impressive by these charts:

![](https://thompsonsed.co.uk/wp-content/uploads/2019/03/prng_perf.jpg)

The article also has some links for more reading on random number generation, in the last paragraph.

PR #3254, opened 2014-11-21
Law and order to lua

impaktor:
# Move Law to Lua

This moves the law enforcement from `Polit.cpp` (which is almost empty now) to lua. I worked on this about 6 months ago, but I got stuck due to a segfault when loading a game, but it seems to have fixed itself now (after updating branch to master).

__Update:__ as of 2015-08 I'm continuing this due to #887 being "fixed". 

## Changed/New features

Some of the changes in how police will behave now:

1. Police number, and how quickly they react to your crimes (police launch probability) depends on system lawlessness.
2. However, police will not launch until player has amassed high enough fines. I.e. threshold isn't `fine = 0` anymore, but `maxFineTolerated=1e3`
3. Lowered fine on piracy to `1e5`, and weapons discharge to `1e2`, allowing the latter to be below the `maxFineTolerated` level.
4. Dumping illegal goods or goods with negative price (close to station) is an offense.
5. Crime record now lists _how many_ of each crime type the player has.
6. Player has two crimerecords for each faction, one with "outstanding fines", and one "Crimerecord" with past history.
7. A police ship destroyed at a station remains so until player hyperjumps back (same as before I believe); i.e. limited number of police ships in station.
8. Police ships are different for different factions. Default: sinonatrix, CIW: pumpkinseed, SolFed: kanara.

## Design / Code

I'm not sure about what code to put in which file, i.e. how to structure it. Current structure reflects a lot of the development history during my work on this.

- I've introduced a `libs/Legal.Lua` that holds the crime constants, and a `notifyOfCrime()` which was previously in `Polit.cpp`.

- I've introduced a  `modules/CrimeTracking` to be symmetric with our `modules/StatsTracking`. It has the event triggers (`onShip<ActionX>`) for killing, firing, jettison, etc. and it holds the `doLawAndOrder()` (previously in `SpaceStation.cpp`).

- `libs/Player.lua` has a local CrimeRecord class, which the player has two copies of for each faction (current crimes, and past crimes). There are also Player:AddCrime() and such.

-  In the old (current master) code, each space station upholds the law (every frame) by calling the function DoLawAndOrder. I could implement it the same way in `libs/SpaceStation.lua`, but thought it needs only be called _once_ (about every 5 seconds), at which point a check is made to find the closest space station (calls a new method: `Ship:GetLawEnforcingStations()`, returns stations within "law range"), thus you will not be fined twice for one single illegal action done within range of both stations. (I still have to check the distance of the player to all stations every call.)

[...]

Issue #790, opened 2011-11-10
Custom system specification


Brianetta:
Custom systems could be easier to define. In particuar:
- It should be possible to omit values, and have the engine fill in the blanks
- It should be possible to specify more values:
  - The type of space station
  - The orientation of a space station (for example, locking a door to face down to the planet, or perhaps into the prograde direction)

PR #3485, opened 2015-10-07
Use PCG for random number generation

fluffyfreak:
# Description:
Since I first heard about [PCG](http://www.pcg-random.org/) I have been meaning to give it a try with Pioneer. The maths for understanding exactly how robust this is are a bit beyond me in some ways however it is supposed to be a very fast and very robust random number generator with a very long period and many other fine qualities.

# Motivation:
We used to use a Mersenne Twister RNG but it was deemed "too slow" at some point (#1893) and replaced with a faster but 'less secure' XOR based RNG along with the comment: `As an xorshift generator the worst thing about it is it gets stuck on zero input`.

Recently we've had a lot of comments about objects, seemingly along the sector 0-axis, not being generated correctly, issues with the galaxy being regenerated incorrectly, and the startling similarity between characters at stations. 

So I picked up the `pcg-cpp-0.98` implementation, fixed the VS2015 issues and whipped this up to get us started.

# Results:
It still works.
I don't notice any speed difference personally but I would bet that it's slower than the XOR rng.

# Todo:
I haven't finished yet, although the actual code is in and the Random class uses it and works I haven't done the following:
- [x] Fix Linux build
- [x] Test
- [x] Bump save-game version

Andy


impaktor:
So to solve this problem, described here:
dilbert

My suggestion would be something like:
xkcd

No, but seriously, this builds fine on GNU/Linux for me. I haven't tested it any further than this.

Also, does this close #2108? (which links to http://xorshift.di.unimi.it/)

fluffyfreak:
Essentially yes :)

I'd forgotten about #2108 but hopefully it will fix some of those issues. It would be nice to get better brains than mine to look at it!

fluffyfreak:
ON my PC this turns the current startup time of : `29339.304202 milliseconds` into `27841.688087 milliseconds` ... which is nagging at me as not-quite-right since it was reporting `34000.etc` a few minutes ago.

What I'm saying I guess is that the difference is negligible and is probably more affected by HDD performance than RNG.

impaktor:
I know of another PR with a savegame bump. Would be practical to merge them at the same time.

Please don't finish this too soon. :wink: 

> It would be nice to get better brains than mine to look at it!

Your brain seems pretty fine to me (grey and mushy I assume?). But more eyes doesn't hurt, maybe @johnbartholomew would care to look at it some?

johnbartholomew:
I have no strong feelings one way or the other

I don't have anywhere near enough statistical knowledge to judge the relative merits of these PRNGs. If it fixes a bug with generation of bodies along the zero axis then that seems worthwhile.

joonicks:
my opinions, if anyone wants them;

startup time, a better way to do it would be to compile original and modified, do 10 startups with each, discard the first two of each and do an average on the remaining eight.

breaking the galaxy continuum, hey, whatever, Pioneer is still Alpha, if you want continuioty - dont upgrade. otherwise I think devs are free to break the galaxy every day if they wanted to.

merge - test - analyze. more testers can only give better end results (even if it means reverting)

fluffyfreak:
OK, bumped the save version, flown around a bit, landed at various places, looked at ships on offer, BB, etc.

No problems as far as I can tell.

impaktor:
Btw, this changes the starting station type in London. The reason we switched from LA to London was to get the nice octagonal stations as a first player experience.

So either change starting station on Earth, or #790, or we are not as excited by the octagonal stations anymore. I have no strong feelings one way or the other, I'm just pointing it out.

impaktor:
Also, does this require any testing on Mac?

fluffyfreak:
This should produce the same output in each platform, though you are right that I haven't tried it on that due to the existing issues with building on OSX.

impaktor:
OK, I just thought I should mention it.

I do wonder if there was some way to release the next savebumping pioneer as "unstable/test"-version? On SSC only?

fluffyfreak:
Well built and tested using the cmd line on OSX and it works fine :)

How would I identify the start location for Los Angeles on Earth? It uses the Octagonal station.
At some point we must get rid of the 6 pad station that I hacked together!!!

impaktor:
> How would I identify the start location for Los Angeles on Earth?

 82075ea30e179d96172b66dfd5d8905729383bc7

fluffyfreak:
Yeah I did it by looking at the ordering of the stations in the system info screen... and counting up from London to LA :)

impaktor:
Does this branch conflict with something?

@fluffyfreak I've written up an "announcement"-post for SSC, where I blatantly assume we've merged this PR as well as the following:

#3498
#3505
#3254
#3504
#3503

I'm feeling in the mood to get this merge started, and throw the players in the deep end, either now, or tomorrow. Hm, either way, our build bot robn wont be able to start anything now, seeing how it's in the middle of the night over there.

@fluffyfreak what was your plan? To test and merge today? If not, are you fine with me merging everything tomorrow?

fluffyfreak:
It was conflicting with an unchanged line in `Game.cpp` for some reason, odd.

I was planning on testing and merging but today didn't go to plan :) I'm fine with you doing it tomorrow evening.

impaktor:
OK, I've started merging now, will test all on my local system before I push.

Hold on to your hats people!

PR #3486, opened 2015-10-07
exposing full color values instead of only alphas in skin.ini, getting rid of hardcoded black colors

joonicks:
used where;
normalcolor -- unknown
hovercolor -- table rows (like bbs, commodity, skips and equipment markets)
selectcolor -- unknown


robn:
From a quick skim of the code, I'm ok with this :+1:

PR #3488, opened 2015-10-08
GLSL preprocessor includes

fluffyfreak:
# Description:
This is a primitive, slightly hacky even, `#include` preprocessor for the shader files.
Currently when building a shader we add the other included glsl files for attirbutes, logz and library methods in hardcoded calls even for shaders that will not use them.

In future therefore when we add GPU noise shaders and their support functions everything would include those as well.

Supporting a simple `#include` macro means that each shader can include only what the shaders author wants including.

# This PR:
I have added the necessary `#include` to each file, probably unnecessary ones too as I just copied and pasted the three usual suspects into each file.

I've been a little lazy with the filename construction and used an `sprintf` which I should probably tidy up but it's late and I couldn't think straight about doing it with Pioneers string stuff.

Comments welcome as always.

Andy

### DIFF
--- a/data/shaders/opengl/FresnelColour.frag
+++ b/data/shaders/opengl/FresnelColour.frag
@@ -1,6 +1,10 @@
 // Copyright © 2008-2015 Pioneer Developers. See AUTHORS.txt for details
 // Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
+#include "attributes.glsl"
+#include "logz.glsl"
+#include "lib.glsl"
+
 in vec3 varyingEyepos;
 in vec3 varyingNormal;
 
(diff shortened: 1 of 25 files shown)


impaktor:
> Comments welcome as always.

I wish I had something to say, other than "I need to learn shaders".

fluffyfreak:
@impaktor there's some good courses lying around and I can take you through any of our shaders and the system behind it if you like? Maybe I should make some time to write something on the forum about how it all works in Pioneer.

impaktor:
Although graphics isn't my main interest, I'm always interested in learning. I have (since yesterday) two tabs open in my browser with shader tutorials, for some nice weekend reading. So thanks for the offer, but don't spend too much time on it, since at least I, am not able to go too deep into it, at the moment. 

Documentation is always helpful, if not for me now, then for someone else in the future.

Issue #3477, opened 2015-10-04
hyperspace range is bugged


joonicks:
![hyperbug](https://cloud.githubusercontent.com/assets/14345157/10269419/8a3d9aba-6ad7-11e5-947a-03e96851aa2e.png)
According to the formula in equipment.lua, my hyperspace range should be
625*7*7/4280 = 7.16ly
its not just a visual bug, it allowed me to make a 8.58ly jump without complaints

joonicks:
issue comments are bugged, I wrote 620 star 7 star 7 / 4280 which comes out 625*7*7/4280
meants 625 times 7 times 7 divided by 4280

joonicks:
found it, ship.totalMass does not include fuel (for deltav calculations I guess), but it should be included in hyperspace calcs. however I dont know if its safe to call Game.player.fuelMassLeft because calls can also be made by tradeships?

impaktor:
This issue is fixed by 3489 -> closing. 
Further discussion on hyperrange is here: http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=355

Thanks for reporting (and for pointing out how to fix it)

Issue #3005, opened 2014-07-03
hyperspaceRange appears to be broken


walterar:
![screenshot-20140702-214826](https://cloud.githubusercontent.com/assets/1334989/3464402/0b1ff2fe-024c-11e4-9a18-3c1e9c21b194.png)

PR #3489, opened 2015-10-10
Account for the fuel mass in hyperspace range calculation.

fluffyfreak:
This should fix issue #3477 where the jump range was miscalculated as it did not account for the mass of the fuel aboard the ship.

Also takes into account the fuel for the hyperspace duration.

Changed `totalMass` to `staticMass` since it represents the unchanging mass of the ship rather than the dynamic mass of the fuel which will change through usage.
I have kept the old value of `totalMass` around for backwards compatibility with mods and forks of Pioneer like Scout+/Paragon/etc.

Thanks @joonicks for finding the issue and suggesting the fix.

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -190,7 +190,7 @@ end
 local HyperdriveType = utils.inherits(EquipType, "HyperdriveType")
 
 HyperdriveType.GetMaximumRange = function (self, ship)
-	return 625.0*(self.capabilities.hyperclass ^ 2) / ship.totalMass
+	return 625.0*(self.capabilities.hyperclass ^ 2) / (ship.totalMass + ship.fuelMassLeft)
 end
 
 -- range_max is as usual optional
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -190,14 +190,14 @@ end
 local HyperdriveType = utils.inherits(EquipType, "HyperdriveType")
 
 HyperdriveType.GetMaximumRange = function (self, ship)
-	return 625.0*(self.capabilities.hyperclass ^ 2) / (ship.totalMass + ship.fuelMassLeft)
+	return 625.0*(self.capabilities.hyperclass ^ 2) / (ship.staticMass + ship.fuelMassLeft)
 end
 
 -- range_max is as usual optional
 HyperdriveType.GetDuration = function (self, ship, distance, range_max)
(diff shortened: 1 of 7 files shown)


impaktor:
I was first worried that fuel should be included in `totalmass`. But this satisfied me:
```c
/ *
 * Attribute: totalMass
 *
 * Mass of the ship including hull, equipment and cargo, but excluding
 * thruster fuel mass. Measured in tonnes.
 *
 */
```

joonicks:
I didnt want to submit this simple patch because;

5 lines down totalMass is mistaken again to mean the whole ships complete and total mass, this time in calculating the time it takes to hyperspace. but hey, who will complain that hyperspace jumps take less time than they are supposed to?

in frontier, the hyperspace range calculation is 300 * class^2 / mass, in pioneer its 625 * ... for some reason?

as a programmer the variable name "totalMass" implies to me. total complete nothing excluded mass. a comment hidden in source code is not going to help me much. source is not distributed with the game, but lua code is. bad design, I would rather either change the variable name to totalMassExcludingFuel or change the C++ code so that totalMass really is totalMass not partial mass.

fluffyfreak:
Updated the opening post with description of what else I've just done.

@joonicks Pioneer isn't Frontier ;)

joonicks:
I know its not frontier but shouldnt there be some design choice rationale in doubling the hyperspace range or is it just a "meh. felt like it" ?

impaktor:
> a comment hidden in source code is not going to help me much.

Actually, it's in the official [API documentation](http://eatenbyagrue.org/f/pioneer/codedoc/files/LuaShip-cpp.html#Ship.totalMass), which is compiled from the source code I pasted above.

> but hey, who will complain that hyperspace jumps take less time than they are supposed to?

Could you _please_ stop with these passive aggressive accusations. We are also real human beings. I don't want to hesitate and feel sick every time I see you post something (as I do now), because I know there will be snarky and condescending comments. Should I remind you that you [just complained](http://spacesimcentral.com/ssc/topic/5250-contributions-unwelcome/) that we hadn't fixed the bug, and now that it's being fixed you're still complaining? Please think _long_ and _hard_ what use it is to post comments like that.

walterar:
This definitely fixes #3005 :+1:

fluffyfreak:
Hi @walterar, I should have known you'd already have found this issue before :D

PR #3490, opened 2015-10-11
Bad Terrain, no cookie!

### LINKED ISSUE #3441: Ground station landing and collision missing on Qulia Chou's grave (11 0 0)
[Marcel stumbled upon a ground base in Scout+](http://spacesimcentral.com/ssc/topic/4211-things-related-to-scout/?p=54916), where it's impossible to land. It happens in Pioneer too.
He mentions different station names from what you see in Pioneer and there's only one, but that's most likely because they are generated differently.
You can't land on the ground base Mullen Station on Chou's Grave in the Qulia system (11 0 0). And you can't even collide with the base, or anything it seems. Or with the planet itself.
Enabling the collision mesh display for the station and the ship shows them properly.

[Here's a save file, so you don't have to hunt and travel to it.](https://www.dropbox.com/s/3sbr4hk0xo7i22v/mullen_base_2?dl=0)
Made on windows, build: 20150810 (0bd8149)
screenshot-20150822-235759

The strange thing is that the station, and even the city buildings stutters back and forth, while you are hovering at 0 m/s. Quite visible when you turn up time acceleration, and a bit similar to when you leave the reference frame of an orbital sometimes.
What I noticed that the planet seems to rotate and orbit it's star quite fast.
When you switch off the autopilot (I was landing with it at first), you instantly get a 60 m/s speed.  If you switch the frame of reference to the station, and set speed to 0, the speed indicator tells you you are moving at 61 m/s relative to it, while standing still above it. If you switch the frame to the planet, the station instantly starts to go away from you as the autopilot brakes for a bit, and then stops.

And I noticed that the longitude of the station is constantly changing, while you are hovering above it, and the latitude remains the same. Like if the station/city was orbiting the planet at ground level or something, but it's hard to tell if it's really moving, because the area is quite featureless. Both with Very High and Very Low detail settings. It doesn't seem to be moving away from it's position, when I was watching it from orbit, but it's hard to tell.

The planet is not much larger or heavier than Earth (1.5x, 1.2x), and it has a day length and orbital period of 8.1 days, so it's tidally locked. It orbits it's star, a brown dwarf at a distance of about 5 million km or 0.03 AU.

You can dock with Kennedy Station and the other stations on the other planets, in an even tighter and faster orbits, and are  also tidally locked but the speed or latitude thing I described doesn't happen there. Also you can dock with the station orbiting the planet.

fluffyfreak:
This was a pain to diagnose but this PR should fix #3441

Unlike all of the other terrain generation values the final value of `n` was being calculated like this:
`n = m_maxHeight*2.0*n*n;`

Which seems to have been producing values which caused us NOT to transition to the correct `Frame` as we approached the planet.

The planet is still as flat as a pancake though, it's very odd and I'm wondering if the fractals are behaving as expected or if something is broken with them.

### DIFF
--- a/src/Body.cpp
+++ b/src/Body.cpp
@@ -190,9 +190,9 @@ void Body::OrientOnSurface(double radius, double latitude, double longitude)
 
 void Body::SwitchToFrame(Frame *newFrame)
 {
-	vector3d vel = GetVelocityRelTo(newFrame);		// do this first because it uses position
-	vector3d fpos = m_frame->GetPositionRelTo(newFrame);
-	matrix3x3d forient = m_frame->GetOrientRelTo(newFrame);
+	const vector3d vel = GetVelocityRelTo(newFrame);		// do this first because it uses position
+	const vector3d fpos = m_frame->GetPositionRelTo(newFrame);
+	const matrix3x3d forient = m_frame->GetOrientRelTo(newFrame);
 	SetPosition(forient * GetPosition() + fpos);
 	SetOrient(forient * GetOrient());
 	SetVelocity(vel + newFrame->GetStasisVelocity(GetPosition()));
@@ -217,7 +217,7 @@ void Body::UpdateFrame()
 
 	// entering into frames
 	for (Frame* kid : m_frame->GetChildren()) {
-		vector3d pos = GetPositionRelTo(kid);
+		const vector3d pos = GetPositionRelTo(kid);
 		if (pos.Length() >= kid->GetRadius()) continue;
 		SwitchToFrame(kid);
 		Output("%s enters frame %s\n", GetLabel().c_str(), kid->GetLabel().c_str());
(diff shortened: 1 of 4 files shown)


impaktor:
cookie_monster

But a big :cookie: cookie for you, for solving it!

jaj22:
I suspect the flat surface is caused by the low starting frequency and amplitude feedback in the barren rock fractal. Essentially the fractal is far too dependent on a handful of random values. 

This is a general risk with any fractal that uses a low starting frequency and noise-based amplitude. The non-fracdef functions all have frequency 1.0 on the first octave, so any fractal that uses those is a potential problem.

fluffyfreak:
@jaj22 ah I see, it's a bit disappointing that we're still using the non fracdef versions too since it makes all of those worlds use exactly the same inputs.

I think I'll have to start another PR (or many) to update them.

Issue #2337, opened 2013-06-16
Falling through terrain


Tichy:
This happened to me a couple of times in different places.
The first time was on New Hope, trying to land at the top of a mountain near Epsilon Cove. Unfortunately, I didn't took the exact coordinates.
Last time was in NN 4201 b (-3, -4, 2). You can see the exact plase in this two screenshots.
![screenshot-20130616-131217](https://f.cloud.github.com/assets/2348771/659248/e10f945c-d677-11e2-886e-25676eac9739.png)
![screenshot-20130616-131426](https://f.cloud.github.com/assets/2348771/659249/e12cca54-d677-11e2-8d5f-b116a679fb18.png)

I found another place on the same planet. This is a savefile of it https://www.box.com/s/538yq64rjpjf41l241qp

All you have to do is load, wait for the terrain to stop rendeing, unpause the game and let the ship fall.
My terrain settings are all to the maximum. I don't know if it's relevant.

In Veedeth A c (-5, -5, 3) I was unable to land in every places that I tried to. First, I tried a couple of times upon a mountain, then I moved to a more flat terrain, but without luck.

Three videos showing the issue:
https://www.box.com/s/g8kqm4oohklidop2sugb
https://www.box.com/s/l90s1447e8eub0hgau16
https://www.box.com/s/9bs8n673ipmof7ettbdj

PR #3491, opened 2015-10-12
Generous Minimum PhysRadius.

fluffyfreak:
Set a more generous minimum phys radius around planets so frame transitions happen at greater distance from the surface.

I was finding that transitions for some worlds were happening exactly at, or just below the height of the tallest mountains making landing on them manually a nightmare.


jaj22:
Same fundamental problem as #2337, right?

The maximum feature height calculation in TerrainBody::InitTerrainBody() is too low for most fractals. For persistence 0.5, a factor of 2.0 on m_maxHeight is correct, but the code uses 1.0. I'm not sure where to put this factor though.

More generally, you can calculate the maximum feature height of a fractal by adding up the amplitude of each octave, so for example a 4-octave fractal with 0.4 persistence would have a maximum feature height of 1 + 0.4 + 0.4^2 + 0.4^3.

I'm not sure that 1.05x sbody radius would work around the problem in all cases, as max feature radius+1km can potentially be higher than 1.05x sbody radius but still lower than the actual terrain. More likely to happen on very small planets & asteroids.

fluffyfreak:
frequency should usually be equal `frequency = 1.0 / persistence` if I'm understanding correctly. That could just be added to the `TerrainNoise.h` functions.

The problem with trying to estimate the maximum height from our fractals is that they often use branching logic to decide what fbm's they're going to run, then clamp, band and otherwise limit results from some things. So there's a lot of logic going on in some of them.

Still that might come in handy to know for the future :)

fluffyfreak:
In the case of `Quila 11, 0, 0, 0, 14` you have to be less than 1,300 metres from the surface before you make the frame transition. It has a really tenuous atmosphere almost regardless of what terrain function you use it is completely featureless.

I think that 1km minimum needs to be a much higher value because you can cross a kilometre so quickly it's insane.

fluffyfreak:
Thinking about it more that +1km gets fed into the phys radius, which then becomes the maxHeight for the terrain. So the bigger the planet, and `Chou's Grave` is about 7 times the radius of the Earth, the smaller a different that 1km will be in percentage terms.

fluffyfreak:
Seems that there's still a lot I don't understand about the terrain generation dependencies and maths, need to look into this more thoroughly.

jaj22:
I give up on trying to analytically determine maximum fractal heights. It's workable for some algorithms, but many of them are too complex, as you suggested.

In practice, the fractals don't seem to exceed maxHeight by very much. Also, as terrain heights are generally an order of magnitude lower than Earth-like atmosphere heights, there's a large window that should be usable without encroaching on low orbits. Something like maxHeight*2.0 should be (very) fractal-safe and still unlikely to cause any navigation issues.

The +1km value was chosen assuming that maxHeight was correct for the terrain, but that buildings could potentially rise above it. While that's apparently untrue, I think it's still a valid solution for very small (or very flat) bodies such as asteroids with stations on them. +2km might be safer.

A related issue is that the maxHeight values for heightmaps are broken. That's partly because of faulty code paths and partly because heightmap planets get a ton of extra fractal data on top. It's not a very visible problem because Mars and Earth have atmospheres and the Moon accidentally gets an algorithmic maxHeight that's fractionally higher than the heightmap+fractal. There are some marginal autopilot suicide cases.

Note that this problem shouldn't be confused with the control & instrumentation problem where set-speed functionality and speed readouts behave differently in the rotating frame. They shouldn't, IMO: Frames should be invisible to the user, where reasonable. Blending between body & surface velocity based on altitude is more appropriate. I'd be inclined to leave that side until the imgui HUD is merged.

tl;dr: Maxheight*2 + 2km should be fine for minimum rotating frame size and the ShipAICmd.cpp's MaxFeatureRad function. I'll add a fix for the heightmaps.

jaj22:
On second thoughts, 2*maxHeight may be too high for anything like Olympus Mons. Might need to do this properly and fix the fractals.

jaj22:
Second problem: 2*maxHeight isn't even high enough for TerrainHeightMountainsNormal. It's quite capable of exceeding 2.0, and I found a 1.92 pretty quickly. The main culprit is the giant chunk of river_octavenoise, with a 3.0 multiplier. The results from the early stuff are then cubed, so anything >1 blows up. The later divisors (*0.2 & *0.3) can't drag the result below 1.0.

The river_octavenoise is such a dominant feature that it's not fixable without changing the character of the fractal or rewriting it entirely. I suspect it was written "blind" and most of the fractals either have no effect or are not working as intended, so rewriting it probably wouldn't lose anything of value. Could do with a better terrain viewer first though: The current ObjectViewerView is only much use for the colours.

fluffyfreak:
I think everyone would be fine with the terrain generator being rewritten, we've had to do some drastic things to eliminate bugs in the terrain in the past.

The attitude is fix it now, replace the whole damned thing later :)

fluffyfreak:
@jaj22 ok so I switched to the `MaxHeight*2.0+2km` then I tried to fixup the `TerrainHeightMountainsNormal` a little whilst preserving most of it's character which I think that I've managed.

So I guess this is ready for review now.

PR #3495, opened 2015-10-13
WIP - Frequency from inverse persistence.

fluffyfreak:
Change the frequency based on the value of the persistence passed in.


fluffyfreak:
Going to close this since i have plans for terrain stuff, might as well leaved it alone until they're done and supercede this.

PR #3501, opened 2015-10-22
Eclipse tidy

fluffyfreak:
This moves most of the code for the Eclipse stuff into a separate `eclipse.glsl` library file that can be included with our `#include` macro.

Eventually I would like to change this to use [OpenGL Uniform Buffers](https://www.opengl.org/wiki/Uniform_Buffer_Object) but that's a more extensive rewrite and this is just a neat cleanup to make that future work easier.

All functionality should be exactly the same.

Andy


fluffyfreak:
Tested on Linux too, doesn't make the Eclipse code work magically, that's still dependent on MESA fixes I guess.

Merging.

PR #3504, opened 2015-10-28
Introduce faction specific fuel clubs

impaktor:
## Faction specific fuel clubs

Adds three new fuel clubs (we only had one):
- _The Federal Fuel Foundation_ 400 credits/y, (spawn rate in fed space: 40%)
- _The Confederate Fuel Union_ 300 credits/y, (spawn rate in CIW space: 40%)
- _Haber Propellant and Perishables Division_ 600 credits/y, (spawn rate in Haber space: 100%)

Now we can specify how common (spawn probability) each flavour is depending on faction.

The three new fuel clubs only exist in their own faction space. For other factions they get Interstellar Pilots' Fuel Co-operative (40%).

(In current master, it's Interstellar Pilots' Fuel Co-operative (30%) for all BBS's)

In some SolFed and CIW space, some stations can have two fuel clubs, the other being Interstellar Pilots' Fuel Co-operative. However, Interstellar Pilots' Fuel Co-operative spawns 0% in Haber space, because Haber doesn't allow competition.

## Review

It's good if the text is right, and doesn't need to be changed after merge. The different messages should "feel" like Federation or CIW, or Haber. Suggestions for improvements are welcome.

## Future

There should probably be some kind of list under F3 that shows player memberships, since the player can be member of several fuel clubs. Should also show time remaining on the membership.

### DIFF
--- a/data/lang/module-fuelclub/en.json
+++ b/data/lang/module-fuelclub/en.json
@@ -11,13 +11,49 @@
       "description" : "",
       "message" : "Interstellar Pilots' Fuel Co-operative"
    },
+   "FLAVOUR_1_CLUBNAME" : {
+      "description" : "Fuel club for the Federation faction",
+      "message" : "The Federal Fuel Foundation"
+   },
+   "FLAVOUR_2_CLUBNAME" : {
+      "description" : "Fuel club for the CIW (Commonwealth of Independent Worlds) faction",
+      "message" : "The Confederate Fuel Union"
+   },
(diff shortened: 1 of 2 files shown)


impaktor:
I had screenshots as well
screenshot-20151028-162904

screenshot-20151028-164959

screenshot-20151028-172335

impaktor:
(I label it as savebump, not that there's any save crashing potential, but it changes the pilot clubs on all BBS's and we have a battery of savebumping merges (like #3485) so, this neatly fits in with those )

Issue #3048, opened 2014-07-13
Placed inside another ship on landing pad


GrungeHead:
Sometimes when taking off I seem to get placed on a landing pad inside another ship.

If I notice in time I can wait for the other ship to take off before I do.

Otherwise I naturally "crash into" the other ship, am declared a pirate and then shot down by police.

Slayo:
![broken](https://cloud.githubusercontent.com/assets/11958747/7155634/9a697370-e3a3-11e4-9ab4-c502555a490d.png) [pic of two ships overlapping.  this is the type of station that has just 6 landing pads]

when i approach this type of station and the alert picks up a ship i begin to dread what is probably about to happen.  then when i'm directed to landing pad 2 i know for sure i'm screwed.  what seems to happen is, regardless what landing pad i touch down on (even if i manually land on another pad), i'm teleported to landing pad 1 which is occupied.  i'm guessing a tiny bit because i can't see if the landing pads are actually numbered, unlike the other type of station where the numbered landing pads are very obvious.  

everything is fine as long as i don't launch.  or come out of pause mode when on the normal (F1) view.  i could wait for the other ship to take off, but with active missions this is usually not an option.

atm i have so many active missions carefully chosen for maximum profit and reward i am just searching for a way to delete that ship from the game :p 

just to clarify: nozmajner asked "This kind of station?" [pic of other type of station].  No, not that type ;)

walterar:
This is another flavor of the same bug which also causes #3331

impaktor:
@Slayo Thanks for reporting, that helps. 

> atm i have so many active missions carefully chosen for maximum profit and reward i am just searching for a way to delete that ship from the game :p 

Try removing `data/modules/TradeShips.lua` ?

nozmajner:
Magagnarata had something like this. Not sure if it's the same, but it sure looks the same.
And happens with the Mola here too (from the third post).  Other ships come and go if you wait around for a bit, but these two rotated ones won't fly away ever.
I was able to land, and I got intersected with the Mola similarly to the above image. Undocking exploded/damaged the other ship, so the Police came after me.
I tried to land a Mola on a similar station, and it did it without problem, on autopilot.
https://www.dropbox.com/s/npp7xyxcrrtknh3/haxed_impossible_landing.zip?dl=0
[...]

PR #3506, opened 2015-10-29
Small time acceleration limitation fix

laarmen:
This was at first only a thing about fixing missing braces, but it ended up a real fix \o/

See the commit log for the details. I wouldn't mind someone else's opinion, specifically @fluffyfreak since I don't know how to test because I don't even get what's the angular velocity length thingy. (I mean I get each word separately, but the three together...)

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -306,48 +306,50 @@ bool Game::UpdateTimeAccel()
   		if (m_player->GetAlertState() == Ship::ALERT_SHIP_FIRING)
 			newTimeAccel = std::min(newTimeAccel, Game::TIMEACCEL_1X);
 		
-		if (!m_forceTimeAccel)
+		if (!m_forceTimeAccel) {
 
 		        // if not forced - limit timeaccel to 10x when other ships are close	  
-		        if (m_player->GetAlertState() == Ship::ALERT_SHIP_NEARBY)
-			        newTimeAccel = std::min(newTimeAccel, Game::TIMEACCEL_10X);
+			if (m_player->GetAlertState() == Ship::ALERT_SHIP_NEARBY)
+				newTimeAccel = std::min(newTimeAccel, Game::TIMEACCEL_10X);
 
       			// if not forced - check if we aren't too near to objects for timeaccel
 			else {
-			for (const Body* b : m_space->GetBodies()) {
-				if (b == m_player.get()) continue;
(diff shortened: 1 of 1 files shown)


fluffyfreak:
I think that it means the magnitude of the angular velocity, sometimes in vector maths Length & Magnitude are used interchangeably. Not sure what that value would actually represent in the real world though... because the angular velocity is how fast something is rotating around an axis yes? So it's magnitude/length should just be whatever that value is.

fluffyfreak:
Ah I think it might be the magnitude of the speed of rotation around the xyz axes. So each axis has it's own velocity, and this is the magnitude of them all together. Best as I can tell from the code.

Will try to review this and some other stuff on my laptop tonight :)

impaktor:
+1 for the name of the branch!

fluffyfreak:
looking at that code around [L339](https://github.com/laarmen/pioneer/blob/im_alive/src/Game.cpp#L339) I'm ... well I'm not sure what it does, it's testing the ships angular velocity against a threshold divided by the time acceleration factor.

So, if the player is spinning really fast then don't allow time acceleration? It would have been really helpful for someone to have documented that code when they wrote it so we knew *why* it did it :/ oh well.

I don't see anything wrong with the code itself and it does clean things up a lot in there :)

walterar:
I think this solves other important things too. But I need to test it several hours in real game to be sure.

I suspect that this area is responsible for many "rare" behaviors.

impaktor:
> I think this solves other important things too

Are any of those "important things" / "rare behaviours" documented here on the issue tracker? 

> But I need to test it several hours in real game to be sure.

If you're going to spend hours testing, you could merge in many of our active PRs into the same temporary branch and see how they work together? Specifically all the save_bumping issues (pink/purple label).
(I'll update my law-and-order branch today to remove merge conflict and fix some cosmetic detail in it)

walterar:
"¿Alguna de esas "cosas importantes" / "comportamientos raros" documentados aquí en el seguimiento de incidencias?"

Our old friend in #3048 and others.

laarmen:
Thanks, but I don't think this is related. I didn't even know we had this
angular velocity thing before yesterday.

walterar:
I think the problems of NPC ships stuck is related, somehow, with a check, or reset, than fail sometimes with accelerated time.

fluffyfreak:
Good code works, merge now!

PR #3507, opened 2015-10-31
Compact Use Equip Widget

fluffyfreak:
Separate the missile icons by a fixed size if using the compact scanner for equally compact equip usage page.

Inspired by seeing how fulgy it currently is, also fixed some indentation.


impaktor:
I have not inclueded this in The Big Merge, which is now complete, ff5a49ce6f639f6471ca371075bc90ba5e42c1fc.

But we might have time to squeeze this in before we get a new build out. I'm not sure what it looks like, so I'm compiling now to see.

fluffyfreak:
Load up a ship with four missiles and they'll just appear next to each other with regular spacing rather than stretching across the scanner space depending on how many slot you have.

You can toggle it with the compact scanner option.

PR #3508, opened 2015-11-02
More time acceleration plus increased time acceleration in hyperspace

corundscale:
An extra time acceleration level (x100k) and correspondingly increased  time acceleration while in hyperspace (x1M).
Should reduce the amount of forced downtime and hopefully defuse any further delta-v arguments.

Forcing player to stare blankly at the screen for extended periods of time doesn't accomplish anything and exploration vs realism shouldn't even be a trade-off.


nozmajner:
Wouldn't that break some physics? I remember somebody mentioned that the current max accel already on the verge of doing that, but not sure.

corundscale:
I can't be entirely sure (especially given that it's not the question of if, but how much), but I think it's really needed given that delta-v budgets are going down, while systems stay the same. NPC ships are a bit hard to observe with current trajectory fade, but nothing really jumped at me when watching them. 
It does seem to be a bit hard on the performance, though.

If something bad happens, can't the impact be alleviated by disallowing forcing max time acceleration?

laarmen:
Well, 100k means that if your velocity is bigger than 60m/s, you can get through a body 100km thick. I think it is a bit much. I'm pretty sure you'll have NPC ships that will go through planets when trying to land, for instance.

robn:
If it breaks anything it'll be the AI. From memory the math is different at 10x when chasing down planetsplanets, space stations and other intercept targets. If you make the timestep too large then they'll continuously miss until they rum out of fuel.

It can probably screw the combat AI too, but that's likely already broken at 10x.

corundscale:
Frontier dealt with similar issue by allowing rather generous margins for hitting all sorts of goals at high accelerations. We could have something similar, but guarded by some coarse acceleration and delta-v checks, and assume that NPCs that physically could have docked or intercepted their target, did. 

Also, how hard would it be to add simple sentinel along trajectory offset so that at given velocity and time acceleration, if a ship should have hit something between ticks, the sentinel would catch it?

src/ShipCpanel.cpp
```diff
@@ -280,12 +287,12 @@ void ShipCpanel::Update()
 	int timeAccel = m_game->GetTimeAccel();
 	int requested = m_game->GetRequestedTimeAccel();
 
-	for (int i=0; i<6; i++) {
+	for (int i=0; i<7; i++) {
```
fluffyfreak: I realise that you didn't write it originally but it would be nice if this and other hardcoded numbers used the values from the enum instead of '5', '7' etc. Would you mind making that change?

fluffyfreak:
Tag it as 'WIP' whilst we discuss it, I'll take another look at the code tomorrow and try testing it a bit.

I do think that at 100,000x time acceleration we could probably just skip all of the approach, physics and other stuff and just have a distance test. If less than distance then docked :)

Worth thinking about.

corundscale:
How about 'If less than distance and sufficient acceleration to stop in time and sufficient delta-v to not run out of fuel when doing so'? 

Anyway one thing speaking in its favour is that it is a straightforward extension of what was already in place and took minutes to write, so it should take seconds to revert if we (gah! >.<) can't live with it. ;)

laarmen:
As if it were that simple to revert a published change...

I might be biased here (well, more than usual), but I'm against this change for a simple reason (besides the technical difficulties raised earlier) : it would make space feel less big. That's one thing I've always loved about Pioneer, the "game of lonely space adventure". The point of making players stair at an almost blank screen for quite a bit of time is that it gives us a sense of scale.

corundscale:
Git revert. It reverts a commit by making a commit that undoes changes from that commit.

I definitely agree that space should feel big, but I don't think forcing player to sit minutes at a time with absolutely nothing to do is a good way to go about it. 
If there is a way to keep space physically (distances) and mechanically (reasonable delta-v budgets and accelerations) big, without forcing player to wait much more than they actually play it's a net gain.

Sure travel should take some time IRL too, but the way it is now is overdoing it and the time and resources it takes in game are much more important.

src/ShipCpanel.cpp
corundscale: I was considering using size of array, but decided to keep it close to the original (as sizeof()/sizeof() would make it muddled). Using enum value seems like a better idea than this although I'm a bit wary of this kind of things, because there is no way to ensure that given enum stays last or keeps its value. OTOH it would be better to declare the size of the array itself as something that isn't hardcoded and I definitely agree that magic numbers are bad so enum it is - I will make the change tomorrow.

impaktor: >  it would be nice if this and other hardcoded numbers used the values from the enum instead of '5', '7' etc. Would you mind making that change?

Then please do it as a separate commit (or PR) so we can cherry-pick it if this PR isn't merged.

corundscale: Done.

corundscale:
I have another suggestion regarding time compression - wouldn't it be beneficial to move to base 8 (if my PR is merged) or base 16 (if it is not) time acceleration multipliers? 

Both would yield floats with exact binary representations and might prove easier to optimize from compiler's PoV and possibly give stabler caclulations.

Should be just as easy to implement.

fluffyfreak:
If things work then I have no objection to trying this out in a release.
So I will test it, maybe make some small changes to things I've spotted.

Discussion about how to improve the time accelerated docking and motion of ships etc should go onto the forum unless we're actually going to try to do it all in this PR(?).

corundscale:
Things seem to work comparably to max current acceleration although there seems to be a considerable performance hit.

fluffyfreak:
This hasn't been forgotten but if you've seen the flurry of PR's and other stuff I've been doing I've been enjoying some actual developing rather than reviewing recently.

I'll get onto this asap, hopefully this weekend.

corundscale:
Thanks.

fluffyfreak:
Ok I got a chance to test this at work, on a fairly fast PC, I had no performance issues but holy-crap it removes 95% of the hyperspace and in-system travel times. 
Travel is just gone, completely, journeys to the very limit of my hyperspace range happen in a few seconds and crossing an entire system in less than 5 seconds.

The problem with this is that it doesn't match how we're used to Pioneer, there's no risk or reward in going to a big system and making long journey's. 
It will make a mockery of things like exploration (*something I've worked on and need to return too*), or rescue missions, etc. 
It doesn't fit the goal of *"A game of lonely space adventure"* if you can just whizz around and be at any destination in seconds, one even multiple jumps away.

Recently in Elite:Dangerous I went to [Hutton](https://forums.frontier.co.uk/showthread.php?t=63056) [Orbital](http://huttonorbital.com/) ([one more link](https://steamcommunity.com/app/359320/discussions/0/611702631245430782/)) without at first realising what a huge journey I was embarking on just to collect a bounty reward. Once I realised I had to jump out, strip down my ship, equip with more fuel and then I set off again for a flight that took 1.5 hours. What made it worth doing was the distance, the effort and **the amount of time that it took**.

So my opinion is that I don't want this change, it takes away from Pioneer as a game. However if you want to put this under a configuration option so that players can try it out for themselves (*off by default please*) then I would merge that and we can do a poll of peoples preferences later on as to whether we keep it or not.

nozmajner:
What if it's not a 10x step upwards for the new timeaccel but only a 2x? And keep the hyperspace step intact? (haven't tested, just based on the comment from @fluffyfreak)
Also this might be better discussed after we actually finished rebalancing the ships and their performances.

corundscale:
While I agree that Pioneer is supposed to be a game of **lonely space** adventure, it's also supposed to be a game of lonely space **adventure** - with all due respect, staring blankly and passively at the screen with nothing to do is anything but. 

The problem with Pioneer, as a work in progress, s that current content density and diversity is very low. Sure, "physical" content density is always going to be low (it's called space for a reason), but that doesn't mean there couldn't be a lot of things happening. 

Going somewhere being just a matter of zipping there might be a problem, but forcing player with nothing to do, to waste prolonged periods of time doing nothing is hardly a solution - the problem is the "nothing" part. 

There are many things that could be done to make travel a bigger deal:
* reducing the jump range - current jump ranges in Pioneer are about an order of magnitude greater than in Frontier. If they were reduced a journey from A to B wouldn't involve just selecting the destination and hitting jump but plotting the course through a number of intermediate systems, possibly getting ambushed, having to refuel, having all sorts of events requiring a reaction happen and so on. That might just as well take more real time than a single jump does now, but be actually interactive. That plus having a super-powerful jump drive doesn't really help make getting from A to B a serious matter, all it does is preventing player from having grasp of immediate surroundings as instead of containing a handful of recognizable systems their "jump bubble" contains a "soup" of hundreds of stars.
* reducing delta-v - this will make travel longer again and make orbital motion of celestial bodies more relevant to the gameplay, maybe even allow slingshots to have gameplay impact instead of being a nerdy pastime.
* implementing distance from planetary body before jump is allowed. I would make it much more restrictive than in Frontier - maybe 2 planet radii or more?
* tying throttle to exhaust velocity (for main engine). It's realistic as kinetic energy (your energy requirement) grows with square of velocity but momentum (what you actually want) grows with velocity, so you can achieve higher thrust by being wasteful. Then you can have stuff like getting into orbit with hold full of cargo, then having to refuel (because liftoff needs thrust) before getting anywhere, then gently accelerating a bit away, then jumping, then in-system travel (again low thrust, so minimum coasting), then orbital refueling, then touching down with cargo, etc.
* allowing taking more cargo/equipment at the expense of propellant by making internal tank only a fuel buffer, plus any dead space resulting from ship being odd shaped (wings, etc.) - voluntary delta-v reduction to take more cargo -> longer travel (also susceptibility to intercepts, and so on - gameplay, gameplay, gameplay!)
* allowing fuel allocation mechanics and giving player reason to take more elaborate course than beeline form hyperspace exit to single destination - more stuff happening and less delta-v per single leg of journey.

In short, if you want player to spend their time traveling, let them spend it **playing** the game, not looking at the game playing itself - respect players' time (apologies if I sounded patronizing, that's never been my intention even though I do have a fairly strong opinion on this issue and am strongly convinced that it is the right one ;)).

Lastly, the feature is already optional enough, I think - no one is forced to use any particular time compression level and it has no or close to no effect on game's mechanics.

robn:
You missed a bullet point:

* Have extra fun things to do during travel. Fighting attackers, reading the news, looking at trade prices, etc.

This is a hack around the fact that we lack content. I think that's the wrong way to do it. At this stage of development it's good to highlight where our content holes are, in the hope people will fix them. Without the travel, it's just the station screens and a bit of planet-side flying if you like that sort of thing (and yes, I'm massively exaggerating).

I'm also still really nervous about pushing the AI this hard.

So overall I'm a :-1: on this. Feel free to ignore that; I don't think my powers of veto exist anymore :wink:

corundscale:
>You missed a bullet point:
 >*   Have extra fun things to do during travel. Fighting attackers, reading the news, looking at trade prices, etc.

>This is a hack around the fact that we lack content.

Counterpoint:
If we have interesting things happening, they will pretty much always drop player out of time compression or force them to keep an eye on the game and keep time compression at manageable levels. Time compression in general is what happens when there actually is nothing to do and it is a hack around the inevitable fact that the vast majority of game set in actual space will be non-content (and whether you read magazines while traveling or at some other time makes little difference in game).

>I'm also still really nervous about pushing the AI this hard.

But since the AI is going to be pushed harder and harder if there is ever going to be sensible traffic around populated planets and sensible NPC behaviour around terrain then maybe this would be a good opportunity to optimize and smoke test it?
:)

robn:
Different kind of pushing. It's not performance I'm worried about, it's accuracy. The larger the timestep, the more likely it is to overshoot targets, fly right through planets, etc.

corundscale:
Yes, but if the game is to handle very large number of ships in system then having some working low accuracy mechanics for all those that are less relevant to the player is going to be necessary anyway, except for performance reasons.

walterar:
TIMEACCEL_100X is, roughly, the current limit with reasonable accuracy.

If you're able to work on the code to increase accuracy, the world is yours. :)

corundscale:
I was thinking more about low accuracy code that could still keep stuff behaving more or less correctly and catch major impossibilities (like flying through a planet).

fluffyfreak:
The trouble is that things are meant to take time, that's part of the game and the gameplay itself. 
If, as in this case, hyperspace is just a flicker and you can always just go one-faster and traverse a big system in 5 to 10 seconds then your not playing the game with it's limitations.

Time compression is a conceit to the fact that people would die of old age playing the game without it. However go to far and you're skipping the playing of the game itself. The travel is part of it, the delaying of arrival.

Having used it I don't like what it does to the game, I view the travel as being part of "Pioneer" itself.
I invest my time doing that, like playing a flight sim where there is nothing to do for a few hours after taking off... but that conceit of time compression to make it more game-like.

Hence why I'm suggesting the toggle-on/off option, at the moment I cannot disable the hyperspace time setting so it takes that away from Pioneer for me, and that option will always be sat there on the dashboard.

Games are as much about their restrictions, and limitations, as what they enable you to do. For the sake of a few minutes/seconds I feel this doesn't fit. So if you want it in Pioneer then I want it off by default on a toggle.

impaktor:
> implementing distance from planetary body before jump is allowed. I would make it much more restrictive than in Frontier - maybe 2 planet radii or more?

Yes, this was in the works. Just hasn't been done yet. (In part due to discussions on how/why/consequences of too early jumping). Should be straight forward to implement, please see #2399, once we have decided roughly how it should work. Biggest issue I think is how the player will know jumping is now allowed/safe (a LED on the dashboard?).

> reducing the jump range - current jump ranges in Pioneer are about an order of magnitude greater than in Frontier

This has also been decided to do, to at least cut range in half. Maximum range for the Earth start ship should be around 10 ly. Please feel free to nerf it, but in a way that big ships still retain a reasonable range (fully loaded big ship with good drive should still jump 10 ly I guess?)

> I do have a fairly strong opinion on this issue and am strongly convinced that it is the right one ;))

A lot of what you listed is stuff that should be in the game, but isn't because we don't have time to implement it. I wouldn't mind seeing those strong feelings you have diverted into writing code implementing those suggestions, like the two I quoted above.

> Lastly, the feature is already optional enough, I think - no one is forced to use any particular time compression level 

If you put a button in that's basically a "cheat" button, people _will_ press it, even if it ruins game play for them, and even if they know it does.

> This is a hack around the fact that we lack content. I think that's the wrong way to do it.

I agree, and I think jpab said the exact same thing a few weeks ago.

> Without the travel, it's just the station screens and a bit of planet-side flying

I agree.

> So if you want it in Pioneer then I want it off by default on a toggle.

I wouldn't even want this as an option, to be honest, as I think this goes too much against where we want to go with pioneer. As said above, fill the hole with content instead. It's not uncommon for free software project to end up with a million options, each being a solution of trying to please a contributor. I'd prefer if we could avoid that as much as possible, especially when we all seem to agree on this particular PR. It was an honest try, but this time I think it shouldn't be merged, beyond d7e5dce34e78b8e431cd05311fdbe164d20c4e8e. 

I hope you don't take it personal, we appreciate the attempt.

fluffyfreak:
Right, I am going to close this. I understand the arguments for, but they're mostly arguments about the other things that need fixing or changing about Pioneer.

So lets take this onto the forum and discuss the things that do need doing so that this PR is not required.

impaktor:
@fluffyfreak I could cherry-pick d7e5dce and push to master, if you want.

fluffyfreak:
@impaktor yeah can do.

@corundscale I have created a [thread](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=379) on the forum for us to discuss what we actually want to do. So we can agree before work gets done and anyone's nose gets put out of joint :)

corundscale:
I still think we need it, and I'm a bit disappointed that other than me and fluffyfreak flailing at each other ;)  there was little discussion regarding this feature, but I'm ok with that - if I'm right it will get revisited sooner or later, otherwise there will be no need to.

Although I really need to voice my opinion on one last thing:
 I disagree with calling it a cheat - yes, such "optional" features are indeed not truly optional and end up ruining the gameplay, but since time compression is supposed to be a purely metagame feature without impact on what happens in game, the only thing it lets player cheat themselves out of is some tedium - I know it's very literal minded of me but for me "gameplay" is what involves game being played, not stared at blankly. :P

And yes, I'll happily discuss the particulars of making travel better on the forums.

fluffyfreak:
Well @nozmajner @laarmen @impaktor @robn and @walterar all commented on this, that's pretty much everyone that ever discusses things here except when others get time from their jobs or poked because it's their particular interest.

Pretty solid turnout actually, and all against this particular way of doing it.

PR #3510, opened 2015-11-04
Fix CargoRun stuffing.

impaktor:
When landing first check what mission cargo should be unloaded,
before trying to load potential new mission cargo.

Should solve issue described at [SSC](http://spacesimcentral.com/ssc/topic/5282-announcement-big-update-adventurous-test-pilots-needed/#entry55869):

> 8. Almost forgot. It'd be nice if starports first unloaded all mission wares before attempting to load new ones. For example, if I'm due to pick up 5t of cargo on station A, but arrive there short a couple tonnes of space, they will say "come back when you have space". But I might be carrying mission cargo to that station A at the same time. So they unload me, cargo hold frees up, but now I have to undock/dock (and pay the fees again!) only to be able to load those 5t.

__NOTE:__ I have not tested this, other than in my head.

### DIFF
--- a/data/modules/CargoRun/CargoRun.lua
+++ b/data/modules/CargoRun/CargoRun.lua
@@ -768,8 +768,10 @@ end
 local onShipDocked = function (player, station)
 	if not player:IsPlayer() then return end
 
+	-- First drop off cargo (if any such missions)
 	for ref,mission in pairs(missions) do
-		if (mission.location == station.path and not mission.pickup) or (mission.domicile == station.path and mission.pickup and mission.cargo_picked_up) then
+		if (mission.location == station.path and not mission.pickup) or
+			(mission.domicile == station.path and mission.pickup and mission.cargo_picked_up) then
 			local reputation = mission.localdelivery and 1 or 1.5
 			local oldReputation = Character.persistent.player.reputation
 			local amount = Game.player:RemoveEquip(mission.cargotype, mission.amount, "cargo")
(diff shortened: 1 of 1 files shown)


clausimu:
How should this be addressed for mission cargo that is tracked by different mission scripts?

impaktor:
@clausimu It doesn't, it just makes the management of this module's cargo slightly smarter.

So worst case scenario, if you have another mission module trying to load cargo when you dock to a mission station for CargoRun, you might need to re-dock again once CargoRun has emptied its mission cargo.

clausimu:
Maybe it would be smart to insert a (2 sec?) timer before trying to auto-load any cargo at a station when scripting? That way unloading from other scripts should have completed. Un-dock/re-dock would seriously kill the mood...

impaktor:
@clausimu That would only help if the lua-event hooks are run in parallel, and not at all if serial, which I for some reason though they were, although I'll admitt I don't know. Either way, we're discussing something that can not happen in current pioneer since there is no other module adding cargo than this one, and if it were it would only be an issue in the very rare event of taking one of each mission to the same station.

clausimu:
Fair enough.

impaktor:
@robothauler so I thought you'd like to know I'm messing with your mission.

Also, there's a bug report for you it seems [here](http://spacesimcentral.com/ssc/topic/5282-announcement-big-update-adventurous-test-pilots-needed/page-2#entry55879):

> Also just noticed during the test, the guy who gives the mission "Go there, bring back N tonnes of X" also says "you don't have enough space" if I currently don't have enough space. I.e. he doesn't allow me to take the mission, even though the actual load is at another station. Does he expect me to fly empty? "Flying empty is a waste of space!" 

Do you have time to investigate it? Sounds like it should be a simple fix.

impaktor:
(StanPancakes PM-ed me that it works, and fixes the issue, thus merged)

PR #3513, opened 2015-11-05
Direct buffer population

fluffyfreak:
There are a lot of places where we use the `VertexArray` class as an intermediary buffer whilst building the geometry that we wish to display. 

This is really helpful in some places, such as string buffers, however it can come with some downsides such as:
* Rendering requires `VertexBuffer` objects which must be created, populated, rendered & destroyed per-frame.
* Needless copying, we fill one buffer then copy it straight to another.
* Missed optimisation opportunities, there's a lot of places where we could be using `PointSprites` or other ways of grouping billboard type objects.

There's probably more too, some of these go through multiple stages which could be applied more directly and thus more efficiently.

So far I've identified this lot to change:
- [x] Atmosphere rendering
- [x] Billboard class

Andy

### DIFF
--- a/src/BaseSphere.cpp
+++ b/src/BaseSphere.cpp
@@ -37,16 +37,14 @@ void BaseSphere::OnChangeDetailLevel()
 	GeoSphere::OnChangeDetailLevel();
 }
 
-//static 
 void BaseSphere::DrawAtmosphereSurface(Graphics::Renderer *renderer,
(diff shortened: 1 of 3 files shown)


fluffyfreak:
I'm going to do the other work in a future PR or this one is going to get messy, there's the whole way that the `SFX` class works which is just ... hideous and confusing so I'll keep this one small and simple.

PR #3515, opened 2015-11-05
Normal Mapping for objects (ships, stations etc)

fluffyfreak:
Updated (*optional*) normal mapping based on work that Luomu did several years ago.

To use normal mapping on an object just open it's `.model` file and add a line like `tex_norm NormalMap.dds` too it and of course the normal map with that name.

Job done, object is now normal mapped.

To test this you can delete the `wave.sgm` file within `data/models/ships/wave` and then try my [wave-normap-map.zip](http://quantum-thoughts.net/Pioneer/wave-normal-map.zip) mod which looks really ugly and will prompt @nozmajner into an artist frenzy :P

Andy

### DIFF
--- a/data/shaders/opengl/attributes.glsl
+++ b/data/shaders/opengl/attributes.glsl
@@ -31,8 +31,9 @@ layout (location = 0) in vec4 a_vertex;
 layout (location = 1) in vec3 a_normal;
 layout (location = 2) in vec4 a_color;
 layout (location = 3) in vec4 a_uv0;
-layout (location = 4) in mat4 a_transform;
-// shadows 5, 6, and 7
-// next available is layout (location = 8) 
+layout (location = 4) in vec3 a_tangent;
+layout (location = 5) in mat4 a_transform;
+// shadows 6, 7, and 8
+// next available is layout (location = 9) 
 
 #endif // VERTEX_SHADER
--- a/data/shaders/opengl/multi.frag
+++ b/data/shaders/opengl/multi.frag
@@ -12,6 +12,7 @@ uniform sampler2D texture2; //glow
 uniform sampler2D texture3; //ambient
 uniform sampler2D texture4; //pattern
 uniform sampler2D texture5; //color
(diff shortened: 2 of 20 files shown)


fluffyfreak:
Rats, found a bug :(

fluffyfreak:
Bug fixed, back on track!

corundscale:
Would it be possible to have normal mapped decals?

nozmajner:
What's the point in normal mapped decals?

corundscale:
Combat damage.

nozmajner:
I thought you mean the manufacturer decals.

corundscale:
No, I meant decals as it's often used in 3D game context - additional textures applied, often dynamically, to existing geometry. 

Although normal mapped decals could have other uses for Pioneer as well - some kinds of equipment that should be visible outside but not necessarily protrude much could be added this way (like all sorts of additional shielding and amror), details for infrastructure could be added as decals to allow more variety (combinatorics) than just premade reskins, etc.

fluffyfreak:
@corundscale Yes that'd be possible, but not in this PR. I just want to get the basic system for models in first :)

corundscale:
Sure, I just wanted to raise the issue of such possible application early.

nozmajner:
Started playing with it:
screenshot-20151108-004220

fluffyfreak:
Shiny!

fluffyfreak:
Found some issues with the modelcompiler loading and saving, fixing them up now.

fluffyfreak:
There that should fix it, I hadn't gotten the support correct for SGM files and it was failing to give some of them tangents, they then subsequently failed to load even though it looked like they had been exported.

fluffyfreak:
I might merge this in the next couple of days as it should be quite robust now and I'd like to give it the rest of the month for testing with developers before Geraldines ModDB update.

Of course this doesn't mean that we need normal maps, they are optional and can come later.

Zordey:
Awesome :) 

I really must get back to doing some modelling.

Issue #3450, opened 2015-09-11
Lua function GetEquipFree("cargo") is not behaving as expected


ghost:
I think the lua functions that return the space available for loading commodities don't take into account the loaded "extra" equipment that eats into the available space. The following is the output for a Xylophis outfitted with autospheric shielding and an autopilot (using up 1t of mass each), without any commodity cargo loaded. Real, in-game space to buy/load commodities is 8t.

```
ship:GetEquipSlotCapacity("cargo")
```

Output expected => 10 (maximal capacity of the "cargo" slot)
Output provided => 10

```
ship:GetEquipCountOccupied("cargo")
```

Output expected => 2
Output provided => 0

```
ship:GetEquipFree("cargo")
```

Output expected => 8
Output provided => 10

I think either the "cargo" slot should be used only to describe the space that is available for commodity type cargo, in which case the ship:GetEquipSlotCapacity("cargo") function should take into account the equipment that has already eaten into this space. Or this equipment should be taken into account when returning occupied and/or free space.

A way to tell how much a given ship can load is important as safeguard. I know that trying to load more than what a ship can load will just return "0" and not load the item in question. But I think it would be cleaner if we could test for available space beforehand.

laarmen:
I don't have the code in front of me right now, but this issue sounds
vaguely familiar.

The thing is, we have a slightly schizophrenic approach to the cargo
management. On one hand, we have the metric ton approach, which gives us an
upper limit as to how much *equipment* we can embark, including
commodities. On the other side, we have the cargo hull, which is strictly
designed to hold *commodities* and nothing else. The "cargo" slot describes
the space of this cargo hull, regardless of the mass system. GetEquipFree
deals only with the space, not with the mass. In fact, pretty much none of
the equipment system knows anything about mass. Another thing that makes
matters worse is that we consider that 1t of commodities always takes up
the same volume (a cargo slot) no matter its specificities.

So, there is two things to do to makes things a bit more clear: First,
address the m=v insanity in the commodities. For that, we'd probably need
to convert the masses into kg if it ain't already done (I did it at some
point in the #1719 saga, but I dunno if it remained in the final patchset).
Just that will probably bring a world of small bugs. Once that's done, we
would consider that 1 cargo slot is, say, 1m^3, and then start estimating
the correct masses for the different commodities. And of course we'd need
to modify the commodity market to make these changes apparent.

Second, we'd scrap the mass limit entirely. It doesn't make much sense,
[...]

PR #3516, opened 2015-11-06
Flying empty fix

robothauler:
This fixes the behaviour that the player couldn't take a cargorun pickup mission if he hasn't enough free cargo space at the moment. laarmen advised me to use the equipment API but until the volume/mass thing is improved this PR seems to be the only practical solution IMHO.

### DIFF
--- a/data/modules/CargoRun/CargoRun.lua
+++ b/data/modules/CargoRun/CargoRun.lua
@@ -343,7 +343,8 @@ local onChat = function (form, ref, option)
 		form:SetMessage(howmuch)
 
 	elseif option == 3 then
-		if Game.player.freeCapacity < ad.amount then
+		if (not ad.pickup and Game.player.freeCapacity < ad.amount) or
+			(ad.pickup and Game.player.totalCargo < ad.amount) then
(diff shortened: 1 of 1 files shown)


data/modules/CargoRun/CargoRun.lua
```diff
@@ -343,7 +343,8 @@ local onChat = function (form, ref, option)
 		form:SetMessage(howmuch)
 
 	elseif option == 3 then
-		if Game.player.freeCapacity < ad.amount then
+		if (not ad.pickup and Game.player.freeCapacity < ad.amount) or
+			(ad.pickup and Game.player.totalCargo < ad.amount) then
```
impaktor: No `Game.player:GetEquipFree("cargo")`?

robothauler: I don't know how to deal with #3450. Besides that I have messed up something with this PR.

fluffyfreak:
So this just needs a one line change to `Game.player:GetEquipFree("cargo")` and then it's good for testing/merging yes?

fluffyfreak:
Bump!

robothauler:
I would merge this as is. I think changing this will not work at the moment. I'm not sure.

Issue #3511, opened 2015-11-04
Docking/Launching delay


corundscale:
There is a noticeable delay - around 4s - during launch and touchdown during which the game appears frozen.

The issue occurs both in downloaded binaries and locally compiled version (using MinGW).

nozmajner:
It's because of the autosave feature. Since the json save format saving takes much more time.
(I usually disable autosave, delete the lua file for it, especially when I'm testing)

corundscale:
Then I guess it should be closed. 
Or is there any plan to move to some more compact format at some point?

fluffyfreak:
No the old system was ditched in favour of the JSON format.
It does need to be faster though it's ridiculously slow.

corundscale:
I was thinking of some sort of compact format that still allows nested tagged data, like TLV encoding.
There is a definite advantage to JSON for debugging reasons, though.

impaktor:
Perhaps an option to disable autosave? Perhaps autosave should even be off by default. 

I think saving in general should only be possible when docked, for "The Final Version Of Pioneer", but for now it is useful to save where-ever to catch bugs.

fluffyfreak:
Enable/Disable autosave option.
Also it just needs optimising, profile, identify slow bits, optimise.

clausimu:
> Enable/Disable autosave option.

+1

> I think saving in general should only be possible when docked

-1

impaktor:
Not a bug.

PR #3518, opened 2015-11-07
Make Autosave optional

impaktor:
Since there were complaints of long stalls when docking / undocking, and people not making use of it anyway.

Note: this is untested!

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -279,6 +279,10 @@
       "description" : "",
       "message" : "Employment"
    },
+   "ENABLE_AUTOSAVE" : {
+      "description" : "Allow the game to autosave when docking",
+      "message" : "Enable Autosave"
+   },
    "ENABLE_COCKPIT" : {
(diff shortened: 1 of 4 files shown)


impaktor:
... in #3511

fluffyfreak:
Goes kablooey in `optionCheckBox` from `Settings.lua` line 19

I wish we could copy and paste from that damned error dialog!!!

impaktor:
> I wish we could copy and paste from that damned error dialog!!!

When running pioneer from the command line (as I do) it is also echoed there. 

I'll probably just misstyped something. I'll have a look tomorrow. Thanks!

laarmen:
We could simply put the text in the clipboard (SDL_SetClipboardText) when there's an error. I wish we could add a button to the error dialog but the widget is, by design, too primitive for that.

impaktor:
Wow. Many bugs and typos in it. Fixed them all now.

I've tested both disabled and enabled autosave, and it seems to work now.

PR #3531, opened 2015-11-12
Fix handling of negativ coord numbers for DecimalToDegMinSec function

clausimu:
As @nozmajner pointed out yesterday in IRC - negative lat/long coordinates were not transformed correctly from decimal to degree/minute/second format. This was due to my incorrect use of the floor() function to get the integer part of a number. For positive numbers it doesn't matter, but for negative numbers returning the smallest following integer throws everything off by a little bit.

This merge request fixes the coordinate conversion.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -428,10 +428,10 @@ void StrToAuto(double *pVal, const std::string &str)
 */
 std::string DecimalToDegMinSec(float dec)
 {
-	int degrees = floor(dec);
-	int minutes = floor(60*(dec - degrees));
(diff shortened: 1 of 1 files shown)


impaktor:
Merged. Commit is in master, closed this in a funny way, because I cherry-picked the commit, rather than copying branches and stuff. It was one of those days.

fluffyfreak:
Yeah we're not getting new build at the moment, I've messaged @robn about it (via Twitter) to see if we can just have builds without the `sgm` files for the time being.

PR #3536, opened 2015-11-17
L4 & L5 Lagrange Point display

fluffyfreak:
Since we [might need a way to target](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=379) them, I've added the display of [L4 and L5](https://en.wikipedia.org/wiki/Trojan_(astronomy)) [Lagrange Points](https://en.wikipedia.org/wiki/Lagrangian_point) to the `SystemView` screen.

Still to do:
- [x] Toggle their display


nozmajner:
Here are the icons you mentioned, attached here, sorry for the wait:
--> l4
--> l5
(attaching here, it's easier with my wonky connection.)

I kept them as simple dots, not fully bright. L5 is even a bit duller, to differentiate between them.
I've included small text next to them, but for more bodies it tends to look cluttered, so I ditched it:
l-points_1
Here just dots, but I played with a faint dashed line to make it more clear which body they belong to:
l-points_2

fluffyfreak:
Thanks @nozmajner I've updated the icons and the icon toggling now so that it has OFF, ICON and ICON+TEXT states.

nozmajner:
Err, should I make a button for the toggling?

fluffyfreak:
Doh! Yes, I just realised that I've hacked on together that just says "L4/L5" :/

Sorry, you can see it in the `Files changed` view, it's naff programmer-ness-art!

nozmajner:
How about this?
l45_button

impaktor:
I've tested it earlier today. I also though the button should have different states (like a highlighted). I wonder if the mouse over text should be something shorter and simpler: "Show stable Lagrange points".

Question: how the heck will the players know they can jump to L-points? Obviously, we'll put it in the FAQ, Tutorial, Quickstart, Manual, but I think most people just download and play. And if you have a Frontier background you wouldn't notice that you can do in system jumps. Hmm, tutorial mode.


@nozmajner nice. So it would switch between being high lighted? Or just the small dots going in/out of fade on the icon?

nozmajner:
Good point. How about these?
Off: (faded the text)
l45_button_off
On: (Button brighter, text faded to dark button color)
l45_button_on
On & text: (Text is highlighted too)
l45_button_on_text

impaktor:
Don't beat yourself up over it. The cleanest, and most obvious way was `L4/L5` text-on-button, which is beauty to a programmer, no? :wink:

I'm seeing something strange in Sol. Jupiter, Saturn, Uranus and Neptune are missing their L-points (Pluto is fine). Seems to be reproducible (seeing it both times I
start the game).
screenshot-20151119-093640

fluffyfreak:
Yeah, there's a few other things wrong with that screen I've just seen like the orbit fade not being attached too the planet!

fluffyfreak:
@impaktor good catch! I was checking against the wrong thing for half of the test to see if I should show lagrange points! It was checking against the parent body, instead of the current child object.

impaktor:
Ah, right, you were discriminating against gas giants. 

I wouldn't mind a rebase before merge, so there aren't "dead"/unused unnecessary pngs added to the repo history.

fluffyfreak:
I might nerd your help with a rebase, never had one work so far, might a windows thing.

impaktor:
The Jovian satellites seem to be too fast for their trail to keep up (Ganymede, Europa). Are the trails not drawn from the centre of the body, but some imagined surface?
screenshot-20151119-173138

fluffyfreak:
I don't recall, but lets open a separate issue for the trail-separation because we can fix that independently.

impaktor:
Is this ready for merge, or is anything missing?
I can clean up the commits at merge.

fluffyfreak:
It's ready i think.

impaktor:
merged. squashed the last commits togehter, so 3 in total.

Had the png's been in a separate commit, I would have set @nozmajner as author, but even my obsession over a clean git history has its limits.

PR #3539, opened 2015-11-22
Lagrange Points in turquoise-ish

fluffyfreak:
One little desire to set RGBA values from a hex (32-bit INT) resulted in a lot of files needing changing.

The goal was just to make the Lagrange Points blue-ish but in setting a single 32-bit RGBA value and unpacking it to RGBA in the Color4ub constructor I also had to update a lot of other files.

Added the `Color::BLANK` preset, removed the single `Uint8` ctor, added the 32-bit `Uint32` ctor and RGBA unpacking so you can write values like `0x00d6e2ff` and it will unpack r=`00`, g='d6', b='e2', a='ff'.

Finally... all that made the icons blue, they look nice like that, bit blocky though, think we need higher res icons.

Andy


impaktor:
Tested (for 10 seconds). 
New colors look nice.
(Did the orbits go "goth" again? or is it just that today is the first day in 2 weeks with sun outside so I can hardly see anything my monitor.)

Either way, merge when you feel it's ready.

fluffyfreak:
Must just be a sunny day where you are :)

I had missed a couple of cases where I'd cast it to Uint8 so changed them.

Issue #3449, opened 2015-09-10
MXE build of Modelcompiler not working on Windows


impaktor:
Just opening this to keep track of the, otherwise hard to find, discussion over in:
https://github.com/pioneerspacesim/pioneer/commit/f4c4ddd6cbe1d8735edfc6cb505197d7e88eeab8

fluffyfreak:
Changed the title of this since it's the `modelcompiler` that processes models into our SGM format for faster loading, and not the built-in `pioneer -mv` option for viewing models, that is broken.

fluffyfreak:
@robn I think this is the cause, or at least it's a problem, to do with the recent builds not working, the `modelcompiler` can't rebuild the `sgm` files, so it ships the **old** `sgm` files with the new build.

Those old sgm files cause the game to crash.

robn:
Presumably I screwed something when I crashed us over to Ubuntu build hosts the other week. Will try to look tomorrow.

impaktor:
@robn is it possible to revert the download link on the homepage to a non-crashing version in the mean time?

fluffyfreak:
I can verify this with my own MXE build of `modelcompiler` but not the why, It's very strange.

Pioneer works just fine.

fluffyfreak:
AHA!!! @robn it is running, sort of, but then something it obviously silently failing.

I've had to fall back on a sort of `printf` debugging by writing out txt files when it gets to certain stage in the program which is... primitive to say the least.

...and it all seems to be working actually, it doesn't output anything to the cmd line like it should but with the `-b inplace` parameters it does produce the `sgm` files in the `data/models/` sub-folders. It just does it so fast (`on my ssd`) that I didn't even notice it run, sub 1-second compile time!!!

Anyway Pioneer loads just fine with those generated files now.

In short, I do not know why the build machine is failing to produce valid sgm files :(

fluffyfreak:
Also (@impaktor, @robn) I have a zipped version that I could upload with the fixed `sgm` files and my `modelcompiler.exe` that seems to work. Should I replace the current bad `pioneer-20151109-win32.7z` version with a `pioneer-20151109-win32-fixed.7z` upload?

robn:
Yes, just upload it straight over the top.

I will figure out why the build isn't generating correct `sgm` files before I do the next build. I hope to have time on Friday.

fluffyfreak:
:D

impaktor:
[...]

PR #3542, opened 2015-11-24
Fix the broken win32 cross compile.

fluffyfreak:
ditto


impaktor:
Not related to #3449?

fluffyfreak:
Not yet, this is just getting it to compile!

PR #3543, opened 2015-11-25
Speculative fix for modelcompiler

### LINKED ISSUE #3449: MXE build of Modelcompiler not working on Windows

fluffyfreak:
I reordered the `Makefile.am` so that Pioneer and the ModelCompiler are in the same section, I also made the ModelCompiler use the Pioneer icon and just bundled all of the libs into it the same as Pioneer.

After doing this the version that I built seems to work ok for me but whether this really fixes #3449 or whether my build is just different in some other way I do not know.

Pinging @impaktor and @robn for comment and thoughts please :)

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -314,6 +314,45 @@ if !HAVE_LUA
 pioneer_LDADD += ../contrib/lua/liblua.a
 endif
 
+modelcompiler_SOURCES = \
+	modelcompiler.cpp \
+	Color.cpp \
+	CollMesh.cpp \
+	DateTime.cpp \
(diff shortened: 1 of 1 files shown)


impaktor:
@robn did some magic to get the build back, but did that also fix what this PR is trying to fix? Where are we on this?

fluffyfreak:
Nah this is defunct

PR #3544, opened 2015-11-27
Use PageUp / PageDown to increase / decrease time acceleration

salborrelli:
Player can now use the PageUp / PageDown button for requesting an increase / decrease in time acceleration. The request works under the following conditions, otherwise it is ignored:

1. player is in World View
2. game is not paused
3. player is not in hyperspace
4. if requesting an increase, time acceleration is not already at max (10000X)
5. if requesting a decrease, time acceleration is not already at min (1X)

In order to avoid conflicts, the following default keybindings have also been changed:

BindTopCamera: alternative key has been set to "T" (was "PageUp")
BindBottomCamera: alternative key has been set to "B" (was "PageDown")

Other default keybindings for PageUp and PageDown haven't been changed because they do not affect the World View.

Please test the new feature and let me know if you find bugs.

Even if my proposal works, please let me know if you think there is an easier (or more elegant) solution to the same problem. I am basically starting out with the SDL library so I would be glad to try a more "best practice" way, if any.


impaktor:
Sweet. Thanks!

Tested.

I see two (and a half) questions that should be considered:
 1. Should pause state be excluded from the PgUp/PgDn cycle?
 2. Shouldn't PgUp/PgDn work when in stationView? I can set them with the mouse, but, as the PR description says, this isn't WorldView. Time accel in BBS or ship store is used sometimes to gett better missions/ships spawning.
 3. Potential issue: PgUp/PgDn zooms the F2 map, this is good/OK?. (Why would anyone want to time accel when viewing the star map?) 

So we are OK with PgUp/Dn not always time accel-ing and reverting to mouse in those cases?

@nozmajner, thoughts?

@salborrelli If you want to attack yet another small issue (also nothing open on the tracker for it),
that is very similar to the code you've been working on: ditch the highlighted states of `multifunc_equip_on.png` and `multifunc_scanner_on.png`, since no point in using four separate buttons for two different states. I.e. have one button that toggles between scanner and equip.

src/Game.h
```diff
@@ -98,6 +98,17 @@ class Game {
 
 	void SetTimeAccel(TimeAccel t);
 	void RequestTimeAccel(TimeAccel t, bool force = false);
+
```
impaktor: Nitpick of the day: Technically, there's a trailing indentation on this empty line. I think we usually don't indent empty lines.

corundscale:
Customizable key bindings.
For example I use PgUp and PgDn for other things.

But having keybindings for increasing and decreasing time acceleration would be a big plus.

src/Game.h
salborrelli: Got it! The indentation will be removed in the next commit.

impaktor: I'd prefer it if was updated in the current commit, if possible (I'm in IRC if you need help).
http://pioneerwiki.com/wiki/Using_git_and_GitHub#Fixing.2FUpdating_your_pull_request

salborrelli: Done.

salborrelli:
> But having keybindings for increasing and decreasing time acceleration would be a big plus.

@corundscale agree and I will look into it. I also think that make the keybindings customisable would resolve some of the problems related to the default bindings that I have mentioned above.

@impaktor @nozmajner in response to your questions:

>1. Should pause state be excluded from the PgUp/PgDn cycle?

Not completely sure, but I preferred to exclude it because the system frequently goes in "pause" automatically when some event happens, e.g. player click "Esc" to switch in Settings View. So the "pause" state looks like an exception, but of course it would be very easy to include it into the PageUp / PageDown cycle so let me know your thoughts. 

>2. Shouldn't PgUp/PgDn work when in stationView? I can set them with the mouse, but, as the PR description says, this isn't WorldView. Time accel in BBS or ship store is used sometimes to gett better missions/ships spawning.

I was to aware of this use case so let me know if you think that the PageUp/PageDown should be extended to other views.

>3. Potential issue: PgUp/PgDn zooms the F2 map, this is good/OK?. (Why would anyone want to time accel when viewing the star map?)

Yes I also think that increasing / decreasing the time acceleration in map view makes little sense, to the extent that I would suggest to hide the time controls altogheter in this view.

impaktor:
> to the extent that I would suggest to hide the time controls altogheter in this view.

:+1: 

(although wait for OK from @nozmajner as well, before implementing that)

nozmajner:
So it's not rebind-able, or those are just the default bindings?
I don't think there's really need for time-accel on the sector map, but it is very useful in the orbit map. 

I'm not sure about hiding the time controls away in other views. One reason against it could be that you might be looking at some of those screens while flying on high time-accel, but might want to set it to lower for a reason (like thinking about your route or something).

Excluding pause from it makes sense to me, because sometimes you just want to quickly go back to real-time, and it would stall the game if it would pause it if you press pg-dwn rapidly.

impaktor:
> is very useful in the orbit map

Even the "real world" time accel, not just the right hand side "simulated time accel"?

> Excluding pause from it makes sense to me

Then that's settled.

nozmajner:
Yes, I use time accel all the time for maneuvers for example, even though we have an orbit planner now, it's results aren't that accurate, and you have to watch your orbit.

salborrelli:
Ok so if I understand well we have more or less settled on the following list of adjustments:

1. make the buttons used for requesting a time accel / decel rebindable
2. extend the function to all reasonable views (e.g. Settings View should remain excluded).
3. keep excluding the pause state from it
4. leave the ship panel visibility as it stands

Regarding this request from @impaktor 

>ditch the highlighted states of multifunc_equip_on.png and multifunc_scanner_on.png, since no point in using four separate buttons for two different states. I.e. have one button that toggles between scanner and equip, and uses two pngs instead of four.

I think it's better placed into a new PR so I'll wait for this PR to be closed before looking into it if you don't mind.

impaktor:
> I think it's better placed into a new PR  so I'll wait for this PR to be closed before looking into it if you don't mind.

Oh, absolutely its own PR! I just thought I'd suggest it while you were poking around in this similar code, and I thought maybe you wanted to work on something new in parallel while this is being discussed.

walterar:
This is very useful, and does its job well. It will be welcomed by the players. :+1:

impaktor:
So should Shift+F[1,2,3,..] still map to time accel, as they do now? Or should that be removed? I.e. just use PgUp/PgDn to cycle.

salborrelli:
Hi guys,
 all improvements to the original function have been implemented as described into the commit message. Please retest and advice if you find any issue.

@impaktor I haven't touched the mapping for Shift+F[1,2,3,4] for now but if you need those buttons to be unmapped we can do it quite easily so let me know.

src/LangStrings.inc.h
```diff
@@ -455,3 +455,5 @@ DECLARE_STRING(SHIPS_DISPLAY_MODE_TOGGLE)
 DECLARE_STRING(SWITCHED_TO_ROTATIONAL)
 DECLARE_STRING(SWITCHED_TO_PARENT)
 DECLARE_STRING(L4L5_DISPLAY_MODE_TOGGLE)
+DECLARE_STRING(INCREASE_TIME_ACCELERATION)
+DECLARE_STRING(DECREASE_TIME_ACCELERATION)
```
impaktor: You also need to add the string to data/lang/core/en.json (and _no other_ .json-file after that).

salborrelli: Done added the translations to en.json. BTW if you need it, I can also keep the Italian translation (it.json) in synch whenever I add a line to en.json

salborrelli: :+1:

impaktor:
When I tried this, PgUp/Dn in sector view (F2) both moves forward and time accelerates. Now obviously it's because the key was mapped in my config to do that. Removing my config (emulating a fresh pioneer install) fixed it. So all is good, I guess.

I haven't looked at the code in any detail, but function in game is as expected.

The major thing now, is to __change all your indentation to tabs__, as that is what we're using in pioneer. Your editor should be able to sort it out, or just do a search and replace.
Please see: http://pioneerwiki.com/wiki/Code_style

salborrelli:
>The major thing now, is to change all your indentation to tabs, as that is what we're using in pioneer. Your editor should be able to sort it out, or just do a search and replace.

Done fixed some wrong indentation and re-committed. Please review and let me know if there is anything else that I need to fix before you can merge the PR.

src/WorldView.h
```diff
@@ -132,6 +132,14 @@ class WorldView: public UIView {
...
 	void OnPlayerChangeTarget();
 	void OnPlayerChangeFlightControlState();
+	/**
+	 * Handler for "requestTimeAccelerationInc" event
+	 */
+	void OnRequestTimeAccelInc();
+	/**
+	 * Handler for "requestTimeAccelerationDec" event
+	 */
+	void OnRequestTimeAccelDec();
```
impaktor: Last nitpick: I'd prefer 1 line C++ style comment rather than 3-line C style. I don't know what the others say (speak up if in disagreement).

src/Game.h
```diff
@@ -99,6 +99,17 @@ class Game {
 	void SetTimeAccel(TimeAccel t);
 	void RequestTimeAccel(TimeAccel t, bool force = false);
 
+    /**
+     * Requests an increase in time acceleration
+     * @param force if set to false the system can reject the request under certain conditions
+     */
+    void RequestTimeAccelInc(bool force = false);
```
impaktor: See my other comment just now on C++ vs. C style comments.
(seems like we have C-style comments in other places, but at least not in this header file)

I think C-style comments is (only?) used at least when documenting code for Natrural docs LuaAPI that we use: http://eatenbyagrue.org/f/pioneer/codedoc/files/LuaBody-cpp.html

impaktor:
I'll merge this as soon as this last perfectionist nitpick is fixed. Otherwise, looks good, and very clean commits, thanks!

You're already in authors file, right?

src/Game.h
salborrelli: @impaktor my comments are not meant to be C-style to be true, they are Javadoc style instead. I am used to this system from my Java experience and I tend to use it in any context because I find it very explicative. In addition, there are several tools (e.g. Doxygen) which would allow us to easily generate an online documentation browser from C++ source code that is annotated this way.

That said I have of course no problems to switch to a 1-line C++ style commenting system if you prefer.

Will modify the comments and amend the commit as soon as I can.

impaktor: We have Doxygen support (see COMPILING.txt) in pioneer.

As for comment style, I'll have to ping @johnbartholomew and @fluffyfreak as I'm not sure what the official style is.

johnbartholomew: Given that the rest of the file uses as line comments I would recommend sticking to that. I would also be happy for you to use 'three-slashes' line comments, which Doxygen also recognises. I.e.,

    /// Request an increase in time acceleration.
    /// @param force  If false, the actual acceleration level can be less than the requested level (e.g., due to nearby ships).

salborrelli:
>I'll merge this as soon as this last perfectionist nitpick is fixed. Otherwise, looks good, and very clean commits, thanks!

@impaktor np and I've just fixed the comments by replacing the multi-line delimiters with single-line ones 

>You're already in authors file, right?

@impaktor yes thanks!

impaktor:
Merged. 

Sorry to make you jump through all those hoops, but now you know the dance, so easier next time.

Also, I notice you code from your master branch. I wouldn't recommend that. Work from a branch instead.

    git branch my_new_feature

    git checkout my_new_feature
    // do stuff, commit, push

    git checkout master
    
I suspect your master branch is a bit strange now, since your commits are not in the same place in the git history as the upstream master is now? `git reset --hard upstream/master ` will fix that, but is a dangerous command, so make sure you run it from your master branch, and have nothing to loose on that branch. Also that your `upstream` is the pioneer-team's copy of the pioneer source (`git remote -v`).

fluffyfreak:
for working on a branch I generally do:
> git checkout master
> git checkout -b my_new_branch
> // do stuff, commit, push
> git checkout master
So that I don't accidentally branch from another branch :)

salborrelli:
@impaktor @fluffyfreak thank you for the tips and I will surely take them into account from now on.

PR #3546, opened 2015-11-29
Fix cockpit loading

fluffyfreak:
The Lua code used the name `cockpit` to define what cockpit model to load but in the change to the JSON format it was altered too `cockpit_model` only none of the JSON files used that!

So I've reverted it to `cockpit` and now everything should work as intended.

I've also disabled the `ALLOW_LUA_SHIP_DEF` code since all of our ships and models are now in JSON format.


nozmajner:
Wow, thanks. :)

fluffyfreak:
No problem :) any obstacle to great art Andy removes!

PR #3549, opened 2015-11-29
New-UI galactic view

johnbartholomew:
This is the simplest of the map views, since it just shows the static image of the galaxy. Even so, conversion is not yet finished.

Todo:
  - [x] Add the scale indicator
  - [x] Make 'Zoom' text translatable (possibly make other text translatable too)
  - [x] Centre on the Player's current sector, rather than always centering on Sol
  - [x] Mark the Player's current location with a dot or cross rather than just text
  - [x] Clean up commits (I went down some dead ends in this implementation)
  - [x] Remove the outline font thing (it's broken)


impaktor:
Nice!

All other Lua-scroll-bar thingys (BBS, commodity market, ship market, equipment market) support middle-mouse-scroll wheel. I guess it would be nice to also have it here.

Just seeing how much work it is to move something this "simple" to new UI hints at the daunting task left for the other screens.

Although I wonder if the EconInfo and Demographics tabs could not be combined to a single screen with just text: The 4+1 column for export/import/illegal commodities, and above (or below) those one could put the political / demographical info on the system. Anyway, that's for another day.

By the way, nice to see you back here.

johnbartholomew:
impaktor wrote:

> All other Lua-scroll-bar thingys (BBS, commodity market, ship market,
> equipment market) support middle-mouse-scroll wheel. I guess it would be
> nice to also have it here.
>
Good idea.

> Just seeing how much work it is to move something this "simple" to new UI
> hints at the daunting task left for the other screens.
>
Some of the work for this one (the OverlayStack and LabelOverlay widgets)
should be re-usable and useful for other things too, at least with a bit of
extension. I  expect to use them for the other maps and also the WorldView,
which all have "stuff with labels overlaid". WorldView will probably have
several overlaid layers -- the actual view of space, the cockpit (if it's
enabled), labels, velocity indicators, targeting indicators, plus the
existing HUD elements. But... yeah, it'll be a lot of work to get that all
done.

impaktor:
> Some of the work for this one (the OverlayStack and LabelOverlay widgets) should be re-usable and useful for other things too, at least with a bit of extension.

One of my main problems with NewUI is that there doesn't appear to be any kind of documentation for it? This means that we numbnuts that don't want to dig deep into C++ code (I don't know where), tend to be limited to thinking: "I kind of want it to work like this-and-that existing lua UI, so copy paste and wing it from there".

So is it possible to document OverlayStack and LabelOverlay somewhere? Like in code, formatted so Naturaldocs picks it up?

> WorldView will probably have several overlaid layers

I know robn has working(?) transparency layers for WorldView in some branch, for comms history log.

johnbartholomew:
Current status:
screenshot-20151201-212332

johnbartholomew:
Added a (badly formatted) scale indicator. Made the green dot and its label update to show the player's current position. Now I just need to clean up the commit history and this should be ready to test and merge.

screenshot-20151206-012037

johnbartholomew:
Commits are now all cleaned up. Ready for testing and (hopefully) merging.

impaktor:
SWEET! 

I suppose sometime in the future, scroll wheel support can be added, like rob did in a435eafec147dfe756158408adc6d3be33bc1f26. 

I'll test it tomorrow if it hasn't been merged by then.

PR #3552, opened 2015-12-06
New UI scaling (HiDPI) support

johnbartholomew:
Old UI used a uniform 800x600 unit design grid, which meant that if you had more pixels everything would be stretched to fit. New UI uses a fluid layout system and pixel units, which is more flexible but results in some UI components being very small when shown on a high DPI display.

New UI already has some support for applying uniform scaling in the UI 'Skin'. This is principally used to cope with displaying the user interface on low resolution screens -- squeezing the UI if the game window is smaller than 1027x768. This PR extends that scaling support to apply the scaling factor to more things, and to allow it to be set in the pioneer config.ini.

### DIFF
--- a/src/GameConfig.cpp
+++ b/src/GameConfig.cpp
@@ -15,6 +15,7 @@ GameConfig::GameConfig(const std::map<std::string,std::string> &override_)
 	map["StartFullscreen"] = "0";
 	map["ScrWidth"] = "800";
 	map["ScrHeight"] = "600";
+	map["UIScaleFactor"] = "1";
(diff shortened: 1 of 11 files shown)


fluffyfreak:
I haven't checked out the code apart from a quick read of it online, but I like the feature and it'd be good to get it in. I'm afraid my merge frenzy appears to have introduced a conflict for you to fix now though, I do apologise.

johnbartholomew:
No problem -- a merge conflict between this and the new-ui-galaxy-map
branch was expected. I'll fix it later today.

PR #3557, opened 2015-12-09
Galactic view fixes

johnbartholomew:
Fixes for a couple of bugs noted on IRC.

  * Comms messages in the HUD were visible through the galaxy map.
  * There's a warning about marker_dot.png being non-power-of-two.

See commits for details.

### DIFF
--- a/data/ui/GalacticView.lua
+++ b/data/ui/GalacticView.lua
@@ -55,7 +55,9 @@ local controls =
 		)
 	)
 
-local galacticView = ui:OverlayStack():AddLayer(map):AddLayer(controls)
+local galacticView =
+	ui:ColorBackground(0,0,0,1,
+		ui:OverlayStack():AddLayer(map):AddLayer(controls))
 
 ui.templates.GalacticView = function ()
 	resetCurrentSector(map)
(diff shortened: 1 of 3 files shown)


impaktor:
I wasn't 100% sure it was a bug to begin with, if we compare the behavior with StationView:
output_p3kgkd

(Or stationview has it too?)

Feels like that comms is going to be reworked by _someone_ (a knight in shining armor) at some point, robn has that half finished branch of his.

johnbartholomew:
It definitely looked weird in the galactic view because the comms message
is only partially hidden by the galaxy image itself. As you zoom in the map
it (un)covers more of the text so you can see the square boundary of the
galaxy image. The effect in the station view is not so bad because the same
part of the text (the top half) is always covered.

A quick hack for this would be to just move the comms message up the screen
a little bit, so that it's completely covered instead of just partially
covered. But you're right; at some point the comms message (like everything
else) needs to be reworked.

fluffyfreak:
That sounds like something that can be done in another PR, merging this one.

impaktor:
I consider showing (part of) comms text under the station window a "feature". Then you know something just happened. Alas, no way to see the comms history yet.

PR #3566, opened 2015-12-13
Translation for #3549

radius75:


impaktor:
Looks fine to me.

Ideally, we need the transifex magic of @robn (or @johnbartholomew?) to move translated string when this is merged, to not need re-transltion.

fluffyfreak:
@impaktor Does this mean we're waiting on @robn to do something or do you mean in future we need to do it another way?

impaktor:
> @impaktor Does this mean we're waiting on @robn to do something 

I believe so, yes. Since this moves already translated strings from core to ui-core.

johnbartholomew:
I'm merging this manually now.

johnbartholomew:
Merged, and moved translations pushed to Transifex.

PR #3572, opened 2015-12-18
Deimos Seed

fluffyfreak:
Based on [this SSC thread](http://spacesimcentral.com/ssc/topic/5282-announcement-big-update-adventurous-test-pilots-needed/?p=56349).

I also found a bad NaN which caused the string to inf-loop.

### DIFF
--- a/data/systems/00_sol.lua
+++ b/data/systems/00_sol.lua
@@ -169,6 +169,7 @@ local mars_moons = {
 			:longitude(math.deg2rad(-5)),
 	},
 	CustomSystemBody:new('Deimos', 'PLANET_ASTEROID')
+		:seed(439771126)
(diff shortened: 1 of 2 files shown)


impaktor:
A thought I had reading that SSC post: I would have expected Deimos with the seed used by the old rand gen to make it look different compared to the new one.

fluffyfreak:
Yeah me too, another thing that doesn't look right is on [StarSystemGenerator.cpp - Line 602](https://github.com/pioneerspacesim/pioneer/blob/master/src/galaxy/StarSystemGenerator.cpp#L601).

Using a smaller number for the seed also seems to produce flatter terrain, biggers number always seem to get lumpier (_I played around with the value a bit_) which also doesn't seem very random.

Issue #3574, opened 2015-12-26
Recent Linux build from source fails to start


TeddyDD:
I pulled source recently but new build don't want to start 
I'm using Linux Arch
```
new ModelCache
Shields::Init
BaseSphere::Init
pioneer: GeoSphere.cpp:48: static void GeoSphere::Init(): Assertion `s_patchContext->GetEdgeLen() <= detail_edgeLen[4]' failed.
fish: './pioneer' terminated by signal SIGABRT (Abort)
```

Tichy:
I can confirm this. Same error and same distribution.

fluffyfreak:
I've just done this on my Ubuntu test box and it builds and runs just fine.

Wonderful. I'm not sure how it could ever get into that state given the line preceeding that either.

fluffyfreak:
Perhaps try deleting your `config.ini` from the `./pioneer` folder in your home directory so that it has to recreate it.

johnbartholomew:
This will happen any time the planet detail level is set to "very high", because the code creates a GeoPatchContext passing in `detail_edgeLen[detail]` (with detail between 0 and 4 inclusive), then GeoPatchContext adds 2 to the passed in value, and then the code checks that the result is <= the maximum `detail_edgeLen` value.

TeddyDD:
Removing config.ini fixed this. And I see PR is already waiting :) Thank you @johnbartholomew

PR #3575, opened 2015-12-27
Make DetailLevel a class and initial all members by default.

fluffyfreak:
Whilst investigating #3574 I noticed that the DetailLevel struct was not being fully initialised so I've changed it too a class and fully initialise all members to `0` within the default constructor.


laarmen:
Why make it a class? IIRC you can declare a constructor in a struct as you
would in a class.

PR #3577, opened 2015-12-27
Remove incorrect assertion on max planet detail

johnbartholomew:
For #3574. I don't think these assertions are useful.


fluffyfreak:
Ah yes, I changed it to store the `+2` value direectly rather than keep readding it everywhere, but when you build using Visual Studio in Release it disables asserts.

PR #3584, opened 2016-01-07
Barnard's star update

nozmajner:
This is an update to Barnard's star, I've already posted on the dev forum a couple of weeks or so ago.
Small system compared to Sol, with 5 planets and a bunch of moons. Some of them have settlements, OPLI branches, and an orbital station orbiting the sole gas giant. The two high security prisons are in the L4 and L5 points of the gas giant (faked of course), with the same orbit they had before.
The bodies are named after the twelve causes of suffering, from Buddhism, and the settlement names are on the same line of thought as the two High Security Prisons.
The first two planets are ocassionally inside the habitable zone of the star (according to Universe Sandbox 2, where I designed the system mostly.)

I've kept the original system description, only updated the line about it being empty, and added a line about the naming scheme.

The system is quite small, my aim was to make it a good fit for learning and practicing navigation. This was my main motivation for it. And in the future it could serve as the place for an in-game tutorial with fiction: a former inmate is released, and relearns piloting/makes a certification exam for it.

It's a quite gloomy place, because of Barnard's being a brown dwarf.

screenshot-20160107-175624
screenshot-20160107-175630
screenshot-20160107-175822

### DIFF
--- a/data/systems/custom/00_barnard_s_star.lua
+++ b/data/systems/custom/00_barnard_s_star.lua
@@ -4,10 +4,11 @@
 local s = CustomSystem:new('Barnard\'s star',{ 'STAR_M' })
     :faction('Solar Federation')
     :govtype('EARTHCOLONIAL')
+	:lawlessness(f(0,100)) -- 1/100th from a peaceful eden
     :short_desc('Earth Federation prison colony')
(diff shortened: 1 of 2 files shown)


fluffyfreak:
I'm happy with this update/change.

PR #3585, opened 2016-01-09
Convert geometry to use 32-bit indices

fluffyfreak:
Currently we're limited to 65536 vertices in a models mesh due to using 16-bit indices.

This PR changes the rendering to use 32-bit (_unsigned_) indices which allows for many more vertices in a single mesh.

- [ ] test with a mesh that has that many vertices!!!

Andy

### DIFF
--- a/contrib/vcacheopt/vcacheopt.h
+++ b/contrib/vcacheopt/vcacheopt.h
@@ -67,6 +67,7 @@ public:
 	}
 };
 typedef TVertexCacheData<int, INT_MAX >					VertexCacheDataInt;
+typedef TVertexCacheData<unsigned int, UINT_MAX >		VertexCacheDataUInt;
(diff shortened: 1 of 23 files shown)


nozmajner:
I'll give you a station model that has everything in one .dae file. 220k-ish vertices. As soon as it uploads.

walterar:
Compiled Linux 64 OK
Run faulted by sgm
Sgm delete and regenerate
Run OK
Win32 compiled MXE OK
Remove sgm and run in wine OK
Sgm regenerate in wine
Run in wine OK but the buildings are not visible.
Test sgm generated in Linux64 with identical results.

laarmen:
@walterar, thanks for the detailed testing!

fluffyfreak:
@nozmajner well there seem to be some bits missing, not sure if that's the model or the engine though :/

PR #3586, opened 2016-01-11
Anisotropic Texture Filtering

fluffyfreak:
This adds (_optional_) [Anisotropic Texture Filtering](https://en.wikipedia.org/wiki/Anisotropic_filtering) so that our texture remain crisp and beautiful as the day our hardworking artists crafted them despite viewing them from shallow angles.

This is optional because on my little Linux test machine it reduces the framerate from __20 fps__ to __11 fps__ which some people may not enjoy.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -291,6 +291,10 @@
       "description" : "",
       "message" : "Employment"
    },
+   "ENABLE_ANISOTROPIC_FILTERING" : {
+      "description" : "Use Anisotropic Filtering",
+      "message" : "Anisotropic Filtering"
+   },
    "ENABLE_AUTOSAVE" : {
       "description" : "Allow the game to autosave when docking",
       "message" : "Enable Autosave"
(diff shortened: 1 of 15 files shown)


impaktor:
What should the default value be?
Punish slow computers, or reward fast ones?

In general, Is there some way to indicate in the settings menu which options are recommended for slow computers? Either way, we have the wiki, which should be updated upon merge:
http://pioneerwiki.com/wiki/Settings_Menu

fluffyfreak:
Good point on the default, I'll turn it off by default for now I think since it does have quite a hit on lower end GPUs.

fluffyfreak:
@impaktor we could have a dropdown with default configurations for "__lowest-detail__" through to "__best-graphics__" but that should be a different PR I think.

impaktor:
> should be a different PR I think

Yes, I didn't mean to imply anything else.

> a dropdown with default configurations

Like: "Max performance" and "Max detail", or what ever.

fluffyfreak:
Yeah that sort of thing.

This is ready for testing now by the way, should "just work", but game will require a restart to enable/disable.

impaktor:
Tested. Doesn't crash. Not sure I see anything special. But I don't have the eye for these kind of details.

fluffyfreak:
Easiest way to see it is sat on the launchpad (_cockpit off_) and look out at the far side of the pad areas. With it off you start to see a moire pattern, or blurriness and all the colours merging together.
With it enabled however it should stay much clearer even when being viewed at very shallow angles.

I just figured out that some texture don't need it enabled at all so having it on for UI, Billboards and Cubemaps is a waste and I've disable it for those.

PR #3592, opened 2016-01-16
SGM graceful failure

fluffyfreak:
Fail-back to the model definition files with a warning when we hit a bad SGM file instead of throwing an exception.

As we change the __SGM_VERSION__  we don't want to break mods and other things that can be simply updated or fixed. Especially when we have all the data we need to keep running in the original `.model`, `.dae` and other files present.

So instead of raising and exception we first try to fail-back and load the original data, only if that fails will we raise an exception.

Andy


impaktor:
Sounds reasonable.

Just a thought about future, could pioneer re-bouild the sgm file if it isn't found, or found to be invalid?

fluffyfreak:
I think sgm files should be a manual process. It's something to think about for the future though.

PR #3599, opened 2016-01-19
Text rendering performance improvements

### LINKED ISSUE #3417: More Text = Less Frames
So for the record I'm running Win7 64 bit, 8GB of RAM, an ATI HD7870 and an AMD Phenom II X4 Black Edition @ 3.4Ghz.

Whenever text is rendered on screen I lose frames like crazy, to the point where opening autopilot can dip my framerate from a perfect 60 to sub-15. I'm not sure what's causing this but I have the latest proprietary drivers from ATI installed as well. Other than this the game runs flawlessly.

On latest version: 20150701

fluffyfreak:
Another attempt to fix #3417 by improving the performance of our text rendering.

I've discovered that there were several flaws and shortcomings in the attempts to avoid recreating objects. Some things works fine within `TextLayout` for example, however this was rendered useless by the containing object simply deleting it's entire `TextLayout` object.

There is still more work to be done, several uses of `DrawTriangles` need to be remove to really reduce the buffer creation impact for example:
- [x] Gradient.cpp
- [x] Icon.cpp
- [x] Image.cpp
- [x] Skin.cpp - _a bit hacky though_
- [x] Face.cpp

After that I might be able to remove `DrawTriangles` entirely!

Andy

### DIFF
--- a/src/Color.h
+++ b/src/Color.h
@@ -59,6 +59,7 @@ struct Color4ub {
 	Color4ub operator/(const float f) const		{ return Color4ub(Uint8(r/f), Uint8(g/f), Uint8(b/f), Uint8(a/f)); }
 
 	friend bool operator==(const Color4ub& aIn, const Color4ub& bIn) { return ((aIn.r == bIn.r) && (aIn.g == bIn.g) && (aIn.b == bIn.b) && (aIn.a == bIn.a)); }
+	friend bool operator!=(const Color4ub& aIn, const Color4ub& bIn) { return ((aIn.r != bIn.r) || (aIn.g != bIn.g) || (aIn.b != bIn.b) || (aIn.a != bIn.a)); }
 
 	Color4f ToColor4f() const { return Color4f(r/255.0f, g/255.0f, b/255.0f, a/255.0f); }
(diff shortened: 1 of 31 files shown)


fluffyfreak:
Sadly `DrawTriangles` isn't going away just yet and the use of it in `Skin.cpp` remains however I've hacked in re-use of `VertexBuffer` objects within `DrawTriangles` itself for now.

I think I should move it somewhere saner as soon as possible but it's proof of principle.

So far I've only seen it grow to 5 objects as I've eliminated it from everywhere that I can preferring fixed buffers instead.

impaktor:
Is this still WIP?

Maybe @snwyvern or @SpencerBelleau can test it if they are familiar with git and compiling branches. If not, we can just merge it and see what happens.

fluffyfreak:
I'm wondering about moving the hack-job with `DrawTriangles` from being a local static into the `RendererGL` class itself so it can be monitored (_perhaps_) but maybe I should do that in a separate PR.

fluffyfreak:
Right, that's done so now it's ready for review :)

PR #3600, opened 2016-01-20
Fix #ifdef to allow compilation on various BSDs.

christiank:
The "domain" field of struct utsname is a GNU extension. OS X is not the only system which does not have the "domain" field (actually, almost no systems do). So, it makes more sense to check if we're GNU-like instead.

### DIFF
--- a/src/posix/OSPosix.cpp
+++ b/src/posix/OSPosix.cpp
@@ -107,7 +107,7 @@ const std::string GetOSInfoString()
 	}
 
 	char infoString[2048];
-#if defined(__APPLE__)
+#if !defined(_GNU_SOURCE)
 	snprintf(infoString, 2048, "System Name: %s\nHost Name: %s\nRelease(Kernel) Version: %s\nKernel Build Timestamp: %s\nMachine Arch: %s\n",
 		uts.sysname, uts.nodename, uts.release, uts.version, uts.machine);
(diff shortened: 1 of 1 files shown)


fluffyfreak:
@christiank thanks I'll try to get a look at it this weekend.

impaktor:
Also, thanks @christiank for providing the NetBSD [install script](https://wip.pkgsrc.org/cgi-bin/gitweb.cgi?p=pkgsrc-wip.git;a=commit;h=04f0e216e8be28494e76748ec430bd1fd67dec2b)

As for this PR, I'm not familiar with what it does, so I'll leave it in capable fluffy hands.

christiank:
You will notice that my pkgsrc-wip entry actually contains the diff inside this PR. All I'm doing is trying to come up with a portable way to get Pioneer to work (I could have said "ifdef __NetBSD__" instead, hahah)

fluffyfreak:
Actually @salborrelli should probably look into this since it's an OSX change and he usually does our OSX builds. It should just be fine to merge from the look of it though.

@christiank we usually add anyone how submits patches to our `AUTHORS.txt` file, would you like adding and if so under what name? Is "Christian Koch" ok?

christiank:
Yes, Christian Koch is fine.

Issue #3602, opened 2016-01-22
Fix Commodity market layout 


impaktor:
The commodity market changes were merged (#3467) even though some layout love was needed, so I'll open this a a reminder.

Buy/Sell buttons aught to look and function something more like @nozmajner's mockups:
![Buy "tab"](https://lh3.googleusercontent.com/-FvHgNdJplfY/VgcZP0fsr2I/AAAAAAAAEJg/OCG_p5aBYR8/s2048-Ic42/market-buy.png)
![Sell "tab"](https://lh3.googleusercontent.com/-Q6VEhr_RleI/VgcZP8RUk-I/AAAAAAAAEJc/eXjnmRKYKaY/s2048-Ic42/market-sell.png)

impaktor:
Fixed by #3603

Issue #3517, opened 2015-11-07
Cargo assertion fail


impaktor:
This is due to the re-write, #3467 by @joonicks:

As reported on [SSC](http://spacesimcentral.com/ssc/topic/5282-announcement-big-update-adventurous-test-pilots-needed/#entry55844)

> I start at New Hope with a Pumpkinseed, go to the equipment trade and sell everything except the hyperdrive. Free cargo capacity on that screen shows 11t. But the ship's info says max cargo is 10t total. Now I go to commodity market, and here free cargo capacity is shown as 8t. But the trading buttons allow me to "order" a 11t purchase. And if I confirm that purchase, the game displays Lua assertion and closes:

```
error: [string "[T] @ui/StationView/CommodityMarket.lua"]:300: assertion failed!

stack traceback:

    [C]: in function 'assert'

    [string "[T] @ui/StationView/CommodityMarket.lua"]:300: in function <[string "[T] @ui/StationView/CommodityMarket.lua"]:274>
```

impaktor:
fixed by #3603

PR #3603, opened 2016-01-23
Fix for #3602 and #3517

robothauler:


impaktor:
Fast and sweet, great!

I'd like to run it by @nozmajner before merge, if the different size of buy/sell (to mark which is  selected) is a feature?

screenshot-20160123-145627

Nothing to sell:
screenshot-20160123-145704

nozmajner:
Yeah, that small/big button looks a bit odd, but other than that it looks good in my opinion.

impaktor:
I wonder if "we" could implement an invisible/borderless button type so it would look like in:
Buy "tab"

I guess, just check what makes a gray button gray, and make it look invisible (background blue) instead? I expect that to (maybe) be on the C++ side. Perhaps for another day. This PR was just opened afew hours ago, so let's se what others say.

robothauler:
The different size of the buttons should be a visual indicator of the mode (buy/sell).
Just for completeness, the sell mode.

screenshot-20160123-152621

impaktor:
I've implemented the Invisible state, here is in-game screen shot:
screenshot-20160123-165942
screenshot-20160123-165920

screenshot-20160123-165922

PR on the way.

fluffyfreak:
@impaktor So is this for review or will you merge it at the same time as #3605?

impaktor:
We could either merge it now, or if #3605 goes in first, this PR should make use of that feature and be updated. I'm fine with this being merged now. It does fix a bad bug, so it has that going for it.

fluffyfreak:
done :)

PR #3605, opened 2016-01-23
Invisible disabled button for commodity market

impaktor:
## Purpose

At the moment disabled buttons are always grey. This implements a state where disabled buttons just "blend" into the background. This is needed to make the suggested PR #3603 look like @nozmajner's mockups in #3602.

## Before / After Example
(applied to #3603 )
screenshot-20160123-145627

screenshot-20160123-165920

## Implementation

No more than 3 brain cells were harmed during this implementation. It's a pure copy-paste-rename operation on the `Disabled()` state of `ui/Widget.{cpp,h}`, in the most stupid way possible.

Now Invisible state is done by not having the widget being a `BorderedRectBox` but a  `Color` instead.

## Implement sanity into the code

Like I said, it works, but feels as wrong as those funny feelings you had when seeing that distant hot cousin each Christmas. so I think this should be considered before merge:

- [x] 0. Use more than 3 brain cells

- [ ] 1. How should it be used in Lua? Right now you do `button:Invisible()` instead of `button:Disable()`. Perhaps `button:Disable(arg)` where `arg` sets what kind of disable you want: (grey, or invisible)?

- [x] 2. Does the Skin.ini even need to be edited at all? Just use the flat background colour from code directly? __Done__, (use `Color` instead of `BorderedRectBox`)

- [x] 3. most likely some dead code in there, due to stupid copy-paste-rename operation on my part. Remove it

- [x] 4. Rename member functions and attributes from `Invisible()` (to `Hide`? `DisableHide()`?) to something that doesn't clash with how that word is already used in the code for other things.

- [x] 5. Clean up commit history. 

_PS. I don't have any cousins, but it's better this way_

### DIFF
--- a/data/ui/StationView/CommodityMarket.lua
+++ b/data/ui/StationView/CommodityMarket.lua
@@ -346,7 +346,7 @@ local commodityMarket = function (args)
 		if Game.player:CountEquip(tradecommodity) == 0 then
 			trade_mode = trade_mode_buy
 			showbuysellbutton = confirmtradebuy
-			buyfrommarket:Disable()
+			buyfrommarket:Hide()
 			buyfrommarket:SetFont("LARGE")
 			sellfromcargo:Enable()
 			sellfromcargo:SetFont("SMALL")
@@ -364,7 +364,7 @@ local commodityMarket = function (args)
 		trade_mode = trade_mode_sell
 		tradeamount = 0
 		showbuysellbutton = confirmtradesell --change which confirm button to show
-		sellfromcargo:Disable()
+		sellfromcargo:Hide()
 		sellfromcargo:SetFont("LARGE")
 		buyfrommarket:Enable()
 		buyfrommarket:SetFont("SMALL")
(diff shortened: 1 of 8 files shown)


impaktor:
Almost there. 

I feel bad putting up this crock of shit yesterday. But is almost done now to the best of my current abilities.

> The term "a crock of shit" derives from an ancient Roman custom that coincidentally took place in Roman times. It referred literally to a pot into which people would excrete if they were particularly bored by whichever freelance philosoper happened to be talking rubbish at the time. The Roman empire employed crock-monitors who were each assigned to a philosopher, and it was their job to monitor the pot (or crock). Should the crock become full, it would be presented to the philospher, who was obliged, by law, to announce that it bore a remarkable resembence to himself, thus proclaiming he was full of crap and was, in fact, talking a crock of shit.
"I am talking a crock of shit", Socretes 429 BC

fluffyfreak:
They should bring that back for [prime minister's questions](https://en.wikipedia.org/wiki/Prime_Minister%27s_Questions)...

fluffyfreak:
This code looks fine (**as I've just said on IRC but want a record of it for posterity**), so I think that it's ready for review and merge unless someone disagrees or @impaktor has something else on the todo list?

impaktor:
> or @impaktor has something else on the todo list?

Nope. As I said (also on IRC, but repeated here for the future radioactive ape-squid-men), I had wanted to hear what @robn though would be the "valid" approach for doing this.

PR #3607, opened 2016-01-24
Search&Rescue Script: fix serialization issue (#3606)

### LINKED ISSUE #3606: Crash on save
1. Jump to an inhabited system.
2. save 

`error: Lua serializer '(null)' tried to serialize an invalid 'SpaceStation' object`

clausimu:
During the last update to the script I realized that aircontrol characters were never saved. Fixing this surfaced an old bug with the original script that was hidden before (using station object as table key instead of station.path object).

This should fix  #3606 and allow saving the game again.


walterar:
:+1:

PR #3610, opened 2016-01-25
Fixed narrowing conversion problem when compiling with C++11 support

salborrelli:
When compiling using the C++ Standard Library with C++11 support (environment: xCode on Mac OS X 10.10 - Yosemite), I recently started getting the following errors on several lines of "../contrib/vcacheopt/vcacheopt.h":

**Non-type template argument evaluates to 4294967295, which cannot be narrowed to type 'int'**

After some analysis I found that the error is caused by the narrowing conversions required to convert UINT_MAX to int in several typedef instructions, for example:

*typedef TVertexCacheData<unsigned int, UINT_MAX >		VertexCacheDataUInt;*

The problem is related to the c++11 standard which at 8.5.4 item 3 says:

*If a narrowing conversion (see below) is required to convert any of the  arguments, the program is ill-formed.*

and item 7:

*A narrowing conversion is an implicit conversion ... from long double to double or float, or from double to float, except where the source is a constant  expression and the actual value after conversion is within the range of values that can be represented  (even if it cannot be represented exactly)*

My solution to this problem has been to set an explicit type cast on the UINT_MAX value, wherever it is used as a template argument.

### DIFF
--- a/contrib/vcacheopt/vcacheopt.h
+++ b/contrib/vcacheopt/vcacheopt.h
@@ -67,7 +67,7 @@ public:
 	}
 };
 typedef TVertexCacheData<int, INT_MAX >					VertexCacheDataInt;
-typedef TVertexCacheData<unsigned int, UINT_MAX >		VertexCacheDataUInt;
+typedef TVertexCacheData<unsigned int, static_cast<int>(UINT_MAX) >		VertexCacheDataUInt;
 typedef TVertexCacheData<unsigned short, USHRT_MAX >	VertexCacheDataUShort;
 
 template <typename T, int ERR_VAL>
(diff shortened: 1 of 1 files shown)


salborrelli:
@fluffyfreak @johnbartholomew Hi guys, looks like you are the main contributors for the single file this PR is about (contrib/vcacheopt/vcacheopt.h). Could you please take a quick look at my proposed change and let me know if you see any potential issue?

If you don't then I can merge and close the PR myself. Thks!

fluffyfreak:
I'll take a look this evening :)

No rush on getting it committed.

salborrelli:
:thumbsup:

fluffyfreak:
Looking at the code it seems that it's me that was wrong to have put the `UINT_MAX` and `USHRT_MAX` in the first place. I believe that both of them shuld just be `INT_MAX` as that defines the `ERR_VAL` returned when something has gone wrong.

I'm just double checking that on my linux machine now.

fluffyfreak:
Ok yes that is the problem, I appear to just be an idiot!

I've created another PR #3613 with all of them changed back to `INT_MAX`. I'm surprised we ran for so long with it like this!

Thanks for finding this @salborrelli i'll close this PR and we'll use the other one to avoid more Git history than we need :)

salborrelli:
Great and no problems, the error shows only if the compiler is configured with support for C++11 (and later, I suppose); otherwise you get nothing, not even a warning :smile:

laarmen:
Given that we use C++11 ourselves, it is a bit odd it hadn't blown up in
our faces already.

Issue #3604, opened 2016-01-23
Problems with text messages


walterar:
Example: Try to buy without enough cash.

impaktor:
Maybe @fluffyfreak knows? 

![screenshot-20160123-152733](https://cloud.githubusercontent.com/assets/619390/12530661/02cfbd68-c1e6-11e5-846f-02d61e076ab6.png)

fluffyfreak:
Hmm, probably a bit of over-optimisation.

walterar:
Another problem with the text.

![text-wrong](https://cloud.githubusercontent.com/assets/1334989/12536600/d62a3038-c288-11e5-8cf8-4668773a13a0.png)

fluffyfreak:
How do you get that first screen? I just get a message saying insufficent funds but not as a popup?
That or the screen stops me from buying it at all.

walterar:
Test in Equipment Market

fluffyfreak:
That second problem with the multiline text is completely bizarre. There is no reason that I can fathom for it to be happening, and it goes away after the first jump.

walterar:
The Radar Mapper It looks very ugly too.

fluffyfreak:
ok, new PR :)

walterar:
#3612 fix it, thanks :+1:

PR #3612, opened 2016-01-26
Fix bad text size.

fluffyfreak:
We fail to notice that the size has changed or needs to change.

@walterar this should fix the rest of #3604


walterar:
Yesss... :) :+1:

PR #3614, opened 2016-01-27
Fix number of police ships not being spawned in right amount

impaktor:
@clausimu  pointed out this bug.

Police were spwning in opposite probability as they should, with respect to `lawlessness`.

Just to check sanity: here is the new _average_ number of spawned police ships for a stion with five pads, and a station with nine or more (caped). x-axis stops at 0.99, thus not shown is `lawlessnss=1`, in which case no police will be spawned.
ships

### DIFF
--- a/data/libs/SpaceStation.lua
+++ b/data/libs/SpaceStation.lua
@@ -339,7 +339,7 @@ function SpaceStation:LaunchPolice(targetShip)
 		-- decide how many to create
 		local lawlessness = Game.system.lawlessness
 		local maxPolice = math.min(9, self.numDocks)
-		local numberPolice = math.ceil(Engine.rand:Integer(1,maxPolice)*lawlessness)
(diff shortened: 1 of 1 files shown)


laarmen:
Yay, plots!

clausimu:
:+1:  for showing plotted average values and the quick fix!

fluffyfreak:
I think you should document how you generate those graphs, I think all our PR's should have graphs, even when they're not relevant :)

Also the fix looks ok, feel free to merge when ready!

laarmen:
+1 on the "all our PR's should have graphs"
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Cet
e-mail a été envoyé depuis un ordinateur protégé par Avast.
www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DDB4FAA8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

PR #3617, opened 2016-02-01
Fix Character.lua / Ship.lua import relationship

clausimu:
This is a fix for the Ship:GenerateCrew() method - and the game crash if Character.lua is imported by Ship.lua

Character.lua was importing Ship.lua but didn't use it. Instead, Ship.lua needed to import Character.lua for its Ship:GenerateCrew() method. But importing Character.lua there would cause the game to crash (circular reference...). 

I don't know why this was never noticed before - I guess nobody every used Ship:GenerateCrew(). It is working now.

### DIFF
--- a/data/libs/Character.lua
+++ b/data/libs/Character.lua
@@ -52,7 +52,6 @@ local Game = import("Game")
 local Event = import("Event")
 local NameGen = import("NameGen")
 local Serializer = import("Serializer")
-local Ship = import("Ship")
 
 local Character;
 Character = {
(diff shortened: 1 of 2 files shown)


impaktor:
Looks good. :+1: 

Too lazy to merge now. 

@salborrelli can merge this if he wants to.

fluffyfreak:
To lazy! Merge-from-the-hip! :P

impaktor:
:smile:

PR #3621, opened 2016-02-15
SAR - fix termination of mission for partially completed missions

clausimu:
This is a fix for the SAR (Search and Rescue) mission script where partially completed missions were treated as completed by the target ship (wrong!) without properly closing the mission and paying the player (right!). Taken together, this could lead to missions that the player cannot complete and never close.

This fix does not address the issue where fuel can't be transferred to target ships even though the player has enough on board. I still can't reproduce that bug on my system.

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -1581,50 +1581,60 @@ local interactWithTarget = function (mission)
 		      if actiontime then
 			 
 			 -- pickup crew from target ship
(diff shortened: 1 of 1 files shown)


impaktor:
You're adding `partial` in one commit and removing it in the next?

Looks like a good time to learn how to squash / fixup commits. Check pioneer wiki on editing commit history.

clausimu:
My bad. I just fixed the commit. I hope it is correct and clean now.

fluffyfreak:
@clausimu I don't know how else to conttac you but there's a report, with save games, on the developer of some minor issues http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=449

PR #3625, opened 2016-02-21
Fix terrain cracks

fluffyfreak:
I was a bit too aggressive with the crack avoidance values.
Eased off the multiplier a bit to avoid terrain cracks at low-patch-detail levels.


impaktor:
Looks like it's including the commit from #3624. I guess git is smart enough to handle that.

fluffyfreak:
@impaktor yeah I was just going to merge them in-order :)

PR #3629, opened 2016-02-27
Moar profiling

fluffyfreak:
Profile the modelcompiler and startup process.

As expected the texture creation/loading takes a long time, ~20% of totall startup time. 
Another heavywieght is Factions finding their best-fit home system!

**EDIT**
I've now deleted the auto-generation script for the factions and replaced it with 103 faction scripts!
That sound like a lot but it's also 1/5th of the loading time gone.

The script tiself has been archived with the name `XX_AutoGenerate.OLD` in case anyone wants to ressurect it in future or use it elsewhere.
I also fixed an odd issue where it never bothered to persist the `bodyIndex` value into the SystemPath which meant that Factions with a homeworld still didn't know which one it actually was.


impaktor:
Well, the plan was to remove auto-generated factions all together, and only have hand scripted factions.

So, maybe a future system where we just give the home system coordinates, and then the expansion variable does some kind of sphere around it, weighted in some way where it instersects with other spheres? (disclaimer: I don't know how current code works for these things, maybe that's how it is done today?)

fluffyfreak:
Well 'SetBestFitHomeworld' gets called 115 times and causes 20% of the loading time. 
That's about the same as creating all of the textures!

impaktor:
> I've now deleted the auto-generation script for the factions and replaced it with 103 faction scripts!

Holy moly, I missed that edit. I was quite surprised seeing +100 files in `factions/`-folder. Well, at this point, I wonder why even have them? We just want 3 factions, right? And if any other should be added, it aught to be by hand. Either way, something I'll think about in +2-3 months.

Issue #3631, opened 2016-02-27
linux core dumped error when try to show star system view


sfence:
My Pioneer is builded from git: d3eb57e
Running on linux

Sometimes, it happens to me, that when I try to show information about star system (F7), game failed with core dumped error in terminal output. Without any warning.

http://147.229.208.36/pioneer/SYSTEMVIEWFAIL_LINUX_CORE_DUMPED

fluffyfreak:
That link says access forbidden.

sfence:
i'm sorry. Now file should be aviable.

sfence:
From testing, it looks like it happens when you have selected navigation target like ship and try to show star system view.

fluffyfreak:
Ah good to know, thanks for testing @sfence :)

fluffyfreak:
Yeah I can repeat it now, should have a fix shortly.

fluffyfreak:
There we go, fix PR'd.

fluffyfreak:
This should be fixed by #3632 which has just been merged.

PR #3632, opened 2016-02-28
Only get the SystemBody pointer on system bodies and NOT on ships.

fluffyfreak:
This should fix the #3631 crash which was caused by it dereferencing a nullptr.


fluffyfreak:
Feb 29th merge's come but once every four years!!!

impaktor:
This doesn't prevent the player from targeting ships? Like Ctrl-click to set as frame of reference is useful in combat, and docking missions.

fluffyfreak:
Hmm, I didn't think of that, but I don't think so because if it ever went thise code path then it was trying to view the "SystemInfo" for a ship, which never makes any sense.

PR #3634, opened 2016-03-01
Ability to set Combat and Nav targets to null validly.

fluffyfreak:
This is for @walterar, the game can set combat and nav targets to `nullptr` to disable them, so the Lua should be able too as well.

### DIFF
--- a/src/LuaPlayer.cpp
+++ b/src/LuaPlayer.cpp
@@ -40,7 +40,7 @@ static int l_player_is_player(lua_State *l)
  *
  * Status:
  *
- *   experimental
+ *   stable
  */
 
 static int l_get_nav_target(lua_State *l)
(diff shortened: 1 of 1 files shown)


laarmen:
I dunno about the whole experimental -> stable migration, but other than that, :+1:

impaktor:
> experimental -> stable ...

I suspect that description doesn't have much meaning anyway. Add to that, that the online LuaAPI for pioneer isn't updated very often. (Like once a year).

walterar:
This is exactly what I needed. 
Thanks Andy! :+1:  :)

fluffyfreak:
@walterar that's ok, just shout if you need things like this, sometimes it's just an implementation oversight.

@laarmen / @impaktor I figure something that's been in since Aplha 14 is pretty "stable" ;)

PR #3639, opened 2016-03-05
Ship stat planner addon for Blender. Exports ship.json

nozmajner:
I thought I update that ship stat planner addon I made for Blender a while ago with the changes in the equipments. I also gave it another go with the export functionality, and it seems to be working, so here it comes.

It should be pretty straightforward to use (at least I hope). It can be found in the Scene settings in the Properties window.
You can set up everything in it, no need to edit the output.json apart from renaming it if name's not good.
It calculates accelerations with full and empty ship, and also calculates deltaV.
For most of the equipments it only includes them in the file if they aren't on their default values. I kept some lines around even if they are at their default though, like cargo capacity and so.

I did some testing and it worked fine for me, but I guess I made some mistakes which weren't obvious.
One thing I noted that the thrust values were missing the .00 part of the number, if it was a whole number, but it seems the game didn't freaked out on it.

Hope I put it in the proper directory: Scripts.
shipplanner_screen2

### DIFF
--- a/scripts/ShipPlanner015.py
+++ b/scripts/ShipPlanner015.py
@@ -1,0 +1,682 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software Foundation,
+#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# ##### END GPL LICENSE BLOCK #####
+# This is an addon for Blender, that provides some tools to calculate
+# ship stats, and exports it next to the blend file with the same name.
+
+import bpy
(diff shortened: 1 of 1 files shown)


scripts/ShipPlanner015.py
```diff
@@ -0,0 +1,682 @@
...
+
+	@classmethod
+	def poll(cls, context):
+		return context.active_object is not None
+
+	def execute(self, context):
+		print("Krumpli") 
+		scene = bpy.data.scenes["Scene"]
+#		file = open(filepath, 'w', encoding="ansi", newline = "\n")
+		file = open(os.path.splitext(bpy.data.filepath)[0] + ".json", 'w', newline = "\n")
```
impaktor: The other comments have their "#" indented. I guess that's more sane, considering python's indention-pickiness.

scripts/ShipPlanner015.py
```diff
@@ -0,0 +1,682 @@
...
+		
+		
+	
+		file.close()
+		
+		return {'FINISHED'}
+	
+
+class ShipPlanner(bpy.types.Panel):
+		"""Ship stats planner utility for Pioneer"""
```
impaktor: Is this whole function double indented? (two Tabs?). Is that on purpose?

scripts/ShipPlanner015.py
```diff
@@ -0,0 +1,682 @@
...
+	if dv <=0.0001:
+		result = "<0.0001"
+	else:
+		result = dv
+	return result
+
+
+ #Operator example
+class VesselExport(bpy.types.Operator): # , bpy_extras.io_utils.ExportHelper .ExportHelper feldobja az export ablakot, ha megnyomom a gombot
+	"""Export shipdef.lua"""
```
impaktor: Hm, comment langauge spirals down into some insane vampyre lingo.

PR #3641, opened 2016-03-07
Bowfin fighter for Kaluri/CIW

nozmajner:
Bowfin fighter, about the CIW counterpart of the Kanara. Relatively low deltaV but good acceleration. Small capacity, even smaller cargo hold. A ship of tradeoffs for the player, which ought to make it interesting.
Next: I want to turn this ship into the CIW police craft, but that needs texturing.
screenshot-20160307-153708


laarmen:
I'm not really good at physics, but aren't acceleration and delta-V
basically the same thing? How can one be low and the other one good?

Looks nice anyway :)

fluffyfreak:
@laarmen I had to [look that up](https://www.reddit.com/r/askscience/comments/3sdnzu/why_deltav_and_not_acceleration/) too :)

impaktor:
smbc-id4040

laarmen:
So delta-v would be some kind of "bang-for-your-buck" measure akin to the
fuel efficiency?

nozmajner:
Yupp, something like that. The larger the deltaV, the faster you will go in the end, even with miliG accelerations, so it gets you there sooner. I usually translate this to range in my head, like you don't want to go too far with a low deltaV ship, because it might taake months, instead of days in a another ship. This why the Lunar shuttle has lower deltaV.
And acceleration doesn't matter too much for this, at least on interplanetary ranges. (Once I got a day or so shorter trip with a lower thrust sinonatrix, but same deltaVwhen I was doing some stats comparison.)

laarmen:
So, a fighter would need a powerful acceleration, but not so much a high
delta-V, right? Since they don't need to be good at interplanetary travels,
they usually hang around one specific spot.

nozmajner:
Yupp, that's the idea

PR #3642, opened 2016-03-07
Kanara resize (75%)

nozmajner:
When I was working on the Bowfin, it became apparent, that the Kanara is way too big. Like two people could almost sit next to each other in the cockpit easily.
So I shrinked it by 75%, and adjusted the stats along with it. It's accelerates better than the Bowfin, but has worse capacities and deltaV.
Also while I was there, I fixed a tiny mesh error which put some stray polygons near the engine.
And I made a quick and subtle normal map with the gimp plugin.


impaktor:
Even smaller? Now you'll be able to _really_ explore the space station.

nozmajner:
Damnit, @impaktor is right. Shrinked _it_ to 75%. :)

PR #3643, opened 2016-03-09
Remove redundant LuaAPI function for Disable/Enable buttons.

impaktor:
This PR has three purposes:

1. Show that I'm still alive.
2. Remove code that I think is redundant
3. an excuse to poke at @johnbartholomew, and see if he still lives :smile:

So on point 3: In October 2013 you added an additional way to disable/enable buttons, with `button:SetEnabled(bool)` @bc73254f0252c855751e56fe4532d46be8518020 but I don't see how that method adds anything that isn't already there with `button:Disable()`, and `button:Enable()`. Sure, that method saves you the need to wrap the Disable/Enable in an if-statement in some cases, but two different API functions for doing the same thing seems odd to me, just for that one-line trick in Lua. I could be completely wrong of course.

### DIFF
--- a/data/ui/FileDialog.lua
+++ b/data/ui/FileDialog.lua
@@ -39,10 +39,10 @@ ui.templates.FileDialog = function (args)
 	cancelButton.onClick:Connect(onCancel)
 
 	if #files > 0 then
-		selectButton:SetEnabled(true)
+		selectButton:Enable()
 		list:SetSelectedIndex(1)
 	else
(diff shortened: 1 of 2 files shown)


johnbartholomew:
I'm still alive :-)

I probably just failed to find the existing Enable/Disable methods when I
added that. Or maybe I thought it was worth the extra method at the time, I
don't know. But sure, you can remove it, I have no objection.

impaktor:
Cheers!

PR #3644, opened 2016-03-10
RandomColor Port for better random colours

fluffyfreak:
This is all @robn [fault](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=221) ... weekend my arse.

So I found the [C# port](https://github.com/nathanpjones/randomColorSharped) of the original JavaScript evilness and thought to my self `Well, how long can that take?!?`, about an evening and a bit it turns out!

To see this in action I suggest building Pioneer as usual, using `-mv wave` as the parameter, picking pattern number 6 from it's dropdown, and then repeatedly clicking the `Random Colour` button to see what it throws up.

My only query regarding merging is about the license, it was MIT and so that's what I've put at the header instead of our our Pioneer GPL license for now. I didn't want to restrict it's use by others in some way but I can change that if necessary.

PS: sorry, not got any pictures to show of it here.
PPS: If people (artists) want a build for a platform to test it, then let me know and I'll upload some somewhere.

Andy

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -63,6 +63,7 @@ The Pioneer Developers are:
 * Marcus de Roos
 * Michele Botticelli
 * Miha Vitorovič
+* Nathan Jones
 * Nic
 * Nick Wanstall
 * Oleg Ovcharenko
@@ -199,5 +200,6 @@ and services, without which Pioneer would not be possible:
   assimp        http://assimp.sourceforge.net/
   MXE           http://mxe.cc/
 
-  GitHub        https://github.com/
-  SourceForge   https://sourceforge.net/
+  GitHub        		https://github.com/
+  SourceForge   		https://sourceforge.net/
+  randomColorSharped	https://github.com/nathanpjones/randomColorSharped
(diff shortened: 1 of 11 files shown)


robn:
Obviously when I said "weekend" I wasn't accounting for someone of your skill level and good looks.

MIT is fine, though you should add a thing to AUTHORS about it (the clause about including a permission notice). It's a bit fuzzy but that's probably good enough. Then again, if it's truly copyright, you can do whatever you want with it :)

fluffyfreak:
Done :) added him to `AUTHORS.txt` in the list and at bottom with link to his code. Probably not quite necessary but I think it's right.

impaktor:
tested it. Neat.

fluffyfreak:
Hmm, this has a tiny bit too much randomness in it, it's different everytime for everything. 

So buildings for a city are always different for example. Will need some tweaking before merging.

laarmen:
Well, last time I said "a couple of days at most" at work was two weeks
ago. I've finally finished the thing in question, basically rewriting the
entire app in the process. So you are very lucky, my friends (or skilled
and good looking, apparently ;-) )

Issue #3648, opened 2016-03-11
Game throws up " lua serializer '(null)' tried to serialize an invalid 'ship' " and crashes when attempting to save


wafehling:
This error is reproducible. Here are the following steps:
1. Download this save file and add it to the usual save file location for pioneer (My documents/pioneer/savefiles)
[_autosave2.zip](https://github.com/pioneerspacesim/pioneer/files/168662/_autosave2.zip)
2. Load this up in pioneer. Version where error occurs is 20160307 (33328ed)
3. jump to Barnard's star, which is already selected. All you have to do in-game is unpause and hit F7.
4. Attempt to manually save. Game will crash, giving that exact error. Error will also occur upon docking with "System Administration Resting" orbital starport with autosave enabled. Have not tested with other locations, but assumed it will happen anywhere in system.

wafehling:
Targeting a different ship from the one currently set as combat target in original save file does not prevent the error.

wafehling:
This star system is the first system where I performed a search and rescue mission. Jumping to Sirius and attempting to save produces the same crash.

I highly suspect the crash is caused by the game attempting to save the ship spawned for the search and rescue mission, and failing somehow.

wafehling:
I flew out and destroyed the ship, saving immediately before. The save right before his death was fine, the save right after his death failed. The ship appeared to not have left the system properly from the discardShip function found in searchrescue.lua. 

It seems the error narrows down to a ship not being marked as destroyed/gone when attempting to save. I'll see if I can figure out how to get a debugger working and run it through there, to narrow things down further.

impaktor:
Ping @clausimu

laarmen:
My guess is that it has something to do with the recent changes to the
setCombatTarget Lua method.

wafehling:
I've fully reproduced the bug. Simply accept a fuel resupply emergency mission, successfully complete it, and hyperdrive away to another system shortly after completing it. When you attempt to save in the new system, it. will. crash.

walterar:
@laarmen That's a joke?

The problem is that are trying to serialize a "dead" ship.

clausimu:
@wafehling - thank you. I will look into it. 

There must be a sensor somewhere in the SAR script that sets it up to fail exactly when my work-life is at peak capacity... :)

fluffyfreak:
[...]

PR #3649, opened 2016-03-11
Release discarded_ships on entering or leaving systems

fluffyfreak:
This does three things :
- use nil instead of "NIL"
- release any discarded_ships when entering or leaving a system
- check that we have a target before trying to use it

The first is just me being pedantic, the second I think fixes the problem but may have side effects, and the third stops it crashing which is one of the side effects.

@clausimu this is your baby I just got distracted by it so if you could take a look and tell me what you think then that'd be great :)

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -1836,19 +1836,20 @@ local onLeaveSystem = function (ship)
 end
 
 local onShipDocked = function (ship, station)
-   if ship:IsPlayer() then
-      for _,mission in pairs(missions) do
-	 if Space.GetBody(mission.station_local.bodyIndex) == station then
-	    closeMission(mission)
-	 end
-      end
-   else
-	for i,discarded_ship in pairs(discarded_ships) do
-	if ship == discarded_ship then
-		discardShip(ship)
-		table.remove(discarded_ship,i)
+	if ship:IsPlayer() then
+		for _,mission in pairs(missions) do
+			if Space.GetBody(mission.station_local.bodyIndex) == station then
+				closeMission(mission)
+			end
+		end
+	else
+		for i,discarded_ship in pairs(discarded_ships) do
+			if ship == discarded_ship then
+				discardShip(ship)
+				table.remove(discarded_ship,i)
+			end
+		end
 	end
-   end
 end
 
 local onReputationChanged = function (oldRep, oldKills, newRep, newKills)
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -750,7 +750,7 @@ local onChat = function (form, ref, option)
 	 problem            = ad.problem,
 	 dist               = ad.dist,
 	 flavour            = ad.flavour,
-	 target             = "NIL",
+	 target             = nil,
 	 lat                = ad.lat,
 	 long               = ad.long,
 	 shipdef_name       = ad.shipdef_name,
@@ -1569,7 +1569,7 @@ local interactWithTarget = function (mission)
 		      
 		      -- abort if interaciton distance was not held or target ship destroyed
 		      -- TODO: set the check mark for each mission right
-		      if not targetInteractionDistance(mission) or mission.target == "NIL" then
+		      if not targetInteractionDistance(mission) or mission.target == nil then
 			 Comms.ImportantMessage(l.INTERACTION_ABORTED)
 			 searchForTarget(mission)
 			 return true
@@ -1651,7 +1651,7 @@ end
 function searchForTarget (mission)
    -- Measure distance to target every second until interaction distance reached.
 
-   if mission.searching == true or mission.target == "NIL" then return end
+   if mission.searching == true or mission.target == nil then return end
    mission.searching = true
 
    -- Counter to show messages only once and not every loop
@@ -1813,6 +1813,10 @@ local onEnterSystem = function (player)
 	 mission.target = createTargetShip(mission)
       end
    end
+	  
+      for _,discarded_ship in pairs(discarded_ships) do
+		table.remove(discarded_ships, discarded_ship)
+	  end
 end
 
 local onLeaveSystem = function (ship)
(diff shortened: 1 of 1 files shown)


fluffyfreak:
Oh yes this is to fix #3648

clausimu:
@fluffyfreak - thank you. I will definitely take a look. I will also have to double check on your replacement of nil with "NIL". I do remember specifically deciding against using nil for some reason. I want to make sure putting nil in doesn't break things.

impaktor:
@clausimu 

For stings like these:
```lua
if mission.searching == true or mission.target == nil then
```
You can write it as
```lua
if mission.searching or not mission.target then
```
provided `mission.searching` is either `true` or `false` (any value not `nil` or `false` is evaluated as true, I think), but I guess your way is clearer.

fluffyfreak:
@clausimu any luck looking into this?

impaktor:
@fluffyfreak clausimu was talking about this on IRC the other day. He has changes and stuff he'd like to add.

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -1813,6 +1813,10 @@ local onEnterSystem = function (player)
 	 mission.target = createTargetShip(mission)
       end
    end
+	  
+      for _,discarded_ship in pairs(discarded_ships) do
```
clausimu: Not sure why you are emptying "discarded_ships" here too? Shouldn't the insertion within "onLeaveSystem" (see line 1835 below) take care of it? Or am I missing a situation where you are not leaving a system but entering another?

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -1813,6 +1813,10 @@ local onEnterSystem = function (player)
...
+		table.remove(discarded_ships, discarded_ship)
```
clausimu: I don't think this works. I think you have to do: table.remove(table_to_remove_from, index_of_item) instead of table.remove(table_to_remove_from, item_itself).

Might be faster to just overwrite the whole table with discarded_ships = {}?

clausimu:
@fluffyfreak I just entered a few comments into the file. @impaktor suggested this way. There is also one completely stupid oversight of mine (which is why the failchecks you inserted are even necessary). Whenever a ship is "discarded" it is hyperjumping away - but kept in the "discarded_ships" table!

Can you change this section (line number 1842 in the original file, within the function "onShipDocked"):

`for _,discarded_ship in pairs(discarded_ships) do`
`if ship == discarded_ship then`
`discardShip(ship)`

to this:

`for i,discarded_ship in pairs(discarded_ships) do`
`if ship == discarded_ship then`
`discardShip(ship)`
`table.remove(discarded_ship,i)`

Your replacement of "NIL" with nil should work. I went over the file again and can't really figure out why I chose "NIL" over nil.

data/modules/SearchRescue/SearchRescue.lua
fluffyfreak: This is just a [belt & braces](http://www.phrases.org.uk/meanings/61250.html) approach, did it just in case.

data/modules/SearchRescue/SearchRescue.lua
fluffyfreak: Ok I'll do it that way :)

fluffyfreak:
Ok, implemented all those changes, I really think we should find a lua beautifier and run all our code through it.
It's a horrid mix of tabs and spaces right now.

impaktor:
I'm a space-indent kind of guy, but all C++ code is Tabs (3 chars?) indented, then space for indentation of args and stuff. 

For Lua, I use the same style that the current file uses. Actually, that's what my emacs does by default for C++ files as well.

Since clasimu uses emacs, he might be interested in that I use dtrt-indent, which auto detects and sets the indentation style when you open a file, so I never give this much thought. [my emacs dtrt-indent](http://home.thep.lu.se/~karlf/emacs.html#sec-20-6-2). Only caveat is when starting a new file from scratch of course.

fluffyfreak:
All the C++ is tabs indented (==4-spaces) I believe but there's probably islands of horror within it.

I have no idea what the lua code indentation is besides bad, it's a mix of tabs and N-spaces, but the spacing doesn't seem consistent.

fluffyfreak:
@clausimu I've tried it with those changes, it seems that missions don't complete now :/

I've only had chance to try a "replacement pilot" missions but no matter how close I got it never trasnsferred the pilot.

clausimu:
Let me give it a try. I just downloaded your file.

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -1824,28 +1826,30 @@ local onLeaveSystem = function (ship)
...
+		for _,mission in pairs(missions) do
+			if Space.GetBody(mission.station_local.bodyIndex) == station then
+				closeMission(mission)
+			end
+		end
+	else
+		for i,discarded_ship in pairs(discarded_ships) do
+			if ship == discarded_ship then
+				discardShip(ship)
+				table.remove(discarded_ship,i)
```
clausimu: You need to replace "discarded_ship" with "discarded_ships" in this line.

clausimu:
There is one little change I had to do (see comment I posted a second ago in the file). After that I was able to complete a refueling mission and also save the game successfully after hyperjumping to a different system.

When you tried the "replacement pilot" mission - what kind of message did you see pop up? This issue might be unrelated to what you are trying to fix with this merge request here. I'll have to try one.

fluffyfreak:
I didn't see any message pop up, at all.

laarmen:
Where are we at with this one, @clausimu, @fluffyfreak ?

fluffyfreak:
I got as far as a short test of @clausimu changes and then no further, I was really just trying to help by starting this PR as this amount of Lua is a bit out of my depth.

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -750,7 +750,7 @@ local onChat = function (form, ref, option)
 	 problem            = ad.problem,
 	 dist               = ad.dist,
 	 flavour            = ad.flavour,
-	 target             = "NIL",
+	 target             = nil,
```
laarmen: Just removing the line is equivalent. Any key matches nil by default.

laarmen:
Oh hell, merging this as is. The way I see it, it only makes things better anyway.

PR #3653, opened 2016-03-16
Bowfin: Deleting some duplicate .dae mesh files

nozmajner:
For some reason I had the three LoD levels for the bowfin duplicated: bowfin_hi.dae and bowfin-hi.dae respectively. I noticed it accidentally, so let's delete them.


walterar:
I believe also have a error in the file json
```
	"slots" : {
		"cargo" : 6,
		"engine" : 13, <---
```

nozmajner:
Yeah, I wanted to ask the equipment guys about that, how do you define max engine, if you want them to be military engines? (@laarmen, @impaktor )

walterar:
No need to define it in json. That makes the script.
Deletes the line engine.

fluffyfreak:
Max engine? It's clamped between 0 and 1 in the loading code so `"engine" : 13,` is just limited to 1 anyway.

What are you trying to achieve and I'll see if I can figure it out.

fluffyfreak:
Otherwise is this ready to merge as is? We can sort out the other things separately.

nozmajner:
Yes, apart from that 13 it's good.
I wanted to allow military drives for it too.

fluffyfreak:
I think that if a ship can have hyperspace drives then it can also have military ones. It's just that not everywhere sells them.

walterar:
@nozmajner This is what you want?
bowfin

Issue #3627, opened 2016-02-24
Search&rescue bug - with save game


sfence:
My Pioneer is builded from git: d3eb57e

I have tried fuel and crew rescue missions, but both fails because of transfer problem.
For fuel rescue, game console wrote me that hydrogen miss for transfer. But that time I have 3t of hydrogen in cargo space, ship needed 1t.
And for crew transport on ship. passager transfer failed, because of no aviable space for crew on target ship. Secundary problem was ship location. System orbit view (F6) show the target ship on orbit of sun, instead of planet, where ship was. But, after selecting ship as a target, autopilot works fine. For a situation, it can be helpfull, I have uploaded two saves from crew transport on address: http://147.229.208.36/pioneer/
May be, it will help us to solve the bug. Good luck and thank us for work on this game.

clausimu:
Thank you for reporting the bug! I have so far been unable to reproduce it on my system. I will take a look at the save games and hopefully they will allow me to track down the problem. If you can, could you upload a save game where you had an issue transferring fuel as well? Seeing the problem in action is going to be very valuable!

sfence:
I'am sorry. I have tried to reproduce fuel problem but without success. But I get to different types of errors. On the same address, files with prafix LANDFUEL and FUEL. First one, LANDFUEL, I am not able to land, to finish transfer. Probably because of terrain profile. Second, FUEL, it looks like game not recognized, I am in transfer area to allow transfer.

sfence:
May be, it can be needed information. I am running pioneer under Linux.

walterar:
@clausimu This fixes the problem of missing fuel.
```
local cargoPresent = function (ship, item)
-- Check if this cargo item is present on the ship.
   if item.l10n_key == 'HYDROGEN' then
     item = Equipment.cargo.hydrogen
   end
   if ship:CountEquip(item) > 0 then
     return true
   else
     return false
   end
end
```
The reason, Laarmen maybe best explained ;)

@sfence You can send the savegames in zip format?

sfence:
@walterar Now you should be able to download a zip archive of all saves from the same address.

clausimu:
@walterar - really?? I must be totally on the wrong track then... As far as I understand I am already passing "Equipment.cargo.hydrogen" to the cargoPresent function as "item". If I weren't so, how else would "item.l10n_key" return 'HYDROGEN'? As it is, the (new) first three lines in that function would seem redundant to me...? 

[...]

PR #3659, opened 2016-03-25
SAR: fix fuel transfer

clausimu:
This is in response to one of the issues raised in #3627 - that fuel transfer is not working for some users. I went with a solution based on what @walterar suggested (re-loading the actual cargo equipment object), but packed it a little differently. That way you only have to change one place in the script if you ever add more mission flavors and more types of cargo.

I still don't know why this workaround is necessary, or why it works for some users (such as me...) and doesn't for others. Due to a lack of common language :) @walterar can't explain it to me. If someone else understands, please share. But I tested it with the save game supplied by @sfence and it works.

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -6,7 +6,8 @@
 -- - All station/planet location references in ad/mission are stored as paths for consistency because
 --   some can't be accessed as bodies until the player enters the respective system and getting a path from
 --   systembody needs to go through body. Variable "location" is a SystemsBody.
-
+-- - Any cargo commodity that needs to be delivered/picked up needs to be intered into "verifyCommodity" function.
+--   Some users have issues with commodities not beeing recognized otherwise. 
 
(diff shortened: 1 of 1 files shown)


impaktor:
This is @laarmen's domain.

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -292,6 +293,15 @@ end
 -- basic mission functions
 -- =======================
 
+local verifyCommodity = function (item)
```
laarmen: I don't like it, as it fixes the symptom instead of the underlying problem (which is that it gets lost in the first place)

walterar:
@clausimu Welcome to the madhouse, get comfy :)

fluffyfreak:
So why does this happen in the first place? has `item` been corrupted after save/load?

fluffyfreak:
I'm closing this one as @clausimu seems to have been busy/away and I think we should just accept this fix as a bandage until we know the cause of the issue.

Issue #2043, opened 2013-02-07
Game Freeze/Crash - Nav. Map


radius75:
![bez tytu u11](https://f.cloud.github.com/assets/1929111/136109/99ca3892-713f-11e2-9a10-a390fce9ff2b.PNG)
winXP32
write letters łńśźż... and press [Backspace] key.

it was tested on:
-master, alpha31freeze
-alpha30-win32

PR #3662, opened 2016-03-26
Make GuiTextEntry handle UTF-8 multibyte chars for backspace and delete

### LINKED ISSUE #3655: Backspace on non-ASCII in star map causes crash
1. Set pioneer language to, say russian.
2. Type some weird ass ruski letter in search bar of F2 star map / sector map
3. Realize your horrible mistake, and try to erase it, using backspace --> crash.

...from assert on  src/text/TextureFont.cpp L108, 20160316 build. So should be current master.

__EDIT__ Also realize that spelling is hard.

johnbartholomew:
This should fix #3655 and #2043 (which I think are duplicates?).

However, it's 100% untested (faith-based programming ftw?). So... it would be good if someone could test this before it's merged. If I can remember how to input an accented character on my computer I'll test it myself... :-/

### DIFF
--- a/src/gui/GuiTextEntry.cpp
+++ b/src/gui/GuiTextEntry.cpp
@@ -73,13 +73,14 @@ bool TextEntry::OnKeyDown(const SDL_Keysym *sym)
 		accepted = true;
 	}
 
-	// XXX deleting characters is not UTF-8 safe
 	if (sym->sym == SDLK_BACKSPACE) {
 		if (m_cursPos > 0) {
 			if (m_text[m_cursPos-1] == '\n')
 				--m_newlineCount;
-			m_text = m_text.substr(0, m_cursPos-1) + m_text.substr(m_cursPos);
-			SetCursorPos(m_cursPos-1);
+			const char *cstr = m_text.c_str();
+			const int len = Text::utf8_prev_char_offset(cstr + m_cursPos, cstr);
+			m_text = m_text.substr(0, m_cursPos-len) + m_text.substr(m_cursPos);
+			SetCursorPos(m_cursPos-len);
 			changed = true;
 		}
 		accepted = true;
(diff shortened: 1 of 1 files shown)


impaktor:
Sweet. I'll give this a try next time I'm at the office, which might be today or tomorrow. I trust umlauts will also trigger the bug (I'll test that) in which case I have the perfect keyboard, with å,ä,ö.

impaktor:
PIoneer doesn't let me type any "foreign" letter in that search box, so I can't even trigger the original bug. Maybe @radius75 can test if this PR fixes the original issue, or alternativley we just merge it.

impaktor:
Let's hope it works.

radius75:
tested current master 646f6c0ad7e2a98ac145009567336504b130d180 , and works fine :)

btw 
```right alt + o``` = ó  and ```left alt + left ctrl + o``` = ó ( Polish char.) not inputing in nav. map now.
a shortcut is probably blocking it
but it has no relation to this fix, and it is not necessary at this time

laarmen:
Direct non-ascii input (é, à, ç are directly reachable on French AZERTY,
without even Shift) won't work either.

impaktor:
But why would you want to input characters that don't exist in the game?

_if_ that is done, it shouldn't crash when removing them, thus this fix, but incorrect chars are... incorrect.

radius75:
maybe sometime in the future you can give the name of the planet or system...
using its own set of characters

impaktor:
> maybe sometime in the future you can give the name of the planet or system...
using its own set of characters

Yes, but how would someone find that system? I don't have mandarin, or cyrilic or greek letters on my keyboard.

radius75:
I used the "mod" of the Polish translation of the planets in the Sol.
I could then search for the "Układ Słneczny " = " Solar System"  in nav. Map.
I do not need other characters than Polish :-)

PR #3663, opened 2016-03-30
Don't compress normal maps

fluffyfreak:
This PR avoid compressing normal maps that are not already in DDS format.
DXTn is pretty bad at handling normal maps and so for now we will simply sidestep the issue and use them without it.

Add TextureBuilder::Normal which will not compress a texture if it isn't already compressed.
Use it for all normal maps, existing DDS normal maps will still be compressed.

ping @nozmajner

### DIFF
--- a/src/graphics/TextureBuilder.h
+++ b/src/graphics/TextureBuilder.h
@@ -24,6 +24,9 @@ public:
 	static TextureBuilder Model(const std::string &filename) {
 		return TextureBuilder(filename, LINEAR_REPEAT, true, false, false, true, true);
 	}
+	static TextureBuilder Normal(const std::string &filename) {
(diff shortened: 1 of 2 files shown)


nozmajner:
Thanks ! :)
I wanted to ask about this, but you beat me to it.
@BobTheTerrible might be interested in this too. 

I guess I should begin generating normals for all existing ships then?

fluffyfreak:
I found an [interesting article](http://www.ozone3d.net/tutorials/bump_map_compression.php) about using the swizzled and Ati 3Dc techniques with OpenGL so hopefully in future we will be able to use compression again. When that happens though I will try to make the changeover as seamless as possible :)

PR #3667, opened 2016-03-31
NavLight Billboard Gathering

fluffyfreak:
Attempt to gather all of the NavLight billboards for dispatch with a single draw call, instead of a draw call per-navlight.

For the largest stations especially this could mean removing 800+ draw calls which especially on slower machines could have some impact.

There is an overhead to doing this, it's essentially shifting some processing from the GPU driver onto Pioneer itself. However the work shifted to Pioneer is just some matrix transformations, whereas the draw calls could be much more expensive.

Another possible negative is that currently I don't think that the NavLights look as good as they used too. This is partially due to the solid colourisation whcih means that they don't brighten to white in the centre (_currently_), and also because I am using solid colours which might look better once an artist has edited the `NavLights.ini` file.

- [x] Test on Linux.
- [ ] Test on OSX.


impaktor:
Curious:
```
[~/usr/src/pioneer/src][130]% grep -iR colour * | wc -l                                                [git][master]
370
[~/usr/src/pioneer/src]% grep -iR color * | wc -l                                                     [git][master]
3755
```

I can give it a spin on GNU/Linux

fluffyfreak:
Thank, I hope to test it to see if there's any performance difference too by disabling vsync in the config then running it with something like `export vblank_mode=0 pioneer`.
Have to run the current and this version separately of course but it'd be good to see if there is any change.

If the texturing changes prove to be a show stopper then there are further things that I can try and have been thinking about.

fluffyfreak:
Also why is disabling VSync in Linux so difficult? http://stackoverflow.com/questions/17196117/disable-vertical-sync-for-glxgears

impaktor:
```
Loading took: 32417.275885 milliseconds
Could not find model: mx1
pioneer: VertexBufferGL.cpp:62: Graphics::OGL::VertexBuffer::VertexBuffer(const Graphics::VertexBufferDesc&): Assertion `m_desc.numVertices > 0' failed.
[1]    12444 abort      ./src/pioneer
```

fluffyfreak:
Ah my mistake, should have moved https://github.com/pioneerspacesim/pioneer/pull/3667/files#diff-121a7a511a68ec73a60927e7bb05a875R243 inside the if check :/ I'll fix it now.

What's that `Could not find model: mx1` all about?

fluffyfreak:
@impaktor there, that should run, tested it in debug so I actually got the asserts too!

impaktor:
mx1 was apparently something I had lying around. Said "drop ship". Removed it.

Tested new update. Works. Both London and Torvalds.

fluffyfreak:
Ok, definite framerate improvement on my low-end linux box :)

The question now is do I go a bit further with it? There's a couple of other way of doing this that might perform even better.

laarmen:
I say yes (while I go fetch my netbook).

impaktor:
Reach for the stars?

fluffyfreak:
Ok then, now trying to get custom point sprite rendering working instead.
Been meaning to do this for ages!

fluffyfreak:
**Booyah!!!** Point sprites using a texture.

fluffyfreak:
@nozmajner I switched back to using a texture `halo_4x4.png`, the `NavLights.ini` just defines the UV offsets within it.

nozmajner:
Looking at that file, it uses relative coordinates, right? 0.5 for the half point.

fluffyfreak:
Yep, they seem to render with some slight... blockiness, not sure, I might need to revisit that shader at some point but it's good for now I think.

You could use a bigger texture if you like, the originals were only 128x128, the 4x4 version is only 256x256. Going to 512x512 would be fine.

Off too a party now :)

nozmajner:
Have fun! :)
I did some experiments with hand painted halos a while ago, maybe I revisit them both for this and the engine exhausts. My images for those were bigger anyway.

fluffyfreak:
@salborrelli & @eloquentmess I haven't yet been able to test this on OSX which is the platform that always gives me the most gfx issues. Any chance that one/both of you could try it out?

fluffyfreak:
@salborrelli there's a couple of new files in this one within `graphics/opengl/` called `BillboardMaterial.h/cpp` that will need adding to the `graphics` project/folder in XCode.

PR #3669, opened 2016-04-03
LuaShip accessors for a ships position and velocity.

fluffyfreak:
For @nozmajner to play with :)

@laarmen, @impaktor, @johnbartholomew I'm not sure if this is the right way to do this, should these just be properties of each ship instead?

### DIFF
--- a/src/LuaShip.cpp
+++ b/src/LuaShip.cpp
@@ -4,7 +4,6 @@
 #include "LuaObject.h"
 #include "LuaUtils.h"
 #include "LuaConstants.h"
-#include "LuaVector.h"
 #include "EnumStrings.h"
 #include "Ship.h"
 #include "Missile.h"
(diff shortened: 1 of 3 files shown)


impaktor:
Is it velocity relative to the current selected frame?

fluffyfreak:
Whatever a Ship::GetVelocity is :)
So yes, relative to current frame.

walterar:
And... if we make a SetVelocity too? :)

Maybe we can do this?
```
local min = 10-- km
local max = 20--km
local ship = Space.SpawnNear("viper_police_craft", Game.player, min, max)
local vel = Game.player:GetVelocity()
local plus = 450--mts
vel = vel + plus
ship:SetVelocity(vel)
ship:AIKill(Game.player)
```

laarmen:
I'd rather have properties, indeed. It is pretty easy to imagine UI elements responding to velocity and/or position, which, IIRC, is what properties were initally all about. Plus, it makes the Lua code easier on the eyes.

fluffyfreak:
@walterar I think for that we'd be better of adding the velocity to the `SpawnNear` command, or indeed modifying the command to spawn ship with the targets velocity. It seem like a bug that it doesn't do that already in fact.

@laarmen I'll swap it to be a property instead them.

fluffyfreak:
Swapped to be `Ship` properties

laarmen:
OK, having had a meeting for the last 2 and a half hour, I guess I'll have
done little enough programming for me to review and merge this tonight :)

fluffyfreak:
Yay for meetings?!? ;)

laarmen:
Can anybody merge this? I'm fine with it, but github won't let me push...

Issue #3654, opened 2016-03-21
I can see through Bowfin and space stations


Tichy:
That issue occours only with Bowfin and space stations. Every other ship looks good.

My opengl.txt
https://gist.github.com/Tichy/6ecbe05b46511798619f

My output.txt
https://gist.github.com/Tichy/b755b6465a63783ac136

![2016-03-21-184010_1280x800_scrot](https://cloud.githubusercontent.com/assets/2348771/13928229/278d36ee-ef95-11e5-854c-36cb1362148b.png)

![screenshot-20160321-184645](https://cloud.githubusercontent.com/assets/2348771/13928446/ea6823fe-ef95-11e5-82c0-a42671fefeec.png)

impaktor:
Conversation with @robn Friday March 11 2016:
```
<robn> I'm guessing the big ring station interior isn't supposed to be
       seethrough?  [12:12]
<robn> http://i.imgur.com/aOk0SWa.png
<impaktor> Wow, how did that happen?  [12:13]
<robn> stock build, this very second
<impaktor> Arrrghhh!
<robn> OpenGL version 3.3 (Core Profile) Mesa 11.1.2, running on Intel Open
       Source Technology Center Mesa DRI Intel(R) Haswell Mobile
<robn> I'll try the nvidia
<fluffyfreak> back 5 minutes and already breaking stuff ;)
<robn> devops mate  [12:14]
<robn> that's how we roll
<impaktor> master works fine on my system.  [12:15]
<robn> http://i.imgur.com/yWNMHp2.png
<robn> nvidia is the same, but with prettier lights
<robn> OpenGL version 3.1.0 NVIDIA 352.79, running on NVIDIA Corporation
       GeForce GTX 770M/PCIe/SSE2
* robn squints at GL3
<robn> that's supposed to be GL 4.lots on the nvidia
<fluffyfreak> yeah but nvidia gives you what you ask for  [12:16]
<robn> I dunno what garbage debian is shipping these days though
<fluffyfreak> we request 3.1 interface
<robn> ohh
<robn> ok
<robn> just different to last time I was here
<robn> all good then
<robn> so, you got _bugs_
<robn> but works fine for impaktor, hmm  [12:17]
<fluffyfreak> *features* mean, now you can see where you're gonna land ;)
<robn> that could be anything though, our systems are likely very different
<robn> totes
<robn> the whole hub was lit up for miles
<robn> I wonder what it looks like from the inside
[...]

PR #3671, opened 2016-04-03
Update to whole new Assimp

fluffyfreak:
This updates us to use the new Assimp library in pioneer-thirdparty.

Then I added dumping the warnings found when the modelcompiler is used, then I fixed up the bowfin and orbital station `.model` material names.
There are [many warnings](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=27&start=50#p4692)

**EDIT:** Also it's a fix for #3654

### DIFF
--- a/data/models/misc/cargo/cargo.model
+++ b/data/models/misc/cargo/cargo.model
@@ -1,4 +1,4 @@
-material cargo-material
+material cargo
 tex_diff cargopod.dds
 tex_glow glow.dds
 tex_spec specular.dds
--- a/data/models/misc/default_cockpit/default_cockpit.model
+++ b/data/models/misc/default_cockpit/default_cockpit.model
@@ -1,4 +1,4 @@
-material cockpit-material
+material cockpit
 diffuse 1.0 1.0 1.0
 specular 0.3 0.3 0.3
 shininess 23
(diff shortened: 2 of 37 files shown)


nozmajner:
Do I have to change anything about the models, or in the ways I export them?

fluffyfreak:
@nozmajner I'll get back to you on that, but so far seems to just be the name used in the `.model` file.

fluffyfreak:
@nozmajner when/if this gets merged the `modelcompiler` will output warnings so a good way to find errors and problems with *all* the models is `modelcompiler -b` and then look in the `output.txt` to see what it's said.

Most of the stuff is benign and the engine just silently handles it but you can then check them in the viewer and see whihc ones are real problems or not.

fluffyfreak:
@robn this is the companion code/data update to the pioneer-thirdparty stuff https://github.com/pioneerspacesim/pioneer-thirdparty/pull/19

fluffyfreak:
@robn this has data fixes which might be necessary for an Assimp 3.2 release.

robn:
I'm ready to do a test build against assimp 3.2. I will merge this and @pioneer-thirdparty#19 for the Linux builds in the next hour or two unless there's any objections.

fluffyfreak:
That'd be great, thanks @robn :)

PR #3673, opened 2016-04-04
Equipment: check for commodity copies when saving

laarmen:
This patch should make the game throw an assertion error any time someone tries to serialize a copy of a commodity instead of the actual commodity. It is a hack and should probably be opt-in, but I don't think we have a way (yet) to enable debug feature in Lua.

Poke @impaktor 

Also, I apologize for the branch on the main repo, but the Github UI isn't very practical in that regard. (Well, they DO provide you with an editor, so...)

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -10,6 +10,11 @@ local Timer = import("Timer")
 local Space = import_core("Space")
 local Comms = import("Comms")
 
+local cargo
+local laser
+local hyperspace
+local misc
+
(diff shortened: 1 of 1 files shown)


impaktor:
I wonder if that assert will trigger on any of the existing code in master?

Not to mention scout+?

laarmen:
That's sort of the point, actually. Which is why I'd like to make it opt-in

laarmen:
Replaced by #3675

PR #3675, opened 2016-04-05
Lua debug tools, part one

laarmen:
Hi,

This PR is a bit of a warmup for me. It adds a 'debug mode' to the Lua context to allow us to sprinkle consistency checks and asserts throughout the code without having the average user crash because a modder has had a creative way to use our APIs. Also, it will make available to the user/dev/whoever types in the Lua console/my dog all the cool shit I'm going to try to develop next.

PS: I also added a check when serializing Equipment, because that's how I roll.
PPS: I now realize it probably won't catch the bugs I was thinking about because metatables and whatnot, but hey, c'est la vie!

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -10,6 +10,11 @@ local Timer = import("Timer")
 local Space = import_core("Space")
 local Comms = import("Comms")
 
+local cargo
+local laser
+local hyperspace
+local misc
+
 --
 -- Class: EquipType
 --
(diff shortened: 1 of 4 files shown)


walterar:
"...user crash because a modder has had a creative way to use our APIs."

What do you mean with that? I see crashes that result from the decision of a c++ dev to not use creative ways to build the APIs.

nozmajner:
I think he means that it's impossible to predict every way of usage. It's a bit like I'm using Blender for like 12 years but my students still surprise me with errors I never experienced.

fluffyfreak:
Are we able to attach an external debugger to our Lua process?

laarmen:
A few years back I had an embryo of a debugger in an external repo and I
succesfully used it when working on #1719, however I seem to remember it
failing to compile more recently. (at the time the alternatives either
weren't good enough or needed a full fledged Lua IDE, things might have
changed since)

I'll look into it as part of this series, I guess.

laarmen:
This one looks nice, I'll try it out in the week :

https://github.com/slembcke/debugger.lua

fluffyfreak:
The C++ code here looks fine to me, I'm not very good with the Lua side though so will have to take your word on those parts :)

laarmen:
That's why we have @impaktor ;-)

PR #3679, opened 2016-04-07
Change to GetPosition and GetVelocity LuaShip functions.

### LINKED ISSUE #3678: Segfault on takeoff (autosave?)
[stacktrace.txt](https://github.com/pioneerspacesim/pioneer/files/207341/stacktrace.txt)

I can make the full core dump available if need be, but it's a bit large, so only if somebody says they actually want it.

I'm pretty sure it has to do with saving the game since it stops happening if I turn autosave off.

fluffyfreak:
Revert position and velocity as Properties, expose as GetPosition and GetVelocity LuaShip functions.

This is how I'd prefer to fix #3678 @laarmen

### DIFF
--- a/src/LuaShip.cpp
+++ b/src/LuaShip.cpp
@@ -4,6 +4,7 @@
 #include "LuaObject.h"
 #include "LuaUtils.h"
 #include "LuaConstants.h"
+#include "LuaVector.h"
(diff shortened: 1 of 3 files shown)


laarmen:
Go for it!

PR #3687, opened 2016-04-21
Pre-convert to DDS more textures.

fluffyfreak:
I notice recently that whilst everything looks fine on Windows the LInux drivers are still doing a terrible job of generating DXTn textures. 

This was making landing lights, thrusters etc look worse than they are, and they're not awesome anyway.

This PR just pre-converts them to DDS DXT5 (_8-bit alpha_) with mipmaps format so that we don't have to worry about the quality of the generator on the target machine.


impaktor:
The change in png's is just for better resolution?

fluffyfreak:
Yeah I enlarged the PNG files, some of them anyway, then converted to DXT5+mipmaps as DDS.

walterar:
- projectile_w.DDS
+projectile_w.dds

Issue #3478, opened 2015-10-04
gui warning


alex199568:
When I start the game on Ubuntu 14.04 I have GL_INVALID_ENUM. The game still loads and I can start a new game. I tried to install libtxc-dxtn, but this didn't help. But the gui issue is that the warning about GL_INVALID_ENUM not only prints to console, but also via messagebox every frame. Thus, to start a game I have to close hundreds of them.

fluffyfreak:
Hi @alex199568 could you upload the contents of the `opengl.txt` & `output.txt` files we create when the game is run?

That should give me the information I need about your GPU etc.

impaktor:
```
<magagnarata> hi, I'm getting problems when running pioneer on ubuntu 14.04                                                                  
              x86_64, like a lot of 'GL_INVALID_ENUM' gui messages during                                                                    
              universe simulation startup.
```
And:
```
OpenGL version 3.1 (Core Profile) Mesa 10.1.3, running on Intel Open Source Technology Center Mesa DRI Intel(R) Sandybridge Mobile 
Available extensions:
Shading language version: 1.40
  GL_ARB_ES2_compatibility
  GL_ARB_ES3_compatibility
  GL_ARB_base_instance
  GL_ARB_blend_func_extended
  GL_ARB_clear_buffer_object
  GL_ARB_copy_buffer
  GL_ARB_debug_output
  GL_ARB_depth_buffer_float
  GL_ARB_depth_clamp
  GL_ARB_draw_buffers
  GL_ARB_draw_buffers_blend
  GL_ARB_draw_elements_base_vertex
  GL_ARB_draw_instanced
  GL_ARB_explicit_attrib_location
  GL_ARB_fragment_coord_conventions
  GL_ARB_fragment_shader
  GL_ARB_framebuffer_object
  GL_ARB_framebuffer_sRGB
  GL_ARB_get_program_binary
  GL_ARB_half_float_pixel
  GL_ARB_half_float_vertex
  GL_ARB_instanced_arrays
  GL_ARB_internalformat_query
  GL_ARB_invalidate_subdata
  GL_ARB_map_buffer_alignment
  GL_ARB_map_buffer_range
  GL_ARB_occlusion_query2
  GL_ARB_pixel_buffer_object
  GL_ARB_point_sprite
  GL_ARB_provoking_vertex
  GL_ARB_robustness
  GL_ARB_sample_shading
  GL_ARB_sampler_objects
[...]

PR #3691, opened 2016-04-29
Texture code tidy

fluffyfreak:
# Description:
Tidying up some of the code around texture creation and usage.
Tracking down the cause of https://github.com/pioneerspacesim/pioneer/issues/3478#issuecomment-155578063 so I can eventually fix it.

# Art:
@nozmajner I found something you might find useful/interesting whilst doing this, there appears to be an undocumented piece of patterin functionality:
>	//load as a pattern, allowed:
>	//patternNN.png
>	//patternNN_n.png
>	//patternNN_n_s.png
>	//s = smooth (default for pattern)
>	//n = nearest (default for color)
>	if (name.length() >= 11 && name.compare(10,1, "n") == 0) smoothPattern = false;
>	if (name.length() >= 13 && name.compare(12,1, "s") == 0) smoothColor = true;

In other words you can control the pattern textures smoothing AND the diffuse textures smoothing using the pattern files name.
This might be useful for eliminating the red/green/blue banding that we see on some texture/pattern combinations like the camouflage pattern on the Wave.

Andy


impaktor:
> there appears to be an undocumented piece of ...

Which reminds me that editing the wiki doesn't work since the server-upgrade. Ping @robn (clicking edit-tab, just gives a blank/loading page)

fluffyfreak:
I still isn't enough to fix the Intel driver issues, I think I will actually need to build the mipmaps manually and then submit them to be compressed. There is a good example of this in the SOIL2/stb image helper libraries which all have a very liberal licensing.

Issue #3688, opened 2016-04-27
Crash on jettisoning cargo


icarito:
error: [string "[T] @modules/CrimeTracking/CrimeTracking.lua"]:74: couldn't find constant with name 'SEMI_FINISHED_PRODUCTS' in namespace 'CommodityType

stack traceback:
	[C]: in function 'IsCommodityLegal'
	[string "[T] @modules/CrimeTracking/CrimeTracking.lua"]:74: in function 'cb'
	[string "[T] @libs/Event.lua"]:34: in function '?'
	[string "[T] @libs/Event.lua"]:185: in function <[string "[T] @libs/Event.lua"]:180>

laarmen:
Hi,

Thanks for the report!
@impaktor, that's one of yours isn't it?

impaktor:
@laarmen Depends. Seems like the error is that checking the legal state on custom cargo is not well defined? Or what is the intended behaviour of `LuaStarSystem.cpp:l_starsystem_is_commodity_legal` if the cargo is custom?

laarmen:
This function was written before my time. The whole legal commodity thing has been a pain in my rear, but I didn't dare touch it at the time of #1719 because people were talking about refactoring sysgen and all that.

laarmen:
(so basically its behaviour is undefined for custom cargo)

impaktor:
I guess the logical thing would be to default to "not illegal / legal" for custom cargo?

laarmen:
That'd be a patch, yes, but the more fundamental problem remains: we need to make this moddable!

PR #3692, opened 2016-04-29
Legal custom cargo

laarmen:
This should take care of #3688, at least for the crash part. We still need to export the legality part to Lua somehow.


fluffyfreak:
This looks good to me, I realise that it's a bandage and not what you might want eventually but I think it's good for now.

Issue #3681, opened 2016-04-11
Missiles produce a lot of smoke


walterar:
@fluffyfreak  Help! :) 
![smokemissile1](https://cloud.githubusercontent.com/assets/1334989/14443891/7ce694d0-0018-11e6-96fa-97eeb4327238.png)
![smokemissile2](https://cloud.githubusercontent.com/assets/1334989/14443893/83fd8422-0018-11e6-9050-fd8cba39a74f.png)
![smokemissile3](https://cloud.githubusercontent.com/assets/1334989/14443895/89e74e2c-0018-11e6-97a3-3352424fd39c.png)

Cof! cof!
![smokemissile4](https://cloud.githubusercontent.com/assets/1334989/14443929/b6f61d6c-0018-11e6-9b19-65736b9e1913.png)

laarmen:
Yeah, I guess in vacuum smoke should expand and disappear pretty quickly.

walterar:
Everything points to #3670 (and #3667?)

walterar:
With #3667 works well. The great smoker is #3670

fluffyfreak:
I've been away, gimme a couple of days to catchup.

laarmen:
Here, have 3!

fluffyfreak:
Thanks! :P

This is probably caused by the changes to object size calculation. In the old system it decided how big each quad should be, which was then built and rendered. With billboards however you have to tell it the final size in pixels that it will occupy onscreen.

I probably just didn't get that maths right though I did test firing missiles at stations and made sure it was comparable. Oh well.

walterar:
If the smoke cloud of missile seems too big, wait till you see the explosions of large ships. The size has very much increased.

fluffyfreak:
I think I have an idea of how to tackle this, my brains a bit too tired to get it correct just now but I have it partially working.

I'll try a bit more tomorrow.

walterar:
@fluffyfreak Some precisions of what I have observed:

The dimensions of the explosion are well on normal ships. I like it better now. Only in large statics it seems excessive.

The smoke effect on missile becomes a much more interesting effect with a simple change in these values, try it.

```
 	switch (m_type) {
 		case TYPE_EXPLOSION:
			if (m_age > 3.2) m_type = TYPE_NONE;
 			break;
 		case TYPE_DAMAGE:
			if (m_age > 2.0) m_type = TYPE_NONE;
 			break;
 		case TYPE_SMOKE:
[...]

PR #3695, opened 2016-05-11
Rework SFX rendering to fix giant sprites.

fluffyfreak:
This was just an attempt to fix the SFX rendering of smoke and explosions (#3681)
In that regard it has some improvements but still needs more work.

Whlist doing that though I realised just how awful the whole SFX rendering is so have rewritten it a bit.


src/Sfx.cpp
```diff
@@ -107,152 +88,200 @@ void Sfx::TimeStepUpdate(const float timeStep)
...
+void SfxManager::Cleanup()
+{
+	for(size_t t=TYPE_EXPLOSION; t<TYPE_NONE; t++) 
+	{
+		const size_t numInstances = GetNumberInstances(SFX_TYPE(t));
+		if(!numInstances)
+			continue;
+
+		for (Sint64 i = Sint64(numInstances-1); i >= 0 ; i--)
+		{
```
impaktor: space semicolon?

(Sorry, still in proofreading mindset after thesis handin)

fluffyfreak: That's ok :)

walterar:
@fluffyfreak  How to build the matrix explosions.png?
Manually? or with some utilitary program, from image0...31.png?

fluffyfreak:
@walterar you can use something like [TexturePacker](https://www.codeandweb.com/texturepacker) or you can do it manually. 
The first frame of the animation is in the top left, the 2nd is too it's right, when you reach the end of that row of 6 images you drop down onto the next row of 6 images.
Currently it's hardcoded too a grid of 6x6, but only the first 32 are used for the explosions.

walterar:
@fluffyfreak  TexturePacker worked well for me, thanks.
I'm using 36 images, right 6x6, with some little changes in sfx.cpp. In static ships it seems to work very well. I need to test with small ships in a "real" battle with 4 or 5 pirates.

impaktor:
Changes to png's worries me due to the growing git repository size, so I opt for causion when changes must be made to png. So is the scout+ version objectivly better, or does it not fit the pioneer style? If the former, maybe we should use that version instead?

Just a though.

fluffyfreak:
@impaktor the current 32 image texture atlas is fine, I won't be changing it unless @nozmajner wants to make new explosion animations in which case I'll make some larger code changes to drive it from configuration files. Right now it isn't necessary as it would just add 4 frames to the animation.

impaktor:
But I assume @walterar uses other animations in Scout+, and now wants to know how to update to the new format you introduce here? (Maybe I should just look at scout+)

walterar:
@fluffyfreak I like how it works; smoke needs a little more love. ;)
Have you thought about the posibility to have (a random or directed) explosions_00..99.png?
explosions.png?

fluffyfreak:
@walterar I'm sure then that #3697 will be of interest to you :)

PR #3697, opened 2016-05-13
Sfx ini configuration file

fluffyfreak:
Makes the SFX Explosion, Damage and Smoke effect be driven from a `Sfx.ini` file.

The 3 parameters are:
 - `type` which can be `billboard` or `atlas`
 - `stride` which is the number of times the image is repeat horizontally
 - `count` how many images are in the animation

`count` should always be less than `stride * stride`.

Included in this commit is the file `Smoke45Frames.dds` which is under the CC0 licence from [OpenGameArt](http://opengameart.org/content/smoke-aura) which I have converted to DDS for use in Pioneer.

I have also removed a number of `.png` files that were no longer used and there is a PR in the `pioneer-assets` repository for those original files.


data/textures/Sfx.ini
```diff
@@ -0,0 +1,6 @@
+damageFile=textures/smoke.png
+smokeFile=textures/Smoke45Frames.dds
+explosionFile=textures/explosions/explosions.dds
+damagePacking=billboard,1,1
+smokePacking=atlas,7,45
+explosionPacking=atlas,6,32
```
impaktor: I'm guessing this file doesn't end in a new line? Maybe it should?
(See, still proofreading!)

fluffyfreak: sorted!

fluffyfreak:
Ping @nozmajner to take a look at the new smoke anim and explosion layouts etc.
See if there's anything you dislike.

PR #3505, opened 2015-10-29
Introduce Tech Level to stations - makes them more unique

impaktor:
# Tech Levels

Here I go again with one of my "limit game content exposure to player", but I'm very happy with this one as it really improves the game I think, making each station a bit more unique. This implements what I discussed in the top post on the dev-forum [Tech Levels for ship equipment](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=235), by giving each station a tech level, with some distribution (the [convolution](https://en.wikipedia.org/wiki/Convolution) of two square waves), which then influences what equipment is supported (bought/sold) on in the equipment market. (Note: this does not affect the Second Hand market module on the BBS, which I think is as it should be).

Tech levels lead to more interesting game play I think (pretty much exactly like in Frontier):
- "Oh, this was a good station!"
- "Wow, there's an _Advanced_ ECM as well? Never seen that before!"
And of course:
- "Blast! This station doesn't have <gadget> I desperately need, maybe one city over has it?"

Caveat: last commit adds a "Military" tech level, just like in frontier. And just like in frontier it's "Tech Level 11" that is "special" in being "military", thus no station will show "tech level 11 in the Lobby", since they will instead show "military installation" or what ever (see screen shot).

screenshot-20151027-150622
(I think the text doesn't fit on my 640x800 resolution, but you get the idea.)

## Table of levels

I used the system from Frontier as base mainly because it was (in my opinion) a well constructed system, and also it's what many/most of the pioneer players are used to, and how they expect it to work. So at tech level 12 all equipment is available except those with level "MILITARY" ("MIL" in table below), which are only available at tech level 11 stations. A bit quirky, that Military drive class 2 is only available at these military installations. Just like Frontier. I liked it, so I kept it like that.

```
| Equiptype             | Description                                | pioneer | frontier |
|                       |                                            |    tech |     tech |
|-----------------------+--------------------------------------------+---------+----------|
| MISSILE_UNGUIDED      | unguided rocket (MISSILE)                  |       1 |        - |
| MISSILE_GUIDED        | guided missile (MISSILE)                   |       5 |        5 |
| MISSILE_SMART         | smart missile (MISSILE)                    |      10 |       10 |
| MISSILE_NAVAL         | naval missile (MISSILE)                    |     MIL | 11 / MIL |
| ATMOSPHERIC_SHIELDING | atmospheric shielding (ATMOSHIELD)         |       3 |        3 |
[...]

PR #3701, opened 2016-05-17
Make Autopilot purchasable from everywhere.

fluffyfreak:
The most useful piece of equipment in the game, essential for all newbie players, and for some rason we make it a pain to find?

That's dumb in fact I don't think it should even be equipment, just something that any ship can always do but that's debateable.

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -720,7 +720,7 @@ misc.cargo_life_support = EquipType.New({
 })
 misc.autopilot = EquipType.New({
 	l10n_key="AUTOPILOT", slots="autopilot", price=1400,
-	capabilities={mass=1, set_speed=1, autopilot=1}, purchasable=true, tech_level=6
+	capabilities={mass=1, set_speed=1, autopilot=1}, purchasable=true, tech_level=1
 })
 misc.radar_mapper = EquipType.New({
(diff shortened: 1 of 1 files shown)


impaktor:
:+1:

impaktor:
> The most useful piece of equipment in the game, essential for all newbie players, and for some rason we make it a pain to find?

From table in #3505 I read that it was like that in Frontier. That was the reason.

fluffyfreak:
Yeah I didn't think that it was done with deliberate thought :)

walterar:
Piloting with human hands became obsolete in the early 21st century more than a thousand years ago. Although for some it is a hobby today. I know because I was there.

Brianetta:
Years ago I suggested a very hard game start, where the player started in a ship with no shields, no atmospheric shielding, no jump drive and no autopilot, in an orbital station. A different station sold atmospheric shielding, and an autopilot and jump drive were available from the planet's surface.

This gives the additional challenge of having to actually pilot your space ship before you're allowed to sit back and let it fly itself, including orbital rendezvous and planetary landing.

As long as it's possible to specify, in a custom system, the availability of specific items, this remains possible.

impaktor:
> I don't think it should even be equipment, just something that any ship can always do

I might be late to the party, but some players want autopilot to be _removed_ (based on conversations on IRC), as they find manual flying the true joy of pioneer.

PR #3705, opened 2016-05-20
Cancel a missile order when the ship has jumped

laarmen:
Or crashed, for that matter. Anyway, AIKamikaze will not try and fail to go after invalid bodies anymore.

Ping @fluffyfreak , because i'm guessing the methods I've introduced might be useful elsewhere but I don't know much about this area of the code.

### DIFF
--- a/src/Missile.cpp
+++ b/src/Missile.cpp
@@ -135,6 +135,7 @@ void Missile::NotifyRemoved(const Body* const removedBody)
 	if (m_owner == removedBody) {
 		m_owner = 0;
 	}
+	Ship::NotifyRemoved(removedBody);
 }
 
 void Missile::Arm()


laarmen:
Also, ping @walterar.

walterar:
Testing.

walterar:
This works. thanks @laarmen :+1:

laarmen:
DO NOT MERGE!

While it fixes the problem, it shouldn't have to in the first place. For some reason onDelete doesn't do its job, investigating now.

laarmen:
Oh wow, turns out it was broken ever since #1663 !

laarmen:
OK, a better fix for that.

walterar:
tested, OK :+1:  :)

laarmen:
Cool, thanks!

PR #3709, opened 2016-05-23
Clean up the police tables on ship destruction

### LINKED ISSUE #3512: Game crash because of destroyed police ship
If a police ship gets destroyed, the game stops with the message
"error: Lua serializer '(null)' tried to serialize an invalid 'Ship' object".

You can reproduce this with the following steps:
1. Take off
2. Fire your gun or do other evil things
3. Try to survive until the police ship is destroyed
4. Dock with station

laarmen:
It was not really possible to make the whole module stateless.
Furthermore, I kind of like the idea of doing that in our own modules,
since it shows potential modders (and our future selves) that THOU SHALT
SUBSCRIBE TO onShipDestroyed* FOR THINE SHIP REFERENCES. Nanmého.

(*) OK, and maybe the one for the ships leaving the system, depending on
the context

Fixes #3512

### DIFF
--- a/data/libs/SpaceStation.lua
+++ b/data/libs/SpaceStation.lua
@@ -697,6 +697,17 @@ Event.Register("onLeaveSystem", function (ship)
 	if ship ~= Game.player then return end
 	destroySystem()
 end)
+
+Event.Register("onShipDestroyed", function (ship, _)
+	for _,local_police in pairs(police) do
+		for k,police_ship in pairs(local_police) do
+			if (v == police_ship) then
+				table.remove(local_police, k)
+			end
+		end
+	end
+end)
+
 Event.Register("onGameEnd", function ()
 	destroySystem()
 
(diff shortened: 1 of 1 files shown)


data/libs/SpaceStation.lua
```diff
@@ -697,6 +697,17 @@ Event.Register("onLeaveSystem", function (ship)
 	if ship ~= Game.player then return end
 	destroySystem()
 end)
+
+Event.Register("onShipDestroyed", function (ship, _)
+	for _,local_police in pairs(police) do
+		for k,police_ship in pairs(local_police) do
+			if (v == police_ship) then
```
impaktor: Maybe I'm not thinking straight, but `v` isn't defined, right? It should be `ship` instead?

laarmen: Oooh, right :-)

PR #3710, opened 2016-05-25
Fix crash when player explodes in-dock.

### LINKED ISSUE #2116: Assertion `!Pi::player->IsDead()' failed.
From @jmf:

    <john_minetest> pioneer: WorldView.cpp:1502: virtual void WorldView::Draw(): Assertion `!Pi::player->IsDead()' failed.
    <john_minetest> Program received signal SIGABRT, Aborted.
    <john_minetest> 0x00007ffff4e7e2c5 in raise () from /usr/lib/libc.so.6
    <john_minetest> I get this error when executing Game.player:Explode()

fluffyfreak:
Remove bogus assert, it is not necessary to check that the player is alive.
This fixes #2116

### DIFF
--- a/src/WorldView.cpp
+++ b/src/WorldView.cpp
@@ -1962,7 +1962,6 @@ void WorldView::Draw()
 {
 	assert(m_game);
 	assert(Pi::player);
-	assert(!Pi::player->IsDead());
 
 	m_renderer->ClearDepthBuffer();
 


laarmen:
Why is it not necessary?

fluffyfreak:
Remove it and everything behaves as expect, there's no crash, nothing depends on the player being alive.
You do need a valid player pointer, but we have that so it's fine.

laarmen:
:+1:

PR #3715, opened 2016-06-09
Enable using dedicated nVidia GPUs in Windows laptops with Optimus.

fluffyfreak:
Enable using dedicated nVidia GPUs in Windows laptops with Optimus.

This only affects Windows laptops with nVidia GPUs.
Previously I was seeing that we're always picking the Intel GPU in laptops with an integrated Intel GPu and a dedicated nVidia GPU. 
There are ways around this, like forcing _everything_ to use the nVidia GPU, but that's not ideal.

### DIFF
--- a/src/win32/OSWin32.cpp
+++ b/src/win32/OSWin32.cpp
@@ -16,6 +16,13 @@
 #include <wchar.h>
 #include <windows.h>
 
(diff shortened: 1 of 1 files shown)


fluffyfreak:
I should also add that this doesn't override the users settings, it just let's nVidia's Optimus technology recognise that it **can** use their GPU by choice.

laarmen:
With the later info, :+1

PR #3716, opened 2016-06-09
Minor Noise library improvements.

fluffyfreak:
Removed the component-wise noise and just pass the vector3d directly by reference.
Some other changes which make minor differences to runtime performance.

The effect of this PR could be a bit chaotic as it risks changing the results we get back from the noise library. However I have explicitly **avoided** making any changes which affect the actual values returned for a given input so hopefully that won't be the case.


laarmen:
:+1

(I don't think anybody in the currently more-or-less active dev community
knows more about this stuff than you, so in Fluffy we trust!)

fluffyfreak:
@laarmen thanks! :)
I took before and after screenshots on New Hope and couldn't see a single pixels difference so I think it's ok.

Issue #2176, opened 2013-03-31
Problem with shaders using linux Ati open source drivers (mesa r600). Snow-like sky and invisible planet surface.


Tichy:
Maybe caused by #1353 ?

Earth
![2013-03-31-175359_1920x1200_scrot](https://f.cloud.github.com/assets/2348771/321959/9b157d88-9a1b-11e2-89bd-170de4a84f4f.png)

No glitches, if I disable the shaders.

Using Arch Linux 64bit with ati open source drivers.

Tried with two ATI VGA
ATI Radeon HD6770
VGA compatible controller: Advanced Micro Devices [AMD] nee ATI Juniper XT [AMD Radeon HD 6000 Series]

My opengl.txt
OpenGL version 3.0 Mesa 9.1.1, running on X.Org Gallium 0.4 on AMD JUNIPER
Shading language version: 1.30
http://pastebin.com/PK9jP2XY

ATI Radeon HD4650
VGA compatible controller: Advanced Micro Devices [AMD] nee ATI RV730 PRO [Radeon HD 4650]

opengl.txt
OpenGL version 3.0 Mesa 9.1.1, running on X.Org Gallium 0.4 on AMD RV730
Shading language version: 1.30
http://pastebin.com/K8em2Zwg

Console messages:
EE r600_shader.c:155 r600_pipe_shader_create - translation from TGSI failed !
EE r600_state_common.c:755 r600_shader_select - Failed to build shader variant (type=1) -1
EE r600_shader.c:155 r600_pipe_shader_create - translation from TGSI failed !
EE r600_state_common.c:755 r600_shader_select - Failed to build shader variant (type=0) -1

With mesa-r300-r600-radeonsi-git 9.2.56850 on Radeon HD6770
opengl.txt
OpenGL version 3.0 Mesa 9.2.0 (git-6220cc9), running on X.Org Gallium 0.4 on AMD JUNIPER
Shading language version: 1.30

console messages:
Warning: R600 LLVM backend does not support indirect adressing. Falling back to TGSI backend.
EE r600_shader.c:158 r600_pipe_shader_create - translation from TGSI failed !
EE r600_state_common.c:751 r600_shader_select - Failed to build shader variant (type=1) -1

PR #3721, opened 2016-06-18
Eclipse r600

fluffyfreak:
Re-write the way the shader is setup to avoid using uniforms to control the number of iterations for the loops. This allows it to avoid the problems of #2176.

Also `Mar 31, 2013` it's taken more than 3 years to get this fixed and the MESA "fix" was supposedly imminent all the way back then... (╯°□°）╯ ┻━┻

# Bad Things:
This means generating the shader several times, once for each variant.
It is a little unintuitive to do this manually.
I've pulled in a change to the `01_epsilon_eridani.lua` seed that puts `Itzalean` back on the equator but I think it needs yet another seed to actually get all 4 stations into their intended locations.

# Todo:
There's some unfinished stuff in here.
- [x] limit the number of shadows / programs that can be created.
- [x] more documentation as to what/why is going on.

Andy

### DIFF
--- a/data/shaders/opengl/eclipse.glsl
+++ b/data/shaders/opengl/eclipse.glsl
@@ -3,8 +3,6 @@
 
 #ifdef ECLIPSE
 
-uniform int shadows;
-uniform ivec3 occultedLight;
 uniform vec3 shadowCentreX;
 uniform vec3 shadowCentreY;
 uniform vec3 shadowCentreZ;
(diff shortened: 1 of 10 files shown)


fluffyfreak:
@Tichy you might be interested in this one :)

fluffyfreak:
For that matter @shadmar you might be if you're still involved with Jumpdrive/Paragon.

Tichy:
awesome! wow! I'll try it as soon as I get back to my PC!

Tichy:
screenshot-20160620-161913

it works!!!

laarmen:
That's one hell of an eclipse!

shadmar:
> For that matter @shadmar you might be if you're still involved with Jumpdrive/Paragon.
Nah, I left the ship when they went commercial.

PR #3725, opened 2016-06-24
Keep cargo if buying a new ship

robothauler:
The loaded cargo disappears if the player buys a new ship.
The player loses the money and if there is an active mission with already loaded custom cargo the mission cannot succeed.

This PR should solve this problems.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -1095,6 +1095,10 @@
       "description" : "",
       "message" : "Ship is too small for the current crew."
    },
+   "TOO_SMALL_TO_TRANSSHIP" : {
+      "description" : "",
+      "message" : "Ship is too small to transship your cargo."
+   },
    "TOTAL" : {
       "description" : "",
       "message" : "Total: "
--- a/data/ui/StationView/ShipMarket.lua
+++ b/data/ui/StationView/ShipMarket.lua
@@ -90,6 +90,13 @@ local function buyShip (sos)
 		return
 	end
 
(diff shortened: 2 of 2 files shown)


walterar:
:

data/ui/StationView/ShipMarket.lua
```diff
@@ -106,6 +113,9 @@ local function buyShip (sos)
 	if def.hyperdriveClass > 0 then
 		player:AddEquip(Equipment.hyperspace["hyperdrive_" .. def.hyperdriveClass])
 	end
+	for _, e in pairs(manifest) do
+		player:AddEquip(e)
```
impaktor: Is there a TAB here? after the `for`?

impaktor: Never mind, did a silly comment since my font is too small to read, please disregard.

fluffyfreak:
I like the idea of this :) good implementation too.

data/lang/ui-core/en.json
```diff
@@ -1095,6 +1095,10 @@
       "description" : "",
       "message" : "Ship is too small for the current crew."
    },
+   "TOO_SMALL_TO_TRANSSHIP" : {
+      "description" : "",
+      "message" : "Ship is too small to transship your cargo."
+   },
```
impaktor: "transship"? Shouldn't it be "transfer"?

If so, I can push change ASAP -- no need for new pull request.

impaktor: Also, never mind this. I looked it up, and learned something new.

laarmen: What happens if you can't fit all the cargo?
:-)

laarmen: Hem, nevermind, I just looked at the loop, not the code block above.

PR #3731, opened 2016-06-25
Disable Continue button if no saved game to load.

### LINKED ISSUE #3727: Cryptic error message when attempting to continue game when game has never been played
How to repdroduce:

* Start Pioneer the first time
* Click on “Continue game”
* You see a cryptic error message complaining about a missing save file called “_exit”

I suggest to just disable or remove the “Continue game” button if _exit is not there.

fluffyfreak:
Add method to test if a save game file exists for loading. Use it in the main menu.

Should fix #3727

### DIFF
--- a/data/ui/MainMenu.lua
+++ b/data/ui/MainMenu.lua
@@ -115,6 +115,11 @@ for i = 1,#buttonDefs do
 	button.onClick:Connect(def[2])
 	if i < 10 then button:AddShortcut(i) end
 	if i == 10 then button:AddShortcut("0") end
+	if 1 == i then
+		if not Game.CanLoadGame("_exit") then
+			button:Disable()
+		end
+	end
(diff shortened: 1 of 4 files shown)


impaktor:
I just looked at the code for less than 2 seconds, so I assume I'm wrong, but since this whole "issue" is of nitpicking character, what happens if the player has auto-save, then disables auto-save before the _exit save is created?

Will not the continue button be disabled, even though there are autosaved states, jiust not the _exit one?

fluffyfreak:
Yes it will but the continue doesn't load just any autosave, only the `_exit` autosave, so the point is moot.

walterar:
Is this the inverse of K.I.S.S?

fluffyfreak:
Keeping something simple for the player is sometimes not as clean as keeping it simple in the code but worth doing anyway.

PR #3743, opened 2016-06-26
Make trade analyzer removable again. Undo #3474, close #3742

### LINKED ISSUE #3742: Untranslatable string: Uninstalling trade analyzer fail
It looks like a particular string is untranslatable: If you attempt to uninstall the trade analyzer, the game complains that software can't be sold back. But this message is only in English. It does not look like this string has been exposed for translation.

(search for “system software”)

impaktor:
Re-enable selling of Trade analyzer. As I mentioned in #3742:

> It was implemented by someone who had the distinct idea that this must be "software", and that you can't sell software back. I say: it could be a cartrige, or license, or subscription, or whatever. I want it to have the same behavior as other equipment, just like it used to be. I don't see how the game becomes more "fun" by not being able to sell some equipment that you've bought.

Since this touches some equipment philosophy, ping @laarmen

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -64,7 +64,7 @@ function EquipType:Serialize()
 			ret[k] = v
 		end
 	end
-	
+
 	if debug.dmodeenabled() and self.slots == "cargo" then
 		for _,v in pairs(cargo) do
 			if (v.l10n_key == self.key and v.l10n_resource == self.l10n_resource) then
@@ -73,7 +73,7 @@ function EquipType:Serialize()
 			end
 		end
 	end
-	
+
 	ret.volatile = nil
 	return ret
 end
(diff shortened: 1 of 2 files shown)


sloanonlinux:
If you sell it, wouldn't that be piracy?

fluffyfreak:
No strong feelings one way or the other

impaktor:
@sloanonlinux But it's not necessarily software. Maybe it's a subscription device. Maybe it's like a game console cartridge. Maybe it's a license, that allows reselling. T think treating the trade analyzer (it should have been called "trade computer") like i absolutely _must_ be software and that the licence in 3200 is same as Windows or something, is to me a serious case of overthinking it.

It's not intuitive that the player, for some reason, can't get rid of some piece of equipment on your ship.

sloanonlinux:
@impaktor Why would you want to get rid of it?

laarmen:
That's a bad question from a development POV. If people want to do stuff
it's their business. For one, I'm glad we were able to implement this
limitation, but the limitation itself goes against where I wish our
software/copyright systems were going, so I'm all for merging this PR.

Le dim. 26 juin 2016 21:01, sloanonlinux <notifications@github.com> a
écrit :

laarmen:
While I'm at it, I have to point out that the way it was implemented before was an abuse of the capabilities. The ship doesn't care that it has software installed, this data should have been stored into its own field.

PR #3745, opened 2016-06-27
Improve distance format, by introducing ly and Mm. 

### LINKED ISSUE #218: numbers on the screen
When having selected a target with a large distance, you often see a list of numbers beyond a practical limit of the brain. I often can't read the the super fast changing strings of numbers and make mistakes of large proportions. 
There is no point in having a distance of hundreds of thousands of kilometers, having even the decimals shown.

For readability the game already switches from 14959787070 (you get the issue now ;-) ) to 0.1 AU, how about something similar for the smaller values: 14.9 Mkm, 1.49 Mkm, 0.149 Mkm, and decimals only shown with distances under 10 km.

impaktor:
Fixes #218.

Given a number 0.12345600000000000000, and moving the decimal point ever to the left, 20 times, new algorithm gives:

```
1 m
11 m
112 m
1.12 km
11.23 km
112.35 km
1.12 Mm
11.23 Mm
112.35 Mm
1123.46 Mm
0.08 AU
0.75 AU
7.51 AU
75.10 AU
750.98 AU
0.12 ly
1.19 ly
11.87 ly
118.75 ly
1187.50 ly
```

Noticeable here, besides adding Mm is that I've added "ly", for those who travel in _very_ large systems, or just want to mess about, and also that I favour letting AU go down to the hundredths so first for distances shorter then 0.01 AU does it switch to metric system, with Mega meters, whereas before it switched from AU to km at 0.1 AU.

Actually, both AU and ly aught to be in translation system, no?

Now format is: ly -> AU -> Mm -> km -> m.


impaktor:
Added AU and ly to translation system, (as can be seen). AU is already translated in Lua side with ui-core, but not in core, so a bit of "double banking".

PR #3746, opened 2016-06-28
Passenger warns when jumping to wrong system with right name

### LINKED ISSUE #3739: Taxi mission: Show snarky comment when jumping to wrong system but equal name
I once made a taxi mission where I had to jump to system XYZ. I jumped into system XYZ, but I noticed too late it was the wrong one. There were two systems sharing the name.
Yes, I know that equal names of systems are intentional.

Still I think this can be a bit unfair to new players because they don't know why they can't complete the mission. I suggest to add a snarky comment when the player jumps into a system which has the same name as the target system but it is not the target system.

Something along the lines of: “Hey, that's not the system I was looking for. Have you checked the sector coordinates?”

(And maybe some flavor variants.)

It's just a snarky comment which does not have any other effect (no reputation loss). It can help to get rid of one FAQ. :-)

impaktor:
Closes #3739

I've tested each line in game, but not all lines together in the if-statement, since I did not want to find a mission to a non-unique system, but this should be fine, but maybe @Wuzzy2 like to try/test it.

### DIFF
--- a/data/lang/module-taxi/en.json
+++ b/data/lang/module-taxi/en.json
@@ -583,6 +583,10 @@
       "description" : "",
       "message" : "Will I be in any danger?"
    },
+    "WRONG_SYSTEM" : {
+      "description" : "Shown if system name is right, but coordinates are wrong",
+      "message" : "The system name is right, but not the coordinates."
+    },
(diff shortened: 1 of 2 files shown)


fluffyfreak:
Nice, merged :)

impaktor:
thank!

It can stil be tested retroactivley, from this PR, if anyone is interested, until _next_ release of pioneer.

PR #3747, opened 2016-06-30
Swap LuaVector for a table

fluffyfreak:
Don't use LuaVector instead just return a table with x,y,z components for GetVelocity and GetPosition.

This works should let people actually use these functions now.


fluffyfreak:
@impaktor & @laarmen I've tried this out but a 2 seconds sanity check is always a good idea :)

I think that this should allow CmdrBugBear to get the info that he needs, @nozmajner might also be interested.

PR #3748, opened 2016-07-03
Fix grammar error in module-cargorun

FortressBuilder:
The [Translations guide](http://pioneerwiki.com/wiki/Translations) says translations should go through Transifex, but the error is in the original English source string and I couldn't find an English language option on Transifex.

### DIFF
--- a/data/lang/module-cargorun/en.json
+++ b/data/lang/module-cargorun/en.json
@@ -577,7 +577,7 @@
    },
    "WHYSOMUCH_INFRASTRUCTURE_1" : {
       "description" : "",
-      "message" : "These goods crucial for a construction project."
+      "message" : "These goods are crucial for a construction project."
    },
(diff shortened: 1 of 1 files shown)


impaktor:
Right, only english is handled through git.

Could you please remove all files except the english one? (The rest will auto-update.)

Use `git --amend` followed by a forced push to your github `git push -f origin`.

impaktor:
Thanks. I/we might let this simmer for a while, as pusing strings triggers a re-translation of that string in all languages.

If you see other errors, you can just push those commits to this branch.

fluffyfreak:
@impaktor getting to the end of the month and this has had 22 days so far, I think it's simmered enough how about you?

impaktor:
I have no opinion on the matter.

fluffyfreak:
Merging!

PR #3750, opened 2016-07-06
Comment out cryptic debug string

impaktor:
@fluffyfreak I didn't know if I should remove or keep the string, but I'll comment it out, since I'm getting warnings.

Possibly it should be used in an assert?

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -265,7 +265,7 @@ std::string FloatToStr(float val)
 	static_assert(sizeof(float) == 4, "float isn't 4bytes");
 	fu32 uval(val);
 	char str[64];
-	const int amt = sprintf(str, "%" PRIu32, uval.u);
+	//const int amt = sprintf(str, "%" PRIu32, uval.u);
 	return str;
 #endif
 }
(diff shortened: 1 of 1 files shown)


fluffyfreak:
@impaktor That's not a debug string, that's the actual value assignment.
If you comment it out then the function doesn't do anything.

src/utils.cpp
```diff
@@ -282,7 +282,7 @@ std::string DoubleToStr(double val)
 	static_assert(sizeof(double) == 8, "double isn't 8 bytes");
 	fu64 uval(val);
 	char str[128];
-	const int amt = sprintf(str, "%" PRIu64, uval.u);
+	//const int amt = sprintf(str, "%" PRIu64, uval.u);
```
fluffyfreak: As above.

fluffyfreak:
What is the warning that you're getting? The `static_assert` should catch mismatched sizes at compile time but it might be a GCC vs VS2015 issue where the type needs to be different on GCC instead of `PRIu64` or `PRIu32`.

impaktor:
`amt` unused variable. And I can see why it would think that, since I also thought it. But sure, I haven't looked deeper into the code, so you're probably right.

fluffyfreak:
ah amt is unused, in the other functions it's used for an assert to check that it hasn't overwritten the string, but in those two functions I clearly forgot to add that assert

impaktor:
I'll close this, and forget it ever hapened.

PR #3751, opened 2016-07-06
Faster float convert

fluffyfreak:
After a bit of fiddling around I discovered than using `itoa` was about 4 times faster than `sprintf` for integer to string/char[] conversion.

Also `atoi` is roughly 5.6 times faster than `sscanf` for the inverse.

Of course we're still coping lots of strings around which probably dominates the actual operations but one thing at a time! Plus this silences a large number of warnings under GCC.

Andy


walterar:
After merging this, retrieving a saved game I lost my rear camera. Rare.

PR #3753, opened 2016-07-08
Hyperlaw

impaktor:
We can debate to infinity for very complicated rules on how/when spooling up hyperdrive should be physically possible, and/or legal.

I believe we should keep it simple. This does that. It can be expanded upon a later time.

With this, jumping is illegal at altitude below 1000 m on planets with atmosphere. We can increase altitude based on feedback from players. (How was it in Frontier?). I feel/guess 100 m is much too low, and 100 km is perhaps too high.

(If altitude is a variable depending on planet/ship/atmosphere/speed etc. then there really must be a way for the player to see how much further it is to "allowed" altitude. I'm also not sure that such a complication would add to the game).

## Rules:
- 1 hide button if no valid target selected
- 2 gray inactive button if landed
- 3 "forbidden" button when too low altitude to jump
- 4 "forbidded abort" button when countdown at too low altitude
- 5 engage button when legal jump
- 6 abort button when "legal" countdown

## TO-DO
- [x] Move `ship::isjumpallowed` from C++ to Lua
- [x] __Fix bug__ that causes crash upon illegal jump. Possibly related to leaving the system, and it being thrown away.
- [ ] AI should not break the law when jumping (do they?)
- [x] Test that button is in right state after a jump.
- [x] Clean up in commits. Probably sholdn't auto-clean up old indentation-mistakes
- [x] make countdown longer. 1 second for some ships is about 10 times too short.
- [x] Add professional version of the three new icons, ping @nozmajner when he comes back from camp

## BUGS
2016-08-17 update: have now moved the checking of if a jump is legal or not to Lua, and have discovered new bugs:
- [x] State of button is wrong after a jump
- [x] AI takes over when doing an illega hyperjump (wtf??)
- [x] Segfault after a jump + autopilot to station


walterar:
In some systems claim that the effects of cloud leap is good for health. Is it true?
KISS ;)

laarmen:
This should not be implemented in C++. It is policy, and policies should be
moddable.

Le ven. 8 juil. 2016 16:15, Walter Arnolfo <notifications@github.com> a
écrit :

impaktor:
Well, I don't say any way around the WorldView problem, so there's bound to be C++ code.

Since I feel my head spinning when trying to get Lua and C++ to communicate, I'd rather not. I followed the structure that was in place, like how Ship does checks if hyperspace is possible, and now also if hyperspace is "allowed".

Please feel free to submit code, commits, examples, or whatever.

walterar:
Fiddling with this, here's a mod with some ideas that work. It also improves the functioning of the police and fixes #3512

[](https://dl.dropboxusercontent.com/u/15391580/Pioneer/illegalJump-bestPolice-TestMod.zip)

This are the changes:

```
--- /orig/CrimeTracking.lua
+++ /mod//CrimeTracking.lua
@@ -1,12 +1,13 @@
 -- Copyright © 2008-2016 Pioneer Developers. See AUTHORS.txt for details
 
-local Event = import("Event")
-local Game = import("Game")
-local Engine = import("Engine")
-local Timer = import("Timer")
+local Event      = import("Event")
+local Game       = import("Game")
+local Engine     = import("Engine")
+local Timer      = import("Timer")
 local Serializer = import("Serializer")
-local Legal = import("Legal")
+local Legal      = import("Legal")
+local Comms      = import("Comms")
 
 -- Fine at which police will launch and hunt donwn outlaw player
 local maxFineTolerated = 300
@@ -16,7 +17,7 @@
 
 -- check if we should dispatch police, or call them back
 local function doLawAndOrder ()
-	if Game.player.flightState == "HYPERSPACE" then return end
+	if not Game.system then return end
 
 	local crimes, fine = Game.player:GetCrimeOutstanding()
 	if not policeDispatched then
@@ -96,17 +97,43 @@
 
 local onShipFiring = function(ship)
 	if ship:IsPlayer() then
+ship:SetInvulnerable(true)--XXX
 		Legal:notifyOfCrime(ship,"WEAPONS_DISCHARGE")
 	end
 end
 
 
+local distance
+local onShipUndocked = function (ship, station)
+	if not ship:IsPlayer() then return end
+	distance = 0
+	local system = Game.system
+	Timer:CallEvery(1, function ()
+		if Game.system ~= system then return true end
+		distance = ship:DistanceTo(station)
+		if distance > 15000 then return true end
+	end)
+end
+
+
 local onLeaveSystem = function(ship)
 	if not ship:IsPlayer() then return end
+	local station = Game.player:FindNearestTo("SPACESTATION")
+	local fined = 1999.99--XXX
+	local MinJumpDistance = 10000
+
+	if Game.system.faction.name == 'Solar Federation' then MinJumpDistance = 5000 end
+	if Game.system.faction.name == 'Independent' then MinJumpDistance = 1000 end
+
+	if distance < MinJumpDistance then
+--		Legal:notifyOfCrime(ship,"ILLEGAL_JUMP")--XXX
+		Comms.ImportantMessage("You have been fined $"..fined.." by jumping very close to "..station.label, "Police")
+	end
+
 	-- if we leave the system, the space station object will be invalid
 	policeDispatched = nil
+
 end
-
 
 local onGameEnd = function ()
 	policeDispatched = nil
@@ -114,7 +141,7 @@
 
 
 Event.Register("onShipHit", onShipHit)
-Event.Register("onShipDestroyed", onShipDestroyed)
+Event.Register("onShipUndocked", onShipUndocked)
 Event.Register("onShipFiring", onShipFiring)
 Event.Register("onJettison", onJettison)
 Event.Register("onGameStart", onGameStart)
```
```

--- /orig/SpaceStation.lua
+++ /mod/SpaceStation.lua
@@ -309,6 +309,26 @@
 SpaceStation.lawEnforcedRange = 100000


+
+local Exists = function (ship)
+	local exists = false
+	if ship:exists() then
+		exists = true
+	end
+	return exists
+end
+local ShipExists = function (ship)
+	if ship then
+		ok,val = pcall(Exists, ship)
+		if ok then
+			return val
+		else
+			return false
+		end
+	end
+end
+
+
 local police = {}

 --
@@ -332,10 +352,8 @@
 --
 function SpaceStation:LaunchPolice(targetShip)
 	if not targetShip then error("Ship targeted invalid") end
-
-	-- if no police created for this station yet:
-	if not police[self] then
-		police[self] = {}
+-- police created for this station
+		police = {}
 		-- decide how many to create
 		local lawlessness = Game.system.lawlessness
 		local maxPolice = math.min(9, self.numDocks)
@@ -344,8 +362,8 @@

 		-- create and equip them
 		while numberPolice > 0 do
-			local policeShip = Space.SpawnShipDocked(shiptype.id, self)
-			if policeShip == nil then
+			local policeShip = Space.SpawnShipNear(shiptype.id, Game.player, 6, 8)
+			if not ShipExists(policeShip) then
 				return
 			else
 				numberPolice = numberPolice - 1
@@ -356,12 +374,11 @@
 				policeShip:AddEquip(Equipment.misc.laser_cooling_booster)
 				policeShip:AddEquip(Equipment.cargo.hydrogen, 1)

-				table.insert(police[self], policeShip)
+				table.insert(police, policeShip)
 			end
 		end
-	end
-
-	for _, policeShip in pairs(police[self]) do
+
+	for _, policeShip in pairs(police) do
 		-- if docked
 		if policeShip.flightState == "DOCKED" then
 			policeShip:Undock()
@@ -392,11 +409,10 @@
 --
 function SpaceStation:LandPolice()
 	-- land command issued before creation of police
-	if not police[self] then return end
-
-	for _, policeShip in pairs(police[self]) do
-		if not (policeShip.flightState == "DOCKED") and policeShip:exists() then
-			policeShip:CancelAI()
+	if not police then return end
+
+	for _, policeShip in pairs(police) do
+		if ShipExists(policeShip) and policeShip.flightState == "FLYING" then
 			policeShip:AIDockWith(self)
 		end
 	end
@@ -698,15 +714,6 @@
 	destroySystem()
 end)

-Event.Register("onShipDestroyed", function (ship, _)
-	for _,local_police in pairs(police) do
-		for k,police_ship in pairs(local_police) do
-			if (ship == police_ship) then
-				table.remove(local_police, k)
-			end
-		end
-	end
-end)

 Event.Register("onGameEnd", function ()
 	destroySystem()
@@ -719,13 +726,19 @@
 	shipsOnSale = {}
 end)

-
 Serializer:Register("SpaceStation",
 	function ()
+
+		for k,ship in pairs(police) do
+			if not ShipExists(ship) then
+				police[k] = nil
+			end
+		end
+
 		local data = {
 			equipmentStock = equipmentStock,
 			equipmentPrice = equipmentPrice,
-			police = police,  --todo fails if a police ship is killed
+			police = police,
 			shipsOnSale = {},
 		}
 		for station,list in pairs(shipsOnSale) do
```

impaktor:
## Thoughts regarding the bug

The main difference I see is that in `onLeaveSystem` in Crimetracking you / @walterar call

    local station = Game.player:FindNearestTo("SPACESTATION")
    ....
    distance = ship:DistanceTo(station)

which apparently works (I haven't tested), whereas I call

    station:DistanceTo(Game.player)

from Legal (which in turn was called from `onLeaveSystem` in Crimetracking-module); even though my "station" object appears to be valid it triggers a crash for me.

I'm thinking I could have a `onShipUndocked` like you, but I only need to check which station the player undocked from, which can be stored in a variable.

Regarding your fix in SpaceStation, I'll have to play around with it when I have pioneer in front of me, but looks like the underlying bug might still there, but it's not "allowing"/causing the game to crash (which is good), but I prefer to find the real cause of the crash to begin with.


Also, I'm all for making factions more "different" from each other, but if one is to have different hyperjump minimum altitudes for factions, it should be very clear to the player, who probably doesn't even know he's in another faction. Putting law/rules on Lobby screen feels very "artificial", as they belong on the Police tab, where no one ever clicks anyway.

Also, _if_ having faction dependent minimum jump distances, the police message
aught to be different:

- "You are in violation of SolFed safety protocols",

- "CIW detected breach of safety regulation"

## Thoughts on Lua vs. C++

> This should not be implemented in C++

Maybe the `IsHyperjumpAllowed()` would be useful in lua side, provided it being called every frame from C++ won't cause speed/load issues, but then there are many other ship functions that aught to be exposed to Lua, I know it's not a good argument from my side, but I did it this way, since that's how I know to do it. Didn't even occur to me to do it in Lua to be honest. If called upon to switch it to Lua, then this PR is at risk to go the same way as #2990 (i.e. stalled), which incidentally celebrated 2 years just the other week.

(I wish I understood the Lua/C++ interface better. Every time I tried I either get a suicidal depression or murderous rage. Worst is I don't even understand how to make myself understand.)

walterar:
@impaktor 

Within onLeaveSystem; for DistanceTo, Game.player is no longer a body; for FindNearestTo, yes. Why? That is to another issue. ;)

Thus, I measure the distance onShipUndock, and I use it in onLeaveSystem. That's how it works.

Note that this is only the little engine that drives to work an basic initial idea. You can apply some crazy ideas and expand everything you want.

ship:SetInvulnerable(true) on onShipFiring,  serves to make it more fun the mod test.  If you get shot, you will be invulnerable, and after the battle you can do many things; among them, find out if Jupiter has a solid core, your ship will resist the pressure.

Off topic: FindNearestTo is a very util function. Good job.

"Every time I tried I either get a suicidal depression or murderous rage. Worst is I don't even understand how to make myself understand." To me the same happens many times. ;))

fluffyfreak:
# Thoughts on Lua vs. C++
> This should not be implemented in C++

If it's an issue about allowing mods to control it, then perhaps have an option in a new config file under `data/configs` call it `data/configs/rules.ini` but otherwise I think that it's implementors choice.

It's unreasonable to ask someone to implement something in a way which takes a simple, quick, easy task taht they feel confident about implementing and turn it into a nightmare which they eventually rage-quit over.

# The "Bug"
I have no idea why that's being a problem, I also don't have time or capacity to look into it right now :(
If there's a workaround that simply involves checking something isn't `nil` then use that route.

# The PR
Sorry that I don't have more time to look at this whilst you're developing it, overall I'm happy with the direction that it's going in, and the decisions you're making so stay on track, get it done how you want and I will help review/debug it nearer the end of that process :)

Andy

laarmen:
Oh, I'm not vetoing this or anything. That's just a good candidate for
later hacking :-)

Le dim. 10 juil. 2016 14:03, Andrew Copland <notifications@github.com> a
écrit :

nozmajner:
I uploaded the icons for it in my [hyperspace_icons branch](https://github.com/nozmajner/pioneer/tree/hyperspace_icons). There's only one commit in it: [e9522da](https://github.com/nozmajner/pioneer/commit/e9522dafc5389ec40d7d225b91284a32553464a5) .
Or should I do a PR for them?

impaktor:
@nozmajner  Added it. Thanks!

(I put two line breaks in your commit message, thus it has a different commit-hash)

Also, more gerneral question, is there a point in suffixing these files with "f8" anymore? 

The actual button used for jumping is F7 apparently.

impaktor:
@laarmen don't worry about it, I didn't interpret your comment in any "bad" way, but since your comment was valid, I had to think (and motivate) why I didn't do it as you suggested. Ideally the jumprules should be in Lua, but I leave that as an exercise to the motivated. :wink:

impaktor:
@laarmen did you have any thoughts on how to make player not invalid when called from a OnLeaveSystem hook?

walterar:
@impaktor OnLeaveSystem is activated when an ship jumps into hyperspace. Therefore ship is no longer a valid body in this system since ship.flightState = "HYPERSPACE"
The strange thing here, in my test mod, is that FindNearestTo recognizes the ship (Game.player) as valid body.

impaktor:
@walterar yes, I know. I think the question is how we make it so that the system isn't destroyed until _after_ onLeaveSystem has finished.

walterar:
@impaktor  It seems that you're right.

Seeing the original design of the event ...
-----------------------
**onLeaveSystem**

Triggered immediately before ship leaves a system and enters hyperspace.

local onLeaveSystem = function (ship) ... end
Event.Register("onLeaveSystem", onLeaveSystem)

If the ship was the player then all physics Body objects are invalid after this method returns.

**Parameters**
ship	the Ship that left the system
**Availability**
alpha 10

Status
stable

-----------------------
...things are not working properly.

The ship should enter hyperspace after of onLeaveSystem executed, and before serializer.

laarmen:
Allright, the onLeaveSystem bug has been solved in my fork, @impaktor should pick it up.

Now, about the AI stuff, they certainly don't follow the law since there's no AI call to initiate an hyperjump, there's just the standard one that is also used by the player. The only sensible solution I have here is to track down each of those calls and... act according to what's needed. Like, if there's a ship trying to flee an assassination attempt, I'm guessing they won't bother respecting the law. The only thing I dislike about this is what to do when you can't jump. Do we have an AI call to get to some altitude?

impaktor:
OK, I'm starting to feel done with this. The rules for jumping is now defined in Lua.

At the moment, I set 10 km for planets and 1 km for orbital to be minimum distance for jumping.

What is lacking, or rather, what I'm giving up on for now, is the AI/NPC:

Ideally, `AIFlyTo` could have a second optional argument where one sets the distance to the target (at the moment, it's hard coded to 15000 m I think), then use the current bodyframe as target and minimum legal jumping distance as distance to AIFlyTo. I gave up on that and just call `AIEnterLowOrbit`; however, it hardly works, since the code will not wait for AI to complete before initiating the hyperdrive.

Everything else should work fine.

Also, thanks to @nozmajner and @laarmen for each contributing with a commit.

laarmen:
I say, we merge this and open an issue to track the "jump before reaching orbit" problem. Any objections?

impaktor:
> Any objections?

Not from me.

fluffyfreak:
Good with me

PR #3757, opened 2016-07-17
Point sprite background stars

fluffyfreak:
Whilst working on generating a environment cube map I noticed that the rendering of point stars into the texture is very ugly due to them being points.

This PR switches them over to being a texture an using texture mapped point sprites which should help to avoid the issue.


walterar:
Andy, there are many reasons why I think this should be reversed.

fluffyfreak:
@walter And they are? Also, better to start a new issue, I don't get told when you comment on this.

walter:
@fluffyfreak think you meant @walterar, unsubscribing...

fluffyfreak:
@walterar And they are? Also, better to start a new issue, I don't get told when you comment on this.

walterar:
@fluffyfreak 
1) This breaks what I like about the original system: the effect "twinkling stars". We discussed this when skybox, I thought the issue was over.

2) This has removed the "hyperspace" effect.

impaktor:
> 2) This has removed the "hyperspace" effect.

Then this is worth opening an issue on. Possibly with a screenshot?

fluffyfreak:
@walterar Ah the hyperspace jump being broken is a bug, it's using the wrong shader. I'll fix that now, I usually do something else when hyperspacing so I guess I just missed it.

The stars still twinkle on mine, the smallest ones anyway, in fact it should look *almost* the same as the original effect it just allows an artist to control the look of the stars.

So for Scout+ you can change the file 'textures/star_point.png' to something that you like, at the moment it's a circle, but it could be a fancy cross star, or a small square, or even a smiley face :smiley: 
The idea was to give you more control over it, so maybe make the circle smaller and it will make the twinkling more pronounced (visible).

walterar:
@fluffyfreak  I have set my taste in Background.cpp
```

--- /pioneer/src/Background.cpp
+++ /pioneersp/src/Background.cpp
@@ -244,7 +244,7 @@
 			1000.0f * sqrt(1.0f - u*u) * cos(theta),
 			1000.0f * u,
 			1000.0f * sqrt(1.0f - u*u) * sin(theta));
-		vtxPtr->col = Color(col, col, col,	255);
+		vtxPtr->col = Color(rand.Double(0.3, 0.5)*255, rand.Double(0.4,0.7 )*255, col, 255);
 
 		//need to keep data around for HS anim - this is stupid
 		m_hyperVtx[NUM_BG_STARS * 2 + i] = vtxPtr->pos;

```

How do I make this with the new code?

fluffyfreak:
@walterar I'll need to re-add support for that, I'll do it tonight

fluffyfreak:
@walterar there you go #3774

PR #3760, opened 2016-07-18
Did nothing; please ignore.

jfacoustic:
I'm not sure anything happened with this.

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -1,6 +1,7 @@
 July 2016
    * Fixes
      * SAR: fix fuel transfer (#3755)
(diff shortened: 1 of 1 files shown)


impaktor:
There are three tabs at the top of a git Pull Request (PR): "conversation", "commits" and "Files changes",  you can see that this contains 1 commit, and 1 file changed, which is the Changelog, which we update when we merge your code, so no need to do anything with that.

As the title of this PR points out, this does nothing, but you're learning git fast I see. Btw, you can close PRs, button at bottom, but I'll close this for you now.

jfacoustic:
Thanks impaktor!

PR #3762, opened 2016-07-18
Fixed fullscreen for mac; fixes issue #3348

### LINKED ISSUE #3348: Need a way to exit fullscreen on OS X
There doesn't seem to be a way to use any other app while in Pioneer full screen. I can't cmd-tab to my browser, for example, or use Mission Control or Exposé.

jfacoustic:
Fixes #3348

### DIFF
--- a/src/graphics/WindowSDL.cpp
+++ b/src/graphics/WindowSDL.cpp
@@ -11,7 +11,7 @@ namespace Graphics {
 
 bool WindowSDL::CreateWindowAndContext(const char *name, int w, int h, bool fullscreen, bool hidden, int samples, int depth_bits) {
 	Uint32 winFlags = SDL_WINDOW_OPENGL | (hidden ? SDL_WINDOW_HIDDEN : SDL_WINDOW_SHOWN);
-	if (!hidden && fullscreen) winFlags |= SDL_WINDOW_FULLSCREEN;
+	if (!hidden && fullscreen) winFlags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
 
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 1);


impaktor:
Great! A clean PR. :+1: 

I'll leave it to @fluffyfreak to comment on the code change.

fluffyfreak:
A nice first try but not the correct fix I think.

What this does is allow Mac/Win/Linux users to switch between other programs whilst the game is "fullscreen", this is because the game is not really fullscreen, it's in borderless+foremost and set to the same resolution and colour depth as the desktop. This is a lower performing mode and it also comes with a few drawbacks...

You can no longer set a different (`lower`) resolution than that of the native screen. You are forced to run at full resolution, even if like me that's a massive 2160x1440, or even higher for retina displays.

I think that the original bug report I kind of... odd, it's perhaps a Mac Vs... everything else perspective, but when you're running a fullscreen game, it's the sole thing on most other platforms, it's got 100% of the machine to use.

Perhaps "Fullscreen" needs to have this pseudo-fullscreen setting, so it should be "Fullscreen", "Borderless Maximised" or "Windowed" - with this borderless mode disabling the resolution selection and forcing it to the desktop resolution - instead of just Fullscreen On/Off.

Then everybody would be happy.

@jfacoustic sorry if that complicates the work.

jfacoustic:
I didn't realize that limited it to full resolution.  I'll look into other solutions.  Then again, is this actually an issue?  If on other platforms, it has 100% of the machine to use, then I don't think that you'd want other applications running on Mac either.

impaktor:
> I think that the original bug report is kind of... odd,

I wouldn't mind closing the original issue, if we make the "design decision" that it's not a bug.

fluffyfreak:
Nah the multi-state fullscreen-ness is probably the more user friendly route, some people like to be able to alt-tab away in fullscreen mode. We just need to be aware of the issues around the flags that we're picking.

Other games get around this issue and allow the resolution selection by rendering to an offscreen buffer at the desired resolution, that's not an option for us right now but it could be in the future.

@jfacoustic would you be interested in implementing multi-fullscreen options as I described above? If not then we'll just close and comment in the original bug report what we've found here.

jfacoustic:
@fluffyfreak I'll try to implement those multi fullscreen options.

fluffyfreak:
Closing

PR #3764, opened 2016-08-01
Add function format_duration.

ecraven:
Add a function format_duration that formats a duration in seconds as 1w2d13h28m.
The letters for units are localisable in core (UNIT_...). This is also exported
to Lua as Format.Duration.


ecraven:
This is meant as a preparatory patch for later showing the time remaining until a mission's deadline.

impaktor:
Looking at the code very quickly, it looks good. Assuming you're using correct indentation, github doesn't show that in a hostile way enough for me to see.

impaktor:
....maybe "duration" is too specific? This function would be applicable to any time interval, right?

ecraven:
I've tried hard to insert the correct amount of tabs by hand, fitting the code around it.
I have no issues with changing the name, interval is just as nice.

impaktor:
@fluffyfreak if you give thumbs up, I'll merge by tomorrow, as @ecraven is holding off on other things, to see how difficult we are to accept code :wink: before making more PR.

src/utils.cpp
```diff
@@ -150,6 +150,30 @@ void OpenGLDebugMsg(const char *format, ...)
...
+std::string format_duration(double seconds)
+{
+	std::ostringstream ss;
+	int duration = seconds;
+	int secs = duration % 60;
+	int minutes = (duration / 60) % 60;
+	int hours = (duration / 60 / 60) % 24;
+	int days = (duration / 60 / 60 / 24) % 7;
+	int weeks = (duration / 60 / 60 / 24 / 7);
+	if(weeks != 0)
```
impaktor: you could use `minutes` to compute `hours`, `hours` to compute `days`, etc. I don't care which is which and what is the better approach. Just a though.

fluffyfreak:
Yeah this is fine with me :)

impaktor:
Maybe @walterar might want to make use of this function.

Welcome to the team, I've added you to Authors file. I'm very impressed with the progress you've made on NewUI stuff you're currently working on. (Just letting @fluffyfreak know).

PR #3765, opened 2016-08-02
Make F3/F4 toggle instead of activating the view.

ecraven:
F3/F4 now switch back to the world view, if they are already active.

### DIFF
--- a/src/ShipCpanel.cpp
+++ b/src/ShipCpanel.cpp
@@ -327,6 +327,8 @@ void ShipCpanel::OnChangeInfoView(Gui::MultiStateImageButton *b)
 	Pi::BoinkNoise();
 	if (Pi::GetView() != m_game->GetInfoView())
 		Pi::SetView(m_game->GetInfoView());
+	else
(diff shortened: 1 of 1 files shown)


impaktor:
I like it. Using it I realize I've wanted this a very long time. My only criticism is that the player expects F2 to also be toggable, which it isn't since you need to go F2->F1.

ecraven:
F2 already toggles between the different tabs, so I did not touch that. Maybe later these things should be unified :)

impaktor:
My F2 doesn't toggle, but once in F2 I can F5,F6,F7. And one or two of those toggle tabs. Not sure what would be best. Ping @nozmajner.

ecraven:
F2 > F7 > F2 F2 F2 toggles :)

impaktor:
Sometimes my F-keys aren't working properly. 

But since this only adds features, that will be discovered by players assuming it's there, I'll merge right away. As I said in the first post, I'm one of them, constantly pressing wrong button to get back to Worldview.

PR #3564, opened 2015-12-12
New UI System Info

johnbartholomew:
Very much work-in-progress.

To do:

- [x] Make text translatable
- [ ] Transfer existing translations from 'core' to 'ui-core'
- [ ] Trade info tab, visible only if you have a trade computer, showing imports and exports
- [x] Correctly handle 'unexplored' systems (hide all info)
- [x] Actually fill in the quick info 'Stable system with ...' text.
- [ ] Correctly handle systems containing multiple gravpoints.
- [x] Only use the 'hyperspace target' icon for stars (currently in e.g., Sirius, it gets applied to some planets as well)
- [ ] Auto-select the first body when you first switch to the Body Info tab.
- [ ] Hook up hyperspace target selection action to control the Player's current hyperspace target.
- [x] Make the body icons clickable -- when clicked, that body should be selected and the Body Info tab should show information about it.
- [x] Don't show the 'Notes' heading if the system doesn't have any long description.
- [x] Indicate the selected body somehow. Will probably need to create a new widget type to use to achieve this.
- [ ] Clean up commit history
- [ ] Run luacheck

PR #3766, opened 2016-08-02
Add lua functions for Comms view.

ecraven:
In preparation to move the Comms view into lua, the following are necessary:

- Game.player.AutopilotFlyTo(body)
- Game.player.AutopilotOrbitAt(body, altitude)
- Game.player.AutopilotDockWith(body)
- SystemBody:GetIcon() -> string

### DIFF
--- a/src/LuaPlayer.cpp
+++ b/src/LuaPlayer.cpp
@@ -7,6 +7,8 @@
 #include "Player.h"
 #include "Pi.h"
 #include "Game.h"
+#include "SpaceStation.h"
+#include "Body.h"
 #include "SectorView.h"
 #include "EnumStrings.h"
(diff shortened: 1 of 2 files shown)


impaktor:
Hm, can't you use `ship:AIFlyTo(target)`, and same with the other equivalent functions? Ref: [API](http://eatenbyagrue.org/f/pioneer/codedoc/files/LuaShip-cpp.html#Ship.AIFlyTo)

Also, for GetIcon, check how #3564 does it (maybe you have, I have not), so you're calling the function the same thing, and it works the same way.

ecraven:
Ah, yes, it seems I just did not find these functions, thus I'm not using them :) These already do what I wanted. Great!

PR #3767, opened 2016-08-03
Added navigation button to onClick() and onChat()

robothauler:
This is a second try of #2054 

screenshot-20160803-125904

screenshot-20160803-125929

### DIFF
--- a/data/lang/module-assassination/en.json
+++ b/data/lang/module-assassination/en.json
@@ -191,6 +191,14 @@
       "description" : "",
       "message" : "Return here on the completion of the contract and you will be paid."
    },
+   "SET_AS_TARGET" : {
+      "description" : "",
+      "message" : "Set as navigation target"
+   },
+   "SET_RETURN_ROUTE" : {
+      "description" : "",
+      "message" : "Set return route"
+   },
    "SHIP" : {
       "description" : "",
       "message" : "Ship:"
--- a/data/lang/module-cargorun/en.json
+++ b/data/lang/module-cargorun/en.json
@@ -443,6 +443,14 @@
       "description" : "custom cargo",
       "message" : "Semi-finished products"
    },
+   "SET_AS_TARGET" : {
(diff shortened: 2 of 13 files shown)


walterar:
:+1:

impaktor:
Nice!

Just investigating why the original wasn't merged (bit-rot), and what people had to say about it then:

Brianetta
> Basically, each mission is free to do absolutely anything with this screen. Its onClick function returns a UI object, which can contain absolutely anything at all. The ideal solution to making this look pretty would be for each mission to provide a comprehensive mission-specific interface, rather than trying to resemble the ad.

robn wrote:
> The thought I always had was to have a template that would show a button or some other clicky thing that would take you to the sector map at the appropriate system. We don't quite have the infrastructure to do that yet though... It would be nice to implement this as a custom widget so we don't have to repeat the same code through all the different modules.
 
I'll let this be open over the weekend for others to comment on, if no one else merges it before then.

impaktor:
Actually, this aught to conflict with #3770 for the S&R module. Maybe it's easier to merge this first, then update the other PR with the new files? Or the other way around.

(Also indentation for the new code in S&R-module is spaces on some places? How is it for the other files?)

clausimu:
My original idea was to have the player actually "search" for the target. But I'm overruled. People complain about the tedious manual search and flight. No need to upset players. :) If it removes flavor we can always remove from certain modules. For now it is a great addition to have in general! I'm in favor of merging this. I'd prefer if we could merge #3770 first. But either way is okay.

robothauler:
When the player clicks the button, the target planet is selected as the navigation target, not the target ship. Isn't it?

robothauler:
The lazy player should buy an Advanced Radar Mapper ;-)
Either way of merging this is okay.

walterar:
@robothauler
Or @laarmen / @impaktor can create a NavAssist (0Ton) dependent of autopilot. ;)

PR #3768, opened 2016-08-03
Add a TCP Lua console.

ecraven:
This patch adds a new configure option --enable-remote-lua-repl. If that
is given, When starting a game (not in the menu), pioneer listens to
port 12345 (can be changed with -DREMOTE_LUA_REPL_PORT=...). Connect to
it (telnet 127.0.0.1 12345), and you can evaluate expressions like in
the in-game console. This can be used to interface an external lua IDE
to a running game.

At the moment, the code only works on Linux, but the changes for Windows
should be small (mostly including other header files). This is only relevant if
--enable-remote-lua-repl is given when configuring!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -133,6 +133,7 @@ AC_SUBST([OPTIMISE])
 
 dnl Debug stuff
 AC_ARG_ENABLE([debug], AS_HELP_STRING([--enable-debug], [Enable debugging code (stack checks, debug tools features, etc)]), [EXTRA_CPPFLAGS="$EXTRA_CPPFLAGS -DDEBUG"])
+AC_ARG_ENABLE([remote-lua-repl], AS_HELP_STRING([--enable-remote-lua-repl], [Enable remote lua repl (tcp port 12345 or -DREMOTE_LUA_REPL_PORT=54321]), [EXTRA_CPPFLAGS="$EXTRA_CPPFLAGS -DREMOTE_LUA_REPL"])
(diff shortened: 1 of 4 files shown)


ecraven:
telnet localhost 12345
````
** Welcome to the Pioneer Remote Debugging Console!
> Game = import("Game")

> Game.player:SetMoney(100000)
````

ecraven:
For emacs lua-mode:
````
(defun lua-send-region (start end)
  (interactive "r")
  (setq start (lua-maybe-skip-shebang-line start))
  (let* ((lineno (line-number-at-pos start))
         (lua-file (or (buffer-file-name) (buffer-name)))
         (region-str (buffer-substring-no-properties start end)))
    (lua-send-string (lua-make-lua-string region-str))
    (when lua-always-show (lua-show-process-buffer))))
(setq lua-default-command-switches '("127.0.0.1" "12345"))
(setq lua-default-application "telnet")
````

impaktor:
Most excellent! 

Heads up: I will not be able to test this until next week at the very earliest. Maybe someone else wants to do it before then. @laarmen perhaps? (well, minus the emacs part :smile: )

I assume you'll be using this with your development of Pioneer, so that's testing, I guess.

fluffyfreak:
@impaktor / @laarmen either of you guys able to check this PR out?

laarmen:
The code looked sane enough to me.

impaktor:
I wonder if this perhaps should be documented somewhere? Wiki / "compiling"?

Issue #3694, opened 2016-05-07
Save and land bug - game show error and exit


sfence:
Hello.

I have a mission, from Pluto research base to Sirius. But when I do hyperjump to Sirius system and try to land, game says there is a problem with serializing 'Ship' object. I tried to jump into Sirius system and save game, but same error happend.
I run Pioneer under Linux, from top of git reposity. Last git fetch and merge - today morning (7th May).

My save from Pluto Research base: http://147.229.208.36/pioneer/SIRIUS_LAND_SAVE_BUG

I tried both of target mission bases. No one works.

walterar:
@sfence Do you feel able to replace or modify this function in SearchRescue.lua?
This will solve your problem:

`local serialize = function ()
  for i,ship in pairs(discarded_ships) do
    if not ship or not ship:exists() then
      table.remove(discarded_ships,i)
    end
  end
  return {ads              = ads,
          missions         = missions,
          aircontrol_chars = aircontrol_chars,
          discarded_ships  = discarded_ships}
end
`

sfence:
Fix it primary in main source branch of the game. :)
Because of testing, I want to keep my pioneer source same like source from main branch. :)

laarmen:
Well, you can always modify the file, and use git reset --hard HEAD to go
back to the pristine state.

impaktor:
I was going to suggest `git checkout path/to/changed/file.lua` which is much less brütal. (Usually, `reset --hard` is only for the brave and the bold.)

laarmen:
Well, he/she did say they wanted to keep the experience completely vanilla.
Besides, aren't we all brave and bald^Wbold?

clausimu:
Is this still a problem? I can't replicate it with the current git version (7/25/16). I thought I had solved it and submitted a patch to the module - but so much has happened in "real life" in the past two months that I lost track. Please let me know if I need to look further into it.

walterar:
@clausimu Hello!
This problem is still present while the patch I have suggested are not merged. This is necessary to eliminate the possibility that discarded_ships contain "dead" ships.

clausimu:
@walterar - good catch! Now I understand why you suggested the patch. Logically, I would rather put the fix into the "onShipDestroyed" function though. Any objections to that?

    local onShipDestroyed = function (ship, attacker)

        ...other stuff here.....

        for i, discarded_ship in pairs(discarded_ships) do
[...]

PR #3769, opened 2016-08-04
Search&Rescue module: code style change (indent with tabs instead of spaces)

clausimu:
Cleaned up the code style to use tabs instead of spaces as recommended by @walterar (#3694).


impaktor:
Since this commit is also included in #3770 (altough possibly not on purpose?), I'll close this.

(You probably did #3770 on the same branch as this? Or alternatively, that branch is a derivative of this, instead of your master branch.)

PR #3773, opened 2016-08-10
Fix hyperspace star streaks

fluffyfreak:
I broke the star streaks in hyperspace about 15 days ago and no-one mentioned it?!?


fluffyfreak:
@walterar this fixes the stars in hyperspace rendering issues

PR #3774, opened 2016-08-10
Restore Starfield colourisation

fluffyfreak:
With the change to PointSprites for starfield we lost the varying colours, added these back with @walterar preferred colouring scheme.

I think it looks better :)

### DIFF
--- a/data/configs/Starfield.ini
+++ b/data/configs/Starfield.ini
@@ -1,0 +1,6 @@
+rMin=0.2
+rMax=0.9
+gMin=0.2
+gMax=0.7
+bMin=0.2
+bMax=0.9
--- a/data/shaders/opengl/starfield.frag
+++ b/data/shaders/opengl/starfield.frag
@@ -8,10 +8,12 @@
 uniform sampler2D texture0;
 uniform Material material;
 
+in vec4 v_color;
+
 out vec4 frag_color;
 
 void main(void)
 {
-	frag_color = texture(texture0, gl_PointCoord) * material.emission;
+	frag_color = (texture(texture0, gl_PointCoord) * v_color) * material.emission;
 }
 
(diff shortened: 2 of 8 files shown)


walterar:
He lost some shine, but you can compensate by increasing slightly the value of R and G
With Default.dds monochrome looks ugly; but with the skybox of the Milky Way Hergot it is fantastic.
What do you think of the sky selectable from SkyColor= in config.ini?
Something like this:
if SkyColor
colors [i] = Color (rand.Double (0.5, 0.8) * 255, rand.Double (0.6,0.9) * 255, col, 255);
else
colors [i] = Color 255, 255, 255, 255);

In this way, everyone happy. ;)

fluffyfreak:
Well, I have an idea, I can drive those input from an ini file, and multiply them by the star size, so size will affect brightness, but the colours ranges will come from an ini file.

eg:
> rMin=0.2
rMax=0.9
gMin=0.2
gMax=0.7
bMin=0.2
bMax=0.9

These then get mutiplied by the `size*255`.
`size` itself is still generated as `const double size = rand.Double(0.2,0.9);`.

So you can just edit the file `configs/Starfield.ini` to control the colour selection :)

walterar:
I really like this. :

walterar:
@fluffyfreak  Are you sure you want to Pioneer publish release with stars in colors?

impaktor:
I'm a bit late to the game, but should `Starfield.ini` be capitalized? Looking
in `data/`, I guess it should. 

Also, this adds a `data/configs/` folder, sounds like (maybe) a philosophical change,
or something, as we have *ini files a bit all over the place, like
`data/textures`, `data/ui`. Should they move into that folder? Or one can just
use it for when no other place makes sense?

fluffyfreak:
We've had `data/configs/` for a while now, it stores the `GasGiants.ini` stuff.
I've been using it on an ad-hoc basis, if I create something new, it goes in there, if I work on something existing then I might move it there.

fluffyfreak:
@walterar trying it out, the trouble with asking is that people don't generally get time to check it out, I've pushed it with those colours to see if anyone notices :)

PR #3776, opened 2016-08-12
Disable instancing for city buildings, and enable idle animations.

fluffyfreak:
Disable instancing for city buildings, and enable idle animations.

Does exactly what it says. It seems to be more efficient to submit a few thousand draw calls to the GPU than 14 instanced calls? No idea why but the results are solid, in the worst/best case when viewing an entire city at maximum detail I see the framerate go from 45fps to 90fps.

This also enable's idle animations for City buildings.

EDIT:
One other things this does is batch the drawing of the models to minimise switching between the types of models. So it draws all of one kind, then all of the next. This is what instancing should also do but for some reason it performs poorly.

### DIFF
--- a/src/CityOnPlanet.cpp
+++ b/src/CityOnPlanet.cpp
@@ -192,6 +192,7 @@ void CityOnPlanet::LookupBuildingListModels(citybuildinglist_t *list)
 	for (auto m = models.begin(), itEnd = models.end(); m != itEnd; ++m, i++) {
 		list->buildings[i].instIndex = i;
 		list->buildings[i].resolvedModel = *m;
+		list->buildings[i].idle = (*m)->FindAnimation("idle");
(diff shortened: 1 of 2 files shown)


laarmen:
AMD, Nvidia and Intel alike? And on Linux as well? Or should we test it?

Le ven. 12 août 2016 14:35, Andrew Copland <notifications@github.com> a
écrit :

fluffyfreak:
@laarmen testing would be best, I have a limited range of equipment here.
So far I've tested it on `Windows + nVidia 970` and `Xubuntu + AMD MESA`.
Later on I'll try my laptop which has both an nVidia and Intel onboard.

fluffyfreak:
Actually I have found the real problem!!! CLOSING TIME!

PR #3780, opened 2016-08-15
Add a piece explicitly linking to the FAQ about bug reporting

fluffyfreak:
We still get very poor quality bug reports and always have to ask for the opengl.txt and output.txt files.
So I thought I'd update [the FAQ](http://pioneerwiki.com/wiki/FAQ#How.2Fwhere_do_I_report_my_bug.2Fcrash) a little bit and add this link directly too it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,12 @@ Keyboard and mouse control is found at:
 
 For frequently asked questions, please see
   http://pioneerwiki.com/wiki/FAQ
+  
+  
+## BUG Reporting
+
+Please see the section of the FAQ pertaining to bugs, crashs and reporting other issues.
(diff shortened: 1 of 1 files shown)


impaktor:
To be nit-picky, but the README already has a link to the issue tracker:

> Bugs? Please log an issue: http://pioneerspacesim.net/issues

(So this adds a link to the FAQ entry that adds a link to the issue tracker.)

Other than that, I have no opinions one way or the other. Clarity is a good thing.

fluffyfreak:
Yeah it's just that people obviously aren't looking at it and giving us the detail required, so I also added some more stuff to the specific entry.

PR #3784, opened 2016-08-15
use github templates

impaktor:
Please read, and suggest improvements.

Idea here was that we already have wiki-pages, that are easier to maintain than these templates, thus it links to relevant wiki-pages.

Especially for `contributing` file, this is linked to at the head of a new PR, and thus the info I include is what to consider before pressing "create pull request" button. For _how_ to contribute, we have the wiki, to which also the pioneer home page links.

So, any other "classic" questions to include in issue_template? 

_also, I decided to use lowercase filenames, since these reside in their own folder, and don't need to SHOUT to call attention to them, as README or COMPILING does._

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -1,0 +1,32 @@
+# Contributing
+
+Contributions are welcome!
+
+The following text assumes you're about to submit a new pull request.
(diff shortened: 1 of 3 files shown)


fluffyfreak:
Excellent! :) :+1:

PR #3785, opened 2016-08-15
Search&Rescue module: limit target planet distance

clausimu:
This patch ports a great idea from @walterar and his Scout+ version of the SAR mission script. It limits the target planets of stranded ships to those closest to the station where the request is made. Why request help from Los Angeles for a ship stranded on a moon orbiting Pluto if Pluto Research Base is much closer, right?

I did implement it a little differently though to have less of an impact on ad frequency (there is more effort to find a valid target in my version instead of just dropping the ad if a closer station is found).

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -841,8 +841,6 @@ local isEnabled = function (ref)
 	return true
 end
 
---local nearbysystems
-
 local findNearbyStations = function (vacuum, body)
 	-- Return a list with stations within this system sorted by distance from supplied body (ascending). If vacuum is set to true
 	-- then only return orbital stations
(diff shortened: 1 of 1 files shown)


impaktor:
> It limits the target planets of stranded ships to those closest to the station where the request is made. Why request help from Los Angeles for a ship stranded on a moon orbiting Pluto if Pluto Research Base is much closer, right?

Another way of doing it is to allow the same adverts to be spawned on all BBSes, system wide. But if they're station specific, then sure.

clausimu:
Good idea to have all BBSes get the same emergency alert. But then you run into problems of information sharing between stations. If someone accepted the mission on Pluto Research Base to help someone on one of Pluto's moons it would take a significant amount of time until that information is shared with other stations (via courier?). So in theory you can never be sure if an emergency call is still open - unless it is station specific.

But I'll keep system-wide ads in mind for future missions. Might make for an interesting story.

nozmajner:
Via couriers? In system they could radio it, only about 5 hours signal delay From Earth to Pluto for example.

clausimu:
@nozmajner - fair enough. :) But even with a few hours delay I still have to account for possible double-booking of missions. Not saying it is necessarily bad but something I will have to work into a real story. For now I would like to leave ads station specific.

walterar:
I see nothing wrong in this PR. :+1: 

For future and endless discussions I suggest watching: https://en.wikipedia.org/wiki/Emergency_position-indicating_radiobeacon_station

fluffyfreak:
Ignoring the alternative gameplay considerations, I think this PR is fine and can be merged.

impaktor:
:+1:

PR #3789, opened 2016-08-18
Navigation button update to allow for dynamicBodies as nav targets (SAR and general)

clausimu:
This PR does three things:

1) It updates NavButton.lua to allow setting dynamic objects (such as ships) as nav targets.

2) It allows setting Search&Rescue target ships as nav targets directly either from the BBS ad or from the mission details screen. Before, only the closest planet was selected as nav target.

3) It moves the navbutton in the BBS ad to after mission acceptance, because only then will target ships actually be created.

I know, this changes the original philosophy of the Search&Rescue missions by basically eliminating the "search". But opposed to what I had thought, searching was never really any fun. Especially for out-of-system missions, where one had to go into the F2=>F6 system overview to find the ship (which was usually placed wrongly right on top of the sun of that system), it caused me to lose immersion into the game.

### DIFF
--- a/data/libs/ui/NavButton.lua
+++ b/data/libs/ui/NavButton.lua
@@ -18,7 +18,7 @@ function NavButton.New (text, target)
 	self.widget = ui:HBox(10):PackEnd({ self.button, self.label })
 
 	self.widget.onClick:Connect(function ()
-			if self.target:IsDynamic() then
+			if self.target:isa("Body") and self.target:IsDynamic() then
 				Game.player:SetNavTarget(self.target)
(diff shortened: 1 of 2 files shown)


fluffyfreak:
If it's more enjoyable then I'm all for it.

clausimu:
As @impaktor said in the corresponding dev-forum post - having a real search for stranded ships would be great. But it will take a bit more work to set this up right. Until then I prefer not to annoy players ;).

impaktor:
@clausimu let us know if you're activley waiting for this to be merged. We're lazy.

_That whcih can be done now, can be done tomorrow._ :sleeping:

Issue #3779, opened 2016-08-15
GRAPHICS


BILL2517:
The cursor bumps against but will not hover over the opening menu. Therefore, the user cannot do anything. Perhaps a resolution setting? BUT, the settings are not available. Game always starts with the same resolution. Cannot get to the settings. Why does it remember the settings? Cannot start new.
There is no "readme" or email address for problem reporting.

impaktor:
Sounds strange.

> There is no "readme" ...

There's a wiki, with FAQ. Please see note there on location of config file. Try another resolution by manually editing it, and see if that resolves the problem.

> or email address for problem reporting.

This is the place, you're doing it right now.

impaktor:
[README](https://github.com/pioneerspacesim/pioneer/blob/master/README.md)

impaktor:
@BILL2517  Please let us know if you've solved this.

fluffyfreak:
@BILL2517 also please provide some more detail, what operating system are you running the game on?
If Linux then what distribution and Window Manager.
Also what GPU are you using?

BILL2517:
Win 8.1; I7-4770; GT640 (HP Envy Phoenix 810-160)
I am unable to change the resolution via config-resolution may be the problem.
ESC to config doesn't work until in the game...can't get there.
Updated video driver-no change
Initially, I could move over the menu items. I selected a resolution in setup. Now, can't hover over the menu. It seems to remember the bad resolutiuon even after removing the game and setting it up again.
Is there something hidden...a registry item?

fluffyfreak:
No registry just the contents of the `config.ini` file, try deleting either the two lines:
`ScrHeight=960
ScrWidth=1280`
When it loads it will default to 800x600.

Or the entire `config.ini` file, it will recreate it when the game starts.

BILL2517:
Thanks for the answer….there is no CONFIG.INI file

I created an empty one, which did nothing.

The program I downloaded is: "pioneer-20160814-win32"

 

Bill

 

"Be happy in your work."
- <http://www.imdb.com/name/nm0370564/?ref_=tt_trv_qu> Colonel Saito 

Sent from my toy computer

impaktor:
> there is no CONFIG.INI file

[...]

PR #3790, opened 2016-08-19
Add button to Settings menus so user can open their Pioneer folder.

fluffyfreak:
After numerous situations where the user cannot find their `Pioneer` folder with `config.ini`, `output.txt` and `opengl.txt` files in it I realised that of course we already have that information.

This adds a button to the settings screen that will ask the system to open the folder directly so they do not have to search for it.

I've only implemented the Windows call necessary as that's all I have access too right now.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -795,6 +795,10 @@
       "description" : "",
       "message" : "Ok"
    },
+   "OPEN_USER_FOLDER" : {
+      "description" : "",
+      "message" : "User Folder"
+   },
    "OPTIONS" : {
       "description" : "",
       "message" : "Options"
--- a/data/ui/MainMenu.lua
+++ b/data/ui/MainMenu.lua
@@ -77,7 +77,8 @@ local doSettingsScreen = function()
 	ui.layer:SetInnerWidget(
 		ui.templates.Settings({
 			closeButtons = {
-				{ text = l.RETURN_TO_MENU, onClick = function () ui.layer:SetInnerWidget(ui.templates.MainMenu()) end }
+				{ text = l.RETURN_TO_MENU, onClick = function () ui.layer:SetInnerWidget(ui.templates.MainMenu()) end },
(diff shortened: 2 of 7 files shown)


laarmen:
That's a really good idea!

impaktor:
I assume you're thinking of #3779, where in the first answer the full description how to find the config file, and solve the problem is given, assuming they can find our wiki without us providing the link.

Turned out 11 more posts were needed to reach there.

So, yes, this is probably needed.

Apparently, I'm in a bad mood today.

fluffyfreak:
Not just of #3779 but yes it's what inspired me finally :)

Also now that's it there I'm finding it quite useful!

fluffyfreak:
@laarmen / @impaktor either of you guys want to try implementing a generic Linux version?

laarmen:
Note to future self: xdg-open with inode as mime type should do the trick.

laarmen:
To be honest, I'm not entirely sure it would be useful on Linux. The platform fragmentation will probably make the code not very robust/clean, and our users tend to compile from source, which means they probably have the reflex to look for a $HOME/.pioneer dir.

fluffyfreak:
More for completeness then, Windows will have it and OSX should be pretty easy too.
Is there no simple way of doing it for all Linux distros/WMs?

johnbartholomew:
Drive-by comment:
There's much more variation in how Linux users examine and navigate the file system than on Windows. There might also be more variation in the text editors that would be used to edit config.ini.

But if you just want something for the button to do on Linux, perhaps it could just show a dialog/message box that gives the path to the config directory. (For a message box you could show one in-game, or use SDL_ShowMessageBox or SDL_ShowSimpleMessageBox)

laarmen:
Eh. On my netbook (which actually runs Pioneer, albeit painstakingly slow), I don't have any way to open a directory: I don't have any file explorer installed.

fluffyfreak:
@johnbartholomew thanks :) that's a good idea actually.

fluffyfreak:
Added @johnbartholomew suggestion using a simple message box.
Will merge this once it's passed the check unless there's objections?

PR #3791, opened 2016-08-19
Fix crash when jumping into empty system

### LINKED ISSUE #3788: Reckless jump into non-inhabited system causes crash
### Observed behaviour
Doing a reckless jump from Los Angeles into an uninhabited system. After reaching the system the game crashed with: 

error: [string "[T] @modules/CrimeTracking/CrimeTracking.lua"]:27: attempt to index local 'station' (a nil value)
stack traceback:
        [string "[T] @modules/CrimeTracking/CrimeTracking.lua"]:27: in function <[string "[T] @modules/CrimeTracking/CrimeTracking.lua"]:18>

### Steps to reproduce
Do an illegal jump into a non-inhabited system.

pioneer: git (3c331d0)
openSUSE 42.1 64bit

impaktor:
The crash happens because apparently empty
systems can have a lawlessness > 0
Closes #3788


fluffyfreak:
Looks good to me

PR #3792, opened 2016-08-20
Update GasGiant config, detail and initialisation timing

fluffyfreak:
This varies when the Gas Giant GPU jobs are submitted to prevent them all being requested at once.
The intention is to avoid a large chug happening as all of the 6 cubemap faces are generated for every single Gas Giant in a star system.

I've also taken the opportunity to add different sized texture generation depending on the requested planet detail level, and to control the number of octaves from the config file.

The size of textures used on CPU and GPU can be controlled independently per-detail level (**5**) as can the number of octaves of noise used in the GPU generation.

This should generally improve the performance, but degrade the appearance, on lower detail settings.

### DIFF
--- a/data/configs/GasGiants.ini
+++ b/data/configs/GasGiants.ini
@@ -1,7 +1,17 @@
 texture_size_small=16
 #CPU settings, texture size (limited to 4096) and delay before generating textures when game starts
-texture_size_cpu=512
+texture_size_cpu_very_low=64
+texture_size_cpu_low=64
+texture_size_cpu_medium=256
+texture_size_cpu_high=512
+texture_size_cpu_very_high=1024
 cpu_delay_time=60.0
 #GPU settings, texture size (limited to 4096) and delay before generating textures when game starts
-texture_size_gpu=1024
+texture_size_gpu_very_low=128
+texture_size_gpu_low=128
+texture_size_gpu_medium=512
+texture_size_gpu_high=512
+texture_size_gpu_very_high=1024
 gpu_delay_time=5.0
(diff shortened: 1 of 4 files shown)


Tichy:
Will this prevent the necessity to enable AMD_MESA_HACKS?

fluffyfreak:
@Tichy no that'll still be required. It's a catch-all flag for dealing with several bugs.

impaktor:
I tested it. No crash, and I didn't notice the profound system freeze as I used to, so it seems to be working.

Compiler warning:

```
GasGiantJobs.cpp: In member function ‘void GasGiantJobs::SGPUGenRequest::SetupMaterialParams(int)’:
GasGiantJobs.cpp:181:66: warning: use of old-style cast [-Wold-style-cast]
    m_specialParams.frequency[i] = (float)pTerrain->GetFracDef(i).frequency;
                                                                  ^
In file included from GasGiant.cpp:5:0:
GasGiant.h: In constructor ‘GasGiant::GasGiant(const SystemBody*)’:
GasGiant.h:88:8: warning: ‘GasGiant::m_timeDelay’ will be initialized after [-Wreorder]
  float m_timeDelay;
        ^
GasGiant.h:86:7: warning:   ‘bool GasGiant::m_hasGpuJobRequest’ [-Wreorder]
  bool m_hasGpuJobRequest;
       ^
GasGiant.cpp:296:1: warning:   when initialized here [-Wreorder]
 GasGiant::GasGiant(const SystemBody *body) : BaseSphere(body),
 ^
```

fluffyfreak:
@impaktor thanks, I think I've fixed those warnings/reordering issues now too, I'll check the Travis build log when it's done

Issue #3794, opened 2016-08-23
Segmentation fault after AI ship is set to hyperjump


clausimu:
### Observed behaviour
I hope it is not something stupid I did. But the crash seems to happen on the c++ side.

AI controlled ship is set to hyperjump (ship:HyperjumpTo(system-whatever.path)) out of a non-inhabited system. The corresponding functions that are triggered in Ship.lua and Equipment.lua run fine (as tested by inserting print statements everywhere). Engine present, enough fuel present, hyperjump status is "OK". But after the scheduled warmup time the game crashes with a segmentation fault. Ship never moved.

gdb backtrace:
```
#0  Frame::GetInterpPositionRelTo (this=0x0, relTo=relTo@entry=0x112fd130) at Frame.cpp:182
#1  0x000000000041ef8d in Body::GetInterpPositionRelTo (relTo=0x112fd130, this=0x11eec140) at Body.cpp:142
#2  Body::GetInterpPositionRelTo (this=this@entry=0x11eec140, relTo=<optimized out>) at Body.cpp:154
#3  0x0000000000593237 in WorldView::UpdateProjectedObjects (this=this@entry=0x115c2260) at WorldView.cpp:1769
#4  0x00000000005945a6 in WorldView::Update (this=0x115c2260) at WorldView.cpp:1202
#5  0x00000000004e3f19 in Pi::MainLoop () at Pi.cpp:1357
#6  0x00000000004e4dca in Pi::Start () at Pi.cpp:1222
#7  0x0000000000412765 in main (argc=<optimized out>, argv=<optimized out>) at main.cpp:130
```

### Expected behaviour
Ship should hyperjump.

### Steps to reproduce
I can provide a save game.

__My pioneer version (and OS):__
pioneer version: git (3c331d0)
OS: openSUSE 42.1 64bit

clausimu:
Because @impaktor had asked in irc: removing the CrimeTracking module has no impact.

laarmen:
I'll take that savegame :-)

clausimu:
Savegame here: 

https://dl.dropboxusercontent.com/u/35453311/Pioneer/sar_test_26

To reproduce:

1) move within 49 m of the deep space miner just ahead
2) wait until all three crew members have been transferred to the deep space miner (as shown by the comms)
3) the deep space miner will then attempt to jump (but won't budge)
4) if you wait for 25 sec after the crew transfer the game will crash

walterar:
[...]

PR #3795, opened 2016-08-23
Search&Rescue: bug fixes and more realistic ship simulation

clausimu:
This was supposed to be a simple PR that fixes the crew delivery mission in the Search & Rescue script - but one thing lead to another, depended on another - and here we are. The switch to more realistic ship simulation caused many of the downstream changes. There are more fixes/improvements coming, but I can make a cut here to limit PR size.

What this PR does:

- Simulate AI ships more realistically, with appropriate min/max crew, fuel to fly and hyperjump, engine selection, etc.
- Removed low-level hyperjump calls (ship:InitiateHyperjumpTo).
- Fuel requested for "out-of-fuel" missions is now variable (instead of fixed to 1 ton) and dependent on the target ship fuel tank size, engine, location and future nav-target.
- Crew requested for "missing crew" missions is variable based on crew present (variable too now) and dependent on min-Crew/max-Crew per target ship.
- Maximum of crew and passengers to be picked up or delivered increased from 4 & 2 to 8 & 20.
- For personnel pickup missions there is now a 50:50 chance that it will include passengers in addition to crew. If passenger transport is requested, the number of passengers is dependent on the target ship size (maximum = 20).
- Police ships are excluded as target ships.
- More effort is made to find a suitable target ship type during mission creation to limit the reduction of ad creation due to not finding a possible target ship that fit the mission criteria.
- If the target ship cannot hyp away after mission completion it is now discarded via explosion after flying in high orbit around planet or sun for 10 minutes.

I added "WIP" to the title because bug #3794 may or may not depend on this SAR script. Once the cause for that bug is found this can be merged (unless there are objections, of course).


clausimu:
This can be merged now - as far as I am concerned.

walterar:
@clausimu Look to the future:

"We request the delivery of {amount_of} of {cargo} to the ship."

In this way you will not have to change in the future this line of language.

{cargo}, now = hydrogen
{cargo}, future = liquor, hyperdrive, narcotics, military fuel, missiles, etc.

clausimu:
@walterar - point taken. I like the subtle different flavors in language though for the different mission types in the SAR script. Because others have already done the work and translated the strings I have gone ahead and just replaced the specific "hydrogen fuel" with the more general "{cargo}" per string instead of using the same string altogether for each mission type. Under the hood the correct cargo type is now selected to make future extensions easier.

Since this is a significant update already, I'd like to merge it as is without further additions. There is more coming (for example adjusting reward based on cargo amount delivered, etc.) in following PRs once this one is in. I can do further additions in these smaller PRs.

walterar:
@clausimu Yes, it is better to merge as is. No time for other things. Geraldine is waiting to publish Pioneer (August) in ModDB ;))

laarmen:
Allright, let's merge this bugger.

PR #3796, opened 2016-08-24
Fix nav and combat target removal cleaning

laarmen:
There is nothing preventing the navigation target to be the same as the combat target. Whether that is a good thing or not is not the goal of this patch, which is just to take this possibility into account when the target body is indeed removed.

My apologies for comitting on the master repo, but I don't have my SSH key on this computer and GH didn't let me do that on my (outdated) repo.


laarmen:
That fixes #3794

walterar:
:+1:

Issue #3797, opened 2016-08-24
Assertion failure path.bodyIndex < m_bodies.size()


yump:
### Observed behaviour

After loading [this save file](https://u.pomf.is/wgoabz), taking off and landing again results in a crash with the error message:

    pioneer: StarSystem.cpp:554: SystemBody* StarSystem::GetBodyByPath(const SystemPath&) const: 

[Full log file](http://pastebin.com/HUr0YgRY).

This error is similar to the one from #3502, but it occurs in a different function.

### Expected behaviour

Crash does not happen.

### Steps to reproduce

1. Load provided save file.

2. Unpause game if necessary.

3. Hit launch.

4. Use the autopilot to request clearance.

5. Use the autopilot to land.

__My pioneer version (and OS):__ 20160814-linux64, Fedora 24

walterar:
@clausimu 

 This fix it
```
--- /orig/SearchRescue.lua
+++ /mod/SearchRescue.lua

 local onShipDocked = function (ship, station)
 	if ship:IsPlayer() then
 		for _,mission in pairs(missions) do
-			if Space.GetBody(mission.station_local.bodyIndex) == station then
+			if Game.time > mission.due or Space.GetBody(mission.station_local.bodyIndex) == station then
 				closeMission(mission)
 			end
 		end
```

clausimu:
Good catch, @walterar. Thank you! I actually found another problem with this line - fixed in #3799.

laarmen:
Still, the game shouldn't assert but lua_error out or at least return nil
when the body index is not valid.

clausimu:
@laarmen - I will have to leave that to someone smarter than me to fix. :) I would prefer a lua error myself. But if I remember correctly, whenever trying to index a non-existent body the game would always just crash with pretty much the same error as posted here.

walterar:
@laarmen 

> or at least return nil
> when the body index is not valid.

#3635

laarmen:
Not relevant. Trying to guess what you meant by that, nil is a perfectly good value to return when there's, well, no object to return on a getter. Of course it won't have the method "exists", since it is nil, but I'm skeptical as to the method's utility in the first place.

walterar:
#3799 fixes this correctly.

PR #3799, opened 2016-08-25
Search&Rescue: bugfix for #3797

clausimu:
Bug fix for #3797.

There were actually two things wrong right there (one pointed out by @walterar):

1) Missions were never closed if they were ever overdue until the player landed at the station where the original mission was accepted. Kind of a style choice to now terminate them wherever the player lands. But might be nicer to not clutter the mission screen with failed missions? What if the player never intends to return to the original station? Can't make them suffer forever... :)

2) If the player landed at a station in a different system while a SAR mission was active in the background the game could crash.


walterar:
This fixes #3797 :+1:

PR #3806, opened 2016-09-08
Search&Rescue: adjust mission income

clausimu:
Mission income for the Search & Rescue missions has been adjusted to account for the recently increased variability in commodity/personnel delivery and pickup amounts. The reward is a combination of:

1) a flat fee based on general mission distance (same planet, same system, different system) with some random variability

plus

2) a fee based on necessary cabins (half the current cabin price times number of cabins needed) or commodity (ten times the commodity price).

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -525,7 +525,7 @@ end
 -- extended mission functions
 -- ==========================
 
-local calcReward = function (flavour)
+local calcReward = function (flavour, pickup_crew, pickup_pass, pickup_comm, deliver_crew, deliver_pass, deliver_comm)
 	-- Calculate the appropriate reward for this mission.
 	-- TODO: Extend for other mission types.
 	-- TODO: Adjust based on urgency + risk?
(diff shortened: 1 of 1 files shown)


walterar:
This is what I call continuous improvement. ;)
:+1:

impaktor:
Not sure I like the commit history here. Looks like debug code was added in the "resolve conflict" and then removed in another commit. And the "resolve commit" does a lot. Was there a conflict? It's only you working on the code, right?

Also, the "recoup debug version" commit isn't the one that has the debug code in it, as noted above.

I'd opt for squashing everything to one commit, unless there's something more here that should be tracked? No need to track how debug code is added and removed. While useful when working on the branch, it can be removed once branch is done. 

I wrote a guide on how to fix these things on the git page on our wiki. Also, you know wher to find me if you need help.

Sorry for being pedantic with commit history. But I like it tidy.

laarmen:
impaktor+1 on this one.

clausimu:
More tidy this time... :)

impaktor:
Aren't there fewer changes in it this time?
(I'm just making sure it's doing the same/right thing as before the rebase.)

(In addition to the fix I proposed last night on IRC, you could of course just have drop in the finished file in the new branch, rather than cherry pick each commit, and squash them).

clausimu:
Good catch, impaktor. I missed three little changes I had made in the "messy" branch. I have amended them to the only commit in this branch and tested everything out - it works as intended now.

impaktor:
becoming a git pro now.

laarmen:
Hell yeah!

PR #3808, opened 2016-09-11
WIP - Opengl 2.1 renderer support

fluffyfreak:
# Opengl 2.1

An alternative Opengl 2.1 render in place of the standard Opengl 3.x (_'x' because of extensions etc_) for the following reasons:
- compatibility with older hardware
- improved separation between renderer & engine
- ...I saw that 'jumpdrive' did it and I wanted it...
# Missing bits

I couldn't get all of the Opengl 3.x stuff to work exactly the same in Opengl 2.1 so there are visual differences and some performance differences like:
- No GPU gas giant texture generation
- No Instanced rendering
- Starfield looks strange as points don't appear to resize correctly
- No Ansiotropic rendering
- Probably other bugs / visual issues
# Nasty bits

The program [glLoadGen](https://bitbucket.org/alfonse/glloadgen/wiki/Home)(_ver 2.0.5_) was used to generate the header files used but at the top of each of these it defines a list of identifiers to prevent other Opengl loaders from being used/mixed in.
This is mostly fine except that I want to include 2 instances of these headers, for gl 3.x and 2.1 each.
So at the top of 'Pi.cpp' and 'ModelViewer.cpp' I must first '#undef' those symbols.
I wonder if it would instead be acceptable to just remove them from the generated '.hpp' files instead?

Also of note is how long this has taken in fits and spurts, some of this branch is from 'November 9th 2015'. No doubt some of the rendering changes will have errors and problems.
# Summary

This should allow the game to run lower end GPU's in laptops that cannot be upgraded.
Although due to the lack of instanced rendering in the 2.1 branch it will perform worse around cities.


fluffyfreak:
@robn @laarmen @impaktor this fails to build because it's still trying to build `gl_core_3_x.h` which has been replace by `gl_core_3_x.hpp`. I found that it's because it didn't get-rid-of / replace something in then `/graphics/opengl/.deps` folder.

Otherwise it actually builds fine. So there something going on that I cannot fathom out with the bootstrap/configure process. Any chance one of you could take a look and tell me what I'm doing wrong?

walterar:
this should be here?

src/graphics/opengl/Makefile.am

noinst_HEADERS = \
	gl_core_3_x.h \

fluffyfreak:
@walterar thanks! Turns out that I had local, uncommitted changes on my Linux build machine. For some reason I'd missed them...

walterar:
For me, compiles and works fine now, but it's better to try it someone with a machine that does not support GL3.

image

fluffyfreak:
@walterar I had dv- test it on his laptop with X1300 GPU (opengl 2.1) and it worked ok but it would be good if other could give it a try too.

@impaktor are you able to test this on your laptop with an Opengl 2.1 GPU? Also I've been trying to test the Raspberry Pi Opengl 2.1 support :) but sadly it has problems with my installation :/ it does apparently work better on the RPi2 so it might be worth a try?

impaktor:
Compiled, removed config, start pioneer, I get:
```
SDL Version 2.0.4
Failed to set video mode. (Couldn't find matching GLX visual). Re-trying without multisampling.
Failed to set video mode. (Could not create GL context: GLXBadFBConfig). Re-trying with 16-bit depth buffer
Failed to set video mode. (Couldn't find matching GLX visual). Re-trying with 16-bit depth buffer and no multisampling
error: Failed to set video mode: Could not create GL context: GLXBadFBConfig
```

impaktor:
Before removing config, I tried putting `RendererName=Opengl 2.1` in it, to no effect.

Interestingly, pioneer does not re-create the removed config. Instead, there is no config file at all, after having removed it.

impaktor:
### Conclusion 
 Oops, I only have OpenGL 1.4, apparently. No need to do any extra work on my account.

So I guess this is good to merge?

walterar:
With MXE compiles fine but has some CTD when running, apparently FaceGen. I will review best.

fluffyfreak:
@walterar thanks for testing it, I'm not seeing a CTD or any problems when using FaceGen or anywhere else.

walterar:
@fluffyfreak False alarm. I'm migrating Ubuntu 12.04 to 14.04 and had a problem folders.  No problem so far.

fluffyfreak:
ok thanks for letting me know :)

impaktor:
Do we know why it failed to auto-configure itself on @dvolk's machine (`dv-` on IRC yesterday)? I suspect many/most? players who need this will throw in the towel if it doesn't work out of the box. I'm guesstimating maybe less than 50% will actually start digging through the wiki, FAQ, read forum, read issue tracker, sign on to IRC, or create forum account to get it to work.

My message here is that if the auto-configure doesn't work, then that would be a shame.

Gone are the days when people/kids would bang their heads against the computer for hours/days/weeks if they (we!) got stuck in a game.

Damn spoiled Internet-kids. GET OF MY LAWN!

fluffyfreak:
@impaktor you're right but I don't know what the problem is and don't have a GL2.1 GPU to test with :(
It might be to do with the section in `WindowSDL.cpp`

`// cannot initialise 3.x content on OSX with anything but CORE profile`
`SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);`
`// OSX also forces us to use this for 3.2 onwards`
`SDL_GL_SetAttribute(SDL_GL_CONTEXT_FLAGS, SDL_GL_CONTEXT_FORWARD_COMPATIBLE_FLAG);`

Those flags are really only for OpenGL 3.0+.

I also found this note on the [SDL wiki](https://wiki.libsdl.org/SDL_GLattr):
> SDL_GL_CONTEXT_PROFILE_MASK determines the type of context created, while both SDL_GL_CONTEXT_MAJOR_VERSION and SDL_GL_CONTEXT_MINOR_VERSION determine which version. All three attributes must be set prior to creating the first window, and in general you can't change the value of SDL_GL_CONTEXT_PROFILE_MASK without first destroying all windows created with the previous setting.

Which suggests that there might not be a graceful fallback without some code wrangling to destroy the window first.

impaktor:
Could the error message suggest editing the config file accordingly?

walterar:
@fluffyfreak I do not have NavLight with GL2.1, after applying 3809

fluffyfreak:
@walterar ah the joys of having multiple rendering engines :/
I know what this is, not 100% sure how to fix it right now but I'll think of something.

fluffyfreak:
I'm going to think about this overnight but, there's an increasing number of things that just don't work unless you have some extensions available... and those extensions are almost all things that came in with OpenGL 3.0/1/2 etc.

So basically the only way to make OpenGL 2.1 work, is to use extensions that only work on OpenGL 3.x graphics cards... and if they work, then you can probably run OpenGL 3.x anyway so all of this becomes a bit pointless.

walterar:
I think it's worth working on this. I wish I could help more :(

impaktor:
I've herd scepticism regarding merging this (from its author), here's my 2 cents:

1. Does not his PR do refactoring and separation of rendering routines that is a step in the general right direction?

2. If this doesn't end up in master it would be nice if a once-only version of this got up on sourceforge, since the original 2014-11-18 build got removed. Then we can link to that download from the FAQ, so people with old computers at least get a pioneer build once. And obviously, opengl2.1 renderer need not support any fancy stuff, just not crash. Too often.

fluffyfreak:
1. Not that much and it introduces a few negatives as well for other parts of the code.

2. That'd be one idea, I could (_in theory_) do an OpenGL2.1 **only** build using this renderer which kept some of the nastier changes out of the code base.

The more I think about it the less I like this PR, and I wrote it! 
Time would be better spent elsewhere and it introduces another support burden for me.

HybridDog:
lf you want to test it with opengl 2.1 l think you could temporarily use mesa llvmpipe.

l'm using the cutting edge radeon driver but l don't get the newest opengl version. Does the version string indicate which extensions are available or is it rather just for decoration?

PR #3809, opened 2016-09-14
Made NavLights responsible for their billboard rendering

fluffyfreak:
I've made some odd late-night programming decisions in the past but I don't know what made me put the responsibility for rendering NavLight billboards into the Scenegraph::Model class.

This fixes that questionable decision and moves the responsibility for rendering NavLight billboards from scenegraph/Model into NavLights itself.


impaktor:
Hmm, I can not edit my review above. So no more sloppy-write-> post-> edit-loop for me.

fluffyfreak:
Yeah we'll have to see how the new review system works out.

PR #3811, opened 2016-09-19
Set ship texture patterns from lua

clausimu:
This allows ship texture patterns to be set from lua. Together with the already present "SetSkin" method you can now recreate a specific look for a ship. This is important for persistence in scripts. For example, in Search and Rescue missions, if you leave a system and come back, the target ship will have a different look because the skin and pattern are created randomly when creating a ship. You can now specify them.

C++ experts like @laarmen, @impaktor and @fluffyfreak should take a close look at the additions I made here. I am a total C++ noob and may have easily made mistakes.

While I was at it I also annotated the SetSkin and GetSkin methods. I only stumbled upon them while digging through code looking for other things.

### DIFF
--- a/src/LuaShip.cpp
+++ b/src/LuaShip.cpp
@@ -304,6 +304,10 @@ static int l_ship_set_pattern(lua_State *l)
 {
 	Ship *s = LuaObject<Ship>::CheckFromLua(1);
 	unsigned int num = lua_tointeger(l, 2);
+
+	if (num > s->GetModel()->GetNumPatterns()-1)
+		return luaL_error(l, "This pattern does not exist for this ship");
+
 	s->SetPattern(num);
 	return 0;
 }
--- a/src/LuaShip.cpp
+++ b/src/LuaShip.cpp
@@ -227,6 +227,25 @@ static int l_ship_explode(lua_State *l)
 	return 0;
 }
 
+/*
+ * Method: GetSkin
+ *
+ * Get the current skin object of the ship.
+ *
+ * > ship:GetSkin()
+ *
+ * Parameters:
+ *
+ *
+ *
+ * Example:
+ *
+ * > ship:GetSkin()
+ *
+ * Status:
+ *
+ *  experimental
+ */
 static int l_ship_get_skin(lua_State *l)
 {
 	Ship *s = LuaObject<Ship>::CheckFromLua(1);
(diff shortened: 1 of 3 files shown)


src/LuaShip.cpp
```diff
@@ -243,6 +282,33 @@ static int l_ship_set_skin(lua_State *l)
...
+ * > ship:SetPattern(5)
+ *
+ * Status:
+ *
+ *  experimental
+ */
+static int l_ship_set_pattern(lua_State *l)
+{
+	Ship *s = LuaObject<Ship>::CheckFromLua(1);
+	unsigned int num = lua_tointeger(l, 2);
```
laarmen: Would it be possible to do some parameter sanitization here, and call lua_error if the pattern is invalid? Pretty please!

Note that I'd settle for return false on failure, if it were a common pattern in our API. Sadly, it isn't just yet.

clausimu: @laarmen - is this what you had in mind (see new commit)?

clausimu: @laarmen - okay, it takes into consideration that indexing starts at 0 now (amended my last commit).

Along these lines - wouldn't the following line in Ship.cpp need fixing then too?:
`GetModel()->SetPattern(Pi::rng.Int32(0, GetModel()->GetNumPatterns()));`

Or am I showing my C++ ignorance here?

laarmen:
I wouldn't bet my life on it, but quite often an API asking for an integer
range excludes the maximum value from the range. Especially in C-like
languages.

clausimu:
Okay - thanks.

fluffyfreak:
@clausimu I think in the `ship.cpp` case it does actually fail! but the code handles out-of-bounds values by returning without changing anything so no-ones noticed before :)

src/LuaShip.cpp
laarmen: After checking, the Int32 is inclusive, which means there's a bug there, indeed.

fluffyfreak:
@laarmen I think @clausimu latest changes are valid for the checking etc so I'd like to merge this. Are you happy with it?

Also out of curiosity can you change your review? Or is it a set once sort of thing? Which would be odd.

laarmen:
Looks like I can. :-)

fluffyfreak:
Cool :)

PR #3813, opened 2016-09-20
Additional mesh optimisation

fluffyfreak:
Some additional optimisations applied upon loading using Assimp.

smallerer

It makes the total size of `.sgm` files about 21MiB smaller.
It should reduce the sheer number of vertices along with potential other changes.
I have tested it with stations & taken a look at a number of ships but I think that @nozmajner might want to take a long look at the models and make sure that nothing obvious has been mangled.


nozmajner:
Reducing the number of vertices procedurally/algorithmically? That makes me a bit nervous :D

fluffyfreak:
@nozmajner bit different. 
When it loads it actually unpacks the shared vertices and duplicates them, this flag tells it not to do that and only merges ones that are identical.

nozmajner:
Ah, I see.

PR #3814, opened 2016-09-21
Model pattern range and status checking

fluffyfreak:
This PR does two things:
 - Check that a model actually supports patterns before using them,
 - Limit the range correctly by subtracting 1 from `GetNumPatterns()`

Thanks to @clausimu for discovering that this was wrong!


fluffyfreak:
@laarmen *that guy* eh! :)

Nah it's fine, I'll add that in another mini-PR, @clausimu is already limiting it too a safe range and the `SupportsPatterns()` check is really just a belts-and-braces check. I had forgotten that the method even existed.

PR #3815, opened 2016-09-21
Brighter luminosity for random colours

fluffyfreak:
A frequent comment since the random colour generation was changed is that the colours seem darker and less appealing.

The third parameter of `RandomColor::GetColors` controls the luminance and so I have set this to `LUMINOSITY_BRIGHT` instead of `LUMINOSITY_RANDOM`.

Black can still appear as a "colour" and is obviously not enhanced by this.


impaktor:
Can this be tested in modelcompiler or something? I'm just checking if we know if the end result still looks sane.

fluffyfreak:
Yep just open ModelViewer mode, pick an object and click the "random colour" button a bunch of times :)

PR #3818, opened 2016-09-23
New seed and new locations of star ports for New Hope in Epsilon Eridani

fluffyfreak:
New seed with more varied terrain and repositioned surface stations so that, for example: Itzalean, is where we want it to be.


impaktor:
It's a long time since I played, but will this effect the eclipse 2 days into the game?

fluffyfreak:
That's what I was hoping to restore, you don't get it at the poles!

PR #3819, opened 2016-09-23
RFC/WIP new HUD with ImGUI

ecraven:
**SUPERSEDED by #3826 (the first of multiple PRs that should finish with pioneer looking more or less like the images below)**

This WIP PR moves almost the entire HUD into lua. It adds a ton of new functions to the lua interface.
Lots of C++ rendering code is removed. It uses Dear ImGUI for the UI (C++, exported into Lua).

It is not meant to be merged as-is, but needs more discussion.

docked
orbit
navtargets
radial-menu

I'd appreciate any comments concerning:
- changes in the C++ code
- changes in the Lua code
- the actual look of the UI


impaktor:
YAY! Great!

Some (a lot) squashing of commits would be nice, before it goes in, but that's a final step. Possibly.

In my opinion, when this is merged, the whole master branch becomes "experimental" until either all old Frontier-clone C++ UI code has been booted out, or/and the old "New UI" lua code; since having three UI systems isn't too nice. I suggest not doing any builds during this time.

Pioneer has about 3000 downloads/month, and I'd rather they didn't get something with three different UIs. (how many of those users try out Pioneer for the first time? Many I'd guess, since regular players probably don't update that often). 

ping dormant @robn, @johnbartholomew, who probably want to follow this thread.

Anyone not on IRC: this has been iterated upon between @ecraven and @nozmajner for the past months, and now "aired" out in the open for the first time. I'm very happy for what the future holds.

clausimu:
@ecraven - fantastic effort!
@impaktor - might take a significant amount of time though to remove all of the old C++ and "new UI" code though.

fluffyfreak:
It seems like a lot of files have spurious changes in them, messing up formatting mostly, did you run and auto-formatter over some of the code whilst editing it?

That little aside I've started looking into it and reviewing, it might take me some time! :) good work so far

ecraven:
yes, I have probably auto-indented in emacs unintentionally.. I can clean up whitespace differences, if that would help?

nozmajner:
I guess I start work on updationg the manual on the wiki, so it can be ready when this gets merged.

.dir-locals.el
```diff
@@ -0,0 +1,4 @@
+((nil . ((indent-tabs-mode . t)
```
laarmen: We usually don't include editor preferences into the main repo. Think of all the Vim users who'd be shocked by the mere hint of an ELisp file on their system!

laarmen: Also, I'm fairly certain most of us use tab=4sp, which would account for some of the weird shifts.

impaktor: Agreed, @ecraven you may add a line in .gitignore for this file type, though.

laarmen:
Could you move the imgui lib into the contrib directory? It makes it easier to track external contributions.

Also, there are a few changes that I'd happily cherry-pick right now, but between the commit history mess and the whitespace confusion... Y'know.

.dir-locals.el
fluffyfreak: I use tabs, I think of the code does, but the tab spacing is equal to 4sp.

nozmajner:
@ecraven shared some newer screengrabs. Hope you don't mind if I post one here, so those who don't lurk on irc that much could see the progress. :)
imgui_wip2

johnbartholomew:
I haven't looked at the commits at all, but the screenshots look great. Fantastic progress!

laarmen:
OK, note to self: once this is merged, we need to investigate whether the onHyperspace event is really useful. However, that can wait until after the merge.

However, I'd really like to have the imgui sources moved to contrib/

walterar:
My opinion so far:

Impress good at first glance.
I see a rapid pace of progress in the project.
The old shortcut keys do not work. Too dependent on the mouse.
I would decide what things to use, or have visible on the screen, and what not.

I follow with attention this project but I would like to know the opinion of those who have worked hard in the current UI, particularly @robn

impaktor:
I'm offline the entire next week, so I'll let @laarmen represent my opinion on the commit history needing sanitation before a merge.

@walterar worry not, before this project was started, both @robn and @johnbartholomew were consulted (since this will replace their code), and they both support this. 

Keybindings will come in a later pull request.

fluffyfreak:
@ecraven / @impaktor should this PR be closed now that it's being split up into other smaller PR's?

impaktor:
I dunno. Doesn't really matter to me. I hope the code changes get into master in other future PRs. And this PR is a bit messy. But we want all the functionallity of it.

nozmajner:
I tried this branch yesterday, and it makes traveling a whole lot easier. 
With the help of the deltaV display, current distance and the approx distance to stop display I was able to travel from the Barnard's starting station to one of the High security prisons, some 1.5 AU away, and only overshot a very little. On the first try, and I also did some extra flying around too.
There was only about 250km/s left in my tanks, compared to the about 3000 the autopilot saves, so with practice, one can do faster transfers by hand with much less guesswork.
What I mean, I think we are on the right track. Some parts are certainly cluttered, and need thought, but the extra information we show to the player will be useful, and it's already on a good track to look good too, without needles decorations. :).

fluffyfreak:
So, is this PR going to get rebased/updated to master and then used or will it be factored out into a new PR?

fluffyfreak:
@ecraven ok so I think what's going wrong at the moment is that after you've go into `PiGui::BakeFonts()` it calls `ClearFonts();` and this leaves us with a dangling pointer in `g.Font->` which is... the `ImGuiContext& g = *GImGui;`.

I can't quite figure out how to fix that up yet, there is an internal `SetCurrentFont` function but as I say, it's internal.

Yes in fact if I comment out `ClearFonts();` then it runs and doesn't crash. So perhaps you don't need to destroy the fonts, or you need to rebuild the fonts and then fixup the various global pointers too them :)

ecraven:
Closing this, as its content will be merged in many future PRs.

PR #3820, opened 2016-09-23
Minor performance cleanup

fluffyfreak:
This is just the results of me running CppCheck over the codebase and tackling the performance issues it lists.

Mostly issues around passing by reference instead of value and initialisation in construction. Nothing ina  tight loop that will actually hurt us.


ecraven:
lgtm :)

Issue #3782, opened 2016-08-15
modelcompiler could use more threads to work faster


HybridDog:
l tried ```$ ./src/modelcompiler -b inplace``` and noticed ($ top) that it uses only one thread.

fluffyfreak:
This is true, and it could use more threads to build multiple models simultaneously.

However, this is also a build tool that only really needs to run very infrequently. Usually on our build machine. A player might only do this if building from source and our model format has changed.

fluffyfreak:
I'll leave this open in case someone has an interest in doing it.

HybridDog:
thanks

impaktor:
In my opinion, this is a non-isse if I've ever seen one. Is this something we want to spend time on? If @HybridDog has a patch, then good. Else, my finger is itching to hit close, to let us track real issues, and keep us focued on what matters.

Sorry if I sound anoyed, but I'm hungry and grumpy at the moment.

fluffyfreak:
I figure that it might be a useful introduction for someone wanting to get involved. Maybe we should make a list of these bite-sized tasks and keep it on the wiki instead of the issue tracker?

fluffyfreak:
Yeah going to track as bite-sized intro tasks on the wiki

HybridDog:
>Maybe we should make a list of these bite-sized tasks and keep it on the wiki instead of the issue tracker?

This would make the issues list at github more lucid, to avoid that someone puts a trifle issue here, you could show a link to the guidelines, e.g. https://github.com/minetest/minetest/issues/new

fluffyfreak:
@HybridDog ah, I didn't know you could do that :)

fluffyfreak:
Ah ha!
https://github.com/blog/2111-issue-and-pull-request-templates
https://help.github.com/articles/creating-an-issue-template-for-your-repository/
Brilliant!

fluffyfreak:
@HybridDog I just had a go at this tonight, unfortunately it's not possible at the moment with Pioneer.

In the background we do a lot of things like caches textures to avoid reloading from disk, generate shared buffers and no doubt a number of other things. None of which are compatible with multithreading at the moment so this would actually be quite a bit of work to fix all of those.

fluffyfreak:
@HybridDog I've just opened a PR that enables with this a little help from an `SDL_mutex` and some head scratching :)

HybridDog:
thanks

fluffyfreak:
[...]

PR #3821, opened 2016-09-24
Multi-Threading modelcompiler

fluffyfreak:
# Description:
So a while ago @HybridDog asked if we could use threads to speed up the modelcompiler #3782 and I said no...

...so here's a PR where I have threaded the modelcompiler!

# Of Note:
I'm not 100% sure that this is a great idea, I mean it is faster, *MUCH* faster on my 12-thread PC. However it does introduce a mutex into the texture cache that will not currently be useful in Pioneer.

On the plus side, it's one more thing out of the way of us eventually multithreading the loading, and other sections of the game engine thus speeding that stuff up.

In short, have a look, tell me what you think people.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -428,6 +428,7 @@ modelcompiler_SOURCES = \
 	FontCache.cpp \
 	IniConfig.cpp \
 	GameConfig.cpp \
+	JobQueue.cpp \
 	Lang.cpp \
 	ModManager.cpp \
 	NavLights.cpp \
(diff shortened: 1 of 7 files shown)


fluffyfreak:
@impaktor / @laarmen any comment?

laarmen:
Not yet, maybe tomorrow.

Le sam. 24 sept. 2016 14:04, Andrew Copland <notifications@github.com> a
écrit :

impaktor:
I'm not against this, I just don't see the speed of modelcompiler as an issue. But if it can spin-off to a future where pioneer itself can run multithreaded, then why not.

fluffyfreak:
Ok, i've made it a little more formal, with a static `Init` method for the TextureBuilder rather than relying on static initialisation ordering by the compiler :)

Will merge when the checks have completed.

PR #3823, opened 2016-09-26
Search&Rescue: set ship look and bugfix

clausimu:
This PR does two things:

1) Building on #3811, #3814 and #3817, ship skins and patterns are now actively set and recreated in SAR missions when reloading a game after saving or returning to a system with target ships. Before, the "look" of a ship was random and would change when the same ship was recreated (for example when leaving and returning to a system). This would lead to loss of immersion into the game for players.

2) While debugging this code I also found another bug where target ships for expired missions were not recreated when returning to a system, leading to a crash. Target ships are now created for as long as the mission actually exists (expired or not).

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -44,6 +44,8 @@ local utils = import("utils")
 local Timer = import("Timer")
 local InfoFace = import("ui/InfoFace")
 local NavButton = import("ui/NavButton")
+local Rand = import("Rand")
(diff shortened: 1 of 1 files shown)


impaktor:
I trust this is self-tested. Beyond that I don't have anything to say. I don't have time/ability to test it myself.

fluffyfreak:
Yeah I am probably going to merge it in the next few days unless someone does it first.
I'm just busy at work which is why I haven't given it a quick test yet.

clausimu:
@impaktor - yes, self-tested :) (which does not mean it can't include stupid mistakes...)

PR #3825, opened 2016-09-30
Throw off the Frontier shackles

impaktor:
Renaming some equipment to saner names has been mentioned before, and I thought now would be a good time with the redesigned UI, that really breaks our Frontier heritage. So no real reason to call a target scanner a "radar mapper", which is a name that says nothing at all unless you just put down your 1993 copy of Frontier. And a "Radar" is more descriptive a name than "Scanner".

This should not be merged until #3819 is in, at which time we are "beyond the point of no return", and anything goes.

I wasn't sure if I should settle for just renaming in the lang-file, or also rename all internal variables and functions, so I did both, separated on different commits, thus even commits (starting on "call it...") can be squashed into odd ones (starting with "Rename...") before merge. Note: github doesn't show commits in correct "git" order, but instead on timestamp, which isn't changed when rebasing.

Last change, (to rename Scanner to Radar), turned out to touch more code than a Belgian child molester (do they code?). At least now we can see how much. How much code is being touched by this change, I mean. Either way I will have to redo that since it will collide on multiple points with #3819.

Also: UNTESTED!
Disclaimer: I'm coming out of a cold, so I'll blame any objections to this PR on that.


laarmen:
OVER MY DEAD BODY!!

Just kidding. I never played Frontier, since I was 3 at the time ;-)

PR #3827, opened 2016-10-06
fix ARFLAGS to remove warning

ecraven:
set ARFLAGS to rcD (instead of rcu)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -78,6 +78,11 @@ WARN_CFLAGS=
 WARN_CXXFLAGS=
 WARN_CPPFLAGS=
 
+AC_SUBST(ARFLAGS)
+if test -z "$ARFLAGS"
+then
+	ARFLAGS="rcD"
+fi
 dnl Check for ccache if they want it
 AC_ARG_WITH([ccache], AS_HELP_STRING([--with-ccache], [Compile using ccache]), [AC_PATH_PROG([CCACHE], [ccache])])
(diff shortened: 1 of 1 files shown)


impaktor:
I can't review this due to lack of knowledge. If the author is sure it is right, I can merge it. Doesn't look like it will start a game of global thermonuclear war, so should be safe.

tomyun:
`D` (deterministic mode) is specific to GNU `ar` and causes build failure on macOS:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ar: illegal option -- D
```

Users can circumvent this issue by explicitly setting `ARFLAGS` when running `configure`, but this may not be ideal.

ecraven:
@tomyun does macOS ar have U (Do not operate in deterministic mode)? There's no actual reason for needing deterministic mode (so far), we just need something that works on all platforms and gives no warnings.
Can you try ARFLAGS=crU, if that works, I'll start a PR, should be fixed soon then.

tomyun:
It doesn't accept `U` option either, but non-deterministic behavior seems to be by default according to this [question](http://stackoverflow.com/questions/6741737/ar-library-override-timestamp) posted in SO.

You may want to have a look at the [man page](http://www.manpages.info/macosx/ar.1.html) for the macOS version of `ar`.

PR #3828, opened 2016-10-06
add "override" to silence warnings in clang

ecraven:
add "override" to all functions where clang++ says it's missing
````
grep override * -R
````
before: 307
after: 445


fluffyfreak:
I approve! :)

PR #3830, opened 2016-10-07
Added military drives from 5 to 9.

fluffyfreak:
NB: I originally did a PR for this about 5 yeas ago, so I finally got around to redoing it :)

The idea for these is that they cannot be purchased but that military or Police ships could be fitted with them so that escape is not an option, nor is chasing them through hyperspace a realistic one.

Obviously it also makes it possible for a black market in ships with them fitted, which are of dubious origins, to be created on the bulletin boards... though this PR does not implement that.

### DIFF
--- a/data/lang/equipment-core/en.json
+++ b/data/lang/equipment-core/en.json
@@ -91,6 +91,26 @@
       "description" : "",
       "message" : "Class 4 Military drive"
    },
+   "DRIVE_MIL5" : {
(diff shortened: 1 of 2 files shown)


fluffyfreak:
The idea is for future use, this just enables that future use.

The massive price difference is because, to re-use your example, it's a liquid coolder nuclear reactor and not the civilian diesel generator model ;)

PR #3831, opened 2016-10-08
Thruster upgrades

fluffyfreak:
# Description:
Quick and dirty implementation of thruster upgrades using the equipment system.

The idea being that the player can buy one of 3 levels of upgrade for their ship. You will always have thrusters, the basic package but you can upgrade them by the following (*example*) amounts:
- 10% improvement
- 20% improvement
- 30% improvement

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -755,6 +755,24 @@ misc.hull_autorepair = EquipType.New({
 	l10n_key="HULL_AUTOREPAIR", slots="hull_autorepair", price=16000,
 	capabilities={mass=40, hull_autorepair=1}, purchasable=true, tech_level="MILITARY"
 })
+misc.thrusters_basic = EquipType.New({
+	l10n_key="THRUSTERS_BASIC", slots="thruster", price=30,
+	tech_level=1,
+	capabilities={mass=0, thruster_power=1}, purchasable=true,
+	icon_name="thrusters_basic"
+})
+misc.thrusters_medium = EquipType.New({
+	l10n_key="THRUSTERS_MEDIUM", slots="thruster", price=50,
+	tech_level=1,
+	capabilities={mass=0, thruster_power=2}, purchasable=true,
+	icon_name="thrusters_medium"
+})
+misc.thrusters_best = EquipType.New({
+	l10n_key="THRUSTERS_BEST", slots="thruster", price=95,
+	tech_level=1,
+	capabilities={mass=0, thruster_power=3}, purchasable=true,
+	icon_name="thrusters_best"
+})
 misc.trade_analyzer = EquipType.New({
 	l10n_key="TRADE_ANALYZER", slots="trade_analyzer", price=400,
 	capabilities={mass=0, trade_analyzer=1}, purchasable=true, tech_level=9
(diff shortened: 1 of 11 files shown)


fluffyfreak:
@laarmen can you help me understand the Travis error? It seems to be failing with `./bootstrap && ./configure` but I haven't touched anything that should cause that.
Have I cocked up or has Travis failed for some other reason?

src/PropertiedObject.h
```diff
@@ -11,6 +11,7 @@ class LuaManager;
 class PropertiedObject {
 public:
 	PropertyMap &Properties() { return m_properties; }
+	const PropertyMap &Properties() const { return m_properties; }
```
laarmen: Why do you want ot pass it as const? I actually kind of like it flagged as mutable, since even though the C++ structure won't be changed, the reference Lua table can.

src/Ship.cpp
```diff
@@ -536,13 +536,15 @@ void Ship::SetAngThrusterState(const vector3d &levels)
 
 vector3d Ship::GetMaxThrust(const vector3d &dir) const
 {
+	size_t thruster_power_cap = 0;
```
laarmen: You might want to add the appropriate pi_lua_generic_pull function in LuaPushPull.h for size_t, because on whatever arch travis compiles, it maps to unsigned long, which doesn't exist, hence the compilation fail.

src/PropertiedObject.h
fluffyfreak: It's called from a method which is const and so the `Properties()` method must also be const.

src/Ship.cpp
fluffyfreak: I switched to unsigned int instead.

data/libs/Equipment.lua
```diff
@@ -755,6 +755,24 @@ misc.hull_autorepair = EquipType.New({
 	l10n_key="HULL_AUTOREPAIR", slots="hull_autorepair", price=16000,
 	capabilities={mass=40, hull_autorepair=1}, purchasable=true, tech_level="MILITARY"
 })
+misc.thrusters_basic = EquipType.New({
+	l10n_key="THRUSTERS_BASIC", slots="thruster", price=30,
+	tech_level=1,
+	capabilities={mass=0, thruster_power=1}, purchasable=true,
+	icon_name="thrusters_basic"
```
impaktor: If instead of `thruster_power=1` have the value of it be `=1.1` would that not save you of using the array `power_bands` later on? Oh, well, just a thought. It depends if it is a bug or feature to go beyond the 1.1, 1.2, 1.3 improvements that are now hard coded into c++ side.

impaktor: Feels like the `power_band` should be set in the `<ship>.json` files then? Or that is a future commit?

walterar:
I can think of many uses for this PR but if we had total control of the thrusters power, things get more interesting.
The question is: Is possible to modify, from lua script, the thrusters power "on the fly"?

fluffyfreak:
That's ouside the scope of this one, this is just for upgrades whereas a general system would be different.

fluffyfreak:
I've removed the WIP tags from this, set the prices up a bit and am calling it ready for review.

The increase in thruster power allows for easier landings and takeoffs on higher gravity worlds as well as increased mobility, rate of turn etc. It might slightly reduce in system travel times but not by much.

PR #3832, opened 2016-10-09
Attempt to find the nearest matching exact name.

fluffyfreak:
Whilst attempting to repeat [this bug](http://spacesimcentral.com/ssc/topic/5879-newcomer-some-feedback-and-a-problem/) I noticed that I was constantly finding different instances of the system `Liaphi`.

It's a really common one but I wanted the nearest to my current star system.

This patch attempts to always return not just the first exact match, but the one nearest to the players current system. Sadly it can only do this for those within the `SystemCache` and so if the one you're looking has been discarded you will find whichever one (*or none*) that it can.

### DIFF
--- a/src/SectorView.cpp
+++ b/src/SectorView.cpp
@@ -417,32 +417,34 @@ void SectorView::OnSearchBoxKeyPress(const SDL_Keysym *keysym)
 		return;
 	} catch (SystemPath::ParseFailure) {}
 
-	bool gotMatch = false, gotStartMatch = false;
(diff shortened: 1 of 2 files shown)


impaktor:
Ah, this is how I've been telling it worked. Good to see reality catch up with me.

fluffyfreak:
It could be improved in a number of ways but I think I'll leave them for someone else to do in another PR!
- display a list of systems with the same name and the distance to each.
- keep the volume around the players current system in memory even as the rest of the cache moves.

PR #3833, opened 2016-10-09
Fixing the scaled down collision mesh of New Ground station.

nozmajner:
LjL noticed that the collision wasn't working for the new_ground station. It turns out the collision mesh was there, but very-very tiny - scaled down. Another point for applying all transformations before exporting.
Not sure why it didn't turned up earlier. I haven't touched the collision mesh back then when I cleaned up the geometry in #2950 . Might be that I didn't noticed it back then. Or maybe the assimp update at #3671  could do it?
Anyhow, I've applied all transformation and should be fine now.

Also I have a commit where I delete all the .DAE files and another with the new .dae files, because windows doesn't give a fuck about cases, but I thought they should be consistent now that I'm on them. This is why I updated the .model file too.


fluffyfreak:
Yar it be grand!

PR #3835, opened 2016-10-14
Calculate selling price including equipment

robothauler:
Calculate ship selling price including equipment and damage

### DIFF
--- a/data/ui/StationView/ShipMarket.lua
+++ b/data/ui/StationView/ShipMarket.lua
@@ -20,6 +20,9 @@ local ui = Engine.ui
 
 local l = Lang.GetResource("ui-core")
 
+local shipSellPriceReduction = 0.5
+local equipSellPriceReduction = 0.8
+
 local shipClassString = {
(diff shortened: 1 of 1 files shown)


impaktor:
So this will auto-sell all equipment on the ship when selling it? Why is hyperdrive treated specially, first removed then added back?

Don't we have wear and tear on engines? Should that be included as well?

Also, the SeconHand module also has a mark down value on equipment, 0.8. I wonder if this should be stored in a Station-specific variable? (that is 0.8 for all station), that this can make use of as well. Hmm, I thought we had that. `local sellPriceReduction = 0.8` is in EquipmentTableWidget, possibly all these should be the same variable stored somewhere.

Anyway, I'm fine with it, if it has been tested by author or other.

robothauler:
Regarding the hyperdrive, I prefer to use class 2 hyperdrive with my sinonatrix because I need more firepower ;-) So the base price is wrong if selling the ship.

PR #3837, opened 2016-10-15
Remove autorefuel

robothauler:
Attempt to remove the autorefuel.
This is WIP

### DIFF
--- a/data/modules/StationRefuelling/StationRefuelling.lua
+++ b/data/modules/StationRefuelling/StationRefuelling.lua
@@ -36,7 +36,6 @@ local onShipDocked = function (ship, station)
 	else
 		Comms.Message(l.WELCOME_ABOARD_STATION_FEE_DEDUCTED:interp({station = station.label,fee = Format.Money(fee)}))
 		ship:AddMoney(0 - fee)
-		ship:SetFuelPercent()
 	end
 end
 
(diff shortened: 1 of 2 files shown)


impaktor:
I haven't read the source. or tested it, so I don't konw how it works, but it might be a good idea to do this now with the "Big Change" coming up.

My main concern might be how to help the player to remember to re-fuel. Put a refuel button next to "launch clearance" button, or/and allow the player to subscribe to auto refuel, (and it would cost money).

Also, hydrogen costs 10 credits in Frontier. Seems more reasonable to me.

nozmajner:
I've started working on a UI for refueling, but my lack of coding skills halted the project.
(yes, that mockup image is as old as it's possible, when we had water as propellant back then :) )
station_main01
Here's where I've got, if it's any help:
https://www.dropbox.com/s/5dl5rgwlh7y3ljy/lobby_refuel.zip?dl=0
It works if there are no fractions, and it doesn't recalculates the deltaV.

fluffyfreak:
Erm, the auto-refuel only refuels the internal tank yes? Why would you want to remove that? Since it's like a gift from the station to give you a reason to dock there.

nozmajner:
Yes, but it's a 2000$ gift if you happen to have a Malabar for example, which makest it too trivial to operate in my opinion.

fluffyfreak:
Ok, well I'm ok with it I just wanted to know the reasoning behind it :)

robothauler:
All right, I think it's time to have a look and to comment.
screenshot-20161023-135705

I have tried to shrink it.
This is on a 4:3 screen and with a military drive installed.
screenshot-20161023-135837

PR #3841, opened 2016-10-22
Put imgui.ini file in user's pioneer config folder

impaktor:
Idea is to put the imgui.ini file in the player's pioneer config directory, rather than the pioneer root folder, or player's home folder. I believe that file stores window positions/states between sessions?

Think of this as a feature request, if the code is wrong.


impaktor:
I really winged this one, and I hadn't expected it would compile even, which it did on my computer, but not travis, so I'll just close this.

fluffyfreak:
Nah it'd be good to get it working, I'm not even sure why Travis failed the errors from gcc are bloody appalling to understand these days.

impaktor:
I'll leave it to @ecraven to do. I suspect one should not add code into contrib like I did.

PR #3843, opened 2016-10-23
Put imgui.ini into user config dir

ecraven:
Put imgui.ini into UserConfig.


fluffyfreak:
I don't think it leaking is a problem. It'll be cleaned up on shutdown anyway and it's only allocated the once.

ecraven:
I agree, just wanted to comment that this is something we know, but chose to ignore :)

impaktor:
I'm fine if @ecraven wants to make this his first self-merge.
http://pioneerwiki.com/wiki/Using_git_and_GitHub#How_to_merge_a_PR_the_propper.2Fpedantic_way

laarmen:
Likewise, the leak doesn't bother me much.

PR #3845, opened 2016-10-23
Updating the Bluenose Medium LOD with the correct mesh. 

### LINKED ISSUE #3844: Bluenose High and Medium LODs are the same model
I accidentally noticed that the High and Medium level LOD of the Bluenose is the same model, except the Medium one doesn't have the locators (which is correct BTW).
I was looking at the model file sizes out of curiosity, when we were discussing ship numbers/performance on IRC. 
@BobTheTerrible: Is it an error while exporting, and you have the mid LOD, or you/me should make one?

nozmajner:
Updating the mid LOD of the Bluenose, because it was actually the high LOD sans locators.
It's now 6k tris instead of 15k.
Also updating the .model file: Fuselage-material --> Fuselage, so the ship doesn't throw the material errors in the model viewer. (This changed recently?)
fixes #3844 
(Sorry for jumping early and reporting, before I checked if the correct file is up somewhere)

I've got the Medium LOD from the source I found in the Asset repo of @BobTheTerrible. 
One note to Bob: There was an empty mesh in the Middle lod, which crashed the model viewer (and the game to I'd guess, didn't try). You might want to delete that from the source to avoid errors later.
Also there's room for some optimizations in the mid LOD, like the belly thrusters could be simplified a lot, and also the little windows have their geometry there, but it could be two quads per window, and the color of it on the texture, where the hole would be. (If you ever have this kind of idle time to burn away :) )


BobTheTerrible:
Ah... if the high and low LODs were the same then Im guessing the issue with the empty mesh wouldnt have shown up when I was testing it?

Thanks for pointing out the other issues will get on them as soon as I can!

fluffyfreak:
The material names changed with the update to Assimp which itself had a bug fix in it. Using the "Fuselage-material" was actually a bug with the DAE importer as it should have been just the name of the material "Fuselage" :/

PR #3848, opened 2016-10-26
Avoid crash if player jumps to next system

robothauler:
This is a workaround. Maybe this can be removed after transitation to pigui.


ecraven:
Wouldn't it be easier to *not* call the lobby code if the player isn't in fact in the lobby?

fluffyfreak:
What crash? I haven't been able to get one to happen.

impaktor:
@fluffyfreak I think the IRC conversation was that it triggers a crash when jumping into (an uninhabited?) system.

fluffyfreak:
Merged as it fixes current problem.

PR #3853, opened 2016-11-02
Enhance serializer crash output

laarmen:
I've added some output a bit more useful to the serializer crash thingy.

Also, I've added a way to shoot ourselves in the foot (feet) intentionally with the debug.makenull() function to create a NULL light userdata. That's quite useful to provoke a crash.

### DIFF
--- a/src/LuaSerializer.cpp
+++ b/src/LuaSerializer.cpp
@@ -52,7 +52,7 @@
 // "Deserialize" function under that namespace. that data returned will be
 // given back to the module
 
-void LuaSerializer::pickle(lua_State *l, int to_serialize, std::string &out, const char *key)
+void LuaSerializer::pickle(lua_State *l, int to_serialize, std::string &out, std::string key)
 {
 	static char buf[256];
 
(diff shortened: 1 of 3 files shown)


laarmen:
Do not merge just yet, I'm going to try and improve the thing with some std::typeid trickery to identify the actual type of the userdata, and stuff.

src/LuaSerializer.cpp
```diff
@@ -153,15 +153,15 @@ void LuaSerializer::pickle(lua_State *l, int to_serialize, std::string &out, con
...
-					if (!k) {
-						lua_pushvalue(l, -2);
-						k = lua_tostring(l, -1);
-						lua_pop(l, 1);
-					}
+
+					lua_pushvalue(l, -2);
+					const char *k = lua_tostring(l, -1);
+                    std::string new_key = key + "." + (k? std::string(k) : "<" + std::string(lua_typename(l, lua_type(l, -1))) + ">");
+					lua_pop(l, 1);
```
impaktor: Looks to me like github is complaining about indentation here?

laarmen:
Indentation fixed, and I realized what I wanted to add was actually useless, so this is mergeable.

Issue #3852, opened 2016-11-02
Lua serializer '(null)' tried to serialize an invalid 'SpaceStation' object


impioneer:
### Observed behaviour

The error described in title (or similar, for example it could be 'Ship' instead of 'SpaceStation') and crashing of the game occurs on Save (manual) or autosave (when docking to a station).

### Steps to reproduce

__My pioneer version (and OS):__

Linux 64 - 20161028, both pre-built and compiled from source.
I use Debian Jessie (stable) 64 bit.

laarmen:
Does this crash happen systematically?

impioneer:
It happened many time, but not on every save

ecraven:
could you paste the entire error message here, not just the last line?

laarmen:
Sadly, our serializer errors are succinct, to say the least.

impioneer:
That's the entire message. It's a small window with the title "Pioneer error" and then the line "Lua ..."  and "OK" button. That's all

laarmen:
Just in case, any mod installed?

impioneer:
no

impioneer:
IDK if this helps: I'm looking into the file ~/.xsession-errors
The last line before the serializer error is "TradeShips: Removed 5 ships before serialization"

laarmen:
Thanks. However, this doesn't really help (unless tradeships.lua triggers
autosaves, which I doubt, but cannot be sure of since it is a big can of
worms I won't open again.)

impioneer:
Autosave was disabled in this case, I used a manual Save.

Is this a known bug? Should I try any earlier version? Any known workaround? Really liked this game, but it's not worth much without being able to save :(

laarmen:
What we need is a better error output to pinpoint where exactly do things
go wrong. I'll work on that tonight.

Yes, I will, and not just try to ;-)

laarmen:
You could try an earlier version (like one from maybe 3 months ago) , and
maybe, if you're courageous, bisect from there of it works?

impioneer:
Thanks for your effort.

There were several releases in July, any specific one?

Also, I don't understand your phrase " bisect from there of it works"

impaktor:
> Is this a known bug?

No, not until now. Probably something recent. If you can figure out  how we can reproduce it, or give more info on error, then we can (hopefully) squash the bug.

impaktor:
> Also, I don't understand your phrase " bisect from there of it works"

[...]

PR #3855, opened 2016-11-03
fix pattern selection if no pattern present

clausimu:
This fixes a bug mentioned in #3852 (but is unrelated to #3852), whereby ships that don't have any patterns associated with them crash the game when used as SAR mission targets. It was an oversight by me initially. I thought I had checked to see that every (non-police) ship has at least one pattern. But the Bowfin fighter and the Deep Space Miner do not.


nozmajner:
Nice! Better this way anyhow, because mods could add ships with no patterns too, and you won't have control on that.

PR #3856, opened 2016-11-03
Custom system checking

fluffyfreak:
I noticed that their very little error checking in the `CustomSystem` creation and so when a value like `100,000,000,000,000`, aka `100 trillion`, is used as the denominator for a fixed point value and it gives a planet negative mass or radius this was simply being accepted by the engine and silently used.

I have added some basic error checking for negative values, and a new macro for the Lua fields which must be greater than Zero like the Mass and Radius of a planet.

Next I fixed up the existing data, this might mean that some planets/asteroids are now **WRONG**
This is because I cannot figure out the maths to go from a figure like `21.5 km` into our FIXED structure.
The wiki helpfully states `"mass. For stars, 1.0 is the mass of Sol. For planets, 1.0 is the mass of Earth"` but not the maths required to convert into that format nor how to get the number and denominator parts.
At the least I think that we need that writing down somewhere!

Finally I noticed that NaN's sometimes made it through to the AICmdFlyAround when trying to approach these invalidly sized bodies so I have added assert checks to catch them in the future.

**EDIT**
It seems that the limit for the `fixedf<32>` format we use is somewhere between 10,000,000,000 and 100,000,000,000 which seems fairly big. After that the precision is completely lost and all numbers are just zero.

The value for Metis was off by about 6 orders of magnitude:
`100,000,000,000,000` vs
`           1,000,000,000`
Which seems to get it to approximately the right size. Honestly it is really difficult to tell with this stuff.


ecraven:
I've used the numbers for new planets as follows:
Look up the mass of the earth on wikipedia: https://en.wikipedia.org/wiki/Earth  ->  	5.97237×10^24 kg
Convert your mass to a fraction of that, you get something like 0.00005734
Make that fixed point: f(5734,100000000)
That seems to work, I used it to add a few new planets. I haven't thought it through, but it should be possible to have lua functions that create the fixed point numbers from "normal" floating point numbers, with a bit of magic math :) that would definitely help with checking numbers and with adding new planets and stuff.

PR #3857, opened 2016-11-04
Lobby: Disconnect the property listeners when the docked station changes

### LINKED ISSUE #3852: Lua serializer '(null)' tried to serialize an invalid 'SpaceStation' object

laarmen:
Closes #3852

### DIFF
--- a/data/ui/StationView/Lobby.lua
+++ b/data/ui/StationView/Lobby.lua
@@ -48,12 +48,21 @@ local lobby = function (tab)
 		end
 	end)
 
+	local connections = {}
+
 	local deltaV = ui:Label("")
 	local updateDeltaV = function ()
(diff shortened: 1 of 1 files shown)


fluffyfreak:
Tested and haven't had a problem with taxi missions saving/loading, hyperspacing etc.

impioneer:
Doesn't work for me. Added a comment to  #3852

PR #3859, opened 2016-11-07
Lobby.lua: Add missing return

### LINKED ISSUE #3852: Lua serializer '(null)' tried to serialize an invalid 'SpaceStation' object

laarmen:
Should fix #3852 for real !


impaktor:
Ha! I just merged it on my branch and was trying to push it to upstream master, but was getting denied!

PR #3863, opened 2016-11-08
Docking fee Text update (since station won't refuel you)

### LINKED ISSUE #3862: Autorefuel at docking/landing not working
Observed on 20161028 on Fedora 24.  Seemed to work on 20161022.

nozmajner:
This fixes the english lang file for the autorefuel module, so the station doesn't talk about fuel fee, only docking fee. Translations will need to be updated on transifex.
Because since #3837 the stations won't fill your tanks, just greet you since they are greedy and want to sell you their precious Hydrogen. :) Which makes sense.
There was an imgui.ini file in the first commit, because my git-fu is still weak, but it should be removed now.

One design question that relates here: should anything happen if you don't have the money to pay for the docking fee? It's a very rare case anyhow, but not entirely pointless.

closes #3862

### DIFF
--- a/data/lang/module-stationrefuelling/en.json
+++ b/data/lang/module-stationrefuelling/en.json
@@ -1,10 +1,10 @@
 {
    "THIS_IS_STATION_YOU_DO_NOT_HAVE_ENOUGH" : {
       "description" : "",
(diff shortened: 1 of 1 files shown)


impaktor:
Oh, I think those commits need squashing, to one single.

Issue #3800, opened 2016-08-26
Camera position wiggles spasmodically in particular system


yump:
### Observed behaviour

The position of the camera rapidly jerks around when the ship is in the Esswaze (-4, -4, 8) system.  The effect is most visible when landed, or when the external or sidereal views are used with a close zoom.  If other systems are affected, I haven't run into one yet.
### Expected behaviour

The position of the camera does not wiggle when the ship is in the Esswaze (-4, -4, 8) system, or any other system that shares whatever underlying characteristic causes this.
### Steps to reproduce
1. Load [this save file](https://u.pomf.is/rbhhza).
2. Un-pause the game.
3. Press F1.
4. Observe the wiggling.
5. Make a hyperspace jump to some other nearby system, and observe that the wiggling is gone.

The save file is a convenience. I was also able to produce the effect by starting a new game either at Barnard's Star or Earth and traveling directly to Esswaze.

**My pioneer version (and OS):** 20160814-linux64, Fedora 24

**opengl.txt:** [here](http://pastebin.com/pCt1eeY7)

nozmajner:
My observation:
That system is quite huge. The starport around the 4th star is 0.29ly (18 339 AU) away from the entry point around the first star. If you jump to the first, there's only a slight wiggle, but if you jump to the 4th, the wiggle is quite much larger. The 3D cockpit doesn't wiggle at all.
When landed, the ship and the whole city wiggles in concert, and that's both visible from internal and external view.
And the orbit map behaves wiggly too.
Nothing's wiggling while paused.
Maybe it's due to the size of the system?

yump:
I think you're right about the size of the system.  I just ran into another wiggler, Exphiack (-1,6,-7), and it has a star that's 0.24 ly away from the others.

impaktor:
Didn't Barnard's star have some similar issue?
@fluffyfreak ?

fluffyfreak:
@impaktor it's interesting and yes it looks like a precision issue somewhere, perhaps in the camera? Hard to say really but I think @johnbartholomew has looked into these things before if he's got any insight?

johnbartholomew:
Sorry, I don't have any insight. I agree it's a precision issue but I don't know where in the code the culprit is.

[...]

PR #3872, opened 2016-11-19
Jitter fix

jaj22:
Various fixes for rendering jitter problems, including those in #3800. Description copy-pasted from there:

1. The changes in Frames.cpp enable the use of relative positions in cases where the two frames share a near parent. The "relative to child" case is standard for rendering, because the player's ship and the camera are children of the same parent. The other two cases are less important, but apply when two nearby objects are in different reference frames. I tested those by modding the projectiles.

2. The change in Camera.cpp fixes an old bug where interpolated frame coordinates were not used for rendering, causing heavy rendering jitter and frame transition glitches. The easiest way to demonstrate this is to move in and out of one of the far stations' reference frames at Barnard's Star.

3. The change in WorldView.cpp is a workaround for an issue where the player's ship label is placed exactly between the four centre pixels in external view, causing it to jitter.

Note that some jittering is not fixed, by design. Objects that are very far from their frame's centre will still jitter. The external view will also jitter in the same circumstances, because the camera frame and player ship are both children of the same frame. Many HUD elements also jitter badly: I haven't looked into fixing these, as the HUD is being reworked anyway.

These changes are not thoroughly tested due to several problems: Release mode crashes every few minutes here (possibly GPU driver issue), debug mode is unusably slow and I'm not up to date with current Pioneer functionality. Because the frame code affects everything, I recommend more extensive gameplay testing.


fluffyfreak:
Tested this in the linked #3800 issue and it does seem to fix it in most circumstances. 
When I had to travel 0.29ly there was still some camera jitter on the external camera but as noted that's to be expected.

Probably still needs a bit more testing but the code looks solid and I haven't noted any problems so far.

impaktor:
> Probably still needs a bit more testing but the code looks solid and I haven't noted any problems so far.

Considering @nozmajner also did some basic testing, I think we can put the rest of the testing burden on the players, by merging it into master. Also, welcome back.

fluffyfreak:
@impaktor is there anyway we can test this a bit more thoroughly?
Maybe not now but for things like this in the future where we might want to let an AI test it somehow?

Merging now anyway.

PR #3874, opened 2016-11-21
Render the navlights for orbital stations

### LINKED ISSUE #3873: Continually-accumulating vertex buffer in scenegraph/Billboard.cpp
I'm not sure how this code is supposed to work, but every time Billboard::Render() is called, it adds another vertex to the buffer, renders nothing and then doesn't clear the buffer. The vertex buffer does not appear to be cleared elsewhere, so it continually expands. The orbital station models appear to use this function.

The results here are that if you wait near an orbital station, the 2GB address space limit is exceeded in around 120,000 frames, typically causing a crash with a useless traceback.

**Steps to reproduce:**
1. Start game on Barnard's Star.
2. Wait a bit.
3. Put breakpoint on VertexArray::Add(const vector3f &v, const vector3f &n)
4. Check out size of the position & normal vectors.

**Setup:**
Pioneer version: Compiled from current master with VC2015
OS: Windows 8.1 64-bit
Settings: Default
Mods: None

fluffyfreak:
This fixes #3873 where somehow the call to render the navlights had gone missing.
This meant that instead of being flushed at the end of each frame they were simply accumulating until the game exceeded the 2GB barrier on 32-bit processes.

### DIFF
--- a/src/NavLights.cpp
+++ b/src/NavLights.cpp
@@ -92,7 +92,8 @@ NavLights::NavLights(SceneGraph::Model *model, float period)
 	const std::vector<Node*> &results = lightFinder.GetResults();
 
 	//attach light billboards
-	for (unsigned int i=0; i < results.size(); i++) {
+	for (unsigned int i=0; i < results.size(); i++) 
+	{
 		MatrixTransform *mt = dynamic_cast<MatrixTransform*>(results.at(i));
 		assert(mt);
 		Billboard *bblight = new Billboard(m_billboardTris, renderer, BILLBOARD_SIZE);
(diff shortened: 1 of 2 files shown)


fluffyfreak:
This has been broken since Sept 14th by the look of it, when I did it in PR #3809

impaktor:
So only 2 months and 32bit? That would explain no one reporting it (until now). What's the limit for 64 bit? I have an eerie feeling that question is stupid.

nozmajner:
Since Sept? I noticed yesterday while playing with ecraven's hud branch, but didn't realized that it was broken for the releases too.

impaktor:
If we're interested in adopting gitmoji standard for commit messages:
https://gitmoji.carloscuesta.me/

(sorry for spamming, just wanted to procrastinate some)

laarmen:
Please don't. Pretty please?

jaj22:
Tested, confirmed to fix #3873. Station now lights up like a Christmas tree.

Apparently the Windows releases are 32-bit, so I suspect a lot of users hit this bug and didn't report it. It's aggravated by disabling VSync.

fluffyfreak:
Cheers @jaj22 :)

PR #3877, opened 2016-11-24
Remove GL buffer init fills, add init checks

jaj22:
## Purpose
Fixes a severe performance problem on at least some AMD GL drivers. To avoid potential undefined behaviour, the GL buffer code used to zero-fill the buffers on creation. This meant that a newly-created buffer was often written multiple times per frame, which causes a synchronization stall on some GL drivers even when the buffer wasn't used in a draw call.

To replace the undefined behaviour functionality, asserts were added to the Bind calls to check whether buffers are used before being written.

## Related problems
The two heading labels are forced to update every frame by the `UI::Label::Layout` function. `UI::Label::Draw` then creates a new vertex buffer and draws with it immediately. This patch works around the performance problem, but UI elements should not be updated unnecessarily.

`StaticGeometry::Render` causes a related problem by writing to the same instance buffer each frame. This wouldn't be a problem if frame times were reasonably constant, but Pioneer has asynchronous rendering and physics, so a common pattern looks like this:

1. Short render frame
2. Long render frame due to GPU sync stall
3. Physics frame x3 or x4
4. Repeat 

Double-buffering the instance buffers (or even recreating them every frame as a dirty hack) partially works around the problem and increases nominal frame rates, but a more rigorous approach to physics/render interleaving is necessary for smooth rendering.

### DIFF
--- a/src/graphics/opengl/VertexBufferGL.cpp
+++ b/src/graphics/opengl/VertexBufferGL.cpp
@@ -158,6 +158,7 @@ void VertexBuffer::Unmap()
 		if (m_mapMode == BUFFER_MAP_WRITE) {
 			const GLsizei dataSize = m_desc.numVertices * m_desc.stride;
 			glBindBuffer(GL_ARRAY_BUFFER, m_buffer);
+			glBufferData(GL_ARRAY_BUFFER, dataSize, 0, GL_DYNAMIC_DRAW);
 			glBufferSubData(GL_ARRAY_BUFFER, 0, dataSize, m_data);
 			glBindBuffer(GL_ARRAY_BUFFER, 0);
 		}
@@ -403,6 +404,7 @@ void IndexBuffer::Unmap()
 	} else {
 		if (m_mapMode == BUFFER_MAP_WRITE) {
 			glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, m_buffer);
+			glBufferData(GL_ARRAY_BUFFER, sizeof(Uint32) * m_size, 0, GL_DYNAMIC_DRAW);
 			glBufferSubData(GL_ELEMENT_ARRAY_BUFFER, 0, sizeof(Uint32) * m_size, m_data);
 			glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, 0);
 		}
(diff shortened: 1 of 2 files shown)


fluffyfreak:
Well this seems to work and even on a high end machine you can see some performance difference so hopefully the lower-end hardware will really notice it!

jaj22:
A followup on `StaticGeometry::Render`: Multi-buffering the instance buffers does not help on this GL driver. As far as I can tell, the driver assumes that _all_ written buffers are being drawn from, so any modification of the contents of a buffer causes a sync stall, regardless of streaming hints.

The single buffer update method I've found that doesn't stall is to invalidate the buffer prior to writing it, using the `glBufferData` call again with the `NULL` pointer. Note that `glBufferData` with a data pointer does not work, even though there's no logical reason for that. The driver simply has only one fast path.

The [opengl.org docs on buffer streaming](https://www.opengl.org/wiki/Buffer_Object_Streaming) seem quite keen on this method, so it may not be unusual. However, I'm not happy about pushing it to master without testing on a variety of platforms.

impaktor:
> I'm not happy about pushing it to master without testing on a variety of platforms.

Anythin in particular that we should look out for? Frames / second, or that it even runs?

fluffyfreak:
@impaktor I think just:
- see if it runs
- is everything that you'd expect to see rendered?
- is there visual corruption like rapidly flickering geometry, colours, etc.

fluffyfreak:
Retested in Windows and seems fine. Cannot test OSX or Linux builds at moment.

impaktor:
Tested. Seems to work. I flew about in London, and at an orbital station.

output.txt says I have:
3.1.0 NVIDIA 340.96, running on NVIDIA Corporation GeForce 210/PCIe/SSE2

jaj22:
> Anything in particular that we should look out for? Frames / second, or that it even runs?

Just frames/second. There won't be any risks aside from performance. Without profiling, the ideal test is probably something like this: 

1. Comment out the contents of Geosphere::Render, compile.
2. Disable vsync, set city detail to very low (less rendering per instance buffer write).
3. Load saved game next to a city, preferably a system without much other junk (ships, stations).

To explain the general problem, GL provides several ways to push data to the GPU, and the performance is highly implementation-dependent. In this case, the AMD win64 driver I'm using provides only one fast path: Explicitly invalidating the whole buffer before rewriting it. On other drivers (nVidia, various mobile chips, Linux stuff), there may be faster methods. However, this method is probably reasonably fast on all devices given than Pioneer always rewrites the whole buffer at once.

PR #3878, opened 2016-11-25
Optimisation: Reserve VA capacity correctly for Gui::TextLayout

jaj22:
TextureFont attempts to reserve capacity in vertex arrays, but it doesn't work in cases where the Populate functions are called with a vertex array that already contains vertices. `Gui::TextLayout` populates the vertex array word-by-word, so this happens routinely.

It's not a huge gain for WorldView, but it's significant when there's a lot of text on-screen. With the debug text output (the lower five lines use TextLayout), `TextureFont::PopulateMarkup` CPU usage drops from 8% to 5% here.

### DIFF
--- a/src/gui/GuiTextLayout.cpp
+++ b/src/gui/GuiTextLayout.cpp
@@ -129,8 +129,19 @@ void TextLayout::Update(const float width, const Color &color)
 
 		Color c = color;
 
+		// vertex array pre-assignment, because TextureFont botches it
+		// over-reserves for markup, but we don't care
+		int numChars = 0;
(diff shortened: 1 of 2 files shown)


fluffyfreak:
Tested seems to work just fine :) 👍

jaj22:
Thought I'd made an error by using `std::vector::reserve()` in TextureFont without checking the current capacity, but apparently reserve never reduces capacity by spec, so it should be fine. 

A few (cheap) operations can be saved by checking current capacity and/or assuming the position/diffuse/uv0 vectors are the same size, but it's probably a waste of code as the rest of the function is far more expensive.

PR #3889, opened 2016-12-08
Fix some orbital data in Sol system

ecraven:
Fix the orbits of Metis, Adrastea and Rhea to be correct and synchronous.
Add Saturn III, Tethys
Add the minor planets 2 Pallas, 4 Vesta and 1 Ceres 
(3 Juno is only the 11th largest asteroid)


ecraven:
The orbital data are all taken from wikipedia. The "looks" are most certainly all totally wrong, but the orbits are correct (except for the phases / anomalies, I haven't calculated them for 3200 yet :-/)

impaktor:
Just read the backlog from last night.
1. planet phases are correct for 3200
2. This might need save bump

impaktor:
To elaborate on my point above, in #3135 the correct phases of the planets were included, based on the table I posted there,  so it should be correct in master.

fluffyfreak:
So is this ready for merge then?
I'm afraid I don't understand the conversions enough to comment on whether they're correct or not.

laarmen:
Do we have anything in the tubes requiring a savebump?

fluffyfreak:
I don't think so.

impaktor:
> Do we have anything in the tubes requiring a savebump?

Might be that with the large imgui work something might pop up?

Maybe this shouldn't savebump, because chance of player actually being affected is low? Or is there some actual body-index change?

fluffyfreak:
Yeah this might be ok without the savebump.

impaktor:
There are some terrain gen stuff that have been discussed to be changed. Not sure if/how/when.

Also, from IRC, 
```
<fellowExplorer> visit Groombridge 1618 b, that one rotates so fast that the
                 equatorial regions are moving over escape velocity, your will
                 just lift off into space from it's surface!  bug/feature , or
                 just intended strangeness off space? :)                [00:40]
```

PR #3890, opened 2016-12-08
Error on non-existing translations.

ecraven:
Add missing token PASSENGER_CABIN to equipment-core.

Use l:get from now on to get the translation or nil (no error if an entry
doesn't exist)

If a translation doesn't exist, the missing token and the resource and language
code it was supposed to be in are shown:

error: [string "[T] @modules/CrewContracts/CrewContracts.lua"]:402: missing translation token: `NAVIGATOR' in resource `module-crewcontracts', language code `en'.
stack traceback:
	[C]: in function '__index'
	[string "[T] @modules/CrewContracts/CrewContracts.lua"]:402: in function 'cb'

### DIFF
--- a/src/LuaLang.cpp
+++ b/src/LuaLang.cpp
@@ -10,7 +10,27 @@
 
 static int _resource_index(lua_State *l)
 {
-	return luaL_error(l, "unknown translation token: %s", lua_tostring(l, 2));
+	const char *token = lua_tostring(l, 2);
+	lua_pushstring(l, "resourceName");
+	lua_gettable(l, -3);
+	const char *resourceName = lua_tostring(l, -1);
+	lua_pop(l, 1);
+	lua_pushstring(l, "langCode");
+	lua_gettable(l, -3);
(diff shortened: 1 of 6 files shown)


ecraven:
This would ensure that we actually catch missing translation tokens. If you really need to check for the existence of a token, use l:get("FOO") instead of l.FOO.

src/LuaLang.cpp
```diff
@@ -10,7 +10,27 @@
 
 static int _resource_index(lua_State *l)
 {
-	return luaL_error(l, "unknown translation token: %s", lua_tostring(l, 2));
+	std::string token = lua_tostring(l, 2);
```
ecraven: fixed and force pushed

laarmen:
@impaktor do you wish to test this? Otherwise, I'd say it's good to go.

impaktor:
It someone _wants_ me to test it I can do it tomorrow, otherwise, I have no opinions. Merge when ready.

PR #3892, opened 2016-12-10
Check availability of translation string with l:get()

robothauler:
Since #3890 it is necessary to check translation with l:get() because not all branches have appropriate strings.


ecraven:
Thanks for catching this!

impaktor:
Shouldn't this be the case for all modules, not just cargo run?

ecraven:
@impaktor: I did start the game a few times, and fixed everything I found, the crew module, for example. But it seems I missed quite a few places :-/ no decent machine this weekend, I'll investigate tomorrow night.

PR #3895, opened 2016-12-15
Provide config flag and warning to work around Intel pointsprite bug

### LINKED ISSUE #3888: Stars not visible since 20160814
Hello, firstly thanks a lot to all developers for this sim. Great work.

Now to the problem - two days ago I have installed the sim first time (latest version - 20161129-win32 ) and immediately noticed something is not OK - there were missing stars in the outer space - there were visible just background texture of Galaxy ring - see the screenshot (the galaxy texture on the left).

I tried all various video settings, but I found that stars are OK in older versions, so I focused to stars improvement in change log and identified the last working version - that is 20160724 - where all is OK (test proved that in the next version 20160814 the problem appeared first time).
 - Ah and I just remember another observation - it is not only stars which are missing, but also the flashing lights on the bases and ships which stopped working since that version as well.

My OS is Windows 10 and HW is Lenovo T420 with Intel HD 3000 graphics (OpenGL 3.1 fully supported, latest available video driver). Attached is also the openGL.txt file
I am available for further testing, if it can be fixed by some config tweaking or mods...
(if it is caused by some new star feature, maybe there could be option in menu to use old stars...)

### Observed behaviour
Missing stars in game and in the Main menu, missing stations/ship lights.

### Expected behaviour
Stars should be visible as it is in version 20160724 and older.

### Steps to reproduce
it happens always with version 20160814 and higher.

no_stars_and_lights

[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/637826/opengl.txt)

jaj22:
Fixes #3888. It's a bit messy because there's no way to detect the faulty GL driver until the context has been created, which is currently a separate code branch. It also doesn't set the config parameter itself, to avoid adding a Pi.h dependency in RendererGL.

Another approach would be to move a bunch of init code out of Graphics.cpp and give RendererGL the responsibility for creating the SDL window. In this case, the SDL window can be immediately recreated without the forward compatibility flag after the faulty driver is detected. This would also clear up some of the hacks used for the null renderer.

Another issue is that the code only detects the known-faulty Intel driver. It's quite possible that other Intel drivers have the same fault, and in this case the user will be expected to figure out that the correct workaround. The alternative detection method would require writing a point sprite to a frame buffer texture and then reading it, which may fail on various drivers for other reasons. There's also a question of how much effort we want to expend working around a driver bug.

### DIFF
--- a/src/GameConfig.cpp
+++ b/src/GameConfig.cpp
@@ -54,6 +54,7 @@ GameConfig::GameConfig(const map_string &override_)
 #endif
 	map["EnableGLDebug"] = "0";
 	map["EnableGPUJobs"] = "1";
+	map["GL3ForwardCompatible"] = "1";
 
 	Load();
 
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -467,6 +467,7 @@ void Pi::Init(const std::map<std::string,std::string> &options, bool no_gui)
 	videoSettings.useTextureCompression = (config->Int("UseTextureCompression") != 0);
 	videoSettings.useAnisotropicFiltering = (config->Int("UseAnisotropicFiltering") != 0);
 	videoSettings.enableDebugMessages = (config->Int("EnableGLDebug") != 0);
+	videoSettings.gl3ForwardCompatible = (config->Int("GL3ForwardCompatible") != 0);
 	videoSettings.iconFile = OS::GetIconFilename();
 	videoSettings.title = "Pioneer";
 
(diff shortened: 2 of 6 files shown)


fluffyfreak:
Good clean workaround. I don't think we should spend too much time hacking around driver bugs.
It'd be nice to catch them all, perhaps update a FAQ somewhere (_included with game_), but I don't think that's realistic for our team size.

PR #3896, opened 2016-12-16
Keep quit confirmation messages modern and stylish

impaktor:
Also, this is not just relevant for keeping up to date with 2016 but also for
the year 2020 (if we're still here by then), when there will be a presidential
election one more time!


impaktor:
@laarmen will kill me

laarmen:
Y'know, you could have said "Leave" (with the cap) instead of quit. Might
as well flame the Brexit while we're at it.

Issue #38, opened 2011-04-24
Separate Ship and Player


robn:
Right now `Ship` has too much knowledge and too many methods that are player-specific and should live in `Player`. Two examples:
- `SetNavTarget` & `SetCombatTarget` aren't required by NPCs; they're strictly UI functions. This is reinforced by the fact that these methods cause a sound to be played
- `AddMoney` and friends. NPC ships don't need money; they'll be equipped as necessary by ships. Indeed, `Player` should probably be a `MarketAgent`, but `Ship` should not.

Someone needs to go through and check and move things out of `Ship` into `Player` as appropriate. I think a good rule-of-thumb is that if it is a UI function in any way, then it belongs in `Player`. Another test might be to simply remove it from `Ship` and see if the code compiles.

walterar:
It's a shame that after all this time no one has been interested in solving this. :(

salborrelli:
@walterar FYI I've already "promised" to take a look at this one as soon as I have some time, see PR #3618 :smile:

walterar:
Thank you very much Sal. All my thumb up be for you. :))

salborrelli:
@robn @walterar after a quick look at the code I found that most (if not all) methods of the Ship class seems to be quite specific to the Ship, I haven't found any method which should logically be moved to the Player class.

Sure, some of the methods you mention in the above list (SetNavTarget, SetCombatTarget etc.) are part of the PlayerShipController class. Do you mean moving those methods from PlayerShipController to Player?

Another important thing to note is that, in the current version, the Player class inherits from Ship! Which means a Player doesn't really "owns" a Ship, because he "is" the Ship. Therefore, separating the Player from the Ship is not a matter of moving a few methods from here to there, we should go through a (possibly complex) re-ingegnerization of the code in order to separate the Player from the Ship, and to transform the Inheritance relationship to a Composition.

To summarize, given my far from complete knowledge of the game architecture, I'm probably missing something basic; that said, at the moment this issue is not completely clear to me. Can someone please describe the problem in more details please? Do you want to transform the Player-to-Ship relationship from Inheritance to Composition, or what?

laarmen:
I might putting my fott in my mouth here, but my guess is that the
potential breakage will probably be at the Lua/C++ interface, and in our
Lua scripts. The compiler can tell us when the C++ side mixes the two
notions up, but not when it's in a mod.

impaktor:
[...]

PR #3900, opened 2017-01-05
Override method "IsPlayerShip" for player

mike-f1:
Adding IsPlayerShip in class player

### DIFF
--- a/src/LuaPlayer.cpp
+++ b/src/LuaPlayer.cpp
@@ -19,8 +19,17 @@
 
 static int l_player_is_player(lua_State *l)
 {
-    lua_pushboolean(l, true);
-    return 1;
+	LUA_DEBUG_START(l);
+
+	Ship *s = LuaObject<Ship>::CheckFromLua(1);
+
+	bool player_ship = s->IsPlayerShip();
+
+    lua_pushboolean(l, player_ship);
+
+	LUA_DEBUG_END(l, 0);
+
+    return player_ship;
 }
 
 /*
--- a/src/LuaShip.cpp
+++ b/src/LuaShip.cpp
@@ -56,8 +56,17 @@
  */
 static int l_ship_is_player(lua_State *l)
 {
-    lua_pushboolean(l, false);
-    return 1;
+	LUA_DEBUG_START(l);
+
+	Ship *s = LuaObject<Ship>::CheckFromLua(1);
+
+	bool player_ship = s->IsPlayerShip();
+
+    lua_pushboolean(l, player_ship);
+
+	LUA_DEBUG_END(l, 0);
+
+    return player_ship;
 }
 
 /* Method: SetShipType
--- a/src/Player.h
+++ b/src/Player.h
@@ -47,6 +47,9 @@ public:
 	virtual void StaticUpdate(const float timeStep) override;
 	sigc::signal<void> onChangeEquipment;
 
+    // XXX: isPlayerShip must be redefined
+    virtual bool IsPlayerShip() const override { return true; };
+
 protected:
 	virtual void SaveToJson(Json::Value &jsonObj, Space *space) override;
 	virtual void LoadFromJson(const Json::Value &jsonObj, Space *space) override;


laarmen:
May I ask what is the purpose of this? The overriden method isn't used anywhere AFAICT.

Also, as stated elsewhere, please add more details :-)

mike-f1:
I think is a step ahead for the issue #38 :
because if you don't inherit player from ship then you cannot use their type.

mike-f1:
But if not, I think you have to remove IsPlayerShip from DynBody...

laarmen:
Yeah, that'd be my approach. You'd have to change the interface anyway when
the ship /player divorce happens.

ecraven:
I don't think this is a good idea, with these changes we could have a Player for which IsPlayerShip returns false... We should go through all places that check whether a ship is the player and fix them to use `IsPlayerShip` before merging this.

laarmen:
@ecraven, which one exactly is the bad idea? The original one from the PR, or the IsPlayerShip removal?

For now I'm fairly against merging the PR, but that's mainly gut feeling.

ecraven:
I think this should be part of a larger effort to separate Player from Ship in the inheritance hierarchy. But that is a fairly large undertaking in itself :-/ The changes proposed in this PR to me seem to muddy things further, we would have two ways of checking for the player, inheritance and IsPlayerShip, which can easily disagree :-/

laarmen:
Thank you, that's exactly what caused my uneasiness (is that a word ?) with the PR.

I commend the effort, though.

ecraven:
Absolutely, the idea is very good, but I think we need to think about the general architecture, so we don't run into even more problems than we have already :-/

fluffyfreak:
To me this makes more sense than our existing stuff. Either the `IsPlayerShip` needed removing, or it needs fixing. This fixes it.

ecraven:
I'd advocate removing it (as it seems to be entirely unused) and re-introducing it when we get around to breaking the inheritance chain of Player to Ship.

mike-f1:
I think the problem is (...or I hope is ) not on these few lines (it works either ways...): the problem is if someone would start changing checks btw Player and Ship spread on the code here and there (also for a little commit), then he could start; else you put the whole thing (the "split" and the "checks" ) in only one commit.

laarmen:
I'm not sure I understand this right, but if I do, then when you split Ship
and Player you intend to put the knowledge of being player-owned in the
Ship object? If so, I think that is the wrong location for this
information, the Player should be the one holding the knowledge.

About the content of the PR itself, I'd *really* like it better if it
removed the method.

mike-f1:
@laarmen :  IMHO, ships should know if they are owned by player or NPC...

But even if not, 2 things should be be taken into account if you make player a totally different kind of class:
1) the type based check will not work.
2) all actual stuffs that rely on the fact that Player *IS* a Ship needs changes

About the second point I think something like:
IsPlayerShip() { return ( is_this_ship_own_by_the_player ); };
even as a temporary solution, could be enough to allow actual stuffs works AFTER the split because the check they do is already working (I mean if based on the method).
Then, you could think on how to avoid putting this knowledge on Ship...

That said, for the effort I put doing this change, I think I can remove both methods.

Let me know

ecraven:
> IMHO, ships should know if they are owned by player or NPC...

I am not an expert at game design, but I'd tend to disagree here. A ship should not do other things than be a ship. The Player class (or even better component) should do all that. Every time you need to know whether a Ship is a player inside Ship methods, you should have a component or function outside of Ship to call and handle that. At least that's what I read.

robn:
That's exactly what PlayerController (I think it was called that) is supposed to do. Ship, DynamicBody, etc should represent the physics object. *Controller is what links the physics into other systems, the UI for the player, the AI for NPCs, and so on.

(It was never that clean when I last worked on it, but that was the broad direction we were moving in).

fluffyfreak:
Closing since we chose to remove rather than augment this.

Issue #3331, opened 2015-01-21
Ships stuck in docking phase


clausimu:
There is something strange going on with ships set to dock (ship:AIDockWith(starport)). The AI triggers the event "onAICompleted", but flight status of the ship is stuck in "DOCKING". Either the docking procedure is not complete (in which case the AI should not trigger "onAICompleted"), or the flight status should be changed to "DOCKED". 

This does not happen for every docking event. I have not actually seen the ships stuck in docking (are they hanging in mid-air?). But I notice this because the game immediately crashes when I ask ships to undock that are stuck in this limbo. I can't figure out what exactly triggers this. Has anybody else seen it?

I'm currently using the 20150120 build for 64-bit linux.

corundscale:
I'm still out of dev process, but could it be similar to what happens to player when forcing time compression when landing using autopilot? The ship effectively seems to land on the pad, but the permission times out before that so it never actually completes docking.

clausimu:
That would be in keeping with my observations. I have only seen this when speeding up time. I will try another game where I don't speed up and just keep it running in the background. Let's see if the same happens.

I can also confirm that the ships that are "stuck" stay there for long. I have re-checked their status every minute and they just stay in that "non-docked" state for several minutes (I just aborted the script after that).

In a script modeling ~1000 ships in the Sol system I had 20 ships enter the "non-docked" state within 1.5 hours of game play. The problem is, over time, more and more ships get stuck in this state. Maybe it hasn't been noticed before because the traffic was kept low?

clausimu:
I just did another simulation where I did not speed up time. Within 5 minutes of (realtime) game play 3 ships got stuck. I was using the same simulation (~1000 ships in Sol) as before. 

I am playing with the 20150120 64-bit linux release.

clausimu:
Some more information collected in the last few simulations:

1. Ships get "stuck" in ground bases and starports in space.
2. The ships being "stuck" look like they have landed appropriately. For example, there were 8 ships "stuck" at Gates Starport. All eight had landed in an appropriate spot within the space station (looked pretty cool). They did not hover over the surface or were "stuck" in mid-flight or something like that. No thrusters visible. Position lights were blinking on each ship.
3. The stuck ships don't show up on your scanner when approaching the station where they are "stuck".

All that leads me to believe that the ships are actually docked all right - they just need to change their flight status.

If you want me to test anything else - just let me know.

fluffyfreak:
[...]

Issue #3290, opened 2014-12-29
Docking locators go bananas on the new station


bszlrd:
I'm playing around with the docking locators right now. @impaktor noticed that the malabar doesn't get landing clearance on the station.
At first I thought the game thinks it's larger then the 100m limit I give to the big pads, so I've upped them to 500m, without any effect.
They are set up this way:
there's a port entrance and exit for each area, p04 is for the large docks. I've made it so the port exit locators are right between the given pads, so it the ship doesn't go to the exit on rails, but starts to accelerate inside the station.
The pad locators for the large pads look this way: loc_C001_p04_s0_500_b*
I've started to play around a bit with the port part:
- p01 gives landing clearance, and puts it in the proper place when landed.
- p02, 03, 04 denies landing due to no available port.
- p05 puts it on a small pad (which belongs to p05). The small pads have a 33m limit.

Here's the mod for the station: https://dl.dropboxusercontent.com/u/456051/pioneer/orbital_station_A.zip

And here are 5 different docking daes, the only differences are the big pad port numbers in them.
https://dl.dropboxusercontent.com/u/456051/pioneer/orbital_errordocks.zip

Issue #2656, opened 2014-01-16
Landing pad use shield mesh as collision


NyankoSensei:
![1](https://f.cloud.github.com/assets/5476493/1926674/35862c12-7e48-11e3-91dd-e2530095085d.jpg)

Issue #1683, opened 2012-11-05
About 20% of trade ships is destroyed when landing and the timestep is big


irigi:
I observed this during my work at TradeShips.lua, up to 20% of ships fail to land when the timestep is big. The reason is IMHO obvious: autopilot cannot land with so high timestep. Proposed solution would probably be:
- detect if the timestep is big
- if yes, detect when the ship is in some distance X from the starport attempting to land
- let the ship disapper from current position and appear docked at the starport

I, however, do not know how to do this in lua in a best way.

richardpl:
I do not think I encountered trade ship destruction any more.

fluffyfreak:
Just tried it, over 12 days there were just 3 ships destroyed out of over 1300 events. Seems robust enough to me.

PR #3901, opened 2017-01-05
Slowly position ship to landing tag when landing on planetary bases.

### LINKED ISSUE #2678: Don't position and rotate the player to the dock marker position
I mean if the player lands by hand, it would be look more natural if the ship wouldn't  rotate and translate to the docking marker alignment. Similarly to landing on the ground.
So if you land by hand, then as soon as the ship reaches the pad collision, then the landing tag height of the ship matches the markers height (#2656), and it's close enough to the landing marker, then rights itself, but keeps the heading and position.

Is this possible, and would it worth the effort?

mike-f1:
Partially close #2678


laarmen:
Please, include more details in the git logs and the PRs. We shouldn't have
to go through hoops to find out what this is all about.

laarmen:
After glancing at the code, I have several stylistic/pedantic remarks.

1/ Contrary to most C++/Java code out there, our methods are capitalized: stageReset() => StageReset()
2/ I'd have split the commit into a patchset since there are quite a bit of changes regarding the SpaceStation headers that are not directly related to what you are fixing.
3/ Personally, I'm not a fan of putting getters and setters everywhere. I wouldn't have said anything if it were new code, but the replacement obj.ship -> obj.{get set}Ship really jumped at me as being a loss in readibility, as well as cluttering the diff.

walterar:
I like this. The final coupling maneuver is now as it should be.

mike-f1:
Well, I think is at least an improvement but after a talk with @nozmajner , he ask me to
avoid movement and rotation of the ship if you are near center (I don't know how much far: 1/3 of pad diameter?...). So I would try to improve it.

laarmen:
Cue to @impaktor for git history :-)

src/SpaceStation.h
```diff
@@ -121,27 +121,27 @@ class SpaceStation: public ModelBody {
                 VectorToJson(shipDocking, fromPos, "from_pos");
                 QuaternionToJson(shipDocking, fromRot, "from_rot");
             };
-            Ship* getShip() const { return ship; };
-            int getShipIndex() const { return shipIndex; };
-            int getStage() const { return stage; };
-            double getStagePos() const { return stagePos; };
-            vector3d getFromPos() const { return fromPos; };
-            Quaterniond getFromRot() const { return fromRot; };
+            Ship* GetShip() const { return ship; };
```
laarmen: Note for the future: should we explicitly mark these kind of methods as inline?

fluffyfreak:
Mostly happy with this, double-happy to see someone other than me tackling the docking system! 👍

fluffyfreak:
@mike-f1 I will try to find time to test this tonight.

Is there more you want to do to this PR before it's ready for review and merging?

mike-f1:
@fluffyfreak  : I promised @nozmajner to add a "no repos zone" instead of ever repos, this mean I should do some stuff... Now I have time: I could try to do it right now, so please check if I commit something.

fluffyfreak:
@mike-f1 no rush :)
Work on this stuff whenever you feel like, I will mark it as `WIP` for now.

src/SpaceStation.cpp
```diff
@@ -39,12 +39,8 @@ void SpaceStation::SaveToJson(Json::Value &jsonObj, Space *space)
 	Json::Value shipDockingArray(Json::arrayValue); // Create JSON array to contain ship docking data.
 	for (Uint32 i = 0; i<m_shipDocking.size(); i++)
 	{
-		Json::Value shipDockingArrayEl(Json::objectValue); // Create JSON object to contain ship docking.
-		shipDockingArrayEl["index_for_body"] = space->GetIndexForBody(m_shipDocking[i].ship);
-		shipDockingArrayEl["stage"] = m_shipDocking[i].stage;
-		shipDockingArrayEl["stage_pos"] = DoubleToStr(m_shipDocking[i].stagePos); // stagePos is a double but was saved as a float in pre-JSON system for some reason (saved as double here).
-		VectorToJson(shipDockingArrayEl, m_shipDocking[i].fromPos, "from_pos");
-		QuaternionToJson(shipDockingArrayEl, m_shipDocking[i].fromRot, "from_rot");
+        Json::Value shipDockingArrayEl(Json::objectValue); // Create JSON object to contain ship docking.
```
impaktor: Indentation

src/SpaceStation.cpp
```diff
@@ -167,7 +160,7 @@ void SpaceStation::PostLoadFixup(Space *space)
 {
 	ModelBody::PostLoadFixup(space);
 	for (Uint32 i=0; i<m_shipDocking.size(); i++) {
-		m_shipDocking[i].ship = static_cast<Ship*>(space->GetBodyByIndex(m_shipDocking[i].shipIndex));
+        m_shipDocking[i].PostLoadFixup( space );
```
impaktor: Also, this indentation looks wonky

src/SpaceStation.cpp
```diff
@@ -349,17 +336,17 @@ bool SpaceStation::GetDockingClearance(Ship *s, std::string &outMsg)
...
 		if( !pPort ) continue;
 
 		if( pPort->minShipSize < bboxRad && bboxRad < pPort->maxShipSize ) {
 			shipDocking_t &sd = m_shipDocking[i];
-			sd.ship = s;
-			sd.stage = 1;
-			sd.stagePos = 0;
+			sd.StageClearanceGranted( s );
+			sd.EndRepos( false );
+			sd.EndRerot( false );
```
impaktor: Do we typically use spaces in the `(...)` like this? Looks odd to me, but I could be wrong. Worth checking out elsewhere in the codebase. Just do as others are doing.

src/SpaceStation.cpp
```diff
@@ -437,68 +422,86 @@ bool SpaceStation::OnCollision(Object *b, Uint32 flags, double relVel)
...
-
-		double stageDuration = (dt.stage > 0 ?
-				m_type->GetDockAnimStageDuration(dt.stage-1) :
-				m_type->GetUndockAnimStageDuration(abs(dt.stage)-1));
-		dt.stagePos += timeStep / stageDuration;
+		if ( !dt.GetShip() ) continue;
+		// docked stage is m_type->NumDockingPorts() + 3 => ship docked
+		if (dt.Stage() > m_type->NumDockingStages()+2) continue;
+
+        // if in reposition, stageDuration is 1 sec
```
impaktor: Nitty picky comment on comment not being indented fully. (Meta?)

src/SpaceStation.cpp
```diff
@@ -509,27 +512,64 @@ void SpaceStation::DockingUpdate(const double timeStep)
...
-	assert(dt.ship == ship);
-
-	ship->SetPosition(GetPosition() + GetOrient()*dport.pos);
+	PiVerify(m_type->GetDockAnimPositionOrient(port, dt.Stage(), dt.StagePos(), dt.FromPos(), dport, ship));
+	assert(dt.GetShip() == ship);
 
+	matrix3x3d wantRot = matrix3x3d::FromVectors(dport.xaxis, dport.yaxis, dport.zaxis);
+    matrix3x3d ssPadOrient = GetOrient() * wantRot;
+    vector3d ssDockingPos = GetPosition() + GetOrient()*dport.pos;
 	// Still in docking animation process?
```
impaktor: indentation.

mike-f1:
I had a wrong setting on my editor, now it's ok. I'm sorry for the troubles

mike-f1:
Sorry guys: I changed everything but without a true reason :-/
Now the code is more like the original one: you'd better compare the last version
with the original code...

I'm stuck trying to understand when and where the positions are initialized when you launch
a new game, could someone give me a clue?

And last thing: when ship will aren't repositioned they are "floating" over pad,
I don't think is something related to the code I did...

PS: How to toggle "Savegame bump" label?

laarmen:
Labels can only be assigned by team members. As for the rest, I cannot help
you. I likely won't be able to even glance at the code until next week.

mike-f1:
Think this time it's ready...

impaktor:
Just throwing it out there that I'd like to see a rebase of the commits (see pioneer wiki on git usage), alternatively I'll do it when I merge it. I haven't reviewed the actual code (yet).

walterar:
This is awful. I do not like it now. The first sequence is the best for me.

nozmajner:
I think it's pretty nice, when the ship actually touches the pad:

mike-f1:
@walterar: The initial request was #2678, and so is intended to work: If you land near center your position will remains, else you have "ground staff" repositioning your ship, and an additional "onShipBadDocked" LUA event is triggered (but this is my first time on LUA side, then I may ask you if things are working on this side?)

walterar:
@mike-f1 This does not work for me, sorry.
Unfortunately I lost the files of his first work (SpaceStation.cpp and SpaceStation.h), which did work as I think is correct. If you still have them, can you offer them to me in a zip?

mike-f1:
they are on git: https://github.com/mike-f1/pioneer/commit/6dc78bfe3baa334f642448f396cd34d0ee6ad080 
...but have you tried to land little more far from center of pad?

walterar:
I'm not working with git now. I need the files SpaceStation.cpp and SpaceStation.h that you posted on 6 January 2017. It may be possible?

nozmajner:
@walterar: you can click on the view button, top right to each file, then the raw button to download them.

walterar:
That's the way I do it, but in mike-f1@6dc78bf I only get SpaceStation.cpp

mike-f1:
https://github.com/mike-f1/pioneer/commit/7fecac5675c261309b1e8c5a91cd564b8a63a136
 For *.h

mike-f1:
@walterar : I think it's better if you remain on "master" and on my last commit change line
454 on SpaceStation.cpp:
					if (dist > dt.maxOffset*dt.maxOffset) {
with:
					if (dist > 0) {
so your ship are ever repositioned

walterar:
@ Mike-f1 Thanks. Mike-f1@7fecac5 contains the correct docking sequence for me.

mike-f1:
@walterar : I said that because my previous code was... a mess :-P
This works better and IS better, I don't know how much or if it's enough to have this PR approved, so I mention both ways... And more: in line 454 you decide how far you could be to trigger reposition, so you could try yourself playing with values.

How about "onShipBadDocked"? Did you see it?

walterar:
@Mike-f1 My opinion is based on the results seen by the player, not in the code. :)

walterar:
@mike-f1 In the last proposed code, if we modify the line 454 if (dist> 0) { (as you have suggested)
The docking sequence is completed successfully.

mike-f1:
Hi all,
with this update you no longer have ships transferring from pad to pad: the way I did it is checking distance between your ship and your pad.
This is enough but my thought is that we do things 2 times (because you already traverse all things to check for collisions and could be better if you collect the pads on which you're colliding and)...

I also leave 2 debug message to  allow all to see how it work:
OnCollision: dist sqr = 11.181550 (over max sqr = 100.000000)
DockingUpdate: dist sqr= 11.181550 (over sqr 8.108373)

the first check is to disallow collision when distance is less than 100 (this is maxShipSize/2)
the second is the trigger for reposition (roughly maxShipSize/2 - Ship_radius )

Let me know

impaktor:
Note: I have not reviewed the actual code, I leave that to @fluffyfreak, nor have I tested it. My "approve" was due to indentation fixes.

Also, I'd like to see a rebase before merge. First commit should have a propper description, and those commits that just fix stuff (like indentation, or add/remove debug info) should be squashed with the commits that we want to keep as separate.

As I noted above, how to do this is described [here](http://pioneerwiki.com/wiki/Using_git_and_GitHub#Fixing.2FUpdating_your_pull_request).

Alternatively, I can rebase it at merge, after @fluffyfreak gives 👍 after reviewing the code.

walterar:
I do not understand what they want to do. Everything is very confusing and I suspect that the end result sought is not appropriate for my way of seeing a correct docked with autopilot.

nozmajner:
@walterar It's not for the autopilot, but for the player.
If the player docks a little bit imperfectly, then the ship stays as it landed: not on the center of the pad, and/or not perfectly aligned. If the landing was so bad, that it was on the edge of the pad, then it will be repositioned (by ground crew for example) and rotated to fix the big mistake. So landing is a bit more dynamic, and less on-rails: like the SpaceX landing yesterday:

Not perfectly in the middle of the pad, but good enough to avoid flipping the barge.
Little irregularities like this can insert some life into the game.

The autopilot landed very close to the center, and almost perfectly aligned in most cases while I was testing this.

mike-f1:
...the rest is because I never used github and I'm not a professional coder, so I did more things wrong :-P

walterar:
Manual docking on a space station? In the 31st century? As was done in the 20th century? :)
Well, I guess there will be pilots who like to do it like that, like a high-risk sport. But the NPCs will be in trouble. :(

I think the autopilot docking should first be fixed, which broke when the new orbitals were included. And then make players who want to do it, can docking anywhere in the station, even on the roof if you want, just like on the surface of any planet.

nozmajner:
Yes, but then you could argue about pilots flying ships in general, because that will be done by AI most likely in the 31th century. Or in a hundred years really. 
In this case:  either take away stuff from the player, like proper manual landing, or make it feel a bit more alive and less restricted.

And this interfering/not interfering with NPC ships is a completely different argument really. Why blend the two issues, design vs potential problems together?
They should be able to land all the time for sure, and that needs to be verified if it works.

fluffyfreak:
@mike-f1 I think you'll want to revert/delete that last commit c63952e as you have the remains of a bad merge pushed.

mike-f1:
So, thank's @fluffyfreak for your patience, I hope to find an answer this week and get it to work.

mike-f1:
@fluffyfreak : I tried my best rebasing by myself (also for learning purposes...) and to me seems all good, but let me know if I left some mistake.

Anyway I need to keep this PR on WIP because I asked @nozmajner to model flat pads (in order to avoid ships floating above pads when docked), and this will lead to problems with ships not docking anymore :-(

Yesterday I had a talk with JohnJ and he want to take a glance to better understand what is happening here...

as for reference I found these:
https://github.com/pioneerspacesim/pioneer/issues/2656
https://github.com/pioneerspacesim/pioneer/issues/3290
https://github.com/pioneerspacesim/pioneer/issues/3331

...what to say... I'm unlucky :-/

fluffyfreak:
@mike-f1 well #2656 and #3290 are both fixed already at least :)

mike-f1:
Yes, but for fixes they needed some changes on pad meshes... So they could be related to that.

@fluffyfreak : I'm wondering how actually "collisions" on pads works: are landing tags used or not?
and also: there's a specific need to have "volume" collision on pads? Could they be a simpler surface?

fluffyfreak:
@mike-f1 no need for it to be a volume, we used to use two planes/rectangles crossed in the middle like a `+` and that worked well too. The current collision geometry was just what we switched to using when these new stations were built.

It's been more than a year since I looked at the code so I'll need to refresh my brain on how it all worked.
It does work differently for space stations and ground stations.
Ground stations you must touch the pad collision geometry and then do that tiny positioning.
Space stations you touch the collision geometry at the entrance and it then does the whole animated docking thing.
It's all orchestrated through `SpaceStation::OnCollision` I think.

fluffyfreak:
Also what do you mean by `landing tags`?
Some geometry gets a flag set to identify it as landing-collision geometry, that flag gets tested in `SpaceStation::OnCollision`. That's based on the meshes name and is fixed up in `Loader::GetGeomFlagForNodeName`.

mike-f1:
tag_landing on models, excuse me :-/ (ref here #2656)

fluffyfreak:
Ah yeah, that's explained here https://github.com/pioneerspacesim/pioneer/blob/master/src/Ship.cpp#L279 :)

mike-f1:
Yeah, I saw the line but not the comment.

Thank you and excuse me for ContinuoslyAskingQuestions

jaj22:
There's a general docking problem caused by the Aabb not matching the ship dimensions (or even close), probably with some per-ship variation. For example, the Sinonatrix's Aabb extends 2m below the collision mesh. Apparently this was worked around by extruding the pad collision surfaces upwards, but the core problem remains. Looking into it at the moment, although I don't know the scenegraph code very well.

mike-f1:
@jaj22, @fluffyfreak : I think I found the problem...
in ShipAICmd.cpp, line 966:
m_dockpos -= m_dockupdir * (m_ship->GetAabb().min.y + 1.0);
should be:
m_dockpos -= m_dockupdir * (m_ship->GetLandingPosOffset() );

and then you could delete "+1.0"... So: the problem with AABB still here, but we have a workaround
that could work while AABB will get fixed

I also tried with many ships and the last line of code work almost for all ships even at 10000x,
the ships for which this don't work are:
Venture Star
Bowfin fighter
Xylophis
Varada
...Could be for these ships we don't have any tag_landing locator?

nozmajner:
Only the Venturestar/AC33 and Deneb are the ships that doesn't have a landing tag, but a bit of collision mesh sticking out from the belly of the ship.

mike-f1:
Yes, I think this is the way: you need also a "collision" in order to trigger change on status :-)

mike-f1:
Pumpkinseed have collision around gears...
Sinonatrix have a collision protrusion between rear wheels

Kanara works, and it have collision mesh arond front wheel

jaj22:
This is all just workarounds. If the aabb is wrong then anything that depends on the aabb will have the wrong behaviour, and so there will be further problems in the future.

nozmajner:
Bowfin and Varada and Xylopshis doesn't, not even animated.
And a strange thing: the Xylophis lacks collision model file for some reason, but the model file references one, and the actual collision mesh is in one of the shape dae-s.

mike-f1:
@jaj22 : as @fluffyfreak refer above, there's code to handle this in Ship.cpp:281

So I'm thinking if it is the case to open (another :D ) PR for this, in order to trace all things and have some clue... Also because this is not first time this happens (as referred above) and it could happen
again...
Also (and I don't want to complain) this PR is frezeed for things that aren't related to the "topic" itself

mike-f1:
@nozmajner : May I ask if you could check what happens if you put a collision mesh on a ship that hadn't one or if you tweak an existing collision mesh on a "not landing" ship?

mike-f1:
As for reference:
https://github.com/pioneerspacesim/pioneer/issues/1683

nozmajner:
@mike-f1 : it seems to be working, at least I tried with a Varada. But I had to move that little collision part a bit lower than the tag_landing to make it work. I'll try with the others tomorrow.

mike-f1:
And, after a *small* fight with git, I think I did all correctly...

I also think you could remove "Savegame bumps" label and "WIP"

nozmajner:
Don't know about the Savegame bump tag, so I leave that to the others. Either way, that's not a showstopper I think.
I'll fix up the 6pad station collision meshes for this ASAP.

mike-f1:
I changed a couple of things:
- improved calculations for triggering reposition,
- changed size limits for pads
- other small things...

src/Ship.cpp
```diff
@@ -669,6 +669,12 @@ void Ship::SetFlightState(Ship::FlightState newState)
 		m_launchLockTimeout = std::min(std::max(2.0, 2.0 * (GetMass() / MASS_LOCK_REFERENCE)), 20.0);
 	}
 
+	if (newState == DOCKED) {
+		m_launchLockTimeout = 0.0;
```
laarmen: Out of curiosity, is it really necessary to zero it out, or just for the sake of consistency?

mike-f1: Oh: as now you could pass through docking surface, and it happens you "land" instead... Then if you ask for permission take-off your thruster remains active due to the "if" in ClearThrusterState.

So I would avoid this. But in general you needs not to dock again if your lock timeout is not expired...

So I leave it

src/SpaceStation.cpp
```diff
@@ -440,8 +446,50 @@ void SpaceStation::DockingUpdate(const double timeStep)
 	for (Uint32 i=0; i<m_shipDocking.size(); i++) {
 		shipDocking_t &dt = m_shipDocking[i];
 		if (!dt.ship) continue;
-		// docked stage is m_type->NumDockingPorts() + 1 => ship docked
-		if (dt.stage > m_type->NumDockingStages()) continue;
+		if (dt.stage > m_type->NumDockingStages()) {
```
laarmen: So, let me see if I understand this correctly : we have some sort of correspondance between the ship docking stage number and the station docking stage number, up until a certain point where the station number goes higher to encode the substeps needed for the smooth land ?

mike-f1: Yes: I did it as "extrastage" because I don't know how many stages there are. Orbitals encode each step between waypoints, then your ship are pos is interpolated; surface stations don't have any waypoint

fluffyfreak:
@mike-f1 this now conflicts with the changes made in #3920 I think.
Would you be able to get latest, and fix the conflicts please?

I am hoping to merge this and the other savegame bump tagged PR's soon.

mike-f1:
Hi, I did it: I removed conflicts and I purified history... Sorry for the late, but my hard-disk has gone, so last days I tried to get my pc back to work.

fluffyfreak:
Great, thanks @mike-f1, good luck getting your PC sorted out.

PR #3903, opened 2017-01-08
Do not spawn unarmed pirates

robothauler:
Closes #3902 
I did not touch the police pumpkinseed. Some say it could be a corrupt police officer or it's a stolen police craft.

PR #3906, opened 2017-01-15
Choose enemy ships by role, not hull mass.

robothauler:
During discussion about #3903 it arise that a role field could by handy.

I split the pirates into two roles, pirate and mercenary. Mercenary includes bigger ships like the Nerodia or Deneb. Which reflects, more ore less, the current separation <=150t or <=400t hull mass.
 But a ship type can have both roles for future use.

### DIFF
--- a/data/lang/module-cargorun/en.json
+++ b/data/lang/module-cargorun/en.json
@@ -383,6 +383,10 @@
       "description" : "custom cargo",
       "message" : "Plutonium"
    },
+   "POLICE" : {
+      "description" : "",
+      "message" : "Police"
+   },
    "QUIBBLES" : {
       "description" : "custom cargo, also used in the NewsEventCommodity module, a nod to Star Trek's Tribble (check wikipedia)",
       "message" : "Quibbles"
--- a/data/lang/module-cargorun/en.json
+++ b/data/lang/module-cargorun/en.json
@@ -383,10 +383,6 @@
       "description" : "custom cargo",
       "message" : "Plutonium"
    },
-   "POLICE" : {
-      "description" : "",
-      "message" : "Police"
-   },
    "QUIBBLES" : {
       "description" : "custom cargo, also used in the NewsEventCommodity module, a nod to Star Trek's Tribble (check wikipedia)",
       "message" : "Quibbles"
--- a/data/modules/CargoRun/CargoRun.lua
+++ b/data/modules/CargoRun/CargoRun.lua
@@ -608,9 +608,8 @@ local onEnterSystem = function (player)
 			-- if there is some risk and still no pirates, flip a tricoin
 			if pirates < 1 and risk >= 0.2 and Engine.rand:Integer(2) == 1 then pirates = 1 end
 
-			-- XXX hull mass is a bad way to determine suitability for role
 			local shipdefs = utils.build_array(utils.filter(function (k,def) return def.tag == 'SHIP'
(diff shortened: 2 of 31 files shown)


impaktor:
If I read the code right, only pirate role is actually used?

I'm wondering because you have police role as well, and this is actually set by each faction. I.e. it is up the the faction to define which ship model the police should use, rather than the ship.

impaktor:
It's just that you set "police" role for ships but this is not used anywhere, which might be confusing, since police ships are set in faction. It feels a bit like dead code, or dead config line, that gives the ship modellers the false idea that they can make a ship a police ship by setting that line.

Also, we  must already have `l.POLICE` tag, or?

fluffyfreak:
@impaktor and @robothauler this looks good to me now.

impaktor:
I haven't merged it as I was hesitant about this. I'm not convinced it's the "right way" to go. I don't know what the "right way" is, but it feels slightly odd to me that we on one hand let factions set which police ship to use, and with this PR we make it the ship's responsilility to set who use them. I.e. we're setting the same thing from two different directions:

- Factions decide which ship model to use for police
- Ships decide who use that ship model if pirate, mercenary, merchant (but not police)?

Having said that, if others are fine with merging it, I have no formal objections. It does remove repeated code for filtering "pirate ships" in mission scripts, which is good, I guess.

Maybe @laarmen want's to chime in?

nozmajner:
I think it has merits, since this way a ship mod(der) can decide if the given ship should show up as pirate (and etc), without touching any code.  
Police ships are different in that regard, that they are picked by the faction specifically. But what if police ships would be picked the same way? Like a police_ciw, police_solfed, police_independent roles, and the factions pick accordingly. Could be more mod friendly, because it would be easier to add to the police pool for example.

fluffyfreak:
I think it's ok for the factions to pick any ship as their police ship, but for ships in general to have a defined "role". I don't actually see anything odd about that.

laarmen:
*appears in a puff of smoke*

**You summoned me?**

Let's draw a parallel to cars, because it is well known that car metaphors
in software are great, until they aren't.

Each car has several roles it is intended for : utilitarian vehicle, urban
mobility, military transport, luxury transportation for the rich and
famous, etc... Also, each car has a manufacturer. Usually, when a PD or an
army tries to find a new model for their motorcade, they start by selecting
a subset of manufacturers they want to work with, because, say, they are
local brands, or whatever. And then from this subset of manufacturers they
look at the models fulfilling whatever role they want.

Why don't we copy this behaviour? Each ship already has a defined
manufacturer AFAIK (except maybe the space shuttle?). The factions just
need to define a preferred manufacturer, and voilà.

fluffyfreak:
@laarmen I kind of like that idea, but does it stop a faction from using whatever ship it likes?
I mean in the UK the police just pick a car and use it, regardless of suitability.
https://s-media-cache-ak0.pinimg.com/originals/66/12/54/661254b9b0fe7925cadf041074c505a6.jpg

So I kinda think it's ok as-is.

laarmen:
We can hope for institutional sanity in our dystopian future, can't we?

Le mar. 21 févr. 2017 à 13:34, Andrew Copland <notifications@github.com> a
écrit :

laarmen:
For the record, the IRC consensus was "let's merge this, we'll worry about a better information design later on." And I ended up wondering why the hell were the factions defined in .lua files which were interpreted in their own, separate context. Which led us to sysgen. Which is the path to insanity.

PR #3908, opened 2017-01-16
Add limits when using autopilot SetSpeed

mike-f1:
nothing more, nothing less...

If you are in SetSpeed mode (flying and with autopilot enabled) if you hold the keys
to increase or decrease speed for more than 5 sec, you ask for this:
https://www.youtube.com/watch?v=ygE01sOhzz0

...So I put limits on playerController.cpp on what you could ask to autopilot

### DIFF
--- a/src/ShipController.cpp
+++ b/src/ShipController.cpp
@@ -255,10 +255,14 @@ void PlayerShipController::PollControls(const float timeStep, const bool force_r
 			}
 
 			if (!stickySpeedKey) {
-				if (KeyBindings::increaseSpeed.IsActive())
+				if (KeyBindings::increaseSpeed.IsActive()) {
 					m_setSpeed += std::max(fabs(m_setSpeed)*0.05, 1.0);
-				if (KeyBindings::decreaseSpeed.IsActive())
+					if ( m_setSpeed > 300000000 ) m_setSpeed = 300000000;
+				}
+				if (KeyBindings::decreaseSpeed.IsActive()) {
 					m_setSpeed -= std::max(fabs(m_setSpeed)*0.05, 1.0);
+					if ( m_setSpeed < -300000000 ) m_setSpeed = -300000000;
+				}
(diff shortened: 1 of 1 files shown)


impaktor:
> If you are in SetSpeed mode (flying and with autopilot enabled) 

If autopilot enabled, can one change speed then? Aren't "set speed"/Manual and "auto pilot" mutually exclusive? Sorry, my pioneer-memory is getting rusty.

So idea is you shouldn't be able to (approximatly) go faster than light?

I don't see any "5 second" rule in the code?

Well, those we just my thoughts.

For user interface, I pass the ball to @nozmajner and @ecraven. I'm fine with it if they are.

nozmajner:
I'm not sure how important is this, because most ships can't reach even 0.1c, but it sure looks better having a limit. And you could still go over c manually firing your thrusters. :D (I know, I tried)

mike-f1:
@impaktor I mean if you (on master) start a new game, takeoff, then F5 to enable autopilot... So, as now, if you hold the "carriage return" as long as you wish there's no limit on the speed you could set...

Then I choose to put a (even large, if you want...) limit for the speed.
Idea us you shouldn't be able to ASK a speed faster then light (and yes: I choose speed of light as speed limit)

mike-f1:
...I think we could put "UI" label on this commit...

mike-f1:
thx's

mike-f1:
@nozmajner : Excuse me: 
1) if you set speed to 123456789123456 Km/s, then you need 2 lines:
speedissue

2) if you want to change speed then you ever are inside a range

nozmajner:
That's actually three lines in your screenshot. 
I think it's a good change, but maybe @ecraven wants to chime in, since he's working on that part of the UI. Just to make sure this won't be a wasted effort in a few months when the flight UI becomes gutted and replaced. But I guess having a limit of c then won't be a problem really.

Issue #3910, opened 2017-01-17
Collision mesh bounding boxes include non-collision meshes


jaj22:
The Aabb (axis-aligned bounding box) built by CollisionVisitor currently includes all visible geometry, including shield meshes which extend well beyond the collision mesh. As far as I know, all usage is for physics purposes, so this feels like a bug. Last time I checked, only the radius was used for rendering. The Aabbs (and radius) also don't appear to take account of animation, although that's a much trickier problem and not an immediate one.

The docking AI routines assume that the Aabb describes the extent of a ship's collision mesh (which was true before the new model system), so the landing position is typically far too high, with ship dependence. This is currently worked around by extruding the collision mesh of the landing pads upwards, which would be unnecessary if the landing position was correct.

There are two main directions for a fix:

1. tag_landing could be used instead of Aabb.min.y for determining landing positions, although this is more error-prone than using the true collision mesh extent, and requires modellers to add the tag for the game to function. If people do prefer this solution then I'd recommend removing the Aabb entirely, as it will have no useful functionality.

2. Alternatively, changing CollisionVisitor to only include collision mesh geometry in the Aabb is a trivial function removal. However, the same Aabb radius is currently used for physical and render/clip radius (by different routes), so this functionality would need separation. Note that for performance reasons, physical and clip radius should ideally be separated even if tag_landing is used for the landing offset.

nozmajner:
The point of tag_landing as far as I know is to indicate the lowest point of the lowered gears, which doesn't have animated collision meshes in most ships. Do the animated collision meshes taken into account for generating the bounding box?

fluffyfreak:
@nozmajner is correct, it's to avoid using the collision bounding box for determining the lowest part of the ship which might not be the feet depending on animations etc. `tag_landing` is therefore just to position the ship a distance about the pad.

I don't think we take animated meshes into account when generating the AABB.

Removing the shield from the collision system could be good, honestly that whole thing was a fucking ball ache to work on. (_pardon my language_), It would be good to toggle collision with a shield on and off, it would also help the shield impact rendering effect to look better.

jaj22:
> The point of tag_landing as far as I know is to indicate the lowest point of the lowered gears, which doesn't have animated collision meshes in most ships.

Are there currently any ships which do have animated collision meshes for landing gear? Because that would make this an immediate problem by breaking option #1.

nozmajner:
@jaj22 : Lunar shuttle.

[...]

PR #3911, opened 2017-01-17
AIDockShip could use tag_landing

mike-f1:
Because tag_landing are added to models, then you need to use them
when you use autopilot.

This let autopilot use them because GetLandingPosOffset() return
tag_landing.y if it's present, else return GetAabb().min.y as usual
(check ship.cpp:280)

Before merge this, there's a need to tweak ships collision geometry
to reach the eight of tag_landing

Note: seems my editor do lot of changes, but the only interesting line
is 966...

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -1,4 +1,7 @@
 January 2017
+   * New Features
+     * Toggle cursor position when flying (#3909)
+
    * Internal Changes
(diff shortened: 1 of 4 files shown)


mike-f1:
https://github.com/pioneerspacesim/pioneer/issues/3910

https://github.com/pioneerspacesim/pioneer/pull/3901

PR #3917, opened 2017-02-01
Modelviewer use correct names tag_gunmount_XXX

mike-f1:
ModelViewer actually use wrong tag names for gun attachments,
as in line 251 (sorry: my editor still delete tabs...).

Correct names are as in wiki:
http://pioneerwiki.com/wiki/Model_system#Attachments

and they should be used in code.

...I think there's no need for review because I tested it and it
work with actual models, I think is mergeable.

### DIFF
--- a/src/ModelViewer.cpp
+++ b/src/ModelViewer.cpp
@@ -107,7 +107,7 @@ ModelViewer::ModelViewer(Graphics::Renderer *r, LuaManager *lm)
 
 	m_log = m_ui->MultiLineText("");
 	m_log->SetFont(UI::Widget::FONT_SMALLEST);
-	
+
 	m_logScroller.Reset(m_ui->Scroller());
 	m_logScroller->SetInnerWidget(m_ui->ColorBackground(Color(0x0,0x0,0x0,0x40))->SetInnerWidget(m_log));
 
@@ -248,9 +248,9 @@ bool ModelViewer::OnToggleGuns(UI::CheckBox *w)
 
 	m_options.attachGuns = !m_options.attachGuns;
 	SceneGraph::Model::TVecMT tags;
-	m_model->FindTagsByStartOfName("tag_gun_", tags);
+	m_model->FindTagsByStartOfName("tag_gunmount_", tags);
 	if (tags.empty()) {
-		AddLog("Missing tags \"tag_gun_XXX\" in model");
+		AddLog("Missing tags \"tag_gunmount_XXX\" in model");
 		return false;
 	}
 	if (m_options.attachGuns) {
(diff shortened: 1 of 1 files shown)


ecraven:
@mike-f1: this PR is incorrect, the tag_gun_ is *meant* for something different than tag_gunmount... I don't think this PR should be merged.
tag_gunmount is restricted and can only be used for one forward and one backward gun. tag_gun_.. is meant for more guns, as I understand it. I would rather we fix the system and support multiple guns correctly.

mike-f1:
...so restricted nobody knows about it :P : models actually have "tag_gunmount_XXX" and no "tag_gun_XXX", and this is the reason because it works with actual models...

In my opinion this *needs* to be merged because it's only a fix for actual code, then
you could decide if we should use old names and remove limits (if presents...) or add new one's...

As stated in IRC, actual tags ("tag_gunmount_XXX") are used in InitGun for projectile positioning (Ship.cpp, line 224)

mike-f1:
...or also for a "soft transition" use both

fluffyfreak:
@mike-f1 this is known though it's not written down anywhere, we were discussing it in IRC the other day and @ecraven has an interest in this area.

I'm going to close this for now because:
- The tags on the DSMiner point in the wrong direction.
- We don't have support for using them.
- We don't have a variety of models for use with ships.
- There's interest from others in using this as the basis for a more complex system.

mike-f1:
@fluffyfreak : don't know, to me (as I said) it look as a fix...

But in IRC, if I'm not wrong, @jaj22 ask me to provide code to support multiple guns per direction, and so
I did it: https://github.com/mike-f1/pioneer/tree/multiple_guns
...I wasn't aware of "complex system" you wish to add, but I (hope I) could do it, if someone
tell me how it is supposed to work.

Someone could give me a clue?

fluffyfreak:
@mike-f1 that looks like a great branch, very interesting, I didn't know @jaj22 and you were discussing or doing that.

However the only ship with a `tag_gun_left/rightXXX` setup is the `dsminer` and it only has those tags as an experiment. All of the other ships use `tag_gunmount_XXX` to place the origin(s) of the "laser" fire from the ships.

The `tag_gun_XXX` setup is meant to be completely separate, deliberately, so that changes can be made to ships without interfering with the existing `tag_gunmount_XXX` feature.

It is a WIP feature from the older alpha days when we did things a little bit differently.
If you're interested in working on it then there's more to consider than just enabling it in code, we need to consider the art impacts as well. This is probably best discussed on the forum and/or IRC.

mike-f1:
On IRC @jaj22 encourage me to move on: then I change code to handle a vector instead of a simple array[2]... It was (more or less (...as usual in programming)) trivial, also because all code is now in a shorter file thus it's more manageable.
It is working C++ side, although it needed some utility functions + to remove limits (LUA side?) when you buy guns + UI functions for placing/removing guns in the right slot... If you have also to add tags_* on models and have to start from scratch for LUA and for UI ...only to obtain... "fixedguns"? ...err...

Well, I think there are other improvements I could do before doing this.

Thank you for your answer @fluffyfreak

laarmen:
You might want to hold off on the UI bits =)

Issue #3913, opened 2017-01-25
CTD asking DistanceTo in LUA


mike-f1:
![issueinluatrade](https://cloud.githubusercontent.com/assets/13293916/22294126/5f9417ee-e312-11e6-9fb7-f77b46ce6b5e.png)

### Observed behaviour
Results are documented in attached picture

### Expected behaviour
Trade ships ask DistanceTo nearest starport and then
decide if hyperspace or reach starport

### Steps to reproduce
try to kill a trade ship: you could use attached save game
[Test_ShipKill_SaveLoadError.zip](https://github.com/pioneerspacesim/pioneer/files/729964/Test_ShipKill_SaveLoadError.zip)

Pioneer version: January 2017, on Linux/Ubuntu x86_64

walterar:
@mike-f1 This will solve the problem.

```
--- /old/TradeShips.lua
+++ /new/TradeShips.lua
@@ -786,7 +786,7 @@
 	if trader.no_jump ~= true then
 		if #starports == 0 then
 			trader['no_jump'] = true -- it already tried in onEnterSystem
-		elseif Engine.rand:Number(1) < trader.chance then
+		elseif trader.starport and Engine.rand:Number(1) < trader.chance then
 			local distance = ship:DistanceTo(trader.starport)
 			if distance > 149598000 * (2 - trader.chance) then -- 149,598,000km = 1AU
 				if getSystemAndJump(ship) then
```

mike-f1:
@walterar  It work! :) ...this needs to be pushed in master I think

mike-f1:
@walterar , do you want me to open a new PR for this?

walterar:
@mike-f1 Yes, do it yourself, I do not get along with github, excuse me.

mike-f1:
Ok, this needs to remains open because we don't know exactly what it happens...

Needs further investigations

impaktor:
Has been fixed on master.

mike-f1:
@impaktor : you're right, this has been fixed on master but with a "patch" (as we wrote in related PR).

The real reason why this happens is not yet investigated, so I suggest to 
left this open until we ended up with solution.

PR #3919, opened 2017-02-02
Avoid CTD when trade ship try to decide hyperjump

mike-f1:
Solve #3913

...Seems LUA would know a distance to a "nil" object, but I don't
know: you need to ask @walterar , he did the work ;)

### DIFF
--- a/data/modules/TradeShips.lua
+++ b/data/modules/TradeShips.lua
@@ -786,7 +786,7 @@ local onShipHit = function (ship, attacker)
 	if trader.no_jump ~= true then
 		if #starports == 0 then
 			trader['no_jump'] = true -- it already tried in onEnterSystem
-		elseif Engine.rand:Number(1) < trader.chance then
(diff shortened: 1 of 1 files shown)


impaktor:
Do we/you know why a trade ship ends up with no startport when trying to jump? And what happens with it if it doesn't have a starport set?  It will just linger in the system? Clog up stations? Just for curiosity, I wonder if we understand why this happens?

mike-f1:
@impaktor: err... just see the screenshot on #3913, there are line number and a little explanation...

Regarding what TradeShips are doing all time... yes: this is a good question :)

mike-f1:
@impaktor: err... just see the screenshot on #3913, there are line number and a little explanation...

Regarding what TradeShips are doing all time... yes: this is a good question :)

fluffyfreak:
I'd prefer to merge this and then investigate the consequences afterwards as currently there's a crash bug ingame

PR #3923, opened 2017-02-10
Skip children of geom & "Hole" for SpaceStations

mike-f1:
This PR would add 2 feature:
1. Geom could have a parent-child relationship, and this is used to avoid collision detection on related Geom's
2. Geom could have "an hole": a cylindrical volume in which collision detection is avoided

These methods are used in CollisionSpace::Collide where you do a first loose check.

All these features are used in SpaceStation, in which some changes happens:
some parameter could be added in Json file to specify dimensions for this cylinder
and if an "hit" on that must trigger docking sequence.
More, in SpaceStation I provided methods to put ships (or better geoms of ships) in the not-colliding
list of SpaceStation, and I changed Ships to have collision even when docked.

So: parent-child relation could be useful in order to implement ships mounts, then you
could avoid ship-hull deeply checked with ship-gun or with ship-generator.

Rationale for "hole" on SpaceStations instead is because all ships (even AI ships) docking
on orbitals will be collision checked and this replace an expensive deep check with
simple math. But I think could be better if someone could do a profiling to have a confirm


ecraven:
Wouldn't a simple implementation of [CSG](https://en.wikipedia.org/wiki/Constructive_solid_geometry) be a better and more general solution? This seems way too specialized to me (but I'm prone to implementing general solutions :-/)

mike-f1:
I think you could view this as the first brick: whole model (an orbital), in which you subtract a centered cylinder.

mike-f1:
PS: During tests it seems a collision with a docked ship is more intense and lead to explosion of the
lighter ship very fast... I don't know why this happen and I would investigate the reason.

Just tell me if there's a need to merge this because a way is to disable (again) docked-to-docked ship collision.

mike-f1:
The issue was beacuse onCollision on SpaceStation return false, so if you don't bounce back then damage happens every time step resulting in very fast explosions...

Solved: no needs for WIP, ready to be merged

fluffyfreak:
We have some built-in profiling that can be enabled by defining `PIONEER_PROFILER` during compilation, this can be triggered on a per-frame basis by pressing `Ctrl+Shift+p`.

So when doing optimisation stuff it's always a good idea to do a build with that enabled, then disable VSYNC in the options, profile and save the generated HTML file somewhere.

Then you can implement your optimisations and measure the actual change in performance over time.

This PR git history is messy, it needs updating to the latest master (_`git fetch --all -p`, `git pull upstream master`, etc_), the conflicts need resolving and then the commit history needs cleaning up.

I don't understand why the 2 DAE files have had their collision surfaces removed/modified, or in fact which ones have been removed. I get that you've excluded a volume down the centre of the cylinder but why does that affect the collision geometry of the stations?

mike-f1:
The reason for the deletion of collision surface is that I put a variable ( "central_hole_dock" : 1 in JSON files) to specify if the presence of a ship inside the volume should trigger docking, then I would test and show this feature...

So now you could have spacestations with or without a "hole" in which you don't check for collision,
and you could also specify if this hole could trigger a docking collision or not.
This means also you no longer needs surfaces to trigger the collision.

mike-f1:
History almost cleaned, but I cannot complete rebasing...
Just give me more time :/

mike-f1:
ready to be merged, ping @impaktor to:
repeat {say "thank you" } until false :D

impaktor:
Sweet.

For actual review, I suggest @fluffyfreak or/and @jaj22; as I don't understand anything of this code, and don't have time to go into it.

mike-f1:
@impaktor : I thank you for explanation on how to "repair" branches,

then I ping @fluffyfreak for the review.

jaj22:
And @fluffyfreak pinged me in return.

Two or three separate things in this PR. Looks like there's a bugfix on docking logic which should probably be merged ASAP (although I haven't checked the logic), but the rest is more contentious:

**Parent/child geom system**
This doesn't appear to have any advantages over the group ID mechanism within the current collision system, and it's more complex. Conventionally you'd make use of a parent/child geom relationship to exclude children from testing if the parent didn't intersect, but the BVH collision system in Pioneer is an alternative to that.

It looks like the code also doesn't set the parent/child relationship for animated ship geoms. Last time I checked, no current ship model uses animated collision geometry, but if I'm reading this correctly then it'd have a substantial performance impact with this code. Game start/load logic would also need to be checked.

The group ID system should be easy enough to use instead: Set the group ID equal to the station on docking, and then get a new group ID on undocking. Could implement this by adding an optional group parameter to the SetColliding() function, and then everything should work correctly with a few lines of extra code. 

Note that the group ID overflow concept is shaky, but the probability of an accidental ID collision is extremely low and the impact would be minor. A better system would simply use an ID of 0 for all static objects (or the orbital station - they're dynamic) in a CollisionSpace.

**Central hole system**
Well, it's special-case, and it probably doesn't save much CPU because at maximum time acceleration, each docking ship will typically only spend one frame inside the station radius. Probably not worth the complexity.

There's a better case for having a generalised docking-collision system with simple convex shapes that works in parallel with the current collision system. The advantage is that you could turn off the fancier collision system outside the player's current CollisionSpace if short on CPU time, and it wouldn't stack functionality onto a system that's already barely readable. However, I'm not sure if there's any good way to add this to the current model definition system.

**General concept of enabling collisions for docked ships**
Note that changing the collision status of docked ships may have a significant performance impact, even if there's nothing near their radius. Every frame, a new BvhTree is built for all dynamic geoms within each reference frame. That's problematic for planets with many ground stations, because every docked and undocked ship in every station on the same planet is in the same reference frame. However, this may still be an order of magnitude cheaper than building the BvhTree for the deep-space ships.

An alternative is to set ground-docked ships as static (as is done for the Landed state), but this is probably a bad idea because it causes a rebuild of the entire static geom BvhTree, which includes every building on the planet.

My guess is that enabling collisions for docked ships and leaving them dynamic won't hurt much overall (assuming docked-ship grouping), but it should be tested.

fluffyfreak:
Thanks @jaj22 

Ok @mike-f1 if there's a bug fix in this PR (_I don't know where it might be_) then that should be a separate PR that we could merge to master after reviewing separately.

The other parts about the central hole: If this is a performance issue then you can profile it and tell us what has changed or why it's good. I can't profile it for you and know that I am testing the correct thing.

Profiling: We have some a built-in intrusive profiling that you can use by defining the `PIONEER_PROFILER` define when building on Linux. Search for `PROFILE_SCOPED()` in the source code and you'll see that you just need to define that at the top of a function/method.
Once built you can press `Ctrl+P` to capture a frames profile in html format in the `./.pioneer/profile/` folder.

You can of course also use another profiling tool of your choice and just explain the results to us :)

mike-f1:
@jaj22 : if "Logic change" on docking system refer to the "return true" instead of "return false" ok: I should open another PR, but I not yet tested the logic.

# About parent/child relation:
You're right when you speak about dynamic geoms: I left them "as is" and there's a need to handle them.
but I think the interface I did could be reused, so AddChild / RemoveChild are easy to understand:
could be an improvement to refer internally to group handles, but it keep the "dynamicity" needed when you need to push/pop geoms.

A thought I had about dynamic geoms is they needs a better handling by the code... Don't know if I'm wrong, but to me it seems strange many function in ModelBody work on m_geom only, while others works on both (m_geom and m_dynGeoms).

# Central hole system
I think the gains should increase with the number of ships docking/undocking in spacestations: you do simple math (...a distance check instead of checking for collision triangle by triangle...) and this (sure I could bet) is faster.

# General concept of enabling collisions for docked ships
I can postpone collision check for docked/undocking/docking ships to when I have a way to do "fast math" for ground stations.
But I don't think performance will be so worse if you check for docked/undocking/docking ship, because if you find a way to skip the deep test with intersecting object (the ground stations) then you have to only check for AAbb distance.

# Conclusion
I think I could move internally from building a list of children back to handles, keeping the "interface"
(this should avoid problems with "movable" geoms).
Then I can avoid to put docked ship collision tested, at least until I don't find a way to avoid a deep test with ground stations.
Last: I would better understand how "movable" geoms interact with the system

Let me know if what I think to do is ok

mike-f1:
ok: pinging @jaj22, to say...

# About parent/child relation:
I did the changes in Geoms: now (preserving the interface) Geoms rely on handles.

> Game start/load logic would also need to be checked.

I put an "assert" to assure load/save and start don't call `AddChild` twice (there was a bug...), and I leave it with a comment to avoid others to do the same mistake I took.
I changed also `SetColliding` to take care of "dynG": this seems wrong (why you want only enable/disable the main geom?), but let me know.

# Central Hole

Nothing changed, I would only add a note: every ship in a system needs to be checked if it is near a spacestation, *even* at low timesteps and then you can bail out of collision checks before the first triangle (...of a long list) is checked.

# General concept of enabling collisions for docked ships

...I think when I wrote the first answer, I was drunk :P, let my re-try:
a collision check between each ships and each spacestations happen every timesteps, but this is not so slow because you check for aabb before doing a deep test and "bail out".
However I think it is better to enable collision only when ship are docked: this could be changed when a refined docking system will avoid ships colliding.

# Conclusion

well, hoping I'm not doing a mess, when you review this I will clean history.

fluffyfreak:
@mike-f1 ok I've let this one stew for longer than I had intended, I apolgise for that but real-life has been happening too me lately.

I think that I understand the concept at work here, but I don't feel that it's right. It adds some complexity and I don't believe that there are noticeable benefits to that complexity. The physics doesn't show up much in profiling, and determining if this helps or hinders is very difficult.

For the time being - because it might become a useful idea in the future depending on how Pioneer changes - I am going to close this PR without merging your changes.

My apologies for doing that as the code itself seems sound and you've fixed up all of the history issues etc and I do hate rejecting code, but probably not as much as you hate having it declined.

PR #3925, opened 2017-02-14
GLEW resurrection, initial code and project changes

fluffyfreak:
# What is this?
Some time ago I switched us from GLEW to glLoadGen generated files (#3265).
This PR takes us in the opposite direction back into the sticky arms of GLEW.

# Why is this?
Well there's been another PR (#3808) that cannot progress because supporting two opengl libs is a bit of pain using glLoadGen unless you lump everything together into a single super-lib.

That was one of the original arguments against GLEW itself.
GLEW however potentially has some benefits that glLoadGen doesn't have such as EGL support.

This PR is me exploring whether it makes sense to go back and I've written it in about 20 minutes so it's easy to check.

# Ready?
Yes!
This now builds on the target platforms.


laarmen:
For those who don't quite remember what GLEW was:

https://en.wikipedia.org/wiki/OpenGL_Extension_Wrangler_Library

PR #3930, opened 2017-02-17
Updated MacOS Xcode project

salborrelli:
Updated Xcode project for MaxOS X to include all the latest changes.


impaktor:
Yay! Thanks! :fireworks:

fluffyfreak:
You are a star @salborrelli thankyou so much!

fluffyfreak:
Ah, it fails to link for me, complaining about library not found `-lz 1.2.8`, I've seen this before where you point it to the generic library reference but XCode goes and dereferences it to the specific version library.

I can fix mine copy up locally but we'll use your's as mine ends up with crazy paths. This is another reason I dislike Xcode! :)

Now to figure out how to package it all up as an `.app` again too!

PR #3933, opened 2017-02-25
Avoid jousting when docking on orbitals at high time steps

mike-f1:
This is to avoid a bug (...I put on code :( ...): when ships (even AI) are docking on
orbitals at high time step they remains stuck in "level ship" because,
if in time accel, their position change is less than the movement
of the station itself (due to its rotation).

So you need to jump directly out on the right stage.

### DIFF
--- a/src/SpaceStation.cpp
+++ b/src/SpaceStation.cpp
@@ -464,6 +464,14 @@ void SpaceStation::DockingUpdate(const double timeStep)
 			float dist = 0.0;
 			switch ( extraStage ) {
 				case 1: // Level ship & Reposition eval
+					// PS: This is to avoid to float around if dock
+					// at high time steps on an orbital
+					if (!IsGroundStation()) {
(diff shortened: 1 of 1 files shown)


src/SpaceStation.cpp
```diff
@@ -464,6 +464,14 @@ void SpaceStation::DockingUpdate(const double timeStep)
 			float dist = 0.0;
 			switch ( extraStage ) {
 				case 1: // Level ship & Reposition eval
+                    // PS: This is to avoid to float around if dock
+                    // at high time steps on an orbital:
+                    if (!IsGroundStation()) {
+                        dt.fromPos = vector3d(0.0); //No offset
+						dt.fromRot = Quaterniond(1.0,0.0,0.0,0.0); //Identity (no rotation)
+                        dt.stage += 2;
```
impaktor: Indentation looks off

PR #3941, opened 2017-03-05
Refactor the Vertex, Index, etc buffers

fluffyfreak:
This is some generic refactoring, fixing, cleaning of the VertexBuffer, IndexBuffer, Texture classes along wiht more `override final` usage etc.

- Exposing generic textureID accessor, 
- virtual Bind/Unbind methods, 
- override final, 
- naming tweaks 
- BufferData methods

### DIFF
--- a/src/PiGui.cpp
+++ b/src/PiGui.cpp
@@ -59,7 +59,7 @@ ImTextureID PiGui::RenderSVG(std::string svgFilename, int width, int height) {
 		if(strTex.first == svgFilename) {
 			// nasty bit as I invoke the TextureGL
 			Graphics::OGL::TextureGL *pGLTex = reinterpret_cast<Graphics::OGL::TextureGL*>(strTex.second);
-			Uint32 result = pGLTex->GetTexture();
+			Uint32 result = pGLTex->GetTextureID();
  			Output("Re-used existing texture with id: %i\n", result);
 			return reinterpret_cast<void*>(result);
 		}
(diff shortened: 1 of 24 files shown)


src/scenegraph/Label3D.cpp
```diff
@@ -54,7 +54,7 @@ void Label3D::SetText(const std::string &text)
 		// Most noticeably, this means text consisting of entirely Cyrillic
 		// or Chinese characters will vanish when rendered on a Label3D.
 		if (m_geometry->IsEmpty()) { return; }
-		
+
```
impaktor: I thought you didn't like white space edits? :grinning:

fluffyfreak: I'm about to do a whole codebase one just for whitespace so I'm letting these ones slide ;)

src/PiGui.cpp
```diff
@@ -284,7 +284,7 @@ void *PiGui::makeTexture(const std::string &filename, unsigned char *pixels, int
 	Pi::renderer->CheckRenderErrors(__FUNCTION__, __LINE__);
 	// nasty bit as I invoke the TextureGL
 	Graphics::OGL::TextureGL *pGLTex = reinterpret_cast<Graphics::OGL::TextureGL*>(pTex);
-	Uint32 result = pGLTex->GetTexture();
+	Uint32 result = pGLTex->GetTextureID();
```
laarmen: Question : should we keep the Uint32 type (which I assume is "homemade" or use the uint32_t type ? I see the GetTextureId() declaration uses the latter.

impaktor: I wonder if this is a more general question, should we migrate to C++11 standard, and avoid pioneer home made stuff? e.g. [RefCountedPtr to std::shared_ptr](https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=267)

fluffyfreak: I think we could switch from `Uint32` to `uint32_t`, and their ilk.
I've heard that `std::shared_ptr` has a [lot of performance](http://stackoverflow.com/questions/3628081/shared-ptr-horrible-speed)([2](http://blog.davidecoppola.com/2016/10/performance-of-raw-pointers-vs-smart-pointers-in-cpp/) and [usage issues](http://seanmiddleditch.com/dangers-of-stdshared_ptr/) so it might be worth benchmarking our `RefCountedPtr` vs `std::shared_ptr`.

PR #3942, opened 2017-03-05
Eliminate the evil trailing whitespace

fluffyfreak:
Eliminate the evil trailing whitespace that has crept into the project over time... probably mostly my own fault with Visual Studio.

I have a plugin now that removes any whitespace upon saving a file so here a single big hit to eliminate all of the existing whitespace in one go.

I'm sure @mike-f1 will appreciate this since it means I'll stop complaining about it in his PR's ;)


mike-f1:
https://www.youtube.com/watch?v=YYOKMUTTDdA :)

@fluffyfreak , just wait a moment with #3923 : I will check if there are any and blow them up ;)

PR #3943, opened 2017-03-05
All white spaces be gone! (also from lua)

impaktor:
In the words of the insane Roky Erickson:
```
White spaces
Always haunt me so beautifully
Always scare the white of the devil in me
```


impaktor:
This is a sequel to #3942. Sure, people argue sequels aren't as good as the originals, but Aliens, and Terminator 2, are in a class of their own.

impaktor:
...right, sadly this sequel isn't one of them, as I left out important plot points: like I didn't do the data/pigui file, since that's @ecraven's playground, and I thought that will be fixed by him, when time comes.

fluffyfreak:
I'm happy for you to merge this. I deliberately didn't remove whitespace in our `contribs` folder.

PR #3944, opened 2017-03-12
Removing some include of ship.h

mike-f1:
Here and there I saw `#include "Ship.h"` in unexpected places, and after removing these lines
Pioneer still compiling... So I removed these lines.

I did also some additional things:
- HeatParameter_t moved in "src/graphics/opengl/MultiMaterial.h", because I think ships (and not only them) could use a material, not the material be a "slave" of the ship. But I did something not completely good, because then I created a dependencies in  "src/graphics/gl2/GL2MultiMaterial.cpp".
I feel this is the way and the problem with *dependency* could be solved with a separate file shared from renderers, but I ping @fluffyfreak to share thoughts (also because I don't know how it works...).
- Removed a line generating a warning in PiGui.cpp.

...Some thought I had were about how `#include` is used: to me and IMHO seems there are some case in which `#include` is not used *exactly* the right way: the fact is that without the lines I removed, the files involved weren't recompiled. I think I could do a better job if I'm aware of reasons or the way you (devs) are *including*.

Here a link in order to have some clue:
http://www.cplusplus.com/forum/articles/10627/

What do you think?

### DIFF
--- a/src/Camera.cpp
+++ b/src/Camera.cpp
@@ -9,7 +9,6 @@
 #include "Pi.h"
 #include "Sfx.h"
 #include "Game.h"
-#include "Planet.h"
 #include "graphics/Graphics.h"
 #include "graphics/Renderer.h"
 #include "graphics/VertexArray.h"
--- a/src/PiGui.cpp
+++ b/src/PiGui.cpp
@@ -68,7 +68,6 @@ ImTextureID PiGui::RenderSVG(std::string svgFilename, int width, int height) {
 	NSVGimage *image = NULL;
 	NSVGrasterizer *rast = NULL;
 	unsigned char* img = NULL;
-	int w, h;
 	// size of each icon
 	//	int size = 64;
 	// 16 columns
@@ -83,8 +82,7 @@ ImTextureID PiGui::RenderSVG(std::string svgFilename, int width, int height) {
 	if (image == NULL) {
 		Error("Could not open SVG image.\n");
 	}
-	w = static_cast<int>(image->width);
-	h = static_cast<int>(image->height);
+	int w = static_cast<int>(image->width);
 
 	rast = nsvgCreateRasterizer();
 	if (rast == NULL) {
--- a/src/Shields.cpp
+++ b/src/Shields.cpp
@@ -6,7 +6,6 @@
 #include "scenegraph/FindNodeVisitor.h"
 #include "scenegraph/SceneGraph.h"
 #include "scenegraph/CollisionGeometry.h"
-#include "Ship.h"
 #include "Pi.h"
 #include "json/JsonUtils.h"
 #include <sstream>
--- a/src/Ship.h
+++ b/src/Ship.h
@@ -19,6 +19,7 @@
 #include "LuaTable.h"
 #include <list>
 #include <unordered_map>
+#include "graphics/opengl/MultiMaterial.h"
 
 #include "Propulsion.h"
 #include "FixedGuns.h"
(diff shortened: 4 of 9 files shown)


impaktor:
(Ideally, removing ship.h is one commit, removing the warning in PiGui.cpp is one, and moving material is one, so this _could_ be three commits. Anyway, I'm fine with them all as one, assuming we want all three changes).

The `#include` dependencies are all over the place. I imagine there is some tool that could go through and remove dead includes, and such. 

My philosophy, is I (explicitly) include a file if I use something from it,  even if it is (implicitly) included from some other include, so it might have compiled without the include.

Nice to see you're still hacking on pioneer.

fluffyfreak:
Ah, ok this can't be merged. You've moved the `HeatParameter_t` into and implementation specific location, namely only the `opengl` one. What about GL2, or a new gles/vulkan/dx12 implementation?

It's ok for the renderer to depend on a piece of the game engne code, it's not ok for the game to have dependencies on the specific implementations which may or may not be available on another platform.

mike-f1:
@fluffyfreak : I think you're right, but how this can be solved? I mean: you're right, I put this on wrong location but ship.cpp is not the right place, isn't? Tell me how I can share a material definition between implementation and I will do it, np.

But I would point the talk about the way `#include` are used, because I want to understand if there's something to improve: in the link he state you should prefer to include in .h file, but I don't think this is the way it is done here... There's a reason? Have you thought about this?

fluffyfreak:
@mike-f1 for the `HeatParameter_t` I would create a separate header file with the struct definition in it. Then `Ship.h` and `MultiMaterial.h` can `#include "HeatParameter.h"` that and share the definition.

There's a lot of mess with the current include files.
Where possible things should be forward declared in the header and then `#include` within the `.cpp` file. This avoids having too many nested `#include "XXX.h"` calls where header files include other header files.

Ideally you avoid using `#include` within header files at all. This is what the `[pimpl](http://en.cppreference.com/w/cpp/language/pimpl)` idiom is all about. It can drastically improve compilation and linking times.

mike-f1:
@fluffyfreak : Where's this file has to be placed? I mean: in which folder? ...and, I or you do this? Let me know.

About pimpl, I never heard of it... But this is obvious from me :-/ ...
But as far as I can understand, there's nothing in code that implement things like that, not in the class I learned to know :/ ... or I'm wrong? So there's many steps before we could see this implemented, although I think (at least from my PoV: I work on ships and related stuffs...) this could reduce compiling time if Ship.* use this technique... And for sure I will try to use this :)

A thing I don't understand is about why or when pimpl need to `#include` only on *cpp: probably a mine misunderstood because I didn''t got it completely, but this is completely against what my link point to... And I noticed things works as mine pointed: when I removed these "dead dependencies" then the no more involved files stop to be recompiled (and this made my happy :) ).

...Well: I will learn and read more before stating something, then I would try.

fluffyfreak:
@mike-f1 I only linked ot the pimpl-idiom as an example of what lengths you can go too to avoid including files within the headers. It's not something we bother using on Pioneer :)

A new header for `HeatParameter_t` could just go in `/pioneer/src/` along with `ship.h` and other files.

fluffyfreak:
Ok I've taken a look at this and I'm of a mind (thinkg about) closing it.

The reason is that all of those files use the objects defined within either `ship.h` or `planet.h` etc. There is no compilation cost to them being included - thanks to the use of compilation guards - but they do expressly show what classes are actually being used. Which is the way they should be.

We do need to tidy up the includes in the game but this should be by removing the usage of header within other headers and _not_ by removing the correct usage of them here.

mike-f1:
Please, left this open for some time: there's no hurry to merge or close it soon: I only would the talk to proceed and collect other people opinion...

I won't bother people but I'm thinking that better "include" could lead to a better encapsulation or to things like the one I did for Propulsion & Friends: a way you could be aware of the needs of a (in-)partial data struct shared between classes should be through these header (...I can state they *accidentally* could track dependencies...).

mike-f1:
...I agree with you: this could be closed...

mike-f1:
Oh @fluffyfreak : no problem, I realised this should be better understood in IRC...

For `HeatParameters_t` I should do a PR... Not soon, because I have a trouble I would solve
sooner.

PR #3947, opened 2017-03-12
Slot name typo

### LINKED ISSUE #3945: Cannot buy Trade Computer
Cannot buy a Trade Computer. There is always the message "Trade Computer is not supported on this ship model". It seems no ship type supports it.

robothauler:
Closes #3945


fluffyfreak:
Doh, good catch!

impaktor:
Funny, I was looking at that exact line yesterday, to find the bug, to make sure it wasn't a typo there, but I must have been blind, or in a LSD haze. Good catch, thanks!

PR #3948, opened 2017-03-14
Thruster colors can be modified globally or per direction

### LINKED ISSUE #961: More varied thruster colours
In #945 @Luomu showed us some possible colour variation for the new thrusters:

We've had a few ideas about when variations might occur:
- Based on drive tech (hydrogen or military). Simple but not much variation.
- Based on drive class. Lots of variation, but does it give too much away? We do already have that for laser colours though.
- Based on model (and defined by the model). What's the in-game explanation for this?
- Blended with atmosphere colour, perhaps with intensity influenced by pressure.

Your task: implement one of these :)

mike-f1:
coloredthrusters

This add customizable thrusters color for each model (partially close #961) and
also allow to choose customized colors *per direction*, as in screenshot.

...Ping @nozmajner to provide feedback about colors: mine are too
much psychedelic :D 

As you can see in sinonatrix.json, a structure in which each value is a channel
is used (standard "r", "g", "b" are used, values are [0,255] ).

Two global keyword are provided to choose the settings you wish to customize:
- `thruster_global_color` for global color
- `thruster_direction_color` to define colors per direction

Inside `thruster_global_color` you then have values for each channel,
inside `thruster_direction_color` you need to define a struct value for
direction(s) you will overwrite, like:
```
thruster_direction_color {
    [keyword] : {
        "r" : value_for_red,
        "g": value_for_green,
        "b": value_for_blue
    }
    .....
}

```
Recognized keywords are:

1. forward
2. retro
3. left
4. right
5. up
6. down

### DIFF
--- a/src/Intro.cpp
+++ b/src/Intro.cpp
@@ -37,6 +37,20 @@ Intro::Intro(Graphics::Renderer *r, int width, int height)
 	for (auto i : ShipType::player_ships) {
 		SceneGraph::Model *model = Pi::FindModel(ShipType::types[i].modelName)->MakeInstance();
 		model->SetThrust(vector3f(0.f, 0.f, -0.6f), vector3f(0.f));
+		if (ShipType::types[i].isGlobalColorDefined) model->SetThrusterColor(ShipType::types[i].globalThrusterColor);
+		for (int j=0; j<THRUSTER_MAX; j++) {
+			if (!ShipType::types[i].isDirectionColorDefined[j]) continue;
+			vector3f dir;
+			switch (j) {
+				case THRUSTER_FORWARD: dir = vector3f(0.0, 0.0, 1.0); break;
+				case THRUSTER_REVERSE: dir = vector3f(0.0, 0.0, -1.0); break;
+				case THRUSTER_LEFT: dir = vector3f(1.0, 0.0, 0.0); break;
+				case THRUSTER_RIGHT: dir = vector3f(-1.0, 0.0, 0.0); break;
+				case THRUSTER_UP: dir = vector3f(1.0, 0.0, 0.0); break;
+				case THRUSTER_DOWN: dir = vector3f(-1.0, 0.0, 0.0); break;
+			}
+			model->SetThrusterColor(dir, ShipType::types[i].directionThrusterColor[j]);
+		}
 		const Uint32 numMats = model->GetNumMaterials();
 		for( Uint32 m=0; m<numMats; m++ ) {
 			RefCountedPtr<Graphics::Material> mat = model->GetMaterialByIndex(m);
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -229,7 +229,6 @@ void Ship::Init()
 	p.Set("fuelMassLeft", m_stats.fuel_tank_mass_left);
 
 	// Init of Propulsion:
-	// TODO: Separe the enum used in ShipType and use it both on Propulsion and ShipType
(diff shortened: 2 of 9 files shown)


impaktor:
Yeah, my opinion will be whatever @nozmajner says.

However, until then I'm wondering how best this is used?

1. Set in json file of ship, like this PR does?
2. Set in blender model file?
3. Is it even a trait of the ship, and not the engine?
    - Military drive have different colour than "regular"? (which maybe
      doesn't make sense because this is related to thrusters and not
      hyperdive engine)
    - age / condition / capability / power of thrusters?

mike-f1:
@fluffyfreak : to me seems Travis was unable to connect to the site, but let it try again: I see a bug on my code ( :( ) I would correct, so I commit again.

@impaktor : Here I tried the third way listed in #_961 and this is (simply) a way to provide customized colors. This could be seen as an "appetizer" before doing something better integrated.
The reasons I choose this ways are that the firsts two ways refer (and link) hyperdrives to thrusters (and I prefer to avoid such link, as you already argued) while the forth make a link with atmosphere (...but is still possible for someone to change Thruster.* in order to add a blending function with athmosphere color).
Instead I'm more prone to link thrusters color to ships or (when time come) better to its sub-systems.
The way I did it is with json file for its simplicity and its flexibility, and this make it a promising starting point where we could put some "infrastructure"/"complexity".

I have another branch (...still WIP but working) in which I choose the same place to put data and it is easy to see it is working... However some problems arised: I experienced indeed some troubles when I needed to pass data between `ShipType`, `Ship` (intended as the whole or as the sum of its parents) and `Model`.
My guess is there's a need to find a way in which we can "merge" data from `Model` with data from `ShipType` sooner in code, and then let `Ship` work with results (same but less needed with other kind of objects)... But for now I think I need time to think to this: for now this is just a "feel" I had.

Another problem arise if you would have 

> age / condition / capability / power of thrusters?

this could be (for me) a *fantastic* solution, but there's no infrastructure supporting it as now, so we should provide one and then use it.

...Hoping my english is good enough and *you* are good enough to understand it, I say

Thank you

nozmajner:
Putting it in the json seems logical, even though I did object putting such visual things there. But this way there could even be versions of the same ship, with like different engines, and the color is closely put to the engine data anyaway, and also you don't have to set up separate models for each "engine-ed" ship. Maybe there could be a bit of name difference, like "Sinonatrix (model 3175)" and so on. (Until ship descriptions are implemented of course.)
And if they will go into somewhere engine related in the future, then it's even better.

HybridDog:
Can you set pitch black thruster colours?

mike-f1:
> Can you set pitch black thruster colours?

No: Take a glance at Thruster.cpp line 92:
`m_tMat->diffuse = m_glowMat->diffuse = currentColor * power;`

color of thruster materials are multiplied with `power` that represent the
amount of thrust you have per frame, so what you're asking for needs
a rewrite of the class :/

nozmajner:
I don't think black thrusters would be any useful anyway.

mike-f1:
well: I meant that what remain is that if you reduce the amount of color values, then you reduce the size of the thrusters exhaust... Anyway, I open this #3949

laarmen:
Does this mean this PR is obsoleted by #3949 ? In any case, awesome work!

mike-f1:
@laarmen, thank you, however #3949 is intended for a long term issue where implement something "deeper" than this...

Issue #1065, opened 2012-03-12
Allow the model system to dynamically change thrust values/locations


Brianetta:
This allows for two things:
1. Tail-sitting rocket ships, such as Gernot's atomic rocket (see image)
2. Ships like Frontier's Panther, which has (or should have) thrust vector changes (see video)

**Gernot's atomic rocket:**
![Gernot's Atomic Rocket](http://www.spacesimcentral.com/forum/download/file.php?id=2109&t=1)

**Panther's undercarriage (click for video):**
[![](http://i2.ytimg.com/vi/95WoVjx00Uk/hqdefault.jpg)](http://www.youtube.com/watch?v=95WoVjx00Uk)

PR #3952, opened 2017-03-18
Vectorial thrusters

mike-f1:
This add the capacity to have vectorial thrusters for ships (or missile) as in picture below:
vectorialthrusters

Something is still WIP, because of course there's a need to change Tsiolkovsky equation used
in Propulsion to take into account different eev, there's a need for ShipType to have right data
accordingly with vectorial thrusters, ecc... ecc...

...Sure I won't this Sinonatrix pushed in master ;) ...although I must admit I grow fond of it...

I will clean commit history when time come; comments&critics are welcome, patches&testing even more :) ...working ships with nacelles, Alleluya :))) ...

(Edit: solve #1065 )


nozmajner:
Will there be any rotation limits? Like if I want to do something like a Harrier, which only points the engines down and aft, not up at all? Or on a ship that has four nacelles, so the fwd ones rotate to become retros, the aft ones rotate to become mains?
Is it possible to put multiple thrusters on the same nacelle? Like one on each end, so your retros point upwards when your mains point down.

mike-f1:
@nozmajner : well, I forgot to do this :/ :

> Will there be any rotation limits?

...but a way to have 4 nacelles not frying each other is placing the tag_thruster rotated outside, and this should provide more reliability because you then have all four nacelles burning and providing thrust (although this is not *well* handled by the code because there's a reduction factor not (yet?) implemented).
But you're right: this don't works in order to avoid nacelle frying your hull.

While for this:
> Is it possible to put multiple thrusters on the same nacelle? Like one on each end, so your retros point upwards when your mains point down.

Not yet implemented, but are you referring to a way to have more than a single thruster *per direction* or to a nacelle with thrusters pointing in *different directions*?

Err...: I have also to put code to save/load nacelles data, so this needs "WIP" and "Savegame Bumps";

Err2...: All this works with *normal* (=static) Collision Mesh: so avoid to have too protruding nacelles.

nozmajner:
Yes, thrusters firing in different directions. I'm asking these to know what limitations I have to keep in mind designing ships with this feature.
Apart from limiting rotation, that is important in my opinion for this to work above the example case you provided with that sinonatrix.

mike-f1:
@nozmajner : I edit the comment just on your comment...

fluffyfreak:
Not ignoring this (_or other PR's_), just busy at moment, will review next week I hope.

mike-f1:
Np @fluffyfreak , this is WIP: I must add at least code to handle save/load of nacelle data so you could delay the review.

impaktor:
I'm happy to just nag about commit log, before this goes in.

Just looking at the pictures, it looks like a great contribution.

impaktor:
http://newatlas.com/darpa-xplane-vtol-successful-flight-test/48823/

mike-f1:
> http://newatlas.com/darpa-xplane-vtol-successful-flight-test/48823/

...They use my code ;)

src/Propulsion.cpp
```diff
@@ -36,65 +37,143 @@ void Propulsion::LoadFromJson(const Json::Value &jsonObj, Space *space)
...
+}
+
 Propulsion::Propulsion()
 {
 	m_power_mul = 1.0;
 	m_fuelTankMass = 1;
 	for ( int i=0; i< Thruster::THRUSTER_MAX; i++) m_linThrust[i]=0.0;
 	m_angThrust = 0.0;
-	m_effectiveExhaustVelocity = 100000.0;
+	m_effectiveExhaustVelocity = 1000.0; // This should have an impact on game...
```
fluffyfreak: What is the rationale for this change in the `m_effectiveExhaustVelocity`?

fluffyfreak:
Ok I really like this feature. I've only read the code so far as I have other uncomitted breaking changes local to my machine.

Overall the code changes look good, some good renaming of unclear variables is always good to see.

How do you specify what nacelle model you use? Is this part of the ship mesh? Or from a separate set of nacelle models? @nozmajner might be able to answer that if you've been talking with him.

I am confused about the effect of changing `m_effectiveExhaustVelocity` by two orders-of-magnitude! That's a huge change and I wonder what the effects would be.

Anyway I like it and the code looks good so far.

src/Propulsion.cpp
mike-f1: m_effectiveExhaustVelocity need to be initialized to a value.

But when you play you will overwrite this default value with data loaded from
json files: if this not happen then something is wrong :/

Here I would provide a feedback to players/modders/ecc: if this value is
used when they play, then your tank will empty very very fast.

...Probably a better comment should be useful...

mike-f1:
> How do you specify what nacelle model you use?

Nacelle geometries to be used are specified in json file.
Each nacelle (defined with "vect_thrusters") have custom parameters,
and you use "tag" for geometry (...although a better choice should be "geometry").
They rotate around the origin of the sub-geometry it self.

> Is this part of the ship mesh?

They needs to be defined as separate objects ( @nozmajner could answer
better)

> I am confused about the effect of changing m_effectiveExhaustVelocity by two orders-of-magnitude! That's a huge change and I wonder what the effects would be.

Hope nothing :smile: : the default value needs to be ever overwrited with json
data, here I would give a feedback if this is not overwrited.

src/Propulsion.cpp
laarmen: Yep, a comment along the lines of what you just said would be swell! :-)

fluffyfreak:
@mike-f1 are you still working on this PR?
Because I like what you've done with it so far and once the hsitory is cleaned up.

Thanks for replying to the questions.

impaktor:
@mike-f1 I'd be hapy to help you learn some git.

Or, my constant offer still stands: I could squash the whole thing to a single illogical big commit.

mike-f1:
@fluffyfreak : I'm trying to find time and inclination to do this... But if you think it is better, I can close this PR and open another when I'm ready, let my know.

@impaktor : well, probably I'm still learning (and that how I asked fluffyfreak about the "*illogical* way he used when he did another PR), but I could squash (...I only did this until now...)

Please, don't use my code if I don't explicity open a PR: there's something wrong there I would solve by myself.

Thank's

fluffyfreak:
@mike-f1 leaving it open is fine, we've had PR's that were open and being actively worked on for over a year! We do try to avoid that, but only because I worry it can feel like a burden to the person working on it, and it isn't supposed to be that way. Working on Pioneer should be fun!

PR #3953, opened 2017-03-20
Ever damage ship when docking

mike-f1:
As seen by @jaj22 , this is the bugfix he asked for...

It's not tested but the meaning is: if a collision with docking surface happens,
then *probably* you want the ship bounced back...

Waiting for review


fluffyfreak:
Thanks @mike-f1 👍

PR #3955, opened 2017-03-22
Configure cleanup

laarmen:
Remove the LDB thingy I wrote back in the days of 1719, and make the Server Agent an opt-in compile-time option.

Note that the agent will also be disabled in the Windows and Mac builds, but I haven't disabled the libcurl dependency for those yet.

The rationale for the server agent thing is quite simple : we don't use it, and the libcurl dependency is a pain because some distros ship with a SONAME of 3 by default (Debian) whereas others only ship the SONAME 4.
For LDB, well, it's just broken and unused.

I didn't actually test the build WITH the server agent, but the default builds fine, and doesn't crash in the first 10 seconds of gameplay.


fluffyfreak:
Looks good. Shame the LDB stuff never went anywhere or the ServerAgent stuff for that matter.

impaktor:
Maybe a line about this flag, and debug and profiling flags in COMPILING.txt?

laarmen:
Indeed...

fluffyfreak:
So is this ready to merge?

laarmen:
I guess so, but I'd be reassured if you did a test build and run on
Windows, though.

Le ven. 24 mars 2017 à 11:54, Andrew Copland <notifications@github.com> a
écrit :

fluffyfreak:
Yep, just tested it, builds and runs fine. Merging!

PR #3969, opened 2017-04-13
vs2017 solution and project files

fluffyfreak:
MS have released their biennial Visual Studio release with vs2017 and it has a new Community (free) edition.

As usual I'm being forced to update for work reasons (_console stuff_) and so I'm spreading the cheer for people using the new one.

Builds fine with vs2015 DLLs in thirdparty so far.

I'll continue to support both vs2015 and vs2017 for ... well I dunno, quite a while probably.


impaktor:
Well, this is all on you. No one else can review it, I guess.

But this only adds the configuration files for vs2017, right? So why are there gif-icons in there?

fluffyfreak:
Yeah that's fine, it's just I like to go through review 😄 

Also debugging doesn't work yet, might need new DLLs for vs2017 specifically 😞

impaktor:
@fluffyfreak I'm thinking a big merge tonight, with imgui branch. Are you happy with this as is?

fluffyfreak:
No this is WIP as it cannot be debugged etc, it relies on using the vs2015 libraries, which works for doing a build but not for useful things like debugging etc 😢 

Hence the WIP tag at moment.

laarmen:
BTW, I had to use the integrated CMake support in VS 2017 for work, and it works quite nicely. Might be worth a(nother) try.

fluffyfreak:
I think I'll close this for now and redo it in another way.

PR #3973, opened 2017-04-16
Fix name of option for server agent in configure.ac file

senderghost:
This allows to use with_server_agent variable.
Fixes #3955 pull request.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -312,7 +312,7 @@ if test "x$cross_compiling" = "xno" ; then
 fi
 
 dnl Check for libcurl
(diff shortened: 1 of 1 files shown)


impaktor:
ping @laarmen

senderghost:
To check this, you may need to use --with-server-agent configure argument.

Current --with-serveragent option (as described in `./configure help`) produces following warning: "configure: WARNING: unrecognized options: --with-serveragent" and no "checking for CURL... yes" message.

laarmen:
Yeah, I didn't actually test the enabled version. My bad.

Le dim. 16 avr. 2017 à 21:17, senderghost <notifications@github.com> a
écrit :

senderghost:
Also removed "WITH_SERVER_AGENT=no" part, because it is not used.

PR #3975, opened 2017-04-18
VS2015 win32 & x64 plus reorganisation

fluffyfreak:
WARNING: This will not work without the corresponding https://github.com/pioneerspacesim/pioneer-thirdparty/pull/21 files checked out.

I have reorganised and updated some of the files in [pioneer-thirdparty](https://github.com/pioneerspacesim/pioneer-thirdparty) and added an x64 build target which means that I now get to see all the same compilation warnings as the Linux/OSX folks 😄 

This, hopefully, shouldn't affect Linux users too much, but I guess we'll see.

Problems:
- 64-bit DEBUG and PRERELEASE don't work 😢 
- MANY warnings about type conversion, problably others lost in the noise

I honesrly don't care about those details, they're things that will come out in-the-wash as they say and 32-bit debug and prerelease still work fine.


fluffyfreak:
@walterar and @radius75 just a quick prod so that you don't get confused when you get the latest (_after this is merged_) and it doesn't build.

You'll need to get the latest of the pioneer-thirdparty if you're using Visual Studio 2015 to build Pioneer.

PR #3977, opened 2017-04-19
Add screenshot support for OpenGL 2.1 renderer

### LINKED ISSUE #3976: OpenGL 2.1 print screen triggers assertion
```
PngWriter.cpp:45: void write_png(FileSystem::FileSourceFS&, const string&, const Uint8*, int, int, int, int): Assertion `0' failed
```

fluffyfreak:
This adds support for screenshots when using OpenGL 2.1 and should fix #3976


fluffyfreak:
@impaktor if you get a chance to test this it might still silently fail, it does for me on Windows, however it shouldn't crash/assert.

ecraven:
lgtm

Issue #3924, opened 2017-02-12
Fonts become unreadable when playing through Steam


v1nsai:
![capture](https://cloud.githubusercontent.com/assets/410443/22864619/cdd2f130-f122-11e6-9abd-79ae050d5858.PNG)
[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/769497/opengl.txt)

### Observed behaviour
When running Pioneer as a non-Steam game, fonts become unreadable.  I've tried creating a simple mod to switch fonts and have tried several but they are always fine without Steam and unreadable when run through Steam.  Trying to run through Steam in order to get more options to reconfigure Xbox or Steam Controller.

### Expected behaviour
Fonts to be readable.

### Steps to reproduce
Add pioneer.exe to Steam through the Add a non-Steam game button, then run it.  Fonts are unreadable.

__My pioneer version (and OS):__
Pioneer build 20161129 (0c423b2)
Windows 10 fully updated

nozmajner:
That seems to be the same issue as #2931 visually, which was caused by some MSI overlay that displayed FPS and such.
I had the guess that if you disable the steam overlay, that might fix the issue, but there isn't any overlay, nor any option for it, so that won't work.

v1nsai:
Yea!  That's exactly like what I'm seeing.  Sorry for the dupe post, I searched for 'font' but should have also checked for 'text'.  I've been looking through the Steamworks API to see if compiling a few components in with the main branch might fix the problem but it's a bit over my head.......I'll contribute anything useful I can figure out to the other thread.

fluffyfreak:
@v1nsai I think if you disable the Steam overlay it should be fine.
I believe that it's because we request a strict forward compatible OpenGL context which forbids older style/purpose OpenGL calls which a lot of overlays rely on.

fluffyfreak:
Re-opened issue as I'm wondering if we should put some comments into the readme, wiki or FAQ?

v1nsai:
It would definitely be helpful to have something in the wiki.  I did lots of Googling for "Pioneer Space sim Steam" and variations on that and never found anything that said it was the overlay's fault until noz pointed me that way after I posted here.

I would imagine disabling this strict forward compatible OpenGL context would just require unticking a few flags in ./configure before compilation, right?  I've been coding for a few years but never tried game programming of any sort before, this seems like a good excuse to start learning :-D  

[...]

PR #3980, opened 2017-04-20
Initialise the TextureFont to 0

### LINKED ISSUE #2954: Fonts are "underlined" when multisampling is on
This happens mostly in game and sector view. Sometimes it's quite evidents, sometimes it's more subtle.
underline

Arch Linux with Mesa 10.2.1 Ati r600 drivers.
With my system the artifacts appears when mutisampling is enabled.

With Catalyst 13.35 drivers they appear with multisampling >= 2x

fluffyfreak:
Initialise the TextureFont to 0 to eliminate underlining even at high AA levels.

Fixes #2954 acceptably.

### DIFF
--- a/src/text/TextureFont.cpp
+++ b/src/text/TextureFont.cpp
@@ -624,6 +624,13 @@ TextureFont::TextureFont(const FontConfig &config, Graphics::Renderer *renderer,
 	m_mat.reset(m_renderer->CreateMaterial(desc));
 	Graphics::TextureDescriptor textureDescriptor(m_texFormat, vector2f(ATLAS_SIZE), Graphics::NEAREST_CLAMP, false, false, false, 0, Graphics::TEXTURE_2D);
 	m_texture.Reset(m_renderer->CreateTexture(textureDescriptor));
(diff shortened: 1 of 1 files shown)


impaktor:
related to #2954?

Tichy:
Unfortunately, now I had to use the integrated intel. I remembered that the issue appeared also on it, but apparently I was wrong, since I cannot reproduce it before trying the patch... 😞 
I applied it anyway, so at least I can tell you that it doesn't seem to break anything, but at the moment I can't see if it solves the problem on ATI R600.

fluffyfreak:
@impaktor it is indeed about #2954, hence why I think it fixes it 😁 

@Tichy thanks for testing it anyway, I'll merge it and hopefully it should fix/kludge it for everyone.

impaktor:
Que? Did you edit your top post after I posted, or was I really that confused/daft yesterday? I though this issue was about #3924. Oh, well. 

Sorry for confusion.

fluffyfreak:
no problem :)

PR #3986, opened 2017-04-25
Implement an alternative 2D Azimuthal Equidistant Projection Radar.

ecraven:
Change by right-clicking onto the 3D radar.

Mouse wheel changes the scanning distance.

radar-select
radar

I'd be glad about any comments and suggestions.

### DIFF
--- a/data/icons/icons.svg
+++ b/data/icons/icons.svg
@@ -704,9 +704,9 @@
      borderopacity="1.0"
      inkscape:pageopacity="1"
      inkscape:pageshadow="2"
-     inkscape:zoom="0.25000001"
-     inkscape:cx="1381.252"
-     inkscape:cy="2679.8317"
+     inkscape:zoom="1.4142136"
+     inkscape:cx="3635.9405"
+     inkscape:cy="3634.0335"
      inkscape:document-units="mm"
      inkscape:current-layer="layer1"
      showgrid="true"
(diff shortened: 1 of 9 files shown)


impaktor:
My thinking is that the end goal is to offer the player two different scanners to buy in equipment store? It will also be more clear that there are two different scanners available.

(alternative, one could have a simple/single scanner, and one "dual", that allows switching back and forth)

I wonder if the scanner could be like a movable and resizable widget? Probably then with all that extra (future) information on the sides anchored to it? Just throwing it out there. Then people could choose the size they want, that fits their eyes / monitor.

Also, perhaps when selecting a target, that should then auto-set the distance of the radar, to follow the target? (maybe it is like that now? I haven't tested).

BUG: Also, mouse continues to be "active" when holding in the right mouse button to navigate, in spite of it being hidden. For instance, when in combat, I manoeuvre, and then all of a sudden Pioneer opens the F3 screen, since, apparently, that's where the mouse ended up (hidden) when I swivelled my craft around. And I've seen before that hidden mouse activates tooltips, not sure if that's still the case though.

fluffyfreak:
I'd prefer it if the scanner was just a single instrument, that only thing this changes is how it's displayed on screen. It doesn't feel like that should require a completely different piece of hardware for the ship.

nozmajner:
Yeah, having two spearate equipment for them feels odd. Maybe if it's justified as different software, but even then it might be too much.

laarmen:
If we use today's standards, then yes. But by today's standards much of
Pioneer is doing things in an obsolete way!

impaktor:
OK.

Another question: how will players know about the new radar? I guess 95% don't read the wiki or documentation. One way is to make the new one default. That will cause some racket, I assume. 

Another is just use the new one in all screenshots, then players will wonder how to get it.

clausimu:
I like the new scanner in general. But it makes the UI surface appear very flat all of a sudden. Maybe just my impression?

impaktor:
Also, is blue the colour we want? If so, should not the colour of both radars be the same? Or is that just the thing we don't want?

fluffyfreak:
Colour: Blue seems to match the rest of the asthetic used by the imgui-UI and the new-UI screens.
The green is an oddball leftover from the early days of Radar which always used that green phospherous style screen we all think of.

Aside from the minor mouse click bug I'd happily merge this now :)

PR #3987, opened 2017-04-25
Do not set ARFLAGS on Darwin

### LINKED ISSUE #3839: Linking fails on macOS, ar has no flag D
As @tomyun mentioned in #3827, macOS does not have the `ar' flag D, so linking with our default flags crD fails.
Can anyone with macOS experience recommend a good set of ar flags?

ecraven:
Fixes #3839


ecraven:
@tomyun can you check whether this fixes building on mac os?

tomyun:
I confirm `HEAD` built with no problem on my mac. Thanks.

PR #3990, opened 2017-04-26
Add :space_station_type to CustomSystemBody

ecraven:
You can now use :space_station_type("new_ground") etc. to specify
the exact type of station model to use. The name is the json filename
without the extension.

### DIFF
--- a/src/galaxy/CustomSystem.cpp
+++ b/src/galaxy/CustomSystem.cpp
@@ -94,6 +94,14 @@ static int l_csb_new(lua_State *L)
 		lua_settop(L, 1); return 1;                        \
 	}
 
+#define CSB_FIELD_SETTER_STRING(luaname, fieldname)        \
+	static int l_csb_ ## luaname (lua_State *L) {          \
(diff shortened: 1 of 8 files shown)


fluffyfreak:
Nice

src/galaxy/CustomSystem.cpp
```diff
@@ -94,6 +94,14 @@ static int l_csb_new(lua_State *L)
 		lua_settop(L, 1); return 1;                        \
 	}
 
+#define CSB_FIELD_SETTER_STRING(luaname, fieldname)    \
+	static int l_csb_ ## luaname (lua_State *L) {        \
+		CustomSystemBody *csb = l_csb_check(L, 1);         \
+		std::string value = luaL_checkstring(L, 2);				 \
+		csb->fieldname = value;                            \
+		lua_settop(L, 1); return 1;                        \
+	}
```
impaktor: Nitpick, compared to the other `#defines` their line breaking `\` are straight, yours are "out of line".

ecraven: should be fixed now

PR #3991, opened 2017-04-26
Replace switch default with explicit list in RandomColor.

ecraven:

### DIFF
--- a/src/RandomColor.cpp
+++ b/src/RandomColor.cpp
@@ -102,8 +102,8 @@ namespace RandomColorGenerator
 			sMax = 55;
 			break;
 
-		default:
-			Output("RandomColor::PickSaturation: Unhandled case in switch: %i\n", luminosity);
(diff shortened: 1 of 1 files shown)


impaktor:
Was it @fluffyfreak who implemented this? I'll leave it to him to comment

fluffyfreak:
I did the port from C# in #3644 but that was a straight port, no attempts to understand the logic etc.
I'd guess that it's ok to leave the default values in those cases but that is a total guess. 

It's quite possible from the way it's organised that those cases are never possible/used in those methods but without running it I couldn't say for certain.

ecraven:
> It's quite possible from the way it's organised that those cases are
> never possible/used in those methods but without running it I couldn't
> say for certain.

The problem is they are actually used, and with the Output they produce
a lot of line noise on the console. Thus the proposed change ;)

PR #3993, opened 2017-04-27
Text background, indicator tooltips, delta-v numbers

ecraven:
Add a transparent black background to text
Add tooltips to various indicators
Add Delta-V percentage and number to reticule


nozmajner:
Preview:
blackbg
I think it looks good, and can be useful.

fluffyfreak:
Should be much clearer, merge it!

PR #3994, opened 2017-04-28
Add dashed circles and lines to imgui, use for 2d radar

ecraven:

### DIFF
--- a/contrib/imgui/imgui.h
+++ b/contrib/imgui/imgui.h
@@ -1177,6 +1177,7 @@ struct ImDrawList
 
     // Primitives
     IMGUI_API void  AddLine(const ImVec2& a, const ImVec2& b, ImU32 col, float thickness = 1.0f);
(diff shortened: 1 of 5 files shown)


nozmajner:
A preview of it:
radar_dashes
I think it looks good.

fluffyfreak:
It's nice, merge it!

ecraven:
hm.. ocornut of imgui mentioned that there might be performance issues, I'll need to investigate a bit more, not quite ready for a merge yet...

fluffyfreak:
@ecraven, care to elaborate on the issues?

ecraven:
See https://github.com/ocornut/imgui/pull/1118 for details.
The gist is (afaiu), this creates too many draw calls, it should be done with texturing

nozmajner:
If it's too expensive, then ditch it, it's not that important.
(could be nice for the orbit display too, but not a requirement)

impaktor:
Ah, I've been meaning to make this suggestion to the radar. Great!

laarmen:
So, any news ? (None on the imgui PR itself)

fluffyfreak:
I think we should close this one if it's not going anywhere.

We could do the dashed radar with just a regular square texture for this purpose.

ecraven:
not going anywhere at the moment ;)

PR #3995, opened 2017-04-28
Show in-space body indicators with lua

ecraven:


nozmajner:
[Reworked icons for ship](https://www.dropbox.com/s/27u6g8b6e21nlww/icons.svg?dl=0)s: (I kept the icon file layout, just replace and go :)
shipclass_minimal2
Their minimum size is 16x16, that's where they have 1px line thickness. (right three columns)
Left columns are showing them at 32x32
18x18 feels usable for the HUD:
22_icons

[Here are them for the ship market](https://www.dropbox.com/s/xor9qht59kr1uud/ship_icons.zip?dl=0) (until those screens got themselves on the newnew UI)
shipmarket

ecraven:
Should be ok to merge now, please test and comment

nozmajner:
Lookin' nice!
I noticed one thing: It shows the asteroid icon for Earth and Venus and Mars

data/pigui/game.lua
```diff
@@ -16,6 +16,9 @@ local icons = ui.theme.icons
 local reticuleCircleRadius = math.min(ui.screenWidth, ui.screenHeight) / 13
 local reticuleCircleThickness = 2.0
 
+-- settings
+local ASTEROID_RADIUS = 1500 -- rocky planets smaller than this are considered an asteroid, not a planet
```
impaktor: Maybe @fluffyfreak knows a better way/place to define this, since he did some asteroid work on a branch of his? Anyway, just wanted his eyes on that line. I have no complaint.

fluffyfreak: meh, i don't care.

ecraven:
@nozmajner, wrong icon for asteroid/planet should be fixed now, 1500km is the threshold for now ;)

nozmajner:
Just tried it, and I think it looks good. Didn't find any issues either. [So offblast!](https://www.youtube.com/watch?v=MNyG-xu-7SQ)

impaktor:
Btw, when I see the  [collision alert] in the moc up above by @nozmajner, I wonder if there will be something similar if a NPC locks you, or targets you. Maybe you need equipment to detect if someone has targeted you? Or built in? Where would that be shown?

ecraven:
> Btw, when I see the  [collision alert] in the moc up above by
> @nozmajner, I wonder if there will be something similar if a NPC locks
> you, or targets you. Maybe you need equipment to detect if someone has
> targeted you? Or built in? Where would that be shown?

How would you actually detect that they have targetted you, assuming
they use passive sensors (and if they are close enough for shooting at
you, they probably don't need active sensors).

nozmajner:
For passive detection, no way to spot. 
For active there are methods already IRL. Like you could place detectors on your hull all over the place whom would pick up active methods, like laser illumination or active radar sweep.

impaktor:
> How would you actually detect that they have targetted you,

I don't know, but it works like that in Iron Eagle  and Top Gun. My two fix points in life.

PR #3914, opened 2017-01-26
Modular ships

mike-f1:
Solve (at least): https://github.com/pioneerspacesim/pioneer/issues/2012, and I think it'is a step forward for https://github.com/pioneerspacesim/pioneer/issues/38

It works defining "modules" that are a subset of actual ships functionality, and can be used by
different objects (at least from Missile that now take advantage of re-using AI functions).

I post on issue 38 asking for feedback or comment but there aren't... Now I think it's time to see it merged ;-) , even if I think this could be a starting point for a (long) talk about what issue 38 imply.

TODO: Here and there I leave some comments on how things could be improved,
but here I would recap biggest things:

1. Actual modules are 2: Propulsion and FixedGuns, but I think could be better if others sub-system will be splitted out from ship, eg: shields, sensor and docking/landing routines (...as @robn suggested on IRC)

2. Actually I moved some AI handling things from Ship to DynamicBody, but it's not a "clean" split, it could be improved

3. Propulsion now take a reference for the whole Model only to use SetThrust() [Propulsion::Render]

...but all these things could be discussed and implemented separately.

impaktor's edit: closes #2012

PR #4000, opened 2017-04-29
Fix sqrt exception

### LINKED ISSUE #3996: Crash - exact same time each reload
### Observed behaviour
Current master ([9c8c112](https://github.com/pioneerspacesim/pioneer/commit/9c8c11276eaa10b7db020790093798ce150143c2)) crashes sometimes. It seems random, sometimes I can play for 10-20 minutes, sometimes it happens after a few minutes. Sometimes it only happens if you jump out to another system.
And if it happens, it will happen the exact same time, if you load back the save you did before the crash.
Win7 64bit, Decent nVidia GPU. Latest thirdparty, building with VS2015. No mods.

crash_tradeship

### Steps to reproduce
[Here's a save that should crash](https://www.dropbox.com/s/ni0itw54eqgcy6s/_quicksave_crash?dl=0) at 14:17:07 Jan 1. Just time accelerate and let the autopilot fly to Ceres (the already issued command). The crash happens exactly the same time, down to the second, if you just enable time acceleration. It differs a little if you alter the course to somewhere else, or force time acceleration.

I tried jumping to Epsilon Eridani - no crash there.
If I jump back to Sol, it crashes again, but at different times each time.

I tried removing the Tradeships.lua module. This only postponed the crash, and the crash also happens around the same time, but with little differences in this case, especially with forced time accel (around 06:25 Jan3).

If I remove the Tradeships.lua, jump out and jumped back, then the crash doesn't happen.
(I issued a goto to Ceres at each try for consistency)

Also tried removing every modul that would generate a ship, but that made no difference.

Output.txt seem to end on different stuff each time, and no error message.

I tried watching the other ships from the orbital map, but didn't saw anything out of ordinary. One ships achieving escape velocity seem to coincide, (PI-5143), but I guess that's just coincidence.

fluffyfreak:
Fixes #3996

### DIFF
--- a/data/lang/core/ga.json
+++ b/data/lang/core/ga.json
@@ -21,7 +21,7 @@
    },
    "AND_INDIGENOUS_MICROBIAL_LIFE" : {
       "description" : "",
-      "message" : "\" and indigenous microbial life.\""
+      "message" : "\" agus beathra miocróbach bundúchasach.\""
    },
    "AND_INDIGENOUS_PLANT_LIFE" : {
       "description" : "",
@@ -37,7 +37,7 @@
    },
    "AR_ATMOSPHERE" : {
       "description" : "",
-      "message" : "\" Argon atmosphere\""
+      "message" : "\" Atmaisféar argóin\""
    },
    "ASTEROID" : {
       "description" : "",
@@ -49,11 +49,11 @@
    },
    "AUTOPILOT_CONTROL" : {
       "description" : "Indicating that ship is flying under autopilot control",
-      "message" : "Autopilot control"
+      "message" : "Rialú uathphíolóta"
    },
    "AUTOPILOT_DOCK_WITH_STATION" : {
       "description" : "",
-      "message" : "Autopilot: Dock with space station"
+      "message" : "Uathphíolóta: Téigh chun glais le stáisiún spáis"
    },
    "AUTOPILOT_ENTER_HIGH_ORBIT_AROUND" : {
       "description" : "",
@@ -105,7 +105,7 @@
    },
    "BUTTON" : {
       "description" : "",
-      "message" : "\" Button \""
+      "message" : "\" Cnaipe \""
    },
    "CAMERA_BOTTOM_VIEW" : {
       "description" : "",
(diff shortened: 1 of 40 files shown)


fluffyfreak:
Doh! Just realised this is #4000

impaktor:
Looks like you included a commit too much.

`git reset HEAD~2`

mike-f1:
I would add 2 things:

1. It seems there's a "pre-division" by mass in Propulsion.h (so it's better to use `GetThrustUp`)... And this lead to improvements because in "pre-merge" of d2aab0530b there aren't members to GetThrust*

2. GetThrust* is even positive or negative, because mass is even positive: so the problem is on value parsed from ShipType (`m_linThrust[THRUSTER_UP]`): I don't know but to me it seems more logical let `GetThrust*` member family to return positive values, although this is more complicate to obtain

fluffyfreak:
@mike-f1 I'll leave cleanup to another PR to restrict this to just the fix but thanks for the suggestions 😄

mike-f1:
@fluffyfreak: No problem: I agree with you also because *I* did the mess, and I should solve it...

Only tell me if you think there's reasons to not ever return only positive value on GetThrust*...
Because (working on Propulsion) I get little confused (evidently :/ ) by the way this work and a step
ahead is to simplify it...

jaj22:
Thrust values should always be positive. Paired negative values may work accidentally for some routines, but others will react very poorly. I wouldn't have recommended merging this "fix", as it's just covering up another bug and other less visible problems will be caused elsewhere.

mike-f1:
Hi @jaj22 , nice to see you again... Has been time I miss you :/

> Paired negative values may work accidentally for some routines, but others will react very poorly

I'm afraid but I don't think it was like you say: before (evidently poorly designed) #3914 thrust values were in m_linThrust[] and (you can check) 3 of them were negative. More: they *are* negative... :/
...This probably cause an headache to me while I was coding, and *evidently* I did something wrong...

> another bug and other less visible problems will be caused elsewhere

I'm sorry for the troubles...

jaj22:
Yeah ok, forgot that forward/left/down were set negative in ShipType.cpp. It corresponds to the axis directions: x-right, y-up and z-backwards are the positive directions in the coordinate system.

Either way, if the convention changes then nothing will work correctly :P

PR #4005, opened 2017-05-01
Simple atmospheric crush pressure

fluffyfreak:
Crush the ship if it can't take the pressure as, optionally, defined in the ship type.

To set different crush depths for ships add a parameter like, eg: `atmospheric_pressure_limit : 17.0`, to their ship definition. This defaults to 5 * Earths Atmosphere at sea level, aka 5 bar.

Titan should have a surface pressure of 1.45 bar, and should be viable for a ship landing.
Venus should have a surface pressure of 90 bar, and is totally out of landing!

PR formerly known as: "*Pop goes the weasel!*"

### DIFF
--- a/src/Player.cpp
+++ b/src/Player.cpp
@@ -91,6 +91,24 @@ void Player::InitCockpit()
 		m_cockpit.reset(new ShipCockpit(cockpitModelName));
 }
 
+bool Player::DoCrushDamage(float kgDamage)
+{
+	bool r = Ship::DoCrushDamage(kgDamage);
+	// Don't fire audio on EVERY iteration (aka every 16ms, or 60fps), only when exceeds a value randomly
+	const float dam = kgDamage*0.01f;
+	if (Pi::rng.Double() < dam)
+	{
+		if (!IsDead() && (GetPercentHull() < 25.0f)) {
+			Sound::BodyMakeNoise(this, "warning", .5f);
+		}
+		if (dam < (0.01 * float(GetShipType()->hullMass)))
+			Sound::BodyMakeNoise(this, "Hull_hit_Small", 1.0f);
+		else
+			Sound::BodyMakeNoise(this, "Hull_Hit_Medium", 1.0f);
+	}
+	return r;
+}
+
 //XXX perhaps remove this, the sound is very annoying
 bool Player::OnDamage(Object *attacker, float kgDamage, const CollisionContact& contactData)
 {
--- a/src/Player.h
+++ b/src/Player.h
@@ -20,6 +20,7 @@ public:
 	Player(const ShipType::Id &shipId);
 	Player() {}; //default constructor used before Load
 	virtual void SetDockedWith(SpaceStation *, int port) override;
+	virtual bool DoCrushDamage(float kgDamage) override final; // overloaded to add "crush" audio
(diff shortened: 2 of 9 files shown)


impaktor:
Do we show pressure? Would be nice if it got "in the red" before "pop".  

Also, some squeeky submarine hull crushing sounds would be _really_ nice, for future.

fluffyfreak:
@impaktor We do indeed show pressure and it's scary how quickly that can go up when doing something like fuel scooping. I may have to raise the default somewhat.

fluffyfreak:
@impaktor I quickly implemented the colour changing for the overlay text, so it progresses from green->yellow->red->KABOOM

It would be good if people could try fuel scooping, and seeing how it plays when doing stuff like trying to descend into Venus's atmosphere... you will **NOT** be able to land on Venus anymore as your default crush "*depth*" is `5 bar`, and Venus surface pressure should be `~90 bar`.

impaktor:
Didn't try any fuel scooping. Just tested I could die. What's the noise that keeps playing after Kaboom?

fluffyfreak:
I heard nothing, in fact, I didn't even get a kaboom :/

nozmajner:
I got a kaboom yesterday skimming Hercules (around atm 4.5, it gets red). I tried with the Mola Ramsayi (didn't set the pressure toleranc of it.). But I'll try it again in the evening to make sure I did'nt actually hit the ground. But I think I didn't because the fuel scoping didn't started.

ecraven:
worked fine for me just now, about 50km above jupiter, pressure red, kaboom!

ecraven:
Usable atmospheric range on Jupiter is about 80km (0.01 atm) to 45km (5 atm). That is not very much, considering you have to aim very exact to go fast too...

fluffyfreak:
FFS, ok I tried this:

- Doubled the crush pressure to 10 atmosphere (_~90 metres beneath the sea on Earth_)
- halved the speed needed 1000
- halved the density to 0.5
- reduced the angle dot product to 0.9 fro 0.95

Fuel scooping on Jupiter is **still** insanely difficult.

I would like to have this in there since an infinitely strong crush-proof spaceship doesn't seem right but teaching people how to scoop fuel again is... well I dunno, what do people think?

impaktor:
I agree with your sentiment, @fluffyfreak. 

I still haven't tried scooping, but just throwing out a random idea: what if there's a very large region where you're in "the yellow" (and red) before being crushed, and have them extend to higher pressures. I.e. still maybe be in the green for 10-100 atmosphere (if it's a ship where it makes sense), and then large yellow and red zone that will allow player to adjust.

nozmajner:
What if scooping is made "easier"? Like you would need a lot less pressure for it to start, but you would scope proportionally less H2? It feels odd that you just pick up a ton of H2 in a matter of seconds.

fluffyfreak:
I've changed it so that you are damaged rather than instantly destroyed, a change that I might also apply to the heating later on.

This gives you time to try and recover from your situation.

@nozmajner I've already widened the scope of when and how you fuel scoop for this, and doubled the atmospheric pressure you can withstand from my original 5 atmospheres (_5 * Earth at sea level_) 😞 not sure what values make sense at this stage.

fluffyfreak:
I have decided to be a bit naughty with this and push it into the build 😁

PR #3989, opened 2017-04-26
The Great Disinheritance: make Propulsion a component

ecraven:
This is the start of an effort to remove the inheritance relationships and replace them by composition.

PR #4009, opened 2017-05-04
Move main menu buttons to lua

ecraven:
mainbuttons

This is temporary, most of these buttons should be moved to other UI elements soon.


data/pigui/game.lua
```diff
@@ -789,24 +795,24 @@ local function displayOnScreenObjects(should_show_label)
...
-				player:SetNavTarget(v.mainBody.body)
-			end
-			for k,v in pairs(v.others) do
-				ui.icon(getBodyIcon(v.body), size, colors.frame)
-				ui.sameLine()
-				if ui.selectable(v.label, v.body == navTarget, {}) then
-					player:SetNavTarget(v.body)
-				end
-			end
+							 local size = Vector(16,16)
```
fluffyfreak: What's going on with this random re-indentation?

ecraven: that's emacs' automatic indentation, I ran it over the entire file (must have forgotten to do it before)

impaktor:
crash: Camera.cpp L52, assert failed, when clicking "launch", after mucking about in various star system / map views. Might be unrelated to that, haven't _not_ gotten the crash yet.

Also, normally, you can cycle through the three tabs in "system overview" icon like econ info, demography, planetary info, by repeatedly clicking the icon; if I remember master branch correctly.

impaktor:
Also, I miss not seeing if I'm in set speed, autopilot, or manual flight mode.

...and the hyperspace icon, used to have three states: 
1. grayed out, i.e. target set, but not physically possible to engage, e.g. landed state
2. forbidden
3. allowed

seems like we only have 2. and 3. now. Is this a concious decision?

ecraven:
> crash: Camera.cpp L52, assert failed, when clicking "launch", after mucking about in various star 
> system / map views. Might be unrelated to that, haven't not gotten the crash yet.

thanks for finding it. should be fixed ;)

ecraven:
@impaktor: thanks for checking, should all be fixed now.

impaktor:
Tested again. 

I still don't see my flight state, is this on purpose? E.g. if I'm in manual flight mode.

ecraven:
> I still don't see my flight state, is this on purpose? E.g. if I'm in manual flight mode.

This should be fixed now.

fluffyfreak:
@ecraven merging #3989 has caused a conflict which will need fixing, I can't easily do this from work, so once it's fixed go ahead and merge

PR #4012, opened 2017-05-08
The Great Disinheritance: Make FixedGuns a component.

ecraven:
Continue with what #3989 started.


fluffyfreak:
Some good changes, I hadn't noticed that FixedGuns (_or Propulsion for that matter_) were inherited via DynamicBody, probably I should have queried that before in the original PR 😞 my bad.

impaktor:
With a 5 second drive-by review, I couldn't see anything ojectionable.

PR #4021, opened 2017-06-04
Translated a portion of the messages

WKFO:

### DIFF
--- a/data/lang/module-searchrescue/tr.json
+++ b/data/lang/module-searchrescue/tr.json
@@ -1,15 +1,15 @@
 {
    "AU" : {
       "description" : "",
-      "message" : "AU"
+      "message" : "AB"
    },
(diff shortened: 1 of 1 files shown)


impaktor:
Thanks, but this can not be merged, as it will be over written by our automatic bot that pulls down the translations from transifex. If you want to translate you must do it through transifex, (please see our wiki).
https://pioneerwiki.com/wiki/Translations
You can upload translated file there.

nozmajner:
Thank you for the effort! I see you translated quite a few lines.
Didn't you knew that our localizations are done trough Transifex? Much simpler and comfortable that way for the translators. No need to do PRs, just translate the strings, and they are pulled each day automatically. And new strings go up quite fast for translation too.
You can find it here: https://www.transifex.com/pioneer/pioneer/dashboard/

PR #4023, opened 2017-06-10
Modify jump range so time is linear, not quadratic with distance

clintonmead:
I found it a bit silly micromanaging small jumps to get to my destination on time, so I've just modified the jump time formula to be a simple linear function of distance. It's set so that the distance to jump the max range is the same as previously, but as the function is quadratic all smaller jumps now take longer. This does make the game slightly more difficult although I'm having no trouble completing missions in the early game with the hard start at Barnard's Star.

This behavior more closely matches what was previously in Frontier and FFE and hence will probably be less surprising to those who have experience from them games.

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -212,7 +212,7 @@ end
 HyperdriveType.GetDuration = function (self, ship, distance, range_max)
 	range_max = range_max or self:GetMaximumRange(ship)
 	local hyperclass = self.capabilities.hyperclass
-	return 0.36*distance^2/(range_max*hyperclass) * (3600*24*math.sqrt(ship.staticMass + ship.fuelMassLeft))
+	return 0.36*distance/hyperclass * (3600*24*math.sqrt(ship.staticMass + ship.fuelMassLeft))
 end
 
 -- range_max is optional, distance defaults to the maximal range.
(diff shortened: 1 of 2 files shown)


impaktor:
> but as the function is quadratic all smaller jumps now take longer.

"but as the function is __linear__ all smaller jumps now take longer." you mean?

Not sure I like this, though. 

I think it's more interesting to string small jumps together and plan the trip to "hack" a faster arrival in time, while actually going longer in space.

clintonmead:
Yes sorry now it's linear the shorter jumps take longer. 

The current situation just amounts to solving a shortest path problem in your head repeatedly, and I'm not sure how that adds to gameplay. It's a rather mechanical process which a computer program can solve trivially.

Frontier didn't add this complication and I don't think it lost anything from not doing so.

fluffyfreak:
I disagree with this, there's a branch I ahve with route plotting that does as you say:
>a rather mechanical process which a computer program can solve trivially

Then the player doesn't need to do the calculations and you can optimise it for different definitions of shortest path, e.g: number of  jumps, time, fuel etc.

Also we're not Frontier. Pioneer started out as a clone of it but we're moving ever further from it as we grow into something new and different from the direction that Frontier are going with the Elite series.

impaktor:
What I like with current system is:

1. if you're lazy then just click your destination, but if pressed for time
you can "micro-manage" the navigational path and thus find a shorter (in
time). I like that. And that to me is interesting game play. Sure "it's
mechanical", but so is trading. Or (a little less so) bounty hunting.

2. OK, this isn't really an argument but one could "tech-talk" some motivation
for this, like surface of a sphere goes like "r**2, energy-bubble, reverse
shield polarity, beam me up Scott!".

OK, so feels like this PR is overruled, as I'm not the onlyone feeling this s a major change in game dynamics and nothing we want to merge "willy nilly".

Issue #4019, opened 2017-06-03
Crash when jump to Gliese 783


robothauler:
### Observed behaviour
Crash when jump to Gliese 783 B
Output:
error: [string "[T] @pigui/game.lua"]:674: invalid order function for sorting
stack traceback:
        [C]: in function 'sort'
        [string "[T] @pigui/game.lua"]:674: in function 'displayOnScreenObjects'
        [string "[T] @pigui/game.lua"]:746: in function 'fun'
        [string "[T] @pigui/pigui.lua"]:42: in function 'window'
        [string "[T] @pigui/game.lua"]:739: in function 'fun'
        [string "[T] @pigui/pigui.lua"]:71: in function 'withStyleColors'
        [string "[T] @pigui/game.lua"]:738: in function 'game'
        [string "[T] @pigui/pigui.lua"]:83: in function <[string "[T] @pigui/pigui.lua"]:81>

### Steps to reproduce
Start at Earth
Jump to Gliese 783 B

__My pioneer version (and OS):__
pioneer git (363e44165)
Debian 9.0 x86_64

fluffyfreak:
Can confirm, I just managed to repro it, works fine if you jump to `Gliese 783 A` but crashes if you jump to `Gliese 783 B`

fluffyfreak:
@ecraven the offending block of code seems to be

>	for k,v in pairs(onscreen) do
>
>		table.sort(v.others, function(a,b) return a.body:IsMoreImportantThan(b.body) end)
>
>		local coords = v.mainBody.screenCoordinates
>
>		ui.addIcon(coords, getBodyIcon(v.mainBody.body), colors.frame, iconsize, ui.anchor.center, ui.anchor.center)`

fluffyfreak:
Fixed by #4024

PR #4024, opened 2017-06-11
Fix for incorrectly implemented comparison function crashing Pioneer upon entering system

clintonmead:
Entering certain systems (it seemed to be binary ones) caused Pioneer to crash with this message:

error: [string "[T] @pigui/game.lua"]:674: invalid order function for sorting
stack traceback:
	[C]: in function 'sort'
	[string "[T] @pigui/game.lua"]:674: in function 'displayOnScreenObjects'
	[string "[T] @pigui/game.lua"]:746: in function 'fun'
	[string "[T] @pigui/pigui.lua"]:42: in function 'window'
	[string "[T] @pigui/game.lua"]:739: in function 'fun'
	[string "[T] @pigui/pigui.lua"]:71: in function 'withStyleColors'
	[string "[T] @pigui/game.lua"]:738: in function 'game'
	[string "[T] @pigui/pigui.lua"]:83: in function <[string "[T] @pigui/pigui.lua"]:81>

I haven't coded in Lua before, but it seems the error is regarding the sorting function.
A quick google search revealed the sorting function should return false when the items are equal.

I've reordered the cases in "l_body_is_more_important_than" and this seems to resolve the issue.

### DIFF
--- a/src/LuaBody.cpp
+++ b/src/LuaBody.cpp
@@ -224,10 +224,10 @@ static int l_body_is_more_important_than(lua_State *l)
 	bool result = false;
 
 	if(a == b && a != Object::Type::PLANET) result = body->GetLabel() < other->GetLabel();
-	else if(a == Object::Type::STAR) result = true;
 	else if(b == Object::Type::STAR) result = false;
-	else if(a_gas_giant) result = true;
+	else if(a == Object::Type::STAR) result = true;
 	else if(b_gas_giant) result = false;
+	else if(a_gas_giant) result = true;
 	else if(a == Object::Type::HYPERSPACECLOUD) result = false;
 	else if(b == Object::Type::HYPERSPACECLOUD) result = true;
 	else if(a == Object::Type::MISSILE) result = false;
@@ -240,8 +240,8 @@ static int l_body_is_more_important_than(lua_State *l)
 	else if(b == Object::Type::SPACESTATION) result = true;
 	else if(a_planet && b_planet) result = body->GetLabel() < other->GetLabel();
 	else if(a_moon && b_moon) result = body->GetLabel() < other->GetLabel();
-	else if(sb_a && sb_a->IsPlanet()) result = true;
 	else if(sb_b && sb_b->IsPlanet()) result = false;
+	else if(sb_a && sb_a->IsPlanet()) result = true;
 	else Error("don't know how to compare %i and %i\n", a, b);
 
 	LuaPush<bool>(l, result);


clintonmead:
I'm not sure how the changes break anything... the only time the changes make a difference is when the objects are equal, in which case it now returns false instead of true. When the objects are not equal the return value is no different to previously.

laarmen:
I can't really comment on the validity of the fix, but if we know that equality implies returning false, why not just checking for it explicitly?
The function isn't really readable right now (I'm not sure how it works after reading it twice), so I'd rather have a patch enhancing the readability as well as fixing a bug.

Plus, the format of the patch (which is basically switching whether we test a or b first) makes me think something is wrong : the return value is not supposed to be modified by such changes when a and b are the same.

clintonmead:
The return value is changed when both are equal because in an if-elseif chain we only use the first matching clause in the chain, so when two match (such as when a and b are equal) order does matter.

laarmen:
I don't think my point came through : relying on a elseif chain order is
inherently dangerous and a relatively bad idea. I do see why the patch
works, I just really don't like it.

impaktor:
@clintonmead Btw, it's not fun to have your first two PRs come under fire, so please don't take it personally. I'm really happy to see you open PRs and work on the code, so contributions are very welcome. And here it seems you've identified the cause of the bug, so almost there.

What @laarmen says sounds logical, but beyond that I have no opinion on this PR.

laarmen:
@impaktor is wise beyond his years. I apologize if I made you feel attacked in any ways, it wasn't my intention at all. It's just that lately I find myself caring much about code quality, mostly because I like to follow what's going on in Pioneer but I haven't time to dig through the code to understand a diff :-).

I'd rather like my concerns addressed before we merge this code, but that is in no way blocking, nor mandatory. As long as your patch works and fixes bugs, I'm fairly sure somebody will merge it.

fluffyfreak:
Ok my 2p since I'll probably pull the merge trigger:

@ecraven I don't think we can reorder the enum easily as other parts of the code depend on it's numerical order already.

The if-else mess looks like it's the only straight forward way of doing that equality test since the "equality" or ordering of a random body is itself an odd concept.

@laarmen this is a reasonable fix if it works, I think for minor bug fixes that just repair already bad code that is all that should be required. Otherwise it puts too much pressure on new/casual contributors.

@clintonmead thanks for the bug fix, I'll try to test and merge it today 😄

fluffyfreak:
This fixes #4019 as well.

fluffyfreak:
@clintonmead what would you like to be known as in the AUTHORS.txt?
We add everyone who has contributed too it if they like.

clintonmead:
"Clinton Mead" will be fine as the name thanks.

Didn't feel attacked at all btw, and whilst there probably is a nicer way to do the ordering other than an if-else chain I don't know the code enough not to break new things with that approach.

fluffyfreak:
Done, you're officially one of our contributors :D
Thanks again.

laarmen:
See, I told you somebody would merge it ;-).

Welcome !
And yes, I know, I should have yelled at @ecraven for this function, not @clintonmead , but to be fair it probably came in a huge PR and passed through the cracks. To be fairer, the equipment code is probably much uglier. And to be the fairest of them all, I'm annoyed with the overall architecture that makes it necessary to write such code, more than with the code itself.

I still profundly dislike the way we use the enums in Lua. But haven't found a better way just yet.

</rant>

PR #4025, opened 2017-06-12
Move time and time accel window to lua

ecraven:
Moved the time and time acceleration display to lua.

time-window

### DIFF
--- a/data/pigui/modules/time-window.lua
+++ b/data/pigui/modules/time-window.lua
@@ -1,0 +1,68 @@
+local Engine = import('Engine')
+local Game = import('Game')
+local ui = import('pigui/pigui.lua')
+local Vector = import('Vector')
+local Color = import('Color')
+local Lang = import("Lang")
+local lc = Lang.GetResource("core");
+local lui = Lang.GetResource("ui-core");
+local utils = import("utils")
+local Event = import("Event")
+
+local player = nil
+local pionillium = ui.fonts.pionillium
+local colors = ui.theme.colors
+local icons = ui.theme.icons
+-- names of the keys in lang/core/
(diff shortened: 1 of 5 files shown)


data/pigui/modules/time-window.lua
```diff
@@ -0,0 +1,65 @@
...
+	local year, month_num, day, hour, minute, second = Game.GetDateTime()
+	local month = lc[months[month_num]]
+	local date = string.format("%04i %s %i - %02i:%02i:%02i", year, month, day, hour, minute, second)
+	local current = Game.GetTimeAcceleration()
+	local requested = Game.GetRequestedTimeAcceleration()
+	function accelButton(name, tooltip)
+		local color = bg_color
+		if requested == name and current ~= name then
+			color = colors.white
+			tooltip = 'Requested ' .. tooltip
```
laarmen: I don't know if we had it before, but what about the localization?

data/pigui/modules/time-window.lua
```diff
@@ -0,0 +1,65 @@
...
+		end
+		local res = false
+		if ui.coloredSelectedIconButton(icons['time_accel_' .. name], button_size, current == name, frame_padding, color, fg_color, tooltip) then
+			Game.SetTimeAcceleration(name)
+			res = true
+		end
+		ui.sameLine()
+		return res
```
laarmen: I don't see this return value used anywhere.

data/pigui/modules/time-window.lua
ecraven: Thanks for the catch, I supposed this was going away anyway, and didn't look at it again.
I'll fix it.

data/pigui/modules/time-window.lua
ecraven: Ah, this was used in an alternative version, which I removed. I'll delete it ;)

ecraven:
For Shift-F1..F5 to work, we need to merge #4026 first ;)

laarmen:
About the two commits, rebase gone wrong? Please squash them before
@impaktor sees them and goes berserk ;-).
Otherwise, there's another bit of l10n missing, the 10x time labels.

PR #4026, opened 2017-06-13
Support modifier keys in lua.

ecraven:
Add support for the missing modifier keys in lua.

Only run the main menu keys when pressed without any modifier (as a preparation for S-F1 etc. for time accelleration).


laarmen:
Looks good to me. One question, though. Why systematically wrap the pigui
variables / calls?

ecraven:
> Why systematically wrap the pigui variables / calls?

The idea was to have some functions that are *not* exposed to the user directly, and to avoid having to use . and : both in lua (by just wrapping :-needing functions in a closure). I've noticed too late that we normally capitalize lua functions, I'll refactor that in one huge go later ;)

PR #4029, opened 2017-06-22
Fix all GCC 7.1.1 warnings.

ecraven:
- Use explicit Json::Value::Int.
- Resize some snprintf buffers.
- Ignore -Wfloat-equal where necessary.


fluffyfreak:
Cool, looks good to me

PR #4031, opened 2017-06-22
Decreased density for speedlines

WKFO:
Incredibly minor change; less denser generation for speedlines, avoiding speedlines to look like they appear at the same relative location to the ship. Looks almost the same as before. Let's avoid making the area too large though, it also looks bad.
As you see I didn't touch the velocity-relevant parts.

### DIFF
--- a/src/SpeedLines.cpp
+++ b/src/SpeedLines.cpp
@@ -7,8 +7,8 @@
 #include "Pi.h"
 
 const float BOUNDS     = 2000.f;
-const int   DEPTH      = 8;
-const float SPACING    = 500.f;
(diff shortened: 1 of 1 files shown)


nozmajner:
It's a subtle effect especially on lower speeds, but I think it's a good change (bit cleaner visually).
For those who don't want to build:
hudlines
Original:
hudlines_orig

@ecraven Maybe in the future there could be a slider for this?

laarmen:
LGTM on the code side.

impaktor:
:+1:

I think a slider would be overkill, but exposing it in some config file, could be nice.

PR #4033, opened 2017-06-24
Stricter check for complex ecosystems (...and I need help)

WKFO:
Complex ecosystems can no longer appear on worlds with atmospheres that cannot support "complex" life as we know it (photosyntetic/kemosyntethic plants, multicellular animals with specialised segments and organs etc... Biology, you know...). Some forms of life may still exist; perhaps we need a new description for these worlds. Also, some optimization for the code would be nice. I am not very good at working in a clean way. Throwing a random boolean value into the code feels really absurd.

I got no errors with the build but you may want to check again, just in case.

I examined StarSystemGeneration.cpp/.h and found a part that checks if life can exist or not on some planet; but I couldn't figure out where I could add THIS to that part. If someone can do this, it would be better to keep the project better organised.

### DIFF
--- a/src/galaxy/StarSystem.cpp
+++ b/src/galaxy/StarSystem.cpp
@@ -186,7 +186,8 @@ const float StarSystem::starScale[] = {  // Used in sector view
 	4.0f  // Supermassive blackhole
 };
 
-bool supportComplexLife;
+int supportComplexLife;
+int LifeChanceMultiplier = (rand() % static_cast<int>(4)); //random value between 0 and 3
 
 SystemBody::BodySuperType SystemBody::GetSuperType() const
 {
@@ -338,35 +339,37 @@ std::string SystemBody::GetAstroDescription() const
 			else thickness = Lang::VERY_DENSE;
 
 			if (m_atmosOxidizing > fixed(95,100)) {
-				supportComplexLife = true;
+				supportComplexLife = 9/10 * LifeChanceMultiplier;
 				s += Lang::WITH_A+thickness+Lang::O2_ATMOSPHERE;
 			} else if (m_atmosOxidizing > fixed(7,10)) {
-				supportComplexLife = true;
+				supportComplexLife = 7/10* LifeChanceMultiplier;
 				s += Lang::WITH_A+thickness+Lang::CO2_ATMOSPHERE;
 			} else if (m_atmosOxidizing > fixed(65,100)) {
-				supportComplexLife = true;
+				supportComplexLife = 6/10* LifeChanceMultiplier;
 				s += Lang::WITH_A+thickness+Lang::CO_ATMOSPHERE;
 			} else if (m_atmosOxidizing > fixed(55,100)) {
-				supportComplexLife = true;
+				supportComplexLife = 6/10;
 				s += Lang::WITH_A+thickness+Lang::CH4_ATMOSPHERE;
 			} else if (m_atmosOxidizing > fixed(3,10)) {
-				supportComplexLife = true;
+				supportComplexLife = 3/10;
 				s += Lang::WITH_A+thickness+Lang::H_ATMOSPHERE;
 			} else if (m_atmosOxidizing > fixed(2,10)) {
-				supportComplexLife = true;
+				supportComplexLife = 0;
 				s += Lang::WITH_A+thickness+Lang::HE_ATMOSPHERE;
 			} else if (m_atmosOxidizing > fixed(15,100)) {
-				supportComplexLife = false;
+				supportComplexLife = 0;
 				s += Lang::WITH_A+thickness+Lang::AR_ATMOSPHERE;
 			} else if (m_atmosOxidizing > fixed(1,10)) {
-				supportComplexLife = false;
+				supportComplexLife = 0;
 				s += Lang::WITH_A+thickness+Lang::S_ATMOSPHERE;
 			} else {
-				supportComplexLife = true;
+				supportComplexLife = 2/10;
 				s += Lang::WITH_A+thickness+Lang::N_ATMOSPHERE;
 			}
 		}
-		if (m_life > fixed(1,2) && supportComplexLife) {
+		
+		lifeCahnceMultiplier
+		if (m_life > fixed(1,2)) {
 			s += Lang::AND_HIGHLY_COMPLEX_ECOSYSTEM;
 		} else if (m_life > fixed(1,10)) {
 			s += Lang::AND_INDIGENOUS_PLANT_LIFE;
--- a/src/galaxy/StarSystem.cpp
+++ b/src/galaxy/StarSystem.cpp
@@ -186,6 +186,8 @@ const float StarSystem::starScale[] = {  // Used in sector view
 	4.0f  // Supermassive blackhole
 };
 
+bool supportComplexLife;
+
 SystemBody::BodySuperType SystemBody::GetSuperType() const
 {
 	PROFILE_SCOPED()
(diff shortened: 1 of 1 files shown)


WKFO:
I think I the new decription we need would be something like: "...planet with primitive multicellular life". I don't want to go too scientific but I don't know what else I can say.

shammyx:
What do you think about an integer, increasing the better/more similar to earth an atmosphere is? That way we could also add multiple "stages" of life with different descriptions.

WKFO:
I think that would be better than this, I will have a look.

impaktor:
Please use `commit --amend` and `push -f` when re-doing a commit, see our wiki for git help.

ecraven:
For the long term, I'd love to go with generating the actual composition
of the atmosphere, and then deducing things from that. But that's
probably a scientific long-term project, nothing quick or easy ;)

WKFO:
> Please use commit --amend and push -f when re-doing a commit, see our wiki for git help.
Sorry, I didn't know. I will be more careful later on.

impaktor:
No problem. Thanks for contributing.

WKFO:
How do we generate random integers in Pioneer? The LifeChanceMultiplier must be a random integer value between 0 and 3 (including 0 and 3). I don't think rand() or srand() will work here. I need someone to do that, please....

I can't add Windows libraries because it is the 'src' folder, not the win32.

impaktor:
You'll find plenty of functions in src/Random.h

WKFO:
Can't we do the commit --append thing on the website... 

git terminal is having trouble confirming that pioneer or pioneer-master is a github repository. I just need to change a single line, can't I do this here?
Further deep info, I will change 190th line with:
`int LifeChanceMultiplier = Int32(const int 0, const int 3);`
...as I found in random.h as `Int32(const int min, const int max);`

I am really sorry for creating trouble out of nowhere.

WKFO:
Oh my god, I misclicked...

impaktor:
Don't worry too much, without doing mistakes no progress is made, and nothing is learned.

> I am never going to touch anything one more time.

Doing mistakes is part of learning. No one will be upset with you.

github isn't as powerful as *"driving git directly". 

Depending on how much git you want to learn, ideally, these commits aught to be rebased (see pioneer wiki on git for description), but we/I can do that at merge time, if you don't do it. As I said, don't worry about it, unless you want to play around with git. It can be daunting and much to take in at once. 

We can always fix the commits for you at merge time.

WKFO:
TO DO:

Check failed because I miserably failed to remember to put a semicolon at line: 380.

Language files in lang\core has to be updated as a dependancy of the new description (addition of "AND_PRIMITIVE_MULTICELLULAR_LIFE", empty descrpition, message "and primitive multicellular life".

The random number generator at line: 190, may be completely deleted in case of a compilation error.

src/galaxy/StarSystem.cpp
```diff
@@ -186,6 +186,9 @@ const float StarSystem::starScale[] = {  // Used in sector view
 	4.0f  // Supermassive blackhole
 };
 
+int supportComplexLife;
+int LifeChanceMultiplier = (rand() % static_cast<int>(4)); //random value between 0 and 3
```
ecraven: Why are these global variables and not local to GetSuperType?

src/galaxy/StarSystem.cpp
```diff
@@ -186,6 +186,9 @@ const float StarSystem::starScale[] = {  // Used in sector view
...
+
```
ecraven: spurious newline

src/galaxy/StarSystem.cpp
```diff
@@ -336,34 +339,45 @@ std::string SystemBody::GetAstroDescription() const
...
+				supportComplexLife = 0;
 				s += Lang::WITH_A+thickness+Lang::S_ATMOSPHERE;
 			} else {
+				supportComplexLife = 2 * LifeChanceMultiplier;
 				s += Lang::WITH_A+thickness+Lang::N_ATMOSPHERE;
 			}
 		}
-
-		if (m_life > fixed(1,2)) {
+
```
ecraven: spurious newline

src/galaxy/StarSystem.cpp
```diff
@@ -336,34 +339,45 @@ std::string SystemBody::GetAstroDescription() const
...
 		}
-
-		if (m_life > fixed(1,2)) {
+		
+		if (m_life > fixed(1,2) && supportComplexLife > 7) {
 			s += Lang::AND_HIGHLY_COMPLEX_ECOSYSTEM;
-		} else if (m_life > fixed(1,10)) {
+		} else if (m_life > fixed(1,10) && supportComplexLife > 4) {
 			s += Lang::AND_INDIGENOUS_PLANT_LIFE;
+		} else if (m_life > fixed() && supportComplexLife > 2) {
```
ecraven: Not about this PR specifically, but isn't this a very bad idea, to just return a string? Shouldn't we capture what sort of live we have at least in an `enum` and use that for generating the string?

laarmen: Seconded.

Meanwhile, I'll mumble and grumble about adding more gameplay enums to the C++ core, "this should all be in Lua", etc... ;-)

src/galaxy/StarSystem.cpp
laarmen: Depends. Is the intention here that at equivalent oxidizing level all planet should support complex life the same way, or should there be some variation ? I'm enclined to have the latter. If so, then yes the variables should definitely be inside the function body, as otherwise all bodies would have the same multiplier.

But in any case, this should be computed only once during the body initialization, not in the getter.

src/galaxy/StarSystem.cpp
fluffyfreak: Thirded, also I like having enums in where they're used, which here is the C++ core :P

laarmen:
I say "Move the enums, the code will follow!" ;-)

Of course I'm not volunteering to do any of this. Last time I tried it took
me around 2 years.

fluffyfreak:
@WKFO aside from the off-topic talk is there any reason that you closed this?

It might need some tweaks and changes but it seems like a reasonable idea.

WKFO:
Despite having the some scientific knowledge I couldn't handle the coding part of it but someone else can fix it and make it a proper commit of course.

laarmen:
I'm with @fluffyfreak on this one. The idea is good, and it's by making mistakes that one learns and improves one's ability to code, for instance.

There are several of us that are willing to help you get through this contribution, both on the Git side of things ( @impaktor is usually quite good at this) and on the more concrete coding aspects. Once the code is in a PR, it often isn't in its final step and needs reworking to address the comments of reviewers, it is an integral part of the process.

Going back to your code, first I need to know : is it intentional that there is no variation of life support between planets with equivalent atmosphere ? If so, why do we need a random number ?
Once I'm entirely sure we're talking about the same thing it'll get easier.

Oh, and one last thing : do not take our comments as a way to criticize you. Our goal is both to improve the quality of the code and to help you improve your coding skills so that you'll keep helping us in the future.

WKFO:
Yes, the if the atmospherical composition is the same the life support would be the same but life doesn't evolve the same way in different biomes on Earth. We are talking about whole different planets here. You know, atmosphere isn't the only thing that affects how the life will exist. But until we can implement all the soil/biome etc. (maybe in the next ceuntry?) we still need a bit randomness I think. But a much "proper randomness".

Instead of murdering Github I think we can talk about the code in... I don't know, the dev forum?

I want to keep contributing the best way I can, of course, and it is OK to criticize me as much as anyone wants.  I am not going to get upset about making mistakes. I have made mistakes my whole life :)

PR #4038, opened 2017-06-28
New music for ship-destroyed

WKFO:
Hungarian Dance No. 5 with minor modifications from me. Could also be a docking music with slower speed and no strings.

I am just bored of hearing the same things over and over in the game so I will keep throwing things in the dev forum post. I can't bring back s20dan or emajogi but I will do my best.


fluffyfreak:
I'm not familiar with music licensing but do you know the origin of the piece? Who performed it etc? 

It's very discordant to my ears but then we all have different tastes and it's is for the tombstone loop! :)

WKFO:
The piece is royalty free like classical pieces we already have in Pioneer. Composed by Johannes Brahms at the end of 19 ceuntry. The stuff we use is digitally produced by us so it is impossible for any orchestra to claim rights.

nozmajner:
Shouldn't we merge this?

fluffyfreak:
Yes

impaktor:
I realized what sound I'd like for the death scene in pioneer: 
https://www.youtube.com/watch?v=ji2Wcc2ZHJY&t=1m39s
😉 

Honestly, would be a fun easter egg, like 5% chance of playing that, instead of actual?

PR #4039, opened 2017-06-29
Added Swap button for sector map

shammyx:
Added a button below "Selected system" info to swap the selected system
and the hyperspace target in the sector map.


src/SectorView.cpp
```diff
@@ -645,6 +652,20 @@ void SectorView::SetSelected(const SystemPath &path)
...
+{
+	SystemPath tmpTarget = GetHyperspaceTarget();
+	SetHyperspaceTarget(GetSelected());
+	if (m_automaticSystemSelection) {
+		GotoSystem(tmpTarget);
+	} else {
+		RefCountedPtr<StarSystem> system = m_galaxy->GetStarSystem(tmpTarget);
+		SetSelected(system->GetStars()[0]->GetPath());
+	}
+	//GotoSystem(tmpTarget);
```
laarmen: Commented code ? Either use it or remove it ;-)

Or, at least, I'd very much like a comment as to why it needs to be there.

shammyx: Oooh, you're right, I'm sorry, just forgot to remove it :D

laarmen:
@ecraven have you started working on the sector map view in imgui ? If not, except for the small nitpick I'm 👍

PR #4040, opened 2017-06-30
Planet surface description - WIP

shammyx:
Still WIP. Now rocky planet/oceanic world/ice world are determined based
on the most dominant. Do we want more? Or just clean up my changes with
Lang:: strings and maybe replacing variables/constants?


shammyx:
Wanted to push this, because the current determination of the surface is, begging your pardon, nonsense. Corrected also the ice cover of earth. What do you think about it? The temperature could be used to call the planet 'ice world' although it is a rocky planet. We could use ocean either just for water or any fluid. Then additional temperature checks should be added... Also the atmosphere of earth and teh determination should be corrected/improved with elements. Earth has a nitrogen atmosphere with oxygen, not an oxygen atmosphere.

shammyx:
Oh and my previous commit doesn't belong here. Thought it would work if I'd open a new branch from that last branch.

laarmen:
Ah, OK, now I understand better the question. Well, no, you'll need to
rebase the branch onto master in order to remove the offending commit.

shammyx:
I'm wondering if this is enough? Is there something missing? Are the short Lang strings ("and", "much", "little", ...) ok or should we use longer ones with the whole term in order to be more compatible to translations?

ecraven:
I am sure this will create problems with localization, adding words together will surely break in other languages. I'm not sure how to best proceed, probably it would be best to create longer translation strings (though we'll have many more of these :-/)

shammyx:
Could anybody review this? I think this small PR should be fine for now, but I discovered so much more that could be improved. Also finer details in here, or maybe even new ideas. Anyhow, for now, that should be alright.

impaktor:
I'm a bit split. It looks like it doesn't add that much, more than a slightly more fine grained description of planets? (no offence).

I suspect the translation strings will still be problematic in other languages, at the very least to understand the context of "and ice" and such. the `description` field could be used in some way to let translator know in what context it will be used.

> Wanted to push this, because the current determination of the surface is, begging your pardon, nonsense. 

Care to elaborate? Does this PR, in addition to adding more ways to combine descriptive text of planets fix something that can be considered a bug? I'm mainly trying to understand what it does, after skimming the code.

fluffyfreak:
I think it tries to changes Earth and Mars values to provide better descriptions as well, and I like that it extends the descriptions to offer more detailed explanations but it does have the problems that @impaktor  lists.

It will also conflict mightily with #4214 by @joonicks, perhaps @shammyx and @joonicks might be good to work together on this topic as you are both interested in it?

It's good to use IRC or the forums to bounce your ideas around and I think you're both interested in tackling the same types of problem 😃

impaktor:
In general, I don't like to push a lot of strings to our translation team if they're bound to have to be re-translated or don't add much value. But that's me.

Do we know some language where these strings would break / be difficult to translate? Specifically, building sentences and stringing them together with "<str1> and <str2>"?

joonicks:
considering the gui will be replaced, maybe it would be good to split the single line description into:

<Name>: Tiny/Small/Earth sized/Large/Very large body ... with a molten core/active tectonics ...
Composition: Highly metallic/Silicon based rock/Carbon based compunds ... with surface deposits of water ice/liquid water/liquid hydrocarbons/liquid ammonia/etc
Atmosphere: Tenuous/thin/dense/very dense ... hydrogen/nitrogen/co2 ...

joonicks:
I found this quote

"Only 12 percent of the Earth's surface is permanently covered in ice and snow, the majority of which is found in the polar regions."

nozmajner:
Just a note from the fiction side "[Earth is ravaged, it's environment is in ruins](https://pioneerwiki.com/wiki/Timeline_of_the_Pioneer_universe)", so that 12 % might be either under or overshooting. Maybe the ice world part of the current description is accurate after all, as an effect of a long nuclear winter a thousand years ago.
Not that the visual part mirrors that, but I thought it's worth mentioning.

joonicks:
excellent point.
in view of this fictional future, earth in pioneer is currently at +15C average temperature (288K) which is the average today. so if global warming increased the temperature or nuclear winter lowered it, that should change.

personally I believe more in global warming than nuclear winter (which is more temporary (maybe a decade vs global warming century)).

impaktor:
> personally I believe more in global warming than nuclear winter

(You also have natural occuring ice age, which I think might set in within 1000 years.)

nozmajner:
I read somewhere that we are actually heading out from an ice age still (2017). Nicely acceleration out with our penchant to disregard our environment I guess.

fluffyfreak:
This has conflicts with a later PR that attempted to do several similar things.

It might be best to open a new PR once issues have been resolved.

Sorry @shammyx it is _not_ a rebuffing of your work just trying to cleanup the vast number of inactive PR's on here at the moment.

PR #4042, opened 2017-07-01
Fix or silence 64-bit warnings in VS2015

fluffyfreak:
Mostly a case of doing the following.
- return size_t where appropriate
- cast to Uint32/Sint32 where necessary

miniz still has some 64-bit portability warnings but I'll do them at another time.


fluffyfreak:
Going to do this another way

PR #4044, opened 2017-07-02
Removing collision mesh for those little fin things top and bottom.

### LINKED ISSUE #4041: Buying Ship Breaks Game
### Observed behaviour
After Selling my first ship on earth and buying the lunar shuttle i could not controll the spaceship at all, the only things that could be done were buying fuel and using menus. 

When compiling i did not have any compile warnings or errors.

### Expected behaviour
The ship should have been flyable

### Steps to reproduce
Compile a new version of the current commit and click the earth start. Then use the money cheat to give yourself extra money and buy a ship. Then undock and try to take off. This happened as well when i tested it by giving myself extra money through the console and buying any ship.

[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/1116344/opengl.txt)
[output.txt](https://github.com/pioneerspacesim/pioneer/files/1116343/output.txt)

One thing i might have done wrong was doing a multi-thread build. i am trying again with single thread.

Lubuntu Yakkety new install
Pioneer commit 8e1aa996bbf0e27ab074ee4df611ccc753ed0415
I compiled it normally with the required programs

nozmajner:
I've removed the collision geometry for those little fin thingies on the top and bottom of the ship, since it seemed to cause issues as seen in #4041.
So hopefully fixes #4041


fluffyfreak:
I need to test this but I need to find the time to do so first!

fluffyfreak:
Which I just did, forgot to delete the `sgm` file initially and it didn't work. However once removed it was much better.

There was still an odd collision though once I was quite high above the pad! Not sure what's going on there, I'll need investigate a little further.

This does however improve the situation, even though a problem is still there. So if further investigation is taking a long time i'll merge this as a temporary fix.

nozmajner:
Odd collision above the pad?

fluffyfreak:
@nozmajner do you use SGM files?

impaktor:
Is modifying the ship collision mesh the right fix for this bug to begin with?

fluffyfreak:
@impaktor If it's the cause of the bug, then yes :)

nozmajner:
Nope, no sgm at all. And the debug tags don't show anything off like some leftover collision geometry far from the ship.

fluffyfreak:
Merging as no negative effects and partially fixes. The actual cause needs more investigation though.

PR #4045, opened 2017-07-03
WIP - replace local_stars.lua and bright_stars.lua with stars from the HYG catalog

ecraven:
Use the stars in https://github.com/astronexus/HYG-Database.

Conversion done by `scripts/convert-hygdata.scm` (look at the comments in the file).
The generated file is already included, **you don't need to generate it yourself**.

This is *not* meant for inclusion, but to look at the map (F2) on how actual stars would look.

Note: the in-space star background is *not* real stars (yet).


impaktor:
_"My god! It's full of stars!"_

Does it / will it give the same-ish positions as our previous "real" stars?

ecraven:
> Does it / will it give the same-ish positions as our previous "real" stars?

Hard to say exactly, I just checked a few, Sirius has moved a bit and is
now right outside the initial jump sphere, but I've tried to align
things so that they are kind of in the same direction ;)

starling13:
Will this commit be merged? This is very interesting step toward realism which not requires code modification and complexity increment.

Hoimar:
I agree, would love to see this in game!

impaktor:
@Web-eWorks I suspect we can close this, since #4879 is a more relevant re-implementation of this PR?

PR #4049, opened 2017-07-10
Fail hyperdrive when wheels are down

WKFO:
On gameplay side: this gives me another reason to not forget my wheels down, aside from the atmospheric effects.
Allows countdown to finish beforehand so player has time to retract it if he/she didn't already.
Fails to enter hyperspace if the gear is down at the moment the countdown finishes; plays the abort sound, continues flight status of FLYING.

On science side: a very slight shift in geometry and center of mass causes unpredictabilities for hyperjump calculations etc.

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -1206,7 +1206,7 @@ void Ship::StaticUpdate(const float timeStep)
 			AbortHyperjump();
 		} else {
 			m_hyperspace.countdown = m_hyperspace.countdown - timeStep;
-			if (!abort && m_hyperspace.countdown <= 0.0f) {
+			if (!abort && m_hyperspace.countdown <= 0.0f && m_wheelState != 1.0f) {
 				m_hyperspace.countdown = 0;
 				m_hyperspace.now = true;
 				SetFlightState(JUMPING);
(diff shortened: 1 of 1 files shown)


laarmen:
I don't really like that such a feature is implemented on the C++ side, but OTOH I'm not even sure we can do it on the Lua side yet.

nozmajner:
I remember Frontier having this feature, and we were also talking about it. 
How NPC ships working with it?

laarmen:
Oh I like the feature, just not where it's implemented.

impaktor:
> How NPC ships working with it?

Autopilot raises wheels by auto, which is what AI uses, so should be fine.

fluffyfreak:
@laarmen I don't mind where it's implemented, especially when there's code that's already in C++. If we want something to be in Lua, but the existing code is in C++, then it's not the responsibility of the person adding a small feature to take on that whole task.

@WKFO I like it, I'll give it another day for comments then I'll merge it if there's no argument, thanks for the contribution :)

laarmen:
Actually, the Lua implementation would be a 1line patch, although relying
on an undocumented feature. I had more details on that on IRC a few days
ago, but I'm currently away from... Well, everything, in the Alpes.

Le dim. 16 juil. 2017 à 17:35, Andrew Copland <notifications@github.com> a
écrit :

WKFO:
If I remember correctly laarmen told me a 1 line solution as a Lua implementation but I don't like it since it deactivates the hyperdrive without even activating it. And it has a slight chance of displaying the countdown message for a fraction of a second on computers with low computation speeds. I am not complaining, I am sure all these can be solved in Lua but I am not familiar with it.

PR #4050, opened 2017-07-13
New default keybindings

nozmajner:
I thought that since there's a pretty big UI update already in master, this is the good time to update the default keybindings in a HOKAM (Hands On Keyboard And Mouse, sorry :D ) manner. 
[There were some discussion about it a while ago](https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=373&hilit=keyboard), but then it dwindled. I thought I'd revive it in a less talk, more "code" way.
I blatantly put in my keymap mostly, because I've been playing with it for a while now (and used something similar in almost every space game) and it proved really usable to me. And I don't even use the autopilot these days, so I fly a lot by hand.
This way you can steer with the mouse and move on all axes with the left side of the keyboard, but still can fly keyboard only. Same goes for joysticks. Practically all 6 axes are under your hands all the time this way.
And it's more in sync to what games usually provide these days (that alone won't be a good reason though).

Admittedly it's biased for right handed people, but us southpaws are only 10%, and I suspect that most lefties are actually using the mouse with the right hand anyway (I always did). That's my observation at least, and I'm teaching at an art school, where are more lefties usually.

The new setup:

- **WASD:** Forward/backward, left/right movement (laterals), typical WASD config
- **RF:** Up and down movement.
- **TG, Return-Right Shift:** Set speed (kept the defaults too)
- **IJKL:** Pitch and Yaw
- **QE, UO:** Roll. Sometimes in games QE is used for up-down movement, but I think RF is more intuitive and comfortable. And all 6 axes are available this way. UO is to bunch the rot keys together too.
- **V:** Toggle rotation damping. 
- **X, P**: Kill rot ( I usually put this to C, but then you are toggling it when you are centering the sector map, which is annoying, and C for that makes a lot of sense)
- **Y:** Select target under reticule. (Had to move from T). **Note: I noticed that this feature doesn't work on master.**
- Removed the TB for top and bottom camera, now they are only accessible from the numpad. (And T was conflicting with targeting too). 

I've kept everything else the same. 

I've also updated the Sector map controls the same way. The only difference is that WASD moves along the plane and RF moves up and down, it is more intuitive that way in my opinion.
Also I've put camera rotation to the arrow keys in this case, but kept them as camera controls for the Worldview.


fluffyfreak:
I will try to test this out asap, nice work

PR #4053, opened 2017-07-18
Fix comparison for "most important" object.

ecraven:


nozmajner:
Works. Meant to fix a crash related to deciding either a planet or a box of cargo is more important in term of UI.

PR #4057, opened 2017-07-24
Fix segfault when exiting Pioneer

pcercuei:

### DIFF
--- a/src/graphics/gl2/GL2Renderer.cpp
+++ b/src/graphics/gl2/GL2Renderer.cpp
@@ -128,7 +128,7 @@ static Renderer *CreateRenderer(const Settings &vs) {
 
 	SDL_GL_SetSwapInterval((vs.vsync!=0) ? 1 : 0);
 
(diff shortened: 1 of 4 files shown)


laarmen:
Thanks for the patch. Glancing at the diff, I however don't see the
relationship between the patch itself and its title? Granted, it's a
cursory look.

Le lun. 24 juil. 2017 à 12:24, Paul Cercueil <notifications@github.com> a
écrit :

pcercuei:
The segfault happens when `SDL_GL_DeleteContext(m_glContext);` is called, because m_glContext is never initialized.

ecraven:
looks good to me, however, wouldn't it be simpler to just remove `m_glContext`, which seems to be never used anyway?

pcercuei:
Well, we have to keep it around so that the GL context of SDL is properly destroyed.

ecraven:
fixes #4057 for me.

PR #4062, opened 2017-08-01
Sort order fix

radcapricorn:
A continuation of #4053: it didn't account ordering guarantees expected by Lua sort, resulting in Lua "invalid order function" when comparing two gas giants.

### DIFF
--- a/src/LuaBody.cpp
+++ b/src/LuaBody.cpp
@@ -212,7 +212,10 @@ static int l_body_is_more_important_than(lua_State *l)
 	// bodies are sorted by importance in menus
 
 	if(body == other)
-		return false;
+	{
+		LuaPush<bool>(l, false);
+		return 1;
+	}
 
 	Object::Type a = body->GetType();
 	const SystemBody *sb_a = body->GetSystemBody();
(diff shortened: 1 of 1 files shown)


src/LuaBody.cpp
```diff
@@ -212,7 +212,10 @@ static int l_body_is_more_important_than(lua_State *l)
 	// bodies are sorted by importance in menus
 
 	if(body == other)
-		return false;
+	{
+		LuaPush<bool>(l, false);
+		return 1;
+	}
```
ecraven: thanks for catching this ;) must have been late when I wrote this.

src/LuaBody.cpp
```diff
@@ -236,8 +239,10 @@ static int l_body_is_more_important_than(lua_State *l)
 	else if(a == Object::Type::STAR) result = true;
 	// any (non-star) object is smaller than a star
 	else if(b == Object::Type::STAR) result = false;
-	// a gas giant is larger than anything but a star
-	else if(a_gas_giant) result = true;
+	// a gas giant is larger than anything but a star,
+	// but remember to keep total order in mind: if both are
+	// gas giants, order alphabetically
+	else if(a_gas_giant) result = !b_gas_giant || body->GetLabel() < other->GetLabel();
```
ecraven: ah, I totally ignored that gas giants are planets too... thanks

radcapricorn: Yup, the planet sub/super-typing adds to the complexity of this comparison chain. Just coming in to the code base, it took me a good while to figure out why it was failing.

src/LuaBody.cpp
radcapricorn: This looks mostly to be a sanity check anyway, I'm almost of the opinion it should rather be an assert: it **seems** unlikely for this function to be called for the same pointers. What do you think?

laarmen:
A comparison *must* be defined in the identity case. You're likely to break
quite a few sorting algorithms if you assert on equality.

Le mar. 1 août 2017 à 17:44, Stanislav Blinov <notifications@github.com> a
écrit :

radcapricorn:
Ah, good to know, thanks.

PR #4065, opened 2017-08-02
Setspeed indicator, larger hud size, more frame info

ecraven:
set-speed

### DIFF
--- a/data/pigui/modules/radar.lua
+++ b/data/pigui/modules/radar.lua
@@ -84,8 +84,8 @@ local function displayRadar()
 	-- only display if there actually *is* a radar installed
 	if #radar > 0 then
 
-		local cntr = Vector(ui.screenWidth / 2, ui.screenHeight - ui.reticuleCircleRadius - 15)
-		local size = ui.reticuleCircleRadius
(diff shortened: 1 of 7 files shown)


nozmajner:
Nice, it does feel better at this scale.
As @StanPancakes mentioned on IRC, the azimuthal radar shares the scale setting, which makes it quite big.
radartoobig

data/pigui/modules/radar.lua
```diff
@@ -85,7 +85,7 @@ local function displayRadar()
 	if #radar > 0 then
 
 		local cntr = Vector(ui.screenWidth / 2, ui.screenHeight - ui.reticuleCircleRadius - 15)
-		local size = ui.reticuleCircleRadius
+		local size = ui.reticuleCircleRadius * 0.66
```
radcapricorn: perhaps keep the 15 pixel offset fixed, i.e. `ui.screenHeight - size - 15`?

ecraven: ah, thanks for the pointer, of course ;) I'll push that in a minute.

ecraven:
set-speed

PR #4066, opened 2017-08-02
Hide mouse pointer when GUI is not being drawn

radcapricorn:
Mouse pointer with no UI has little practical use, and just gets in the way of screenshots/capture.

### DIFF
--- a/src/PiGui.cpp
+++ b/src/PiGui.cpp
@@ -1,3 +1,4 @@
+
 // Copyright © 2008-2017 Pioneer Developers. See AUTHORS.txt for details
 // Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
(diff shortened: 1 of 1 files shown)


src/PiGui.cpp
```diff
@@ -336,6 +336,13 @@ void PiGui::EndFrame() {
 }
 
 void PiGui::NewFrame(SDL_Window *window) {
+	// Ask ImGui to hide OS cursor if GUI is not being drawn:
+	// it will do this if MouseDrawCursor is true. After the frame
+	// is created, we set the actual cursor draw state.
+	if (!Pi::DrawGUI)
+	{
+		ImGui::GetIO().MouseDrawCursor = true;
```
radcapricorn: No. It's reversed here, it means that ImGui is responsible for the pointer and not the OS. See `ImGui_*_NewFrame` implementations.

PR #4067, opened 2017-08-05
Add imgui gauges and planetary information 

ecraven:
Updated Screenshot:
gauges

Video: https://nofile.io/f/RmxMJxqKkWL/gauges.mp4

### DIFF
--- a/data/lang/core/en.json
+++ b/data/lang/core/en.json
@@ -699,6 +699,22 @@
       "description" : "Short for latitude",
       "message" : "Lat:"
    },
+   "LATITUDE_NORTH_ABBREV" : {
+      "description" : "Abbreviation letter for latitude 'north'",
+      "message" : "N"
+   },
+   "LATITUDE_SOUTH_ABBREV" : {
+      "description" : "Abbreviation letter for latitude 'south'",
+      "message" : "S"
+   },
+   "LONGITUDE_EAST_ABBREV" : {
+      "description" : "Abbreviation letter for longitude 'east'",
+      "message" : "E"
+   },
+   "LONGITUDE_WEST_ABBREV" : {
+      "description" : "Abbreviation letter for latitude 'west'",
+      "message" : "W"
+   },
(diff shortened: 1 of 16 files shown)


radcapricorn:
Would the inner text on those progress bars be readable in lower resolutions?

ecraven:
Not ideal, but still readable to me ;) 1024x768
small

impaktor:
Just gave it a quick spin. Looks good. Here's what I noticed:

1. the temperature window is still visible when in options menu

2. I'm not sure, but I suspect the position of the bars move if one of the bars isn't needed and therefore not drawn? I'd expect the xy of each bar relative the screen be fixed regardless if, say preassure, is drawn or now. I might be wrong, but I thought something moved when I went into space.

ecraven:
I can't reproduce 1., seems to be fixed ;)
As for 2., I'd prefer a neatly stacked list of gauges to one with holes in it (especially if modules start adding their own gauges). Is that incorrect reasoning?

fluffyfreak:
It looks good! Also I'd prefer continuous gauges instead of gaps too.

impaktor:
Just looking in pigui, your tabs are 2 big?

@laarmen what tab width do we use for lua?

Personally, I use 2, but for pioneer, I think it's 4 on C++ side?

Sorry, don't mean to go into the most boring part of coding...

Tested this. I can no longer tell if the temperature/planet info window is on top of the options window, as they no longer overlapp (and it's not movable by default), so I assume it's good.

ecraven:
> Just looking in pigui, your tabs are 2 big?
> @laarmen what tab width do we use for lua?
> Personally, I use 2, but for pioneer, I think it's 4 on C++ side?

I'd be glad if we could decide on *one* indentation width for lua and
C++, I'd like to put in a cleanup PR that fixes indentation of all files
some time in the future ;) clang-format is a great tool for getting
C++ code in line ;)

fluffyfreak:
1 tab = 4 spaces in C++

PR #4073, opened 2017-08-10
Fix occasional nil dereference

radcapricorn:
Presumably, traders may have a nil starport, causing seldom graceful crashes.

### DIFF
--- a/data/modules/TradeShips.lua
+++ b/data/modules/TradeShips.lua
@@ -837,7 +837,7 @@ local onShipDestroyed = function (ship, attacker)
 	if trade_ships[ship] ~= nil then
 		local trader = trade_ships[ship]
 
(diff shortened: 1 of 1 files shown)


impaktor:
When is a trader.starport.label == 'N/A'? Also, do we/you understand why they might not have a valid starport?

ecraven:
It is not == 'N/A', but if the trader.starport == nil, it will be 'N/A' (instead of trying to access the label of nil).
I don't think we understand why it can be nil, if someone has the time to investigate, it would of course be better to fix the root cause instead of the symptoms ;)

impaktor:
> It is not == 'N/A',

Right, by bad. Also, I see now that this PR is just for correcting a debug print statement. (well, not "just" all contributions are welcome)

ecraven:
yea, however it also crashes pioneer, even if it is only a debug statement, so this is almost as good as fixing the root cause ;) I'll have a look for the cause when I find a bit of time, but not very soon :-/

PR #4074, opened 2017-08-11
Properly warn the player about the landing gear (when trying to hyperspace)

WKFO:
A small incident I encountered in IRC proved that the lack of a warning causes confusion and discomfort. Actually I had this in mind and I was going to implement it (and this was accelerated today :)); however I was trying to "explore the code" to pop-up a code like the hyperspace countdown. In fact I did acheive this (I believe ecraven and nozmajner saw the screenshot) but then I immediately thought it was a very bad idea to hard-code this thing into C++ as this is not a good move since we are trying to get rid of unpleasant things that shouldn't be done with C++ (and this would require me to change sth like 5-6 .cpp and .h files (don't even ask, it was horrible)). So instead I am now messaging the player through comms.

Oh, also, we have a warning sound!


src/LangStrings.inc.h
```diff
@@ -455,3 +455,4 @@ DECLARE_STRING(LONGITUDE)
 DECLARE_STRING(NOW)
 DECLARE_STRING(TOGGLE_RADAR_VIEW)
 DECLARE_STRING(TOGGLE_EQUIPMENT_VIEW)
+
```
impaktor: poop

impaktor:
in need of a squash / rebase. before/at merge.

impaktor:
So player shouldn't be able to start hyperjump countdown, and then raise wheels before it reaches zero?

General Q: why aren't we playing the hyperdrive charge audio during countdown?

Follow Q: then it would be natural to play the hyperdrive abort sound, rather than missle_inbound sound when aborting the jump.

data/libs/Equipment.lua
```diff
@@ -294,6 +296,9 @@ HyperdriveType.HyperjumpTo = function (self, ship, destination)
 	if not fuel_use then
 		return "INSUFFICIENT_FUEL"
 	end
+	if wheels == 1 then
```
ecraven: technically, this should be != 0, otherwise you can initiate a jump while the gears are moving (while moving the gears float goes through the values from 0 to 1).

src/Ship.cpp
```diff
@@ -1215,11 +1215,9 @@ void Ship::StaticUpdate(const float timeStep)
 				// after the whole physics update, which means the flight state on next
 				// step would be HYPERSPACE, thus breaking quite a few things.
 				LuaEvent::Queue("onLeaveSystem", this);
-			}
-			//give player some time to check the wheels
-			else if (m_hyperspace.countdown <= 0.0f && m_wheelState == 1.0f) {
+			} else if (m_wheelState == 1.0f) {
```
ecraven: sam as above, should probably be != 0.0f

data/lang/equipment-core/en.json
```diff
@@ -326,5 +326,9 @@
    "UNOCCUPIED_CABIN_DESCRIPTION" : {
       "description" : "",
       "message" : "Required for transport of a single passenger"
-   }
+   },
+   "ERROR_LANDING_GEAR_DOWN": {
+      "description": "",
+      "message": "ERROR: Landing gear is down!"
```
ecraven: Maybe we could change this to something along the lines of: "Cannot initiate hyperjump while landing gear is down".

data/libs/Equipment.lua
ecraven: sorry about that, obviously lua has ~= instead of != :-/

laarmen:
FYI this whole affaire is on my TODO list for this week (I have a week off
work and a busted knee, which mean I'll likely spend a lot of time on my
computer...)

impaktor:
> General Q: why aren't we playing the hyperdrive charge audio during countdown?

I've been told we do, and apparently, I'm oblivious to many of the sounds of pioneer. I only hear a handful, a very small subset of the ogg files we have in data/sounds.

WKFO:
Don't mind my previous comment. Just made a fresh built with the current commits and it stopped doing that random warning pop-ups.
My bad. Should've tried this before. We're good to go.

ecraven:
Could you rebase onto current master and squash all commits into a
single commit? Then we'd be good to merge, I think.

data/lang/equipment-core/en.json
```diff
@@ -329,6 +329,6 @@
    },
    "ERROR_LANDING_GEAR_DOWN": {
       "description": "",
-      "message": "ERROR: Landing gear is down!"
+      "message": "Cannot initiate hyperjump; landing gear must be retacted!"
```
zonkmachine: > retacted

retracted

WKFO: Damn, thank you!

data/libs/Equipment.lua
```diff
@@ -277,6 +277,8 @@ HyperdriveType.HyperjumpTo = function (self, ship, destination)
 	-- First off, check that this is the primary engine.
 	local engines = ship:GetEquip('engine')
 	local primary_index = 0
+	local l = Lang.GetResource(self.l10n_resource)
```
ecraven: can you move this and the next block into data/libs/Ship.lua?

data/libs/Equipment.lua
```diff
@@ -296,6 +298,9 @@ HyperdriveType.HyperjumpTo = function (self, ship, destination)
 	end
 	ship:setprop('nextJumpFuelUse', fuel_use)
 	local warmup_time = 5 + self.capabilities.hyperclass*1.5
+	if ship = Game.player and wheels ~= 0 then
```
ecraven: see above.

ecraven:
almost ready to merge ;) thanks for adding this!

data/lang/equipment-core/en.json
```diff
@@ -326,9 +326,5 @@
    "UNOCCUPIED_CABIN_DESCRIPTION" : {
       "description" : "",
       "message" : "Required for transport of a single passenger"
-   },
-   "ERROR_LANDING_GEAR_DOWN": {
-      "description": "",
-      "message": "Cannot initiate hyperjump; landing gear must be retracted!"
```
nozmajner: Nitpick: isn't it "can not" ?

WKFO: It is now.

WKFO:
I made quite a mess while working on this and I though you could use some explanation about the recent ones:

- Moved the messaging code to Ship.lua from Equipment.lua

- Moved the warning message to ui-core instead of equipment-core. Both is fitting in my opinion but you don't need to import the ui-core once more from Ship.lua so it has this slight... advantage (?)

- Solved a conflict that involves Ship.cpp, although I have no idea why that happened in the first place.

- Fixed the word "cannot" which nozmajner pointed out

- Fixed a small spacing problem to align the lines in the ui-core.

src/Ship.cpp
```diff
@@ -1215,11 +1215,10 @@ void Ship::StaticUpdate(const float timeStep)
 				// after the whole physics update, which means the flight state on next
 				// step would be HYPERSPACE, thus breaking quite a few things.
 				LuaEvent::Queue("onLeaveSystem", this);
-			}
-			//give player some time to check the wheels
-			else if (m_hyperspace.countdown <= 0.0f && is_equal_exact(m_wheelState,1.0f)) {
+			} 
+      		else if (m_wheelState != 0.0f) {
```
impaktor: Not to torture you,  but don't you get compiler warning for this, assuming you're on GNU/Linux using GCC? See #4090.

PR #4079, opened 2017-08-12
Show alerts near the reticule with imgui.

ecraven:
yellow
red


nozmajner:
I'll come up with proper icons for it.

nozmajner:
[Here are the alert icons.](https://github.com/pioneerspacesim/pioneer/files/1220947/alert.zip)
I've also reduced their size to 24*24 in the game.lua

alerts

PR #4083, opened 2017-08-13
Fix Star/planet/gas giants popping into view.

### LINKED ISSUE #4077: Stars Popping into view
### Observed behaviour
Stars Pop into view long after their halo has already appeared.
This might also be happening to Planets and Gas Giants as they use some of the same code and distances can be similar.

### Expected behaviour
Stars should not pop but should replace their billboards at a far distance, then smoothly grow larger as the player approached.

### Steps to reproduce
Head to Sirius from New Hope start point, fly towards Sirius-A and watch.

This appears to be caused by the near/far clipping planes, must investigate.

On current master.

fluffyfreak:
I have extended the far clip plane to 100,000,000 units, and reduced the near plane to 0.001 units. In truth the near plane could probably go out even further than that but I've tested this and it works so am leaving it alone.

This fixes #4077 as the star is no longer clipped against the far plane, increasing the near plane keeps the range the similar and so avoids precision issues (_I hope_).

### DIFF
--- a/src/graphics/opengl/RendererGL.cpp
+++ b/src/graphics/opengl/RendererGL.cpp
@@ -152,8 +152,8 @@ RendererOGL::RendererOGL(SDL_Window *window, const Graphics::Settings &vs, SDL_G
 //the range is very large due to a "logarithmic z-buffer" trick used
 //http://outerra.blogspot.com/2009/08/logarithmic-z-buffer.html
 //http://www.gamedev.net/blog/73/entry-2006307-tip-of-the-day-logarithmic-zbuffer-artifacts-fix/
-, m_minZNear(0.0001f)
-, m_maxZFar(10000000.0f)
+, m_minZNear(0.001f)
+, m_maxZFar(100000000.0f)
 , m_useCompressedTextures(false)
 , m_invLogZfarPlus1(0.f)
 , m_activeRenderTarget(0)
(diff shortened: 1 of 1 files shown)


ecraven:
hm.. seems it's still not enough :-/ https://nofile.io/f/DYABpyTbjQ2/sun.mp4

ecraven:
also, probably unrelated, but why is sirius as oblate as this: https://nofile.io/f/kJM64aLMdBC/sirius.mp4

fluffyfreak:
We should probably switch to an infinite far clip plane, but I don't know how it interacts with our logZ depth stuff

ecraven:
> We should probably switch to an infinite far clip plane, but I don't know how it interacts with our logZ depth stuff

I'm guessing there's no easy way to have different clip planes for
different kinds of objects? Everything smaller than a sun is impossible
to see from a certain distance anyway, I think mainly suns are the
problem here, imho.

fluffyfreak:
@ecraven for now there's no problem with just changing the range, but where is that Ackay system you showed the enormous star for? i can find several `Ackay` systems but don't know which is the right one.

fluffyfreak:
@ecraven I think that I will leave the `Ackay` problem to another patch as I think it is caused by related by different problems. The change in this patch should fix the original issue but a new patch will be required for `Ackay` scale problems.

On my machine it doesn't matter what I change the near/far plane values to as `Ackay` never becomes visible until I'm in a low orbit around it. Does yours work the same or differently?

macksting:
If you want an even bigger one, Waioeth A at -13, -9, -13 is reportedly about 1.7 AU wide. Pretty impressive. A little difficult to get to, though.

PR #4084, opened 2017-08-14
Add Eros and Phoebe.

ecraven:


impaktor:
...addendum: important to have because The Expanse. 
Plan: attract Expanse fans.

nozmajner:
Reel them in! I mean the fans. :D Maybe they start a mod too and contribute.

PR #4085, opened 2017-08-14
Adding TRAPPIST-1

nozmajner:
Adding [TRAPPIST-1](http://www.trappist.one/) to the game. Coordinates (-4, 1, -3) should be good (distance checks out).
No moons or starports.
You won't find it under that name though, I named it Regula Benedicti, after the Trappist monk order. (so it's not obvious, and people can discover it), 
All planets are named after trappist beer breweries.
trappist


impaktor:
Should the planetary info description mention the name TRAPPIST? Or no?

nozmajner:
Hmm, might be good to have that.

nozmajner:
How about "Regula Benedicti was among the first star systems where habitable exoplanets were discovered, also known as TRAPPIST-1 after the telescope which discovered it."

PR #4089, opened 2017-08-15
Attack outlaws immediately

robothauler:
Attack outlaws immediately and show no mercy.

### DIFF
--- a/data/lang/module-policepatrol/en.json
+++ b/data/lang/module-policepatrol/en.json
@@ -19,6 +19,18 @@
       "description" : "",
       "message" : "No illegal cargo detected. You may proceed."
    },
+   "OUTLAW_DETECTED_1" : {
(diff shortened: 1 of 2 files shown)


impaktor:
So this makes the police patrol attack the player's ship not only if carrying contraband, but also if an outlaw?

I've always imagined that once a player is an outlaw, there will be a contract on your head, thus bounty hunters will be going after you.

fluffyfreak:
How do you now land to pay off a fine if you've got on? Everywhere you go will just attack you before you can pay it.

robothauler:
I think you are talking about the Station Police. The Police Patrol is only sparsely spawned and shouldn't be a problem here.

fluffyfreak:
Ah ok then. Will merge in a few days so it doesn't mess up the August changelog :)

PR #4091, opened 2017-08-16
Orbital stations provide (some) orbit info in lobby

impaktor:
This is "nice to have", as it differentiates stations a bit more,
and you become more aware of being in an orbital station.

I wanted to show it here, to see if this is something we want, before I start fixing the time format.
screenshot-20170816-180235

### DIFF
--- a/data/ui/StationView/Lobby.lua
+++ b/data/ui/StationView/Lobby.lua
@@ -216,12 +216,22 @@ local lobby = function (tab)
 		tech_certified = string.interp(l.TECH_CERTIFIED, { tech_level = station.techLevel})
 	end
 
+	local orbit_period = station.path:GetSystemBody().orbitPeriod
+
+	local station_orbit_info = ""
+	if station.type == "STARPORT_ORBITAL" then
+		station_orbit_info =
+			string.interp(l.STATION_ORBIT, { orbit_period = string.format("%.2f", orbit_period),
+											 parent_body = station.path:GetSystemBody().parent.name})
+	end
+
 	return
 		ui:Grid({60,1,39},1)
 			:SetColumn(0, {
 				ui:VBox(10):PackEnd({
 					ui:Label(station.label):SetFont("HEADING_LARGE"),
 					ui:Align("LEFT", tech_certified),
+					ui:Align("LEFT", station_orbit_info),
 					ui:Expand(),
 					ui:VBox(10):PackEnd({
 						ui:HBox(10):PackEnd({
--- a/src/LuaSystemBody.cpp
+++ b/src/LuaSystemBody.cpp
@@ -276,6 +276,28 @@ static int l_sbody_attr_apoapsis(lua_State *l)
 	return 1;
 }
 
+
+/*
+ * Attribute: orbitPeriod
+ *
+ * The orbit of the body, around its parent, in days, as a float
+ *
+ * Availability:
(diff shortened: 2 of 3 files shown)


impaktor:
OK, exposed orbit period to LuaAPI, and get the right numbers. Not sure if I stuck it in the logically right file (LuaSystemBody). Is the orbit period of a "system body" a body property or space property? 

Should it return period in days or seconds? Rotational period is in days, so sticking to that.

Possibly time should be formated more sanely, as the decimal of days can get up to .99, which might look strange? E.g. "6.98 days"? 

I'm keeping this as WIP until reviewed by @laarmen or ecraven (no ping) after he's back from vacation.

screenshot-20170822-073951

src/LuaSystemBody.cpp
```diff
@@ -276,6 +276,28 @@ static int l_sbody_attr_apoapsis(lua_State *l)
...
+ *   201708
+ *
+ * Status:
+ *
+ *   experimental
+ */
+static int l_sbody_attr_orbital_period(lua_State *l)
+{
+	SystemBody *sbody = LuaObject<SystemBody>::CheckFromLua(1);
+	lua_pushnumber(l, sbody->GetOrbit().Period() / float(60*60*24));
```
laarmen: Eh, that's an elegant way to do the cast, I'd forgotten about this syntaxic possibility :-)

src/LuaSystemBody.cpp
```diff
@@ -276,6 +276,28 @@ static int l_sbody_attr_apoapsis(lua_State *l)
 	return 1;
 }
 
+
+/*
+ * Attribute: orbitPeriod
+ *
+ * The orbit of the body, around its parent, in days
```
laarmen: Just so that the interface is clear, it might be a good thing to say whether the return value is an integer or not. I know it's not (or I've missed a step) but only by looking at the code, which isn't very user friendly.

data/ui/StationView/Lobby.lua
```diff
@@ -216,12 +216,22 @@ local lobby = function (tab)
 		tech_certified = string.interp(l.TECH_CERTIFIED, { tech_level = station.techLevel})
 	end
 
+	local orbit_period = station.path:GetSystemBody().orbitPeriod
+
+	station_orbit_info = ""
```
laarmen: This variable should be local, shouldn't it?

src/LuaSystemBody.cpp
impaktor: Fixed. Said float now. (although what "days in float" mean might be confuding, due to babylonian base 60 system)

src/LuaSystemBody.cpp
impaktor: All in hour of glorious copy-paste!

data/ui/StationView/Lobby.lua
impaktor: Right you are. Thanks! (fixed)

src/LuaSystemBody.cpp
laarmen: Eh, it's an API description. If the reader doesn't understand what "float" means they've got other problems...

laarmen:
LGTM.

impaktor:
OK, then I'll remove the WIP, thanks!

PR #4092, opened 2017-08-17
Move game start location for Earth back to London

impaktor:
__Primary reason:__ make use of the `space_station_type()` from #3990

__Secondary reason:__ In 2014, when Nyankosensei made the new ground station, we
moved starting location from the more original Elite/Frontier/European start
of London, to L.A (#3485); because that's where the random number generator
happened to give us the new octagonal station. Well, now we can specify
explicitly the station type, so no reason to be in L.A. anymore (other than
celebrity spotting?).

Also, development contributions from US have been minimal, so why be there?

__Reason against:__ Unnecessary save bump (I suspect?), 25% of our downloads are
from US, last time I looked.

### DIFF
--- a/data/systems/00_sol.lua
+++ b/data/systems/00_sol.lua
@@ -77,7 +77,8 @@ local earth_starports = {
 		:longitude(math.deg2rad(99)),
 	CustomSystemBody:new('London', 'STARPORT_SURFACE')
 		:latitude(math.deg2rad(51))
-		:longitude(0),
+		:longitude(0)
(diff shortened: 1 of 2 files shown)


radcapricorn:
I actually like the current start, LA and Mexico City being closer to ecliptic than London make it for easier orbital start. Not that 99% of players would consider that a reason though :)

fluffyfreak:
Maybe we should start picking an start location closest to where the user is? or based on language?
Anyway I'm ok seeing this merged so will do it now.

PR #4093, opened 2017-08-17
Add various new UI elements, move from C++ to lua

ecraven:
Move a few UI elements around, add the thrust/velocity indicator, add the low thrust button, move comms window to lua. Also remove the wide radar, only ever show the compact radar. If the wide radar is wanted, we can bring it back in a separate window. 
one
two

Video: https://nofile.io/f/AzPyBo9zfmB/one.mp4


nozmajner:
Don't forget to mention that the low thrust button reacts to the mouse wheel :)

impaktor:
This might be the best day in a long time! YAAAaaaaAAAY! 

For comms, should is show who is saying it? Or it is not like that on master now?

laarmen:
Just glanced at it over the web interface. Could there be some comments on
the public methods for documentation (even just a - - all TODO would make
me happy). Also, tab vs space in utils.lua.

I haven't actually read the code, but since it doesn't crash...

impaktor:
Testing it now. 
The thruster sound is borked. All engine sound sounds like a machine gun. Like the sound being interupted multiple times per second. Could this be an artefact of the new thrust indicator?

Also, there's no mouse over tooltop for thrust and velocity indicators.

impaktor:
Correction. Sound seems to work now. So maybe just something un-deterministic on my setup.

impaktor:
Buying Rockets, then taking off and toggling from scanner to equipment, I don't see any rockets, which I expected I wouldn't since I sort of knew that's not implemented. But it does make me less prone to want to advertise this comming release. Some players actually do combat, so they would find it a bug, I think.

nozmajner:
Missile icons: 
[missiles.zip](https://github.com/pioneerspacesim/pioneer/files/1236406/missiles.zip)
Unguided, Guided, Smart, Naval in that order. They are two icon wide indeed.
I think they could go to the left side, aligned horizontally with the reticule.
Also, the equipment icons could either go there too, or to the low left icon row, which is already representing ship functions. This also eliminates the need for a radar/equipment display toggle, which is a plus. 
For the ECM, the thunder icon could be used. Not sure if it's important to differentiate between the simple and advanced ECM, since you can have only one of them at any time.
09-missiles

PR #4096, opened 2017-08-19
Law-bugfix: No, the victim is not a criminal

impaktor:
For shooting or murdering ships, the victim would get the fine
of murder or piracy.

(Untested)

(bug reported on IRC by @nozmajner)

Semi-unrelated sidenote: there is no event trigger exposed to lua for missile fired.


laarmen:
Our very own "blame the victim"!

Issue #4095, opened 2017-08-19
Game crashes if you instruct autopilot to enter orbit to a body with 0g gravity


nozmajner:
### Observed behaviour
If you instruct the autopilot to enter low orbit around any body that has 0.000m/s^2 gravity.

According to @Stan_Pancakes, it's a floating point exception (it didn't said that on windows):
> 0  CheckCollision (r=6403479.2363868402, endvel=<optimized out>, tpos=<synthetic pointer>..., pathdist=143420200833.03012, pathdir=<synthetic pointer>..., dBody=0x244bdd60)
    at ShipAICmd.cpp:664
1  AICmdFlyTo::TimeStepUpdate (this=0x2a3f1190) at ShipAICmd.cpp:820
2  0x0000000000563aba in AICommand::ProcessChild (this=0x307fdc20) at ShipAICmd.cpp:80
3  AICmdFlyAround::TimeStepUpdate (this=0x307fdc20) at ShipAICmd.cpp:1181
4  0x000000000054ef20 in Ship::AITimeStep (this=0x244bdd60, timeStep=timeStep@entry=0.166666672) at Ship-AI.cpp:36
5  0x000000000056d64f in PlayerShipController::StaticUpdate (this=0x244ad570, timeStep=0.166666672) at ShipController.cpp:178
6  0x0000000000558121 in Ship::StaticUpdate (this=this@entry=0x244bdd60, timeStep=timeStep@entry=0.166666672) at Ship.cpp:1051
7  0x0000000000523633 in Player::StaticUpdate (this=0x244bdd60, timeStep=0.166666672) at Player.cpp:274
8  0x000000000058acd8 in Space::TimeStep (this=0x217d20d0, step=step@entry=0.166666672) at Space.cpp:962
9  0x000000000045df23 in Game::TimeStep (this=0x217b9540, step=0.166666672) at Game.cpp:248
10 0x0000000000505e7f in Pi::MainLoop () at Pi.cpp:1414
11 0x0000000000507879 in Pi::Start () at Pi.cpp:1327
12 0x0000000000412a8c in main (argc=<optimized out>, argv=<optimized out>) at main.cpp:130

### Steps to reproduce
Select any of the outer moons of Jupiter (above ), or the moons of Mars, and instruct the autopilot to enter low orbit. It might not crash right away, because first the autopilot takes off and leaves the planet, but it will, if you enable time acceleration. Fly to vicinity command works.

__My pioneer version (and OS):__
[...]

PR #4099, opened 2017-08-19
Fix for #4095

radcapricorn:
This needs review and testing, I'm not 100% sure this wouldn't break something else.

### DIFF
--- a/src/ShipAICmd.cpp
+++ b/src/ShipAICmd.cpp
@@ -565,11 +565,11 @@ static double MaxEffectRad(Body *body, Propulsion *prop)
 	}
 	const double gMulMassDivAccelUp = G * body->GetMass() / prop->GetAccelUp();
 #if 0
- 	if(gMulMassDivAccelUp>0.0)
- 		return std::max(body->GetPhysRadius(), sqrt(gMulMassDivAccelUp));
- 	return 0.0;
+	if(gMulMassDivAccelUp>0.0)
+		return std::max(body->GetPhysRadius(), sqrt(gMulMassDivAccelUp));
+	return 0.0;
 #else
-    return std::max(body->GetPhysRadius(), sqrt(gMulMassDivAccelUp));
+	return std::max(body->GetPhysRadius(), sqrt(gMulMassDivAccelUp));
 #endif
 }
 
--- a/src/ShipAICmd.cpp
+++ b/src/ShipAICmd.cpp
@@ -564,9 +564,13 @@ static double MaxEffectRad(Body *body, Propulsion *prop)
 		return static_cast<SpaceStation*>(body)->GetStationType()->ParkingDistance() + 1000.0;
 	}
 	const double gMulMassDivAccelUp = G * body->GetMass() / prop->GetAccelUp();
+#if 0
  	if(gMulMassDivAccelUp>0.0)
  		return std::max(body->GetPhysRadius(), sqrt(gMulMassDivAccelUp));
  	return 0.0;
+#else
+    return std::max(body->GetPhysRadius(), sqrt(gMulMassDivAccelUp));
+#endif
 }
 
 // returns acceleration due to gravity at that point


radcapricorn:
On a second look, this does contradict #4002.

jaj22:
#4002 was a bad fix. If GetAccelUp returns negative then there's an error elsewhere. Up thrusters are supposed to thrust upwards.

If you need to protect the autopilot code against thrust-direction errors then you should do it at a higher level. Preferably with an assert, as it's a nonsense result.

radcapricorn:
The issue isn't that `GetAccelUp()` returns negative, it's that `body->GetMass()` returns 0, making `alt` on line 1112 become 0, causing division by zero on line `1096`, which makes m_vel a NaN, which eventually leads to `sqrt(-NaN)` call on line 664, triggering a domain error.

jaj22:
GetAccelUp() returning negative was the original reason for #4002, I think. As you've identified, the original version of the code worked fine for mass 0.

On the other hand, I'm not sure why there are mass 0 objects. That seems like another higher-level bug.

fluffyfreak:
by fixing it this way we get back the other bug #3996. #4002 was always just a patch-it-and-hope fix, so does anyone know how to actually fix it?

jaj22:
Ok, this is weirder than expected. Turns out that Deimos has a massively negative mass. I traced this to `l_fixed_new` in `LuaFixed.cpp` using luaL_checkinteger for >32bit integer values. In context, this has a straightforward workaround:

Old code:
`int num = luaL_checkinteger(L, 1);`
`int denom = luaL_checkinteger(L, 2);`

Fixed code:
`Sint64 num = (Sint64)luaL_checknumber(L, 1);`
`Sint64 denom = (Sint64)luaL_checknumber(L, 2);`

I've stepped through and this works correctly for every call. However, a side effect is that the terrain generator now throws a height assertion on the Earth start. I'm guessing that this is because an asteroid terrain generator function doesn't work correctly for the now-correct mass values. I'm having trouble getting a valid breakpoint due to terrain threading, so I'm going to bounce this back for the moment.

impaktor:
That's great finding. I wonder if @fluffyfreak, with that fix applied, can detect the weirdness in the terrain generator (when you have time).

jaj22:
Ok, some followup. The Terrain assert is likely to be partially caused by this calculation in Terrain.cpp:

`m_maxHeightInMeters = std::max(100.0, (9000.0*rad*rad*(m_volcanic+0.5)) / (body->GetMass() * 6.64e-12));`
`if (!is_finite(m_maxHeightInMeters)) m_maxHeightInMeters = rad * 0.5;`

I believe this is a physical calculation intended to determine how spherical planets should be. For planets with small but non-zero mass, it blows up so that m_maxHeightInMetres is thousands of times greater than the planet radius. The terrain generators generally assume that m_maxHeightInMetres is somewhat related to the planet radius, and they're not exactly solid at the best of times. Capping m_maxHeightInMetres to the planet radius works around this problem. I didn't check whether it ended up with reasonable-looking asteroid terrains.

A related glitch I ran into is that asteroid masses are mis-specified or mis-adjusted. The SystemBody constructor for custom bodies (StarSystemGenerator.cpp, line 497) transfers the mass with the following code:

`kid->m_mass = csbody->mass;             // straight 32.32 fixed-to-fixed copy`
`if (kid->GetType() == SystemBody::TYPE_PLANET_ASTEROID) kid->m_mass /= 100000;`

Whoever wrote the masses for Sol apparently did not know about this line, as all the ASTEROID bodies are specified in earth masses. However, this adjustment is currently pointless, as the lua specification is just as (in)accurate as the SystemBody mass, given the Sint64 fix above.

Note that mass 0 is still a valid case due to the lack of accuracy in 32.32 fixed point values, but I think it's handled correctly everywhere except #4002.

jaj22:
More glitches. Ignoring the asteroid mass miscorrection, the masses of Phobos and Deimos appear to be incorrectly specified in the lua by a factor of ~100:

Real Phobos: 1.775e-9
00_Sol Phobos: 1.875e-11
Real Deimos: 2.47e-10
00_Sol Deimos: 2.5e-12

This seems to also be true of Jupiter's asteroids. Not sure why. Too consistent to be accidental. Who usually deals with this stuff?

nozmajner:
I did some of the Jupiter moons, and I had a hard time figuring out exponent. Not the Mars moons though.

jaj22:
OK, fixed a bunch of moon masses. Pasted for reference:

// Mars moons. Deimos has the smallest representable non-zero mass
	CustomSystemBody:new('Phobos', 'PLANET_ASTEROID')
		:radius(f(21,10000))
		:mass(f(1775,1000000000000)) -- 10.6e15 kg = 1.775e-9 EM
	CustomSystemBody:new('Deimos', 'PLANET_ASTEROID')
		:radius(f(12,10000))
		:mass(f(247,1000000000000)) -- 1.48e15 kg = 2.47e-10 EM - comes out as 1

// Jupiter inner moons, mostly tiny.
	CustomSystemBody:new('Metis', 'PLANET_ASTEROID')
		:radius(f(337,100000))
		:mass(f(633,100000000000)) -- 3.6e16 kg = 6.33e-9 EM
	CustomSystemBody:new('Adrastea', 'PLANET_ASTEROID')
		:radius(f(338,100000))
		:mass(f(352,1000000000000)) -- 0.2e16 kg = 3.52e-10 EM
	CustomSystemBody:new('Amalthea', 'PLANET_ASTEROID')
		:radius(f(13,1000))
		:mass(f(348,1000000000)) -- 208e16 kg = 3.48e-7 EM
	CustomSystemBody:new('Thebe', 'PLANET_ASTEROID')
		:radius(f(773,100000))
		:mass(f(72,1000000000)) -- 43e16 kg = 7.2e-8 EM

// Jupiter outer moons.
	CustomSystemBody:new('Themisto', 'PLANET_ASTEROID')
		:radius(f(627,1000000))
		:mass(f(115,1000000000000)) -- 6.9e14 kg, 1.15e-10 EM - comes out as zero
	CustomSystemBody:new('Leda', 'PLANET_ASTEROID')
		:radius(f(156,100000))
		:mass(f(837,1000000000000)) -- 5e15 kg, 8.37e-10 EM
	CustomSystemBody:new('Himalia', 'PLANET_ASTEROID')
		:radius(f(1334,100000))
		:mass(f(1122,1000000000)) -- 670e16 kg, 1.122e-6 EM
	CustomSystemBody:new('Lysithea', 'PLANET_ASTEROID')
		:radius(f(282,100000))
		:mass(f(1055,100000000000)) -- 6.3e16 kg, 1.055e-8 EM
	CustomSystemBody:new('Elara', 'PLANET_ASTEROID')
		:radius(f(675,100000))
		:mass(f(1457,10000000000)) -- 87e16 kg, 1.457e-7 EM

// wat
	CustomSystemBody:new('Aega', 'PLANET_ASTEROID')
		:radius(f(313,1000000))
		:mass(f(1055,100000000000))

// Couple of Neptune moons. These correctly used the old asteroid adjustment
	CustomSystemBody:new('Proteus', 'PLANET_ASTEROID')
		:radius(f(310,10000)) 
		:mass(f(843,100000000)) -- 5035e16 kg, 8.43e-6 EM
	CustomSystemBody:new('Nereid', 'PLANET_ASTEROID')
		:radius(f(2668,100000))
		:mass(f(452,100000000)) -- 2700e16 kg, 4.52e-6 EM

impaktor:
Merged fix discussed above. @radcapricorn no hard feelings I hope if we "stole" your PR, your debugging effors and insights were very valuable, and got the ball rolling really good. Thanks!

PR #4101, opened 2017-08-20
Add Earth g comparison in system info view

radcapricorn:
Adds information on how surface acceleration on a particular body relates to Earth g.


ecraven:
Ideally, the translation string should be updated, but this will all (soon-ish) move to lua anyway, so should be ok.

Issue #1123, opened 2012-03-29
Build system with CMake?


Sukender:
Hi everyone,

I was wondering if creating a CMake script would be interesting for Pioneer.

For those who don't know what I'm talking about, it's a meta-make (kind of cross-platform automake), used to create makefiles, XCode or MSVC projects, etc. Language is quite easy, and most important: once the main scripts are written and robustly tested, there are few things to know to maintain it (two or three neurons will be enough ^^).
Moreover, you could support multiple versions (eg. MSVC9 / MSVC10...) with no effort at all, since updating the centralized system will update all versions, all platforms, all makefiles, etc.

After a discussion with robn and Ae, Here is a summary:
- There was a CMake script (once upon a time) which has been dropped due to lack of maintenance.
- I'm to write a prototype script, to validate things, and especially the fact that the script must grab the GIT head SHA to includ in the build.
- Whatever the build system, it must handle weird cases if needed.
- Generated makefiles/project files have to be in the GIT repository. That way CMake is optional.
- Of course, there is no point supporting both automake+MSVC+CMake, as CMake is intended to work for all. There would be a period of transition where multiple build systems would cohabit, but in the end CMake would be the only maintained (if successful, of course!).

Thank you for your comments.

PR #4104, opened 2017-08-22
Add CMake build system

pcercuei:

This PR adds support for CMake as an alternative to autotools as the build system for Pioneer.

This is a first version, known to work on top of commit e313948441f85f16323f522c0412e745e0875a7a. Whether or not it works on current master is unknown. The reason for me to submit a PR now is that I can't work on it anymore, since I started a year-long computer-free trip around the world; but I hope it will light up some interest from upstream devs.

The benefits of using CMake over autotools are numerous, but the biggest improvement IMHO is the ability to generate project files for building on OSX and Windows (MinGW or MSVC), so that you don't have to maintain those separately.


impaktor:
We've been talking about adding cmake, (mostly due to the nicer C++ environment in Emacs that will follow 😄), so this is very interesting.

I know @Sukender tried to take us down this route 5 y ago, so discussion (top post(s)) in #1123 and previous un-merged PR #1134 are interesting.

Thanks, and have fun on your adventure!

laarmen:
Anyone looking into this?

pcercuei:
I found some time to work on it a bit more. Now it properly installs and supports external liblua.

I don't have a Windows PC with me, so it would be great if somebody could try building it on Windows. I think it would be easier using msys2 (because you can just use the package manager to install all the precompiled dependencies) but it should be doable with Visual Studio too.

pcercuei:
Builds fine under Windows (mingw):
https://ci.appveyor.com/project/pcercuei/pioneer/build/1.0.45

johnbartholomew:
@fluffyfreak Is this something you want?

Philbywhizz:
> For my part, working on Linux, switching from autotools to cmake would not cause a problem for me, and would probably change my workflow less than a Windows/VS based workflow. I agree CMake has some advantages over autotools, and if I were starting a C++ project from scratch I would most likely use CMake. I disagree that those advantages are particularly significant when looking at Linux alone, so I think this comes down to how devs on other platforms feel about it.
> 
The only concern I have from an OSX point of view is I’ve never used CMake before, so it will be new to me to learn (which isn’t a bad thing tbh).

laarmen:
VS 2017 supports CMake projects (as in integration within the IDE, no need
to generate any manually the sln files.

Le dim. 31 déc. 2017 à 01:53, Phil Howlett <notifications@github.com> a
écrit :

fluffyfreak:
I'm still not sure how this will work with visual studio but if CMake can handle it then I'm ok with it.

johnbartholomew:
> VS 2017 supports CMake projects (as in integration within the IDE, no need
> to generate any manually the sln files.

Neat! I guess that is new since the last time I touched CMake with VS.

> I'm still not sure how this will work with visual studio but if CMake can handle it then I'm ok with it.

I think the only way we'll find out if it works ok with VS is to try it and see (a) if VS can open it as laarmen suggests, and (b) if it builds ok. And that should probably be done before merge.

impaktor:
I'm pretty neutral, with a slight inclination towards the positive side for a change, as Cmake will unleash some improved emacs-integration with C++ mode.

With that said, I'm unlikely to touch any C++ code anytime soon, so I'm in the backseat on this one.

I'm not sure which dev is _passionate_ enough to maintain the switch. @ecraven mentioned CMake before I think?

johnbartholomew:
I have rebased this onto master on my copy of the branch, and fixed a minor problem but very few changes were needed.

https://github.com/johnbartholomew/pioneer/tree/cmake

The root CMakeLists.txt file pulls in all source files from (known) source directories so it doesn't need updating for new files. (It does need updating if new directories are added).

It would be good if someone who uses Visual Studio (2017) could try loading it and see if it works ok. This blog post (and laarmen's note above) suggests you should be able to just open the main directory (the root of the git repository) and VS should see the CMakeLists.txt file and do magic:
https://blogs.msdn.microsoft.com/vcblog/2016/11/16/cmake-support-in-visual-studio-the-visual-studio-2017-rc-update/

I don't know if you need CMake installed separately for that to work (maybe VS comes with CMake bundled now?)

Other things that may be missing:
- Not sure what needs to be done on OS X, there might be some additional stuff needed in the CMakeLists.txt file for that? (I see there is a special 'main' function wrapper in the osx/ directory, I don't know whether that's required)
- It doesn't build the modelcompiler.
- It doesn't build the "test" binaries (not sure I care though)

pcercuei:
The only (minor) problem I can foresee for building with VisualStudio is that the CMakeLists.txt should be able to find the pre-built dependencies. How are you guys dealing with them? Are they installed inside a specific directory?

I can add options to build the modelcompiler and the test binaries.

The-EG:
The prebuilt dependencies for VS builds are in pioneer-thirdparty/win32/<lib &bin>/<arch>/<vs2015 or vs2017>
VS2015 and VS2017 builds are setup with the assumption that pioneer-thirdparty exists alongside the pioneer directory.
I'll see if I can test this today, since I use VS2017.

The-EG:
Yep, it can't find the deps yet:
> 1> CMake Error at C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.9/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
1>   Could NOT find Freetype (missing: FREETYPE_LIBRARY FREETYPE_INCLUDE_DIRS)

fluffyfreak:
Yeah the VS201X projects require the paths to `pioneer-thirdparty` to be setup as well.
That goes for include, libs and binary folders. It will also need to copy across the relevant DLLs post-build.

The-EG:
pcercuei  and I worked through some of this over chat. One issue so far:
> <The-EG> does the library name matter? It's freetype271MT
> <pcercuei> yes, it does matter
> <pcercuei> it is looking for freetype.lib or libfreetype.lib
Renaming the importlib to freetype.lib got cmake through the above error, but it may be required to rebuilt to actually link and run something. 
Also, renaming it would require updates to VS201* solutions.

laarmen:
Assuming we can drop VS support for VS < 2017, which is fine by me, there won't be any VS solution to update. For the record, i'm using the CMake support when I work on https://github.com/Framstag/libosmscout for $work, and it works fairly well. Usually the only problem is with dependency handling, but that's exactly the same as our current sln.

Ping @fluffyfreak for the dropping VS2015 part...

fluffyfreak:
Just don't go deleting any of the existing stuff (_projects/DLLs/etc_) until it works please 😄

pcercuei:
I'm having the problem now that the libsigc++ headers are not inside the pioneer-thirdparty repo, only <sigc++config.h>, but not <sigc++/sigc++.h>, which is included in Pioneer from src/libs.h.

fluffyfreak:
they're in `pioneer-thirdparty\source\libsigc++-2.3.1\sigc++`

pcercuei:
Yes, but they're not in ```pioneer-thirdparty\win32\include\sigc++``` where they should be...

fluffyfreak:
`win32\include\???` is for prebuilt stuff, `source` is for ... well source code that needs to be built by the projects of whichever platform.

I had hoped one day to switch everything over to `source` and have the VS201X solutions build everything as static and link it that way. Thus doing away with the `win32` folder, but that's probably (_/definitely_) beyond the scope of just moving to CMake.

pcercuei:
Yes, but the headers should be installed in the sysroot ("prebuilt stuff" directory) too, CMake shouldn't have to look elsewhere than inside this directory...

pcercuei:
Actually I'm also missing the .lib files for libsigc++. Shouldn't these be included?

laarmen:
I believe the sigc++ is built from source?

laarmen:
Confirmed. There is a sigc++ in the VS2017 solution

The-EG:
It is; it's currently included directly in the Vs201* solutions as a project and gets linked statically.

fluffyfreak:
Yeah, it's built by the project rather than being prebuilt & included. Which is why it's not in the `win32` subfolder, but is just used directly from the `source` folder.

pcercuei:
OK, is there a reason why it's done differently for libsigc++ vs. the other libraries?

fluffyfreak:
Mixture of time, history, some performance issues.
Upgrading it was a pain but it was easier to integrate that way at the time etc.

pcercuei:
Rebased on top of master (de58718d7).

PR #4113, opened 2017-08-26
Fix autopilot getting being fed bad data.

### LINKED ISSUE #4100: Crew member pilot descends to 500m, then ascends to 15 km, repeats ad infinitum
### Observed behaviour
My pilot will descend sharply to reach 500m above the target spaceport, then slowly rise a few meters per second until she achieves 15 km, then descend again. When I turn on manual piloting this ceases, and gravity takes its proper effect.

### Expected behaviour
She probably should either feather the ventral thrust to keep the descent slow, or just let the ship fall for a few moments before braking.

### Steps to reproduce
Well, it happens with great reliability at Pallier's Folly in Arar (-13, -9, -11), a planet with roughly three times Earth's mass. I'll try to reproduce later under other circumstances, including with autopilot and at a planet closer to Sol.
Edit: Might help to have bespoke thrusters.

git (858629b) on lubuntu 16.04

fluffyfreak:
The problem seems to have been caused by incomplete modification of the places that use the thrust values to calculate times, forces and other values.

The fix I've gone with adjusts the linThrust and angThrust instead of trying to modify everywhere that uses them.

This fixes #4100 and I've tested it with bespoke thrusters on the affected planet/system.

### DIFF
--- a/src/Propulsion.cpp
+++ b/src/Propulsion.cpp
@@ -42,7 +42,6 @@ void Propulsion::LoadFromJson(const Json::Value &jsonObj, Space *space)
 
 Propulsion::Propulsion()
 {
-	m_power_mul = 1.0;
 	m_fuelTankMass = 1;
 	for ( int i=0; i< Thruster::THRUSTER_MAX; i++) m_linThrust[i]=0.0;
 	m_angThrust = 0.0;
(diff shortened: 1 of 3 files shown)


jaj22:
Looks good to me.

impaktor:
Doing a bit of an "emergency merge" as I assume there's a build comming anytime soon, and adressing point 4 mentioned above would have been made as a new commit (I suspect), so doesn't matter anyway, -- it can be made as a new PR.

Issue #4121, opened 2017-08-30
I think time in Pioneer must be UTC


manolollr:
Now, starting point in Earth is moved to London, but there is still Los Angeles timezone. At 13:30 in London is night!!
Well, it's not a severe issue but it's not aesthetical to see night in London at 13:30.
I think the best time to use is UTC, it's universal :-)

nozmajner:
Actually there's an event in Pioneer role. When the scientists worked on the first hyperdrive prototype there was a huge hyperspatial explosion that shunted Earth back in time about half a day. That's the reason for it. 
But nah, I'm joking, and will check if I can spin back and forth Earth a bit to match up UTC. After I figured out where the Sun should be on the sky at that time.

PR #4133, opened 2017-09-02
Move Earth to synchronize Pioneer time with UTC

manolollr:
With this change, at 12:00 UTC, in London, sun is at the zenith.

This is related to issue https://github.com/pioneerspacesim/pioneer/issues/4121


ecraven:
Not sure this is the right way to go, maybe we should just adjust the displayed time, not move any planets around?

impaktor:
It might also be more "spacey" to start game at night, under a star sky, then boring blue sky?

manolollr:
I've changed data/ui/MainMenu.lua to set start time in Earth at 18:30, but Sun moves 5 hours, it is not were it was at 13:30, so midday in London is at 8:30 :-(

PR #4138, opened 2017-09-03
PushStyleVar now pushing also vector variables

ColdSpirit0:
`pigui.withStyleVars()` did not worked with vector variables, attempt caused error. Now fixed, But no so clearly. Code needs to clean, because I dont know what types for what.


ColdSpirit0:
I think that PR can be much elegant, I will fix it when I have free time, if it still needed.

ColdSpirit0:
@ecraven I did the changes and amended commit.

impaktor:
@ecraven do you have time to check this out? I think @ColdSpirit0 needs these merged for further work.

fluffyfreak:
Made a call on this, can't see it doing anything bad and @ColdSpirit0 might require it so merging now.

PR #4139, opened 2017-09-05
New tracks

WKFO:
Combat is, obviously, combat music. A bit faster tempo compared to the other 2 we have but fits perfectly to the situation if you ask me.
Minuet is a docking/space station services music.


impaktor:
I'm OK with this, but I have no strong feelings either way for pioneer music.

PR #4146, opened 2017-09-07
Allow planets to have rings.

fluffyfreak:
Allow (~10%) terrestrial planets to have ring systems.

Planets where the rings would overlap with the orbit of a Moon will have them removed again.


fluffyfreak:
No comments in 9 days? Merging!

impaktor:
Sorry, but I don't have skills to review it, but we did talk about it on IRC with you when you opened it. At least I asked about "ring frequency" for rockyplanets (as in Sylvester Stallone?) vs. gas giants.

PR #4147, opened 2017-09-08
Spread the terrain skirts wider to avoid some sparklies

fluffyfreak:
Spread the terrain skirts wider to avoid some [sparklies](https://www.katsbits.com/smforum/index.php?topic=762.0) due to parallel, edge-on triangles.

This widens the base of the skirts drawn around terrain causing the skirts of adjacent terrain patches to overlap and interpenetrate. This should prevent the occurrance of further terrain sparklies.


impaktor:
I didn't know we had this problem. Well, good to fix it then. 😄

Issue #4098, opened 2017-08-19
AI Ship falling down


impaktor:
### Observed behaviour
Parked AI ships failing to hyperspace

### Expected behaviour
1. that AI ships are able to jump to hyperspace (they should know to raise their wheels in time)
2. That player doesn't hear sounds meant for the AI player

### Steps to reproduce
Sit on landingpad of e.g. New Hope. Wait for an AI Ship to take off. Hear the warning sound for entering hyperspace with wheels down (bug number one), then see ship fall down, as it fails to jump (bug number 2).

### Origin
I strongly suspect #4074 and #4049 

__My pioneer version (and OS):__

impaktor:
Possibly, @WKFO might want to have a look at this.

WKFO:
Maybe we should tell the AI ships to gain some altitude before hyperspacing, not just because of this bug; because it is illegal too.

WKFO:
About the warning sound we will have to use sth else than BodyMakeSound I think.

impaktor:
or just add an if in the code checking that it's the player and not AI that is taking off?

ecraven:
playsfx, maybe in luaengine. maybe just add an if for now?

PR #4150, opened 2017-09-10
Stop tradeships crashing into landing pads while trying to hyperspace

WKFO:
Makes things better, for now (their jumps may still be illegal, idk). Maybe we can add a delay so ships can travel even further away from the starport before jumping.

Gotta admit it is enjoyable watching the AI screw up :D

### DIFF
--- a/data/modules/TradeShips.lua
+++ b/data/modules/TradeShips.lua
@@ -307,6 +307,12 @@ local jumpToSystem = function (ship, target_path)
 end
 
 local getSystemAndJump = function (ship)
+	local body = Space.GetBody(trade_ships[ship].starport.path:GetSystemBody().parent.index)
+	local port = trade_ships[ship].starport
+	-- boost away from the starport before jumping if it is too close
+	if (ship:DistanceTo(port) < 20000) then
+		ship:AIEnterLowOrbit(body)
+	end
 	return jumpToSystem(ship, getSystem(ship))
 end
 


macksting:
I know it wouldn't actually help anything, but can somebody record a video of this? It sounds hilarious.

WKFO:
https://youtu.be/dxQpHUjW-lQ

impaktor:
Why do they crash into the pad to begin with? I thought #4098 was solved, as they should no longer be refused to hyperspace due to landing gear being down, right? That code checks if ship is player.

I suspect @laarmen would argue that hyperspace rules should be implemented in Lua (and then i Equipment.lua)?

I suspect @jaj22 would argue this is behaviour the AI should be able to "know" about, and impelemnted in c++?

WKFO:
As soon as ships lift off they attempt to hyperspace the way you (impaktor) do (begin countdown and hope you have enough time before hitting the ground) Most ships acheive this but some ships with low accel on high-ish G worlds unfortunately hit the ground as far as I've observed

jaj22:
The old issue is that there's no function exposed at Ship level for either flying to an arbitrary point or flying away from a body. Not an ideological thing. If you have scripting then people will want fine control, even if full-journey functionality also existed at the C++ level.

However, as far as I can tell, the Lua side has no understanding of frames or relative positions, so some abstraction is required. Maybe something like AIEscape(body, dist), that attempts to fly the specified distance away from the target body. Could implement that with the FlyTo code, although the calculation is a little fiddly. There's also a question of how much fuel to use if the function is used for long distances.

I thought people used to fudge this either by using FlyTo on another planet in the system, or one of the orbits. I don't know what the hyperspace rule is at the moment.

fluffyfreak:
This just patches the problem and I'm ok with that for this PR.

To make the departing ships fit our "legal" rules on hyperspacing they would need to reach a certain distance from the port. So fly to low orbit, then jump. Perhaps just having a delayed hyperspace jump whilst they fly-to-orbit would be enough?

fluffyfreak:
@impaktor if we had tail-sitter style launching this would be a non-issue since the main engines are much more powerful and the ships would still be going upwards when they jumped :)

PR #4152, opened 2017-09-12
Bring the XCode project file up to date with the latest files and config.

Philbywhizz:
I've done these type of pull requests in the past (a few years ago) when building the OSX binary app bundle, and they have just been merged in.

If anyone wants to provide any feedback then please do so. All being well, I'll merge this in at a later date just before i release an OSX binary.


fluffyfreak:
Just a question about the `modelcompiler`. 

Do you mean that it gets built, and is used to generate the `.sgm` files, then is stripped from the bundle so it is not distributed to end users?

The idea is that they can use it themselves to build `.sgm` files for mods so it's not really redundant.

Philbywhizz:
Perhaps _redundant_ is too harsh a word to explain it in a one line git commit message. This newer XCode project creates an app bundle, and as far as I am aware you can only link a single binary to the app icon. As it currently stands, the current XCode project was including the `modelcompiler` binary as well, but with no means to run it without a user diving into the command line and messing with the internals of the app bundle itself (which isn't the best way to do it). I'm just tidying up some lose ends by removing the 'redundant' binary.

My current thought however is to provide 2 types of binary packages for OSX of pioneer. One is an .dmg file that contains an app bundle for people who just want to play pioneer.The data directory and all associated files are bundled within in the .app bundle sandbox. Click the icon and go!

The 2nd binary package is more _unix like_ and will be a command-line based binary of pioneer, and modelcompiler and has easy access to the data directories (exposed) . Ideal for power users or modders who would already be familar with the command line.

Both binaries are easy to generate via a script when ever there is a release.

Open to any feedback (we can talk about it outside of this pr if it creates too much noise).

impaktor:
> Open to any feedback (we can talk about it outside of this pr if it creates too much noise).

This seems to be a good place to talk about it.

> Both binaries are easy to generate via a script when ever there is a release.

So there will be two Mac downloads available on the home page? If so, we should perhaps add some description to the home page (this one or the new version of the web page that @ecraven has made), guiding the player what the difference between the versions is.

fluffyfreak:
It sounds good! Thanks for clarifying :)

Philbywhizz:
> So there will be two Mac downloads available on the home page?

There would be 2 different downloads, depending on what you want. The one most people would download would be the .dmg file as it is the "I just want to play" version, so this should be the default.

PR #4161, opened 2017-09-19
Fix engineers constantly repairing engines, and negative jump counter

### LINKED ISSUE #4158: Hyperdrive less likely to fail the longer the time since warranty expired?
### Observed behaviour

Hope I am understanding this correctly, apologies if not. Been flying around in a Venturestar fitted with a class 9 hyperdrive. Over 4 years have passed since the warranty expired.

This is the console output from a sequence of jumps, plus the latest jump at the end. 

DEBUG: Jumps since warranty: -546, chance of failure (if > 0): 1/801
DEBUG: Jumps since warranty: -549, chance of failure (if > 0): 1/804
DEBUG: Jumps since warranty: -561, chance of failure (if > 0): 1/816
DEBUG: Jumps since warranty: -568, chance of failure (if > 0): 1/823
DEBUG: Jumps since warranty: -568, chance of failure (if > 0): 1/823
DEBUG: Jumps since warranty: -578, chance of failure (if > 0): 1/833
Latest jump...
DEBUG: Jumps since warranty: -1311, chance of failure (if > 0): 1/1566

a614c6f

Linux Mint 18 64-bit

impaktor:
Fixes #4158

Problem is the better the engineer the more the engine will break down and be
fixed,  spamming the player with "your drive was fixed" messages.

This bug must have been there for many years, but only noticable when player has a (good) Engineer, leading to more "true" dice rolls. So, the code used to do a dice roll at each jump for each crew member  and check if they should repair the engines, if yes/true, they repair it, even if it hasn't broken down, and even if warranty hasn't expired yet.

In addition, when the jumpcount is set back, it can reach negative values.

I also changed how probability approaches unity for breakdown after warranty expires.

## EDIT 
A little plot to show how probability of engine breakdown increases after service warranty expired (about after 1 year):
jump
 
Now let's see that on a lin-log-scale, to better resolve the old method:
jump_log

(Then subtract from that the crew that can fix it.)

### DIFF
--- a/data/modules/BreakdownServicing/BreakdownServicing.lua
+++ b/data/modules/BreakdownServicing/BreakdownServicing.lua
@@ -248,27 +248,36 @@ local onEnterSystem = function (ship)
 	else
 		return -- Don't care about NPC ships
 	end
-	local saved_by_this_guy = savedByCrew(ship)
-	if (service_history.lastdate + service_history.service_period < Game.time) and not saved_by_this_guy then
+
+	-- Jump drive is getting worn and is running down
+	if (service_history.lastdate + service_history.service_period < Game.time) then
 		service_history.jumpcount = service_history.jumpcount + 1
-		if (service_history.jumpcount > max_jumps_unserviced) or (Engine.rand:Integer(max_jumps_unserviced - service_history.jumpcount) == 0) then
+	end
+
+	-- Test if the engine will actually break
+	if Engine.rand:Number() < service_history.jumpcount / max_jumps_unserviced then
+		print("Player hyperdrive is breaking down!")
+
+		-- The more engineers the more likely to be saved (rule doesn't apply to cooking):
+		local crew_to_the_rescue = savedByCrew(ship)
+
+		if crew_to_the_rescue then
+			-- Brag to the player
+			if crew_to_the_rescue.player then
+				Comms.Message(l.YOU_FIXED_THE_HYPERDRIVE_BEFORE_IT_BROKE_DOWN)
+			else
+				Comms.Message(l.I_FIXED_THE_HYPERDRIVE_BEFORE_IT_BROKE_DOWN,crew_to_the_rescue.name)
+			end
+			-- Rewind the servicing countdown by a random amount based on crew member's ability
+			local fixup = math.max(0, crew_to_the_rescue.engineering - crew_to_the_rescue.DiceRoll())
+			service_history.jumpcount = service_history.jumpcount - fixup
+		else
 			-- Destroy the engine
 			local engine = ship:GetEquip('engine',1)
 			ship:RemoveEquip(engine)
 			ship:AddEquip(Equipment.cargo.rubbish, engine.capabilities.mass)
 			Comms.Message(l.THE_SHIPS_HYPERDRIVE_HAS_BEEN_DESTROYED_BY_A_MALFUNCTION)
 		end
-	end
-	if saved_by_this_guy then
-		-- Brag to the player
-		if saved_by_this_guy.player then
-			Comms.Message(l.YOU_FIXED_THE_HYPERDRIVE_BEFORE_IT_BROKE_DOWN)
-		else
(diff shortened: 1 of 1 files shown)


impaktor:
Btw, this is WIP, because I haven't yet decided the most sane way the probability of engine breakdown should approach unity as function of jumps since service expired.

Did some one say "plotting time?" 😉 
(It's been a while since I went on a plotting rampage, last time)

impaktor:
> Did some one say "plotting time?"

...aaand done. Updated my initial post with plots.

fluffyfreak:
I like the pictures

data/modules/BreakdownServicing/BreakdownServicing.lua
```diff
@@ -248,28 +248,37 @@ local onEnterSystem = function (ship)
 	else
 		return -- Don't care about NPC ships
 	end
+
+	-- The more engineers the more likely to be saved (rule doesn't apply to cooking):
 	local saved_by_this_guy = savedByCrew(ship)
```
laarmen: The only line unchanged, and, bizarrely, the one I dislike the most. This name is crap! Is it a bool, is it a table/object?

data/modules/BreakdownServicing/BreakdownServicing.lua
```diff
@@ -248,28 +248,37 @@ local onEnterSystem = function (ship)
...
+	if (service_history.lastdate + service_history.service_period < Game.time) then
 		service_history.jumpcount = service_history.jumpcount + 1
-		if (service_history.jumpcount > max_jumps_unserviced) or (Engine.rand:Integer(max_jumps_unserviced - service_history.jumpcount) == 0) then
+	end
+
+	-- Test if the engine will actually break
+	if Engine.rand:Number() < service_history.jumpcount / max_jumps_unserviced then
+		print("Player hyperdrive is breaking down!")
+
+		if saved_by_this_guy then
```
laarmen: I don't get it. Why do we identify the savior before we know we need saving?

data/modules/BreakdownServicing/BreakdownServicing.lua
impaktor: I agree, I had to read Character.lua to see what it was. Not to mention: IT'S SEXIST!!!

data/modules/BreakdownServicing/BreakdownServicing.lua
impaktor: Good catch, it's a remnant from the code I modified, as that (i.e. current master) tries to "save the ship" every jump, even if engine isn't busted. Which is why I think it's strange that this old code's strange behaviour hasn't been noticed before. Anyway, I'll fix it.

impaktor:
Fixed the objections. `saved_by_this_guy` now chaned to `is_saved_by_crew`, (was thinking to just have it `is_saved`, but meh).

Regarding the plots above: I realize what is perhaps more interesting is the cumulative probability. In the plots the jumps are treated as independent, so might want to know:
-  "what is the probability that I can make one _and_ two _and_ three jumps?"

Instead, the plots just show:
- "what is the probability of the _i_ th jump leading to a break down.

Taking that into account, the old method might not be that unreasonable. (Although it was crazy to also have an if-statement to force it to break down after 256 jumps, as that would happen anyway as is obvious from the plots).

data/modules/BreakdownServicing/BreakdownServicing.lua
```diff
@@ -244,32 +244,41 @@ end
...
+	-- Test if the engine will actually break
+	if Engine.rand:Number() < service_history.jumpcount / max_jumps_unserviced then
+		print("Player hyperdrive is breaking down!")
+
+		-- The more engineers the more likely to be saved (rule doesn't apply to cooking):
+		local is_saved_by_crew = savedByCrew(ship)
+
+		if is_saved_by_crew then
+			-- Brag to the player
+			if is_saved_by_crew.player then
```
laarmen: If it is a bool, then why does it have an attribute? ;-)

impaktor: Because, "oops"? I'm half right though, because it's `false` if no helping crew is found.

Now called `crew_to_the_rescue`. Hope that's better. Updated.

PR #4165, opened 2017-09-22
Don't hard code when combat rank changes in stats tracking module

### LINKED ISSUE #4162: Combat rating remains the same.
### Observed behaviour

After 48 kills, received message along the lines of "congratulations commander, your combat rating has improved".
Combat rating iterated to Average.
After 64 kills, received message along the lines of "congratulations commander, your combat rating has improved".
Combat rating iterated to Above Average.
After 80 kills, received message along the lines of "congratulations commander, your combat rating has improved".
Combat rating remains at Above Average.

a614c6f
Linux Mint 18 64-bit

impaktor:
...instead use the methods in Character to determine when rank (e.g. 'HARMLESS' etc.) changes.

Fixes #4162 

100% untested, so need to verify there are no typos and works.

### DIFF
--- a/data/modules/StatsTracking/StatsTracking.lua
+++ b/data/modules/StatsTracking/StatsTracking.lua
@@ -23,14 +23,18 @@ local PlayerDamagedShips = {}
 
 local onShipDestroyed = function (ship, attacker)
 	if attacker:isa('Ship') and attacker:IsPlayer() then
+		-- check current/old combat rating
+		local rating_rank = Character.persistent.player:GetCombatRating()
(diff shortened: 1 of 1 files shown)


impaktor:
Dammit! I misread how `Character:IsCombatRated` works, so this won't work in its current form. Will need to have a 'Character:IsCombatRating` method.

impaktor:
> Will need to have a Character:IsCombatRating method.

No, I'm an idiot. Fix was way simpler. Good for testing now.

PR #4166, opened 2017-09-23
Fix autopilot tooltip when (nav)targeting other ships

### LINKED ISSUE #4163: "Unkown AI Command" Tooltip on Autopilot icon
### Observed behaviour
Reported by @crem on IRC
When you instruct the autopilot to fly to a ship which is selected as a nav target, the Autopilot button will switch to a rear crosshair, and the tooltip says "UNKOWN AI COMMAND"
screenshot-20170920-215148

### Expected behaviour
It should say and show the same as it would for Fly to to a body.

### Steps to reproduce
Select a ship from the Orbit map (pressing the little ship icon on the top right will turn show them) or the target of a SAR mission.
Take off and instruct the autopilot to Fly to the vicinity of...

Win7, 20170828 Build

impaktor:
Autopilot currently doesn't have any icon for formation flying,
if indeed that is a nomenclature we want to use for the tooltip.

Closes #4163

Possibly, I'll update this if we get/want a "formation" icon.


impaktor:
I've removed the WIP of this, and rebased. Discussing with @nozmajner on IRC, we'll make no distinction between targeting body or ship, as the internal "formation flying" is misleading.

fluffyfreak:
I'm happy with it

PR #4167, opened 2017-09-23
Don't fail assassins when entering mission system

### LINKED ISSUE #4154: Assassination missions fail
### Observed behaviour

Every time I take on a termination mission, the mission status reverts to "failed" as soon as I have hyperjumped into the star system the assassination is to take place, this behaviour has happened every time without fail, different ships, different star systems, different builds. All other mission types work flawlessly.

screenshot-20170916-181353

Several recent builds, most recent build downloaded and compiled 8th sept 2017

Linux Mint 64 bit

impaktor:
'FAILED' occurs when mission ship either lands or jumps, after taking off at
said mission time point.

But for some reason it looks like a spawned dock ship now will trigger a
`OnShipDocked` event, as the ship is spawned when the player enters the target
system, they will immediately get failed.

Closes #4154.

I'll remove the debug printout commits when I get confirmation this is
working, and ideally, we'll understand why this bug is triggered to begin
with, all of a sudden.

### DIFF
--- a/data/modules/Assassination/Assassination.lua
+++ b/data/modules/Assassination/Assassination.lua
@@ -358,7 +358,11 @@ local onShipDocked = function (ship, station)
 			Event.Queue("onReputationChanged", oldReputation, Character.persistent.player.killcount,
 				Character.persistent.player.reputation, Character.persistent.player.killcount)
 		else
-			if mission.ship == ship then
+			-- Fail occurs when mission ship either lands or jumps,
+			-- after taking off at said mission time point. Spawned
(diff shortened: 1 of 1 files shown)


impaktor:
I've removed the debug-printing commits now, and consider this a OK fix for the symptom.

It is currently unclear to me if it's a bug or feature that `Space.SpawnShipDocked()` triggers the `OnShipDocked()` event, but there must have been a change recently (within the last months? years?) in how this works. I haven't looked further into it. 

Anyway, this should at least get the assassination mission back in working order.

fluffyfreak:
Looks good

PR #4169, opened 2017-09-23
WIP Ship json parser

impaktor:
I've ported the lua scripts for generating the ship stat output to python, since the ship source is no longer in lua. The old scripts didn't work.

I do however have a problem with not quite knowing where this `Infobox_Ship` thingy is. I assume it's some script that is on the wiki server side to typset the info box of each ship? Thing is this is outdated, e.g. it doesn't know about cargoscoop slots and fuelscoop slots. So I wonder where that script resides? Ping @robn 

I'm opening this PR as WIP "prematurely", as I wanted to give a heads up to what I'm doing, and ping rob, or anyone "in the know".

__UPDATE__ 2018-01-14
- [x] single ship json -> wiki infobox (obsolete)
- [x] all ship json -> csv
- [x] all ship json -> wiki table
- [ ] csv -> json files (half done)
- [ ] fix infobox to be up-to-date with current slots / capability


scripts/wiki_ship_stat_parser.py
```diff
@@ -0,0 +1,215 @@
...
+def thrust_empty(thrust, hull_mass, fuel_tank_mass):
+    this_thrust_empty = thrust / (9.81*1000*(hull_mass + fuel_tank_mass))
+    return this_thrust_empty
+
+def thrust_full(thrust, hull_mass, fuel_tank_mass, capacity):
+    this_thrust_empty = thrust / (9.81*1000*(hull_mass + fuel_tank_mass + capacity))
+    return this_thrust_empty
+
+def myround(number, significand):
+    return math.floor(number)
```
mooskagh: factor = 10.0**sig
  return factor * int(x / factor + 0.5)

scripts/wiki_ship_stat_parser.py
```diff
@@ -0,0 +1,215 @@
...
+
+def thrust_empty(thrust, hull_mass, fuel_tank_mass):
+    this_thrust_empty = thrust / (9.81*1000*(hull_mass + fuel_tank_mass))
+    return this_thrust_empty
+
+def thrust_full(thrust, hull_mass, fuel_tank_mass, capacity):
+    this_thrust_empty = thrust / (9.81*1000*(hull_mass + fuel_tank_mass + capacity))
+    return this_thrust_empty
+
+def myround(number, significand):
```
mooskagh: I'd do significant=1  as it's with what it's usually called

scripts/wiki_ship_stat_parser.py
```diff
@@ -0,0 +1,215 @@
...
+        shipjson = json.load(data_file)
+
+    return shipjson
+
+def thrust_empty(thrust, hull_mass, fuel_tank_mass):
+    this_thrust_empty = thrust / (9.81*1000*(hull_mass + fuel_tank_mass))
+    return this_thrust_empty
+
+def thrust_full(thrust, hull_mass, fuel_tank_mass, capacity):
+    this_thrust_empty = thrust / (9.81*1000*(hull_mass + fuel_tank_mass + capacity))
```
mooskagh: Wrong variable name "_empty".
Also, instead of having two functions, I'd just have one, with capacity=0 default parameter value.

scripts/wiki_ship_stat_parser.py
```diff
@@ -0,0 +1,215 @@
...
+                                               (shipTable['hull_mass'] + shipTable['capacity']))
+
+    deltaV_empty = myround(deltaV_empty, -3) / 1000
+    deltaV_full = myround(deltaV_full, -3) / 1000
+
+    print('{{Infobox_Ship')
+    print("|name = %s" % shipTable['name'])
+    print("|type = %s" % shiptype[shipTable['ship_class']])
+    print("|manufacturer = %s" % manufacturers[shipTable['manufacturer']])
+    print("|forward_thrust = empty: %sG full: %sG" % (forward_thrust_empty, forward_thrust_full))
```
mooskagh: Large portion of code above and those few lines may be simplified to:

```
for d in ['forward', 'reverse', 'up', 'down', 'left', 'right']:
   empty = myround(thrust_empty(shipTable[d + '_thrust'], shipTable['hull_mass'], shipTable['fuel_tank_mass'], shipTable['capacity']), 1)
   full = myround(thrust_full(shipTable[d + '_thrust'], shipTable['hull_mass'], shipTable['fuel_tank_mass']), 1)
   print("|%s_thrust = empty: %sG full: %sG" % (d, empty, full))
```

scripts/wiki_ship_stat_parser.py
```diff
@@ -0,0 +1,215 @@
...
+import math
+import sys
+import copy
+import json
+import numpy as np
+import operator                 # for sorted
+import sys
+from pprint import pprint
+
+shiptype = {
```
mooskagh:
```
def GetShipType(s):
  return "%s [[File:%s.png]]" % (s.replace('_', ' ').capitalize(), s)
```

scripts/wiki_ship_stat_parser.py
```diff
@@ -0,0 +1,215 @@
...
+    "medium_passenger_transport" : "Medium passenger transport [[File:medium_passenger_transport.png]]",
+}
+
+manufacturers = {
+    "albr" : "Albr Corp.",
+    "auronox" : "Auronox Corporation",
+    "haber" : "Haber Corporation",
+    "kaluri" : "OKB Kaluri",
+    "mandarava" : "Mandarava-Csepel",
+    "opli" : "OPLI-Barnard Inc."
```
mooskagh: I suggest adding a trailing comma into that line. That way if it's formatted using yapf, it will format it vertically (like it is now) rather than in one line.

impaktor: done

scripts/wiki_ship_stat_parser.py
impaktor: Thanks, my code was just a stub, needed to see if script would run, and to-do / fix later.

scripts/wiki_ship_stat_parser.py
impaktor: True. This is just a copy-paste inheritance from the original lua-script. Fixed.

scripts/wiki_ship_stat_parser.py
impaktor: True, but these strings are generated by hand "upstream", thus no  guartantee the patter will always be true. In fact, I don't know if it is true even now, as I haven't tested it for all ships/icons.

mooskagh: Then I'd still keep the dict of exceptions, and use generated ones for the rest.

johnbartholomew:
Infobox_Ship is here:
https://pioneerwiki.com/wiki/Template:Infobox_Ship

@impaktor What else needs to be done to take this from WIP to mergeable? (Some commit history cleanup perhaps?)

impaktor:
> What else needs to be done to take this from WIP to mergeable?

0. get thumb out of butt
1. convert csv -> <ship>.json (will help noz test rebalanced specs from spreadsheet)
2. test interaction with wiki
3. clean up commit history.

Thanks for pointing out the wiki-link!

Priority will be point 1. I'm unlikely to have time this week. Hope to pick this up in January.

ecraven:
Can you rebase this, and comment out the csv options for now? Then we'll merge it, I can fix it to work with the current wiki formatting, then we'll continue work on the csv stuff.

impaktor:
On it now. 

Btw, I seem unable to re-gain my registered nick on irc, thus am mute for some time there.

impaktor:
Ye, let it be know, that this was merged knowing parsing of Lodos doesn't quite work, but we needed it in master asap.

PR #4171, opened 2017-09-25
Don't show 'None' for drive specs when asking to service engine

impaktor:
## Problem
When asking for service but without a drive (as in Bernards):
2017-09-25-171251_942x666_scrot

## Solution

I split the intro message for "Break down and service" module, to separate the string that specifies the `drive` of the player, so that this sentence can be replaced by the "you don't have a drive" message.
2017-09-25-171144_939x666_scrot


impaktor:
Also, I've tested all six flavours both with and without hyperdrive. Scream if you don't like it. I'll merge in the comming days.

Issue #4174, opened 2017-10-01
Pioneer crash at start


Tichy:
After updating my local git repo, I compiled Pioneer but the new version crash as soon as it starts.
Visually I just see the window appear then disappear.

Launching from the terminal I obtain this:

    ver git (8e90cc7f9) on: Linux
    
    System Name: Linux
    Host Name: ono-sendai
    Release(Kernel) Version: 4.13.3-1-ARCH
    Kernel Build Timestamp: #1 SMP PREEMPT Thu Sep 21 20:33:16 CEST 2017
    Machine Arch: x86_64
    Domain Name: (none)
    
    SDL Version 2.0.6
    Initialized OpenGL 3.1, with extensions, renderer
    invalid axis binding '' in config file for BindAxisPitch
    invalid axis binding '' in config file for BindAxisRoll
    invalid axis binding '' in config file for BindAxisYaw
    started 7 worker threads
    ShipType::Init()
    Lua::Init()
    Fonts:
    font orbiteer:
    - DejaVuSans.ttf 1.200000
    - wqy-microhei.ttc 1.000000
    - Orbiteer-Bold.ttf 1.000000
    font pionillium:
    - DejaVuSans.ttf 0.928571
    - wqy-microhei.ttc 1.000000
    - PionilliumText22L-Medium.ttf 1.000000
    import [libs/autoload.lua]: libs/autoload.lua did not return anything
    import [ui/FileDialog.lua]: ui/FileDialog.lua did not return anything
    Errore di segmentazione (core dump creato)

I tried to clone again from remote and compile, and I tried to rename my ~/.pioneer directory to exclude my settings.

__My pioneer version (and OS):__

ver git (8e90cc7f9)
Release(Kernel) Version: 4.13.3-1-ARCH
Kernel Build Timestamp: #1 SMP PREEMPT Thu Sep 21 20:33:16 CEST 2017
Machine Arch: x86_64
NVIDIA nasty proprietary drivers ver. 384.90
KDE Plasma 5.10 as DE

ColdSpirit0:
Hmm... works good on my 4.12.13-1-ARCH. Tried to compile exact same version of pioneer and it runs without fail for me.
Maybe path to pioneer directory have some no-english characters? If so, try to move it like to `/data/pioneer/` and run from here.

[...]

PR #4176, opened 2017-10-04
isFile check for ReadFile (fix for macos crash)

ColdSpirit0:
For posix now, I'll test win32 and then will be possible to merge.

### DIFF
--- a/src/posix/FileSystemPosix.cpp
+++ b/src/posix/FileSystemPosix.cpp
@@ -149,33 +149,36 @@ namespace FileSystem {
 	RefCountedPtr<FileData> FileSourceFS::ReadFile(const std::string &path)
 	{
 		const std::string fullpath = JoinPathBelow(GetRoot(), path);
-		FILE *fl = fopen(fullpath.c_str(), "rb");
-		if (!fl) {
-			return RefCountedPtr<FileData>(0);
-		} else {
-			Time::DateTime mtime;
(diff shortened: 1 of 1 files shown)


ColdSpirit0:
That bug does not exists on windows 7, but better to test it also on other windows versions.

Philbywhizz:
Tested under MacOSX and it works as designed.

jaj22:
Minor bug: This code doesn't call fclose if it opens a directory.

I think it's cleaner to use stat_path before opening rather than stat_fd after opening.

ColdSpirit0:
All edits done

impaktor:
@jaj22 do you have any further objections, or good to merge?

@ColdSpirit0 was this bug reported on the tracker somewhere? (Doesn't look like it.)

ColdSpirit0:
@impaktor no, it was not reported as a separate issue, but you can see some info about that bug in #4174

jaj22:
@impaktor good to merge.

@ColdSpirit0 I would also recommend a related fix to LuaUtils.cpp: If ImportInfo.isFullName is false, it shouldn't attempt to open the file as if it's a full name. That's the i==0 part of the loop at line 342.

ColdSpirit0:
@jaj22 im think i did not understand you. When isFullName is true, that means import should load only by given name, when it's false, means import should build relative paths and try to load each of them.

PR #4177, opened 2017-10-04
Don't call SDL_FreeSurface until refcount is zero

jaj22:
Workaround for a crash bug in SDL 2.0.6, where SDL_FreeSurface doesn't work correctly when called with refcount >1. There's a [development patch for SDL](https://hg.libsdl.org/SDL/rev/a09c3f87a12f) but it may not hit release for a year or more.

This patch changes the behaviour of the SDLSurfacePtr wrapper to not call SDL_FreeSurface until the surface reference count reaches zero. As SDLSurfacePtr already adjusts the refcount manually, there should be no additional side effects with other SDL versions.

I don't fully understand the SDLSurfacePtr code, but this shouldn't make it any worse.

Works around the primary issue in #4174. There's other stuff in there so don't close it yet.

### DIFF
--- a/src/SDLWrappers.h
+++ b/src/SDLWrappers.h
@@ -22,7 +22,9 @@ public:
 	SDLSurfacePtr() {}
 	SDLSurfacePtr(const SDLSurfacePtr& b): base_type(b.Get())
 	{ if (this->m_ptr) this->m_ptr->refcount += 1; }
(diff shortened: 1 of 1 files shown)


ColdSpirit0:
Works with sdl 2.0.6.1 and 2.0.5.3

PR #4178, opened 2017-10-05
Fix SAR selection of ship sizes for missions. Fixes #4175.

### LINKED ISSUE #4175: Problem with save and rescue mission.
### Observed behaviour
Accepted a mission to save and rescue 1 crew and 1 passenger from a ship that had crash landed on Vesta

With 2 passenger cabins on board, flew to Vesta, landed <50m from ship, received comms message "cannot complete transfer, passenger missing"

Launched from surface, re-landed next to mission ship, missing passenger still missing.

Returned to Gates starport, comms message "Thank you for your efforts, I hope you will return to complete the task" 
Mission still active, returning not an option with just 2 hours remaining. 

Savegame:
[Missing_Pass.pdf](https://github.com/pioneerspacesim/pioneer/files/1348908/Missing_Pass.pdf)

Linux 64bit
8e90cc7

clausimu:
This fixes #4175. I'm now taking prospective hyperdrive size into account when selecting ships for passenger rescue missions. The crude way I did this before led to rare occasions where passenger cabins could not be fitted on the ship because of lack of space.

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -613,7 +613,20 @@ local createTargetShipParameters = function (flavour)
 	---- loading drive, weapons etc.
 	if flavour.id == 1 or flavour.id == 6 then
 		for i,shipdef in pairs(shipdefs) do
-			if shipdef.capacity / 10 < 1 then shipdefs[i] = nil end
+
+			-- get mass of hyperdrive if this ship has a default drive
+			-- if no default drive assume lowest mass drive
+			-- higher mass drives will only be fitted later at ship creation if capacity is huge
+			local drive = Equipment.hyperspace['hyperdrive_'..tostring(shipdef.hyperdriveClass)]
+			if not drive then
+				local drives = {}
+				for i = 9, 1, -1 do
+					table.insert(drives, Equipment.hyperspace['hyperdrive_'..tostring(i)])
+				end
+				table.sort(drives, function (a,b) return a.capabilities.mass < b.capabilities.mass end)
+				drive = drives[1]
+			end
+			if (shipdef.capacity-drive.capabilities.mass) / 10 < 1 then shipdefs[i] = nil end
 		end
 	elseif flavour.pickup_pass > 0 then
 		for i,shipdef in pairs(shipdefs) do
@@ -661,7 +674,16 @@ local createTargetShipParameters = function (flavour)
 	if flavour.id == 1 or flavour.id == 6 then
 		local any_pass = rand:Integer(0,1)
 		if any_pass > 0 then
-			pickup_pass = rand:Integer(1, math.min((shipdef.capacity / 10)+1, max_pass))
+			local drive = Equipment.hyperspace['hyperdrive_'..tostring(shipdef.hyperdriveClass)]
+			if not drive then
+				local drives = {}
+				for i = 9, 1, -1 do
+					table.insert(drives, Equipment.hyperspace['hyperdrive_'..tostring(i)])
+				end
+				table.sort(drives, function (a,b) return a.capabilities.mass < b.capabilities.mass end)
+				drive = drives[1]
+			end
+			pickup_pass = rand:Integer(1, math.min(((shipdef.capacity-drive.capabilities.mass) / 10)+1, max_pass))
 		else
 			pickup_pass = 0
 		end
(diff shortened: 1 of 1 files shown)


data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -1930,7 +1940,7 @@ local onCreateBB = function (station)
 	-- local num = 3
 	-- for _ = 1,num do
 	-- 	makeAdvert(station, 1, closestplanets)
-	-- 	makeAdvert(station, 2, closestplanets)
+	--	makeAdvert(station, 2, closestplanets)
 	-- 	makeAdvert(station, 3, closestplanets)
```
impaktor: Does this replace a space for tab or the other way around?

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -1967,6 +1977,18 @@ local onUpdateBB = function (station)
...
+	--	makeAdvert(station, 1, closestplanets)
+	--	makeAdvert(station, 2, closestplanets)
+	-- 	makeAdvert(station, 3, closestplanets)
+	-- 	makeAdvert(station, 4, closestplanets)
+	-- 	makeAdvert(station, 5, closestplanets)
+	-- 	makeAdvert(station, 6, closestplanets)
+	-- 	makeAdvert(station, 7, closestplanets)
+	-- end
+
 	-- trigger new ad creation if appropriate
```
impaktor: I'm not sure we normally save debug print out in the finished file, but I guess it doesn't hurt. I have no real opinion one way or the other

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -613,7 +613,14 @@ local createTargetShipParameters = function (flavour)
 	---- loading drive, weapons etc.
 	if flavour.id == 1 or flavour.id == 6 then
 		for i,shipdef in pairs(shipdefs) do
-			if shipdef.capacity / 10 < 1 then shipdefs[i] = nil end
+
+			-- get mass of hyperdrive if this ship has a default drive
+			-- if no default drive only 4t will have to be assumed for lowest tech drive
+			-- higher tech drives will only be fitted later if capacity is huge
+			local drive_mass = 4
+			local default_drive = Equipment.hyperspace['hyperdrive_'..tostring(shipdef.hyperdriveClass)]
```
impaktor: technically, I can imagine we sometimes in the future rebalance ships and hyperdrives, so hard coding `drive_mass` might only work up til then. Is it possible to have query / sort all hyperdrives, and then get the smallest number out? Would be a more stable solution

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -661,7 +668,10 @@ local createTargetShipParameters = function (flavour)
 	if flavour.id == 1 or flavour.id == 6 then
 		local any_pass = rand:Integer(0,1)
 		if any_pass > 0 then
-			pickup_pass = rand:Integer(1, math.min((shipdef.capacity / 10)+1, max_pass))
+			local drive_mass = 4
```
impaktor: same as above. (do you need to set it twice?)

impaktor:
Also, when/if you fix this, please do `git add <file>`, `git commit --amend`; `git push -f`, to squash fix into existing commit.

data/modules/SearchRescue/SearchRescue.lua
clausimu: I need to do it twice. First time shipdefs are filtered based on projected need for equipment capacity and several other things (depending on mission specifics). Second time the actual amount of passengers to be rescued is calculated and fed back to the mission. Both have to happen before the ad is created.

data/modules/SearchRescue/SearchRescue.lua
clausimu: Okay, I understand your point. I'll query for the lowest mass engine instead of hardcoding its current mass.

data/modules/SearchRescue/SearchRescue.lua
clausimu: I've always left this in for faster debugging. But I can remove it if you prefer.

clausimu:
I just amended with the new fixes. I also changed the hyperdrive selection for the actual ship creation slightly to match the ship parameters fixed here. This should prohibit weird surprises when the ship is actually being built.

PR #4179, opened 2017-10-05
Removed unnecessary symlink + pioner now executes from any place

ColdSpirit0:
Symlink did not work by default from custom repo directory (not in /home/). Im think it fixes that problem.
Also you can to move, in example, `data/pigui/` and run `../../pioneer` from there.

### DIFF
--- a/generic-exec
+++ b/generic-exec
@@ -1,8 +1,0 @@
-#!/bin/sh
-# Executable wrapper script
-# Ensures the program is up-to-date, then hands over
-
-PROGRAM_NAME=$0
-
-EXEC_BIN=src/$PROGRAM_NAME
-make && exec $EXEC_BIN $*
--- a/pioneer
+++ b/pioneer
@@ -1,1 +1,13 @@
-generic-exec
+#!/bin/sh
+# Executable wrapper script
+# Ensures the program is up-to-date, then hands over
+
+PROGRAM_NAME=$(basename $0)
+
+EXEC_BIN=src/$PROGRAM_NAME
+START_DIR=$(pwd)
+WORK_DIR=$(dirname $0)
+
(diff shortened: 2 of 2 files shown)


impaktor:
Does anyone know why we have the `generic-exec` to begin with? I don't like removing stuff that seems to have had some motivation for being there at some point. I'd bet @robn would know.

ColdSpirit0:
It might be helpful:
https://github.com/pioneerspacesim/pioneer/commits/master/pioneer
https://github.com/pioneerspacesim/pioneer/commits/master/generic-exec

impaktor:
well, ping @johnbartholomew then.

ColdSpirit0:
As i can see the history, all was so:
1. Was created `generic-exec`, and symlinks for that `pioneer`, `modelviewer`, each symlink calls own component through `generic-exec` (614005afb)
2. `modelviewer` was renamed to `lmrmodelviewer` (c505ffdd5)
3. `lmrmodelviewer` was deleted (0c09a75c5)
4. ???
5. Now we have that PR =D

johnbartholomew:
I added this because I didn't like having to run two commands to build and then run. It has never been necessary, only a mild convenience. But I stopped using it later on anyway, even when I was still an active contributor. I certainly have no objection to it being removed; it's a weird little thing that probably no one else uses(?).

ColdSpirit0:
> run two commands to build and then run
Yo still can do this with same command ./pioneer, just it not a symlink now, it's a shell script.

johnbartholomew:
Oh. Yeah, sorry, I should have looked at the pull request more carefully. Still, the end result is the same: no objection from me.

fluffyfreak:
No objections? merging.

PR #4180, opened 2017-10-05
Trigger onShipDestroyed on fatal crash into city

### LINKED ISSUE #3512: Game crash because of destroyed police ship

robothauler:
Hope this fixes #3512 once and for all :-)


fluffyfreak:
Have noticed this, will test asap

fluffyfreak:
Tested, doesn't crash on save anymore.

PR #4182, opened 2017-10-13
Begin removing direct libpng usage

fluffyfreak:
We have a dependency on `libpng` but we also use `SDL_image`, this seems to work ok on Linux machines where there is a single global version of liubpng. 

However for the Windows (VS2015/et al) version we have different versions of libpng that are linked against and different version of the DLLs get loaded!
At time this causes screenshots to fail, along with several other uses and warnings which can interfere with development.

This is pretty nasty and given that there is a function in SDL_image to save PNG files it also seems unnecessary.

So this PR uses the `IMG_SavePNG` to save an image instead of doing all the work ourselves using libpng.

### DIFF
--- a/src/PngWriter.cpp
+++ b/src/PngWriter.cpp
@@ -21,11 +21,31 @@ void write_png(FileSystem::FileSourceFS &fs, const std::string &path, const Uint
 	bmask = 0x00ff0000;
 	amask = (bytes_per_pixel == 3) ? 0 : 0xff000000;
 #endif
+
 	// create a surface
-	SDL_Surface *surface = SDL_CreateRGBSurfaceFrom((void*)bytes, width, height, bytes_per_pixel * 8, width * bytes_per_pixel, rmask, gmask, bmask, amask);
-	const std::string fname = FileSystem::JoinPathBelow(fs.GetRoot(), path);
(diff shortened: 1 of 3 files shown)


impaktor:
Net effect: less code.

fluffyfreak:
@impaktor yep just one of the benefits. Slight issue is that I believe it converts all images to 32-bit PNGs regardless so it might be a bit slower. Need to look into that actually.

impaktor:
> Slight issue is that I believe it converts all images to 32-bit PNGs regardless so it might be a bit slower. 

1. on windows or all platforms? (I assume the latter)

2. just on upstart? I can live with that.

fluffyfreak:
1. on all platforms
2. on usage, it converts from RGB to RGBA when called. So long as we're passing RGBA it makes no difference which is what I will check now.

nozmajner:
Most, if not all of the pngs are icons, so they should be RGBA already.

fluffyfreak:
This is only for saving so doesn't affect the loading of icons or other images which we already use SDL_image for :)

nozmajner:
Oh, OK then :D

laarmen:
All in all, I don't think it's such a big deal if the PNG are bigger. We're
only talking about the screenshots, right?

fluffyfreak:
Ideally they shouldn't even be bigger since the alpha channel will be blank.

I'm now flipping the image vertically so it matches the OpenGL screen format. Also I'm just reading the format in RGBA instead of RGB so it avoids another conversion step.

fluffyfreak:
Merged, this is just step one, I'll do makefiles/projects later.

PR #4185, opened 2017-10-18
Added -skipmenu/-sm flag

ColdSpirit0:
Use './pioneer -skipmenu' or './pioneer -sm' to skip main menu and load game on Earth.
If all ok I can squash these commits.

### DIFF
--- a/src/LuaEngine.cpp
+++ b/src/LuaEngine.cpp
@@ -950,7 +950,7 @@ static int l_engine_world_space_to_screen_space(lua_State *l)
 	vector3d pos = LuaPull<vector3d>(l, 1);
 
 	std::tuple<bool, vector3d, vector3d> res = lua_world_space_to_screen_space(pos); // defined in LuaPiGui.cpp
-	
+
 	LuaPush<bool>(l, std::get<0>(res));
 	LuaPush<vector3d>(l, std::get<1>(res));
 	LuaPush<vector3d>(l, std::get<2>(res));
@@ -961,7 +961,7 @@ static int l_engine_world_space_to_ship_space(lua_State *l)
 {
 	vector3d vec = LuaPull<vector3d>(l, 1);
 	auto res = vec * Pi::game->GetPlayer()->GetOrient();
-	
+
 	LuaPush<vector3d>(l, res);
 	return 1;
 }
--- a/data/ui/MainMenu.lua
+++ b/data/ui/MainMenu.lua
@@ -97,11 +97,17 @@ local doQuitConfirmation = function()
 	end
 end
 
+local loadOptions = {
+	function () Game.StartGame(SystemPath.New(0,0,0,0,6),48600) ; setupPlayerSol() end,
+	function () Game.StartGame(SystemPath.New(1,-1,-1,0,4)) ; setupPlayerEridani() end,
+	function () Game.StartGame(SystemPath.New(-1,0,0,0,16)) ; setupPlayerBarnard() end,
+}
+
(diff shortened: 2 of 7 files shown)


impaktor:
Sweet. Maybe one commit. Or two. I don't know. 

Personally, I start at Bernards, because it's fewer buildings to render, and faster. I know others prefer Earth start, as it is good to test redering for said reasons. (Just thinking out loud).

ColdSpirit0:
Not tested in windows.

@impaktor yeah, im thought about other planets, but im did it for ui tests... maybe somehow I'll make choice, like '-sm=3', but later (or maybe tomorrow if I have a good mood).

ColdSpirit0:
Now you can to set load option what you want. In example: `./pioneer -sm=number` or `./pioneer -skipmenu=number` where 'number' is planet number in main menu (starts from 1)

- [-sm=1] and [-sm] Loading at Earth
- [-sm=2] Loading at New Hope
- [-sm=3] Loading at Barnard Star
- [-sm=<-999, -1, 0, 4, 42>] Does not work for now, but easy to add new starting option and allow that and other numbers to load within lua code.

Tested in Windows 7 and Arch linux.

fluffyfreak:
@ColdSpirit0 nice! I'd like to test it out but this looks ready to merge already 👍 

You don't have to do it for this PR but in the future an option to supply the start system `<-999, -1, 0, 4, 42>` would be very useful when I'm testing/developing things.

src/LuaEngine.cpp
```diff
@@ -1105,6 +1113,7 @@ void LuaEngine::Register()
 		{ "CameraSpaceToScreenSpace", l_engine_camera_space_to_screen_space },
 		{ "WorldSpaceToScreenSpace",     l_engine_world_space_to_screen_space },
 		{ "WorldSpaceToShipSpace",     l_engine_world_space_to_ship_space },
+		{ "GetLoadOption", l_get_load_option },
```
laarmen: Not aimed at this contribution, but this module is becoming way too much of a kitchen sink.

ColdSpirit0: Where it should be? Game.lua? Im thought about it, and selected Engine as best choice for me)

ColdSpirit0:
@laarmen nice review, thanks! I'll try to do that.

ColdSpirit0:
@laarmen changes added, please check. It still stores all game time, now in the Pi class. Dont know how to avoid it without make the solution dirty =)

laarmen:
To avoid the global, I would pass the start option to Pi::Start() instead of Pi::Init(), in main.cpp around line 130, and from there just pass it as an extra argument all the way 'til the MainMenu call.

ColdSpirit0:
@laarmen for some reason Im thought that the passing argument to function each time costs more than saving it in memory) But yeah, that better. Please check again

laarmen:
It's a tad more complicated than that. It depends, amongst other things, on
the size of the values, the number of arguments on the stack, the length of
the functions using it, the number of registers in the proc, the size of
the different caches, the format your executable is in, and the compiler
you used. And the age of the captain's daughter.

ColdSpirit0:
@laarmen can I squash commits already?

laarmen:
I haven't looked closely at the code but from an architectural standpoint, LGTM. Since the others seem happy with it, let's open the merging gates! (which I won't do for lack of having a git client on my phone)

ColdSpirit0:
@laarmen thanks!

Ready to merge

PR #4195, opened 2017-11-25
Rounded rewards

caligari87:
This (very simple) pull request changes the assassination, cargo run, deliver package, and taxi missions to use rounded (whole dollar) reward amounts. It also changes the amount readouts to truncate the `.00` cent part. In my opinion this feels more "human" and cleans up the look of the BBS somewhat.

image
image

The rounding is *upward* in case anyone cares. Extremely slight modification in the player's favor but it's negligible in my opinion.


fluffyfreak:
Thanks @caligari87

PR #4198, opened 2017-11-30
Fix the MainMenu nil args crash

laarmen:
I also did a bit of cleaning up.

It's late, I'm tired, and on windows, so I can basically only tell you that it builds (and should work, of course ;-) )

### DIFF
--- a/data/ui/MainMenu.lua
+++ b/data/ui/MainMenu.lua
@@ -189,18 +189,11 @@ local menu =
 				)
 		})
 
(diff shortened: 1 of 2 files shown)


fluffyfreak:
I was going to test this just now but because of the path/branch name I can't check it out on windows! 😁 `fix/4197_startmenu_nil_args` is not a valid path-spec apparently... fuck it I'll just merge it.

PR #4201, opened 2017-12-16
Refactor buffers to have a capacity and size

fluffyfreak:
Refactor buffers to have a capacity and size so they can be reused instead of discarded.

The idea is to reuse Vertex/Index/Instance buffers when the data they're to be populated with is smaller than their current capacity. If the buffer is too small then it must still be discarded and a new one created.

This helps to avoid thrashing the drivers to create and destroy buffers all the time which can be very expensive.

This might alleviate performance problems on some ATi/AMD GPU drivers that have been reported in the past. Although we did a lot of work to fix those with @jaj22 already tuning up our OpenGL usage.

### DIFF
--- a/src/NavLights.cpp
+++ b/src/NavLights.cpp
@@ -200,7 +200,7 @@ void NavLights::Render(Graphics::Renderer *renderer)
 
 	const bool bVBValid = m_billboardVB.Valid();
 	const bool bHasVerts = !m_billboardTris.IsEmpty();
-	const bool bVertCountEqual = bVBValid && (m_billboardVB->GetVertexCount() == m_billboardTris.GetNumVerts());
+	const bool bVertCountEqual = bVBValid && (m_billboardVB->GetSize() == m_billboardTris.GetNumVerts());
 	if( bHasVerts && (!bVBValid || !bVertCountEqual) )
 	{
 		//create buffer
--- a/src/NavLights.cpp
+++ b/src/NavLights.cpp
@@ -198,10 +198,10 @@ void NavLights::Render(Graphics::Renderer *renderer)
 		m_billboardRS = renderer->CreateRenderState(rsd);
 	}
 
-	const bool bVBValid = m_billboardVB.Valid();
-	const bool bHasVerts = !m_billboardTris.IsEmpty();
-	const bool bVertCountEnough = bVBValid && (m_billboardTris.GetNumVerts() <= m_billboardVB->GetCapacity());
-	if( bHasVerts && (!bVBValid || !bVertCountEnough) )
+	const bool isVBValid = m_billboardVB.Valid();
+	const bool hasVerts = !m_billboardTris.IsEmpty();
+	const bool isVertCountEnough = isVBValid && (m_billboardTris.GetNumVerts() <= m_billboardVB->GetCapacity());
+	if( hasVerts && (!isVBValid || !isVertCountEnough) )
 	{
 		//create buffer
 		// NB - we're (ab)using the normal type to hold (uv coordinate offset value + point size)
@@ -217,7 +217,7 @@ void NavLights::Render(Graphics::Renderer *renderer)
 
 	if(m_billboardVB.Valid())
 	{
-		if(bHasVerts) {
+		if(hasVerts) {
 			m_billboardVB->Populate(m_billboardTris);
 			renderer->SetTransform(matrix4x4f::Identity());
 			renderer->DrawBuffer(m_billboardVB.Get(), m_billboardRS, matHalos4x4.Get(), Graphics::POINTS);
--- a/src/NavLights.cpp
+++ b/src/NavLights.cpp
@@ -200,8 +200,8 @@ void NavLights::Render(Graphics::Renderer *renderer)
 
 	const bool bVBValid = m_billboardVB.Valid();
 	const bool bHasVerts = !m_billboardTris.IsEmpty();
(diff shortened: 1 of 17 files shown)


impaktor:
Beyond my understanding.
Merge and see if anyone reports a bug, or improvement, I guess.

So December is getting some work.

fluffyfreak:
@impaktor I think this needs a little testing by a few hands, just people looking at the UI, debug info, something they're familiar with etc.

...then merge 👍

impaktor:
Sadly, I don't have my "pioneer"-computer at the moment (i.e. comming weeks), as that's at work.

src/graphics/VertexBuffer.h
```diff
@@ -56,19 +56,30 @@ struct VertexBufferDesc {
...
 public:
 	virtual ~Mappable() { }
 	virtual void Unmap() = 0;
 
+	inline Uint32 GetSize() const { return m_size; }
+	inline Uint32 GetCapacity() const { return m_capacity; }
+
 protected:
-	Mappable() : m_mapMode(BUFFER_MAP_NONE) { }
+	Mappable(const Uint32 size) : m_mapMode(BUFFER_MAP_NONE), m_size(size), m_capacity(size) { }
```
johnbartholomew: Nit: Would be good to mark this constructor as 'explicit' otherwise Uint32 can be magically silently converted to a Mappable.

src/graphics/VertexBuffer.h
```diff
@@ -56,19 +56,30 @@ struct VertexBufferDesc {
...
 	BufferMapMode m_mapMode; //tracking map state
+
+	// size is the current number of elements in the buffer
+	Uint32 m_size;
+	// capacity is the maximum number of elements that can be put in the buffer
+	Uint32 m_capacity;
+
+private:
+	Mappable() {}
```
johnbartholomew: Nit: Why is this private constructor is needed? It looks like it can just be deleted.

fluffyfreak: Probably can be deleted, at the time I was trying something out.

src/NavLights.cpp
```diff
@@ -200,8 +200,8 @@ void NavLights::Render(Graphics::Renderer *renderer)
 
 	const bool bVBValid = m_billboardVB.Valid();
 	const bool bHasVerts = !m_billboardTris.IsEmpty();
-	const bool bVertCountEqual = bVBValid && (m_billboardVB->GetVertexCount() == m_billboardTris.GetNumVerts());
-	if( bHasVerts && (!bVBValid || !bVertCountEqual) )
```
laarmen: Random note, but at some point we're going to have to be consistent in our local variable naming schemes. I'm guessing the b is for boolean though :)

src/Pi.cpp
```diff
@@ -1542,6 +1531,19 @@ void Pi::MainLoop()
 			}
 		}
 
+#if WITH_DEVKEYS
+		// NB: this needs to be rendered last so that it appears over all other game elements
+		//	preferrably like this where it is just before the buffer swap
```
laarmen: Ooooh, yay for the comments!

fluffyfreak: 😁

src/NavLights.cpp
fluffyfreak: yes, really I should have just renamed them to something like `isValid`, but I have a bad habit of just going along with whatever the local (file/function local) scheme already happens to be instead of enforcing ours!

fluffyfreak:
@johnbartholomew & @laarmen thanks for the feedback! Changes made.

src/graphics/Drawables.cpp
```diff
@@ -231,7 +231,7 @@ void Lines::SetData(const Uint32 vertCount, const vector3f *vertices, const Colo
 	m_refreshVertexBuffer = true;
 
 	// if the number of vert mismatches then clear the current vertex buffer
-	if( m_vertexBuffer.Valid() && m_vertexBuffer->GetVertexCount() != vertCount ) {
+	if( m_vertexBuffer.Valid() && m_vertexBuffer->GetSize() != vertCount ) {
```
laarmen: Shouldn't this match the multiple colors version below with a capacity check? I'm confused.

laarmen:
Oh, I wasn't asking for a change on the prefix stuff, mostly thinking out loud. The change is much better though IMHO.

src/graphics/Drawables.cpp
fluffyfreak: Yes it absolutely should. 
Not all these got changed because I did a search/replace for GetVertexCount() -> GetSize() and then did a highly scientific way of finding where buffer allocations occurred by sticking a breakpoint in the constructor and waiting until it got hit :)

jaj22:
Needs performance testing with various drivers. I forget the precise rules here, but GPU buffering optimizations are counter-intuitive: One of the problems with ATI/AMD drivers was that writing to previously-used buffers forces a pipeline flush, so "thrashing the drivers" by discarding the old buffer was frequently the only way to get decent performance.

fluffyfreak:
Unfortunately I don't have any Ati/AMD hardware to test with.

fluffyfreak:
Ok having thought about this it shouldn't really matter to the affected driver except to reduce buffer turnover.

If it is optimal to use double-buffering or to discard buffers frequently then that's a different behaviour entirely which we should persue intentionally.

PR #4202, opened 2017-12-17
WKFO's Military hyperdrive sound effects

johnbartholomew:
Should no longer break save-games. Although this isn't structured the way I'd like for the long term, since it only makes a minor change to the save format (should be backward/forward compatible) I think this is ready for review to bring the new sound effects into use.

Sound effects courtesy of @WKFO:
https://pioneerspacesim.net/forum/viewtopic.php?p=5332#p5332

Implementation notes:
* C++ remains in charge of triggering the sound effects at the right moment.
* Lua gains control of *which* sound effects are played (for warmup, abort, jump)
* Each hyperjump can use different sounds (if you're crazy); Lua passes a table with sound effect selection to the C++ side when calling InitiateHyperjumpTo.
* Sound effects are chosen at jump time on the Lua side based on which fuel type is being used. This is arguably a bit of a hack but means that the equipment *data* (which is embedded in existing save games) doesn't need to change, only the script code (which is *not* in the save files) changes.
* The set of sound effects for any *in-progress* jumps are saved, which is a bit ugly but I don't care so much because that state is effectively ephemeral (as it only applies to the *current* jump event). So not such a big deal if we drop it or change it again in the future.

I did consider exposing some functions to Lua to let Lua directly trigger sound effects and exposing events or callbacks for different stages of hyperjump so that Lua can trigger these sounds on its own... but (a) that would have been a larger change I think, and (b) I think of these sound effects as being on the 'simulation' side (they correspond with physical things happening in the game world) so my gut tells me it's better for C++ to trigger them and Lua to just specify which effect files to use.

Merge notes:
* You might want to squash the code changes. Best to keep commit 0b170d43e539247ca4695249aa224381ecf8ac7f (where the sound files are imported) on its own though.

-----

**Previously:**
Probably needs changing to avoid breaking saves. It breaks them because Lua equipment implementation details and values are embedded into the save files, so changes to Lua code or to the game's Lua API don't always interact nicely with the stuff that is still stored in people's saves.

Just looking for initial feedback on whether this makes any sense at all or if I should scrap it and find a different way.

Sound effects courtesy of @WKFO:
https://pioneerspacesim.net/forum/viewtopic.php?p=5332#p5332

Anyway, basic implementation:
- Lua specifies a set of sounds to use for a particular hyperjump, passed in with Ship:InitiateHyperjumpTo().
- C++ side keeps those values and plays the sounds at the relevant points.

Alternatively we could try to make Lua play the sounds itself (would need to expose mechanisms for Lua to play sound effects), but that has the disadvantage that there is still a bunch of Hyperspace logic on the C++ side and Lua might not always get triggered at the right point to play a sound.

Referring to the sounds as strings (the name of the file, minus the extension) and copying those around is kind of horrible but hyperjumps happen rarely enough that I doubt anyone will notice. Also this exacerbates the weird-stuff-embedded-in-save-files situation by saving the raw sound-effect names in the save if you save *during* a hyperjump countdown (I haven't actually tested this bit).


fluffyfreak:
The uglier, but way I was probably going to hack it in, was just to check what kind of hyperdrive the player has fitted when playing the sounds. 
Something nasty like `isMilitary ? "Hyperdrive_Jump_Military" : "Hyperdrive_Jump"` in each of the three locations. 
The main complication being that I don't know how to tell if a drive is military or not... maybe the refueling stuff can clarify that. Check what type of fuel it consumes.

Other than that I don't mind it breaking the save games, so long as you remember to also rev' the save game version number 👍

laarmen:
The engine (as in C++) shouldn't even *have* the notion of hyperdrive, much
less its class.

johnbartholomew:
Adding WIP tag because I think I need to find some better way to avoid weirdness with saves. Preferably I'd like to avoid breaking existing saves, but if that's not easy then at least I'd like to avoid embedding sound effect names into *new* saves.

johnbartholomew:
Made some changes to avoid breaking saves. Removing the WIP label now; I think this is ready for review and testing.

(I did some testing on my own, but more is always useful)

johnbartholomew:
Rebased on top of master.

impaktor:
Nitpick! first commit authorship could be set to WKFO. We do that sometimes, but not concistently.

PR #4204, opened 2017-12-18
Move sector view (F2) to lua

ecraven:
This moves the entire sector view UI to lua. Search is interactive (and finds all systems that match nearby, not just the "best").
Left upper corner is the "Search" window, right upper corner is the "Factions" toggle.
Please comment on things that should be improved ;)

f2

### DIFF
--- a/data/pigui/modules/map-sector-view.lua
+++ b/data/pigui/modules/map-sector-view.lua
@@ -1,0 +1,236 @@
+local Engine = import('Engine')
+local Game = import('Game')
+local ui = import('pigui/pigui.lua')
+local Vector = import('Vector')
+local Color = import('Color')
+local Lang = import("Lang")
+local lc = Lang.GetResource("core");
+local lui = Lang.GetResource("ui-core");
+local utils = import("utils")
+local Event = import("Event")
+
+local player = nil
+local colors = ui.theme.colors
+local icons = ui.theme.icons
+
+local mainButtonSize = Vector(32,32) * (ui.screenHeight / 1200)
+local mainButtonFramePadding = 3
+local function mainMenuButton(icon, selected, tooltip, color)
+	color = color or colors.white
+	return ui.coloredSelectedIconButton(icon, mainButtonSize, selected, mainButtonFramePadding, colors.buttonBlue, color, tooltip)
+end
+
+local function showSystemInfo(label, current_systempath, systempath, othersystempath)
+	if systempath then
+		local starsystem = systempath:GetStarSystem()
+		local clicked = false
+		ui.withID(label, function()
+								local jumpData = ""
+								if not current_systempath:IsSameSystem(systempath) then
+									local jumpStatus, distance, fuelRequired, duration = player:GetHyperspaceDetails(current_systempath, systempath)
+									jumpData = jumpStatus .. "  " .. string.format("%.2f", distance) .. lc.UNIT_LY .. "  " .. fuelRequired .. lc.UNIT_TONNES .. "  " .. ui.Format.Duration(duration, 2)
+								end
+								if ui.collapsingHeader(label .. ': ' .. starsystem.name .. '  ' .. jumpData .. " (" .. math.floor(systempath.sectorX) .. ", " .. math.floor(systempath.sectorY) .. ", " .. math.floor(systempath.sectorZ) .. ")", { "DefaultOpen" }) then
+									local stars = starsystem:GetStars()
+									for _,star in pairs(stars) do
+										if ui.selectable(star.name, star.path == systempath, {}) then
+											clicked = star.path
+										end
+									end
+									if clicked then
+										Engine.SetSectorMapSelected(clicked)
+										Engine.SectorMapGotoSystemPath(clicked)
+									end
+									local numstars = starsystem.numberOfStars
+									local numstarstext = ""
+									if numstars == 4 then
+										numstarstext = lc.QUADRUPLE_SYSTEM
+									elseif numstars == 3 then
+										numstarstext = lc.TRIPLE_SYSTEM
+									elseif numstars == 2 then
+										numstarstext = lc.BINARY_SYSTEM
+									else
+										numstarstext = starsystem.rootSystemBody.astroDescription
+									end
+									ui.text(numstarstext)
+									ui.text(starsystem.shortDescription)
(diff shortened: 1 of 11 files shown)


src/SectorView.cpp
```diff
@@ -144,43 +144,14 @@ void SectorView::InitObject()
 	m_lineVerts.reset(new Graphics::VertexArray(Graphics::ATTRIB_POSITION, 500));
 	m_secLineVerts.reset(new Graphics::VertexArray(Graphics::ATTRIB_POSITION, 500));
 	m_starVerts.reset(new Graphics::VertexArray(
-		Graphics::ATTRIB_POSITION | Graphics::ATTRIB_DIFFUSE | Graphics::ATTRIB_UV0, 500));
+																							Graphics::ATTRIB_POSITION | Graphics::ATTRIB_DIFFUSE | Graphics::ATTRIB_UV0, 500));
```
fluffyfreak: That's some very odd reformatting you've got there!

src/SectorView.cpp
```diff
@@ -37,8 +37,8 @@ static const float FAR_MAX       = 46.f;
 
 enum DetailSelection {
 	DETAILBOX_NONE    = 0
-,	DETAILBOX_INFO    = 1
-,	DETAILBOX_FACTION = 2
+	,	DETAILBOX_INFO    = 1
```
fluffyfreak: Odd indentation

src/SectorView.cpp
```diff
@@ -402,119 +211,14 @@ void SectorView::SaveToJson(Json::Value &jsonObj)
...
-}
 
 #define FFRAC(_x)	((_x)-floor(_x))
 
 void SectorView::Draw3D()
 {
 	PROFILE_SCOPED()
 
-	m_lineVerts->Clear();
+		m_lineVerts->Clear();
```
fluffyfreak: Ok last one I'll comment on but I think this is caused by the `PROFILE_SCOPED()` above which has no `;` at the end.

data/pigui/modules/map-sector-view.lua
```diff
@@ -0,0 +1,229 @@
...
+local Event = import("Event")
+
+local player = nil
+local colors = ui.theme.colors
+local icons = ui.theme.icons
+
+local mainButtonSize = Vector(32,32) * (ui.screenHeight / 1200)
+local mainButtonFramePadding = 3
+local function mainMenuButton(icon, selected, tooltip, color)
+	if color == nil then
```
laarmen: idiomatically : color = color or colors.white

data/pigui/modules/map-sector-view.lua
```diff
@@ -0,0 +1,229 @@
...
+end
+
+local function showSystemInfo(label, current_systempath, systempath, othersystempath)
+	if systempath then
+		local starsystem = systempath:GetStarSystem()
+		local clicked = false
+		ui.withID(label, function()
+								local jumpData = ""
+								if not current_systempath:IsSameSystem(systempath) then
+									local jumpStatus, distance, fuelRequired, duration = player:GetHyperspaceDetails(current_systempath, systempath)
```
laarmen: Note for the future : it might be time to revisit the API here. I put it together in haste back in the day, but I find the multiple return value dissatisfying.

data/pigui/modules/map-sector-view.lua
```diff
@@ -0,0 +1,229 @@
...
+			Engine.SetSectorMapAutomaticSystemSelection(automatic_system_selection)
+		end
+	end
+end
+
+local hyperspaceDetailsCache = {}
+
+local function getHyperspaceDetails(path)
+	local p = path.sectorX .. "/" .. path.sectorY .. "/" .. path.sectorZ .. "/" .. path.systemIndex
+	local it = hyperspaceDetailsCache[p]
```
laarmen: The cache should be invalidated in case of equipment change (and ship change, of course).

fluffyfreak:
Aside from the caching mechanism I'm happy for this to go in, it's much better than our existing implementation.

The-EG:
The old (current) sector map has a button for 'Swap selected and hyperjump target.' This new one doesn't seem to have that. Is that intended or am I missing something?

Having the swap button was nice when the hyperjump target is locked. You can click around to find a good in-between system to jump to (multiple short jumps is quicker than 1 long one), hit swap to put that into your hyperjump target and the original target would get 'saved' into the selected system (ie. I don't have to go find it again later).
With this, it seems I have to unlock the selection and select the system I want.

impaktor:
OK, just to summarize (for my self, at the very least): I guess three minor things have been raised, before a "perfect merge":

## missing feature
> The old (current) sector map has a button for 'Swap selected and hyperjump target.' This new one doesn't seem to have that. 

Valid objection. If it's easy to add it back, then that would be good.

## comments
laarmen objected in general to missing comments, and specifically:

> Please look into the hyperspace cache thing?

## indentation
I think in general lua files are indented with 4(?) char wide tabs, and some stuff looks odd above, as fluffy and me pointed out.

Oh, and Fröhliche Weihnachten!

nozmajner:
Nice!
I noticed that the Lock Hyperspace target while works, the UI doesn't reflect it with that checkbox.
All other shortcuts worked for me.

nozmajner:
Another thing I noticed: you can't search for sector coordinates. In the old map, you could do that either by separating the numbers by space or ,

The-EG:
Another minor issue: draw vertical lines gets turned on when it shouldn't be in some cases:

- Start Game (select start from London)
- Push F2, turn vertical lines off (if on)
- Exit game (not Pioneer, only get back to the main menu!)
- Start a new game again (select start from London)
- Push F2, vertical lines are shown but checkbox in settings is off.

ecraven:
OK, I've fixed the indentation problems.
Cache invalidation:
````
local function clearHyperspaceCache(ship)
	if ship and ship == player then
		hyperspaceDetailsCache = {}
	end
end

Event.Register("onShipEquipmentChanged", clearHyperspaceCache)
Event.Register("onShipTypeChanged", clearHyperspaceCache)
````
I've not added the button so far, if we still want it after @The-EG's changes, I'll add it then.
I've also not added the (1,2,3,4) syntax, I'll do that later, so we can finally merge this and the exceedingly nice route planner now!
Is there anything else I should address before we can merge this?

PR #4205, opened 2017-12-20
Save as gzip-format file

johnbartholomew:
Doesn't change save version. Should still be able to load the previous
format. Looks for a GZip header when loading the game to determine
whether it is new or old format.

This just adds a GZip header to the save output by default. This makes
it possible to decompress using 'gzip' on the command line instead of
needing a custom tool.

### DIFF
--- a/src/GZipFormat.cpp
+++ b/src/GZipFormat.cpp
@@ -1,0 +1,181 @@
+#include "GZipFormat.h"
+#include <string>
+#include <cstring>
+#include <cstdint>
+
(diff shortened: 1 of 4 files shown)


johnbartholomew:
Missing features:
* Good error messages.
* Efficiency (it shouldn't be too bad though).

joonicks:
gold to you.

fluffyfreak:
As always the code looks very good! Haven't had chance to test it yet.

PR #4207, opened 2017-12-21
Pickle Lua values to JSON

johnbartholomew:
*Updated 2018-09-02*

I've only done very superficial tests.

This replaces some of the custom bytestream serialization of Lua objects to generate JSON values instead. Long term I don't like this because it's likely to be very slow (both on save and on load), but short term it means that much more of the save file can be examined and interpreted by a person with a text editor (after decompressing the save), which is potentially valuable for debugging save problems.

When saving it just outputs the JSON version. When loading it will prefer to load the JSON version if available, but can load bytestream serialised data (ie, existing saves from before this PR) otherwise.

Lua->JSON structure:

- nil -> null
- bool -> bool
- integer -> integer
- string -> string (will definitely break for strings containing nul characters)
- userdata (LuaObject) -> `{ "userdata": { "cpp_class": "<...>", "inner": <...> } }`
- number (float) -> `{ "number": <... uses DoubleToStr ...> }`
- plain table -> `{ "ref": <unique ID number for the object>, "table": [ <table slots in array form; an array because JSON doesn't allow non-string keys for objects> ] }`
- Lua "class" (table with a metatable & Lua serialiser functions) -> `{ "ref": <unique ID>, "table": [ <table data> ], "lua_class": "<...>" }`
- Lua back-reference (reference to table that has been serialised previously; required to support circular references in tables) -> `{ "ref": <unique ID> }`

LuaRef structure: `{ "lua_ref_json": <... Lua value structured as above ...> }`

LuaObject structure: Depends on the type of object (but see "userdata" above).

----

**Previously:**

Current problems/remaining todo:
- Doesn't bump save version but won't be able to load files generated from previous versions. This could be fixed pretty easily by branching to the old or new unpickle code based based on what JSON value type is being unpickled. Or just bump save version and break all old saves again for simplicity.
- LuaObject values (e.g., various C++ objects that are exposed to Lua) are still pickled to a mostly-opaque bytestream. The reverse (LuaRef objects, which are Lua objects that are exposed to C++) are pickled to JSON.

Would work best combined with #4205 so that a custom extraction program isn't required to decompress the save files.

### DIFF
--- a/src/LuaObject.h
+++ b/src/LuaObject.h
@@ -73,18 +73,29 @@ struct SerializerPair {
 	typedef std::string (*Serializer)(LuaWrappable *o);
 	typedef bool (*Deserializer)(const char *stream, const char **next);
 
+	typedef void (*ToJson)(Json::Value &out, LuaWrappable *o);
+	typedef bool (*FromJson)(const Json::Value &obj);
+
 	SerializerPair() :
 		serialize(nullptr),
-		deserialize(nullptr)
+		deserialize(nullptr),
+		to_json(nullptr),
+		from_json(nullptr)
 	{}
 
-	SerializerPair(Serializer _serialize, Deserializer _deserialize) :
-		serialize(_serialize),
-		deserialize(_deserialize)
+	SerializerPair(
+			Serializer serialize_, Deserializer deserialize_,
+			ToJson to_json_, FromJson from_json_):
+		serialize(serialize_),
+		deserialize(deserialize_),
+		to_json(to_json_),
+		from_json(from_json_)
 	{}
 
 	Serializer serialize;
 	Deserializer deserialize;
+	ToJson to_json;
+	FromJson from_json;
 };
 
 
@@ -147,6 +158,9 @@ protected:
 
 	std::string Serialize();
 	static bool Deserialize(const char *stream, const char **next);
+
+	void ToJson(Json::Value &out);
+	static bool FromJson(const Json::Value &obj);
 
     // allocate n bytes from Lua memory and leave it an associated userdata on
     // the stack. this is a wrapper around lua_newuserdata
--- a/src/LuaSerializer.cpp
+++ b/src/LuaSerializer.cpp
@@ -518,6 +518,8 @@ void LuaSerializer::unpickle_json(lua_State *l, const Json::Value &value)
 			break;
 		case Json::objectValue:
 			if (value.isMember("number")) {
+				lua_pushnumber(l, StrToDouble(value["number"].asString()));
+				LUA_DEBUG_CHECK(l, 1);
 			} else if (value.isMember("userdata")) {
 				if (!LuaObjectBase::FromJson(value["userdata"])) { throw SavedGameCorruptException(); }
 				LUA_DEBUG_CHECK(l, 1);
--- a/src/LuaSerializer.cpp
+++ b/src/LuaSerializer.cpp
@@ -451,7 +451,8 @@ void LuaSerializer::pickle_json(lua_State *l, int to_serialize, Json::Value &out
 			if (!o)
 				Error("Lua serializer '%s' tried to serialize an invalid '%s' object", key.c_str(), lo->GetType());
 
-			out["udata"] = lo->Serialize();
+			const std::string pickled = lo->Serialize();
+			BinStrToJson(out, pickled, "udata");
 			break;
 		}
 
@@ -486,14 +487,17 @@ void LuaSerializer::unpickle_json(lua_State *l, const Json::Value &value)
 		case Json::uintValue: // fallthrough
 		case Json::realValue:
 			lua_pushnumber(l, value.asDouble());
+			LUA_DEBUG_CHECK(l, 1);
 			break;
 		case Json::stringValue:
 			// FIXME: Should do something to make sure we can unpickle strings that include null bytes.
 			// However I'm not sure that the JSON library actually supports strings containing nulls which would make it moot.
 			lua_pushstring(l, value.asCString());
+			LUA_DEBUG_CHECK(l, 1);
 			break;
 		case Json::booleanValue:
 			lua_pushboolean(l, value.asBool());
+			LUA_DEBUG_CHECK(l, 1);
 			break;
 		case Json::arrayValue:
 			// Pickle doesn't emit array type values.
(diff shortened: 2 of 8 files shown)


src/LuaSerializer.cpp
```diff
@@ -328,6 +328,261 @@ const char *LuaSerializer::unpickle(lua_State *l, const char *pos)
...
+			}
+		}
+	}
+
+	switch (lua_type(l, idx)) {
+		case LUA_TNIL:
+			break;
+
+		case LUA_TNUMBER: {
+			out = Json::Value(lua_tonumber(l, idx));
```
johnbartholomew: Probably need to change this to something that doesn't lose precision. Except if that something generates a JSON string then there will also need to be a way to differentiate in the JSON output between a Lua string and a Lua number... which is a bit of a PITA.

src/LuaSerializer.cpp
```diff
@@ -328,6 +328,261 @@ const char *LuaSerializer::unpickle(lua_State *l, const char *pos)
...
+		}
+
+		case LUA_TUSERDATA: {
+			LuaObjectBase *lo = static_cast<LuaObjectBase*>(lua_touserdata(l, idx));
+			void *o = lo->GetObject();
+			if (!o)
+				Error("Lua serializer '%s' tried to serialize an invalid '%s' object", key.c_str(), lo->GetType());
+
+			const std::string pickled = lo->Serialize();
+			BinStrToJson(out, pickled, "udata");
```
johnbartholomew: Might need a LuaObject::SerializeToJSON method (and corresponding DeserializeFromJSON).

laarmen:
How about serializing to an object with __class="number" and the data in
the data field?

johnbartholomew:
Update:
- (Rebased).
- Implemented serialising userdata (LuaObject wrapped things) as JSON.
- Should now be able to load both bytestream serialised data and the new JSON data (JSON data preferred if the save contains it).
- Floats now serialised using `DoubleToStr` to avoid potential precision loss problems or problems with NaN values.

Currently when saving it *only* writes the JSON format; it could be pretty easily updated to write both JSON and the old bytestream format so that previous builds can load the new files. Not sure how valuable that is (perhaps useful for testing)

impaktor:
How was it, did this trigger strangeness in Tradeship, that you wanted to investigate before this is merged?

johnbartholomew:
I haven't been able to reproduce the thing I saw with repeated alerts in the comms log. I still get repeated messages to stdout from TradeShips.lua (from some saves at least), but that's not a new bug.

At this point I'm not investigating anything that I think is caused by this PR.

src/LuaObject.h
```diff
@@ -74,18 +74,29 @@ struct SerializerPair {
 	typedef std::string (*Serializer)(LuaWrappable *o);
 	typedef bool (*Deserializer)(const char *stream, const char **next);
 
+	typedef void (*ToJson)(Json::Value &out, LuaWrappable *o);
+	typedef bool (*FromJson)(const Json::Value &obj);
+
 	SerializerPair() :
 		serialize(nullptr),
-		deserialize(nullptr)
+		deserialize(nullptr),
```
laarmen: I'm confused. Why do we need to have 4 functions now?

johnbartholomew: This PR does not remove the current serialisation (read/unserialize) codepath because it still supports reading files that are written with that format.

I haven't removed the write path for it because I wanted it to be easy to dual-write both old & new formats which would make new saves backwards compatible. However I haven't actually implemented dual-write so if we don't want that then the bytestream write path can be deleted.

johnbartholomew: Also for more context, I do expect to break old saves "soonish" if I proceed with (a) [serializing to CBOR instead of JSON](https://pioneerspacesim.net/forum/viewtopic.php?f=7&p=5361#p5361), and (b) changing Lua side structures like Equipment to stop embedding so much "static" game data into saves (as you suggested in the [forum thread](https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=531))

*But* I don't want to prematurely break saves before I'm confident of completing at least the JSON->CBOR change.

fluffyfreak:
@johnbartholomew is this ready for review now or is there more you would like to do on it first?

johnbartholomew:
It's ready for review (or I've forgotten anything else I was going to do :-P)

I would like opinions on whether it's worth dual-writing both the JSON and previous bytestream format for a while (e.g., until the next savebump) to try to support some backwards compatibility. That's not implemented in this PR but could be added quite easily I believe.

impaktor:
> to try to support some backwards compatibility.

not worth it, is my own highly personal opinion. It's not like we have swarms of players keeping up with master, (that we know of).

fluffyfreak:
Ok so this needs testing and then if all good merging.

With a save game bump?

impaktor:
I'm happy to see this merged, assuming it's in merge ready state?

fluffyfreak:
@impaktor sadly this doesn't work for me right now, it crashes when trying to unpause a game after loading.

Something to do with pasing pigui an invalid number

fluffyfreak:
Sorry for the brief PR mess-up I made, thanks to @impaktor for helping me get it fixed.

Pushed a few fixes for a couple of bugs that this exposed.

impaktor:
> Pushed a few fixes for a couple of bugs that this exposed.

Would we not want them in master ASAP, regardless of this PR?

fluffyfreak:
Probably, but we can cherry pick them if we decide not to take this PR

johnbartholomew:
I've cherry-picked fluffy's fixes (PR #4430) and rebased to master.
Now this needs testing again...

johnbartholomew:
Well, it does not immediately crash. There is still a high chance of bugs, but we may have to let them be found after merge unless someone else has time to test this PR more thoroughly on its own.

fluffyfreak:
Added the savegame bump label as I don't seem to be able to open any older save games

johnbartholomew:
Loading old saves (well, old as in master, just before this PR) is definitely supposed to *work*. It might not be *important* for it to work, but let me take a look at it anyway.

(Unfortunately I don't think I'll be able to do that until the weekend)

fluffyfreak:
@johnbartholomew merged mate!

johnbartholomew:
Thanks! Sorry I didn't get back to this to work out what was wrong with it
:-/

fluffyfreak:
No worries, Loading and saving worked in testing after your rebase. There's been other fixes in between as well, maybe it was unrelated.

impaktor:
Nice! Almost a year being open, but made it in.

fluffyfreak:
Yeah I'd like us to do better than that, get a PR and get it done asap. Author able of course.

PR #4209, opened 2017-12-22
Library version information logging to aid debugging.

fluffyfreak:
Extra version information logging.

Output the versions of some of the libraries that are used to build the game.

Useful when the game is built on Linux/OSX where libraries other than those in `pioneer-thirdparty` might be used instead and cause compatibility problems.


src/versioningInfo.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include <curl/curlver.h>
+#endif
+
+void OutputVersioningInfo()
+{
+	SDL_version ver;
+	SDL_GetVersion(&ver);
+	Output("\n--------------------\n");
+	Output("SDL Version %d.%d.%d\n", ver.major, ver.minor, ver.patch);
+	Output("SDL_image ver: %d.%d.%d\n", SDL_IMAGE_MAJOR_VERSION, SDL_IMAGE_MINOR_VERSION, SDL_IMAGE_PATCHLEVEL);
```
johnbartholomew: There's also the IMG_Linked_Version() function, which might be preferable.
SDL_IMAGE_*_VERSION is the version we've compiled against, IMG_Linked_Version should be the version being used at runtime.

src/versioningInfo.cpp
```diff
@@ -0,0 +1,31 @@
...
+	Output("Assimp ver: %u.%u.%u\n", aiGetVersionMajor(), aiGetVersionMinor(), aiGetVersionRevision());
+	Output("FreeType ver: %d.%d.%d\n", FREETYPE_MAJOR, FREETYPE_MINOR, FREETYPE_PATCH);
```
johnbartholomew: There's FT_Library_Version, but it takes an FT_Library pointer as its first parameter so might be a pain to use here (but it's basically the same deal as IMG_Linked_Version).

imgui doesn't use freetype anyway (at least I think it doesn't) so I guess this will be less relevant in the future.

src/versioningInfo.cpp
fluffyfreak: Ah interesting on both of those. 
I think it might be good to keep the compiled version and display what you're actually using at runtime so I'll add that.
It might highlight mismatches and other issues.

johnbartholomew: You could do the same for SDL itself:

    Output("SDL Version (build) %d.%d.%d\n", SDL_MAJOR_VERSION, SDL_MINOR_VERSION, SDL_PATCHLEVEL);
    Output("SDL Version (dynamic) %d.%d.%d\n", ver.major, ver.minor, ver.patch);

fluffyfreak: I was so tired the other day that I was actually trying to do this and didn't understand that I was already getting the dynamic library version 🙄 so tired and so glad that it's the holidays now!

impaktor:
Can be merged, if not for conflict in visual studio-stuff.

...and Merry Christmas!

PR #4214, opened 2017-12-24
systeminfo temperature misconceptions

joonicks:
m_averageTemp misconceptions taken care of. earth +15C is not an ice world, etc.

some source formatting for readability + comment

### DIFF
--- a/src/galaxy/StarSystem.cpp
+++ b/src/galaxy/StarSystem.cpp
@@ -246,7 +246,8 @@ SystemBody::BodySuperType SystemBody::GetSuperType() const
 std::string SystemBody::GetAstroDescription() const
 {
 	PROFILE_SCOPED()
-	switch (m_type) {
+	switch (m_type)
+	{
 	case TYPE_BROWN_DWARF: return Lang::BROWN_DWARF;
 	case TYPE_WHITE_DWARF: return Lang::WHITE_DWARF;
 	case TYPE_STAR_M: return Lang::STAR_M;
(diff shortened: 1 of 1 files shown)


fluffyfreak:
@joonicks Because of all of the formatting changes it's really hard to see what's going on in here.

Could you point out, using comments, where the logic changes are and not formatting changes please?

If not I'll just try to work it out later, thanks.

impaktor:
> Because of all of the formatting changes it's really hard to see what's going on in here.

I had the same problem the other day when looking at this. If formating changes are needed, having them in a separate commit helps review, and the actual feature / change in another commit.

joonicks:
logic changes is just changing the fixed(250) to 273. temp is stored in K. to get C (freezing point of water) K needs to be above 273. plus comment. the rest is just formatting.

fluffyfreak:
OK, thanks @joonicks

impaktor:
Does the re-formating make sense? 

I know we're inconsitent, but usually, functions have opening and closing curly brackets on separate lines, but _not_ `if`, `else`, `switch` statements. 

And yes, you can certainly find plenty of exceptions but I'm not sure all this re-formating is making anything better. Rather it's pushing for higher entropy in our style.

joonicks:
sorry imp, but I adhere to the one true style of formatting curly braces for optimal clarity and readability.

fluffyfreak:
It is very verbose, but it's the same as they use at Epic for Unreal Engine. All curly braces on new lines etc.
I have to say that it is much clearer and easier to read.

PR #4217, opened 2017-12-26
Add VS2017 Build files

The-EG:
Add build files for Visual Studio 2017.
This was copied from VS2015 and converted to VS2017. The Windows SDK version was not changed.

Currently it's quite difficult (impossible?) to download an installer for VS2015 from Microsoft without having some sort of subscription. 
When I was trying to get a build environment for Pioneer setup the first time, I clicked around for about 15-20 minutes and couldn't find an actual download for VS2015, even after creating a free account.
Luckily I realized I still had an old installer for it archived somewhere, but I think the project should support what's available.


fluffyfreak:
@The-EG this is a good point, I had held off doing it because you still need VS2015 installed anyway. Which you seem to have so wouldn't have noticed this, because it must build and link using the VS2015 compiler backend.

This is because some of the libraries are built using VS2015 and are not compatible with the VS2017 linker. I keep meaning to fix this up but... it's a pain and I got distracted so maybe I should just get it done asap.

fluffyfreak:
Also I should say that I am glad to see someone else getting involved with these files. It's exactly how I started with Pioneer many years ago 👍

The-EG:
@fluffyfreak ok, that makes sense. Is it just a matter of the pioneer-thirdparty libs? I see various things in win32/.../.../vs2015 there. I could take a swing at that, too.

fluffyfreak:
@The-EG yes that's the one, there needs to be a VS2017 folder with basically everything rebuilt using/for 2017. It's a bit of a slog to get it up but if you're up for it then I'm happy to comment on stuff.

The-EG:
@fluffyfreak Now that I've had some more time to think about this I think the problem is more limited than it seemed to be at first.

I went back and checked, and I was indeed compiling and linking within VS2017. I verified that it wasn't falling back to the VS2015 toolchain at all. 

I thought it was strange that it would work at all, but then I remembered that binary incompatibilities between compiler versions revolve around ABI differences, which is usually only a problem with C++ (C ABI is very well defined). 
Version differences shouldn't matter for pure C libraries and the only libraries that are C++ that I found are Assimp and libsigc++. 

Now, as to linking to those working: I found [this](https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017).

If I read that right, this falls under caveat 2, so it's still not 'guaranteed,' but at least it makes sense that it actually worked.

It's probably still a good idea to recompile those (I've already got Assimp going!), but if the rest are pure C it shouldn't be needed for them.

fluffyfreak:
There's a sigcpp project within the pioneer solution... I honestly don't remember doing that but I did apparently 😁 it uses relative pathing into `pioneer-thirdparty` and is statically linked.

So yes it looks like we might just need Assimp!

The-EG:
I created a PR for pioneer-thirdparty: [here](https://github.com/pioneerspacesim/pioneer-thirdparty/pull/24).

If that looks good, I'll modify this to point to the new VS2017 folders instead.

fluffyfreak:
Nice work, I've merged that PR now @The-EG

The-EG:
@fluffyfreak Done. I also removed PersistSystemData.h, since I saw a commit for that on the 2015 side.

fluffyfreak:
Wow, I knew vs2017 was faster to build but it's a *lot* faster to build!

fluffyfreak:
Merged, great work @The-EG and thankyou!

PR #4218, opened 2017-12-26
AlertState cache cleared on load

joonicks:
Loading a savegame would leave alertstate cache values undefined possibly triggering an alert or postponing further checks hundreds of years into the future.

This fixes it, clearing the cache on load and adding a sanity check in UpdateAlertState()

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -124,6 +124,11 @@ void Ship::LoadFromJson(const Json::Value &jsonObj, Space *space)
 	m_launchLockTimeout = StrToFloat(shipObj["launch_lock_timeout"].asString());
 	m_testLanded = shipObj["test_landed"].asBool();
 	m_flightState = static_cast<FlightState>(shipObj["flight_state"].asInt());
+	
+	m_lastAlertUpdate = 0.0;	// alertstate check cache timer
+	m_shipNear = false;			// alertstate check cache value
+	m_shipFiring = false;		// alertstate check cache value
+
 	m_alertState = static_cast<AlertState>(shipObj["alert_state"].asInt());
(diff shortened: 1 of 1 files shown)


fluffyfreak:
nice one!

src/Ship.cpp
```diff
@@ -944,6 +954,16 @@ void Ship::UpdateAlertState()
 	}
 
 	bool ship_is_near = false, ship_is_firing = false;
+
+	// sanity check: m_lastAlertUpdate should not be in the future.
+	// reset and re-check if it is.
+	if (m_lastAlertUpdate > Pi::game->GetTime() + 2)
```
laarmen: Out of curiosity, where does the +2 come from?

src/Ship.cpp
```diff
@@ -213,6 +218,8 @@ void Ship::InitMaterials()
 
 void Ship::Init()
 {
+	Output("Running Ship::Init()\n");
```
laarmen: I'm assuming this is a leftover debug printout? In any case, I'd rather not have it, as it clutters the output without adding much information.

src/Ship.cpp
joonicks: arbitrary. could be +1 but. meh.

fluffyfreak: Shouldn't even need to plus anything since it's using `>`

fluffyfreak:
Ok, deal with those two minor comments please and then I think it's good to merge!
Another good fix @joonicks :)

impaktor:
For future reference, that last commit should ideally have been squashed with the previous.

At moments like these:
1. do the change
2. `git add` the change
3. `git commit --amend`
4. `git push -f` to your branch.

We try to have meaningful commit messages. "pimplesqueezer" doesn't say much...

But good that the requested changes were fixed fast, thanks.

Issue #4216, opened 2017-12-26
Atmosphere temperature calculation bug (Planet.cpp::GetAtmosphericState)


joonicks:
### Observed behaviour
Fuel scooping a small gas giant fails. After lots of digging I find that the pressure is too low and the pressure is calculated on, amongst other things, temperature. As you rise through the atmosphere, temperature is supposed to drop. This calculation is done in Planet.cpp::GetAtmosphericState. But the heavy math apparently makes it difficult to get everything right.

With current code, temperature actually rises as you go up. Because a sign that should be - is a +.

Changing the sign fixes the rising temperature and the fuel scooping. However, that might not be all. The calculated air pressure might also be wrong. I just dont know enough about the formula used to see if its correct.

Both Planet::InitParams and Planet::GetAtmosphericState contains potentially buggy code. Preferably a person who knows the formula should take another look at it.

All I can do is to quickfix and throw in a few extra comments.

pioneer-20171001-win32

The-EG:
Fixed, or at least improved by #4946

PR #4219, opened 2017-12-26
atmospheric temperature fix

joonicks:
temperature affects density. if temperature rises with altitude, density drops much more than it should, making fuel scooping harder or impossible.

comments plus a single char fix, a + to a -

### DIFF
--- a/src/Planet.cpp
+++ b/src/Planet.cpp
@@ -121,6 +121,11 @@ void Planet::GetAtmosphericState(double dist, double *outPressure, double *outDe
 	// Note: some code duplicated in InitParams(). Check if changing.
 	if (dist >= m_atmosphereRadius) {*outDensity = 0.0; *outPressure = 0.0; return;}
 
+	// molar mass of hydrogen gas = 2.016g/mol
+	// molar mass of helium gas = 4.002g/mol
+	// molar mass of earth air = 28.97g/mol
+	// molar mass of hydrogen gas(85%) and helium(15%) mix = 2.3139g/mol
+
 	double surfaceDensity;
 	double specificHeatCp;
 	double gasMolarMass;
@@ -139,8 +144,8 @@ void Planet::GetAtmosphericState(double dist, double *outPressure, double *outDe
 	// lapse rate http://en.wikipedia.org/wiki/Adiabatic_lapse_rate#Dry_adiabatic_lapse_rate
 	// the wet adiabatic rate can be used when cloud layers are incorporated
 	// fairly accurate in the troposphere
-	const double lapseRate_L = -m_surfaceGravity_g/specificHeatCp; // negative deg/m
-
+	// m_surfaceGravity_g is calculated and stored as a negative number
+	const double lapseRate_L = -m_surfaceGravity_g/specificHeatCp; // negative deg/m << wrong? -(-)/+ = +, should it be -abs(G)/hcp?
 	const double height_h = (dist-sbody->GetRadius()); // height in m
 	const double surfaceTemperature_T0 = sbody->GetAverageTemp(); //K
 
@@ -156,10 +161,11 @@ void Planet::GetAtmosphericState(double dist, double *outPressure, double *outDe
 	if (height_h < 0.0) { *outPressure = surfaceP_p0; *outDensity = surfaceDensity*gasMolarMass; return; }
 
 	//*outPressure = p0*(1-l*h/T0)^(g*M/(R*L);
+	// WARNING! lapserate is a positive number, not negative as commented
 	*outPressure = surfaceP_p0*pow((1-lapseRate_L*height_h/surfaceTemperature_T0),(-m_surfaceGravity_g*gasMolarMass/(GAS_CONSTANT*lapseRate_L)));// in ATM since p0 was in ATM
 	//                                                                               ^^g used is abs(g)
 	// temperature at height
-	double temp = surfaceTemperature_T0+lapseRate_L*height_h;
+	double temp = surfaceTemperature_T0-lapseRate_L*height_h; // << lapserate is a positive number and needs to be subtracted for temperature to drop the higher you get.
 
 	*outDensity = (*outPressure/(PA_2_ATMOS*GAS_CONSTANT*temp))*gasMolarMass;
 }


fluffyfreak:
Edited just so it will auto-close the related bug report :)

joonicks:
its possible that the positive/negative bug also causes abnormally thin atmospheres. I just scooped a gas giant with 57,000km radius and only ~30km atmosphere. that doesnt seem right.

fluffyfreak:
@joonicks gas giants in pioneer are a bit shitty. Fine to look at, and a good implementation for technology/ability at the time, but they aren't really gas giants like you'd expect. They're rendered as a solid shell instead of layers of clouds. There's no real volume to them. So everything is a bit of a hack.

fluffyfreak:
Ok so in light of #4222 does it make sense to merge this or has it revealed that there needs to be more work?

I mean I looked at all those calculations and I'd hate to guess what order all of the maths is actually being applied in. Someone really needed to bracket the hell out of those equations just for reading and sanity checking.

joonicks:
this fixes #4216 but #4222 is a massive math problem thats way over my head.

src/Planet.cpp
```diff
@@ -139,8 +144,8 @@ void Planet::GetAtmosphericState(double dist, double *outPressure, double *outDe
 	// lapse rate http://en.wikipedia.org/wiki/Adiabatic_lapse_rate#Dry_adiabatic_lapse_rate
 	// the wet adiabatic rate can be used when cloud layers are incorporated
 	// fairly accurate in the troposphere
-	const double lapseRate_L = -m_surfaceGravity_g/specificHeatCp; // negative deg/m
-
+	// m_surfaceGravity_g is calculated and stored as a negative number
+	const double lapseRate_L = -m_surfaceGravity_g/specificHeatCp; // negative deg/m << wrong? -(-)/+ = +, should it be -abs(G)/hcp?
```
impaktor: Looks like github is warning that this line isn't indented correct? (we use tab, 4 char wide).

impaktor:
I'm not very comfortable as it looks to me like no one actually understands what the code should be, something that is emphasized further by the code comments.

From #4216 

> As you rise through the atmosphere, temperature is supposed to drop. 

I'm not sure it's that simple at all, as you rise density and preassure drops, giving higher speed to the molecules. For Earth temperature drops only the initial 10 km, then increases again up to 50 km:
atmo

(Ideally, I'd like to see graphs of how this changes our temperature  before / after this change, as that is how I understand things the best).

Or does this magically make everything better for all planet and atmosphere types?
Or how do we notice this change?

joonicks:
pioneer earth atmosphere stops at 25km. so from sea level to 25km, the temperature should fall.
without this fix, it rises. sealevel = 15C, at 25km = 300C.

I had to apply this fix to scoop a hot jupiter.

joonicks:
stratosphere - "The increase of temperature with altitude is a result of the absorption of the Sun's ultraviolet radiation by the ozone layer."
ozone, typically only present on planets with life. as in 1/100th of any planet in pioneer.
stratosphere ends at ~50km.
mesosphere is considered part of the atmosphere, on earth extends to ~80km
nowhere in the (earth) atmosphere is it hotter than sea level.

venus' atmosphere, temperature steadily declines up to 100km where pressure is 0.01, after that it slowly rises but at 200km its still only 1/5th of ground level temperature.

wikip doesnt give any info about mars' atmosphere except "upper atmosphere is hot"

jupiter lacks mesosphere, in troposphere temperature drops, in thermosphere (usually not considered part of the "atmosphere") temperature increases again

saturn no mention of temperature

joonicks:
I think I have found the formulas used. give me a bit.

fluffyfreak:
I am struggling to test this one. Think that I will need to add some more logging and info display, maybe you'll get your graphs @impaktor 

I've tried going over the original code as well, it's horrible.

joonicks:
youre not ready to take my word for it?

jaj22:
There's currently no code change in this commit and I'm pretty sure a +/- flip would be a miscorrection. As the surface gravity is stored as a negative number, lapseRate_L comes out as positive and the formulas are used correctly. If anything's wrong it's the comment, "negative deg/M".

There's a separate issue that the formulas assume a fully convective (non-layered) atmosphere so the temperature falloff is constant. I'm not sure how much difference this makes given that there's a cutoff at 0.001 surface pressure anyway.

For anyone trying to understand this code, lapseRate_L is the dry adiabatic lapse rate described in https://en.wikipedia.org/wiki/Lapse_rate. The final atmosphere height is calculated by inverting the height to pressure formula here at 0.001 surface pressure: https://en.wikipedia.org/wiki/Density_of_air#Altitude.

fluffyfreak:
> youre not ready to take my word for it?

It's not that. 
As you've pointed out there seem to be some problems and it's complex maths so I'm trying to understand what it does and what it is supposed to be doing.

Changing the sign to a +/- doesn't seem to make much difference on planets like I would have expected and Gas Giants I'm struggling to scoop at any of them anyway.

joonicks:
Ive read up on the various gas pages on wikipedia while looking in to this.

lapserate is a change in temperature as altitude rises, its supposed to be a negative value.
I have seen an example in wikipedia with a value of -0.0065 for lapserate.

the formula on that page was for calculating the height z where pressure equals P0.
it might be the formula they are trying to use in the pioneer code, but its too messy to be able to tell for certain.

lapserate = g/Cp where g is gravity (a positive number) and Cp is the specific heat capacity of the gas

https://en.wikipedia.org/wiki/Vertical_pressure_variation#cite_note-9

thats the formula I tried.

but with that formula, you can never have an atmosphere thicker than -(surfacetemperature/lapserate), or in the case of earth -(288/-0.009801649) = 29382meters. the formula will never give an answer above that value.

if thats the formula used in pioneer, it would show as cool planets having a very thin atmosphere while hot planets have a very thick atmosphere. ive seen indications that that is the case.

the vertical pressure variation formula can therefor not be used to determine the thickness of any atmosphere.

also, if you want this:
// temperature at height
double temp = surfaceTemperature_T0+lapseRate_L*height_h;

to produce temperatures that drop the higher you get, lapserate must be a negative number, which it is not right now. right now, temperature rises until it peaks at the top of the atmosphere that is twice the surface temperature.

yes sure in reality atmospheric temperature varies, but on earth the hottest place is the surface. and it sure as hell is not 300C at 25km.

joonicks:
jaj, the article https://en.wikipedia.org/wiki/Density_of_air#Altitude specifically states:
Temperature at altitude h {\displaystyle h} h meters above sea level is approximated by the following formula (**only valid inside the troposphere**):

the troposphere is only the first 10-15km on earth.

joonicks:
untitled-1 copy

jaj22:
> the formula on that page was for calculating the height z where pressure equals P0.
it might be the formula they are trying to use in the pioneer code, but its too messy to be able to tell for certain.

It's a few lines of code. How can you not tell for certain? As far as I can tell it's a direct copy of the formula from the wiki, correctly transposed and with minus signs in the correct places. As there's no actual code change in the current commit, I'm guessing that you've flipped the minus sign yourself, changed it back and decided that it was wrong in the first place.

There's no dispute on the troposphere point, but you're talking about a difference of maybe 0.003 vs 0.001 atmospheres at 25km, which isn't going to make much noticeable difference. Also unless someone wants to develop a superior functional model of atmosphere pressure that applies to general extraterrestrial planets, it's an academic point.

src/Planet.cpp
```diff
@@ -156,10 +161,11 @@ void Planet::GetAtmosphericState(double dist, double *outPressure, double *outDe
 	if (height_h < 0.0) { *outPressure = surfaceP_p0; *outDensity = surfaceDensity*gasMolarMass; return; }
 
 	//*outPressure = p0*(1-l*h/T0)^(g*M/(R*L);
+	// WARNING! lapserate is a positive number, not negative as commented
 	*outPressure = surfaceP_p0*pow((1-lapseRate_L*height_h/surfaceTemperature_T0),(-m_surfaceGravity_g*gasMolarMass/(GAS_CONSTANT*lapseRate_L)));// in ATM since p0 was in ATM
 	//                                                                               ^^g used is abs(g)
 	// temperature at height
-	double temp = surfaceTemperature_T0+lapseRate_L*height_h;
+	double temp = surfaceTemperature_T0-lapseRate_L*height_h; // << lapserate is a positive number and needs to be subtracted for temperature to drop the higher you get.
```
fluffyfreak: @jaj22 flipped sign is here

joonicks:
this is not intended to fix the whole issue with crappy formulas. simply a quick fix to flip the temperature gradient to its correct way.

joonicks:
Atmosphere at 8000 meters, temperature = -63.4C, with + temperature (unfixed) = 93.4C

top of mount everest: boiling alive.

jaj22:
The line 168 fix is definitely correct. I checked all the other formulas before noticing which line was changed. Just needs the tab spacing fixed on line 148 and preferably some comment cleanup. Should make it clear on line 68 and 148 that lapseRate_L is a positive value.

fluffyfreak:
Ok a lot changed here since I last looked at it

joonicks:
I tried to clean it up best I could, output is exactly the same as old messy function, but a bit optimized and commented up

impaktor:
I don't feel comfortable enough to merge this, as I get an uneasy feeling by it (and I don't have time to get under the hood) . But if anyone else understands it, then feel free to merge.

joonicks:
@impaktor do you read only the diffs, or do you watch the end result as well?

PR #4220, opened 2017-12-26
tradeships.lua unexplored crashfix

joonicks:
game crashes in tradeships.lua when entering an unexplored solar system

this = fix

### DIFF
--- a/data/modules/TradeShips.lua
+++ b/data/modules/TradeShips.lua
@@ -564,6 +564,11 @@ local cleanTradeShipsTable = function ()
 end
 
 local onEnterSystem = function (ship)
+	-- dont crash when entering unexplored systems
+	if Game.system.explored == false then
+		return
+	end
+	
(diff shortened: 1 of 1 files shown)


impaktor:
Good find!

Out of curiosity, do we know why exactly a crash is triggered? What function/line is it that causes the crash?

joonicks:
in getSystemAndJump, line 310.

but afaik, unexplored system = no tradeships, so mite as well skip the whole thing.

laarmen:
I don't remember if this is actually possible, but couldn't a tradeship use an unexplored solar system as a waypoint towards a distant, explored system?

joonicks:
if any ship entered an unexplored system for any reason, it wouldnt be unexplored anymore.

sure some classes of ships like pirates or military wouldnt want to report their movements, but tradeships would. especially if a system was safe to travel, theyd report it = no more unexplored.

fluffyfreak:
I'm happy with it, going to merge

PR #4224, opened 2017-12-28
Query the maximum (MS)AA samples

### LINKED ISSUE #4192: Show graphic options, that are supported by hardware.
For example I set multisampling to x16 and it didn't work. In console output Iv'e found that it couldn't find visual and retried with multisampling disabled.

In glxinfo output I've found that multisampling is supported up to x8. After changing setting to this value, it works fine.

So I think that x16 should be grayed out, or somehow it should be reported in settings window, that it won't work.

fluffyfreak:
Add the ability to query the maximum number of (MS)AA samples the GPU can support.

fixes #4192

@impaktor thanks!

### DIFF
--- a/data/ui/Settings.lua
+++ b/data/ui/Settings.lua
@@ -69,9 +69,19 @@ ui.templates.Settings = function (args)
 		end
 		local modeDropDown = optionDropDown(GetVideoMode, SetVideoMode, l.VIDEO_RESOLUTION, videoModeLabels, videoModeLabels)
 
-		local aaLabels = { l.OFF, "x2", "x4", "x8", "x16" }
-		local aaModes = { 0, 2, 4, 8, 16 }
-		local aaDropDown = optionDropDown(Engine.GetMultisampling, Engine.SetMultisampling, l.MULTISAMPLING, aaLabels, aaModes)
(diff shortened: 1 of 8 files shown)


data/ui/Settings.lua
```diff
@@ -69,9 +69,20 @@ ui.templates.Settings = function (args)
...
-		local aaDropDown = optionDropDown(Engine.GetMultisampling, Engine.SetMultisampling, l.MULTISAMPLING, aaLabels, aaModes)
+		local maxSamples = Engine.GetMaximumAASamples()
+		local aaLabels = { l.OFF, "x2", "x4", "x8", "x16", "x32", "x64" }
+		local aaModes = { 0, 2, 4, 8, 16, 32, 64 } -- I honestly had no idea you could get these values before today!
+		local aaTempLabels = {}
+		local aaTempModes = {}
+		-- many thanks/blame to impaktor for the following loop code
+		for i,value in ipairs(aaModes) do
+			if value <= maxSamples then
+				print(i,value, aaLabels[i])
```
impaktor: You want the print out? 
`git commit --amend`
`git push -f`

impaktor: Also, you don't have to give me credit for the code, in the comment...

PR #4225, opened 2017-12-28
Force time acceleration if right mouse button is held down.

### LINKED ISSUE #4223: Accessibility: time acceleration access mouse/keyboard only not mixed
yeh. Im disabled. paralyzed left hand. on a good day I can glue a left hand finger to control key and apply pressure, but aside from that, please make an alternative for ctrl-click to force acceleration.
right click is unused for instance.

pioneer-20171001-win32

fluffyfreak:
As an alternative to holding down the Ctrl key, you can also hold down the right mouse button & click to force time acceleration.

Should fix #4223

@joonicks how does that sound?


joonicks:
sounds great

joonicks:
tested it, works great

Issue #4226, opened 2017-12-29
Lack of verbose debug when loading json data


joonicks:
Currently when loading json data, there is no feedback at all. Resulting in the game loading and then crashing as soon as a script tries to use a resource from a buggy file.

Any small issue and the whole json file is silently discarded.

Makes it very painful to try and figure out when all it is, is a misplaced \ or " in one of the strings.

Would be a big improvement if the json loader simply said which line it quit on.

johnbartholomew:
Note for whoever works on this: The JSON library we're using has a getFormattedErrorMessages() method... we probably just need to call that and print out the message as appropriate.

PR #4230, opened 2017-12-29
Handle JSON errors in Ship Type, Station Type definitions

johnbartholomew:
For #4226. Makes the game abort on startup when it tries to load a ship-def or station-def that fails parsing. There should be a not-totally-useless error message written to stdout, or (on Windows) to output.txt.

Code structure is pretty bad (sorry).

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -827,7 +827,10 @@ Game *Game::LoadGame(const std::string &filename)
 		const char *pdata = plain_data.data();
 		Json::Reader jsonReader;
 		Json::Value rootNode;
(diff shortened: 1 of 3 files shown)


impaktor:
> For #4226. 

Does it [fix](https://help.github.com/articles/closing-issues-using-keywords/) the issue as well?

johnbartholomew:
> Does it fix the issue as well?

Yes. The commit has the tag to close the issue.

impaktor:
Doh, sorry. I should have checked.

fluffyfreak:
Yes this looks good to me.

PR #4233, opened 2017-12-30
Allow custom factions to make claims

### LINKED ISSUE #4231: Feature request: Through lua setting faction on individual stars or sectors
By way of setup scripts, setting the faction of individual star systems or whole star sectors to custom values without otherwise altering the star systems.

joonicks:
Adds a new command to use when creating custom factions, setting the faction for any individual starsystem or sector.

When creating custom factions, either when doing the Faction.New(), add a :claim(x,y,z,idx) or after the f = Faction.New() add rows of f:claim(x,y,z,idx)

f:claim(x,y,z) for every star in a sector,
f:claim(x,y,z,idx) for an individual starsystem

claims are checked before expansionrate claims
custom systems can not be claimed, only randomly generated ones.

fixes  #4231

### DIFF
--- a/src/Factions.cpp
+++ b/src/Factions.cpp
@@ -553,7 +553,7 @@ bool FactionsDatabase::MayAssignFactions() const
 	return m_may_assign_factions;
 }
 
-const Faction* FactionsDatabase::GetNearestFaction(const Sector::System* sys) const
+const Faction* FactionsDatabase::GetNearestClaimant(const Sector::System* sys) const
 {
 	PROFILE_SCOPED()
 	// firstly if this a custom StarSystem it may already have a faction assigned
--- a/src/Factions.cpp
+++ b/src/Factions.cpp
@@ -569,7 +569,7 @@ const Faction* FactionsDatabase::GetNearestFaction(const Sector::System* sys) co
 	for (ConstFactionIterator it = candidates.begin(); it != candidates.end(); ++it)
 	{
 		if ((*it)->IsClaimed(sys->GetPath()))
-			result = *it;
+			return *it; // this is a very specific claim, no further checks for distance from another factions homeworld is needed.
 		if ((*it)->IsCloserAndContains(closestFactionDist, sys))
 			result = *it;
 	}
--- a/src/Factions.cpp
+++ b/src/Factions.cpp
@@ -230,6 +230,27 @@ static int l_fac_colour(lua_State *L)
 	return 1;
 }
 
+static int l_fac_claim(lua_State *L)
+{
+	Faction *fac = l_fac_check(L, 1);
+
+	Sint32 sector_x = luaL_checkinteger(L, 2);
+	Sint32 sector_y = luaL_checkinteger(L, 3);
+	Sint32 sector_z = luaL_checkinteger(L, 4);
+
+	SystemPath path(sector_x, sector_y, sector_z);
+
+	path.systemIndex = -99; // magic number, claim whole sector
+
+	if (lua_gettop(L) > 4)
+	{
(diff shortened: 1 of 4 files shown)


impaktor:
Please consider using `git commit --amed` when fixing bugs, as we don't want to track the bug-fixing history of PRs prior to merge.

Not sure if @laarmen want to review, although not Lua.

joonicks:
im using the desktop client, not the commandline

impaktor:
OK, I can squash this at merge, but I have not tested this, as I don't have my pioneer computer at hand.

src/Factions.cpp
```diff
@@ -544,8 +566,12 @@ const Faction* FactionsDatabase::GetNearestFaction(const Sector::System* sys) co
 	double closestFactionDist = HUGE_VAL;
 	ConstFactionList& candidates = m_spatial_index.CandidateFactions(sys);
 
-	for (ConstFactionIterator it = candidates.begin(); it != candidates.end(); ++it) {
-		if ((*it)->IsCloserAndContains(closestFactionDist, sys)) result = *it;
+	for (ConstFactionIterator it = candidates.begin(); it != candidates.end(); ++it)
+	{
+		if ((*it)->IsClaimed(sys->GetPath()))
```
impaktor: Granted, I haven't digged into the code here, so my understanding is very low (sorry) , but just at a glance I assumed this would be an `if-else`, not `if-if`, so the isClaimed should take precedence over what ever was there before?

joonicks: yeah youre right

joonicks:
The inner workings of this:

As the parser reads in custom factions, any claim for that faction is stored in a list within that faction.

Once the game starts, any time a new starsystem is created, it checks if the star is within the sphere of growth of any faction. This code adds going through each factions internal claim list to see if anything matches. The list can never be dropped because the star system cache might forget the system and the list needs to be intact in case the system is generated again.

Fortunatly, each claim is only a single SystemPath

The code can be optimized if speed is essential.

src/Factions.cpp
impaktor: In the fix, to replace it with a `return`, you're breaking out of the for-loop, which, at a glance, looks unsymmetric with the other if-statment in the loop, so just checking if it is right that is should break out of the for-loop if `isClaimed`?

laarmen: Yes, it's correct: as I understand it, any explicit claim takes over any proximity-induced implicit claim.

However, why is this mechanism in a function called "GetNearestFaction" ? The name implies very specific semantics, straying from those isn't a very good idea and will lead to confusion further down the road.

joonicks: GetGeneratedFaction, FindBestFaction, ... hmm err whats suitable?

joonicks: function is called twice, once by AssignFaction, once by StarSystemFromSectorGenerator::Apply,
so.. call it FindFactionToApply? GetFactionToApply?

laarmen:
I'd go with GetNearestClaimant. IMHO a method name should reflect what it
does, not what we're supposed to use it for.

impaktor:
So a quick name change of that function, then good to merge? (and git squash at merge)

@fluffyfreak I noticed the merge button has a drop down menu, that might interest you: "squash and merge"!

fluffyfreak:
CRUSH AND MERGE!

impaktor:
@fluffyfreak yes, I thought I'd offer this oportunity to you to use the drop down menu of the merge-button? If not I'll do it tomorrow. 

(I'm tryinng to condition you, now all I need is a bell, and... pavlov's dogs? That doesn't even make sense.)

PR #4236, opened 2017-12-31
Get a systems form of government in Lua

joonicks:
new attribute to StarSystem lua objects;
system.govtype is the form of government used in the local system.

like, "Game.system.govtype" is the current systems govtype.

### DIFF
--- a/src/LuaStarSystem.cpp
+++ b/src/LuaStarSystem.cpp
@@ -6,7 +6,6 @@
 #include "LuaTable.h"
 #include "LuaUtils.h"
 #include "EnumStrings.h"
-#include "enum_table.h"
 #include "LuaUtils.h"
 #include "galaxy/StarSystem.h"
 #include "galaxy/Economy.h"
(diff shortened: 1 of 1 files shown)


src/LuaStarSystem.cpp
```diff
@@ -6,6 +6,7 @@
 #include "LuaTable.h"
 #include "LuaUtils.h"
 #include "EnumStrings.h"
+#include "enum_table.h"
```
johnbartholomew: enum_table.h not needed if you use EnumStrings::GetString.

src/LuaStarSystem.cpp
```diff
@@ -542,6 +543,37 @@ static int l_starsystem_attr_faction(lua_State *l)
...
+	int j = 0;
+	for (; ENUM_PolitGovType[j].name != 0; j++)
+	{
+		if (ENUM_PolitGovType[j].value == s->GetSysPolit().govType)
+			break;
+	}
+	if (ENUM_PolitGovType[j].name == 0)
+		return 0;
+
+	lua_pushstring(l, ENUM_PolitGovType[j].name);
```
johnbartholomew: You don't need any of the above loop, you can just do:

    lua_pushstring(l, EnumStrings::GetString("PolitGovType", s->GetSysPolit().govType));

johnbartholomew:
Squashed and merged as commit e61c7de6a18aa488bd5f324f497bd8e444648b91.

PR #4238, opened 2018-01-01
CustomSystem input syntax adapted for easier data entry

### LINKED ISSUE #4237: CustomSystem absurd input syntax requirements
To me its absurd that CustomSystem doesnt accept float numbers and similar but instead forces the human user to do the conversion to "fixed()".

It should rather be the computer that converts "1.2345" to the appropriate internal format. Massive work overload and excellent potential for mistakes to happen.

Also for some values, allowing alternate units to be used shouldnt be much work. Instead of only accepting earth masses and eath radii, etc.

joonicks:
No longer do you have to calculate the values into a fraction and preprocess with the f() function.

Now all attributes that took f() values, also accept real numbers. It accepts both.

Extra addition: you can use radius_km() instead of radius() to set the radius value in kilometers instead of size relative to earth. The proper value will be calculated by the computer, which is how it should be.

Example from 00_sol.lua;

local ceres = CustomSystemBody:new('Ceres', 'PLANET_TERRESTRIAL')
	:radius(f(074,1000)) -- 473km
	:mass(f(15,100000)) -- 9.393e20 kg
	:temp(168)
	:semi_major_axis(f(27675,10000)) -- 2.7675 AU
	:eccentricity(f(758,10000)) -- 0.075823
	:inclination(math.deg2rad(10.593)) -- 10.593°
	:rotation_period(f(3781,10000)) -- 9h
	:axial_tilt(fixed.deg2rad(f(4,1))) -- 4°

Can now be changed to;

local ceres = CustomSystemBody:new('Ceres', 'PLANET_TERRESTRIAL')
	:radius_km(473) -- 473km
	:mass(0.00015) -- 9.393e20 kg
	:temp(168)
	:semi_major_axis(2.7675) -- 2.7675 AU
	:eccentricity(0.0758) -- 0.075823
	:inclination(math.deg2rad(10.593)) -- 10.593°
	:rotation_period(0.3781) -- 9h
	:axial_tilt(fixed.deg2rad(4)) -- 4°

It will work either way. Hopefully making data entry less of a headache.

*EDIT* closes #4237

### DIFF
--- a/src/galaxy/CustomSystem.cpp
+++ b/src/galaxy/CustomSystem.cpp
@@ -56,14 +56,35 @@ static int l_csb_new(lua_State *L)
 	return 1;
 }
 
+static double *getDoubleOrFixed(lua_State *L, int which)
+{
+	static double mixvalue;
+
+	const char *s = luaL_typename(L, which);
+	if (strcmp(s, "userdata") == 0)
+	{
+		const fixed *mix = LuaFixed::CheckFromLua(L, which);
+		mixvalue = mix->ToDouble();
+		return(&mixvalue);
+	}
+	else if (strcmp(s, "number") == 0)
+	{
+		mixvalue = luaL_checknumber(L, which);
+		return(&mixvalue);
+	}
+	return nullptr;
+}
+
 // Used when the value MUST not be NEGATIVE but can be Zero, for life, etc
 #define CSB_FIELD_SETTER_FIXED(luaname, fieldname)          \
 	static int l_csb_ ## luaname (lua_State *L) {           \
 		CustomSystemBody *csb = l_csb_check(L, 1);          \
-		double value = luaL_checknumber(L, 2);				\
-		if (value < 0.0)									\
-			Output("Error: Custom system definition: Value cannot be negative (%lf) for %s : %s\n", value, csb->name.c_str(), #luaname);\
-		csb->fieldname = (fixed)value;						\
+		double *value = getDoubleOrFixed(L, 2);				\
+		if (value == nullptr)								\
+			return luaL_error(L, "Bad datatype. Expected fixed or float, got %s", luaL_typename(L, 2)); \
+		if (*value < 0.0)									\
+			Output("Error: Custom system definition: Value cannot be negative (%lf) for %s : %s\n", *value, csb->name.c_str(), #luaname); \
+		csb->fieldname = fixed::FromDouble(*value);			\
 		lua_settop(L, 1); return 1;                         \
 	}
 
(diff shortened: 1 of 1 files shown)


impaktor:
Maybe @jaj22 has interest to review. I can't test, as I'm away from pioneer-computer. Seems like a feature we want.

nozmajner:
Somebody said a couple of months ago, that those values are handled like this, so they are deterministic. Does that point matters/affected in this case?

jaj22:
What's probably the reasoning behind the original behaviour is that text->FP and FP->integer operations are not strictly platform-independent. Small changes to physical parameters could have large effects on the terrain generator, which could lead to a starport that appeared in different positions on different platforms (or with different compiler settings).

I'm not sure how large the risks are in practice. Most relevant platforms will be using SSE2, and text->FP may be standardised. It's also possible that the terrain noise has no dependence on physical parameters, in which case single bit errors are unlikely to make any difference. I'd need to check.

The safe way to handle float input is to do the string->fixed conversion manually, but this is of course a pain.

joonicks:
jaj, if text->FP can mess things up youd better go through the whole galaxy generation bit that uses plenty of Fixed::FromDouble values internally.

jaj22:
Yeah ok, apparently someone broke that rule a while back. Notably StarSystemGenerator.cpp, line 813+. Original version didn't use doubles at all.

m_radius having a dependency on the pow function is about as bad as you can get for determinism, so there's unlikely to be any additional harm from bouncing the custom system params through doubles.

nozmajner:
That's kinda yay then, the current method is a pain to use.

fluffyfreak:
@jaj22 yeah that was me I'm afraid, I just couldn't work out the Fixed stuff for that.

impaktor:
I'm not sure what the final verdict is? Are people happy if we merge this as is? 

Any objectrions @jaj22 @fluffyfreak ?

(maybe @laarmen or @johnbartholomew want to have a chance to voice in, if this is of any interest)

Relevant IRC-discussion from yesterday:
```
<impaktor> If I understand thing correctly, it will not introduce any more
           platform specific behaviour than what was already there, but it
           will cement it further, I guess.

<JohnJ> If you don't want to cement it further, then you could instead pass
        values as strings, and then fix it later if desired.
<JohnJ> essentially do the string->whatever conversions strictly on the C++
        side.

<joonicks> there is no string->fixed code to use

<JohnJ> yeah, you'd just placeholder with string->double->fixed.
<JohnJ> I think my fake-fp code did have string input functionality :P

<joonicks> not entirely necessary
<joonicks> the lua interpreter could be altered to store numbers in fixed
           and/or doubles and pass the relevant one
<joonicks> but i think the platform independence is a pipedream
<joonicks> its just way too much complexity to be made perfect

<JohnJ> It used to be essential because x87 was so compiler-dependent.
<JohnJ> Issue with system generation is that it's a series of dependent
        operations, and small differences early in the pipe get magnified.
<JohnJ> but maybe with SSE2 you can largely ignore it.

<joonicks> well, right now it accepts both f() and number input. to make it
           "compliant" again you just undo the number input
<joonicks> in any case, if custom systems appear differently on different
           platforms, its not the end of the world
<joonicks> however the use of doubles in galaxy generation has the potential
           to fuck up the whole galaxy
<joonicks> custom systems vs the whole galaxy is a bit different issues

<JohnJ> other related issue is that the scales are not self-documenting, like
        stars and planets being measured in different units.
<JohnJ> There used to be a third category for asteroids which apparently
        no-one knew about.
<JohnJ> I just checked the terrain code and errors in planet radius do
        directly affect the noise RNG, although single bit errors in a double
        are not likely to have a significant effect.
```

fluffyfreak:
I'm good with it as-is

jaj22:
Yeah, in retrospect anything in custom system definitions should be fine to specify as doubles. Just keep them out of galaxy generation and preferably system accretion calcs.

Would prefer more transparent handling of earth vs solar units, but as the interpretation happens at another level it's not within the scope of this patch.

laarmen:
I have no idea what this is all about, so merge at will ;-)

fluffyfreak:
MEEEEEEEEEEEEEEEEEEERGING

PR #4240, opened 2018-01-01
Happy New Year!

impaktor:
Ten years in the making now, but how can we improve upon perfection
for ten more years?


joonicks:
there is no such thing as a perfect program

fluffyfreak:
Merge at will!

PR #4241, opened 2018-01-01
Hyperjump Route Planner

The-EG:
Work in progress!

This adds a hyperjump route planner to the new sector map in PR #4204. This PR is currently based on that branch and not master (not ready for merge yet!). 

The planner (lower right):
base profile screenshot 2017 12 29 - 13 03 59 53

base profile screenshot 2017 12 29 - 13 41 35 41

The planner has functions for adding the currently selected system, removing jumps, reordering and a *very* simple 'auto-route' feature. The option to 'remove jump when completed' will automatically remove the first jump from the list if you hyperjump to that system.

In addition to adding the planner itself:
- Add lines for the current route to the sector map (yellow lines)
- Altered the display of the 'far' (faction colored boxes) view. Smaller boxes and slightly more transparent
- Lowered the font size used on the sector map (and a few other places like the auto pilot screen), which makes the sector map much more readable.

Open issues/todo:
- [ ] Integrate the planner into the main sector map window (left) (tabs?)
- [X] Move the auto-route feature to c++ and use an actual networking algorithm
- [ ] Custom icons
- [ ] Setting button for displaying route lines
- [ ] Favorites/Saving route between game sessions
- [x] Internationalization/Translations (there's some done already for unit names and what-not, but most of the strings are just literals right now).

### DIFF
--- a/data/fonts/OverlayFont.json
+++ b/data/fonts/OverlayFont.json
@@ -4,14 +4,14 @@
    "faces" : [
       {
          "fontFile" : "PionilliumText22L-Medium.ttf",
-         "pixelWidth" : 14,
-         "pixelHeight" : 14,
+         "pixelWidth" : 11,
+         "pixelHeight" : 11,
          "advanceXAdjustment" : -1
       },
       {
          "fontFile" : "DejaVuSans.ttf",
-         "pixelWidth" : 14,
-         "pixelHeight" : 14,
+         "pixelWidth" : 11,
+         "pixelHeight" : 11,
          "advanceXAdjustment" : -1,
          "ranges" : [
             [ "0x0400", "0x04FF" ],
@@ -24,8 +24,8 @@
       },
       {
         "fontFile" : "wqy-microhei.ttc",
-        "pixelWidth" : 14,
-        "pixelHeight" : 14,
+        "pixelWidth" : 11,
+        "pixelHeight" : 11,
         "ranges" : [
             [ "0x4E00",  "0x9FFF"  ],
             [ "0x3400",  "0x4DFF"  ],
--- a/src/LuaEngine.cpp
+++ b/src/LuaEngine.cpp
@@ -1041,6 +1041,21 @@ static int l_engine_get_model(lua_State *l)
 	return 1;
 }
 
+static int l_engine_sector_map_clear_route(lua_State *l)
+{
+	SectorView *sv = Pi::game->GetSectorView();
+	sv->ClearRoute();
+	return 0;
+}
+
+static int l_engine_sector_map_add_to_route(lua_State *l)
+{
+	SectorView *sv = Pi::game->GetSectorView();
+	SystemPath *path = LuaObject<SystemPath>::CheckFromLua(1);
+	sv->AddToRoute(path);
+	return 0;
+}
+
 static int l_engine_get_sector_map_zoom_level(lua_State *l)
 {
 	SectorView *sv = Pi::game->GetSectorView();
@@ -1062,21 +1077,21 @@ static int l_engine_get_sector_map_center_sector(lua_State *l)
 	return 1;
 }
 
-static int l_engine_get_sector_map_current_system_path(lua_State *l) 
+static int l_engine_get_sector_map_current_system_path(lua_State *l)
 {
 	SectorView *sv = Pi::game->GetSectorView();
 	LuaObject<SystemPath>::PushToLua(sv->GetCurrent());
 	return 1;
 }
 
-static int l_engine_get_sector_map_selected_system_path(lua_State *l) 
+static int l_engine_get_sector_map_selected_system_path(lua_State *l)
 {
 	SectorView *sv = Pi::game->GetSectorView();
 	LuaObject<SystemPath>::PushToLua(sv->GetSelected());
 	return 1;
 }
 
-static int l_engine_get_sector_map_hyperspace_target_system_path(lua_State *l) 
+static int l_engine_get_sector_map_hyperspace_target_system_path(lua_State *l)
 {
 	SectorView *sv = Pi::game->GetSectorView();
 	LuaObject<SystemPath>::PushToLua(sv->GetHyperspaceTarget());
@@ -1121,6 +1136,79 @@ static int l_engine_set_sector_map_automatic_system_selection(lua_State *l)
 	bool value = LuaPull<bool>(l, 1);
 	sv->SetAutomaticSystemSelection(value);
 	return 0;
+}
+
+static int l_engine_sector_map_get_route(lua_State *l) {
+	SectorView *sv = Pi::game->GetSectorView();
+	std::vector<SystemPath> route = sv->GetRoute();
+
+	lua_newtable(l);
+	int i = 1;
+	for (const SystemPath j : route) {
+		lua_pushnumber(l, i++);
+		LuaObject<SystemPath>::PushToLua(j);
+		lua_settable(l, -3);
+	}
+	return 1;
+}
+
+static int l_engine_sector_map_get_route_size(lua_State *l) {
+	SectorView *sv = Pi::game->GetSectorView();
+	std::vector<SystemPath> route = sv->GetRoute();
+	const int size = route.size();
+	LuaPush(l, size);
+	return 1;
+}
+
+static int l_engine_sector_map_auto_route(lua_State *l)
+{
+	SectorView *sv = Pi::game->GetSectorView();
+	SystemPath current_path = sv->GetCurrent();
+	SystemPath target_path = sv->GetSelected();
+
+	std::vector<SystemPath> route = sv->AutoRoute(current_path, target_path);
+	sv->ClearRoute();
+	for (auto it = route.begin(); it != route.end(); it++) {
+		sv->AddToRoute(*it);
+	}
+
+	return l_engine_sector_map_get_route(l);
+}
+
+static int l_engine_sector_map_move_route_item_up(lua_State *l) {
+	SectorView *sv = Pi::game->GetSectorView();
+	int element = LuaPull<int>(l, 1);
+
+	// lua indexes start at 1
+	element -= 1;
+
+	bool r = sv->MoveRouteItemUp(element);
+	LuaPush<bool>(l, r);
+	return 1;
+}
+
+static int l_engine_sector_map_move_route_item_down(lua_State *l) {
+	SectorView *sv = Pi::game->GetSectorView();
+	int element = LuaPull<int>(l, 1);
+
+	// lua indexes start at 1
+	element -= 1;
+
+	bool r = sv->MoveRouteItemDown(element);
+	LuaPush<bool>(l, r);
+	return 1;
+}
+
+static int l_engine_sector_map_remove_route_item(lua_State *l) {
+	SectorView *sv = Pi::game->GetSectorView();
+	int element = LuaPull<int>(l, 1);
+
+	// lua indexes start at 1
+	element -= 1;
+
+	bool r = sv->RemoveRouteItem(element);
+	LuaPush<bool>(l, r);
+	return 1;
 }
 
 static int l_engine_set_sector_map_selected(lua_State *l)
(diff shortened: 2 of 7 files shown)


nozmajner:
I'll make the icons the coming days, or next weekend.

I think it could be integrated into the left side panel, since with this, the need for two separate system selections are moot, so there'd be more room available. The question is, how that would work on smaller screens.

fluffyfreak:
This is a thing of beauty

The-EG:
Some more todo's that I thought of:
- Favorites/Saving route between game sessions
- Internationalization/Translations (there's some done already for unit names and what-not, but most of the strings are just literals right now).

impaktor:
> Favorites/Saving route between game sessions

That sounds (over?) ambitious. In generall we don't, or try not to, let information from one game influence the next.

nozmajner:
Yeah, it sounds more useful to be from save to save.

fluffyfreak:
I think he means savegames, not between starting a new game.

The-EG:
Sorry that wasn't clear; I do mean between save games and not separate games.

laarmen:
For that, I think Elite:Dangerous has the right approach : simply store the
intended destination, and require the user to go back in the planner to
compute it again.

impaktor:
> For that, I think Elite:Dangerous has the right approach : simply store the intended destination, and require the user to go back in the planner to compute it again.

This makes sense, especially once we've implemented the system-destroying death star / Vogon attack fleet, that could potentially annihilate entire systems (jump points) needed to get to the destination; because this is how we measure how good a game is, right? Both Sid Meier's Alpha Centauri and Master of Orion II, allowed the destruction of entire cities / planets, respectively.

joonicks:
I agree with laarmen on the saving routes issue, bookmarks would be very nice to have though.

impaktor:
> Move the auto-route feature to c++ and use an actual networking algorithm

Here's a 20 minute video I enjoyed watching back when I learned some graph theory: [Dijkstra 's Shortest Path Algorithm](https://www.youtube.com/watch?v=jRlNVmRjdRk)

The-EG:
> Here's a 20 minute video I enjoyed watching back when I learned some graph theory.

Cool, I'll take a look. I've been reading up on that algorithm some, too. 
I'll probably tinker with it this weekend. Luckily, the changes I made to the sector map for drawing the route lines taught me some about accessing sectors/systems and how the caching works, so I think I have a good idea of how to approach it in an efficient manner.

impaktor:
@The-EG I took that stanford MOOC-course by that guy, I really loved it, he's an awesome teacher. Have fun! (I hope the video makes sense without having watched the previous ones)

The-EG:
@impaktor I took a go at implementing that algorithm...the routes look odd at times, but they are the fastest I believe.
The only issue I'm having is the duration is calculated within Lua and I can't seem to access it correctly. For now I'm calculating it in the same way, but I know it really should be calling the same Lua function instead.

impaktor:
> For now I'm calculating it in the same way, but I know it really should be calling the same Lua function instead.

Yeah, I think this should probably be addressed before merging. Not sure I'm the right person to ask right now, but I'm sure you can get help in IRC, or here, if you can specify some key questions that will set you on your path.

The-EG:
> if you can specify some key questions

[This part](https://github.com/pioneerspacesim/pioneer/pull/4241/commits/d7484987d207b9c11bd6611b0006489c60d22310#diff-4f612648d55f1a365bb2bc4cd167353aR654) in particular is what I'm having issues with:
> //auto v_stats = LuaObject<Player>::CallMethod<std::string, float, float, float>(Pi::player, "GetHyperspaceDetails", closest, v);
> //float v_dist = std::get<3>(v_stats);

Those two statements should call lua and return the duration, but I get a compile error. It looks like it doesn't know how to push a `SystemPath` back to lua through the generic push methods. ecraven did write some stuff for that, but it doesn't seem to be picking it up.

joonicks:
the-eg, have you ever though about caching the duration for a distance, or pre-generating a list of distance to time values to use to process it all quicker?

The-EG:
@joonicks There's a few things that could improve performance, like caching the durations themselves, but I'm not sure how useful caching the result of a calculation in addition would be since the distance would have to be exactly the same (and how often would that happen?).

joonicks:
skip "exactly the same", close enough is good enough. approximation is good enough.

impaktor:
Is it slow to do a search? I assume it depends on distance, of course...

> the distance would have to be exactly the same (and how often would that happen?).

If wanting to travel back to the original system, then I assume symmetry.

But this sounds like a optimization that could be saved for another PR. Try to keep it simple to get this into master.

The-EG:
> But this sounds like a optimization that could be saved for another PR. Try to keep it simple to get this into master.

That was my intent, just get it working in a simple form first.

The-EG:
I haven't figured out why calling `GetHyperspaceDetails` wasn't working that way, but I did find a way around it. And it should streamline things a bit in the end too. 
Now, the hyperdrive (engine) is pulled from Lua at the very beginning and that's used (`GetDuration`) to get the durations instead.

impaktor:
I take it this is done then? So we're just waiting for #4204 to be merged first, and rebased on that.

The-EG:
For now, yes. There are still the other ToDo's listed, but they could be done with later PRs.

impaktor:
@The-EG 4204 is merged now, so feel free to rebase ontop of new master, and please let us know when this is ready for merge.

The-EG:
@impaktor Ok: one interactive rebase later, and it's back onto master...

This is now ready for merge. The other 'to-dos' can be handled in separate PRs; I'll leave the list here for reference.

impaktor:
A warning:

```
SectorView.cpp: In member function ‘void SectorView::DrawRouteLines(const vector3f&, const matrix4x4f&)’:
SectorView.cpp:704:20: warning: comparison between signed and unsigned integer expressions [-Wsign-compar ]
  for (int i = 0; i < m_route.size(); ++i) {
```

a crash:
```
cleanTSTable:total:63,active:46,removed:0
error: [string "[T] @pigui/modules/hyperjump-planner.lua"]:163: bad argument #1 to 'SectorMapGotoSystemPath' (userdata expected, got nil)
stack traceback:
	[C]: in function 'SectorMapGotoSystemPath'
	[string "[T] @pigui/modules/hyperjump-planner.lua"]:163: in function 'callback'
	[string "[T] @pigui/modules/hyperjump-planner.lua"]:104: in function 'mainButton'
	[string "[T] @pigui/modules/hyperjump-planner.lua"]:160: in function 'showJumpRoute'
	[string "[T] @pigui/modules/hyperjump-planner.lua"]:217: in function 'fun'
	[string "[T] @pigui/pigui.lua"]:65: in function 'window'
	[string "[T] @pigui/modules/hyperjump-planner.lua"]:211: in function 'fun'
	[string "[T] @pigui/pigui.lua"]:99: in function 'withStyleColors'
	[string "[T] @pigui/modules/hyperjump-planner.lua"]:210: in function 'showHyperJumpPlannerWindow'
	[string "[T] @pigui/modules/hyperjump-planner.lua"]:238: in function 'fun'
	[string "[T] @pigui/game.lua"]:677: in function 'callModules'
	[string "[T] @pigui/game.lua"]:926: in function 'fun'
	[st%

```

Above might have been triggered by adding a destination, the clearing the list, then click "zoom to destination"

src/LuaEngine.cpp
```diff
@@ -1307,6 +1395,15 @@ void LuaEngine::Register()
 		{ "SectorMapGotoSystemPath",                l_engine_sector_map_goto_system_path },
 		{ "GetSectorMapFactions",                   l_engine_get_sector_map_factions },
 		{ "SetSectorMapFactionVisible",             l_engine_set_sector_map_faction_visible },
+		{ "SectorMapAutoRoute",                     l_engine_sector_map_auto_route },
+	    { "SectorMapGetRoute",                      l_engine_sector_map_get_route },
+	    { "SectorMapGetRouteSize",                  l_engine_sector_map_get_route_size },
```
impaktor: indentation

The-EG:
> A warning:

Now using `std::vector<SystemPath>::size_type` instead of int.

> a crash:

Ok, specifically this is caused by building a route, _selecting a jump in the list_, clearing it and then clicking zoom to. Should be fixed now (selected is set to `nil` when the list is cleared).

> indentation

Also fixed (this file uses tabs, not spaces!)

I also fixed another issue: the route was not cleared between games. It is now cleared when a game ends.

impaktor:
Added translation strings to your PR. Feel free to review them, as I did it a bit half-fast.

I changed "zoom to system" to "center on system" as that is what it appears to do.

Also, I noticed you have some trailing spaces at quite a lot of places. Can your editor nuke them, and squash into your previous commit? (Yeah, I know, nitpick comment)

(I have set my editor to nuke all trailing spaces each time I save a change in a file)

src/SectorView.cpp
```diff
@@ -532,6 +549,188 @@ void SectorView::DrawNearSectors(const matrix4x4f& modelview)
...
+	std::set<std::vector<SystemPath>::size_type> unvisited;
+
+	std::vector<float> path_dist; // distance from source to node
+	std::vector<std::vector<SystemPath>::size_type> path_prev; // previous node in optimal path
+
+	// nodes[0] is always start
+	nodes.push_back(start);
+
+	// go sector by sector for sec_dist sectors and add systems
+	// if they are within 110% of dist of both start and target
```
mooskagh: Does this feature allow trying to build a rote to a very remote systems? If it is, it feels like it can be problematic, as memory needed is O(d^3), and runtime is O(d^6), where d is distance between source and destination.

laarmen: I don't see any checks against it. However, I don't see this as a blocking problem, the algorithm can be optimized afterwards *if* this proves problematic, which might not even be the case! ;-)

The-EG:
It does currently allow building routes to very remote targets and that could present a blocking issue. 
When I was testing it, routes between systems about 85LY apart would take a noticeable length of time to run (seconds).

I'd argue against restricting it, since I think planning long routes is part of why this is useful. But a warning it something may be warranted.

There is definitely room for optimization. Unfortunately, this is the first time I've ever dealt with this routing algorithm, so my first concert was to write something that functioned proper.
I think it could also be an opportunity to thread it and provide a 'cool' routing animation for really long routes.

impaktor:
@The-EG  `git show <commit>  -W` shows all the culprits.

I can fix the whitespace at merge if you like, as my editor fixes them all by default.

I'd just like to check with @ecraven that he's happy with the imgui-code before merging.

The-EG:
Ok, trailing whitespace should be cleaned up, and I believe I got the commits squashed/reordered correctly.

It sounds like I need to break down and get emacs or VIM setup on Windows...

data/fonts/OverlayFont.json
```diff
@@ -4,14 +4,14 @@
    "faces" : [
       {
          "fontFile" : "PionilliumText22L-Medium.ttf",
-         "pixelWidth" : 14,
```
ecraven: Is this change meant to influence imgui? font sizes are still hardcoded for imgui, these settings don't do anything for it. If it is meant to work for C++, that's correct then ;)

src/LuaEngine.cpp
```diff
@@ -1123,6 +1138,79 @@ static int l_engine_set_sector_map_automatic_system_selection(lua_State *l)
 	return 0;
 }
 
+static int l_engine_sector_map_get_route(lua_State *l) {
+	SectorView *sv = Pi::game->GetSectorView();
+	std::vector<SystemPath> route = sv->GetRoute();
+
+	lua_newtable(l);
```
ecraven: Not sure if this works here, but LuaTable might be a better fit (no low-level lua functions)

src/SectorView.cpp
```diff
@@ -532,6 +549,188 @@ void SectorView::DrawNearSectors(const matrix4x4f& modelview)
 	Gui::Screen::LeaveOrtho();
 }
 
+bool SectorView::MoveRouteItemUp(const std::vector<SystemPath>::size_type element) {
+	if (element <= 0 || element >= m_route.size()) return false;
+
+	SystemPath tmp = m_route[element - 1];
```
ecraven: Would std::swap work here? http://en.cppreference.com/w/cpp/algorithm/swap

data/fonts/OverlayFont.json
The-EG: This changes the fonts that are rendered in the 3D portion of the sector map, like the system labels. It also changes a few others, like the auto-pilot/comms window.
I was originally going to do this as a separate commit, but @nozmajner felt it impacted the sector map enough to include it here.

src/LuaEngine.cpp
The-EG: To be honest, this is the first time I've worked with Lua at all..so I stumbled through moving data back and forth. If there's something already written for doing that, I'm open to it.

data/fonts/OverlayFont.json
impaktor: > but @nozmajner felt it impacted the sector map enough to include it here.

Hey-hey, it's spelled "impakted" please. 😃

nozmajner: Yup, it was my suggestion, to get the text sizes of the map labels closer to the imgui parts. I think it looks way better this way.

src/SectorView.cpp
The-EG: Yes, I imagine so. This is me slipping back into my old C habits...

The-EG:
This is ready to merge.

I fixed the std::swap parts. LuaTable doesn't know what to do with a SystemPath, so it's been left as is for now.
Also, en.json had CENTER_ON_SYSTEM but hyperjump-planner.lua referenced CENTER_TO_SYSTEM. I changed the lua to match the json.

PR #4245, opened 2018-01-03
sectorview zoomout prettier homeworld labels

joonicks:
no more quick and dirty labels. made the point stand out from the background clutter with contrast background. centered the label and decreased the margins slightly. label no longer overlaps the star.

### DIFF
--- a/src/SectorView.cpp
+++ b/src/SectorView.cpp
@@ -764,29 +764,38 @@ void SectorView::PutFactionLabels(const vector3f &origin)
 
 				Gui::Screen::MeasureString(labelText, labelWidth, labelHeight);
 
+				// draw a big diamond for the location of the star
+				static const float STARSIZE = 5;
+				Graphics::VertexArray outline(Graphics::ATTRIB_POSITION);
+				outline.Add(vector3f(pos.x - STARSIZE - 1.f, pos.y, 1));
+				outline.Add(vector3f(pos.x, pos.y + STARSIZE + 1.f, 1));
+				outline.Add(vector3f(pos.x, pos.y - STARSIZE - 1.f, 1));
+				outline.Add(vector3f(pos.x + STARSIZE + 1.f, pos.y, 1));
+				m_material->diffuse = { 0, 0, 0, 255 };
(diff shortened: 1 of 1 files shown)


joonicks:
screenshot-20180101-224218

joonicks:
screenshot-20180101-224218

src/SectorView.cpp
```diff
@@ -764,29 +764,38 @@ void SectorView::PutFactionLabels(const vector3f &origin)
...
-
-				{
-					Graphics::VertexArray va(Graphics::ATTRIB_POSITION);
-					va.Add(vector3f(pos.x - 5.f,              pos.y - 5.f,               0));
-					va.Add(vector3f(pos.x - 5.f,              pos.y - 5.f + labelHeight, 0));
-					va.Add(vector3f(pos.x + labelWidth + 5.f, pos.y - 5.f,               0));
-					va.Add(vector3f(pos.x + labelWidth + 5.f, pos.y - 5.f + labelHeight, 0));
-					m_material->diffuse = labelBorder;
-					m_renderer->DrawTriangles(&va, renderState, m_material.Get(), Graphics::TRIANGLE_STRIP);
-				}
```
laarmen: You removed the explicit code blocks, any reason?

joonicks: any reason for the codeblocks being there in the first place? other than not having to rename the variable?

I dont know if the compiler optimizes away locally created stack frames, but if it doesnt, then thats overhead just for being too lazy to rename a variable.

impaktor: 1. If the code-snippet is repeated, with just variable name change, then why not a nice lambda-function?

2. putting actual feature change/addiont in seperate commit, and then all refactoring changes in a single commit e.g. names "refactoring changes" or "indentation fix" or something, makes it easier to review.

joonicks: Dude... I dont know C++.. but I do know that the new code is in 3 blocks that replace the previous two blocks. none of the three new blocks is identical, they all have some differences. none of the change is "indentation fix".

fluffyfreak:
This is good, I've only got time a quick skim, hopefully merge it tomorrow

PR #4246, opened 2018-01-03
+1 city building

joonicks:
its my house, I live there


joonicks:
myhouse1

joonicks:
cityhouse6

nozmajner:
Some points:
* My nitpicks from IRC still stand. This isn't a good quality model. No visible glitches, but it ever needs modifications or polish, it will be a pain. (And you didn't learned much in the way of good quality models by making it.)
* No LODs. Now this is another point where the model quality will be a hinderance. It's at about the tri count of the mid LOD of the others, but less detailed at the same time, so it's not that big of an issue. Low lods for them are about 20-40 tris, so you might want to make one.
* Why put it to another directory, then duplicate the vlastan textures? Especially for one building.
* the .mtl file is unneeded, materials are set up in the .model file.

nozmajner:
And another one I just noticed: the scaling of the window textures are different from the other buildings. Which throws the sense of scale a bit off.

impaktor:
Cool. But for what it's worth, I've never seen a model go in at first try, nozmajner can always find many things to teach, if modeller is willing, that will significantly improve the final result, plus one learns a lot.

Over at the forum, in [Ship and 3D asset development](https://pioneerspacesim.net/forum/viewforum.php?f=6&sid=d42e130d41c0fe513d730bc5afced319) you can see the process of each asset that was added.

Alternatively, upload it to SSC as a mod, if you're not willing to work on it more.

PR #4247, opened 2018-01-03
clarification of the unexploredgray patchnote

joonicks:
in the nearview mode i think unexplored has always been gray


impaktor:
I'm not sure we're that picky with changelog, but sure, right is right.

PR #4248, opened 2018-01-04
Gamestart money

joonicks:
changed the amount of startup cash from 100 to 1000.

why?-----

with only 100 money, the player is FORCED to run missions and the only kind of missions available at the start seems to be deliveries. a player who likes trading can not trade from the start, its impossible.

the impact of this change is very minimal, the player is in theory skipping ahead into the game by 1 or 2 missions.

however, I am also looking at the starter ships. having sinonatrix as a starter ship seems very rich. it costs 219000, maybe thats a mistake, but right now anyways.

I believe a better choice would be a lunar shuttle.
pumpkinseed is also an option (but already used for new hope)
xylophis ~~~ very cheap, if thats good or bad I dont know.
amphiesma is cheaper but compared to sinonatrix its not quite smaller (capacity is higher)

what I do remember from playing frontier/E:D is that you start out with the lowliest of ships, a handful of cash and an open door to the universe. Im not out to copy any game, but I think from a players perspective its a good place to start. its very motivating to have a bigger ship to aim for. currently there arent that many choices.

but. atleast, with 1000 instead of 100. the option of trading to gain cash comes up on the table. even though trading is pretty much a non option right now because of the low commodity prices.

### DIFF
--- a/data/ui/MainMenu.lua
+++ b/data/ui/MainMenu.lua
@@ -27,7 +27,7 @@ local setupPlayerSol = function ()
 	Game.player:AddEquip(misc.autopilot)
 	Game.player:AddEquip(misc.radar)
 	Game.player:AddEquip(cargo.hydrogen, 2)
-	Game.player:SetMoney(100)
+	Game.player:SetMoney(1000)
 end
 
 local setupPlayerEridani = function ()
@@ -39,7 +39,7 @@ local setupPlayerEridani = function ()
 	Game.player:AddEquip(misc.autopilot)
 	Game.player:AddEquip(misc.radar)
 	Game.player:AddEquip(cargo.hydrogen, 2)
-	Game.player:SetMoney(100)
+	Game.player:SetMoney(1000)
 end
 
 local setupPlayerBarnard = function ()
@@ -50,7 +50,7 @@ local setupPlayerBarnard = function ()
 	Game.player:AddEquip(misc.autopilot)
 	Game.player:AddEquip(misc.radar)
 	Game.player:AddEquip(cargo.hydrogen, 2)
-	Game.player:SetMoney(100)
+	Game.player:SetMoney(1000)
 end
 
 local loadGame = function (path)


nozmajner:
Ship prices, and ship stats in general need a rebalance, don't base much assumptions on them.
The three starting places are intended to provide somewhat different difficulties.
For the Barnards/Xylophis start I'd say it should be kept at 100$. So the player has to do those missions, or figure out that the scanner could be sold for enough cash to buy a drive. And the ship isn't fit for trading anyway.

fluffyfreak:
Yeah the start locations can also be thought of as "easy", "medium" and "hard" difficulty so varying the amount of money between them makes sense.

joonicks:
> So the player has to do those missions, or figure out that the scanner could be sold for enough cash to buy a drive.

This, I object to. As a completely new player, starting the game by selling startup equipment feels wrong. For someone who doesnt know anything, selling startup equipment could create more problems than it fixes.

> Yeah the start locations can also be thought of as "easy", "medium" and "hard" difficulty so varying the amount of money between them makes sense.

Personally I dont start on earth because its "easy", I start there because its the only place in the universe Im familiar with.

If the intention with the three starting positions IS to provide easy/medium/hard, the I think labeling them as such should be a thing.

clausimu:
Why don't we add a label to the starting positions then [easy, medium, hard] and adjust starting money in addition to the (already) different starting ship? I would really like to keep the hard setting in Barnard's star - that kind of game start appeals to me. But it might throw off players that don't want to dismantle their ship for survival. Players like @joonicks obviously dislike the delivery missions and would rather start trading right away. With three different starting positions we could easily cater to all those needs.

nozmajner:
Quite a few space games offer some choices for game start, where you can choose from three ships, one combat, one trading and one generally inclined one. This might work for Pioneer too:
Choose between some systems and three ships in each place. Maybe even different available ships at the starting points, like at Earth: Lunar shuttle, Sinonatrix, Varada, at Epsilon Eridani it's Pumpkinseed, Mola Mola and the Lunar shuttle here too.  I'd keep the options lower in Barnard's, like Xylophis and Varada.

I could even imagine a random start somewhere in the bubble with a randomly picked ship and a random amount of money between some set bounds. So you could end up in a very uncomfortable place or strike lucky with another start.

caligari87:
Pending a complete overhaul / rebalance of the economy and rewards system, I do think adding some extra differentiation to the starts would help in the interim.

Some paradigms I think would be good to follow:

* "Easy" should have means to jump into whatever you prefer nearly immediately. Bump starting rep somewhat, enough starting money to buy an extra passenger cabin, make sure there's a trade computer esaily available, stuff like that. 

* "Medium" would require a bit of extra work to get into what you want to do. This is the level where all the current starts seem to be, barring some mild variations which can be overcome in an hour or two of gameplay. Maybe give a *little* more cash or starting rep than already exists, but not much.

* "Hard" would basically be starting destitute. No jumpdrive, no cash, no rep, no radar, maybe even no autopilot. The "this isn't even remotely fair" skill level. 

Additionally, all starts could definitely do with some randomization of cash, rep, equipment, and location, just to make things more interesting. I like the "start floating in space" idea and possibly picking from a couple optional ships, @nozmajner

impaktor:
> Why don't we add a label to the starting positions then [easy, medium, hard]

This is planed, once the main menu moves to the new pigui system.

As for the PR, I prefer the 100 credits start for now.

fluffyfreak:
How about a compromise?
- Earth 2000
- New Hope 1000
- Barnards Star 100

Alternatively if there's objections too changing it at all then I'm of a mind to just close this PR.

clausimu:
@fluffyfreak - I would support that.

impaktor:
I think that's too much for Earth. And the problem isn't starting money but balance between mission rewards relative trading profits. And the idea is to start trading the cheaper commodities, to build up money to then afford buying the more expensive, until you have enough money to fill your hold with the most expensive / profitable per unit tonne in absolute money.

If we want to oversee the starting options, my plan was to do that when we move start menu to imgui, and then have for instance descriptions for each starting location.

If the plan is to make it easier for ourselves to try/test stuff in pioneer easier, then that could be done by adding a cheating script in `modules/`

> Alternatively if there's objections too changing it at all then I'm of a mind to just close this PR.

That was my thinking also.

fluffyfreak:
ok, closing

joonicks:
with a starting pool of 94 credits (remember, 6 is deducted before the game even starts)
you need a minimum profit of 14% on each run to break even (fuel costs, jumpfuel cost, docking fees)

if you are extremely lucky and get an absolute dream of a trade route, you can make maybe 25 credits per run.

so yeah. Id say starting the game by trading is impossible.

its super frustrating to hear people going on about "it shouldnt be easy, we shouldnt give money away, just do some missions"

just because your preference works doesnt mean everyone else MUST FOLLOW YOUR LEAD.

its not a sandbox if players are forced to do stuff they dont want to.

trading is not feasible, not because profits are unbalanced, but because the starting fund is inadequate.
once you have starter funds you can start discussing the commodity prices, but as long as the starting fund is only 94 credits, it will always be unfeasable to start the game by trading.

joonicks:
> If the plan is to make it easier for ourselves to try/test stuff in pioneer easier, then that could be done by adding a cheating script in modules/

@impaktor are you serious? do you seriously think 1000 money is enough to allow you to do anything you want in pioneer? did you even think before you wrote this?

answer me this, what can you do with 1000 money that you can not do with 100?

clausimu:
I don't think I would phrase it as violently as @joonicks :wink:, but I think he has a point here. Pioneer is a sandbox type game and with the option of several starting positions/ships, why not offer different starting funds that cater to a different playing style. I myself am more conservative and would rather limp along with just the 12 credits necessary for the first two landing fees, but why impose that on others.

Since no one is into really balancing economics + starting conditions right now, we just have to agree on a limit above which it would probably take the fun out of the game. I don't think 1000 or even 2000 is getting close to that (not having hard data to prove that, sorry @impaktor :blush:, just impression from playing). But maybe @joonicks could add a "hard", "medium", "easy" tag to the starting options even in the old gui.

Just my 2 cents. Not important enough to me to :rage: over though.

joonicks:
also consider that for any dev, "hard mode" would be easy. because we know everything about the game, all the internals, everything. for someone who just heard about the game though, pioneer would be hard. exceptionally hard with just 100 credits. 100 credits is enough for about 3 trips. if you dont make money in 3 trips, you are screwed and have to start over.

laarmen:
Nope. Pioneer dev are statistically terrible at playing the game.

joonicks:
> Nope. Pioneer dev are statistically terrible at playing the game.

show me that statistic.

fluffyfreak:
It's true, most of us are shit actually playing Pioneer. I smash into planets all the time because I'm not paying attention, overshoot, run out of funds even after I've cheated to get lots of cash and have to cheat again because I'm not playing I'm just wizzing about having fun (_& crashing into planets_).

Now I did suggest increasing the amounts, see above, but @joonicks you're super aggressive in your responses. No-one deserves that kind of anger over minor disagreements.

If you want to discuss stuff that has proven contentious please take it to the forums, discuss it there and then create an appropriate PR for it once it's been worked through.

joonicks:
Ive provided a fairly strong argument for why 100 money is inadequate at start. yet impaktor has made up his mind that no-one is allowed to change his game. I ask for reasons why 100, but they are ignored and dismissed outright. my opinion dont matter, and thats why Im agressive.

nozmajner:
I suspect increasing the starting money alone wouldn't really offer a way into/save trading I think. As far as I know, prices are almost completely random, there's no sense of logical imports/exports, and the UI isn't too good for finding trade routes either, even with the trade computer. _(Side note, maybe the trade computer could offer some kind of trade visualization in the map, like filters or a tool-tip that shows the trade info on a system.)_

With current of starter money and reputation, you can gain the money required for a cabin relatively quickly, so you can take the more lucrative taxi missions. I've just checked, and there was one 700$ courier mission at the BBS on Earth start for example, among the couple of lowly local courier runs.
From that point of view, it's only like giving thirds of the cabin money to the player in advance.
Or if you get lucky with a SAR mission, you can get some very good starting money soon.

SAR missions are few and far, and you aren't even able to take a lot of them (hold or cabin limitations for example), so I doubt you could built a career onto that alone.

Assassination, I never tried that, so I don't know how hard they are, and if I'm correct you'd need some combat rating to be able to take one, which could be quite hard to build up currently.

The other problem: ship prices. I've put them quite high when I did the last rebalance. I tried to calculate it based on stats and the price of metal alloys, so I might overdone it a bit. Especially compared to equipment prices I think. You'd need to do a lot of missions or trading to be able to buy a new ship in most cases.
But on the other hand, I think it would serve the game better, if buying a ship would be a big undertaking and decision for the player, instead of a clear route of progression towards the "bestest ship, let's graduate to the next best along the line". Instead it would be nice if the player could grow attached to the ship, kitting it out, surviving some close calls, etc, but that would need a quite deep customization/equipment and equipment quality system.

But instead of arguing and grrrrs, let's take this onto the forum, as @fluffyfreak suggested, and maybe we should try different approaches mods for example. I agree with @impaktor that changing starting money won''t alone do much in the long run. I think, instead we should rebalance the game properly: ship stats, mission rewards calculation, commodity prices/differences. (Maybe move more stuff to lua regarding trading, so it's easier to try different solutions via mods.). (Request for that: acceleration caps for ships, maybe even overrideable for combat situations)

impaktor:
> @impaktor are you serious? do you seriously think 1000 money is enough to allow you to do anything you want in pioneer? did you even think before you wrote this?

> answer me this, what can you do with 1000 money that you can not do with 100?

We do not address each other in this manner in our community. Do you really think I deserve that attitude simply because I have an opinion that is not identical to yours? You seem to not understand how this will reflect back upon yourself.

For the matter at hand: What the starting budget should be is not an absolute science with absolute right or wrong, but it's a matter of opinion, and with several people voicing in, it's not surprising there are different opinions.

By maintaining status quo, with 100 credits, the player will probably do one or two delivery missions, and see their wealth increase tenfold in a matter of tens of minutes of game play. That's a pretty good motivation, seeing payment scaling up, and seeing there might be a path to that cool upgrade they want. (Unfortunately that up-scaling doesn't hold after the initial missions from 100 to 1000. E.g. from 1,000 to 10,000.) 

There's not necessarily any reason to think the starting budget should include money to invest in cargo. You can just as well think of the starting money for covering just fuel and fees. Player will have to find a way to get the budget to start trading commodities. Compare to real life: we usually need to work (sign a contract) before earning money to start investing in stock (commodities), or open a small import/export business.

Similarly, since combat is half-broken - just like trade - you could just as well argue that the starting ships should also be equipped with heat seeking missiles, or else combat is impossible, and all combat-missions closed to the player, like you argue trade is. And then we need to include radar as well. And target scanner. And shields. And ECM. Sooner or later, the player will start with a half-finished upgraded ship, instead of a rusting bucket and a shoestring budget, which was the future direction we've discussed in the community.

> yet impaktor has made up his mind that no-one is allowed to change his game.

I've merged quite a number of your PRs the past days, thereby "changing pioneer". I'll extend an olive branch and assume the _"his game"_-part is a miraculous typo.

> I ask for reasons why 100, but they are ignored and dismissed outright. 

- By _"ignored"_ I assume you mean that I haven't replied (until now) to your rude post? Reason: I don't post when mad, as that is only counterproductive. Thus I've waited a day.
I recommend it.

- By _"dismissed outright"_ you seem to ignore it was closed only _after_ discussion, quite the opposite to _"dismissed outright"_.

> my opinion dont matter, and thats why Im agressive.

Will you react like this for every PR that's rejected after having been discussed? For both your own, and our sake, please don't post when aggressive.

PR #4250, opened 2018-01-05
Found three unused strings

joonicks:
three unused strings in core lang


impaktor:
As for the fine payed string, I believe that was back when we could pay fines remotely. That was temporarily lost when we moved from oldUI to newUI, with the intent to implement it back, but that have yet to happen, and now we're ditching newUI for pigui, so hope to be able to pay fines through the comms remotely.

joonicks:
A wise man once said "We usually don't merge 'dead code' or unused assets."

If they need the string its easy enough to put back in when there is code for it.

impaktor:
> If they need the string its easy enough to put back in when there is code for it.

Yes, it's easy for us to add or remove strings, but when adding new strings they're pushed to all our translators, so it's equivalent to putting a lot of work on other peoples' desk. Work they've already done once.

If a string is already translated, and we suspect it will be used again in the future, removing it will also remove all translated version of that string, and adding that back either requires a lot of work with transifex-API by devs, or all translators to re-translate what was already translated, but removed.

E.g. I think all our equipment have currently unused `_DESCRIPTION` strings, that were likewise lost in the newUI transiton, but will be brought back in pigui, so by the logic of removing currently unused strings, they couls also be removed.

Anyway, I'm not sure it's a good idea to remove the three fine-related strings. The others can probably go, as indeed we don't want unused dead code, or assets, or strings that will no longer be needed.

joonicks:
the three fine strings is because it moved to lua, but lua uses different strings of its own device

impaktor:
> the three fine strings is because it moved to lua,

That's true, and I assume we don't have the energy to move them through the transifex API, so OK, they can go.

PR #4251, opened 2018-01-05
Hyperdrive diversification project

joonicks:
added attributes to individual hyperdrives to allow for diversification/specialization
E.capacity.speed = the speed with which it travels through hyperspace, base number = 100 to indicate 100% normal speed. military drives travel at 160% normal speed.
E.capacity.warmup, instead of calculating warmup for all drives equally, warmup is now an explicit attribute that can be changed on a per drive basis. normal drives have pretty much exacly the same warmup, military drives have shorter warmup, 5 vs 7 seconds class 1, 10 vs 19 seconds class 9. this isnt much of an issue since any impatient player can time accelerate past the warmup.
E.stock is a value of approx how many of each of this drivetype to stock in a spacestations inventory

hyperdrive prices were changed to raise the significance of it as part of a ship. I aimed at 10% the price of the ship for class 1, then incremented prices linearly by mass. military drives have an additional element of cost, exponential to the drive class.

added one example diversified engine;
Corello Corp hyperdrive, Class 2
"better quality, lighter, 20% shorter traveling time, shorter warmup, but quality costs"
normal class 2 = 17,000 money, corello class 2 = 51,000 money and higher tech, small stocks.


nozmajner:
Nice!
Question: how fuel consumption is handled currently? Can't recall it. Maybe that could be another stat: ly/t. So there could be very efficient, but slower drives for example, and inefficient, but brutally fast.

joonicks:
so far, fuel calculations is untouched, but I had the same idea, twice the fuel, up the speed

nozmajner:
Hmm, want to sit down and talk/think through this?

joonicks:
if you have drives with different specs maybe you dont need to allow different classes to be mounted on a ship. maybe ships are limited to one size only and have to search for better performing drives if they want more range.

impaktor:
> Hmm, want to sit down and talk/think through this?


My general opionion: yes, this needs to be discussed very thoroughly, as this is not something we want to merge willy nilly. Looks like this changes all kinds of things, that we have been carefully discussing within the community, for a long time. Given how much @nozmajner has on his plate now, I don't see this going anywhere this month.

Random thought: having faction names incorporated in the ship drive name, raises questions about translation, as the faction-names outside of this currently can't be translated.

fluffyfreak:
@joonicks you seem to have merge in a bunch of other branchs into this PR?

PR #4252, opened 2018-01-05
Hyperspace jump range

joonicks:
this formula brings hyperspace ranges down below 30 lightyears in most cases, it ignores ship cargo and equipment and calculates range based on "volume" to give more consistent ranges making it easier to balance ship stats against eachother.

adjusted the pre-set hyperdrive class for some ships where it was badly set before

amphiesma/class 3/9t cargo = 22.5ly range
wave/class 3/1t cargo = 27.7ly
mola mola/class 4/24t cargo = 21.8ly
sinonatrix/class 3/6t cargo = 21.2ly


impaktor:
This is the direction I want to move in, regarding jump range, and reducing complexity. 

1. I think you and @nozmajner discussed getting the ship-volume from the model files, (incorporating it into the ship.json files) and using that?

2. I like to see charts and statistics in general to get a better overview what the change in a PR does, especially when it's a systematic / game mechanics chaning feature. For this PR, I'm thinking a list of all ships and ranges for all drives. I assume it should be trivial, and very useful for future use, if you could just do a nested for-loop in lua over all ships and all engine types each ship can fit and print resulting range to command line, and post output here? (and post code might be good too, so people can play around with stats now, or in future if they want)

data/modules/print_stat.lua:

```lua
for <all_ships> do
    for <engine_types> do
        if <engine_ successfully_fitted_to_ship> then
			print(<ship_name>, <engine_class>, <range>)
		end
	end
end
```

nozmajner:
I was thinking about it a bit more yesterday. Maybe instead of volume, we could use surface area. Or a bounding box/sphere volume. I can give the volumes of each from Blender.

joonicks:
@impaktor yes I can do that

@nozmajner surface area of the shield maybe. some ship models are not smooth and might get very unfavorable values (pumpkinseed?)

joonicks:
@impaktor post the script or data here or where or...?

nozmajner:
Yeah, I was thinking about the shield areas too. Can do.

impaktor:
> Maybe instead of volume, we could use surface area. 

What would the benefits be, over volume? Then we want ships to be more spherical, and thus similar-looking, and not have any longish-ships? Even more extreme: depend on area (or volume) of the smallest sphere containing the ship (I'm not suggesting this, just mentioning it). I liked the volume thinking, it will let modellers still be free to shape the ship how ever they want (but still be believable, and take acceleration and such).

> post the script or data here or where or...?

Yeah. to both. Unless you have some other preference. You can play around with the [markdown](https://guides.github.com/features/mastering-markdown/),  if you like.

nozmajner:
One benefit of shield surface area is that it won't go up as steep as volume with size, so the range of larger ships won't go down as fast.

impaktor:
If the reason is to have less sensitive dependence on ship size, then I'd go with square root of volume or some other fractional power of volume. Just thinking, that then we get dependence on ship size, but not ship model shape, and we can tweek the exponent of `Volume^gamma`, with `gamma < 1`

nozmajner:
Sounds reasonable. I can give any of the data anyway. I should already have volumes of all ships, but I'd have to get the areas.

joonicks:
an option is to take the volume of the ship shield bubble instead of the volume of the ship itself, sort of like saying "my computer cant calculate every paint flake of space around the ship that I want sucked into the wormhole so ill settle for some excess and suck in everything inside the shield",

it would suck to hyperjump only to find the tip of your antenna didnt make it so now you dont have comms...

joonicks:
I had a thought. it seems a bit funny that, within a nanosecond of jumping into hyperspace, X tonnes of hydrogen simply vanishes from your cargo hold. what happens? how is hydrogen stored in the cargohold and how is it transferred to the hyperdrive so quickly?

then I thought, what if the hyperdrive slowly consumes hydrogen, not from the cargohold, but from the main fuel tank, like, slowly during the countdown.

if you aborted the jump, you would lose the fuel already spent.

but the good thing is, the size of the main fuel tank would limit how big a hyperdrive you could fit. if the tank is only 5t, you cant fit a class 3 because it needs 9t for a max jump.

nozmajner:
Hydrogen storage is a sour point in general, if we look it from realism pov: the density of liquid H2 is like 70kg/m3. There must be some very advanced storage method in those ships, that can squish it to the currently used water density. And if it can be stored that nicely, one could also imagine that the storage units could be hooked up to the drive somehow.

Making the hyperdrives to draw hydrogen from the tanks seem a bit strange, but who knows. I mean it would surely eat into the deltaV reserves, so you would have to take that into account. You'd have to bring along additional hydrogen in your hold, so the freeing up the hold from jump fuel storage would be a moot point. Anyhow, it should have HUD feedback on the deltaV bar, so the player can make an informed decision.
But on the other hand, the tanks are already there, so it doesn't make much sense that the drive can't draw from them.
But what would happen to military fuel/drives then?

Another solution could be to have special hyperdrive fule cells or something: There's a given amount of slots for these, like slots in the drive. _(not sure what would be a good reason to limit their number though. Maybe a drive could only use a certain amount of them, (depending on class and tech maybe) directly connected to it, because you can't just plug them into an extension cord due to the high power levels or something, or they become part of the "emitter" of the drive_). And a jump would spend a number of them depending on distance, etc, and you could replace them from your hold after you spent them, if you need extra jumps. Could even be fairly compact, so one could load up quite a few of them, but could only use the amount the drive is capable of, so no excessive jump ranges. 
And the fuel scoop, or another equipment could be able to refill those cells in the field, or you could buy them at stations. But if you refilled them you couldn't retrieve the hydrogen used for refilling. (_don't know what could be a reason for it, maybe it stores power in a special way, so the drive can draw it at the speed it needs to, but that's only possible in a non-reversible way_). So you could convert propellant for hyperdrive fuel, but not back.
This way there could even be more types of fuel cells, for special drives, like military ones. These could turn the cells radioactive for example, so you would not be able to refill them. And the availability of the different cell types could fluctuate, like there's a more expensive type of drive which has nice range and speed, but it is hard to find the proper type of cells to it at the fringes, so you need to stock up. Especially if there's a chance for the drive to destroy a cell after jump, or they could only last a given number of jump/refill cycles before going bad. Which might make the player to make a decision: either I stick to my fancy drive and risk that I run out of cells out there, or buy a cheaper drive with easier to obtain cells. Especially if the fancy drive won't sell at a good price, because the cell availability is bad at those regions.

Heh, I guess I'm overthinking again :)

joonicks:
hmm, what about having "fuel cells" kind of like batteries, you can charge them in ship in space but it takes time. you charge them by feeding them fuel. then the drive has "battery holders" where you put N of these cells that can provide the energy for the jump

you can have spare cells, precharged, for faster consecutive jumps... different cells would be for different kind of fuel... hydrogen, military fuel, water... fertilizer... rubbish.. i even thought about making a drive run on radioactives

the deltav allowance is a pet peeve of mine...

nozmajner:
Mine too. DeltaV is like one of the strongest point of Pioneer.

joonicks:
my peeve is that I cant go to a far out planet for a look, and then back, at full speed, without refuelling

nozmajner:
Mine is the need to be on the lookout for your deltaV/fuel levels if you don't want to end up wooshing out on an escape trajectory. Having deltaV is one point which sets apart Pioneer from almost all space games, especially among the more adventure inclined ones. And it's like the most important thing for spaceflight.

impaktor:
> what if the hyperdrive slowly consumes hydrogen

Not sure we need to fiddle around with this. Keep it simple, and don't add in more complexity than needed. (My general point:) Especially considering our existing complexity is unbalanced, and adding more of it will just make it even harder to balance the game to be fun.

Specifically: regarding fuel use in hyperspace, I think it's works fine the way it is now. You can think of it like you need to convert the fuel to energy to rip open the entrance to hyperspace, once there, you're "falling" back into normal space, and all your fuel (for the jump) is spent.

If anything, I think the fuel could be consumed during the count down, as that's where we charge up energy for the drive. Also, it's more likely the player will want to abort a countdown, than abort an ongoing hyperspace jump.

joonicks:
> If anything, I think the fuel could be consumed during the count down, as that's where we charge up energy for the drive. 

thats what I was trying to say

clausimu:
Once this is merged (or agreed on) we may need to go through all types of missions to adjust them to the new average jump range.

joonicks:
assassination maxrange = 30
cargorun = 15
packagedelivery = 30
searchandrescue = 30
taxi = 40

fluffyfreak:
@nozmajner & @joonicks I don't quite understand how this one uses the volume of the ship instead of mass? It looks like it still uses the mass?

However whilst I don't quite grasp the changes you both do seem to think it will improve the situation, is it therefore ready to merge?

fluffyfreak:
@joonicks any reason to close this? I was under the impression that it was almost ready to merge?

PR #4255, opened 2018-01-05
Change lua_pushinteger for lua_pushnumber to preserve value of seed

### LINKED ISSUE #4249: Earth tech certificate differs in 32-bit vs. 64-bit compiles
When compiling pioneer in visual studio as Release/Win32, everything is normal, but if you compile it as release/x64, London, earth, gets downgraded to a level 1 tech certificate.

previously on #4238 it was discussed and presumed that pioneer should be platform independent. this issue points out that its not.

fluffyfreak:
Fix #4249 by changing lua_pushinteger for lua_pushnumber to preserve value of seed

### DIFF
--- a/src/LuaBody.cpp
+++ b/src/LuaBody.cpp
@@ -71,7 +71,7 @@ static int l_body_attr_seed(lua_State *l)
 	const SystemBody *sbody = b->GetSystemBody();
 	assert(sbody);
 
-	lua_pushinteger(l, sbody->GetSeed());
+	lua_pushnumber(l, sbody->GetSeed());
 	return 1;
 }
 


jaj22:
Tested with and without #4257 on MSVC x86 and x64, works fine.

fluffyfreak:
Great, thanks @jaj22

impaktor:
I don't dare merge these two PRs, so I leave it up to you guys who know better what you're doing, and who have tested it.

fluffyfreak:
@impaktor since this will affect the tech levels of stations I am wondering what to do about them myself...

impaktor:
I don't see that as much of a problem. We're in eternal alpha! 

So this will only suck for those who have mapped out exactly which stations they like due to military tech level, or some other less common level, I assume. I think most players (who notice) will understand that this is the cost of upgrading their game. It's not 1:1.

PR #4256, opened 2018-01-06
access all(?) systembody attributes in lua

joonicks:
to be used when systemview moves to lua, and for my mysterious explorer mod...

new attributes accessible:

metallicity
volatileGas
atmosOxidizing
volatileLiquid
volatileIces
volcanicity
life
hasRings

pretty much all(?) attributes you need to determine visual and textual descriptions of system bodies

### DIFF
--- a/src/LuaSystemBody.cpp
+++ b/src/LuaSystemBody.cpp
@@ -399,6 +399,174 @@ static int l_sbody_attr_average_temp(lua_State *l)
 }
 
 /*
+* Attribute: metallicity
+*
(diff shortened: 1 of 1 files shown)


jaj22:
Looks uncontroversial to me. Any objections to merging?

impaktor:
> Any objections to merging?

No, not really. In general, there are infinite C++-side functions that are not exposed to lua, and we normally only expose those that we think are needed for missions/modules, or that people working on modules need, and thus request. I assume this will be used at some time, if we expose them.

PR #4257, opened 2018-01-07
Some hash_random related fixes

johnbartholomew:
Warning: This will change some procedurally generated stuff due to changing hash_random output and changing Rand seeding.

Tries to make hash_random a little less broken, and stop using it to seed another RNG. Also restricts the output range to 32-bit (signed or unsigned)... though there's a good chance that code is still buggy as I didn't test it thoroughly. Also adds comments explaining the numbers in the hash_random random-double code. Also makes it an error to pass nil as the input to hash_random.

I have not tested the Lua side changes. I have opened up the game and checked that Rand:New('some string') doesn't completely break and that calling util.hash_random rejects some out-of-range values.

### DIFF
--- a/data/libs/SpaceStation.lua
+++ b/data/libs/SpaceStation.lua
@@ -26,7 +26,7 @@ local l = Lang.GetResource("ui-core")
 
 function SpaceStation:Constructor()
 	-- Use a variation of the space station seed itself to ensure consistency
-	local rand = Rand.New(util.hash_random(self.seed .. '-techLevel', 2^31)-1)
+	local rand = Rand.New(self.seed .. '-techLevel')
 	local techLevel = rand:Integer(1, 6) + rand:Integer(0,6)
 	self:setprop("techLevel", techLevel)
 end
--- a/data/ui/StationView/Police.lua
+++ b/data/ui/StationView/Police.lua
@@ -22,7 +22,7 @@ local police = function (tab)
 
 	local station = Game.player:GetDockedWith()
 
-	local rand = Rand.New(util.hash_random(station.seed .. '-police', 2^31-1) - 1)
+	local rand = Rand.New(station.seed .. '-police')
 	local face = InfoFace.New(Character.New({
 		title  = l.CONSTABLE,
 		armour = true,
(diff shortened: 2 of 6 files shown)


johnbartholomew:
For #4249, but not intended to be a 'fix' as such (#4255 is still needed anyway)

jaj22:
**l_rand_new** looks good, although it's more transparent to use lua_tonumber and do the integer conversion explicitly rather than using lua_tounsigned. Where there are potential limit/type-size issues, it's probably best to avoid lua integer functions entirely. Lua storage is all doubles internally.

**l_hash_random:**
1. Probably need lua_pushnumber rather than lua_pushunsigned to return correct results for negative ranges. As above, I'd recommend using it anyway. Note that this changes the true range limits to +/-2^53, although INT32_MIN to UINT32_MAX are sensible enough for the specified limits.
2. Range checks slightly overkill after n>m. Only need to check m >= min and n <= max.
3. Also don't need the check on n-m+1 if using lua_pushnumber.
4. Should probably have hashA on the bottom of the register if it's the better one.
5. Do people really want the two-parameter version with the implicit m=1? Looks like a vehicle for bad assumptions to me.

johnbartholomew:
I've fixed various things that were broken in my first attempt (particularly in the LuaRand change which was *completely* broken). Commits updated (fixes squashed in).

Thanks for the review!

> Where there are potential limit/type-size issues, it's probably best to avoid lua integer functions entirely. Lua storage is all doubles internally.

Yeah... this changes with Lua 5.3 which adds actual integers (probably 64-bit though it depends on settings in luaconf.h) to the language. But we're still using Lua 5.2 :-)

> Probably need lua_pushnumber rather than lua_pushunsigned to return correct results [...]

Done.

> Range checks slightly overkill after n>m. Only need to check m >= min and n <= max.

Changed the range checks to allow either both inputs in range for signed 32-bit or both in range for unsigned 32-bit which is what I originally intended but then messed up.

> Also don't need the check on n-m+1 if using lua_pushnumber.

Done.

> Should probably have hashA on the bottom of the register if it's the better one.

Done.

> Do people really want the two-parameter version with the implicit m=1?

I originally included it because Lua defaults to 1-based indexing for arrays (not that Lua actually *has* arrays, but tables get used like arrays) and I was expecting hash_random to be used for pick-an-item-from-an-array-of-options. But keeping it explicit is probably better, so sure, let's drop the two-parameter version.

jaj22:
Ok, code looks good. I stepped through #4257 + #4255 with x86 and x64 builds and they work as expected. Also ran some hash_random tests with the console, results look fine.

PR #4259, opened 2018-01-07
no radar in hyperspace

joonicks:
very simple patch


impaktor:
@nozmajner thoughts?

nozmajner:
Can't check it right now. Is it hiding the radars during hyperjump? Since we already hide the HUD, it would make sense to hide it as well.

impaktor:
@nozmajner yeah, check the code, I suspect you can render its effect in the brain.

nozmajner:
:D

PR #4262, opened 2018-01-07
Updatefuel code duplication

joonicks:
offending lines exist in UpdateFuel() already.
the exact same lines.

```
void Ship::UpdateFuel(const float timeStep )
{
	GetPropulsion()->UpdateFuel( timeStep );
	UpdateFuelStats();
	Properties().Set("fuel", GetFuel()*100); // XXX to match SetFuelPercent

	if ( GetPropulsion()->IsFuelStateChanged() )
		LuaEvent::Queue("onShipFuelChanged", this, EnumStrings::GetString("ShipFuelStatus", GetPropulsion()->GetFuelState()));
}
```

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -865,9 +865,6 @@ void Ship::TimeStepUpdate(const float timeStep)
 	// fuel use decreases mass, so do this as the last thing in the frame
 	UpdateFuel( timeStep );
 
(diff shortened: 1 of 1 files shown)


fluffyfreak:
Ah so this is deduplication? Right gotcha

Issue #4136, opened 2017-09-03
Appears planar and azimuthal radar at same time


manolollr:
![planar_azimuthal_radar](https://user-images.githubusercontent.com/8089376/30004560-263c374c-90d2-11e7-9101-3a5a1eea081c.png)

Steps to reproduce:
1. Load or start a new game.
2. Set azimuthal radar.
3. Exit the game.
4. Load or start a new game.

PR #4265, opened 2018-01-08
planar and azimuthal radar at same time #4136

joonicks:
takes care of this bug. todo: radar type preference is not saved in savegames.

### DIFF
--- a/data/pigui/modules/radar.lua
+++ b/data/pigui/modules/radar.lua
@@ -154,6 +154,15 @@ Event.Register('onChangeMFD', function(selected)
 								 end
 end)
 
+Event.Register("onGameStart", function ()
+	if selected == "radar" then
+		shouldDisplay2DRadar = true;
+		Game.SetRadarVisible(false)
+	else
+		shouldDisplay2DRadar = false;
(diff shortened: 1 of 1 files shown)


impaktor:
It's good if you make use of [auto-closing](https://help.github.com/articles/closing-issues-using-keywords/) either from your first PR-comment above, or the commit message itself. I've edited it now, but good to use for future PRs.

joonicks:
this is ready to merge

data/pigui/modules/radar.lua
```diff
@@ -17,7 +17,7 @@ local icons = ui.theme.icons
 local MAX_RADAR_SIZE = 1000000000
 local MIN_RADAR_SIZE = 1000
 
-local shouldDisplay2DRadar = false
+local shouldDisplay2DRadar
```
fluffyfreak: I would have left this with a safe default of `false`, since the later cases are `elseif` not just `else` you could in theory end up with it being `nil`... I think

joonicks: Ive changed the later ones to "if then else end", no more elseif, it defaults to old radar unless new is activated

Issue #2625, opened 2014-01-05
possible bug in TradeShips.lua file?


impaktor:
I've been looking in data/modules/TradeShips.lua

Or more specifically line 341-359 this is looking like a bug to me, that table.insert is only used in two of the ifs:

```
local prices = Game.system:GetCommodityBasePriceAlterations()
local import_score, export_score = 0, 0
imports, exports = {}, {}
for k,v in pairs(prices) do
    if k ~= 'RUBBISH' and k ~= 'RADIOACTIVES' and Game.system:IsCommodityLegal(k) then
        -- values from SystemInfoView::UpdateEconomyTab
        if      v > 10  then
            import_score = import_score + 2
        elseif  v > 2   then
            import_score = import_score + 1
            table.insert(imports, k)
        elseif  v < -10 then
            export_score = export_score + 2
        elseif  v < -2  then
            export_score = export_score + 1
            table.insert(exports, k)
        end
    end
end
```

then there's and 878-887, which I don't understand yet, when comparing with the code above, but will at some point.

```
local prices = Game.system:GetCommodityBasePriceAlterations()
for k,v in pairs(prices) do
    if k ~= 'RUBBISH' and k ~= 'RADIOACTIVES' and Game.system:IsCommodityLegal(k) then
        if v > 2 then
            table.insert(imports, k)
        elseif v < -2 then
            table.insert(exports, k)
        end
    end
end
```

joonicks:
fixed by #4268

richardpl:
Is this still relevant?

PR #4268, opened 2018-01-09
Tradeships getSystemAndJump checking (crash bug)

### LINKED ISSUE #4253: Tradeships module causes crash
### Observed behaviour
Jumping from Epsilon Eridani to Kaptain's Star. Game crashed when entering the new system with the following error message:

error: [string "[T] @modules/TradeShips.lua"]:310: attempt to index field 'starport' (a nil value)
stack traceback:
        [string "[T] @modules/TradeShips.lua"]:310: in function 'getSystemAndJump'
        [string "[T] @modules/TradeShips.lua"]:592: in function 'cb'

Is this related to #4220?

### Steps to reproduce
No idea. Seemed random. I jumped Epsilon Eridani => Kaptain's Star several times to test a new module I'm working on. Never had a problem except this one time.

pioneer built from git on 01/03/18 (should include #4220 fix, right?)
OS: openSUSE 43.2

joonicks:
if a tradeship sets course for an uninhabited system and arrives, while the player goes to the same system, once both are in system, the game will crash.

in tradeships.lua, there is a codepath that takes ships to getSystemAndJump with trade_ships[ship].starport set to nil. getSystemAndJump dereferences that value without first checking and so it crashes.

also, another bug. tradeships would only trade in minor imports and minor exports because of a poorly written if then elseif chain. its fixed here.

plus a bit of reformatting and optimising.
  
fixes #4253

### DIFF
--- a/data/modules/TradeShips.lua
+++ b/data/modules/TradeShips.lua
@@ -342,14 +342,19 @@ local getAcceptableShips = function ()
 end
 
 local spawnInitialShips = function (game_start)
+	-- quicker checks first
+	-- dont spawn tradeships in unpopulated systems
+	local population = Game.system.population
+	if population == 0 then return nil end
+
 	-- check if the current system can be traded in
 	starports = Space.GetBodies(function (body) return body.superType == 'STARPORT' end)
 	if #starports == 0 then return nil end
 	vacuum_starports = Space.GetBodies(function (body)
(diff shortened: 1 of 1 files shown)


impaktor:
Might also fix #2625, I assume?

joonicks:
fixes #2625

fluffyfreak:
Squashed and merged

PR #4269, opened 2018-01-09
Flatten terrain within Cities - WIP

fluffyfreak:
This is an old idea and variation on the initial implementation (#1476) which has long ago fallen out of date.

It's purpose is to flatten the terrain within the radius of a city then falloff/blend into the surrounding terrain and I thought it might be of interest to @joonicks 

This can look ok in some places:
screenshot-20180109-233047

and then very much not in others:
screenshot-20180109-233320
screenshot-20180109-233446

-Change GetHeight to GetHeightInner for implementations.
-Add ApplySimpleHeightRegions call it for all uses of GetHeight.
-Initialise cities within ctor of BaseSphere.

### DIFF
--- a/src/BaseSphere.cpp
+++ b/src/BaseSphere.cpp
@@ -7,8 +7,50 @@
 #include "GeoSphere.h"
 #include "GasGiant.h"
 #include "graphics/Material.h"
+#include "CityOnPlanet.h"
 
-BaseSphere::BaseSphere(const SystemBody *body) : m_sbody(body), m_terrain(Terrain::InstanceTerrain(body)) {}
+BaseSphere::BaseSphere(const SystemBody *body) : m_sbody(body), m_terrain(Terrain::InstanceTerrain(body))
+{
+	std::vector<vector3d> m_positions;
+	std::vector<RegionType> m_regionTypes;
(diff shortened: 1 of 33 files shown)


joonicks:
yes. very interesting. there is definatly a use for it.

being tired and lazy, ill just ask you; can it be modified to flatten only parts, like the area near the starport, instead of the whole city?

impaktor:
I suspect something like this is needed to make city generation work within reasonable limits. (Not that I have looked into city gen strategies.)

fluffyfreak:
Yes it could just flatten in the centre, or do a lot less flattening overall etc.
Probably discuss in the forum https://pioneerspacesim.net/forum/viewtopic.php?f=4&t=353&p=5405#p5405

The-EG:
It looks like the flattening is done by raising everything up to the maximum height in the area (or something similar). Would it be possible to use the minimum or an average? I think if it were recessed a bit, it would look less strange than sticking out of the side of the moon/asteroid/rock like that.

impaktor:
yeah, average sounds good, or median, for even more stability. I assume there is then a grid of points x,y that one tests to see which is highest/lowest.

PR #4272, opened 2018-01-10
Asteroid mining results controlled by lua

joonicks:
this code breaks out to lua the logic that detemines what you get when you blast away at an asteroid with a mining laser

original code spawned water, this code spawns coal ore instead

this is the bare code, my startup idea was to make a small chance of something rare spawning, like "strange crystal formation" which you could turn in somewhere for a reward, however if that should be integrated into the game or somehow split out as a module, I dont know.

also the name "SpawnMiningContainer" might be debated...though it does spawn a container with the returnvalue commodity.

### DIFF
--- a/data/modules/Mining.lua
+++ b/data/modules/Mining.lua
@@ -1,0 +1,23 @@
+local Player = import_core("Player")
+local Engine = import("Engine")
+local Equipment = import("Equipment")
+
+function Player:SpawnMiningContainer(body)
+	-- this function is called from C++ when a mining laser shot hits the surface of an asteroid body and chips off a peice
+	-- body is a valid SystemBody object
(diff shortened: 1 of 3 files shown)


joonicks:
screenshot-20180110-184301

fluffyfreak:
I'm a little confused by the statement 

> original code spawned water, this code spawns coal ore instead

Does that apply for all asteroids?
Looking at the code it looks like it should behave the same as the original but migrated to Lua.

joonicks:
this code runs when any mining blast hits any asteroid. the lua returns the commodity to place in the container that is created.

fluffyfreak:
Ok thanks, just wanted to check.

PR #4275, opened 2018-01-11
fixes #4274 - Segfault on collision

### LINKED ISSUE #4274: Segfault when flying around starport
To replicate: Fly around New Hope starport for a while (or probably any terrestrial starport) and crash into the ground

I am tracking the latest git version. It seems like Space::hitCallback is being passed a CollisionContact struct which contains a null pointer to po2, causing a segfault. I have attached a full backtrace report from GDB.

I have played for several hours so far, this is the only reason the game has crashed for me. I traced the bad variable, and from what I can tell, the offending object's m_data (defined in Geom.h) is an unallocated null void pointer. It appears as if there are two, exactly two places in the code where this variable is set: ModelBody.cpp, lines 149 and 161. From what I can tell, this void pointer chicanery is intended to give the ability to implement some sort of "generic" mesh. It seems to work well enough... until it doesn't.

At any rate, I have applied a quick fix which I will commit shortly. I lost the trail after ModelBody.cpp. I would rather have the crashing fixed and leave the task of finding the deeper cause to someone with more C++ chops than I have. It seems to fix things just fine if I only perform the collision callback in Geom.cpp if both objects to be collided have allocated m_data pointers. 

[pioneer_full_backtrace.txt](https://github.com/pioneerspacesim/pioneer/files/1623828/pioneer_full_backtrace.txt)

CampNowhere:
Basically, I already described this in the issue report. 

Sometimes, collision logic is attempted on an object with a null pointer instead of a pointer to m_data, which causes a sefgault.

### DIFF
--- a/src/collider/Geom.cpp
+++ b/src/collider/Geom.cpp
@@ -227,7 +227,11 @@ void Geom::CollideEdgesTris(int &maxContacts, const BVHNode *edgeNode, const mat
 			// contact geomFlag is bitwise OR of triangle's and edge's flags
 			contact.geomFlag = b->m_geomtree->GetTriFlag(isect.triIdx) |
 				edges[ edgeNode->triIndicesStart[i] ].triFlag;
-			callback(&contact);
+			//Don't attempt collide callback if either m_data pointer is NULL
+			if((m_data != 0) && (b->m_data != 0))
+			{			
+				callback(&contact);
+			}
 			if (--maxContacts <= 0) return;
 		}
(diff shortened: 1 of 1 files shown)


fluffyfreak:
Thanks for this PR @CampNowhere but with the other fix for this I'm going to close this one.

CampNowhere:
What was the other fix?

jaj22:
Fix was #4284, referenced above.

PR #4279, opened 2018-01-14
Small display changes

CampNowhere:
Added registration number to ship info screen
Added new icon for systems management, not used yet
Changed color param so orbits are visible in system view

 Changes to be committed:
	modified:   data/icons/icons.svg
	modified:   data/pigui/themes/default.lua
	modified:   data/ui/InfoView/ShipInfo.lua
	modified:   src/SystemView.cpp


impaktor:
Do you have a screenshot?

CampNowhere:
Actually, this pull request isn't quite ready, I need to make sure the labels are correct.

Quick question: What is the preferred way to handle things if I'm adding a new label? It's easy enough to add an English one in the en.json file, but from what I'm seeing in the game logic, it looks like the game will crash if someone is using a different language and the label is not defined for that language.

Is there a preferred way to add a label to every language file in a directory? Because I wrote a Perl script to do just that and I'll share it if it'd be useful.

In the meantime, here is a preview of the newly visible orbits.

display1
display2

nozmajner:
Nice!
Translations handled automagically, you just need to add it to the english, the rest goes trough Transifex  daily, and gets back to the other lang files (and the string stays english if there's no translation yet) 
Aren't those orbits too bright? I see you removed the fading too (which never really worked as intended)

What are your plans with the new icon?
Note, that there's an UI migration in progress, moving everything to imgui, so any changes on the two old UI systems will eventually be moot.

CampNowhere:
Uploaded patch to add label for registration number.

The orbits might be a bit bright, but they're visible. It was almost impossible to see your projected orbit using the transfer planner before this patch. 

display-3

fluffyfreak:
@nozmajner this looks good to me, is there anything else that needs doing for getting new strings?
Or is this good to merge?

fluffyfreak:
@CampNowhere what name would you like to be known as within our `AUTHORS.txt` file?
Everyone who's got PR merged gets to go in it :)

CampNowhere:
Jon Jenkins is fine! I look forward to committing many more patches.

PR #4282, opened 2018-01-17
Make the black hole system navigatable

WKFO:
Trying to fix sth nobody even knows about in the game :) (Who else would spend months trying to reach the center of the galaxy deep in unexplored sectors?)

The current black hole is much more massive than it should be. It is impossible to navigate in the system. None of the ships have enough Delta-V (but somehow we have stations orbiting it). Also, scientifically inaccurate (Well, I mean, our real-life info about that region of space isn't that accurate, but still... it is better to have it this way (https://en.wikipedia.org/wiki/Sagittarius_A*)).

Maybe we can name the system too. Plain "Black Hole" sounds wrong, as if we only have a single black hole in the galax... Oh.


fluffyfreak:
Feel free to rename it too 😄

impaktor:
I'd say +1 on renaming it, if the specs like coordinates and mass are taken from astro data.

Could also add more black holes, e.g. [Cygnus_X-1](https://en.wikipedia.org/wiki/Cygnus_X-1)

> ### Sagittarius A in popular culture
> In the 2014 space-sim videogame Elite: Dangerous players are able to travel to Sagittarius A*, with an achievement tied to it in the Xbox One version of the game.

WKFO:
Hah, after we can project nearby stars to the skybox, it would also be very nice to have ambient lighting increase with the number of stars nearby. The black hole is black, you know... You literally can't see planets.
Same with some brown dwarves.

Just saying.

data/systems/custom/02_black_hole.lua
```diff
@@ -1,15 +1,17 @@
...
 -- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
-local s = CustomSystem:new('Black Hole', { 'STAR_SM_BH' })
+local s = CustomSystem:new('Sagittarius A*', { 'STAR_SM_BH' })
 	:govtype('CISLIBDEM')
-	:short_desc('The End of the line.')
-	:long_desc([[This is what holds it all together baby, Yeah!]])
+	:short_desc('A supermassive black hole.')
+	:long_desc([[The supermassive black hole which holds all the Milky Way Galaxy together.
+	Astronomers were aware of the existance of this black hole since the 21st ceuntry, but the first expedition to the system did not depart until 2827 when the famous explorer Tom Morton decided to attempt such a dangerous journey. After a few exploration missions, humans established permanent research bases in the system for further observations.
```
mooskagh: I'm pretty bad with English punctuation, but it feels like the comma is needed before "when", and also before "which" in the previous line.

PR #4283, opened 2018-01-19
Route planner optimisations

fluffyfreak:
The route planner (@The-EG) is great but it can be a little bit slow when you're plotting a route over hundreds of light years.

- [x] Avoid calls to GetSector by testing is the sectors in question can be reached by the player
- [x] Limit the total Sectors added as nodes using a tighter bounding volume
- [x] Limit the nodes to within a point-point cylinder/range

There's a few other minor optimisations that snuck in whilst I was doing stuff too.

### DIFF
--- a/src/SectorView.cpp
+++ b/src/SectorView.cpp
@@ -603,31 +603,12 @@ void SectorView::AutoRoute(std::vector<SystemPath> &outRoute, const SystemPath &
 	// nodes[0] is always start
 	std::vector<SystemPath> nodes;
 	nodes.push_back(start);
-#if 0
-	// go sector by sector for sec_dist sectors and add systems
-	// if they are within 110% of dist of both start and target
-	for (Sint32 sx = -sec_dist; sx <= sec_dist; sx++) {
-		for (Sint32 sy = -sec_dist; sy <= sec_dist; sy++) {
-			for (Sint32 sz = -sec_dist; sz < sec_dist; sz++) {
-				const SystemPath sec_path = SystemPath(start.sectorX + sx, start.sectorY + sy, start.sectorZ + sz);
-				RefCountedPtr<const Sector> sec = m_galaxy->GetSector(sec_path);
-				for (std::vector<Sector::System>::size_type s = 0; s < sec->m_systems.size(); s++) {
-					if (start.IsSameSystem(sec->m_systems[s].GetPath()))
-						continue; // start is already nodes[0]
 
-					if (Sector::DistanceBetween(start_sec, start.systemIndex, sec, sec->m_systems[s].idx) <= dist * 1.10 &&
-						Sector::DistanceBetween(target_sec, target.systemIndex, sec, sec->m_systems[s].idx) <= dist * 1.10)
-					{
-						nodes.push_back(sec->m_systems[s].GetPath());
-					}
-				}
-			}
-		}
-	}
-#else
 	const Sint32 minX = std::min(start.sectorX, target.sectorX)-2, maxX = std::max(start.sectorX, target.sectorX)+2;
 	const Sint32 minY = std::min(start.sectorY, target.sectorY)-2, maxY = std::max(start.sectorY, target.sectorY)+2;
 	const Sint32 minZ = std::min(start.sectorZ, target.sectorZ)-2, maxZ = std::max(start.sectorZ, target.sectorZ)+2;
+	const vector3f start_pos = start_sec->m_systems[start.systemIndex].GetFullPosition();
+	const vector3f target_pos = target_sec->m_systems[target.systemIndex].GetFullPosition();
 
 	// go sector by sector for sec_dist sectors and add systems
 	// if they are within 110% of dist of both start and target
@@ -640,8 +621,12 @@ void SectorView::AutoRoute(std::vector<SystemPath> &outRoute, const SystemPath &
 					if (start.IsSameSystem(sec->m_systems[s].GetPath()))
 						continue; // start is already nodes[0]
 
+					bool isWithinLineSegment = false;
+					const float lineDist = MathUtil::DistanceFromLine(start_pos, target_pos, sec->m_systems[s].GetFullPosition(), isWithinLineSegment);
+
 					if (Sector::DistanceBetween(start_sec, start.systemIndex, sec, sec->m_systems[s].idx) <= dist * 1.10 &&
-						Sector::DistanceBetween(target_sec, target.systemIndex, sec, sec->m_systems[s].idx) <= dist * 1.10)
+						Sector::DistanceBetween(target_sec, target.systemIndex, sec, sec->m_systems[s].idx) <= dist * 1.10 &&
+						/*isWithinLineSegment &&*/ lineDist<(Sector::SIZE*3))
 					{
 						nodes.push_back(sec->m_systems[s].GetPath());
 					}
@@ -649,7 +634,6 @@ void SectorView::AutoRoute(std::vector<SystemPath> &outRoute, const SystemPath &
 			}
 		}
 	}
-#endif
 	Output("SectorView::AutoRoute, nodes to search = %lu\n", nodes.size());
 
 	// setup inital values and set everything as unvisited
--- a/src/SectorView.cpp
+++ b/src/SectorView.cpp
@@ -25,6 +25,7 @@
 #include "GameSaveError.h"
 #include <algorithm>
 #include <sstream>
+#include <unordered_set>
 #include <SDL_stdinc.h>
 
 using namespace Graphics;
@@ -586,7 +587,7 @@ std::vector<SystemPath> SectorView::GetRoute()
 	return m_route;
 }
 
-void SectorView::AutoRoute(std::vector<SystemPath> &outRoute, const SystemPath &start, const SystemPath &target) const
+void SectorView::AutoRoute(const SystemPath &start, const SystemPath &target, std::vector<SystemPath> &outRoute) const
 {
 	const RefCountedPtr<const Sector> start_sec = m_galaxy->GetSector(start);
 	const RefCountedPtr<const Sector> target_sec = m_galaxy->GetSector(target);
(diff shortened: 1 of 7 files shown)


The-EG:
Unfortunately, I had no idea that `GetSector` was so slow!

The cylinder is 6 sectors (48LY?) in diameter? That seems like a reasonable value, maybe even a little high still. I wonder if this could be based on the range of the players hyperdrive (like half of max or something). 
I'll do some testing to see if there are any major differences in what gets generated. 

I wouldn't expect there would be differences, but it is possible that the optimal route starts out by jumping perpendicular to the line between the start and the target, and I guess there could be some rare cases where the first jump could be slightly 'behind' the start.
Does `DistanceFromLine` only give the perpendicular distance? Is that what `isWithinLineSegment` signifies? 
If it's the closest distance to the line, regardless if that's perpendicular or not, then the above is less of a concern...because the initial set of nodes would be more of a sausage shape than cylinder.

In any case, good stuff!

impaktor:
Any ballpark estimate on how much faster this is? (Just curuous).

@clausimu said [on the forum](https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=546)

> I just planned a route from Epsilon Eridani to the home of the Haber Corporation. About 200 ly of travel distance. Pioneer froze up completely for a little more than 10 minutes until the calculation was done.

Oh, I see later i the forum: it' skips 13 million `Getsector`-calls for a 96ly distance.

src/SectorView.cpp
```diff
@@ -585,48 +586,60 @@ std::vector<SystemPath> SectorView::GetRoute()
 	return m_route;
 }
 
-std::vector<SystemPath> SectorView::AutoRoute(const SystemPath &start, const SystemPath &target)
+void SectorView::AutoRoute(std::vector<SystemPath> &outRoute, const SystemPath &start, const SystemPath &target) const
```
mooskagh: I don't think that returning the path by value caused the vector copy. In C++11 it surely will move it rather than copy, but even before that, the RVO would probably kick in. And even if it didn't, it's rather a small vector, so I'd expect no visible slowdown because of that.
But if you'd still like to keep it as an out param, I'd recommend to move it to the last param (it's just more popular way, it all style guides that I saw that's the recommendation), and to convert from reference to a pointer (that way the calling function will know that that parameter is probably modified).

src/SectorView.cpp
```diff
@@ -585,48 +586,60 @@ std::vector<SystemPath> SectorView::GetRoute()
...
+		for (Sint32 sy = minY; sy <= maxY; sy++) {
+			for (Sint32 sz = minZ; sz < maxZ; sz++) {
+				const SystemPath sec_path = SystemPath(sx, sy, sz);
 				RefCountedPtr<const Sector> sec = m_galaxy->GetSector(sec_path);
 				for (std::vector<Sector::System>::size_type s = 0; s < sec->m_systems.size(); s++) {
-					if (start.IsSameSystem(sec->m_systems[s].GetPath())) continue; // start is already nodes[0]
+					if (start.IsSameSystem(sec->m_systems[s].GetPath()))
+						continue; // start is already nodes[0]
+
+					bool isWithinLineSegment = false;
```
mooskagh: Not used.

src/SectorView.cpp
```diff
@@ -585,48 +586,60 @@ std::vector<SystemPath> SectorView::GetRoute()
...
+						Sector::DistanceBetween(target_sec, target.systemIndex, sec, sec->m_systems[s].idx) <= dist * 1.10 &&
+						/*isWithinLineSegment &&*/ lineDist<(Sector::SIZE*3))
+					{
 						nodes.push_back(sec->m_systems[s].GetPath());
 					}
 				}
 			}
 		}
 	}
+	Output("SectorView::AutoRoute, nodes to search = %lu\n", nodes.size());
```
mooskagh: Is it intended to be here, or it's a leftover from the debugging code?

src/SectorView.cpp
```diff
@@ -585,48 +586,60 @@ std::vector<SystemPath> SectorView::GetRoute()
...
 
 	// setup inital values and set everything as unvisited
+	std::vector<float> path_dist; // distance from source to node
+	std::vector<std::vector<SystemPath>::size_type> path_prev; // previous node in optimal path
+	std::set<std::vector<SystemPath>::size_type> unvisited;
```
mooskagh: std::set operations while being O(log N), are often slow because of memory allocations.
If we don't need the order (and it doesn't seem that we do), unordered_set will perform better.
Furthermore, as we don't need to check whether something is in the set, but only iterate over it, I'd replace that std::set with std::vector (and when element is erased, just copy the last element of a vector to a "hole")

src/SectorView.cpp
```diff
@@ -636,64 +649,74 @@ std::vector<SystemPath> SectorView::AutoRoute(const SystemPath &start, const Sys
 	// distance to the start is 0
 	path_dist[0] = 0.f;
 
+	size_t totalSkipped = 0u;
 	while (unvisited.size() > 0) {
 		// find the closest node (for the first loop this will be start)
 		std::vector<SystemPath>::size_type closest_i = *unvisited.begin();
-		for (auto it = unvisited.begin(); it != unvisited.end(); it++) {
+		for (auto it = unvisited.begin(), itEnd = unvisited.end(); it != itEnd; ++it) {
```
mooskagh: I don't think that caching end() iterator will actually make any difference. In fact, I expect them to produce identical code by most compilers.
By the way, is there any reason range-based for loop is not used here?
i.e:
` for (auto x : unvisited) { ... } `

mooskagh: Also, if instead of iterating over the vector to find the current minimum, heap is used (e.g. std::priority_queue, or it may be easier to use std::set for that [as it can insert/remove/get minimum, all at O(log N), just as heap], the algorithm becomes O(N log N) rather than O(N^2)), see for example here:
https://www.geeksforgeeks.org/dijkstras-shortest-path-algorithm-using-priority_queue-stl/

It requires some rework, so feel free to ignore that comment (or I can rewrite it).

src/SectorView.cpp
```diff
@@ -636,64 +649,74 @@ std::vector<SystemPath> SectorView::AutoRoute(const SystemPath &start, const Sys
...
 
+		RefCountedPtr<const Sector> closest_sec = m_galaxy->GetSector(closest);
+
 		// if not, loop through all unvisited nodes
 		// since every system is technically reachable from every other system
 		// everything is a neighbor :)
-		for (auto it = unvisited.begin(); it != unvisited.end(); it++) {
-			SystemPath v = nodes[*it];
-			RefCountedPtr<const Sector> v_sec = m_galaxy->GetSector(v);
+		for (auto it = unvisited.begin(), itEnd = unvisited.end(); it != itEnd; ++it) {
```
mooskagh: Also, why not range-based for loop?

src/SectorView.cpp
```diff
@@ -636,64 +649,74 @@ std::vector<SystemPath> SectorView::AutoRoute(const SystemPath &start, const Sys
...
 				// if our calculated duration is less than a previous value, this path is more efficent
 				// so store/override it
-				path_dist[*it] = v_dist;
-				path_prev[*it] = closest_i;
+				path_dist[idx] = v_dist;
+				path_prev[idx] = closest_i;
 			}
 		}
 	}
+	Output("SectorView::AutoRoute, total times that nodes were skipped = %lu\n", totalSkipped);
```
mooskagh: Is that a leftover from the debug code? If yes, consider also removing totalSkipped.

src/SectorView.cpp
```diff
@@ -636,64 +649,74 @@ std::vector<SystemPath> SectorView::AutoRoute(const SystemPath &start, const Sys
...
 			break;
 		}
 	}
 
-	// Build the route, in reverse starting with the target
-	while (u != 0) {
-		route.push_back(nodes[u]);
-		u = path_prev[u];
+	// It's posible that there is no valid route
+	if(foundRoute) {
```
mooskagh: Nitpicking: space after "if", to be consistent with the rest of the code.

src/SectorView.cpp
mooskagh: Ah, it's actually used as an output parameter of a function.. Then it's ok to keep it then, or consider removing that altogether from DistanceFromLine() function

src/SectorView.cpp
fluffyfreak: It was just my preference, and like a few of the others after this it's because I don't trust the compiler to actually do that. Sure sometimes it will use move instead of copy but not always, whereas this way it will always work 👍

impaktor: If being explicit and clear in the code doesn't add any downside, then I'm all for it, as at least I am very foggy on what magic goes on in the compiler.

src/SectorView.cpp
fluffyfreak: Deliberately left there, I want to see this info and it'll give me more detail from the `output.txt` when people complain about it being slow.

src/SectorView.cpp
fluffyfreak: Deliberate, more info for the logs when people complain about it being slow.

src/SectorView.cpp
fluffyfreak: This is where I'd actually like to put in a spatial locality search instead of this iterate over everything approach.

fluffyfreak:
> The cylinder is 6 sectors (48LY?) in diameter? That seems like a reasonable value, maybe even a little high still. I wonder if this could be based on the range of the players hyperdrive (like half of max or something).
I'll do some testing to see if there are any major differences in what gets generated.

It's purely to restrict the number of sectors/starsystems to be considered so I deliberately didn't base it on what jump range you might have. Otherwise the bigger the range the worse the performance.
Also you might have a tiny ship with a reasonable range, but you have to expand the distance to some deliberate minimum or it would fail to plot many routes dues to over restricting the radius of the search cylinder.

In practice it doesn't quite make it to 6 sectors because you can travel on a diagonal from (0,0,0) to (12,12,12) for example and sectors being whole indexed might well fall out making it only 4 or 5 wide in places.

Just repeated the test of plotting from Earth to Haber Corportation, original it meant a 10 minute hang, now it takes 4 seconds... much improved!

fluffyfreak:
@mooskagh thanks for the nitpicking, good nits, well picked.

PR #4287, opened 2018-01-20
Keepers orbital station texture update

fluffyfreak:
Keepers [orbital station texture update](https://pioneerspacesim.net/forum/viewtopic.php?f=6&t=539) with pattern textures converted to DDS


impaktor:
Will this then double the amount of "unique" stations, as there will be a 50-50 chance if they use these textures or original?

fluffyfreak:
I cannot honestly remember right now. I think it will always use a pattern if there is one available.

fluffyfreak:
@nozmajner what do you think?

nozmajner:
I think it's good. Are they the ones made from the dds textures? Those were quite artifact-laden.

fluffyfreak:
@nozmajner they're from whatever was used in that zip file

nozmajner:
Then that's from the dds-es. The artifacts were quite obvious on the hud, so maybe we shouldn't merge them yet.

fluffyfreak:
So, what's going with this one?

nozmajner:
I think so

PR #4289, opened 2018-01-23
cityonplanet setuserdata optimized

joonicks:
setuserdata for collision handling,
create building geom in a single call instead of three


fluffyfreak:
nice

PR #4294, opened 2018-01-24
Minor optimisations around avoiding isqrt

fluffyfreak:
Two sets of really low level optimisations:
- Avoid using isqrt at all wherever possible by using the square of the comparison value instead.
- Replace soft isqrt with cast from sqrt which is 7.3 to x15 times faster.

Suggested in a post-commit [comment](https://github.com/pioneerspacesim/pioneer/commit/208065b95a187c8da10f6bb47cfd7eb21f2bd3e8#commitcomment-27075091) by @joonicks 👍

### DIFF
--- a/src/galaxy/SectorGenerator.cpp
+++ b/src/galaxy/SectorGenerator.cpp
@@ -9,6 +9,8 @@
 #include "Factions.h"
 #include "GameSaveError.h"
 
+#define Square(x) ((x)*(x))
+
 static const unsigned int SYS_NAME_FRAGS = 32;
 static const char *sys_names[SYS_NAME_FRAGS] =
 { "en", "la", "can", "be", "and", "phi", "eth", "ol", "ve", "ho", "a",
@@ -17,11 +19,10 @@ static const char *sys_names[SYS_NAME_FRAGS] =
 
 bool SectorCustomSystemsGenerator::Apply(Random& rng, RefCountedPtr<Galaxy> galaxy, RefCountedPtr<Sector> sector, GalaxyGenerator::SectorConfig* config)
 {
-	PROFILE_SCOPED()
-
 	const int sx = sector->sx;
 	const int sy = sector->sy;
 	const int sz = sector->sz;
+	const Sint64 dist = (1 + sx*sx + sy*sy + sz*sz);
 
 	if ((sx >= -m_customOnlyRadius) && (sx <= m_customOnlyRadius-1) &&
 		(sy >= -m_customOnlyRadius) && (sy <= m_customOnlyRadius-1) &&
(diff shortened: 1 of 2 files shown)


jaj22:
Due to rounding, the square comparison substitutions are not maintaining the current behaviour. Consider `if(isqrt(101) > 10)` vs `if(101 > 100)`.

I think the sqrt + cast is safe up to a point. Float to int casts round down in C (as does isqrt, apparently), and square roots of adjacent numbers need similar precision to distinguish them, so values towards 2^53 should give the correct results.

Danger area is if this is used for values >2^53, which may happen in cases where it's used on fixed-point values. Rather than trying to figure out whether all of those are sub-2^53, it might be better to provide an isqrt_fast and substitute in the cases where you need it.

fluffyfreak:
@jaj22 the only Fixed usage is https://github.com/pioneerspacesim/pioneer/blob/master/src/galaxy/StarSystemGenerator.cpp#L767 so I'll do a fast/slow version

joonicks:
and then when you start thinking about it, comparing (x*x + y*y + z*z) with Square(x) seems a bit peculiar..

fluffyfreak:
@joonicks no that still makes sense since it's replacing the test against `isqrt` with a test against the square of the value.

joonicks:
@fluffyfreak its because I dont know math. I didnt know a^2 = b^2 + c^2 + d^2 is a valid thing.

fluffyfreak:
It isn't:
> `a^2 != b^2 + c^2 + d^2`

However:
> `isqrt(1 + sx^2 + sy^2 + sz^2) <= 90`

Is equivalent too:
> `(1 + sx^2 + sy^2 + sz^2) <= 90^2`

Which is what changed because, as you pointed out, a single `mul` is much cheaper than an `isqrt`.

joonicks:
> Which is what changed because, as you pointed out, a single mul is much cheaper than an isqrt.

Its a constant, zero muls

>It isn't:

>>   a^2 != b^2 + c^2 + d^2

I was referring to pythagoras theorem, a^2 = b^2 + c^2. If you introduce a third axis d, its still a^2 = b^2 + c^2 + d^2, I always though yed need ta a^2 = (sqrt(b^2 + c^2)^2) + d^2 to get the correct answer but even now writing it out I see my folly.

jaj22:
This needs a save bump due to the `if(isqrt(101) > 10)` vs `if(101 > 100)` issue. Some systems will change as a result. Shouldn't those be saved up?

fluffyfreak:
@jaj22 there may be minor differences in the universe... now that I've fixed the mistake that I'd made.

However they shouldn't affect existing save games and hopefully most people won't notice them.

I try to only use the save game bump for things that would break previous save game files, or cause massive changes to the universe. This probably won't be noticeable without some effort so I don't think it needs a bump.

joonicks:
actually if (isqrt(dist) <= 90) is different from (dist <= Square(90)), an additional rng.Int32() will be called and followup effects of that will be that the currently generated system will have different star types and the system will be differently named, and as starsystems are generated one sector at a time, if the current system is not the last in that sector, any systems in that sector generated after will be completely different.

but its also true that noone will ever notice.

Issue #1644, opened 2012-10-29
Hot Key request for Heading Hold Flight Modes


Baobobafet:
Having the Heading Hold and retro functions is a great feature, but it's not really as user friendly as it  could be and can lead to some unanticipated maneuvers when one is least expecting.

If this control were more consistent in it's responses, it would be terrific, but in this instance, I think it is unfortunately failing to live up to expectation.
Worse, it may be creating confusion for the average player not to mention any possible new users.

This control (as critical as it is) should be made a lot more user friendly via adding at least, an additional (more reliable) means of cycling thru the two heading hold options.

The hot key(s) solution for engaging such an important control function(s) reliably would be one possible alternative giving the player more decisive and predictable responses from it's use.

Hot key approaches also offer increased flexibility when mapping critical game functions onto various types of game controllers.

IMO, It's one of the most important nav control functions to have in Pioneer - I hope this can be easily resolved.

baobob

Keeper1st:
Old request, but I'd like to second it. As a joystick user, having to reach over to the keyboard or mouse in order to switch between full manual and set-speed modes is an inconvenience. The ability to map it not only to another key, but also to a controller button, would be quite useful.

As it stands, when I'm manually flying I tend to try never having to use it, making things more difficult than they should be, simply because it's a hassle to activate or deactivate it. We can adjust the set speed with a controller, but not actually turn on the set-speed mode. It's a bit silly.

PR #4299, opened 2018-01-28
Main Menu converted to ImGui

### LINKED ISSUE #847: Load / Save game screen suggestions
When mouse hovers over a game save file name, may I suggest a summary of the game state is displayed.

Data to display:
                Current ship
                Current system
                Ship state (Landed / InFlight)
                Game date/time
                Credits
                Combat rating

Just a few details to help keep track of multiple game save files.

### LINKED ISSUE #3146: controller name sometimes is too long
bez tytulu
suggestion: small fonts for joyname

The-EG:
Includes:
Main Menu
Quit Confirmation
Settings Window (both main menu and in game)
Load/Save Dialog
Key/Joystick Binding Dialog

Additional Changes:
Starting locations are defined in a single spot, making modding easier.
Starting locations also include a brief description that is shown on
  mouse hover.
It is no longer necessary to hold the joystick hat in the desired
  direction while confirming the binding.
Save games now include some basic information that is shown on mouse
  hover in the load game window (fixes #847). Newer save games will show:
    - Game Time
    - System
    - Ship
    - Credits
    - Docking location (if docked)
  This is backwards compatible with older saved games, which show:
    - Game Time
    - Ship
    - The current frame of reference

Screenshots!
Main Menu

Start location info shown on tooltips:
Start Location Info

desktop screenshot 2018 01 27 - 20 12 57 98

Also fixes #3146.


impaktor:
Awesome!

I assume the starting location tooltip is built dynamically, so if we change anything, say starting credits, or ship, it picks that up.

For save/load, do we also want to include the output of `GetFlightState()`? (DOCKED, FLYING, LANDED, HYPERSPACE, etc.), so if docked show port, if not, just show flightstate?

For video configuration screen, we can now have tooltip, so, maybe we want to include some description as to what e.g. VSync and such does, and specifically if for each option higher/lower means more or less demands on hardware? E.g. _"set to low for faster rendering"_. Maybe @FluffyFreak want's to suggest good tooltip texts?

An alternative is to have a little Internet-help icon that takes the player to: https://pioneerwiki.com/wiki/Settings_Menu (but when we have mouse over tooltip, it feels like we could make use of it).

Will this also close #3146? 

Open question:
What's #1644 talking about? Something of interest, or maybe we should close that too?

The-EG:
> I assume the starting location tooltip is built dynamically, so if we change anything, say starting credits, or ship, it picks that up.

I started to do that, but I changed it to static because I thought dynamically building it would cause issues with translation. It'd be easy enough to change back though.

> For save/load, do we also want to include the output of GetFlightState()? (DOCKED, FLYING, LANDED, HYPERSPACE, etc.), so if docked show port, if not, just show flightstate?

Sure.

> For video configuration screen, we can now have tooltip, so, maybe we want to include some description as to what e.g. VSync and such does

Sure, just need the strings! I can take a first pass.

> Will this also close #3146?

Not exactly. We could make the buttons wider or try to wrap text, but I think it's always going to be a possibility. Adding in tooltips (not currently there yet) and the fact that it is displayed in full on the 'capture' window does help.

> What's #1644 talking about? Something of interest, or maybe we should close that too?

This sounds like a request to be able to configure the keys/actions handled elsewhere in the UI. It's a valid point and 2 other people brought up similar requests in chat about hard coded keys. In general, this would require adding key bindings for most things that don't have them and converting over the hard coded stuff.

impaktor:
> Adding in tooltips (not currently there yet) and the fact that it is displayed in full on the 'capture' window does help.

Yes, that sounds like a solution that could warrant closing that issue. So tooltip on all / or only when text owerflows the box.

> This sounds like a request to be able to configure the keys/actions handled elsewhere in the UI. 

OK, so deemed outside the scope of this PR.

ecraven:
> This sounds like a request to be able to configure the keys/actions
> handled elsewhere in the UI. It's a valid point and 2 other people
> brought up similar requests in chat about hard coded keys. In general,
> this would require adding key bindings for most things that don't have
> them and converting over the hard coded stuff.

I've was thinking about this a few months ago, we should really not
hard-code any keys, but make everything configurable, but we should
probably make some kind of plan before just going off and implementing
that willy-nilly ;)

I'll have to look into the existing key-binding machinery, on whether it
already has everything we would need. Or maybe we should think about
moving to a keybinding -> command -> action system, that might work well
for multiplayer too :P :D

The-EG:
A few changes:
 - Added in some tooltips for the video settings and control binding buttons (for #3146). I took most of the descriptions from the wiki.
 - Changed the starting location tooltip back to auto-generated. The first line is still static, and requires translation, but that's where you can add some flavor too. It's structured so that it should still translate decently.

impaktor:
I'll have a look at the tooltips later, and we should give @ecraven some time if he wants to review, once he's regained his health.

Also, looks like we aught to save-bump anyway with the optimization changes, so backwards compability might be a non-issue.

data/pigui/mainmenu.lua
```diff
@@ -0,0 +1,213 @@
...
+
+local showQuitConfirm = false
+local quitConfirmMsg
+local max_flavours = 19
+
+local startLocations = {
+	{['name']=lui.START_AT_EARTH,
+	 ['desc']=lc.START_AT_EARTH_DESC,	 
+	 ['location']=SystemPath.New(0,0,0,0,6,48600),
+	 ['shipType']='sinonatrix',['money']=100,hyperdrive=true,
```
ecraven: just for consistency's sake, could you make this ['hyperdrive']?

data/pigui/mainmenu.lua
```diff
@@ -0,0 +1,213 @@
...
+	 ['shipType']='xylophis',['money']=100,hyperdrive=false,
+	 ['equipment']={
+		{misc.atmospheric_shielding,1},
+		{misc.autopilot,1},
+		{misc.radar,1},
+		{cargo.hydrogen,2}}}
+}
+
+local function dialogTextButton(label, enabled, callback)
+	local bgcolor = enabled and colors.buttonBlue or colors.grey
```
ecraven: indentation

data/pigui/mainmenu.lua
```diff
@@ -0,0 +1,213 @@
...
+		button = ui.coloredSelectedButton(label, dialogButtonSize, false, bgcolor, nil, enabled)
+	end)
+    if button then
+        callback()
+    end
+    return button
+end
+
+local function mainTextButton(label, tooltip, enabled, callback)
+	local bgcolor = enabled and colors.buttonBlue or colors.grey
```
ecraven: indentation

impaktor:
Tootips it was, right! @The-EG once you fixed up stuff commented on above, I could squash a commit of mine with edits to tooltip (I assume I'll tweak something) into yours.

Ideally, once this is merged, it would be nice if F3/F4 menues are also migrated to Imgui before next release, as they're visually similar. No hurry, though.

data/pigui/mainmenu.lua
The-EG: This should be indented...it's a function...or am I missing something?

data/pigui/mainmenu.lua
The-EG: same as above?

data/pigui/mainmenu.lua
The-EG: fixed.

The-EG:
From #pioneer chat:
> \<ecraven\> only thing I noticed is that Escape in the game pauses and opens the settings window, but pressing escape again doesn't close the window and unpause the game. but that can be fixed easily in another commit

This is also now fixed.

impaktor:
I've read through the strings, I have nothing to add, except nit-pick:

- I prefer some punctuation around the `however`:
```"Samples the texture across multiple mip levels to produce a much clearer texture when viewed at shallow angles. Stops the textures from becoming as blurred; however, it can come at a heavy performance cost."```

- But more, I wonder if these strings really should be in `core/en.json`? That's usually only for the C++ side strings, I think? `ui-core/en.json` is where we put lua-side strings, I think. Please correct me if I'm wrong, but I assume the strings you've added are mostly/all for Lua-side code?

File: data/pigui/mainmenu.lua (hunk 2 above)
Comment (The-EG) [reply data/pigui/mainmenu.lua]: From remarks in chat, this sounds like it's tabs vs spaces. I'll take a look.

**Comment by The-EG** [comment]
> I prefer some punctuation around the however:

Ok

> But more, I wonder if these strings really should be in core/en.json?

Ok. I started to put them there but I over thought it. Easy enough to move them.

**Comment by impaktor** [comment]
Maybe some of these warnings are related to your stuff?

```
LuaFixed.cpp: In function ‘int l_fixed_new(lua_State*)’:
LuaFixed.cpp:11:45: warning: use of old-style cast [-Wold-style-cast]
  Sint64 num = (Sint64) luaL_checknumber(L, 1);
                                             ^
LuaFixed.cpp:12:47: warning: use of old-style cast [-Wold-style-cast]
  Sint64 denom = (Sint64) luaL_checknumber(L, 2);  // use checknumber for >32-bit precision
   
PngWriter.cpp: In function ‘void write_png(FileSystem::FileSourceFS&, const string&, const Uint8*, int, int, int, int)’:
PngWriter.cpp:39:23: warning: use of old-style cast [-Wold-style-cast]
     ((Uint8*)surface->pixels)[dst_index + channel] = bytes[src_index + channel];
                       ^~~~~~
LuaPiGui.cpp: In function ‘int l_pigui_get_axisbinding(lua_State*)’:
LuaPiGui.cpp:649:21: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   for (int a = 0; a < axes.size(); a++) {
                   ~~^~~~~~~~~~~~~
LuaPiGui.cpp:647:24: warning: unused variable ‘id’ [-Wunused-variable]
   const SDL_JoystickID id = js.first;
                        ^~
LuaPiGui.cpp: In function ‘int l_pigui_get_keybinding(lua_State*)’:
LuaPiGui.cpp:709:22: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
    for (int b = 0; b < buttons.size(); b++) {
                    ~~^~~~~~~~~~~~~~~~
LuaPiGui.cpp:710:89: warning: use of old-style cast [-Wold-style-cast]
     Output("%s: %d -> %d\n", Pi::JoystickGUIDString(js.first).c_str(), b, (int)buttons[b]);
                                                                                         ^
LuaPiGui.cpp:717:22: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
    for (int h = 0; h < js.second.hats.size(); h++) {
                    ~~^~~~~~~~~~~~~~~~~~~~~~~
LuaPiGui.cpp:707:25: warning: unused variable ‘id’ [-Wunused-variable]
    const SDL_JoystickID id = js.first;
                         ^~
LuaPiGui.cpp: In function ‘int l_pigui_combo(lua_State*)’:
LuaPiGui.cpp:1209:81: warning: use of old-style cast [-Wold-style-cast]
  bool changed = ImGui::Combo(lbl.c_str(), &selected, &items[0], (int)items.size());
                                                                                 ^
LuaPiGui.cpp: In function ‘int l_pigui_listbox(lua_State*)’:
LuaPiGui.cpp:1229:83: warning: use of old-style cast [-Wold-style-cast]
  bool changed = ImGui::ListBox(lbl.c_str(), &selected, &items[0], (int)items.size());
```

**Comment by The-EG** [comment]
> Maybe some of these warnings are related to your stuff?
Some of them are, yes. I'm not getting all of them in VS, weird...I think I took care of them though.

File: data/pigui/mainmenu.lua (hunk 2 above)
Comment (The-EG) [reply data/pigui/mainmenu.lua]: This should be better now...there was a mix of spaces and tabs in these functions, I converted to all tabs.

**Comment by impaktor** [comment]
I was about to merge, but I'm not sure I understand the logic division between putting strings in `core/en.json` vs. `ui-core/en.json`. My understanding was that all things you've added are on lua-side, so I expected the strings to be in `ui-core`, but looks like most are in `core`. 

ping @laarmen opinions?

Also, copyright header missing in saveloadgame, and settingswindow.lua

Other than that: let's get this sucker merged!

**Comment by The-EG** [comment]
Anything I've added should be in ui-core, unless I got it backwards somehow. There are still some strings referenced in core, but I didn't move those because I have no idea where they are all used.

I'll fix up the headers this evening.

**Comment by The-EG** [comment]
> Also, copyright header missing in saveloadgame, and settingswindow.lua

Now fixed.

>  I expected the strings to be in ui-core

That should be the case...double checked!

File: data/pigui/mainmenu.lua [hunk 4]
```diff
@@ -0,0 +1,213 @@
...
+			if showQuitConfirm then confirmQuit() end
+		end)
+	end)
+
+	callModules('mainMenu')
+end -- showMainMenu
+
+ui.registerHandler('mainMenu',function(delta)
+	showMainMenu()
+end)
```
impaktor: File needs a new closing line

data/pigui/modules/saveloadgame.lua
```diff
@@ -0,0 +1,138 @@
...
+				showSaveFiles()
+			end)
+		end)
+	end
+end
+
+ui.registerModule("game", savedGameWindow)
+ui.registerModule("mainMenu", savedGameWindow)
+
+return {}
```
impaktor: newline needed.

data/pigui/modules/settings-window.lua
```diff
@@ -0,0 +1,508 @@
...
+		if showKeyCapture then
+			captureBinding(keyCaptureId, keyCaptureNum)
+		end
+	end
+end
+
+ui.registerModule("game", optionsWindow)
+ui.registerModule("mainMenu", optionsWindow)
+
+return {}
```
impaktor: newline

The-EG:
> My only nitpick now is that the new files end on a new line.

Either git is eating them (through CR/LF conversion?) or it's just not displaying in GitHub. I've put 2 extra lines at the bottom of each of those files:

capture

impaktor:
Github showed that there were no linebreak.

Now there's two, so I think it's working.

Anyway, merged, thanks!

PR #4300, opened 2018-01-28
new hotkey for landing gear

joonicks:
default: g


nozmajner:
G is already used for decreasing the set-speed setting (T increases).

fluffyfreak:
`F6` is the current Toggle landing gear key

nozmajner:
I like the notion of moving it from the F-keys to somewhere rebindable. Freeing them up for UI function and such is a good direction in my opinion.
But I think  T and G for Set Speed should stay there, so you can reach every important control on that side of the keyboard.
How about defaulting it to H or V? V is closer to WASD, so might be more comfortable. True, it doesn't rhymes that nicely as with G, but I doubt there's any other hotkey in Pioneer that lines up with its first letter anyway.

joonicks:
z and x seems unused.... check y0 staging!

nozmajner:
Z sounds good to me.

joonicks:
I was suggesting z and x for setting thrust while keeping g for gear

nozmajner:
Z and X isn't as comfortable to reach in my opinion, and Z and Y tend to switch places on some international keyboard layouts too.
And T and G have a bit of directionality to them, like WASDRF for the lateral directions.

clausimu:
I agree with @nozmajner regarding the "z" - switches position with "y" on US vs. German keyboard. That sometimes messes up hotkey layout things.

mooskagh:
I think key bindings should not (and currently do not, I guess) depend on keyboard layout. Otherwise it's complete mess for Dvorak users, and also for people with multiple layouts (e.g. Russian) finding out that keyboard doesn't work because you didn't change layout is pretty annoying.

The-EG:
Why not leave the default at F6, but have this allow it to be changed for those users that want to?

As commented on in #3799, key bindings are going to have to be cleaned up/overhauled at some point with moving everything to ImGui, so we will likely be talking about it more in the future.

fluffyfreak:
Closing

PR #4308, opened 2018-02-09
Star Icons overhaul

mruk:
Giants, super giants, hyper giants rotting on my HD for years... The
rest of the famous 'Mysibrat's Collection':
http://spacesimcentral.com/forums/topic/plans-for-alpha-11/#post-80380

All files are Icons - no screenshots needed.

(Almost) all Icons are the same size, with Alpha - source files are on my HD (if we need source files for future editions for eg.)


impaktor:
Nice. However, we're currently thinking about re-doing the whole thing, so we're not sure there's much meaning to replace something that's on its way out anyway. 

We'll see what happens. Either way, thanks for taking the time to make the PR.

nozmajner:
[Maybe this could be handy too:](https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=261&p=3073&hilit=icons#p3073)
But as @impaktor said, I'd say we should wait with this kind of things until the map screens are migrated to imgui.

fluffyfreak:
These are still good I think. Improvements over the existing.

As for "they might change", well, they haven't yet. I'm always of the opinion that we takes positive changes we have now instead of changes that we might have in the future.

PS: Yes this means I'm back from the US 😁

impaktor:
I'm reluctant to merging graphics that might soon get replaced anyway, since that inflates the git repo size unnecessarily.

Don't get me wrong, I'd like to see this PR kept open, until we know where we stand on the new screens. I think @The-EG has some proof of concept he's working on (when time permits).

fluffyfreak:
No-one is currently working on the `SystemInfoView`, it's a 1.4MB increase in repo size at maximum and subjectively the icons look nicer than what we have.

Maybe someone will popup and port it to IMGUI tomorrow but they haven't since February when this was opened. Merging.

PR #4315, opened 2018-02-24
 Keepers orbital texture updates and patterns

nozmajner:
Adding the updated textures Keeper made, with solar panels on the outside of the orbital station, and 7 patterns too:
modelviewer-20180224-191246

I've added Keeper as commit author, but I turn up there too for some reason (hope it's not a problem)


fluffyfreak:
Beat me too it ;)

nozmajner:
Figured I'd do something useful with the new laptop. :)

PR #4328, opened 2018-04-25
Throw lua error when specific Lang is not found.

### LINKED ISSUE #4327: Game won't launch when choosing a localized language
### Observed behaviour
I previously selected the Italian language from the setting screen. When launching Pioneer the game exits with this error
screenshot_20180425_183511

If I modify Lang=it to Lang=en in config.ini, the game starts correctly.

### Expected behaviour

That the game starts. :)

I saw this issue earlier. It happened when there were untranslated strings. In this case I don't see untranslated strings in transifex for my language. Anyway, I don't even see string that have to be translated for the easter egg module.
What I expect is that the game starts correctly even when there are untranslated strings, and reverts to default english for those.

### Steps to reproduce

Change Lang=en to it, de, pl or others in config.ini and launch the game.

Pioneer version 20180203.r33.g22063c289-1
OS: Arch Linux

ecraven:
When lua Lang.GetResource does not find a resource, it throws a lua error,
mentioning the exact missing file, and does *not* just silently return nil.

Partially fixes #4327.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -66,7 +66,7 @@ file_filter = data/lang/module-crewcontracts/<lang>.json
 source_file = data/lang/module-crewcontracts/en.json
 source_lang = en
 
-[pioneer.module-easteregg]
+[pioneer.module-easteregg-message]
 file_filter = data/lang/module-easteregg-message/<lang>.json
 source_file = data/lang/module-easteregg-message/en.json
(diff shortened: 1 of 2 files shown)


impaktor:
Yeah, the module on transifex has the `-message` part, so that's probably the missing piece.

PR #4334, opened 2018-05-20
EasterEgg: Call onDelete if removed

### LINKED ISSUE #4329: Saving game fails after hyperspacing away from Barnard's star
### Observed behaviour

With the (alomst) latest pioneer version (commit 78e140c from Apr 26) I get the following bug:

If you leave from Barnard's Star to any other system, e.g. to sol, then trying to save the game after the jump fails with a Lua serialization error:

`Lua serializer '.Message.ads.91.station' tried to serialize an invalid 'SpaceStation' object`

This happens both with "normal" saving via the settings menu and with ctrl-f9 quick save (I have autosave diabled)

I experience no saving problems as long as I stay away from Barnard's Star.

__latest bug-free version__

The bug is not present in the last release version from february, i.e. tag 20180203

### Steps to reproduce

Start at Barnard's Star
Sell all initial equipment and buy class 1 hyperdrive
Hyperspace to sol
Quick save the game with ctrl-f9

Same behaviour if you start at sol, then go to Barnard's Star, then back to sol, then save

__log files__

[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/1958634/opengl.txt)
[output.txt](https://github.com/pioneerspacesim/pioneer/files/1958635/output.txt)

pioneer build from github sources: commit 78e140c from Apr 26
OS: Debian stretch 9.4

robothauler:
Fixes #4329

### DIFF
--- a/data/modules/EasterEgg/Message.lua
+++ b/data/modules/EasterEgg/Message.lua
@@ -35,6 +35,9 @@ local onChat = function (form, ref, option)
 	form:SetMessage("\n" .. flavours[ad.flavour].body)
 end
 
+local onDelete = function (ref)
+	ads[ref] = nil
+end
 
 local makeAdvert = function (station, flavour_index)
 
@@ -47,6 +50,7 @@ local makeAdvert = function (station, flavour_index)
 	local ref = station:AddAdvert({
 		description = ad.desc,
 --		icon        = "lightning",
+		onDelete    = onDelete,
 		onChat      = onChat})
 	ads[ref] = ad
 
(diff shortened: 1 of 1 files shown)


impaktor:
Ah, good. I had a feeling I've missed something. In fact, I was thinking, it shouldn't need saving/loading state at all, since it's fixed to the station, and never goes away, or changes, so I had a thought some time ago, to remove the seralize/unserialize functions.

robothauler:
I pushed it again because I forgot onGameStart.

fluffyfreak:
Code looks good to me, but I'll leave this review to @impaktor :)

impaktor:
:+1: for branch name!

impaktor:
So, the bug was, if visiting a system with an "egg", then visiting another system, the BBS would try to use that egg, I think, (just running the code very sloppily in my brain).

PR #4335, opened 2018-06-02
Audio alarms

WKFO:
This PR adds audio alarms to spaceships, which warns the pilot about

- Excessive temperature (red gauge 80 percent)

- Excessive atmospheric pressure (9 atm)

- Low fuel (compared to the vessel's current speed)

The alarm doesn't loop, but it rearms itself once the danger is over.
You may not like my coding style in this PR but this is how it worked without any bugs. For example, using elseif instead of using lots of if's caused the alarm to not rearm itself, game crashed after leaving a planet's SoI, alarm kept going continuously etc... So, I assure you, I know what I am doing.

I put the script as a pigui module because playSfx is a part of pigui. Also why not, audio is another part of user-vessel interface.

I didn't quite like the alarms we already have, so I just made a few myself.


nozmajner:
Cool!
For fuel, I think it would be more useful if the alarm sounds when the current speed is close to the remaining deltaV, since if you go over that, then you are screwed big time.

WKFO:
Right. I will probably change it when I get home

WKFO:
Done! The system now checks if the difference between remaining Dv and current speed is lower than 1/10th of the maximum Dv.

data/pigui/modules/master-alarm.lua
```diff
@@ -0,0 +1,52 @@
...
+		end
+		if not pressure or pressure < 9 and alreadyAlertedPres then
+			alreadyAlertedPres = false
+		end
+	end
+end
+
+ui.registerModule("game", alarm)
+
+return {}
```
impaktor: File doesn't end with a new line.

data/pigui/modules/master-alarm.lua
```diff
@@ -0,0 +1,52 @@
...
+local Game = import('Game')
+local ui = import('pigui/pigui.lua')
+
+local alreadyAlertedTemp = false
+local alreadyAlertedFuel = false
+local alreadyAlertedPres = false
+
+local function alarm ()
+	--check hull temperature
+	local t = Game.player:GetHullTemperature() * 100
```
impaktor: Personally, I would have kept  ` 0 < t < 1`, rather than convert to percent.

data/pigui/modules/master-alarm.lua
WKFO: It does now.

WKFO: It should now

PR #4338, opened 2018-06-15
Fix various isuess when saving/loading games

richardpl:
* Do not attempt to show saved games if there is none yet
* Ignore case when user gives empty string for saving game

Signed-off-by: Paul B Mahol <onemda@gmail.com>

### DIFF
--- a/data/pigui/modules/saveloadgame.lua
+++ b/data/pigui/modules/saveloadgame.lua
@@ -78,23 +78,25 @@ local function showSaveFiles()
 		saveFileCache = {}
 	end
 
-	table.sort(files, function(a,b) return (a.mtime.timestamp > b.mtime.timestamp) end)
+	if _ then
+		table.sort(files, function(a,b) return (a.mtime.timestamp > b.mtime.timestamp) end)
 
-	ui.columns(2,"##saved_games",true)
-	for _,f in pairs(files) do
-		if ui.selectable(f.name, f.name == selectedSave, {"SpanAllColumns"}) then
-			selectedSave = f.name
-		end
-		if Engine.pigui.IsItemHovered() then
-			local tooltip = getSaveTooltip(f.name)
-			Engine.pigui.SetTooltip(tooltip)
-		end
+		ui.columns(2,"##saved_games",true)
+		for _,f in pairs(files) do
+			if ui.selectable(f.name, f.name == selectedSave, {"SpanAllColumns"}) then
+				selectedSave = f.name
+			end
+			if Engine.pigui.IsItemHovered() then
+				local tooltip = getSaveTooltip(f.name)
+				Engine.pigui.SetTooltip(tooltip)
+			end
 
-		ui.nextColumn()
-		ui.text(Format.Date(f.mtime.timestamp))
-		ui.nextColumn()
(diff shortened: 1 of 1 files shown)


impaktor:
ping @laarmen or @johnbartholomew whom I think wrote the code.

laarmen:
Just glancing at it (on my mobile...) , the _ variable test is waaaaay too
obscure and should be replace with an explicit variable. Not readable, and
prone to breakage due to simple variable scope leakage.

The rest looks good to me.

richardpl:
I dont get it,, with what it should be replaced?

fluffyfreak:
@laarmen the use of '_' seems to be a common convention amongst Lua coders [apparently](https://stackoverflow.com/questions/34475995/variable-in-lua-has-a-special-meaning/34476033) though I did have to Google it myself!

ecraven:
> @laarmen the use of '_' seems to be a common convention amongst Lua
> coders [apparently](https://stackoverflow.com/questions/34475995/variable-in-lua-has-a-special-meaning/34476033)
> though I did have to Google it myself!

Though as I understand it, _ should be used to indicate something is
*ignored*, which is not the case here. Maybe just rename it, then this
PR seems fine?

fluffyfreak:
No it makes sense, that loop iterates over a pair.
So the first value '_' is being ignored, the second 'f' is what gets used.

If both values were to be used then '_' would be renamed something useful like 'e' ;)

fluffyfreak:
@laarmen & @ecraven merging as the usage of '_' was there from the original code so that part remains consistent. 
It's just been indented further due to the way it's guarded now.

johnbartholomew:
I think there has been some confusion here. There are *two* uses of `_` as a variable name in this function.

From the original code, the FileSystem.ReadDirectory call returns two values (and wrapping it in pcall gives you three return values - success/failure boolean, then the return values from the ReadDirectory call, or error string if it failed). The last value is assigned to `_` because in the original code that value was not used:

    local ok, files, _ = pcall(FileSystem.ReadDirectory, "USER","savefiles")

The second use of `_` as a variable name also comes from the original code, in the for loop. In this case, again, we have pairs looping through pairs of values (key and value from the table), and we don't care about the key so it is discarded by assigning to `_`:

    for _,f in pairs(files) do

The *new* code adds this if statement, which *uses* the previously unused value from the FileSystem.ReadDirectory call:

    if _ then

This is the contentious point - the value is being used, not just discarded; therefore the variable should not be called `_`. The two return values from the ReadDirectory function, according to its doc comment are (1) a table of files, (2) a table of (sub-)directories.

Finally, the if statement is weird anyway. It's checking the third result from pcall, which happens to be nil if the ReadDirectory call fails and non-nil if it succeeds, but that is not at all an obvious thing to do. The if statement should instead be checking the `ok` variable which is the intended and explicit indication of whether the call succeeded. Also immediately after the `pcall` there's an `if not ok then` block; you could just return from the function inside that block. Or you could make it `if not ok then ... else ... end` instead of checking `ok` a second time.

johnbartholomew:
Oh, I should warn though: I haven't actually run the code, all the above is just from reading it, so I could be wrong :-)

PR #4341, opened 2018-06-16
Reinstate searching by Sector coordinates

### LINKED ISSUE #4340: Move SectorView view to coordinates using the search box
### Observed behaviour
Moving to coordinates entered in the sector view search box does not work. It used to work before moving Sector View to Lua.

### Expected behaviour
Move the sector view camera to the coordinates entered into the search field as "pathX, pathY, pathZ"
Like before.

Latest master, Linux Mint Cinnamon.

fluffyfreak:
SectorView lost the ability to search by coordinates in the conversion to Pi-IMGUI, this simply restores that functionality.

It works by attempting to parse a string into a SystemPath, if that fails then it falls through to the existing search functionality. However if it succeeds then it moves the Sector screens viewing location to those coordinates.

This is intended to fix #4340

### DIFF
--- a/data/pigui/modules/map-sector-view.lua
+++ b/data/pigui/modules/map-sector-view.lua
@@ -8,6 +8,7 @@ local lc = Lang.GetResource("core");
 local lui = Lang.GetResource("ui-core");
 local utils = import("utils")
 local Event = import("Event")
+local SystemPath = import("SystemPath")
 
 local player = nil
 local colors = ui.theme.colors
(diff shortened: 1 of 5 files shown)


data/pigui/modules/map-sector-view.lua
```diff
@@ -120,33 +121,38 @@ local function showSearch()
...
+										 if ui.selectable(label, false, {}) then
+											 Engine.SetSectorMapSelected(item.path)
+											 Engine.SectorMapGotoSystemPath(item.path)
+										 end
 									 end
-								 end
-			end)
+				end)
+			end
 		end
```
impaktor: Is indentation correct here? I could imagine everything is too indented in the inner block,  or something? the `end` don't seem equidistant in indentation. I could be wrong. Just voicing concern.

fluffyfreak: There is a block that's more indented that the rest, but that seems common throughout the file and I just indented it by an extra tab with this PR.
It seems to follow the indentation set by the parantheses that wrap it.

impaktor:
Maybe @ecraven or @The-EG want to be in the loop on this.

data/pigui/modules/map-sector-view.lua
impaktor: OK.

PR #4343, opened 2018-06-17
Small saveloadgame script cleanup

fluffyfreak:
Avoid using the _ variable
Cleanup the safe to use test


richardpl:
Looks fine to me.

PR #4344, opened 2018-06-18
Add MSVC2017 to COMPILING.txt

WKFO:
I noticed that we didn't mention VS2017.
I broke the lines because that's what we did through the whole file. It's easier to read I think.


impaktor:
Is 2015 still supported by us?
@fluffyfreak ?

fluffyfreak:
Hmm... I've not actually tried it in a very long time. 
Just tried a build now and it needs some fixup but it works! Debug Win32 anyway.

I'll submit a PR to fix the remaining project issues in a few minutes 😁

fluffyfreak:
That PR is #4343

PR #4347, opened 2018-06-18
Systemview

richardpl:

### DIFF
--- a/src/SystemView.cpp
+++ b/src/SystemView.cpp
@@ -215,6 +215,7 @@ SystemView::SystemView(Game* game) : UIView(), m_game(game)
 	m_objectLabels = new Gui::LabelSet();
 	Add(m_objectLabels, 0, 0);
 	m_shipLabels = new Gui::LabelSet();
+	m_shipLabels->SetLabelColor(Color(255,155,0,200));
 	Add(m_shipLabels, 0, 0);
 	Gui::Screen::PopFont();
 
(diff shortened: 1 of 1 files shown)


fluffyfreak:
Ah ha! So that was what caused that bug! Nice work on figuring it out

src/SystemView.cpp
```diff
@@ -938,7 +939,14 @@ void SystemView::RefreshShips(void) {
 void SystemView::DrawShips(const double t, const vector3d &offset) {
 	m_shipLabels->Clear();
 	for(auto s = m_contacts.begin(); s != m_contacts.end(); s++) {
-		const vector3d pos = offset + (*s).second.OrbitalPosAtTime(t) * double(m_zoom);
+		vector3d rel_pos;
+		if ((*s).first->IsDocked() || (*s).first->IsLanded()) {
```
fluffyfreak: I think that if you change this test too
'if ((*s).first->GetFlightState() != Ship::FlightState::FLYING) {'
Then it should catch all of the non-flying states including when docking is in progress. Hyperspace, etc.

impaktor:
Ths must close at least one open issue? I know I've seen this on the issue tracker somewhere.

@richardpl I hope you keep sending us PRs (pull requests), so for future reference
1. try to give a descriptive title longer than one word (e.g. "fix ships stuck at centre of star")
2. try to have at least one sentence describing the issue/pr, (unless 1. does the job fully)
3. in PRs, make use of https://help.github.com/articles/closing-issues-using-keywords/ so it will auto close the issue it's fixing when it's merged.

Anyway, fantastic to have this bug fixed!

fluffyfreak:
The problem with this way of doing it, using `GetPositionRelTo`, is that we lose the prediction.

When it was using `OrbitalPosAtTime` it forward predicted the orbit of the ships

richardpl:
It predicted wrong because it put ships on wrong location.

PR #4355, opened 2018-06-22
add flyby camera view

### LINKED ISSUE #2115: Request, flyby camera
I know I love those flybys..
- Press a key, put it 1000 meters in front of ship in the direction of movement offset by a little in the x,y.
- Shouldn't be enabled at insane speeds? Or use some sort of inital velocity to make it work at any speed.

richardpl:
Closes #2115

Signed-off-by: Paul B Mahol <onemda@gmail.com>


impaktor:
Could you please hold on merging this until @fluffyfreak checks the coding style, so you agree on where to put `{` and such. But looks straight forward.

nozmajner:
Also, could you provide a bit more verbose description? Like how does it work, how do you trigger it, etc. For the sake of clarity.

richardpl:
Player triggers it with F1, like external, sidereal and internal view.

nozmajner:
Just tried it. It shows the ships in a quite unusual angle sometimes, even though you could still choose the same pow from the current external views. I guess you'd rarely rotate it that way. Which is quite nice in my opinion, less static

I think putting it on F1 isn't ideal. On one hand it is the only view related key, but this way we just have another view to cycle through. (and in my opinion the current two externals could be treated as one, and a key to switch between the sideral and other mode.)

Also the random 'dropping' of the camera feels odd. I mean when it resets while you fly, or even just stay put. Sometimes it even feels like the ship is stationary for half a second, before flying away, which is quite odd. (might be unrelated, but I never experienced any lag in change of speed from internal view)
This is where having a dedicated key for it would be beneficial, I think. (alt+F1 maybe?). You drop to the external view by pressing it, and if you press it again, it drops the camera again. Most games work this way in my experience, and it works better this way, since the camera doesn't jump around out of the blue. Would be more predictable.

Also a speed limit would be useful in my opinion. Even at several hundred m/s, you zip away quite fast.

Basing the view distance on the ship size might be useful too, for a Xylophis, I was unable to get a good view, the ship only showed as a dot, no matter what I did.

What does the +/- keys do? I'd think it's a zoom/focal length adjustment, but mostly rotates the camera a bit.

Another thought: I could try enlarging the main engine plumes a bit, would look good from this view.

richardpl:
-/+ just changes distance of camera from initial ship position. Using another key combination would be better. You can also roll camera left/right.

impaktor:
Regarding Keys, would it be possible to have F1+<number> to select view?

F1 + 1 back view
F1 + 2 left
...
F1 + 0 landing pad view

Would also have graphical interface, so when pressing F1, it would list in a sub-menu all icons for different views.

richardpl:
F1, and than pressing numerical keypad keys do this already.
Regarding submenu, not bad idea, but we are missing icons for all those 6 internal views.

nozmajner:
The [directional crosshair icons](https://pioneerwiki.com/wiki/Flight_UI#Direction_indicators:) should be good for that in my opinion. But reversing them might be more intuitive for the submenu. In worldview they aren't representing the direction, but point ship forwards, but as icons, that could be confusing I think.

richardpl:
OK, here is with alt+f1 way to toggle.

nozmajner:
Cool

impaktor:
Don't forget to type some text / instructions for the wiki, as we seem to be good att adding more keys, but not to document them. 

Looks like this could be more detailed:
https://pioneerwiki.com/wiki/Keyboard_and_mouse_controls

impaktor:
I think @richardpl is itching to merge this, so speak up now or forever hold your peace if wanting to object to this.

fluffyfreak:
I don't have, and won't get, time to test this for a few more days. 

Have other people used it?
How does it work?
What does it actually add since the description is non-existent?

richardpl:
@fluffyfreak This view, triggered with Alt+F1 will drop camera near ship and will always be oriented toward ship and center of screen will point to ship but camera otherwise will not move.

PR #4358, opened 2018-06-24
add racing missions

richardpl:
Try it, it is fun!


nozmajner:
Sorry it took this much to check it, sometimes we need some nudging. I just flew a race and here are my observations:
- The advert texts are very minimalistic, lacking any kind of description.
- I won the race I flew, but other than the end result, there's not much to know about the whole thing, which robs a lot of the race-ness of the race. No info on the opponents for example, nothing you can compare your performance to.
- Are there any other ships flying the race, or it's just against a deadline?
- It's kinda odd, that the race is just an advert you pick up, then fly away. I guess it could be a "best time between X any Y" kind of thing that's always on. But then there should be a list of the best times you could beat. Like ten entries for example, and if you beat one of them, you get a reward, and the better place you beat, the more money you get. These could be hardcoded races, and we could fill up the best-of list with actual data, and players could share theirs, and we could incorporate them to each release, updating it.
- It would be nice, if the advert would just say, where and when the race is starting, and the player could get there. Like they get some coordinates to fly to, where all the other participants are waiting, and when the time is up, they all blast off. And if it's close enough to a station, then they won't even jump away until they leave the restricted zone. And all the ships would actually fly the race. And this way, the player could even witness race starts, even if not participating.
 
And a thing to note: our ships have quite diverse deltaVs, and I suspect, there will be one or two top choices that would blew anybody out of the water. Not that it's a problem, but it's good to be aware of it.

impaktor:
I haven't tested this (sorry), but based on what nozmajner, here are some ideas I got (maybe that's how it already works?):

- it would be cool if your racing results were either displayed in your Personal Info / F3 page "Flew the Kessel run in under 3 Parsec" (yeah-yeah, I know not a time measurement, I couldn't resist the star wars reference), or have the same result permanently listed on the BBS (faction specific for race flavour?), with the 3 best pilots?

- Maybe the different flavours are each annually?

Again, sorry if this is how it already works, and I'm not saying this is how it aught to work, just throwing out some ideas to differentiate it from normal missions.

Oh, also for what nozmajner writes, about a common starting location, sounds like we, ideally, need buoys for that?

WKFO:
I think this is a good mission. I just read the comments above, and people have good points. There could be a starting time and place where all participant ships will have to go and the race will begin there at that time, so you won't (start) fly(ing) alone.

Other than that, I don't have anything to say. And if Pioneer ever gets any networking feature, this is one good place where we can use it.

fluffyfreak:
Ok the only reason this hasn't been merged yet is because there's been no response about the feedback.

Races might be a great idea, but this implementation is very bare bones, and I don't want to merge something that is then just going to be abandoned.

Web-eWorks:
Going to close this PR as abandoned/changes required due to three years of inactivity. The original author is free to reopen, or other module authors are welcome to copy the code and open a new PR if they're willing to carry it the rest of the way.

PR #4359, opened 2018-06-26
Fix showing Pale-Blue-Dot and other planets again.

### LINKED ISSUE #4190: The pale blue dot is the Earth
punt_blau

I suggest to draw distant planets as dots, so you can say: "the pale blue dot is the Earth", or the red dot is Mercury, or the White dot whit tiny dots is Jupiter and its satellites. And when you are aproaching, the dots are getting bigger until a texture is rendered.

I think it's an easy way to add realism to the game. (My old Spectrum ZX80 was very good drawing dots :-) ). Now, the effect is a little strange, you are watching a point in the universe and instantly a planet is rendered, spontaneous generation? :-( I wish to see planets as dots.

fluffyfreak:
Fix a bug whereby the billboards were not being drawn for planets once they had gotten far enough away for that to be an option.

We have an onscreen pixel threshold below which we render planets and other objects as billboard.
However I broke that rendering when I rewrote the billboard code back in Apr 3, 2016!

To fix this the commit uses the billboard shader and the onscreen pixel size for a planet.

Fixes #4190

### DIFF
--- a/src/Camera.cpp
+++ b/src/Camera.cpp
@@ -85,6 +85,7 @@ Camera::Camera(RefCountedPtr<CameraContext> context, Graphics::Renderer *rendere
 	m_renderer(renderer)
 {
 	Graphics::MaterialDescriptor desc;
+	desc.effect = Graphics::EFFECT_BILLBOARD;
 	desc.textures = 1;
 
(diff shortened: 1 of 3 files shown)


richardpl:
Tested, works as intended.

PR #4360, opened 2018-06-26
Split what needs to be done for pick up mission for cargo run

### LINKED ISSUE #4027: 2 trips missions not showing progress
### Observed behaviour
When i open the mission it shows me the 2 trips i'll have to make but doesnt show which is completed.

### Expected behaviour
When the first trip is completed it shows so in the mission details. 

### Steps to reproduce
Accept a missions that requires you to go to one planet, pick cargo and deliver to anoher planet.

pioneer 20170415-1 from AUR. Archlinux.

richardpl:
So user do not need to memorize next step.

Closes #4027.


impaktor:
Could you please hold off on merging this until we gotten a bit more feedback.

At least from those who either made it @robothauler and / or someone who played it?

Thanks.

robothauler:
I think, now we have insufficient information if the cargo is not loaded.
If the player has a long list of local cargo missions and a ship with limited cargo space, he needs more info to plan ahead.
I would only add some indicator if the cargo is already loaded or not.
e.g. Amount: 2t loaded/not loaded
or something like that.

richardpl:
@robothauler is this ok now?

richardpl:
Will apply soon.

PR #4362, opened 2018-06-30
Turn ship towards destination at end of voyage

### LINKED ISSUE #293: Autopilot should turn the ship towards the destination at the end of the voyage
This is no bug, but i'd like to have my starship facing the destination at the end of an autopilot travel instead of manually turning my ship everytime to face the planet i have arrived.

richardpl:
Closes #293.


impaktor:
I guess @fluffyfreak or @jaj22 are the ones who might be interested in voicing an opinion.

src/ShipAICmd.cpp
```diff
@@ -910,7 +910,18 @@ Output("Autopilot dist = %.1f, speed = %.1f, zthrust = %.2f, state = %i\n",
...
 
 	// face appropriate direction
-	if (m_state >= 3) m_prop->AIMatchAngVelObjSpace(vector3d(0.0));
+	if (m_state >= 3) {
+		vector3d pos;
+		if (m_target)
+			pos = -m_dBody->GetPositionRelTo(m_target);
+		else
+			pos = -m_dBody->GetPosition();
+		m_prop->AIMatchAngVelObjSpace(vector3d(0.0));
```
fluffyfreak: Won't this just get overridden by the subsequent call too `AIFaceDirection`?

fluffyfreak: You could but then you'd just override it the other way around.
I think that what needs to happen is that either the ship turns around and *then* calls `AIMatchAngVelObjSpace`, or it needs to call `AIMatchAngVelObjSpace`, wait for that to complete and then call `AIFaceDirection`

richardpl: Done.

richardpl:
I think I nailed everything, so this should be ready for merge.

jaj22:
Should lose the same-frame check on line 915. As far as I can tell, the code gives incorrect results for cases where the target isn't the frame origin (you'd want to point at the target, right?).

This does also make ships with retro < ventral thrusters much more dangerous at high time accel, although that was already an unsupported case. Ideally you wouldn't turn towards the target unless it's the final stage of navigation, but there's no way to know that in AICmdFlyTo.

richardpl:
I can make this work only for player ship. I tested both cases when target frame match and when it does not.

richardpl:
Hopefully this one is ok?

jaj22:
Probably won't break anything. However, isn't this only wanted in the fly-to-vicinity case, while docking and orbit waypoints should really avoid it? Unfortunately I think you'd need to add state (or a wrapper class) for that, as there's currently no clean way to distinguish.

richardpl:
Sure, is this now OK?

jaj22:
You'd need to serialize that parameter too. Doesn't strictly need a save bump as long as the serialization code can handle default params (default false is fine), but I'm not sure how you write that.

richardpl:
Done, anything else?

Issue #2698, opened 2014-01-30
flight controls remain active during typing save game name


Baobobafet:
While I was busy typing a name for my game save, I heard what sounded like a missile launch after incidentally typing the letter 'M' as part of the save title I chose.

Sure enough, when I checked, I was down one missile. 
When I loaded the save, it could be seen flying away from my ship.

Wouldn't leaving flight controls enabled on the save game screen cascade to other unintentional save and piloting errors?

richardpl:
This does not happen any more. Closing.

PR #4364, opened 2018-07-02
Do not toggle bindings when in settings view

### LINKED ISSUE #538: Function keys still respond while changing controls
Repro steps:
1. Start game.
2. ESC to game menu, and select Controls tab.
3. Press one of the GUI buttons to remap a control.
4. Press F3.

What happens: Control is remapped, _and_ the view changes to the ship info view.
What should happen: Control is remapped, and nothing else happens.

richardpl:
Previously keys like missile fire: M would trigger when changing keys. Closes #538.

### DIFF
--- a/data/pigui/game.lua
+++ b/data/pigui/game.lua
@@ -933,9 +933,11 @@ ui.registerHandler(
 			if not ui.showOptionsWindow then
 				Game.SetTimeAcceleration("paused")
 				ui.showOptionsWindow = true
+				Engine.DisableBindings();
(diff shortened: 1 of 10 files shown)


impaktor:
I think we have other open(?) bug reports that flight control keys are still active when poking around in menues or typing save game names? Or that might have been fixed? Anyway, just thinking out loud.

richardpl:
The typing saving games with m is fixed, and I closed issue report #2698.

PR #4365, opened 2018-07-04
translation updates. running everything through jq, whitespace, sort

ecraven:
Update current translations from transifex.
Run everything in `data/lang` through `jq`, to auto-format and sort correctly.


impaktor:
```
<ecraven> en.json is never fetched from transifex, only ever uploaded to there   [16:22]
```
When a new asset is created, one has to add the en.json manually, but then any change in the master repo will migrate up to transifex, and then back to the other files.

ecraven:
> When a new asset is created, one has to add the en.json manually, but
> then any change in the master repo will migrate up to transifex, and
> then back to the other files.

Also, new modules need to be added to .tx/config (the transifex module
configuration). Any time now I'll put all this together into some sort
of developer documentation ;)

impaktor:
Yes, that's what I do when new resources are added, and rob / jpab before me.

PR #4366, opened 2018-07-04
StarTypeInfo cleanup

fluffyfreak:
StarTypeInfo does not need or use the additional information that each entry is a star.

Noticed this redundant information a while ago.


impaktor:
So it's never a black hole or dyson sphere? Or those are also star types?

fluffyfreak:
'StarTypeInfo' only includes info about stars, black holes counts as those down at the end of the list.

fluffyfreak:
However the info field was never used anyway

PR #4369, opened 2018-07-05
Radial menu

ecraven:
Right-click and hold on in-space bodies (planets and stations). You should see a radial menu, with working commands.

The new system overview window (toggle with f4 in-flight, or click the button in the right upper corner) replaces the right side of the old f4 menu.

foo


richardpl:
I can not use this radial thing. My touchpad have only one button. So it can not trigger actual operation because at same time needs right and left button pressed.

The target window info window is always on even on BB - this needs changing.

What about replacing current functionality of F4, just displays possible stations, in transparent window.

impaktor:
"Fly to" angry police ship:

```
error: [string "[T] @pigui/modules/info-window.lua"]:46: attempt to index local 'sb' (a nil value)
stack traceback:
	[string "[T] @pigui/modules/info-window.lua"]:46: in function 'fun'
	[string "[T] @pigui/pigui.lua"]:66: in function 'window'
	[string "[T] @pigui/modules/info-window.lua"]:18: in function 'fun'
	[string "[T] @pigui/pigui.lua"]:111: in function 'withStyleVars'
	[string "[T] @pigui/modules/info-window.lua"]:17: in function 'fun'
	[string "[T] @pigui/pigui.lua"]:103: in function 'withStyleColors'
	[string "[T] @pigui/modules/info-window.lua"]:16: in function 'fun'
	[string "[T] @pigui/game.lua"]:677: in function 'callModules'
	[string "[T] @pigui/game.lua"]:1006: in function 'fun'
	[string "[T] @pigui/pigui.lua"]:66: in function 'window'
	[string "[T] @pigui/game.lua"]:994: in function 'fun'
	[string "[T] @pigui/pigui.lua"]:103: in function 'withStyleColors'
	[string "[T] @pigui/game.lua"]:993: in function 'game'
	[string "[T] @pigui/pigui.lua"]:123: in function <[string "[T] @pigui/pigui.lua"]:%   
```

richardpl:
Needs disabling old f4 view, also it needs to be toggle able with F4 key. Otherwise one can not access station screen. Disable titlebar and increase height of window. With all mentioned changes it is ready for merge.

nozmajner:
@ecraven 
A bit of mocking up:
The icons is a bit hacky right now, rotating and mirroring the system overview one, but it does work.
It's button is in the top right, and opens it up/closes it, similarly to the nav log (which could have a similar presentation instead of the plain imgui window)
Closed:
nav-panel-off

Open:
nav-panel-on
The heigth is half the screen, but I could imagine it to be resizeable, but the default should be good anyway.
nav-panel-resized
Width might need some extension though, for longer names. But there should be a limit to that.

I've put icons in for the filters (Moons, Settlements, Dist), and a funnel icon for the filtering.
Also, a window title.
I've also put a settlement and orbital icon after the names of according bodies (so it's visible, which does have one, even if that filter is off.
I've also put [Earth] in brackets to indicate that it is the current frame of reference.

Semi related: I guess it would be somewhere around F8, if the F keys are updated to the icon positions.
I think it would make sense, but also that would mean that autopilot, landing gear and hyperdrive  commands would be moved from the F keys then. Which might be a good thing, since none of that is UI related. (H could be hyperspace, B could be landing gear, autopilot could be Z for example.)

richardpl:
There is already an icon for F4.

nozmajner:
OK, I thin I misread. F4 does have an icon, but that icon is more comms related

ecraven:
This is reaching a state where I'd consider removing the WIP tag (and getting it ready to be merged).
Please test latest, and comment with any issues.
Notes:
- The radial menu does not work on the overview window (right side) right now, still looking into how to fix that.
- It *does* however work on the nav target name displayed to the right of the reticule

nozmajner:
Finally it finished building.  Looking nice.
Is it intentional that you can't open it while landed? I'd imagine it would be useful to be able to pick your target before blastoff.
I noticed that the tooltip doesn't pick the name of the target. I right-moused on the Moon, and it said Fly to the vicinity of Jupiter (which I also targeted before targeting the Moon). It still flew to the Moon though.

nozmajner:
New icons:
radialhoz
[And the svg file for them.](https://www.dropbox.com/s/ze45flp8grhb3cm/icons.svg?dl=0)

nozmajner:
Looking good to me mostly.
I noticed two things:
- The alt+ orbital direction menu is a bit inconsistent. It pops up on any body indicator and the menu, but does not pop up when clicking on the name on the reticule. But now that I thought about it, it doesn't really relate to your target at all, but your current frame and orbit. I think it could be a general alt+RMB that you could use anywhere in worldview (possibly even in sysview later), and not on the targets. Or maybe at the frame indicator on the reticule only.
- It's somewhat cosmetic, but I'd also rearrange the orbit direction icons in the radial too:
radialmenu-orbiticons-crop
This way each "axis" refers to one of the orbital directions + -.

ecraven:
Should be ok to merge, cleaned it up a bit further and nicened up history.

PR #4371, opened 2018-07-06
Deny docking requests too far from station

WKFO:
I thought reserving a docking port from 11AU was insane and space stations' air traffic controllers shouldn't grant such clearances. If I were them, I wouldn't. Would you? (I would be suprised if someone requested clearance from 500.000km away and made it in time.) I might just need a more proper excuse to deny those requests, though (current message is a placeholder). Come up with a good excuse and I will change the message :)

'Too far' as in 'more than 100km'. Which is a long enough braking/slow-travelling distance in my opinion.


WKFO:
WHOA! Conflict?

fluffyfreak:
@WKFO Nothing to worry about, just that the language files got updated.

WKFO:
Okay, it was just unexpected.

nozmajner:
@WKFO: does the 'Dock with" autopilot knows about it?

WKFO:
'Dock with' autopilot doesn't know about it on the code side but it requests for docking at 15km distance.
Tested it tens of times, had no problem so far.

nozmajner:
Cool.

richardpl:
This can not be merged as branch have conflicts.

WKFO:
Okay, will try to fix.

WKFO:
I don't know what I just did but it showed like the entire files was replaced. Reverting...

richardpl:
Merged.

PR #4373, opened 2018-07-06
Uninvert system map camera

Garmelon:
This inverts the vertical axis of the system map camera and makes it behave similar to the sector map camera.


nozmajner:
Wow, I guess I filtered this out during the years. Thanks for the fix!

PR #4374, opened 2018-07-06
Add a new ship 'Lodos'

WKFO:
So I think it is time we have the PR for this.
Thanks a lot @nozmajner and everyone else involved in this.

Designed as a cargo freighter. It is a compact design, no equipment scattering around the hull, which makes it a good choice for atmospheric flight. It lacks a bit of capacity compared to other medium freighters, but it has a plenty of Delta-V and not-bad-at-all thrusters. Also it is cheaper than most (if not all) other medium ships.

Autopilot works well with the ship as far as I've observed.

Fresh out of the shipyard. Take her out for a flight. Report any problems you encounter!

https://youtu.be/f_JRmXwpsLk


WKFO:
@impaktor 
https://imgur.com/a/SwQEGmY

WKFO:
Pity I still couldn't come up with a paint pattern other than a full body single color paint.

nozmajner:
Leave that to me, I'll paint up something cool.

nozmajner:
To make sure that they won't get lost:
[Lodos patterns](https://www.dropbox.com/s/47q6rdmz82fmwin/lodos_patterns.zip?dl=0)
modelviewer-20180707-223212_s
modelviewer-20180707-223223_s
modelviewer-20180707-223235_s
modelviewer-20180707-223241_s

fluffyfreak:
@WKFO and @nozmajner so is this all ready to review/merge now?

WKFO:
It is ready.

data/models/ships/lodos/lodos.model
```diff
@@ -14,5 +14,4 @@ anim gear_down 1 121
 collision lodos-col.dae
 
 lod 20
-mesh lodos-lo.dae
-
+mesh lodos-lo.dae
```
impaktor: No end of line here? Is this generated by blender (if so, I guess it's OK) or by hand (if so, then maybe a "RET")?

nozmajner: It's by hand.

WKFO:
EOF should have a new line now

PR #4375, opened 2018-07-07
Combat mission icon

nozmajner:
Adding the combat mission icon, and setting it up in the Combat.lua mission file.
The icon is the light figter ship class icon, as picked by @robothauler :) 
screenshot-20180707-164534


nozmajner:
Wow, that screenshot brought it's alpha transparency for sure

PR #4376, opened 2018-07-08
Make it possible to change position of flyby camera

richardpl:
With cursor key or with 4682 numpad keys.


fluffyfreak:
Looks good

PR #4377, opened 2018-07-09
Add heavy atmospheric shielding, allows easier fuel scooping

richardpl:
"Stolen" from @impaktor old branch.

### DIFF
--- a/data/lang/equipment-core/en.json
+++ b/data/lang/equipment-core/en.json
@@ -15,6 +15,14 @@
     "description": "",
     "message": "Shields your spaceship from the heat of atmospheric re-entry."
   },
+  "ATMOSPHERIC_SHIELDING_HEAVY": {
+     "description": "Will withstand higher pressures than normal Atmospheric Shielding",
+     "message": "Heavy Atmospheric Shielding"
+  },
+  "ATMOSPHERIC_SHIELDING_HEAVY_DESCRIPTION" : {
+     "description": "",
+     "message": "A more capable version of Atmospheric Shields"
+  },
   "AUTOPILOT": {
     "description": "",
     "message": "Autopilot"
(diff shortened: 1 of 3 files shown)


fluffyfreak:
This doesn't look like it help with pressure damage it just **increases** the drag factor which would make things worse for fuel scooping (_for example_) and for travelling in atmospheres.

It looks like it should instead affect the calculation around line [Ship.cpp LN 1092](https://github.com/pioneerspacesim/pioneer/blob/master/src/Ship.cpp#L1092) perhaps put the multiplier factor there instead?

richardpl:
My math skills tell me this decreases the drag factor...

fluffyfreak:
@richardpl dammit, sorry you are correct.

Is it still worth using the shield to increase the pressure resistance too?

richardpl:
Yes, but by how much?

fluffyfreak:
I'd guess that for now we try it with the same multiplier. If it needs to become a separate parameter later that's not too much work.

fluffyfreak:
@richardpl are you going to make the change to the [relevent line](https://github.com/pioneerspacesim/pioneer/blob/master/src/Ship.cpp#L1092)?

richardpl:
Not yet.

fluffyfreak:
I added the changes so it increase the pressure resistance according to the `atmo_shield_cap`. Capped it with a minimum so it respects the default atmospheric pressure resistance otherwise you start taking damage immediately upon takeoff.

fluffyfreak:
I think that ships without an atmo' shield will behave in exactly the same way as they used too.

The old pressure test never took into account having an atmospheric shield at all so multiplying pressure resistance by `1` is the same as the old test. It only improves if you have a heavy shield.

PR #4378, opened 2018-07-09
Acceleration cap

Garmelon:
Add the option to cap the maximum acceleration for each thruster on a ship.
Can be set via the `<direction>_acceleration_cap` field in the ship.json.
If no such field is set for a thruster, it is not acceleration capped.

### DIFF
--- a/src/LuaShip.cpp
+++ b/src/LuaShip.cpp
@@ -1189,7 +1189,7 @@ static int l_ship_is_hyperspace_active(lua_State *l)
 static int l_ship_get_thruster_state(lua_State *l)
 {
 	Ship *s = LuaObject<Ship>::CheckFromLua(1);
-	vector3d v = s->GetPropulsion()->GetThrusterState();
+	vector3d v = s->GetPropulsion()->GetLinThrusterState();
 	lua_newtable(l);
 	pi_lua_settable(l, "x", v.x);
 	pi_lua_settable(l, "y", v.y);
--- a/src/Missile.cpp
+++ b/src/Missile.cpp
@@ -125,10 +125,10 @@ void Missile::StaticUpdate(const float timeStep)
 	//Add smoke trails for missiles on thruster state
 	static double s_timeAccum = 0.0;
 	s_timeAccum += timeStep;
-	if (!is_equal_exact(GetPropulsion()->GetThrusterState().LengthSqr(), 0.0) && (s_timeAccum > 4 || 0.1*Pi::rng.Double() < timeStep)) {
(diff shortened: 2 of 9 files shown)


Garmelon:
The autopilot is still not working correctly with acceleration caps.

impaktor:
This will for sure need discussion. Is UI satisfactory, is it a function we want, are there sideffects (e.g. autopilot?)

jaj22:
So I think one problem here is that when the autopilot sets a thruster to 1.0, it expects to get the thrust returned by GetThrust. With this patch, that's not the case unless acceleration is uncapped.

Also some issues with function names and optimization, but let's try to get it working first.

Garmelon:
But GetThrust returns the capped value and ClampLinThrusterState adjusts the thrust down so that setting the thruster to 1.0 means getting the thrust from GetThrust (the capped thrust).

jaj22:
0.0 and 1.0 are the only inputs that give the correct result. Consider if the autopilot requests 0.9 when the cap is 0.5: Expected value is 0.45, actual value 0.5.

Garmelon:
Sorry, I'm probably being dumb here and missing the obvious.

In ClampLinThrusterState, if the cap was set to 0.5, that means that `GetThrust(thruster)` is half of `m_linThruster[thruster]` with the ship's current mass.
First, the autopilot's 0.9 is clamped to [-1.0; 1.0] (which I forgot in the second `ClampLinThrusterState` function), then multiplied by `GetThrust(thruster) / m_linThruster[thruster]`, which should be equal to 0.5.
Thus the result would be 0.9 * 0.5 = 0.45.

jaj22:
Ok, you're right. I probably assumed that ClampLinThrusterState would just clamp.

It looks like `SetLinThrusterState(vector3d)` is double-translating in that case though? Line 154+ appears to be translating the `m_linThrusters` values which are already translated by line 153.

Garmelon:
Thanks a lot, that fixes all the autopilot bugs so far!

fluffyfreak:
@Garmelon / @impaktor  @jaj22 
I get that this can limit the acceleration for specific thrusters but I don't know why you would want to do that and how the player would know what it was doing.

Can someone explain the why of all this and how I can test it before it gets merged?

nozmajner:
@fluffyfreak It's to keep the ships from reaching the likes of 11-15G accelerations at low fuel/load, to aid in the rebalancing of ship stats. A bonus is that the brake gauge becomes more accurate.

nozmajner:
@Fluffyfreak Actually I asked for this when @Garmelon came around and asked what stuff is needed :)
I've already started rebalancing the stats utilizing this feature, but I will ne out of sight for a month at least. I'll finish em up after I get back.

PR #4379, opened 2018-07-09
Fix spotted issues with SAR mission

richardpl:
Fix crash if target ship is destroyed in middle of transfer.
Change mission location after completed transfer.


impaktor:
I'd like feedback on this from @robothauler, if possible.

impaktor:
Ops, my bad, it looks like this was written by @clausimu actually, so if you have time,  @clausimu, please give some feedback.

PR #4381, opened 2018-07-11
Units library

Garmelon:
As talked about on IRC: https://github.com/nholthaus/units/

> a compile-time, header-only, dimensional analysis and unit conversion library built on c++14 with no dependencies.

`<ecraven> I'd like at least fluffy, johnj and durandal to chime in with an opinion`


Garmelon:
Damnit, the attempt to change to c++14 didn't work with travis. C++14 or above is necessary for the units library.

> However, you'll have to properly set up the necessary compilation flags for C++14 (-std=c++14 on gcc).

ecraven:
Personally, I'd welcome this, I myself sometimes have trouble telling which exact unit something is, this would enforce that at least at the C++ level without any performance overhead. I'm very willing to aid with converting the existing codebase to actually using `units`.

impaktor:
We can discuss if we want/object to this here (@fluffyfreak @laarmen ), but there would have to be commits in this PR that actually make use of it --- probably not a good idea to merge libraries without using them, as contributors come and go.

impaktor:
I have no real opinion. I'll leave it up to you professional programmers if we need/should have this.

fluffyfreak:
I'll need to try this in VS201X to see if it works there. I'd imagine that it will mean removing support for VS2015 which at least one person still uses.

fluffyfreak:
Also is there any example of using this? We don't normally just pull in a library without it being used somewhere.

fluffyfreak:
Ok, checked with VS2017 and VS2015. Both seem ok with it. Although I don't have any code changes to check that it works which makes me a little nervous.

laarmen:
I'm all for the idea itself, and looking at the doc it seems nice, so I'd
go ahead and add a few conversions to this PR :)

ecraven:
> I'm all for the idea itself, and looking at the doc it seems nice, so I'd
> go ahead and add a few conversions to this PR :)

OK, I'll add it in a few places later, then we can discuss it concretely
;)

Garmelon:
What version of gcc should travis use? 5.4, 5.5, 8.1, ...

ecraven:
After working a bit with units in Orbit.h/cpp, I really like it. I think it serves as excellent documentation of what all the different numbers are, and it ensures that you don't mix incorrect units. After my experience with it I'd suggest merging this and using it for the entire codebase.
One caveat, the compilation time does go up by quite a bit, maybe there is a way to use something like precompiled headers to mitigate this?

impaktor:
Given both andy and ecraven are afk for a week, we'll let this sit for a while.

fluffyfreak:
@impaktor well it's been a week! ;)

fluffyfreak:
From our IRC conversation it seems that this really is a good idea... but it comes with a 30% longer build time. Our build times are already worse than they should be.

Dicsuss

pirogronian:
I noticed that 'make' build everything after every 'git pull', no matter how many files was modified. On the other side I know ever headers can be in precompiled state (*.gch files). Remembering build state should greatly reduce rebuild time.

fluffyfreak:
Closed as I don't think is going anywhere

ecraven:
I will revisit this topic later, after trying to improve build times with existing code / refactoring. I still believe that there is a lot to be gained by introducing (some sort of) units library, but I first need to figure out how this would influence lua (I'd much prefer to handle everything possible in C++ with strict unit type checking than in lua, which won't have any sort of units library ever, I think).

Issue #4383, opened 2018-07-13
Crash when removing hyperjump routes


pcercuei:
### Observed behaviour
When removing the hyperjump routes one by one, once all the routes have been removed, pressing the remove button again crashes Pioneer.

### Expected behaviour
Shouldn't crash.

### Steps to reproduce
Add one route, then another one. Select the last one, click the remove button to remove it. Click the remove button again, this will remove the first route. Clicking the remove button once again will crash Pioneer.

__My pioneer version (and OS):__
Latest master on Linux

PR #4384, opened 2018-07-13
Fix issue #4383

Garmelon:
This fixes issue #4383 by bounds checking before removing an element.


ecraven:
Looks good to me, merge away ;)

PR #4387, opened 2018-08-03
Add additional tooltips to the settings menu

impaktor:
# Add additional tooltips to the settings menu
Having mouse over tooltip for only half of the options in settings menu bugged me. Now, I'm just bugged by some of the tooltip starting with capital letter, and some with lowercase. Not sure what is the most sane? 

@fluffyfreak care to skim through the strings I've added and see that stuff does what I say they do?

## Three for one, one for all!
Planet, City and Fractal detail all use the same generic description: 
```json
  "DETAIL_DESC": {
    "description": "tooltip description in settings menu",
    "message": "Lower gives better performance, at cost of visual aesthetics"
  },
```

## Still lacking description
There are still two options lacking description. Not sure what to put here.

- `Fullscreen` - does this have any impa**k**t on performance?
- `Starfield density` - does this have any impakt on performance?

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -279,6 +279,10 @@
     "description": "",
     "message": "Destroy enemy ship"
   },
+  "DETAIL_DESC": {
+    "description": "tooltip description in settings menu",
+    "message": "Lower gives better performance, at cost of visual aesthetics"
+  },
   "DISMISS": {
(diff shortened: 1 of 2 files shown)


fluffyfreak:
- Fullscreen: might impact performance on some machines
- Starfield Density: might impact performance on very low end machines

fluffyfreak:
These all look good to me.

PR #4388, opened 2018-08-09
Provide :other_names to add multiple names to custom systems.

ecraven:
They can by accessed in lua with `starsystem.other_names` (which returns
`{ "name a", "name b" }`), but are not used anywhere yet.


fluffyfreak:
So is this WIP?

PR #4391, opened 2018-08-11
Add support for using an external libglew

pcercuei:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -341,6 +341,13 @@ if test "$with_external_liblua" = yes; then
 fi
 AM_CONDITIONAL([HAVE_LUA], [test "$HAVE_LUA" = yes])
 
+dnl Optionally use external libglew
+AC_ARG_WITH([external-libglew], AS_HELP_STRING([--with-external-libglew], [Use external libglew in place of our internal copy]), [], [with_external_libglew=no; HAVE_GLEW=no])
+if test "$with_external_libglew" = yes; then
+  PKG_CHECK_MODULES([GLEW], [glew >= 2.0.0], [HAVE_GLEW=yes], [AC_MSG_ERROR(["External libglew was requested but is not available"])])
+fi
+AM_CONDITIONAL([HAVE_GLEW], [test "$HAVE_GLEW" = yes])
+
 LUA_CFLAGS="$LUA_CFLAGS -DLUA_USE_APICHECK"
 case "$host" in
   i686-w64-mingw32.static)
--- a/contrib/glew/Makefile.am
+++ b/contrib/glew/Makefile.am
@@ -1,6 +1,11 @@
 #  Process this file with automake to produce Makefile.in
+
+if !HAVE_GLEW
+
 include $(top_srcdir)/Makefile.common
 
 noinst_LIBRARIES = libglew.a
 libglew_a_SOURCES = glew.c
 noinst_HEADERS = glew.h glxew.h wglew.h eglew.h
+
+endif
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -315,12 +315,11 @@ pioneer_LDADD = \
 	../contrib/PicoDDS/libpicodds.a \
 	../contrib/jenkins/libjenkins.a \
 	../contrib/json/libjson.a \
-	../contrib/glew/libglew.a \
 	../contrib/profiler/libprofiler.a \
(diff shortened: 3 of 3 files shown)


impaktor:
@fluffyfreak anything "libglew"/graphics-related, and I'm out. I assume you have deeper understanding on this stuff.

I'm not sure why Travis tries to build this against the third-party repo?

```
254.47s$ git clone --depth 1 git://github.com/pioneerspacesim/pioneer-thirdparty.git pioneer-thirdparty
Cloning into 'pioneer-thirdparty'...
fatal: unable to connect to github.com:
github.com[0: 192.30.255.113]: errno=Connection timed out
github.com[1: 192.30.255.112]: errno=Connection timed out
The command "git clone --depth 1 git://github.com/pioneerspacesim/pioneer-thirdparty.git pioneer-thirdparty" failed and exited with 128 during .
```

fluffyfreak:
@impaktor Nope, in general this is Linux makefile madness 😛 
However I thought that we always link against thirdparty since it's the stuff that we know works with our code?

@pcercuei this bugs me. You can't just post a PR like this without telling us *why* you want this PR merged. 
Having a "description" like:
> Signed-off-by: Paul Cercueil paul@crapouillou.net`

Is not very helpful. It doesn't tell me why you want to make these changes, what effect it will have on builds or users, or who else might want it.

We have a template instructions for new PR's so that other people can understand why you would want to merge something.

laarmen:
I agree with @fluffyfreak on this one, even though I know exactly why one
would want external libglew support on Linux.

Basically, bundling dependencies together with the main app is frowned upon
in most distributions, as it makes it much harder to apply system-wide
security fixes and other global improvements. Regardless of the merits of
this position (it has merits, I assure you!), I don't see why we shouldn't
allow it : I used to link against the system libs when I worked on Pioneer,
but that was a looong time ago. IIRC we even have some support for using
the system Lua :). For a Debian package, this patch would make things
easier, so I'm assuming that's also the case for the newer, shinier
packaging thingies that came along recently?

Glancing at the code, the external lib support is opt-in, so there
shouldn't be much changes for most users, only more possibilities.

pcercuei:
I thought the usecase was obvious, sorry. Thanks @laarmen for explaining.

fluffyfreak:
Ok thanks for clarifying @laarmen, I'll leave this to yourself or @impaktor to resolve.

impaktor:
Well, I can push the merge button, but I wouldn't understand what I'm doing, but as pointed out, this only adds additional options, so _should_ be harmless. Mostly.

johnbartholomew:
Re: the Travis failure - that looks like an unrelated transient network problem somewhere between Travis and github, rather than anything specific to this CL. The Travis config has cloned pioneer-thirdparty and configured to build against it since it was first added (see `before_install` commands in the Travis config: https://github.com/pioneerspacesim/pioneer/blob/master/.travis.yml )

I don't think Travis will give any useful signal about whether this option works as expected, since we don't try to get Travis build coverage of all different build configurations. Let me do a couple of test builds...

johnbartholomew:
Looks like it works ok. Builds with and without external GLEW, ldd shows it linking to libGLEW.so in the external-GLEW build (and not when using the bundled contrib version). Both versions start up and display things as expected.

pcercuei:
Thanks!

johnbartholomew:
Unfortunately I just found a problem with this. For most of the build, the contrib/ directory is in the include path. We have source files that include "glew/glew.h", which works for the local glew because it finds it (in contrib/glew/glew.h). However, the system header is /usr/include/GL/glew.h. That obviously won't be found at "glew/glew.h"; it would be found at <GL/glew.h>.

So I think with this patch and when configured to use system glew this will actually link against the system library but use the header from the contrib/ library. That will cause a version skew problem except for people who are "lucky" enough that their installed system glew version exactly matches the one we've bundled.

We'll need to move the headers around in contrib/ so that the include structure is more uniform and then conditionally add contrib/glew (or whatever) to the include path when we're using the bundled copy. Or some other fix.

impaktor:
> We'll need to move the headers around in contrib/ so that the include structure is more uniform and then conditionally add contrib/glew (or whatever) to the include path when we're using the bundled copy. 

Remind me: was this done, or is it still an issue? If the latter, I suggest opening an issue so we track this.

PR #4394, opened 2018-08-12
Reduce stock to 0 for non-purchaseable items

### LINKED ISSUE #4276: Stations won't allow selling of equipment they aren't having due to their tech level
### Observed behaviour
I noticed that you can't sell your equipment, if the station doesn't have it due to it's tech level being low. 

### Expected behaviour
One should be able to sell equipment in my opinion, even if the station doesn't trade in it. I might even imagine they would be more than happy to put their hands on it, so they'd pay a bit more for it. Or less, because it's more work for them, and can't really use it anyway.

### Steps to reproduce
Open up the equipment store.

2017 Oct 1 official build, Linux64.

fluffyfreak:
This fixes #4276 or at least tries to avoid the main problem. 
It reduces the available stock of parts to zero for stations that don't have a high enough tech level.

This means that ships can still sell their equipment, but there is never any stock available even after a sale.

### DIFF
--- a/data/ui/StationView/EquipmentMarket.lua
+++ b/data/ui/StationView/EquipmentMarket.lua
@@ -33,7 +33,15 @@ local equipmentMarket = function (args)
 		stationColumns = { "name", "buy", "sell", "mass", "stock" },
 		shipColumns = { "name", "amount", "mass", "massTotal" },
 
-		canTrade = function (e) return e.purchasable and hasTech(e) and not e:IsValidSlot("cargo", Game.player) end,
+		canTrade = function (e) return e.purchasable and not e:IsValidSlot("cargo", Game.player) end,
+		-- how much of this item do we have in stock?
+		getStock = function (e)
+			if hasTech(e) then
+				return Game.player:GetDockedWith():GetEquipmentStock(e)
+			else
+				return 0
+			end
+		end,
 	})
 
 	return


johnbartholomew:
Looks like this works as intended. Note that if you sell a piece of equipment that has a tech level higher than the station then it still won't appear in the station's stock list so you cannot buy it back. I guess that's acceptable?

johnbartholomew:
Uh.. I don't know why I pointed that out, fluffy already noted it in the original PR comment. Sorry.

impaktor:
I haven't tested this (My pioneer computer is bricked until tuesday), but will this then show all available equipment in the game, regardless of tech level?

Meaning: dock to a tech level 1 station, and you see all equipment the game offers, but just that all over tech level 1 is 0 in stock?

johnbartholomew:
> Meaning: dock to a tech level 1 station, and you see all equipment the game offers, but just that all over tech level 1 is 0 in stock?

Yes; it shows all items, with unavailable items listed as having 0 stock:

Tech Level 1:
screenshot-20180818-203900

Tech Level 8:
screenshot-20180818-203908

fluffyfreak:
@impaktor if wanted to dull/grey out the items that couldn't be bought but I just couldn't figure out the Lua for it. Might be one way of dealing with it.

impaktor:
@fluffyfreak well, my objection would be similar for that as well, as it would "show all the game" at once, and not leave any surprises for the player in future docking to rare space stations (military tech level), by discovering there's also an "X equipment".

Now, I realize I'm whining, and I'm not objecting to merging this, since I have no alternative "fix".

## Possible future solution 

But as this UI will be ditched soonish (I/we hope), it's worth trying to figure out other ways this could be handled, and assess if they're better.

Maybe the ship yard could remove the "high tech" equipment, and it would become rubble, because they're too incompetent? Or maybe you have that (in)competence yourself? So you can transfer equipment from installed slot to cargo hold? Then you've freed up the slot it took (if that was your goal), and you can sell it once at a high tech station. And possibly you could convert the equipment to rubble? (assuming you're desperate for the cargo hold capacity)

Lucas art-point and click games: use hammer on military hyper drive -> rubble? (semi-joke)

We've talked about being able to put equipment in the cargo hold so it's just another commodity.

What I think would be a bad solution (from a game mechanics perspective) is to be able to sell the high tech equipment on a low tech world for price above purchase price, as it would maybe be natural that it would be very valuable, but then that would mean everybody would mainly trade in high tech ship equipment instead of commodities, which is what was initially intended to be the main "items of trade" / commerce.

johnbartholomew:
Perhaps we can move the general discussion of game mechanics for sale (or not) of high tech equipment back to the parent issue #4276?

I don't really want to merge this PR as is; I agree with impaktor that it's not great to show all equipment types and levels on every station.

fluffyfreak:
Closing for future solution for the problem.

PR #4395, opened 2018-08-14
Two new custom systems

SpifforOmnivore:
Added two custom systems :
-61 Virginis
-1 Orionis
and edited the 02_local_stars.lua to reflect this change.


ecraven:
Could you comment on what makes these systems special, why you think they should be custom, and what they now look like?

fluffyfreak:
@ecraven there's been some discussion on the forum here: https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=578

SpifforOmnivore:
I'm adding my new system right now, 61 cygni.

Like my other systems, here's what I do:
- I think of some lore compatible with Pioneer's existing lore
- I mimic the original system : Almost all bodies are identical in my custom than in the original one.
- I do little tweaks (in this case, an icy uninhabited planet has turned into a little Venus with settlements).
- I give a name to every body. I usually look for public domain names (in dictionaries, maps etc.)

61 Cygni's description:
_"Terraforming gone wrong": these words describe well planet Eden, its runaway greenhouse gases, its scorching heat, its unbreathable atmosphere. But these words don’t begin to capture the spirit of its people. 
Back in 2317, as they disembarked from colony ship « Enlightenment » after nearly 30 years of a slower-than-light journey, the first settlers knew there would be no such thing as outside help. They had to turn their icy planet into an habitable one, and were confident it would become the lush paradise Earth ceased to be.
As things went wrong, Eden’s citizens witnessed their world turning slowly into the scorching hell it is today. In order to survive their new environment, they radically embraced genetic engineering. Besides creating plant species, they also massively engineered the human body, thus breaking an almost universal taboo.
When hyperdrive technology reached 61 Cygni in the 27th century, Eden’s people stopped being desperate for survival, and the system resumed an ordinary development. However, three centuries of isolation have shaped society until today : mankind is quite different here than anywhere else in the galaxy. And only here does the Solar Federation tolerate an exception to its ban on human genetic engineering.
Nowadays, the system is a hotspot for gene-fashionistas (green skin and horns are all the rage right now), and hosts many private or government-run research facilities. Tourists are commonplace too, enjoying the sights of this world, so exotic yet so close._

WKFO:
While we're doing this, why are we not writing Sol a proper long description?

impaktor:
> While we're doing this, why are we not writing Sol a proper long description?

Because [Sapiens](https://www.goodreads.com/book/show/23692271-sapiens) has already been written? Doing Sol justice seems like a tough task, but I assume the focus should be on how humanity spread from Earth to other planets and systems, generation ships, etc.

In general I prefer not too long descriptions, considering the atention span people have today. A paragraph or two. Basically not longer than the two systems here suggested. But I'm not the boss, and if there's a lot to say for Sol, then I think people will read it. (we like to read about ourselves the most, right?)

WKFO:
@impaktor Long description as in "long_desc:". My bad.

impaktor:
I assume these changes are save-bumping, since systems change.

PR #4396, opened 2018-08-15
Update COMPILING.txt

impaktor:
Got somoe complaints in IRC from n00b that lost hours to figure this out. 

Something similar could be added to GNU/Linux and OSX build perhaps.


fluffyfreak:
This is caused by MSVC loading the configuration and picking the first one it finds ALPHABETICALLY... amazing... and approved

Issue #171, opened 2011-06-03
La Musica


s20dan:
A few friends and I have been working on some music for Pioneer, some of you guys probably know.

 But its getting to a stage where we have quite a few tunes now and really I think its time to decide how we will deal with the music.

 I'll give a quick rundown of what I have in mind, hopefully you can argue against it for some other points of view or ideas on how to proceed.
 Im kind of thinking that we should play music based on situation and location, so that means music for combat, docking ect, and different music for systems, perhaps faction specific? Providied we have enough music, we could even have selection influenced by seed somewhat, so that specific systems would have the same music like X3, it could give a bit of character to systems.

 Unexplored space: Eerie ambience
 Core worlds:    Classical scores, such as Beethoven
 Fringe worlds:   Theres one song so far that could fit in here, its a mix of classical and electronic ambience with some structure to it.

 There's the old frontier remixes too, at least 2 of which are legally acceptable (they are sufficiently different) the rest would require some more modification.
 Anyway I'd love to hear anyone's thoughts on how you think we should proceed with this.

PR #4398, opened 2018-08-19
Unexplored systems soundtrack

WKFO:
It is basically about giving character to systems. Also something once @s20dan thought of, but was never implemented, if I understand correctly. I just wanna add this link again for some reason:
https://github.com/pioneerspacesim/pioneer/issues/171

Modified the MusicPlayer so that it will play the unexplored systems soundtrack whenever the player is wandering off more than 500ly away from Sol (this is where unexplored systems start appearing and human settlements start getting really sparse).

I had some talk with @impaktor and he thinks it would be better if we checked the current system's exploration status instead of hard coding 500ly. I'm fine either way, I don't disagree with him. Little inconveniece though: once the player enters an unexplored system it ticks the "explored" flag, so I don't think we can use that method. So if anyone has a bright idea, tell us!

Enough about the technical part, the comparison between normal space soundtrack and unexplored systems soundtrack is basically the following:

Human-populated space: Relatively brighter electronic tracks with beats, electronically reproduced classical scores. Typical Pioneer soundtrack people used to listen.

Unexplored space: Slower paced, more minimalist (compared to the already-kinda-minimalist normal space soundtrack), calmer electronic tracks, with some tracks having dark ambience elements.

Also moved one of @s20dan 's tracks to the unexplored space soundtrack. It was literally named "unexplored_space_1" and was quite fitting too.


ElonSatoshi:
Suppose newly explored systems would play the unexplored-system music until a certain amount of time after the time it was explored? I think that would make sense, almost as if you were getting used to the newly explored system.

impaktor:
> Suppose newly explored systems would play the unexplored-system music until a certain amount of time after the time it was explored? I think that would make sense, almost as if you were getting used to the newly explored system.

Not sure I agree, as that would mean you hear the track even less. Also, it's not like the system will be teeming with life after discovery, but rather, you'll have it to yourself.

WKFO:
I suggest I increase the unexplored space distance to somewhere near 650-700ly and we won't bother finding a solution to check for exploration status of systems (read about the problem on my PR description), as anthing beyond that point won't be inhabited. (except for custom systems like Sag A*, which should still play the unexplored systems soundtrack.)

About the above comments:
Hearing the tracks less is not a problem itself, but changing the music's mood a lot may be. So different points of view, but I think I agree with impaktor. (Additionally, we would need to get real-life time instead of in-game time for that.)

WKFO:
I found (and fixed) a bug where MusicPlayer crashed the game when it tried to change the music during a hyperspace jump. 

Now, whenever the music needs changing while in hyperspace, it will go to the unexplored systems soundtrack and if you arrive in a core (populated) system it will change to standart space soundtrack, but if you arrive to an unexplored system it will play unexplored systems soundtrack.

An alternative and perhaps better solution could be getting the hyperspace details, but IMO it would add way too much complexity to such a script like the music player.

Another likely question to be asked: Why would it not default to the standart space soundtrack?
Answer: Because hyperspace should make you feel the vast emptiness between star systems. Additionally, it is not a very probable event the music needs changing during the hyperspace period, so it will be more like a small easter egg-y feature for music. (For reference, it took me this long to come across this event since I made the PR)

impaktor:
> Now, whenever the music needs changing while in hyperspace, it will go to the unexplored systems soundtrack and if you arrive in a core (populated) system it will change to standart space soundtrack, but if you arrive to an unexplored system it will play unexplored systems soundtrack.

I imagine this would, for most users, mean that the track would change again when exiting, as most travels are to core systems, I think. 

Is this ready for merge?

WKFO:
Yes, and yes.

PR #4401, opened 2018-08-22
Add hyperdrive breakdown sound effect

WKFO:
Adding hyperdrive breakdown sound effects. Only plays when the hyperdrive actually breaks down, not when it was repaired on the edge of breaking down.

Hopefully no player gets to find out what triggers these sound effects.


WKFO:
For some reason the Travis CI build seems to be stuck on yellow.

data/modules/BreakdownServicing/BreakdownServicing.lua
```diff
@@ -282,6 +284,11 @@ local onEnterSystem = function (ship)
 		else
 			-- Destroy the engine
 			local engine = ship:GetEquip('engine',1)
+			if engine.fuel == Equipment.cargo.military_fuel then
```
impaktor: Actually, in Equipment.lua there are structures:
```lua
local HYPERDRIVE_SOUNDS_NORMAL = {
	warmup = "Hyperdrive_Charge",
	abort = "Hyperdrive_Abort",
	jump = "Hyperdrive_Jump",
}

local HYPERDRIVE_SOUNDS_MILITARY = {
	warmup = "Hyperdrive_Charge_Military",
	abort = "Hyperdrive_Abort_Military",
	jump = "Hyperdrive_Jump_Military",
}
```

where one could possibly add a "critical failure" / "breakdown" noise. But it looks like there would need to be some C++ code for this.

Would need to just add some lines to LuaShip.cpp, and Ship.cpp, I assume. Maybe me or @fluffyfreak could help you if you don't feel like digging into c++ side.

I'm not sure what is best. (ping @laarmen ?)

WKFO: I was not willing to mess around with code a lot so I just went for the least-lines-of-code solution. Whatever you suggest is fine, but I could need help.

PR #4403, opened 2018-08-24
Spice up the error message with a dose of retro-sillyness

impaktor:
I can think of no better to review this than @fluffyfreak :grinning: 

guru3-00c06560


ElonSatoshi:
lol go for it

fluffyfreak:
I can't wait until you embed a fully programmable ships computer into the game and in reality it's actually just a fork of [UAE](https://en.wikipedia.org/wiki/UAE_(emulator)) running a 68k OS

PR #4405, opened 2018-08-25
Move JsonUtils.h/cpp from contrib/json/ to src/

johnbartholomew:
These files aren't part of the jsoncpp library and are custom for Pioneer. They also directly include various other Pioneer source files. It's awkward having dependencies flowing in both directions between contrib/ and src/. Unfortunately this isn't the only place we've done it.

I have tried to update the VC++ projects as necessary, but I haven't tested them. I have not updated the XCode project.

### DIFF
--- a/contrib/json/CMakeLists.txt
+++ b/contrib/json/CMakeLists.txt
@@ -2,7 +2,7 @@ project(json LANGUAGES CXX)
 
 set(BUILD_SHARED_LIBS OFF)
 
-add_library(${PROJECT_NAME} jsoncpp.cpp JsonUtils.cpp)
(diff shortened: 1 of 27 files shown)


impaktor:
I think XCode is dead? I haven't seen anyone so much as mention it in a year.

This is the first PR showing changes needed in both CMake, Make and VC++, so I'll keep this in the backpocket.

johnbartholomew:
Travis says no.
Looks like I haven't got the order or dependencies right between src/JsonUtils.cpp and some of the sub-src/ libraries.

johnbartholomew:
Apparently JsonUtils.cpp is required for the modelcompiler too (not surprising really). Let's see if Travis likes this version.

PR #4407, opened 2018-08-25
Delete older build system files (premake/codeblocks, XCode)

johnbartholomew:
I suspect no one uses these. CodeBlocks never had much usage by Pioneer developers as far as I know (it had *some* but not much). XCode is only useful on OS X and we have very few OS X developers, which makes it difficult to maintain the XCode project files.

CMake can generate both CodeBlocks and XCode project files for people who want them. The workflow is perhaps more complicated than if the project files are pre-existing in the repository, but running CMake is probably better for a new potential contributor than opening a project file and then finding that it doesn't work because it's out of date and missing some files.

Unfortunately I can't put proper CMake+XCode instructions in COMPILING.txt, because I don't have access to a Mac to try things out on.


impaktor:
Compiling.txt needs to be updated as well

johnbartholomew:
> Compiling.txt needs to be updated as well

I don't think COMPILING.txt mentions premake or codeblocks as options.

johnbartholomew:
Ping @Philbywhizz 
Heads up that the XCode project files are going away. Shout if it's a bad idea. I might end up merging this before you see it -- but we can always pull stuff back from git history if needed.

johnbartholomew:
Also ping @salborrelli 
I'm not sure who has been doing the OS X release builds most recently...

Philbywhizz:
The Xcode related files are old legacy. I stopped using them as it was far too hard to maintain (a new project had to be re-merged every time a new source file was added/removed). You will still need them at some point in the future if you want to create an .app bundle.

I've been doing the recent OSX builds using a different method (autotools). Looks like I might have to update all my scripts if you are going down the Make path.

I'm still happy to generate the OSX builds when I can (because I still like to play pioneer now and then). Happy for someone else steps up and takes over as well. ;)

PR #4414, opened 2018-08-26
Fix bundled vs. external GLEW build configuration options

### LINKED ISSUE #4409: Building with system GLEW library may still use bundled header files
Copying from my comment on #4391 (after I had merged it):

> Unfortunately I just found a problem with this. For most of the build, the contrib/ directory is in the include path. We have source files that include "glew/glew.h", which works for the local glew because it finds it (in contrib/glew/glew.h). However, the system header is /usr/include/GL/glew.h. That obviously won't be found at "glew/glew.h"; it would be found at <GL/glew.h>.
>
> So I think with this patch and when configured to use system glew this will actually link against the system library but use the header from the contrib/ library. That will cause a version skew problem except for people who are "lucky" enough that their installed system glew version exactly matches the one we've bundled.
>
> We'll need to move the headers around in contrib/ so that the include structure is more uniform and then conditionally add contrib/glew (or whatever) to the include path when we're using the bundled copy. Or some other fix.

johnbartholomew:
Fixes #4409.

The normal system include path (on Linux -- assuming my system is representative of 'normal') for GLEW is /usr/include/GL/glew.h. Previously the Pioneer code has included glew/glew.h, which is incompatible with using the system GLEW library since it's explicitly looking in a glew/ directory. Instead, we need to include either "GL/glew.h" or just "glew.h" and set up the include path in the build commands as needed.

It is a pain to actually test that this is right in all configurations, but I have done my best to validate:
- CMake with bundled GLEW
- CMake with system GLEW
- autotools with bundled GLEW
- autotools with system GLEW

I've updated the visual studio project files but I may have made a mistake (I can't test them).

Unfortunately if there's a mistake here it may not be visible but could produce subtle problems. In particular, if the header used doesn't match the library that's actually linked then it will be totally fine unless we do something like call a function that has a difference in declaration between the two places. In most cases the declarations will match anyway so we'll never notice (and there would be no problem in that case).


johnbartholomew:
Travis say yes --> merging.
Of course as soon as I merged I realised I probably broke the imgui build in the visual studio projects.

impaktor:
> Of course as soon as I merged I realised I probably broke the imgui build in the visual studio projects.

Maybe that will prove a good incentive to lean on CMake also for Windows?

PR #4417, opened 2018-08-27
Fix dual fire weapon origin problems

### LINKED ISSUE #4397: Dual weapons don't fire from correct positions
When equipped with dual pulse weapons, the shots fired eminate from unexpected locations on the ship or outside of the ship.

This can easily be seen here where one shot seems to be coming from outside of the ship entirely.
badpewpew

In truth they're BOTH incorrect because they are centred around, and then offset from, the `gun_mount_0` tag within the model heirarchy which is beneath the right wing.

In this particular situation the `gun_mount_0` tag's position should/could be mirrored around the ships local position in model space and no offset applied. Then each shot would behave as though it was coming from beneath each wing.

This might not apply to all models, and some models have the `gun_mount_0` along the centre-line of the ship therefore requiring the existing fixed offset.

fluffyfreak:
Replace the gunmount system with multiple gun mount locations for front/rear guns.

This could be extended later to allow for multi-barrel ships/systems.

Intended to fix #4397 

Problems:
- legacy ships/mods do not support dual-fire weapons

### DIFF
--- a/src/FixedGuns.cpp
+++ b/src/FixedGuns.cpp
@@ -4,6 +4,7 @@
 #include "FixedGuns.h"
 #include "GameSaveError.h"
 #include "Beam.h"
+#include "StringF.h"
 
 FixedGuns::FixedGuns()
 {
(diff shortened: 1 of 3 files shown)


fluffyfreak:
NB: There is no new ship data in this PR, so if anyone wants to test it they will need to get modifed ships from @nozmajner or edit the existing `tag_gunmount_X` locations to use the new format `tag_gunmount_X_multi_Y`

nozmajner:
[Here's one Sinonatrix example.](https://www.dropbox.com/s/0vi5iujcxtafj56/sinonatrix.dae?dl=0)

fluffyfreak:
Rebased to latest before I add some code to handle old models etc

nozmajner:
Will do the tag update for the ships in the weekend or so. (But I have to figure out, why the animations aren't exporting from Blender first)

fluffyfreak:
Ok this is updated so that older ships will load with their old gunmount slots. 
However these will not work for dual fire weapons which will fire a single shot.

fluffyfreak:
@nozmajner these will still need the new `tag_gunmount_X_multi_Y` doing for all the ships but there's a half-arsed fallback so single fire weapons will still work for now.

PR #4419, opened 2018-08-27
Make visual studio build again after recent changes

fluffyfreak:
Make visual studio build again after recent changes


johnbartholomew:
Sorry I broke it! It's difficult to avoid mistakes when editing the project files by hand :-/

fluffyfreak:
That is one of the advantages of generating the files! This is the odd bit of maintenance i don't mind though. This way of building for windows is just an odd solution.

PR #4424, opened 2018-08-28
CMake: Make GLEW as external target when using the system one

### LINKED ISSUE #4409: Building with system GLEW library may still use bundled header files

pcercuei:
This avoids an error when USE_SYSTEM_LIBGLEW is OFF then ON from the
GUI of CMake.

Fixes #4409.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,7 @@ endif (WIN32)
 
 option(USE_SYSTEM_LIBGLEW "Use the system's libglew" OFF)
 if (USE_SYSTEM_LIBGLEW)
+	add_library(GLEW::GLEW INTERFACE IMPORTED)
 	find_package(GLEW REQUIRED)
 endif (USE_SYSTEM_LIBGLEW)
 


impaktor:
@fluffyfreak speaking of CMake, this was on hackernews just now:
https://cliutils.gitlab.io/modern-cmake/

PR #4429, opened 2018-09-01
Added a clang-format harness.

Web-eWorks:
Per my conversation on IRC, I have implemented a harness for clang-format, mostly adapted from Godot's implementation.

The general gist of it is that, when properly hooked up in `travis.yml`, it will check all changed files in a PR to ensure that they adhere to the code style of the project, generating an error in the CI check if they are wrongly formatted. This will happen alongside the existing code compilation check and an error here will not affect the compilation step.

As the Pioneer code base is wildly inconsistent in and of itself, I have attempted to set sane and compliant defaults for clang-format. However, some limitations of the tool are insurmountable via mere configs - for example, the alignment of pointers and references are both controlled by the same clang-format switch. I would appreciate a review of the clang-format style from the core maintainers.

You can test the current rules by changing a `.h/.cpp` file and running `./scripts/clang-format.sh`. This is designed to run on a *nix / Mac system (e.g. Travis), so it is unfortunately unavailable to run locally on a non Unix-capable Windows development environment.

Running clang-format on the existing codebase is well beyond the scope of this PR, and will need to be carried out by the core team.

TODO:
- Get core maintainers to sign off on the `.clang-format` style.
- [x] Hook up to `travis.yml`.
- [x] Add a pre-commit hook to run clang-format automatically, giving the user an option to automatically apply the generated diffs.

Comments / critiques are welcome and requested.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,116 @@
+# Commented out parameters are those with the same value as base LLVM style
+# last sync: Clang 6.0.1
+---
+Language:           Cpp
+BasedOnStyle:       LLVM
+AccessModifierOffset: -4
+AlignAfterOpenBracket: DontAlign
+#AlignConsecutiveAssignments: false
+#AlignConsecutiveDeclarations: false
+#AlignEscapedNewlines: Right
+AlignOperands:   false
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+#AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: true
+#AllowShortFunctionsOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: true
+#AllowShortLoopsOnASingleLine: false
+#AlwaysBreakAfterDefinitionReturnType: None
+#AlwaysBreakAfterReturnType: None
+#AlwaysBreakBeforeMultilineStrings: false
+#AlwaysBreakTemplateDeclarations: false
+#BinPackArguments: true
+#BinPackParameters: true
+BraceWrapping:
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  AfterExternBlock: false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: false
+#BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+#BreakBeforeInheritanceComma: false
+BreakBeforeTernaryOperators: false
+#BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: AfterColon
+#BreakAfterJavaFieldAnnotations: false
+#BreakStringLiterals: true
+ColumnLimit:     0
+#CommentPragmas:  '^ IWYU pragma:'
+#CompactNamespaces: false
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: false
+#DerivePointerAlignment: false
+#DisableFormat:   false
+#ExperimentalAutoDetectBinPacking: false
+#FixNamespaceComments: true
+#ForEachMacros:
+#  - foreach
+#  - Q_FOREACH
+#  - BOOST_FOREACH
+#IncludeBlocks:   Preserve
+IncludeCategories:
+ - Regex:           '^".*"'
+   Priority:        1
+ - Regex:           '^<.*\.h>'
+   Priority:        2
+ - Regex:           '^<.*'
+   Priority:        3
+#IncludeIsMainRegex: '(Test)?$'
+#IndentCaseLabels: false
+#IndentPPDirectives: None
+IndentWidth:        4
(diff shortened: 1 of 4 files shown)


Web-eWorks:
Added a pre-commit hook (that has to be manually installed) that runs clang-format and refuses to allow commits that are non-compliant to the style. It can be bypassed with `git commit --no-verify` if need be.

Web-eWorks:
Barring a review of the clang-format style, I believe this PR is feature complete.
CC @ecraven @fluffyfreak (I was informed you were the core C++ people).

ecraven:
I like this, however I'd prefer to keep the .clang-format file much more minimal. Removing all commented lines, and also relying more on the existing style presets.

Web-eWorks:
The commented lines were included as a future-proofing measure, so that if clang-format changes the default styles, a maintainer can quickly return to the correct coding style by un-commenting the appropriate line. I can remove them if there is a consensus on the matter, however.

fluffyfreak:
I've requested a review from a few others who are better able to test this at the moment. I'm on a Windows PC during the day which means that I can't easily see the outcome of the scripts etc.

Hopefully it'll get feedback fairly quickly.

impaktor:
> I have attempted to set sane and compliant defaults for clang-format. 

"Sane" as in "minimizes changes to current code base", I assume? Specifically, it would be interesting to see how common different styles for which line to put `{}` on are? E.g.

```c++
void foo()
{

}

if(1){

}

```

I think we have (in part of the code base) different styles for if-statements and functions/methods? @fluffyfreak you might want to check that these changes adhere to what ever we use.

I guess we should find some short cpp/h file and run the script and make sure it only does sane changes. For instance: `UseTab: Always` makes me slightly nervous, does this always indent with tabs? Because the entire main c++ code base only uses tab for the "main" indentation, and then spaces.


When this is merged, I'd favor running it on the full code base, as a giant commit, like we do with the header copyright New Years Eve update each year (when we remember it :smile:).

Anyway, we (the future I, when I have a working computer?) should test these settings on some files and see how much will be changed, before this PR is merged.

@Web-eWorks if feeling restless you could check that the [Editorconfig](https://editorconfig.org/) file [here](https://github.com/pioneerspacesim/pioneer/blob/master/.editorconfig) is in agreement with the clang definitions introduced in this PR.

Web-eWorks:
@impaktor 
You assume correctly.

>I think we have (in part of the code base) different styles for if-statements and functions/methods?

As far as I can tell, there's an even split over function braces on a new line vs on the same line. I prefer on a new line, as it allows indenting long parameter lists (there's a lot of this) without the list visually flowing into the function body. If-statements are universally same-line braces, which includes stacking the else keyword onto the ending brace.

>Because the entire main c++ code base only uses tab for the "main" indentation, and then spaces

NEVER MIX TABS AND SPACES. DOOM LIES THAT WAY.
All joking aside, the majority of the codebase I've touched has used only tabs, the editorconfig file mandates tabs, and so does the wiki. It's pretty clear that the style is tabs only, though you can insert spaces to line things up at your own peril.

The editorconfig is in agreement with these changes.

I'm not 100% happy with the formatting of class initializer lists, but I think it's miles better than before.

@fluffyfreak 
I have attached a sample diff below:

<details>

```
--- a/src/Aabb.h	2018-09-07 16:53:20.928818746 -0400
+++ b/src/Aabb.h	2018-09-07 16:53:30.032843455 -0400
@@ -9,22 +9,24 @@
 struct Aabb {
 	vector3d min, max;
 	double radius;
-	Aabb()
-		: min(DBL_MAX, DBL_MAX, DBL_MAX)
-		, max(-DBL_MAX, -DBL_MAX, -DBL_MAX)
-		, radius(0.1)
-	{ }
-	void Update(const vector3d &p) {
+	Aabb() :
+		min(DBL_MAX, DBL_MAX, DBL_MAX),
+		max(-DBL_MAX, -DBL_MAX, -DBL_MAX),
+		radius(0.1)
+	{}
+	void Update(const vector3d &p)
+	{
 		if (max.x < p.x) max.x = p.x;
 		if (max.y < p.y) max.y = p.y;
 		if (max.z < p.z) max.z = p.z;
 		if (min.x > p.x) min.x = p.x;
 		if (min.y > p.y) min.y = p.y;
 		if (min.z > p.z) min.z = p.z;
-		if (p.Dot(p) > radius*radius) radius = p.Length();
+		if (p.Dot(p) > radius * radius) radius = p.Length();
 	}
-	void Update(float x, float y, float z) {
-		if  (max.x < x) max.x = x;
+	void Update(float x, float y, float z)
+	{
+		if (max.x < x) max.x = x;
 		if (max.y < y) max.y = y;
 		if (max.z < z) max.z = z;
 		if (min.x > x) min.x = x;
@@ -32,12 +34,14 @@
 		if (min.z > z) min.z = z;
 	}
 	template <typename T>
-	bool IsIn (const vector3<T> &p) const {
+	bool IsIn(const vector3<T> &p) const
+	{
 		return ((p.x >= min.x) && (p.x <= max.x) &&
-		    (p.y >= min.y) && (p.y <= max.y) &&
-		    (p.z >= min.z) && (p.z <= max.z));
+			(p.y >= min.y) && (p.y <= max.y) &&
+			(p.z >= min.z) && (p.z <= max.z));
 	}
-	bool Intersects(const Aabb &o) const {
+	bool Intersects(const Aabb &o) const
+	{
 		return (min.x < o.max.x) && (max.x > o.min.x) &&
 			(min.y < o.max.y) && (max.y > o.min.y) &&
 			(min.z < o.max.z) && (max.z > o.min.z);


```

</details>

fluffyfreak:
That looks good to me, close to the style I keep intending to write 😁

laarmen:
If Fluffy is good, I'm good :)

ecraven:
Thanks!

Issue #3362, opened 2015-03-17
Joystick inversion, deadzones, axis bindings


fluffyfreak:
# The Problem:

For a long time we've had some [issues with Joysticks/joypads](https://github.com/pioneerspacesim/pioneer/search?q=joystick&state=open&type=Issues&utf8=%E2%9C%93) which make playing Pioneer less than fun using these times of controls. It's a common post on the [SSC forum](http://spacesimcentral.com/ssc/topic/4925-help-with-flight-system/) and the [developers forum](http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=318). 
# Task overview:

This feature request would evaluate the work done in the [Paragon 10.5 Alpha](https://github.com/MeteoricGames/pioneer/commit/c93e14c8cbb629b3aa2a88dfe92bb445510f5e98) to see if it's a good solution that fits Pioneer and then to reimplement it for Pioneer. 
We cannot directly backport this code, despite it being GPL licensed and based on Pioneer, due to the way that our code bases have diverged.
Therefore the deadzone, axis inversion, and controller mappings would need to be written using the different code present in Pioneer. This will require a mix of C++ for the game side enumeration, access, etc and Lua for the options/setting screen and binding of axes, buttons, and d-pads.
# Alternatives:

Despite referencing the Paragon solution to the problem it is not necessary to use their code or method of fixing the issue if whoever picks up this issue doesn't want to. Other solutions are valid it is the problem that needs solving. The Paragon solution is merely presented as one option.

fluffyfreak:
@nick-the-ninja, here's the issue since you mentioned interest in it :)

victort:
I just noticed I cannot bind joystick axes to thrusters, so I thought i'd :+1: this,

:D

PR #4432, opened 2018-09-03
Improve Joystick / Input Handling

### LINKED ISSUE #27: 3D thruster control with joystick
I think the 3D thruster control already available by keys on keypad should also be available from the joystick, in the following way:
The main thruster/break thruster control might be bound to custom keys, or the joy's thruster controls, but the thrusters controlling the thrusters in the plane perpendicular to the axis of the main thrusters, should be controllable by the "hat switch" of the joy. Possibly inverse config should be available for those who would like it that way...

Web-eWorks:
The goal of this PR (or series of PRs, as needed) is to fix the input handling. We're moving from a statically-defined key-binding system that requires hard-coding everything in KeyBindings.inc.h to a dynamic, extensible binding system that is much easier to use and maintain.

Fixes #27, #3362.

Done:
- [X] Move the input handling subsystem into its own module, decoupling it from the monolithic main class as much as possible.
- [x] Rework the input system to support inversion, deadzone, and sensitivity on a per-axis/binding level.
- [x] Allow defining and handling key / axis bindings without having to hard code them in C++ - this needs to die as soon as possible.
- [x] Update the Lua-side key-binding logic to use the new system.

I've split the input subsystem out so that I can more efficiently implement the rest of the features without adding to the horrible bloat of the main class.


impaktor:
I guess you've seen:
https://github.com/pioneerspacesim/pioneer/labels/Input

Web-eWorks:
I've gotten the new binding system functional on the C++ side, and I'm now going to handle the Lua implementation. Fixing joystick hats is an optional target at this point - I'm starting to run out of steam.

CC @fluffyfreak for code review while I'm still working in the C++ side of things.

I also had a question - why are the `Binding::FromString` methods using C-style string operations instead of C++? Any performance boost seems to be entirely lost by the fact that they're only called about once per session.

impaktor:
Maybe @kuroneko wants to be aware of this PR? If you have any thoughts on this, please share.

fluffyfreak:
@Web-eWorks two reasons, first is C-style is usually more recognised by people or the C++ is usually verbose and ugly. The second is that the whole code base is like that, usually referred too as C-with-classes rather than true C++.

There's a lot of legacy stuff in here that's been in place for 10+ years and is endlessly repurposed. So even if it appears to be new code it might well have been copied/moved from elsewhere in the code.

fluffyfreak:
Only had chance to skim read it now but it looks good so far!

Been needing moving out out Pi for ages so glad that you're doing all this.

fluffyfreak:
@impaktor I think it should. the CMakeLists just grabs all cpp files in the src folders, then removes a few specific one. This should just be picked up automatically.

Web-eWorks:
I think I'm done here. Everything is hooked up and working, joystick sensitivity works fine, deadzone is confirmed working, inputs are working perfectly.

There are a few things which will need to be taken care of by another PR / maintainer:
- The input system is designed so that localizations of page/group/binding names are looked up by the binding ID, instead of including the localization strings in the binding info. This is not yet implemented on the Lua side, but I have left TODOs where this needs to be handled. It shouldn't be difficult, but I don't have enough experience with the Lang system to set it up right now.
- Axis events are currently not triggered, nor are Action events from anything other than a key binding. This is an artifact of the old implementation and I'd be more than happy to address it in another PR, but as it's currently not used by anything, I'm leaving it alone.
- The UI needs some better styling to distinguish the last line of an Axis binding from an Action binding; just a simple margin underneath should work.

Since this PR is sitting at +1300 insertions and it significantly affects a core system, I added myself to the AUTHORS.txt. I also added CMake compilation instructions to COMPILING.txt, because I didn't know that the codebase was CMake compatible until you said something.

Web-eWorks:
I forgot to mention it, but I left the old binding system in, as the old UI makes use of it in a few places. Once that is fully converted to the new UI, the old binding system can be deleted.

fluffyfreak:
Only thing i would say is to try and keep each PR to a distinct purpose.
So this one is about input, there shouldn't be changes to document CMake or fixup other things since we'dd add those separately to keep a PR easier to review.

Not a problem this time, just making sure that you know for future 😄

Web-eWorks:
Moved the CMake stuff to a new PR.

PR #4317, opened 2018-03-21
WIP: Update imgui

ecraven:
Work in progress, to pull in current imgui.
Problems:
- crashes sometimes
- text is not rendered
- some windows show scroll bars (might be related to text not being rendered)

PR #4437, opened 2018-09-06
Update bundled IMGUI to v1.64

Web-eWorks:
This PR supersedes #4317, as it actually gets IMGUI rendering again.

Due to changes in how WindowBorders work, the "ShowBorders" flag has been removed, and will need to be replaced with code that sets the StyleVar "WindowBorderSize". In addition, the styling has changed enough that some windows render scrollbars where they should not - the main menu progress bar, for one.

The `ImGui::IO.KeysDown` vector is using SDL_Scancode instead of SDL_Keycode - this was changed upstream. Use `SDL_GetKeycodeFromScancode` or `SDL_GetScancodeFromKeycode` if you need to swap between the two.

The radial menu is currently non-optimal - when it opens, the cursor snaps to the middle of the screen (usually causing an item to be hovered), and it's position is used to calculate the radial rather than the relative movement.

Otherwise, I believe everything works currently, though I would appreciate it if numerous people could test this PR.


ecraven:
Did you try languages with many glyphs, like chinese? The glyph tracking was never a part of IMGUI, always a local patch, and the reason why we added it was precisely because we *need* it. Without that, I don't think we should merge, otherwise we'll run into problems with machines that don't support the huge textures that will be generated.

Web-eWorks:
I have not tested this in any other language than English. I would appreciate it if you could test this in as many languages as you are able and let me know what needs fixing.

Regarding the glyph issues, I'd appreciate it if you could take a look at https://github.com/ocornut/imgui/releases/tag/v1.61, https://github.com/ocornut/imgui/issues/1671, https://github.com/ocornut/imgui/issues/1859 and let me know if and/or how any of those impact Pioneer's custom font code.

I do want to get the character range code working again, but it was keeping IMGUI from rendering _any_ characters, so it needed to be struck and rewritten.

ecraven:
@Web-eWorks: none of these IMGUI changes is relevant for us. We really do need the original patch. I'll try to port it to your current branch.

ecraven:
1) CMakeLists.txt needs to be updated, it does not compile any longer.
2) I haven't looked closely, but running `./pioneer -g Lang=zh` or `Lang=ru` do not work (and I guess all other languages that use a different font don't work either).
I'll look into fixing this later, but it seems you changed enough to make it not work at all any longer :-/

Web-eWorks:
In order:
1. Fixed.
2. Fixed.
3. Fixed.

Any other issues you've found?

Web-eWorks:
Rebased against latest master and squashed.

Web-eWorks:
I've fixed the issues with window borders and erroneous scrollbars, for the most part. We still have some issues (the initial progress bar renders with a window background) but as they're not thing's I've broken with these commits, I'm leaving them alone.
I've also switched the default theme to Dark instead of Classic - Classic is no longer actively developed or recommended, and I personally think it fits Pioneer's existing color scheme slightly better.

fluffyfreak:
Just pushed a fix for the vs2017 so that I could test.

It loads but I cannot get beyond the main menu screen due to a lua callstack into C++ crash. Just double checking to see if that's on master branch as well now.

ecraven:
@Web-eWorks can you please split this into one commit with upstream imgui, then one or more commits with our changes? The way it is now it's impossible to see which changes are local and which are due to upstream's update.
Thanks!

Web-eWorks:
@ecraven 
Done.

impaktor:
Version 1.65 released 4 days ago :laughing: 

I suggest do nothing, so this can be reviewed tomorrow as planned, and merged.

Web-eWorks:
1.65 is a code-shuffle release, basically. I can integrate it later with (very probably) no effort.

ecraven:
I've looked at the changes in `src` and `data`, they look ok. I haven't fully looked through the changes in `contrib/imgui`, but things seem to work fine for me. Could anyone test this on windows and mac?

fluffyfreak:
I might need to do some fixup for VS201X but don't worry about that and merge away if it's all good

ecraven:
Thank you!

PR #4440, opened 2018-09-07
More Input Improvements

Web-eWorks:
I have updated WorldView.cpp to use the new input system, marking (almost) the last of the non-oldUI systems to be switched over.

I have updated the way joystick hats are handled, allowing diagonals to be properly registered as two cardinal directions at the same time. This comes with the caveat that we can no longer bind diagonals as their own buttons, but due to the way diagonals are generated, this is an improvement to the input semantics (as you would have to not bind anything to the cardinal directions instead).

I also changed how keybindings are displayed so that they can be collapsed or expanded at will. Obviously, these new strings need to be properly added to the translation and localized.

There's some funny stuff going on with keybindings getting through even when having the Controls window open - it might be from the imgui switch, or it might be a method I forgot to upgrade. I'll address it in another PR.


impaktor:
I assume @fluffyfreak is best apt to OK this, sorry, I don't mean to shuffle work on others. 

...he said while shuffeling work onto others.

PR #4441, opened 2018-09-10
[WIP] Move the Info View over to the new UI

Web-eWorks:
Started work on moving the Info screen over. The first two panels are done, with the other three left to go. The model viewer and face display widgets need to be ported as well - I'm probably not going to be able to do that myself at the moment.


impaktor:
Oh, cool!  Anyone with a screenshot?

nozmajner:
Ping me if you need any mockups!

Web-eWorks:
@nozmajner 
I'm just doing a 1:1 translation of the old InfoView right now, but the UI theming could use a definite facelift. Defining a set of colors (and specifically noting them in the theme file) would go a long way.

If you want to mock something up, the Flight UI could use some sections for orbital info - apoapsis, periapsis, time to AN/DN/AP/PE/etc, planned maneuver info, etc. I'm planning to get to that at some point on my long list of TODO.

nozmajner:
Eventually I want to rethink the InfoView screens anyway, because they aren't that optimal in their current state, no matter the UI system.
I already did some mocks for the orbit display a couple of weeks or so ago, when we were talking about it with @ecraven 
Like:
Collpased:
worldview04
Open, and the body list is open as well, to show the space usage.
worldview03
I played a little bit of Orbiter, and I got most of my inspiration from the orbit display of it, but I aimed fo a much cleaner one.
I imagine it to be resizable from the top (if imgui allows), so you could just have the graphics display or the data at hand. Ap and Pe and time to them are always displayed when applicable along with your coordinates at the bottom.

Web-eWorks:
Minor nitpick - I believe that the up and down arrows are/were/should be used to show ascending node and descending node. We should probably use something better tailored for the AP/PE nodes. Additionally, I'm not entirely sure what you mean by the chevron over the bar and the down arrow below it - AN/DN have the same angle (inclination), and argument of periapsis or longitude of ascending node need much better iconography to distinguish them.

nozmajner:
I think the apse icons work as they should: down triangle is Pe, up is Ap. Compact and directional. 
I tried to avoidcluttering the view with other icons, the AN/DN are indicated by the fadeing on the orbit line. Although that alone might fail to indicate which is which. A litttle down and up chevron could alleviate that.
 Although if the current position icon is rotated to the direction of travel, that could make it more clear. (like on the image, you are heading towards AN)
Also, tha chevron Over the line icon among the data is just a placeholder I put in ther for the Longtitude of AN. The down triangle is the Argument of Pe, paralleling the Pe icon (also a placeholder

PR #4427, opened 2018-08-29
Generate an installer for Windows in the cloud with AppVeyor CI

pcercuei:
Hi,

This PR adds a few scripts that allows the AppVeyor CI to create Windows builds of Pioneer on each tag, PR or commit. This is really useful to make sure that Windows builds don't get broken in the development process. Besides, creating releases is now very easy, as one can just redistribute the installer generated by AppVeyor that corresponds to the tagged commit.

For now, the generated installer only contains a 64-bit version of the game, but it shouldn't be too hard to alter the script to pack both 32-bit and 64-bit versions in one installer (I don't have time anymore to do it myself) or alter the scripts to do additional things (like pre-compile the models).

To use the AppVeyor CI, visit https://www.appveyor.com/ and sign up with your GitHub account.
In the projects list, click "New project", select "Github" and the Pioneer repository.

No extra configuration should be required, everything is contained within the ```appveyor.yml``` script. The only thing to verify in the project's configuration is that the option "Skip branches without appveyor.yml" is checked.

PR #4442, opened 2018-09-11
Build the modelcompiler in CMake

fluffyfreak:
Build the modelcompiler as well as Pioneer executable using shared object library files to reduce build time.

I think the there needs to be some executable stripping options added to the build but CMake is woefully opaque abot how you pass compiler options like that through.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,7 @@ foreach (each IN LISTS SRC_FOLDERS)
 endforeach (each IN LISTS SRC_FOLDERS)
 
 list(REMOVE_ITEM CXX_FILES
+	src/main.cpp
 	src/modelcompiler.cpp
 	src/tests.cpp
 	src/textstress.cpp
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -171,9 +200,15 @@ set_target_properties(${PROJECT_NAME} PROPERTIES
 	CXX_EXTENSIONS ON
 )
 
-install(TARGETS ${PROJECT_NAME}
-	RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+set_target_properties(modelcompiler PROPERTIES
```
pcercuei: You can specify both ```${PROJECT_NAME}``` and ```modelcompiler``` in set_target_properties(), to avoid duplicating code.

CMakeLists.txt
```diff
@@ -142,7 +143,15 @@ if (NOT USE_SYSTEM_LIBLUA)
 	include_directories(contrib/lua)
 endif (NOT USE_SYSTEM_LIBLUA)
 
-add_executable(${PROJECT_NAME} WIN32 ${CXX_FILES})
+if (WIN32)
```
pcercuei: I assume this shouldn't be here, it belongs to #4427.

PR #4443, opened 2018-09-12
CMake: Support building model compiler

pcercuei:
The CMake script will now build both the Pioneer executable and the
model compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,7 @@ list(REMOVE_ITEM CXX_FILES
 	src/tests.cpp
 	src/textstress.cpp
 	src/uitest.cpp
+	src/main.cpp
 )
 
 if (WIN32)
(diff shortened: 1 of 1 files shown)


pcercuei:
Woops, just noticed #4442... Sorry about that xD

fluffyfreak:
That's ok, so what are the differences between these two approaches?

pcercuei:
From what I can see they do basically the same: create a static library that contains all of Pioneer's code, and link both the main executable and the model compiler to it. The main difference is that I link all the dependencies to my 'libpioneer' so that I don't have to do it again when building pioneer and the model compiler.

fluffyfreak:
I used the [OBJECT](https://gitlab.kitware.com/cmake/community/wikis/doc/tutorials/Object-Library) keyword approach so it's not a true library, just creates the obj files and then reuses them for linking.

fluffyfreak:
Rather than duplicate effort I've closed #4442 so we can focus on just one PR

pcercuei:
Ah, you should have closed mine, I'm going to disappear again for a couple of weeks ;)

ecraven:
So can we merge this? Anything not working with it?

fluffyfreak:
Closing this as I picked #4447 to use instead. I can't really tell the difference so...

PR #4444, opened 2018-09-12
Move Hyperspace exit locations into the heart of systems

fluffyfreak:
Does what it says on the tin and makes the hyperspace exit point be uncomfortably close to the systems hyperspace target star.

- [x] Should add a velocity for safe orbital speed
- [x] Orients the ship in direction of travel

I'm actually going to say that this is ready for testing now. It needs people to go about the galaxy jumping from place to place. Some stars are already pretty close in, others far out.

Ready for testing and feedback.

### DIFF
--- a/src/CameraController.cpp
+++ b/src/CameraController.cpp
@@ -8,6 +8,7 @@
 #include "Game.h"
 #include "GameSaveError.h"
 #include "JsonUtils.h"
+#include "MathUtil.h"
 
 CameraController::CameraController(RefCountedPtr<CameraContext> camera, const Ship *ship) :
 m_camera(camera),
@@ -488,15 +489,6 @@ void FlyByCameraController::Reset()
 	SetPosition(vector3d(0, 0, 0));
 }
 
-static matrix3x3d LookAt(const vector3d eye, const vector3d target, const vector3d up)
-{
-	const vector3d z = (eye - target).NormalizedSafe();
-	const vector3d x = (up.Cross(z)).NormalizedSafe();
-	const vector3d y = (z.Cross(x)).NormalizedSafe();
-
-	return matrix3x3d::FromVectors(x, y, z);
-}
-
 void FlyByCameraController::Update()
 {
 	const Ship *ship = GetShip();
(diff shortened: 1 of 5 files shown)


WKFO:
Why though? I'm not against it, just curious.

fluffyfreak:
It was discussed as an option to try on the forum and I quite liked it so it's here to try whilst discussion is ongoing.

impaktor:
I think this PR was triggered by a conversation on IRC by some player saying they often end up +15 AU from the target, and have very long travel times. To me it sounds like they targeted the wrong binary star.

Either way, this is an interesting de-Frontier-clonify step for Pioneer, and it might lead to similar traveling times as being dropped the typical 8 AU from a system. 

> Needs to be a formula or agreed safe distance from the star

It could be interesting game play if some ships/ship configurations had their exit point too close to the star to escape it. However, the player must understand this, or else it will feel like the game is giving them the middle finger, but I don't know how that would be done in a good way.

But I do like the potential haggle in the ship market:
_"With this fine engine you can escape the gravity well of the infamous exit point in Cygnus X-1" but it will cost ya laddie!"._

WKFO:
Gameplay-wise I liked it quite a lot. I don't really know if it will have a serious effect on mission deadlines in either way, but it seems to work quite nicely. Most of the time the exit velocity puts you more or less in a nice trajectory. OK, "nice" might be a little too good, but at least it is not immediately alarming.

Also you get to see stars.

fluffyfreak:
Also, and this just occurred to me as I flew away from a star, it means that instead of being spread across the surface of a 20+ AU wide sphere all of the ships entering a system are now much closer together near the stars themselves!

That could make some kinds of nefarious gameplay far more probable.

richardpl:
Can you still escape blackhole in system wiith such body?

WKFO:
As I jumped into Sagittarius A* (-3125, 0, 0) I had a incredibly eccentric suborbital orbit and I was heading outwards from the center.

I didn't wait long enough to see myself but this is what the orbit display told me.
If you jump near large stars it seems to put you into a suborbital orbit of some kind, with the ship heading towards apoapsis.

PR #4445, opened 2018-09-13
RFC: Run clang-format on src/**/*.{cpp,h}

ecraven:
Just ran `clang-format` on all cpp and h files under src. Please comment on whether there are any problems with this, otherwise we probably just merge something like this commit in one big swoop?
We might also want to add one or more of the hooks @Web-eWorks kindly provided.


Web-eWorks:
@ecraven 
At the moment, there's just the one hook - `scripts/pre-commit-hook.sh`. It's a client-side hook, so you'll need to provide instructions in COMPILING.txt or similar.

johnbartholomew:
+1 to setting up a pre-commit hook or whatever, because there's approximately zero chance that I'll remember to do this manually for every commit.

laarmen:
Won't change much if you don't manually install the precommit hook, though
😉

Web-eWorks:
Obligatory reminder that the hook is already written - you just need to copy it from `script/pre-commit-hook.sh` into `.git/hooks` and remove the `.sh` on the end. You can do this now, on current master.

PR #4450, opened 2018-09-15
Remove OpenGL 2.1 code from the project

fluffyfreak:
Removed Opengl 2.1 again (#3219)


impaktor:
I'm fine with this. (I suspect it was kept in for this long due to me).

PR #4454, opened 2018-09-20
Save game optimisations

fluffyfreak:
With the merging of #4207 saving has gotten a little bit slower. This reminded me of the fact that there's lots of save optimisations we left on-the-table after #3680 and #3684

So here's a couple of little changes that together reduce the size of (_already compressed_) save games by ~26%

~Also this absolutely breaks all existing saves and thus far I'm making no effort to prevent that.~
Contrary to my initial statement this should load most recent save games alright.

### DIFF
--- a/src/Beam.cpp
+++ b/src/Beam.cpp
@@ -153,7 +153,6 @@ void Beam::LoadFromJson(const Json::Value &jsonObj, Space *space)
 	if (!jsonObj.isMember("projectile")) throw SavedGameCorruptException();
 	Json::Value projectileObj = jsonObj["projectile"];
 
-	if (!projectileObj.isMember("dir")) throw SavedGameCorruptException();
 	if (!projectileObj.isMember("base_dam")) throw SavedGameCorruptException();
 	if (!projectileObj.isMember("length")) throw SavedGameCorruptException();
 	if (!projectileObj.isMember("mining")) throw SavedGameCorruptException();
--- a/src/DynamicBody.cpp
+++ b/src/DynamicBody.cpp
@@ -101,10 +101,6 @@ void DynamicBody::LoadFromJson(const Json::Value &jsonObj, Space *space)
 	if (!jsonObj.isMember("dynamic_body")) throw SavedGameCorruptException();
 	Json::Value dynamicBodyObj = jsonObj["dynamic_body"];
 
-	if (!dynamicBodyObj.isMember("force")) throw SavedGameCorruptException();
-	if (!dynamicBodyObj.isMember("torque")) throw SavedGameCorruptException();
-	if (!dynamicBodyObj.isMember("vel")) throw SavedGameCorruptException();
-	if (!dynamicBodyObj.isMember("ang_vel")) throw SavedGameCorruptException();
 	if (!dynamicBodyObj.isMember("mass")) throw SavedGameCorruptException();
 	if (!dynamicBodyObj.isMember("mass_radius")) throw SavedGameCorruptException();
 	if (!dynamicBodyObj.isMember("ang_inertia")) throw SavedGameCorruptException();
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -28,10 +28,8 @@ Frame::Frame(Frame *parent, const char *label, unsigned int flags)
 	Init(parent, label, flags);
 }
 
-void Frame::ToJson(Json::Value &jsonObj, Frame *f, Space *space)
+void Frame::ToJson(Json::Value &frameObj, Frame *f, Space *space)
 {
-	Json::Value frameObj(Json::objectValue); // Create JSON object to contain frame data.
-
 	frameObj["flags"] = f->m_flags;
 	frameObj["radius"] = DoubleToStr(f->m_radius);
 	frameObj["label"] = f->m_label;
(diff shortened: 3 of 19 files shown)


impaktor:
> Also this absolutely breaks all existing saves and thus far I'm making no effort to prevent that.

:+1: 

I say: walk all over that code like a dinosaur! Given the enormous changes/improvements since last release, I don't think players will mind us borking their old saves.

fluffyfreak:
I've got some more commits coming for but much of the remaining bulk of the data is down to two things: 
 - "matrix_transform"
 - "lua_class": "EquipType",

Each with their own associated followup data. I'm going to tackle the "matrix_transform" stuff next but the "Equip_type" stuff has me baffled.

In a single save file that I'm working from there's 36123 instances of this, and I have no idea why there's so much. A lot of it is in cargo, or serialising the slots of a ship that's associated with a hyperspace cloud.

Can anyone @laarmen @impaktor @johnbartholomew chime in with advice about reducing the amount of this output, or information about why it's needed please?

laarmen:
Normally there shouldn't be that many instances of EquipType, only one per
equipment type (duh). On the other hand you should have a ton of references
to these instances.

Of course, we cannot exclude me having been young and stupid when I first
wrote the current equipment system. I think I would do it very differently
nowadays.

Le sam. 22 sept. 2018 à 16:11, Andrew Copland <notifications@github.com> a
écrit :

fluffyfreak:
@laarmen ah references... well this is what a snippet looks like

```
 "body": {
	"clip_radius": "4629927121953751040",
	"dead": false,
	"index_for_frame": 39,
	"label": "NC-5764",
	"lua_ref_json": {
		"ref": 3049564573728,
		"table": ["autopilot_cap", {
				"number": "4607182418800017408"
			}, "hullPercent", {
				"number": "4636737291354636288"
			}, "label", "NC-5764", "shipName", "", "totalMass", {
				"number": "4659431211351932928"
			}, "fuelMassLeft", {
				"number": "4652007308841189376"
			}, "equipSet", {
				"lua_class": "EquipSet",
				"ref": 3049564576416,
				"table": ["slots", {
						"ref": 3049564576672,
						"table": ["trade_computer", {
								"ref": 3049564583072,
								"table": ["__limit", {
										"number": "4607182418800017408"
									}, "__occupied", {
										"number": "0"
									}
								]
							}, "engine", {
								"ref": 3049564576032,
								"table": [{
										"number": "4607182418800017408"
									}, {
										"lua_class": "HyperdriveType",
										"ref": 3049564563360
									}, "__limit", {
										"number": "4607182418800017408"
									}, "__occupied", {
										"number": "4607182418800017408"
									}
								]
							}, "target_scanner", {
								"ref": 3049564577312,
								"table": ["__limit", {
										"number": "4607182418800017408"
									}, "__occupied", {
										"number": "0"
									}
								]
							}, "cargo", {
								"ref": 3049564575008,
								"table": [{
										"number": "4607182418800017408"
									}, {
										"lua_class": "EquipType",
										"ref": 3049564506400
									}, {
										"number": "4611686018427387904"
									}, {
										"lua_class": "EquipType",
										"ref": 3049564506400
									}, {
										"number": "4613937818241073152"
									}, {
										"lua_class": "EquipType",
										"ref": 3049564506400
									}, {
										"number": "4616189618054758400"
									}, {
										"lua_class": "EquipType",
										"ref": 3049564506400
									}, {
										"number": "4617315517961601024"
									}, {
										"lua_class": "EquipType",
										"ref": 3049564506400
									}, {
```

If I can avoid some of that spam, or if the data is necessary then save it in a different format, it would massively reduce both the file size and the time taken to save/load.

johnbartholomew:
A few miscellaneous comments:

- There are some things in the JSON-serialized Lua data that are super ugly because of the limitations of JSON and Lua and text based formats. E.g., all numbers, even if *in practice* they're small integers (such as table indices), are serialized as though they're a double precision floating point value, because that's what Lua exposes (prior to 5.3). JSON also has no differentiation between integers and floats. And then because round-tripping floats through a normal decimal format is (a) slow, and (b) a bit finicky if you want it to be bit-exact (*), the 8-byte floats are reinterpreted as 8-byte integers and then written out as `{ "number": 123456 }` (e.g., `{ "number": 4607182418800017408 }` is actually just the integer 1). So that's longer than it needs to be, and more awkward.

- Lua doesn't distinguish between tables and arrays, and it's not that uncommon to have something that is *mostly* an array (i.e., small integer indices), but which has a few non-integer-key entries too, holding metadata. With hindsight, I should have structured a serialized Lua table as something like: `{ "__tablepart": {}, "__arraypart": [ ... ] }` so that Lua tables that are mostly arrays would actually be mostly stored as JSON arrays. Instead, I did a direct serialization so each table turns into a JSON object, and each small integer key is individually serialized (instead of being implicit in the fact that it's an array). And the small integer keys turn into big `{ "number": ... }` things, as explained above. Also, JSON doesn't support non-string keys in objects, which is why I serialized a Lua table as a JSON array where each slot in the Lua table becomes *two* elements of the JSON array (key, then value).

- The equipment data structure (see EquipSet.lua) doesn't store (item-type, count), it stores *each instance* of each item individually. See this loop in [EquipSet:__Add_NoCheck](https://github.com/pioneerspacesim/pioneer/blob/master/data/libs/EquipSet.lua#L218-L228). Have a big cargo hauler carrying 500x "Super Fun Widget" items? Congrats, we're going to save and load an array of 500x identical elements. These *should* all be references. So they look like `{ "number": "4611686018427387904" }, { "lua_class": "EquipType", "ref": 3049564506400 }`.

- I don't actually remember why I'm recording the `lua_class` value for every reference. There might be a reason... or it might be unnecessary. I'd have to look through the code a bit.

(*) I mean... it's fine as long as you have a float-to-decimal and decimal-to-float routine that you really trust and as long as you don't care about NaNs and infinities and so on... float<->decimal is still slow though)

johnbartholomew:
Btw, a lot of the serialization-side awkwardness could go away if we serialize [to CBOR](https://pioneerspacesim.net/forum/viewtopic.php?p=5350#p5350) instead of JSON. Or some other format, obviously, but when I was looking at this CBOR seemed like a reasonable option. (Unless we want something where we explicitly define a schema for the data, which IMO would be great, but would be a fairly big change...)

Anyway, I think the more important problems come from how the underlying data is structured, with a lot of redundancy and cross-references. Those structures need to change, or we need to stop directly serializing ~internal data structures and build a cleaner format to save. In Lua because it's garbage collected and tables are passed around by reference it's trivial to build up highly cross-linked data structures. This means we end up writing a lot of what is really *static* game data (e.g., the [static Equipment data](https://github.com/pioneerspacesim/pioneer/blob/master/data/libs/Equipment.lua#L535-L539)) directly into the save files, because the *dynamic* game state *directly* references those tables. It would be good to be able to split this static data vs dynamic game state and serialize just the dynamic game state.

Web-eWorks:
@johnbartholomew 
> It would be good to be able to split this static data vs dynamic game state and serialize just the dynamic game state.

I am (slowly) working on converting the Equipment system over to something that better follows that pattern. Right now I'm working on the best way to distinguish cargo from equipment, as an interim measure, with the goal of allowing us to store cargo as simple key-value pairs - e.g. `"hydrogen": 12`, instead of serializing 12 Lua equipment instances.

The main reasoning behind splitting cargo from equipment is that the two are not logically the same - cargo is something transported by the ship, while equipment is something (semi)-permanently affixed _that affects the ship in some significant way_. It also has the secondary benefit of namespacing cargo commodities - allowing us to just use a string key to refer to them in save files.

...I should probably open this as a new issue, shouldn't I?

laarmen:
And *à propos de* EquipSet storing a reference per individual item, it was because of the previous C++ APIs which allowed reordering and individual access to elements. I didn't have what it took to investigate why that was needed, I just mirrored it in Lua.

This actually makes sense for the equipment themselves, but I agree that cargo should be a separate category. That would allow for having actual objects as equipment, with individual values (performance boost, damage, etc...). Of course that's out of scope.

Issue #4415, opened 2018-08-26
The game crashes when hovering over corrupt save files


42triangles:
### Observed behaviour

The game crashes when hovering over save files that are either corrupt or can't be opened for read access.

### Expected behaviour

The game doesn't crash, but displays a warning instead.

### Steps to reproduce

Manually corrupt a save file or `chmod -r` it, and hover over a save file in the main or pause menu.

__My pioneer version (and OS):__
commit 86cfa671887e57cca70983f8b05a6ea1918c1f65 in this repository, NixOS (Linux)

johnbartholomew:
Fixed by #4416.

PR #4459, opened 2018-09-25
Update JSON to a better library, save games in CBOR

### LINKED ISSUE #4458: Upgrade JsonCpp to nlohmann/json
As I briefly discussed on IRC, I may have found the magic bullet here. [nlohmann/json](https://github.com/nlohmann/json) is an alternative Json implementation featuring a much more intuitive API and some really really tempting features over our current JsonCpp.

Before I get into pros and cons, I'd also like to remind anyone reading this that our JsonCpp hasn't been updated since 2010, which is a very long time.

**Pros:**

- It's much faster, more standards-conformant, uses much less memory, and has a smaller binary size. (https://github.com/miloyip/nativejson-benchmark)
- It has a more intuitive API; there's no need for the explicit asX methods, as it uses C++ conversion operators.
- It supports a number of features we don't really use right now but might in the future (JsonPointer, JsonPatch, etc.)
- It supports automatically transforming a JSON tree into CBOR. This means we don't *ever* need to write _yet another_ serialization path when we switch.
- And the best of all: it supports half/float/double internally, only converting them at the serialization step. This means that a round-trip from C++ to CBOR will be bit-correct for all our floating point numbers, allowing us to entirely drop the DoubleToStr calls.

**Cons:**

- There are 598 lines of JSON-related code in `src/` I'd have to upgrade. Not really a big deal, as I can get it done fairly quickly.
- It's a header-only library, so there's now 18,800 lines of C++ header to compile instead of 1800 for our current JsonCpp. This might result in slightly increased compilation times, but it shouldn't be too bad.

I'm... really not seeing a downside here, are you?

cc @fluffyfreak @johnbartholomew

Web-eWorks:
This PR fixes #4458, moving us to a new (and better) JSON implementation, removing the old DoubleToStr/StrToDouble path, and storing all data in CBOR. Also included is a CBOR->JSON dumper for debugging save file issues.

A note - this change _will crash_ with old save files, due to the lack of DoubleToStr / etc. I don't think this is going to be too big of an issue, as we're already breaking savefile compatibility with a couple other PRs.


Web-eWorks:
This PR has been rebased and updated. Now it also loads uncompressed JSON files to allow for save-debugging.

Web-eWorks:
I've moved some stuff that was previously duplicated all over the place into a new JsonUtils namespace, exposed Json data loading to Lua via the `Json` object, and added a simple system to take advantage of the library's Json Merge Patch support. Now, mods can _update_ data files by writing `<name>.<ext>.patch` files using the [RFC 7386](https://tools.ietf.org/html/rfc7386) syntax, in addition to overriding them.

I've got a little bit more cleanup left to do tomorrow, and then I believe this can be merged safely.

ecraven:
Seems to be OK, just one assert in JsonUtils.cpp:216 that fails to compile and doesn't make any sense to me. Might be a copy&paste error?

Web-eWorks:
It was a merge error. Should be fixed now.

Web-eWorks:
I think I finally got all of the references to the old json library out of the build system, and I even threw in a fix for the extremely verbose logspam thrown by Travis when building with thirdparty.

impaktor:
With the ` -Wno-old-style-cast` that will not only silence the thirdparty warnings but also our own pioneer/src-source tree, I assume?

Web-eWorks:
I'm working on a better method. Ideally, I'll use `#pragma GCC diagnostic` around our assimp includes.

Web-eWorks:
I don't know why Autotools keeps failing. It builds just fine under CMake, so I'm going to delegate fixing that to someone more knowledgeable with the build system. Otherwise, I believe this PR is ready to merge.

CMakeLists.txt
```diff
@@ -146,6 +150,19 @@ add_library(pioneerLib OBJECT ${CXX_FILES})
...
+	src/savegamedump.cpp
+	src/JsonUtils.cpp
+	src/FileSystem.cpp
+	src/utils.cpp
+	src/StringF.cpp
+	src/GZipFormat.cpp
+	src/DateTime.cpp
+	src/Lang.cpp
+	${FILESYSTEM_CXX_FILES}
+	# This one's in here because there's no such thing as separation in the codebase.
```
fluffyfreak: 👍

fluffyfreak:
Further testing has found issues.
- Loading errors out
- If you have old save games in the list them moving the mouse over them crashes the game

impaktor:
> If you have old save games in the list them moving the mouse over them crashes the game

Sounds like #4415

Web-eWorks:
@fluffyfreak 
>Loading errors out

If you could set your debugger to break on SavedGameCorruptException, and get me a backtrace (or just a save file), I'll take a look at it. I thought I caught all of those, though.

>If you have old save games in the list them moving the mouse over them crashes the game
That... shouldn't be happening. Do you have a backtrace? Or a save file? It should throw exceptions, probably, but not crash.

>loading/init screen seems corrupted, progress bar box is truncated to the left
I haven't seen this locally, and I'm not sure what you mean.

johnbartholomew:
> > If you have old save games in the list them moving the mouse over them crashes the game
> 
> That... shouldn't be happening. Do you have a backtrace? Or a save file? It should throw exceptions, probably, but not crash.

Fluffy could be seeing some other crash, but the one I see is a Json type_error exception in `LuaGame.cpp:l_game_savegame_stats`, when extracting the details from the rootNode Json object (i.e., after `Game::LoadGameToJson` completes). The Json exception types aren't caught in that path.

Web-eWorks:
@johnbartholomew Thanks for the info. That should be fixed now.
@fluffyfreak Does this take care of the issue, or are you experiencing any more crashes?

Web-eWorks:
I think I finally fixed the autotools build, no thanks to my IDE. Turns out that tabs are fine for indenting sources, but not if statements.

Web-eWorks:
I believe this PR is finally ready for merge. Invalid save files are cached along with valid save files, removing a giant slowdown, and invalid saves are properly caught.

fluffyfreak:
I get an assert in [Frame.cpp ln 72](https://github.com/pioneerspacesim/pioneer/pull/4459/files#diff-76f89b751d681404c52704a8e267f8f0R72) when trying to load a save that I've just saved.

Seems that it doesn't like the `child_frames` array being optional?

Web-eWorks:
@fluffyfreak That absolutely should not be happening. That line should generate a `null` json value, not error out.

I'm assuming that you're compiling this on a C++11 compliant MSVC, i.e. VS2015 or greater? I was not getting that error at all when I tested this, and I loaded a number of saves.

impaktor:
(personally, I wonder if all those "Fix..." commits could be squashed?)

fluffyfreak:
loadscreencorruption
That's what I see when loading with this branch, and I'm still unable to load a save game that I've saved with it.

I'm debugging it now to see if I can provide more info or figure out that problem.

fluffyfreak:
[newjson.zip](https://github.com/pioneerspacesim/pioneer/files/2453154/newjson.zip)
Save file for testing @Web-eWorks

Web-eWorks:
All loading errors have been resolved, and better error checking has been implemented. I believe that this is _finally_ ready for merge.

Web-eWorks:
@jaj22 are you _sure_ about that? I've tested the save Fluffy sent me, and it no longer crashes. If you've still got a crash even after the latest commit, please send me the save file so I can verify it.

fluffyfreak:
@jaj22 @Web-eWorks I still get that error too. RL has gotten in the way of doing anything on it today but I just tested with the latest code. It gets past the `child_frames` assert but still fails at unpickle time in the Lua

jaj22:
Ok, so I enabled the lua stack checking (define DEBUG or edit LuaUtils.h to use _DEBUG). Turns out that the stack check on line 578 of LuaSerializer.cpp is coming out as 0 rather than 1. 

Looking at the code, this seems to be as-written: If a table doesn't have a class type then the table is returned on the lua stack in unpickle_json, but if it does have a class type then it isn't. This seems utterly broken so I'm not sure that I'm reading it correctly, particularly given that it pre-dates this pull request.

Sanity check required plz.

johnbartholomew:
unpickle_json was only merged fairly recently; I would not be surprised if
it was broken from the start. Having said that, with a shallow look through
the code it looks like it's doing the same thing as the much older unpickle
function.

Build a table, and then if the object is a (Lua) class, pass the table to
the class Unserialize function (consuming the table) and get back some
object on the stack which gets returned from unpickle_json.

This looks wrong maybe?

    if (lua_isnil(l, -2)) {
        lua_pop(l, 2);

That would leave the stack with nothing if I'm counting right (which I may
not be). Though that check is also in the older unpickle code.

Web-eWorks:
While I still couldn't get it to crash, I have fixed the serialization code. @jaj22 @fluffyfreak if you could test the latest commit, it'd be much appreciated.

fluffyfreak:
moreluaissues

I am afraid not.

Also it hasn't been commented upon but I uploaded a screenshot of the loading screen which is totally corrupt on Windows. Any idea why?

I wouldn't have thought this PR would affect it at all.

fluffyfreak:
@jaj22 what compiler are you seeing the issues with? I'm using VS2017.

Web-eWorks:
@fluffyfreak this PR doesn't touch mainmenu.lua (AFAIK), which is responsible for that. I really don't know what's going on there.
I still cannot reproduce _any_ savegame errors; this seems to be a platform-specific issue restricted only to Windows.

fluffyfreak:
Perfect @jaj22 just tried it locally and that works.

Web-eWorks:
And now I feel kinda dumb. :man_facepalming: Thanks @jaj22 for tracking that down. If you would push your fix (and the savegamedump VS files) to this PR, I believe we can finally merge it.

laarmen:
Don't worry. We usually feel dumb when @jaj22 comes by and saves the day ;-)

jaj22:
@fluffyfreak That project file probably isn't 100% up to standard, and it's currently making iobj and ipdb files. I just rushed it into a "basically working" state.

fluffyfreak:
It saves, it loads.
If @impaktor doesn't do it first then I'll merge it tomorrow when my brain is working again.

impaktor:
@fluffyfreak could you please hold as we (me and @laarmen?) want ideally as many as possible of those "fixed xy"-commits to be squashed into the commit that introduced the "bug" xy.

laarmen:
Well, I'm not as... Let's say thorough ;-) as you are but that's true that
the history of this MR looks a bit... patched up :)

impaktor:
OK, so appart from the commit history, does anyone want to have time to review the code of this PR, or is it good to merge? I'm specifically thinking of @johnbartholomew and maybe @laarmen who have been working on the save-code before, or in the latter case: just can't stop thinking about Lua/C++ integration in Pioneer.

If no objections we will suffice with "it works on my end", and squash, rebase, and merge.

Web-eWorks:
Travis is red because it's not passing the clang-format check; I didn't want to clutter the PR even further with unrelated formatting changes.

At some point in the near future all of the source files are going to be upgraded for compliance with the format specification, and at that point I'll start caring about that check.

impaktor:
I'm happy with the rebase. Can be merged as is :+1:

PR #4463, opened 2018-10-02
Expose GetWindowSize() and GetContentRegion() to Lua.

Web-eWorks:
A very simple PR, but it allows creating windows that react to being resized. GetContentRegion() returns an ImVec2 with the available size in the current container (window, column, whatever) calculated from the current cursor pos. I've also exposed GetWindowPos() counterpart as well.


impaktor:
@ecraven if you just OK this, me or @fluffyfreak can merge it ASAP. Looks trivial.

PR #4465, opened 2018-10-04
Switch to CMake.

### LINKED ISSUE #4461: Drop Autotools, move to CMake
As you may have seen from my struggles in #4459, autotools is very finicky about certain things, difficult to maintain, and prone to occasional breakage. Having two build systems is decidedly sub-optimal in that it slows down anyone writing new code, as they have to know both build systems _and_ take the time to update and resolve any bugs arising in either system.

Given the conundrum, I propose that we drop autotools in favor of CMake. The reasoning for this is threefold:
- Autotools is old (not necessarily a con), complicated to configure, requires manually listing the sources and headers, very finicky about whitespace, and requires quite a bit of dev time to maintain. It's also *extremely* verbose, at 1721 total lines of configuration - counting *only* Makefile.am and configure.ac.
- CMake has much better C/C++ support, including automatic header discovery, better linking and CFLAGS support, and is much less verbose and easier to configure, coming in at only 262 lines.
- CMake *also* has much better build tool support, generating Ninja, Make, and even Visual Studio build configurations. It also has built-in support for generating `compile_commands.json`, allowing code completion in editors without built-in C/C++ support, like vim or Atom.

From my perspective, I'm not seeing any reason to keep autotools around. If any of you have objections or new information, please voice them soon, as I'm going to start work on implementing this by next week at latest.

As an aside, @fluffyfreak you do know you can use CMake to generate the VS solutions, right? I'm in no way a Windows dev, but it seems we might be able to remove the need to manually update the windows build files by automating it.

Web-eWorks:
This PR (partially) resolves #4461, by moving travis builds to CMake over autotools.

We're still using the thirdparty repo for travis - is that needed? It would shave a good 5-10 minutes off our already lengthy compile times if we were to drop it.

If I don't get significant pushback here, I'll be dropping the autotools build as part of this PR.


Web-eWorks:
So CMake doesn't like looking up packages in our thirdparty directory when they already exist on the system. As this neatly dovetails in with dropping the thirdparty for travis builds, I can live with that.

But then, CMake also doesn't like finding SDL2, despite specifically installing the package that (should) have the FindSDL2.cmake file in it.

At this point, I'm going to have to ask someone more knowledgeable to chime in here.

Web-eWorks:
So I got most of this to work, but travis apparently has some issues that no-one else is experiencing. I estimate a few more commits before this is ready to merge.

laarmen:
Just a heads up, there are conflicts plus a travis build fail.

I'm not against using CMake as a default build system, but I'd rather keep the autotools around for a bit longer, just to be sure we're not breaking anybody's setup (everybody doesn't follow the github project and simply pull regularly from master).

Thanks for the hard work on the issue!

Web-eWorks:
@laarmen thanks for the report. I plan on keeping autotools in a deprecated state for a while before we remove it outright.

I think I may have a solution to the latest round of errors, so this might actually be ready for a merge soon. I think.

Web-eWorks:
Huzzah, huzzah! It builds! Finally, everything works again.
@impaktor: pending a quick review (and maybe a squash), this PR is ready to merge. The deploy script needs updating to use the new binary location, but I'll take care of that when we add the API key.

impaktor:
I'm without pioneer-computer for the rest of this week, so I can't test it, but if @fluffyfreak and @ecraven, who are the main guys working on the C++ source the past year(s), are happy, then so am I (assuming squash :smile:).

impaktor:
So we already have cmake in master, right? So this moves travis to also use cmake as well, and fixes the aforementioned bug? Is there anything for us to test? If not, we could merge now-ish, or by the weekend at the latest.

Web-eWorks:
@impaktor This just makes Travis build with CMake instead of autotools, updates the CMakefile so that we can build with thirdparty, and restores compatibility with versions of CMake < 3.12. It is ready to merge now, if you wish.

PR #4470, opened 2018-10-12
Some linguistic fixes for world names

gunchleoc:
- Fix spelling error in Portuguese placename
- Use Scottish Gaelic spelling for Scottish Gaelic placenames
- Replace invented placenames with ones that work in scottish
- Carndow was a duplicate, so I replaced one of those with a new name

@SpifforOmnivore I am allowing edits from maintainers, in case you don't agree with my changes. Keep up the good work! :)

### DIFF
--- a/data/systems/custom/06_61virginis.lua
+++ b/data/systems/custom/06_61virginis.lua
@@ -26,7 +26,7 @@ local virginisa = CustomSystemBody:new('61 Virginis a', 'PLANET_GAS_GIANT')
 	:inclination(math.deg2rad(5.64))
 	:rotation_period(f(327,10))
 	:axial_tilt(fixed.deg2rad(f(0,10)))
-	
+
 local virginisb = CustomSystemBody:new('61 Virginis b', 'PLANET_GAS_GIANT')
 	:radius(f(10402,1000))
 	:mass(f(108204,1000))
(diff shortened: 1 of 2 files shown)


impaktor:
Cool. 

@nozmajner are we OK with non-ASCII characters? There is a question of how will letters be pronounced by the non-native speakers of that language? I.e. translate foreign name to how it would be spelled for an English speaking person vs. stay true to the original language?

nozmajner:
@impaktor They work, just checked. No missing characters. Although to my non-welsh eyes this one would look like if there's a characters missing: "Mac an t-Sealgai", but it does not.

SpifforOmnivore:
@guncheloc Thank you very much for your corrections :+1:

I'm really glad you changed the words of made up names so that they actually work in Scottish, and changed planet Uabhar to Pròis. 

I'm really happy that my clumsy attempts at using the Scottish language (which I don't speak at all) are corrected :)

However, I have a general question about the name of the actual towns and islands (not only for 1 Orionis, but for other custom systems as well).

I write those custom systems in English (which is not my native language, I'm French), because it's the base language of Pioneer.
As such, when I browse for real location names, I browse OpenStreetMap or google maps, looking for the English names.
The most prominent example is the Arabic-themed system Sirius. When looking at the map of Syria and Iraq with the Arabic alphabet, I had absolutely no idea how to say the locations. However, the English name sounds familiar enough.
That problem would be just as manifest if I wrote about a Chinese themed custom system. I have absolutely no idea how to say 武汉 (well, actually yes because I searched: that's Wuhan).

Now, Scottish is written with the latin alphabet, so each reader can imagine how to say the words, in genuine Scottish rather than in its anglicization ;)

SpifforOmnivore:
:+1:

(because Scottish is written in Latin alphabet)

gunchleoc:
> 
> 
> @impaktor They work, just checked. No missing characters. Although to my non-welsh eyes this one would look like if there's a characters missing: "Mac an t-Sealgai", but it does not.

That should be "Mac an t-Sealgair" with an r at the end, so if that's missing and not just a typo in your post, it is indeed missing.

Does the Pioneer file format support adding comments for translators? I could give some guidance as to pronunciation and meaning then, rather than just having comments in the data files.

BTW we already managed to smuggle a reference to [Táin Bó Cúailnge](https://en.wikipedia.org/wiki/T%C3%A1in_B%C3%B3_C%C3%BAailnge) into the Irish and Scottish Gaelic mission texts :)

nozmajner:
That was just me screwing up the copy/paste from the sys file.

PR #4474, opened 2018-10-17
Fix the unused variable warnings

fluffyfreak:
Fix the unused variable warnings caused by (Json::type_error &e)

This only seems to affect VS201X, are unused variable warning suppressed in our makefiles?


impaktor:
@fluffyfreak I think @Web-eWorks added supression of warnings in makefiles, due to the noise from thirdparty or contrib

Web-eWorks:
I was originally going to do *something* with those error objects (likely pass some info to `SavedGameCorruptException`), but this is a good interim measure.

PR #4478, opened 2018-10-19
Support 'module.submodule' import syntax.

Web-eWorks:
Extend our `import()` to support Lua 5.2+ `require()` semantics - specifically, allow using `a.b.c` as a shortcut for `a/b/c.lua` or `a/b/c/init.lua`.

This change is seamless and does not require updating any existing code in either the C++ or Lua.

CC @ecraven for review / merge.

### DIFF
--- a/src/LuaUtils.cpp
+++ b/src/LuaUtils.cpp
@@ -274,6 +274,23 @@ static bool get_cached(lua_State *L, const std::string& name)
 }
 
 /**
+	Convert a module path in the form `module.submodule.abc` into a file path
+	in the form `module/submodule/abc`.
+*/
+#include <algorithm>
+static std::string table_path_to_file_name(const std::string &path) {
+	const char separator = '.';
+	std::string out = path;
+
+	// If we have an old-style 'module/path/file.lua' path, return it normalized.
+	// Otherwise, transform 'module.path.file' -> 'module/path/file'.
+	if (path.find('/') == std::string::npos) {
+		std::replace(out.begin(), out.end(), '.', '/');
+	}
+	return FileSystem::NormalisePath(out);
+}
+
+/**
 * Finds field in the cache.
 * Require a existing cache table in the stack at index -1.
 * Returns true if field not nil (exists), puts value to lua stack.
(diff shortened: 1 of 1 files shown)


Web-eWorks:
Per a request from @ecraven on IRC, I'll add a little explanation here for what this PR accomplishes.

In Lua 5.1, package management was a fair bit of a mess. You had the `module()` syntax, which was prone to security issues, didn't support returning values from the file, etc. In Lua 5.2, this syntax was removed in favor of a new mechanism, called `require()`.

`require()` was called with a specific syntax - it was passed one string argument with the name of the module to load, and returned whatever the module returned, or threw an error (with one special exception - if the module didn't return anything, `require()` returned `true`).

Now, the Lua designers had a bit of a conundrum on their hands - how do you handle looking up these modules? Especially with C modules (which could wrap several submodules), and system-level modules, you couldn't just use raw file paths - there were too many ways that could break.

They came up with a generic, cross-platform syntax that closely resembled normal table access - if there was a module (let's say `lext`), and a submodule (let's say `random`), you simply wrote your call as `require ('lext.random')`, and the Lua runtime handled all of the fiddly, platform-specific bits, including loading C modules and looking for packages in multiple directories.

At the end of the day, that call to require was essentially translated into a `pcall( dofile, '/usr/local/share/lua/5.2/lext/random.lua' )`, which would run the file and pass the return value back to the calling code. The directories for lookup were controlled by the `package.path` and `package.cpath` variables, and you could insert values into `package.searchers` or `package.preload` or `package.loaded` to control various stages of the lookup process.

We're circumventing the `require()` functionality because we don't want the user loading arbitrary C or Lua code into Pioneer's execution context from system-level files, but the fact remains that our current syntax of directly referring to files by path is both non-elegant and directly contradictory to what most experienced Lua programmers are used to.

This specific change allows using the `module.submodule` syntax to load Lua files, in addition to the existing `module/submodule.lua` syntax. It additionally brings in a little utility feature from Lua 5.2 (or perhaps 5.3), by which a file `module/submodule/init.lua` may be considered if `module/submodule.lua` is not present.

I will be putting both of these features to *great* use in my infoview PR, so I would appreciate it if this PR could get merged as soon as possible.

impaktor:
> I will be putting both of these features to great use in my infoview PR

OK, so we hit merge button sometime during the weekend, unless people protest, or anounce they want time to review. (@laarmen look! shiny-lua stuff! :wink: )

> so I would appreciate it if this PR could get merged as soon as possible.

You can work off a branch off of this branch in the mean time, I assume.

Web-eWorks:
@impaktor 
> You can work off a branch off of this branch in the mean time, I assume.

Way ahead of you. :smile:

impaktor:
@laarmen Oh, there's no pleasing me! I'm high maintenence.

On a serious note, thanks for review!

PR #4482, opened 2018-10-24
Disable the SDL_CaptureMouse path in IMGUI code.

### LINKED ISSUE #4481: Mouse target and pointer misalignment
mousepointer
[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/2499632/opengl.txt)
[imgui.ini](https://github.com/pioneerspacesim/pioneer/files/2499634/imgui.ini.txt)

Bug is absent in 7d64219900, and present in 9f60062f99f and all descendants.

### Observed behaviour
All parts of the game consistently behave as if the mouse pointer were lower than it actually is.

### Expected behaviour
The tip of the mouse pointer should be the same as the program's internal representation of the mouse position.

### Steps to reproduce
Run Pioneer

__ver git (b360ab0) on: Linux__

Web-eWorks:
Fixes #4481.

For some reason, the way the IMGUI implementation handles capturing the mouse causes it to snap to the center of the screen whenever we right-click. Additionally, the polling of global coordinates instead of window-local coordinates is causing #4481.

I plan on revisiting this to implement it correctly, but for now this fix should work just fine.

@fluffyfreak has cursorily confirmed that this change is functional on IRC.
@clausimu a sign-off that this PR doesn't cause any undue problems (e.g. when dragging outside of the window, etc.) would be appreciated if you have time.

### DIFF
--- a/contrib/imgui/examples/imgui_impl_sdl.cpp
+++ b/contrib/imgui/examples/imgui_impl_sdl.cpp
@@ -271,7 +271,11 @@ static void ImGui_ImplSDL2_UpdateMousePosAndButtons()
     io.MouseDown[2] = g_MousePressed[2] || (mouse_buttons & SDL_BUTTON(SDL_BUTTON_MIDDLE)) != 0;
     g_MousePressed[0] = g_MousePressed[1] = g_MousePressed[2] = false;
 
-#if SDL_HAS_CAPTURE_MOUSE && !defined(__EMSCRIPTEN__)
+// Using this code-path causes several issues in Pioneer - when capturing the
+// mouse, the mouse snaps back to the center of the window, and there are some
+// issues with the WM's decoration extents being calculated into the window's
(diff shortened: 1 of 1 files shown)


laarmen:
I'm wondering how it happened that we use a file that's in the examples
directory of imgui. Wouldn't it be a good idea to import it in our own
sources ?

fluffyfreak:
@laarmen Typically you'd write [your own interface](https://github.com/fluffyfreak/pioneer/tree/imgui-pi-impl) that uses your own renderer interfaces. But that wasn't done at the start and so we're still relying on the examples.

Web-eWorks:
Better integration of IMGUI into our rendering loop is planned. In the mean time, @impaktor I belive this PR can be merged - @fluffyfreak has signed off on IRC, and both clausimu and quickdudley have tested it.

PR #4485, opened 2018-10-27
Restore 3 action bindings to event dispatch sequence.

quickdudley:
Fixes unreported bug: restores functionality to the actions under View/Miscellaneous.

In PR#4440: the handlers for these actions had been attached under `WorldView::InputBindings` (before this they were under `KeyBindings`), but there was no control flow path to the bindings in that object from the main event loop.


src/WorldView.cpp
```diff
@@ -1132,3 +1132,9 @@ vector3d WorldView::GetMouseDirection() const {
 		mouseDir = -mouseDir;
 	return (Pi::player->GetPhysRadius() * 1.5) * mouseDir;
 }
+
+void WorldView::HandleSDLEvent(SDL_Event &event) {
+	InputBindings.toggleHudMode->CheckSDLEventAndDispatch(&event);
+	InputBindings.increaseTimeAcceleration->CheckSDLEventAndDispatch(&event);
+	InputBindings.decreaseTimeAcceleration->CheckSDLEventAndDispatch(&event);
+}
```
impaktor: github says you need to press RET here.

laarmen:
We really need to have a clang-format flag day, because this PR is really hard to review *because* of the tool. That's a shame (and not the OP's fault, obviously)

fluffyfreak:
I am assuming that the large number of diff changes are from the clang-format?
It's hard to tell what's actually different, ah well.

What name would you like to be known under in the `AUTHORS.txt` file after I merge this?

quickdudley:
The large number of diff changes is indeed from clang-format: but fortunately my own changes are all in the first two commits and the last two commits are just formatting.

I would like to be known as Jeremy List in `AUTHORS.txt`

PR #4486, opened 2018-10-30
NewBuildings-01

Jay-Cup:
screen1
screen2
screen3

Six new buildings, two domes, two towers, and two smaller apartment style buildings.


impaktor:
Do you have images / screenshots?

Jay-Cup:
I added a few images.

impaktor:
I don't know if you've seen discussions on domes (a bit old):
https://pioneerspacesim.net/forum/viewtopic.php?f=6&t=240
https://pioneerspacesim.net/forum/viewtopic.php?p=2967#p2967

Ideally, on vacuum worlds, all buildings are of dome kind (we would need more different sized domes, and linking them together), maybe something you want to contribute. 

Btw, the contributor in the forum posts have not been heard from in years, so feel free to copy and steal.

Jay-Cup:
I do remember seeing this a while back. How would the dome linking system work? either way I would like to do more domes, for sure.

impaktor:
One way is to have separate dome models, and then build new models from these. Just like ships can be assembled from sub-model parts, like landing wheels, or cockpits, I think. @nozmajner knows more.

Alternative is to do it in code, which would require work at undefined time, by undefined person, meaning: it won't get done at all.

nozmajner:
Sorry for the delay. They look good, and could be merged.

impaktor:
Can models have tags, indicating it's suitable for vacuum worlds as well? (just thinking ahead)

fluffyfreak:
@impaktor there's nothing in the code at the moment. Better to add tags when the code and structure is put in place rather than add something in advance that is not used.

fluffyfreak:
Squashed and merged

fluffyfreak:
@Jay-Cup sorry it took so long to get this merged, real-life getting in the way.

What name do you want to be known as in the `AUTHORS.txt` file? (_if you're not already in there_)

PR #4487, opened 2018-10-30
WIP: Add a data collection mission

Jummit:
**Warning: This is my first contribution here, so be careful what you merge :D**
I added a data collection mission type, which is quite simple: You have to click collect in the mission info in a low orbit around the wanted planet/moon and then return to the station where the scientist waits.

# Things I think could be improved, but I need feedback on:
* I suggest changing the reward (400) to scale with the distance traveled
* Better "is near planet to inspect" calculations
* More translations
* Add a due system like in every other mission
* Better translation variables (I currently concatenate strings like this: FLY_TO..planet..AND_DO_STUFF)
# Suggestions to extend this:
* Add data collection missions that reach out of the current system
* Add science reputation
* Extend this by adding surface/gas sample missions

### DIFF
--- a/data/lang/module-collectdata/en.json
+++ b/data/lang/module-collectdata/en.json
@@ -26,5 +26,9 @@
   "INDEPENDENT_SCIENTISTS": {
     "description": "",
     "message": "Independent Scientists"
+  },
+  "VERY_GOOD_I_WILL_SEND_THIS_DATA_DIRECTLY_TO_THE": {
+    "description": "",
+    "message": "Very good, I will send this data directly to the "
   }
 }
--- a/data/lang/module-collectdata/en.json
+++ b/data/lang/module-collectdata/en.json
@@ -1,0 +1,14 @@
+{
+  "DATACOLLECTION": {
+    "description": "",
+    "message": "Data collection"
+  },
+  "THANKS_I_CANT_WAIT_FOR_THE_RESULTS": {
+    "description": "",
+    "message": "Thanks, I can't wait for the results!"
+  },
+  "OK_AGREED": {
+    "description": "",
+    "message": "Ok, agreed."
+  }
+}
--- a/data/lang/module-collectdata/en.json
+++ b/data/lang/module-collectdata/en.json
@@ -5,20 +5,26 @@
   },
   "THANKS_I_CANT_WAIT_FOR_THE_RESULTS": {
     "description": "",
-    "message": "Thanks, I can't wait for the results!"
+    "message": "Thanks, we can't wait for the results!"
   },
   "OK_AGREED": {
     "description": "",
     "message": "Ok, agreed."
   },
-  "SCIENCELABS": {
-    "INTERGALACTIC_RESEARCH_FOUNDATION": {
-      "description": "",
-      "message": "Intergalactic Research Foundation"
-    },
-    "EXTRATERESTIAL SCIENCE LABS": {
-      "description": "",
-      "message": "Extraterestial Science Labs"
-    },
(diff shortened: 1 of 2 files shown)


impaktor:
> Better translation variables (I currently concatenate strings like this: FLY_TO..planet..AND_DO_STUFF)

yes, we can't merge unless strings are translatable. I.e. no "clever" sentence building like how original Elite used the fact that "COMMANDER" and "COMPUTER" star with the same three letters, thus freeing up memory by exploiting it. Strings must be able to be translated for our translators.

Also, we/I have a scout mission that's 98% done. It sends player flying close to surface of planet for X minutes. I don't know how overlapping that mission would be with this one. 

Also, welcome, and thanks for contributing :)

Jummit:
> Also, we/I have a scout mission that's 98% done. It sends player flying close to surface of planet for X minutes. I don't know how overlapping that mission would be with this one.

Darn, that's unfortunate :). I think these missions can be further developed, just heading in a different direction. For example, the scout mission could also be applied for space station or (abandoned?) cities. The data collection mission can, as mentioned in the OP, also be extended to surface or gas sample gathering.

data/lang/module-collectdata/en.json
```diff
@@ -0,0 +1,74 @@
...
+    "description": "",
+    "message": "Ok, agreed."
+  },
+  "INTERGALACTIC_RESEARCH_FOUNDATION": {
+    "description": "",
+    "message": "Intergalactic Research Foundation"
+  },
+  "EXTRATERESTIAL_SCIENCE_LABS": {
+    "description": "",
+    "message": "Extraterestial Science Labs"
```
zonkmachine: Extraterrestrial

data/lang/module-collectdata/en.json
```diff
@@ -0,0 +1,74 @@
...
+    "message": "The "
+  },
+  "WANTS_YOU_TO_COLLECT_DATA_FROM": {
+    "description": "",
+    "message": " wants you to collect data from "
+  },
+  "SET_STATION_TO_BRING_THE_DATA_TO_AS_TARGET": {
+    "description": "",
+    "message": "Set station to bring the data to as target"
+  },
```
zonkmachine:
```
  "SET_STATION_TO_BRING_THE_DATA_TO_AS_TARGET": {
    "description": "",
    "message": "Set station to bring the data to as target"
  },
```
This is a duplicate

data/lang/module-collectdata/en.json
```diff
@@ -0,0 +1,74 @@
...
+    "description": "",
+    "message": "Data from "
+  },
+  "COLLECTED": {
+    "description": "",
+    "message": " collected"
+  },
+  "SET_TO_BE_INSPECTED_PLANET_AS_TARGET": {
+    "description": "",
+    "message": "Set to be inspected planet as target"
```
Keeper1st: Took a few times reading this to figure out what it means. Hyphens are important. "Set to-be-inspected planet as target" is understood easily. And for the next entry, perhaps "Set delivery station as target" as the "station to bring the data to" is a bit awkward.

fluffyfreak: Or "_Set planet as target to be inspected_" for a sentence that makes a bit more sense :D

craigo-: Is the current standard "Set destination as navigation target" not accurate for this mission type?

Otherwise, for consistency: "Set planet to be inspected as navigation target"?

Is the destination always a planet? The initial description suggests not... Perhaps change the word "planet" to "body"?

Jummit:
Closing as I have no plans on finishing this.

PR #4491, opened 2018-11-04
Further extend soundtrack

WKFO:
If there is a better way to accomplish what I did in `data/modules/MusicPlayer.lua` at line 158, let me know. Because I don't want to break anything.

Basically I don't want music to change whenever a ship comes by a starport while you are docked.


WKFO:
It would be nice if we had music modules for some mission types, which would override the default player. For example when the player finds the broken ship in a SAR mission, we don't need to play a suspenseful track (the player knows that the ship is not hostile); but in most other encounters we do need to play it (because most deep-space encounters are hostile).

laarmen:
@WKFO Do you still have the source files for all the tracks you made for Pioneer ? If so, would you be willing to share them, and how big are they ?

WKFO:
@laarmen No, but I think I still have some good portion of them. I would be willing to share most of what's left, but size depends on the format because I don't use the same production method for every track.

fluffyfreak:
Ok, this PR looks good to merge for me.
Is there anything else you want to add @WKFO ?

impaktor:
(ideally the "code improvement" commit aught to be squashed into the first)

WKFO:
I can just rebase everything into a single commit when I get home.

WKFO:
@fluffyfreak No, that's all, there is nothing else I've planned to do.
I squashed everything into a single commit for convenience.

PR #4495, opened 2018-11-24
Adjust hyperspace exit calculations

cwyss:
A few tweaks to the calculations of the hyperspace exit location,
inspired by fluffyfreak's recent changes from #4444.

This PR
- makes the calculations of exit direction and position mathematically more rigorous,
- places the hyperspace exit point further away from the system body,
though still in the centre of the system.

### Some details/comments

At two places in the code a vector orthogonal to a given vector in 3-dim space is needed. For this I added a function OrthogonalDirection() in MathUtil.h, which computes it in a mathematically save way.

The calculation of the exit distance is based on the mass and the radius of the star:
The distance is at least ten times the radius of the star and such that the velocity of a circular orbit at this distance is at most 100km/s. The second condition makes life easier for small ships with low thrust and/or small amount of delta v. Finally a third condition is that the exit distance is at least 0.2 AU regardless of the star. For small stars like sol or Sirius it is always this third condition which actually determines the distance.

Given the above "comfortable" exit distance, it is not critical to provide the ship with the velocity for an orbit, and therefore I kept the original fixed 100m/s.

I tested my code with about a dozen of stars of different sizes. Typically you end up inside the orbit of the innermost planet of the system.

The "right" choice of exit distance is certainly subject to debate. (In the code the distance can be easily tweaked by adjusting some constants)
I found fluffyfreak's choice to close to the star given my experience when flying with a Xylophis: After one jump to a bigger star (some giant, forgot which one), the planet I wanted to go to lay opposite to the direction that I was orbiting the star. I made the mistake to turn around and accelerate towards the planet. My velocity relative to the planet decreased, but since I was so close to the star the ship could not build up speed towards the planet fast enough and I crashed into the star. To fly to the planet, I actually had to orbit half way around the star and then accelerate to the planet. (I wonder how an autopilot would have solved this situation..)
Because of this episode I added the delta-v condition.

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -490,62 +490,9 @@ void Game::SwitchToNormalSpace()
 	m_space->AddBody(m_player.get());
 
 	// place it
-	vector3d position = m_space->GetHyperspaceExitPoint(m_hyperspaceSource, m_hyperspaceDest);
-	m_player->SetPosition(position);
-
-	// calculate orbital velocity
-	vector3d orbVelocity(0, 0, -100.0);
-	matrix3x3d orientation(matrix3x3d::Identity());
-	{
-		auto dest = m_hyperspaceDest;
-
-		// find the primary body in this system
-		Body *primary = nullptr;
-		if (dest.IsBodyPath()) {
-			assert(dest.bodyIndex < m_space->GetStarSystem()->GetNumBodies());
-			primary = m_space->FindBodyForPath(&dest);
-			while (primary && primary->GetSystemBody()->GetSuperType() != SystemBody::SUPERTYPE_STAR) {
-				SystemBody* parent = primary->GetSystemBody()->GetParent();
-				primary = parent ? m_space->FindBodyForPath(&parent->GetPath()) : 0;
-			}
-		}
-		if (!primary) {
-			// find the first non-gravpoint. should be the primary star
-			for (Body* b : m_space->GetBodies()) {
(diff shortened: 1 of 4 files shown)


fluffyfreak:
@cwyss Ok this is mostly good.

I would prefer the exit velocity calculation left in as I felt it contributed to the sense of having come from _another_ system and thus tied together the game universe a little more.

Also it meant that, for example, your difficulty in getting too a target planet is a consequence of the route you took to get there.
Altering your jump route to approach the system from another angle might take much jumps but less time overall due to being a better angle of entry into your target system.

So with this we lose that, but all of the other work is good. I just need to find some time to sit down and play it a bit.

fluffyfreak:
Also I apologise for taking so long to reply, real life is - as usual - crazy right now.

cwyss:
@fluffyfreak I agree that an exit velocity which puts you in an orbit makes things more dynamic and more challenging. And it let's you see from where in hyperspace you were coming from.

So I changed the exit velocity to that of a circular orbit. 
In fact it's only almost circular since the calculation does not use the actual randomised distance to the body, but the distance before randomisation. Though in practise this does not seem to make the orbit particular elliptic. If we want that we should perhaps apply some explicit randomisation to the velocity too.
Happy testing :)

fluffyfreak:
This looks good to merge, I just haven't had much chance to test it ingame yet. Sorry for the hold-up.

PR #4499, opened 2018-12-09
Don't show military allegiance info for player

WKFO:
It serves literally no purpose right now and might even mislead players into thinking that being a part of a military possible.

I wanted to delete the strings as well but I don't know if they are used somewhere else in the code or if they can be any useful in the future so I left them for now. Can delete them on your wish.

That UI window got a little empty for sure. Maybe we can fill that space with the crime statistics of the player. For example: 

- Wanted in Solar Federation. 5534 Cr fine for unlawful weapons discharge (x3).
4432 Cr fine for illegal hyperspace jump.
6622 Cr fine for black market activities (x2).
- Outlaw in Obsidian Horde.

(I'm not volunteering for now.)


fluffyfreak:
Yes this is a good idea and probably ready to merge.

PR #4505, opened 2018-12-28
cmake: Test for a .git directory.

orbea:
This prevents the following error when git is found and the source directory is not a git repo.
```
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```


orbea:
Travis seems to be failing with something similar to issue https://github.com/pioneerspacesim/pioneer/issues/4506 which is not related to this PR.

Issue #4504, opened 2018-12-27
configure --with-strip fails


orbea:
### Observed behaviour

When configuring pioneer with `./configure --with-strip` it will fail.
```
config.status: executing depfiles commands
config.status: error: in `/tmp/pioneer':
config.status: error: Something went wrong bootstrapping makefile fragments
    for automatic dependency tracking.  Try re-running configure with the
    '--disable-dependency-tracking' option to at least be able to build
    the package (albeit without support for automatic dependency tracking).
See `config.log' for more details
```
Full log: https://pastebin.com/x9ERwCPP

If `--with-strip` is omitted or if `--disable-dependency-tracking` is also used then configure will succeed, but neither is ideal. 

I bisected this and found the first bad commit.
```
ec70cd27bd7e3d50db97c07f85ea8693b22dad25 is the first bad commit
commit ec70cd27bd7e3d50db97c07f85ea8693b22dad25
Author: Webster Sheets <webster@web-eworks.com>
Date:   Sat Sep 29 12:15:56 2018 -0400

    Fix some crashes relating to loading old saves.
    
    Catch JSON type errors in LuaGame.cpp.
    
    Apparently, Autotools doesn't like tabs. Who'da thunk it?
    
    Fixup the saved game loading code.
    
    Fix save loading errors, patch JSON library to throw instead of asserting.
    
    Theoretically finally fix the serialization code.
    
    String lifespan fix in unpickle_json

:040000 040000 cd403501c8b38853d57d934faace3117390a7b88 49c367d279268936e84b145c73078f578b04f0d4 M	contrib
:040000 040000 32b9b72d164525ed7d0207b2357bfbc485cc57b4 d382017f500705624cf2fbfde433c0a6dcd32f5d M	data
:040000 040000 4e44c821d5228d390c1c36b7aac00c8b0ee92bcc 625dccd4d10702c1dda2a9418cfa1a0fad49998b M	src
```
ec70cd27bd7e3d50db97c07f85ea8693b22dad25

### Expected behaviour

pioneer should configure with `--with-strip` enabled.

### Steps to reproduce

1. `./bootstrap`
2. `./configure --with-strip`

__My pioneer version (and OS):__

pioneer: https://github.com/pioneerspacesim/pioneer/commit/3fdd945ee43ae73b43cf61f8f41e6cf6eeb31f9a
OS: `Slackware64-current`

orbea:
@Web-eWorks Do you have any ideas about this?

[...]

PR #4507, opened 2019-01-01
Happy New Year!

impaktor:


Web-eWorks:
Clang format does not like Automake files. You'll need to revert those, @impaktor.

Alternatively, we could just drop Autotools support entirely and close #4504 and #4461 while we're at it...

impaktor:
I included a clang reformat, and swapped order of commits, so both should go through travis, but clang format also wants to change Makefile.am files, so it still fails.

@Web-eWorks so should clang-format be applied to Makefile.am or not?

Web-eWorks:
However you ran clang-format, it didn't actually get applied to the files correctly. There's still some issues with pointer alignment (`type*` vs `type *`), and you definitely shouldn't run clang-format on Autotools files (e.g. Makefile.am, etc.)

impaktor:
> However you ran clang-format, it didn't actually get applied to the files correctly.

From top pioneer directory I'm running:
`clang-format -i -style=file src/**/*.{cpp,h}`

I've re-done both commits, still failing.

When I run `./scripts/clang-format.sh` it returns:
```
Files in this commit comply with the clang-format rules.
```
So I suspect it's something on travis that's wrongly configured?

Web-eWorks:
@impaktor By default, the clang-format script in `scripts/` checks only the files currently un-committed. It's meant as a sanity-check hook for when contributors are committing changed code. (Alliteration, ho!)

I'm not sure what's going wrong on your end - #4508 passes the clang-format check just fine. Go ahead and revert that commit; I've got it taken care of.

Issue #4506, opened 2018-12-28
sandbox violation - modelcompiler segfaults during cmake


orbea:
### Observed behaviour

When running the cmake build with sandbox causes the `modelcompiler` to segfault when it fails to access the mesa shader cache.
```
[100%] Building CXX object CMakeFiles/modelcompiler.dir/src/modelcompiler.cpp.o
[100%] Building CXX object CMakeFiles/pioneer.dir/src/main.cpp.o
[100%] Linking CXX executable modelcompiler
Optimizing models
[100%] Linking CXX executable pioneer
 * ACCESS DENIED:  open_wr:      /home/orbea/.cache/mesa_shader_cache/index
make[2]: *** [CMakeFiles/modelcompiler.dir/build.make:100: modelcompiler] Segmentation fault
make[2]: *** Deleting file 'modelcompiler'
make[1]: *** [CMakeFiles/Makefile2:122: CMakeFiles/modelcompiler.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[100%] Built target pioneer
make: *** [Makefile:130: all] Error 2
 * --------------------------- ACCESS VIOLATION SUMMARY ---------------------------
 * LOG FILE: "/tmp/sandbox-30854.log"
 * 
VERSION 1.0
FORMAT: F - Function called
FORMAT: S - Access Status
FORMAT: P - Path as passed to function
FORMAT: A - Absolute Path (not canonical)
FORMAT: R - Canonical Path
FORMAT: C - Command Line

F: open_wr
S: deny
P: /home/orbea/.cache/mesa_shader_cache/index
A: /home/orbea/.cache/mesa_shader_cache/index
R: /media/data/home/.cache/mesa_shader_cache/index
C: /tmp/pioneer/build/modelcompiler -b inplace 
 * --------------------------------------------------------------------------------
```
And I can reproduce it manually.
```
$ sandbox build/modelcompiler -b inplace
 * ACCESS DENIED:  open_wr:      /home/orbea/.cache/mesa_shader_cache/index
Sandboxed process killed by signal: Segmentation fault
 * --------------------------- ACCESS VIOLATION SUMMARY ---------------------------
 * LOG FILE: "/tmp/sandbox-4706.log"
 * 
VERSION 1.0
FORMAT: F - Function called
FORMAT: S - Access Status
FORMAT: P - Path as passed to function
FORMAT: A - Absolute Path (not canonical)
FORMAT: R - Canonical Path
[...]

PR #4510, opened 2019-01-04
Fix crash in modelcompiler and add missing DESTDIR.

orbea:
This fixes the crash in modelcompiler described in comment https://github.com/pioneerspacesim/pioneer/issues/4506#issuecomment-451266374. The solution is thanks to sturnclaw in irc.

It also adds a missing `${DESTDIR}` to `CMakeLists.txt` so the data files are installed to the correct location.

The remaining problem is that the modelcompiler doesn't have a way to specify an output directory and doesn't respect `${DESTDIR}` as a result.


Web-eWorks:
Specifically, the issue is that the modelcompiler doesn't have a way to process an arbitrary directory. The output directory (especially for an inplace build) is the same as the input directory, which assumes that the `data/` dir has already been installed to it's final resting place.

Web-eWorks:
@impakor: I'll be addressing the modelcompiler issue in another PR; this one's good to merge.

orbea:
I think this might of also fixed the gcc travis build, only the static check is failing.

pcercuei:
From my tests, and confirmed by @orbea over IRC, the addition of ```${DESTDIR}``` is not needed, as:

- the data files properly installed to the expected directory when running ```DESTDIR=/tmp/foo ninja install``` even before this addition;
- ```${DESTDIR}``` probably resolves to an empty string, as it is an environment variable and not a cmake variable (it should have been ```$ENV{DESTDIR}```

Therefore I suggest that we revert b97fe6147.

Web-eWorks:
Can confirm that `${DESTDIR}` does not produce any observable result and is contrary to the way the DESTDIR functionality works. +1 for revert.

pcercuei:
Reverted in #4513.

PR #4508, opened 2019-01-02
Clang-format all the things!

Web-eWorks:
(Finally) Apply clang-format to the entire code base. Will likely need another pass to resolve hidden dependency errors (and as such the build will likely fail), but the first bit is done.

CC @impaktor: I'll get the clang-format stuff, you just handle updating the names and dates in your PR.

Now without Autotools files getting messed up!

Closes #4445.

PR #4514, opened 2019-01-06
Faster build after commit

pcercuei:
This set of commits allows for a much faster build time after a commit is made. Instead of recompiling the whole ~420 source files, only about ~10 are now recompiled.


fluffyfreak:
Removing the `buildopts.h` means Visual Studio won't build. Still need to get it working with CMake somehow.

If anyone has any ideas that'd be great 'cos I'm out of my depth in this area.

jaj22:
Well, you could chuck a buildopts.h into win32/include as a workaround. Given that the MSVC build didn't pass version numbers before, it seems reasonable.

Web-eWorks:
@fluffyfreak visual studio isn't working with CMake because of pioneer-thirdparty, right? Not sure how easy that is to resolve, but I'd be willing to take a crack at helping you get the CMake side working.

@pcercuei changes look pretty good. The undefined ObjectViewerView type gives me the shivers, but that's destined for the garbage heap anyways, so I can look past it. It looks like this doesn't actually improve the first-compile time, but if it makes a partial recompile faster, I'm all for it.

Let's hold off on this until #4508 is merged - it'll make it a bit easier if you're rebasing onto the properly formatted files instead of me having to rebase my giant PR again.

Web-eWorks:
@pcercuei you're good to rebase and merge now!

Web-eWorks:
@laarmen commit hook checks all changed files. I plan on editing it to only check files added to the index, but I can't make it only check a specific diff or section of a file.

@pcercuei It looks like clang-format is (mostly) complaining about out-of-order #includes in your commit. Not sure about Pi.cpp; might have been something that got merged ontop of my commit.

Go ahead and apply the clang-format changes to your PR.

pcercuei:
It complains about out-of-order includes, but that was before my commits :)

I'll fix the PR.

pcercuei:
It should work now.

impaktor:
Looks like there are still complaints.

Just out of curiosity, why do you add `Signed-off-by: Paul Cercueil <paul@crapouillou.net>` to all commits? The commit already has the name and email address you've configured in the git config file (either local or global).

pcercuei:
It's the tag for the Developer Certificate of Origin. See: https://developercertificate.org/
```git commit -s``` adds it automatically. I use it for all my contributions.

**Comment by fluffyfreak** [comment]
@jaj22 good idea, either @pcercuei can do it in this PR or I'll sort the `win32/include` header out for Visual Studio afterwards.

**Comment by pcercuei** [comment]
What about we make Visual Studio work with our CMake instead? It would be time. I guess I can help with that, too.

**Comment by laarmen** [comment]
I tried it last spring, it's... Not easy 😕

Le mar. 8 janv. 2019 à 18:07, Paul Cercueil <notifications@github.com> a
écrit :

**Comment by pcercuei** [comment]
Sorry for the noise - it finally passes the static analyzer check now.

**Comment by impaktor** [comment]
autotools fails when building pioneer, after change done by this PR. Will that effect OSX or MSVS builds? Or are all systems on cmake now?
```
BVHTree.cpp:5:10: fatal error: buildopts.h: No such file or directory
 #include "buildopts.h"
```

pcercuei:
It was fixed for MSVC in #4521.

PR #4517, opened 2019-01-13
Non-instant system exploration

WKFO:
Instead of instantly exposing the celestial bodies, it will take a few seconds to explore unexplored systems. The more major bodies (stellar bodies, moons, planets) in a system, the more seconds it will take for sensor data to be gathered and analysed. The overview window (the new (pigui) one, as well as the old one) will be unusable for that duration.

Same with hidden bases, the more bases there are, the longer it will take for the bases to be revealed. Bases will not be revealed before the major bodies are discovered (because where is logic in that?)

Also adding a few short 'rare discovery' tracks to play when you explore a system with a rare stellar object, such as a black hole or a wolf-rayet star. (Just these two, actually.)

You can still probably hyper-hop around systems, because it still doesn't take long at all. To be honest, if Pioneer was my personal game, I would make the system exploration process last at least a bit more than few hours with the retro-futuristic approach of Pioneer (that reveals planets one by one starting with the biggest ones), but that could make people angry.

I tried to introduce the least amount of bugs, but feel free to yell at me.


fluffyfreak:
Code looks fine, I haven't had a chance to play-test this yet but I'm gonna trust that the music is upto your usual excellent standard ;)

fluffyfreak:
Regarding:

> To be honest, if Pioneer was my personal game, I would make the system exploration process last at least a bit more than few hours with the retro-futuristic approach of Pioneer

I agree, system exploration should be multi-layered. First find the planets, then scan each of the planets, then find the asteroids (_perhaps leave probes for weeks or months, return to collect results_) etc etc etc.

So maybe view this is a first step in a better direction eh?

PR #4511, opened 2019-01-04
Remove the (unmaintained) Autotools build method.

Web-eWorks:
Closes #4461, closes #4504.

This PR finishes off the series of CMake improvements. CMake is now stable enough that it's the main compilation method for non-Windows platforms, and Travis is now exclusively using CMake as it's build tool. Considering the issues we have with maintaining the autotools port (no one wants to do it, it's 1700 lines of DSL, etc...) I feel it is time that we finally struck support for Autotools from the project entirely.

You are welcome to consider this a happy New Year's gift. :smile:

PR #4520, opened 2019-01-20
Include decouple

mike-f1:
Speed up compile time decoupling includes
and using forward declaration.


Web-eWorks:
You will need to run clang-format on your changed code and squash your commits before we can merge this.

Clang-format will only reorder includes in the same block, so if header reordering is causing issues for you, you can place a blank line between blocks of includes.

fluffyfreak:
@Web-eWorks Can we disable the header reordering, I don't see a case where it's the correct behaviour.

@mike-f1 you removed the file `pioneer`, can your restore that please. It needs updating for use CMake (_which someone else can do, you don't have too_) rather than removing it.

Web-eWorks:
@fluffyfreak We can do that. I'll disable that in a new PR, and update the pioneer script in #4511.

mike-f1:
Should I do something about conflict?

pcercuei:
@mike-f1, you should rebase on top of master, and fix the conflicts, yes.

fluffyfreak:
@mike-f1 if you can resolve it and push the fixes that'd be helpful.
Either via rebase or merge. I don't mind but @impaktor might have a suggestion/preference.

mike-f1:
Well, probably I did something wrong when I run "clang-format" on my own local
branch and I will amend that commit.

But I note it Travis fail because `#endif //something` is not conform to actual rules...
It shouldn't be the opposite or this is what you want?
(I say this only because AFAIK this tool seems just introduced)

PR #4521, opened 2019-01-21
Fix VS2017 buildopts

fluffyfreak:
Re-add buildopts.h in Win32 folder for VS only. Supply path too it in relevant project files.


impaktor:
As usual, this is 100% your stuff. Btw, as of last night, we now have hooked in appveyor into pioneer, so we have automagic build checks by both travis and appveyor, as you can see on this PR, as well as on master:

 https://ci.appveyor.com/project/pioneerspacesim/pioneer

PR #4529, opened 2019-02-07
Add ship name and class to intro screen.

ecraven:
screenshot

### DIFF
--- a/src/Intro.h
+++ b/src/Intro.h
@@ -14,6 +14,7 @@ public:
 	Intro(Graphics::Renderer *r, int width, int height);
 	~Intro();
 	virtual void Draw(float time);
+	SceneGraph::Model *getCurrentModel() const { return m_model; }
 
 private:
 	void Reset(float time);
--- a/data/pigui/mainmenu.lua
+++ b/data/pigui/mainmenu.lua
@@ -168,6 +168,20 @@ local function showMainMenu()
 			ui.withFont("orbiteer",36 * (ui.screenHeight/1200),function() ui.text("Pioneer") end)
 		end)
 	end)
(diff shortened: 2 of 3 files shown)


nozmajner:
Nice idea!

impaktor:
Your code doesn't conform to clang thingy.

I've compiled and tested. Works as advertised.

nozmajner:
Erm, on that screenshot, the ship and the text doesn't match.

laarmen:
Soooo... Not as advertised, then :)

impaktor:
> Soooo... Not as advertised, then :)

I'm required to know the name of all the ships in game? What do you ask of me? It didn't crash and that was good enough for me.

nozmajner:
I don know, but didn't spotted it at first.

laarmen:
You didn't know ? All ships, as well as the names of all systems less than
100ly away from Sol.

Nice catch, noz :)

src/Intro.h
```diff
@@ -14,6 +14,7 @@ class Intro : public Cutscene {
 	Intro(Graphics::Renderer *r, int width, int height);
 	~Intro();
 	virtual void Draw(float time);
+	SceneGraph::Model *getCurrentModel() const { return m_model; }
```
mike-f1: `const` ?

ecraven: does not modify the object, thus cunst.

Keeper1st:
I found the font size of 24 instead of 30 looked a little better. It would be really slick if the name could be displayed only when the ship is fully zoomed in. (Centering the name would be nice too...)

ecraven:
> I found the font size of 24 instead of 30 looked a little better. It would be really slick if the name could be displayed only when the ship is fully zoomed in. (Centering the name would be nice too...)

Does #4534 work for you?

PR #4534, opened 2019-02-11
Intro: Font size of ship name -> 24, show only if fully zoomed.

ecraven:
As requested by Keeper1st ;)

Reduce font size from 30 to 24.
Only show the ship name while the ship is fully zoomed in.


Keeper1st:
Neat! That is a lot slicker. For my own installation I went a step further, putting the class name on a new line at a slightly smaller font (20). See this image [http://www.keeper1st.com/pioneer/pseedmenu.png](url)

PR #4535, opened 2019-02-11
Add instructions on building pioneer with CMake and MSYS2

dlshcbmuipmam:
The documentation lacks the instruction of CMake build in Windows, but it is strict forward to do so in MSYS2. I have added a few lines to describe that.

### DIFF
--- a/COMPILING.txt
+++ b/COMPILING.txt
@@ -5,11 +5,12 @@ Table of Contents
     1.1.a Autotools
     1.1.b CMake
   1.2 Windows - MSVC
-  1.3 Windows - Autotools (Linux cross-compile)
-  1.4 OS X - Autotools
-  1.5 OS X - Homebrew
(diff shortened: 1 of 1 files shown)


impaktor:
Nice, thanks!

pcercuei:
What's about the CMAKE_SH trick? I never had to do that.

dlshcbmuipmam:
> What's about the CMAKE_SH trick? I never had to do that.

This is a strange limitation of CMake and MSYS2. There is a sh.exe under MSYS2 path, and you will get an error because of that when you run CMake in MinGW shell. "CMAKE_SH="CMA KE_SH-NOTFOUND" tells CMake to ignore the sh.exe.

fluffyfreak:
I don't know why this hadn't already been merged so I've done it now :)

Thanks for the instructions @dlshcbmuipmam

PR #4536, opened 2019-02-15
Add InputFrames for contextual input, split ship view controller.

### LINKED ISSUE #3032: Cannot look around using middle mouse button while game is paused
When the game is paused the middle mouse button lookaround feature does not function. The arrow keys still change the view correctly. 

(issue on asus g75vw laptop using external mouse on version 20140704-win32)

Web-eWorks:
Fixes #3032 as a side-effect, and resolves the unreported issue of the ship view rotating when dragging MMB in another view.

I've written a preliminary implementation of the InputFrame contextual input system, which will let us enable/disable inputs and input code as needed with a minimum of hassle. It's intended to piggyback ontop of the View system and other context-sensitive enable/disable triggers, with a simple interface for adding and removing Frames.


src/Input.h
```diff
@@ -115,6 +153,8 @@ class Input {
 	std::map<std::string, BindingPage> bindingPages;
 	std::map<std::string, KeyBindings::ActionBinding> actionBindings;
 	std::map<std::string, KeyBindings::AxisBinding> axisBindings;
+
+	std::list<InputFrame *> inputFrames;
```
mike-f1: I don't see any `new` or `delete` around, so there's a reason preferring that way
instead of:
`std::list<InputFrame &> inputFrames;`

(Note compiler will allow `delete (InputFrame *) ptr_to_InputFrame;` ...)

Second question is: why a `std::list` and not a `std::vector`?

Web-eWorks: InputFrames are managed by other code not owned by the Input system. It is expected (and required) that the code responsible for pushing an InputFrame pops the frame before destroying it. However, the non-nullable property of a reference would be better. I'll get that fixed up. (I'm much more fluent in C/Lua than C++ - I forgot references were a storable thing.)

`std::list` was chosen for the constant-time insertion / deletion property, but considering that it's something small like a reference, I will probably swap it to a `std::vector`.

Thanks for the review!

Web-eWorks: EDIT: I also didn't know that STL containers hate references and they're more trouble than they're worth. Pointers it is!

mike-f1: > EDIT: I also didn't know that STL containers hate references and they're more trouble than they're worth. Pointers it is!

Sorry, I had to check my brain was on :/ ...

impaktor:
"checks have failed"

Web-eWorks:
@impaktor I'm not done with this PR yet. I've still got to refactor PlayerShipController - once I'm done with that, then I'll care about the checks.

Web-eWorks:
@Keeper1st I think you'll like these changes. I've added a pair of independent axes for head-tracking, and fixed the choppiness in regular mouselook ontop of that. You'll want to configure Opentrack to send X and Y axis data as joystick axes; each axis rotates 90 degrees left and right of center (`-1 .. 1`), and center is set with the normal 'switch view' keys (front, back, etc.)

Note: normal mouselook is disabled while the headtracking axes are != 0. You can configure the deadzone of each axis in-game, and switching away from the internal camera disables the headtracking. If you *need* head-look in external mode I can set that up, but try this first.

Web-eWorks:
To whom it may concern: I've refactored the CameraController API fairly heavily. Now there's one function for each axis instead of two, and values are passed in radians instead of degrees (as before you had to convert to degrees, only for the values to be internally converted back to radians).

@impaktor I think this PR is ready for merge - I haven't refactored PlayerShipController yet, but this PR is long enough already, so moving it around will simply have to do.

Keeper1st:
Awesome, I'll give it a go as soon as I can. Working 12-hour overnight shifts for the next week or so...

PR #4538, opened 2019-02-15
Another decouple

mike-f1:
Remove includes here and there; Move ShipAICmd methods in *.cpp in
order to allow a deeper includes decoupling

Minor changes:

- in LuaShip.cpp: little variable naming change in order to avoid misread
- in Ship.cpp: little rewording in a comment


impaktor:
@fluffyfreak Ideally this should have been squash-merged, as now we have yet another commit that just moves a bunch of white space around...

src/Pi.cpp
```diff
@@ -789,6 +786,7 @@ void Pi::Quit()
 		_pclose(Pi::ffmpegFile);
 	}
 	Projectile::FreeModel();
+	Beam::FreeModel();
```
mike-f1: @fluffyfreak , just because in PR description I forgot to mention this and change in src/LuaShip.cpp#678

PR #4540, opened 2019-02-17
Add OSX to travis test matrix, upgrade Linux tests to 16.04 Xenial

Web-eWorks:
I've updated our travis testing environment to test against Mac OSX 10.13 High Sierra and upgraded the linux test environment to Xenial 16.04. This has the fortunate side-effect of getting rid of the `pioneer-thirdparty` dependency, as well as reducing compile times and simplifying our travis setup. 

This should have no effect on Appveyor builds, and is ready for merge immediately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,44 +1,59 @@
-dist: trusty
+dist: xenial
 language: minimal
+compiler: gcc
 addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
-    packages: &packages
-      - gcc-4.9
-      - g++-4.9
+    packages:
       - mesa-common-dev
       - libfreetype6-dev
       - libfreeimage-dev
       - libglew-dev
       - libsigc++-2.0-dev
       - libvorbis-dev
-      - libpng-dev
+      - libpng16-dev
+      - libassimp-dev
+      - libsdl2-dev
+      - libsdl2-image-dev
+  homebrew:
+    packages:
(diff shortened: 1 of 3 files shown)


Web-eWorks:
Pinging @impaktor @fluffyfreak for review.

@Philbywhizz the process to build with CMake on OSX is incredibly simple - just follow the wiki for CMake on linux, installing the corresponding packages through Homebrew. You can take a look at the travis build log for OSX if you need details, or open a new issue with your questions. Hopefully we won't need manually-generated OSX builds any more, but I don't have a Mac to test this stuff on.

impaktor:
Wow, if this works then that would be amazing! :fireworks: 

We could try to merge, and then push a tag and see if we get a binary for the Mac people to try; then they won't even need to setup CMake on their own end, just try the binary OSX release. 

@Web-eWorks perhaps before doing that, I wonder if it's possible to have travis give the binary build a better name, e.g.: `pioneer_<yyyymmdd>_<os>.tar.gz`, as with the last release I manually renamed the file that travis uploaded.

pcercuei:
We can go further and create a regular OSX bundle from CMake. It's not complicated, I did it in the past.

Web-eWorks:
@pcercuei You're gonna have to be the guy on deck for that - I've never touched a mac in my life.

@impaktor I'll mess with the build script and see what I can do.

PR #4543, opened 2019-02-19
Use squared distance in BodyNearFinder

mike-f1:
...And if a `double` overflow happens, print a specific message on console
so users know what happens and this change can be reversed (second commit).

Comments are welcome

### DIFF
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -32,7 +32,7 @@ void Space::BodyNearFinder::Prepare()
 	m_bodyDist.clear();
 
 	for (Body *b : m_space->GetBodies())
-		m_bodyDist.push_back(BodyDist(b, b->GetPositionRelTo(m_space->GetRootFrame()).Length()));
+		m_bodyDist.push_back(BodyDist(b, b->GetPositionRelTo(m_space->GetRootFrame()).LengthSqr()));
 
 	std::sort(m_bodyDist.begin(), m_bodyDist.end());
 }
@@ -46,10 +46,10 @@ void Space::BodyNearFinder::GetBodiesMaybeNear(const vector3d &pos, double dist,
 {
 	if (m_bodyDist.empty()) return;
 
-	const double len = pos.Length();
+	const double len = pos.LengthSqr();
 
-	std::vector<BodyDist>::const_iterator min = std::lower_bound(m_bodyDist.begin(), m_bodyDist.end(), len - dist);
-	std::vector<BodyDist>::const_iterator max = std::upper_bound(min, m_bodyDist.end(), len + dist);
+	std::vector<BodyDist>::const_iterator min = std::lower_bound(m_bodyDist.begin(), m_bodyDist.end(), len - dist*dist);
+	std::vector<BodyDist>::const_iterator max = std::upper_bound(min, m_bodyDist.end(), len + dist*dist);
 
 	while (min != max) {
 		bodies.push_back((*min).body);
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -24,16 +24,23 @@
 #include "graphics/Graphics.h"
 #include <algorithm>
 #include <functional>
+#include <cfenv>
 
 //#define DEBUG_CACHE
 
 void Space::BodyNearFinder::Prepare()
 {
+	std::feclearexcept(FE_ALL_EXCEPT);
+
 	m_bodyDist.clear();
 
 	for (Body *b : m_space->GetBodies())
 		m_bodyDist.push_back(BodyDist(b, b->GetPositionRelTo(m_space->GetRootFrame()).LengthSqr()));
 
+	if (std::fetestexcept(FE_OVERFLOW)) {
+		Output("Seems a 'FE_OVERFLOW' happens in 'Space::BodyNearFinder::Prepare()'...\nProbably you should not have used 'squared' distances\n");
+		exit(1);
+	}
 	std::sort(m_bodyDist.begin(), m_bodyDist.end());
 }
 


jaj22:
This is technically incorrect, as (a+b)*(a+b) != a*a + b*b. Could be fixed, but as it'd require retaining the sqrt in GetBodiesMaybeNear, it may not be an optimisation. The vector inserts are probably the bulk of the algorithm cost anyway.

mike-f1:
> This is technically incorrect, as `(a+b)*(a+b) != a*a + b*b`

Yes, you're right. It needs some explanations:
I did it without worrying too much that this change will decrease `std::lower_bound` and will
increase `std::upper_bound` of `2*a*b`... and so this code may increases the number of objects
returned by `GetBodiesMaybeNear`, as you saw.
_May_, because the distribution of bodies in space is "clustered": they often are near planets
(which are `Body`ies and have their `Starport`s with their inbound or outbound `Ship`s) or
really sparse (some ship from other systems). So I was betting there's no needs to take into
account the `2*a*b` term.

Note the term `2*a*b` is often little because it requires (in order to be of the same magnitude
of a^2 or b^2) that `a` and `b` are close in size.

If you require 'correct' results I could add (and subtract) `2*a*b` on lines 58-59,
where `std::lower_bound` and `std::upper_bound` are called. Doing this way there's
still a gain of speed because term `2*a*b` is calculated once in `GetBodiesMaybeNear `
while `sqrt` is called in `Prepare()` for the entire list.

> The vector inserts are probably the bulk of the algorithm cost anyway

There's a `std::sort` right after insertion and in `GetBodiesMaybeNear` you (may) fill a vector
starting from 0 size, but I agree with you: no major changes here, this is what it seems:
a little work for a little gain.

fluffyfreak:
I've got to ask because it's not obvious from the code but why does that function even work like that at all?

The thing asks for `GetBodiesMaybeNear`, and gives a position & radius. 
So what are the upper and lower bounds for?
Why copy all bodies into the vector rather than just the ones within the given radius?
Also bodies move but that thing seems to setup an initial distance.

The whole thing is completely baffling too me, and it's uncommented, wtf does that class do?

@mike-f1 / @jaj22 this isn't saying you need to fix/change it just asking for some clarity.

laarmen:
Another question on top of that : are there actual numbers regarding the
performance increase?

Le mer. 20 févr. 2019 à 15:10, Andrew Copland <notifications@github.com> a
écrit :

mike-f1:
More or less (jaj22 may corrects my answer if necessary) every timeStep `Space::TimeStep(float step)`
call `Prepare()` which calculate distances from system center for each body in `Space`, then put a pointer
and a double in a dedicated structure (see `BodyDist` definition) and then sort that list.

So you got a vector with all bodies and their distance from center, sorted.

When something asks for `GetBodiesMaybeNear` then its position act as a mean value:
you are looking for bodies placed in an spherical shell with a thickness of `2*dist`, in which
`std::upper_bound` find the most distant and `std::lower_bound` find the inner (returning
iterators to these elements)

> Why copy all bodies into the vector rather than just the ones within the given radius?

That's is just what happens:
```
while (min != max) {
		bodies.push_back((*min).body);
                ++min;
}
```
act as `std::for_each(min, max, bodies.push_back(.......))` for a given range

Note this code copy pointers, not objects (though as I noted in previous comment,
it is pushing_back on a vector which may be 0 in size...)

mike-f1:
> Another question on top of that : are there actual numbers regarding the performance increase?

No: it is all based on an assumption (largely used in code: see for example `vector*.h` and their members `Length()` and `LengthSqr()`) which is that comparing squared distances is faster than comparing distances because you haven't to perform `sqrt()` for both members, without any side effects.

mike-f1:
So? What'up? ...Do I confused you with my ~~chatter~~ fancy explanations?

Do you want me let this open or you prefer I close this PR?

And if you are to go forward, how do you want me to proceed?

Thank's

jaj22:
I'd shelve it personally. I'm normally fine with squared comparison optimisations but in this case it adds confusion with the function split, and one or two fewer sqrts per ship per frame isn't worth it.

> If you require 'correct' results I could add (and subtract) `2*a*b` on lines 58-59,
> where `std::lower_bound` and `std::upper_bound` are called. Doing this way there's
> still a gain of speed because term `2*a*b` is calculated once in `GetBodiesMaybeNear `
> while `sqrt` is called in `Prepare()` for the entire list.

I think the best case saves one sqrt out of two, and that sqrt could instead be saved by re-using the distance values in the list generated by Prepare. I imagine that's also not done because it increases the visual complexity.

I also wouldn't worry about the double overflow. The game visibly misbehaves, several orders of magnitude before distances of 2^512.

For people wondering what this algorithm is actually for: It's a quick workaround for O(n^2) ship vs ship sensor checks eating a lot of CPU in common cases. Essentially you can throw away a lot of checks early if the distance of objects from an arbitrary point isn't within the given range. The implementation is nowhere near optimal but IIRC it did make a substantial difference at the time. I don't remember why the sensor checks were expensive, however.

mike-f1:
> I'd shelve it personally

...well, good: so I will return playing with my toys then :wink:  ...

> I think the best case saves one sqrt out of two

Sure it seems I'm pedantic, but I would be sure we understand
each other well, so please allow me to check if math may work:
Now `Prepare()` calculate a distance from sys center, say '`p`' for
each Body in Space. Subsequently (in `GetBodiesMaybeNear`) you
calculate the same distance (say '`d`') for the argument of that funcion
(which is a body) and a 'threeshold', say '`t`'.
Next you check for `upper_bound` and for `lower_bound`,  and these
iterate through elements finding the first element which do not satisfy:
`p < d + t` and `p < d - t` respectively

IF `Prepare()` use squared distances, then you should square both
members; and you get:
`p^2 < (d + t)^2 = d^2 + t^2 + 2 * d * t`
`p^2 < (d - t)^2 = d^2 + t^2 - 2 * d * t`

...and doing like that saves both `sqrt`... Or I'm wrong?

> I imagine that's also not done because it increases the visual complexity.

Let me say that (IMHO) comments are THE way

> I also wouldn't worry about the double overflow. The game visibly misbehaves, several orders of magnitude before distances of 2^512

I can confirm: that code is just skimming, comparing "space|sidereal" distances with
km, thus the approximation I initialli left...

> For people wondering what this algorithm is actually for: It's a quick workaround for O(n^2) ship vs ship sensor checks eating a lot of CPU in common cases. Essentially you can throw away a lot of checks early if the distance of objects from an arbitrary point isn't within the given range. The implementation is nowhere near optimal but IIRC it did make a substantial difference at the time. I don't remember why the sensor checks were expensive, however.

...I couldn't have put it better myself  :clap:

Last thing you should take care of when you shelve this whole thing, is the one I
spot during this useful conversation: you should address the issue on having the
third argument of `GetBodiesMaybeNear` so "naked".
What I mean is that any client of this function could prepare a `std::vector` without any
`reserve()` triggering memory reallocation several times...
You can check yourself with a 'grep', but for example in file `src/LuaPiGui.cpp`, in function
`l_pigui_get_targets_nearby`, `nearby` is provided without calling `reserve()` :disappointed_relieved: 

IMHO you should prepare that vector inside function (reserving... at least
something) and passing it back as a return value, like so:
`std::vector<Body *> GetBodiesMaybeNear(...)`

impaktor:
> Let me say that (IMHO) comments are THE way

@mike-f1 I haven't followed this PR, just wanted to say you're doing fine work, and if this PR doesn't get merged, then - and this is a general thing - please feel free to at least add comments to the code you've been looking at if there's anything profound or clarifying to say, so at least something good comes out of the time you spent trying to understand the code base.

Just my thoughts at least.

mike-f1:
> IMHO you should prepare that vector inside function (reserving... at least
something) and passing it back as a return value, like so:
`std::vector<Body *>  GetBodiesMaybeNear(...)'

 #4546

PR #4546, opened 2019-02-24
Avoid reallocation of vector using a list of nearby bodies inside Bod…

mike-f1:
…yNearFinder

Relevant changes are in `Space.h` and in `Space.cpp`,
remaining changes are to use the new prototypes
of `GetBodiesMaybeNear`

NOTE: Feel free to add comments & commits, because I feel
I'm not the best one at this...

### DIFF
--- a/src/LuaPiGui.cpp
+++ b/src/LuaPiGui.cpp
@@ -1212,13 +1212,12 @@ static int l_pigui_get_targets_nearby(lua_State *l)
 {
 	int range_max = LuaPull<double>(l, 1);
 	LuaTable result(l);
-	Space::BodyNearList nearby;
-	Pi::game->GetSpace()->GetBodiesMaybeNear(Pi::player, range_max, nearby);
+	Space::BodyNearList nearby = Pi::game->GetSpace()->GetBodiesMaybeNear(Pi::player, range_max);
 	int index = 1;
-	for (Space::BodyNearIterator i = nearby.begin(); i != nearby.end(); ++i) {
-		if ((*i) == Pi::player) continue;
-		if ((*i)->GetType() == Object::PROJECTILE) continue;
-		vector3d position = (*i)->GetPositionRelTo(Pi::player);
+	for (Body *body : nearby) {
+		if (body == Pi::player) continue;
+		if (body->GetType() == Object::PROJECTILE) continue;
+		vector3d position = body->GetPositionRelTo(Pi::player);
 		float distance = float(position.Length());
 		vector3d shipSpacePosition = position * Pi::player->GetOrient();
 		// convert to polar https://en.wikipedia.org/wiki/Spherical_coordinate_system
@@ -1235,7 +1234,7 @@ static int l_pigui_get_targets_nearby(lua_State *l)
 		vector3d aep(rho * sin(theta) / (2 * M_PI), -rho * cos(theta) / (2 * M_PI), 0);
 		LuaTable object(l);
 		object.Set("distance", distance);
-		object.Set("label", (*i)->GetLabel());
+		object.Set("label", body->GetLabel());
 
 		//		object.Set("type", EnumStrings::GetString("PhysicsObjectType", (*i)->GetType()));
 		//		object.Set("position", position);
@@ -1243,7 +1242,7 @@ static int l_pigui_get_targets_nearby(lua_State *l)
 		//		object.Set("polar_position", polarPosition);
 
 		object.Set("aep", aep);
-		object.Set("body", (*i));
+		object.Set("body", body);
 		result.Set(std::to_string(index++), object);
 		lua_pop(l, 1);
 	}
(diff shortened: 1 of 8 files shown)


mike-f1:
[Did I miss something?](https://stackoverflow.com/questions/5232198/about-vectors-growth)

[...Did I do something wrong?](https://i.stack.imgur.com/B8G9U.png)

fluffyfreak:
It's probably fine but not been reviewed as I've been out of the country for a week without a computer.

Reviews only happen when people have time to do so, that might mean only doing them every weekend (_7 days_), or even every other weekend (_14 days_) due to real life.

Please be patient ;)

jaj22:
There's still one allocation per GetBodiesMaybeNear call due to the vector copy return, right? As vector::clear doesn't actually reduce the capacity of a vector (C++11 weirdness), you could get a performance reduction for the UpdateAlertState usage. Return by reference plus lifespan comment should be fine.

Otherwise this seems like a general improvement.

mike-f1:
> There's still one allocation per GetBodiesMaybeNear call due to the vector copy return, right?

Actually return is by ref (see line 81 in `Space.h`: `std::vector<Body *> &`) so I can avoid moving copying and so on, but yes: the purpose of this PR was to reduce memory reallocations, then if doing it by `move` one allocation for each call (must?) be fine. So `move` could be retained as an option...
I colud add that actual drawbacks are that 

- compilers  (at least mine: gcc) don't stop with an error if someone push pop or either free the container received but only give warnings and that 

- that list behaves as it is a "`static`" so it is "freed" only in `~Space ` (...when hyperjumping or when game exits...)

I'm fine also with moving that list (explicitly: like with a `std::move` in funtion return) as it is safer... Let me know your opinion.
A note: I speak of `move` and not `copy`: it will perform better because move should 'steals' the
memory reserved by `m_nearBodies`, thus decreasing its capacity to 0 (...but I didn't any checks
this way...)

> As vector::clear doesn't actually reduce the capacity of a vector (C++11 weirdness)

I wanted to trade that reduction with less memory reallocations, because
I don't see cons in having that list only growing between `~Space` calls.

> you could get a performance reduction for the UpdateAlertState usage

I didn't get the reason, please: could you better explain your concern?

> Otherwise this seems like a general improvement.

:pray: 


**General OT**

> ~~Return by reference plus~~ lifespan comment should be fine.

Well, I hope you understand my thoughts: it is not I do not want
to comment code, it is that I don't think to be the one's that
may|should|could|must do that thing, expecially here where
comments could stay for long and there's a lot of stuff that needs
to be explained (...really a "lifespan comment" may suffice?).

Greetings

mike-f1:
> ...really a "lifespan comment" may suffice?

...Indeed: how MP will behave?
(...I think in that case we should pass to "move" or a processor
could have its ref changing while working on it...)

jaj22:
I didn't spot that you'd changed the typedef of BodyNearList. The functionality here is fine but I'm really not a fan of hiding a reference in a typedef.

I wouldn't worry about MP. I suspect this code is low on the list of threading candidates even before this optimisation.

mike-f1:
> I didn't spot that you'd changed the typedef of BodyNearList. The functionality here is fine but I'm really not a fan of hiding a reference in a typedef.

Done...

> I wouldn't worry about MP. I suspect this code is low on the list of threading candidates even before this optimisation.

...but it may worth a mention in a comment

> > you could get a performance reduction for the UpdateAlertState usage

> I didn't get the reason, please: could you better explain your concern?

? :)

jaj22:
UpdateAlertState was already using static storage (although per-ship, so a fat source of cache misses), hence if BodyNearList wasn't returned by reference then you'd be adding allocations with the change. With return-by-reference it should be equal or better.

Any static temporary storage is going to be an obstacle to MP, but then it reduces the need for it. In a game where IIRC about half the CPU time is used for drawing the UI (and most of the rest for planets), parallelising object updates seems vanishingly far down the performance priorities.

On the other hand, one allocation per GetBodiesMaybeNear call isn't likely to make an impression on the profiler either, so feel free to remove the static storage if you'd prefer. I'm ambivalent on the point as long as there's a reserve() in there.

mike-f1:
> UpdateAlertState was already using static storage (although per-ship, so a fat source of cache misses), hence if BodyNearList wasn't returned by reference then you'd be adding allocations with the change. With return-by-reference it should be equal or better.

I removed that caching as it seems unused.

> [snip] half the CPU time is used for drawing the UI (and most of the rest for planets) [snip]

Thank's for the hint, and thank's for your answers and patience.

fluffyfreak:
So, is this PR ready for review and merge?

mike-f1:
Yes

fluffyfreak:
I only ask because the work looks good to me already but if there's still some tweaks you want to make @mike-f1 then I'll hold off until they're done.

fluffyfreak:
Ok cool, i'll review in shortly, thanks for doing it!

mike-f1:
> Ok cool, i'll review in shortly, thanks for doing it!

Np, I hope only I used the right syntax (...as you shown in your #_4549,
a lot things may happens under the hood, even under the 'thin' hood
C++ provide...).

PR #4548, opened 2019-03-02
Speedup pilua

mike-f1:
While trying to wrap my head around how PiLuaGui works, finding how I
could speed up that code, I ended up founding `ui.pointOnClock`.

It is used 35 times in "data/pigui/game.lua", it is short but "math-heavy".

Thus I tried to move that funtion on C++ side, thinking that that change could
be a good starting point.

When I wrote that function in LuaPiGui, compiler told me there's no way it can
push/pull a `vector2` in Lua (commit eb44e86a80bf6e1af7f9e0b60c0680115ff65d91).
That thing amazed me but, ~~after deep thoughts~~ I simply copy-pasted some things
to provide these wrappers (remaining commits)... I compiled everything and... 

luacrash_15-17-22

Boom!

I finally got what's happened: there's no usage of either `vector2` or `vector3` classes
on Lua-side :fearful: : all the math it's on Lua (there's also a `Vector.lua` file but it don't
use any c++ facility)

...To me that seems like a "big-code-defect", but I would ask if this is know, if it is wanted
and if or how I could proceed...


impaktor:
Let's consult the experts:
ping @ecraven for pilua stuff, ping @laarmen for lua/c++/vector stuff...

laarmen:
My reply will be short: do we have actual profiling/performance numbers?

That's something that should have been done before any actual coding :-)

Web-eWorks:
@mike-f1 those functions (`math.cos`, `math.sin`, etc) are using the C standard library functions internally. There's probably little benefit to moving that kind of math C++-side other than reducing the Lua-to-C-to-Lua round-trip overhead, which is already fairly low. I'm happy to be proven wrong, but you'll need to do a profiling pass first.

The crash you're running into is due to the fact that the `LuaPull<>` function groups expect the pulled value to be tagged (via metatable) according to the class type you're trying to pull. A `Vector2` on the lua side is simply a table with a metatable pointing to the Vector2 object (probably defined in `libs/vector.lua` or similar), which causes it to be interpreted as a raw table instead of a `vector2d`.

mike-f1:
Apart from the fact that being asked about profiling when in a _Draft PR_ title I just wrote:

> Thus I tried to move that funtion on C++ side, thinking that that change could
be a good starting point.

may seems a little steep, but

Apart from the fact that moving a function that calls a lot of math from within a VM
entirely in C++ increase performance by itself (..as peoples wrote on Lua manuals,
or I'm wrong?) and may lead to further improvements made easily by compiler;

Apart from the fact that that function is sub-optimized as it is wrote (or
LVM is so smart to throw away "zeroed" variables...?);

Apart from the fact that I found that that function it is used widely (35 times)
so a little improvement here count as x35 (well, I felt really lucky when I first spot
that thing...)

I would know how you explain to yourselves why someone's feels there was a need
to provide a facility for `vector3` in Lua, apart from having a speed up or at least a
cleaner way to work with `vector3`... (Plz note that `LuaVector` is older than UI rewrite...)
I'm wrong or those Push/Pull method are faster than pushing every single variable?
(not considering every calculus that may be done in c++...)

So I opened this PR because I had doubts about on how to proceed, and I'm asking
again if you have some ideas because (obviously) my plan was to begin moving some
"internal function" to c++ and **then** do some profiling, if needed...
But that thing change everything; do you really think it is not an improvement if every
vectorial calculations is done in C++?

@Web-eWorks :
> reducing the Lua-to-C-to-Lua round-trip overhead, which is already fairly low

...so in that code what is slow, given function length and usage? I mean: given
how VM works, where is the bottleneck of that function?

> The crash you're [...snip...] as a raw table instead of a vector2d.

I would thank you for the explanation but, more or less I understand what happens, and
I know I may work around accepting that table in c++, but the question is if it's the right
way to handle these things?

@laarmen : I'm just been told that UI code absorb a big chunk of processor time...
May suffice as appetizer, isn't?

Let me know, this time I'm getting curious about your opinions|ideas|whatever...

Greetings

ecraven:
I have not measured things, however I'd guess that simple operations are faster in lua than moving them to C++.. for complicated calculations, it's probably worth it.
I think I added the vector3 support, not for efficiency but to make things simpler. it's much harder to keep pushing three values (and receive them on the C++ side). I haven't had time to look at the code, but it should be straightforward to make the code work for vector2 too (I just didn't need that).

mike-f1:
> I think I added the vector3 support,

I feel/fear you did the opposite: There are files in src/ called LuaVector* that do the same
as file in data/libs/Vector.lua, but instead of redefining "vector" they provide a facility
for math operations...

If I click on "Blame" those file seems made by @ johnbartholomew ...
May he give us some rationale? ...or even tell us the story behind?

NOTE: I checked again: `ui.pointOnClock` is called 35 times, **of which** ~5 are in loops...

fluffyfreak:
@mike-f1 and is that expensive?
Without profiling you don't know, lots of things get called lots of times that can look expensive but which the game will happily call for thousands of times without it even using 1% of the CPU time.

Whereas if you'd profiled, and by "profiled" I mean added profiling throughout the PiGui stuff, you'd see that `l_pigui_get_projected_bodies` actually takes 8% of the CPU time for a single call per-frame. That's madness and needs fixing.

You can look at code all you like, "fix" something, optimise the hell out of it and make it super quick, then not notice any difference to performance anywhere. I know **because I've done that a lot**.

If you're going to optimise, you start by profiling, then you improve the profiling, either by using multiple tools or by extending our own intrusive profiling call (_search for `PROFILE_SCOPED` in the code_), then you profile some more until you have identified exactly where performance is _really_ being lost.

Then you optimise, then you profile it again to get hard answers to tell you **if** you've improved it because optimisation will surprise you sometimes.

Maybe you've chanced upon a real performance culprit with `ui.pointOnClock`, but no-one can tell you whether it's worthwhile without more data.

mike-f1:
Before 'vector2' use
```
Lua PiGUI took 0.002607
Lua PiGUI took 0.002352
Lua PiGUI took 0.002348
Lua PiGUI took 0.002408
Lua PiGUI took 0.002444
Lua PiGUI took 0.002515
Lua PiGUI took 0.002382
Lua PiGUI took 0.002848
Lua PiGUI took 0.002570
Lua PiGUI took 0.002751
```

After 'vector2' use
```
Lua PiGUI took 0.002316
Lua PiGUI took 0.002270
Lua PiGUI took 0.002187
Lua PiGUI took 0.002226
Lua PiGUI took 0.002412
Lua PiGUI took 0.002336

```
but it is still incomplete and unstable, I didn't complete the thing.

I think now you get the point and may speak about that:
neither `vector2` or `vector3` were used through `LuaVector`
and all the math come from `Vector.lua`

To work with these classes in Lua, I `Register`ed both `LuaVector`
and `LuaVector2`

In Lua, due to a FTL (and boring) coding, I loose the capacity to perform
some operation. These operation are "sub-optimized".

For people wondering on how I got the numbers above,
I should remember them that a "cross-C-to-Lua-profiling" is
(AFAIK) not possible, so I did the simpler one I found over Internet
(see line changed at the bottom of `Pi.cpp`)...

But results aren't so good (or at least 'till now) as I expected, so it is
more a matter of consistency.

PS: If this is not good, then you should consider to remove unused file
like LuaVector, LuaMatrix and so on to recover compiling time... And to
avoid other people wondering why these files are there...

jaj22:
Wasn't sure that would help at all, although the difference isn't large. If my reasoning is correct, the bulk of the performance cost will be in creating vectors, as lua tables are a truly awful container for them. Unfortunately most of these vector creations are unavoidable and will have a similar cost from either C or lua.

Oddly enough, ui.PointOnClock is a good example of a function with unnecessary vector creations (3 where 1 would read fine). Whether this function is a significant part of the PiGui cost is another matter. Maybe try commenting out displayReticuleCompass first.

mike-f1:
> Wasn't sure that would help at all, although the difference  isn't large. If my reasoning is correct, the bulk of the performance  cost will be in creating vectors, as lua tables are a truly awful  container for them.

I agree with you, but then an almost not measurable performance gain should be there when you do some vector math directly in C.

> Unfortunately most of these vector creations are unavoidable and will have a similar cost from either C or lua.

I don't know... True is that
- adding some member (like `vector2.set()` ) should help decreasing number of `new`, thus encouraging re-use
- and the above means there's some space for an optimization (be that in C or Lua doesn't matters)
- code is more readable: you know exactly if that function needs a vector or not (there's doc about that thing?)

NOTE: I lose the ability to write:
```
p = vector2.new(0,0)
p.x = p.x + 10
```
thus I can't write (but only read) a single member... I don't know if there's a way of doing
this, it may help when reading code... Backup plan is to add some specializations of a `vector2.set()`

> Oddly enough,

:+1: 

> ui.PointOnClock is a good example of a function with unnecessary vector creations (3 where 1 would read fine). Whether this function is a significant part of the PiGui cost is another matter. 

...but now (with vector2/3 in Fast-C) you have your first brick and so...

> Maybe try commenting out displayReticuleCompass first.

...we can build our Wall :crossed_fingers:

mike-f1:
I worked another little on that and I obtain something stable enought to be played,
I hope to provide another commit within next days.

I would ask for your feedback because profiling tell that times are (as argued by @jaj22 )
more or less the same. So, before proceeding in following task and thus move
`ui.PointOnClock` & friends in C++ (starting with those who seems to be library
functions), I would know if you agreed that with `vector2` and `vector3` code is
more readable and may be better understood and ported in C++.

Let me know your opinion and thoughts.

fluffyfreak:
This looks like it could be useful when porting complex Lua functions too C++.

The one thing that I'm not sure of is the value in moving `ui.PointOnClock` to C++, for example it can be simplifed down to this:

`ui.pointOnClock = function(center, radius, hours)
	-- 0 hours is top, going rightwards, negative goes leftwards
	local a = math.fmod(hours / 12 * two_pi, two_pi)
	return Vector(center.x + (radius * math.sin(a)), center.y - (radius * math.cos(a)))
end`

or with your new syntax

`ui.pointOnClock = function(center, radius, hours)
	-- 0 hours is top, going rightwards, negative goes leftwards
	local a = math.fmod(hours / 12 * two_pi, two_pi)
	return vector2.new(center.x + (radius * math.sin(a)), center.y - (radius * math.cos(a)))
end`

Which is just an `fmod`, `sin` and `cos` plus a smattering of maths operators. 

The `vector2` and `vector3` code is more readable and **I think that might help**.

mike-f1:
> The one thing that I'm not sure of is the value in moving ui.PointOnClock to C++

`displayReticulePitch` (with `pitchline`), `displayReticuleHorizon` (with a "naked" version) and `displayReticuleCompass` are all using a "flavour" of `ui.addLine` in which the line is pointing
to a center, thus I would need a C++ version of that function in order to have an
`ui.addClockLine` (or `ui.pitchline`, I can't choose: which name do you prefer?) which do the
same but C++ side (I think I also would store a state in that function so I remove the
needs of pushing centers in every call)

> The vector2 and vector3 code is more readable and I think that might help.

:+1: 

> EDIT: why is the code formatting so bad!?!

mine => FTL coding doesn't not help :)
previous => ...I guess the same :)

Seriously: may I ask you to be more specific, so I will try to do things better...

impaktor:
@fluffyfreak I took the liberty to edit your last post.

fluffyfreak:
Thanks @impaktor

laarmen:
Would it be possible to have something a bit more explicit/documented on the new LuaTable constructor ? Right now one has to read the implementation (and the lua_createtable docs) to understand what the new arguments do.

Web-eWorks:
I'm probably a bit late to the party, but I would _strongly_ recommend that you use the syntax `Vector2()` instead of `vector2.new()`. Among other reasons, this follows existing syntax, prevents an extra C-function invocation (slightly circumvented by not using the LuaObject code), and is easier to see and reason about in a convoluted block of code.

Inserting it into the global namespace is a good idea, though.

mike-f1:
@laarmen : I will document that ctor before merge, np

@Web-eWorks : I don't understand if you ask for a different "syntax" or "way"...
I mean: syntax has been changed since vectors (2 or 3) in Lua are now using
`vector3.h` and `vector2.h` C++ code trought `LuaVector.cpp` and `LuaVector2.cpp`.
Thus these types are now `userdata` in Lua, because they are "registered" as types
in  `Pi.cpp` (see lines around 314) and every calculation is done in C++.

So what I don't understand is if your question is "Could you **restore** `Vector.lua`" or
if is "Could you **rename** `vector2` to `Vector2`...

In late case I don't know (yet) how I can drop "`new`" (it seems that if I remove
`new`, Lua interpret the vector as a table...).

In first case... well, probably I spent a lot of time doing something not a lot useful
although I don't see drawbacks avoiding C++ being using a Lua library (you could
check diffs of `LuaPiGui.cpp`: `Vector.lua` was "imported"...), in using old files
someone already made, and in having type checking in Lua...
(Yeah: in that case you're a bit late ;) )

> Among other reasons

? ...May I ask if you elaborate this answer?

src/LuaVector2.cpp
```diff
@@ -0,0 +1,256 @@
...
+};
+
+const char LuaVector2::LibName[] = "vector2";
+const char LuaVector2::TypeName[] = "vector2";
+
+void LuaVector2::Register(lua_State *L)
+{
+	LUA_DEBUG_START(L);
+
+	luaL_newlib(L, l_vector_lib);
```
Web-eWorks: To make `Vector2()` create a new table, you'll need this syntax after `luaL_newlib(...)`:
```c++
lua_newtable(L);
lua_pushcfunction(L, &l_vector_new);
lua_setfield(L, -2, "__call");
lua_setmetatable(L, -2);
```
You'll also need to rewrite your function to support passing the `Vector2` lib table as the first argument, as the syntax of the metamethod is `function __call(tab, ...)`.

mike-f1:
What I address is the use of libraries (Vector2 and Vector3) so anyone may
change if something should be in Lua or in C++; there's also a stronger "type checking"
in Lua and `ImVec2` usage is now limited to C++.

...All that thing is growing and growing and the time saved is not soo big.

I think it is coming the time of merge and wait for users feedback (lot of
CTD are expected :/ ). The remaining things could be addressed later.

Just some things to pin:

1) Vector2 uses `double` but are now used only for screen coordinates, is this worth?

2) `TScreenSpace` have `direction`, which is calculated but used only in `l_engine_world_space_to_screen_space` in order to display the target indicator inside the reticule: may be worth doing a separate function

3) `getProjectedBodiesGrouped` uses hash part of Lua tables, may be worth using the array part?

4) `getProjectedBodiesGrouped`, as it is wrote, forces Lua to iterate on each group to call `IsMoreImportantThan`, and this happens because (IIRC) hash part of Lua table don't preserve order.

Point 3 and 4 are the most important site for an improvement.

mike-f1:
Finally I get some time to address point 3 and 4 and to perform profiling and do some stats...


After adding the same lines I added in this PR at `Pi.cpp`, I compared execution times of this PR with the ones of master... And it seems I was wrong :smile: 

I used [this save](https://github.com/pioneerspacesim/pioneer/files/3019662/debug.zip) in order to have the same number of objects in Space and the same number of objects displayed both in this branch and in master.

I drop the first results because they were obtained during planet fractal calculations.

What I get is here:

[SpeedUp_PiGui.txt](https://github.com/pioneerspacesim/pioneer/files/3019635/SpeedUp_PiGui.txt)

So, the overall improvement is about ~30%:
LuaPi took ~0.0025 while in master it took ~0.0040 millisecond.

Then I commented some function call in `game.lua`, disabling
only some part of the UI, specifically each of these:
https://github.com/pioneerspacesim/pioneer/blob/3322ed4f8281934c905fd52557c6416c7832f188/data/pigui/game.lua#L967-L973

Considering 0.0025 and 0.0040 as update time, then commenting out
each function means that improvements may be isolated for each function.

So the change between commenting or not `callModules("game") ` is
4.0 - 2.7 = 1.3 millisec in master, while in this branch is
2.50 - 1.58 = 0.92 millisec

Thus the gain is about 30%... And I don't get how this can be achieved, considering
that that time reduction is on the _whole_ code and not only on calculations...
I think Lua is breaking something in pipeline and/or the push/pop of each component and
each operation is far more expensive than pushing the Vector2 and then the operation,
but these are just thoughts...

The gain due to new funtions `l_pigui_pointOnClock` and `l_pigui_lineOnClock` seems
the reason for the results in part "LuaPiGui without reticule", where the difference is
0.0040 - 0.0031 = 0.0009 (master) 
0.0025 - 0.0018 = 0.0007 (this branch)

Not so much then. In fact it is "on top" of vector changes, but I can't understand why 
is not any better (times here are small, so they may be not accurate).

The change due to the new `getProjectedBodiesGrouped` function is
4.0 - 2.6 = 1.4 millisec (master)
2.5 - 1.8 = 0.7 millisec (this branch)

Althogh it is "on top" of vector changes, it is mainly on C++ side then the
gain could be considered of 50%... Not bad, thought a different number of
displayed objects could potentially change this improvement because in C++
there are 5 "`for`" in that function.

Tip, thoughts and critics are welcome

NOTE: I didn't thank @Web-eWorks for he's suggestion, and others for theirs, so

Greetings

laarmen:
I haven't really looked at the code, but even without that I think I can
safely say "nice work!".

If I get around to it I'll probably have some stylistic nitpicks, so will
others. And we might be slow to merge. That doesn't mean the work is bad,
just that we are busy elsewhere 😉

jaj22:
I feel like I'm missing something here, but wouldn't it be straightforward to implement Vector2  as userdata rather than a table? Use some sort of pool allocator for fixed-size objects and everything should operate at the speed of a lua->C function call.

Serialization would probably be a nightmare, but nothing new there.

laarmen:
This sounds reasonable.

For those in the sidelines, userdata is an opaque blob of memory allocated
and managed by Lua but whose content is at the discretion of the C side of
things, the Lua side cannot do much except call methods on it. IIRC.

mike-f1:
> Never mind, missed a step. Current implementation should be good as long as the lua allocator doesn't suck for that allocation size.

...mmmh, may I ask if you can explain better? ...I don't get what you meant, or (worst) if there's a problem...

I'm also curious about what happens in `l_pigui_pointOnClock` and friends: why or how I can get any better?

And also in your thougts about how actual Vector2 are so fast, I expect an improvement but not so big: 30% is far from my understands...

Web-eWorks:
@mike-f1 regarding Vector2, C++ operations are subject to innumerable compiler optimizations ranging from loop-unrolling to vectorization of data (using a single instruction to perform 4 or 8 simultaneous floating-point operations). Additionally, I believe the FPUs of most modern processors support native operations like vector length, etc. which the compiler can emit instructions for instead of having to perform each step of the operation as its own, slower, instruction.

Lua, on the other hand, must perform stack-juggling for each and every operation that it performs - a method like so:

```lua
function Vector2:Length()
    return math.sqrt(self.x^2 + self.y^2)
end
```

Will compile to something like this (note: not actual Lua opcodes nor a 100% accurate representation of the Lua stack or how literals are used):

```
GETGLOBAL 'math'
GETFIELD 'sqrt' -1
GETFIELD 'x' $arg1
PUSHLITERAL 2
CALLOP POW
GETFIELD 'y' $arg1
PUSHLITERAL 2
CALLOP POW
CALLOP ADD
CALLFN -2 1
```

Each of these pseudo-opcodes are read by Lua, upon which the VM does some stack fiddling, table lookups, or calls some underlying math. In all, it's maybe 100-200 CPU instructions and at least one probable cache stall (from looking up values in tables), costing around 500 CPU cycles for each invocation. (Remember, a cache stall is a _minimum_ 50-cycle delay, and very few instructions on x86 are single-cycle.)

For the userdata version of Vector2, it's a bit simpler. There's a single metatable lookup (for 'length'), a C-function call (fairly simple, but does need to create a stack frame), a pointer-indirection in the C code (calling `v->length`), and a few cycles for function prolog and the actual vector operation. The big saving here is that instead of using the Lua operators for addition and exponentiation (which are quite a few cycles each due to the need to test for and invoke metamethods), you're invoking the CPU's facilities for addition and exponentiation directly.

jaj22:
My remaining concern with the userdata Vector2 is that an apparently-trivial lua expression like vec1+vec2 is still likely to be calling a malloc, as it creates a new Vector2. In C that would be considered a gross performance bottleneck, although compared to the other lua overheads it may not stand out. It's certainly preferable to creating a table.

Note that the Vector2.Length example doesn't actually create a Vector2 at any point, and so may be much faster than a vec1+vec2. Would be interesting to compare in practice.

Web-eWorks:
We can evaluate the performance benefit of using a custom lua allocator function. If we maintain a block pool for objects with common sizes (eg 16 for vector2d, 24 for vector3d) and choose between allocation from one of those pools or a malloc (for odd sized objects) we might mitigate the performance cost even further.

jaj22:
@Web-eWorks I'm guessing that a vector2d would actually allocate something like 16 bytes + typeinfo + metatable reference?

I didn't figure that lua had a function you could hook for a custom allocator. Looks like it'd be easy enough to hack in an allocator with a single exception for vector2d, whatever size they are.

mike-f1:
I would postpone these further improvements for Vector2 and Vector3 in a different PR:
although interesting I feel a little tired of this and I prefer/suggest to let these changes be
tested at least some weeks/months before any additional change...

I would also suggest that a really important speed up may be achieved "`local`izing"
functions instead of only variables (...I would remember you that a local is 4 times
faster than a global in Lua)...

But apart from the above, I'm puzzling about two questions:
1) `lineOnClock`&friends aren't fast as I expect, but I cannot explain why... May I miss something on C-side?
2) LuaTable new ctor: it seems it allocate only Lua datatypes, which are stored in an union. Thus Lua should then allocate also the c-data part which is allocated for each entry...
As a result it not working so fast :/

mike-f1:
About

> still likely to be calling a malloc, as it creates a new Vector2.

and 
> We can evaluate the performance benefit of using a custom lua allocator function

I noted, skimming trough code, that `LuaObject` pushes light user data because it use `lua_pushlightuserdata`, you can check this line:
https://github.com/pioneerspacesim/pioneer/blob/19ee935b233b9f6c0066f08f9ed4f46f9add1761/src/LuaObject.cpp#L721

 which is in function `void LuaObjectBase::Register(LuaObjectBase *lo)`, which is in turn called
by `void LuaObject<T>::PushToLua(DeleteEmitter *o)`...And so it is used for everything

So the exception, AFAIU, is LuaVector2 and LuaVector which are using Lua-managed standard
userdata, calling

https://github.com/pioneerspacesim/pioneer/blob/19ee935b233b9f6c0066f08f9ed4f46f9add1761/src/LuaVector.cpp#L203

in function `vector3d *LuaVector::PushNewToLua(lua_State *L)`

mike-f1:
> If I get around to it I'll probably have some stylistic nitpicks, so will others.

So, are these "stylistic nitpicks" coming or there's something else that is wrong here?

impaktor:
>> If I get around to it I'll probably have some stylistic nitpicks, so will others.

> So, are these "stylistic nitpicks" coming or there's something else that is wrong here?

Here's looking at @laarmen 

Obviously, the branch also has merge conflicts, and messy commit history that could use a lot of squash-ing, using interactive rebase.

laarmen:
Well, yeah, the messy history is one point. And it's much easier to do code review with a sane git history : here, we have a huge diff, which makes reviewing the entire patchset at once difficult, but I don't fancy reading 30+ commits, especially if the later commits undo some changes in the first commits.

Also, to be clear : I didn't commit myself to do any sort of review, hence the "if I get around to it", and said review would be relatively superficial ("stylistic nitpicks").

fluffyfreak:
When time comes this can just be Squashed and Merged, so commit history isn't a big deal for me.

I'll try it out but @mike-f1 if you could rebase/merge to the current master to fix the conflict then that'd be good and I'll merge it in asap after testing.

impaktor:
Also, I didn't follow if the whole discussion on how to best implement/make use of vector2 got resolved, so long as those who objected are happy, or hold their peace...

fluffyfreak:
@impaktor I think it was decided that if it gets done, then it gets done as another PR in the future before @mike-f1 gets burnt out doing this PR.

mike-f1:
> Also, to be clear : I didn't commit myself to do any sort of review, hence the "if I get around to it", and said review would be relatively superficial ("stylistic nitpicks").

...Well, I'm sorry: I would have liked someone (...like you...) reading the code before but it seems you are really looking elsewhere. Not a problem though

> Obviously, the branch also has merge conflicts, and messy commit history that could use a lot of squash-ing, using interactive rebase.

I will solve conflicts, and I would clean commit history because I think it is better to have at least some sort of "incremental changes" stored. But I don't get why this commit history is so messy, even if

> especially if the later commits undo some changes in the first commits

...Aren't commits there to store the history of changes? Thus later commits are undoing firsts because a better understanding come in: vectors and their calculations are ever faster in C++, and this was unclear at the beginning...

Anyway, are there some criteria I should apply? How I can choose/identify commits to be squashed?

NOTE: I would pin here major changes (something that IMO should be retained in history):
1) LuaVector forward and use instead Vector in Lua
2) ImVec2 replaced by LuaVector
3) All the thing of ui.pointOnClock
4) "Hierarchical Agglomerative Clustering" (aka `getProjectedBodiesGrouped`) in C++

> @impaktor I think it was decided that if it gets done, then it gets done as another PR in the future before @mike-f1 gets burnt out doing this PR.

Exactly: this PR, IMO, is big enough... But some hint for future PR are:
- Caching functions in Lua
- Use of a pool for Vector2 and Vector in C++ to avoid malloc-ing them (or use lightuserdata instead of userdata
- Return separate arrays in `getProjectedBodiesGrouped` so element allocation is done ahead of time

impaktor:
> and I would clean commit history because I think it is better to have at least some sort of "incremental changes" stored. 

By all means. As long as it's not "this is what I did on thursday", "This is what I did until next episode of Game of Thrones was released" kind of commits. (I've worked with people that just made a single commit each day when they went home from work. That's sub-optimal).

> But I don't get why this commit history is so messy, even if the later commits undo some changes in the first commits

Yes, during review keeping the new modifications separate can be of value: e.g. consider a PR with three commits:
``` 
41119f9 A
bc327ba B
1383267 C
```
reviewer reads those, then complains the indentation is wrong in all three, and a thing should be changed in commit B. then you can push new commits
``` 
b24fa4d D - fix indentation
d63501d E - fix thingy in B
```
Now reviewer can easily see the isolated changes, and don't need to re-read all commit history to re-review (e.g. due to changed indentation). To be honest, it's rare we review someone that thoroughly that this is relevant, maybe if it's some part of the code where @laarmen is invested and he has time to review. Most of the time we rely on self-review, i.e. write the code at the top of ones abilities from the start.

> ...Aren't commits there to store the history of changes? 

Yes, but that would only be interesting if/when under an _ongoing_ review. In the master-branch, we really don't want to see all the jumps and hoops that it took to develop the PR as that is just ugly, bloaty, and confusing for the next person, so before merging it should be "clean", i.e. no: commits of the type:
``` 
ca09b8e Now volume goes to 11
3bdb429 Fix bug in volume control
7b26338 negative volume doth not make sense!
```
in this case last two should be marked "fixup" in interactive rebase (or sqashed if you want to change commit message at the same time), so we only get:
``` 
fe87a71 Now volume goes to 11
```
So in master we see the "perfect" code for the new feature.

> Anyway, are there some criteria I should apply? How I can choose/identify commits to be squashed?

Yes, preferably code that is re-doing/changing previous commit in the same PR should be fixup/squashed.

A commit should preferably be for each feature, if that helps reading it, especially when something revolves changes in multiple files, e.g. in .cpp and .h files. 

I'm also fine with squashing all to one commit and calling it "optimizations" or similar, but I suspect there are some things that logically belong as a commit.

At least this is how I think about it, but (my) git commit fascism shouldn't ruin the joy of contributing to pioneer, but at least having this in the back of ones mind helps, I think.

> this PR, IMO, is big enough... 

Yes.

src/Pi.cpp
```diff
@@ -1615,9 +1619,35 @@ void Pi::SetMouseGrab(bool on)
 void Pi::DrawPiGui(double delta, std::string handler)
 {
 	PROFILE_SCOPED()
+	bool debug = false;
```
fluffyfreak: This debug/profiling code needs removing

src/LuaPiGui.cpp
```diff
@@ -1275,31 +1365,215 @@ TScreenSpace lua_world_space_to_screen_space(const Body *body)
...
+		// First element should always be the "media";
+		// so push twice this element:
+		//printf("Body 1 = %s (%f, %f)\n", (*it)._body->GetLabel().c_str(), (*it)._screenPosition.x, (*it)._screenPosition.y);
+		group.push_back((*it));
+		// Zeroed body so you could distinguish between a "media" element and a "real" element
+		group.back()._body = nullptr;
+		group.push_back((*it));
+
+		// Find near displayed bodies in remaining list of bodies
+		std::vector<TScreenSpace>::reverse_iterator current = std::reverse_iterator<std::vector<TScreenSpace>::iterator>(it + 1);
```
fluffyfreak: I get a crash here, which I'm just investigating. 
When playing (Mars start) I take off and start flying straight ahead, then I look around with the middle mouse button

fluffyfreak:
@mike-f1 That crash is stopping me getting any further with testing.
It seem the reverse iterator tries to read beyond the beginning of the vector.

src/Pi.cpp
mike-f1: I did it that way to get it compiled: reason is that `#ifdef` clause don't check code
thus after years you have code which may not compile...

...I also would let that code "in" so next time an optimization is needed then
I (or some other) will get the same numbers.

@impaktor : that code, if I follow the rules you explained , must be fixed in
the first commit or I'm wrong?

src/LuaPiGui.cpp
mike-f1: It seems your S.O. is more susceptible to memory leaks, so I hope
you will do another test-fly...

mike-f1:
> It seem the reverse iterator tries to read beyond the beginning of the vector.

`it` still a `forward_iterator` so, in reality, I'm trying to read from past the end;
Thus I let try to group the last element with... Nothing :smiley:

mike-f1:
@impaktor : is the history ok for you or do you think there's some "mandatory" to do before merge?

impaktor:
Maybe dumb question, but:

- why is LuaMatrix.cpp/h renamed to have trailing "__"? If these files aren't used, should they not be removed? (I'm not saying you should, just like to know the reason for keeping them, if we don't want LuaMatrix to be compiled)

> @impaktor : is the history ok for you or do you think there's some "mandatory" to do before merge?

Nah, at a quick glance, I don't see anything to exert my git-fascism on, this PR has been open long enough.

mike-f1:
> why is LuaMatrix.cpp/h renamed to have trailing "__"? If these files aren't used, should they not be removed? (I'm not saying you should, just like to know the reason for keeping them, if we don't want LuaMatrix to be compiled)

Good point. My 2 cents are:
all this PR originate because nobody removed `LuaVector.h|cpp`, so I took these file and
I "only" used them... I'm not prone to remove code: IMO they are like a book not (yet) read.

Given that a code file more is not a big space waste, the remaining problem is to save compiling
time (which is a small save, but not for hundreds of thousands compilations...), thus the renaming.

impaktor:
I don't think we've ever done this (renaming files to keep unused code around) before, so thus it feels a bit weird to me, I'd love to hear what e.g. @laarmen thinks of the "LuaMatrix.cpp__"-rename?

Web-eWorks:
We specifically use Git to have version control history. If the files are no-longer used by anything, delete them and do not keep them around. If by some chance you want them back later to use for something else, you can always check them out from a previous commit.

The reasoning behind this is twofold: it requires more time and effort to mentally segregate `LuaMatrix.h__` from actual useful files, and you are effectively creating garbage files for no other reason than a duplication of functionality that already exists in the version control tool.

laarmen:
What Webster said.

Le lun. 29 avr. 2019 à 05:50, Webster Sheets <notifications@github.com> a
écrit :

src/Pi.cpp
fluffyfreak: Is there a reason that you didn't just use the builtin profiling which is already there? Meaning the call to `PROFILE_SCOPED()` above your own profiling code?

src/LuaPiGui.cpp
fluffyfreak: I've done a lot of test flights, with latest code, and I still get the problem.
It's a 100% repro' rate when using VS2017 x64|PreRelease (_which has Debugging & light optimisations_)

fluffyfreak: It runs in Release|x64 but this means no-one will be able to run debug/prerelease builds using VS201X. It's definitely not correct.

src/Pi.cpp
mike-f1: In general, is easier to collect times because it print some message and you only copy-paste these ones...

In particular, I follow the instructions you gave in your previous PR but I can't get anything.

But you're right: I will remove this code by removing the related commits.

src/LuaPiGui.cpp
mike-f1: Well, I will investigate this issue, np.

I wonder if there's a way to get the same behaviour with with gcc...

fluffyfreak: Great, thankyou

mike-f1: @fluffyfreak , I'm wondering if you have time and will to collect some feedback on this issue:
may I ask you if you could uncomment the "printf" lines on this function and send me the output?

jaj22: I think the problem is that pop_back() technically invalidates any iterators pointing to the last element (or the end), and it2 can potentially be pointing to the final element when it's popped. Real vector implementations probably wouldn't invalidate the iterator, but MSVC debug will complain.

I couldn't get this working without switching to forward iterators in addition to the obvious solution of moving the increment inside the loop. I'm guessing that reverse iterators refer to the previous element internally, and so get invalidated if they're one-off-the-back too.

mike-f1: I was thinking more or less the same, but blaming the `std::swap` which is moving/copying a previously moved (thus invalidated) iterator.

Anyway I pushed a "debug" version with (useful?) messages and a not "swap-and-pop" workaround.

Another solution, if the problem is on `pop_back()`, is [erase](http://www.cplusplus.com/reference/vector/vector/erase/) which return a valid iterator.

jaj22: I actually removed the std::move on sight, and forgot about it. It still needed both of the other modifications (forward iterator, decrement it2 before pop), so I don't know if the std::move is technically a problem or not.

Erase wouldn't necessarily return the correct iterator for the algorithm. it2 may or may not point to the last element.

Your copy-instead-of-swap code has the same iterator invalidation crash, with or without the std::move.

mike-f1: Sorry for the late and thank for your support, I'm a little under pressure these days...

I would try a version in which `it2` increment is inside the loop (as per your suggestion) and the order of things is:
1) copy on `it2` (`(*it2) = filtered.back();`)
2) increment `it2` (which is "point `it2 `to previous element") (`it2++`)
3) `pop_back()`

This way, If I understand the problem correctly, `it2` will never point past-the-end and thus will never be invalidated... And the problem should be solved...

But I would add that the previous version (the one with forward iterator) was so unnatural and so full of checks and workaround that I prefer to propose a different route:
if the problem is as outlined above, to me a possible solution/workaround is to disable
those smart-iterator around the blamed code as explained [here](https://docs.microsoft.com/en-us/cpp/standard-library/debug-iterator-support?view=vs-2019).

Note we are working around not for a "bug" but for something less...

mike-f1: Uploaded but not tested (I used some time at work), it should work anyway.

NOTE:

> forward iterator, decrement it2 before pop

equals to "increment reverse it before pop": So I did what you suggest, though not exactly

jaj22: Crashes. I would guess that the problem is that reverse iterators refer to the next element internally, and so the pre-incremented reverse iterator is still invalidated. I don't understand the problem with the implementation of the forward iterator version. As far as I can tell it's just one line shorter. Code for reference:
https://gist.github.com/jaj22/c89d213ec44c242f54c8016538b7b1fd

As far as I can tell, MSVC is correctly interpreting the C++ spec by throwing the errors. It's more that the C++ spec isn't well aligned with how vectors are implemented, so you'd never get a real crash from a wide variety of technically incorrect code.

I also get an unrelated crash in l_pigui_radial_menu, by opening the radial menu. It's a lua_istable assertion caused by the line `LuaTable xuv0(l, -1);`. I'm guessing that this was overlooked in the ImVec2 replacement.

mike-f1: > Crashes.
😤

> I don't understand the problem with the implementation of the forward iterator version

I thought "reverse iterators" were designed (and they exist) to _solve_ this kind of situation :/ ... And for readability: I feel it was easier to get the idea when you saw a "reverse iterator" going "normally" instead of a forward iterator going backward (though in C this happens often), but if there are so many "nitpick" then it is definitely not worth the effort.

Well, enough is enough and you write that code for me: thank you. If you didn't find the remaining error (`LuaTable xuv0(l, -1);`) I had already pushed everything: I will push if not tomorrow this weekend when I hopefully will solve that issue (...there aren't iterator around that line ;) ).

mike-f1:
> Exactly: this PR, IMO, is big enough... But some hint for future PR are:
.....

-  use a `LuaColor.*` as a bridge between C++ and Lua (more stylistic than for performance,
though the benefit may worth coding).

- check `radar.lua` so it use `lineOnClock` & push/pop center and some data so calculous are entirely in C++ (and, as a side effect, split objects handling from background graphics)

...Done? (:crossed_fingers: )

fluffyfreak:
@mike-f1have we added you to the 'AUTHORS.txt' file yet? Would you like your name in there?

mike-f1:
@fluffyfreak  Yes, thx's

PR #4549, opened 2019-03-02
PiGui profiling changes.

fluffyfreak:
In testing using `PIONEER_PROFILER` defined and our intrusive profiler (_V-sync off, multiple runs, ctrl+shift+p ingame once terrain generated_) this reduces the load of `l_pigui_get_projected_bodies` from ~8% down to ~1% of CPU time.

@ecraven has suggested there are other ways of doing this so this PR might only be useful as reference.

### DIFF
--- a/src/LuaEngine.cpp
+++ b/src/LuaEngine.cpp
@@ -821,11 +821,11 @@ static int l_engine_world_space_to_screen_space(lua_State *l)
 {
 	vector3d pos = LuaPull<vector3d>(l, 1);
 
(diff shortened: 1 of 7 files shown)


jaj22:
I have trouble believing that tuples are that slow. Are you sure this was a fair test, ie. same object count?

fluffyfreak:
@jaj22 yup, same object count, 3 runs each. The difference was small enough that I wouldn't normally bother, but since I also like clearly named structs I made the change.

jaj22:
Ah, I was looking at the wrong commit. Makes sense now. General rule would be that creating lua tables and adding keys to them is extremely expensive.

PR #4550, opened 2019-03-03
Terrain Cleanup

fluffyfreak:
Some Terrain changes and tidying up:
- Remove fractal and terrain "texture" options - fix them as always on and average
- Change terrain patch geometry to have low/med/high options only, and to be (power-of-2)+1 sizes
- Move generation methods into request classes
- Remove all old GL2 shaders since they're defunct
- Detail mapping is mandatory!
- Add standardised BiCubicInterpolation method for sampling from heightmaps
- Cache the inverse of m_planetRadius and use it for mul instead of div as minor optimisation

There's a lot of changes in here but they all follow the same theme of removing things and cleaning up a little bit.

The goal is to simplify terrain stuff a bit and make it easier to work on in the future.


impaktor:
I suspect only you can review this, so merge when you feel ready.

laarmen:
I'm curious, what's the rationale behind the (2^n)+1 ? I'm more used to having a -1 ;-)

jaj22:
@laarmen (2^n)+1 is natural for quadtree terrain vertex counts. For example, a patch with 8x8 quads will have 9 vertices per side. It was always annoying for storage.

PR #4554, opened 2019-03-05
Planet texturing

fluffyfreak:
NB: This is hilariosuly WIP but there's reasons for making it a draft PR now.

Add texturing to planets based on a Look-Up-Table (_aka LUT_) and Texture Arrays (_aka Atlas_)

Here are two shots of Mars with the current code for example:
snap
screenshot-20190305-222720

Work still to do:
- [x] Replace hardcoded textures with per-planet-type LUT and unique textures
- [x] Replace the Gas Giant texture generation with shader ported code
- [x] Remove all of the `GetColor` system from `Terrain` classes (_maybe re-use colourisation if viable_)
- [x] See if Gas Giant `HueShift` from `gen-gas-giant-colour.frag` shader can be used to add variety
- [ ] Split Star generation/rendering into it's own thing just like I did for Gas Giants
- [ ] Get textures, cc-by-sa / created / etc, to replace the test ones I "found"
- [ ] Document the height sampling madness a lot
- [ ] Look into using Git LFS for storing the textures
- [ ] Make this a longer list...

Questions, suggestions, thrashing and screaming all welcome.

The textures used for this test so anyone can have a go.
Contained in the ZIP file is a folder with textures, extract it into `/data/textures/` so that you have `/data/textures/terrain/`
[terrain.zip](https://github.com/pioneerspacesim/pioneer/files/2935227/terrain.zip)

\<impaktor\> So how many textures would one need to create an infinite universe?
\<impaktor\> Something there I don't understand, unless textures are randomly generated.... which I guess counteracts the idea of using a texture.
\<fluffyfreak\> a finite number of textures, but with a variety of combinations, plus there's the possibility of some hue shifting at runtime
\<fluffyfreak\> basically you've got a pool of textures... lets say 128 of them in total which is about 80MiB of textures given the sizes etc used so far
\<fluffyfreak\> you pick upto 16 for a single planet
\<fluffyfreak\> and those are chosen/blended in the terrain shader using a LUT table
\<impaktor\> OK.
\<fluffyfreak\> we may want to do some stuff like use the Git LFS stuff to manage these textures


Web-eWorks:
Oh hey, it actually looks like a planet and not a giant ball of vomit. Imagine that.

In other news, I could not be happier that this is finally getting done. For suggestions, I'd throw in distance-dependent detail texturing, e.g. having a high-res texture blend in when you're close to the ground. That'd reduce the tiling-pattern effect of having a higher-res texture at long distances, and mitigate the previous problem of textures being 1:1 pixels to meters.

nozmajner:
Neat! 
Could you shoot the textures towards me? I want to see, how complex they are, and want to take a shot on them. I'd prefer having custom made in the end.

fluffyfreak:
@nozmajner I've added a zip of the textures, they're currently horrible :)

They're all 1024x1024 and DDS DXT1 compressed, I think that there might need to be some other changes in the future such as adding specular & normal map support but that starts to get very texture heavy so for now they're just diffuse. One option might be to bake the specular data into the alpha channel, that will need some thought and possibly custom art process or code tooling.

There are 16 textures + 1 LUT, that plus the two detail texture maps that were already there bring the total number of texture channels used to 19 textures total.

It doesn't have to be that many, we could have just two or three for a planet (_barren, airless... like the Moon for example_) but for now it's 16 total for a planet.

The LUT is made of 16 shades, starting at black through to 255. 

NB: I need to double check this next part but I think that it's correct ;)
The origin is in the TOP LEFT.
Height runs vertically (_Y axis_), so lowest at the top of theimage, max height at the bottom. 
The horizontal (_X axis_) is for the slope, so flat is on the left, vertical surfaces are on the right. Very few surfaces actually get to vertical ingame it seems.

fluffyfreak:
Well that took a LOT longer than I'd hoped. 
Ported over the shader gas giant stuff to CPU, removed all of the crazy templated `GetColor` implementations, discovered some stuff on the way through so random other changes happened.

There's still a bug when building Saturns texture that I cannot claim to even begin to understand yet but I'll figure it out eventually.

impaktor:
Nice to see you back!

Let us know when you need testing

fluffyfreak:
Having used LFS at work on 2 projects now I think I can safely say... I'd really rather use a submodule or another repository instead. LFS works but we've had issues with it.

Web-eWorks:
Haven't tested the gas giants yet (and I just realized I haven't tested the new textures), but planet texturing appears to be improved. I'm noticing a few issues:

Screenshot from 2019-05-18 10-19-29

You can see in the upper right of this screenshot that there's noticable pop-in where two patches have different UV scale instead of smoothly blending - this might be related to the second issue, but ideally we want a consistent UV scaling that blends in the 'detail' UV scale smoothly over distance instead of a binary switch per-patch.

Screenshot from 2019-05-18 10-20-53

You can clearly see in this screenshot that it appears that we're doing vertex-based coloring / texturing, which yields incredibly unrealistic results and may also be the root cause behind the UV pop-in.
EDIT: haven't tried the new textures on Earth, this _may_ fix things.

fluffyfreak:
@Web-eWorks that's the existing terrain, it's vertex based colouring, there's already a couple of detail textures.

What does this have to do with this PR?

Web-eWorks:
>What does this have to do with this PR?

I pulled the PR and tested it (albeit accidentally without the new texture files) and wanted to point out the issues I saw under what I thought was the new texturing method... On taking a second look, I apparently forgot to change branches before compiling. Please disregard.

Web-eWorks:
Something you might want to look into (assuming you haven't already) is [this post](http://johnwhigham.blogspot.com/2015/03/derivative-mapping.html) on derivative-based surface normals - from the looks of it, it could help improve the quality of terrain texturing even further than this PR already does.

fluffyfreak:
@Web-eWorks 

> <sturnclaw> fluffyfreak: ...you do realize glsl's mix() function blends two textures together without affecting the resulting luminance, right? trying to finish the terrain texturing PR and the color = (s1 * s2) * 2.0; bit threw me for a loop there just use mix(s1, s2, 0.5) and it doesn't darken the resulting texture at all"

yes, I wasn't expecting anyone to try and contribute to the branch so things aren't as well documented as they should be!
mix is an option but I was porting the code from elsewhere
originally it was from https://forum.unity.com/threads/improved-terrain-texture-tiling.116509/
but it might also be the cause of the visible texture seams when close to the surface

> <sturnclaw1> doesn't help that the default LUT doesn't use texture slot 0 for some things

it's all work in progress, the LUT is supposed to be one of a selection of LUTs, and this one isn't very good yet, just for testing

> <sturnclaw1> There are a couple of low-effort things I want to improve about planet texturing
> First of all, it should be absolutely possible to define e.g. only 4 textures for a planet instead of needing to fill out the giant 16-texture JSON file
>  pass the number of textures in the array as a uniform or use GLSL's textureSize(texArray).z to figure out the multiplier

yes, it's supposed to be variable eventually

> <sturnclaw1> Secondly, if I'm right in thinking that our LUT starts at top-left for flat sea-level, I'd recommend changing that around

probably should, just used it as-is to get it working

> <sturnclaw1> we should also do some form of multi-tap sampling and blend different textures together
> not sure the best way to do that, but transitions between textures should absolutely be much softer than they are new
> because LUTs have discontiguous values (e.g. you could tap twice and get 1 and 15) it's probably better to blend the resulting texture values, but you're looking at triple the texture bandwidth in some cases (especially with displacement maps et. al.)

splatting would be ideal I think, use N textures with weighting and blend between them, not figured out _how_ yet

> <sturnclaw1> for the record, our current shader code is an absolute mess and I really really really really really want to facedesk every time I read it
> don't statically #if on whether or not textures are bound to the material, please!
> If it's a material that needs textures, bind our unknown.png to samplers which don't have assigned textures
> if it doesn't need textures, don't provide slots for them in the first place
> but above all, don't complicate the shader any further!

this is just dealing with how it works currently so as not to break things, hopefully it gets simpler because there's only one code path after this is done and it will always use a texture, but development is messy

> <sturnclaw1> We need some way of realistically varying the coloration of planets. At the moment, I understand that's what the different color fractal jsons are for, and the idea behind setting up alternate textures based on metallicity, etc.

No Man's Sky did something interesting with this, they defined a palette for each world, the textures were palettised and the palette was simply replaced with this worlds new palette!

> <sturnclaw1> I stalled out on loading array sub-textures from PNG last time, though it was more because TextureBuilder is kind of internally an abomination

There's other formats which support texture arrays but fuck trying to use anything standard from DDS, and it has to be pre-generated DDS due to Intel DDS generation issues on some linux versions with some GPUs

> <sturnclaw1> When close to the surface of the planet there are visible texture seams; it appears there's an issue with uv scale of our "close" texture sampling

yeah, it's got bugs, lots of bugs

> <sturnclaw1> regarding where to source textures, cc0textures.com looks like a wonderful resource; they're the GameTextures of the open-source world

Any source that suits our needs is good by me, I'm just using some textures from... actually I don't even remember but they might not be open source!

> <sturnclaw1> re: git LFS, I'm glad to see you've seen the light with it; I don't think it's a good fit for pioneer either

ugh, the project I'm on that uses it drives me insane!

> <sturnclaw1> fluffyfreak: planet texturing height sampling is based on sea-level reference height, right? 
> How are you encoding that into 0..1 uv space? What's the scale / metric?

horribly, if you look in GeoSphere::SetUpMaterials you'll see at the bottom a height sampling loop. 
It goes through and finds approximate min/max heights for the world, then in GeoPatch::UpdateVBOs this is used to normalise(ish) the height and puts it into the second uv slot of the vertex

Web-eWorks:
@fluffyfreak: I've left quite a few links to implementation papers in IRC, most of which are from Frostbite. I've taken another stab at UV scale, and I'm currently doing debug visualizations to track down a few bugs.


[Here](https://imgur.com/a/oDKcycz) are the mentioned debug visualizations, with a bit of commentary for scale.

The-EG:
I'm happy to see this, the stations always seemed good to me but landing on a planet always felt wrong. 

This makes me think back to MS Flight Sim 5 / 95; even with the textures turned off (a 386 with 4MB couldn't really handle them!), you still felt like you were flying over different terrain. A city (light grayish) stretched out around you with roads (dark gray). And you could tell when the terrain changed due to changes (albeit abrupt) in color: dark green (wooded), light green (grassy plains), brown (barren, sandy), white (snow), etc. Textures added to that effect: cities had rooftops and roads, wooded areas (sort of) looked like tree canopies, etc.

I get the impression that Cydonia is supposed to be a large (maybe even huge?) settlement. But, looking at it in game I feel like I'm only seeing the downtown/industrial area where the spaceport is, and that there must be another 80% of the city somewhere:
image
I think simple texturing could go a long, long way there. I think I'd rather see more texturing than modeled buildings!

Another thought that just occurred to me: if you do end up with textures that are to be areas around cities/populated areas, can they have areas pixels that are lit (maybe not specifically give off light, but are visible as the same color and shade regardless of lighting)? I would imagine that would be a layer/map on top of the texture.
Another thing that's always bothered me is flying near a populated planet (ie Earth) and it's completely dark...and landing in a large city and there's only a few lights coming from the spaceport and modeled buildings.

The-EG:
I played with this some this evening....it's cool!

It looks like the LUT isn't actually getting loaded and it's using unknown.png instead (which makes for an interesting LUT). I had to change the LUT entry in the terrain JSON to the path of the png file including the extension to get it to work (ie. textures/terrain/terrainLUT.png).

Some simple textures I came up with playing around:
image

I haven't figured out the LUT enough to get mountains/snow showing properly yet, but it's exciting as is.

Some things I wasn't expecting:
I had to create a water texture and make sure that 0,0 in the LUT was pointing to it.
The Y axis on the LUT isn't really intuative. The x axis seems to work like I'd think, but the scale for y seems weird or insanely huge. What is the max height? 

I also had to increase the LUT size from 256x256 to 512x512 to get something workable at the coast. Otherwise a single pixel for 'beach' would run from the water almost up to L.A. This is at 512 with a single pixel which is much more reasonable, I think (near L.A. on Earth):
image

I'm not real sure why that lighter texture is showing black edges. It doesn't actually have that in the texture...and the other textures aren't doing that...

Very cool stuff though.

The-EG:
In case anyone else wants to mess with this, unzip this into pioneer/data which will overwrite some of the files from this PR and the zip that fluffyfreak attached above:
[terrain textures.zip](https://github.com/pioneerspacesim/pioneer/files/4842255/terrain.textures.zip)

I think it's a decent start for EarthLikeHeightMapped:
image
image

The textures aren't great, but I created them from scratch so there are no issues with licenses.  And, more importantly, they can serve as placeholders until something better is created. 

I approached this by defining terrain types and creating textures for those types: water, light-grass, green-brush, forest, snow and stone (mountains). So in this setup, Earth and Mars are currently using 7 layers of the 16 available.
With this approach, we can create a library of terrain types, and profiles can be built by choosing the applicable types for the type of planet/body. 

With that, the 'atlas' for a profile contains a layer for each type of terrain. The order doesn't strictly matter, but the LUT will reference the layers by index:

```
...
   "atlas" : [
      {
         "description" : "sand / beach",
         "textures" : [
            {
               "texture" : "lightsand.dds",
               ...
            }
         ]
      },
      {
         "description" : "grassy plains",
         "textures" : [
            {
               "texture" : "lightgrass.dds",
               ...
            }
         ]
      },
```
As described above, the LUT texture maps the terrain's height (y) and slope (x) to the layer. It looks like only the red channel is used, so I created a reference table like this to help when modifying the LUT:

Layer | Red (0 / 255) | Earthlike
-- | -- | --
0 | 0.000 | sand
1 | 17.000 | grassy
2 | 34.000 | brush
3 | 51.000 | forest
4 | 68.000 | stone
5 | 85.000 | snow
6 | 102.000 | water
7 | 119.000 |  
8 | 136.000 |  
9 | 153.000 |  
10 | 170.000 |  
11 | 187.000 |  
12 | 204.000 |  
13 | 221.000 |  
14 | 238.000 |  
15 | 255.000 |  

It's important to note that even though the profile only has 7 layers, you still index the LUT as if there were 16 (obviously don't use the ones that don't exist).
My earthlikeLUT.png looks like this (can definitely be improved):
image

The bright red dot in the top left is water (last layer, it's the brightest). The basic idea is that as slope increases (farther right), less vegetation can grow, so it goes from forest, to brush, to grass and then finally to stone. Similar for height (top to bottom): there's water at sea level, followed by a little bit of sand, grass, brush, forest, then stone and finally snow.

fluffyfreak:
@The-EG this is exactly what HOPED would happen with this PR, I've really struggled to pull togethher the textures and visuals but you've done a great job!

The-EG:
@fluffyfreak Thanks! My intent is that this might help get others started with this. 

I also created 2 test profiles, one for height and one for slope, that show how the terrain is actually mapped out and allows one to see what's interesting about the terrain. Here's the height (left) and slop (right) test for the moon (red is min , green is max):
image

So, most of the interesting variation is within the height. The changes in slope mostly occur within the first layer (so using slope won't in the LUT won't do much). To get something interesting looking (for the moon at least), we'll probably need at least 7-8 textures of varying shades, most likely darker shades to lower heights, with most of the variation centered towards the max height.

A similar approach might work for something like Pluto (Rock) (this is the height test, red=min, green=max):
image
I think something with different colors (tan, pinkish, blue-gray) could actually look interesting! (not like pluto probably, but interesting)

nozmajner:
@The-EG The ctrl+F10 shortcut might be handy for this, it views the selected nav target, so you don't need to fly to it. Works for planets and ships as well.
image
You can even play with the planet settings if they aren't using height maps. Handy for custom system making.

impaktor:
> Ctrl+f10

I've added it to my very ugly first draft of development tricks/tools that have, so far, been un-documented:
https://pioneerwiki.com/wiki/Development_Tools#Debug

The-EG:
Here are the test profiles I created. Just replace the profile you want to test out (ie. EarthLikeHeightMapped.json) with one of the profiles (testHeight.json or testSlope.json):
[test_profiles.zip](https://github.com/pioneerspacesim/pioneer/files/4869747/test_profiles.zip)

The-EG:
Also, I put together a page on the Wiki showing how to create some basic textures from scratch: https://pioneerwiki.com/wiki/Creating_Simple_Pattern_Textures_in_GIMP

The-EG:
In case it's helpful to anyone: here's a branch where I was able to get the conflicts resolved and merged with master a month or two ago. I have no idea if my choices on the conflicts were right, but it did compile, run and display the planet texturing stuff properly.

https://github.com/The-EG/pioneer/tree/planet-textures

Web-eWorks:
@The-EG thank you for the rebased branch; I wound up rebasing this one instead (and did a fair amount of cleanup along the way). I've been doing some exploratory work into texture blending, so far I have this:

slope_blending

In this image, a transition between horizontal LUT texels driven by slope is shown as a transition from bright red to black; you can think of it as a contour map as used in orienteering, except it shows the change in *slope*, not in height. The green indicates (artificially scaled by 8x) actual slope value.

Once I get blending between two different textures working correctly, I'm going to merge this branch into a work branch here on the main repository for you, myself, and @zesterer to work on terrain generation and rendering; I'd like to be able to replace all existing generation and texturing code before we merge it into master, but also be able to collaborate.

Web-eWorks:
**HEADS UP**: this branch is going to be force-pushed with the newly rebased branch. To make up for the work you'll have to do locally, have some nice terrain texture blending:
blending
blending2

I think we're almost to the stage where we can merge this to a work branch and start building out the "procedural" aspect of the texturing; I'll take a look at the TODO before evaluating that, so please let me know if there's anything left to implement or fix @fluffyfreak!

impaktor:
Look what I've found (from early 2014):
https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=166

fluffyfreak:
Nice work @Web-eWorks

fluffyfreak:
This needs some discussion about biomes, temperature distribution and movement, along with weather, etc.
Just making a note to myself here more than anything.

impaktor:
> This needs some discussion about biomes, temperature distribution and movement, 

Just general thought: just merge bare minimum working thingy, then we can iterate, and people/players/contributors who feel it's an eye-sore get provoked to improve it for us.

ConfedMajor:
@fluffyfreak I want this feature too, i want to make pioneer more realistic!
But something that i want to change, its... the mars texture... **mars is red**

But if we are talking about terraformed mars, it would look something like this:
imagen
(Fictional Image)
Or this:
imagen(Fictional Image)

Web-eWorks:
Mars right now looks like the second image from the left... the green is just much harder to spot from orbit. That's something we'll look into when we're building the texturing lookup tables, as we definitely want to be able to visually communicate that a planet is being terraformed.

fluffyfreak:
Ok the main issues preventing this being merged are content related.
1. there aren't enough textures
2. variations needed for planet types

Keeper1st:
> Ok the main issues preventing this being merged are content related.
> 
>     1. there aren't enough textures

What sort of additional surface textures are needed?

fluffyfreak:
I am slowly resurrecting this, it's just the codebase has drifted in the... erm, _years_ since I last touched this.

bszlrd:
If you need textures, I think I will have time for them during the winter break. Will it be that lookup table thing?

impaktor:
@fluffyfreak Good to hear! This deserves to be in master (once minimum working version). 

I don't know what is easiest for you, but merge conflicts aren't fun, and especially when you have to propagate them all through 12 commits. Perhaps start a new branch from master, and cherry pick each commit from this branch as needed into it?

(To continue on the branch for this PR, rather than opening a new, you could then reset this branch to the newly made one).

fluffyfreak:
Urgh, I'd ask what I fucked up with that commit history but it's after midnight and I'm going to bed.
I built this atop #5979 because I realised that I needed those changes as that's quite a refactor of the GeoPatch/SPhere code etc. However I appear to have messed that up 🤷 

Anyway, hopefully some Git wizard can help me fix that later with rebasing or something magical involving history rewriting.

It's compiles, builds, runs, it's generally glorious!
Or at least it's as good as it was several years ago and ready for fresh eyes to look at it again.

I've attached a zip with the original textures and LUT, plus the lovely work that @The-EG did.
[terrain.zip](https://github.com/user-attachments/files/17963592/terrain.zip)

fluffyfreak:
It's probably because of commit [1924465] which I did and then brought in the #5979 changes 🤦 anyway, problem for another day

fluffyfreak:
The removal of TerrainColorFractal, and the deletion of the vector3d color arrays reduces the size of the Terrain class from 2800 bytes to 512.

fluffyfreak:
I need to work out how to actually `// do the choosing...` as described on line 633 of `GeoSphere.cpp` so that I can support texture selection from variants based on temperature, life, atmosphere etc.

Suggestions welcome.

Also I'm not sure if the choice of slop & height is really useful for the UV coordinates of the LUT texture. They can be based on anything and I went with them as it seemed logical at the time. However as people have pointed out above slope isn't great and perhaps we could derive some other value to range from 0..1 to replace it?

Maybe the pairing could be height and  `normalised(slope * avg-temperature)`??? it's really about making the 2 axes we have to feed into the UV LUT lookup more useful.

Again suggestions welcome as I'm making this up as I go along.

@bszlrd more textures will be very welcome, but I don't have a single coherent answer for what's needed yet thanks 👍

sturnclaw:
Hmm... I didn't get a Github notification from this PR - this is the first I'm seeing the new activity. Good work @fluffyfreak!

Regarding the point above, from a technical perspective I'd probably define texture sets as JSON-based "biomes", defining

sturnclaw:
Whoops, Github had a major issue there. Not sure what happened, but typing in the editor also triggered the "close with comment" button. Editing the first post to contain the full comment...

MirceaKitsune:
Glad to see more progress on this. It's such a good yet long awaited change. I think the blurry planets are one of the biggest visual limitations remaining that need an urgent upgrade: I understand the focus for the next release is on other aspects, but I sure hope this can be merged too.

ollobrains:
Just some ideas 

Suggestions & Additional Ideas
1. Dynamic Biome Definitions
Why: A planet might combine multiple “biome” configurations (like tundra, deserts, temperate forests) rather than just a single LUT.

Idea: Let the JSON define multiple “biomes,” each with a distinct set of textures. Then, when generating a planet, the game picks which biome set(s) to use based on climate variables (temperature, atmospheric density, latitudinal belt, etc.).
Result: Variety across different latitudes: an equatorial zone with desert or rainforest textures, mid-latitudes with forests or plains, polar caps with ice or snow.
2. Layer Blending With More Than Two Axes
Why: Height + slope alone can be limiting, especially if slope rarely hits extremes.

Idea: Introduce a “temperature–humidity” or “latitude–elevation” approach. For each planet, compute a simple climate model: the near-equator is hotter/wetter, poles are colder/drier, etc. This can be your LUT’s two axes instead.
Result: More intuitive planet-level distribution, allowing seamless transitions from deserts to grasslands to tundra.
3. Multi-Step / Weighted Blending
Why: Hard cutoffs for texture selection can create visible seams or abrupt transitions.

Idea: In the shader, sample adjacent “biome blocks” from the LUT and smoothly interpolate if the terrain falls near a boundary. For instance, if you’re near a slope or height threshold, you blend 70% texture A and 30% texture B.
Implementation: A quick approach is using linear interpolation based on proximity to a LUT cell boundary. Another is sampling multiple LUT cells and blending them proportionally.
4. Distance-Dependent Texture Detail
Why: Planets often look good from orbit, but up close, you see repetitive tiling or low-res textures.

Idea: Introduce “detail maps” that scale differently when the player’s near the surface. For instance, large-scale color variation is from the main LUT, but you overlay a high-frequency detail layer that becomes visible only within a few hundred meters.
Implementation: The code might do a second texture fetch using finer UV scaling, then blend it into the main result (e.g., detailMap * 0.5 + mainTexture * 0.5).
5. Weather & Seasonal Tints (Future Scope)
Why: If a planet has seasons or dynamic weather, terrain coloration or “snow coverage” might shift.

Idea: A param like “snow coverage factor” or “rainfall level” could fade in an appropriate texture layer over time. This might be expensive, so a simplified approach would be random per-visit or triggered by an event in the system’s timeline.
Result: Adds dynamic flavor—players returning to the same planet weeks later might see more snow or dryness.
6. Normal & Specular Map Support
Why: Flat color alone can make surfaces look dull. Adding normal maps provides more “3D” detail.

Implementation: Add an alpha channel or second texture array for each biome’s normal map. If GPU memory is a concern, let players enable or disable normal maps in graphics settings.
Specular: Some planets might have shiny patches (like ice or wet surfaces). You could store specular data in the alpha channel of the diffuse map or in a separate texture if performance/budget allows.
7. Planetary “Recipe Book” for Variation
Why: A single set of textures repeated across many planets can feel repetitive.

Idea: Define a handful of “planet recipes” (e.g., Earthlike, Desert, Iceworld, Tropical, Volcanic, Gas Giant, etc.), each referencing a small set of textures. Then apply random or rule-based color shifts.
Implementation:
For desert planets, pick from a palette of oranges, browns, maybe basaltic black.
For ice planets, pick from whites, grays, tinted blues.
For gas giants, keep code-based swirling color generation but allow random shifts in hue/saturation for variety.
8. Height-Dependent “City Lights” Layer
Why: One user mentioned flying over populated planets at night feels empty.

Idea: A separate “city-lights” texture that only shows up in night regions of the planet. This can be masked by altitude or biome (e.g., only in “urban” zones).
Implementation: Possibly use the same LUT approach—one channel or sublayer for “lights intensity.” The shader checks if we’re on the night side and within an “urban” altitude/biome threshold, then blends city lights in.
9. Tools & Debugging
Why: Creating and tuning LUTs is tricky.

Idea: A small in-game debug overlay that displays the real-time slope/height/temperature/humidity values under the player’s ship, letting you see which portion of the LUT is active.
Implementation: Possibly a toggle in developer mode, with textual readouts or color-coded overlays.
10. Thoughtful Long-Term Structure
Why: The ultimate goal is a coherent terrain system that’s both flexible and well-documented for modders.

Idea: Provide a modular code structure that’s easy to extend:
A “terrain pipeline” that builds out fractals, normal maps, biome logic, etc.
Each step can be replaced or customized with minimal disruption to the rest of the system.
Result: New features (like city lights, weather effects) become simpler to slot in once the pipeline is well-defined.

Just some creative idea fodder, can ignore at will if not useful have a good new years day and 2025

MirceaKitsune:
Seems the new release is out! It contains the #5979 patch which was merged a day prior: I played with it and everything seems to be working perfectly (Linux). How close is this to making it in Git master now that any feature freeze from the release is done? Very eager to see Pioneer with high quality planets over unrealistic low-resolution noise!

impaktor:
@MirceaKitsune The release is literally the master branch, at the time of the release. 
Does the fix not work when you compile from master?

impaktor:
@MirceaKitsune Ah, OK, I misunderstood your __"this"__ to refer to the previously mentioned patch, not this PR:

> Seems the new release is out! It contains the #5979 patch which was merged a day prior: I played with it and everything seems to be working perfectly (Linux). How close is __this__ to making it in Git master

impaktor:
> Very eager to see Pioneer with high quality planets over unrealistic low-resolution noise!

If this PR is in playable state, you could merge it in with a branch from master, and play test it, report bugs back here. That would help developing this.

fluffyfreak:
This branch hasn't been touched in 3 months, it desperately needs a rebase against master which I'll do asap but not today as I'm working and have plans later. Then it needs the history cleaning up, possibly just flattening, and more development work before it could be considered for merging. Currently it really only supports a single planet type, and it's... ok-ish, but not upto release quality yet.

fluffyfreak:
[LUT-textures.zip](https://github.com/user-attachments/files/25246588/LUT-textures.zip)
Textures as a mod that can just be dropped into your `/pioneer/mods/` folder. Includes work from @The-EG 👍

PR #4555, opened 2019-03-06
Fix game load crash with unattached hyperspace clouds

### LINKED ISSUE #4552: Segfault while loading any saved game
### Observed behaviour
When I try to load a savegame, pioneer works for several minutes, slowing down my computer, then exits, printing `Segmentation fault (core dumped)` in the terminal I ran it in.

### Expected behaviour
Pioneer loads the saved game successfully

### Steps to reproduce
1.

```
git clone https://github.com/pioneerspacesim/pioneer
cd pioneer
git checkout 74e38aa37
./bootstrap cmake && cmake --build ./build
./build/pioneer
```

2. Try to load the [save file](https://github.com/pioneerspacesim/pioneer/files/2928515/search.and.rescue.json.gz) mentioned in #4551 or make your own by starting a new game and saving it, and then trying to load it.

74e38aa37 Ubuntu 18.04 LTS

jaj22:
Fixes #4552, maybe.

When loading a saved game with an HS cloud that doesn't have an attached ship, the m_ship pointer is unintialized, causing an immediate crash in MSVC debug and undefined behaviour everywhere else. This patch sets the value of m_ship to 0 in this case.

### DIFF
--- a/src/HyperspaceCloud.cpp
+++ b/src/HyperspaceCloud.cpp
@@ -44,6 +44,7 @@ HyperspaceCloud::HyperspaceCloud(const Json &jsonObj, Space *space) :
 		m_due = hyperspaceCloudObj["due"];
 		m_isArrival = hyperspaceCloudObj["is_arrival"];
 
+		m_ship = 0;
 		if (hyperspaceCloudObj["ship"].is_object()) {
 			Json shipObj = hyperspaceCloudObj["ship"];
 			m_ship = static_cast<Ship *>(Body::FromJson(shipObj, space));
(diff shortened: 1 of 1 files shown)


src/HyperspaceCloud.cpp
```diff
@@ -44,6 +44,7 @@ HyperspaceCloud::HyperspaceCloud(const Json &jsonObj, Space *space) :
 		m_due = hyperspaceCloudObj["due"];
 		m_isArrival = hyperspaceCloudObj["is_arrival"];
 
+		m_ship = 0;
```
fluffyfreak: This should be `nullptr` rather than `0` which we're trying to discourage usage of.

fluffyfreak:
Cool, i'll either merge this and change it to `nullptr` or you can push the change.

impaktor:
@fluffyfreak then please do:
2019-03-07-090638_331x279_scrot

impaktor:
:+1: :smile:

PR #4556, opened 2019-03-07
Atmospheric flight

WKFO:
Adds simple atmospheric flight capabilities to ships. It might not be as good as X-Plane or DCS right now, but it can easily win against Star Wars physics. I had a lot of more work in my mind, but if I tried to implement everything I wanted, I couldn't open this PR before I got bored of the idea already. So instead of trying out complex formulas, I nerfed everything down to their simple and safe forms. (Less complex, less likely to break the game, can actually be reviewed and merged before the sun goes supernova)

Most people can understand the code at first look (I took my time writing descriptive variable names), but here is a more human-friendly summary of the simplest flight dynamics I came up with:

- Atmospheric lift

- Passive aerodynamic control -> ship tends to go where you point it at

- Aerodynamic stability -> ship tends to rotate nose-first into the velocity vector

I'm not an expert on the subject, so if there is something wrong, don't hesitate to yell.

Some experimenting is needed to find viable ship constants for aerodynamically-designed ships. Give me some time and I can come up with those values.

Some bits I might need to explain:

- Atmospheric torque is not applied below 150 m/s because although it is not unrealistic, it could be a bit too annoying.

- Yes, you can scoop fuel. This even makes it a bit easier if you know what you are doing.

- Drag is calculated by multiplying the same things (density, speed squared) you do when calculating lift, so there is nothing wrong with that part.

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -386,7 +386,7 @@ vector3d Ship::CalcAtmosphericForce() const
 	double m_shipLiftCoeff = GetShipType()->shipLiftCoefficient;
 
 	// By converting the velocity into local space, we can apply the drag individually to each component.
-	auto m_localVel = GetOrient().Inverse() * GetVelocity();
+	auto m_localVel = GetVelocity() * GetOrient();
 	auto m_lVSqr = m_localVel.LengthSqr();
 
 	// The drag forces applied to the craft, in local space.
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -31,6 +31,7 @@
 static const float TONS_HULL_PER_SHIELD = 10.f;
 HeatGradientParameters_t Ship::s_heatGradientParams;
 const float Ship::DEFAULT_SHIELD_COOLDOWN_TIME = 1.0f;
+const double Ship::DEFAULT_LIFT_TO_DRAG_RATIO = 0.001;
 
 Ship::Ship(const ShipType::Id &shipId) :
 	DynamicBody(),
@@ -212,6 +213,10 @@ void Ship::Init()
 	p.Set("shieldMassLeft", m_stats.shield_mass_left);
 	p.Set("fuelMassLeft", m_stats.fuel_tank_mass_left);
 
+	fLift = vector3d(0.0);
+	fAtmoTorque = vector3d(0.0);
+	fDragControl = vector3d(0.0);
+
 	// Init of Propulsion:
 	GetPropulsion()->Init(this, GetModel(), m_type->fuelTankMass, m_type->effectiveExhaustVelocity, m_type->linThrust, m_type->angThrust, m_type->linAccelerationCap);
 
@@ -359,6 +364,65 @@ void Ship::SetPercentHull(float p)
 void Ship::UpdateMass()
 {
 	SetMass((m_stats.static_mass + GetPropulsion()->FuelTankMassLeft()) * 1000);
+}
+
+//calculates atmospheric lift
+vector3d Ship::CalcAtmoLift()
+{
+	double m_topCrossSec = GetShipType()->topCrossSection;
+	double m_shipLiftCoeff = GetShipType()->shipLiftCoefficient;
+
+	double m_fDrag = CalcAtmosphericForce(DEFAULT_DRAG_COEFF); //returns 0 if not in atmosphere, so no need to check
+
+	vector3d m_AoAVector = GetOrient().VectorZ().NormalizedSafe() - GetVelocity().NormalizedSafe();
+	double m_AoAMultiplier = m_AoAVector.Length();
+
+	if (m_AoAMultiplier > 1.8) {
+		fLift = vector3d(0, m_fDrag * m_AoAMultiplier * m_topCrossSec * m_shipLiftCoeff * DEFAULT_LIFT_TO_DRAG_RATIO, 0);
+	}
+	else {
+		fLift = vector3d(0.0); //stall!
+	}
+	return fLift;
+}
+
+//calculates simplest form of aerodynamic control
+vector3d Ship::CalcAtmoPassiveControl()
+{
+	double m_topCrossSec = GetShipType()->topCrossSection;
+	double m_sideCrossSec = GetShipType()->sideCrossSection;
+	double m_frontCrossSec = GetShipType()->frontCrossSection;
+	double m_aeroStabilityMultiplier = GetShipType()->atmoStability;
+	
+	double m_drag = CalcAtmosphericForce(DEFAULT_DRAG_COEFF);
+
+	fDragControl = GetOrient().VectorZ().NormalizedSafe() * m_drag * -0.25 * ((m_topCrossSec + m_sideCrossSec) / (m_frontCrossSec * 4)) * m_aeroStabilityMultiplier;
+
+	return fDragControl;
+}
+
+
+//calculates torque to force the spacecraft go nose-first in atmosphere
+vector3d Ship::CalcAtmoTorque()
+{
+	double m_topCrossSec = GetShipType()->topCrossSection;
+	double m_sideCrossSec = GetShipType()->sideCrossSection;
+	double m_frontCrossSec = GetShipType()->frontCrossSection;
+	double m_aeroStabilityMultiplier = GetShipType()->atmoStability;
+
+	vector3d forward = GetOrient().VectorZ();
+	vector3d m_vel = GetVelocity().NormalizedSafe();
+	vector3d m_torqueDir = -m_vel.Cross(-forward); // <--- This is correct
+
+	double m_drag = CalcAtmosphericForce(DEFAULT_DRAG_COEFF);
+
+	if (GetVelocity().Length() > 150) { //don't apply torque at minimal speeds
+		fAtmoTorque = m_drag * m_torqueDir * ((m_topCrossSec + m_sideCrossSec) / (m_frontCrossSec * 4)) * 0.05 * m_aeroStabilityMultiplier;
+	}
+	else {
+		fAtmoTorque = vector3d(0.0);
+	}
+	return fAtmoTorque;
 }
 
 bool Ship::OnDamage(Object *attacker, float kgDamage, const CollisionContact &contactData)
@@ -877,6 +941,11 @@ void Ship::TimeStepUpdate(const float timeStep)
 	const vector3d thrust = GetPropulsion()->GetActualLinThrust();
 	AddRelForce(thrust);
 	AddRelTorque(GetPropulsion()->GetActualAngThrust());
+
+	//apply atmospheric flight forces
+	AddRelForce(CalcAtmoLift());
+	AddForce(CalcAtmoPassiveControl());
+	AddTorque(CalcAtmoTorque());
 
 	if (m_landingGearAnimation)
 		m_landingGearAnimation->SetProgress(m_wheelState);
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -396,7 +396,10 @@ vector3d Ship::CalcAtmoPassiveControl()
 	
 	double m_drag = CalcAtmosphericForce(DEFAULT_DRAG_COEFF);
 
-	fDragControl = GetOrient().VectorZ().NormalizedSafe() * m_drag * -0.25 * ((m_topCrossSec + m_sideCrossSec) / (m_frontCrossSec * 4)) * m_aeroStabilityMultiplier;
+	fDragControl = GetOrient().VectorZ() * m_drag * -0.25 * ((m_topCrossSec + m_sideCrossSec) / (m_frontCrossSec * 4)) * m_aeroStabilityMultiplier;
+
+	if (fDragControl.Length() > (m_drag * 0.5)) //don't let any ship fly infinitely, just in case a very wrong ship value is inserted
+		fDragControl = GetOrient().VectorZ() * m_drag * -0.5;
 
 	return fDragControl;
 }
@@ -417,7 +420,7 @@ vector3d Ship::CalcAtmoTorque()
 	double m_drag = CalcAtmosphericForce(DEFAULT_DRAG_COEFF);
 
 	if (GetVelocity().Length() > 150) { //don't apply torque at minimal speeds
-		fAtmoTorque = m_drag * m_torqueDir * ((m_topCrossSec + m_sideCrossSec) / (m_frontCrossSec * 4)) * 0.05 * m_aeroStabilityMultiplier;
+		fAtmoTorque = m_drag * m_torqueDir * ((m_topCrossSec + m_sideCrossSec) / (m_frontCrossSec * 4)) * 0.1 * m_aeroStabilityMultiplier;
 	}
 	else {
 		fAtmoTorque = vector3d(0.0);
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -366,7 +366,7 @@ void Ship::UpdateMass()
 	SetMass((m_stats.static_mass + GetPropulsion()->FuelTankMassLeft()) * 1000);
 }
 
-//calculates atmospheric lift
+//calculates atmospheric lift by difference in pressure on the top and the bottom of the ship
 vector3d Ship::CalcPressureLift()
 {
 	double m_topCrossSec = GetShipType()->topCrossSection;
@@ -416,19 +416,21 @@ vector3d Ship::CalcAirflowRedirection()
 	vector3d m_AoADragVector = GetOrient().VectorZ() - GetVelocity().NormalizedSafe();
 	double m_AoADragMultiplier = (2 - m_AoADragVector.Length());
 
-	//caclulate redirection of air, apply lift caused by the wing deflection and increase drag
+	//calculate redirection of air, apply lift caused by the wing deflection and increase drag
 	fDragControl = ((-GetOrient().VectorY() * m_AoALiftMultiplier * m_aeroStabilityMultiplier * m_drag * (m_topCrossSec * m_sideCrossSec / m_topCrossSec * m_topCrossSec)) * 0.15 - (GetOrient().VectorZ() * m_drag * m_AoADragMultiplier)) * DEFAULT_LIFT_TO_DRAG_RATIO;
 
 	return fDragControl;
 }
 
-
-//calculates torque to force the spacecraft go nose-first in atmosphere
+//calculates torque caused by the airflow
 vector3d Ship::CalcAtmoTorque()
 {
+	//cross section values are used to estimate how sleek a ship is
 	double m_topCrossSec = GetShipType()->topCrossSection;
 	double m_sideCrossSec = GetShipType()->sideCrossSection;
 	double m_frontCrossSec = GetShipType()->frontCrossSection;
+
+	//better aerodynamic designs get a higher value
 	double m_aeroStabilityMultiplier = GetShipType()->atmoStability;
 
 	vector3d forward = GetOrient().VectorZ();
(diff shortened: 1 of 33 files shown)


WKFO:
@nozmajner https://blender.stackexchange.com/questions/94657/how-to-calculate-the-cross-section-area-of-an-object

I bet there is an easier method (?)

nozmajner:
That's an interesting method. I just flattened three copies of the deneb and cleaned them up to a single ngon. Tht method wasn't quick though, but also it did not need the 3d printing addon, which is still broken in 2.8 I think.

src/Ship.cpp
```diff
@@ -361,6 +366,65 @@ void Ship::UpdateMass()
...
 
+//calculates atmospheric lift
+vector3d Ship::CalcAtmoLift()
+{
+	double m_topCrossSec = GetShipType()->topCrossSection;
+	double m_shipLiftCoeff = GetShipType()->shipLiftCoefficient;
+
+	double m_fDrag = CalcAtmosphericForce(DEFAULT_DRAG_COEFF); //returns 0 if not in atmosphere, so no need to check
+
+	vector3d m_AoAVector = GetOrient().VectorZ().NormalizedSafe() - GetVelocity().NormalizedSafe();
```
mike-f1: I may suggest to find the angle between these vectors?

`GetOrient().VectorZ()` don't needs `NormalizedSafe()`
Instead, in `GetVelocity` you may have a "0" vector,
but in that case I think you could "bail out" because your
ship is almost immbile, isn't?

mike-f1:
I would note there are already some calculations of these forces (specifically "Drag"):

https://github.com/pioneerspacesim/pioneer/blob/169f95461d1c8415b9c200fc58f677af2bd30a82/src/DynamicBody.cpp#L190-L204

..:Just to get you aware of that, in case you didn't know that already

impaktor:
Sweet!

> So instead of trying out complex formulas, I nerfed everything down to their simple and safe forms. (Less complex, less likely to break the game, can actually be reviewed and merged before the sun goes supernova)

Yes, this is exactly how I/we prefer. Start simple, then iterate improvements. For compilcated PRs, chance is too great the author's interest wanes, and it just goes no where.

> Atmospheric torque is not applied below 150 m/s because although it is not unrealistic, it could be a bit too annoying.

Maybe have a sigmoid (or tanh) that successively turns this on/off, instead of a hard limit? I'm not suggesting it for this PR, but for your next / keep in mind. Note: I don't know what I'm talking about, and I haven't tried to see if I notice anything happening at 150+/-1, so if it doesn't make sense, please ignore.

src/Ship.cpp
WKFO: It doesn't have to be directly proportional to the angle, this method makes the derivative of the torque applied change with the angle between vectors, so I that's why I used it this way. Feels less artifical to me. Also, the AoA vector isn't directly used in anywhere right now, so I though I could keep it like that until I got my head into the actual complex stuff.

Thanks, I will remove the NormalizedSafe().

The AoA Vector itself isn't directly used anyway, so it doesn't really matter when the velocity vector approaches 0.

WKFO:
> 
> > Atmospheric torque is not applied below 150 m/s because although it is not unrealistic, it could be a bit too annoying.
> 
> Maybe have a sigmoid (or tanh) that successively turns this on/off, instead of a hard limit? I'm not suggesting it for this PR, but for your next / keep in mind. Note: I don't know what I'm talking about, and I haven't tried to see if I notice anything happening at 150+/-1, so if it doesn't make sense, please ignore.

The force is so weak at 150 m/s that I don't actually think I needed that. I just wanted to put it there just to be safe, out of paranoia. But I think you have a pretty good point, there must be a better way instead of hard coding the value.

WKFO:
I added the ship-specific values to the *.json files. I think I've tested them all, I think most of them can be rated "not bad" at this state.

WKFO:
I've tweaked the values so that most aerodynamically-designed ships are best flown at high speed cruise. (At around 2 km/s to 7 km/s in Earth's skies. For reference, the cruise speed of a 737 is around 300 knots, which is only about 155 m/s.)

Ships which can produce lift are Wave, Deneb, AC33 Dropstar, Venturestar, and Lodos. By far, Wave produces the most proportional lift among all the vessels. (The ship itself is a wing with an engine strapped on below.)

I must say that our atmosphere generation is not very good. They are too thin.

Web-eWorks:
@WKFO 
>I must say that our atmosphere generation is not very good. They are too thin.

How so, and at what height are you talking? IIRC most atmospheres have 90% of their density near the bottom 20% of height, and it drops off fairly quickly after that.

nozmajner:
Cool! I'v just tried it, and it feels like you are actutally in an atmospehere.
I noticed that the thrusters are lighting up (on the wrong side even), as the ship turns into the flight direction. (I'm not touching the controls. I'd guess that's not intentional.
screenshot-20190310-221348

WKFO:
@Web-eWorks For example the Earth's atmosphere could "start" at around 70 - 100km altitude instead of 20km. No matter how thin the atmosphere is, it makes a lot of difference at orbital speeds.

For reference, most of the US space shuttle's reentry phase is considered to begin around 120km altitude, and the dramatic heating and deceleration happens mostly above 20km before the shuttle commander takes control. https://www.nasa.gov/centers/johnson/pdf/584730main_Wings-ch4d-pgs226-241.pdf

Web-eWorks:
@WKFO do you mean 'low' instead of 'thin'? Thin refers to the density, while low refers to the height at which the atmosphere starts. I do agree that we need to move the upper-bound of the atmosphere up if it's at 20km, probably to the Kármán line.

WKFO:
@Web-eWorks Oh yes, I meant "low". Sorry for the confusion.

WKFO:
@nozmajner I think it is the ship trying to compensate for the atmospheric torque.

Web-eWorks:
@WKFO Your implementation is subtly incorrect, mostly due to the fact that drag is internally calculated with the radius of a sphere for area _because reasons_. The other part lies in the fact that some of the values you're applying to the generated drag scalar have no root in physical properties and are completely arbitrary (`DEFAULT_LIFT_TO_DRAG_RATIO` and `mAoAMultiplier`).

***However***, it's good enough to fool the eye and worth merging anyways. @impaktor consider this my endorsement of merge.

I've got a fixup PR in testing, but I'd rather WKFO get the credit for the initial implementation.

impaktor:
I should read the IRC backlog from yesterday, but seems like @jaj22 raised some valid points: 
- now ships will be under forces the AI doesn't know about
- i.e. autopilot / NPC ships will be wonky when under high time acceleration

I assume this leads to problems with mission ships not docking as they should, or similar.

> consider this my endorsement of merge.

Yeah, we iterate improvements once merged, with input from those in the loop, e.g. original author, and who ever feels like it

WKFO:
Well, I could use some guidance on what to do about that. I know don't know much about time accel and the autopilot behaviour other than the "field tests" I've conducted. Maybe I can do something similar to the fudge in dynamicbody.cpp

fluffyfreak:
@WKFO one option might be to simply disable it, or scale it down massively towards zero effect, at higher time accelerations? We can accept approximations as it gets faster.

mike-f1:
> The other part lies in the fact that some of the values you're applying to the generated drag scalar have no root in physical properties

I would say that there some root beyond that math: @WKFO simply don't convert values to angles...
Then you could find lot of papers over internet speaking about how lift to drag coefficent works.
Also the thing about drag may be easily solved putting some 'evolved' formulas.

But apart from math stuffs, I would say that I feel like this PR isn't completed because I see, as a natural step further, a little refactoring.
IMHO You should move all that thing in a separate class with at least a ctor in which you store data parsed from shiptype, and an uodate in which you pass a DynBody. Then you could use that class as a component in ship or whatever.
I would also recomend "update" will return a struct or tuple so you may call it once, and a check in that function against pressure and speed before any more math (below some pressure and speed, return nothing...)
Lastly, you could also use the trick mentioned by @jaj22 to disable updates or the whole component

Web-eWorks:
@mike-f1 most airplanes have lift-to-drag ratios of 7+, even at suitably hypersonic speeds. I'd rather lift be correct (with respect to the altitude / atmosphere as well) and we properly implement parasitic drag instead.

Also, @WKFO upon further inspection, your PR is blatantly physically incorrect - under your model, the lift forces are _highest_ at 0 AoA, whereas in reality they fall off to 5-10% and are highest around 20 degrees. You're also not taking into account the case in which the nose is below the AOA, which should be generating _downward_ thrust due to wing deflection.

I've got the requisite algorithms mostly re-written, so don't worry too much about fixing it right now.

@mike-f1 
I can't say I agree with you about the refactor right now. While it's certainly tempting to consider, I'm going to have to come along later and duplicate all of that work anyways when we fully move to the ECS paradigm. Additionally the DynamicBody "component" system is... the less said about it, the better.

WKFO:
CalcAtmoLift ONLY calculates lift caused by difference in pressure; total lift is a combination of fDragControl and fLift.

WKFO:
The thing I did very wrong is that fLift makes the most of the lift instead of fDragControl. fDragControl should be much much more powerful than fLift.

Web-eWorks:
@WKFO here's a lift graph under my revised flight model. (Lift in this case means force in the local Y axis, including drag at high AoA.) This takes into account the various resources on lift physics I found, including the link you sent me.

flight_lift_model

Green indicates lift from wing pressure/displacement in horizontal flight, red indicates drag forces acting on the Y axis, and blue indicates "lift efficiency", or the lift force divided by the overall loss of velocity.

The "peak lift" point can be adjusted with the AoA falloff (e.g. if 25 or 15 degrees is better), while the drag contribution equation is strictly physically accurate, as I understand it.

(I would have sent this to you on IRC, but you kept quitting before I got a chance.)

WKFO:
@Web-eWorks you should be able to fly upside down now

fluffyfreak:
@WKFO if you wanted too you could pull @Web-eWorks PR/branch #4558 `Web-eWorks/atmo-sim` into this PR and then continue to work on it.

Or would you prefer us to merge this and then his immediately afterwards? That's not generally how we'd prefer to do things as it's better if everything is done in a single PR

WKFO:
@fluffyfreak I'm okay with what you suggest, if it is okay with @Web-eWorks.

WKFO:
I merged two branches into one.

Web-eWorks:
@WKFO looks pretty good to me. I'll get the other PR dealt with when I get back to my main box.

WKFO:
@Web-eWorks Are there any other core mechanics you want to modify?
Otherwise, I (we?) just need to address the possible issues with autopilot and time acceleration.

Web-eWorks:
If the autopilot is landing at 90deg, the new calculations shouldn't affect it at all. I think @jaj22 was going to handle updating the autopilot, but feel free to correct me if I'm wrong.

@WKFO I don't think there's anything else I need to change, though I do want to get better simulation of hypersonics at some point.

impaktor:
> I think @jaj22 was going to handle updating the autopilot, but feel free to correct me if I'm wrong.

I don't think he's said anything to that effect. He mostly keeps an eye on us, and comes to our rescue when we have difficult bugs.

src/Ship.cpp
```diff
@@ -361,6 +364,87 @@ void Ship::UpdateMass()
...
+	// By converting the velocity into local space, we can apply the drag individually to each component.
+	auto m_localVel = GetOrient().Inverse() * GetVelocity();
+	auto m_lVSqr = m_localVel.LengthSqr();
+
+	// The drag forces applied to the craft, in local space.
+	// TODO: verify dimensional accuracy and that we're not generating more drag than physically possible.
+	// TODO: use a different drag constant for each side (front, back, etc).
+	// This also handles (most of) the lift due to wing deflection.
+	vector3d fAtmosDrag = vector3d(
+		CalcAtmosphericDrag(m_lVSqr, m_sideCrossSec, m_sideDragCoeff),
```
mike-f1: I think that if `CalcAtmosphericDrag` will return a `vector3d` and accept as second and third elements a `vector3d`
then you could avoid to treble calls and code become more readable.

src/Ship.cpp
```diff
@@ -361,6 +364,87 @@ void Ship::UpdateMass()
...
+	// TODO: vary drag coefficient based on Reynolds number, specifically by
+	// atmospheric composition (viscosity) and airspeed (mach number).
+	double m_topDragCoeff = GetShipType()->topDragCoeff;
+	double m_sideDragCoeff = GetShipType()->sideDragCoeff;
+	double m_frontDragCoeff = GetShipType()->frontDragCoeff;
+
+	double m_shipLiftCoeff = GetShipType()->shipLiftCoefficient;
+
+	// By converting the velocity into local space, we can apply the drag individually to each component.
+	auto m_localVel = GetOrient().Inverse() * GetVelocity();
```
mike-f1: IIRC, `auto m_localVel = GetOrient().Inverse() * GetVelocity();` is the same as
`auto m_localVel = GetVelocity() * GetOrient();` but without the needs of `Inverse`

mike-f1:
I know it seems I'm getting pedantic, but:
https://github.com/pioneerspacesim/pioneer/blob/12e5be173a9790a82f8cb89329bbfd0bb7a20b40/src/vector3.h#L134-L144

So, the use of this function should be reserved **only** when there aren't alternative.

And this is not the case (IMO), because when velocity decrease then all aerodynamic things can be
thrown away as they became small enought to be trashed.

Second: I don't think it is good to postpone something "good" because there's something
"big" that is expected to change things: even more so because in this case for an "ECS" you
should then start refactoring and grouping...

Third: I note a lot of "jump" between function in different file (from Ship to DynBody then back),
it make difficult understand code, at least for me...

Anyway, to me is not mandatory to implement these changes and the last comment I did
on code. I'm just suggesting for future reference.

src/Ship.cpp
Web-eWorks: I think overloading the function to accept `vector3d` would be best suited in this case, and it'd make things a little easier to read.

src/Ship.cpp
Web-eWorks: My matrix math admittedly isn't quite up to par, but `operator* (Vector<T>, matrix4x4)` does a *transpose* operation, not a matrix inversion. We're specifically converting a vector into local space (the Orient matrix is local-to-world), so the inverse of the world-space orientation matrix is what is required here.

jaj22: The rotation/orientation matrices are orthogonal. For an orthogonal matrix, the inverse and transpose are identical. Hence the Inverse() has the same results as Transpose() but is far slower.

WKFO:
I had to force push my last commit because I screwed something up in my local repo. Nothing to worry about.

src/Ship.cpp
WKFO: I replaced the line with what @mike-f1 suggested.

WKFO:
The last commit fixes the increasing atmospheric torque on the craft when in time accel.

impaktor:
Feels like this PR might need some serious rebasing, and clean up. Prefereably, all "fixes" should be mered into the commits they fix/tweak.

Web-eWorks:
@WKFO I've got a couple of tweaks I want to make to the PR, then we should be ready for squash+merge.

WKFO:
@Web-eWorks Okay.

WKFO:
@Web-eWorks Asking just out of curiosity, what sort of tweaks do you have in mind?

Web-eWorks:
> Asking just out of curiosity, what sort of tweaks do you have in mind?

I want to go through each ship and set them up with appropriate drag and lift constants. I've already done a cleanup pass on the new code to finalize comments and optimize the function (early out when no velocity, etc.).

Unfortunately, it looks like I'm not going to be able to get to that before next Sunday at the very earliest, so feel free to merge this PR and I'll come along later with the final tweaks in a separate PR.

fluffyfreak:
I would prefer to wait until thing were tweaked before merging this PR.
Don't get me wrong, it's a great PR already, but if we know that there's a few more pieces coming soon (_-ish_) then I say we wait for them and everything goes in together.

We don't have a "release date" to hit or publishers to molify just ourselves :)

impaktor:
> I would prefer to wait until thing were tweaked before merging this PR.

I don't necessarily agree with that, we have too long history of getting PRs stuck in perfection-limbo.

...however, the messy commit history is making me cold-sweat.

(yes, I'm aware of the inherent contradiction of my two statements)

laarmen:
On Tue, Apr 16, 2019 at 2:03 PM Karl F <notifications@github.com> wrote:

> I would prefer to wait until thing were tweaked before merging this PR.
>
> I don't necessarily agree with that, we have too long history of getting
> PRs stuck in perfection-limbo.
>
True, but usually due to lack of feedback. Here my email history and
Fluffy's comment tell me it's something else.

fluffyfreak:
Damn perfection, merging

starling13:
Hello. I think there is a little mistake in calculations of the drag basis components.

This is current implementation:
```cpp
...
auto m_lVSqr = m_localVel.LengthSqr();

// The drag forces applied to the craft, in local space.
// TODO: verify dimensional accuracy and that we're not generating more drag than physically possible.
// TODO: use a different drag constant for each side (front, back, etc).
// This also handles (most of) the lift due to wing deflection.
vector3d fAtmosDrag = vector3d(
	        CalcAtmosphericDrag(m_lVSqr, m_sideCrossSec, m_sideDragCoeff),
	        CalcAtmosphericDrag(m_lVSqr, m_topCrossSec, m_topDragCoeff),
	        CalcAtmosphericDrag(m_lVSqr, m_frontCrossSec, m_frontDragCoeff));

// The direction vector of the velocity also serves to scale and sign the generated drag.
	fAtmosDrag = fAtmosDrag * -m_localVel.NormalizedSafe();
...
```
Here we use whole vector squared length to calculate each component, then scale by normalized airspeed vector which seems incorrect.

I propose this change:
```cpp
// Get current atmosphere parameters
double pressure, density;
GetCurrentAtmosphereParameters(pressure, density);

// Precalculate common part of drag components calculation (rho / 2)
const double rho_2 = density / 2.;
// Vector, which components are square of local airspeed vector components
const vector3d lv2(m_localVel * m_localVel);
vector3d fAtmosDrag(
	       -sign(m_localVel.x) * rho_2 * lv2.x * m_sideCrossSec * m_sideDragCoeff,
	       -sign(m_localVel.y) * rho_2 * lv2.y * m_topCrossSec * m_topDragCoeff,
	       -sign(m_localVel.z) * rho_2 * lv2.z * m_frontCrossSec * m_frontDragCoeff);
```
(I'v also added method **GetCurrentAtmosphereParameters** to class **DynamicBody**.)
Next I place the debug output with comparision of old and new drag calculation.
I used **sinonatrix** on Mars with modified drag coefficients.

Values from one time step:
`
Cross section: 40 97 26
Drag coefficients: 0.4 1.1 0.2
Local velocity: 0.143433 -0.35287 -332.503
Pressure: 0.39416 Density: 0.476633
Old drag:  -181.853        2983.52  137009
New drag: -0.0784467    3.16627  137009
`
Look at the huge side and top drag of -181.853 and 2983.52 N with much less then 1 m/s airspeed on that axes in old value. I think new values are more realistic.

impaktor:
@starling13 How does the ship handle under the new change? So head-on drag is unchanged? Then I assume this is a corner case, thus it not having been noticed?

ping @WKFO, thoughts?

(if no further response on this is made, within the next few days, you can open a PR and we take the discussion there)

(also bonus: if you get the drag debug output into the debug Imgui window, hmm, maybe even as graphs, that would be cool)

WKFO:
Glancing at it, @starling13 's change looks correct to me. Head-on drag should stay unchanged, but the ships' response to Angle-of-Attack would change.

I would like to see a PR.

starling13:
I'll do PR in short time.

> also bonus: if you get the drag debug output into the debug Imgui window, hmm, maybe even as graphs, that would be cool

I afraid that I'm not that cool at the moment, but it's worth to try :)

Web-eWorks:
There are sadly still a number of hurdles to cross (mostly related to newUI / oldUI) before debug visualization in C++ is possible - mostly because opening a new Imgui frame before input handling happens has issues. It's on my backlog...

PR #4557, opened 2019-03-10
Refactor sound code to it's own folder, delete Pi::BoinkNoise.

Web-eWorks:
**Pi::BoinkNoise is dead! Long live pigui.playBoinkNoise!**
Continuing my recent PRs, this moves the sound-related code into it's own source folder to cut down on the top-level `src/` bloat, and deletes the abomination formerly known as `Pi::BoinkNoise()`. Only one instance of the underlying call to `Sound::PlaySfx()` remains in C++ source code, but this time it's actually worth keeping as it's being driven by user input.

There's a new function in the `pigui` / `ui` Lua namespace, `ui.playBoinkNoise()`. This is where the functionality should have been at the start, instead of the top-level engine namespace.

EDIT: also removed the unused function `WorldView::OnClickBlastoff()`, which apparently was sticking around from a previous round of refactoring.


impaktor:
Looks like there isn't much to object to here, so assuming @fluffyfreak or anyone else is fine with this, I suggest we merge this soonish (one day/now?)

fluffyfreak:
Stuffs moved so I might have to fixup the vs projects but that's on me so merging now.

PR #4558, opened 2019-03-13
Improve the atmospheric simulation.

Web-eWorks:
Building on PR #4556, this PR refactors the various calculations to be more physically accurate with respect to the ship's topology.

Specifically, this PR replaces the previous, *extremely* broken method of calculating atmospheric drag with a new model that takes into account the difference between drag experienced by an object flying nose-first or belly-first. It also changes the equations used to calculate lift, et al to something simpler and more easy to both maintain and verify as physically accurate.

The end result is that you can actually _fly_ on Mars now, though you have to be going pretty fast and have a very aerodynamic ship. I've not tested this on Earth yet, which is why it's still in the Draft stage.

This PR is dependent upon WKFO's PR, and should be merged after his.

Note: I've not updated all of the ships with new values for drag. You'll be wanting to test-fly the Wave, AC33/VentureStar, or Deneb.

### DIFF
--- a/data/ships/ac33.json
+++ b/data/ships/ac33.json
@@ -32,10 +32,15 @@
    },
    "thruster_fuel_use" : -1.0,
    "up_thrust" : 60000000.0,
+
    "front_cross_section" : 120,
    "side_cross_section" : 280,
    "top_cross_section" : 1190,
-   "lift_coeff" : 1,
+
+   "front_drag_coeff": 0.06,
+   "side_drag_coeff": 0.21,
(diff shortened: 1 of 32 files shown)


fluffyfreak:
@Web-eWorks you know that you can do PR's into someone elses repository and branch right?
I'd think that was the correct way to do this, collaborate with @WKFO on their existing PR so we don't have multiple PR's for the same thing like this.

Web-eWorks:
Merged into #4556

PR #4561, opened 2019-03-21
Fix game load crash typo

### LINKED ISSUE #4560: Segmentation fault when loading a game
Compiled pioneer version: newest (commit 1ca6015)
OS: Arch Linux x86_64
cMake version: 3.14.0
gcc version: 8.2.1

Until commit f4186da it works perfectly, you can load any game without problem. Since commit 34e8965 there is a segmentation fault any time you want to load a game, regardless the game is saved with newer or older versions.
I've tried to debug it and I think the error is in file WorldView.cpp line 64
shipView.LoadFromJson(jsonObj);

I've attached game output, saved game and GDB backtrace.
[output.txt](https://github.com/pioneerspacesim/pioneer/files/2989378/output.txt)
[Marte.zip](https://github.com/pioneerspacesim/pioneer/files/2989380/Marte.zip)
[gdb_backtrace.txt](https://github.com/pioneerspacesim/pioneer/files/2989383/gdb_backtrace.txt)

jaj22:
Fixes #4560.

Looks like a refactoring typo. Crashes on every game load, at least with asserts enabled. Otherwise possibly undefined behaviour.

I haven't tested any other behaviour but this prevents the crash at least, and the rest looks superficially ok.

### DIFF
--- a/src/WorldView.cpp
+++ b/src/WorldView.cpp
@@ -61,7 +61,7 @@ WorldView::WorldView(const Json &jsonObj, Game *game) :
 
 	InitObject();
 
-	shipView.LoadFromJson(jsonObj);
+	shipView.LoadFromJson(worldViewObj);
 }
 
 WorldView::InputBinding WorldView::InputBindings;


Web-eWorks:
This should be correct. In fact, I believe that the preceding line (that sets `shipView.m_camType`) can be moved to `ShipViewController::LoadFromJson` as well.

fluffyfreak:
thanks @jaj22

PR #4569, opened 2019-04-03
Gracefully handle destroyed target ships in SAR missions

### LINKED ISSUE #4547: Search and rescue is impossible when vessel to be rescued gets destroyed
### Observed behavior
Search and rescue missions are sometimes impossible because the vessel you're rescuing gets destroyed shortly after arriving in the system they're in. In my case there's a vessel in 96 Ceti that free-falls into a gas giant when I get there. After hyper-spacing out of that system and then back into it, the same vessel is there again, alive and well as if nothing happened, and free-falls into the gas giant again. There's no way to abort or complete the mission.

### Expected behavior
Either all search and rescue missions include ships in stable orbit, or if they free-fall into a gas giant they get permanently destroyed and the mission is failed. Bonus points if you get to take the spare crewman back to a spaceport safely.

### Steps to reproduce
Download and run the latest stable Pioneer version from feb 2019, start doing search and rescue missions until you do one where the spacecraft you're trying to rescue mysteriously disappears while you're beginning the maneuver towards it. Hyperspace out of that system and back into it, observe that the spacecraft is alive and well again. Use the system map to find out what's happening to the spacecraft.

Pioneer 20190203 (425c372) Ubuntu 18.04.2 LTS x86_64

clausimu:
This should fix #4547. The SAR missions now handle destroyed target ships without leaving a "zombie" mission in the player's mission roster.

If the target ship is destroyed by accident (i.e. gravity pull, etc), the player is commended for trying to help the target ship and payed 1/2 the agreed upon amount.

If the target ship is destroyed by the player, the mission source is very upset, the player gets punished by removing 2x the reputation he would have otherwise gotten, and, obviously, there is not payout. 

In both cases the mission is closed and removed from the player's mission roster. The player has to return to the mission source for this to take effect.

### DIFF
--- a/data/lang/module-searchrescue/en.json
+++ b/data/lang/module-searchrescue/en.json
@@ -694,5 +694,14 @@
   "WHERE_IS_THE_TARGET": {
     "description": "",
     "message": "Where do I find the target?"
+  },
+  "PLAYER_DESTROYED_TARGET": {
+    "description": "",
+    "message": "We have heard of your deeds, captain! Rest assured that we will report you to the local authorities and get you flagged for completely unacceptable behavior. A criminal lawsuit will follow."
+  },
+  "ACCIDENT_DESTROYED_TARGET": {
+    "description": "",
+    "message": "The sad news has reached us that {shiplabel} has been destroyed. We will pay half of the agreed price for you efforts."
   }
+
 }
--- a/data/lang/module-searchrescue/en.json
+++ b/data/lang/module-searchrescue/en.json
@@ -697,7 +697,7 @@
   },
   "PLAYER_DESTROYED_TARGET": {
     "description": "",
-    "message": "We have heard of your deeds, captain! Rest assured that we will report you to the local authorities and get you flagged for completely unacceptable behavior. A criminal lawsuit will follow."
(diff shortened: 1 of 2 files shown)


data/lang/module-searchrescue/en.json
```diff
@@ -694,5 +694,14 @@
   "WHERE_IS_THE_TARGET": {
     "description": "",
     "message": "Where do I find the target?"
+  },
+  "PLAYER_DESTROYED_TARGET": {
+    "description": "",
+    "message": "We have heard of your deeds, captain! Rest assured that we will report you to the local authorities and get you flagged for completely unacceptable behavior. A criminal lawsuit will follow."
```
impaktor: Rather than " get you flagged for completely unacceptable behavior", it might be good if it specified why, e.g. "for destroying the very ship you were sent out to aid" or similar.

clausimu:
Darn... I did not just change the language of the message but also the SAR module for the next coming fix (crashing with pilots < 1). I'll have to fix that later and remove the latest SAR module changes. They are meant for another pull request. Sorry.

impaktor:
I guess just copy the old lua file over the new one, if you only want the language change.

fluffyfreak:
@clausimu / @impaktor is there pending work to be done before this can be merged?

impaktor:
I haven't tested this, and will not have time to, so if @clausimu has self-reviewed (which I assume you've done), then we can squash-and-merge this.

PR #4571, opened 2019-04-20
Manual player face generator

vakhoir:
This is a manual face generator in the player personalInfo screen. It let's you pick and change each feature of the portrait. All other character portraits are still generated randomly (or "randomly").

pioneer-facegen-1

pioneer-facegen-2

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -127,6 +127,10 @@
     "description": "Center on selected star system",
     "message": "Center on System"
   },
+  "CHANGE": {
+    "description": "Face feature option",
+    "message": "Change: "
+  },
   "CHANGE_BINDING": {
     "description": "Key binding option",
     "message": "Change Binding"
@@ -426,6 +430,50 @@
   "FAILED": {
     "description": "When failed mission",
     "message": "Failed"
+  },
+  "FEATURE_SPECIES": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Species"
+  },
+  "FEATURE_RACE": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Race"
+  },
+  "FEATURE_GENDER": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Gender"
+  },
+  "FEATURE_HEAD": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Head"
+  },
+  "FEATURE_EYES": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Eyes"
+  },
+  "FEATURE_NOSE": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Nose"
+  },
+  "FEATURE_MOUTH": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Mouth"
+  },
+  "FEATURE_HAIRSTYLE": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Hairstyle"
+  },
+  "FEATURE_ACCESSORIES": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Accessories"
+  },
+  "FEATURE_CLOTHES": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Clothes"
+  },
+  "FEATURE_ARMOUR": {
+    "description": "Face feature for FaceGenerator in the PersonalInfo view",
+    "message": "Armour"
   },
   "FEMALE": {
     "description": "",
@@ -1187,6 +1235,10 @@
     "description": "",
     "message": "Negotiate"
   },
+  "NEXT_FACE_FEATURE": {
+    "description": "Face feature switch button for FaceGenerator in the PersonalInfo view",
+    "message": "Next"
+  },
   "NEXT_PAID": {
     "description": "",
     "message": "Next paid"
@@ -1343,6 +1395,10 @@
     "description": "",
     "message": "Press a key or controller button or escape to clear"
   },
+  "PREVIOUS_FACE_FEATURE": {
+    "description": "Face feature switch button for FaceGenerator in the PersonalInfo view",
+    "message": "Previous"
+  },
   "PRICE": {
     "description": "",
     "message": "Price"
@@ -1374,6 +1430,10 @@
   "RATING": {
     "description": "",
     "message": "Rating:"
+  },
+  "RANDOM_FACE": {
+    "description": "Generate random face",
+    "message": "Random"
   },
   "REAR_WEAPON": {
     "description": "",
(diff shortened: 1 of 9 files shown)


impaktor:
Cool! 

- Please remove all *.json-files except en.json, and do a `git add ; git commit --amend`
- for future / down the line, the way I imagine it is for manual faces is that you can only change it when starting a new game. I imagine it like "New Game" -> takes you to a menu where you can choose face parts, name, sex, and starting location (and in future: role playing traits like strenth, charisma, etc.). This would obviously be done in the PiGui of the main menu, instead of "New"UI (BBS-screens) that this is done in. 

FYI the "NewUI" is getting kicked out of pioneer, so this code will have to be re-written in PiGui at some point, but I'd say we still merge this as is, as code re-write can take weeks or many years.

impaktor:
Wait: there's no longer a "create random face" feature for the user? Because I think we still want that, no? Feels silly to remove a feature, when they shouldn't be mutually exclusive?

vakhoir:
- Ah, makes sense. I was thinking it has to be tedious to modify every language whenever something is added, good to know you don't actually have to do that.
- Yeah, I had the same idea.
- Yeah I removed it because I didn't want to shrink the portrait to fit another button. Maybe I can still fit it right under the previous/next buttons? I hope it won't get too cluttered.

impaktor:
> Yeah I removed it because I didn't want to shrink the portrait to fit another button.

I'm quite sure the rest of us would be more than happy if the face gen images got smaller, as they're quite ugly. There has even been talk of removing them (but I think they're now so comically ugly it's almost become a trope/campy part of pioneer)

vakhoir:
All I want to know is what was in the head of the guy who thought a bunch of flies would make a good accessory...
But I'm actually looking around for an artist that's going to draw something more reasonable to look at.

Anyways, I'll play around with the gui and post the screenshots, so we can pick the best one.

vakhoir:
Well, turned out there's plenty of space. How about this?

pioneer-facegen-3

fluffyfreak:
Yeah that looks good to me

nozmajner:
Nice!
I'd like to suggest a better UI layout though, that one seems a bit clunky.
I think something like this would be better and easier to use:
generator-ui-suggestion
If you can do an additional column, that is.
I can cook up some proper icons for you if you want.
And lighter on translators, not that it's a huge burden to translate an additional few words.

vakhoir:
> I think something like this would be better and easier to use

Nice! I'll give it a go, and see if it works out

nozmajner:
Oh, and if you can actually find somebody who would remake the face sets, that would be awesome. It is like if the whole thing is cursed. There have been quite a few starts at it the last couple of years (including some low-momentum ones from me), but nobody really worked trough it yet, not even close to it. There's a [whole section for them on the dev forum](https://pioneerspacesim.net/forum/viewforum.php?f=8).

vakhoir:
Hm, I might be missing something, but it seems the icons are in an svg file, and the method to load these is only in pigui.

Seems messy / like a waste of time to either extract the icons just for this, or  write a method for the old gui to handle svg files. Maybe it makes more sense to leave this as is, and I'll try to rewrite the Personal  Info view for pigui?

nozmajner:
There are a bunch of icons already in png format for newUI, I could extend that.
But yeah, the question is, how much it worth anyway, especially if you want to take a stab at the infoscreen pigui conversion anyway.

impaktor:
Yeah, pigui conversion would be the absolute best. I'd suggest that as a new branch, and if that project stalls we can merge this as is.

fluffyfreak:
Can I just jump in to say that whilst a PiGui branch would be great, this can (_and should_) be merged pretty much as is.

data/lang/ui-core/en.json
```diff
@@ -427,6 +431,50 @@
     "description": "When failed mission",
     "message": "Failed"
   },
+  "FEATURE_SPECIES": {
+    "description": "",
+    "message": "Species"
+  },
+  "FEATURE_RACE": {
+    "description": "",
```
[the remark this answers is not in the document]
impaktor: Note: This is just a suggestion, not something I demand you change. I just write all that comes to my mind as I check the code.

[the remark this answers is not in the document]
vakhoir: I went ahead and added a basic description

src/gameui/LuaFace.cpp
```diff
@@ -9,39 +9,59 @@ namespace GameUI {
 
 	class LuaFace {
 	public:
-		static inline Uint32 _unpack_flags(lua_State *l)
+		static inline bool _get_int(lua_State *l, int table, const char *key, int &output)
```
mike-f1: NOTE: Instead of writing a new function (which I think should have a name more like "_find_int" if I guess correctly its complexity), may it be better if you find a way to use existing code:

https://github.com/pioneerspacesim/pioneer/blob/1847087e21a43e3675b066793639c09caef8ce60/src/LuaTable.h#L46-L59

...It seems it do the same without reinventing the wheel

vakhoir: Oh, good point. Still getting familiar with the code, and haven't come across that class, thanks!

Just pushed another commit. I'll squash them all later once it's good to merge.

mike-f1: My 2 cents are that this should go as is: changing the overall design cost
time and is not worth the effort... It should be better done when (and if :( )
this will be rewritten in ImGui.

src/gameui/Face.h
```diff
@@ -4,6 +4,7 @@
 #ifndef GAMEUI_FACE_H
 #define GAMEUI_FACE_H
 
+#include <src/FaceParts.h>
```
mike-f1: Should you use quotation marks?

src/gameui/LuaFace.cpp
vakhoir: > My 2 cents are that this should go as is

Right, I meant it in the "if nothing else comes up"... like the using #include with quote marks thing ;)

src/gameui/Face.h
vakhoir: Yeah good point. I pushed an update.

vakhoir:
Just noticed another problem that I also just fixed.

I forgot to update the ChatForm.lua fule, which was still using "MALE" and "FEMALE" strings to setup gender, and this was leading to a crash. I switched that to an int now.

I also decided to rename the field from character.face to character.faceDescription, so it doesn't get confused with face widget used in the chat/bulletin board

impaktor:
Nitpick perhaps, but why not boolean for male/female and armor 1/0, etc?

vakhoir:
> Nitpick perhaps, but why not boolean for male/female and armor 1/0, etc?

The FaceDescriptor descriptor struct actually used integer for every single feature all along, and the FaceParts class generally allows to pick from an arbitrary (well, INT_MAX) number of features. I don't see a reason to arbitrarily limit that for specific features.

With Armors in particular, I think even in vanilla there's several types, so there's no good reason to limit that. I will have to make another change to the lua scripts, because right now if a character is armored it will always select the first armor from the list

impaktor:
Fair point

vakhoir:
>  I will have to make another change to the lua scripts, because right now if a character is armored it will always select the first armor from the list

I just pushed the mentioned update

fluffyfreak:
@vakhoir ok so this is ready for review and possible merging? 
No other tweaks or changes you want to make?

vakhoir:
Yeah, good to go from my perspective, the only other thing I wanted to do is squash the commits, which I just did.

Web-eWorks:
Other than the changes I requested above, this code looks good to go.

PR #4572, opened 2019-04-20
Autofit main menu

### LINKED ISSUE #4570: Buttons in start menu do not fit their box upon resizing
[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/3064919/opengl.txt)
2019-04-10 18_49_39-Pioneer

Verison: Pioneer20190203

### Observed behaviour

Start menu buttons no longer fit their box after resizing the screen.

### Expected behaviour

Either the box where start menu buttons are supposed to be enlarges to accomodate their new, longer width, or the buttons are rescaled (font and frames) to fit in the pre-determined box.

### Steps to reproduce

1. Install Pioneer
2. Start Pioneer
3. Change resolution to 1768x992 (I also changed a bunch of other stuff, not sure if it is relevant, but I feel it's not)
4. Quit Pioneer
5. Start Pioneer again

Freshly installed Pioneer 20190203
Windows 10 Pro v 1803

vakhoir:
Hopefully I got this right, went by the comment in imgui.h

`// set next window size. set axis to 0.0f to force an auto-fit on this axis`

Fixes #4570

### DIFF
--- a/data/pigui/mainmenu.lua
+++ b/data/pigui/mainmenu.lua
@@ -198,6 +198,7 @@ local function showMainMenu()
 	end)
 
 	ui.setNextWindowPos(winPos,'Always')
+	ui.setNextWindowSize(Vector(0,0), 'Always')
 	ui.withStyleColors({["WindowBg"] = colors.lightBlackBackground}, function()
 		ui.window("MainMenuButtons", {"NoTitleBar", "NoResize", "NoFocusOnAppearing", "NoBringToFrontOnFocus"}, function()
 			mainTextButton(lui.CONTINUE_GAME, nil, canContinue, continueGame)


impaktor:
Maybe @ecraven would like to review this one-liner change?

vakhoir:
For anyone testing this - I noticed a funny thing, even after checking out a different branch where my fix was not merged, the bug did not come up again for me.

In fact, no matter what I do to the script now, I cannot replicate it the bug anymore, even though it was happening every single time before.

nozmajner:
Imgui creates a settings file, imgui.ini ~Pioneer (or Documents ó/Pioneer), that could be what is making it into a heisenbug.

vakhoir:
Ah, that got me on the right track!
The menu size is stored in the .ini, that's why it ends up wrong after the resolution is changed. I was expecting the size to get messed up right after removing the fix, but to reproduce the resolution needs to be changed again.

PR #4573, opened 2019-04-27
Prepare InfoView for move to PiGui

vakhoir:
So this is pretty much @Web-eWorks 's branch (from #4441 ) but rebased to resolve all conflicts with master, and adjusted to make gradual work easier.
The way this is supposed to work is that both the old and new UI are processed at the same time, but depending on the `showView` field in the PiGui version either the old or new UI is actually displayed.
This is what it looks like, the ShipInfo tab is old gui, and the PersonalInfo is PiGui (I haven't actually done any work on it yet, it's just to show the idea).

infoview-pigui-migration-01

infoview-pigui-migration-02

My idea was that as a first step, I'd switch off the showView for PersonalInfo after I get feedback, and we merge the whole thing as is. After that it will be possible to split work on the InfoView and review/merge each tab separately.

PR #4576, opened 2019-05-03
Refactor game UI, refactor flag handling, move InfoView to pigui.

### LINKED ISSUE #4590: Improve Pigui lua flag operations.
This has been bugging me for a while, ever since @ecraven and I were chatting about it. Currently pigui handles passing ImGui flags (e.g. `ImGuiWindowFlags_NoResize`) from Lua by passing a table containing strings and looking those strings up in a `std::map` to convert them to the integer enum values ImGui expects. It looks a little like this:

```lua
ui.window("<NAME>", { "NoResize", "NoBringToFrontOnFocus", ... }, function() ... end)
```

On the C++ side, this entails copying the string contents from the table into a `std::string`, copying them _once again_ into another stack-allocated `std::string` (though my testing shows that this is mostly optimized out at `-03`), hashing the string with `std::hash`, looking the value up in the map, and then returning the integer enum value.

This is (obviously) comparatively expensive - copying lots of small strings around isn't good for performance, and while the `LuaTable` implementation helps a little with cache coherency it's still not great. This got me thinking... in Lua, there's a little implementation detail: strings are compared and passed around by their hash, which is calculated _once_ (strings are immutable in lua). What if we made Lua do the hashing and map lookup instead?

So I wrote a quick profiling case of 100,000 invocations and got some interesting numbers out of it. `std::map` style is as described above. `lua table` uses a Lua-hosted lookup table to translate string keys into integer constants, and `lua integer` just passes a table of integer values directly to the function, which are error-checked against a `std::set` for validation.

```
std::map style took 44ms.
lua table style took 27ms.
lua integer style took 18ms.
```

Surprisingly, raw integers aren't much faster than doing a lua-side table lookup, coming in at only 50% faster. Depending upon optimization level, lua tables were 2-3 times faster than our existing implementation, probably more as my test case is a small, easily optimized C-style function. Clearly we should just pass raw integers around, right?

Not quite. The reason we pass strings around instead of integers is because it makes for much better error detection. When you pass `NoResize` into a function instead of `CharsUppercase`, a string will fail, while an integer will be unknowingly accepted as correct. Considering that we have very little in the way of static validation for our lua code and pioneer is by its nature a project for rookies and part-time developers, it is usually better to err on the side of resiliency instead of speed.

So, a long winded explanation later, I'm proposing to create a lua lookup table for each flag type instead of a `std::map` to prevent extraneous string creation and speed up flag parsing - quite a number of UI constructs make heavy usage of flags, and speeding that up means more cycles that we can use on other things.

Another related change I'd suggest is to store our flag constants in a table (`ui.WindowFlags` etc.) and refer to them by name instead of raw string - this helps to prevent misspellings and will provide at least a little error detection if we later decide to pass integer constants instead.

Web-eWorks:
This PR will close #4590.

I've split the monolithic `views/game.lua` file into several topical files to reduce cognitive load and increase maintainability, as well as provide a better interface for mods to register their own elements of the game UI. Eventually, I want to standardize the UI elements and provide 'hardpoints' for mods to add their own UI code to, but this is a good first start.

I've fixed a nasty order-dependent segfault - calling `ui.getProjectedBodies()` in the same frame after switching to the WorldView would cause a segfault due to the camera's Frame not existing (and a complete absence of code testing for that case, as well as passing raw pointers around without validating them). To fix that, I've changed `Pi.cpp` to unconditionally create and destroy the camera frame each frame regardless of whether the WorldView is active; this is only a few cycles and fixes the issue quite decisively.

~~I intend to port the face display over to pigui before I mark this PR as ready, though if someone wants to do that on their own recognizance I'd really appreciate it.~~
**EDIT:** thanks @vakhoir!


vakhoir:
> I intend to port the face display over to pigui before I mark this PR as ready, though if someone wants to do that on their own recognizance I'd really appreciate it.

Yeah, I'm already on it. I've been approaching it since #4571, and #4573 is the first step for that (didn't want to rewrite the entire info view at once, so I set it up so views can be worked on one at a time). I might have a PR ready by next week if I don't run into any problems.

As for this branch, it looks like you might need to rebase, I did an amend commit on #4573.

mike-f1:
It seems you included changes from #4567...

Web-eWorks:
@vakhoir Yeah, I'll rebase once that PR gets merged into master.

@mike-f1 that's on purpose, yes.

Web-eWorks:
I've implemented the proposed changes from #4590, with the exception that I haven't yet added the `ui.WindowFlags` et. al. functions. I haven't run a profiling pass on this, partly because I have no clue how to enable profiling in pioneer, and partly because I profiled the changes separately in isolation and they're a very small amount of frametime.

src/LuaPiGui.cpp
```diff
@@ -65,6 +66,13 @@ void *pi_lua_checklightuserdata(lua_State *l, int index)
 	return nullptr;
 }
 
+void pi_lua_generic_pull(lua_State *l, int index, ImVec2 &vec)
```
mike-f1: Mmmmhh... I don't like too much mixing these types...
I mean that when I removed ImVec2 pull/pop I explicitly copy/pasted
the conversion to avoid this kind of things to kick in: it was IMO a
(primitive?) way to confine ImVec2 usage.

Anyway, if that thing is not appreciated and thus you would this function
in master, sure you should use it extensively: thus changing all the involved
lines in order to standardize things and increase visibility, thus avoiding any
subsequent programmer from implementing its owns methods.

Web-eWorks: Had I caught it earlier, I would have suggested that you define a conversion function instead of going to the work of rewriting every function that needs an ImVec2, but it is what it is. I intend to go back through and fix up each function which does the conversion manually, so don't worry; I'll establish a common convention.

Considering that the alternatives are to add a UDC to either ImVec2 or vector2d, or to manually instantiate multiple ImVec2s in every single function that needs to pull a vector from lua, I think this is an acceptable solution.

mike-f1: ...Well, as you wish... Not a problem but you answered yourself:

`draw_list->AddLine(ImVec2(p1.x, p1.y), ImVec2(p2.x, p2.y), color, thickness);`

(line 329, `LuaPiGui.cpp`)

Why this shouldn't suffice? Compiler will throw away useless bytes of code and you're
away from any mixing of these types or from `ImGui` types proliferation...

.gitignore
```diff
@@ -79,3 +79,5 @@ result
 *.exp
 *.VC.db
 /release
+# Core dumps shouldn't be tracked by git.
+.[09][09]*
```
impaktor: Github says it wants a RET here

src/LuaPiGui.cpp
Web-eWorks: Because it's a significant amount of extra work and complexity, both for anyone maintaining the existing code, and anyone adding new code to the PiGui implementation. The entire point of a conversion function like `LuaPull` is to hide and consolidate the verbose code that takes one input (e.g. a Lua value) and outputs the same content in a different format (e.g. an ImVec2).

.gitignore
Web-eWorks: Yeah, I realized that after I committed it.

src/LuaPiGui.cpp
mike-f1: Good, so it seems I'm the only one having the will to prevale over that copy-paste-hell. :sweat_smile: 

I'm prone to suggest you the use of a couple of "helper function"; I mean something like
```
vector2 ImVec2vector( ImVec2 )
vector3 ImVec2vector( ImVec3 )
```
and back, with a name that helps separating their purposes from `LuaPush/Pop(****)` methods, because in general

> The entire point of a conversion function like LuaPull 

...is to pull/pop from Lua, and if you hide that complexity (IMHO), you end having no idea of what type of data are you working with (...a side effect of overloading) as demostrated in recent PR where has been was really difficult to explain to all people (including myself) what happened there...

Web-eWorks: And yet, using those functions still adds extra complexity (and work!) to what is solved by the existing adapter function, `LuaPull`. Considering that the specific overload is defined in only one CPP file, it's not like we're hiding where the conversion is happening or what underlying datatype is being pulled in - we're just reducing the amount of work it takes to pull a vector type from Lua and pass it to ImGui.

I'm sure @impaktor and @fluffyfreak will back me up when I say that something that makes the code more maintainable and reduces the amount of unnecessary copy-and-paste boilerplate you need to write for _every single function_ is something much more preferable for Pioneer then requiring our already busy contributors to manually remember and perform the conversion every time.

Web-eWorks:
Added `ui.WindowFlags` and cohorts, finishing out the tasks set out in #4590. Also cleaned up the leftover manual vector conversions; many thanks to @mike-f1 for his hard work on optimizing the core UI code!

At this point I think this PR is ready for merge; I'll be working on the InfoView and ModelSpinner code in a separate PR based on this one, I think.

impaktor:
looks like there are clang format errors?

Web-eWorks:
@impaktor it's an include-order thing, you can merge as-is. IIRC we were going to turn that specific clang-format check off anyway, so it doesn't really matter.

impaktor:
would be good to have @laarmen and/or @ecraven voice opinion on this, as it's related to what was discussed in the issue this PR will close.

fluffyfreak:
Ready to review?
I was really hoping to have some more Lua/IMGUI voices involved but in the interest of getting PR assessed I'd rather move it along.

fluffyfreak:
Ready to review?
I was really hoping to have some more Lua/IMGUI voices involved but in the interest of getting PR assessed I'd rather move it along.

Web-eWorks:
@fluffyfreak ready for review. It's a bit of a mishmash because I realized that the Travis static-checks step had been broken for some time during the middle of the PR, but otherwise it does what it says on the tin.

vakhoir:
Oh this was still WIP last I checked. I'll take a look at it later today, if you want another set kof eyes on it.

data/modules/DebugShipSpawn/DebugShipSpawn.lua
```diff
@@ -0,0 +1,107 @@
...
+  end
+  return table.sort(ship_defs)
+end
+update_ship_def_table()
+print(#ship_defs)
+local selected_ship_type = 0
+local draw_ship_types
+draw_ship_types = function()
+  for i, ship in ipairs(ship_defs) do
+    if ui.selectable(ship, selected_ship_type == i) then
```
vakhoir: After the refactoring of flag handling this line causes a crash. Flags either need to be explicitly passed here, or `l_pigui_selectable` in `LuaPiGui.cpp` should be changed to use `LuaPull` with a default value.

src/LuaPiGui.cpp
```diff
@@ -16,42 +16,43 @@
...
-			theflags = static_cast<Type>(theflags | table.at(flag));
-		else
-			Error("Unknown %s %s\n", name.c_str(), flag.c_str());
+	Type theFlags = Type(0);
+	if (lua_isnumber(l, index)) {
+		theFlags = static_cast<Type>(lua_tointeger(l, index));
+	} else if (lua_istable(l, index)) {
+		theFlags = lookupTable.LookupTable(l, index);
+	} else {
+		Error("Expected a table or integer, got %s.\n", luaL_typename(l, index));
```
vakhoir: Would be a good idea to add `luaL_traceback` here, and in all the other parsing functions, to the error output, hunting down the source of the error without it is a bit tedious without it.

data/modules/DebugShipSpawn/DebugShipSpawn.lua
Web-eWorks: Hmm. I thought I changed `l_pigui_selectable` to use the default value. Must not have survived the rebase.

Regarding indentation, that's because this file is autogenerated from `DebugShipSpawner.moon` in the same directory - I'll see what I can do about that.

Web-eWorks:
That should fix travis compilation, and there's nothing I can do about appveyor right now.

fluffyfreak:
Damn, sorry @Web-eWorks but merging other stuff has caused some conflicts here. If you can fix 'em up I'll merge.

Web-eWorks:
@fluffyfreak rebased, fixed up and ready for merge. Feel free to squash these commits if you want, but I've rebased this two (three?) times already so I'm very done with it.

PR #4578, opened 2019-05-05
Fix crash on main menu

CylonicRaider:
The refactoring in #4548 missed one instance of `Vector`; this causes a Lua panic as soon as the main menu is shown.

I hope I'm not racing against someone else's fix; as of this writing, I see no mentions of this in the issues.

### DIFF
--- a/data/pigui/mainmenu.lua
+++ b/data/pigui/mainmenu.lua
@@ -197,7 +197,7 @@ local function showMainMenu()
 	end)
 
 	ui.setNextWindowPos(winPos,'Always')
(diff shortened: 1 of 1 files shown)


CylonicRaider:
Tangentially, it appears that the reticule is broken as well, displaying like follows:

reticule

...This probably should go into a separate issue.

fluffyfreak:
Thanks! I dunno how I missed that when testing

CylonicRaider:
See also #4581.

CylonicRaider:
Also see also #4580 and #4583.

PR #4579, opened 2019-05-05
[Workaround to avoid CTD] Initialize variables to be integers

mike-f1:
Get an error in master:
```
Could not find model: sinonatrix_cockpit
error: [string "[T] @modules/SearchRescue/SearchRescue.lua"]:544: attempt to perform arithmetic on local 'pickup_crew' (a nil value)
stack traceback:
	[string "[T] @modules/SearchRescue/SearchRescue.lua"]:544: in function 'calcReward'
	[string "[T] @modules/SearchRescue/SearchRescue.lua"]:1333: in function 'makeAdvert'
	[string "[T] @modules/SearchRescue/SearchRescue.lua"]:1984: in function 'cb'

```

...Thus I pushed the simpler solution.
It is quick and dirty: ~~not tested~~ and it should be checked if works for every mission

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -648,7 +648,9 @@ local createTargetShipParameters = function (flavour)
 	local shipdef = shipdefs[rand:Integer(1,#shipdefs)]
 
 	-- number of crew
-	local crew_num, pickup_crew, deliver_crew
+	local crew_num=0
+	local pickup_crew=0
+	local deliver_crew=0
 	if flavour.id == 1 or flavour.id == 6 then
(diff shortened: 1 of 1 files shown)


vakhoir:
I had a look, and tested it several times as well, seems ok to me. Didn't get any more crashes so I suppose no other missions are affected.

mike-f1:
> I had a look, and tested it several times as well, seems ok to me. Didn't get any more crashes so I suppose no other missions are affected.

Yeah: it is working because it took "nil" which wasn't a number: My doubts are about 
the correctness of "0" as default value thus the "workaround" in PR title.

fluffyfreak:
I'm going to merge this as it avoids a crash to desktop. Correctness can come later ;)

mike-f1:
Link to #3795 (or more probably #4569) which seems to introduce the bug...

PR #4581, opened 2019-05-05
Replace leftover lua Vectors with Vector2

vakhoir:
Not sure if we use rotated icons anywhere, but if we do, the users would be in for a surprise.
I searched all lua files for occurrences of the old Vector class, this should be the last leftover, but maybe someone can doublecheck.

### DIFF
--- a/data/pigui/pigui.lua
+++ b/data/pigui/pigui.lua
@@ -335,7 +335,7 @@ ui.addWideIcon = function(position, icon, color, size, anchor_horizontal, anchor
 	  local up_right = up_left:right()
 	  local down_left = up_left:left()
 	  local down_right = -up_left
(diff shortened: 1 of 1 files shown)


mike-f1:
First of all, thank's for the fix ;)

I note that that function (`get_wide_icon_tex_coords`) is pretty the same as `get_icon_tex_coords`, thus these can be simplified down to one parameterized function.

I would also pin that [there](https://github.com/ocornut/imgui/issues/1982) seems uv coord should not be rotated... I may be wrong, but probably this code needs some test (or some test-case).

fluffyfreak:
@vakhoir have we added you to the 'AUTHORS.txt' file yet? Would you like your name in there?

vakhoir:
I'd be honored :)
I'll add myself with the next PR. I should be finished with the pigui face generator this week.

fluffyfreak:
@vakhoir ok or you can tell me what name you'd like to be added under (_screenname or real name, we accept either_) and I can just edit the file directly.

PR #4583, opened 2019-05-06
Fix flight UI direction and vector indicators

### LINKED ISSUE #4580: In-flight UI broken
The latest code changes — i.e. #4548 — have snuck in a few UI bugs, aside from the crash I noticed in #4578.

<p align=center>
Broken reticule
</p>

I observed that:
1. The movable portion of the horizon indicator is... doing strange things.
2. The approach speed indiator constantly shows zero (even after accelerating the ship in various directions).
3. The direction indicators are missing. (The left-pointing chevron appeared to me spuriously while rotating the ship.) When turning the camera (and not the ship), the direction to front indicator appears; I cannot definitely say whether it points whither it ought to.
4. The small version of the prograde vector indicator appears even while the indicator itself is inside the reticule.
5. Not visibly on this image, the background of the body name (the dark area behind e.g. “Earth”) is shifted upward w.r.t. the body name itself.

I've prepared a fix for the horizon indicator in CylonicRaider@7e014150ae0f1ce625593dd7bd12c799be21cbf7; the other points are beyond my current motivation to debug.

mike-f1:
Closes #4580

### DIFF
--- a/data/pigui/pigui.lua
+++ b/data/pigui/pigui.lua
@@ -426,8 +426,8 @@ ui.addStyledText = function(position, anchor_horizontal, anchor_vertical, text, 
 									position.y = position.y - font.offset
 								end
 								if bg_color then
-									pigui.AddRectFilled(Vector2(position.x - textBackgroundMarginPixels, position.y - size.y + textBackgroundMarginPixels),
-														Vector2(position.x + size.x + textBackgroundMarginPixels, position.y + textBackgroundMarginPixels),
+									pigui.AddRectFilled(Vector2(position.x - textBackgroundMarginPixels, position.y - textBackgroundMarginPixels),
+														Vector2(position.x + size.x + textBackgroundMarginPixels, position.y + size.y + textBackgroundMarginPixels),
 														bg_color,
 														0,
 														0)
--- a/data/pigui/views/game.lua
+++ b/data/pigui/views/game.lua
@@ -66,12 +66,16 @@ local function displayReticuleHorizon(roll_degrees)
 	local radius = reticuleCircleRadius - offset
 	-- left hook
 	ui.lineOnClock(center, hrs, width, radius, colors.navigationalElements, 1)
-	ui.lineOnClock(nil, hrs + height_hrs, width, radius, colors.navigationalElements, 1)
-	ui.lineOnClock(nil, -3, width/2, radius, colors.navigationalElements, 1)
+	ui.addLine(ui.pointOnClock(nil, radius, hrs),
+	           ui.pointOnClock(nil, radius, hrs + height_hrs),
+	           colors.navigationalElements, 1)
+	ui.lineOnClock(nil, -3, -width/2, radius, colors.navigationalElements, 1)
 	-- right hook
 	ui.lineOnClock(nil, hrs + 6, width, radius, colors.navigationalElements, 1)
-	ui.lineOnClock(nil, hrs + 6 - height_hrs, width, radius, colors.navigationalElements, 1)
-	ui.lineOnClock(nil, 3, width/2, radius, colors.navigationalElements, 1)
+	ui.addLine(ui.pointOnClock(nil, radius, hrs + 6),
+	           ui.pointOnClock(nil, radius, hrs + 6 - height_hrs),
+	           colors.navigationalElements, 1)
+	ui.lineOnClock(nil, 3, -width/2, radius, colors.navigationalElements, 1)
 end
 
 -- display the compass at the top of the reticule circle
@@ -216,7 +220,7 @@ local function displayDirectionalMarkers()
 	local function displayDirectionalMarker(ship_space, icon, showDirection, angle)
 		local screen = Engine.ShipSpaceToScreenSpace(ship_space)
 		local coord = Vector2(screen.x, screen.y)
-		if screen.z <= 0 then
+		if screen.z <= 1 then
 			ui.addIcon(coord, icon, colors.reticuleCircle, Vector2(32, 32), ui.anchor.center, ui.anchor.center, nil, angle)
 		end
 		return showDirection and (coord - center):length() > reticuleCircleRadius
@@ -230,8 +234,7 @@ local function displayDirectionalMarkers()
 		end
 	end
 	aux.z = -1
-	local forward = Engine.ShipSpaceToScreenSpace(aux)
-	local forward2 = Vector2(forward.x, forward.y) - center
+	local forward = Engine.ShipSpaceToScreenSpace(aux) - Vector3(center.x, center.y, 0.0)
 	local showDirection = displayDirectionalMarker(aux, icons.forward, true)
 	aux.z = 1
 	showDirection = displayDirectionalMarker(aux, icons.backward, showDirection)
@@ -243,7 +246,7 @@ local function displayDirectionalMarkers()
 	aux.y = 0
 	aux.x = 1
 	showDirection = displayDirectionalMarker(aux, icons.right, showDirection)
-	aux.y = -1
+	aux.x = -1
 	showDirection = displayDirectionalMarker(aux, icons.left, showDirection)
 
 	if showDirection then
@@ -285,13 +288,13 @@ local function displayIndicator(onscreen, position, direction, icon, color, show
 		ui.addIcon(position, icon, color, size, ui.anchor.center, ui.anchor.center)
 		if tooltip then
 			local mouse_position = ui.getMousePos()
-			if (mouse_position - position):magnitude() < indicatorSize:magnitude() then -- not size on purpose, most icons are much smaller
+			if (mouse_position - position):length() < indicatorSize:length() then -- not size on purpose, most icons are much smaller
 				ui.setTooltip(tooltip)
 			end
 		end
 	end
 	-- only show small indicator if the large icon is outside the reticule radius
-	if showIndicator and (center - position):magnitude() > reticuleCircleRadius * 1.2 then
+	if showIndicator and (center - position):length() > reticuleCircleRadius * 1.2 then
 		ui.addIcon(indicator, icon, color, indicatorSize, ui.anchor.center, ui.anchor.center)
 	end
 end
@@ -306,7 +309,7 @@ local function displayDetailButtons(radius, navTarget, combatTarget)
 	if combatTarget or navTarget then
 		local color = reticuleTarget == "frame" and colors.reticuleCircle or colors.reticuleCircleDark
 		ui.addIcon(uiPos, icons.display_frame, color, Vector2(size, size), ui.anchor.left, ui.anchor.bottom, lui.HUD_SHOW_FRAME)
-		if ui.isMouseClicked(0) and (mouse_position - (uiPos + Vector2(size/2, -size/2))):magnitude() < size/2 then
+		if ui.isMouseClicked(0) and (mouse_position - (uiPos + Vector2(size/2, -size/2))):length() < size/2 then
 			reticuleTarget = "frame"
 		end
 		uiPos.x = uiPos.x + size
@@ -314,7 +317,7 @@ local function displayDetailButtons(radius, navTarget, combatTarget)
 	if navTarget then
 		local color = reticuleTarget == "navTarget" and colors.reticuleCircle or colors.reticuleCircleDark
 		ui.addIcon(uiPos, icons.display_navtarget, color, Vector2(size, size), ui.anchor.left, ui.anchor.bottom, lui.HUD_SHOW_NAV_TARGET)
-		if ui.isMouseClicked(0) and (mouse_position - (uiPos + Vector2(size/2, -size/2))):magnitude() < size/2 then
+		if ui.isMouseClicked(0) and (mouse_position - (uiPos + Vector2(size/2, -size/2))):length() < size/2 then
 			reticuleTarget = "navTarget"
 		end
 		uiPos.x = uiPos.x + size
@@ -322,7 +325,7 @@ local function displayDetailButtons(radius, navTarget, combatTarget)
 	if combatTarget then
 		local color = reticuleTarget == "combatTarget" and colors.reticuleCircle or colors.reticuleCircleDark
 		ui.addIcon(uiPos, icons.display_combattarget, color, Vector2(size, size), ui.anchor.left, ui.anchor.bottom, lui.HUD_SHOW_COMBAT_TARGET)
-		if ui.isMouseClicked(0) and (mouse_position - (uiPos + Vector2(size/2, -size/2))):magnitude() < size/2 then
+		if ui.isMouseClicked(0) and (mouse_position - (uiPos + Vector2(size/2, -size/2))):length() < size/2 then
 			reticuleTarget = "combatTarget"
 		end
 	end
@@ -342,8 +345,9 @@ local function displayDetailData(target, radius, combatTarget, navTarget, colorL
 	-- current distance, relative speed
 	uiPos = ui.pointOnClock(center, radius, 2.75)
 	-- currently unused: local distance, distance_unit = ui.Format.Distance(player:DistanceTo(target))
-	local approach_speed = position:dot(velocity) / position:magnitude()
-	local speed, speed_unit = ui.Format.Speed(approach_speed)
+	local approach_speed = position:dot(velocity) / position:length()
+
+	speed, speed_unit = ui.Format.Speed(approach_speed)
 
 	ui.addFancyText(uiPos, ui.anchor.left, ui.anchor.baseline, {
 										{ text=speed,      color=colorLight, font=pionillium.medium, tooltip=lui.HUD_SPEED_OF_APPROACH_TO_TARGET },
@@ -356,6 +360,7 @@ local function displayDetailData(target, radius, combatTarget, navTarget, colorL
 	local altitude = player:GetAltitudeRelTo(target)
 	local ratio = brake_distance / altitude
 	local ratio_retro = brake_distance_retro / altitude
+
 	speed, speed_unit = ui.Format.Speed(velocity:length())
 
 	uiPos = ui.pointOnClock(center, radius, 3)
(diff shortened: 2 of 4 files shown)


CylonicRaider:
See https://github.com/pioneerspacesim/pioneer/issues/4580#issuecomment-489709602 for my (mis-posted) review as of commit 2d9fdfeda221a6750114a9ea2dceba74e9ce6328.

mike-f1:
@CylonicRaider , are things working now?

CylonicRaider:
As of commit c87de1c90a857bccc797fdce2c2fa4e4f2286d4f, all direction indicators are present, but their rotation is computed incorrectly: When I reset the view to default, then free-look such that the dorsal indicator is in the middle of the reticule, and let the camera pivot around its (approximate) view axis, it takes two rotations of the indicator for the stars to perform one rotation.

CylonicRaider:
The to-front indicator points almost correctly, except that it's mirrored along the (screen's) vertical axis.

CylonicRaider:
```diff
 static int l_vector_angle(lua_State *L)
 {
        const vector2d *v = LuaVector2::CheckFromLua(L, 1);
-       lua_pushnumber(L, atan2(v->x, v->y));
+       lua_pushnumber(L, M_PI * 2 - atan2(v->x, v->y));
        return 1;
 }
```
Remark that `atan2()` has the **y** argument first (so, it computes `tan(y)` and uses `x` to tweak the result).

mike-f1:
...Sorry, but I couldn't solve every thing that come out from reviewing (...so well, I would state) all that code:
https://github.com/pioneerspacesim/pioneer/blob/ee84abb8c56c648a947315f6efdecd93126b31ed/data/libs/Vector.lua#L85-L89

The purpose should be limited to address bugs related to the offending PR.
Anyway, if x and y are inverted, then that thing is calculating the co-tangent: thus that inversion is simply adding (or removing) an half-Pi (IIRC what I learned looooong time ago).

CylonicRaider:
Whoever wrote the Lua version of `angle()` might have been confused about `atan2()` as well (remark the seemingly unnecessary normalization operation); however, if this interpretation is deeply entrenched in other code, refactoring that out is indeed a task for another day. Perhaps document that the weird definition is kept for backwards compatibility?

mike-f1:
I feel it is better to open a new issue: it is easy "per-se" but given the facts that a lot of PR are rewriting all UI stuffs and that I'm waiting for other bugs, comments may be forgotten...

But I hope some maintainer will give its opinion...

fluffyfreak:
I'm going to ping @laarmen and @impaktor for advice about Lua stuff (_which I should probably have done a few days ago!_)

CylonicRaider:
@ecraven, Git blames you for the `angle()` method quoted in https://github.com/pioneerspacesim/pioneer/pull/4583#issuecomment-489768391; is there any particular reason for it being structured like that?

CylonicRaider:
@mike-f1, in either case, the nonstandard behavior of the `angle()` method is something to keep in mind.

CylonicRaider:
I also propose to rename this PR to something like “Fix flight UI direction and vector indicators”.

CylonicRaider:
<dl>
<dt>To-front indicator:</dt>
<dd>Still almost broken. </dd>
<dt>Direction indicators:</dt>
<dd>Still partially broken. (BTW, one should be able to reuse the code of the to-front indicator for these — let them be copies of the to-front indicator at fixed locations with chevrons as icons.)</dd>
<dt>Zenith indicator:</dt>
<dd>Still broken. (In my testing game, I can see it together with the body it is supposed to point away from, which can't be good.)</dd>
</dl>
...Ahh, you renamed the commit.

impaktor:
OK, I haven't really followed the latest stream of activity, as I've been away from keyboard for a few days.

Regarding the [arctan2](https://en.wikipedia.org/wiki/Atan2): so this is some issue from the recently merged PR #4548

> Whoever wrote the Lua version of angle() might have been confused about atan2() as well (remark the seemingly unnecessary normalization operation); however, if this interpretation is deeply entrenched in other code, refactoring that out is indeed a task for another day.

I only see `angle()` used in two lines: `aux:angle()` in views/game.lua

Also, this PR now has a conflict.

mike-f1:
> To-front indicator:
    Still almost broken. 
Direction indicators:
    Still partially broken. (BTW, one should be able to reuse the code of the to-front indicator for these — let them be copies of the to-front indicator at fixed locations with chevrons as icons.)
Zenith indicator:
    Still broken. (In my testing game, I can see it together with the body it is supposed to point away from, which can't be good.) 

...I don't have much to say because I can't do any test (last time I tryed things were working)...
Anyway I would like if you try to explain how exatly these are broken, because (in related issue) it seemed you "approve" the changes I did or at least that those changes were working...

> ..Ahh, you renamed the commit. 

Yeah, at last a good idea, isn't? ;)

> so this is some issue from the recently merged PR #4548

Not exactly: It _was_ because I did wrong when I copied from original code (now is correct)...
But I restored the original code and @CylonicRaider is telling there _was_ something
strange before (now is not yet correct).

> I only see angle() used in two lines: aux:angle() in views/game.lua

You should check in files other than those changed by this PR: indeed the change affect
`LuaVector.cpp` and will propagate to every occurence of `Vector2:angle()` ...
Even if @CylonicRaider could be right, it will take time to grasp the situation, thus I suggest
to avoid any additional change (reasons are above); to open an issue to pin this thing (...and
probably some dozens of others that are coming to light...); letting you or whoever of
maintaners decide how to proceed.

Let me know

impaktor:
>>    I only see angle() used in two lines: aux:angle() in views/game.lua

> You should check in files other than those changed by this PR: indeed the change affect
LuaVector.cpp and will propagate to every occurence of Vector2:angle()

Well, I only see 2 lines in _Lua_ in the code base that uses `:angle(...)`.

> letting you or whoever of maintaners decide how to proceed.

If I'm to speak for "the maintainers", I'd say we want the bug resolved, as soon as possible (preferably within a few days), as this seems like a serious incorrect algorithm, if it's done as separate PR or not I don't really care, we've already started the discussion on it here. If you/anyone opens a separate PR or issue report, please summarize the bug-description from this thread, as I still don't quite get what the issue is, or who is right: original code by @ecraven (that is tried and tested), @mike-f1 untested code from 2 days ago, or @CylonicRaider

CylonicRaider:
I can't start a new game because I get lots of Lua crashes (haha), so I can't provide reproduction steps more specific than “load a save file, go into space, and look how it behaves”.

Anyway, I observe the following behavior of the to-front indicator:

When the reticule is above or below the front indicator, the to-front indicator looks normally:
<p align=center></p>

When the reticule is to the left or right of the front indicator, however, the to-front indicator points opposite of where it ought to:
<p align=center></p>

In summary, the to-front indicator turns opposite of how it should, as if there were a sign error while computing its angle. Once it works reliably, it should be easy to correct the orientation of the directional markers (where it matters); looking at the code, I also notice that the “left” and “right” markers are not turned (in contrast to the “up” and “down” ones), although they probably need that, too.

Additionally, it looks like the approach speed indicator is incorrect now; the ship I gathered the screenshots with is actually in orbit around Earth with an approach speed of about 0.

mike-f1:
> If I'm to speak for "the maintainers", I'd say we want the bug resolved, .....

@impaktor , are you serious?

CylonicRaider:
@mike-f1, yes, “the maintainers” do indeed (with overwhelming probability) want that the bug be resolved.

Aside from that, did my latest review help?

CylonicRaider:
@impaktor, the executive summary is that PR #4548 (the time interval between the known-good and the known-bad commit also includes #4569 and #4577, which do not touch the files relating to *this* issue/PR) introduced several bugs in the flight UI, some of which I detailed in #4580. This PR was opened as a response to #4580, and aims to fix the concrete issues mentioned there; it has gone through a few feedback-correction cycles. I coindidentally noticed the strange `angle()` implementation while looking at the difference between two commits in this PR; it is not immediately relevant, but might be investigated in the future.

If there are indeed only two uses of `angle()` in Lua code, it might not have been *that* “tried and tested”; its two consumers could have been adjusted so that they work in spite of the nonstandard definition of `angle()`.

Given that some of the bugs are rather serious, I would propose finishing this PR quickly and fixing other issues elsewhere.

mike-f1:
> ...finishing this PR quickly and fixing other issues elsewhere.

No. I'm opposing... I will do the thing in a different way.

I will solve this increscious situation. Or at least I will try...

I hope you will give me enought time to prepare something working 

My bad, how I could get to that... I'm so sorry :(

CylonicRaider:
While @mike-f1 is dithering I've adjusted their code to fix the bugs noted in #4580 (at least for me) at CylonicRaider@cf970b488e30397c5decb0bb2247c44a32147a79. It would be nice if someone else could verify that the bugs are actually fixed.

mike-f1:
> The commit does not address the zenith indicator or other bugs noted in the issue's follow-up discussion.

@CylonicRaider , I don't know how to thank you for your support, but if also you aren't
able to find a viable solution for that thing of the "zenith indicator", then #4548 is ruining everything... 
And given @impaktor and other maintainers wills... Well, I'm to remove it.

Are the "zenith indicator" or other bugs something able to ticke your curiosity?
I hope you're able to debug that bug and solve all this incresious situation, or
I don't know :(

impaktor:
It would be good to establish if this is a bug reproducable for others as well. (I'm at work and cant test now)

CylonicRaider:
Reverting #4548 and redoing it more carefully sounds like an option worth considering, IMHO. Should it be taken, this PR becomes obsolete, and additional effort may be required to rebase (or drop) PRs merged after #4548.

If #4548 is to be kept, one way to expedite fixing at least some of the bugs it introduced would be to change `Vector` and `Vector2`'s `magnitude()` methods to raise errors (e.g. by dropping them altogether). Before #4548, these methods returned the *length* of the vector, while now they return the *squared* length; since the old `Vector` had no method for a squared length — and unless the new `magnitude()` is deliberately used somewhere —, we can be sure that any use of `magnitude()` is erroneous and fails silently.

***

I can reproduce (or rather demonstrate) the zenith indicator bug (on the current `master` branch as I get crashes on this branch) as follows:
1. Start a new game on Mars.
2. Blast off and use the autopilot to enter a low orbit around Mars (time acceleration may be used to reduce the tester's wait).
3. Point the ship's bow at the zenith indicator: Ensure you are looking forward (w.r.t. the ship); holding the right mouse button, move the mouse until the yellow double circle coincides with the zenith indicator.
4. Look backward (w.r.t. the ship; in a default install, this is bound to the `Down` arrow key).
5. Observe how the icon representing Mars (a dot with the text “Mars” besides it) is *not* in the center of the reticule.

Unless I misunderstand the purpose of the zenith indicator or my spatial reasoning is seriously flawed, I would expect the Mars icon to be in the center of the reticule at Step 5.

jaj22:
For what it's worth, any function called "magnitude" should return the length rather than the squared length, as that's the mathematical convention. I would guess that's just a porting error.

If the intended behavior of the UI is documented somewhere then I can look into it.

CylonicRaider:
As far as checking whether the UI works as inended goes, the [wiki](https://pioneerwiki.com/wiki/Flight_UI) and the indicators' and gauges' tooltips provide information about what is supposed to do what.

Web-eWorks:
@CylonicRaider 
> If #4548 is to be kept, one way to expedite fixing at least some of the bugs it introduced would be to change Vector and Vector2's magnitude() methods to raise errors (e.g. by dropping them altogether)

Given the convention of most frameworks / engines, I would suggest that we drop `magnitude()` and use `length()` and `lengthSqr()` instead. Several calculations (at least with 3d vectors) have a large speed benefit from comparing the squared length of a vector instead of incurring the extra cost of a square-root operation before comparing.

Regarding a revert, I will simply say that these errors absolutely *should* have been caught by the PR author before it was merged, and much greater care should have been taken to ensure all the systems touched by the PR were updated to work with the new changes. It will likely be more work to re-implement the changes brought by that PR than it is to fix the breakage, so my opinion falls mostly on the side of "stay".

The bugs we're seeing now are indicative to me that something is fishy in the underlying calculations and assumptions - I'm not sure if that's the UI code's assumptions, the original vector implementation, or @mike-f1's implementation. I would suggest bisecting the git history and carefully studying the changes to determine what is responsible for breaking things like the zenith indicator, as it will likely be responsible for a whole host of other issues.

CylonicRaider:
The bugs listed in the description of #4580 (that this PR aims to address) were introduced less than 24 hours before I filed #4578 (citation: *I played the game the day before*); while I lost the precise known-good commit, the time window is narrow enough to only include a few “translation updates” (more precisely, actually none), #4548 (which I blame as the culprit), #4577 (whose diffstat looks fairly innocuous), and #4569 (which did introduce regressions fixed in #4579, that, however, look irrelevant to this issue).

I had pegged the zenith indicator as unreliable way before #4548 was merged, which suggests that it is not immediately related #4548; locating the root cause of that issue will probably require additional effort indeed.

mike-f1:
...Thus, summarizing:
1) Zenith bug => Before #4548 
2) Vector2:angle() miss copy => #4548 
3) Vector2:angle() wrong offset => Before #4548 
4) Misuse of magnitude => #4548 
5) Misname of magnitude => #4548 

...mmmh... thus if this PR were merged monday then what was left were...

...Misname of magnitude?

:/
Am I wrong?

CylonicRaider:
Depending on, or, rather, regardless of which Monday you refer to (May 6th or perhaps May 13th), merging this PR (as it is/was up to now) would miss the fixes I prepared in CylonicRaider@cf970b488e30397c5decb0bb2247c44a32147a79, leave the zenith indicator more-or-less as it was before (if my memory is to be trusted), indeed leave a misnamed `magnitude()` in the code, and (most importantly) not work at all because of the conflict. :wink:

mike-f1:
Ahahah (...though if you had avoided jokes it would have helped ;) ...)

6th or 13th didn't have to make a difference because:

> and (most importantly) not work at all because of the conflict. :wink: 

...In #4548 talk I miss where I said to avoid merging things over the files involved, so
probably is my fault to not have warned maintaners that that PR were buggy
_for sure_ : IMO a change like that should be "baked/rested": it touches too many
little things to be perfect at first...

Anyway: the above to introduce what could be a viable solution: rebase master
so this PR (with your [commit](https://github.com/CylonicRaider/pioneer/commit/cf970b488e30397c5decb0bb2247c44a32147a79https://github.com/CylonicRaider/pioneer/commit/cf970b488e30397c5decb0bb2247c44a32147a79) of course) result before the
PRs merged this week (#4573 is the main problem...).

...the naming issue (which **I** did wrong) could be addressed later, or considering
that before my PR there wasn't a method that return a squared length, thus one simply
can use "length" everywhere, or rename that method and see where faults are...
IIRC before the #4573 `magnitude` calls weren't so many and I (not everywhere :( )
should have used "magnitude" in a form like:
`Vector2:magnitude() < AnotherVector2:magnitude()`

...And if some magnitude remain, come on: it is truly a "big bang"?

CylonicRaider:
> a change like that should be "baked/rested"

Given the recent vote for #4548 staying, it sounds like we'll have to speed-bake it in the next (and past!) days.

> rebase master so this PR (with your commit of course) result before the PRs merged this week (#4573 is the main problem...).

While that is technically possible, it would break every fork that was updated since then, and is therefore undesirable. We'll have to accept that `master` was broken at some point and try to ensure that does not happen again.

> ...And if some magnitude remain, come on: it is truly a "big bang"?

Yes, if it happens in the approach speed indicator, it makes manual flying (which Pioneer encourages, if I recall some discussion correctly)... err... rather hard.

In any case, it might be worthwhile to include a `compareLengths()` method instead of `magnitude()` such that the squared length is not exposed to users at all and a few C++/Lua round-trips are avoided. @Web-eWorks, is that idea any good?

Web-eWorks:
>is that idea any good?

It's not _bad_ per se, but I'd much rather expose `lengthSqr` along with it, as it puts more power into the hands of users. We should not sacrifice versatility here, especially when it's being sacrificed in favor of a non-standard method that only accounts for a single use case of squared vector length. While it's certainly the most prevalent use, I'm sure there are one or two other applications that would benefit from having access to the squared length of a vector, and not having to reconstruct it with unnecessary operations is the best outcome.

Web-eWorks:
Regarding the angle method, the existing implementation is designed to calculate the angle of a vector with respect to `0 = up`, increasing counter-clockwise. From a user's point of view, this appears to be the correct origin if not the correct direction, and though I know of at least one game engine where `0 = right` instead.

I'm probably not the correct person to pontificate about this, so I'll ping @fluffyfreak for his input regarding coordinate systems.

mike-f1:
So, if I didn't lost something, this last commit should solve the issues...

I replaced every occurence of `magnitude` with `length` in order to gain time
to choose how to bring the power of `lengthSqr` to users (hope @Web-eWorks don't
complain if I choose a quick&dirty way...). As a conseguene, I found three more
misusages of `magnitude` in `game.lua`: line 312, 320 and 328.

>  I should have used "magnitude" in a form like:
Vector2:magnitude() < AnotherVector2:magnitude()

...Or in a form like:
`Vector2:magnitude() < 1`

Anyway a `lengthSqr` method is present in `LuaVector*` replacing the old wrong
name (as @jaj22 pointed out).

I checked #4573 code and it seems there no usage of offending methods,
so, if not everything is addressed, I feel this is (at least) an improvement.

About the `angle()` method, my 2 cents are to avoid any difference
between Lua and C++ worlds.

The "Zenith bug" could be addressed later (I would encourage @CylonicRaider 
to open an issue... or to try solve it, if he is brave enough :wink:)

fluffyfreak:
Ok so this now fixes _most_ of the issues?
It's been a very confusing PR to follow along with but it sounds like there's just this "Zenith bug" left to fix?

mike-f1:
> Ok so this now fixes most of the issues?

Hope yes :crossed_fingers: 

> It's been a very confusing PR to follow along with but it sounds like there's just this "Zenith bug" left to fix?

...Not only but mainly

Sure this should be merged before any additional change on UI, please :stuck_out_tongue_winking_eye:

CylonicRaider:
I'll create a new issue to track the zenith indicator bug, given that — IMO — it is less critical than the bugs fixed in this PR.

Aside from that, commit be341d43a94136aa669a05f579c1aeb7dbb50443 has my :thumbsup:.

impaktor:
@mike-f1 did this change keyboard/mouse/joystic handling?
if so, don't forget to update the [manual](https://pioneerwiki.com/wiki/Manual)

mike-f1:
@impaktor , I hope not: it should get "broken" things back as manual says.

But as far I can see, there's a lot of picture not updated: they seems older than PiGui.
The most important, IMO, is that there are no new [face generator](https://pioneerwiki.com/wiki/File:Manual_infoview_02_commander.png).

PR #4584, opened 2019-05-07
Pigui personal view

vakhoir:
So... still WIP, don't even bother reviewing unless you have nothing better to do, but wanted to get some feedback on the looks. The buttons don't feel quite right to me, and I'll play around with them a bit more.
Also, @nozmajner, any chance you could add the new icons to the svg file, like the ones you came up with in #4571 ?

pigui-infoview-draft


impaktor:
Cool! I'm sure @nozmajner can give you the icons as a svg. 
(Usually we then make that as a commit with "--author=nozmajner")

@nozmajner and others: to be consistent in not using text for any of the icons (in nozmajners mockup), how about using [this](https://fontawesome.com/icons/random?style=solid) for "Random"-button?

Example:
poop

vakhoir:
> (Usually we then make that as a commit with "--author=nozmajner" and include in the PR)

Will do! 

@nozmajner, FYI the list of features is:

- FEATURE_SPECIES
- FEATURE_RACE
- FEATURE_GENDER
- FEATURE_HEAD
- FEATURE_EYES
- FEATURE_NOSE
- FEATURE_MOUTH
- FEATURE_HAIRSTYLE
- FEATURE_ACCESSORIES
- FEATURE_CLOTHES
- FEATURE_ARMOUR

In the mockup from the pevious image there was no icon for race. I'm wondering if armour should be customizable or not.

@impaktor, I also wanted to ask about the stuff in /src/gameui/ folder. Is this a library we'll be phasing out? I'm wondering whether I should keep the Face stuff there, or start migrating it to /src/ where all the other Lua stuff is, or create yet another ui component folder.

impaktor:
Yes, gameui folder is being fased out/purged, I'm not sure if there's any use of any of it after InfoView, StationView and F6-F8 in sectorView are ported to PiGui. So please remove unused old code from there as you move to PiGui/imgui.

Web-eWorks:
@vakhoir `gameui` folder is where custom widgets for the GUI / newUI are stored. Once those two systems are gone, the folder is going as well.

Regarding folder structure, I think we'd be best served by creating a new folder `src/pigui/`, and storing `LuaPigui.*` `Pigui.*` and the new `PiguiFace.*` / `PiguiModelSpinner.*` files there. It would cut down on the bloat of the top-level `src/` folder, and enable us to better separate the C++ sources so we don't have to recompile as much when things change.

nozmajner:
Will do, but can't right now. Hopefully i can do them in the weekend.

impaktor:
If this is a time to make feature requests, it would be nice if some assets were excluded from player's (and random generator's) ability to use. E.g. we don't want player's or NPCs with police uniform, unless the police station UI is asking Lua for it*. Similarly with the spoof/Easter egg bear costume, which isn't much of an Easter egg if it's always there. 

But I'm not averse to having e.g. a santa hat (only) on christmas or similar.

*from memory, I think giving `armor=true` is how we get a police. Would be better with "police=true", as we hopefully get military, pirate, santa or other variables in the future.

vakhoir:
> If this is a time to make feature requests, it would be nice if some assets were excluded from player's (and random generator's) ability to use. E.g. we don't want player's or NPCs with police uniform, unless the police station UI is asking Lua for it*. Similarly with the spoof/Easter egg bear costume, which isn't much of an Easter egg if it's always there.

> *from memory, I think giving armor=true is how we get a police. Would be better with "police=true", as we hopefully get military, pirate, santa or other variables in the future.

Yeah, like I mentioned I was wondering how to handle it. The thing is, unless I'm missing something, the only relevant question is whether to expose the variable in the UI. Whether we handle it with a value in the FEATURE_ARMOUR variable of the faceDescriptor, armor=true, or police=true doesn't matter.

Since the feature selector, and the game datafiles allows for an arbitrary number of features, the way I'd handle police/military/pirate is to have seperate features for police/military/pirate uniforms. Then we could have something like police=true, and the selector would choose one of the available uniforms. We could even modify it so that every faction could have their own uniform, or something.

impaktor:
I'd say don't let the face-UI picker be able to choose armor, or such, as I think they should be given after missions or similar, so user can't just add uniform to themslves.

(Just for future reference, from Lua-code we want to be able to set armor, uniform, etc. maybe this is done once player finished an important mission, or becomes a police or similar)

_come to think of it, I never got to choose my face_

laarmen:
You never got to fly around space in a newtonian universe, either ;)

impaktor:
@laarmen to where do I send my feature requests?

laarmen:
Depends on $LOCALE and custom features. For me, it'd probably be `>
/dev/null`, for others it'd be god@heaven.org (there's a special queue
if the sender's address starts with Margaret (see
https://en.wikipedia.org/wiki/Are_You_There_God%3F_It%27s_Me,_Margaret.
) )

vakhoir:
So this is pretty much finished, I just need the icons, and I'm waiting for #4583 to be merged so I can rebase.

@Web-eWorks, are you doing other InfoViews? If not, I'll go ahead to finish them, but if so I'll move on to station comm views.

nozmajner:
Sorry, but I don't think I can do the icons in the coming weeks, I'm swamped neck-deep. I'd suggest merging it as is, then I can later make the icons. You could add the lines for it in the theme default.lua, pointing them to the same icon, then I can update that. in a separate PR.

Web-eWorks:
@vakhoir I'm working on the other screens of the InfoView slowly but surely. Go ahead and merge this and I'll pull it into my local branch - I should be able to get it working again now that master is unborked.

Regarding the station comms screens, I'd suggest writing a translation layer that mimics the existing API for the bulletin board system, as otherwise rewriting all the modules that touch the BBS is going to be a massive pain. We can refactor later once the actual pigui rendering stuff is done, and having the intermediate layer will likely help us later on with caching etc.

vakhoir:
So I ended up taking a stab at the icons. Some still look a bit goofy but I wanted to have some reasonable placeholders.

I think I also prefer the "Random" button to be text. I like the way it fills the space at the bottom. Although I guess I can still change that to an icon

I still want to tweak some things and test, but it should be ready by tomrrow.

pigui-infoview-draft-2

fluffyfreak:
It looks great! I also agree with the Random button being text, for some reason it make it clear that's it's separate and the text is always clearer than an icon

impaktor:
> I think I also prefer the "Random" button to be text. I like the way it fills the space at the bottom. Although I guess I can still change that to an icon

I don't agree. Yes, it looks symmetric and nice how the word "random" happens to align nicely _in this layout_ with English. Try it with swedish "Slumpmässig" and it's suddenly not so nice.

The upside with icons instead of text is they give consistent UI, and all buttons aught to have tooltip text when mouse over the button anyway. E.g. the first two, "globe" and "sun shining" are probably not clear to a new user that doesn't know there's races and species.

Yes, icons look great! I'm kind of averse to merging graphic assets that we know will be replaced. I wouldn't mind seeing these icons in the game, I wonder if @nozmajner has time to voice an opinion on them? Maybe the nose is the weakest of the icons. By the way, is it always the same icon for sex? If so, it's showing a man's face (I hope "it" is a man), but icon shows female.

Also, I assume globe = race, and star = species? If so, maybe it would be more logical to fist choose species, before race? (not that either of them does anything given current assets)

vakhoir:
> I don't agree. Yes, it looks symmetric and nice how the word "random" happens to align nicely in this layout with English. Try it with swedish "Slumpmässig" and it's suddenly not so nice.

Hm, point taken. One way I saw this solved is to trim the word a bit and put a dot at the end to indicate trimming, but I agree it doesnt look so nice.

> I'm kind of averse to merging graphic assets that we know will be replaced.

They're in the svg file, so it's not *that* messy.

> By the way, is it always the same icon for sex? If so, it's showing a man's face (I hope "it" is a man), but icon shows female.

Yeah, I just ripped off @nozmajner's design. We could get of those male/female intertwined icons, but I at least need an example to rip off - I'm not much of an artist.

> Also, I assume globe = race, and star = species? If so, maybe it would be more logical to fist choose species, before race? (not that either of them does anything given current assets)

Er... So the ordering is as you suggested, it's just the meaning of the icons that's inverted. When I was looking for an appropriate one for race I figured that skin color changes by exposure to the sun, so... But your interpretation makes more sense. I guess I'll also add tooltips just so the meaning is absolutely clear.

impaktor:
>> I don't agree. Yes, it looks symmetric and nice how the word "random" happens to align nicely in this layout with English. Try it with swedish "Slumpmässig" and it's suddenly not so nice.

> Hm, point taken. One way I saw this solved is to trim the word a bit and put a dot at the end to indicate trimming, but I agree it doesnt look so nice.

If, we for some reason, really want one of the buttons to have text, then it could be placed elsewhere (above/below image), where there's ample room for the text, regardless of window-size, screen-resolution and language.

I think both `random` and `sex` buttons are currently causing us some trouble since they're not actually flipping through a (non-binary) sequence.

I.e. there is technically no meaning to have a back/forward arrow for sex. But I understand it's that way for UI symmetry, and it looks good this way. That said, one could do the same for `random`, with:

`back arrow` / `icon` / `forward arrow`

but I guess then a user expects clicking "back" will go back to a previously shown face, so you'd have to save the seed/face of each "random face" the user has generated, so one would have to do some book keeping (suggested variable name: `facebook`?).

For "feature completeness" one would probably want the "back" button to be disabled until the player has generated a face (or if it's simpler to implement: circular sequence, so back = forward for the first click).

The `facebook` variable could be deleted when leaving the view/personal info page (if it minimizes the coding hassle).

I think in the best of all worlds, that is the solution I like the most of the ones presented so far, but I realize that's asking for "more stuff" potentially leading the PR to be stalled and bitrot, which I don't want.


Hmm, what happens if user clicks forward twice:

- generate random face A
- generate random face B

then back:

- show random face A

then forward again: Show random face B, or generate new? Or generate new face only when at end of list of stored faces?

I'm getting ahead of my self so I should stop now.

Please share your thoughts on the above.

impaktor:
...I've tested it now, works nicely, appart from tooltip (can be added at a later time).

Maybe players will feel that species button is busted, since we only have one, so maybe comment that line out for now? Or keep it there to inspire contributors to make us some more species. Technically, one could have a small species of just single assets. A "they all look the same to me"-species of robots or something. "Service droids", Cylons, etc. Just thinking out loud about the future.

mike-f1:
> If not, I'll go ahead to finish them, but if so I'll move on to station comm views.

Sorry for the noise, may I ask if you start rewriting [EquipmentMarket](https://github.com/pioneerspacesim/pioneer/blob/master/data/ui/StationView/EquipmentMarket.lua)?

Thx's in advance :smiley:

vakhoir:
> I think both random and sex buttons are currently causing us some trouble since they're not actually flipping through a (non-binary) sequence.

> I.e. there is technically no meaning to have a back/forward arrow for sex. But I understand it's that way for UI symmetry, and it looks good this way. That said, one could do the same for random

I disagree about the `sex` button. Every single face feature works the same way - we have a certain amount of them, and you can use the `previous` / `next` buttons to cycle through them. When you reach the beginning/end of the list clicking the button will send you to the front / back of it. Sex is no different, it's just that there are 2 of them so clicking `next` has the exact same result as clicking `previous`. Functionally, both in terms of code, and user intuition on how to use the UI these buttons will work exactly the same as for any other feature, so there's no reason to design the UI any differently for it.

Not to mention, maybe a modder will want add an alien race with multiple sexes (...and this UI issue ended up making me think of all the possible ways that could happen...), right now we support it more or less out of the box, and keep the interface consistent at the same time.

All this does raise a question about the `player.female` variable... is it used for anything other than the face? Name generation I guess, but I'm wondering if there's a point in keeping it, if the player can just change their name. At the moment I'm handling it the old way (changing the face feature changes the variable).

The `random` button, on the other hand, is different fundamentally, everything you wrote about how to fit it into the current design shows that. Now, your idea is pretty good as far as fitting that functionality into that UI goes. Even the final issue you brought up can be handled relatively easy - the `facebook` has a set size, let's say 5, and when you get to the 5th face and click `next`, or 1st and click `previous`, a new face is generated and saved, and the one from the other side of the list is deleted. The problem is that the user is not going to see what's going on. The way faces are stored/deleted might seem arbitrary if you don't know the maximum amount that is saved, and your current position on the list. We could create a widget that shows all these things, I picture a sliding bar with fancy animated effects... but honestly, is it worth it?

I ended up just going with an image button (I had to make a custom version that does not stretch the icons), and tooltips. Seems clear and simple to me, which is just the way I like it:

pigui-infoview-draft-3

> Sorry for the noise, may I ask if you start rewriting EquipmentMarket?

> Thx's in advance

No problem :smile:

impaktor:
> I disagree about the sex button.

Fair point.

> I ended up just going with an image button (I had to make a custom version that does not stretch the icons), and tooltips. Seems clear and simple to me, which is just the way I like it:

Hmm, so doing both? I guess it's not possible to put the icon on the button? Our UI only supports text, I assume?

> We could create a widget that shows all these things, I picture a sliding bar with fancy animated effects... but honestly, is it worth it?

No, keep things as simple as possilbe.

impaktor:
Also, I just realize, if the "random" has a `<` / `>` that will "trick" the player into thinking it's now selecting yet some other part of the face feature, while in fact deleting all the parts they've chosen and serving a new face, so I no longer stand by my initial suggestion, I guess.

vakhoir:
> I guess it's not possible to put the icon on the button?

What do you mean? I appended a screenshot, the icon *is* on the button. You can also see the tooltip (there are also tooltips if you hover over the other icons).

impaktor:
Oh, that "Random"-text is a tooltip? 

Thought: what makes most sense: have the random bottom last or first?

vakhoir:
> Oh, that "Random"-text is a tooltip?

Yup. Screenshots usually remove the mouse cursor, so it looks confusing, but yeah that's just a tooltip.

> Thought: what makes most sense: have the random bottom last or first?

Hm, last made the most sense to me, but either one would be fine, I suppose. Happy to change it if that's what the public demands.

impaktor:
> Yup. Screenshots

Ah, ok, then it's perfect!

> Hm, last made the most sense to me

__having it last:__
pro: random flipps within the constraints set by previous options (race, species, sex)
con: it deletes the face player just made, might be a slightly negative experience first time it happens

__having it first:__
inverse of the above

I'm fine either way.

OK, then I think this PR is feature-wise done. if anyone wants to do code review speak up now.

src/pigui/Makefile.am
```diff
@@ -0,0 +1,20 @@
+include $(top_srcdir)/Makefile.common
```
impaktor: Btw, please remove all autotools stuff, as suport for that was dropped yesterday.

Also squash commits?

vakhoir: Dropped the autotools stuff.

Turns out squashing wasn't the issue, I must have made a mistake rebasing, so I redid it. The first 2 commits that were here originally are now identical to master.

vakhoir:
> random flipps within the constraints set by previous options (race, species, sex)

Keep in mind, at the moment the only variable I constrained for random generation is sex ( for consistency with the old UI). Personally I'd be leaning towards dropping the constraints entirely, but I'm open to suggestions of how that should work.

src/pigui/Makefile.am
vakhoir: Ok, now it should be fine

Web-eWorks:
Regarding random, I'd suggest storing a temporary list of seeds and allowing the user to select one of the previous seeds via a popup/dropdown menu. Specifically, I'd only store the seed when the user clicks Random, as we don't need to track changes the player has manually made. The list can be cleared when the player closes the view.

Additionally, I'd suggest changing the icon for sex based on the current selection - show a male symbol when a male portrait is selected, and vice-versa for female. This more naturally fits the use of that selector, and frees up people implementing custom species (three-caste insectoids etc.) from needing to create all-in-one images. The only sticking point for this is to track information about the species somewhere and maintain a mapping between sexes and icons representing them, but as we only have the one species we can leave that for a later date.

Otherwise this is looking very good - well done @vakhoir!

impaktor:
> Regarding random, I'd suggest storing a temporary list of seeds and allowing the user to select one of the previous seeds via a popup/dropdown menu

I vote that's for a future PR.

> I'd suggest changing the icon for sex based on the current selection

I agree it feels strange to have mismatch between symbol and face. I guess that problem is that the icon doesn't actually show "sex" but rather a single specific option of it. Could the icon incorporate both symbols at the same time, if that would be easier than flipping between male/female symbol?  

fnuff

Now, that isn't very square, so hard to make into an icon. 

I think we should merge this as is now, unless @vakhoir wants to change something with the female/male icon.

vakhoir:
@Web-eWorks:

> Regarding random, I'd suggest storing a temporary list of seeds and allowing the user to select one of the previous seeds via a popup/dropdown menu. Specifically, I'd only store the seed when the user clicks Random, as we don't need to track changes the player has manually made. The list can be cleared when the player closes the view.

Two problems with that. One is that it's a bit more work then it sounds. Having a face generator that allows you to pick a specific feature means the old system of "just store the face as a seed for random generation" cannot be used (easily), because you can't predict how to change the seed to get a feature changed in a specific way. So either I'd have to expose a random number generator to LUA like the one used in FaceParts.cpp, or have FaceParts.cpp return the FaceDescription struct to LUA. It's not a huge change, but not trivial either, so if we opt for that, I'd do it as a new feature.

The second problem is that having a dropbox with seeds to choose from wouldn't be a very clear way to show the user what he's chosing... I guess we could make a dropbox with previously generated face images?

> The only sticking point for this is to track information about the species somewhere and maintain a mapping between sexes and icons representing them, but as we only have the one species we can leave that for a later date.

That's kind of the reason why I don't like doing it like that. But if we stick to human sexes for now it's a fairly trivial change.

@impaktor 

> Could the icon incorporate both symbols at the same time, if that would be easier than flipping between male/female symbol?

Er... again, that's what it *is* right now :smile: 
Although it doesn't look that good in my opinion, but I warned you in advance my artistic skills aren't great.

vakhoir:
Made another force push to fix a stupid indentation discrepancy

impaktor:
>> Could the icon incorporate both symbols at the same time, if that would be easier than flipping between male/female symbol?

> Er... again, that's what it is right now

Dammit. Wasn't like that in my memory. OK, I say we stop with features now and merge in a few days, if there are no objections to the code.

vakhoir:
I discussed this in some previous PR, but forgot to do it - I added myself to the authors list.

vakhoir:
Oh crap - turns out the way I wrote pigui/views/info-views.lua requires both legacy UI and the pigui InfoView to have the same amount of tabs. So when I dropped the old PersonalInfo view, it caused a crash when switching to the Crew Roster.

Fixed now. The InfoView uses the view ID instead of index.

fluffyfreak:
Merged!

PR #4588, opened 2019-05-15
Friction and sinking

### LINKED ISSUE #2337: Falling through terrain

mike-f1:
**Summary:**

1) Close the closed #2337 (as it is still reproducible in master) tweaking the physical engine
2) Add some friction when colliding so a DynamicBody (or more) will stop.
3) Decouple some includes
4) Address some warnings

**Know issue:**
It can't handle more than x1000 time-accel, at least on my PC.

If you would me to split this PR or to open an issue about a possible
improvement of that feature, let me know.

### DIFF
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -848,6 +848,11 @@ static void hitCallback(CollisionContact *c)
 	// collision response
 	assert(po1_isDynBody || po2_isDynBody);
 
+    // Bounce factor
+    const double coeff_rest = 0.35;
+    // Allow stop due to friction
+    const double coeff_slide = 0.700;
+
 	if (po1_isDynBody && po2_isDynBody) {
 		DynamicBody *b1 = static_cast<DynamicBody *>(po1);
 		DynamicBody *b2 = static_cast<DynamicBody *>(po2);
@@ -855,10 +860,6 @@ static void hitCallback(CollisionContact *c)
 		const vector3d linVel2 = b2->GetVelocity();
 		const vector3d angVel1 = b1->GetAngVelocity();
 		const vector3d angVel2 = b2->GetAngVelocity();
-
-		const double coeff_rest = 0.5;
-		// step back
-		//		mover->UndoTimestep();
 
 		const double invMass1 = 1.0 / b1->GetMass();
 		const double invMass2 = 1.0 / b2->GetMass();
@@ -881,9 +882,9 @@ static void hitCallback(CollisionContact *c)
 		const double j = numerator / (term1 + term2 + term3 + term4);
 		const vector3d force = j * c->normal;
 
-		b1->SetVelocity(linVel1 + force * invMass1);
+		b1->SetVelocity(linVel1 * (1 - coeff_slide * c->timestep) + force * invMass1);
 		b1->SetAngVelocity(angVel1 + hitPos1.Cross(force) * invAngInert1);
-		b2->SetVelocity(linVel2 - force * invMass2);
+		b2->SetVelocity(linVel2 * (1 - coeff_slide * c->timestep) - force * invMass2);
 		b2->SetAngVelocity(angVel2 - hitPos2.Cross(force) * invAngInert2);
 	} else {
 		// one body is static
@@ -898,7 +899,6 @@ static void hitCallback(CollisionContact *c)
 			hitNormal = -c->normal;
 		}
 
-		const double coeff_rest = 0.5;
 		const vector3d linVel1 = mover->GetVelocity();
 		const vector3d angVel1 = mover->GetAngVelocity();
 
@@ -920,13 +920,46 @@ static void hitCallback(CollisionContact *c)
 		const double j = numerator / (term1 + term3);
 		const vector3d force = j * c->normal;
 
-		mover->SetVelocity(linVel1 + force * invMass1);
+		/*
+		   "Linear projection reduces the penetration of two
+		   objects by a small percentage, and this is performed
+		   after the impulse is applied"
+
+		   From:
+		   https://gamedevelopment.tutsplus.com/tutorials/how-to-create-a-custom-2d-physics-engine-the-basics-and-impulse-resolution--gamedev-6331
+
+		   Correction should never be more than c->depth (std::min) and never negative (std::max)
+		   NOTE: usually instead of 'c->timestep' you find a 'percent',
+		   but here we have a variable timestep and thus the upper limit,
+		   which is intended to trigger a collision in the subsequent frame.
+		   NOTE2: works (as intendend) at low timestep.
+		   Further improvement could be:
+				1) velocity should be projected relative to normal direction,
+				   so bouncing and friction may act on the "correct" components
+				   (though that a reason for fails and glitches are frames skipped
+				   because relVel is quitting before any further calculation)
+				2) with time accel at 10000x you end in space... probably in order
+				   for that to works correctly some deeper change should kick in
+		*/
+		const float threshold = 0.005;
+
+		const Aabb &aabb = mover->GetAabb();
+
+		vector3d correction = std::min(std::max(c->depth - threshold, 0.0) * c->timestep, c->depth + aabb.min.y + threshold) * c->normal;
+
+		mover->SetPosition(mover->GetPosition() + correction );
+
+		const float reduction = std::max(1 - coeff_slide * c->timestep, 0.0);
+		vector3d final_vel = linVel1 * (1 - coeff_slide * c->timestep) + force * invMass1;
+		if (final_vel.LengthSqr() < 0.1) final_vel = vector3d(0.0);
+
+		mover->SetVelocity(final_vel);
 		mover->SetAngVelocity(angVel1 + hitPos1.Cross(force) * invAngInert);
 	}
 }
 
 // temporary one-point version
-static void CollideWithTerrain(Body *body)
+static void CollideWithTerrain(Body *body, float timeStep)
 {
 	if (!body->IsType(Object::DYNAMICBODY)) return;
 	DynamicBody *dynBody = static_cast<DynamicBody *>(body);
@@ -944,12 +977,7 @@ static void CollideWithTerrain(Body *body)
 	double terrHeight = terrain->GetTerrainHeight(body->GetPosition().Normalized());
 	if (altitude >= terrHeight) return;
 
-	CollisionContact c;
-	c.pos = body->GetPosition();
-	c.normal = c.pos.Normalized();
-	c.depth = terrHeight - altitude;
-	c.userData1 = static_cast<void *>(body);
-	c.userData2 = static_cast<void *>(f->GetBody());
+	CollisionContact c(body->GetPosition(), body->GetPosition().Normalized(), terrHeight - altitude, timeStep, static_cast<void *>(body), static_cast<void *>(f->GetBody()));
 	hitCallback(&c);
 }
 
@@ -972,7 +1000,7 @@ void Space::TimeStep(float step)
 	// XXX does not need to be done this often
 	CollideFrame(m_rootFrame.get());
 	for (Body *b : m_bodies)
-		CollideWithTerrain(b);
+		CollideWithTerrain(b, step);
 
 	// update frames of reference
 	for (Body *b : m_bodies)
--- a/src/collider/CollisionContact.h
+++ b/src/collider/CollisionContact.h
@@ -12,19 +12,29 @@ struct CollisionContact {
 	vector3d normal;
 	double depth;
 	double dist; // distance travelled to hit point
+	double timestep;
 	int triIdx;
 	void *userData1, *userData2;
 	int geomFlag;
 	//	bool vsStatic;		// true => object 2 was in static, else dynamic
-	CollisionContact() :
+	CollisionContact() : // default ctor
 		depth(0),
 		dist(0),
+		timestep(0),
 		triIdx(-1),
 		userData1(nullptr),
 		userData2(nullptr),
 		geomFlag(0)
-	{ /*empty*/
-	}
+		{}
+	CollisionContact(vector3d p, vector3d n, double de, double t, void *u1, void *u2) : // ctor for collision with terrain
+		pos(p),
+		normal(n),
+		depth(de),
+		timestep(t),
+		userData1(u1),
+		userData2(u2),
+		geomFlag(0)
+		{}
 };
 
 #endif /* _COLLISION_CONTACT_H */
--- a/src/collider/Geom.cpp
+++ b/src/collider/Geom.cpp
@@ -2,9 +2,12 @@
 // Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
 #include "Geom.h"
+
 #include "BVHTree.h"
 #include "GeomTree.h"
 #include "collider.h"
+#include "CollisionContact.h"
+
 #include <float.h>
 
 static const unsigned int MAX_CONTACTS = 8;
--- a/src/collider/Geom.h
+++ b/src/collider/Geom.h
@@ -6,8 +6,8 @@
 
 #include "../matrix4x4.h"
 #include "../vector3.h"
-#include "CollisionContact.h"
 
+class CollisionContact;
 class GeomTree;
 struct isect_t;
 struct Sphere;


fluffyfreak:
Usually I would prefer the include decoupling to be a separate PR but I was able to work out which changes were which.

I'll try to test it tonight (_hopefully I'll get time_) but otherwise definitely before next week.

mike-f1:
No hurry: I have to prepare a commit to solve Travis issues...


Note: Reading logs, I note a couple of strange things, would you take a look?

src/Space.cpp
```diff
@@ -944,12 +965,7 @@ static void CollideWithTerrain(Body *body)
 	double terrHeight = terrain->GetTerrainHeight(body->GetPosition().Normalized());
 	if (altitude >= terrHeight) return;
 
-	CollisionContact c;
-	c.pos = body->GetPosition();
-	c.normal = c.pos.Normalized();
-	c.depth = terrHeight - altitude;
-	c.userData1 = static_cast<void *>(body);
-	c.userData2 = static_cast<void *>(f->GetBody());
+	CollisionContact c(body->GetPosition(), c.pos.Normalized(), terrHeight - altitude, timeStep, static_cast<void *>(body), static_cast<void *>(f->GetBody()));
```
Web-eWorks:
```suggestion
	CollisionContact c(body->GetPosition(), body->GetPosition().Normalized(), terrHeight - altitude, timeStep, static_cast<void *>(body), static_cast<void *>(f->GetBody()));
```
Don't use an uninitialized member in the object's own initialization please. I know this one's a simple copy-paste error, but it's still not good.

Web-eWorks:
@mike-f1 the build is failing because `0.0d` is a non-standard extension to the C++ specification - just use `0.0` as it already refers to a double. Additionally, there are some other issues - some code (mine actually, sorry!) is using the C `abs` (which operates on integers) instead of C++ `std::abs` (which has overloads for floats / doubles); and you're using a member before it's initialized (`Space.cpp:968:42`) which is a big no-no.

mike-f1:
Yes: that is what I was speaking about :)

Anyway I would take a look also on the other warnings,
as it is strange to check for unsigned < 0 :/ (Mac, IIRC)

fluffyfreak:
@mike-f1 if it becomes unstable under high time acceleration then it might be a good idea to disable, or scale the values down as the time acceleration goes up?

A lot of things can become unstable at higher time accelerations.

mike-f1:
@fluffyfreak It's difficult to make it stable because planets are moving and because of rounding errors: you could try yourself only taking off; using autopilot to hold your position and increasing time accel. You will note your ship move slowly. I fear this is unavoidable, so I give up 😕
IMO this could be merged as time accel near planets is limited by default and until now there aren't many missions on planets... But if you want I can split this PR and you can merge the other commits and leave this open so we can talk about if this can be handled and how.

mike-f1:
Regarding scaling down values, I would say I tried but I wasn't able to find any solution. Reasons are that, due to planet movement, small value are unuseful because you end sinking.

Disabling lead to same results...

jaj22:
> @fluffyfreak It's difficult to make it stable because planets are moving and because of rounding errors: you could try yourself only taking off; using autopilot to hold your position and increasing time accel. You will note your ship move slowly. I fear this is unavoidable, so I give up 

Planets aren't actually moving when the ship's in the rotating reference frame. With the exception of atmospheric forces, which were broken in a recent patch, all external forces on a dynamic body will be accounted for in DynamicBody::GetExternalForce().

Autopilot functions will not work correctly in an atmosphere due to the atmospheric forces patch.

mike-f1:
> Planets aren't actually moving when the ship's in the rotating reference frame

Right, I was wondering if that counts...

> With the exception of atmospheric forces, which were broken in a recent patch, all external forces on a dynamic body will be accounted for in DynamicBody::GetExternalForce().

Indeed if I print that value, it is equal to GetGravityForce()... Which is natural: given the ship does not move, those aerodynamical forces should be 0.

> Autopilot functions will not work correctly in an atmosphere due to the atmospheric forces patch.

Well, I would say that the autopilot is doing a great job keeping ship in position, but atmo forces are 0... Thus the only explanation left is "rounding errors" :/

Returning to the subject, if I add this code:
```
		if (mover->IsType(Object::PLAYER)) {
			printf("Timestep: %f; Depth: %f; Correction: %f\n", c->timestep, c->depth, correction.Length());
		}
```
around line 940 in `Space.cpp` and print values I note that `correction` match `depth` when `timestep` is big enough... Thus the ship is repositioned in the "correct" position.
But `depth` (in a single update) reaches values of hundreds meters.

The "instability" is (probably) due to the graphics not in sync with physical engine.
This last issue is IMO the most important thing to address if you would this working at high timesteps.

jaj22:
In retrospect I don't know what actually breaks at high time acceleration from reading this PR. What exactly are you trying to do and what happens?

The renderer won't be relevant unless it's purely a visual issue. Physics just does its thing at 60Hz regardless of the renderer.

mike-f1:
> In retrospect I don't know what actually breaks at high time acceleration from reading this PR.
What exactly are you trying to do and what happens?

Basically (in my PC) the ship will explode if I click x10000 time accel (with ctrl pressed, of course).
At the same time, clicking x1000 time accel, I can see some glith but ship will not explode.
And the above aren't an acceptable behaviour (isn't?).

Thus I'm working in order to understand:
1) what is triggering ship explosion;
2) why I see glith;

> The renderer won't be relevant unless it's purely a visual issue. Physics just does its thing at 60Hz regardless of the renderer.

Putting a printf-line just in physical update (`Space.cpp:988`), I found an explanation for that thing: it seems that `hitCallback` is called only once every 2 frames... Thus the rendering may render a ship deep inside a planet.
A first try I did is to change the value of correction in line 939 from `c->depth` to `c->depth * x` where x is a number less than 1. But this change has no effect on the behaviour: the glitch still there.

I will work further on these issue, investigating why the `hitCallback` isn't called every frame
and why the ship explode.

@jaj22, @fluffyfreak are this behaviour reproducible on your pc or only on mine?

jaj22:
> Basically (in my PC) the ship will explode if I click x10000 time accel (with ctrl pressed, of course).
> At the same time, clicking x1000 time accel, I can see some glith but ship will not explode.
> And the above aren't an acceptable behaviour (isn't?).

Yes, but in what scenario? Sitting on station pad? Landed rough? Hovering with setspeed? Something else? I have no idea what to test at the moment.

mike-f1:
Sorry: not landed but on ground (you could also try jettisoning some cargo while hovering, should be the same)...

A save:

[Friction_Save.gz](https://github.com/pioneerspacesim/pioneer/files/3203849/Friction_Save.gz)

Some (early) conclusions:
1) Explosion is triggered in `if (!OnCollision(po1, po2, c, -relVel)) return;`: when timeaccel is big, it seems `GetVelocity` is getting big values leading to `OnDamage` and explosions
2) Improved a little the `correction` calculation:
`vector3d correction = std::min(std::max(c->depth - threshold, 0.0) * c->timestep, c->depth + aabb.min.y + threshold) * c->normal;`: I try to trigger a collision test every frame diminishing glitches.

jaj22:
If the ground isn't applying any upwards force in the physics calculation (and as far as I know, it isn't), then the ship would drop 1.4km and gain 166m/s of velocity per frame, given earth gravity at 1000x time accel. That's one hell of a collision correction.

I wouldn't worry too much though. It's not the sort of thing that needs to work at high time acceleration. If people can scrape the ground at 1-10x and have it behave reasonably then that's a big improvement.

fluffyfreak:
I'm with @jaj22 on this, we don't expect that situation to always work, forcing time acceleration whilst that close too a planet is allowed but... well those are the consequences you can expect ;)

mike-f1:
Well, now it works "reliably" at 100x :smile:

Just some notes (I added some comments on `Space.cpp`, but I would quote also here):
1) `GetVelocity` is, IMO, the main reason for this to not work at 1000x : a frame skipped is (as @jaj22 note) deadly and a more reliable calculation will end in less glitches and increased survival, because of line
`	if (relVel > 0) return;`
thus as in comments, calculation could be improved
2) For 10000x ...well, physics can't helps: it will be better if a `m_moving` mechanics kick in and/or the entire sequence will be rewritten to disable gravity for a frame.

...I'm not saying "it should be done", just pinning results and considerations...

3) If, instead of a single data as in:
https://github.com/pioneerspacesim/pioneer/blob/5f64a2827726350a124806a2925ed9d0f3962248/src/Space.cpp#L944
you have data of the "a quad"/"patch" under your DynBody, the algorithm may be generalized and could handle hills and slopes.

...And here, with some (**a lot**) of hints, I'm saying "it should be done" :wink: 

(**A thought**: I feel a bit sad because I wonder how much effort will be required in order to start thinking at missions in which players should grab cargo left on planets...)

fluffyfreak:
We'll probably merge this PR as-is (_assuming no other code changes required_) but for future work there should probably be what other physics engines have, an "_at rest_" state.

Similar to a ship that has landed it's where no physics are being done to the object, it's not in motion but in a fixed position. Usually there's a threshold where it's gone so slowly that it has "_come to rest_" and switches into this rest state to reduce processing requirements.

fluffyfreak:
Ok I like it, tested it by bouncing my ship too a halt :D
Merging.

mike-f1:
> We'll probably merge this PR as-is (assuming no other code changes required)

Yes, I would say that even if this still a little change, it could be a good point/commit from
where to begin thinking on. _Sure_ I underestimated this challenge as I discovered a
bit too late (I was thinking to *only* implement a trivial change... obviously I was wrong :/ ).

Anyway, my thoughts are that a better improvement (instead of "refining" this) could be
to have a nicer feedback from terrain: I would not say it but it happens too often to go
under the terrain (even without steep slopes) :(

If the "point 3" is so difficult then any further improvents sounds vane, isn't?

Therefore, If you or @jaj22 would give me some hint on how I can get more data from terrain,
it could be useful :) (...may suffice _seeing_ on what you pass through...)

> Similar to a ship that has landed it's where no physics are being done to the object, it's not in motion but in a fixed position. Usually there's a threshold where it's gone so slowly that it has "come to rest" and switches into this rest state to reduce processing requirements.

Yeah, that is what I was talking on: there are already some handling, you could find yourself
in member `m_isMoving` of `DynBody` which is used to skip `m_externalForce` in updates...
I was wondering if adding a similar member, which is set only during collision calls with high timeaccel
and ever reset during updates of bodies, could be an easy task or a rabbit hole...

...Then there's needs for some changes on AI... Which `is` a rabbit hole, but can be done
later...

mike-f1:
> Ok I like it, tested it by bouncing my ship too a halt :D
Merging.

...Not a _deep_ test, I would say :D

fluffyfreak:
Well for:

> If, instead of a single data as in: pioneer/src/Space.cpp
>
>Line 944 in 5f64a28
>
> https://github.com/pioneerspacesim/pioneer/blob/5f64a2827726350a124806a2925ed9d0f3962248/src/Space.cpp#L944
>
>you have data of the "a quad"/"patch" under your DynBody, the algorithm may be generalized and could handle hills and slopes.

You could sample the height again with 3 or 4 offset positions and then derive the slope using cross products. Which is quite easy to do but isn't done by default because each call to `GetTerrainHeight` goes down to the terrain generation algorithm and that can be computationally expensive.

The other problem, and reason that it sometimes goes below the terrain is that because it sample the `GetTerrainHeight` functions directly they don't align with the visible goemetry of the surface. They shouldn't deviate _too much_ but in practice they're always different by a bit.

mike-f1:
> You could sample the height again with 3 or 4 offset positions and then derive the slope using cross products

What is assuring the sample are on the same patch? I mean: if I get 3 positions, and 1 is in an adjacent patch, then you could have a chine in between.

>  Which is quite easy to do but isn't done by default because each call to GetTerrainHeight goes down to the terrain generation algorithm and that can be computationally expensive.

Well, I don't think to have any clue on numbers (or in anything else ;) ), but I would say that IMO it is better to waste some cycle more here and get a nicer effect than having Ships (which are meters tall) flying under a sea of grass :D

mike-f1:
@fluffyfreak and @jaj22 , I'm asking because I'm working on that but before opening a PR, I would know your opinion.

The question is if could be acceptable a "drill" over all "GeoPatch-tree" in order to find the nearest patch .
Then you could store that patch and when `GetHeight` is called make some calculation using _the data you see_.

Obviously the computational cost is not null (though limited: patches are 4 per layer right now) but
what you obtain is a "perfect behaviour" when you collide. I mean that the collision checks are performed  againts the mesh you see, thus resetting gaps.

Let me know if this may be a solution or if you are aware of something different that may help.

fluffyfreak:
@mike-f1 that's the approach that I'd like to see explored (_i.e. tested and experimented with to see if it works ok_).

However for "drilling into" the `GeoPatch` tree you want to use something like bounding sphere tests to find the leaf node that the collision could be in.

You can do that directly, similarly to how the `DEBUG_BOUNDING_SPHERES` stuff works in your #4598 PR. Or you could build a whole [BVH](https://en.wikipedia.org/wiki/Bounding_volume_hierarchy) in parallel to the `GeoPatch` tree.

Only once you've found the leaf node would you actually test against the geometry of the patch itself.

impaktor:
probably unrelated, but someone implemented [Bullet](https://github.com/bulletphysics/bullet3) physics library for collison detection in pioneer a few years ago, just to learn the bullet library, thus never bothered to finish it up. I have it in my IRC backlog... on my IRC computer that I lost connection to a few days ago.

mike-f1:
I did a couple of commits:

https://github.com/pioneerspacesim/pioneer/compare/master...mike-f1:terrain_height

I'm able to get the patch underneath `campos` (which is more or less player position).

>However for "drilling into" the `GeoPatch` tree you want to use something like bounding sphere tests to find the leaf node that the collision could be in.

Err... I tried with a sort of "Traverse", checking for distances: it is straightforward when underneath patch does not change (thus they maintain order) and _little_ more costly if it changes (thus at some level order will break and you ends recalculating distances...). Don't know if what you suggest is faster or better, but I didn't have time to check.

NOTE: Sure this "Traverse" at the end will ends somewhere else or `GeoSphere` should store the right order of patches for every body in its space.

Next step should be to "extract" polygons data from `GeoPatch` and try to convert it to something our existing physic can handle...


@impaktor : what you suggest is way bigger than the changes I (hardly) try to make... I don't know if I can handle a change so big and even if I could try, I don't know if it will works better than existing.

Web-eWorks:
We have a regression introduced by this change - attempting to bring the default sinonatrix to rest on the Cydonia landing pad without requesting landing will cause it to bounce violently and clip through the model, ending up falling through to the underside of the pad.

I'm bringing this up here because I don't think it needs its own issue, it should be relatively straightforward to adjust the changes to collision to rectify this regression.

mike-f1:
Did another commit [here](https://github.com/pioneerspacesim/pioneer/compare/master...mike-f1:terrain_height)... I'm stuck placing that GeomTree in the right place but the bones are visible.

Any hint or consideration may be helpful, considering that it still require a more computation.

fluffyfreak:
@mike-f1 been quite busy, will try to take a look at this terrain collision tomorrow (Sunday)

mike-f1:
@fluffyfreak thx's... I hope I'm not late: just pushed something.

Anyway it is buggy: it sometime crashes in `GeomTree::TraceRay`, but I don't know if it is related to some memory leak or not.
There are assumptions on collision meshes? I mean: like if they should be closed or not?

jaj22:
@mike-f1 There may be assumptions but that one shouldn't be an issue. Functionally it's a simple edge vs triangle collider. All the fancy stuff is in reducing the check count.

mike-f1:
> @mike-f1 There may be assumptions but that one shouldn't be an issue. Functionally it's a simple edge vs triangle collider. All the fancy stuff is in reducing the check count.

Yeah, thx's: thus probably some memory leaks I left :sweat_smile:
I just realize that when you are near a planet, game is making GeoPatch-es also for near planets (obviously), then next step should be to use objects instead of camera coordinates...

@jaj22, btw: do you feel worried if, for example, I need to create and populate a CollisioneSpace with meshes of patch for every object near a planet?

PR #4595, opened 2019-05-23
Remove Color from Lua instead use LuaColor as a userdata

mike-f1:
Speed up (as usual), LuaPiGui. This time using `Color` (`Color4ub`) trough `LuaColor` as for vectors.

Some profiling:
[Profiling.zip](https://github.com/pioneerspacesim/pioneer/files/3213921/Profiling.zip)

...Hoping someone can tell me if I'm reading these files right (:wink:),
it seems (for example) that `l_pigui_push_style_color` MCycles are more or less halved,
so every other function.

TODO: 
1) "standard" colors could be shared between Lua and C++ world.
2) usage of "lightuserdata" instead of "userdata"
3) it seems in `scenegraph/LuaModelSkin` there were a different way for push/pop-ing Colors I left untouched 
4) ...

After this PR and the others, I would know if saying that PiGui is 3 time faster and it is no more the main concern, I'm wrong.


src/LuaColor.cpp
```diff
@@ -0,0 +1,213 @@
+// Copyright © 2008-2019 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+#include "LuaColor.h"
+#include "LuaUtils.h"
+#include "libs.h"
+
+inline void sanity_check(float r, float g, float b, float a)
```
Web-eWorks: These all need to be `float &` or this function is a no-op.

Additionally, given your usage pattern, I'd suggest returning a Color from this function and using it directly instead of manually constructing a `Color()` after calling `sanity_check`.

src/LuaColor.cpp
```diff
@@ -0,0 +1,213 @@
...
+{
+	r = std::max(std::min(r, float(255.0)), float(0.0));
```
Web-eWorks:
```suggestion
	r = std::max(std::min(r, 255.0f), 0.0f);
```
To reduce keyword noise, please use the `f` suffix literal instead of casting doubles to floats. While the compiler likely will optimize out the cast, it's easier and simpler to use the suffix.

src/LuaColor.cpp
```diff
@@ -0,0 +1,213 @@
...
+{
+	LUA_DEBUG_START(L);
+	double r = luaL_checknumber(L, 2);
+	double g = luaL_checknumber(L, 3);
+	double b = luaL_checknumber(L, 4);
+	double a;
+	if (lua_isnumber(L, 5)) a = luaL_checknumber(L, 5);
+	else a = 255.0;
+	sanity_check(r, g, b, a);
+	Color c = Color(r, g, b, a);
```
Web-eWorks: You appear to be doing a no-op here, and I'd strike it from the function entirely.

src/LuaColor.cpp
```diff
@@ -0,0 +1,213 @@
...
+	if (lua_isnumber(L, 1)) {
+		const double s = lua_tonumber(L, 1);
+		const Color4ub *c = LuaColor::CheckFromLua(L, 2);
+		LuaColor::PushToLua(L, *c * s);
+	} else if (lua_isnumber(L, 2)) {
+		const Color4ub *c = LuaColor::CheckFromLua(L, 1);
+		const double s = lua_tonumber(L, 2);
+		LuaColor::PushToLua(L, *c * s);
+	} else {
+		return luaL_error(L, "Cannot multiply a Color time Color");
```
Web-eWorks: ...Yes you can multiply two colors together. It's vector multiplication, you convert `0..255` to `0..1f`, multiply, and convert back to `0..255`.

Additionally, this function assumes that you're multiplying a color with a color - you could be using a boolean by accident. Please use `luaL_typename` to generate a correct error message.

src/LuaColor.cpp
```diff
@@ -0,0 +1,213 @@
...
+			return 1;
+		} else if (!strcmp(attr, "b")) {
+			lua_pushnumber(L, c->b);
+			return 1;
+		} else if (!strcmp(attr, "a")) {
+			lua_pushnumber(L, c->a);
+			return 1;
+		}
+	}
+	lua_getmetatable(L, 1);
```
Web-eWorks: There's a function called `luaL_getmetafield` that does basically the same thing in an optimized one-liner, though it only supports string keys. Given the specific use case, I think that's appropriate.

We might also want to support accessing values by index and/or swizzling color values. E.g. `col[0] = x, col.bgr = col2.rgb`

src/LuaColor.cpp
```diff
@@ -0,0 +1,213 @@
...
+	}
+	LuaColor::PushToLua(L, *c);
+	return 1;
+}
+
+static int l_color_index(lua_State *L)
+{
+	const Color4ub *c = LuaColor::CheckFromLua(L, 1);
+	if (lua_type(L, 2) == LUA_TSTRING) {
+		const char *attr = luaL_checkstring(L, 2);
```
Web-eWorks: This is redundant as you've already checked the type of the value - there's no possible way this will error out. Additionally, you should check that the length of the string = 1, and gate on that as well.

src/LuaColor.cpp
```diff
@@ -0,0 +1,213 @@
...
+		return luaL_error(L, "Cannot multiply a Color time Color");
+	}
+	return 1;
+}
+
+static int l_color_new_index(lua_State *L)
+{
+	Color4ub *c = LuaColor::CheckFromLua(L, 1);
+	if (lua_type(L, 2) == LUA_TSTRING) {
+		const char *attr = luaL_checkstring(L, 2);
```
Web-eWorks: See my comment on `l_color_index` for this function as well.

src/LuaPiGui.cpp
```diff
@@ -338,13 +328,13 @@ static int l_pigui_lineOnClock(lua_State *l)
 	const double hours = LuaPull<double>(l, 2);
 	const double length = LuaPull<double>(l, 3);
 	const double radius = LuaPull<double>(l,4);
-	const ImColor color = LuaPull<ImColor>(l, 5);
+	const Color color = LuaPull<Color>(l,5);
 	const double thickness = LuaPull<double>(l, 6);
 	if (lua_type(l, 1) != LUA_TNIL) {
 		const vector2d center = LuaPull<vector2d>(l, 1);
-		lineOnClock(center, hours, length, radius, color, thickness);
+		lineOnClock(center, hours, length, radius, ImColor(color.r, color.g, color.b, color.a), thickness);
```
Web-eWorks: Please do not do this. You are uselessly duplicating effort that **should** be in the `LuaPull<ImColor>` function instead. I already have to clean up `ImVec2`, please don't add to my workload. Instead, redefine `pi_lua_generic_pull(lua_State*, int, ImColor&)` to pull a `Color` from Lua and convert it to a `ImColor`. The compiler should completely optimize the overhead involved in calling the conversion functions and it's much less work for anyone who later has to implement functions passing Colors to ImGui (i.e. _me_).

src/LuaColor.cpp
mike-f1: Ohoh: don't drink&code :/ Thank's!

src/LuaColor.cpp
mike-f1: Np

src/LuaColor.cpp
mike-f1: > ..Yes you can multiply two colors together. It's vector multiplication, you convert 0..255 to 0..1f, multiply, and convert back to 0..255.

...so use it at your own risk :wink:  

there's lines of code there:
https://github.com/pioneerspacesim/pioneer/blob/4c51f4eff542fde17bde94b9ccb6209da3a57f34/src/Color.h#L114

...probably another hint for future improvement, isn't?

> Additionally, this function assumes that you're multiplying a color with a color - you could be using a boolean by accident. Please use luaL_typename to generate a correct error message.

...I never seen so many checks around... This could be the first time, but then someone should check types for all Lua* types: sure a lot of big trees of if-then-else will kick in :smile:

src/LuaColor.cpp
mike-f1: > There's a function called luaL_getmetafield that does basically the same thing in an optimized one-liner, though it only supports string keys. Given the specific use case, I think that's appropriate.

Done;

> We might also want to support accessing values by index and/or swizzling color values

Really? ...there are use-cases for all these things? ...Sure I miss them at all, may I ask where these use cases are?

src/LuaColor.cpp
mike-f1: > This is redundant as you've already checked the type of the value

Copy-paste error: `LuaVector` has an `else` branch...

>Additionally, you should check that the length of the string = 1, and gate on that as well.

I don't think there's needs for this.

src/LuaPiGui.cpp
mike-f1: >  I already have to clean up ImVec2

...Well, probably not a work that must be done...

Anyway, this PR sure will be merged after pull #4576, thus let's talk of this on that PR
and note I simply followed what I suggested there, waiting for answers.

I would also say that these indirections and layers are a possible reason why these table aren't been spotted before, thus my concern.

mike-f1:
Addressed (more or less) everything I feel I should address...
This should be merged after #4576 when there's a direction on whether to implement changes on how "unusual" types should be handled. I don't care anymore on how it should be done, I only hope I explain my reasons, then I will copy-paste what should be copy-pasted for correctness and style of code.

Thanks @Web-eWorks for your review: you make things works better (retroactively also!), though I don't think I have to implement everything because... Well, Rome didn't became an Empire in a day ;)

src/LuaColor.cpp
```diff
@@ -0,0 +1,212 @@
...
+	if (lua_isnumber(L, 1)) {
+		const double s = lua_tonumber(L, 1);
+		const Color4ub *c = LuaColor::CheckFromLua(L, 2);
+		LuaColor::PushToLua(L, *c * s);
+	} else if (lua_isnumber(L, 2)) {
+		const Color4ub *c = LuaColor::CheckFromLua(L, 1);
+		const double s = lua_tonumber(L, 2);
+		LuaColor::PushToLua(L, *c * s);
+	} else {
+		return luaL_error(L, "Something strange has been happened in 'LuaColor', check your data!");
```
Web-eWorks:
```suggestion
		return luaL_error(L, "Attempt to multiply a Color with an incompatible value (%s * %s)", luaL_typename(L, 1), luaL_typename(L, 2));
```

src/LuaColor.cpp
Web-eWorks: > I don't think there's needs for this.

It saves four invocations of `strcmp` every time someone calls `Shade` or `Tint`, which is going to happen at least once per frame(and likely tens or hundreds of times). Actually, use `lua_tolstring(l, const char *str, size_t *len)` and just check `len` to see if it's a single-character string - then you don't even need to `strcmp`, you can just check `attr[0] = 'r'` directly, saving function-call overhead.

src/LuaColor.cpp
Web-eWorks: >Done
You sure? It doesn't seem to have made it in.

As far as use cases go, I forgot to stress the _might_; we're not using them now, so it's something to consider for later.

src/LuaPiGui.cpp
Web-eWorks: The fact that these were made with tables was most likely due to the fact that it's code that's been around for six years, and nobody had time or willingness to reinvent the wheel at that point.

src/LuaColor.cpp
mike-f1: How you can call this method with a zero-length string, given Lua parser?

Web-eWorks: ...I'm not sure if you're talking about `attr[0]` or checking that it's a single-character string; arrays in C are zero-indexed, and the single-character check is specifically to speed up method lookups for functions like Shade or Tint. We're not concerned with zero-length strings but with strings > 1.

Web-eWorks: Specifically, the reason we check the string length by calling `lua_tolstring` instead of `luaL_checkstring` is because as-written, calling code can use `Color().redABCDDESLDGS` and the current `__index` method will think that you've written `Color().r`. If you use the single-character comparison, you _must_ check the string length.

mike-f1: > you must check the string length.

I don't think so: I can show you a lot of examples here and there on code in which you can ends with CTD or strange behaviour because of this kind of errors not handled so well in C.
...And not in a "critical code" like this, called _a lot_ each frame.

Thus I'm wondering if there are reasons for you to ask for so many changes:
I changed what you asked for inside reasonable limits (...color "swizzling"?) l but
I begin to suspect there may be something else you want to say.

Let me know...

Web-eWorks: Okay, look. I'm not trying to badger you for unreasonable demands, just trying to help you write better, less-error-prone code. Here's an example of that function rewritten to catch the most useful edge cases:

```C++
const Color4ub *c = LuaColor::CheckFromLua(L, 1);
size_t len = 0;
const char *attr = NULL;
if (lua_type(L, 2) == LUA_TSTRING)
    attr = lua_tolstring(L, 2, &len);
if (attr != NULL && len == 1) {
    if (attr[0] == 'x') {
        ...
```

This does two things - first of all, it reduces the overhead involved with calling `luaL_checkstring`, and secondly it correctly handles cases where you might _intend_ to be looking up a method called e.g. `redify`, and do not want to get the red channel of the color instead.

As an aside, give [the wiki article on swizzling](https://en.wikipedia.org/wiki/Swizzling_(computer_graphics)) a read.

Web-eWorks:
Once the issue with the `__index` metamethods is resolved, this PR looks good for merge to me. Travis seems to think it works as well, so only cursory tests should be needed.

fluffyfreak:
Ok so this is ready for someone to review now? @mike-f1

mike-f1:
>Ok so this is ready for someone to review now?

oh, I thought it was been _already_ reviewed... Anyway yes.

fluffyfreak:
@mike-f1 yes it's been reviewed but I'm not that good with the Lua stuff so it's a chance to get others to speak up.

I'm afraid that there's a conflict with `info-view.lua` now due to another PR I just merged. If you're happy to fix that I'll merge it asap

PR #4605, opened 2019-06-01
Remove usage of AAbb in collision, remove noises if speed is less tha…

mike-f1:
…n 1m/sec

Fix what @Web-eWorks notes in #4588

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -493,10 +493,12 @@ bool Ship::OnDamage(Object *attacker, float kgDamage, const CollisionContact &co
 			if (Pi::rng.Double() < kgDamage)
 				SfxManager::Add(this, TYPE_DAMAGE);
 
-			if (dam < 0.01 * float(GetShipType()->hullMass))
-				Sound::BodyMakeNoise(this, "Hull_hit_Small", 1.0f);
-			else
-				Sound::BodyMakeNoise(this, "Hull_Hit_Medium", 1.0f);
+			if (dam > float(GetShipType()->hullMass / 1000.)) {
+				if (dam < 0.01 * float(GetShipType()->hullMass))
+					Sound::BodyMakeNoise(this, "Hull_hit_Small", 1.0f);
+				else
+					Sound::BodyMakeNoise(this, "Hull_Hit_Medium", 1.0f);
+			}
 		}
 	}
 
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -943,9 +943,7 @@ static void hitCallback(CollisionContact *c)
 		*/
 		const float threshold = 0.005;
 
-		const Aabb &aabb = mover->GetAabb();
-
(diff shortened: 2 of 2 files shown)


src/Ship.cpp
```diff
@@ -493,10 +493,12 @@ bool Ship::OnDamage(Object *attacker, float kgDamage, const CollisionContact &co
 			if (Pi::rng.Double() < kgDamage)
 				SfxManager::Add(this, TYPE_DAMAGE);
 
-			if (dam < 0.01 * float(GetShipType()->hullMass))
-				Sound::BodyMakeNoise(this, "Hull_hit_Small", 1.0f);
-			else
-				Sound::BodyMakeNoise(this, "Hull_Hit_Medium", 1.0f);
+			if (dam > float(GetShipType()->hullMass / 1000.)) {
```
fluffyfreak: Shouldn't that be a less than?
> `if (dam < float(GetShipType()->hullMass * 0.0001)) {`

Because it's a miniscule amount of damage that you don't want to hear rather than a huge amount of damage?

src/Space.cpp
```diff
@@ -943,9 +943,7 @@ static void hitCallback(CollisionContact *c)
 		*/
 		const float threshold = 0.005;
 
-		const Aabb &aabb = mover->GetAabb();
-
-		vector3d correction = std::min(std::max(c->depth - threshold, 0.0) * c->timestep, c->depth + aabb.min.y + threshold) * c->normal;
+		vector3d correction = std::min(std::max(c->depth - threshold, 0.0) * c->timestep, c->depth + threshold) * c->normal;
```
fluffyfreak: I need to check but won't this mean that other ships now either don't sit correctly on the pad, or won't look correct landing?

This might be the ship that needs fixing, not the code.

src/Ship.cpp
mike-f1: I read it as "If damage is _more_ than 'amount' go in parentheses" and then it choose if play big or medium sound...

Where 'amount' is ship mass in [g]: `dam` is [Kg]*[m/sec]*[m/sec], but is multiplied at the function begin in [g], thus this is checking if the speed is above 1 m/sec.

What I try to avoid is to have that little sound played continuosly even when Ship is more or less firm... Don't know, but trying these things I started hating that sound :D

src/Space.cpp
mike-f1: Screenshot from 2019-06-01 13-26-33

IIRC there were some issue due to shields meshes but when you are landed  then collision test is disabled.

Anyway I just landed at 1x, 10x and 100x : it seems no issue here...

mike-f1: PS: `AAbb.min.y` is negative, thus breaking things because `std::max` returned a negative value.

src/Ship.cpp
fluffyfreak: ah I see, I fail at reading comprehension today!

PR #4606, opened 2019-06-01
Basic Atmospheric heating re-enabled

### LINKED ISSUE #4594: Fix atmospheric heat calculations
I have realized, very belatedly, that I disabled atmospheric heating while I was trying to nail the calculations for atmospheric flight. It, uh... never got rewritten, and fluffy merged the PR without noticing.

I'm not able to work on that right now, but if this issue goes on long enough without a resolution, I'll take a crack at implementing accurate heating calculations that work with the new flight model. Otherwise, I'd really appreciate it if someone would Ctrl+F the `TODOs` in `Ship.cpp` and fix atmo heating so we can turn it back on.

fluffyfreak:
This is based on the old g-force system, I simply re-enabled it.
Then I made it damage the ship rather than just instantly `Explode()` it like it did previously!

There were some other warnings and they snuck in with the 2nd commit.

Fixes #4594

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -1046,10 +1046,10 @@ double Ship::GetHullTemperature() const
 	int atmo_shield_cap = 0;
 	const_cast<Ship *>(this)->Properties().Get(R"(atmo_shield_cap)", atmo_shield_cap);
 	if (atmo_shield_cap && GetWheelState() < 1.0) {
-		return dragGs / (300.0 * atmo_shield_cap);
+		return dragGs / (150.0 * atmo_shield_cap);
 	}
 	else {
-		return dragGs / 5.0;
+		return dragGs / 15.0;
 	}
 }
 
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -363,7 +363,7 @@ void Ship::SetPercentHull(float p)
 
 void Ship::UpdateMass()
 {
-	SetMass((m_stats.static_mass + GetPropulsion()->FuelTankMassLeft()) * 1000);
+	SetMass(((double)m_stats.static_mass + GetPropulsion()->FuelTankMassLeft()) * 1000);
 }
 
 template <typename T>
@@ -571,7 +571,7 @@ void Ship::Explode()
 	ClearThrusterState();
 }
 
-bool Ship::DoCrushDamage(float kgDamage)
+bool Ship::DoDamage(float kgDamage)
 {
 	if (m_invulnerable) {
 		return true;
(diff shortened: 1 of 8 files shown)


src/Ship.cpp
```diff
@@ -1035,22 +1035,16 @@ void Ship::TimeAccelAdjust(const float timeStep)
...
-double Ship::GetHullTemperature() const
-{
 	// TODO: fix this to properly account for heating due to air friction instead of G-force.
-	return 0.0;
 	double dragGs = GetAtmosForce().Length() / (GetMass() * 9.81);
 	int atmo_shield_cap = 0;
-	const_cast<Ship *>(this)->Properties().Get("atmo_shield_cap", atmo_shield_cap);
+	const_cast<Ship *>(this)->Properties().Get(R"(atmo_shield_cap)", atmo_shield_cap);
 	if (atmo_shield_cap && GetWheelState() < 1.0) {
 		return dragGs / (300.0 * atmo_shield_cap);
```
Web-eWorks: This may be an existing issue, but even with the interim calculations, having an atmospheric shield equipped means you can do pretty much any maneuver in < 1Atm at any speed without consequences, and flying even an aerodynamic ship like the Wave without a shield results in near-instant death. I'd suggest changing the high multiplier from 300 to 150, and the low multiplier from 5 to 15. Some testing shows that these values should allow careful pilots to slowly fly ships without shielding, and make crazy maneuvers like deep aerobraking at 15+km/s more dangerous without a heavy shield.

mike-f1:
@fluffyfreak, what kind of warnings do you get when you compile?

...So I can change my settings and avoid all that work on your side ;)

src/Ship.cpp
fluffyfreak: Ok, i'll try some tweaks

fluffyfreak:
@mike-f1 the `CollisionContact` was being redeclared as a `class` instead of `struct`, and there were some missing members from the initialiser list. Namely the `distance` and `triIdx`.

fluffyfreak:
This probably wants to become configurable from an ini file but there's lots of things like that so it can be a different PR.

mike-f1:
>This probably wants to become configurable...

...If I'm late it is not a problem, but shouldn't `atmo_shield_cap` be used directly?
I mean that it is already "loaded" from `equipments`, thus it is already a customizable property:
instead of multiplying it of 150, give it a value of 150 in `equipments`...

Or even simpler: 
```
return dragGs / (15.0 * (atmo_shield_cap + 1.0 + GetWheelState())

```
...Where `atmo_shield_cap` provide a protection which is "times" the normal.
And the value of normal atmo shields should be 9 and 18 for heavy (if my calculation aren't wrong...)

NOTE: IIRC GetWheelState is 1 when up and 0 when down, or you should change that

fluffyfreak:
@mike-f1 it might be an idea to redo that later but the `atmo_shield_cap` is used elsewhere too

fluffyfreak:
I prefer not to scoop up other changes into a single task PR, this ones already gone over the top a bit

mike-f1:
:ok_hand:

Issue #746, opened 2011-10-28
lights


dewasha:
I like the lights of the ship could disable and enable

thanks

Web-eWorks:
While I too would love a better implementation of ship navlights and general lighting, the realization of that dream lies well in the future.

Closing as unimplemented; feature requests and discussion should go on the [forum](https://pioneerspacesim.net/forum), as we use Github for tracking actively-developed features and issues with currently-implemented functionality.

If you'd like to re-open this, please instead open a new PR with an implementation of the proposed functionality, thank you!

PR #4608, opened 2019-06-01
Change landing lights according to #4, rework the NavLights class.

### LINKED ISSUE #4: Landing light changes on landing pads
Landing pad lights should be red when occupied, yellow when vacant, green when assigned to the player

Web-eWorks:
Closes #4.

Issue #4 is one of our longest standing issues, having remained open for five years. It's also stupendously simple to fix, once you figure out which class owns the navlights for spaceports.

I've taken the suggested light colors and tweaked them a little (blue for landed, yellow for takeoff, green for landing, off for vacant), as well as setting custom blink patterns for each stage. I've tested this on Cydonia and Mars High, the former of which works perfectly, while the latter doesn't actually have any pad lights at all.

I've also changed the way we store the light groups by using a `std::map` instead of a vector of vectors - this simplifies things immensely and also provides an extremely minor performance boost.

### DIFF
--- a/src/NavLights.cpp
+++ b/src/NavLights.cpp
@@ -26,7 +26,6 @@ static vector2f m_lightColorsUVoffsets[(NavLights::NAVLIGHT_YELLOW + 1)] = {
 	vector2f(0.5f, 0.5f)
 };
 
-typedef std::vector<NavLights::LightBulb>::iterator LightIterator;
 static vector2f get_color(Uint8 c)
 {
 	return m_lightColorsUVoffsets[c];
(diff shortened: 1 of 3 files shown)


Web-eWorks:
As an addendum, @fluffyfreak why in the world does the navlight rendering not just use a single emissive texture and tint it in the shader? Just pass a color attribute to the vertex buffer and shade like that...

impaktor:
Is the player treated special? I.e. if player lands same time as someone else, both pads have the same lights?

Web-eWorks:
I could probably treat the player specially, though A) there's never enough traffic to have more than one or two ships land at the same time, and B) stations that you have to manually land at give a UI marker for the pad anyways, so there's no confusion.

impaktor:
Just wanted to know. In general, player shouldn't be "special" - so good, I guess.

The fact that there are few ships is something we want to change, but fot that we need:
-1 queue system for docking
-2 some cheap way of representing ships far away in the system

fluffyfreak:
@Web-eWorks I didn't write it, NavLights is _old_ code. Might as well ask why we don't avoid that nasty hard cutoff where they intersect the pad by using multiple "shells" of textured light billboards and additively blend them so the lights have a deeper effect that's still relatively cheap to render.

Web-eWorks:
@fluffyfreak I figured that. Still trying to wrap my head around the scenegraph code (what, does every Model accumulate a list of transforms with each `Render()` call and render all of them at once with `glDrawInstanced` or something?) but it is something I want to tweak at some point.

fluffyfreak:
I'm confused by this comment:
> I've tested this on Cydonia and Mars High, the former of which works perfectly, while the latter doesn't actually have any pad lights at all.

Testing it in the regular build everywhere has NavLights just fine, Mars High included. Is this somethign that's gotten broken with this PR?

Web-eWorks:
@fluffyfreak The models for the orbital spacestations (including Mars high) just have diamonds of `navlight_red`/`navlight_green` on each pad; they don't have actual `navlight_pad` tags. The spaceport structure at Cydonia has properly-setup `navlight_pad` tags, so it shows the context-aware NavLight colors. Nothing's broken, the models just never supported contextual pad lights.

mike-f1:
[Just as a side note](https://github.com/pioneerspacesim/pioneer/issues/746) ...

And if I may: @fluffyfreak , the rendering system can "produce" lights? [I mean like these](http://blog.aveoengineering.com/wp-content/uploads/2018/02/GE-landing1.jpg)

fluffyfreak:
@mike-f1 no not yet

mike-f1:
> @mike-f1 no not yet

...Just pinned :smile:

PR #4617, opened 2019-06-08
Star rendering optimisation

fluffyfreak:
Minor rendering optimisation, stars were constantly rebuilding their halos per-frame, but using almost exactly the same inputs. The radius varied depending on the distance from the star.

Instead this build a unit (1) sized halo and then just scales the rendering of it per-frame, thus avoiding all of the creation and destruction of the triangle buffers.

A couple of other minor alterations crept in too.


src/Star.cpp
```diff
@@ -1,14 +1,17 @@
 // Copyright © 2008-2019 Pioneer Developers. See AUTHORS.txt for details
 // Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
+#include <memory>
 #include "Star.h"
 
+#include "Frame.h"
```
mike-f1: Is it needed?

src/Star.cpp
```diff
@@ -1,14 +1,17 @@
...
 #include "Pi.h"
 #include "galaxy/StarSystem.h"
 #include "graphics/Graphics.h"
 #include "graphics/Renderer.h"
 #include "graphics/RenderState.h"
 #include "graphics/VertexArray.h"
+#include "graphics/VertexBuffer.h"
```
mike-f1: Same as above

fluffyfreak: Cleaned up these headers now. The frame one was a leftover from some debug rendering I was testing.

PR #4614, opened 2019-06-05
Pigui Model Spinner, move ShipInfo to pigui.

Web-eWorks:
It's what it says on the tin. @fluffyfreak should probably review the rendering code and make sure I'm not doing something stupid, but otherwise this one's ready to go.

I've done something a little _special_ with the infoview - I've written it so that it's ready to be used in the new sidebar layout I've been hashing out with @nozmajner; all you've gotta do is turn `_OLD_LAYOUT` off and call it from a parent which handles the sizing. I'm going to further extend this with a few custom widgets to achieve the look of the prototypes, but that's for another PR.

@vakhoir on the first window opening, the tab-view wants to show the newUI tab underneath the pigui tab - I'm leaving you to figure out what's going wrong with that code.

PR #4620, opened 2019-06-10
Fix hiding/showing new tab-view

vakhoir:
Fixes the bug described in #4614.
New tab-view now checks every frame whether the new or old tabview should be displayed


impaktor:
nitpick for @vakhoir https://help.github.com/en/articles/closing-issues-using-keywords

I.e. `fixes #4614` or `closes #abcd` etc. does the trick.

PR #4626, opened 2019-06-18
Improve Pioneer startup times, part 1.

Web-eWorks:
This PR restores the ability to build Pioneer with the profiler enabled (by calling `./bootstrap -DENABLE_PROFILER=1`), instruments a number of significant functions with `PROFILE_SCOPED`, and lowers Pioneer's startup time from 10 seconds to 8 seconds in release configuration.

I've rewritten the `Serializer` classes and moved them to the scenegraph/ module, as the collision and scenegraph code are the only files which use it. Previously, the serializer involved calling a non-inlineable function for each byte read and written, which was _extremely_ slow. `StaticGeometry::Load` went from 1.3 seconds over 500 calls to a mere 75ms, and several other functions received a commensurate boost!

~~This does have a slight downside. To gain such a performance boost, we must enforce a constraint that two vertex buffers with the same attribute bits (position, normal, uv0, tangent) have the same ordering of vertex data elements, even across renderer backends. Thus `attrib[0]` must be `position` with `offset = 0`, `attrib[1]` must be `normal` and have `offset = 12`, `attrib[2]` must be `uv0` and have `offset=24`, etc.~~

~~This should not present an issue as I understand it; even if we were to implement a DX11 / DX12 / GLES 3 / Vulkan backend, we should be able to easily and effortlessly satisfy this constraint.~~

EDIT: I've lifted this constraint. When building in release mode, the inlined operations are fast enough to handle significant amounts of meshes with little issue.

There's one other caveat, and that is that the current implementation is little-endian only and depends on struct alignment and packing staying the same across architectures. To that end, I've added some `static_assert`s to Serializer.h to enforce this - I'd encourage testing this change as widely as possible before merging to ensure that this is a sane decision.

I've also removed the `#pragma pack()` from vector3.h - I'd appreciate it if someone knowledgeable can confirm whether they are actually still needed.


impaktor:
> This PR restores the ability to build Pioneer with the profiler enabled (by calling ./bootstrap -DUSE_PIONEER_PROFILER=on)

Should be documented somewhere. COMPILING.txt? Maybe add a misc section? Or wiki?

pcercuei:
What's the problem with endianness? Can't you just byte-swap during the deserialization?

Web-eWorks:
@pcercuei there are a couple of issues. First of all, there's no standard library routine to convert endianness (and while I could use SDL's macros it'd still be painful); the way the serializer is implemented it doesn't decompose data structures to component integers / floats but simply `memcpy()`s them; and the most important one: pioneer does not target any big-endian architectures. We support x86/64 little-endian only, with the occasional overture towards supporting ARM in little-endian mode.

It's not horribly difficult to conditionally replace the serialization routines on a big-endian architecture, and there is no particular reason the serializer can't run on a big-endian architecture, but at present the point is moot.

Additionally, this is simply a stopgap measure - I'm tentatively planning to replace the SGM format (and Serializer class) entirely, likely with GLTF. When I do so, I'll rewrite the model serialization code to properly account for endianness, component layout, etc.

Web-eWorks:
On futher reflection, I've made some changes. I've rolled back the change to `StaticGeometry` - we'll still store compiled vertex data per-component instead of enforcing identical layouts on both ends. I've also changed some other functions around to prefer storing data on a per-component basis rather than as a chunk of memory; this will produce better results if alignment or padding is different per-platform.

Once built in release configuration, the new serializer class gets acceptable performance, taking only 64ms to load all of the geometry data in our model files. I've also replaced our `miniz` INFLATE decompression of model data with lz4 compression/decompression - this cuts our compression times down significantly (20s instead of 60s), and we decompress our entire model data payload in about 400ms.

Decompression speed should experience another significant drop once I think my way past using a `std::string` as the target buffer - currently we're decoding a 16KB block and appending to a string, hardly the most efficient method.

CMakeLists.txt
```diff
@@ -155,8 +160,12 @@ pkg_check_modules(ASSIMP REQUIRED assimp)
 pkg_check_modules(SIGCPP REQUIRED sigc++-2.0)
 pkg_check_modules(VORBISFILE REQUIRED vorbisfile)
 
+add_subdirectory(contrib/lz4)
+add_subdirectory(contrib/entt)
```
mike-f1: ...not used (yet)?

Web-eWorks: Nope, that came in accidentally; I'll fix that up with the next commit.

mike-f1: Np. Just a question: are you working on that? Because if you start using that too early, it may make things more difficult to manage as current implementation is incomplete and not clear.

impaktor:
Is the miniz library still used by pioneer after this PR?

Web-eWorks:
Yes. We still use miniz internally to gzip savefiles, as well as (with the next commit) to maintain backwards compatibilty with .sgm files distributed as part of mods.

CMakeLists.txt
Web-eWorks: I'm using some foundational features of entt (families, signal handlers) internally in prototyping, but I haven't seriously started moving things over yet; I'm working on getting a better grasp of the internals before I do.

pcercuei:
Why miniz and not simply zlib by the way?

Web-eWorks:
Legacy decisions, acceptable performance, one less dependency to wrangle on Windows? Take your pick. I honestly don't know; I think it's been around for a number of years.

pcercuei:
Zlib is a dependency of libpng, so it's built and loaded on Windows already.

CMakeLists.txt
mike-f1: Got it. If I may share a though, what I would see (or try by myself...) is a change here:
https://github.com/pioneerspacesim/pioneer/blob/be3d9bed7d0826ac50e2fe5d925c7a643f8dbeb5/src/collider/CollisionSpace.h#L57-L58

...and here:

https://github.com/pioneerspacesim/pioneer/blob/be3d9bed7d0826ac50e2fe5d925c7a643f8dbeb5/src/Space.h#L111-L112

I don't think entt "as a whole" can be the answer, but its [sparse set](https://github.com/skypjack/entt/blob/master/src/entt/entity/sparse_set.hpp) or a just little veil over `std::vector`, in order to allow erasing things with "swap&pop", may lead to big improvements without a big investment in coding, isn't?

fluffyfreak: "big improvements" only exist if the code is actually required to be performant. The things you just linked too don't appear to be used in performance critical ways.
Only optimise where optimisation is needed.

fluffyfreak:
Why did you remove the `#pragma pack()` from vector3 if you don't know what it will do? 
And did you check the sizes before and after doing so of the float and double versions?

CMakeLists.txt
mike-f1: Ahahah, well, "big improvements" is a `__reserved` concept :wink: ... I just note that a `std::list` of pointers is the least cache friendly structure to use and, actually, it is used by the "main" data structures over the codebase, which are `Body` and `Geoms`.

Thus for bodies any iteration on the whole things means picking objects from here and there in memory (think at `StaticUpdate()`; `Update()`; `Render()` ecc...);

For `Geoms` probably you're right: there's also the `BVHTree`. Please forgive if I'm wrong but I'm new to that code.
However I found at least a "for" which iterate on the whole lists:
https://github.com/pioneerspacesim/pioneer/blob/be3d9bed7d0826ac50e2fe5d925c7a643f8dbeb5/src/collider/CollisionSpace.cpp#L390

> Only optimise where optimisation is needed.

Agreed: to me, after the improvements on PiGui, there's no need for further optimizations, but I feel that these changes, if any further optimisation may become necessary, are "on top" of my list.

mike-f1: NOTE: 
>Agreed: to me, after the improvements on PiGui, there's no need for further optimizations, but I feel that these changes, if any further optimisation may become necessary, are "on top" of my list.

@Web-eWorks , the above doesn't mean that this PR is not welcome, obviously... Just that my brain was out of order :smiley:

Web-eWorks:
@fluffyfreak ...if you'll look at the new `Serializer.h` you'll notice a number of `static_asserts` checking that. Both `double` and `vector3d` align to 8 on all the platforms we care about, with no intermediate padding. Additionally, the way `operator[]` is written, we don't need to manually specify alignment as it automatically uses the alignment of the numeric datatype (float or double).

The only way to make `operator[]` more robust is to write it using `if constexpr` and do compile-time testing for 0/1/2 indicies.

Web-eWorks:
I've restored backwards compatibility with the previous DEFLATE-based sgm compression; any mods distributing SGM-v6 files will remain compatible, but newly-compiled models are now significantly faster.

I don't *think* I can speed the LZ4 decompression any further - changing the intermediate buffer size to 1MB netted an overall 10ms (or 1/35th) improvement, so the issue is likely the small batch datum size and the fact that we're outputting (and therefore copying) to a std::string instead of a raw buffer.

I might take another crack at faster decompression in a further PR, but for now I'm very happy with the speedup; overall this PR gives at least two seconds faster startup in a release configuration, and 3-4 in a debug configuration.

Web-eWorks:
Pinging @fluffyfreak @mike-f1 for final review; barring any major issues I think this PR is ready for merge.

mike-f1:
> so the issue is likely the small batch datum size and the fact that we're outputting (and therefore copying) to a std::string instead of a raw buffer

...Probably not worth the effort, but `std::string` has a `reserve()` method, thus even if you can't get an estimate of the size of data, you could use it to pick a buffer of, say, 4KB directly in ctor and you will avoid all the initial reallocation.

>  I'm very happy with the speedup

Me too :+1:

impaktor:
@fluffyfreak anything missing or needed to be fixed to merged this? Any objections?

Web-eWorks:
@mike-f1
> ...Probably not worth the effort, but std::string has a reserve() method, thus even if you can't get an estimate of the size of data, you could use it to pick a buffer of, say, 4KB directly in ctor and you will avoid all the initial reallocation.

Yeah, the other downside with `reserve()` is that I didn't see a way to manually increment the string length, otherwise I'd skip the intermediate buffer and just decompress directly into the string.

mike-f1:
@Web-eWorks , I just saw you're _writing_ on an empty string (`class Writer` have an empty ctor)...

However if your problem is to shrink or enlarge a string internal buffer, [that method](http://www.cplusplus.com/reference/string/string/reserve/) is doing exactly what you need:
`decompress_buffer.reserve(decompress_buffer.capacity + buffer_len)` should works as
you want, triggering a reallocation of `std::string` internal buffer on each call.

Hope it helps

Web-eWorks:
@mike-f1 it's not the internal buffer, it's that `std::string` doesn't (to my knowledge) expose a method to expand the 'used' length `.size()` of a string without copying from an external buffer. If I only needed to resize the internal buffer for copying, that'd be perfect, but I also need to pass the string around and have `.size()` work as expected.

I *might* be able to `string.resize(newlen, 0)` and then just copy into that, but A: it's an extra write over the entire (potentially 30+MB) buffer, and B: honestly I think the small size (and granularity) of the files is the limiting factor here. There's no 500MB file to get *up* to 4000MB/s speeds with.

Also, doing it this way means we can just read the first X bytes of a file for efficient JIT loading of model data. E.g. we store tags, offsets, etc. in the first part, and the actual GPU vertex data at the end. That way we don't need to have all of our model files resident in memory when we only want one of them.

mike-f1:
> it's not the internal buffer, it's that std::string doesn't (to my knowledge) expose a method to expand the 'used' length .size() of a string without copying from an external buffer

Yes, you're right: they are bulletproof, thus no easy way for an "unitialized expansion"...
Though, what I should have said is that having a precise idea of the length of the buffer means you can
cause reallocation when you decided to and you can also _fill_ it with `append` (instead of copying, which should be slower) and then use the final part of the buffer directly, avoiding temporary.

Not a concern, though.

fluffyfreak:
Will review asap, just busy at moment.

fluffyfreak:
A read only review of the code looks ok, lots of noise in the review with files moving and being added etc.
`Serializer.h/cpp` moving to the scenegraph folder for example is fine, but you haven't used `git mv` so the history of the file looks to have been broken.

I'm away now until Monday, but I'm ok with this being merged if people want too.

impaktor:
> but you haven't used git mv so the history of the file looks to have been broken.

Yeah, this has got to be fixed, some (like @laarmen) relies heavily on commit history, to track changes and understand the code.

laarmen:
True, but in the specific case of renaming, in actuality Git doesn't have
any special treatment for git mv, it's mostly a wrapper for `git add` +
`git rm`. It detects renames when it computes diffs from commit to commit.
This means that if you want Git to be able to follow the history of a file
across renames, the "only" thing you have to do is to rename the file in a
separate commit than the one where you actually modify it, and of course
tell git log you want to `--follow` the renames :)

fluffyfreak:
I'm aware of the moving/renaming issues with Git, which is why I started the whole thread on the forum to which everyone replied: "_use git mv_", which I was already aware of...

Anyway this can be merged far as I'm concerned

impaktor:
I get a merge conflict in CMakeList, ping @Web-eWorks

Web-eWorks:
Ping @impaktor, merge conflict should be fixed.

PR #4633, opened 2019-06-25
Remove erraneous whitespace from README.md

ecraven:


impaktor:
You don't have `(add-hook 'before-save-hook 'whitespace-cleanup)`? I assume it would have nuked all whitespaces in README on first try, if so.

ecraven:
> You don't have `(add-hook 'before-save-hook 'whitespace-cleanup)`? I assume it would have nuked all whitespaces in README on first try, if so.

I didn't create that file, I just fixed the bugs :)

PR #4634, opened 2019-06-26
Support building under VS2017 using CMake & add VS2017 to AppVeyor build matrix

pcercuei:
This patchset makes it possible to build Pioneer under Visual Studio 2017 using nothing but the CMake script; none of the scripts in the "win32" folder are used. The pioneer-thirdparty repository must be cloned in the tree and be up-to-date.

This patchset also adds a Visual Studio build in the build matrix of AppVeyor. The pioneer-setup.exe artifact is still generated using MinGW, as it's been working just fine; the point of this build is simply to detect code that breaks compilation under Visual Studio as soon as it appears.


pcercuei:
The static analyzer fails when parsing Color.h, it complains that the opening bracket is on the same line as the function names. But I didn't change that. This file was not updated to the clang-format rules?

Web-eWorks:
@pcercuei During the period the responsible PR was merged, the static-check process was silently failing due to an upstream change in the OSX CI environment, and the clang-format binary could not be used to format certain headers as it misidentified them as Objective-C.

As you can see, static checks are working again, so we should not see a repetition of this issue.

pcercuei:
Okay. I changed it anyway so that the patch compiles with the clang format rules.

impaktor:
@fluffyfreak shall we merge? This is stuff related to your build-flow...

Once this is merged, and we've confirmed it works fine, shall we then remove the win32 folder? Or do we want to keep it for nostalgic reasons?

fluffyfreak:
Yes, leave the win32 folder for a bit until I've switched over to using CMake full time, then I'll remove it.

PR #4636, opened 2019-07-01
Add pioneer logo to README

impaktor:
Also, add link to windows nightly build, because it should be linked from _somewhere_ at least.

I'll squash the commits later, thought to keep separate if people protest against either change.

See what the logo looks like:
https://github.com/impaktor/pioneer/tree/add_badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@
 
 # Pioneer Space Simulator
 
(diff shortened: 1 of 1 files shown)


impaktor:
OK, removed the nightly build commit, as it would be outdated as soon as #4634 was merged.

PR #4638, opened 2019-07-08
Avoid passing PIONEER_PROFILER macro on command line

pcercuei:
Instead of adding -DPIONEER_PROFILER=1 on the command line, we now
define the WITH_PROFILER macro in buildopts.h.

Now, when toggling this option ON/OFF, only the source files using
buildopts.h will be recompiled (~13 files) instead of the whole project
(~390 files).

The option name changed from PROFILER_ENABLED to WITH_PROFILER to
follow the pattern of already existing options.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,7 @@ endif (MINGW)
 
 option(WITH_OBJECTVIEWER "Include the object viewer in the build" ON)
 option(WITH_DEVKEYS "Include various extra keybindings for dev functions" ON)
+option(WITH_PROFILER "Build pioneer with profiling support built-in." OFF)
 
 list(APPEND SRC_FOLDERS
 	src/
@@ -152,11 +153,6 @@ if (USE_SYSTEM_LIBLUA)
 		add_definitions(-DLUA_BUILD_AS_DLL)
 	endif (WIN32)
 endif (USE_SYSTEM_LIBLUA)
-
-option(PROFILER_ENABLED "Build pioneer with profiling support built-in." OFF)
-if (PROFILER_ENABLED)
-	add_compile_definitions(PIONEER_PROFILER=1)
-endif(PROFILER_ENABLED)
 
 if (MSVC)
 	include(msvc-defaults.cmake)
--- a/buildopts.h.cmakein
+++ b/buildopts.h.cmakein
@@ -7,5 +7,6 @@
 
 #cmakedefine01 WITH_OBJECTVIEWER
 #cmakedefine01 WITH_DEVKEYS
+#cmakedefine01 WITH_PROFILER
 
 #endif /* BUILDOPTS_H */
--- a/contrib/profiler/CMakeLists.txt
+++ b/contrib/profiler/CMakeLists.txt
@@ -2,6 +2,8 @@ project(profiler LANGUAGES CXX)
 
 set(BUILD_SHARED_LIBS OFF)
 
+include_directories(${CMAKE_BINARY_DIR})
+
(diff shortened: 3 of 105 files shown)


Web-eWorks:
@pcercuei there's a major problem. The files using `PROFILE_SCOPED` don't include `buildopts.h`, but rather depend on a global definition of `PIONEER_PROFILER` to enable the actual profiling macros. When swapping profiling options, _all code_ using those macros _must_ be recompiled, or you won't get profiling output from those functions. That's mainly why it's a global define and not in `buildopts.h`.

mike-f1:
> all code using those macros must be recompiled

@Web-eWorks , however, I can't find a reason `Lua`, `glEW` (...in a word: libraries) must be recompiled.

Last time I tried enabling `PIONEER_PROFILER` the entire project _including_ libraries (those internals like graphics or terrain, and even those external) were recompiled.
I'm not an expert, but I higly doubt there's a need for that. I'm wrong?

A good deal is to cut recompilation time from something like 30min to 5/10min, thus some effort
could be spent in doing this...

Can `cmake` be configured to parse files looking for `PROFILE_SCOPED()` string?

Web-eWorks:
@mike-f1 the code in `contrib/` doesn't need to be recompiled, as none of it is instrumented with `PROFILE_SCOPED()`. However, everything in `src/` will, including the `terrain` and `graphics` _modules_ (not libraries).

Enabling profiling is semantically similar to changing from a release to a debug build; you have to recompile all the project files for it to take effect. Similarly, enabling profiling only causes one project-wide recompile, after which normal incremental builds take over.

pcercuei:
Force-pushed. It should work properly now.
```<profiler/Profiler.h>``` now includes ```<buildopts.h>```.

I removed the ```<profiler/Profiler.h>``` include from ```libs.h``` and ```utils.h``` since those are included basically everywhere. Now the C++ files which use the ```PROFILE_SCOPED()``` macro should add the proper include.

With this change, only 114 files are recompiled when toggling the option ON/OFF. It's still about 1/4 of the project.

fluffyfreak:
I added the profiler with the intent of it being a separate build.
So in the Visual Studio project it's:
- release (_full optimisations etc_)
- profile (_same as release but with profiling enabled_)
- pre-release (_same as release but with extra debugging information_)
- debug

That would be my preference, to have it as a separate build target.

impaktor:
OK, so in summary, @Web-eWorks objects to the original PR not working as intended, this was fixed by @pcercuei with a force-push on 9th of july.

@fluffyfreak prefers having profiling be a separate build, so I guess he/you votes "no"?

@Web-eWorks opinions on new force pushed fix?

Personally, I think we should make development and contributions (e.g. compiling) as easy as possible, that said, I don't know which way this should go, __so the one who screams the most wins?__ :stuck_out_tongue_winking_eye: 

in-tough-times-mother-birds-feed-the-healthiest-chick

Web-eWorks:
I'd vote with Fluffy on this one; the profiler is something that should probably be its own build type, but it should be the default development build configuration. Instead of replicating the `#include <profiler/Profiler>` line across the entire project, I'd suggest we use `target_compile_definitions` in the CMakeLists file; this will prevent recompilation of modules in `contrib/` and avoid extra work in instrumenting newly-contributed code.

impaktor:
Sounds like a plan. What ever the default is changed to, we need to update build script (I assume?) and documentation (COMPILING.txt) to reflect the changes, and for how to change out of the default selection.

fluffyfreak:
I'm going to close this one, it's not a bad idea as such but it's not how I ever envisaged the profiler being used

PR #4639, opened 2019-07-11
Update pioneer wrapper script to use CMake

CylonicRaider:
During the Autotools removal in #4511 and #4635, this file was somehow forgotten. Since it is more appealing to use than some opaque `(cd build && make) && build/pioneer`, this PR opts to keep the script and to adapt it to use CMake.

The tangentially related `bootstrap` file is left untouched.

### DIFF
--- a/pioneer
+++ b/pioneer
@@ -2,12 +2,20 @@
 # Executable wrapper script
 # Ensures the program is up-to-date, then hands over
 
-PROGRAM_NAME=$(basename $0)
+set -e
+cd "$(dirname "$0")"
+PROGRAM_NAME="$(basename "$0")"
 
-EXEC_BIN=src/$PROGRAM_NAME
-START_DIR=$(pwd)
-WORK_DIR=$(dirname $0)
+# Create build directory
+[ -d build ] || make build
+cd build
 
-cd $WORK_DIR
-make && exec $EXEC_BIN $*
-cd $START_DIR
+# Generate build files
+[ -f Makefile ] || cmake ..
+
+# Generate binaries
+make
+
+# Run main binary
+cd ..
+exec build/$PROGRAM_NAME "$@"
--- a/pioneer
+++ b/pioneer
@@ -7,15 +7,15 @@ cd "$(dirname "$0")"
 PROGRAM_NAME="$(basename "$0")"
 
 # Create build directory
-[ -d build ] || make build
-cd build
+[ -d build ] || mkdir build
 
 # Generate build files
-[ -f Makefile ] || cmake ..
+[ -f build/CMakeCache.txt ] || ./boostrap
 
 # Generate binaries
-make
+# Users of GNU Make can enable parallel builds on N cores by setting
+# MAKEFLAGS=-jN when calling this script.
+cmake --build build
 
 # Run main binary
-cd ..
 exec build/$PROGRAM_NAME "$@"


pioneer
```diff
@@ -2,12 +2,20 @@
...
-WORK_DIR=$(dirname $0)
+# Create build directory
+[ -d build ] || make build
+cd build
 
-cd $WORK_DIR
-make && exec $EXEC_BIN $*
-cd $START_DIR
+# Generate build files
+[ -f Makefile ] || cmake ..
```
pcercuei: You can run "cmake .." unconditionally.

pioneer
```diff
@@ -2,12 +2,20 @@
...
+
+# Generate binaries
+make
```
pcercuei: Let's make it a bit faster: ```make -j $(nproc --all)```

pioneer
CylonicRaider: Doing so regenerates `buildopts.h`, which, in turn, results in rebuilding some files, relinking Pioneer, and re-running the model compiler. The Makefile automatically re-runs CMake whenever necessary.

pioneer
CylonicRaider: A bare `nproc` should suffice unless the `make` can somehow access more CPU cores than `nproc`. Aside from that, I wonder whether we should hog all of the user's cores with compiler processes unconditionally.

pioneer
CylonicRaider: OTOH, my previous observation seems to be related to me switching commits before rerunning the script. It might be still better to print a message (like `warn: version may be stale; re-run cmake manually if you need it up-to-date`) instead of re-running CMake to eliminate those compile times if one does not care about Pioneer displaying the right commit.

impaktor:
Do you need to `cd build`? You can compile it with cmake from pioneer's root folder (`make -C  build -j`), unless I'm missing something?

CylonicRaider:
I don't immediately recall a way to make `cmake` generate build files in a directory different from the current one; after that, one can as well run `make` in the already-entered directory.

CylonicRaider:
Apparently, CMake indeed always generates in the current directory and no-one can even *imagine* it doing otherwise. What one *can* do, however, is use it to build the project rather than hard-coding Make — and this mode *requires* specifying the directory where the build files are. :laughing:

pcercuei:
```cmake -Bbuild -H.```

Web-eWorks:
@CylonicRaider please prefer calling `./bootstrap` instead of `cd`ing into build and invoking CMake directly. `./bootstrap` sets up default options for cmake, whereas a bare call to cmake will not.

Once you're calling the bootstrap script, building is as simple as invoking `make -C build -j$(nproc)`.

CylonicRaider:
While @pcercuei's command works, I can't find any documentation on it. There is a (documented) `cmake -Ssrc -Bbuild` mode which was added in 3.14 (i.e. this March), but that version is way above what the CMakeLists declare. In any case, @Web-eWorks' advice renders all of that moot.

pcercuei:
Yep. Hurray for undocumented features :p

CylonicRaider:
Portable parallelization support (`CMAKE_BUILD_PARALLEL_LEVEL`) was introduced in CMake 3.12, i.e. a year ago, and still way above 3.4. :disappointed:

It's also curious how no-one noticed that it says `make build`.

CylonicRaider:
Another way to (hopefully) speed up the build would be to split the currently single `modelcompiler` invocation into one per model so that it can be parallelized and managed by the build system.

fluffyfreak:
So is this ready to merge?

CylonicRaider:
**Yes**. The outstanding issues (parallelization as default, in particular parallelizing the model compiler) might be worth addressing in a follow-up, but are not required for the script to work, and all of the script is optional anyway.

CylonicRaider:
An additional issue is that the script may result in Pioneer displaying a stale commit hash when run after changing commits; however, one might argue that this is an issue with the build infrastructure itself (which ought to detect commit changes). I have left this unresolved as the Autotools incarnation of the script displayed the same behavior (IIRC), and as the current behavior is slightly faster.

fluffyfreak:
Merged, thanks!

PR #4624, opened 2019-06-13
Pigui equipment market for station view

vakhoir:
So I've been working on this for a while, and it's almost ready, but like I said in another PR, I won't be available for the next couple weeks. I thought I'll open a PR and maybe someone can take a look, I'll finish it when I'm back.

PR #4640, opened 2019-07-12
Fix transparency on menu items

pcercuei:
Fix a bug introduced in c19aab4e4 ("Color.h: Fix conversion from float
to u8"). The order of operations was wrong.

The resulting behaviour was broken transparency on menu items, and probably other things.

### DIFF
--- a/src/Color.h
+++ b/src/Color.h
@@ -131,15 +131,15 @@ struct Color4ub {
 	Uint8 GetLuminance() const;
 	void Shade(float factor)
 	{
-		r *= static_cast<Uint8>(1.0f - factor);
-		g *= static_cast<Uint8>(1.0f - factor);
-		b *= static_cast<Uint8>(1.0f - factor);
+		r = static_cast<Uint8>(r * (1.0f - factor));
+		g = static_cast<Uint8>(g * (1.0f - factor));
+		b = static_cast<Uint8>(b * (1.0f - factor));
 	}
 	void Tint(float factor)
 	{
-		r += static_cast<Uint8>((255.0f - r) * factor);
-		g += static_cast<Uint8>((255.0f - g) * factor);
-		b += static_cast<Uint8>((255.0f - b) * factor);
+		r = static_cast<Uint8>(r + (255.0f - r) * factor);
+		g = static_cast<Uint8>(g + (255.0f - g) * factor);
+		b = static_cast<Uint8>(b + (255.0f - b) * factor);
 	}
 
 	static const Color4ub BLACK;


src/Color.h
```diff
@@ -131,15 +131,15 @@ struct Color4ub {
 	Uint8 GetLuminance() const;
 	void Shade(float factor)
 	{
-		r *= static_cast<Uint8>(1.0f - factor);
-		g *= static_cast<Uint8>(1.0f - factor);
-		b *= static_cast<Uint8>(1.0f - factor);
+		r = static_cast<Uint8>(r * (1.0f - factor));
```
mike-f1: I don't think this is mandatory, but I feel better if `r` is explicitly converted to float before its use.

Pros are that anyone is aware of conversions at a glance, cons are... none as the conversion must be
done anyway(?)

vakhoir:
In the issue you mentioned that the screenshot issue is also fixed, but in my tests that was something related to #4626 (still haven't found the exact source). Did you test it yourself? I think it only happens when in station or player info views.

src/Color.h
pcercuei: You mean another static_cast<float>()? 😅

pcercuei:
I just did one screenshot at the same spot as always and it resulted perfect.

src/Color.h
mike-f1: >You mean another static_cast()?  :sweat_smile:

`r = static_cast<Uint8>(float(r) * (1.0f - factor));`

I know is "stylistically" uglier but implicit conversion still bite C/C++ programmers, and to me a better 'style' is a worse deal than a better 'grasp'...

I'm for merge anyway :smile:

pcercuei: By writing ```r = static_cast<Uint8>(r * (1.0f - factor));``` then it's obvious that the inner ```r``` is casted to float in the multiplication, I think.

I'm more a C developer myself, but I know that in C++ static_cast<> is preferred. I'd rather not mix the two cast styles.

Web-eWorks: IIRC a C-style cast would be `(float) r` not `float(r)` though I might be mistaken. The first reads as 'interpret r as a float', while the second reads 'convert r to float and use it in this expression', being an initializer instead of a cast.

Of course, the C++ spec is perfectly clear as to what's happening here; the byte gets implicitly converted to float as it's the larger datatype. I don't think we need to explicitly specify the cast.

mike-f1: @Web-eWorks , @pcercuei : No needs, of course
> but I feel better if r is explicitly converted to float before its use.

And:
> I'm for merge anyway

thus, hurry up :+1:

pcercuei:
Actually, before the screenshots were completely corrupted here, most of the time they resulted completely blank. With this commit I cannot reproduce it anymore. However, on some screenshots the overlay UI is shown with a chessboard pattern grey background. But not all of them.

vakhoir:
Can you append some of them? I'm going to start testing this branch now, I'm pretty sure the transparency issue should still be there, because of the tests I did yesterday, but I haven't seen the other issues you mention.

pcercuei:
Heh. I reverted the fix and compiled it again, and I can't reproduce it...

vakhoir:
Ok, this is interesting...
This is from my tests yesterday (when I reverted the change from Color.h):

screenshot-20190711-203034

And this is from today:

screenshot-20190712-195239

So yesterday the buttons were fixed but the background was not. An now both seem mostly (the black part at the bottom and on the outside is still grey/transparent) fixed.

Come to think of it, when was the last time anyone used the in-game screenshots? I always took the via the OS, not the game. Maybe this was broken for a while?

vakhoir:
Yeah, it looks like the background issue was there for a long time. First of all the problem only appears in PiGui views. That's why it seemed like reverting to a certain commit was fixing the problem somehow, but in fact it was just going back to the NewGui views, which were rendered more or less properly. This screenshot is from a 100 or so commits ago:

screenshot-20190713-125540

You can see the background at the bottom is transparent, I'm pretty sure it's a symptom of the exact same issue. I'll open an issue for this, so it's not forgotten, but it's not relevant to this PR.

impaktor:
> You can see the background at the bottom is transparent, I'm pretty sure it's a symptom of the exact same issue. I'll open an issue for this, so it's not forgotten, but it's not relevant to this PR.

I'm confused. Does this PR fix the transparency issue, and if not, then what does this PR fix?

vakhoir:
It fixes the Shade function. The bug caused buttons to be rendered as near black. Take a look at your own screenshot in #4624

PR #4643, opened 2019-07-14
Remove unused virtualization of Space and 'emplace_back' elements

mike-f1:
As title says, remove `virtual` on Space,
`emplace_back` in Prepare (thus no copy or move are needed),
and use `nullptr` instead of 0

...little by little...

### DIFF
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -31,7 +31,7 @@ void Space::BodyNearFinder::Prepare()
 	m_bodyDist.clear();
 
 	for (Body *b : m_space->GetBodies())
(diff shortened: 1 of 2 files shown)


impaktor:
Not sure I see what's going on here, so some of you C++ hackers have to give feedback.

mike-f1:
@impaktor , I don't think that is an ["hack"](https://stackoverflow.com/questions/43596430/push-back-vs-emplace-back-to-a-stdvectorstdstring), or that that [de(_virtual_)ization](https://stackoverflow.com/questions/300986/when-should-you-not-use-virtual-destructors) have conseguences... Anyway the code should be reviewed, and there's no hurry

fluffyfreak:
Yep it looked fine to me so merged

mike-f1:
:)

Issue #4545, opened 2019-02-23
Incorrect hyperspace range shown in comm screen


ElonSatoshi:
### Observed behaviour
![2019-02-23-095424_1366x768_scrot](https://user-images.githubusercontent.com/24940510/53288942-e6c59600-3754-11e9-8815-aeab6e859ad5.png)
![2019-02-23-095513_1366x768_scrot](https://user-images.githubusercontent.com/24940510/53288943-ecbb7700-3754-11e9-809b-3180dfe6fb22.png)

Hyperspace range is shown as way too much in this screen. The sector map, however, knows what the correct lightyear range is and will say INSUFFICIENT_FUEL if a star system is too far away.

The number displayed decreases when I buy more hyperdrive fuel.

### Expected behaviour
Comms screen shows actual lightyear range, which increases when I add more fuel.

### Steps to reproduce
1. Git clone pioneer, checkout the commit hash mentioned below, compile Pioneer using the cmake method.
2. Open pioneer, start a new game as easy or intermediate
3. Open comms, observe the hyperspace range shown that is way too high, is more than 0 when there's no fuel, and decreases when more fuel is added.

__My pioneer version (and OS):__
Pioneer commit [0072a09](https://github.com/pioneerspacesim/pioneer/commit/0072a0959c026bca3939a70014c84780eb4fc714) on Ubuntu 18.04.2 LTS x86_64

PR #4644, opened 2019-07-15
Consider thruster upgrade

robothauler:
Uses methods from LuaPlayer.cpp to calculate acceleration and delta v.
Fixes also the atmo shield info.

### DIFF
--- a/data/pigui/modules/info-view/01-ship-info.lua
+++ b/data/pigui/modules/info-view/01-ship-info.lua
@@ -77,15 +77,10 @@ local function shipStats()
 	rearWeapon =  rearWeapon  or nil
 
 	local mass_with_fuel = player.staticMass + player.fuelMassLeft
-	local mass_with_fuel_kg = 1000 * mass_with_fuel
 
-	-- ship stats mass is in tonnes; scale by 1000 to convert to kg
-	local fwd_acc = -shipDef.linearThrust.FORWARD / mass_with_fuel_kg
-	local bwd_acc = shipDef.linearThrust.REVERSE / mass_with_fuel_kg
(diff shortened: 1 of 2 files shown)


impaktor:
Code looks OK to me.

Have you tested that you get the same DeltaV now as before? (suggestion: use `print()` on version of script that has both old and new way of computing it). Asking as a way to make sure the Get method does the same thing as the manual calculation.

mike-f1:
There's plenty of those methods, but evidently not all are working: see #4545

robothauler:
The Get method and the manual calculation return the same DeltaV.

mike-f1:
> The Get method and the manual calculation return the same DeltaV.

@robothauler , may I ask if you have time/will to update methods used in "comm screen"?
Thus with this PR issue #4545 can be closed :wink: ...not a "must", anyway.

robothauler:
> > The Get method and the manual calculation return the same DeltaV.
> 
> @robothauler , may I ask if you have time/will to update methods used in "comm screen"?
> Thus with this PR issue #4545 can be closed wink ...not a "must", anyway.

Sorry, I think this is beyond my skills at the moment.

mike-f1:
Ops, I didn't understand what happen here... But not a problem:
I did some investigation and it seems that the "right" method cited in [that](#4545) issue is:
https://github.com/pioneerspacesim/pioneer/blob/3a314cf28ba1ecbd97c4c2a9d8c638d58a3d89bb/src/SectorView.cpp#L1169-L1171

Thus I feel is worth to check if this should be used instead.

>Sorry

For what? I give the wrong hint, I should be sorry :wink:

PR #4648, opened 2019-07-25
WIP - Modified all VS projects to use the new sigcpp files in thirdparty

fluffyfreak:
WIP - I'll need to double check this but...

I hadn't realised that `pioneer-thirdparty` had changed recently and the Visual Studio project files haven't actually worked in a while once you got latest from it.

As part of fixing it up I'm removing the `win32` version of sigc++ because I didn't use it anyway and just build a statically linked lib directly from the common source files.

This will need the corresponding PR in `pioneer-thirdparty` to be merged simultaneously:
https://github.com/pioneerspacesim/pioneer-thirdparty/pull/27

### DIFF
--- a/win32/vs2015/collider/collider.vcxproj
+++ b/win32/vs2015/collider/collider.vcxproj
@@ -142,7 +142,7 @@
       <GenerateDebugInformation>true</GenerateDebugInformation>
     </Link>
     <ClCompile>
-      <AdditionalIncludeDirectories>../../../src;../../src;../../../pioneer-thirdparty/win32/include;../../../../pioneer-thirdparty/win32/include;../../../../pioneer-thirdparty/source/libsigc++-2.3.1;../../win32/include;../../include;../../../contrib;../../contrib</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>../../../src;../../src;../../../pioneer-thirdparty/win32/include;../../../../pioneer-thirdparty/win32/include;../../../../pioneer-thirdparty/source/libsigc++-2.10.2;../../win32/include;../../include;../../../contrib;../../contrib</AdditionalIncludeDirectories>
       <AdditionalOptions>-Zm137 %(AdditionalOptions)</AdditionalOptions>
     </ClCompile>
     <Lib>
@@ -156,7 +156,7 @@
       <GenerateDebugInformation>true</GenerateDebugInformation>
     </Link>
     <ClCompile>
-      <AdditionalIncludeDirectories>../../../src;../../src;../../../pioneer-thirdparty/win32/include;../../../../pioneer-thirdparty/win32/include;../../../../pioneer-thirdparty/source/libsigc++-2.3.1;../../win32/include;../../include;../../../contrib;../../contrib</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>../../../src;../../src;../../../pioneer-thirdparty/win32/include;../../../../pioneer-thirdparty/win32/include;../../../../pioneer-thirdparty/source/libsigc++-2.10.2;../../win32/include;../../include;../../../contrib;../../contrib</AdditionalIncludeDirectories>
       <AdditionalOptions>-Zm137 %(AdditionalOptions)</AdditionalOptions>
     </ClCompile>
     <Lib>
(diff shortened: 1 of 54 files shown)


pcercuei:
I'm not really in favor of updating the VS project files, now that CMake can generate them...
Did you try to build with CMake?

fluffyfreak:
@pcercuei just did, and there's a few issues.

They don't currently work because the `USE_PIONEER_THIRDPARTY` section makes the `pioneer-thirdparty` folders a child of the `pioneer` folder path instead of a sibling. 
Eg:
`D:\Code\PioneerGit\pioneer\pioneer-thirdparty\win32\include`
instead of the correct path:
`D:\Code\PioneerGit\pioneer-thirdparty\win32\include`
I've tried fixing it up with relative pathing etc which gets it past that problem, but then it fails to link.

Also it regresses the sigc++ stuff to being a DLL, whereas we had it as a static library called `sigcpp`, and I was trying to move away from the win32 specific folder and just use the `pioneer-thirdparty\source\` directly. For some reason a lot of Linux projects that do MSVC versions revert to DLLs instead of static libs for no discernible reason.

Also it doesn't do multiprocessor compilation.

So they're still useful to keep around, it might be worth taking a look at them in fact to see what the differences are.

fluffyfreak:
@pcercuei none of that above is a criticism of the CMake stuff you've done, and this PR will probably get changed a bit now that I understand the CMake work a bit better thanks to the #4649 PR.

I'll probably switch fully to the CMake generated builds **asap**, it's just that keeping the existing ones around has some benefits still. 
I prefer there to be a long overlap where both work so my plan will likely be the following:
- remove VS2015 projects (_within days_)
- once VS2019 is out of beta, remove the VS2017 projects
- maintain VS2019 build alongside CMake for a short while (_months_)
- work on converting CMake & VS2019 to use static libs
- deprecate VS2019 entirely in favour of CMake (_unknown_)

pcercuei:
The pioneer-thirdparty is expected inside the "pioneer" folder in case we want to make it a git submodule at some point.

How is linking dynamically to libsigcpp a *regression*? Looks more like an improvement to me. I don't agree that static libs should be used.

Web-eWorks:
On Windows, at least, loading a DLL requires it to be linked into the process's address space in a complex manner (the loader calculates all the program offsets on a per-process basis instead of ELF-style position-independent code (as I remember; this might be older pre-WinPE object format)), requires shipping and maintaining another file, and most importantly requires an indirect call everytime we call into sigcpp. Statically linking reduces startup complexity, failure points, and most importantly allows the linker to further (and possibly significantly) optimize calls into the sigcpp routines.

impaktor:
>    - remove VS2015 projects (within days)
>    - once VS2019 is out of beta, remove the VS2017 projects
>    - maintain VS2019 build alongside CMake for a short while (months)
>    - work on converting CMake & VS2019 to use static libs
>    - deprecate VS2019 entirely in favour of CMake (unknown)

If VS build can be made to work though CMake, then the "backup" could just be on your local side, as I think it's mostly you using VS (and JohnJ).

pcercuei:
These reasons seem bogus to me. Classic case of over-optimization.

Anyway, you're free to change the pioneer-thirdparty libs to static, I don't care. As long as you don't enforce it as the general rule in CMake.

Btw, for the pioneer-thirdparty path, on Windows you could call [AddDllDirectory](https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-adddlldirectory) to avoid copying the DLLs.

fluffyfreak:
> The pioneer-thirdparty is expected inside the "pioneer" folder in case we want to make it a git submodule at some point.
> 

That's never been how we've done it before, nor how the instructions tell us anyone to do it.
The folder structure is expected (_and documented_) to be:
./pioneer
./pioneer-thirdparty

At least part of the reason for not using submodules is that every project I've seen use them has had problems with them.
Referencing another sibling heirarchy project introduces issues like this one, where I hadn't noticed things had been updated, but it is a simple failure case and easy to learn from or fix.

> How is linking dynamically to libsigcpp a _regression_? Looks more like an improvement to me. I don't agree that static libs should be used.

Statically linked libs can be optimised along with the rest of the code whereas that's obviously impossible for DLLs. In the case of SIGC++ I changed it to a static lib because previously it showed up in profiling as an occasional performance hog which doesn't happen when statically linked because it can be properly optimised away.
So not over-optimisation, just the classic profile->find cause->optimise for a real performance problem on this game ;)

Also we don't benefit from any of the usual uses of DLLs as we're not sharing them with other programs, no-ones distributing updated versions of them for Pioneer etc. It also makes for yet another difference between Win, Linux & Mac which are (_or were_) mostly statically linked.

pcercuei:
Everything can appear as a performance hog with the right benchmark. The real question is, how many more FPS do you get by linking statically?

fluffyfreak:
That is not the right question since I have a very fast desktop PC and don't see any performance issues ever. If that was the benchmark I'd never have optimised it or anything else. On lower end hardware it made a difference, small but meaningful for those playing it. Unless you're lucky and find something badly coded you have to make a lot of those kind of optimisations. It's just one of many.

PR #4649, opened 2019-07-26
Fix VS CMake

fluffyfreak:
Does a few things to get Visual Studio builds going.
- Enable multithreaded compilation for VS, 
- prevent it Optimizing the models per-build, 
- fix up the pioneer-thirdparty paths

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,6 +105,8 @@ list(APPEND SRC_FOLDERS
 )
 
 foreach (each IN LISTS SRC_FOLDERS)
+	file(GLOB header_files RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${each}/*.h)
+	list(APPEND HXX_FILES ${header_files})
 	file(GLOB src_files RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${each}/*.cpp)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -293,11 +300,14 @@ endif (CMAKE_CROSSCOMPILING)
 if (MODELCOMPILER)
 	# Optimize the models.
 	# This really shouldn't be done inside the source tree...
+if (WIN32)
+else (WIN32)
```
pcercuei: ```If (NOT MSVC)```
WIN32 includes MinGW builds.

impaktor:
> fix up the pioneer-thirdparty paths

Sounds like we (you guys) should agree on where the propper path should be. I see the logic in both sides of the argument.

fluffyfreak:
What I'm most interested in this PR is that I'm not breaking anything non-MSVC related... am I?

pcercuei:
If you move the location of the pioneer-libs folder, you'll have to update the appveyor script as well.

fluffyfreak:
> You just changed that in your second commit, why do you change it again? Just edit you commit.

because I'm going to squash them all anyway

pcercuei:
Hey, review wasn't done :(
Indentation is still wrong...

Issue #4222, opened 2017-12-27
Atmosphere calculations


joonicks:
Bad news. The whole atmosphere calculation functions Planet::InitParams and Planet::GetAtmosphericState seems to be fubar.

InitParams calculates how thick the atmosphere is for all bodies with an atmosphere. The result is most often much too thin compared to reality.
earth, 25km vs reality 480km (google)
venus, 80km vs reality 250km (wikipedia)
jupiter gets stuck with 79km of atmosphere
saturn gets 143km
explore a bit and you'll run into gas giants with only 10-30km of atmosphere

in any case, it seems calculating the thickness of the atmosphere isnt as easy as the current code tries to make it out to be.

so the question is, what to do about it.

caligari87:
Would this be part of the reason I don't seem to get any atmospheric heating gear-up even at orbital velocity, and barely anything worth mentioning even with gear down? Was recently thinking that seemed odd, like there was no reason to ever be concerned with temperature.

joonicks:
this is not about the temperature. this is about how far the atmosphere extends and the atmospheric pressure at different altitudes. basically everything about atmospheres.

nozmajner:
@caligari87 You'd have way more heating without the Atmospheric shielding equipment.

jaj22:
Surely the "atmosphere height" is somewhat arbitrary, given that the density is an exponential falloff? Given how the frames work, it's a reasonable approximation to set the density threshold relatively high.

Now if the density is way off within the threshold range that Pioneer simulates, that's a reasonable issue to fix.

joonicks:
atmosphere height also influences the haze drawn around atmospheric planets

impaktor:
25 km doesn't sound that far off in my mind. Planes fly at 10 km, at 20 km you see earth curvature and those flying there report it (thus) feels like being in space. Where to put the limit for where atmosphere stops is arbitrary, as there's not a hard limit. 

> earth, 25km vs reality 480km (google)

I'm guessing 480km is in the higher end of the spectrum. Also, this would mean the international space station, isn't even in space, assuming "space" and atmosphere don't overlap.

[Wikipedia:](https://en.wikipedia.org/wiki/Atmosphere_of_Earth)
> The mass of Earth's atmosphere is distributed approximately as
follows:
>
> - 50% is below 5.6 km (18,000 ft).
>
> - 90% is below 16 km (52,000 ft).
> 
[...]

PR #4653, opened 2019-07-28
Use an equation in order to calculate HullTemperature

mike-f1:
It should behave as before, except that with wheels down the temperature is little more (it treble without any shielding).

From now on, an atmospheric shield value of 1 will halve the temperature,
which is quite straightforward.

EDIT: Ops, It seems actually "Atmospheric shields" automatically goes down if wheels are down...
I can do like that but I would ask if it's not too much: if a ship has wheels up and is travelling fast, then hitting the wrong button (landing gear) means that the ship will overheat and probably will explode. Is this right/wished?

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -702,12 +702,12 @@ misc.missile_naval = EquipType.New({
 })
 misc.atmospheric_shielding = EquipType.New({
 	l10n_key="ATMOSPHERIC_SHIELDING", slots="atmo_shield", price=200,
-	capabilities={mass=1, atmo_shield=1},
+	capabilities={mass=1, atmo_shield=9},
 	purchasable=true, tech_level=3
 })
 misc.heavy_atmospheric_shielding = EquipType.New({
 	l10n_key="ATMOSPHERIC_SHIELDING_HEAVY", slots="atmo_shield", price=900,
-	capabilities={mass=2, atmo_shield=2},
+	capabilities={mass=2, atmo_shield=19},
 	purchasable=true, tech_level=5
 })
 misc.ecm_basic = EquipType.New({
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -1039,11 +1039,7 @@ double Ship::GetHullTemperature() const
 	double dragGs = GetAtmosForce().Length() / (GetMass() * 9.81);
 	int atmo_shield_cap = 0;
 	const_cast<Ship *>(this)->Properties().Get(R"(atmo_shield_cap)", atmo_shield_cap);
-	if (atmo_shield_cap && GetWheelState() < 1.0) {
-		return dragGs / (150.0 * atmo_shield_cap);
-	} else {
-		return dragGs / 15.0;
-	}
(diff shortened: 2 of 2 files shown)


WKFO:
Maybe a gradual increase in heat by getting wheelstate would be more preferable. Otherwise, if you accidentally lower the gear during high-speed flight, it causes instant death.

vakhoir:
As far as I can tell, the increase is gradual.

In fact, I'd increase the drag factor. I can't seem to increase the temperature to a dangerous level when going into the Earth's atmosphere, direct from orbit, without atmospheric shields.

WKFO:
> 
> 
> As far as I can tell, the increase is gradual.
> 
>

> 
> As far as I can tell, the increase is gradual.
> 

https://youtu.be/ZKDVbfBmhNY <-----It does this in my build. I need someone to confirm though.

mike-f1:
@vakhoir : With these few lines I would only code something without branching (the `if`), and let Lua to have more granularity for `atmo_shield` value.
Then I ended handling (better?) the case for wheels moving as the removal of that `if` avoid discontinuities on calculation.

Probably you're right about the increase of "drag factor", but there's a lot of things involved.
An example is #4222, which can be the reason you're able to do what you did whitout any issue
(which is a good news, as every AI ship do the same...)

> I'd increase the drag factor

May I ask how? ...As there's a lot of comments but the equation I see don't make much sense (to me at least...)

vakhoir:
Well, too late for my answer it seems, but in case anyone wants to tackle this in the future:

> May I ask how?

I may have phrased it badly, I mean just a brute force ramp-up of the output of the equation so the temperature is higher. For example I see the denominator right now has an arbitrary 15.0 factor... just lower it.

> which can be the reason you're able to do what you did whitout any issue
(which is a good news, as every AI ship do the same...)

I assume (most?) AI ships have basic equipment like atmospheric shields. The sweet-spot, in my opinion, would be for ships to be able to pull of these maneuvers with an atmospheric shield, but the same maneuver being lethal unless the pilot is extremely careful.

The motivation for this is my recent high-difficulty playthrough (Bernard's Star beginning) where I started by selling off all the "useless crap" like the autopilot and atmospheric shield, in order to buy a hyperdrive. I think that should be a valid strategy, but it should come at a cost, but while the disatvantages of not having an autopilot are obvious, I never once missed my atmospheric shield.

mike-f1:
> Well, too late for my answer it seems

I don't think so: I think @fluffyfreak o the right thing, but even if this got merged your point is valid.
This is the reason behind the issue I opened

>I may have phrased it badly

...or **I** misunderstand... or my english lacks some care :(

> For example I see the denominator right now has an arbitrary 15.0 factor... just lower it.

...Which is a _gameplay_ change and needs more test and effort, then I open an issue trying to address that change as, before any effort will worth, a reliable and meaningfull model should be in place.

> The motivation for this is my recent high-difficulty playthrough (Bernard's Star beginning) [snip] obvious, I never once missed my atmospheric shield

...which is a gameplay issue and a lieu for an improvement ;)

PR #4654, opened 2019-08-03
PiGui commodity market

vakhoir:
As promised, the pigui commodity market:

Screenshot from 2019-08-03 18-19-51

I noticed that a lot of the ui does not scale well for other resolutions, so I'm diverting my efforts to have some kind of semi-automatic rescaling

EDIT: What's up with appveyor? I don't think I changed anything that would crash the build.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -87,6 +87,10 @@
     "description": "",
     "message": "Cancel"
   },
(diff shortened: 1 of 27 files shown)


fluffyfreak:
@vakhoir you haven't done anything to AppVeyor, it's been throwing that or some other error for a while now. Safe to ignore it.

PR #4655, opened 2019-08-04
Remove the rings from Venus and Ariel

### LINKED ISSUE #4652: Venus (ringless) has rings in pioneer
Venus is a ringless planet. It has rings in the game though.
Pioneer version: git master d72e4e9dc9dbd1f48109419d36ca322ebc7ec2d2

venus

fluffyfreak:
Fixes #4652 by marking the them as not having rings.
I checked all of the other bodies for rings, these were the only ones

### DIFF
--- a/data/systems/00_sol.lua
+++ b/data/systems/00_sol.lua
@@ -48,6 +48,7 @@ local venus = CustomSystemBody:new('Venus', 'PLANET_TERRESTRIAL')
 	:ice_cover(f(0,1))
 	:life(f(0,1))
 	:orbital_phase_at_start(fixed.deg2rad(f(248,1)))
+	:rings(false)
 
 local earth = CustomSystemBody:new('Earth', 'PLANET_TERRESTRIAL')
 	:radius(f(1,1))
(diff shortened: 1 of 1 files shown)


impaktor:
> I checked all of the other bodies for rings,

Probably stupid question, but you mean in the code, right? So with this PR, all bodies have explicit true/false on rings? If so, merge when ready.

fluffyfreak:
Nope this just changes the custom system definintion for Sol to prevent Venus & Ariel from having rings explicitly. No code changes.

impaktor:
So next time the seed or random number generator changes, any planet in Sol that doesn't have explicitly defined ring state, might get rings?

mike-f1:
>might get rings?

[...probably? :smile:](https://en.wikipedia.org/wiki/Murphy%27s_law)

impaktor:
Just to be clear, when I said "code" above, I meant "Sol custom system script", and I think a good fix would be to explicity set rings = false for all planets except Saturn, I guess. No change in C++ for this.

PR #4651, opened 2019-07-27
PiGui Station Lobby view

vakhoir:
Takes into account @mike-f1 hyperspace range fix from #4650
Was branched from #4624, so it also contains it's commits

EDIT: Will also close #4545

Screenshot from 2019-07-27 17-48-18

PR #4663, opened 2019-08-10
Improved UI resolution rescaling

### LINKED ISSUE #4545: Incorrect hyperspace range shown in comm screen

vakhoir:
_EDIT by @impaktor_ [this](https://help.github.com/en/articles/closing-issues-using-keywords#closing-multiple-issues) closes #4624 closes #4651 closes #4654, fixes #4545

Like I mentioned in the previous pull request, I noticed the ui doesn't seem to rescale very well. We typically use code like this to scale to current resolutin:

`local mainButtonSize = Vector2(32,32) * (ui.screenHeight / 1200)`

It usually works well enough. but only within the same aspect ratio. Calculating these sizes has also been annoying, because the formula assumes 1920x1200 as the base resolution, and I usually work on 1600x900. So I wrote the `ui.rescaleUI = function(val, baseResolution, rescaleToScreenAspect)` function, which will automatically rescale any value, or all values in a table, using the relatively smaller dimension (so if the width changed more relative to the base resolution, it will use that). The `rescaleToScreenAspect` parameter works only for vectors, and if it's set it will use both dimensions (so a shape that is a square in 4:3 will become more rectangular in 16:9, and vice versa).

Other than rescaling I added some resolution-dependant rearrangements in the lobby.

So, without further ado, some comparisons:

**Station lobby - 1600x900 (my base resolution):**

1600x900_lobby

**Station lobby - 1152x864, old rescaling:**

1152x864_lobby

**Station lobby - 1280x960, new rescaling**

1280x960_rescale_lobby

**Station lobby - 800x600, old rescaling:**

800x600_lobby

**Station lobby - 800x600, new rescaling:**

800x600_rescale_lobby

**EquipmentMarket - 1600x900, old rescaling:**

1600x900_eq_market

(note I originally did not notice that with the sizes provided I cut off the "In Stock" column)
**EquipmentMarket - 1600x900, new rescaling:**

1600x900_rescale_eq_market

**EquipmentMarket - 800x600, old rescaling:**

800x600_eq_market

**EquipmentMarket - 800x600, new rescaling:**

800x600_rescale_eq_market

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -87,6 +87,10 @@
     "description": "",
     "message": "Cancel"
   },
+  "CANNOT_SELL_ITEM": {
+    "description": "",
+    "message": "This item cannot be sold"
+  },
   "CAPACITY": {
     "description": "Capacity of ship to fit more equipment",
     "message": "Capacity"
@@ -1650,6 +1654,10 @@
   "STATION_ORBIT": {
     "description": "Information shown in the station loby",
     "message": "We are in a {orbit_period} day period orbit around {parent_body}."
+  },
+  "STATION_TECH_TOO_LOW": {
+    "description": "",
+    "message": "Station tech level too low to sell this item"
   },
   "STATUS": {
     "description": "",
--- a/data/libs/ui/PiImage.lua
+++ b/data/libs/ui/PiImage.lua
@@ -1,0 +1,40 @@
+-- Copyright © 2008-2019 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local Image = import 'PiGui.Modules.Image'
+local ui = import 'pigui/pigui.lua'
+local colors = ui.theme.colors
+
+local PiImage = {}
+
+function PiImage.New (filename)
+
+	local piImage = {
+		widget = Image.New(filename)
+	}
+
+	piImage.texture = {
+		id = piImage.widget.id,
+		size = piImage.widget.size,
+		xy = piImage.widget.size.x / piImage.widget.size.y,
+		uv = piImage.widget.uv
+	}
+
+	piImage.Draw = function(self, size)
+		if (size.x == 0 and size.y == 0) then size = self.texture.size
+		elseif (size.y == 0) then size.y = size.x / self.texture.xy
+		elseif (size.x == 0) then size.x = size.y * self.texture.xy
+		end
+
+		ui.image(self.texture.id, size, Vector2(0.0, 0.0), self.texture.uv, colors.white)
+	end
+
+	setmetatable(piImage, {
+		__index = PiImage,
+		class = "UI.PiImage",
+	})
+
+	return piImage
+end
+
+return PiImage
--- a/data/pigui/libs/market.lua
+++ b/data/pigui/libs/market.lua
@@ -175,6 +175,11 @@ local defaultFuncs = {
 local MarketWidget = {}
 
 function MarketWidget.New(id, title, config)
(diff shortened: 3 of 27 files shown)


mike-f1:
First of all: nice work, not only for this PR but also for the other one's (...so many I can't count :) ).
This thing of rescaling is a big improvement as first issues are already opened: the old UI has not this hard limit (IIRC), and iy can change scaling far better.
So centralizing all the calls to the same function is a good step in the right direction but I think that _The_ final solution is to change all functions in LuaPiGui.cpp to request a different kind of Vector specifically designed to be used for screen coordinates.

...As I said the above are a future topic, for now this will suffice.

Well done!

impaktor:
Nice! And thanks for the many screenshots. Helps a lot to show the effect of the PR.

impaktor:
I've been a bit absent (still am), but does anyone want to voice opinion on this, or any hinderence to merging?

pcercuei:
I think it'd look better to have each fuel gauge below its set of buttons. There is quite a lot of vertical space available.

fluffyfreak:
It looks good.
That doesn't look like font downsizing, it's a higher resolution so that font is proportionally scaled by the looked of it.

pcercuei:
It is downsizing, if you look at the last two screenshots, they are both 800x600.

fluffyfreak:
It looks good to me, although I admit that I do not play in 800x600... does anyone?

Web-eWorks:
I'll review the Lua in a couple hours, otherwise I'm very happy with better rescaling!

Web-eWorks:
Lua looks good to me! Only thing I'd comment on is that we might want to make `screenSize` an optional parameter defaulting to 1600x900 for `rescaleUI`, to establish a common base screen size.

pcercuei:
If you do that, I won't contribute anymore, because I won't be able to run the game anymore :)

Web-eWorks:
@pcercuei the idea is to specify a common 'base' resolution to develop for, and allow the game to handle resizing / scaling the UI at higher or lower resolutions (or different aspect ratios). I intend (and I'm sure @vakhoir intends as well) to make the game playable at any resolution you can think of.

pcercuei:
Ah, ok.

impaktor:
Sweet!

I've edited the first message, to auto-close all other PRs this is branchin off from, as I've understood it.

Will merge shortly.

impaktor:
> Lua looks good to me! Only thing I'd comment on is that we might want to make screenSize an optional parameter (defaulting to e.g. 1600x900) for rescaleUI, to establish a common base screen size.

@Web-eWorks / @vakhoir shounds like a PR that would be merged fast ;), if it's something you'd like to have.

Sorry for the long delay on this barrage of PRs, but it's very nice to have these features in the game now! Let's see how many bugs we introduced now... (given the amount of code)

PR #4664, opened 2019-08-10
CMake: Ship Debian's FindGLEW.cmake script

### LINKED ISSUE #4662: Appveyor MINGW failures
When building the MINGW versions Appveyor seems to fail regularly, if not always, with the following error message:
```
CMake Error at C:/msys64/mingw64/lib/cmake/glew/glew-config.cmake:47 (message):
  find_package(GLEW) was called the second time with different
  GLEW_USE_STATIC_LIBS setting.  Previously, `glew-config.cmake` created
  GLEW::GLEW as a copy of _glew_previous_src_target-NOTFOUND.  Now it
  attempted to copy it from GLEW::glew.
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake/share/cmake-3.15/Modules/FindGLEW.cmake:62 (find_package)
  CMakeLists.txt:157 (find_package)
```

I have also seen it fail to checkout the repository during the MINGW build as well, no errors or information on why.

@pcercuei

pcercuei:
This will be used only when compiling under MinGW, since MinGW's GLEW
config script seems to be really broken.

Fixes #4662.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,11 @@ endif()
 
 include(GNUInstallDirs)
 
+if (MINGW)
+	# Fix build errors on AppVeyor with MinGW due to a broken GLEW config script
+	list(INSERT CMAKE_MODULE_PATH 0 ${CMAKE_SOURCE_DIR}/cmake/Modules)
+endif (MINGW)
+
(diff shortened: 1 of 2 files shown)


fluffyfreak:
nice one!

PR #4666, opened 2019-08-12
Update en.json

craigo-:
Correct spelling (payed -> paid)

I'm not sure if some of the traders out there are supposed to have bad spelling (e.g. for flavour)... If so, feel free to decline this PR. Otherwise, this is just a simple spelling correction.

Message Typo

### DIFF
--- a/data/lang/module-cargorun/en.json
+++ b/data/lang/module-cargorun/en.json
@@ -185,7 +185,7 @@
   },
   "FAILUREMSG_3": {
     "description": "",
-    "message": "Listen, we were clear about the deadline. You can forget about getting payed."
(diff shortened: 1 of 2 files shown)


impaktor:
Thanks. We want this. But not rush to merge, as 1. the file has already been translated, so "damage is done", 2. merging this will trigger re-translation of that string, so we sometimes (admittedly, rarely) let corrections in en.json accrue and then do transifex magic to not force re-translation of corrected strings in every language. Hope that answers @Web-eWorks's questions (from IRC).

craigo-:
OK, thanks for the explanation.

craigo-:
Another English correction, another commit (there -> their):

image

Issue #4288, opened 2018-01-22
Floating point arithmetic exception on ShipAICmd.cpp


CampNowhere:
To replicate: Go to star system referenced in the attached file. Target the first planet orbiting the black hole (BH1b1), hit autopilot, and "enter low orbit around BH1b1".

I re-implemented the std::max function locally to make debugging easier, but the same exception happens with current master (4d32739b69). It looks like there's a max operation being done on the xmm1 and xmm2 registers being done before returning, for some reason. I'm not sure why this is happening. The greater than comparison is being done in the ucomisd instruction a couple of  lines above. And nonsense data (0x800000....) seems to be loaded into xmm1.

I'll work on this further, I just wanted to put it out there for everyone else to look at.

![crash](https://user-images.githubusercontent.com/13381835/35248854-4cc4fb42-ff95-11e7-9082-72628b41b973.png)

[07_bh1.txt](https://github.com/pioneerspacesim/pioneer/files/1653867/07_bh1.txt)

jaj22:
The black hole is coming out of the sysgen with zero radius. The terrain code has no idea what to do with a zero-radius body, so the terrain feature height comes out as a NaN which is passed through to the black hole's m_physradius. This throws an exception in the autopilot, because we enable exceptions around it.

I don't know what physical radius black holes are supposed to have. Maybe the event horizon?

impaktor:
> I don't know what physical radius black holes are supposed to have. Maybe the event horizon?

Sounds reasonable. So what would the best fix be for this issue? Why does the black hole have zero radius? It doesn't in the lua-file, as far as I can see.

richardpl:
Lua file doesn't set radius of black hole, so it defaults to 0 for some reason.

mike-f1:
> So what would the best fix be for this issue?

Correct system generation code to force radius (or the default) to be more than [Schwarzschild radius](https://en.wikipedia.org/wiki/Schwarzschild_radius)?

Issue #4164, opened 2017-09-21
No initial Frame of Reference when jumping from Sol to Sirius


kennworl:
I don't know if this is a code bug, design flaw or meant to be like this, but when I jump from Sol system to Sirius system, I have no initial Frame of Reference set, yet I have heading, pitch and roll values around the reticule.

Happens every time.  Using the Windows 28 Aug std build, start a new game in London.  Launch, select Sirius in the sector map.  Go back to World view, and hit the hyperspace icon.  When you arrive in the middle of nowhere in the Sirius system, the reticule has no frame or nav target info.

I would expect there to always be frame of reference, being the biggest star in the system, given that you are never in interstellar space.

nozmajner:
Just checked, I can confirm it on windows.
It seems to happen for binary systems. For Epsilon Eridani it shows the star as frame, but for Toliman or Sirius it shows nothing.
Ctrl+i info screen shows "System" as reference in all of them.
Maybe because binary+ sytems have a common gravitational center?

fluffyfreak:
N-ary systems use a grav-point as their root. It doesn't have a name or anything since it's a notional/logical construct that we have in the code to organise systems.
Does pose an interesting question of what we should actually display in this case!

fluffyfreak:
Or disable the heading, pitch and roll display when it's a grav point

kennworl:
Ok, so what is the consensus here?  The reason I bring this up is that my hardware console I'm building expected there to always be an FoR.  I like the idea of "System centre of mass" being displayed as we could assume it is a arbitrary body which would then allow h/p/r angles displayed to make sense.  Just don't leave it blank.  This would seem to be the easiest logical fix for this one.

impaktor:
Those with understanding of c++ core would be the right ones to decide, so @jaj22 / @fluffyfreak or @ecraven / @laarmen I presume.

jaj22:
There are (at least theoretically) gravpoints that are not the "system centre of mass". The game has a convention for gravpoint names already (eg. "Sirius A,B") but only uses it for child bodies at the moment. Original reason might just be to avoid drawing the labels on-screen.

impaktor:
> Original reason might just be to avoid drawing the labels on-screen.

Interesting, as we since a couple of weeks back only draw icons.

jaj22:
Ok, here's how it works:

1. Gravpoints do have labels. The root sbody in Sirius is called "Sirius A,B".
[...]

PR #4668, opened 2019-08-22
[WIP] Compute gravity from each gravity source

mike-f1:
First commit tries to address what emerge in #4288, basically I did what @jaj22 suggested, and I replaced those `macro` with some more readable and centralized check, which can be extended to be used for more checks in `CustomSystem` and, if useful, also to avoid `GalaxyGenerator` to be completely free to play with numbers (which lead to a lot of strange results).

The second commit is addressing (partially) what emerge in #4164: it add a (more expensive) way to apply gravity forces, leading to a realistic behaviours when DynamicBody(ies) are exposed to more than a single gravitational body.

Given that the second commit doesn't solve #4164, I put some effort trying at least to show what is happening: the third add a grid (and a way to reset orientation) in `SystemView`. It still useful, and can be used to better understand what `GravPoints` meants, giving also a "reason" to keep feedback when in an empty frame. If enough to close the above issue, I let you decide.

Hope @fluffyfreak find some time to check if the graphic code in third commit is ok, and @jaj22 to check if math in second commit is ok (...which is hard to check given that gravity is a relatively weak force).

@nozmajner , feel free to replace the icons I did: they are really a shame :P

TODO:
1) `Orbit` aren't updated, so they can show a "not real orbit": I think is a little flaw, as they still approximate the behaviours.
2) Autopilot (apart from other improvements) should be aware of changes on gravity.

NITPICKS:
1) Optimize the code in Space as actually is not optimized (thoughts it does not show itself in profiler)
2) Add a "sub-grid" around `m_selectedObject`
3) Limit zoom to show only the entire system
4) Draw "legs" for ships (when system is the actual one's and ships are displayed)

### DIFF
--- a/src/DynamicBody.cpp
+++ b/src/DynamicBody.cpp
@@ -233,13 +233,6 @@ void DynamicBody::CalcExternalForce()
 		m_externalForce += m_atmosForce;
 	} else
 		m_atmosForce = vector3d(0.0);
-
-	// centrifugal and coriolis forces for rotating frames
-	if (GetFrame()->IsRotFrame()) {
-		vector3d angRot(0, GetFrame()->GetAngSpeed(), 0);
-		m_externalForce -= m_mass * angRot.Cross(angRot.Cross(GetPosition())); // centrifugal
-		m_externalForce -= 2 * m_mass * angRot.Cross(GetVelocity()); // coriolis
-	}
 }
 
 void DynamicBody::TimeStepUpdate(const float timeStep)
--- a/src/DynamicBody.cpp
+++ b/src/DynamicBody.cpp
@@ -213,16 +213,9 @@ void DynamicBody::CalcExternalForce()
 {
 	// gravity
 	if (!GetFrame()) return; // no external force if not in a frame
+	m_externalForce = m_gravityForce;
+
 	Body *body = GetFrame()->GetBody();
-	if (body && !body->IsType(Object::SPACESTATION)) { // they ought to have mass though...
-		vector3d b1b2 = GetPosition();
-		double m1m2 = GetMass() * body->GetMass();
-		double invrsqr = 1.0 / b1b2.LengthSqr();
-		double force = G * m1m2 * invrsqr;
-		m_externalForce = -b1b2 * sqrt(invrsqr) * force;
-	} else
-		m_externalForce = vector3d(0.0);
-	m_gravityForce = m_externalForce;
 
 	// atmospheric drag
 	if (body && GetFrame()->IsRotFrame() && body->IsType(Object::PLANET)) {
--- a/src/DynamicBody.h
+++ b/src/DynamicBody.h
@@ -52,6 +52,8 @@ public:
 	void AddRelTorque(const vector3d &);
 	vector3d GetExternalForce() const { return m_externalForce; }
 	vector3d GetAtmosForce() const { return m_atmosForce; }
+	// This should be called *only* from Space to apply gravity.
+	void SetGravityForce(const vector3d &g) { m_gravityForce = g;}
 	vector3d GetGravityForce() const { return m_gravityForce; }
 	virtual void UpdateInterpTransform(double alpha) override;
 
--- a/src/LuaSpace.cpp
+++ b/src/LuaSpace.cpp
@@ -675,6 +675,17 @@ static int l_space_get_bodies(lua_State *l)
 	return 1;
 }
 
+static int l_space_dump_frames(lua_State *l)
+{
+	Space *space = Pi::game->GetSpace();
+	if (!Pi::game && !space) {
+		luaL_error(l, "Game is not started");
+		return 0;
+	}
+	space->DebugDumpFrames();
+	return 0;
+}
+
 static int l_space_attr_root_system_body(lua_State *l)
 {
 	if (!Pi::game) {
@@ -707,6 +718,7 @@ void LuaSpace::Register()
 
 		{ "GetBody", l_space_get_body },
 		{ "GetBodies", l_space_get_bodies },
+		{ "DbgDumpFrames", l_space_dump_frames },
 		{ 0, 0 }
 	};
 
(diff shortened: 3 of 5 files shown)


jaj22:
Unless you're also removing frames in general, you need to at least ignore the gravity contribution of the direct parent(s) of the current frame object, as that's already applied by the frame-relative movement. Test with engines-off near a space station.

Performance may also be an issue. It's checking every dynamic body against every gravity source each frame. Most likely to be an issue when star systems have a ton of small moons and something spams cargo canisters or debris.

mike-f1:
> Unless you're also removing frames in general

not my intention: they works well, though I feel gravity could/should be applied without them as it's a "concept" above them.

> as that's already applied by the frame-relative movement

I missed it, sorry: may I ask where? ...or how?

> Test with engines-off near a space station

:P 

> It's checking every dynamic body against every gravity source each frame

...which is the only way I fear :P (...worth mentioning that there's a third level which are the frames themselves)

If you don't think it can do the job fast enough, I can drop that commit... But before doing such thing, please consider that numbers (even in worst situations) are small compared to colliders and that all this can be optimized (collecting a direct reference to "gravity frames" or to their absolute position and force).
Obviously if you have better ideas, please share them and I will try.

jaj22:
Frames implicitly apply the gravity from the frame's parent bodies. Consider if you're floating near Mars High, within its reference frame: No dynamic forces are being applied, as the station itself has no gravity. However, you're still orbiting Mars, and you're still orbiting Sol.

If you apply Mars or Sol's gravity in this case, you will rapidly drift out of the station's frame, as their gravity is now being incorrectly applied twice.

Assuming that this is fixed, there's still a more general problem that the station and Mars are only following gravity from the frame parents, but your ship is also getting a substantial contribution from Jupiter, so there will be some drift. Hence it's questionable whether the visible inaccuracies from including additional gravity sources are more or less important than the benefit.

mike-f1:
**Update:** I split original pull request so #4669 is first and third commits in order to let those be independent: this become related only on gravity changes.

> Assuming that this is fixed, there's still a more general problem [...snip...] Hence it's questionable whether the visible inaccuracies from including additional gravity sources are more or less important than the benefit.

...I don't have an answer: I feel it is a matter of taste to decide whether this is a wanted general gameplay/simulation feature.
I should pin that going for a "complete gravity/centrifugal/coriolis way" can lead to these improvement:
1) Realistic orbits around more than 1 body (eg gravpoints)
2) Realistic orbits **between** double or treble system bodies (orbit eg between them)
3) Realistic orbits near Lagrange points (which exist as a solution of a two body problem)
4) Realistic orbits such as [this](https://en.wikipedia.org/wiki/Polydeuces_(moon))
5) Realistic gravitational sling effect (...although I had doubts it is not already "included"...)

I would also pin that because gravity force are declined in "full form" (...squared in distance), I fell the case of big gas giant is mitigated by their distance (I don't think at a specific case, I mean that having, in general, a square law enforce the "indipendence" of frames), and if in time accel and near planets (or whatever), should be unsafe to not use autopilot.

What I worried about is if you fell the calculus are too complex for the reward, as I guess completely different (...I'm probably taking a mistake): actually (after UI improvements) the big users of computational time are (...sure you guess :) ) graphics and collision detection.

Here an updated profiler output without logs in Sol system:
[pioneer-profile-20190827_1708.zip](https://github.com/pioneerspacesim/pioneer/files/3546875/pioneer-profile-20190827_1708.zip)

> Frames implicitly apply the gravity from the frame's parent bodies. Consider if you're floating near Mars High, within its reference frame: No dynamic forces are being applied, as the station itself has no gravity. However, you're still orbiting Mars, and you're still orbiting Sol.

...basically this means that Frames are orbiting, isn't?

> If you apply Mars or Sol's gravity in this case, you will rapidly drift out of the station's frame, as their gravity is now being incorrectly applied twice.

Could the application of Centrifugal forces (and Coriolis) lead to stability: I mean, centrifugal forces of Mars on objects in orbits (including children frames) and also from Sol on their children (which are what I try to address in the newest commit)

Thank's for feedback... Hope you have time to check the weight of this change.

mike-f1:
[Just a couple of things I would pin here...]

[The normal approach to orbits](http://www2.mae.ufl.edu/~uhk/PLANETARY-MOTION.pdf)

[A deeper approach](https://en.wikipedia.org/wiki/Three-body_problem)

[Hills sphere and Lagrange points](https://ocw.mit.edu/courses/aeronautics-and-astronautics/16-07-dynamics-fall-2009/lecture-notes/MIT16_07F09_Lec18.pdf)

[An (approximate) solution to three body problem](https://arxiv.org/ftp/math-ph/papers/0402/0402016.pdf)

[Perturbations](https://en.wikipedia.org/wiki/Perturbation_theory)

impaktor:
I know @ecraven developed a N body simulation for eventual inclusion in pioneer, his simulations went well, I think I remember, perhaps there's code somewhere to share?

mike-f1:
> I know @ecraven developed a N body simulation 

Nice but this change *is* a N body simulation: what is missing is a way to (at least) diminish differences between the simulation (= frame by frame integration) and [calculated orbits](https://github.com/pioneerspacesim/pioneer/blob/master/src/Orbit.h), which are used for planets and frames updates.

That should diminish drift between dynamic bodies and frames.

mike-f1:
After scratching the surface, this become a quite big change...

I fear is time to close, back to the drawing board :(

impaktor:
Sadly, I haven't had time to follow the discussions here in-depth, but I applaude you for trying.

Rember, best way to get stuff into master, especially when reviewers are lazy, is small baby-step commits.

Also, if real gravity is implemented, how can we maximize that the player understands that, compared to our current impolementation? A gravity-well visualizer?

mike-f1:
> Also, if real gravity is implemented, how can we maximize that the player understands that, compared to our current impolementation? A gravity-well visualizer?

[Orbit](https://github.com/pioneerspacesim/pioneer/blob/master/src/Orbit.h) : this code should be updated accordingly, actually it provide the right behaviour both from physical than from graphical point of view. If a mathematical solution will come out, then someone should update this file.

PR #4612, opened 2019-06-03
Multiple Internal Guns

mike-f1:
Screenshot from 2019-06-03 20-58-43

...Multiple guns at works...

**Left-click** on gun gauge change activation (grey is off)
**Right-click** on gun gauge change the number of barrels which will fire (check numbers on tooltips)

In `infoview`, it show mounts on player ship and, in equipment section, it show guns installed with their mount number.

Clicking over a gun open a pop where is possible to change the mount for that gun

Guns_Swap

Models and some tweaked file are in the last commit, which should be dropped before merge.

~~TODO:
waiting for UI in order to provide (and test) a `SwapGuns(GunId a, GunId b)` member
function and a suitable way _to see_ the swap, but it could be merged and that will
follow.~~

PR #4380, opened 2018-07-11
WIP: move systemview UI to pigui

ecraven:
Not done yet, still working on it.

PR #4669, opened 2019-08-27
Little improvements

mike-f1:
This is a rib of what was #4668, here first and third commit from that PR:
First commit add centralized checks, while second (ex third) add grid & legs to SystemView.

Let @fluffyfreak check code as in #4668 and @nozmajner (if necessary) optionally replace these icons


mike-f1:
Added 3 commits:
1) `DbgDumpFrame` is now part of Lua Space functions
2) Rework a little ctor/Init on Frames
3) Store Frames (not pointers or references) in container (which, for now, is a `std::list`)

impaktor:
Cool, just tested the grid, it's an improvement.

The commit message says something about "legs"? Does this mean vertical bars connecting body to plane of grid? (I did not see those)

mike-f1:
> Does this mean vertical bars connecting body to plane of grid? (I did not see those)

exactly: you should click the button again, there are three "mode": no grid -> grid -> grid & legs.

> Cool, just tested the grid, it's an improvement.

thx's, little but useful... Not a big change as I saw #4380 which would move all in Lua.

impaktor:
> you should click the button again, there are three "mode": no grid -> grid -> grid & legs.

Doesn't work on my end. I have some fancy Nvidia card, not sure, what, Dell XPS15. Anyway, might be good to see if anyone else can reproduce either it working or not.

mike-f1:
If you start in Sol you should look at Pluto: the other bodies are really near ecliptic...

impaktor:
Ah, I see it now, I was looking at space ships, since these typically are above the orbital plane.

So feature-wise, this should go in, just need to check if someone cares enough to look at the code, if not we'll merge without if no complaint.

impaktor:
What's the use case for the lua method `DbgDumpFrames`?

Also, would be good if someone could glance at the restructure? of the `Frame::` stuff.

Web-eWorks:
@impaktor I'll handle review; I've finally got some time to get on this (and more importantly got to a point where I can swap git trees without losing core-refactor).

src/Frame.cpp
```diff
@@ -90,32 +119,13 @@ void Frame::PostUnserializeFixup(Frame *f, Space *space)
...
-	ClearMovement();
-	m_collisionSpace = new CollisionSpace();
-	if (m_parent) m_parent->AddChild(this);
-	if (label) m_label = label;
-	m_sbody = nullptr;
-}
-
 Frame::~Frame()
 {
-	m_sfx.reset();
```
fluffyfreak: So what handles this now?

src/galaxy/CustomSystem.cpp
```diff
@@ -638,7 +626,7 @@ static void RegisterCustomSystemsAPI(lua_State *L)
 	register_class(L, LuaCustomSystemBody_TypeName, LuaCustomSystemBody_meta);
 }
 
-void CustomSystemsDatabase::Init()
+void CustomSystemsDatabase::Load()
```
fluffyfreak: Init was the convention for this kind of class, you don't need to know what it's doing inside just that things are being initialised. Changing it to Load is currently correct for this method but doesn't declare the intent of it, nor conforms to the others that are like it.

fluffyfreak:
There's some good stuff in there, there's also some stuff that I just don't know why you changed it because it no longer matches the conventions established in the rest of the code.

Also, why change the `Frame` stuff to be allocated/freed within the class rather than as it was? Places still hold pointers but they don't own them. If you were concerned about tracking/freeing usage then just change them use `unique_ptr`, or if you wanted multiple references and checking then our own `RefCounterPtr`. 

Third option would be to use handles, so the `CreateFrame` returns an `int` which is the handle too an internally allocated and managed list of Frames. This change no means that people can forget to free frames just as much if not more than before.

So I'm a little confused by that change. I might mallow it in anyway since it's a nitpick, but I think it's possibly for the worse from an API interface design point of view.

src/Frame.cpp
mike-f1: ...dtor it self, if I'm not wrong: it's a unique_ptr, thus it should call dtor of pointed object automatically

impaktor:
Sounds like we just should cherry pick the commits we want then? Like the grid stuff I like.

src/galaxy/CustomSystem.cpp
mike-f1: I would highlight it _load_ files, not only initialize some value...
Don't know: I can restore it or (rather?) call it 'InitAndLoad'.

mike-f1:
> Also, why change the Frame stuff to be allocated/freed within the class rather than as it was?

I read (and I think) it is better to avoid having `new` and `delete` called around, and 

>Third option would be to use handles, [...snip...]

...that could be the final step (and I move to it), but before doing so, sure I should know if you like that thing.

This only use a (sort of) common pattern on a class which partially use it, which is a little improvement.

src/galaxy/CustomSystem.cpp
fluffyfreak: Nah keep it ,I'm just wondering as to your motivation behind the change, since it doesn't seem necessary or related to the rest of the PR... just something you did that got caught up in it.

fluffyfreak:
Ok if it's going somewhere (_to handles in the future, for example_) then I'm ok with it going in 👍

mike-f1:
> to handles in the future, for example

Indeed, I did some checks: it seems that `Get*RelTo()` members are called thousand of times a frame, thus I wonder if there's a way to measure performance gain in having frames packed tight (in a `std::vector`, for example)... However I fear a `std::vector` can't be used directly because handles are fixed while a vector may change order (due to swap&pop or memory reallocation)... Any suggestion?

mike-f1:
> Sounds like we just should cherry pick the commits we want then? Like the grid stuff I like.

Feel free to do it, I hope you would pick at least the first 3.

> What's the use case for the lua method DbgDumpFrames?

It was handy for this PR to have such a function, also given recent changes which involve Frames.
Anyway, it's matter of policies: if you feel unconfortable with a "Dbg*" function in master you may drop it.

fluffyfreak:
This PR is probably fine to merge, I've just got so much going on with RL at the moment that I have no chance to do Pioneer stuff at the moment.

mike-f1:
> (to handles in the future, for example)

...not that far ;)

mike-f1:
Done, handles here and some "little" improvements on C++ side (or at least I think so...).

@fluffyfreak , if you still haven't time, there's someone else voluentering? ...Don't know, I think to @jaj22 (last commit, especially...) or @Web-eWorks

Web-eWorks:
@mike-f1 this PR has ballooned to a pretty large size. Would you mind separating out the FrameId changes into a separate PR so we can review them there? The smaller improvements seem fine, but I can't actually find them under 1.2k lines of Frame refactoring.

impaktor:
@Web-eWorks if any commit looks like a no-brainer / good to go, I can just cherry pick it into master, if you let me/us know. But sure, separate PRs, and minimum edits for easier review is preferred.

mike-f1:
I can't split things right now, but I will do next days... If you want I suggest to cherry 8c653bd first which is mostly no brainer and useful (...there a change in Space.cpp that could needs more attention but I tested it letting my ship floating over terrain at 1000x and bouncing over a pad).

Once merged 83780e4 and df7146d can be cherried as well.

Remaining commits are on Frame things

mike-f1:
Due to time needed for review(s), I put these changes in that branch:

https://github.com/mike-f1/pioneer/tree/experiments

So people can "Download & Compile & Play", and helps find bugs (due to this change) which come out by testing :)

Please note that that branch also include changes from #4612, as I'm putting all my rubbish in the same bin (not recycling anything, obviously 👼 )

Ping @craigo- & @Cyborgscode , if they have time to take a look and check it will not crash too soon.

Thx's in advance

fluffyfreak:
If you can fix the formatting which seems to be causing issues again then this looks like it's ready to merge, probably.

ElonSatoshi:
If it works locally, why are github's checks failing?

ElonSatoshi:
I'd like to playtest this branch before it's merged. If there's one thing I can do for pioneer, maybe it's to test things out. What improvements should I be looking out for?

mike-f1:
> Merging it once it builds.

Sorry @fluffyfreak : I really prefer to see #4612 merged first, as it is there for much more and I already rebased this on top of that for my shared personal branch... And, after all, it is there from spring and nobody raise cons, so why it still there?

> But the PR is good, I've tested it locally

Nice, it was there for so long and I was thinking if there were reasons :) ...And so lucky: I just fix a bug and you get ready for review! :D

> If you can fix the formatting

err... Frankly I don't think I have time (and will) to go in every files I touched (61!) checking for things like include re-order or namespaces without a "named" closing brace (eg. `}` instead of `} // namespace Graphics`)... You really should put back `autoformat`: It's too much for my little shoulders :/

>I'd like to playtest this branch before it's merged. If there's one thing I can do for pioneer, maybe it's to test things out. What improvements should I be looking out for?

@ElonSatoshi : Thx's :) and If I may suggest, do it on my local branch: so you can help (me) finding bugs and having feedback also for the couple of PRs.

impaktor:
> Nice, it was there for so long and I was thinking if there were reasons 

Probably the many many changes, all bundled up into one PR, makes it difficult to even muster up the energy to review it.

So now that it's been reviewed, _please_ don't push any more new changes, unless it's fixes for the existing commits. This getting merged pretty much depends on fluffy having the energy to follow the myriads of small changes done here.

@Web-eWorks We might need a step by step guide for how to fix formating errors, the steps you outlined in IRC some weeks ago didn't work for me; so basically: how do we use the format checker to fix these files?

@ElonSatoshi That would be most welcome!

PR #4677, opened 2019-09-22
WIP Add "sponsor us" button to github account

impaktor:
I discovered github has something called "[sponsor button](https://help.github.com/en/articles/displaying-a-sponsor-button-in-your-repository)", so I thought I open this PR to discuss if/what to put in there. 

Ideally, we would have a better/shorter paypal.me/user link. ping @robn?

For an example, see top middle on this project:
https://github.com/atlas-engineer/next

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+custom: https://www.paypal.com/donate/?token=T4jFnbKQST8inqHgzVR4hewrB7g9lfemmzP3P79Zd_ec6wL_XXvS0i0CYgBmdvTlBkkGl0&country.x=AU&locale.x=


fluffyfreak:
The only problem with allowing monetary donations is the question, what do we do with it?

robn:
The rare donations that came in from the old homepage were spent on domain name renewals and a beer or two per year to pay for my sysadmin time. Or they would, if they reached that far!

impaktor:
@fluffyfreak well, we already have "donate to us" on the home page, but I guess what rob said.

@robn OK, good to know. If we want a "support us" button, it would be good if you could login to paypal and create some nicer donate url, like a [paypal.me/pioneerspacesim](https://www.paypal.me/) or similar, than the super-long one I have in this PR.

PR #4679, opened 2019-09-25
added a sponser butten, this should help you guys get some more funds for the next update

Burnsedia:


impaktor:
How is this different from #4677? We know the syntaxt for the sponsor button, problem is which acconuts we want to use, if we should use our paypal, and if we should generate a short url for paypal donations.

PR #4682, opened 2019-09-30
Fix for medical emergency mission

### LINKED ISSUE #4601: Medical emergency is broken
### Observed behaviour
Landing next to the target ship where medical attention is needed does trigger a "please allow 0.0 minutes for transfer" message, and then nothing happens.

### Expected behaviour
Something happens.

### Steps to reproduce

Start a new game on Mars. Accept a medical emergency mission. Navigate to the ship in distress. Land asides it.

This is my savefile to make it easier to reproduce the bug: https://app.box.com/s/m0t8kx4mlq7pk9obutgynmwtsshtch74

master branch (4c51f4eff), Debian 9 64-bit

robothauler:
Fixes #4601

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -666,6 +666,7 @@ local createTargetShipParameters = function (flavour)
 	elseif flavour.deliver_crew > 0 then
 		crew_num = rand:Integer(flavour.deliver_crew+1,shipdef.maxCrew)
 		crew_num = crew_num - flavour.deliver_crew
+		deliver_crew = flavour.deliver_crew
 	elseif flavour.pickup_crew > 0 then
 		crew_num = pickup_crew
 	else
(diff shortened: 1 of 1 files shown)


impaktor:
Looks like a no-brainer, so should be safe to merge.

The second fix, is just a forgotten debug line, I take it.

Good work!

Btw, from a code symmetry perspective, there's not a similar bug with `pickup_crew`? (I haven't checked thoroughly, but just seems like those three variables are set/loaded together, but not in that if-statment).

robothauler:
Indeed. Looks like there is also something missing. But at the moment there are no missions with a fixed number of crew people to pick up.

mike-f1:
If I may, I would suggest to glance again https://github.com/pioneerspacesim/pioneer/pull/4579#issue-276010655...

Here the relevant line:

>error: [string "[T] @modules/SearchRescue/SearchRescue.lua"]:544: attempt to perform arithmetic on local **'pickup_crew'** (a nil value)

...Just to pin & share toughts...

robothauler:
OK, just in case. Last commit should work. Not tested. :-)

PR #4709, opened 2019-10-13
Fix midhyperjump game crash

### LINKED ISSUE #4703: midhyperjump game crash..   attempt to index field "system" ( a nil value )
OS: os independent
Version: 2019-10-09

while in a long hyperjump, pressing the "set as navigation target" from the mission windows, crashes the game with "Navbutton.lua:24: attempt to index field "system" ( a nil value )":

Pioneer-midwarpcrashjpg

vakhoir:
Fixes #4703.

The way the NavButton was written, it looks like it was not expected to be pressed in hyperspace. I added some extra checks that fix the issue. I tried to show the option is disabled in the UI as well,but the :Disable method doesn't seem to do anything.

One more thing that came up is a SIGABRT that's thrown when you're not in the main game view as the hyperjump begins (you can switch to a different window when it's in progress). I tried to find the source but it's a pain to debug. I'm pretty sure that's a relatively new issue (wasn't there a PR that let you jump with the sector map open?), but maybe it's something to do with my setup. Can someone test this?

### DIFF
--- a/data/libs/ui/NavButton.lua
+++ b/data/libs/ui/NavButton.lua
@@ -21,11 +21,11 @@ function NavButton.New (text, target)
 		self.widget.onClick:Connect(function ()
 				if self.target:isa("Body") and self.target:IsDynamic() then
 					Game.player:SetNavTarget(self.target)
-				elseif self.target:IsSameSystem(Game.system.path) then
+				elseif Game.system and self.target:IsSameSystem(Game.system.path) then
 					if self.target.bodyIndex then
 						Game.player:SetNavTarget(Space.GetBody(self.target.bodyIndex))
 					end
-				else
+				elseif not Game.InHyperspace() then
 					Game.player:SetHyperspaceTarget(self.target:GetStarSystem().path)
 					-- XXX we should do something useful here
 					-- e.g. switch to the sector map or just beep


Web-eWorks:
@vakhoir I'll try to find the time to track that bug down. SIGABRT seems to be thrown most commonly in the case of a bad alloc or sometimes a double-free; so it's likely that not being in the game view is making *something* allocate a negative-sized array or whatever.

mike-f1:
Just a note:
hyperspace 'thing' is controlled by these two methods in [`Game.cpp`](https://github.com/pioneerspacesim/pioneer/blob/master/src/Game.cpp#L482):

https://github.com/pioneerspacesim/pioneer/blob/653b1a88ff2ac71a9acfac442a076bc3c5372640/src/Game.cpp#L407

https://github.com/pioneerspacesim/pioneer/blob/653b1a88ff2ac71a9acfac442a076bc3c5372640/src/Game.cpp#L482

Thus I wonder if there's a way to fix everything (I mean all the views from Lua) finding what's change between these calls and, more specifically, the calls made to `Space` ctors.

fluffyfreak:
I'm going to merge this for now. Looks like there is other work that needs to be done as far as the `SIGABRT` fault but that can come separately.

PR #4717, opened 2019-10-15
Decouple

mike-f1:
A(nother) rib of #4669:

...decouple includes, remove a couple of variables, fix an equation in Space.cpp (already tested, but who knows? ;) ).

EDIT: It took also commit 95d80c9... I hope isn't a problem...


fluffyfreak:
@mike-f1 the static checks failed but once those are fixed I'll merge it 👍

Web-eWorks:
Looks pretty good to me, the only issue I have is that the `sort_BodyAttrs` function may not be able to be inlined, which would hurt performance. I'd profile that change if I were you, see how it affects drawing in high-complexity scenes.

mike-f1:
> once those are fixed I'll merge it +1

...And I will fix those when someone fix the hook :ok_hand: :weary: or give some hints: `reset HEAD^1` and re-commit (68 files) don't work and either `./autoformat` don't have any effect...

>sort_BodyAttrs function may not be able to be inlined, which would hurt performance

Yeah, it can't be inlined but I don't think it will "hurt" performances: it is called only on visible and not billboard-ed Body which rarely will be more than a couple... I tried to find a way to persuade C++ that that funtion would be inlined but I didn't find the "correct reasoning" in order to call a friend in a nested class of a (nested) namespace of a....
Too much at that time, a lambda may suffice next time.
I left a comment to highlight what you found.

@Web-eWorks : pls consider that including the `Body.h `in `Camera.h` means you build a bridge between physical objects and rendering things, so compilation times are **really** growing...

fluffyfreak:
@mike-f1 I'm honestly not sure, this is the log https://travis-ci.org/pioneerspacesim/pioneer/jobs/598815426 and when it happened to me I went through and checked the formatting around the lines where each log section ended, but I only had one and got lucky.

@Web-eWorks any hints on how to figure out what it's complaining about? The output saying that it fails isn't very helpful to an untrained eye.

impaktor:
@Web-eWorks I tried:
```
TRAVIS_BRANCH=master ./autoformat
```
from this branch, but I still get the git diff-index help dialouge.

fluffyfreak:
If I format the code in MSVC it rearranges the order of the include's in `Model.cpp`, `Geom.cpp` etc.

So is that not being done by the the autoformatter script?

fluffyfreak:
src/collider/Geom.cpp | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/collider/Geom.cpp b/src/collider/Geom.cpp
index ff3d2a9ee..6ddce3ac7 100644
--- a/src/collider/Geom.cpp
+++ b/src/collider/Geom.cpp
@@ -4,9 +4,9 @@
 #include "Geom.h"
 
 #include "BVHTree.h"
-#include "GeomTree.h"
 #include "CollisionContact.h"
 #include "CollisionSpace.h"
+#include "GeomTree.h"
 
 #include <float.h>
 
----------------------------------

 src/scenegraph/Model.cpp | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/src/scenegraph/Model.cpp b/src/scenegraph/Model.cpp
index 8c0397309..c19ec409e 100644
--- a/src/scenegraph/Model.cpp
+++ b/src/scenegraph/Model.cpp
@@ -10,13 +10,13 @@
 #include "NodeCopyCache.h"
 #include "StringF.h"
 #include "Thruster.h"
-#include "graphics/Renderer.h"
 #include "graphics/RenderState.h"
+#include "graphics/Renderer.h"
 #include "graphics/TextureBuilder.h"
 #include "graphics/VertexArray.h"
 #include "scenegraph/Animation.h"
-#include "scenegraph/MatrixTransform.h"
 #include "scenegraph/Label3D.h"
+#include "scenegraph/MatrixTransform.h"
 #include "utils.h"
 
 namespace SceneGraph {

mike-f1:
Yes, it may be and this time it will suffice... But this is the first of several commits which could or couldn't be well formatted.

Thus I wonder if that tool is not working for me or for everyone.
In the first case, I can read these logs and intervene where needed.
In the latter case, I wonder if that tool should be better disabled until fixed, because a lot of effort and time is needed to manually format everything.

fluffyfreak:
Taken the view that I'm just going to merge it and we can fix the formatting afterwards.

impaktor:
Don't we have the same problem? I.e. get the script to show us which lines are the offending?

mike-f1:
[Wiki](https://pioneerwiki.com/wiki/Coding_Conventions) state that a `git commit` should format things.

But I copied the git hook in the directory and it doesn't work, apparently.

Web-eWorks:
@mike-f1 Yeah, I need to sit down and go over the format hook again. Seems things are breaking a little. In related news, if you're using `clang-format` version 7.0.x, it has some breaking issues with misidentifying header files; updating LLVM and/or clang-format should help.

mike-f1:
I can provide some more info on my setup, hope it helps:

`clang-format --version` output is:
`clang-format version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)`

but I use `gcc` (not `clang`), and its version is:
`gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0`

PR #4721, opened 2019-10-17
Add % to "HUD_DELTA_V_PERCENT"

### LINKED ISSUE #4684: Percent Fuel: HUD value different to Lobby value
### Observed behaviour
When the ship's fuel level is not completely full or completely empty, the percent fuel shown in the lobby is not the same as the percent fuel shown on the HUD. The value displayed on the HUD is always a little bit more.

### Expected behaviour
I think they should agree - unless I am missing something? 

### Steps to reproduce
1. Start on Mars (or anywhere)... The percent fuel displayed on the HUD is 100%
image

2. Go to the lobby, reduce fuel from 100% to 50%
image

3. Switch to world view and take note of the percent fuel displayed on the HUD (56%)
image

Shouldn't they be the same? Or is the HUD value something else? The [wiki's description](https://pioneerwiki.com/wiki/Flight_UI) of the value is "It shows the % of the tanks"...

Pioneer 20190203 (latest release)
Pioneer master (6fd8f02f69f51ed772e6d0f6888491feba60dbf6), approx. 20190915

craigo-:
Fixes #4684. Differentiating the messages for "HUD_DELTA_V_PERCENT" and "HUD_DELTA_V" to hopefully make it more clear that the percentage value is not the fuel load as shown in the lobby.

Initial commit changes only those 12 language files where the string is in English.

Happy to discuss other wording if people think this is suboptimal.

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -5,6 +5,7 @@ October 2019
    * Fixes
      * Fix for medical emergency mission (#4682)
      * Fix midhyperjump game crash (#4709)
+     * Hang up chat on visibility change (#4710)
 
    * Internal Changes
      * cmake: Don't install .gitignore files (#4683)


impaktor:
You need only update the `en.json` file, the rest are automagically fixed by translation system, so would be good if you could remove the other files from this pull request, thanks!

Maybe the text should even read `(% of maximum delta-v)` or similar? @nozmajner - thoughts?

nozmajner:
@impaktor Sounds good to me.

craigo-:
Sorry, I couldn't seem to roll back my fork without wiping out this PR; I'll raise another one.

PR #4722, opened 2019-10-17
Optimise sector view and docked face generation

fluffyfreak:
Profiling some unrelated behaviour showed that we had some expensive behaviour going on.

`SectorView` was doing a lot of work displaying labels even when labels weren't being rendered. When they were being rendered there was still a lot of needless work happening. This mostly just moves around some of the tests and checks so it can early out more often.

Second we were rebuilding the station managers face every single frame!
That was obviously very bad and could account for anywhere from 10% to 25% of a frames CPU time! Not only that but it was thrashing texture generation etc.
This was a little tricky to track down and I had to add some additional profiling code and a Lua utility function called `pi_lua_stacktrace` that let me see how the call came up through the Lua callstack.

I've added a semi-dummy call to `pi_lua_stacktrace` in the LuaManager. That way I can use the "immediate" window in the MSVC debugger to get any Lua callstack that I like whenever I want it. Which is amazing and I wish I'd had this for years! 😁


impaktor:
Good work!

PR #4725, opened 2019-10-26
Missing Debian/Ubuntu Dependency

h54:
A build from source is not successful without `libfreetype6-dev`.
This package is available in in the Debian stable repos as well as Ubuntu repos back to 16.04.


Web-eWorks:
Looks good to merge to me!

PR #4726, opened 2019-10-27
Extend NameGen name lists

WKFO:
Mostly self-explanatory.

- The name 'Ring' is added because all orbital spaceports currently in Pioneer have at least one. If other starport types are added in future, may need to change.

- Can add more (or delete some) on request


data/libs/NameGen.lua
```diff
@@ -328,6 +329,10 @@ NameGen.orbitalStarportFormats = {
 	"{name} Outpost",
 	"{name} Citadel",
 	"{name} Platform",
+	"{name} Ring",
+	"{name} Residance",
```
Web-eWorks: Bit of a typo here...

```suggestion
	"{name} Residence",
```

WKFO: Oops. Got confused because of my mother language :P

robothauler:
Here are a few more orbital station names:
"{name} Trade Center",
"{name} Science Station",
"{name} Facility",
"{name} Site",
and 2, more ore less, odd city names:
"{name} Oasis",
"{name} Castle",
and for the small rocky planets e.g.
"{name} Settlement",
"{name} Commercial Area",

Please pick up if you like it.

WKFO:
If only there was a way to assign station names according to the system's economic activities...

impaktor:
1. To fluffy: if you merge, make sure to select "squash and merge".
2. FYI, I have a branch with _lots_ of [new names](https://github.com/impaktor/pioneer/tree/cultures/data/culture), idea was to make different regions have names from different places, so it's more like landing in a strange place where eveyrone has "pappadoppulus"-names, eg. Needs some code to assign culture mix to each system (or planet/station), if anyone wants to pick it up.
3. I assume this will be a save breaking change? If stations change name in the middle of a mission.

WKFO:
I don't think the change affects previously generated galaxies (my old save worked) but there is a good chance that I am wrong.

impaktor:
@WKFO the scenario I'm thinking of is: when this is merged, the stations will change name (I think), and if you have a saved mission to "Hopper High", that might now be called something else.

WKFO:
Tested a few times. Yes, the stations do change names... However it should be relatively okay(?) since it changes name on the mission screen as well.

impaktor:
> However it should be relatively okay(?) since it changes name on the mission screen as well

Ah, OK, then it should be fine.

impaktor:
```
"{name} Residence",
"{name} Trade Center",
"{name} Commercial Center",
"{name} Science Station",
"{name} Observation Post",
"{name} Industrial Center",
"{name} Refinery",
```

If I'm to play devil's advocate, then these names will lead the player to believe industrial goods are cheaper / more available at a "Industrial Center", and high tech stuff at a "Science Station", etc. (Luckily, there's no `"{} Military Station"`, as that really aught to have military tech level).

With that said, I don't recommend any further action, other than squashing the commits (can be done by person who pushes green button).

WKFO:
There must be a way to generate those station names based on system's economic activities

impaktor:
> There must be a way (...right?) to generate those station names based on system's economic activities.

Yes, quesiton is how much C++/Lua plumming is needed. If you find the time to look into it, that would be very nice.

Thanks for the PR!

WKFO:
It will take some free time and a smack of enthusiasm to get me going. I hate it when there is no Lua equivalent/reference of something in C++ and vice versa.

Hope this added some nice change to the game in the meanwhile.

PR #4735, opened 2019-11-16
Use system time at start

fluffyfreak:
When starting a new Game offset the start time from the year 2000, specifically Friday, 31 December 1999 23:59:59 so that the date and time are exactly 1200 years from whenever they start the game.

This is only done when starting a new game, saved games are not affected.

When testing it today my starting date and time were: 3219, Nov 16, 14:47

### DIFF
--- a/src/Game.h
+++ b/src/Game.h
@@ -59,8 +59,8 @@ public:
 	void ToJson(Json &jsonObj);
 
 	// various game states
-	bool IsNormalSpace() const { return m_state == State::STATE_NORMAL; }
-	bool IsHyperspace() const { return m_state == State::STATE_HYPERSPACE; }
+	bool IsNormalSpace() const { return m_state == State::NORMAL; }
+	bool IsHyperspace() const { return m_state == State::HYPERSPACE; }
 
 	RefCountedPtr<Galaxy> GetGalaxy() const { return m_galaxy; }
 	Space *GetSpace() const { return m_space.get(); }
@@ -171,8 +171,8 @@ private:
 	std::unique_ptr<Player> m_player;
 
 	enum class State {
-		STATE_NORMAL,
-		STATE_HYPERSPACE,
+		NORMAL,
+		HYPERSPACE,
 	};
 	State m_state;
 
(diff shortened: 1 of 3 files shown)


fluffyfreak:
Good point, guess that would need updating, wonder if there's any other hardcoded date entries it might mess up

src/Game.h
```diff
@@ -170,7 +170,7 @@ class Game {
 
 	std::unique_ptr<Player> m_player;
 
-	enum State {
+	enum class State {
 		STATE_NORMAL,
 		STATE_HYPERSPACE,
```
impaktor: I don't know what the naming convention is, but perhaps the variable don't need "STATE" prefix when it's given by the class, e.g. `state::STATE_NORMAL`?

src/LuaGame.cpp
```diff
@@ -63,8 +63,13 @@ static int l_game_start_game(lua_State *l)
 	}
 
 	SystemPath *path = LuaObject<SystemPath>::CheckFromLua(1);
-	const double start_time = luaL_optnumber(l, 2, 0.0);
```
impaktor: A short API documentation compatible paragraph might be nice, or is this only for internal use? Any chance someone else might want to use this?

impaktor:
I should say, I think checking system time might be fun/useful for easter eggs, like putting a santa hat on BBS NPCs if it's 24-26 december, and things like that.

src/LuaGame.cpp
fluffyfreak: It's an old function that is documented just above. It's always been in there but I don't think anyone uses it.

src/Game.h
fluffyfreak: Good point I'll change that!

fluffyfreak:
As for the forum, just change the question to ask what century? Then we'll be good for the next 81 years ;)

fluffyfreak:
Or just change it every year, so this right now it'll be 3219, next year it will become 3220!

zonkmachine:
> Or just change it every year, so this right now it'll be 3219, next year it will become 3220!

So this is why I had problems registering to the wiki and forum. :)
Maybe instead ask what ship you start with on Mars?

impaktor:
@zonkmachine indeed, both wiki and forum need annual upadte of the security question. You've been activated on the forum now, btw.

PR #4737, opened 2019-11-17
Fix OSX Travis deployment

fluffyfreak:
OSX cp doesn't allow use of -r or -t, -r can be replaced with -R and -t can be avoided

This was causing the OSX Travis build to fail to deploy, or at least failing to copy files
https://travis-ci.org/pioneerspacesim/pioneer/jobs/613081290#L1006

### DIFF
--- a/scripts/build-travis.sh
+++ b/scripts/build-travis.sh
@@ -25,15 +25,19 @@ fi
 echo "Copied binaries."
 
 # Copy the text files.
-ls *.txt | grep -v -E '(COMPILING\.txt|SAVEBUMP\.txt|CMakeLists\.txt)' | xargs cp -t $COPY_DIR
+cp AUTHORS.txt $COPY_DIR
+cp Changelog.txt $COPY_DIR
+cp Modelviewer.txt $COPY_DIR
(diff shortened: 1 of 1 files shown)


scripts/build-travis.sh
```diff
@@ -25,15 +25,19 @@ fi
...
+
 # Copy the licenses
-cp -r licenses -t $COPY_DIR
+cp -R licenses $COPY_DIR
 
 echo "Copied text files."
 
 # Copy the text files
-cp -r data -t $COPY_DIR
+cp -R data $COPY_DIR
```
impaktor: As far as I know, capital and lowercase `r` should do the same thing. Or OSX is different?

fluffyfreak: OSX doesn't officially support lowercase version

impaktor: OK

orbea: Seems osx doesn't follow the posix spec here which does include `-r` and `-R`.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/rm.html

PR #4738, opened 2019-11-17
Add AUTH token to Appveyor

fluffyfreak:
This adds a token for the Appveyor build, no idea if that's correct or not.

Is this how you do this? I have no fucking idea. [This page](https://github.com/forexample/github-binary-release) seems to indicate that it might be?


appveyor.yml
```diff
@@ -11,4 +20,4 @@ install:
     - C:\msys64\usr\bin\bash -l /c/projects/pioneer/CI/appveyor/%TARGET%/prepare.sh
 
 build_script:
-    - C:\msys64\usr\bin\bash -l /c/projects/pioneer/CI/appveyor/%TARGET%/build.sh
+    - C:\msys64\usr\bin\bash -l /c/projects/pioneer/CI/appveyor/%TARGET%/build.sh
```
impaktor: No new line?

impaktor:
The commits make it unclear what the token is. Is it an appveyor encryption of the travis token? I assume not, as that probably wouldn't make much sense.

fluffyfreak:
The token is a personal access token I generated and encryted using the appveyor encrypter page, so it's only useable by appveyor, the thing is that it's almost completely undocumented as to **_how_** you add it to the yaml script.

appveyor.yml
fluffyfreak: Yeah I don't know what happened there, I'll fix it when/if I have to update the file agian

fluffyfreak:
Closing as the bug is unrelated to auth tokens, which have also been revoked from my account

PR #4739, opened 2019-11-17
ASAN/valgrind fixes

johnbartholomew:
Fix a few `new[]`/`delete` mismatches and places where we read from memory that hasn't been initialized. Mostly found by valgrind. I tried to get an ASAN build working too but it was a pain(*).

There are probably a lot more - running under valgrind is too slow to interact with the game really so this is just stuff that showed up when running modelcompiler and right at the start of a new game. I didn't, e.g., click around all the different views.

(*) E.g., sigc++ 2 is not ASAN-clean. https://github.com/libsigcplusplus/libsigcplusplus/issues/10

### DIFF
--- a/src/LZ4Format.cpp
+++ b/src/LZ4Format.cpp
@@ -62,14 +62,14 @@ std::string lz4::DecompressLZ4(const char *data, size_t length)
 	return out;
 }
 
-std::unique_ptr<char> lz4::CompressLZ4(const std::string &data, const int lz4_preset, std::size_t &outSize)
+std::unique_ptr<char[]> lz4::CompressLZ4(const std::string &data, const int lz4_preset, std::size_t &outSize)
 {
 	PROFILE_SCOPED()
 	LZ4F_preferences_t pref = LZ4F_INIT_PREFERENCES;
 	pref.compressionLevel = lz4_preset;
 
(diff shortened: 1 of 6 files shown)


src/LZ4Format.cpp
```diff
@@ -62,14 +62,14 @@ std::string lz4::DecompressLZ4(const char *data, size_t length)
...
-std::unique_ptr<char> lz4::CompressLZ4(const std::string &data, const int lz4_preset, std::size_t &outSize)
+std::unique_ptr<char[]> lz4::CompressLZ4(const std::string &data, const int lz4_preset, std::size_t &outSize)
 {
 	PROFILE_SCOPED()
 	LZ4F_preferences_t pref = LZ4F_INIT_PREFERENCES;
 	pref.compressionLevel = lz4_preset;
 
 	std::size_t compressBound = LZ4F_compressFrameBound(data.size(), &pref);
-	std::unique_ptr<char> out(new char[compressBound]);
+	std::unique_ptr<char[]> out(new char[compressBound]);
```
Web-eWorks: ...ah, I've been writing too much C. I forgot that `char[]*` and a `char *` to some length of `char`s are two different types.

fluffyfreak:
Thanks @johnbartholomew

PR #4743, opened 2019-11-19
Modal save/load window

### LINKED ISSUE #4718: Can click through Save dialogue and select objects behind
### Observed behaviour
When viewing outside and you bring up the Save dialogue, clicks on the dialogue can select objects behind the dialogue.

### Expected behaviour
The Save dialogue should be 'on top' and be the recipient of all clicks.

### Steps to reproduce
1. Here I am on Tranquility Base. Note the position of Pluto:
image

2. I've pressed `Esc` to bring up the in-game menu, then pressed `Save`:
image

3. I'm trying to click the `Save` button, but instead it is clicking through and bringing up the menu for clicking on Pluto:
image

Pioneer: master (3dff8caea6bad11d0e49a0afec513ee6a3ebb68b), approx. 20191017, Release x64

[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/3737057/opengl.txt)

### LINKED ISSUE #4720: Crash when Load Game dialogue selected after New Game started
### Observed behaviour
Crash to desktop when you are in the Load dialogue and you select one of the `Start on` buttons, but continue to try to load a game.

### Expected behaviour
That the game does not CTD. Probably the best way to deal with this is to not continue to present the initial buttons when presenting the 'Load' dialogue (as it was in the old UI in version 20180203).

### Steps to reproduce
1. Start game, with save games present
2. Click `Load game`
3. Click `Start on Mars` (you probably shouldn't be able to do this):
image

4. The game starts, but the 'Load game' dialogue persists. Select a saved game and click `Load`:
image

5. The game crashes to desktop with the following guru meditation error:
image

NB: clicking the `Cancel` button at Step 4 works fine.

Pioneer: master (3dff8ca), approx. 20191017, Release x64

NB: a very similar guru meditation error happens with Pioneer 20190203. You cannot reproduce the error in Pioneer 20180203, because the `Start on` buttons are not there when the Load dialogue is present.

vakhoir:
Fixes #4720 and fixes #4718

Had to test it on an old branch because of the DeleteFrame issue (I can't even start a new game as is), so maybe someone can double check, but it's pretty straight forward.

Come to think of it, maybe this can wait a day or two more for the merge. and I'll see if there are any other windows that should be modal.

### DIFF
--- a/data/pigui/modules/saveloadgame.lua
+++ b/data/pigui/modules/saveloadgame.lua
@@ -17,6 +17,7 @@ local player = nil
 local colors = ui.theme.colors
 local icons = ui.theme.icons
 local pionillium = ui.fonts.pionillium
+local popupOpened = false
 
 local mainButtonSize = Vector2(40,40) * (ui.screenHeight / 1200)
 local optionButtonSize = Vector2(125,40) * (ui.screenHeight / 1200)
@@ -80,7 +81,7 @@ local function showSaveFiles()
 		table.sort(files, function(a,b) return (a.mtime.timestamp > b.mtime.timestamp) end)
 		ui.columns(2,"##saved_games",true)
 		for _,f in pairs(files) do
-			if ui.selectable(f.name, f.name == selectedSave, {"SpanAllColumns"}) then
+			if ui.selectable(f.name, f.name == selectedSave, {"SpanAllColumns", "DontClosePopups"}) then
 				selectedSave = f.name
 			end
 			if Engine.pigui.IsItemHovered() then
@@ -97,8 +98,10 @@ local function showSaveFiles()
 end
 
 local function closeAndClearCache()
+	ui.closeCurrentPopup()
 	ui.showSavedGameWindow = nil
 	saveFileCache = {}
+	popupOpened = false
 end
 
 local function closeAndLoadOrSave()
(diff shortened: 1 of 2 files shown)


vakhoir:
I went over the windows, and it seems the only one that would make sense to modalize is the in-game options menu, which could be combined with an auto-pause. Any opinions on that?

I also found a bug with this PR. Turns out that if you open the load game window in the main menu, clicking save/load in-game will open it in the main menu again (so: it will not open in game, but if you quit, it appears in the main menu!), and vice versa. I'll need to spend a bit more time here.

vakhoir:
Ok, the bug is solved, turned out to be just an issue with when `openPopup` was called.

If there's no objections I'd also do the modal options window with auto-pause thing.

Web-eWorks:
I'd actually prefer that the in-game options menu was full-screen rather than a modal window, though I suppose the two can coexist. I'd like something more along the lines of the Elite: Dangerous options menu than what we have now.

impaktor:
Many of the reasons for the UI is just because "that's how frontier was", because pigui or newUI just duplicated whatever it was replacing, which was a Frontier UI clone.

I guess there's logic in making game controling windows full frame, as no information / game views are needed simultaneously.

Web-eWorks:
Controls
This is what I was talking about as the fullscreen menu goes - it's not actually a separate screen, but it does blur the underlying game view quite heavily.

Full album of the menu [here](https://imgur.com/a/HpVWRQ4).

vakhoir:
Turned out the autopause was already built in into the options window (though I could have sworn it wasn't originally, oh well), so I just made the options window modal.

PR good to go as far as I'm concerned.

@Web-eWorks, yeah that's definitely the way to go, but we'll have to take it one day at a time ;)

Web-eWorks:
Code looks good to me; @impaktor this should be good to merge as-is.

mike-f1:
I'm now unable to open a "Save/Load" window... Don't know but it can be related to this PR.

Would you check if I'm wrong?

Web-eWorks:
Yeah, I can confirm the bug. I'd bet it's related to the pigui `openPopup` implementation being written with only a single popup in mind. I'm in the process of upgrading the ImGui implementation over the weekend, so I'll get that fixed along the way.

I thought this PR was simple enough that I didn't need to pull it locally and test it myself; it's literally fifteen new lines.
I thought wrong.

vakhoir:
Ah, crap. I tested the save/load **before** modifying the options window.
I can bring back the options window to be non modal, a modal save/load should be enough to prevent any major problems.

Issue #4741, opened 2019-11-18
Segmentation Fault when requesting launch


beltsonata:
Hi,
I've not been having much luck with stability so far but I managed to reproduce this one and get a stack trace. 

**Issue**

The game crashes with a segfault.

**Steps to reproduce**

1. Load up the the save game file attached to this ticket, called '8.zip'. 
2. Unpause time and make a launch request. 
3. The game crashes immediately with a segmentation fault. 

**Pioneer Version**

I built Pioneer from source. According output.txt, version is 20191117 (4dd3918b9). I'm running Xubuntu 18.04.1 on an old Dell Latitude E-6420. Specs:

8Gb RAM
Intel Core i5-2540M @ 4x 3.3GHz
NVS 4200M / nvidia proprietary driver 390.116
Resolution: 1600x900 (I run Pioneer fullscreen)

The stacktrace is from gdb. After hitting 'r' all I did was perform the steps above to reproduce the crash, then when the segfault occurred I got the backtrace.

Let me know if you need any other files or any other information.

[stacktrace.txt](https://github.com/pioneerspacesim/pioneer/files/3860787/stacktrace.txt)
[output.txt](https://github.com/pioneerspacesim/pioneer/files/3860789/output.txt)
[8.zip](https://github.com/pioneerspacesim/pioneer/files/3860797/8.zip)
EDIT: added [opengl.txt](https://github.com/pioneerspacesim/pioneer/files/3860849/opengl.txt)

Web-eWorks:
Oh, I was wondering how fast we'd find the first bug related to the FrameID changes. Thanks for tracking this one down!

The cause of the segfault is fairly simple - `DeleteCameraFrame()` doesn't handle being passed invalid frames or frames without parents. However, this behavior is being caused by a deeper problem - we're double-deleting the camera frame in the `SaveGame()` routine.

I'll get this one cleaned up and ready to go in a day or two.

@fluffyfreak in retrospect, maybe I should have dug into that FrameID PR further... I'm thinking we need to update the contribution guidelines wrt. enforceable PR size.

impaktor:
Is this related to the PR we merged from @mike-f1?

fluffyfreak:
Yeah, I'll take a look shortly unless @Web-eWorks beats me too it

mike-f1:
[...]

PR #4745, opened 2019-11-21
Promote FrameId to wrapper and fix crash, remove unused test

Web-eWorks:
In the process of fixing #4741, I've cleaned up a couple pain-points related to the FrameId type and made it much more difficult to inadvertently write incorrect code.

- Removed a test that hadn't been updated in quite some time and isn't used
- Fixed another test to refer to the new render. I don't know if it runs, but at least my editor isn't complaining.
- Normalized `<stdint.h>` to `<cstdint>`
- Made FrameId a wrapper over an int instead of a raw integer
- Fixed a crash in Frame::DeleteCameraFrame()
- Made CameraContext::GetCamFrame() return the invalid frame index when there
  is no camera frame
- Fixed a typo related to LuaEngine's autosave binding

The autosave typo fix is here because it's a one-line fix and I found it while tracking the original bug down.


Web-eWorks:
@vakhoir try this PR - if you still get a crash, I'd appreciate it if you could send me the stacktrace.

src/FrameId.h
```diff
@@ -1,10 +1,37 @@
...
+	constexpr int id() const { return m_id; }
+
+private:
+	int m_id;
+};
+
+static_assert(sizeof(FrameId) == sizeof(int) && alignof(FrameId) == alignof(int),
+	"Error: FrameId sized differently than the underlying type on this platform!");
 
 constexpr FrameId noFrameId = -1;
```
mike-f1: would it be better to move `noFrameId` definition on top and use it to define `FrameId.Invalid`?

As:
```

constexpr FrameId noFrameId = -1;

struct FrameId {
	static constexpr int Invalid = noFrameId;
```

src/FrameId.h
```diff
@@ -1,10 +1,37 @@
...
+	constexpr bool operator!=(FrameId rhs) const { return m_id != rhs.m_id; }
+
+	constexpr bool valid() const { return m_id > Invalid; }
+	constexpr int id() const { return m_id; }
+
+private:
+	int m_id;
+};
+
+static_assert(sizeof(FrameId) == sizeof(int) && alignof(FrameId) == alignof(int),
```
mike-f1: it is needed? ...just wondering why

Web-eWorks: Ah, that one's future proofing to ensure that nothing increases the size or alignment requirements of the struct. Functionally a `FrameId` is just an int with some extra methods, and I want to make sure it stays that way.

src/FrameId.h
Web-eWorks: Somewhere between writing and bugfixing, I forgot that I intended to go through the rest of the code and get rid of IsIdValid and noFrameId. The current incarnation is just a patch to keep the rest of the code working while I rewrote FrameId.

src/Frame.cpp
```diff
@@ -269,17 +268,22 @@ FrameId Frame::CreateCameraFrame(FrameId parent)
...
-			Error("DeleteCameraFrame: seems camera frame is not the last frame!\n");
-			abort();
-		};
-	#endif // NDEBUG
+	if (parent)
+		parent->RemoveChild(camera);
+
+// Call dtor "popping" element in vector
+#ifndef NDEBUG
+	if (camera != s_frames.size()) {
```
vakhoir: I have to disclaim that this was me punching the code until it compiled and the game stopped crashing, and I have no idea what the broader context of these changes is, but shouldn't this be something like:
`if (camera != s_frames.back().GetId()) {`
Similarly at line 200:
`assert((s_frames.size() - 1) != f->m_thisId);`
should be:
`assert(s_frames.back().GetId() != f->m_thisId);`

Seems like otherwise the comparisons are to the size of the vector rather than the frameId?

In any case the game seems to work for me after I made these changes.

mike-f1: > ...what the broader context...

Is a check made because `camera` is recreated every frame, thus I would check it was the last on that `vector`.

> Similarly at line 200:

This was to ensure save and load use the same ordering, as comment on previous line.

Probably I got a case of "off-by-one" error... But I don't know why I don't spot it before :/

Web-eWorks: @vakhoir the FrameId as currently implemented is a raw index into the vector of Frame pointers. So to check if a FrameId is the last one in the list, a comparison of `camera == s_frames.size() - 1` would be correct; `s_frames.size()` is the index one-past-the-end of the vector. There's no need to dereference the pointer into the actual Frame to check if it's the last one.

That being said, I'll go over it with a fine-toothed comb to see if I can find off-by-one errors like in DeleteCameraFrame().

mike-f1: > That being said, I'll go over it with a fine-toothed comb to see if I can find off-by-one errors like in DeleteCameraFrame().

@Web-eWorks You can narrow that search: it seems both are under debug clause (one in an `assert`-ion and one in `#ifndef NDEBUG`)...

vakhoir: I just checked, and yeah, it's on off-by-one in line 282, the other line I quoted even has a -1 when comparing to the vector size. 

The DEBUG clause could explain another thing I was wondering about - how was this compiling for you guys. Because the comparison seems to be vector size to FrameId (as in: int to struct), rather than index to index. My compiler protested, anyway.

mike-f1: > My compiler protested, anyway.

May I ask if with some warning or if it stop until you change something?

vakhoir:
```
pioneer/src/Frame.cpp:200:32: error: ambiguous overload for ‘operator!=’ (operand types are ‘std::vector<Frame>::size_type {aka long unsigned int}’ and ‘FrameId’)
   assert((s_frames.size() - 1) != f->m_thisId);
          ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
pioneer/src/Frame.cpp:200:32: note: candidate: operator!=(std::vector<Frame>::size_type {aka long unsigned int}, int) <built-in>
pioneer/src/Frame.cpp:200:32: note: candidate: operator!=(std::vector<Frame>::size_type {aka long unsigned int}, size_t {aka long unsigned int}) <built-in>
```

I missed the int() and size_t() cast operator overloads earlier, the code does make sense at fase value. It might be a gcc thing, or a question of some obscure settings.

mike-f1: > It stopped until I explicitly called the .id() on the FrameIds

...which allow compiler to perform an **implicit** conversion outside FrameId thing... :D 

(or at least I guess so, I'm AFK right now...)

Anyway, W C++!!!

Web-eWorks: I *could* have added an implicit conversion to `std::size_t`, but the problem with that is that invalid values would map onto the valid range by accident. It might work to `return m_id < 0 ? size_t(-1L) : m_id`, but it's easier to make the conversion explicit so that the programmer has to think about what they're doing.

mike-f1: Well, I would say that I'm really happy when I see this kind of code. Personally I deferred it as I didn't know how much it will take to get it right so I give it up and use only "plain-old-basic functionality"...

That said, I would ask if you mind to open a dedicated PR only for the bugfix as it should merge sooner. You then gain time to refine this an have a glance on my experimental branch and include some (or all) the optimization I did there...

Briefly: I put 0 (zero) in FrameId as RootFrame and this let me decouple Space and lead to some more optimizations

Web-eWorks:
@mike-f1 I'd be more than happy to include your improvements in in a future PR!

With the resolution of vakhoir's bugs, this PR is now feature-complete and I'd like to merge it as-is. I'll be coming over the code in another, future PR to double-check that things are working as intended and make some minor improvements (a free-list or a better way of preserving/updating FrameIds across serialization, etc.) but that's beyond the scope of this one.

@impaktor this should be ready for merge; in lieu of fluffy taking a look, I'll ask @vakhoir to confirm that it builds and runs fine.

vakhoir:
> I'll ask @vakhoir to confirm that it builds and runs fine.

Yeah, looks good now!

Issue #4744, opened 2019-11-20
Crash on launch - pigui error - appveyor installer used for updating


nozmajner:
### Observed behaviour
The 2019 11 20 appveyor build (Appveyor says, it's 1.0.629) crashes with a guru meditation error:
![crash_pigui](https://user-images.githubusercontent.com/4182678/69267191-f9c6a300-0bcc-11ea-9aec-af6b91b1a295.png)
It did this when I updated the 20190203 release with the installer from appveyor. Uninstalling, and installing the appveyor build fresh fixed the problem, so I guess some files aren't updated if you install on an existing install.
No mods were active.
### Steps to reproduce
Install appveyor build on top of 20190203.

__My pioneer version (and OS):__
Windows 10.

Web-eWorks:
Oh yeah, this is happening because I moved `pigui/modules/gauges.lua` into `pigui/modules/flight-view/gauges.lua` and therefore the old file is kicking around in your install.

The installer really should delete the current install when upgrading to avoid this kind of issue in the future.

Web-eWorks:
CC @pcercuei...? I think you're the one who handling the windows installer.
For good measure, cc @fluffyfreak too.

pcercuei:
Disagreed. Pioneer should just ignore old files.

impaktor:
This might be resolved now, with fix to setup script. We'll notice.

PR #4748, opened 2019-11-25
Inno Setup script: Remove old Pioneer files before new install

pcercuei:
This makes sure that we don't have leftover files from an old install of
Pioneer inside the new install.

UNTESTED.

Suggested in bug report #4744.


impaktor:
Well, I'll merge and we'll see what happens, :see_no_evil:

PR #4752, opened 2019-12-02
Upgrade ImGui to version 1.74

Web-eWorks:
Among other features, this brings with it the Tab Container widget, several internal ImGui improvements, and keeps us less out-of-date when the bigger features like Tables and Docking become stable in the ImGui repository.

CC @vakhoir; I've tested pretty much all the widgets we have in game, but I'd appreciate it if you could try to break things some more.


vakhoir:
Will do, but I'm a bit swamped at the moment. I should be able to do it over the weekend.

Web-eWorks:
@fluffyfreak will you have time to go over this PR this week? If not, I'm good to just merge as-is; it was a very painless upgrade and I don't think there are any regressions on the lua side.

impaktor:
Maybe just verify it works on Windows? Or just merge, assuming a breaking change on windows can be fixed swiftly.

PR #4753, opened 2019-12-05
Restore a lost feature

mike-f1:
Game can again show no icons and no label,
show only icons or show both.

### DIFF
--- a/data/pigui/views/game.lua
+++ b/data/pigui/views/game.lua
@@ -144,10 +144,12 @@ local function displayOnScreenObjects()
 		local mainBody = group[2].body
 		local mainCoords = group[1].screenCoordinates
 		local count = #group - 1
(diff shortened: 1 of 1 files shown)


Web-eWorks:
@mike-f1 as a suggestion, have you tried using the `if` statement to conditionally call the `ui.addStyledText` that displays the label? Seems to me like it'd be much better to not call that at all instead of calling it with an empty string to display.

mike-f1:
Done.

Please keep commit history, thank's

Web-eWorks:
Looks good to me, thanks @mike-f1! I'll give this a thorough testing tomorrow and hopefully it'll be good to merge.

Web-eWorks:
Sorry about the testing delay, I had an unexpected hardware failure. Seems very straightforward and I uncovered no bugs.

PR #4754, opened 2019-12-11
Fix append of CameraFrame on saves

mike-f1:
Restore symmetry of [`Begin`/`End`]`CameraFrame` during save/load routine and avoid potential memory leaks

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -205,8 +205,8 @@ void Game::ToJson(Json &jsonObj)
 	jsonObj["hyperspace_end_time"] = m_hyperspaceEndTime;
 
 	// Delete camera frame from frame structure:
-	if (m_gameViews->m_worldView->GetCameraContext()->GetCamFrame())
-		m_gameViews->m_worldView->EndCameraFrame();
+	bool have_cam_frame = m_gameViews->m_worldView->GetCameraContext()->GetCamFrame().valid();
+	if (have_cam_frame) m_gameViews->m_worldView->EndCameraFrame();
 
 	// space, all the bodies and things
 	m_space->ToJson(jsonObj);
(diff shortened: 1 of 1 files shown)


src/Game.cpp
```diff
@@ -205,8 +205,8 @@ void Game::ToJson(Json &jsonObj)
 	jsonObj["hyperspace_end_time"] = m_hyperspaceEndTime;
 
 	// Delete camera frame from frame structure:
-	if (m_gameViews->m_worldView->GetCameraContext()->GetCamFrame())
-		m_gameViews->m_worldView->EndCameraFrame();
+	bool have_cam_frame = m_gameViews->m_worldView->GetCameraContext()->GetCamFrame().valid();
```
Web-eWorks: By the by, the FrameId struct has an implicit conversion to `bool`. So it's not required to put the `.valid()` on there, unless you want it for code clarity :D.

mike-f1: Yes, you're right :P .., On "the Book" too" :D

Web-eWorks:
@mike-f1 there's only one issue left with the PR:
```diff
--- a/src/Game.cpp	2019-12-12 08:22:01.741117762 +0000
+++ b/src/Game.cpp	2019-12-12 08:22:01.873389857 +0000
@@ -273,7 +273,7 @@
 	Pi::luaSerializer->UninitTableRefs();

 	// Bring back camera frame:
-	if (have_cam_frame)	m_gameViews->m_worldView->BeginCameraFrame();
+	if (have_cam_frame) m_gameViews->m_worldView->BeginCameraFrame();
 }
```

clang-format is complaining about there being a rogue TAB character in that single-line if-statement; if you can append that change into the most recent commit and push again, I'll merge it immediately.

mike-f1:
Done... Sorry for the noise ;P I didn't realize what's happened there...

PR #4757, opened 2019-12-15
Fix calculation of latitude and longitude for PlanetaryInfo

cwyss:
Simple fix to the conversion of lat/lon from floating point values to degree/minutes/seconds format, which is done during output formating for the PlanetaryInfo window (bottom right in flight UI).

The old code only yielded minutes in the range of 0 to 29 where the missing range from 30 to 59 was somehow also mapped to 0 to 29, but "flipped". (Will add some examples)
AFAIK this bug existed already for some time (>=1 year) but was apparently not recognised yet.

### DIFF
--- a/data/pigui/pigui.lua
+++ b/data/pigui/pigui.lua
@@ -231,26 +231,24 @@ end
 
 ui.Format = {
 	Latitude = function(decimal_degrees)
-		local deg = math.floor(decimal_degrees + 0.5)
-		local dec = math.abs(decimal_degrees - deg)
 		local prefix = lc.LATITUDE_NORTH_ABBREV
-		if deg < 0 then
+		if decimal_degrees < 0 then
 			prefix = lc.LATITUDE_SOUTH_ABBREV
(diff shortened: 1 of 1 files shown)


cwyss:
Here is an example of what the old code did/does. When crossing the equator from south to north you get this for the latitude:
```
float   old     correct
-1.00  S 1°0'   S 1°0'
-0.75  S 1°15'  S 0°45'
-0.50  N 0°30'  S 0°30'
-0.25  N 0°15'  S 0°15'
 0.00  N 0°0'   N 0°0'
 0.25  N 0°15'  N 0°15'
 0.50  N 1°30'  N 0°30'
 0.75  N 1°15'  N 0°45'
 1.00  N 1°0'   N 1°00'
 1.25  N 1°15'  N 1°15'
 1.50  N 2°30'  N 1°30'
```
So with the old code you see the minutes of the latitude increase upto 30, then _decrease_ again down to zero, then up again to 30 and so forth.

PR #4758, opened 2019-12-22
Make continue button load _quicksave if autosave not active

cwyss:
Simple additional feature for the continue button in main menu:

- Load the _quicksave file if the autosave option is _not_ activated (and _quicksave exists).
- If autosave _is_ active, then continue works as before and loads the autosave _exit file.

This added functionality could be handy whenever you are not using autosave.
In fact I did not use it for quite some time :)

### DIFF
--- a/data/pigui/views/mainmenu.lua
+++ b/data/pigui/views/mainmenu.lua
@@ -134,7 +134,19 @@ local function quitGame()
 end
 
 local function continueGame()
-	Game.LoadGame("_exit")
+	if Game.CanLoadGame('_exit') and ( Engine.GetAutosaveEnabled() or not Game.CanLoadGame('_quicksave') ) then
+		Game.LoadGame("_exit")
+	else
+		Game.LoadGame('_quicksave')
+	end
+end
+
+local function canContinue()
+	if Game.CanLoadGame('_exit') or Game.CanLoadGame('_quicksave') then
+		return true
+	else
+		return false
+	end
 end
 
 local function startAtLocation(location)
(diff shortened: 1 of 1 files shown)


cwyss:
In #4760 it was pointed out that the code from `showMainMenu()` in `data/pigui/views/mainmenu.lua`, which checks whether or not to make the Continue button active, polls the filesystem at every frame. This is -as I understand it- the case with all versions of `showMainMenu()`, both in upstream, in this PR, and in #4760. E.g., upstream checks for the existence of `savefiles/_exit`, while this PR checks for `_exit` and (if not found) `_quicksave`. 

Some questions/thoughts about this:
- Is `showMainMenu()` actually called if the main menu is _not_ shown, i.e. when flying? If not then the FPS impact would happen only when showing the main menu (and could thus be acceptable?)
- A simple alternative would be to have the Continue button always active and only upon clicking check for existence of whatever files, possibly giving the user a "Sorry. No game to continue from".

Opinions?

impaktor:
> ... polls the filesystem at every frame. 

I think @johnbartholomew did some work on the save/load menu in the past, perhaps he's interested in voicing concerns about this.

data/pigui/views/mainmenu.lua
```diff
@@ -134,7 +134,19 @@ local function quitGame()
...
-	Game.LoadGame("_exit")
+	if Game.CanLoadGame('_exit') and ( Engine.GetAutosaveEnabled() or not Game.CanLoadGame('_quicksave') ) then
+		Game.LoadGame("_exit")
+	else
+		Game.LoadGame('_quicksave')
+	end
+end
+
+local function canContinue()
+	if Game.CanLoadGame('_exit') or Game.CanLoadGame('_quicksave') then
```
Web-eWorks: While I'd prefer you use idomatic lua (see below), this style doesn't have any major problems and can be easily refactored later. Please do keep in mind for the future that there are much more elegant and simple ways to express a construct like this, though!
```lua
return Game.CanLoadGame('_exit') or Game.CanLoadGame('_quicksave')
```

cwyss: Thanks for the hint! Reminds me that I still need to learn lua properly :)

PR #4760, opened 2019-12-24
Add a 'FindMostRecentSaveGame' function in Game and use it in MainMenu

mike-f1:
iIn order to generalize the "Continue" button functionality
in `MainMenu`, pick the name of the most recent filename
and return that name to Lua. If it's not possible, return
a boolean `false` to disable that button.

Tested:
1) saving with Ctrl-F9 and reloading > it works
2) renaming save folder > it works (button disabled)

@cwyss , don't get me wrong: I hope you agree this way
is better. And if so, this will supersedes #4758

Oh, I forget it: merry Christmas  to everyone!

### DIFF
--- a/data/pigui/views/mainmenu.lua
+++ b/data/pigui/views/mainmenu.lua
@@ -36,6 +36,8 @@ local showQuitConfirm = false
 local quitConfirmMsg
 local max_flavours = 22
 
+local canContinue = false
+
 local startLocations = {
 	{['name']=lui.START_AT_MARS,
 	 ['desc']=lui.START_AT_MARS_DESC,
@@ -134,7 +136,7 @@ local function quitGame()
 end
 
 local function continueGame()
-	Game.LoadGame("_exit")
+	Game.LoadGame(canContinue)
 end
 
 local function startAtLocation(location)
@@ -157,7 +159,7 @@ local function callModules(mode)
 end
 
 local function showMainMenu()
-	local canContinue = Game.CanLoadGame('_exit')
+	canContinue = Game.FindMostRecentSaveGame()
 	local buttons = 4
 
 	local winPos = Vector2(ui.screenWidth - mainButtonSize.x - 100, ui.screenHeight/2 - (buttons * mainButtonSize.y)/2 - (2*mainButtonSize.y)/2 - 8)
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -908,6 +908,31 @@ bool Game::CanLoadGame(const std::string &filename)
 	// file data is freed here
 }
 
+//static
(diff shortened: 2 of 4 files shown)


cwyss:
Great! I totally support this. As it covers the functionality of my PR #4758 and extends it to regular saved game files.
#4758 is then obsolete.

cwyss:
@mike-f1 
Just tested your continue_game branch and it seems that, while pioneer is running, your code does not always check what the newest file in savefiles is. It seems to pick the one that was most recent when starting pioneer.
What I did:

1. Have a most recent game file in savefiles dir which is not _quicksave, e.g., "game1"
2. Hit ctrl-f9 to produce a new _quicksave file, which is now the most recent file in savefiles
3. End game to return to the main menu. Do no quit pioneer.
4. Continue button now still loads "game1", not _quicksave
5. Quit pioneer and restart: now Continue loads _quicksave

Can you confirm?

mike-f1:
@cwyss , now it should work: I forgot `canContinue` (as I wrote it) behave like a `static` thus it updates only once.

Thank's for your time!

cwyss:
Ok, seems to work good now **if** you start pioneer with at least one file already present in savefiles.

**But** if savefiles dir is empty at program startup, you start a game, do quick save, go back to main menu, then the continue button remains inactive...

mike-f1:
Done: now is polling :P so it reads that directory each frame.

I also changed funtion names so their complexity is better reflected.

NOTE: There's a similar function in [`savegame.lua`](https://github.com/pioneerspacesim/pioneer/blob/0881148a85e247a69bca9477bf6995561fccfd2e/data/pigui/modules/saveloadgame.lua#L75) which can be made 1 but I feel it's OT for this PR.

laarmen:
I'm reading this on my phone so I might be missing something, but are you
saying that with this PR we'd hit de filesystem on every frame?

mike-f1:
Yes... Is it not allowed?

laarmen:
It's not so much not allowed as not advised. For instance, consider that
the directory that you are polling might be a network drive. FPS would go
down the toilet.

There's a number of other reasons why this method would cause serious
performance issues, this is just the first and most obvious one that came
to mind.

mike-f1:
OMG... You're right :P So the code I spot above ([Here again the link](https://github.com/pioneerspacesim/pioneer/blob/0881148a85e247a69bca9477bf6995561fccfd2e/data/pigui/modules/saveloadgame.lua#L75-98) ), which is already in master, should be corrected A.S.A.P!

I'm just sorry I can't undertake a so critical task, but sure there's a better way to do such a thing and you will enlighten us with your code :)

Thus, hurry up and have good coding! :+1:

PR #4761, opened 2019-12-26
Restore feature: maximum display distance for ship indicators

cwyss:
Make ship indicators be displayed in game UI _only_ if closer than a certain distance, currently 1000 km.

This feature seems to have been gone since d90e14ec1 (5 May 2019, #4548). I went through the comments, but couldn't find a hint that it was dropped on purpose. Moreover, in `data/pigui/views/game.lua` the constant `IN_SPACE_INDICATOR_SHIP_MAX_DISTANCE` from the old code is still present, but not used anymore.
Hence my conclusion that the feature was lost by accident.

In my restore, I don't use a hard coded value for the maximum distance on the c++ side, but instead pass the value of `IN_SPACE_INDICATOR_SHIP_MAX_DISTANCE` (which is 1000km) in `game.lua` to `getProjectedBodiesGrouped()` as an additional parameter. Code-wise, this is similar to the current radar size being passed to `getTargetsNearby()` in `radar.lua`.

I hope that I did not undo something that was deliberately deleted :-)
Any comments or suggestions welcome!

### DIFF
--- a/data/pigui/views/game.lua
+++ b/data/pigui/views/game.lua
@@ -138,7 +138,7 @@ local function displayOnScreenObjects()
 	local iconsize = Vector2(18 , 18)
 	local label_offset = 14 -- enough so that the target rectangle fits
 	local collapse = iconsize
-	local bodies_grouped = ui.getProjectedBodiesGrouped(collapse)
+	local bodies_grouped = ui.getProjectedBodiesGrouped(collapse, IN_SPACE_INDICATOR_SHIP_MAX_DISTANCE)
 
 	for _,group in ipairs(bodies_grouped) do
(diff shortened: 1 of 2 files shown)


Web-eWorks:
Looks good to me!

For future reference, I'd appreciate it if you could document Lua-facing API changes like extra parameters using the Natural Docs syntax, but as this function doesn't have existing docs I won't require that. I know the documentation is in a bit of a sorry state now, but whatever you can do to improve it would be accepted with great joy!

impaktor:
> I'd appreciate it if you could document Lua-facing API changes like extra parameters using the Natural Docs syntax,

@cwyss that's an invitation for another PR, :)

Issue #4755, opened 2019-12-14
Save game dialog does not appear


cwyss:
### Observed behaviour

The save game dialog does not open when pressing the save button, i.e. flight UI -> Esc -> "Save" button.
After pressing the save button once, also Load Game from the main menu stops working.
quicksave via Ctrl-F9 works fine.

__My pioneer version (and OS):__
d0f26555 (current master)

last bug free:
d016f7a5 (Nov 16 2019)

impaktor:
Yep, it's a known bug, I hope it will be short lived, as it's pretty serious for user experience. 
Thanks for reporting!

cwyss:
Thanks for the info.

I also noticed just now that the dialog box to change a key binding opens _behind_ the options dialog box. But this is probably also known already... :)

impaktor:
....or if you @cwyss have time to look at this, that would be nice. Seems like a pretty major serious bug.

vakhoir:
> @vakhoir do you have time to look into this?

I finally do!
Looking into it now

impaktor:
Great! Do note, there's a PR for it, #4755, but author seems to indicate it's not working as expected. 

I think @Web-eWorks said he knew what was amiss with this bug:

```
<sturnclaw> [08:51:30] yeah, the save issue is just the code in pigui.lua
            being designed for only one modal
<sturnclaw> [08:52:02] I can rip that out and rewrite it in an afternoon, but
            I haven't been able to *think* much less work on Pioneer for the
            past week or so
```

Web-eWorks:
Hmm, on second thought we don't seem to be wrapping the `ImGui::BeginPopupModal()` in any custom code, so it may just be an incorrect use of `ui.openPopup()` / `ui.closeCurrentPopup()`; I'd recommend investigating the imgui docs and looking at the code in `imgui.cpp` and related files...

vakhoir:
Yeah, it's fixable in lua, but I have to say I'm not sure what the ImGui people were going for when they decided the separete openPopup from beginPopup, and then having openPopup be called only once on open, when everything else in the UI has to be called once per frame...

Anyways I have something worked out, I'll be testing and polishing it tomorrow.

vakhoir:
Oh, should've hit refresh

> Hmm, on second thought we don't seem to be wrapping the ImGui::BeginPopupModal() in any custom code, so it may just be an incorrect use of ui.openPopup() / ui.closeCurrentPopup();

[...]

PR #4765, opened 2020-01-06
fix issue  #4755 (save game dialog)

Gliese852:

### DIFF
--- a/data/pigui/modules/settings-window.lua
+++ b/data/pigui/modules/settings-window.lua
@@ -559,7 +559,10 @@ local function optionsWindow()
 
 				if Game.player then
 					ui.sameLine()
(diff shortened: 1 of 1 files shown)


Gliese852:
everything worked for several days, but just found that if you press ESC on the keyboard with the save dialog open, it disappears and does not appear anymore, until restart.

impaktor:
@Gliese852 We appreciate your effort, and applaud you for the submission, but this time, we might go with the other PR that will fix the same issue, linked above, in a more complete way.

We hope this will not discourage you from any future contributions.

Gliese852:
I am glad to see that the problem is solved by people who know code better.

PR #4767, opened 2020-01-10
ModalWindow class that allows nesting

### LINKED ISSUE #4755: Save game dialog does not appear

vakhoir:
Fixes #4755 closes #4765 

I created a class that keeps track of and draws modal windows, so each window can be open from any script without having to worry about what was opened before. It should make writing modals a bit more intuitive.

### DIFF
--- a/data/pigui/libs/modal-win.lua
+++ b/data/pigui/libs/modal-win.lua
@@ -1,0 +1,77 @@
+-- Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local ui = import 'pigui/pigui.lua'
+
+local modals = {}
+local modalStack = {}
+
+local ModalWindow = {}
+local defaultModalFlags = ui.WindowFlags {"NoTitleBar", "NoResize", "AlwaysAutoResize", "NoMove"}
+
+function ModalWindow.New(name, innerHandler, outerHandler, flags)
+	local modalWin = {
+		name = name,
+		flags = flags or defaultModalFlags,
+		stackIdx = -1,
+		isOpen = false,
+		innerHandler = innerHandler,
+		outerHandler = outerHandler or function(self, drawPopupFn)
+			drawPopupFn()
+		end,
+	}
+
+	setmetatable(modalWin, {
+		__index = ModalWindow,
+		class = "UI.ModalWindow",
+	})
+
+	modals[name] = modalWin
+
+	return modalWin
+end
+
+function ModalWindow:open()
+	if self.stackIdx < 0 then
+		table.insert(modalStack, self)
+		self.stackIdx = #modalStack
+	end
+end
+
+function ModalWindow:close()
+	for i=#modalStack, self.stackIdx, -1 do
+		modalStack[i].stackIdx = -1
+		modalStack[i].isOpen = false
+		ui.closeCurrentPopup()
+		table.remove(modalStack, i)
+	end
+end
+
+local function drawModals(idx)
+	if idx <= #modalStack then
+		local win = modalStack[idx]
+
+		if (not win.isOpen) then
+			win.isOpen = true
+			ui.openPopup(win.name)
+		end
+
+		win:outerHandler(function ()
+			if ui.beginPopupModal(win.name, win.flags) then
+				win:innerHandler()
+				drawModals(idx+1)
+				ui.endPopup()
+			end
+		end)
+	end
+end
+
+ui.registerModule('game', function()
+	drawModals(1)
+end)
+
+ui.registerModule('mainMenu', function()
+	drawModals(1)
+end)
+
+return ModalWindow
--- a/data/pigui/modules/saveloadgame.lua
+++ b/data/pigui/modules/saveloadgame.lua
@@ -10,6 +10,7 @@ local lc = Lang.GetResource("core")
 local lui = Lang.GetResource("ui-core")
 local FileSystem = import("FileSystem")
 local Format = import("Format")
+local ModalWindow = import 'pigui/libs/modal-win.lua'
 
 local utils = import("utils")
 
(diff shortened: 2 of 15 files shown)


vakhoir:
In testing the ShipMarket view I noticed the Market widget I wrote wasn't blocked by modal popups, this is now fixed. Also added a check for not double-opening popups, which screws up to popup stack on the c++ side, and leads to all sorts of issues.

data/pigui/libs/modal-win.lua
```diff
@@ -0,0 +1,81 @@
...
+	for i=#modalStack, self.stackIdx, -1 do
+		modalStack[i].stackIdx = -1
+		modalStack[i].isOpen = false
+		ui.closeCurrentPopup()
+		table.remove(modalStack, i)
+	end
+end
+
+local drawModals
+drawModals = function(idx)
```
Web-eWorks: @vakhoir per the lua reference manual, `local function drawModals(idx)` is syntactic sugar for `local drawModals; drawModals = function(idx)`, so this construct is redundant.
Just FYI!

data/pigui/libs/modal-win.lua
```diff
@@ -0,0 +1,81 @@
...
+
+local modals = {}
+local modalStack = {}
+
+local ModalWindow = {}
+
+function ModalWindow.New(name, innerHandler, outerHandler, flags)
+	local self = {
+		name = name,
+		flags = flags or {"NoTitleBar", "NoResize", "AlwaysAutoResize", "NoMove"},
```
Web-eWorks: I'd recommend you use the new WindowFlags helper class to create this set of flags once at file scope and simply set `flags = flags or defaultFlags`. This avoids the extra memory cost of the flags table, and prevents the table from having to be parsed once a frame!

data/pigui/libs/modal-win.lua
```diff
@@ -0,0 +1,81 @@
...
+		stackIdx = -1,
+		isOpen = false,
+		innerHandler = innerHandler,
+		outerHandler = outerHandler,
```
Web-eWorks: I'd also recommend that you set `outerHandler = outerHandler or function(drawFn) drawFn() end` instead of deferring the check until the end of the table creation. You save about five lines that way :D

data/pigui/modules/saveloadgame.lua
```diff
@@ -98,50 +99,40 @@ local function showSaveFiles()
...
 
-local function savedGameWindow()
-	if ui.showSavedGameWindow then
-		if not popupOpened then
-			popupOpened = true
-			ui.openPopup("LoadGame")
-		end
+ui.saveLoadWindow = ModalWindow.New("LoadGame", function()
+	local mode
+	mode = ui.saveLoadWindow.mode == 'SAVE' and lui.SAVE or lui.LOAD
```
Web-eWorks: Is there a reason why you're declaring `mode` and then assigning a line later? This should be all one line.

data/pigui/modules/saveloadgame.lua
```diff
@@ -98,50 +99,40 @@ local function showSaveFiles()
...
+	optionTextButton(mode, nil, selectedSave~=nil, closeAndLoadOrSave)
```
Web-eWorks: Minor nitpick, but please space around your operators like this. Also if selectedSave is guaranteed to be a "truthy" value when present, use `not selectedSave` instead; it's a little shorter and much easier for the reader to parse.
```suggestion
	optionTextButton(mode, nil, selectedSave ~= nil, closeAndLoadOrSave)
```

data/pigui/modules/settings-window.lua
```diff
@@ -509,76 +507,62 @@ local optionsTabs = {
 	["controls"]=showControlsOptions
 }
 
-local optionModalOpen = false
-
-local function optionsWindow()
-	if ui.showOptionsWindow then
-		if not optionModalOpen then
-			ui.openPopup("Options")
```
Web-eWorks: Thank you for cleaning up this style of code in particular! I'd much rather we call one function to open rather than have to check it every frame.

data/pigui/modules/station-view/01-lobby.lua
```diff
@@ -49,7 +50,15 @@ local hydrogenIcon = PiImage.New("icons/goods/Hydrogen.png")
 local hyperdriveIcon = PiImage.New("icons/goods/Hydrogen.png")
 
 local popupMsg = ''
-local popupId = 'lobbyPopup'
+local popup
+popup = ModalWindow.New('lobbyPopup', function()
```
Web-eWorks: Same issue here, please prefer the `local popup = ModalWindow.New` style over split lines.

data/pigui/libs/modal-win.lua
vakhoir: Not exactly - it's not possible to do a recursive call if the function isn't already declared. The game crashes if you try to do it in a single line.

Web-eWorks: That's weird, because the 5.2 manual explicitly states that here: https://www.lua.org/manual/5.2/manual.html#3.4.10

If it's crashing like that, there may be a deeper problem...

Web-eWorks: Ah. I think I understand. Please prefer the use of `local function name()` rather than `local name = function()` - it's much easier to read and it gives more reliable results for Lua's debugger.

data/pigui/modules/station-view/01-lobby.lua
vakhoir: `local function funcName()` worked for allowing a recursive call, but I'm not sure there's a way around it here. `popup` needs to be at least declared to be referenced inside the render function.

data/pigui/libs/modal-win.lua
vakhoir: I don't know what I was thinking :D

data/pigui/modules/saveloadgame.lua
vakhoir: I think that's still from the pre-modal save window days, I just didn't touch it when I made the window modal. Fixed now.

data/pigui/modules/saveloadgame.lua
vakhoir: I fixed the spacing, but now I'm wondering if the check shouldn't be changed to `selectedSave ~= nil and selectedSave ~= ''` like what is used in `closeAndLoadOrSave`

data/pigui/modules/settings-window.lua
vakhoir: Yeah, that was the part that was driving me up the wall with popups. It was very awkard to have to check both if the window should be rendered and if it was opened already.
It's a bit silly we have to keep track of the popup stack when ImGui is doing it too, but the code should be more manageable.

data/pigui/modules/saveloadgame.lua
vakhoir: I went ahead and added the `selectedSave ~= ''` check.

vakhoir:
I applied all the changes. They're in a separate commit for ease of review, I'll squash them once there's a green light to merge.

data/pigui/libs/modal-win.lua
```diff
@@ -7,22 +7,19 @@ local modals = {}
...
+		flags = flags or defaultModalFlags,
 		stackIdx = -1,
 		isOpen = false,
 		innerHandler = innerHandler,
-		outerHandler = outerHandler,
-	}
-
-	if not outerHandler then
-		self.outerHandler = function(drawPopupFn)
+		outerHandler = outerHandler or function(slf, drawPopupFn)
```
Web-eWorks: This is the world's tiniest nitpick, but I'd recommend using `self` for consistency.
```suggestion
		outerHandler = outerHandler or function(self, drawPopupFn)
```

vakhoir: Fixed and squashed

vakhoir:
Fixed and squashed

data/pigui/libs/modal-win.lua
vakhoir: I also renamed the new instance variable to modalWin, it was the whole reason for naming the function argument "slf" - didn't want to confuse people.

impaktor:
I assume this can be merged now? I'll merge in 24 hours, if this is still open, and no objections.

I think I'll need this for the Police-tab stuff I've been working on.

Web-eWorks:
Sorry about the delay... I've been up to my neck in sandboxing and forgot to hit the shiny merge button!

Issue #4719, opened 2019-10-17
Overview window stays atop Mission screen, obscuring buttons


craigo-:
### Observed behaviour
In flight, when the Overview window is open and you open the Missions screen, the Overview window stays atop the Missions screen and you cannot select the `More info...` button.

### Expected behaviour
In this instance, I believe the Missions screen should be on top, obscuring the Overview window. Don't close the Overview window in this instance, just don't show it at all and allow the Missions screen to receive the clicks.

### Steps to reproduce
1. Accept a local mission
2. Launch
3. Open the Overview window
4. Select `Show personal info`, then `Missions`
5. Try and click on the `More info...` button
![image](https://user-images.githubusercontent.com/4467963/66972098-f7fe5100-f0ef-11e9-9f90-24e8025a01e8.png)

I often use this workflow, because (a) the Overview window distance information is very useful; (b) I use the Mission window's `More info...` button to select a navigation target.

__My pioneer version (and OS):__
Pioneer: master (3dff8ca), approx. 20191017, Release x64
Windows 10 Enterprise 1809 x64

[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/3737093/opengl.txt)

Web-eWorks:
All of the full-screen windows (station, ship, etc) are slated for a rework. However, I agree that the Overview should be disabled in a fullscreen context.

PR #4768, opened 2020-01-10
Fix issue #4719 (overview obscuring buttons)

Gliese852:
Just copied the behavior of 'Commslog' window.
Actually added only two lines, I just had to indent a lot.


Gliese852:
Sorry, some kind of confusion with commits, I'd rather create a new PR

PR #4770, opened 2020-01-12
Fix mission screen obscuring buttons

### LINKED ISSUE #4719: Overview window stays atop Mission screen, obscuring buttons

Gliese852:
Fixes #4719

Another try, this time commits are fine.
Now if the player is docked or we are not looking into space, the overview window function does nothing at all. In addition, I thought that button "Show Comms" should be removed when it does nothing, that is, when a player is not docked or landed.

### DIFF
--- a/data/pigui/modules/fx-window.lua
+++ b/data/pigui/modules/fx-window.lua
@@ -116,20 +116,16 @@ local function button_info(current_view)
 end
 
 local function button_comms(current_view)
+	if player:IsDocked() then
 	ui.sameLine()
 	if mainMenuButton(icons.comms, current_view == "space_station", lui.HUD_BUTTON_SHOW_COMMS) or (ui.noModifierHeld() and ui.isKeyReleased(ui.keys.f4)) then
-		if player:IsDocked() then
 			if current_view == "space_station" then
 				Game.SetView("world")
 			else
 				Game.SetView("space_station")
 			end
-		else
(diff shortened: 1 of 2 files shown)


Web-eWorks:
@Gliese852 there appears to be one more conflict. If I could ask, could you rebase your branch on current master and separate it into two commits - one that corrects the indentation, and one that applies the new code? It's much easier to review the code if I'm only checking a few lines instead of almost 100...

Gliese852:
@Web-eWorks Please see if I did the right thing.

impaktor:
Looks readable to me, thanks!

PR #4772, opened 2020-01-16
CMake profiler build option: support older versions of cmake

cwyss:
Let's you build pioneer with profiler also under cmake versions less than 3.12.

Thanks to @laarmen for pointing out that the command used so far,

`add_compile_definitions(PIONEER_PROFILER=1)`

actually is a cmake macro, 

> but only available in 3.12 and later, which was released in July 2018.

As it happens, I am still running Debian stretch (old stable), which has cmake 3.7.2. The new stable Debian "buster", released July 2019, has cmake 3.13.4.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,7 +173,7 @@ endif (USE_SYSTEM_LIBLUA)
 
 option(PROFILER_ENABLED "Build pioneer with profiling support built-in." OFF)
 if (PROFILER_ENABLED)
-	add_compile_definitions(PIONEER_PROFILER=1)
+	add_definitions(-DPIONEER_PROFILER=1)
 endif(PROFILER_ENABLED)
 
 if (MSVC)


Web-eWorks:
Yup, this one's my bad. I reference off of the latest CMake docs due to the higher quality of later versions, and either I didn't see the "available in" or it simply wasn't on the page.

impaktor:
@cwyss I suspect you're not in [AUTHORS.txt](https://github.com/pioneerspacesim/pioneer/blob/master/AUTHORS.txt). Let us know if/what to put (we seem to have forgotten that in the last year :P)

Thanks for good work!

laarmen:
@Web-eWorks don't beat yourself up, the CMake doc seldom mentions when a
particular feature was introduced. I only happened to know this particular
one
because I hit the exact same situation on another project at work this week.

It might be worth it to add
https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html at
the top with the version currently used by @cwyss. It would document the
state of things.

cwyss:
> @cwyss I suspect you're not in [AUTHORS.txt](https://github.com/pioneerspacesim/pioneer/blob/master/AUTHORS.txt). Let us know if/what to put (we seem to have forgotten that in the last year :P)

@impaktor Yes, you can add me as `Christian Wyss`. That would be nice :smile:

PR #4746, opened 2019-11-23
New mission: Find person

robothauler:
The player has to find a person or a person with a ship and hand over a message. Which means hopping from station to station. Then return, or transport the person back. The ideas are taken from https://forum.pioneerspacesim.net/viewtopic.php?f=3&t=487. I tried to add some surprises like pirates firing missiles or mercenaries who try to intercept if close to the planet.

Unfortunately, merging this will cause two issues.

1. ~~The missile launch is not detected. No warning.~~ ~~No police action.~~
2. Close to a station, the mercenary can fire what he can. But the player is immediately attacked by the police.

While 2. can be solved later. ~~It would be great if someone can take care of 1. e.g. play a warning sound and trigger an event so that other modules can take action.~~

PR #4773, opened 2020-01-18
Add number of total docking pads at station to lobby

impaktor:
I've been playing with the lobby (motivated by comments in other [PR](https://github.com/pioneerspacesim/pioneer/pull/4746#issuecomment-575055727)), trying to see how much of that code I can muster. So in experimenting, I added one more info sentence. 

(I imagine these Lobby-sentences, sort of how the pilot annonuces the weather when you land at a new airport)

I'm doing this as WIP, because maybe wording can be better (suggestions?), or I find other stuff to add to it.

Possibly we'd also like to add information about how many ships are currently
docked, as this only shows the total capacity.

2020-01-18-102555_1600x900_scrot
2020-01-18-102616_1600x900_scrot

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -1651,6 +1651,10 @@
     "description": "",
     "message": "Station manager"
   },
+  "STATION_DOCKS": {
+    "description": "Information shown in the station lobby, for total number of docking pads",
+    "message": "Our total capacity is {total_docking_pads} docking pads."
+  },
   "STATION_ORBIT": {
     "description": "Information shown in the station loby",
     "message": "We are in a {orbit_period} day period orbit around {parent_body}."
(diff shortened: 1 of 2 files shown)


WKFO:
Didn't look at the code, but if possible, we could also separate medium and small docking pads, like "...total of 4 medium and 12 small docking pads." or something.

impaktor:
Yeah, that would be better, but that information isn't as easily accessible. @fluffyfreak would know for sure where to get that info. Not sure it's even exposed on Lua side, but would have to do some digging.

(I'm also thinking police station would say how many police ships there are, and how many are docked).

Web-eWorks:
I'm not 100% sure the game even makes the distinction between pad sizes. I might be wrong, but AFAIK the landing and station code doesn't really care, it just knows "pad" and "not pad".

fluffyfreak:
Internally the pad sizes matter, you can't be allocated a pad smaller than your ship for example. I don't know if that's exposed anywhere though.

PR #4774, opened 2020-01-18
Ship market

vakhoir:
Added the Ship Market view.
Also contains a minor fix for the model spinner, and a fix for the alignment of the Request Launch button.
Based off the ModalWindow fix branch (#4767)

ShipMarket

### DIFF
--- a/data/pigui/libs/market.lua
+++ b/data/pigui/libs/market.lua
@@ -5,6 +5,7 @@ local Game = import 'Game'
 local Lang = import 'Lang'
 
 local ui = import 'pigui/pigui.lua'
+local ModalWindow = import 'pigui/libs/modal-win.lua'
 
 local l = Lang.GetResource("ui-core")
 
@@ -46,8 +47,8 @@ local defaultFuncs = {
         if not self.funcs.onClickBuy(self, e) then return end
 
         if self.funcs.getStock(self, e) <= 0 then
-            self.popupMsg = l.ITEM_IS_OUT_OF_STOCK
-            ui.openPopup(self.popupMsgId)
+            self.popup.msg = l.ITEM_IS_OUT_OF_STOCK
+            self.popup:open()
             return
         end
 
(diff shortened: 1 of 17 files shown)


impaktor:
Most excellent! Slowly we're moving closer to be exclusively pigui in our UI.

I've tested it, and seems to work as advertised. I'll be running my pioneer from this branch, for other work, to see if all is well.

(For future: we might want to incorporate a short seller's pitch on each ship (~ a paragraph in length), like it's done for commodity market, but not sure how a string in ship.json would be put into the translation system).

vakhoir:
>  but not sure how a string in ship.json would be put into the translation system

Just have the json contain the translation key to rather than the string itself?

Web-eWorks:
> (For future: we might want to incorporate a short seller's pitch on each ship (~ a paragraph in length), like it's done for commodity market, but not sure how a string in ship.json would be put into the translation system).

Likely we'd make a new translation resource where the raw names of the ships (sinonatrix) is the key, and the message is the sales pitch. Could likely even have two or three variants for each ship, say one for a 'shady' second-hand ship dealer, one for the megacorp dealer, and maybe one for a rimworld chop shop ship dealer.

impaktor:
continuing on this tangent, our current thinking is it should be possible to add ships as modules, so one would have to provide a en.json file as well for each ship (and/or fallback string for missing sales-pitch).

Web-eWorks:
I'm not 100% sure it doesn't already function like this, but it's very easy to allow mods to provide .patch.json extensions to a language resource, whether that's adding additional key-value pairs, changing an existing value, or removing one or more entirely.

So the 'base mod' would provide the translation resource 'ship-sales/en.json' with strings for all the vanilla ships, and mods would include a file with the extra messages that gets merged with the original json file by the loading process.

Web-eWorks:
Everything looks good to me, sorry it's taken so long to merge!

PR #4776, opened 2020-01-26
Rewind PiGui stack in case of an error

Web-eWorks:
I've added a method to "rewind" the ImGui stack in case of an error being thrown, ensuring we don't accidentally leave unterminated Begin() calls in unprotected code that threw an error.

Next on the docket is being able to selectively purge the import cache, deprecating `import()` in favor of `require()` to more closely align with standard lua practices (most importantly, some Lua dialects use `import` as a keyword, which naturally causes problems with our implementation), and figuring out if we actually want to watch the filesystem for changes.

CC @vakhoir; I'd appreciate your insight into what your use case for reloading would look like - I'm thinking a `package.reimport('pigui.views.XXX')` API might work, though if you'd like something different now's the time to get the request in!


Web-eWorks:
I messed with the build system a bit and now `buildopts.h` isn't getting found properly; pinging @pcercuei because I think you were the last person to touch that file - can you advise what's different between the `pioneer-lib` and `pioneer-lua` targets that one is working properly and the other has no clue how to find the file?

src/lua/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+	${CMAKE_CURRENT_SOURCE_DIR}
+	core
+)
+
+# Creates variables LUA_CXX_FILES and LUA_HXX_FILES
+add_source_folders(LUA LUA_SRC_FOLDERS)
+
+# Creates a library, adds it to the build, and sets C++ target properties on it
+define_pioneer_library(pioneer-lua LUA_CXX_FILES LUA_HXX_FILES)
+target_include_directories(pioneer-lua PRIVATE ${CMAKE_BINARY_DIR}/src)
```
pcercuei: You don't include ```${CMAKE_BINARY_DIR}```.

vakhoir:
> CC @vakhoir; I'd appreciate your insight into what your use case for reloading would look like - I'm thinking a `package.reimport('pigui.views.XXX')` API might work, though if you'd like something different now's the time to get the request in!

The typical workflow is that I'm testing a newly implemented view, decide to adjust someting, switch over to the IDE to do so, and when I'm done I normally have to restart the game. If I could call something `package.reimport('pigui.views.XXX')` from the game console, it would be perfect.

src/lua/CMakeLists.txt
Web-eWorks: As in I shouldn't include that director, or I'm not doing it right now? Testing shows a copy of `buildopts.h` is being generated at several levels in `build/`, and I'm not 100% sure why it's being created in those different places...

Web-eWorks:
Working on `package.reimport()` now!

Also there are about 570 calls to `import()` in our lua scripts, yay... So much double-checking to do.

Web-eWorks:
So I did some blocking out of exactly how I can accomplish reimporting files, and wound up entirely changing the semantics of our import functionality. Luckily I don't think anything was using the previous semantics to their extent so there should be very little breakage.

Basically, we're going to get rid of the implicit file-relative lookup, and allow the calling file to specify if they'd like to enable that instead. We're also dropping the filename-based caching, and using fully qualified module names instead (as they're pretty much indistinguishable).

A previous scenario that looked like this:
```
+ folder
|- file.lua
|+ awesome
||- myMod.lua
```

Where `file.lua` might have called `import("awesome/myMod.lua")`, the new behavior is to call `require '.awesome.myMod'` instead - the leading `.` is a directive for the implementation to append the module name for the source directory of the calling file to the module name, forming a *fully-qualified* module name.

This removes a lot of wiggle room for unanticipated behavior in our implementation, and allows us to throw some extra functionality in later down the line... we can let mods register their own prefixes, do interesting things with trusted vs. untrusted code, etc.

I haven't put this to code *yet*, so if anyone has suggestions or criticisms, please throw them in the ring now; otherwise I'm going to be moving full-steam-ahead on this.

Web-eWorks:
Some addenda I didn't write before hitting Comment:

The reason I'm not improving our existing way of doing things instead is because it's all but impossible to sanely reimport a file under the current method. There are numerous ways to turn a module name into a file, there's no way to specify which specific file you want to evict if there happen to be two files with the same basename in a file's lookup scope, and I'm hitting the limits of what I can do because we accept both file paths and module names to `import`.

*"What About Mods?"*

As far as I know (and I've spent a few hours picking at the filesystem code), all of the mod files are merged with the game's virtual FS at the root, and it's up to the mod author to prevent name collisions. We don't have sub-root mount points or anything like that, and we've never had that.

Yes, this change will probably break existing lua mods. I'm pretty sure we don't actually have any still existing that a: use Lua scripting and b: aren't already incompatible with the latest version of the Scenegraph format.

Yes, I'll have to update literally every `import()` call that uses paths instead of module names. Luckily, there aren't many of those (about 100 or so), almost all of them are in the newUI code (so I can strike them entirely), and I'm going to have to update everything to use `require()` instead, so that's not a big problem.

impaktor:
> I haven't put this to code yet, so if anyone has suggestions or criticisms, please throw them in the ring now; otherwise I'm going to be moving full-steam-ahead on this.

:bell:  @laarmen :bell: care to have an opinion?

laarmen:
Initial thought is "can we split this in several PRs?".

I'm liking the idea of reloading the UI code, this should make the UI dev
much easier.

The require vs import sounds good to me. I just have one question : is it
possible to get to the ".." directory in this proposal, and if so, what's
the syntax? I'm not convinced it is needed, but this case should at least
be explicitly mentioned either way.

Web-eWorks:
@laarmen
> Initial thought is "can we split this in several PRs?".

This is being merged to a work branch I'm maintaining myself and not `master`, so I'm good with the PR size. If absolutely necessary it can be split, but most of the PR size is from splitting up existing files and not the actual changes.

> I just have one question : is it possible to get to the ".." directory in this proposal, and if so, what's the syntax? I'm not convinced it is needed, but this case should at least be explicitly mentioned either way.

Most of this is paraphrased from the header doc comment I wrote but didn't commit yet. There's no way to ascend directories or specify `..` or anything like that; the `.submodule.name` syntax is just sugar to avoid having to type a really long mod name like `mods.bjornbjarn.eagleLander.lib.displayLandingRadar` (in a totally nonspecific case, of course...)

We don't let require() ascend directories to a: avoid escaping the virtual FS, and b: because then you have to canonicalize *everything* and it winds up being a massive pain. Just tack the source folder's module name on the beginning and you're good to go.

Eventually, I'll probably allow mods to register their own "virtual root" directory that gets used by `require '@lib.displayLandingRadar'` in the example above, but that's for a completely different PR.

Web-eWorks:
@impaktor 
> **EDIT** will this also reload themes? E.g. If someone want to play around with the `pigui/themes/default.lua` stuff.

Yes it will, although the way our theme-using code is set up you're better off restarting the game instead, as 90% of the code won't even notice there's a new theme object. Once everything gets implemented and properly refactored you'll be able to change theme colors and values from *inside* the game with a pigui widget, but that's a task for another time and another PR.

laarmen:
Ok, this all sounds good to me. (haven't looked at the code)

Le mar. 28 janv. 2020 à 16:14, Webster Sheets <notifications@github.com> a
écrit :

Web-eWorks:
Going to split this into two PRs to be on the safe side; I've marked this one as ready for review and am merging it into the work branch as soon as it finishes building.

PR #4777, opened 2020-01-27
Fix / improve orbit calculations.

Gliese852:
# Motivation
If you start a new game, open the system map, and turn on the display of ships, you can see their flickering, which lasts several minutes. Then the situation seems to improve, but still some ships sometimes begin to flicker. Sometimes the ship begins to fly away with acceleration to infinity, sometimes it jerks back and forth. All this ruins the impression of the game, in my opinion.

# Investigation

The problem of the entire instability of the positions of the ships turned out to be that, to display them on the map, first the orbit parameters are calculated by their coordinates, and then their position is calculated again by these parameters. This is necessary in order to rewind time back and forth.
А simple solution for the present time would be to simply store the ship's coordinates inside the orbit at t = 0 and issue them when requested, but this would not solve the problem of coordinate prediction in the future.
so I decided to leave it as it is, just improve the accuracy and stability of the calculations.

# What was done

dc05993

1. it turned out that numerical methods are implemented with a fixed number of iterations (5), and they are not always enough. Moreover, testing showed that in some cases the solution does not converge and even 10,000 iterations are not enough. Therefore, a more slowly converging bisection method was added, in case the main method fails in 10 iterations.
2. When the ship has low speed, its orbit is a very elongated ellipse, and its true anomaly is very very close to pi. Trimming the variable to 1 - 1e-6 before taking the arccos does not allow it to get close to pi enough, so there is a strong shift from the true position. Therefore, it was removed.

633e996 

1. It was found that the orbit orientation matrix is calculated through the Euler angles, which are calculated from the velocity and position vectors, and many divisions are made in which the denominator can unpredictably approach 0. All this was removed since the matrix is easily calculated directly from the vectors.
2. When moving from an elliptical orbit to a hyperbolic one, the eccentricity passes through 1 (parabola), when the average anomaly is zero everywhere, except for infinity. Therefore, the ship at this time is attracted to the point of the pericenter, because there is not enough accuracy to go from the true anomaly to the mean anomaly and back, and stay in the same coordinates. Therefore, I suggest jumping from an ellipse with e = 0.9999 to a hyperbole with e = 1.0001. These orbits are very similar, almost identical. Moreover, ellipses with 0.9999 < e < 1 are also considered as hyperbolas, so that there are no distortions at a low speed of the ship.

# Result

After these changes in the code, the positions of the ships on the system map are stabilized, no problems have been identified.

### DIFF
--- a/src/Orbit.cpp
+++ b/src/Orbit.cpp
@@ -71,8 +71,36 @@ static void calc_position_from_mean_anomaly(const double M, const double e, cons
 		// eccentric anomaly
 		// NR method to solve for E: M = E-e*sin(E)  {Kepler's equation}
 		double E = M;
-		for (int iter = 5; iter > 0; --iter) {
-			E = E - (E - e * (sin(E)) - M) / (1.0 - e * cos(E));
+		int iter;
+		for (iter = 0; iter < 10; iter++) {
+			double dE = (E - e * (sin(E)) - M) / (1.0 - e * cos(E));
+			E = E - dE;
+			if(fabs(dE) < 0.0001) break;
+		}
+		// method above sometimes can't find the solution
+		// especially when e approaches 1
+		if(iter == 10){ // most likely no solution found
+			//failsafe to bisection method
+			//max(E - M) == 1, so safe interval is M+-1.1
+			double Emin = M - 1.1;
+			double Emax = M + 1.1;
+			double Ymin = Emin - e * sin(Emin) - M;
+			double Ymax = Emax - e * sin(Emax) - M;
+			double Y;
+			for(int i = 0; i < 14; i++){ // 14 iterations for precision 0.00006
+				E = (Emin + Emax) / 2;
+				Y = E - e * sin(E) - M;
+				if((Ymin * Y) < 0)
+				{
+					Ymax = Y;
+					Emax = E;
+				}
+				else
+				{
+					Ymin = Y;
+					Emin = E;
+				}
+			}
 		}
 
 		// true anomaly (angle of orbit position)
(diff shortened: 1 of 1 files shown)


impaktor:
> open the system map, and turn on the display of ships, you can see their flickering, which lasts several minutes. Then the situation seems to improve, but still some ships sometimes begin to flicker.

There should be an open issue on this somewhere, as there's been some head scratching trying to solve it. (was ships appearing to be spawning inside the star, as seen in orbit view, the same issue, or another?).

Anyway, this sounds very promissing, although I didn't follow in the details, in spite of me once having known all this stuff, having read through [Goldstein](https://en.wikipedia.org/wiki/Classical_Mechanics_(Goldstein_book)). 

(Since you describe each commit, you could copy paste those paragraphs to their respective commit message, if you feel like it).

Also, this sounds like stuff that might interest @jaj22 (ping!).

Gliese852:
@impaktor 
>There should be an open issue on this somewhere, as there's been some head scratching trying to solve it. (was ships appearing to be spawning inside the star, as seen in orbit view, the same issue, or another?).

I think this is another problem. I looked for an issue on this topic, but could not find it, and was surprised, to be honest.
> (Since you describe each commit, you could copy paste those paragraphs to their respective commit message, if you feel like it).

Aren't they too big? Looking through the git log I have not seen such long comments on commits.

Web-eWorks:
I'd like to preface this by saying the last time I saw a copy of Classical Mechanics it was laying in a thrift store, so I'm *extremely* grateful for your contribution @Gliese852!

> Aren't they too big? Looking through the git log I have not seen such long comments on commits.

Please definitely take the long explanations and amend your commit messages with them - I'd love to have them available and the git history doesn't really associate commits with pull requests. We don't care about the length of commit messages - the more useful detail you can cram in there, the better!

I'm going to leave the PR open for about a week so @fluffyfreak and @jaj22 can take a close look and have some time to test; if there are no objections or improvements to register at that time, this will be getting immediately merged.

impaktor:
> I think this is another problem. I looked for an issue on this topic, but could not find it, and was surprised, to be honest.

Basically, what @Web-eWorks said. First sentence should encapsulate what it's about. The body can then elaborate in detail. Reason people write short commit messages is usually lazines, or "the code documents itself / is obvious (although only to the author at commit time)". 

Examples:

5f0c8ca151a8da43145eb07580eb5b6681eb1637
c0f820fa527c5032aa0d362a7d69241c29d5d914
77fc68eb3e24cc1f0129dba00706a1577928849e
c63b0feb99e3f6e84039294cb2d38b3e2779d23b

Web-eWorks:
I've pulled this and tested it locally and everything looks good; waiting for @fluffyfreak or @jaj22 to sign off and then I'll merge.

fluffyfreak:
I'm impressed, going to merge now 👍 
Great explanation helped me follow along with the maths too and I appreciate the long comments in the commits

impaktor:
@Gliese852 this warranted mention in the release-tweet :+1: as it's been bugging many for years:
https://twitter.com/pioneerspacesim/status/1224399814483480576

PR #4779, opened 2020-01-29
Fix on-demand glyph loading

### LINKED ISSUE #4778: Bug in font charset detection causing memory leak
Detected while  testing #4695 

### Observed behaviour

Certain unicode characters are not displayed properly with some of the fonts, even when the font contains the character. This leads to constant re-baking of the font, and a huge (~100MB/s) memory leak.

### Expected behaviour

Characters that are in a font are rendered correctly. If they aren't handled by the font, a fallback is used or an error is thrown. In either case, there should be no memory leaks.

### Steps to reproduce

1. Set language to German

2. Start new game

3. Open the Equipment Market (or any view that displays a formatted money amount > $1 000). The formatted amount of money displayed should contain a '?' like in the following screenshot: 

FontBug

### Causes

This is one of those bugs where a number of things has to happen at the same time for it to come up. First of all in the `format_money` function, if the locale has a white space set as the thousand-separator, we overwrite that with a unicode fixed space:

https://github.com/pioneerspacesim/pioneer/blob/99a2e0b452e9bd3c559f3240c50eca6d0eb88c44/src/utils.cpp#L43-L45

That's why it's necessary to switch to German, or another language that use a white space as the thousand-separator, in order to reproduce the bug.

When we try to render the character ImGui's `FindGlyph` function says it's not there, so a fallback character is displayed instead, and the fixed space is "reported missing", therefore PiGui tries to add it explicitly to the font, and rebake it. The twist is that PiGui says `PionilliumText22L-Medium.ttf` contains the `\u00a0` fixed space, so it wants to keep using that font:

https://github.com/pioneerspacesim/pioneer/blob/99a2e0b452e9bd3c559f3240c50eca6d0eb88c44/src/PiGui.cpp#L139-L146

But when ImGui calls `stbtt_FindGlyphIndex` it can't find it, 0 is returned and the whole thing starts over in the next frame

https://github.com/pioneerspacesim/pioneer/blob/99a2e0b452e9bd3c559f3240c50eca6d0eb88c44/contrib/imgui/imstb_truetype.h#L1501-L1518

So is it a Pioneer bug or an ImGui bug? A quick test seems to indicate that the "fixed space" character is indeed defined in the Pionillium font:

```
$ hb-shape PionilliumText22L-Medium.ttf `echo -ne "\u00a0"`
[space=0+235]
```

So it seems the problem is on ImGui side, but at the moment I can't make heads or tails out of what's going on in `stbtt_FindGlyphIndex`. Maybe someone else knows off the top of their head, if not I'll resume work on this later.

### P.S.

I'm pretty sure the bug is in font-parsing now. When I was debugging this I originally misunderstood the meaning of the `containsGlyph` function here:

https://github.com/pioneerspacesim/pioneer/blob/99a2e0b452e9bd3c559f3240c50eca6d0eb88c44/src/PiGui.cpp#L139-L146

So I thought "Wait, why are we adding a glyph when the font already contains it? Shouldn't it be the other way around?" and added a negation. It actually helped! Kind of. The character ranges on Pigui-side are hardcoded (we ~~should probably~~ have to (see below) change that somehow, as well as the name of the `containsGlyph` function and `m_ranges` variable), so the alternative fallback fonts like DejaVuSans.ttf jumped into the AddGlyph block. On ImGui side the fixed space was also detected and displayed correctly, that's valid because that font also contains the character:

```
$ hb-shape DejaVuSansMono.ttf `echo -ne "\u00a0"`
[nonbreakingspace=0+1233]
$ hb-shape DejaVuSans.ttf `echo -ne "\u00a0"`
[nonbreakingspace=0+651]
```

Although it looks like it's defined differently, maybe that has something to do with it? Can't say I know much about the ttf format.

### P.P.S

After thinking about it more, I guess the bug is also on the Pioneer side.

https://github.com/pioneerspacesim/pioneer/blob/99a2e0b452e9bd3c559f3240c50eca6d0eb88c44/src/PiGui.cpp#L745-L748

This is the bit I mentioned, where the charsets are hardcoded, the last argument in each `PiFace` constructor is the range of valid characters the font is supposed to support, and Orbiteer and Pionillium have it set to from 0x0000 to 0xffff. Even if the fonts contained every single unicode character, this would still be wrong to have the range hardcoded. But since they don't have every character, this means we're treating that range as a preference - "ImGui, try to use this font to handle these characters, if you can, please". In that case we need to also handle the case when ImGui tells us it cannot handle that character with that font.

### Conclusions

- We need to remove the hardcoding of `m_ranges` in `PiFace`. Either try to read the valid character ranges from the file directly, or to save the glyphs we already tried loading with ImGui in a variable, and use it to return false in `containsGlyph` the next time the same glyph comes up.

- Should we drop that whitespace / unicode fixed space replacement in format_money? We seem to be able to render whitespaces just fine.

- (Optional) figure out why ImGui can't read the fixed space character for Pionillium

- (Optional) Rename `m_ranges` to something like `m_valid_ranges` and `containsGlyph` to `canHandleGlyph`, or describe what they do in comments. It's hard to figure out what these are actually for at first glance.

vakhoir:
Fixes #4778

For all the talk in the issue, the fix seems relatively simple, the following description is also in the commit message:

Change the approach to checking if a font can handle a glyph. The original approach was to pass the valid glyph range to the constructor. When it's not found by ImGui we call AddGlyph where we check if it's valid for the font. In the event where ImGui couldn't handle a glyph that was defined as valid in the constructor, that led to constant re-baking of fonts. The new approach is to change m_ranges (which used to contain valid glyph ranges) to m_invalid_glyphs, which contains only invalid glyphs that we tried to add to the font but failed. The next time AddGlyph is called, we skip the font since it's unable to handle the glyph.

But I have to say, I'm not entirely confident about what I did here. @ecraven I think you're the one that wrote all this, I also saw an issue you opened in the ImGui repo, is there any chance you can chime in on this?

### DIFF
--- a/src/PiGui.cpp
+++ b/src/PiGui.cpp
@@ -138,7 +138,7 @@ void PiGui::AddGlyph(ImFont *font, unsigned short glyph)
 	}
 	PiFont &pifont = pifont_iter->second;
 	for (PiFace &face : pifont.faces()) {
-		if (face.containsGlyph(glyph)) {
+		if (face.isValidGlyph(glyph)) {
 			face.addGlyph(glyph);
 			m_should_bake_fonts = true;
 			return;
@@ -743,9 +743,9 @@ PiGui::PiGui() :
 	m_should_bake_fonts(true)
 {
 	PiFont uiheading("orbiteer", {
-									 PiFace("DejaVuSans.ttf", /*18.0/20.0*/ 1.2, { { 0x400, 0x4ff }, { 0x500, 0x527 } }), PiFace("wqy-microhei.ttc", 1.0, { { 0x4e00, 0x9fff }, { 0x3400, 0x4dff } }), PiFace("Orbiteer-Bold.ttf", 1.0, { { 0, 0xffff } }) // imgui only supports 0xffff, not 0x10ffff
+									 PiFace("DejaVuSans.ttf", /*18.0/20.0*/ 1.2), PiFace("wqy-microhei.ttc", 1.0), PiFace("Orbiteer-Bold.ttf", 1.0) // imgui only supports 0xffff, not 0x10ffff
 								 });
-	PiFont guifont("pionillium", { PiFace("DejaVuSans.ttf", 13.0 / 14.0, { { 0x400, 0x4ff }, { 0x500, 0x527 } }), PiFace("wqy-microhei.ttc", 1.0, { { 0x4e00, 0x9fff }, { 0x3400, 0x4dff } }), PiFace("PionilliumText22L-Medium.ttf", 1.0, { { 0, 0xffff } }) });
+	PiFont guifont("pionillium", { PiFace("DejaVuSans.ttf", 13.0 / 14.0), PiFace("wqy-microhei.ttc", 1.0), PiFace("PionilliumText22L-Medium.ttf", 1.0) });
 	AddFontDefinition(uiheading);
 	AddFontDefinition(guifont);
 
@@ -759,14 +759,10 @@ PiGui::PiGui() :
 	GetFont("pionillium", 14);
 };
 
-const bool PiFace::containsGlyph(unsigned short glyph) const
+const bool PiFace::isValidGlyph(unsigned short glyph) const
 {
 	PROFILE_SCOPED()
-	for (auto range : m_ranges) {
-		if (range.first <= glyph && glyph <= range.second)
-			return true;
-	}
-	return false;
+	return (m_invalid_glyphs.count(glyph) == 0);
 }
 
 void PiFace::addGlyph(unsigned short glyph)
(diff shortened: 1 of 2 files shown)


src/PiGui.cpp
```diff
@@ -759,12 +759,11 @@ PiGui::PiGui() :
...
 };
 
-const bool PiFace::containsGlyph(unsigned short glyph) const
+const bool PiFace::isValidGlyph(unsigned short glyph) const
 {
 	PROFILE_SCOPED()
-	for (auto range : m_ranges) {
-		if (range.first <= glyph && glyph <= range.second)
-			return true;
+	if(m_invalid_glyphs.find(glyph) == m_invalid_glyphs.end()) {
```
Web-eWorks: Can I recommend using `!m_invalid_glyphs.count(glyph)` here instead? It's shorter, still maintains the same time-complexity, and easily converts to bool.

vakhoir: > One thing stands out to me; is there a reason for using `std::unordered_map` over `std::unordered_set`?

Yeah, the reason is my C++ got rusty ;)

> Can I recommend using !m_invalid_glyphs.count(glyph) here instead? It's shorter, still maintains the same time-complexity, and easily converts to bool.

Done!

Web-eWorks:
Alright, I'll pull this PR locally in about twelve hours and try to dig into it to double-check everything is working. Thank you @vakhoir for this bugfixing (and bugfinding too!)

PR #4782, opened 2020-01-31
Add keybinding "Toggle Lua Console" to settings menu

impaktor:
We need to be able to select this, as default "`" can be located in retarded
position at some keyboard layouts.

I'd like to have a discussion as if there's a reason why a load of key-bindings aren't exposed to lua? Is this on purpose? E.g. `SECTOR_MAP_VIEW`, `EXTERNAL_VIEW`, `RADAR_CONTROL`?

1. do we want them in settings menu?
2. is that to be done for this PR?

__EDIT__ I'd like to have `WITH_DEVKEYS` as well, so players/devs easily can see what keys are available, and I think having dev tools easily accessible encourages dev-ing.

### DIFF
--- a/data/lang/input-core/en.json
+++ b/data/lang/input-core/en.json
@@ -43,6 +43,10 @@
     "description": "Descriptive name for the CycleCameraMode action.",
     "message": "Cycle Camera Mode"
   },
+  "BIND_DECREASE_SCAN_RANGE": {
+    "description": "Descriptive name for the BindDecreaseScanRange action.",
+    "message": "Decrease Radar Scan Range"
+  },
   "BIND_DECREASE_TIME_ACCELERATION": {
     "description": "Descriptive name for the DecreaseTimeAcceleration action.",
     "message": "Decrease Time Acceleration"
@@ -50,6 +54,10 @@
   "BIND_FRONT_CAMERA": {
     "description": "Descriptive name for the FrontCamera action.",
     "message": "Switch to Front Camera"
+  },
+  "BIND_INCREASE_SCAN_RANGE": {
+    "description": "Descriptive name for the BindIncreaseScanRange action.",
+    "message": "Increase Radar Scan Range"
   },
   "BIND_INCREASE_TIME_ACCELERATION": {
     "description": "Descriptive name for the IncreaseTimeAcceleration action.",
@@ -115,6 +123,10 @@
     "description": "Descriptive name for the ToggleRotationDamping action.",
     "message": "Toggle Rotation Damping"
   },
+  "BIND_TOGGLE_SCAN_MODE": {
+    "description": "Descriptive name for the BindToggleScanMode action.",
+    "message": "Toggle Radar Scan Mode"
+  },
   "BIND_TOGGLE_SET_SPEED": {
     "description": "Descriptive name for the ToggleSetSpeed action.",
     "message": "Toggle Speed Control Mode"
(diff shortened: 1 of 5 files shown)


Web-eWorks:
The simple answer is that it's my fault. :D

The long answer is that the new input system isn't 100% complete yet. I need to expose bindings to lua in a way that we don't have to synthesize them via string manipulation (see the current pigui code for keybinding and marvel...), automatically distribute SDL events, implement key chording (so shift+Ctrl+Joy 5 etc.) and convert the last uses of the previous keybinding system over.

That last one in particular is the biggest reason a good number of keys aren't exposed, because they're in use by systems that (hopefully) are getting the boot and being redesigned. I simply haven't had the time or mental energy to determine which systems are worth patching to the new input paradigm and which aren't worth the effort, and so I've temporarily filed them all under the latter.

@impaktor as an aside, you'll probably need to hook the lua console binding up to the console code in Pi.cpp unless I'm totally off base. I'm on the road today so I can't easily double check...

impaktor:
OK, I interpret it as:
1. this is scheduled for total re-design (at some point), so might not be adviseable to patch old code, like I've done here.
2. I've tested this PR, and it works, I tested it in Worldview, StationView, and SectorView. I did have to re-start the game after putting lua-console on an umlaut-key.

Web-eWorks:
> OK, I interpret it as:
> 
>     1. this is scheduled for total re-design (at some point), so might not be adviseable to patch old code, like I've done here.
> 
>     2. I've tested this PR, and it works, I tested it in Worldview, StationView, and SectorView. I did have to re-start the game after putting lua-console on an umlaut-key.

Not as such! It's more "I don't have time to *think* about it, much less try to do it", so I'm very happy to see you're picking things up here! I'd love to get as many inputs as possible converted over to the new system, and if I don't have to do that, it's even better.

impaktor:
OK, I could do the same as I've done here with more key-bindings, but I'm not sure if that's correct or not.

What I did here was just a copy-paste of existing code-patterns and trial and error, although how to do that took +30 minutes to figure out since I spent quite some time to try to understand the settings-window.lua, which turned out to be not where this stuff goes (as that just iterates over what pre-existing bindings have been exposed).

impaktor:
I've now added another commit, where I also added Radar control group as well, I put it in `PlayerShipController.cpp`, as I assume it aught not be in worldview directly.

I have a lingering feeling that we should be able to remve a number of key-control related strings from `LangStrings.inc.h`, but the're needed for `Keybindings.inc.h`, which still makes C++ side aware for the strings, I assume, although it shouldn't need the translation strings, I think.

Web-eWorks:
@impaktor on second look, you will need to update the relevant code to use the new keybindings; just declaring a new binding with the same name doesn't make the existing code use them. See e.g. this line in Pi.cpp:

https://github.com/pioneerspacesim/pioneer/blob/68cc2ea05718402443bbb07eb79da9746757ffcd/src/Pi.cpp#L694

As you can see, that's referring to the input binding declared in the `KeyBindings` namespace, which is the one declared in KeyBindings.inc.h - not the new binding you've created. The only reason why it seems to work is that the new binding code is sharing the same name in the config file as the old binding; everything in C++ is still using the old binding method and it just so happens to coexist with the new one by accident.

impaktor:
Interesting, so the code works, but for "ghost" reasons :P. I can't quite figure out what the correct code for toggling lua console would look like, and I've tried following the code path, but that just gave me vertigo, so I'm closing for now. I'll re-open if I figure it out.

Plus you say you're working on a new input system, so... Piority one for me is to get `SECTOR_MAP_VIEW`, `EXTERNAL_VIEW`, `RADAR_CONTROL` into settings window.

PR #4784, opened 2020-02-01
Small additional improvements to orbit calculation

Gliese852:
Sorry for not noticing this right away. When the ship takes off from the planet, the eccentricity of its orbit is for some time in the range 0.9999 .. 1 and turns into a hyperbole 1.0001. The velocity vector is directed from the planet, so a line is drawn that goes from the ship to infinity. It is not right. These changes solve the problem.
Everything seems to be taken into account now. Feel free to kick.

### Was:
orbit with 0.9999 < e < 1.0001 forced to e = 1.0001 (hyperbola)
### Now:
orbit with 0.9999 < e < 1      forced to e = 0.9999 (ellipse)
orbit with 1   <=   e < 1.0001 forced to e = 1.0001 (hyperbola)

In an elliptical orbit, when the eccentricity is fixed at 0.9999, and
the speed increases, the semi-major axis shrinks and the apocenter moves
toward the star. Therefore, a condition check has been added that prevents
the ship from shifting in this case.

### DIFF
--- a/src/Orbit.cpp
+++ b/src/Orbit.cpp
@@ -332,7 +332,8 @@ Orbit Orbit::FromBodyState(const vector3d &pos, const vector3d &vel, double cent
 	if (ret.m_eccentricity < 0.0) ret.m_eccentricity = 0.0;
 	ret.m_eccentricity = sqrt(ret.m_eccentricity);
 	//avoid parabola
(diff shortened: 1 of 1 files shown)


impaktor:
Cool

Btw, it looks like github doesn't understand who the author of your commit is, i.e. it's not linking your github account with the commits, might be differently set `name` in your `~/.gitconfig` - doesn't matter, so long as you're fine with it.

PR #4786, opened 2020-02-02
New require() implementation

Web-eWorks:
We now load files via Lua 5.2 require() semantics instead of raw paths.
This means we've dropped the current directory implicit lookup, and
instead we use module names to semi-uniquely identify a package.

We mimic the behavior of package.path for the data/libs/ directory,
with an eye to allowing client code to extend that behavior later on.

Loaded modules are cached in a two-stage cache; the first stage maps
module names to file paths to avoid duplicate loads, and the second
stage maps file paths to the return value of a module.

package.reimport(name) can be used to purge the cache of a specific
name and force it to be reimported (as the name suggests). It returns
the imported module, similarly to require().

This implementation deprecates import() and translates old-style calls
to that function to the new module names. It complains loudly in the log
when it does so, so we can keep track of what modules need to be updated.

-----

This is stage two of the Great Lua Refactor Mk. VII (or so). ~~It's not quite running; there's a FIXME that explains why, and I still have to double-check and commit my updates to `pi_lua_import_recursive` et. al. before you can attempt to test it,~~ but I still want to get eyes on it overnight and tomorrow while I'll be away from the workstation.

Stage three is going to involve updating *all the things*. It will be coming in a separate joint PR with @vakhoir, after which I will rebase it all and merge directly back into `master` without a PR. (That's what these PRs are for!)

Still... @vakhoir this is it; `package.reimport('pigui.views.xxx')` is now perfectly possible!


Web-eWorks:
@vakhoir can I ask you to take a look at the C++ code and do a quick review? I'm mostly looking for functions which don't have adequate comments to describe all of the stack-based black magic that's going on; I've mangled this code pretty heavily with GDB so you don't have to worry about correctness.

impaktor:
I call upon thee, @laarmen, to let thy opinions rain down from the heavens in
bountiful measures, and let thy judgement be swift and fair, ...if time be.

Web-eWorks:
Alright, everything is bugfixed and confirmed working. I'm working on packaging up only the breaking changes to the lua files for this PR so that testers can actually... *run* it, and then I'll mark this as ready for review.

vakhoir:
> @vakhoir can I ask you to take a look at the C++ code and do a quick review? I'm mostly looking for functions which don't have adequate comments to describe all of the stack-based black magic that's going on; I've mangled this code pretty heavily with GDB so you don't have to worry about correctness.

Will do, I should have time to take a look tomorrow.

src/lua/core/Import.cpp
```diff
@@ -48,17 +95,267 @@ static std::string get_caller(lua_State *L)
...
+* This function discards the trailing component of the directory path. If you
+* wish to retain it, ensure the path string ends with '/'.
+*/
+static std::string path_to_module(std::string path)
+{
+	std::string module_name;
+	// simple allocation optimizations
+	if (path.size() > 8 && path.find_first_of('/') != std::string::npos) {
+		module_name.reserve(8);
+	}
```
laarmen: I'd be curious to know the story behind this optimisation :)

src/lua/core/Import.cpp
```diff
@@ -48,17 +95,267 @@ static std::string get_caller(lua_State *L)
...
+	while (end != std::string::npos) {
+		auto fragment = path.substr(start, end - start);
+
+		if (fragment.size() > 0) {
+			// Paths containing the separator character cannot be translated into a module name
+			// this means you can't do implicit lookup using a path starting in a hidden directory, e.g.
+			// /data/.local/mylib.lua
+			if (fragment.find('.') != std::string::npos)
+				return "";
+			module_name += module_name.empty() ? fragment : '.' + fragment;
```
laarmen: In case anybody's wondering, the ternary is evaluated after the + here. I had to look it up to be sure -_-'

Two minor optimizations (not worth it here IMHO but the general pattern is worth bearing in mind) would be to

  * Not allocate a temporary string for the `'.' + fragment`, you'll need to split the operation in two appends.
  * Append the dot *after* the fragment name, unconditionally, and after the loop simply remove the extraneous '.' at the end of the string.

In the general case, the latter avoids paying the cost of testing for the initial case at each loop iteration. As I said, not really worth it here, but a nice trick nonetheless :-)

src/lua/core/Import.cpp
```diff
@@ -48,17 +95,267 @@ static std::string get_caller(lua_State *L)
...
+		// Store the module in the import cache
+		lua_pushvalue(L, -1);
+		lua_setfield(L, cacheIdx, path.c_str());
+
+		LUA_DEBUG_END(L, 1);
+		return true;
+	}
+
+	LUA_DEBUG_END(L, 0);
+	return false;
```
laarmen: Minor nitpick: If you're to have multiple returns in the function, I find it more readable to have the common case in the main indentation level and the error cases in the indented blocks (early return pattern). You'd have

```cpp
if (!fileInfo.Exists()) {
    LUA_DEBUG_END(L, 0);
    return false;
}

auto data = fileInfo.Read();
//etc...
```

laarmen:
The design is exactly as advertised, I like it a lot. I have a few comments on the code itself, mostly readability and minor optimisations, but feel free to ignore them, or better yet, turn them into cleanup tasks to take up for would-be contributors in search for easy tasks.

src/lua/core/Import.cpp
```diff
@@ -48,17 +95,267 @@ static std::string get_caller(lua_State *L)
...
+
+	lua_getfield(L, LUA_REGISTRYINDEX, MODULE_MAP_CACHE_NAME);
+	lua_getfield(L, -1, moduleName.c_str());
+	std::string path;
+	if (lua_isstring(L, -1))
+		path = lua_tostring(L, -1);
+	lua_pop(L, 2);
+	LUA_DEBUG_END(L, 0);
+	return path;
+}
```
laarmen: Dang, when I read code like this I'm glad to be mainly programming in Rust and C these days. I'm guessing the cache miss signal is the empty string ?

impaktor:
Thanks for review @laarmen!

laarmen:
Well, I'm not finished on the nitpicks yet, I just had to get back to work,
so I figured I'd give the overall conclusion

src/lua/core/Import.cpp
Web-eWorks: This one comes from the discussion about stdlib containers and your post on optimal sizing for `reserve()` in #4764... however, on second thought the small-string optimization renders this point completely moot. :(

src/lua/core/Import.cpp
Web-eWorks: I'm extremely used to Lua's string semantics, where strings are immutable and thus changing them by even one character triggers a copy-and-allocate operation. Under Lua's semantics it's most efficient to merge two small strings together and then into the larger string than it is to merge them individually, which would be what is tripping me up here.

Thanks for the tip though, I'll definitely use it!

src/lua/core/Import.cpp
Web-eWorks: Yeah, I consider it semantically valid to signal the lack of a cached path for the module by simply returning an empty string.

src/lua/core/Import.cpp
```diff
@@ -60,390 +104,393 @@ static std::string get_caller(lua_State *L)
...
+		return true;
+	} else
+		lua_pop(L, 1);
 
-	bool isImported = false;
-	std::string realName = importInfo.importName;
+	// FileSystem::JoinPathBelow can't join a path starting with '/' for some reason
+	// TODO: find out why and remove that restriction.
+	std::string filePath = (path[0] == '/') ? path.substr(1) : path;
+	auto fileInfo = FileSystem::gameDataFiles.Lookup(filePath);
```
vakhoir: That would be because a leading `/`is interpreted as the file system's root folder, and is only allowed if the basepath is set to in `src/FileSystem.cpp` is also set to root. However we explicitly set it to the game's data folder, and user data folder on init so we don't end up crawling all over the file system.

vakhoir:
> I'm mostly looking for functions which don't have adequate comments to describe all of the stack-based black magic that's going on

Hm, the problem here is that unless someone has the C++ lua API fresh in their mind, even with comments some of this stuff looks like black magic. But I haven't spotted anything that seems to come out of the blue so far. I'll go through everything again tomorrow to make sure.

> Still... @vakhoir this is it; package.reimport('pigui.views.xxx') is now perfectly possible!

Hm, I tried playing around with it, and while the console is giving me messages about file(s) being reloaded, it seems to have no effect on the ui.

Web-eWorks:
@vakhoir you're still going to have to hook it up in the actual UI code; this doesn't e.g. remove modules from GameView or anything, but it does let you basically re-execute the source file. Either you'll want to make your reimported file call e.g. `infoView:removeModule('ship-info')` before registering the module, or you'll need some other way to de-register code before re-importing it. This is just the basic building block that makes all of that possible.

Web-eWorks:
As an addenda, if you put a file-scope `print()` call in the re-imported file, you'll see in the console that it is being re-executed when you call `package.reimport()`, so my job is done :D

src/lua/core/Import.cpp
Web-eWorks: The semantics behind that decision boggle my mind though; the name `JoinPathBelow` reads as though the function was purposefully *intended* to translate an absolute virtual-root path into a real filesystem path, and not throw `std::invalid_argument` if passed a non-relative second path.

Oh well, just another reason to re-design the interface of the filesystem abstraction. Just pin that on the docket in about five year's time... :D

Web-eWorks:
Going to merge this by Sunday evening EST, so get your reviews / comments in please! I've already started work on the great `require()`ing, and I haven't noticed any issues or crashes so as far as I'm concerned this PR is good to merge.

impaktor:
Ah, this wasn't merged into master, but it's own branch. (Just a note to self)

PR #4790, opened 2020-02-07
Police to pigui

impaktor:
## Introduction
This is my first pigui contribution. I hope to get some eyes on this, there's at least one "breaking issue" that needs fixing before merge (see below), that I haven't managed to figure out.

Other than that, it has all the same functionality as the original, which wasn't very flashy to begin with, since that was also something I barely managed to hack together.

## Description
So I'm sure the data here can be presented better, (e.g. remove gray color for crime record, use some collapsing header or similar instead), but the question is how much more work should be done before this is merged.

My plan was to do another PR after this where I add some sort of dialogue button as well, so the player can ask for missing persons (needed for #4746), but if this PR looks OK, I could attack the ShipRepair screen next, which should be trivially similar to this.

I've made a little Debug-module that allows for adding _crime_ and _money_ to the player's record that's useful for testing this PR, if anyone wants to play (it also allows playing with _hull damage_, swap _ships_, adjust _reputation_ and _killcount_). See __Appendix__-section below.

## Know issues
1. When trying to pay fines higher than what the player has, the modal popup window saying "not enough money" crashes the game when clicking "OK". I'm not sure what the correct way of implementing this is. Pointers are most welcome, ping @vakhoir

2. `widgetSizes` table is a copy-paste from lobby, that I haven't put that much thought into. Ideally I should rename it (and think through what it's doing)

3. I'm not sure how pioneer-style compliant I am with colors, etc. Please point out if it's too uncceptably deviant.

4. __edit__ also, I added a as a separate commit the faction name and police faction name. Not sure we want this, if we want it I'll add it to translation system and squash the commit

## Screenshots
Compare below with original shown in #3254

2020-02-07-162203_3200x900_scrot

2020-02-07-162052_3200x900_scrot

2020-02-07-162110_3200x900_scrot

2020-02-07-162132_3200x900_scrot

## Appendix: Debug-window
Drop into `data/modules/` adds a window on worldview.
screenshot-20200207-160544

```lua
-- Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details

--[[
	data/pigui/modules/add_crime.lua

	Written for @impaktor as a tutorial on how to get started with pigui by @sturnclaw, extended by @impaktor
--]]

local Game = import('Game')

local Format = import "Format"

-- pigui is traditionally imported as 'ui' for simplicity.
local ui = require 'pigui'
-- gameView is needed to dispatch to our Hello World window
local gameView = require 'pigui.views.game'
-- we want all displayed text to be localized
local lui = require 'Lang'.GetResource('ui-core')

-- ui.WindowFlags is an acceleration structure to allow composing window flags
-- once and reusing them with each call.
local window_flags = ui.WindowFlags {
	-- We don't want to be able to collapse the window.
	"NoCollapse"
}


local amount = 1000
local selected = 0

-- a small function, to add some offences to player,
-- (for testing ui layout of police tab)
local Legal = import "Legal"
local utils = import "utils"
local Lang = import 'Lang'
local l = Lang.GetResource("ui-core")

-- build list of all crime types:
local crime_types = {}
for k, v in pairs(Legal.CrimeType) do
	table.insert(crime_types, k)
end

local Character = import("Character")

-- build list of all ships
local ShipDef = import("ShipDef")
-- local shipdefs = utils.build_array(utils.filter(function (k,def) return def.tag == 'SHIP' end, pairs(ShipDef)))
local selected_ship = 0
local changed_ship = false

local shipdefs = {}
for k,v in pairs(ShipDef) do
	if v.tag == "SHIP" then
		table.insert(shipdefs, v.id)
	end
end


-- Register this window as a Game-view module. It will be displayed when the
-- player is in the World View.
-- registerModule takes two parameters, the unique key of the module, and the
-- module descriptor table.
gameView.registerModule("add-crime", {

	-- Pretty self explanatory, if this is true, draw() is called when the
	-- ship is in hyperspace as well as in normal space.
	ShowInHyperspace = true,

	-- Called once per frame while the module is active. The function is passed
	-- the module object (this table) and the frame delta time
	draw = function(self, deltaTime)
		-- ui.window takes three parameters: the window title, the window flags,
		-- and a function containing the body of the window.

		-- the window title is an IMGUI string ID - two windows cannot share
		-- the same ID at the same stack position. To work around this,
		-- everything after ## is not part of the title, but instead used to
		-- make the window ID unique.
		ui.window("Debug!##id42", window_flags, function()

			if ui.collapsingHeader("Money", {"DefaultOpen"}) then
				-- args: label, default, min, max, (optional: str format)
				amount = ui.sliderFloat("Amount", amount, 0, 1000000)
				if ui.button("Give money", Vector2(100, 0)) then
					Game.player:AddMoney(amount)
				end
				ui.sameLine()
				ui.text(Format.Money(Game.player:GetMoney()))
			end

			ui.separator()

			if ui.collapsingHeader("Crime", {"DefaultOpen"}) then
				local changed, ret = 0

				ui.text("ADD CRIMINAL CHARGES:")
				local selected_crime = 0
				for i, v in pairs(crime_types) do
					if ui.selectable(v:lower(), selected_crime == i, {}) then
						Legal:notifyOfCrime(Game.player, crime_types[i])
					end
				end

				ui.dummy(Vector2(0,10))
				-- CRIME
				local crimes, fine = Game.player:GetCrimeOutstanding()
				if #utils.build_array(pairs(crimes)) > 0 then
					ui.text(l.OUTSTANDING_FINES)
					for k,v in pairs(crimes) do
						ui.text(v.count.."\t"..Legal.CrimeType[k].name)
					end
					ui.text("Fine:\t".. tostring(fine))
				end

				ui.dummy(Vector2(0,10))
				local past_crimes, _ = Game.player:GetCrimeRecord()
				if #utils.build_array(pairs(past_crimes)) > 0 then
					ui.text(l.CRIMINAL_RECORD)
					for k,v in pairs(past_crimes) do
						local s = v.count.."\t"..Legal.CrimeType[k].name
						ui.text(s, Vector2(100, 0))
					end
				end
			end

			-- Ship
			if ui.collapsingHeader("Ship", {}) then
				changed_ship, idx = ui.combo("Ships", selected_ship, shipdefs)
				if changed_ship then
					selected_ship = idx
					local i = selected_ship + 1
					Game.player:SetShipType(shipdefs[i])
				end

				-- hyperdrive maintenence
				local hyperdrive = Game.player:GetEquip('engine',1)
				if hyperdrive then
					ui.text(hyperdrive.l10n_key)
				end

				local hull = ui.sliderInt("Hull", Game.player.hullPercent, 0, 101)
				Game.player:SetHullPercent(hull)
			end

			-- Reputation & kills
			if ui.collapsingHeader("Reputation & kills", {}) then
				Character.persistent.player.reputation = ui.sliderInt("Reputation", Character.persistent.player.reputation, 0, 512)
				ui.sameLine()
				ui.text(Character.persistent.player:GetReputationRating())

				Character.persistent.player.killcount = ui.sliderInt("Kills", Character.persistent.player.killcount, 0, 6000)
				ui.sameLine()
				ui.text(Character.persistent.player:GetCombatRating())
			end

		end)
	end
})

```

### DIFF
--- a/data/pigui/modules/station-view/07-police.lua
+++ b/data/pigui/modules/station-view/07-police.lua
@@ -1,17 +1,161 @@
 -- Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details
 -- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
+local Game = import "Game"
+local Rand = import "Rand"
 local ui = import 'pigui/pigui.lua'
 local StationView = import 'pigui/views/station-view'
 local Lang = import 'Lang'
-
+local Legal = import "Legal"
+local utils = import "utils"
+local InfoFace = import 'ui/PiguiFace'
+local Format = import "Format"
+local Character = import "Character"
 local l = Lang.GetResource("ui-core")
+
+local ModalWindow = import 'pigui/libs/modal-win.lua'
+
+local rescaleVector = ui.rescaleUI(Vector2(1, 1), Vector2(1600, 900), true)
+
+local pionillium = ui.fonts.pionillium
+local orbiteer = ui.fonts.orbiteer
+local colors = ui.theme.colors
+local face = nil
+local stationSeed = false
+
+local popup = ModalWindow.New('policePopup', function(self)
+	ui.text(l.YOU_NOT_ENOUGH_MONEY)
+	ui.dummy(Vector2((ui.getContentRegion().x - 100*rescaleVector.x) / 2, 0))
+	ui.sameLine()
+	if ui.button(l.OK, Vector2(100*rescaleVector.x, 0)) then
+		self:close()
+	end
+end)
+
+local gray = Color(100, 100, 100)
+
+local widgetSizes = ui.rescaleUI({
+	itemSpacing = Vector2(4, 9),
+	windowPadding = Vector2(14, 14),
+	faceSize = Vector2(586,565),
+	crimeRecordColumnWidth = 55,
+	buttonSize = Vector2(100,0),
+	dummySpaceSmall = Vector2(0, 10),
+	dummySpaceMedium = Vector2(0, 50),
+}, Vector2(1600, 900))
+
+
+local function payfine(fine)
+	if Game.player:GetMoney() < fine then
+		popup:open()
+		return
+	end
+	Game.player:AddMoney(-fine)
+	Game.player:ClearCrimeFine()
+end
+
+
+local function crime_record()
+	local past_crimes, stump = Game.player:GetCrimeRecord()
+	if #utils.build_array(pairs(past_crimes)) > 0 then
+		ui.withFont(orbiteer.xlarge.name, orbiteer.xlarge.size,
+			function() ui.textColored(gray, l.CRIMINAL_RECORD) end)
+		ui.withFont(pionillium.medlarge.name, pionillium.medlarge.size,
+			function ()
+				for k,v in pairs(past_crimes) do
+					ui.textColored(gray, v.count)
+					-- start second column at this position:
+					ui.sameLine(widgetSizes.crimeRecordColumnWidth)
+					ui.textColored(gray, Legal.CrimeType[k].name)
+				end
+		end)
+	end
+end
+
+
+local function outstanding_fines()
+	local crimes, fine = Game.player:GetCrimeOutstanding()
+	if #utils.build_array(pairs(crimes)) > 0 then
+
+		-- headline
+		ui.dummy(widgetSizes.dummySpaceSmall)
+		ui.withFont(orbiteer.xlarge.name, orbiteer.xlarge.size,
+			function() ui.text(l.OUTSTANDING_FINES) end)
+		-- wrap list in medlarge font
+		ui.withFont(pionillium.medlarge.name, pionillium.medlarge.size,
+			function()
+				for k,v in pairs(crimes) do
+					ui.text(v.count)
+					-- start second column at this position:
+					ui.sameLine(widgetSizes.crimeRecordColumnWidth)
+					ui.text(Legal.CrimeType[k].name)
+				end
+				local pay_fine_text = string.interp(l.PAY_FINE_OF_N,
+					{ amount = Format.Money(fine) })
+				ui.text(pay_fine_text)
+
+				if ui.button(l.PAY, widgetSizes.buttonSize) then
+					payfine(fine)
+				end
+		end)
+	else
+		ui.withFont(pionillium.large.name, pionillium.large.size,
+			function() ui.text(l.WE_HAVE_NO_BUSINESS_WITH_YOU) end)
+	end
+end
+
+
+local function drawPolice()
+
+	local intro_txt = string.interp(l.THIS_IS_FACTION_POLICE,
+		{ faction_police = Game.system.faction.policeName, faction = Game.system.faction.name})
+
+	ui.withStyleVars({WindowPadding = widgetSizes.windowPadding, ItemSpacing = widgetSizes.itemSpacing},
+		function()
+			local infoColumnWidth = ui.getContentRegion().x
+				- widgetSizes.faceSize.x - widgetSizes.windowPadding.x*3
+
+			ui.child("CrimeStats", Vector2(infoColumnWidth, 0), {"AlwaysUseWindowPadding"},
+				function()
+					ui.withFont(pionillium.large.name, pionillium.large.size,
+						function () ui.text(intro_txt) end)
+
+					-- 1. If outstanding fines, show list & offer to pay
+					outstanding_fines()
+
+					ui.dummy(widgetSizes.dummySpaceMedium)
+
+					-- 2 If old payed fines, show grayd out list
+					crime_record() end)
+			ui.sameLine()
+			ui.child("PoliceOfficer", Vector2(0, 0), {"AlwaysUseWindowPadding", "NoScrollbar"},
+				function ()
+					if(face ~= nil) then
+						face:Draw(widgetSizes.faceSize)
+			end end)
+	end)
+end
+
 
 StationView:registerView({
 	id = "police",
 	name = l.POLICE,
 	icon = ui.theme.icons.shield_other,
-	showView = false,
-	draw = function() end,
-	refresh = function() end,
+	showView = true,
+	draw = function()
+		ui.child("StationPolice", Vector2(0, ui.getContentRegion().y
+			- StationView.style.height), {}, drawPolice)
+		StationView:shipSummary()
+	end,
+	refresh = function()
+		local station = Game.player:GetDockedWith()
+		if (station) then
+			if (stationSeed ~= station.seed) then
+				stationSeed = station.seed
+				local rand = Rand.New(station.seed .. "-police")
+				face = InfoFace.New(Character.New({ title = l.CONSTABLE, armour=true }, rand),
+					{windowPadding = widgetSizes.windowPadding, itemSpacing = widgetSizes.itemSpacing})
+			end
+		end
+	end,
 })
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -1347,6 +1347,10 @@
     "description": "Player owing money to crew members",
     "message": "Owed"
   },
+  "PAY": {
+    "description": "Pay for a transaction of services, e.g. for paying fines, (or repairs)",
+    "message": "Pay"
+  },
   "PAY_FINE_OF_N": {
     "description": "",
     "message": "Pay fine of {amount}"
@@ -1686,6 +1690,10 @@
   "THE_SHIP_IS_UNDER_STATION_CONTROL_COMMANDER": {
     "description": "",
     "message": "The ship is under station control, Commander."
+  },
+  "THIS_IS_FACTION_POLICE": {
+    "description": "Greeting message in police station. 'faction' is one of (English, non-translatable) strings: 'Solar Federation', 'Commonwealth of Independent Worlds', or 'Haber Corporation', and 'faction_police' is one of: 'SolFed Police Force', 'Confederal Police' or, 'Haber Enforcement Division'",
+    "message": "This is the {faction_police} of the {faction}"
   },
   "THREE_KPC_ARM": {
     "description": "Arm of the Milky Way galaxy",
--- a/data/pigui/modules/station-view/01-lobby.lua
+++ b/data/pigui/modules/station-view/01-lobby.lua
@@ -50,7 +50,7 @@ local hydrogenIcon = PiImage.New("icons/goods/Hydrogen.png")
 local hyperdriveIcon = PiImage.New("icons/goods/Hydrogen.png")
 
 local popupMsg = ''
-local popup = ModalWindow.New('lobbyPopup', function()
+local popup = ModalWindow.New('lobbyPopup', function(self)
 	ui.text(popupMsg)
 	ui.dummy(Vector2((ui.getContentRegion().x - 100*rescaleVector.x) / 2, 0))
 	ui.sameLine()


data/pigui/modules/station-view/07-police.lua
```diff
@@ -1,17 +1,157 @@
...
+
+local rescaleVector = ui.rescaleUI(Vector2(1, 1), Vector2(1600, 900), true)
+
+local pionillium = ui.fonts.pionillium
+local orbiteer = ui.fonts.orbiteer
+local colors = ui.theme.colors
+local face = nil
+local stationSeed = false
+
+local popup = ModalWindow.New('policePopup', function()
```
vakhoir: > When trying to pay fines higher than what the player has, the modal popup window saying "not enough money" crashes the game when clicking "OK". I'm not sure what the correct way of implementing this is. Pointers are most welcome, ping @vakhoir

Sorry, this is my fault, I must have tested the modal window changes with a view where a window  reference was used rather than `self:close()`, this can be fixed by just adding `self` to the handler parameters. I'll go over the other views to fix this.

```suggestion
local popup = ModalWindow.New('policePopup', function(self)
```

data/pigui/modules/station-view/07-police.lua
```diff
@@ -1,17 +1,157 @@
...
+
 StationView:registerView({
 	id = "police",
 	name = l.POLICE,
 	icon = ui.theme.icons.shield_other,
-	showView = false,
-	draw = function() end,
-	refresh = function() end,
+	showView = true,
+	draw = drawPolice,
```
vakhoir: I usually put the station screens in a wrapper, then you can use the `StationView:shipSummary()` to render the cash/capacity/cabins/legal satus infos at the bottom.

```suggestion
	draw = 
		ui.child("StationPolice", Vector2(0, ui.getContentRegion().y - StationView.style.height), {}, drawPolice)

		StationView:shipSummary()
	end,
```

data/pigui/modules/station-view/07-police.lua
```diff
@@ -1,17 +1,157 @@
...
+
+
+local function drawPolice()
+
+	-- Todo: add to language translation system
+	local intro_txt = "This is the ".. Game.system.faction.policeName ..
+		" of the "  .. Game.system.faction.name .. "."
+
+	ui.withFont(pionillium.large.name, pionillium.large.size,
+				function () ui.text(intro_txt) end)
```
vakhoir: I'd move this inside the "CrimeStats" child window so window padding is applied

vakhoir:
> widgetSizes table is a copy-paste from lobby, that I haven't put that much thought into. Ideally I should rename it (and think through what it's doing)

I don't think you need to rename it, I use it with than name with a lot of views. What I'm not sure is if using`ceil` on all the vectors is necessary. I think I tried it out to align the widgets better but I think it might not be needed.

In any case it's just a local variable for storing widget sizes, for consistency, and so they don't need to get recalculated every frame, etc.

> I'm not sure how pioneer-style compliant I am with colors, etc. Please point out if it's too uncceptably deviant.

_Maybe_ we could use a slightly lighter shade of grey for the criminal record, but otherwise looks good.

data/pigui/modules/station-view/07-police.lua
impaktor: thanks, that fixed it. The missing `self` can be found in Lobby (and maybe others as well)

data/pigui/modules/station-view/07-police.lua
impaktor: Fixed it, thanks!

data/pigui/modules/station-view/07-police.lua
impaktor: Yes, that looks way better, thanks.

impaktor:
Thanks for review!

I've:

- Fixed popup-bug

- Made gray tex a bit brighter / visible (from `(76,76,76)` to `(100,100,100)`, where 255=white)

- Removed "ceil" call, seems to work fine. My goal with the police tab was to do as little copy-paste as possible from previous pigui screens, so I understand all that's going on in the code, so this is good for me.

- I've put top/first greeting sentence into translation system @nozmajner: do we want that commit? Problem here is that _faction names_ and _faction police name_ is not translatable, as they're pulled from the factions/<faction>.json stuff. Still as separate commit, in case we don't want it.

## Example:
2020-02-08-140745_1600x900_scrot

Web-eWorks:
@impaktor looking good, I like what you've done with the previous criminal record! A couple of tips / suggestions for you:
- Increase the padding between the offense count and the name of the offense. Right now e.g. Murder is offset, and it spoils the even, table-like look of the display. You could do it like so (with variable names as appropriate):
```lua
local countPlusPadding = 40 -- or however much space you want for the count and padding
ui.sameLine(countPlusPadding - ui.calcTextWidth(offenseCountText))
ui.text(offenseName)
```
- Add an icon to past criminal offenses to indicate that it has been paid off. This could be a checkmark, a money symbol, whatever, but it should indicate that your criminal record has been dealt with.
- There should be a bit more padding between the "this is the police" text and the "Outstanding fines" heading. (Speaking of, "fines" in that heading should be capitalized!)
- The "Pay" button and the money count is misaligned, but that's not something that's easy to fix as a novice (I don't even know the trick to that one, and I definitely don't have the time to go digging right now).

Otherwise, this looks pretty good! I'll try to go over the code tomorrow and see about merging.

As an aside @vakhoir can you do something about the need to load all of the newUI screens to have proper tab ordering? From what I'm looking at, if I start deleting *some* but not all of the newUI station view screens it'll mess up the pigui-to-newUI hosting because we rely on tab indicies to properly display the correct newUI screen.

impaktor:
thanks for review and pointers, @Web-eWorks!

- Regarding padding, I intentionally gave my self > 9 offences to see what it looked like, and if I would get complaints :smiling_imp:. But there's no `ui.calcTextWidth` so I tried `ui.sameLine(countPlusPadding - ui.calcTextSize(v.count).x)`, but that led to mayhem, and I'm too low energy to figure that out at the moment. Would it not be more natural to use the `ui.columns`, assuming that aligns width of the columns properly to begin with?

- I think I'll ignore adding icon for now, if that's OK, as I'm not sure we have a fitting one, and this screen will be re-worked quite a bit soonish, by me, and I cant see it looking good, in my minds eye.

- Added more padding between line 1 and 2, as suggested, but will not capitalize "... fines" or "... record", as those are strings already in the translation system, plus it's simply a [matter of style](https://english.stackexchange.com/questions/76479/where-can-i-find-a-list-of-capitalisation-rules-for-pure-british-writing), although I know US English ("Chicago Manual") likes to capitalize every word in a heading.

- I've moved the "Pay" button to its own line, thus the difficult alignment issue is "fixed".

- Fixed indentation for line-broken argument lists and functions (my Lua-indentation mode doesn't do as we have agreed it should be done)

impaktor:
Well, using `ui.columns` to build a table, gives me:
2020-02-09-184834_1600x900_scrot

I'm out of energy for this weekend now.

data/pigui/modules/station-view/07-police.lua
```diff
@@ -1,17 +1,154 @@
...
+	if #utils.build_array(pairs(crimes)) > 0 then
+
+		-- headline
+		ui.withFont(orbiteer.xlarge.name, orbiteer.xlarge.size,
+					function() ui.text(l.OUTSTANDING_FINES) end)
+		-- wrap list in medlarge font
+		ui.withFont(pionillium.medlarge.name, pionillium.medlarge.size,
+					function()
+						for k,v in pairs(crimes) do
+							ui.text(v.count.."\t"..Legal.CrimeType[k].name)
```
Web-eWorks:
```suggestion
							ui.text(v.count)
							-- TODO: scale this spacing value with ui.rescaleUI
							ui.sameLine(40 - ui.calcTextSize(v.count))
							ui.text(Legal.CrimeType[k].name)
```

impaktor:
Thanks for the code; but the way I read the source: `ui.sameLine(pos_x, spacing_w)` it's enough to just give `pos_x`-coordinate on screen where to start the column? Seems to work fine, see screenshot below. (the code you posted inversed the problem, moving the text too much to the left if the number to the left of it was high, probably becuase the width some some variable wasn't scaled correctly, or something).

However, I've only tested this on the resolution I use (1600x900), so it sounds reasonable that I should make use of some `ui.rescaleUI`, alas I don't understand that code. If I send in a number, and a resolution, I'd expect to get something similar, but rescaled, back. But I get something x1000 times bigger.

Well, the code path for `ui.rescaleUI` for my example (sending a `number`) would be:

```lua
ui.rescaleUI = function(val, baseResolution)
	local rescaleVector = Vector2(pigui.screen_width / baseResolution.x, pigui.screen_height / baseResolution.y)
	local rescaleFactor = math.min(rescaleVector.x, rescaleVector.y)
        return val * rescaleFactor
end
```
In summary: At the moment I don't quite understand what/how it is used, although the name of the function sounds like it's something useful.

## So what remains is:
1. Am I wrong in thinking I don't need any `calcTextSize`? The vaiability of one-two characters between different resolutions might not be that big? (famous last words?)
2. Do I need / how do I use `rescaleUI`

## Status:
2020-02-10-141519_3200x900_scrot

vakhoir:
> However, I've only tested this on the resolution I use (1600x900), so it sounds reasonable that I should make use of some `ui.rescaleUI`, alas I don't understand that code. If I send in a number, and a resolution, I'd expect to get something similar, but rescaled, back. But I get something x1000 times bigger.

I might need to rename the parameters, but I can't think of anything better. Or maybe I'll just add a comment explaining how it works. The idea is that you pass the current resolution you're working with, and the size of the widget has at that resolution, tand the function with rescale it to whatever the player's resolution is. The `rescaleToScreenAspect`determines if the shape of the widget should should change with the aspect ratio of the screen (whether what was a square at 4:3 should become more of a rectangle at 16:9).

> Am I wrong in thinking I don't need any calcTextSize? The vaiability of one-two characters between different resolutions might not be that big? (famous last words?)

I think you can skip it, but I'd leave more room than 2 digits. 4 just in case?

Web-eWorks:
@impaktor
> Am I wrong in thinking I don't need any calcTextSize? The vaiability of one-two characters between different resolutions might not be that big? (famous last words?)

You are in fact correct; this one's my bad. For some reason, even after reading the source, I was thinking that `ui.sameLine` took a relative offset instead of an absolute one. There's no need for ui.calcTextSize at all, and from what you're showing this works perfectly fine.

It looks like you've left room for three digits of offenses; although you might want to add just a little more space, it's totally optional and we adjust it later if it ever becomes necessary.

data/pigui/modules/station-view/07-police.lua
impaktor: Resolved as discussed below. Went for the fixed position, increased it from 50 to 55. 

If a player has +999 offences, me thinks we should aggregate illegal discharge offences differently. "I can't imagine anyone needing more than 999 crimes (64 kb)"

impaktor:
OK, pushed final(?) changes, basicaly settled to what we said above, just use `ui.sameLine(55)`.

I also couldn't remove adding a small fix as a second commit, ping @vakhoir (or should I remove it?)

If anyone cares, you might want to check my indentation style, for breaking long lines.

data/pigui/modules/station-view/07-police.lua
```diff
@@ -1,17 +1,157 @@
...
+	end
+end
+
+
+local function outstanding_fines()
+	local crimes, fine = Game.player:GetCrimeOutstanding()
+	if #utils.build_array(pairs(crimes)) > 0 then
+
+		-- headline
+		ui.dummy(Vector2(0, 10))
```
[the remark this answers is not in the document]
Web-eWorks: Yes and yes. Ideally, call `local v = Vector2()` as few times as possible, and when you're updating the vector call `v(x, y)` instead of `v = Vector2(x, y)` to set the values without allocating a new vector object.

Overall gains will be incredibly tiny, considering that the entire pigui call tree runs in 1-2ms or so, but it is "best practice" for whatever that's worth.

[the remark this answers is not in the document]
vakhoir: > and when you're updating the vector call `v(x, y)` instead of `v = Vector2(x, y)`

Oh, I didn't realize that was an option, sweet!

impaktor:
@vakhoir I've fixed in accordance to your suggestions

I've pushed it as a separate commit, so you can check that it looks OK. 

I'll squash it and rebase when you give thumbs up, so this can be merged afterwards.

Thanks for suggestions

PR #4791, opened 2020-02-08
Move ShipRepair screen to pigui

impaktor:
## Intro

OK, so I've done the ship repair shop as well. The original had this silly option, to either repair 1 percent-point, or all the damage. I replaced it with a slider.

I'm posting a screenshot with two different styles of hull integrity progress bars, the default ImGUI (yellow), and the pigui gauge (the "pioneery"), please tell me what you think, @nozmajner.

## Further future improvements
- I'm thinking if it's possible to make certain stations be authorized service shops for ships from certain manufacturers. Would give a discount on repairs. (HA! compare that to real "licensed" repair shops, they charge x2 the price!)

- The original Frontier you did both hull and engine maintenance in the ship repair shop, but engine "health" is in the _BreakdownService_ module, so not available to the station screens. Either way, all equipment will have their own health/damage system in the future, so the _BreakdownService_ module will be kicked out/moved at some point, but that's for another day.

## NewUI: What it used to look like 
2020-02-08-201724_1600x900_scrot

## PiGUI: What it looks like now
2020-02-08-201927_1600x900_scrot

## Frontier: What it looked like in the original to our clone
(screenshot from Commodore Amiga version of Frontier)
2020-02-08-194332_1600x900_scrot

### DIFF
--- a/data/pigui/modules/station-view/06-shipRepairs.lua
+++ b/data/pigui/modules/station-view/06-shipRepairs.lua
@@ -3,15 +3,155 @@
 
 local ui = import 'pigui/pigui.lua'
 local StationView = import 'pigui/views/station-view'
+local ShipDef = import("ShipDef")
+local Game = import "Game"
+local Rand = import "Rand"
+local InfoFace = import 'ui/PiguiFace'
+local Format = import "Format"
+local Character = import "Character"
+local ModalWindow = import 'pigui/libs/modal-win.lua'
 local Lang = import 'Lang'
-
 local l = Lang.GetResource("ui-core")
+
+local rescaleVector = ui.rescaleUI(Vector2(1, 1), Vector2(1600, 900), true)
+
+local colors = ui.theme.colors
+local icons = ui.theme.icons
+
+local pionillium = ui.fonts.pionillium
+local orbiteer = ui.fonts.orbiteer
+
+local face = nil
+local stationSeed = false
+local damageToRepair = nil
+local repair_cost = 0
+
+local widgetSizes = ui.rescaleUI({
+	itemSpacing = Vector2(4, 9),
+	windowPadding = Vector2(14, 14),
+	faceSize = Vector2(586,565),
+	buttonSize = Vector2(100, 0),
+	verticalDummy = Vector2(0, 50),
+	gaugeWidth = 600,
+}, Vector2(1600, 900))
+
+
+local popup = ModalWindow.New('ChiefMechanicPopup', function(self)
+	ui.text(l.YOU_DONT_HAVE_ENOUGH_MONEY_FOR_THAT_OPTION)
+	ui.dummy(Vector2((ui.getContentRegion().x - 100*rescaleVector.x) / 2, 0))
+	ui.sameLine()
+	if ui.button(l.OK, Vector2(100*rescaleVector.x, 0)) then
+		self:close()
+	end
+end)
+
+
+local tryRepair = function (damage, price)
+	if Game.player:GetMoney() >= price then
+		Game.player:AddMoney(-price)
+		Game.player:SetHullPercent(Game.player:GetHullPercent() + damage)
+	else
+		popup:open()
+	end
+end
+
+
+local getRepairCost = function (percent, shipDef)
+	-- repairing 1% hull damage costs 0.1% of ship price
+	shipDef = shipDef or ShipDef[Game.player.shipId]
+	return math.ceil(shipDef.basePrice * (percent * 0.1)) * 0.01
+end
+
+
+local getRepairMessage = function (damage, price)
+	return string.interp(
+		l.REPAIR_X_HULL_DAMAGE_FOR_X, {
+			damage = string.format('%.1f', damage),
+			price = Format.Money(price)
+		})
+end
+
+
+function round(num, numDecimalPlaces)
+	local mult = 10^(numDecimalPlaces or 0)
+	return math.floor(num * mult + 0.5) / mult
+end
+
+
+local function drawShipRepair()
+	local hullPercent = round(Game.player:GetHullPercent())
+	local damage = 100 - hullPercent
+
+	local intro = string.interp(l.YOUR_HULL_IS_AT_X_INTEGRITY, {value = string.format('%.1f', hullPercent)})
+
+	ui.withStyleVars({WindowPadding = widgetSizes.windowPadding,
+		ItemSpacing = widgetSizes.itemSpacing}, function ()
+			local infoColumnWidth = ui.getContentRegion().x - widgetSizes.faceSize.x - widgetSizes.windowPadding.x*3
+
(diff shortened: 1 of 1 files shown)


nozmajner:
I think the pigui one would be better. It might be a bit more cluttered, but it is consiszent with the HUD that way.

Web-eWorks:
@impaktor a couple of tips for you:
- Sliders allow you to specify the format string, so you could replace that `12` with a `12% Hull` text fairly easily. (AFAIK it uses `snprintf()` internally, so you can `man snprintf` to see all the options you have available)
- When the player doesn't have enough money to pay for the selected level of hull repair, turn (at least) part of the "Repair X hull damage for $Y" red to indicate that it's more expensive than the player can currently afford. Remember `ui.sameLine()`; it lets you append more text to the current line rather than starting a new one with every `ui.text()` call.
- The "Pay" button should be wrapped in a `ui.withFont()` call that uses a larger font; it's a little small right now.

I'm not a fan of either of those gauges; neither of them fit the look of the screen or indicate useful information like how many tons of hull you have remaining, the type of hull material (which is moot right now, because we don't have e.g. military composite hulls), or anything beyond the current hull percentage.

*However*, because this screen is on it's way out the door anyways, I don't mind a few rough edges in the styling and presentation, so use whatever gauge is the best in your eyes.

Regarding the discounts, I like the idea of a faction-owned shipyard being able to repair their faction's ships cheaper - mostly because they have the required parts easily to hand and are (probably) getting them at a slight discount from the ship manufacturer. In reality, yeah, that's not a thing; dealers and licensed repair shops tend to be more expensive and less honest than a (good!) third party repair shop, but that's neither here nor there.

impaktor:
- I've added a "%" symbol to the slider, and fixed font of pay button.
- For your second point:
  - I don't understand what it is you think should be on the same line as what?
  - I'll leave it as it is, since no where else do we show the player that he can't afford it, at least not in ship store or equipment market. (Commodity shows yellow text "insufficient funds").

__Note:__ I'll rebase this to master after #4790 is merged, which adds a `l.PAY` string, that I can use here.

2020-02-09-094250_1600x900_scrot

impaktor:
OK, I've fixed this up now, there was a bug, due to "get percent hull" function being floating point and applying many small incremental integer repairs got som very small `0.0000001` round thingy, very vague description, but I added a `round()` function to:
```lua
local hullPercent = round(Game.player:GetHullPercent())
```

which fixed it. 
__Q__ do we have any round function in lua? Couldn't find any in `utils.lua`

## beware:
I have a `dummy(Vector2(0, 50))` which I should move to `widgetSizes`, as I did yesterday with #4790, but I'm not sure I want that dummy there at all, or might want to do something smarter. Basically: how do I place the hull percent bar closer to the footer, like the NewUI currently is? Get the (relative?) distance from the bottom? Knowing the height of the station foot / `StationView:shipSummary()` stuff? 

Pardon a vague post.

vakhoir:
I haven't looked at the code yet, but just from the screenshot, wouldn't it look better if the hull integrity gauge and the repair slider were to same size?

Web-eWorks:
@vakhoir that's because the hull integrity gauge is using the `ui.gauge` method. Ideally we want to redesign gauge drawing entirely (so it's not completely static and making bad assumptions), but that's a topic for another PR methinks.

impaktor:
Yeah, I know, it looks a bit wonky. If you have suggestions on how to fix, I'm all ears. I plan to do another PR touching this ship repair screen, within a few weeks (ship manufacture specific repair), so maybe I have stronger pigui-fu then. 

I just removed that hard-coded `Vector2` in the `ui.dummy` for vertical space, and moved it into `widgetSizes`, as I suspect we're not going to drill down on that any more, so unless anyone wants to iterate on this more, this should be good to merge now.

data/pigui/modules/station-view/06-shipRepairs.lua
```diff
@@ -3,15 +3,153 @@
...
+local stationSeed = false
+local damageToRepair = nil
+local repair_cost = 0
+
+local widgetSizes = ui.rescaleUI({
+	itemSpacing = Vector2(4, 9),
+	windowPadding = Vector2(14, 14),
+	faceSize = Vector2(586,565),
+	buttonSize = Vector2(100, 0),
+	verticalDummy = Vector2(0, 50),
```
vakhoir: Step 1: Add gauge size to widget sizes:

```suggestion
	verticalDummy = Vector2(0, 50),
	gaugeWidth = 600,
```

data/pigui/modules/station-view/06-shipRepairs.lua
```diff
@@ -3,15 +3,153 @@
...
+			local infoColumnWidth = ui.getContentRegion().x - widgetSizes.faceSize.x - widgetSizes.windowPadding.x*3
+
+			ui.child("ShipStatus", Vector2(infoColumnWidth, 0), {"AlwaysUseWindowPadding"}, function ()
+				if hullPercent > 99.9 then
+					ui.withFont(pionillium.medlarge.name, pionillium.medlarge.size,
+						function () ui.text(l.SHIP_IS_ALREADY_FULLY_REPAIRED) end)
+				else
+					ui.withFont(pionillium.medlarge.name, pionillium.medlarge.size, function ()
+						ui.text(getRepairMessage(damageToRepair, repair_cost))
+						damageToRepair = ui.sliderInt("", damageToRepair, 1, damage, "%d%%")
```
vakhoir: Step 2: call pushItemWidth to set the sliderSize

```suggestion
						ui.pushItemWidth(widgetSizes.gaugeWidth)
						damageToRepair = ui.sliderInt("", damageToRepair, 1, damage, "%d%%")
```

data/pigui/modules/station-view/06-shipRepairs.lua
```diff
@@ -3,15 +3,153 @@
...
+					end)
+				end
+
+				ui.dummy(widgetSizes.verticalDummy)
+
+				local gaugePos = ui.getCursorScreenPos()
+				gaugePos.y = gaugePos.y + 50
+
+				local text = string.format(hullPercent)
+				ui.gauge(gaugePos, hullPercent, '', l.HULL_INTEGRITY, 0, 100, icons.hull, colors.gaugeEquipmentMarket, l.HUD_HULL_STRENGTH)
```
vakhoir: Step 3: Pass the width to ui.gauge():

```suggestion
				ui.gauge(gaugePos, hullPercent, '', l.HULL_INTEGRITY, 0, 100, icons.hull, colors.gaugeEquipmentMarket, l.HUD_HULL_STRENGTH, widgetSizes.gaugeWidth)
```

impaktor:
@vakhoir thanks! That did the trick! I've squashed and rebased.

2020-02-16-165602_1600x900_scrot

Web-eWorks:
Looks good to me! Merging this in.

PR #4792, opened 2020-02-10
PiGui docstrings and fix to ui.columns

impaktor:
OK, I'm throwing this on the wall to see if it sticks.

## Default argument to columns
Looking at pigui code, they finish their column-block with just a single `Columns(1)`
```c++
ImGui::Columns(3, NULL, false);
static bool selected[16] = {};
for (int i = 0; i < 16; i++)
{
    char label[32]; sprintf(label, "Item %d", i);
    if (ImGui::Selectable(label, &selected[i])) {}
    ImGui::NextColumn();
}
ImGui::Columns(1);    // <--- THIS
```

So made second argument optional, defaulting to `""` (is this correct? It's default is `NULL` in ImGui), so we can do:
```lua
ui.columns(2, "Mycolumntest", true)
ui.text("row 1, col 1")
ui.nextColumn()
ui.text("row 1 1, col 2")
ui.nextColumn()
ui.text("row 2, col 1")
ui.columns(1)   -- <--- THIS
```

## Documentation strings
Also, added suggestion of how documentation could look like. I imagine we want to discuss what it should look like, but idea is we add documentation to where the function is defined, `src/LuaPiGui.cpp`, but use the function names as defined in `data/pigui/pigui.lua`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,6 +258,14 @@ add_executable(savegamedump WIN32
 	src/PngWriter.cpp
 )
 
+find_program(NATURALDOCS NAMES naturaldocs)
+if (NATURALDOCS)
+	add_custom_target(codedoc
+		${CMAKE_COMMAND} -E make_directory codedoc
+		COMMAND naturaldocs -i src/ -i data/libs/ -xi src/data/ -o HTML codedoc/ -p nd/ -do -ro -s Default Local
+	WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+endif()
+
 list(APPEND pioneerLibs
 	pioneerLib
 	${ASSIMP_LIBRARIES}
--- a/src/LuaPiGui.cpp
+++ b/src/LuaPiGui.cpp
@@ -398,12 +398,17 @@ static int l_pigui_lineOnClock(lua_State *l)
 /*
  * Interface: PiGui
  *
- * Various functions for the imgui UI. Do *not* use these directly, use the interface that import('pigui') provides.
+ * Various functions for the imgui UI. The below documentation are for the method names as implemented in the 'pigui' module
+ *
+ * Example:
+ *
+ * > local ui = import 'pigui/pigui.lua'
+ *
  */
 
 /* ****************************** Lua imgui functions ****************************** */
 /*
- * Function: Begin
+ * Function: begin
  *
  * Availability:
  *
@@ -423,6 +428,28 @@ static int l_pigui_begin(lua_State *l)
 	return 1;
 }
 
+/*
+ * Function: columns
+ *
+ * Initiates drawing of a table with n columns
+ *
+ * > ui.columns(n_columns, label, border)
+ *
+ * Example:
+ *
+ * > ui.columns(n_columns, label, has_border)
+ * > ui.text("row 1, column 1")
+ * > ui.nextColumn()
+ * > ui.text("row 1, column 2")
+ * > ui.columns(1)
+ *
+ * Parameters:
+ *
+ *   n_columns - integer, the number of columns,
+ *   label - optional string, the label of the object
+ *   border - optional bool, draws a vertical border between columns. Defaults to false.
+ *
+ */
 static int l_pigui_columns(lua_State *l)
 {
 	PROFILE_SCOPED()
@@ -466,6 +493,24 @@ static int l_pigui_get_scroll_y(lua_State *l)
 	return 1;
 }
 
+/*
+ * Function: progressBar
+ *
+ * Make a progress bar widget
+ *
+ * > ui.progressBar(fraction, size, overlay)
+ *
+ * Example:
+ *
+ * > ui.progressBar(0.3, Vector2(0.0, 0.0), "%")
+ *
+ * Parameters:
+ *
+ *   fraction - float, text on button
+ *   size - Vector2, defining size
+ *   overlay - string, text to overlay
+ *
+ */
 static int l_pigui_progress_bar(lua_State *l)
 {
 	PROFILE_SCOPED()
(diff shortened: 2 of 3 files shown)


src/LuaPiGui.cpp
```diff
@@ -423,11 +423,30 @@ static int l_pigui_begin(lua_State *l)
...
+ * > ui.text("row 1, column 1")
+ * > ui.nextColumn()
+ * > ui.text("row 1, column 2")
+ * > ui.columns(1)
+ *
+ * Parameters:
+ *
+ *   n_columns - integer, the number of columns,
+ *   label - optional string, the label of the object
+ *   border - optional, bool, draw vertical column border or not (default)
```
Web-eWorks: @impaktor API documentation should use strong, clear, and concise language. Please prefer something like the below to document this parameter:
```suggestion
 *   border - optional bool, draws a vertical border between columns. Defaults to false.
```

src/LuaPiGui.cpp
```diff
@@ -476,6 +495,19 @@ static int l_pigui_progress_bar(lua_State *l)
...
+ * Function: nextColumn
+ *
+ * Moves drawing position to next column
+ *
+ * > local ui = require 'pigui'
+ * > ui.columns(n_columns, label, has_border)
+ * > ui.text("row 1, column 1")
+ * > ui.nextColumn()
+ * > ui.text("row 1, column 2")
+ * > ui.columns(1)
```
Web-eWorks: See [the NaturalDocs syntax reference](https://www.naturaldocs.org/reference/formatting/index.html#code_and_text_diagrams) for more information on how to format text blocks. Ideally we want to prefer the `--- Lua ---` format instead of prefixing each line with a `> ` symbol.

src/LuaPiGui.cpp
```diff
@@ -423,11 +423,30 @@ static int l_pigui_begin(lua_State *l)
 	return 1;
 }
 
+/*
```
Web-eWorks: You should probably add an `/* Interface: ui */` block here at the start of the documentation for pass-through functions, as otherwise all of these functions will be grouped under the `PiGui` interface.

src/LuaPiGui.cpp
impaktor: Oh, didn't know that. I don't think we've used it anywhere else, which of course isn't an excuse to not start now.

impaktor:
Seems like we lost ability to build naturaldocs in the move to CMake?

Web-eWorks:
@impaktor that's because NaturalDocs 3 didn't build our docs under autotools anyways (due to command line argument changes) so I neglected it. It's on the docket somewhere under "I'll get to it when I have time".

impaktor:
Added a commit to bring back `codedoc` to the buld process. As might be evident from the IRC backlog, I bloodly lost my mind trying to understand CMake, so `make codedoc` isn't possible, instead it builds always, ... because reasons. Suggestions welcome to what I'm missing.

And yes, I'm opting to get v.1.x building as that's what we were using, higher versions seem to pull down 200 MB of dependencies.

CMakeLists.txt
```diff
@@ -258,6 +258,13 @@ add_executable(savegamedump WIN32
 	src/PngWriter.cpp
 )
 
+add_custom_target(codedoc ALL)
+add_custom_command(TARGET codedoc POST_BUILD
+	COMMENT "Building naturaldoc for LUA API"
+	COMMAND bash -c "mkdir -p codedoc"
+	COMMAND bash -c "naturaldocs -i src/ -i data/libs/ -xi src/data/ -o HTML codedoc/ -p nd/ -do -ro -s Default Local"
+	WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+
```
laarmen: Thoroughly untested, but you might want to simply add a check for naturaldocs :
```cmake
find_program(NATURALDOCS NAMES naturaldocs)
if (NATURALDOCS)
add_custom_target(...)
endif()
```

impaktor: Added. "works on my system". Thanks.

(Still interested in how to do a `make codedoc`)

laarmen:
Add a PROPERTIES EXCLUDE_FROM_ALL perhaps ?

laarmen:
Hem... This time with proper syntax highlighting as well as a value for the property :)

```cmake
set_target_properties(codedoc PROPERTIES EXCLUDE_FROM_ALL ON)
```

impaktor:
I assume that needs to go after `add_custom_target(codedoc ALL)`, and I should remove the `ALL`. Seems like it's not building it at all now.

laarmen:
OK, we seem to have lost the threading there.
So, reading the doc, I don't think you *need* the `add_custom_target`, `add_custom_command` should take care of things by itself.

impaktor:
Welcome to my nightmare!

```
CMake Error at CMakeLists.txt:263 (add_custom_command):
  No TARGET 'codedoc' has been created in this directory.
-- Configuring incomplete, errors occurred!
```

impaktor:
OK, CMake issue resolved, thanks to @laarmen and @pcercuei. Updated `COMPILING.txt` documentation for how to build the docs.

src/LuaPiGui.cpp
impaktor: In version 1.x this is not supported. We haven't done it anywhere else, so to be consistent, I'm not doing it here either.

src/LuaPiGui.cpp
impaktor: I tried with both the above suggested, and 
```
/* 
 * Interface: ui
 */
```

But doesn't insert any such section.

impaktor:
OK; I think I'm done playing with this now. I've added doc strings to most of the functions I've learned and used. I'll add more as I learn, later down the line. You can see what current PR looks like here:
https://impaktor.github.io/pioneer/files/LuaPiGui-cpp.html

src/LuaPiGui.cpp
Web-eWorks: More NaturalDocs v2 functionality then, I suppose.

Web-eWorks:
@impaktor there's a slight issue with clang-format regarding some trailing whitespace; I'm going to ignore that and merge as-is.

PR #4794, opened 2020-02-15
Some fixes to the handling of in-space bodies in flight-ui

cwyss:
I noticed a few issues that remained after #4764:

- In certain (rare) situations it is possible that two clusters of bodies are displayed very close to one another, closer than the cluster size. The problem is not the algorithm which groups the bodies into clusters, but the fact that after creating the clusters, when the bodies of each cluster are sorted according to their importance, the most important body becomes the new main body. That is, the clusters remain unchanged, but the main body which is used to put the body icon on screen might not be the centre of the cluster anymore.

  As an example, in the following savefile Ceres and Jupiter are very close together on screen and cause the described problem: [debug_grouped.gz](https://github.com/pioneerspacesim/pioneer/files/4209073/debug_grouped.gz)

  The fix I implemented is to make sure that already when grouping the bodies into clusters a new body immediately becomes the new main body if it is more important than the old one.

- The old cluster area was rectangular. This makes the clusters depend on the viewing angle: If you roll your ship around a cluster, so that a body's relative position to a cluster changes e.g. from left of the centre to upper-left of the centre, the body may be absorbed into the cluster and then pop out again once it is directly above the centre. 

  I found this behaviour odd, and so I changed the cluster area to a circle.

- When using setspeed mode with a setspeed target different than the frame of reference (i.e. ctrl-selecting that target), the setspeed target is reset to the frame of reference once you select any other target. This is rather unpleasent when you want to use setspeed to accelerate towards some planet, and then want to set a different nav target - temporarily only - to check e.g. your relative speed against it.

  I implemented the following new behaviour:
    - Don't clear the setspeed target when in setspeed mode and a
      different nav target is selected.
    
    - Change or clear the setspeed target in setspeed mode when
      ctrl-selecting a new nav target or ctrl-deselecting the old
      setspeed target, respectively.
    
    - When not in setspeed mode, implicitly clear the setspeed target on
      any target change.

  The last one is to keep you from surprises in case you forget that some setspeed target was ctrl-selected and you are now on some different body and turn on setspeed mode to hover, e.g. Since you probably selected some other target in the meantime, the old setspeed target was implicitly cleared and setspeed is thus again relative to body over which you want to hover.

- The last commit is not really related to in-space bodies but only a small lua code clean-up suggested in https://github.com/pioneerspacesim/pioneer/pull/4758#discussion_r373876072

### DIFF
--- a/data/pigui/views/game.lua
+++ b/data/pigui/views/game.lua
@@ -139,12 +139,12 @@ local function displayOnScreenObjects()
 	local should_show_label = ui.shouldShowLabels()
 	local iconsize = Vector2(18 , 18)
 	local label_offset = 14 -- enough so that the target rectangle fits
-	local collapse = iconsize -- size of clusters to be collapsed into single bodies
-	local click_radius = collapse:length() * 0.5
+	local cluster_size = iconsize.x -- size of clusters to be collapsed into single bodies
+	local click_radius = cluster_size * 0.5
 	-- make click_radius sufficiently smaller than the cluster size
 	-- to prevent overlap of selection regions
 
-	local bodies_grouped = ui.getProjectedBodiesGrouped(collapse, IN_SPACE_INDICATOR_SHIP_MAX_DISTANCE)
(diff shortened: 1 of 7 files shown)


cwyss:
Rebased to resolve merge conflicts due to advanced master branch

PR #4775, opened 2020-01-23
Bulletin Board view

vakhoir:
Ho boy, this is going to be a big one - every ad has it's own UI code. This is just a WIP PR if anyone wants to keep track of the progress, and give feedback.

In the mean time I'm refactoring the the Market widget a bit. I already felt I'm pushing it with the Ship Market, because it wasn't using half the handlers the other markets were, it just needed a table with item highlighting, so out it goes into it's own widget.

The Ship Market now just uses a new Table widget, and there's a basic Bulletin Board that disaplays the ads, but not much more.

PR #4796, opened 2020-02-19
Make police non-persistent in Goodstrader

impaktor:
Idea: one should be able to recognize the trustworthy black market merchants,
by them being persistent to each station. So you learn to recognize the name
of the market, e.g. "Honest Jim".

However, if _both_ police and the _true_ black market merchant are persistent,
you will recognize the name of both, so now you need to remember _which_ was
the fake one.

I much prefer the way it was in Frontier: the undercover police had a new name
every time you docked, so when returning to a station you would only recognize
one of the trader-names.

### DIFF
--- a/data/modules/GoodsTrader/GoodsTrader.lua
+++ b/data/modules/GoodsTrader/GoodsTrader.lua
@@ -117,7 +117,10 @@ local onCreateBB = function (station)
 	end
 	if not has_illegal_goods then return end
 
+	-- Real trader is persistant for each station, police is random every time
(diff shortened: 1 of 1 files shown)


impaktor:
I assume appveyor failing is unrelated to this PR? 

Either way, I suggest wating to merge this until #4775 is in, then I'll rebase on top of that, as I suspect we might get a merge conflict, since we edit almost the same lines.

(Also, this PR will keep the same "true" goodstraders in _most_ stations, only where a police trap is spawned before the "honest" one, will it change the name of the trader, I think).

PR #4813, opened 2020-02-24
Draw performance information with ImGui, new performance counter backend

Web-eWorks:
Screenshot from 2020-02-24 16-48-07

I've moved `src/PiGui.*` into the appropriate source folder, and improved the way we handle calling `PiGui::EndFrame()`. This has opened up the possibility to render ImGui widgets from more than one Lua handler, and let C++ render widgets like the performance counter.

I've also moved the block of `snprintf()`-based performance stats to its own class responsible for rendering performance information, and written a generic, `std::atomic`-based performance counter class to better facilitate tracking important information like drawcalls and memory usage. That work isn't quite finished; I need to figure out a good API to access a central instance of the class, start tracking more information, and add automatic stat discovery to `PerfInfo.cpp`, as currently we're hardcoding which stats we want to display.

- Explicitly call PiGui::Render at the end of the frame
- PiGui::Render renamed to RunHandler, now only dispatches to lua handlers
- PiGui::RenderImGui renamed to Render(), now calls EndFrame()
- Rewrote the Ctrl+i performance display using ImGui
- Added a new generic performance counter implementation using atomics
- Added physics- and frame-time display to performance info
- Rewrote Graphics::Stats to use the new PerfStats impl internally
- Added Profiler::Clock, a `std::chrono::steady_clock`-based timer that can be converted into an actual time without needing to know the CPU bus frequency (the pitfall of `rdtsc`).

@fluffyfreak I've provided a fairly foolproof implementation of GetMemoryInfo() for Linux systems; if you could replace that `TODO` in `PerfInfo.cpp` with a windows implementation that returns the used and peak used memory information for the whole process, that'd be great.


impaktor:
Looks very cool! The physicist in me wonders what the min/max of x/y axis are in those graphs...

fluffyfreak:
Nice, I'll try to take a look today when I'm not working

Web-eWorks:
Separate PR would probably be the best, I think. This one is throwing errors for some reason; looks like there's an extraneous move construction in travis' version of the C++ STL. I'll have to track that down in a bit.

fluffyfreak:
Just trying to test it on Windows, how do I bring up the new perf window? What key combo etc?

fluffyfreak:
Ok, I made a start on the Windows memory stuff, just need to find a way to tidy it up and squirrel it away in the Win32 stuff so it's tidier, but that can wait until all of this is merged in.

Web-eWorks:
@impaktor
> Looks very cool! The physicist in me wonders what the min/max of x/y axis are in those graphs...

I would display that if I could; but I'm limited to ImGui's PlotLines widget which doesn't give me much in the way of rendering over top. Something for the future, I suppose...

To answer your question, frame time goes from 0 to 33ms, and UPS (aka physics time) goes from 0 to 10ms. Values higher than that are clamped to prevent them going out of range, though you can hit the pause button and manually inspect it with the mouse cursor.

...I should probably track the amount of time we spend in rendering updates as well, but that's going to require a histogram that displays more than one line at a time.

Web-eWorks:
Figured out why Windows and OSX builds were failing; apparently `contrib/lua/lua.h` was being found before `src/lua/Lua.h` which was causing us... issues. I've hot-fixed that here, and I'm going to commit the full fix (move lua sources to `contrib/lua/src/`) as a follow-on to this PR.

impaktor:
> I would display that if I could; but I'm limited to ImGui's PlotLines widget which doesn't give me much in the way of rendering over top. 

If you know the y_max, y_min (same for x) of the graph, one could just print it below on a single line, if it's information that is of interest. If only relative fluctuations are interesting, no scale is needed.

PR #4795, opened 2020-02-16
Captains log, star date 2020...

impaktor:
## Introduction
Inspired by this [flight log](http://www.planetmic.com/orbit/fdl-file.htm) add-on for original Frontier, I've implemented something similar. This is highly work in progress. I've used the FlightLog that automatically logs departures from _stations_, and arrival/departure times into _systems_. I've also added a _custom_ log, that can be created by writing some fitting string, and pressing enter; and show all three in a Flight Log tab in InfoView.

~~Currently text can only be entered into the _custom_ logs, but I'll extend the other two logs (_system_, and _station_) so one can enter, as @nozmajner suggested: "this was a smelly station".~~

## Potential: Custom entries at start give backstory
I've played around with adding a custom log entry from the mainmenu, when starting a new game, so the first entry could be:

_"Just picked up the ship that uncle Nick left for me to inherrit. This rusty old bucket isn't much, but together we'll write our destiny in the stars, and it's my new home. Can't wait to undock!"_

*"Now I've cleared my debt with Wu, and sold all my belongings, and all I got was this lousy ship. By it's mine, and with it I can escape this god forsaken world and build a brighter future, forever onwards!"*

## Biggest hurdle
- Currently, it's to get the three logs (_system_, _station_, _custom_) to be printed in date-order, instead of printing each one in turn. I think the FlightLog method might have some method that sorts them, or combines them, but they have different table fields, so... need to think. 

- Multiline input. I've experimented with adding multiline text-input field, but the game crashes if I try to use it, so for now, it's just a dead commit.

- Also, this is a savebumping PR, since the updated FlightLog library will be incompatible with the previous, since I'm adding more fields (also log player's money), and a custom log.

## To do
- [x]  0 Allow re-editing old entries in custom logs
- [x]  1 Clear flightlog upon new game (bug of old?)
- [x]  2 Remove (only?) custom entries
-  3 Do we also want to be able to remove SYSTEM and STATION auto-logs as well?
- [x]  4 Add startup custom entry, backstory of how the ship ended up in the player's possession
- [x]  5 Edit/update Entries of SYSTEM and STATION auto-logs as well?
- [x]  6 ~~sort them on dates~~, not printing SYSTEM, STATION & CUSTOM logs separately. EDIT: Solved this by printing each log in a separate tab, thus #4893
- [x]  7 Better "Location" strings for CUSTOM, for if docked, landed, in orbit etc. 
- [x]  8 Document new methods in FlightLog
- [x]  9 Put strings in to translation system
[...]

PR #4817, opened 2020-02-27
WIP: missions view

vakhoir:
I'm (_slooooowly_) chipping away at this, and I'd like some feedback. I thought that the Bulletin Board mockup would also fit quite nicely with the Missions View, so I started designing it with that in mind. If you think it's a bad idea, I'll just move that to the BB view, though in that case I'd like a new mockup (I guess I can also just base it on the newGUI Missions View)

missions-01

For reference - the mockup from the forums:

BBS2

It's still WIP, I'm pretty sure I can make it much closer to the mockup.


Web-eWorks:
Looking really good so far! I've got a few suggestions, most of which are cosmetic:

- The selected mission should have a more pronounced background color than current; as it is right now it looks like alternating shading on a bank ledger.
- On the mission information, I'd like a button with a "pick target" or "crosshair" icon next to the destination to select the destination as the nav target. If we don't have an icon with those semantics, a bit of placeholder text will do until I can nudge @impaktor or @nozmajner into drawing us up one.
- Ideally if a target is in another system that button calls something to the effect of `ui.requestLaunchSectorView(SystemPath(x, y, z, ...))`; but we don't actually have anything like that functionality so I'm simply dreaming out loud.
- The field headings on the mission information (Destination, etc.) should ideally be bolded; not 100% sure we have bold font support in our current font chain.
- The mission deadline should show in parentheses the days (and hours) left until the deadline in addition to the deadline timestamp.
- Delivery missions which involve actual physical cargo and not a variable set on the player's ship should have an icon communicating such, as well as a field in the mission information showing the cargo the player is required to transport to the destination. (E.g. `Cargo: Battle Weapons (5t), Atmosphere Processors (27t)` )
- You should probably double-check the Character object (or at least the face seed) that generated the mission is being properly serialized and re-created across view changes and loads/saves. I know we were having an issue or two with that earlier.
- If a mission is in-system, replace the 0.00LY text with the straight-line distance to the target in Mm or AU. This should be fairly simple to retrieve; see the system outliner.

Overall, I really like the direction you're taking this. If you really want to improve on the concept, replace the `DELIVERY` text in the mission entry with an icon and move the `PACKAGE DROP` or `MOVING HOME` text into the top line of the entry; this will free up the bottom line for the mission description. Alternatively (and this might be a better use of the space) put the mission reward on the top line of the mission entry next to the icon at the end.

Maybe also play with the margin / padding a bit; some of it feels like the text is about to fall off the page, the left side of the mission information in particular...

impaktor:
> The field headings on the mission information (Destination, etc.) should ideally be bolded; not 100% sure we have bold font support in our current font chain.

I'm not so sure, look at the screenshot in the flight log PR: #4795 there the heading is grayed out, and it looks pretty good. The heading is there to show context, the text that follows it is the interesting part, so that's the part that should stand out (relative the headings field), me and nozmajner reason. In this case, just using gray and white does the trick.

> Delivery missions which involve actual physical cargo and not a variable set on the player's ship should have an icon communicating such,

Not sure I understand, aren't all cargo delivery missions for a fixed/non-negotiable amount of cargo?

> Overall, I really like the direction you're taking this. If you really want to improve on the concept, replace the DELIVERY text in the mission entry with an icon

Yeah, I don't like the DELIVERY. To me, it feels like it's exposing the underlying algorithm that's generating the mission, so the player learns that "all these missions are the same", a bit too fast. Sure, we kind of do the same today with icons, but it's less "in your face". To me, the thing that gives the illusion of it not all being the same, is the text "MOVING HOME", "GOING TO the {system}", "WANTED", "POSTAL SERVICE" and "PACKAGE DROP", as these are different flavors of the same module.

vakhoir:
> * The field headings on the mission information (Destination, etc.) should ideally be bolded; not 100% sure we have bold font support in our current font chain.

I thought the same, but we don't. ImGui itself does not support it out of the box, we'd need to load the bolded version as a separate font. Maybe I'll work on that when we're done with the migration to pigui, but in the mean time, I'd go with Impaktor's solution. The greyed out headings look pretty good to me.

> Overall, I really like the direction you're taking this. If you really want to improve on the concept, replace the DELIVERY text in the mission entry with an icon and move the PACKAGE DROP or MOVING HOME text into the top line of the entry;

Will do, but that's a bit trickier than it sounds. PACKAGE DROP / MOVING HOME are just a part of the longer mission description string. Not all missions even have that kind of all caps summary in the description, so I'll need to extract them / add them for each mission.

Thanks for the other suggestions!

impaktor:
> Will do, but that's a bit trickier than it sounds. PACKAGE DROP / MOVING HOME are just a part of the longer mission description string.

I'd like to object to this. The point of the strings is to _not_ sound algorithmically generated, but something a human posted to a bulletin board. As I understand it, what is being proposed is to take the strings generated by our algorithm to sound natural, and non-algorithmic, to then make them sound algorithmic again, by deconstructing them and splitting them up? Sounds like this will be a lot of work, and it would be best if we agree on what should be done, and what sould not be done before a lot of time is sunk into this.

Web-eWorks:
Considering that we're designing the "new" mission display here, I'd recommend we take the chance to add some new functionality to the mission display; let's separate the heading from the short and long description, so that missions can put whatever they want in those fields if they want to sound non-algorithmically generated.

cwyss:
I'd suggest a search-for-destination functionality, like search for a specific starport (in the current system), or a star system. 
Imagine you have a delivery to some other star system. And you just docked to some starport in your current system. If you are lucky, there is again a mission to the star system you already plan to go. Then a search option will be very helpfull if the list of missons is large (and in some systems it can be quiet large!)

In upstream, we have a generic search field for this, but search-for-destination would be enough here I guess, given that there is a filter-by-mission-type thing (if I understand the screenshots correctly)

impaktor:
I have been thinking in similar terms, but I think a more generic would be better, where one could sort BBS based on direction or something similar.

Web-eWorks:
@vakhoir how are things going? I'm assuming life has kept you busy the past few months, but if you get a chance to let us know the status of this PR I'd appreciate it!

PR #4819, opened 2020-02-28
NewUI Cleanup, part 1

Web-eWorks:
Removed the entirety of the newUI StationView, removed the GalaxyMapView (per discussion on IRC), and moved our bundled Lua runtime one folder deeper to avoid confusion with the equally-valid `src/lua`.

This PR is waiting on one more set of commits to remove the unused LuaFace and LuaModelSpinner newUI classes, then it will be ready for merge.


fluffyfreak:
Erm, removed the GalaxyMapView?
This is why we shouldn't do things _just_ on IRC, even when I can get on I can't always pick out bits from the inredibly long conversation history.

Care to summarise?

impaktor:
Yeah, best would be to move galaxymap to pigui.

As I mentioned on IRC: I prefer to keep galaxymap, but also to kick out NewUI asap, those are obviously conflicting. Maybe @vakhoir  or @Gliese852 feel motivated to move it to pigui. I have other corners of pioneer I'd like to work on at the moment.

Web-eWorks:
To summarize: the galaxymap has but one purpose, and that is to show how far the player is from Sol.

Showing the density of stars (as brought up in IRC) is fundamentally the responsibility of the SectorView; zoom all the way out and scroll around and you'll notice nice voronoi-map faction borders and plenty of stars, and if there were an actual need to plan around stellar density as a player, you would be able to observe different densities in the distribution of stellar bodies. Right now, it's a mostly-uniform mass of star soup.

The GalaxyMap and the SectorView lack a defined shared coordinate system or any obvious way to use the former to navigate in the latter (other than some experimentation with sector coordinates) making its other discussed purpose - that of planning large trips across portions of the galaxy, say Sol to the galactic center - difficult if not impossible to accomplish currently.

You *cannot* interact with the GalaxyMap. It's just there to look pretty. If you want to plan a trip, you have to figure out the coordinate mapping between GalaxyMap and SectorView, then zoom out and scroll in what you hope is the correct direction, plot a series of jumps, try jumping, and then see if your dot on the GalaxyMap has moved that one pixel in the correct direction.

Suffice to say, I find that to be the responsibility of the SectorView instead, and I do not find the GalaxyMap so important as to be worth the effort of re-implementing in pigui.

That being said, if there is someone willing to port it to pigui, and others contributors prefer to keep it for reasons beyond duplicating functionality that SectorView should be providing, I will retract the commit in question.

However, in my view, the effort spent porting this screen to pigui could be put to much better use improving the SectorView. Literally just switch between 3D and 2D display when you zoom out far enough and draw our galaxy texture in the 2D mode and you will have already re-implemented the GalaxyMap in its entirety.

Web-eWorks:
Removed all of the now-unused classes in `src/gameui/`. The only thing left over is the newUI face display class, which will likely stick around until all of the missions are converted to using pigui faces.

Web-eWorks:
Per discussion on IRC, the galaxy map view removal stands. We'll be merging the useful parts of its previous function into the SectorView as part of the next iteration of that view.

PR #4820, opened 2020-02-28
Perf counter Windows memory info

fluffyfreak:
Grabs the process memory information on Windows, also fixes up the VS2019 (_yes I'm still maintaining it_) build.


Web-eWorks:
Looking pretty good; only thing I can mention is making sure CMake builds have the necessary defines (e.g. do we need to specify we're linking against `psapi.lib`?)

impaktor:
Travis complains about placement of curly bracket, and header file order. I think that should be fixed before this goes in.

fluffyfreak:
Ugh, I'm going to squash that history when I merge it

PR #4822, opened 2020-02-28
Adding icons for future PRs: BBS, systemview, imgui migration, bar

nozmajner:
Updating the icons.svg with a few icons that will be needed for future PRs by @impaktor @Gliese852 @Web-eWorks 
Some of the icons are by Gliese852 (I don't know how I tag as author for the PR)
I've also named all icons in the file with them names from the default.lua for consistency.
I've also changed the display nav target data icon from the reticule from the crosshair to that google-maps like pin. I've added that pin for the BBS set destination as nav target function as well.
I've also removed the old apoapsis and periapsis icons in favor of the new ones. They weren't in use anyway.

icons

### DIFF
--- a/data/icons/icons.svg
+++ b/data/icons/icons.svg
@@ -2,6 +2,7 @@
 <!-- Created with Inkscape (http://www.inkscape.org/) -->
 
 <svg
+   xmlns:osb="http://www.openswatchbook.org/uri/2009/osb"
    xmlns:dc="http://purl.org/dc/elements/1.1/"
    xmlns:cc="http://creativecommons.org/ns#"
    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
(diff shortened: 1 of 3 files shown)


impaktor:
75589766-f5d3d800-5a7b-11ea-88f9-07ac74ff39ce

- Why add a new "dot"?
- Why are there two positions markers
- Also, I note, that we have, from old, 
  - two "saturns" (first one is used `gas_giant`, second one called `map` I don't think is used anywhere)
  - two skulls, the first one isn't used anywhere. Not sure which skull is best, but do we need two?

Also, @robothauler check second to last row, this is inteded for the Find Missing person mission.

Gliese852:
> Why add a new "dot"?

One dot for the BBS and other for the Lagrange point.

impaktor:
Could one not just call it "dot", and map that to both `default` and `lagrange`?

Gliese852:
@impaktor Not sure about the BBS - default (did not find a mention in the default.lua),the lagrange point is called `lagrange`

impaktor:
Well, the "default" is what is used when no icon is specified for a BBS module.

impaktor:
OK, added icons to themes.

- I'll squash the first indentation fix commit after review.

I've changed names on some of your icons @Gliese852, to be slightly more generic, as none of the other icons specify exactly where they're being used. The naming convention is a bit mixed. Should it describe what the icon is, or what it is being used for (I prefer the former)?

Well, to stay faithful with the rest of the code, I was consistently inconsistent.

So with this, we agree to have double symbols in SVG, because some could change. Like make a more refined marker for lagrange points, or default BBS symbol.

PR #4828, opened 2020-03-05
Add histogram to pigui

impaktor:
I'm using this in my [debug commodity branch](https://github.com/impaktor/pioneer/tree/debug_commodity_prices), please review, as I wrote some code that didn't quite work, and then @ecraven helped by writing the code below, but I renamed all the variables, and added doc string, because I think they were sub-optimally named.

I also added some default variables, if user doesn't specify any explicitly themselves.

After thumbs up on this, I can add `plotLines()` in the same commit, as it's just copy paste, I think.

## Example
2020-03-08-083815_1600x900_scrot

### DIFF
--- a/src/lua/LuaPiGui.cpp
+++ b/src/lua/LuaPiGui.cpp
@@ -494,6 +494,57 @@ static int l_pigui_get_scroll_y(lua_State *l)
 }
 
 /*
+ * Function: plotHistogram
+ *
+ * Make a histogram
+ *
+ * > ui.plotHistogram(label, data, display_count, offset, overlay, y_min, y_max, graph_size)
+ *
+ * Example:
+ *
+ * Plots 5 bars, with height 17,3,15,5, and 8, respectively
+ *
+ * > ui.plotHistogram("Test", {17,3,15,5,8})
+ *
+ * Parameters:
+ *
+ *   label - string, text on button
+ *   data - table of values
+ *   display_count - optional, to limit number of pionts to plot
+ *   offset - optional x-axis offset, default: 0
+ *   overlay_text - optional title string, to put on histogram
+ *   y_min - optional float, setting min y-value displayed
+ *   y_max - optional float, setting max y-value displayed
+ *   graph_size - optional Vector2, defining size
+ *
+ */
+static int l_pigui_plot_histogram(lua_State *l)
+{
+	PROFILE_SCOPED()
+	std::string label = LuaPull<std::string>(l, 1);
+	LuaTable vals = LuaTable(l, 2);
+	float values[vals.Size()];
+	float max = FLT_MIN;
+	float min = FLT_MAX;
+	for (int i = 1; i <= vals.Size(); i++) {
+		values[i - 1] = vals.Get<int>(i);
+		if (values[i - 1] > max)
+			max = values[i - 1];
+		if (values[i - 1] < min)
+			min = values[i - 1];
+	}
+	int display_count = LuaPull<int>(l, 3, vals.Size());
+	int values_offset = LuaPull<int>(l, 4, 0);
+	const char *overlay_text = LuaPull<const char *>(l, 5, NULL);
+	float y_min = LuaPull<float>(l, 6, min);
+	float y_max = LuaPull<float>(l, 7, max);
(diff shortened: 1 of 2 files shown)


pcercuei:
I wonder how that will reflect on the gameplay. Doesn't that make things a bit too easy?

nozmajner:
@pcercuei It's more like a debug/balancing tool for trading.

impaktor:
@pcercuei  I agree, which is why this PR is only for the histogram widget. But I do think we need a average-price column in the commodity market. Possibly one could just take the commodity base-price, before system specific market modifier is applied. Maybe something @vakhoir might want to know about / tinker with.

src/lua/LuaPiGui.cpp
```diff
@@ -493,6 +493,55 @@ static int l_pigui_get_scroll_y(lua_State *l)
...
+	float max = FLT_MIN;
+	float min = FLT_MAX;
+	for (int i = 1; i <= vals.Size(); i++) {
+		values[i - 1] = vals.Get<int>(i);
+		if (values[i - 1] > max)
+			max = values[i - 1];
+		if (values[i - 1] < min)
+			min = values[i - 1];
+	}
+	int display_count = LuaPull<int>(l, 3);
```
Web-eWorks: Ideally the default behavior should be to display all values in the data array; but we offer the caller more control as to how many values they'd like to display.

```suggestion
	int display_count = LuaPull<int>(l, 3, vals.Size());
```

src/lua/LuaPiGui.cpp
```diff
@@ -493,6 +493,55 @@ static int l_pigui_get_scroll_y(lua_State *l)
...
+	const char *overlay_text = LuaPull<const char *>(l, 4, NULL);
+	float y_min = LuaPull<float>(l, 5, min);
+	float y_max = LuaPull<float>(l, 6, max);
+	ImVec2 graph_size = LuaPull<ImVec2>(l, 7, ImVec2(80, 80));
```
Web-eWorks: If the calling code isn't passing a value for the histogram, let ImGui size it appropriately instead of constraining it to a 80px by 80px square.

```suggestion
	ImVec2 graph_size = LuaPull<ImVec2>(l, 7, ImVec2(0, 0));
```

src/lua/LuaPiGui.cpp
impaktor: I'm not sure that variable does what I think it does. From Imgui demo example:

Here, it's called `display_count`:
``` c++
ImGui::PlotLines("Lines", func, NULL, display_count, 0, NULL, -1.0f, 1.0f, ImVec2(0,80));
ImGui::PlotHistogram("Histogram", func, NULL, display_count, 0, NULL, -1.0f, 1.0f, ImVec2(0,80));
```

Here it's called `values_offset`:

```c++
ImGui::PlotLines("Lines", values, IM_ARRAYSIZE(values), values_offset, overlay, -1.0f, 1.0f, ImVec2(0,80));
ImGui::PlotHistogram("Histogram", arr, IM_ARRAYSIZE(arr), 0, NULL, 0.0f, 1.0f, ImVec2(0,80));

```
Here's an experiment I did:
2020-03-14-104449_1600x1800_scrot

And in the screenshot above, explicitly setting it to 10, looks to just shift the plot (still plots 31 points), so it's an offset? Then we should probably expose the third argument on c++ side as well, assuming that is the _real_ display_count? 

(I'll test that now)

src/lua/LuaPiGui.cpp
impaktor: I did try the 0,0 option, but thought the window became strangely elongated, see screenshot in previous comment

impaktor:
I've pushed new changes: added a new argument, for the offset, and made the `display_count` do what it's supposed to do.

This code:

```lua
ui.text("only give 2 args (label, data)")
ui.plotHistogram("##price distribution1", data)

ui.text("only give 4 args, explicitly say to display all data, and offset by half")
ui.plotHistogram("##price distribution2", data, #data, math.floor(#data/2))

ui.text("only give 3 args, explicitly say to display half the points")
ui.plotHistogram("##price distribution3", data, math.floor(#data/2))
```

gives this output, which seems to be correct:
2020-03-14-104449_1600x1800_scrot

PR #4833, opened 2020-03-11
Stars that are brighter are now bigger and have a brighter color on the Starfield

tsoj:
To control the appearance there are new properties in data/config/Starfield.ini, I wrote some comments describing roughly what each parameter does. Here are three possible configurations:
screenshot1
screenshot2
screenshot3


fluffyfreak:
Nice work @tsoj

tsoj:
Ok, I noticed two issues that I need to solve first. When doing a hyperjump, the animation looks incredibly ugly. Also, after arriving at a system I needed to pause first and then resume to get the starfield filled properly. I just wanted to say why I closed this pull-request.

fluffyfreak:
@tsoj we don't close them for problems, just mark it as WIP in the title and added the WIP tag ;)
That way you can work on it, or it can be fixed by other people.

Web-eWorks:
@tsoj I saw you dropped by IRC earlier today; if you're having issues with speaking in the channel, you'll need to authenticate with [NickServ](https://freenode.net/kb/answer/registration) before you can talk on freenode due to the influx of spam across the entire IRC network in recent years.

impaktor:
Sorry, but this really should have been squash-merged. Now we have three commits that are just iterating settings in starfield. (I haven't looked at the other 5 commits, as it's amoot point now)

PR #4835, opened 2020-03-13
Texture Cache Visualizer and Memory Usage Tracker

Web-eWorks:
I've implemented a visualizer for our texture caching system that allows inspecting each of the textures stored in the cache (currently only works on Texture2Ds), as well as added performance tracking for memory usage and number of cached textures.

Screenshot from 2020-03-13 15-02-39

I eventually want to separate the texture cache into a two-stage indirect cache instead of lumping everything into the same `std::map`; among other reasons, I'd like to remove the static mutex in `TextureBuilder` and instead lock on the specific texture cache bucket we're assigning the texture to.

I've also updated our clang-format script to avoid leaking the generated patch file, and make it easier to manually run the formatter.

Most of this PR has been developed on the terrain-texturing branch, but I've endeavored to cut out everything not relevant to the texture cache changes.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -9,7 +9,7 @@ AlignAfterOpenBracket: DontAlign
 #AlignConsecutiveDeclarations: false
 #AlignEscapedNewlines: Right
 AlignOperands:   false
-AlignTrailingComments: false
+AlignTrailingComments: true
 AllowAllParametersOfDeclarationOnNextLine: false
 #AllowShortBlocksOnASingleLine: false
 AllowShortCaseLabelsOnASingleLine: true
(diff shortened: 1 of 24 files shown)


Web-eWorks:
There are still some rough edges; non-compressed textures don't report their memory size, the DDS texture loading doesn't handle non-POT textures very well right now, and I'm sure there are a few extra things we should probably throw into the visualizer.

However, as it stands right now it *works*, and I'll be continuing to work on our texture handling in the terrain texturing branch, so I think this is worth merging.

fluffyfreak:
@Web-eWorks wow that is so cool!
MERGING!

impaktor:
Wow, looks very cool! Moar debug-tools!

impaktor:
More bling bling with 2D IMGUI plotting: 
https://github.com/epezent/implot

PR #4838, opened 2020-03-21
Starfield moving with player

### LINKED ISSUE #4837: Background stars don't track sector position
### Observed behaviour
As the player moves through the galaxy the stars don't track or change with the players position like they should

### Steps to reproduce

fluffyfreak:
We populate the background star field with stars, this should track the players position through the galaxy but it wasn't until this PR.

Fixes #4837


richardpl:
Cool!

PR #4841, opened 2020-03-24
Move clang-format to Github Actions

Web-eWorks:
This change splits our clang-format step out into its own check, makes it a bit easier to run clang-format (for several reasons including that it's included into the default Github Actions image...), and allows us to more clearly see when there's an actual error with the build instead of clang-format complaining that you haven't adhered to the proper syntax.


impaktor:
@Web-eWorks looks like github action is complaining about there being errors?

Issue #4797, opened 2020-02-19
Commodity market activities not reset


Cyborgscode:
Version: 20200203
OS: Fedora 30

Steps to reproduce:

Dock.
Commodity Market
select i.e. Radioatives to be bought, but don't buy them.
(i.e. take a transport mission)
Fly to a different station ( i think redocking would result in the same result)
Dock.
Open Commodity Market.

You will notice that the ui still shows the selection to buy the commodity you had earlier.
Conclusion: Market window is not reset to default on undock.

impaktor:
Good catch! @vakhoir might want to include a fix for it in #4775

vakhoir:
This might be just a display issue - the selected item is not reset when the view is refreshed. I'll fix in that PR.

PR #4844, opened 2020-03-25
Fix Commodity Market Message Lag

### LINKED ISSUE #4826: Commodity market state - yellow and red text messages lag "one action"
### Observed behaviour
The commodity market has "incuficient funds" (yellow) and "none available" (red text) messages when trying to buy too expensive or sold out commodities. However, one has to click twice on the commodity for it to appear, and when clicking the next product that warning text is still there, until one clicks the commodity one more time. 

### Expected behaviour
Either that the text doesn't show up at all, until the player tries to buy it, or that it shows up upon the first click on the commodity. (latter preferred)

### Steps to reproduce
1. start pioneer, new game, go to commodit market
2. click on something cheap, that's in stock, make usre all is well.
3. click on something that costs more than you have, or something that's sold out. Notice that if you click a second time, then the appropriate message pops up.
4. repat 2. notice that the warning message is still shown, unless you click that commodity a second time.

Maybe @vakhoir want to have a look at this.

Current master: 4ade8f86fd9200dfbd0847bb84f87201df63cbf3

vakhoir:
Fixes #4826

A small issue possibly introduced with the fix to #4797. The method that updates the commodity market message was called before the selected item was set on the object, and so the message was always lagging.

To avoid confusion of what the Refresh method actually does, the selection clearing has to be done explicitly, and has been moved to the appropriate StationView tab.

### DIFF
--- a/data/pigui/libs/commodity-market.lua
+++ b/data/pigui/libs/commodity-market.lua
@@ -85,10 +85,10 @@ function CommodityMarketWidget.New(id, title, config)
     end
     config.canDisplayItem = config.canDisplayItem or function (s, e) return e.purchasable and e:IsValidSlot("cargo") and Game.system:IsCommodityLegal(e) end
     config.onClickItem = config.onClickItem or function(s,e,k)
+        s.selectedItem = e
         s.tradeModeBuy = true
         s:ChangeTradeAmount(-s.tradeAmount)
         s:Refresh()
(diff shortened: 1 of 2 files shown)


impaktor:
I wonder if any change is needed in Goodstrader as well?

vakhoir:
Just double checked, looks good to me as is.

PR #4849, opened 2020-04-01
Refactor the ModelViewer to use PiGui

Web-eWorks:
I've re-implemented the ModelViewer in PiGui, using the GuiApplication class and the Input system to simplify a fair amount of the handling.

To accomplish this, I've done a fair amount of refactoring:
- `PiGui` is now `PiGui::Instance`, and all of the static drawing methods have been moved out of the class, opening up the possibility to add more C++-side widgets for both Lua and C++ to use.
- `PiGui::Instance` is now only responsible for handling ImGui - all Lua-side initialization is handled by `PiGUI::Lua::Init()` instead.
- As an addenda, the (separate) PiGUI namespace will be merged into the (new) PiGui namespace in a future PR; I didn't do that here because I was focused on other work.
- `PiGui::Instance` has its renderer dependency injected by its owner rather than relying on `Pi::renderer`. This actually reduced complexity a bit, because we no longer had to manually specify the SDL window every time we started an ImGui frame.
- `Input` had its last ties to `Pi` removed, with the Config dependency injected by the owner, similarly to PiGui.
- Of note: `KeyBindings::AxisBinding` and `ActionBinding` still depend on the `Pi::input` singleton variable, so there's an ugly hack in the ModelViewer to work around that. I'll address that in a later PR that refactors bindings (once again!).
- `GuiApplication` now owns an Input instance and a PiGui instance, decoupling management of those two classes away from `Pi.cpp`. It's not as automatic as I'd like, due to oldUI's issues with label drawing, but it's 90% of the way there.
- `Input` now has methods `IsKeyPressed()` and `IsKeyReleased()` to check whether a specific key was pressed or released during the current frame. This does come with some more runtime complexity, but the overall effect should be mostly negligible.

Overall, this is a 1:1 replacement of the previous ModelViewer, with only one regression: the "snap view to direction" keybindings are not currently working because they're not being fed SDL_Events to trigger the `onPress` callbacks. This will also be resolved in a future Input PR, but is outside of the scope of this work.

### DIFF
--- a/src/Input.cpp
+++ b/src/Input.cpp
@@ -9,7 +9,13 @@
 
 #include <array>
 
-void Input::Init(GameConfig *config)
+Input::Input(const GameConfig *config) :
+	m_capturingMouse(false),
+	mouseYInvert(false),
+	joystickEnabled(true),
+	keyModState(0),
+	mouseButton(),
+	mouseMotion()
 {
 	joystickEnabled = (config->Int("EnableJoystick")) ? true : false;
 	mouseYInvert = (config->Int("InvertMouseY")) ? true : false;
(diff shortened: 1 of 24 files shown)


impaktor:
Did we find out what's happened to travis?

Web-eWorks:
Nope, still not sure what's going on there. The config file works fine, travis is building correctly, but it's not showing up on the repo. I think you'll have to dig around in the Travis settings to figure out what's going on.

Web-eWorks:
Merging this in, as I think we've caught all of the related bugs!

PR #4856, opened 2020-04-13
Docking fee not applied if fuel tank is at least 99.9 % full

### LINKED ISSUE #4804: Landing fee charged when starting a new game
### Observed behaviour

I am being charged 6 credits at the very beginning of the game, for "landing". This is unacceptable. How could I have been landing before starting the game?

### Expected behaviour

The player should start with 100 credits, as advertised in the main menu. Anything else is just false advertising.

### Steps to reproduce

Start a new game. Get unjustly charged. Complain.

Latest master (e25517163) on Debian x64.

Echoes91:
Silly fix proposed to avoid docking fee as onShipDocked is triggered when starting a new game, and possibly other "unfair" fees too (eg. re-docking after accidental launch).

Fixes  #4804

### DIFF
--- a/data/modules/StationRefuelling/StationRefuelling.lua
+++ b/data/modules/StationRefuelling/StationRefuelling.lua
@@ -29,6 +29,9 @@ local onShipDocked = function (ship, station)
 		ship:SetFuelPercent() -- refuel NPCs for free.
 		return
 	end
+	if ship.fuel > 99.9 then -- No docking fee when starting a new game
+		return
+	end
 	local fee = calculateFee()
 	if ship:GetMoney() < fee then
 		Comms.Message(l.THIS_IS_STATION_YOU_DO_NOT_HAVE_ENOUGH:interp({station = station.label,fee = Format.Money(fee)}))


Gliese852:
Imho if you took off and then re-dock again, you need to be fined additionally.

impaktor:
Well, that's a pretty ugly fix...

Don't get me wrong, I don't have any better suggestion.

Peraps, one can use number of ticks/date? Should be very low when starting a new game?

Gliese852:
Maybe we do not need to generate the event "onShipDocked" when the game starts?

pcercuei:
That would be the proper fix, yes.

Keeper1st:
Oh, I don't know. When you're starting the game, you've got a ship and you have asked it to be delivered to your home station. Of course they're going to charge you the docking fee, even if you've never flown it yet. You can think of it as a hangar storage fee in that case.

WKFO:
Although Keeper1st has a good point, I want to say that losing 4 credits right away (from the advertised 100 credits in the main menu) looks like a scam.

impaktor:
When starting a game one does not dock. There is no telling where the ship comes from, or how long it's been docked (although backstory is/will be given in #4795). Either way, triggering the docking event is wrong when starting a new game.

I think the propper fix is to not trigger the docking event.

And correct me if I'm wrong, but isn't there a problem that the current way isn't even deterministic? I.e. it's not even triggering the docking event reliable on every new game start? Thus, IMO, the times docking event is triggered is the bug.

Gliese852:
@impaktor IIRC, every start of a new game triggers this event. Now there is no method outside the SpaceStation class namespace, that can place a ship in a station without triggering ondocking event. In fact, I seem to fixed it, created a PR to this branch, but @Echoes91 doesn't react yet.

Echoes91:
Sorry for absence, I see many valid point to not implement this fix (from not triggering the event on start to apply fee when re-docking) so I'm closing the PR.

impaktor:
@Echoes91 thanks for the effort. :+1:

joonicks:
IMHO, instead of a bunch of extra code just to initialize a new game, I think it would be better to just rip that out and transparently load a savegame edited to the specific starting conditions.

Issue #4858, opened 2020-04-13
Sell price reduction kills commodity trading


Echoes91:
Maybe this could  be part of [Economy Rebalance](https://github.com/pioneerspacesim/pioneer/projects/2)

### Observed behaviour
Currently `sellPriceReduction = 0.8` is applied both to equipment and commodities, making most trade routes unprofitable. For example:

- Base/buy price of _Mining machines_ is 302 on Sirius (Minor export)
- Base/buy price of _Mining machines_ is 346 on Delta Eridani (Major Import)
- 1 ton bought for 302 on Sirius can be sold for 346 * 0,8 ~ 279 on Delta Eridani, resulting in a loss while trading from a minor export to a major import.

### Expected behaviour
I think any trade from minor export to minor import or better should never result in making a loss.

### Steps to reproduce
Buy any commodity and resell it in the same station to observe the price difference

### Possible fixes
- Reduce or remove `sellPriceReduction` for commodities at least (maybe decoupling from equipment). Seem like this was [done within old UI](https://github.com/pioneerspacesim/pioneer/pull/3161) but not carried over.
- Increase price deltas between import and export to increase profits (0,8 : 1 should be the minimum ratio)

__pioneer 20200410 (609273394) on Arch Linux__

impaktor:
> I think any trade from minor export to minor import or better should never result in making a loss.

In general, this should be true, not sure about the "never" part, since "normal" price could be very different in different regions of space.

I agree that there should not be a price reduction when selling commodity. The reduced price is mainly for equipment as it's now second hand / used.

PR #4859, opened 2020-04-16
Reduce hydrogen price back to 1 credit

WKFO:
̷R̷a̷i̷s̷i̷n̷g̷ ̷l̷o̷c̷a̷l̷ ̷d̷e̷l̷i̷v̷e̷r̷y̷ ̷r̷e̷w̷a̷r̷d̷s̷.̷
After countless protests across the galaxy, fuel prices are reduced again.

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -494,7 +494,7 @@ local CARGOLANGRESOURCE = "commodity"
 
 cargo = {
 	hydrogen = EquipType.New({
-		l10n_key = 'HYDROGEN', l10n_resource = CARGOLANGRESOURCE, slots="cargo", price=10,
+		l10n_key = 'HYDROGEN', l10n_resource = CARGOLANGRESOURCE, slots="cargo", price=1,
 		capabilities={mass=1}, economy_type="mining",
 		purchasable=true, icon_name="Hydrogen"
 	}),


WKFO:
Drafted because still not profitable.

impaktor:
Ah, yes, I've been planning to do this. In Frontier the rewards were typically: 
- Package delivery ~ 100-500 (no time limit)
- Taxi missions ~ 1000-5000
- Assassination ~ 7000-20,000

I'd like to see the actual rewards generated from these new levels. E.g. put each code snippet in a for loop and generate 1000 rewards for each, and show distribution for each (or mean/median/variance). (I have code for that, although it's trivial.)

but as you've pointed out (elsewhere), fuel cost is too high, as I've assumed using ~1t fuel per travel, i.e. ~10 credits in travel cost. The way I see it, fix is to either lower in-system fuel consumption e.g. 10 fold, or put fuel price (temporarily) back to 1 credit / tonne, until we've come up with something better.

Zireael07:
I think the latter makes more sense, as I've pointed out elsewhere, hydrogen is the most abundant element in the universe. Even if it's processed somehow to become fuel, 10cr/tonne is too much.

nozmajner:
@Zireael07 Yes, but storing Hydrogen is anything but trivial. It seeps trough the wall of tanks even in cryogenic form.

Zireael07:
What year is Pioneer set in? Somewhere in 22nd century or even 23rd? I'm sure it's a solved problem by then, if it's widely used as fuel.

WKFO:
Pioneer is in 3200's.

WKFO:
So would you rather have this PR reduce the hydrogen price until we balance everything and re-implement the new economy, or...?

Zireael07:
By reducing hydrogen price, you both make it more sensible from a point of view of a galaxy-spanning economy, AND don't have to change fuel consumption.

impaktor:
@WKFO yes. I think putting Hydrogen back to 1 credit, would again bring cost to ~10 credits per trip, which is what the trade is scaled for. And missions as well.

I'm not against hydrogen at 10 credits, but then either ship fuel consumption needs to go down, or trade and mission rewards need rebalancing, the latter of which seems to be the most work, which I'll gladly stay a way from.

WKFO:
Makes my job a lot easier :)
Side note, trade is still broken because of #4858

WKFO:
please review this, thanks

PR #4860, opened 2020-04-19
Scoop mission

robothauler:
~~As you can see, this is not finished yet. Reward and mission time are not balanced so far.
Just to show something and play with the possibilities.~~

This module creates 4 mission types.
1. A debris field with cargo containers. The player has to pay for this information.
2. A debris field with illegal cargo. The player has to pay for this information.
3. A rescue mission. Pick up Rescue Capsules and land at a station.
4. Pick up a drifting container and deliver it to a ship in orbit around a star.

Additionally it creates random debris fields depending on population and lawlessness when entering the system.

I added an additional lua event to identify which cargo is related to which mission (2 similar missions) and for the NavButton.

Comments are welcome!

### DIFF
--- a/data/lang/module-scoop/en.json
+++ b/data/lang/module-scoop/en.json
@@ -1,0 +1,198 @@
+{
+  "ACCEPTED_ARMS_DEALER": {
+    "description": "",
+    "message": "Great! Thank you."
+  },
+  "ACCEPTED_ILLEGAL_GOODS": {
+    "description": "",
+    "message": "Thanks. You won't regret it!"
+  },
+  "ACCEPTED_LEGAL_GOODS": {
+    "description": "",
+    "message": "Thank you. You know a good deal when you see one!"
+  },
+  "ACCEPTED_RESCUE": {
+    "description": "",
+    "message": "Thank you very much. The crew will appreciate that."
+  },
+  "ADTEXT_ARMS_DEALER": {
+    "description": "",
+    "message": "PICK UP: Prompt recovery of drifting container required."
+  },
+  "ADTEXT_ILLEGAL_GOODS": {
+    "description": "",
+    "message": "EASY MONEY: Valuable information for sale."
+  },
+  "ADTEXT_LEGAL_GOODS": {
+    "description": "",
+    "message": "INFORMATION FOR SALE: Location of a debris field with valuable cargo."
+  },
+  "ADTEXT_RESCUE": {
+    "description": "",
+    "message": "HELP NEEDED: Urgent escape capsule rescue operation."
+  },
+  "CLIENT": {
+    "description": "",
+    "message": "Client:"
+  },
+  "DANGER": {
+    "description": "The risk level",
+    "message": "Danger:"
+  },
+  "DEADLINE": {
+    "description": "Must be delivered by",
+    "message": "Deadline:"
+  },
+  "DENY_1": {
+    "description": "",
+    "message": "Excuse me, sir? I don't think your current qualification is sufficient."
+  },
+  "DENY_2": {
+    "description": "",
+    "message": "Sorry, I think this task is beyond your ability as a pilot."
+  },
+  "DETONATORS": {
+    "description": "",
+    "message": "Detonators"
+  },
+  "DOCKING_INSTRUCTION": {
+    "description": "",
+    "message": "Nice to meet you. Please approach to within 100m to complete the cargo transfer!"
+  },
+  "FAILURE_MSG_ARMS_DEALER": {
+    "description": "",
+    "message": "You fool. The client's not gonna like this!"
+  },
+  "FAILURE_MSG_RESCUE": {
+    "description": "",
+    "message": "Your unreliability has caused great harm. Get out of my sight!"
+  },
+  "HOW_MUCH_TIME": {
+    "description": "",
+    "message": "How much time do I have?"
+  },
+  "HOW_MUCH_TIME_ARMS_DEALER": {
+    "description": "",
+    "message": "My business partner is in orbit around {star} to avoid detection. It would be great if you could reach him before {date}."
+  },
+  "HOW_MUCH_TIME_ILLEGAL_GOODS": {
+    "description": "",
+    "message": "Don't waste time. Only the early bird catches the worm!"
+  },
+  "HOW_MUCH_TIME_LEGAL_GOODS": {
+    "description": "",
+    "message": "I don't know. But if I were you, I would be quick!"
+  },
+  "HOW_MUCH_TIME_RESCUE": {
+    "description": "",
+    "message": "It would be good if you could reach the location before {date}."
+  },
+  "INTROTEXT_ARMS_DEALER": {
+    "description": "",
+    "message": "Hello Commander. My name is {client}. It would be great if you could help me out. An unreliable freighter captain dropped my cargo near {planet} and has disappeared since. Your task would be to pick up one container and deliver it to ship {shipid} in orbit around {star}. I would pay you {cash}."
+  },
+  "INTROTEXT_ILLEGAL_GOODS": {
+    "description": "",
+    "message": "Hey buddy, I know about a debris field with illegal goods. The police caught a smuggler close to {planet}. They haven't had time to clean up the area yet. If you are quick, you can make easy money! I can give you the coordinates. It only costs you {cash}. Don't let the chance slip by!"
+  },
+  "INTROTEXT_LEGAL_GOODS": {
+    "description": "",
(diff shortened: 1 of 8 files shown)


impaktor:
I haven't looked at the code. I assume numbr 3 can use the SAR BBS icon?

> Additionally it creates random debris fields depending on population and lawlessness when entering the system.

This almost feels like it might be in another module. Not sure. If so, a create debris field could be moved to some lua library. Just thinking out loud.

Gliese852:
Got an error while loading my savegame:
```
error: [string "[T] @modules/Scoop/Scoop.lua"]:762: bad argument #1 to 'pairs' (table expected, got nil)
stack traceback:
	[C]: in function 'pairs'
	[string "[T] @modules/Scoop/Scoop.lua"]:762: in function 'cb'
error: Frame instance deletion outside 'DeleteFrame' [0]
```
My savegame:
[arrive.zip](https://github.com/pioneerspacesim/pioneer/files/4520168/arrive.zip)

A new game starts successfully.

Gliese852:
@robothauler And what if you do this: if the cargo is located beyond a certain distance from the player for too long, it disappears, the mission is deleted, and a message is sent to the commlog that the ephemeris is out of date.

impaktor:
Will need a rebase or just a squash merge.

impaktor:
@robothauler will this PR depend on #4877 first being merged?

I haven't followed the details in IRC, but as I understand it @Gliese852 and you have made it so one will scoop up an escape pod (that is basically a modified cargo container)?

Gliese852:
@impaktor I just made the object have orbital speed. Definitely #4877 should be merged earlier, otherwise a "ERROR" model will appear instead of the capsule.

data/lang/module-scoop/en.json
```diff
@@ -165,7 +165,7 @@
   },
   "SUCCESS_MSG_RESCUE": {
     "description": "",
-    "message": "Thank you!"
+    "message": "Thank you very much! The crew would also like to express its thanks."
```
Web-eWorks, line 168: In English, that would read "The crew would also like to express their thanks", as 'crew' is a plural noun referring to several persons and 'its' is a singular pronoun typically used to refer to an object, not a person

Gliese852:
@robothauler it may be worth adding a button "set as navigation target" for the station to which you need to return the capsule?

robothauler:
@Gliese852 you can return to any station. Maybe a button to the nearest station?

impaktor:
Thanks for updating, I think the last commit should be squashed into some appropriate previous commit (the ones implementing those things).

src/CargoBody.cpp
```diff
@@ -89,6 +102,7 @@ void CargoBody::TimeStepUpdate(const float timeStep)
 	if (m_hasSelfdestruct) {
 		m_selfdestructTimer -= timeStep;
 		if (m_selfdestructTimer <= 0) {
+			LuaEvent::Queue("onCargoDestroyed", this, NULL);
```
Web-eWorks, line 105: Please prefer the use of `nullptr` over `NULL` everywhere in the code - NULL is not type-safe and doesn't follow our coding style. In addition, if you're trying to pass a `nil` parameter to Lua, simply leave the value off and only push one parameter - the LuaEvent mechanism works perfectly fine with 0..3 parameters in addition to the event name.

Web-eWorks (changes requested):

Gliese852:
@robothauler if you fixed what the @Web-eWorks  was asking for, you now need to re-request the review.

robothauler:
@Gliese852 thanks for the hint!

data/lang/module-scoop/en.json
```diff
@@ -0,0 +1,190 @@
...
+    "description": "",
+    "message": "I don't know. But if I were you, I would be quick!"
+  },
+  "HOW_MUCH_TIME_RESCUE": {
+    "description": "",
+    "message": "It would be desirable if you could reach the location before {date}."
+  },
+  "INTROTEXT_ARMS_DEALER": {
+    "description": "",
+    "message": "Hello Commander. My name is {client}. It would be great if you could help me out. An unreliable freighter captain dropped my cargo near {planet} and has disappeared since. Your task would be to pick up the cargo and deliver it to ship {shipid} in orbit around {star}. I would pay you {cash}."
```
Web-eWorks, line 88: The arms dealer text should ideally mention the fact that this is only one ton of goods (e.g. with a form option similar to the Cargo missions) but this is a low-priority papercut.

data/modules/Scoop/Scoop.lua
```diff
@@ -0,0 +1,837 @@
...
+	mission:Remove()
+	missions[ref] = nil
+end
+
+-- Cargo transfer to a ship
+local transferCargo = function (mission)
+	Timer:CallEvery(9, function ()
+		if not mission.pirate then return true end
+
+		if Game.player:DistanceTo(mission.pirate) <= 50 then
```
Web-eWorks, line 313: This function ideally needs a Comms message when the player closes to within ~5km informing the player that they need to fly close enough to transfer cargo. It's otherwise non-obvious how the player should complete the mission, and even after skimming through the code I found myself having to double-check the completion condition.

Additionally, a 50m distance can fail for some ship types (dsminer, malabar) especially if the player is also in a larger ship. The *proper* solution here is to add the player and the target's clip radius to the distance, but that's not currently exposed to Lua. Failing that, increasing the distance to e.g. 100m should work correctly for all ships.

data/modules/Scoop/Scoop.lua
```diff
@@ -0,0 +1,837 @@
...
+end
+
+Event.Register("onCreateBB", onCreateBB)
+Event.Register("onUpdateBB", onUpdateBB)
+Event.Register("onShipEquipmentChanged", onShipEquipmentChanged)
+Event.Register("onShipDocked", onShipDocked)
+Event.Register("onShipUndocked", onShipUndocked)
+Event.Register("onShipHit", onShipHit)
+Event.Register("onShipDestroyed", onShipDestroyed)
+Event.Register("onCargoDestroyed", onCargoDestroyed)
```
Web-eWorks, line 827: I'd recommend adding a function that reacts to the mission cargo being jettisoned from the player's hold with `Event.Register("onJettisoned")`. I leave it at your discretion whether that is grounds for failing the mission or simply a strongly-worded CommsMessage encouraging the player to scoop it back up.

data/modules/Scoop/Scoop.lua
```diff
@@ -0,0 +1,837 @@
...
+			planet            = ad.planet,
+			id                = ad.id,
+			debris            = debris,
+			amount            = #debris,
+			reward            = ad.reward,
+			due               = ad.due,
+			return_to_station = ad.return_to_station,
+			deliver_to_ship   = ad.deliver_to_ship,
+			pirate            = ship,
+			pirate_label      = ad.pirate_label,
```
Web-eWorks, line 440: Unless I'm horribly missing something, the ship you deliver goods to isn't actually a pirate (doesn't attack you, stumble when he walks, or smell like rum) and I don't think anything in the Mission Support Framework that is our ad-hoc class setup actually needs this ship to be referred to with a variable named `pirate`.

It's a bit of a papercut, but calling this ship something more descriptive would greatly ease the maintenance burden of this module!

Web-eWorks (changes requested):
Thank you @robothauler for all of your hard work on this mission! Other than a few UI/UX-related issues everything looks good and the mission works remarkably well given the amount of old code I've been pulling out from under you during the past few months.

I tested the full lifecycle of an arms dealer mission and from looking at the code the rest of the missions are similar variants on the mission - if there's anything that needs specific testing, please let me know.

I'm sorry it's taken me this long to get around to an in-depth review; with newUI gone and oldUI almost there as well I finally feel like I can get back into the content side of the game. :smile:

data/lang/module-scoop/en.json
```diff
@@ -17,19 +17,19 @@
   },
   "ADTEXT_ARMS_DEALER": {
     "description": "",
-    "message": "PICK UP drifting container."
+    "message": "PICK UP. Prompt recovery of drifting container required."
```
impaktor, line 20: `PICK UP: ` with `:` I suspect? When fixed, I think this text improvement commit can be squashed into your scout.lua

robothauler:
I squashed the commit history a bit. #5110 is mandatory for the negative mission reward (pay for the valuable info).

impaktor:
After review, the last 3 or 4 commits aught to be squashed into the "New Scoop mission" commit, if possible, as they look (to me, after a quick glance) to just add on further fixes to this original commit.

Web-eWorks (approved):
This just needs some squashing of the bugfix / minor change commits and I'm good to merge this in. Thank you for all of your hard work hauler!

impaktor:
I squashed and force pushed to your origin, hope that's OK. 

I also took the liberty to make some minor re-wording of the language file. 
Thanks, and sorry for the long review process.

PR #4875, opened 2020-05-07
Commodity name clean-up

WKFO:
Radioactives are mostly useful and should be paid for.
Radioactive waste mostly isn't and... you get the idea.
(I think it has been that way before, maybe it changed during pigui conversion project, idk)

Commodity "hand weapons" are called "small arms" in market but "hand weapons" in system economy data (and, in turn, trading computer). I changed all into "small arms", as it is in the commodity market.

### DIFF
--- a/data/lang/commodity/en.json
+++ b/data/lang/commodity/en.json
@@ -201,7 +201,7 @@
   },
   "RADIOACTIVES": {
     "description": "",
-    "message": "Radioactives"
+    "message": "Radioactive waste"
   },
   "RADIOACTIVES_DESCRIPTION": {
     "description": "This is an artistic description of the commodity. Translate as you deem appropriate or come up with your own description, long or short, its up to you.",
--- a/data/lang/core/en.json
+++ b/data/lang/core/en.json
@@ -489,7 +489,7 @@
   },
   "HAND_WEAPONS": {
     "description": "",
-    "message": "Hand weapons"
+    "message": "Small arms"
   },
   "HARD_CAPITALIST": {
     "description": "",


impaktor:
I'm not sold on the last modification.

When I worked with the commodity prices, I discovered the "Hand weapons" - "small arms" discrepancy in the code (and Frontier), but after speaking to @fluffyfreak I/we decided it was best to keep it as it was. At least, that's how I remember it then.

Web-eWorks:
I'll side with Small Arms under one condition - if Battle Weapons are intended to be things like APCs and rocket launchers, machine guns, power armor, etc. then it makes sense to have the definition between small arms like pistols, carbines, knives, etc and the bigger, more prohibited stuff that's intended to be outfitting a PMC or rebellion.

If "Battle Weapons" is the small stuff like pistols, knives, etc. then why do we need hand weapons? Knives are subsumed into consumer goods already, and I highly doubt that a system government blocking all form of firearms is going to let shipments of medieval poleaxes in (although that's its own pickle, as when firearms are Battle Weapons, there's almost nothing left for the Hand Weapons category...)

WKFO:
In game description for battle weapons:
`Heavy military-grade weapons, like armoured hover tanks, and surface-to-space missiles. Trading in them is usually a governmental monopoly. On the black market, however, these can often fetch a hefty sum of money.`

In game description for hand weapons/small arms:
`Providing insurance beyond the long arm of the law, hand weapons are essential for space traders in far-out systems. However, many governments see them as a threat to their own control over their population and strictly control their trade and possession.`

PR #4877, opened 2020-05-09
Escape pod for the scoop mission

nozmajner:
This PR adds an escape pod model intended for the scoop mission of @robothauler 
It has three LoDs: 7500, 3742, 1290 tris, Collision mesh is 444 tris.
Has DDS textures and one pattern. Most of the textures are 512x1024, the normal map is 1024x2048. The source textures are 2048*4096, so there's some room for better quality later.
Especially because surface details are baked from decals.
No animation, and only tags for navlights (which don't light up yet, since cargo doesn't have a gear down state).
It has some very small RCS thrusters on the texture, but no tags for it.

If you want to see it in-game, replace the misc/cargo/ dir  with it, and rename the escape_pod.model to cargo.model. Then you can jetisson any cargo to see it.

It has an array of solar panels and radiators, a high-gain collapsible articulated antenna and an inflatable habitat sphere for some extra leg room (Hence it's affectinatelly called "The loo of the Little Prince" by SAR personell"). The top part has a size 0 docking collar.
It is about 5 meters long, and the habitat sphere has a 3.25m diameter. The main cylinder has a 1.5m diameter, and about 1m diameter room inside.

[The model on sketchfab. ](https://sketchfab.com/3d-models/escape-pod-1dcd03840fd8443ab795a0dcf1267f66)
(These screenshots and the sketchfab model have the full res textures)
modelviewer-20200508-221016
modelviewer-20200508-221029
screenshot-20200509-105925
screenshot-20200509-105954
screenshot-20200509-110056
screenshot-20200509-110113


impaktor:
@nozmajner what's the status on this? Are we/you waiting further coordination for how it will be included on the code side?

nozmajner:
I have to fix the texture file capialization issue

fluffyfreak:
> I have to fix the texture file capialization issue

Once you've fixed that just merge it or ping me to do so.

nozmajner:
Gliese fixed it with the latest commit if I'm correct

PR #4878, opened 2020-05-10
LuaMetaType + LuaObject 2.0

Web-eWorks:
If you've not been following my progress on IRC, let me introduce this PR before I get down to the changes.

For quite some time, I've been stymied in my efforts to make sweeping changes to the structure of the codebase by several factors. The largest of these factors has been (obviously) the code hierarchy and the extreme degree of interconnection between seemingly disparate concerns.

This PR doesn't do anything about that.

This PR is aimed at fixing the *second* largest factor, that being the reams of boilerplate one must write to expose a struct with five public members to Lua. Given a struct like so:
```C++
struct MyAwesomeStruct {
    bool aBooleanValue;
    SomeLessAwesomeStruct sadStruct;
    int anIntegerValue;

    void ThisDoesSomething();
    float AwesomeGetter();
};
```
The code required to bind this structure to Lua is so long I have to put it in a foldout for readability (and I only wrote 15% of the code required in this situation!):

<details>

```C++
// ================================================================

static int l_set_aBooleanValue(lua_State *l)
{
    MyAwesomeStruct &obj = LuaObject<MyAwesomeStruct>::CheckFromLua(l, 1);
    bool value = LuaPull<bool>(l, 2);
    obj.aBooleanValue = value;
    return 0;
}

static int l_get_aBooleanValue(lua_State *l)
{
    MyAwesomeStruct &obj = LuaObject<MyAwesomeStruct>::CheckFromLua(l, 1);
    LuaPush<bool>(l, obj.aBooleanValue);
    return 1;
}

// Rinse and repeat for the rest of the parameters...
// - two methods for each two-way value
// - a larger method for each function call
// Six lines of boilerplate minimum for each method

// ================================================================

template<>
const char *LuaObject<MyAwesomeStruct>::s_type = "MyAwesomeStruct";

template<>
void LuaObject<MyAwesomeStruct>::RegisterClass()
{
    static luaL_Reg l_methods[] = {
        { "SetABooleanValue", l_set_aBooleanValue },
        { "GetABooleanValue", l_get_aBooleanValue },
        // + 2 entries for each value
        // + 1 entry for each function
        { NULL, NULL }
    };

    LuaObjectBase::CreateClass(s_type, l_methods, 0, 0);
}
```
</details>

Suffice to say, a ten-line structure takes about 150 lines of tedious, tiresome boilerplate to bind to Lua. So much for rapid prototyping, eh?

Not all hope is lost though. I've spent the past year thinking, and the past week bashing my head against C++'s templates, and managed to completely invert the paradigm used to create Lua bindings. Behold all twenty lines needed to bind our struct:
```C++
template<>
const char *LuaObject<MyAwesomeStruct>::s_type = "MyAwesomeStruct";

template<>
void LuaObject<MyAwesomeStruct>::RegisterClass()
{
    static LuaMetaType<MyAwesomeStruct> metaType(s_type);

    metaType.StartRecording()
        .AddMember("aBooleanValue", &MyAwesomeStruct::aBooleanValue)
        .AddMember("sadStruct", &MyAwesomeStruct::sadStruct)
        .AddMember("anIntegerValue", &MyAwesomeStruct::anIntegerValue)
        .AddFunction("ThisDoesSomething", &MyAwesomeStruct::ThisDoesSomething)
        .AddMember("awesomeGetter", &MyAwesomeStruct::AwesomeGetter)
        .StopRecording();

    LuaObjectBase::CreateClass(metaType);
}
```

Looks completely different, right? "What's the catch?", I'm sure you're asking...

Other than another 500 lines of templates needed to bind pointer-to-members, the 'catch' is very simple - because this depends *heavily* on the `LuaPush` / `LuaPull` mechanism, you'll have to make sure appropriate overloads of `pi_lua_generic_push` et. al. are available for all of the types (including integral types like `uint8_t`) you're binding to Lua, or you might get reams of template deduction failure errors from the compiler. Compile times for the `Lua*.cpp` files might be slightly increased as well, but I'm more than happy to trade hours of programmer time for less than a minute of compile time.

Runtime performance is (almost) completely unaffected - the `__index` metamethod is *slightly* more complicated in some spots, but now much less complicated in others. There's a `__newindex` metamethod, but it's impact is completely negligible... because we had no mechanism to set attributes before.

Another side benefit is that this simplifies working with PropertyMaps as well - once a property is defined, you can simply call `object.hyperdrive_cap = 4` instead of `object:setprop('hyperdrive_cap', 4)`. Obviously, the old methods are still alive and well. (To avoid confusion, a name not present in the property map will instead run through the attributes first... there's some room for improvement here, like assigning to the property map if no attribute with that name was found.)

#### What if I need custom argument checking?

If you need the flexibility that writing every method manually gave you (for e.g. argument-dependent behavior), fear not - you can do that here too! Lambdas and regular functions are supported in all cases - unlike regular `lua_Cfunction`s, the `self` argument is automatically provided for you to reduce the boilerplate needed.

Support for binding 'regular' functions without a self argument is limited right now, though it's certainly possible through the existing `luaL_Reg` mechanism.

```C++
metaType.AddMember("sadStruct", [](lua_State *l, MyAwesomeStruct *st) -> int {
    if (lua_gettop(l) > 2)
        // setter ...
        auto param = LuaPull<MyParameterType>(l, 2);
    else
        // getter ...
});
```

### Changes:

- Add a new system to bind member functions and data types to Lua without requiring the programmer to spend hours writing boilerplate to pull arguments and data out of Lua.

- Replace our existing `__attribute_x` system for attrs with a separate `attrs` table, (slightly) simplifying lookup. Support attribute setters via a `__newindex` metamethod (attribute handlers will be called with an extra parameter corresponding to the setter).

- Don't lean on `luaL_newmetatable` anymore, explicitly store metatypes in the registry table `LuaMetaTypes` for better namespacing and reliable lookup.

### DIFF
--- a/src/lua/Lua.cpp
+++ b/src/lua/Lua.cpp
@@ -57,6 +57,8 @@ namespace Lua {
 	void InitModules()
 	{
 		PROFILE_SCOPED()
+		lua_State *l = Lua::manager->GetLuaState();
+
 		LuaObject<PropertiedObject>::RegisterClass();
 
 		LuaObject<Body>::RegisterClass();
@@ -110,7 +112,6 @@ namespace Lua {
 		SceneGraph::Lua::Init();
 
 		// XXX load everything. for now, just modules
-		lua_State *l = Lua::manager->GetLuaState();
 		pi_lua_dofile(l, "libs/autoload.lua");
 		lua_pop(l, 1);
 
--- a/src/lua/LuaCall.h
+++ b/src/lua/LuaCall.h
@@ -1,0 +1,103 @@
+// Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+#pragma once
+
+/*
+	This file implements a generic method to call a C++ member function with
+	arguments passed in from Lua. It's intended to back a generic method of
+	binding code to Lua with minimal boilerplate needed.
+*/
+
+#include "Lua.h"
+#include "LuaPushPull.h"
+
+// Backend to bind call a C++ member function with arguments from Lua.
+// Parameter `index` points one-behind the first argument, where the object
+// pointer should traditionally be
+
+template<class T, typename Ret>
+Ret pi_lua_multiple_call(lua_State *l, int index, T *ptr, Ret (T::*fn)())
+{
+	return (ptr->*fn)();
+}
+
+template<class T, typename Ret, typename Arg1>
+Ret pi_lua_multiple_call(lua_State *l, int index, T *ptr, Ret (T::*fn)(Arg1))
+{
+	auto arg1 = LuaPull<Arg1>(l, index+1);
+	return (ptr->*fn)(arg1);
+}
+
+template<class T, typename Ret, typename Arg1, typename Arg2>
+Ret pi_lua_multiple_call(lua_State *l, int index, T *ptr, Ret (T::*fn)(Arg1, Arg2))
+{
+	auto arg1 = LuaPull<Arg1>(l, index+1);
+	auto arg2 = LuaPull<Arg2>(l, index+2);
+	return (ptr->*fn)(arg1, arg2);
+}
+
+template<class T, typename Ret, typename Arg1, typename Arg2, typename Arg3>
+Ret pi_lua_multiple_call(lua_State *l, int index, T *ptr, Ret (T::*fn)(Arg1, Arg2, Arg3))
+{
+	auto arg1 = LuaPull<Arg1>(l, index+1);
+	auto arg2 = LuaPull<Arg2>(l, index+2);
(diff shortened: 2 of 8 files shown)


impaktor:
Nice write up :+1: 

Dunno if @laarmen would be interested in seeing what's going on in his favourite place: lua/C++ integration in Pioneer.

laarmen:
Funny, I knew just by looking at the title that you'd tag me on it 😋

I'll take a look tomorrow if I can remember.

laarmen:
Ok, just reading the write-up it seems like the logical step beyond
LuaTable, which makes me damned happy. That's exactly what the generic push
pull functions were designed for.

Great work! I look forward to reading the code.

Web-eWorks:
Merging this, it's been a bit and no objections have come up!

Issue #4866, opened 2020-04-26
4 Fala system in Sector (19, -6, -2)


Gliese852:
![Screenshot from 2020-04-26 15-07-35](https://user-images.githubusercontent.com/18342621/80307268-9f132200-87d0-11ea-8f01-67c56bb9f5a9.png)

### Observed behaviour
When searching for the Fala system, it turned out that there are 4 or even more.

### Expected behaviour
Uniquie name for each star system.

### Steps to reproduce
Slide to sector (19,-6,-2), search Fala

__My pioneer version (and OS):__
Current master 8c1a495f6f046f8e8432720bd3b5d683b6d7b6c6

impaktor:
This has been reported before, and is not a bug. 

Consider: 
- there are two major university cities called Cambridge (UK / US)

fluffyfreak:
Suggestions and improvements are always welcome to the random name generator though

fluffyfreak:
The system generation of names is dictated by https://github.com/pioneerspacesim/pioneer/blob/master/src/galaxy/SectorGenerator.cpp#L17

Which has a collection of sub-strings it combines to generate system names.

fluffyfreak:
Does anyone know how/why those fragments were chosen? Is it based on the phonetics/something of existing system naming?

impaktor:
I suspect it comes from original Elite, where they constructed sentences by combining such similar strings consisting of vowels and consonants.

One alternative way is to do it like a markov chain, where for each letter, (or letter pair) gives a vector of probablies of each next letter in the alphabet, and space/stop (sum to 1), which can be trained on some text corpus (e.g. english wikipedia in raw text). I did this a few years back but didn't think the results were that impressive, although I might have used a too small text sample for getting the letter frequencies. 

(Maybe training on geographical locations would be better, but on the other hand, I doubt it)

impaktor:
But as I said, I really don't think this is a bug. Fala was the 134th Glorius Emperor of the Stallman faith, so only natural many systems have taken name from him.

Zireael07:
Cambridge example, however, isn't that relevant as they are far from each other. Those four Fala, however, are in a single sector (!) which makes it especially egregious.

[...]

PR #4879, opened 2020-05-12
Load stars from the HYG Star Database

Web-eWorks:
This PR closes #4045 - with many thanks to @ecraven for the initial implementation in Scheme!

This is a long PR with many changes along the way, considering it's been in development for half a year now (*cough* due to my own laziness *cough*). I'll summarize quickly here, but the commit descriptions are where the meat of the PR is:

- Moved some functionality to the `core/` module (compression, IniConfig, etc.)
- Added external libraries to `contrib/`: `csv-parser` to simplify reading the HYG database, `nonstd::string_view` for an easier way to pass around strings without copying them, `argh` to replace our ad-hoc and ugly command line parsing
- Added some more performance information and restructured the way it's displayed
- Made the performance info widget available in the main menu
- Moved the modelcompiler run step into its own target - `make -C build build-data` invokes this as well as the HYG database fetch and parse steps

A couple of notes:
- Fetching the HYG database requires a `curl` instance to be available on the command line (sorry Windows users!)
- About 1200 stars from the DB aren't imported into the game; this is probably an oversight on my part, if someone wants to add logic to bind the remaining spectral classes to Pioneer's star types, feel free!
- There's tons of data in the DB that we're not using; the absolute magnitude and star color temperature are two that I've added to the parser but don't have anything in-game to hook them up to.
- The new modelcompiler step isn't currently run in the `build-travis.sh` script (easy to fix).
- It's not really feasible to remove `02_local_stars.lua`, as you wind up with an empty bubble about ~80ly around Sol. Whether this is intended behavior is up for discussion (as well as whether the galaxy should be a *lot* less dense), but for now it's staying.
- System name generation is *really* bad. I did some cursory searching and within about 200ly we have 10+ duplicate systems all named "Solia". That's something to improve on next time though...


impaktor:
Quick first note: first commit aught to also add an entry to AUTHORS due to the csv import lib. 

Last issue is already tracked in #4866 

Will test later.

Web-eWorks:
Thanks for catching that, I didn't even notice that I forgot to add entries to AUTHORS for csv-parser and string_view_lite!

Web-eWorks:
Because I'm my own worst enemy, I pushed the logging changes to this branch as well. I'll split that off if people don't want to review both of them at the same time...

impaktor:
Have compiled, confirm that custom systems went from 500 100,000 in debug info, after `make -C build build-data` 1. (that needs to be documented in COMPILING), 2. Confed faction has moved so that needs fixing. i.e. now only the 3(?) custom CIW worlds are left, in a sea of Federation systems, + swarm of CIW systems very far away from CIW homeworld.

Nice progress!

impaktor:
I don't think we need to support procedural assignment of systems to factions. We decided some time ago to not have procedural factions, at which point the code generating them was removed, but only after its output had been dumped as +100 faction files in `data/factions`.

This step was the first taken in the attempt to reduce factions from hundreads, down to only three: SolFed, CIW, and Haber, (plus "independent"). The idea is that factions should be hand written, with lore, illegal-goods rules, positions, wiki, ship manufacturers, etc.

In summary: I wouldn't mind if each faction explicitly listed which systems (or sectors) it controlled, and then that's what's used in sector view. Either way, currently with this PR it sort of breaks CIW. I'll hold off on further review until some dicision / action on how to fix that has been taken.

Web-eWorks:
@impaktor as far as I know, nothing in this PR has touched the faction code; if loading custom systems is causing issues with faction alignment, the best solution I see is to temporarily disable the `build-hyg-data` step until the faction code can be brought up to spec.

While it's certainly an issue that the factions are not showing up as intended, I think that's out of scope for this PR.

impaktor:
Also: starting positions are messed up. See for instance New Hope (in orbit), or Bernard Star (crashes the game)

Web-eWorks:
*Sigh...* Looks like this needs some more work then. I'll split the QOL changes into its own PR and this can bitrot for a bit while I work on something else.

impaktor:
Sorry if I've come off as negative, I really like moving to using the Hueygens star catalogue, so I really appreciate the work you've done. I agree that merging the code - but leaving it inactive - could be a good idea, to prevent bitrot.

Brianetta:
I stand by every word. Alpha Centaurus is s great example; Toliman (as was; the stars' names have been moved around IRL since the Pioneer project began) is at such a threshold distance from Proxima that it is unclear whether to call it a separate system. In the galactic core distances of that order are the norm.

Incidentally, I'm no sunk cost fallacist. If something is better or more fun, use it, even if what it replaced cost a lot of time and effort.

Web-eWorks:
Yeah, this PR was mostly an effort to finish up ecraven's work and get it going; I'm going to take a break from this and work on another section of the code instead while I figure out the best way to handle it.

From what I can see, there are a few major tasks that need to be done before this can be merged:
- [ ] Process star coordinates to match Pioneer's coordinate system
- [ ] Merge close (how close?) stars into binary / trinary stars
- [ ] Add an override / modify mode to CustomSystem to allow easier editing of imported stars
- [ ] Add a "star blacklist" to the parser / importer to filter out specific cases of incorrect stars

src/galaxy/CustomSystem.cpp
```diff
@@ -3,18 +3,27 @@
 
 #include "CustomSystem.h"
 
+#include "EnumStrings.h"
 #include "Galaxy.h"
+#include "Json.h"
+#include "LZ4Format.h"
```
Gliese852, line 9: this line causes me a compilation error

impaktor:
Probably not of interest, but I imagine this has (pytong) code for parsing HYG data base (although projecting to 2D sky, I assume),
https://github.com/eleanorlutz/worldstars_atlas_of_space

syonfox:
Based #6021  I made a calculator tool that does the -90 ra rotation from j2000

https://foxi.eapi.space/j2000_to_pioneer_sector_pos_calculator

```js
    let pioneer_equ = a.approx_j2000_to_pioneer_equatorial(ra, dec)
    const cartesian = a.calculateCartesian(pioneer_equ.ra, pioneer_equ.dec, dist);
    const sectorPos = a.get_sector_pos(cartesian.x, cartesian.y, cartesian.z);
```

Seems to be the working code  https://github.com/syonfox/eapi-space/blob/main/_foxi/web/js/lib_astro_crs.js#L228

Lots of stuff does not work in that file and needs more clean up

PR #4881, opened 2020-05-15
Improve Core Functionality

Web-eWorks:
This was split from #4879, as that PR needs some more time to cook.

I've upgraded our logging functionality to remove the need for `OS::RedirectStdio` as well as adding better formatting and more logging levels. I've also added `argh` to improve our command-line argument parsing (actually upgrading the parsing is going to happen in a separate PR, this just adds the module).

- Moved some functionality to the core/ module (compression, IniConfig, etc.)
- Added external libraries to contrib/: nonstd::string_view for an easier way to pass around strings without copying them, argh to replace our ad-hoc and ugly command line parsing
- Added some more performance information and restructured the way it's displayed
- Made the performance info widget available in the main menu
- Moved the modelcompiler run step into its own target - make -C build build-data invokes this.


Web-eWorks:
In this PR I've also included fixes for LuaMetaType outliving the lua instance, removed the old and unneeded `atexit` handler for LuaObject, fixed the handling of const member functions, and *finally* got the clang-format check working again!

Web-eWorks:
As there hasn't seemed to be any pushback on this, merging now!

PR #4882, opened 2020-05-18
Prevent unwanted font face changes

### LINKED ISSUE #4807: UI font suddenly changes to smaller font size
### Observed behaviour
While in flight-ui, when using the system overview window, the font used there and in the reticule and planetary info suddenly changes to a smaller font size, which is harder to read.

Similar thing happens when in main menu - load game dialog: while hovering the list of saved games with the mouse, eventually the font used in the list changes to a smaller font size.

### Expected behaviour
Font stays at the original, bigger size

### Steps to reproduce in flight-ui
1. Start at Mars, undock.
2. Open system overview window
3. Alternatively
  - click on "Toggle showing stations", or
  - select Earth, then Moon
4. Observe change of font size

current master e255171
OS & DE: Debian 9.12 (as well as 10.2) with Gnome
pioneer gfx config: fullscreen 1920x1080

This bug appears to have been introduced in 8ff50803, which is in fact related to font management. The version before, 105d06e3, does not have this bug.

Gliese852:
Before PR -> After PR
Screenshot from 2020-05-18 13-37-17
Screenshot from 2020-05-18 13-36-36

When a character appears in the text, that is not in the default font face, it is looked up in the spare faces of the font. Moreover, the default range also began to be baked from the spare face. I made that *only* characters that are not in the default face are baked from the spare face.

To visually see what is this all about, start a new game and click on Toliman in the sector map. In it's description there is an "α" (alpha) symbol, this leads to a complete change of the font face. This PR fixes this issue.

Fixes #4807

### DIFF
--- a/src/pigui/PiGui.cpp
+++ b/src/pigui/PiGui.cpp
@@ -121,17 +121,31 @@ Instance::Instance() :
 {
 	// TODO: clang-format doesn't like list initializers inside function calls
 	// clang-format off
+
+	// The main face of the font should go first in the list, because:
+	//
+	// - during the first initialization, only the first face is used to bake the
+	// font ( see Instance::AddFont )
+	//
+	// - when a non-standard glyph is found in the text, a search is started in
+	// the faces, and the faces are scanned in the order of this list
+	// ( see Instance::AddGlyph )
+	//
+	// - the default imgui glyph range ( 0x20 .. 0xFF ) is always baked from the
+	// first face, that has some ranges defined
+	// ( see Instance::BakeFont )
+
 	PiFont uiheading("orbiteer", {
+		PiFace("Orbiteer-Bold.ttf", 1.0), // imgui only supports 0xffff, not 0x10ffff
 		PiFace("DejaVuSans.ttf", /*18.0/20.0*/ 1.2),
-		PiFace("wqy-microhei.ttc", 1.0),
-		PiFace("Orbiteer-Bold.ttf", 1.0) // imgui only supports 0xffff, not 0x10ffff
+		PiFace("wqy-microhei.ttc", 1.0)
 	});
 	AddFontDefinition(uiheading);
 
 	PiFont guifont("pionillium", {
+		PiFace("PionilliumText22L-Medium.ttf", 1.0),
 		PiFace("DejaVuSans.ttf", 13.0 / 14.0),
-		PiFace("wqy-microhei.ttc", 1.0),
-		PiFace("PionilliumText22L-Medium.ttf", 1.0)
+		PiFace("wqy-microhei.ttc", 1.0)
 	});
 	AddFontDefinition(guifont);
 	// clang-format on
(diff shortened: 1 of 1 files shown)


Gliese852:
I think it's worth explaining in more detail the logic of what is happening there.

The font contains several faces:
https://github.com/pioneerspacesim/pioneer/blob/97f06419c9333f8a389112b0efdede3d9944502f/src/pigui/PiGui.cpp#L131-L135
it can be seen that the main font is at the _end_ of the vector. How does he become basic? A range is added to it (when a glyph is added, a range is created for it if it does not fall into an existing one, and in the beginning we don't have any ranges at all)
https://github.com/pioneerspacesim/pioneer/blob/97f06419c9333f8a389112b0efdede3d9944502f/src/pigui/PiGui.cpp#L202
In the next function, the font texture is created, apparently
https://github.com/pioneerspacesim/pioneer/blob/97f06419c9333f8a389112b0efdede3d9944502f/src/pigui/PiGui.cpp#L347-L386
It iterates over all faces of the font, in turn (note that the main face comes last)
https://github.com/pioneerspacesim/pioneer/blob/97f06419c9333f8a389112b0efdede3d9944502f/src/pigui/PiGui.cpp#L352
but it ignores faces without ranges, and in the beginning we have a range _only for the main face_.
https://github.com/pioneerspacesim/pioneer/blob/97f06419c9333f8a389112b0efdede3d9944502f/src/pigui/PiGui.cpp#L359
А default range is added to each face
https://github.com/pioneerspacesim/pioneer/blob/97f06419c9333f8a389112b0efdede3d9944502f/src/pigui/PiGui.cpp#L363
that's what this range is
https://github.com/pioneerspacesim/pioneer/blob/97f06419c9333f8a389112b0efdede3d9944502f/contrib/imgui/imgui_draw.cpp#L2225-L2234
If after rendering, the dear ImGui lacks a glyph, this function is called (I omitted the details, there are a lot of calls, as I understand it)
https://github.com/pioneerspacesim/pioneer/blob/97f06419c9333f8a389112b0efdede3d9944502f/src/pigui/PiGui.cpp#L161-L184
It checks all faces of the font for the presence of a given glyph, and also sets the flag that fonts need to be rebuilt
https://github.com/pioneerspacesim/pioneer/blob/97f06419c9333f8a389112b0efdede3d9944502f/src/pigui/PiGui.cpp#L176-L181
Аnd now we have a range not only in the main face, but also in some other. When baking, it will go _before_ the main face, and now the default range will be added to it. When it comes to the main face, the range will already be occupied.

Therefore, I simply rearranged the main face to the top of the list, and the default range will always be taken from it, and everything that cannot be found in it will be taken from other faces.

Web-eWorks:
@Gliese852 can I ask you to *please* transcribe that excellent write-up to function comments? The overall architecture is something I've had to reverse-engineer myself when doing imgui upgrades, and I'd really really appreciate having this overview as function comments so future maintainers (and future me!) don't have to recreate it.

Gliese852:
@Web-eWorks Done, tried to adapt this information, because in this topic it is written how it works now, and in the comments in the code, I wrote how it will work after the patch.

PR #4887, opened 2020-06-17
Update MSVS2019 Project (#4884)

### LINKED ISSUE #4884: Visual Studio Project Doesn't Build
I know about the Visual Studio project file not building.

The CMake version is probably ok.

I'm just making this issue to force myself to find the time to update it.

The-EG:
Add subproject for fmt
Add include directory for contrib/fmt/include to most subprojects

This will get the project to build for x64, but Pioneer will not run yet. There are a couple of other issues that aren't related to the actual project files that need to be fixed for that. If you want to run it to test it out, set the subsystem to console in the linker options, and set SIZE_FMT equal to "%zu" (libs.h: 44-48).

Additionally, x86 building has an issue in FrameId.h: it seems that size_t and uint32_t are being evaluated to the same value and causing a duplicate declaration. 

Fixes #4884


The-EG:
It's also worth noting: I'm not sure why it's actually building successfully against the VS2017 pioneer-thirdparty libs...I guess there wasn't an ABI change between v141 and v142. 

But, that's why the whole program optimization is off for the new fmt project in the release configuration. If that is turned on, it triggers the use of /LTCG which then causes the linker to fail on linking libogg_static.lib, complaining it's compiled with an old version(v141).

The-EG:
Fixed a few things for PreRelase and Profile builds. All (Release, PreRelease, Profile, Debug) should all now build for x64.

x86 may require some tweaks once the FrameId.h issue is addressed.

impaktor:
This is up to @fluffyfreak to OK / merge, if it's all done.

PR #4889, opened 2020-06-18
Cleanup Sector Map View

The-EG:
This is some cleanup I meant to do to the sector map view after I added the hyperjump planner...better late then never?

image

The hyperjump planner has been integrated into the sectormap module. There was no real reason to have them separate other than it made it easier for me to write it while the sector map was being ported to ImgUI.

Everything is now on separate tabs, which should be closer to some of the original mockups and I think it ends up being less cluttered/crowded.

Info:
image

Factions:
image

Search:
image

Hyperjump Route:
image

Settings:
image

Clicking the tab that is already selected will hide the window. This is similar to hiding all of the sectormap ui with the tab key (this still functions for now), but the zoom controls and tab buttons are still shown:
image
Click a tab to show it again.

Because it's using less screen space, and it's easy to hide, I changed the window background to something darker. It was difficult to read at times with the darker background.

Todos:
- [ ] It looks like I need to cleanup the sectormap lua a bit...indenting is strange in some spots
- [ ] Translation strings - some of the sectormap stuff is hard coded
- [ ] Add Jump in the route planner still needs a better icon
- [ ] Settings: 'Remove jump when completed' should probably be changed to 'Remove jump from route when completed' 
- [ ] Should 'Remove jump..' setting also be persistent like the other settings?
- [ ] Enable font scaling? (I noticed some of the new UI now does this in places...)

Also, the fact that the sector map is rotated by holding the right mouse button, but the system map now uses the middle button drives me bonkers! I don't really mind which, but I feel like they should use the same button.


Gliese852:
Please see #4852

impaktor:
Yeah, this will clash hard with @Gliese852's branch. Better if @The-EG could review and test #4852 and then add wanted changes after it's merged, or collaborate with @Gliese852 for additional commits.

Thanks for the effort though!

The-EG:
@Gliese852 & @impaktor Fair enough! I looked for anything open first, but didn't find that one for some reason.

Gliese852:
@The-EG It would be very interesting to know your opinion about #4852 !

The-EG:
@Gliese852 I'll have to get your branch and test it out...but on first glance, it looks like what you're doing is much more in depth then what I was trying to accomplish, which was just to unclutter it and get it looking more towards what nozmajner's mockups look like. 

I'll reply there, so it's all in one spot.

PR #4890, opened 2020-06-19
Update SIZET_FMT and catch fmt::system_error exception

### LINKED ISSUE #4888: fmt can't write to stderr/stdout on Windows
### Observed behaviour
Pioneer built on Windows (/subsystem:Windows) fails to run with an unhandled `fmt::v6::system_error` exception within `Log::Logger::WriteLog`. This appears to be due to the usage of `fmt::print(stderr` and `fmt::print(stdout` when `stderr`/`stdout` are not open/valid in a Windows application.

Building as a console application allows Pioneer to run, but with an additional console window.

### Expected behaviour
Pioneer should be able to run when built for Windows and output any output to output.txt.

### Steps to reproduce
Build Pioneer for x64 Windows
Run pioneer-x64.exe

Current master (https://github.com/pioneerspacesim/pioneer/commit/91d1caa8a647b512cbca23e58c975659efe0c6d9)
Windows 10 (x64)

The-EG:
Catch and silently ignore fmt::system_error exception when using fmt::print on stdout and stderr. This will occur if there is no console (ie. for Windows builds).
Set SIZET_FMT to "%zu" on all builds now, fmt does not understand the MSVC "%Iu" format.

This and https://github.com/pioneerspacesim/pioneer/pull/4887 should get Pioneer building and running for x64 again in Visual Studio.

Fixes #4888

### DIFF
--- a/src/core/Log.cpp
+++ b/src/core/Log.cpp
@@ -84,13 +84,17 @@ void Log::Logger::LogLevel(Severity sv, nonstd::string_view message)
 void Log::Logger::WriteLog(Time::DateTime time, Severity sv, nonstd::string_view msg)
 {
 	std::string &svName = s_severityNames.at(sv);
-
-	if (sv <= Severity::Warning) {
-		fmt::print(stderr, "{}: {}", svName, msg);
-	} else if (sv <= m_maxSeverity) {
-		fmt::print(stdout, "{}", msg);
-		// flush stdout because it might have a different cache size than stderr
-		fflush(stdout);
+	try {
+		if (sv <= Severity::Warning) {
+			fmt::print(stderr, "{}: {}", svName, msg);
+		} else if (sv <= m_maxSeverity) {
+			fmt::print(stdout, "{}", msg);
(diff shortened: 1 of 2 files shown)


Web-eWorks:
This certainly works, though I'd rather `#ifdef` and disable writing to stdout / stderr on windows instead of adding a try/catch overhead every time we're logging. Not a major issue, just a little nitpick!

The-EG:
Comments from chat (for documentation purposes):
<The-EG> yeah, I had toyed around with different things on that. you could set one manually i guess (like _CONSOLE) but I couldn't come up with one that's automatically set/not set across the board
<The-EG> the other idea I had was testing if stderr was valid on init, and then only doing it if it is...but that's messy too
<The-EG> also, with /subsystem:windows you can still have a console. If i run it from bash (git env) i get the console output
<The-EG> so excluding it completely with a preprocessor directive seemed a bit like using a sledge hammer
<The-EG> oddly enough, running it directly from CMD didn't work though. I think it detects it's a windows program and forks it off instead

PR #4891, opened 2020-06-21
Surface impact alerts

WKFO:
Adds HUD alerts for dangerous planetary (and stellar?) approaches. Has 3 modes.

- "DESCENT RATE": If the player is flying close to the asteroid/planet/star and has a dangerously high descent rate, a HUD warning appears with a short audio tone. If ignored for longer than several seconds (descent rate not reduced), it usually guarantees a crash (especially if there is no atmosphere) or maybe a soft bump.

- "IMPACT TRAJECTORY": If the player is far away from the body but is unable to decelerate before hitting the ground. Usually does not require immediate action, and disappears once the velocity vector is pointed away from the body. No audio tone.

- "IMPACT IMMINENT": If the player is far away from the body but is on a dangerous impact trajectory. May appear if IMPACT TRAJECTORY is ignored for too long. The player is unlikely to survive unless immediate action is taken, and even then, it is hard to recover from. No audio tone, because we will let people die in peace.

Those are all calculated independently from the braking distance gauge, because it takes the regular ship velocity while I take approach velocity only. They have different purposes.

I did a lot of play-testing myself, but others should probably do some too.
The mathematics are pretty simple motion formulas, but if you can't read my handwriting I will gladly explain. They just look long because variable names are long.

descent

impact

imminent
(The last image might look survivable but no, you can't divert quickly enough. I tried.)


WKFO:
Found a bug where IMPACT IMMINENT alert shows when you are passing by a planet at high speed. I know how to solve it, but it is past midnight so I will solve it tomorrow probably.

Web-eWorks:
@WKFO I'd definitely love it if you'd include a link to a breakdown of the motion equations used, if only for future maintainers to be able to understand which specific formula you're using (and to give future contributors a stepping-off point if they want to improve it!)

impaktor:
> I'd definitely love it if you'd include a link to a breakdown of the motion equations used,

Preferably this should be included as comment and/or commit message as well.

WKFO:
The bug is fixed, I made a few changes.

Here are the horribly drawn visualizations of what I tried to do. I think the equivalent of most equations can be found: 

https://en.wikipedia.org/wiki/Projectile_motion
https://en.wikipedia.org/wiki/Projectile_motion#Maximum_height_of_projectile

And I have this bookmarked which helps you predict orbits by velocity + position:
https://space.stackexchange.com/questions/1904/how-to-programmatically-calculate-orbital-elements-using-position-velocity-vecto/1919#1919

Descent rate:
descent_rate
If tangential speed is high, it becomes a warning that tells you getting any lower any faster is dangerous.

Impact trajectory/imminent:
impact

Impact warnings are accurate themselves, only the escape possibility estimations have estimations in them. At this point it is accurate enough, anything more accurate in escape estimations does not earn the player anything (and gives the player less time to react), and also requires complex integrals which might become expensive to calculate every frame and are hard on my brain.

And why would you need an "IMPACT IMMINENT" warning showing up just when impact becomes 100% unavoidable anyway?

However, even then, I did make a few calculations for theoretically improved versions, trying to come closer to maximum physically accurate models.
formm

This is my whiteboard while I was trying to model "approach velocity over time depending on periapsis altitude or initial velocity vector". Another instance was me trying the same thing but using kinetic and gravity potential energy. 

None of them ended up turning into code, because I could as well just go on and simulate every atomic nuclei making up the ship.

WKFO:
Alerts are now red instead of green.

WKFO:
Fixed an issue. The descent rate alert no longer triggers when you force extreme time accelerations when the autopilot is docking.

laarmen:
Oh wow. Lots of graphics, that ought to help understanding it all.

On the subject of having a warning for impacts that are unavoidable: they are indeed relatively useless from a gameplay perspective, but in the real world you'd want them, in a "brace for impact" sort of thing. Sure, the player cannot "brace", but it'd still make sense to have the warnings for immersion.

However, I'm quite partial to not implementing stuff that make my brains hurt, too ;-)

WKFO:
Impact trajectory does not represent immediate danger so I put it as a small notifier on bottom right as suggested. (Sits just above GPS data box)

For some reason UI background color in screenshots are messed up.

traj
danger

impaktor:
Would be good if you could squash the bug fix commits

WKFO:
Squash done!

nozmajner:
A first draft of a impakt trajectory icon I mentioned on IRC
impact_trajectory_01

impaktor:
Is this ready for test/merge?

WKFO:
@nozmajner was making an icon, so it should wait

nozmajner:
Right, I forgot about this, sorry.
How about this one?
[icons_2.zip](https://github.com/pioneerspacesim/pioneer/files/4941901/icons_2.zip)
image
(I thin I prefer this one)



Or this:
[icons.zip](https://github.com/pioneerspacesim/pioneer/files/4941883/icons.zip)
image
A bit clearer though.

WKFO:
Ugh, I will have to fix this.

WKFO:
I did something and en.json turned into a binary file apparently!

WKFO:
Spaces fixed, two conflicts left.

nozmajner:
Up-to-date icons file, pulled from master, added the collision icon to the end.
[icons.zip](https://github.com/pioneerspacesim/pioneer/files/4943492/icons.zip)

WKFO:
This PR should be ready for test/merge now.

WKFO:
Oh, and this is where the icon is located, mirror symmetry to the combat alert icons.
screenshot-20200721-152715

fluffyfreak:
Ok this PR is all done and ready to review+merge etc?

WKFO:
@fluffyfreak yes, it is ready.

impaktor:
I'd squash the last commit into the first, as they look to edit the same files. 
You could then add another commit that removes tve icon file and squash that commit also, and then make a new commit that adds only the icon file and commit with `--author=nozmajner`

...um, make sure you don't loose the icon file in the exercise!

nozmajner:
I don't mind if you don't do the thing with the icon and crediting. No need for the hassle.

WKFO:
> 
> 
> I'd squash the last commit into the first, as they look to edit the same files.
> You could then add another commit that removes tve icon file and squash that commit also, and then make a new commit that adds only the icon file and commit with `--author=nozmajner`
> 
> ...um, make sure you don't loose the icon file in the exercise!

Done.
I think.

impaktor:
The git fascist in me is happy.

As for the code, I assume that blob that does the magic in lua is mostly heuristic, as such I don't intend to review it at any lenght, maybe someone else want to do it? If this has been tested and works, I'm happy to merge (or "rebase and merge" if using github's gui) as is.

PR #4894, opened 2020-06-22
Spelling Updates

craigo-:
Spelling and grammatical changes to:

- module-advice/en.json
- module-newseventcommodity/en.json
- equipment-core/en.json
- module-assassination/en.json
- module-breakdownservicing/en.json

### DIFF
--- a/data/lang/module-advice/en.json
+++ b/data/lang/module-advice/en.json
@@ -8,8 +8,8 @@
     "message": "Know when to leave town"
   },
   "ADVICE_1_BODYTEXT": {
-    "description": "Point made is: the real illegal goods traders are fix to each station, but police's fake illegal goods traders will be random every time.",
-    "message": "As you return to a station, be weary of those alternative market places you do not recognize, and you will stay out of trouble with the law."
+    "description": "Point made is: the real illegal goods traders are fixed to each station, but police's fake illegal goods traders will be random every time.",
+    "message": "As you return to a station, be wary of those alternative market places you do not recognize, and you will stay out of trouble with the law."
   },
   "ADVICE_1_HEADLINE": {
     "description": "",
--- a/data/lang/module-newseventcommodity/en.json
+++ b/data/lang/module-newseventcommodity/en.json
@@ -37,22 +37,22 @@
   },
   "FLAVOUR_13_NEWSBODY": {
     "description": "Imply that nobody wants to buy ROBOTS anymore.",
-    "message": "It is now man against machine in the {system} system ({sectorx}, {sectory}, {sectorz}), as \"behavioural malfunction\" causes home consumers to loose faith in robot servants."
+    "message": "It is now man against machine in the {system} system ({sectorx}, {sectory}, {sectorz}), as \"behavioural malfunction\" causes home consumers to lose faith in robot servants."
   },
   "FLAVOUR_14_HEADLINE": {
     "description": "",
-    "message": "NEWS {date}: Solar flare causes biggest black out in a millennium in the {system} system."
+    "message": "NEWS {date}: Solar flare causes biggest blackout in a millennium in the {system} system."
   },
   "FLAVOUR_14_NEWSBODY": {
     "description": "Avoid using word Tupperware, say tuppenware, blooperware, tooperwar or plastic containers etc. If translation doesn't work, feel free to be creative, as long as it is implying a high demand of PLASTICS.",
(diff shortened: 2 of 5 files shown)


data/lang/module-advice/en.json
```diff
@@ -8,8 +8,8 @@
     "message": "Know when to leave town"
   },
   "ADVICE_1_BODYTEXT": {
-    "description": "Point made is: the real illegal goods traders are fix to each station, but police's fake illegal goods traders will be random every time.",
-    "message": "As you return to a station, be weary of those alternative market places you do not recognize, and you will stay out of trouble with the law."
+    "description": "Point made is: the real illegal goods traders are fixed to each station, but police's fake illegal goods traders will be random every time.",
```
impaktor, line 11: I'm not convinced that 'fixed' is preferred over 'fix', as it not referring to something that is "put in order"/fixed, but rather, something that is constant. Native speakers can chime in on this if I'm wrong.

data/lang/module-newseventcommodity/en.json
```diff
@@ -37,22 +37,22 @@
...
-    "message": "The system wide power outage wrecks havoc to the {system} system ({sectorx}, {sectory}, {sectorz}). Consumers line up to buy the last Trapperware to replace their fridges, as producer has run out of raw material."
+    "message": "The system wide power outage wrecks havoc to the {system} system ({sectorx}, {sectory}, {sectorz}). Consumers line up to buy the last Trapperware to replace their fridges, as the producer has run out of raw material."
   },
   "FLAVOUR_15_HEADLINE": {
     "description": "",
     "message": "NEWS {date}: Police raids crime syndicate in the {system} system, key leaders arrested."
   },
   "FLAVOUR_15_NEWSBODY": {
-    "description": "imply high demand of NARCOTICS.",
+    "description": "Imply high demand of NARCOTICS.",
```
impaktor, line 55: Technically, description fields are just for the internal transifex system, and not used in-game, so doesn't matter if upper/lower case. Just mentioning it for future reference.

impaktor (commented):
So in general, I (we?) are reluctant to mess about with the english translation strings once in master, since any change will trigger re-translation of that string in all other languages. 

I know there's at least one post on the forum that has done a lot of "nitpicking" on the en.json files, but that also never made it into master, for that reason. That being said, I'm OK with merging this, except for the "fix"/"fixed" swap that I don't agree with.

craigo-:
Thanks for looking at this, @impaktor. I am aware of the translation process that informs the need to touch these files as little as possible. This PR started with the definite improper use of a word that appeared mid-game (weary), but I expanded my checking to catch as much as I could in one hit. If any of this is straying into nit-picking territory, I'm happy to revert. And you could make the argument that the imperfect language adds flavour and authenticity - I'm sure people's spelling and grammar in 3200 isn't going to be perfect 😄

data/lang/module-advice/en.json
craigo-: Native speaker here (although school is a distant memory 😄)

The two definitions of the verb "to fix" still use the same conjugation and past participle (I think that's the right term for its use here). Replace with the verb "paint" and see what you get:

"The products are paint to each wall" / "The products are painted to each wall"

data/lang/module-newseventcommodity/en.json
craigo-: Fair enough and good to know. This was merely an inconsistency I picked up.

data/lang/module-advice/en.json
Web-eWorks: "Fixed" in this instance is a shorter and more common form of the verb "affixed", as in "to pin/nail/glue in place" or "to permanently attach". "Fix" in this instance doesn't have any semantic meaning other than a mis-conjugation of "fixed".

impaktor: So, e.g. "The North star is affixed / fixed in the northern sky"?

Web-eWorks: Yes, "The North Star is fixed at a constant heading in the northern sky" would be valid - here "fixed" means it's in a constant position. It's a little weird because "affixed" is an action verb and implies that "we" (for some definition of we) *did* the affixing, but that's more of an English idiosyncrasy, and likely why the "fixed" form tends to be more popular - because it doesn't imply the subject of the verb so strongly.

impaktor:
I've squashed the commits (they're all the same "spelling fix", don't need a commit for every file, or change), and rebased to current master, to get rid of old merge-commit, and force pushed it to your github branch, so beware, your local copy is now in conflict with your origin / github repo branch. Will merge shortly.

laarmen:
Oh wow, when they say that the owner can edit the PR branch they're not
kidding around!

PR #4902, opened 2020-07-01
Fix MSVC cmake build (#4658)

### LINKED ISSUE #4658: MSVC CMake Debug (x64) builds cannot run due to sigc++
### Observed behaviour
Use CMake to generate Visual Studio 2019 project
Open it and choose Debug x64
Build All
...
Once built attempt to run the `Pioneer` project,
It will stall on https://github.com/pioneerspacesim/pioneer/blob/master/src/Pi.cpp#L524 and if it ever progresses beyond it then it must take at least 10 minutes on a high-end machine but I've never waited that long...

This makes it much harder to debug under MSVC since we cannot run a full Debug build

### Expected behaviour
That it works like it used too.

### Steps to reproduce
see above

Windows 10, Visual Studio 2019

The-EG:
This requires https://github.com/pioneerspacesim/pioneer-thirdparty/pull/29

Update SIGCPP_LIBRARIES to use debug version for debug builds
Fix variable name reference in post_build command
Output executables to the root dir when doing MSVC build

fixes #4658 

With this, the cmake build should work again in MSVC for both Release and Debug.

I added the output dir thing so that you can run debug pioneer from within Visual Studio without having to move files around or attaching it manually. 
I may be using it differently, but when I open the CMakeLists.txt in MSVS (File->Open->Cmake) and build, it outputs everything under out/ without this change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,11 @@ if (WIN32)
 	set(CMAKE_INSTALL_BINDIR ${CMAKE_INSTALL_PREFIX})
 endif (WIN32)
 
+# Put the output into the root dir so it can be run from Visual Studio
+if (MSVC)
+    set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR})
+endif (MSVC)
+
 if (MSVC)
 	# Avoid annoying warnings from Visual Studio
 	add_definitions(-D_CRT_SECURE_NO_WARNINGS)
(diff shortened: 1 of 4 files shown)


impaktor:
> With this, the cmake build should work again in MSVC for both Release and Debug.

Will merge as soon as Windows-devs can confirm

The-EG:
Also, I can include a CMakeSettings.json, which is what Visual Studio is using to define the build configurations (x64-Debug, x64-Release, etc.). I hesitated to include this because I wasn't sure if it would affect building on other systems.

fluffyfreak:
I'm to merge this then test, I trust @The-EG

PR #4903, opened 2020-07-02
Fix crash in-flight if hj-planner has never opened

Gliese852:
When the amount of fuel changes on the ship, the function of rebuilding the route strings is called, because colors can change there. The function uses a pointer to the player, which is calculated when the planner window is displayed. If the window never been shown, a reference to the null pointer occurs and the game crashes.

### DIFF
--- a/data/pigui/modules/hyperjump-planner.lua
+++ b/data/pigui/modules/hyperjump-planner.lua
@@ -139,7 +139,8 @@ end --mainButton
 
 local function buildJumpRouteList()
 	hyperjump_route = {}
-	local start = current_path
+	local player = Game.player
+	local start = Game.system.path
 	local drive = table.unpack(player:GetEquip("engine")) or nil
 	local fuel_type = drive and drive.fuel or Equipment.cargo.hydrogen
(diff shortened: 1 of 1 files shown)


data/pigui/modules/hyperjump-planner.lua
```diff
@@ -139,7 +139,8 @@ end --mainButton
 
 local function buildJumpRouteList()
 	hyperjump_route = {}
-	local start = current_path
+	local player = Game.player
```
impaktor, line 142: Are you sure it should be `local` ? looks like player is a global to me. It's already defined on L14, and assigned in L326

Gliese852: @impaktor since this is a pointer, there is no difference inside this function. But the definition on line 326 only works when the hj-planner window is visible. It may happen that this function `buildJumpRouteList()` will be called _before_ the hj-planneer is shown. and global `player` will be `nil`.  In the other hand, other functions should not rely on the assignment in `buildJumpRouteList()`. Therefore, I decided not to touch the global variable `player` in this function.

PR #4906, opened 2020-07-04
Fix some things with regards to the sector map

### LINKED ISSUE #4353: Do not "lock" camera when searching with coordinates in the Sector View
Thanks a lot @fluffyfreak for #4341 . :)
Though I just have a minor complaint.

### Observed behaviour
When moving the SectorView camera by entering coordinates to the search field, the camera position gets "locked" to the coordinates and you can't move it. I think it might be caused by the camera position being updated every frame. If you delete the coordinates in the search field the camera "unlocks" and you are free to move again.

But, while deleting (or writing) coordinates, if the z coordinate has more than one digits, the camera parses every string and bounces around trying to go everywhere you write. 

For example, if you want to go to (11, -5, -28), the camera first goes to (11, -5, -2) and then bounces to (11, -5, -28). Same while deleting.

### Preferable fix
Only try to move the camera ONCE when 'Enter' is pressed. It seemed to work pretty well before.

Latest master, Windows 10 Enterprise

### LINKED ISSUE #4667: Sector Map: Auto Route button crashes game when no hyperdrive fitted
### Observed behaviour
With a ship that is not fitted with a hyperdrive, pressing the "Auto Route" button on the sector map results in a game crash:

image

### Expected behaviour

That the game doesn't crash 😄

Perhaps for ships without a hyperspace drive, pressing the "Auto Route" button could generate an informational alert (e.g. "No hyperdrive"), or perhaps simply do nothing at all.  

### Steps to reproduce

1. Start at Barnard's Star (Xylophis, no hyperdrive)
2. Switch to Sector Map
3. Select any system distant from Barnard's Star
4. Click "Auto Route"

NB: when doing the above with a hyperdrive-equipped ship, the game does not crash.

Pioneer 20190203, and compiled from master (commit 4bc31eef4700b343256cb1dae604622680c534d8)

Gliese852:
Screenshot from 2020-07-05 21-20-05

- clear route strings when enging the game, otherwise, when the game restarts, a segfault occurs

- it turns out that a hyper jump is possible not only to the stars, but also to the "substellar objects", but they do not return in the method `GetStars`. Since a specific body should be shown in the route planner, implemented an additional method `GetJumpable`, that gets all the bodies available for jumping `(IsJumpable == true)`, and this is synchronized with what you can click in the systeminfo view

- also I had to change the route check, now it’s not the star index that is checked, but just the path (this check is performed for multisystems, so that only the specific body that registered in the route is always selected)

- fix that the "remove jump" button continued to remove jumps even when the route became empty

- reordered sections in the object information widget, added icons to the names of stars in the multisystem

- remove method `SectorView::SwapSelectedHyperspaceTarget`, because it is not used anywhere

 - **add a module for creating message boxes**

- fix that Auto Route button crashes game when no hyperdrive fitted. Add 3 possible options - successful route building, no drive, no valid route. Also turn the last static function in `LuaSectorView.cpp` into a lambda, because it turned out that it was not needed in other functions. Fixes #4667 

- fix that when moving the SectorView camera by entering coordinates to the search field, the camera position gets "locked" to the coordinates and you can't move it (now it moves only when the search bar changes) Fixes #4353 

P.S. The system closest to the Sol, in which there is "substellar object", that is jumpable, but not in the list of the stars - Altair.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -1311,6 +1311,10 @@
     "description": "",
     "message": "No missions."
   },
+  "NO_VALID_ROUTE": {
+    "description": "Message in the sector map, issued by the autoroute function.",
+    "message": "No valid route to the selected system."
+  },
   "N_LIGHT_YEARS_N_MAX": {
     "description": "",
     "message": "{range} light years ({maxRange} max)"
--- a/data/pigui/libs/message-box.lua
+++ b/data/pigui/libs/message-box.lua
@@ -1,0 +1,25 @@
+-- Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local ui = require 'pigui'
+local ModalWindow = require 'pigui.libs.modal-win'
+
+local OK_BUTTON_WIDTH = 100
+
+local msgbox = {}
+
+msgbox.OK = function(msg)
+	ModalWindow.New('PopupMessageBox', function(self)
+		ui.text(msg)
+		local width = ui.getContentRegion().x
+		if width > OK_BUTTON_WIDTH then
+			ui.dummy(Vector2((width - OK_BUTTON_WIDTH) / 2, 0))
+			ui.sameLine()
+		end
+		if ui.button("OK", Vector2(OK_BUTTON_WIDTH, 0)) then
+			self:close()
+		end
+	end):open()
+end
+
+return msgbox
--- a/data/pigui/modules/hyperjump-planner.lua
+++ b/data/pigui/modules/hyperjump-planner.lua
@@ -176,6 +176,7 @@ local function updateHyperspaceTarget()
 		sectorView:SetHyperspaceTarget(hyperjump_route[1].path)
 	else
 		sectorView:ResetHyperspaceTarget()
+		selected_jump = nil
 	end
 end
 
@@ -225,7 +226,6 @@ local function showJumpRoute()
 		mainButton(icons.retrograde_thin, lui.CLEAR_ROUTE,
 			function()
 				sectorView:ClearRoute()
-				selected_jump = nil
 				updateHyperspaceTarget()
 		end)
 		ui.sameLine()
@@ -359,6 +359,13 @@ function hyperJumpPlanner.onEnterSystem(ship)
 			end
 		end
 	updateHyperspaceTarget()
+end
+
+function hyperJumpPlanner.onGameEnd(ship)
+	-- clear the route out so it doesn't show up if the user starts a new game
+	sectorView:ClearRoute()
+	-- also clear the route list, saved in this module
+	buildJumpRouteList()
 end
 
 return hyperJumpPlanner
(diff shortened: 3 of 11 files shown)


impaktor:
Your second point makes me nervous. What are the jumpable object? As far as I know, we do not consider grav points to be jumpable. 

Also, it's unclear to me how one selects which star, in a binary system (e.g. Sirius), one will jump to. I know I can set it from "Planetary Info" (C++/OldUI), but how would one set if from sector view?

Gliese852:
@impaktor 
> Also, it's unclear to me how one selects which star, in a binary system (e.g. Sirius), one will jump to. I know I can set it from "Planetary Info" (C++/OldUI), but how would one set if from sector view?

You can select a specific body in the list at the objectinfo widget, and it will also be displayed in the route, (if added), and the jump will be made on it:
Screenshot from 2020-07-05 11-11-03

> ...What are the jumpable object? As far as I know, we do not consider grav points to be jumpable.

If you look at the Altair system, this is a single system, and there is only one star in the list:
Screenshot from 2020-07-05 11-12-00

But in the system info screen, you can see that there is a substellar object Altair f, and you can also select it for a jump.
Screenshot from 2020-07-05 11-12-32

The problem is that this is not visible in the sector map, and in this PR I fix it, now all the objects that can be selected for the jump are displayed in the list in the sector map, and are fixed in the route.
Screenshot from 2020-07-05 11-27-05

impaktor:
> But in the system info screen, you can see that there is a substellar object Altair f, and you can also select it for a jump.

I wonder if that is a bug, that we even make the thingy jumpable? Let's say Jupiter was ignited (like in the plot of the 2001 sequel), making it a brown dwarf, should that be considered a "jumpable" object?

Gliese852:
> I wonder if that is a bug, that we even make the thingy jumpable? Let's say Jupiter was ignited (like in the plot of the 2001 sequel), making it a brown dwarf, should that be considered a "jumpable" object?

I don't know much about astronomy, apparently Altair f is not so big that it was a binary system, but big enough to jump on it. By the way, it is 3 times larger than Jupiter, and 14 times heavier.
You can disable the ability to select such objects for a jump, but then in this particular system you will have to fly very far to the star port.
Probably, you can turn such systems into binary, but it seems that sooner or later the opportunity will be implemented to jump not only to the stars. (L-points, jumpgates?)

P.S. or maybe add the ability to jump to planets like Jupiter too?

WKFO:
Brown dwarfs should be jumpable, because there are systems that only has brown dwarves. It would make sense that we could jump to them even if they shared the system with a more massive body.

I would disagree with jumping to Jupiter or sub-Jupiter sized gas giants though, if you can jump everywhere, what do you need all that Delta-V for?

nozmajner:
What if all bodies above a certain mass are jumpable, and as a backup measure the root body of the system is always jumpable?

impaktor:
Oops, seems I've opened pandoras box, I suggest we take further discussions elsewhere. 

What this PR addresses, is that things work consistently in different UIs, so all good.

Gliese852:
I think I caught UB related to the route list, and I'm going to fix it here.

impaktor:
@Web-eWorks do you want to review this? Do you have time the comming days?

Gliese852:
@impaktor oh, I also wanted to fix here #4353 and #4667, almost done

Web-eWorks:
I should be able to review this over the weekend. Thanks @Gliese852 for your work on this!

PR #4907, opened 2020-07-06
No more crashes when hyperjumping

### LINKED ISSUE #4901: Game crashes when hyper jumping while having any ship screen open
### Observed behaviour
Game crashes when hyper jumping while having any ship screen open

### Expected behaviour
Game should not crash

### Steps to reproduce
1. Select any destination on your hyper jump computer
2. Take off and activate hyper jump
3. Before hyper jump occurs open any ship info screen

__My pioneer version (and OS):__ 20200203 (Windows 7)
Error messages and PC specs included in Screenshots

Error
Error_2
Specs

Gliese852:
## Commits

<strike>

### Notify the camera when the ship changes the frame

Since the camera is not tied to the ship, but to the frame of the ship, when the frame changes, the information in the camera becomes irrelevant. If the worldview is active, it is quickly updated, but if the worldview is not active, an accident occurs.
This patch fixes a crash when hyper jumping with an open info screen.

### Rename CameraContext vars to cameraContext (optional)

Many variables of type CameraContext are called camera, and this seems to be confusing because there is also a type Camera, and variables of this type, also called "camera" or something like that.
</strike>

_The problem with info screens was resolved in #4941_
### Fix crashes when hyperjump with a map opened
ezgif-2-f0135cd12af1
Fixes #4901

**System map:** add a check to see if we are in hyperspace

**Sector map:**
 - move the drawing of the player's indicator from `DrawNearSector` up to `Draw3D`
 - calculate the interpolated position of the player's indicator during a hyperjump
 - don't display current system during hyperjump (lua-side)
 - fix selected system indicator (green circle), it pointed not to the the selected system, but to the current system

**Also:**
 - add easing function into `utils.cpp` to smooth the movement of the player indicator
 - move fuel reduction from the start of the hyperjump to the end
 - force handling of lua events immediately after entering system

## TODO:
- [x] Fix crash in the system map.
- [x] Fix crash in the sector map.
- [x] Show player movement between systems in the sector map

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -320,7 +320,7 @@ HyperdriveType.HyperjumpTo = function (self, ship, destination)
 	return ship:InitiateHyperjumpTo(destination, warmup_time, duration, sounds), fuel_use, duration
 end
 
-HyperdriveType.OnEnterHyperspace = function (self, ship)
+HyperdriveType.OnLeaveHyperspace = function (self, ship)
 	if ship:hasprop('nextJumpFuelUse') then
 		local amount = ship.nextJumpFuelUse
 		ship:RemoveEquip(self.fuel, amount)
--- a/data/libs/Ship.lua
+++ b/data/libs/Ship.lua
@@ -1015,12 +1015,9 @@ local onEnterSystem = function (ship)
 			end
 		end
 	end
-end
-
-local onLeaveSystem = function (ship)
 	local engine = ship:GetEquip("engine", 1)
 	if engine then
-		engine:OnEnterHyperspace(ship)
+		engine:OnLeaveHyperspace(ship)
 	end
 end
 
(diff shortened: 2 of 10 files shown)


Gliese852:
Apparently, this PR should wait for #4908 to merge.

Web-eWorks:
Not necessarily, but I'd much prefer seeing: a) the frame-switching logic moved to ShipViewController instead of duplicating it in the (soon-to-be-deprecated) CameraController subclasses, and b) the logic related to reacting to a change in frames attached to the Player ship or, more efficiently, in the ShipViewController (a simple check once per frame to compare FrameIds is literally a single cycle, something we can well afford compared to adding 16+ bytes and an irrelevant signal handler call to every single ship's datum in the world.)

If we have another impending use-case for an `onChangeFrame` signal attached to all ship objects I stand corrected on the latter point, but I'd rather keep fixes like this well contained and work on *removing* methods and logic from the monolithic object hierarchy we have now.

Gliese852:
@Web-eWorks I touched the framework in this query, because there was a crash in hyperspace, as a result of calling function `Pi::game->GetWorldView()->BeginCameraFrame();`. Since it will be deleted in #4908, existing commits (and signals) will most likely not be needed.

Web-eWorks (approved):

Issue #4818, opened 2020-02-27
Decoupling Cameras from the Player's Ship


Web-eWorks:
As promised in #192, this issue tracks the work needed to detach the view controller from the Player and allow centering / orbiting arbitrary objects or indeed any point in space at all.

We have several use-cases for this but they ultimately boil down to two:
- An "editor camera" used to fly around the solar system and inspect / verify that procedural generation, AI behavior, and custom definitions are working as intended. This in particular will be extremely important in implementing terrain generation.
- A "screenshot camera" used to take screenshots of the player's location and surroundings from arbitrary angles and positions, or to follow other ships or objects.

This naturally requires that the camera is able to "fly" on its own, with systems such as terrain generation and model rendering reacting appropriately to the presence of the camera and not the ship. As far as I am aware, most rendering-related systems use the camera's position by default, so I anticipate very little clean-up beyond decoupling such systems from vestigial dependencies on the player object.

As part of this work, we would like to improve handling of the existing cameras:
- Sidereal and External cameras need a "spring" behavior that prevents them from clipping inside world and ship geometry. This spring should offset the camera to push it out of geometry it would otherwise be clipping into, but should not limit the camera's zoom.
- The External camera needs exponential zoom scaling; the further the distance from the orbit point, the faster zoom should act.

This is not going to be a small undertaking, but the groundwork for it has been laid in #4536. This work will involve further decoupling ship and view controller, allowing PiGui to transition between a ShipViewController and a non-Ship ViewController, updating WorldView to properly handle the disconnect between Player and ViewController, and passing the responsibility for which object to orbit / center on to the ViewController instead of the individual CameraControllers.

Gliese852:
How do you look at making a camera relative to the global coordinate system when you are in space, as in eve online, for example, but switching on relative to the surface of the planet when you enter a rotating frame? (get close to surface)

impaktor:
Someone somewhere once requested missile camera. Maybe that's implied in one of the above listed points.

PR #4908, opened 2020-07-07
WIP: Improve Camera Controls

Web-eWorks:
This is a draft PR encapsulating my work on #4818 and some bugs fixed along the way.

**2020-07-06:**
Fixed the top/bottom directional indicators pointing the wrong way, removed the need to have an active camera frame during pigui code, made `WorldView::shipView` a pointer instead of a value in the class (why did I do it that way in the first place?), and made target indicator drawing more accurate when the camera is outside the ship.

**2020-07-10:**
Moved the old debug display code from WorldView to the PiGui debug code, cleaned up the last of the newUI references in WorldView, moved it to a PiGuiView base instead of UIView, and removed two signals from Pi that were used for almost nothing.

I'm getting closer to achieving the original goal, but I want a good understanding of the intricacies of WorldView first. Dropping 200 lines from the file is a good side-goal too!


Web-eWorks:
...I'm still working on this I promise, github just won't let you have an open PR without commits attached.

Web-eWorks:
Some fun things are happening on my WIP branches. I've got nothing worth pushing *yet*, but there are exciting possibilities on the horizon once I figure out how to tie everything together...

https://user-images.githubusercontent.com/4218491/117588616-66e83400-b0f2-11eb-8bf7-d6a9ff351451.mp4

nozmajner:
Cool! 
How those displays are done? Rendered to texture, and one could  select which parts to display with the UV? 
Asking because if this stuff goes live, I need to finish up the default cockpit replacement. And for future cockpits as well.

Web-eWorks:
> Cool!
> How those displays are done? Rendered to texture, and one could select which parts to display with the UV?
> Asking because if this stuff goes live, I need to finish up the default cockpit replacement. And for future cockpits as well.

I am aiming to have it directly render the imgui primitives (triangles) superimposed above the surface in the world, to avoid the need for a costly mipmap generation step and to generally produce far more crisp results. There are some *issues* with that however (lack of clip rect support for one, lack of renderer API features for several more) so for right now it's drawn to a low-res render target (300px square) and the render target's texture is mapped to the screen.

For right now, I'd suggest setting up all screens in the cockpit so that each one is a single plane with its own material, covering the 0..1 UV space. It's extremely cheap (at low resolutions and right now) to actually render to the screen so I don't think we need worry about atlasing. It also makes it a lot easier to then select specific atlas regions via push constants (uniforms) rather than trying to manually pack the screens into UV space during creation.

PR #4910, opened 2020-07-10
Set hyperdrive last service date to game start date for new games.

The-EG:
This sets the hyperdrive last service date to the game start date for new games. Currently, the hyperdrive service date is still set to Jan 1 3200 for new games:
image

Currently, that means that new games start with a hyperdrive that is 19 years out of service! 

A new game after this change:
image

Also in regards to the likelihood of the hyperdrive breaking:
Perhaps I'm particularly unlucky, but it would seem that a breakdown becomes fairly likely above 15 jumps after the service period ends. Just doing some quick testing, I got the drive to break 4 times in a row (out of 4 test runs) in the 13-17 jump range. I don't think this is necessarily bad, but it's not anywhere close to the maximum in BreakdownServicing.lua (255), and the likelihood of a random number being less than 0.0588 (~ 15/255) seems like it wouldn't be a given (the test is `Engine.rand:Number() < service_history.jumpcount / max_jumps_unserviced`). 
Actually, if I'm thinking about this correctly, the chance that the drive will break after 15 jumps should be about 5.8% (if random is perfectly random , etc.). Each jump after the service period ends should increase the likelihood of the drive breaking by just under 0.4% each time.

Out of curiosity, I ran 1 million random numbers (`Engine.rand:Number()`) and it's not favoring either end, so perhaps I am just that unlucky?
image

### DIFF
--- a/data/modules/BreakdownServicing/BreakdownServicing.lua
+++ b/data/modules/BreakdownServicing/BreakdownServicing.lua
@@ -225,7 +225,7 @@ local onGameStart = function ()
 
 	if not loaded_data then
 		service_history = {
-			lastdate = 0, -- Default will be overwritten on game start
+			lastdate = Game.time,
 			company = nil, -- Name of company that did the last service
 			service_period = oneyear, -- default
 			jumpcount = 0, -- Number of jumps made after the service_period


impaktor:
> I got the drive to break 4 times in a row (out of 4 test runs) in the 13-17 jump range. I don't think this is necessarily bad, but it's not anywhere close to the maximum in BreakdownServicing.lua (255), and the likelihood of a random number being less than 0.0588 (~ 15/255) seems like it wouldn't be a given 

You've considered that it's commulative? 1/255 + 2/255...., so by 15 jumps the random number generator has with high probability drawn a low number, in your case 15/255. I think if one made a plot of how the code actually works, you'd see what's going on.

The-EG:
@impaktor Yes, but not in the correct way. That's what I get for posting things late at night with no sleep! 
Even at 15 jumps, the probably is low (I think around 5% is correct). But the probability of something that can happen 5% of the time happening over 4-5 trials is pretty good: about 20% by the time you get to 5, and probably more than that since the probability increases each trial.

PR #4911, opened 2020-07-10
Cleanup Win32 Build (VS 2019)

### LINKED ISSUE #4899: Windows (Visual Studio) 32bit builds broken
The Visual Studio project can't be build in Win32 (x86) configuration. The build fails in FrameId.h, on the overloaded definitions of `operator const` (member function already defined or declared).

In Win32 configuration, `size_t` and `uint32_t` both evaluate to `unsigned int`.  In x64 `size_t == unsigned long long` and `uint32_t == unsigned int`

Windows 10 (VS 2019)
current master

The-EG:
This should get 32bit builds working again for Windows (using VS2019).

Remove const operator uint32_t() from FrameId (fixes duplicate member declaration on 32bit - fixes: #4899):
See conversation in https://github.com/pioneerspacesim/pioneer/issues/4899. I removed the uint32_t operator instead of size_t because it will not compile without size_t. This needs to be tested for Linux builds.

Remove calls to fmt::print from WriteLog instead of catching an exception:
Per conversation: https://github.com/pioneerspacesim/pioneer/pull/4890

Linker/Library options cleanup:
The 32bit pioneer-thirdparty libs are named a bit differently than the 64bit ones...

### DIFF
--- a/src/FrameId.h
+++ b/src/FrameId.h
@@ -14,7 +14,6 @@ struct FrameId {
 		m_id(new_id) {}
 
 	constexpr operator bool() const { return m_id != Invalid; }
-	constexpr operator uint32_t() const { return m_id; }
 	constexpr operator size_t() const { return m_id; }
 
 	constexpr bool operator==(FrameId rhs) const { return m_id == rhs.m_id; }
--- a/src/core/Log.cpp
+++ b/src/core/Log.cpp
@@ -84,18 +84,19 @@ void Log::Logger::LogLevel(Severity sv, nonstd::string_view message)
 void Log::Logger::WriteLog(Time::DateTime time, Severity sv, nonstd::string_view msg)
 {
 	std::string &svName = s_severityNames.at(sv);
-	try {
-		if (sv <= Severity::Warning) {
-			fmt::print(stderr, "{}: {}", svName, msg);
-		} else if (sv <= m_maxSeverity) {
-			fmt::print(stdout, "{}", msg);
-			// flush stdout because it might have a different cache size than stderr
-			fflush(stdout);
-		}
-	} catch (fmt::system_error) {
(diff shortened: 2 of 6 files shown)


The-EG:
Alternatively, I can remove the 32bit configurations instead of modifying FrameId. It's probably worth noting that the cmake windows build only has 64bit.

Web-eWorks:
Will test this on linux tomorrow, thank you for the work @The-EG! If I don't find any major issues this should be an easy merge.

fluffyfreak:
> Alternatively, I can remove the 32bit configurations instead of modifying FrameId. It's probably worth noting that the cmake windows build only has 64bit.

That's more of a cmake-is-lacking rather than meaning that it should be removed

The-EG:
> That's more of a cmake-is-lacking rather than meaning that it should be removed
True, I'm just pointing out that no one seems to be missing it either.

fluffyfreak:
> > That's more of a cmake-is-lacking rather than meaning that it should be removed
> 
> True, I'm just pointing out that no one seems to be missing it either.

They don't miss it now, but they will when someone adds in BGFX / Vulkan renderer and it can then run on an ARM/Risc-V CPU. 

It did build on ARM for Raspberry Pi as 32-bit (_but not run_) and it keeps getting brought up again for various SBC's which are a mix of 32-bit and 64-bit. It takes minimum effort to maintain, but it's a nightmare to re-introduce support for 32-bit if abandon it for long enough.

Anyway I'm happy with this PR, and I take the long view

Web-eWorks:
Sorry this took so long to get around to, I've got a half-finished commit that is taking longer than I thought to clean up my working tree. I should be able to test this either tonight or tomorrow.

Web-eWorks:
Looking good on the linux side, merging in now!

PR #4912, opened 2020-07-10
Resize Load/Save Window (#4674)

### LINKED ISSUE #4674: Load game: UI a bit small
### Observed behaviour
The Load game dialogue is a bit small, truncating the text of both the file name and the date. (I do like the popup detail, though.)

Screenshot below, taken with game resolution at 1920x1080 fullscreen (image cropped):
image

For reference, this is what it used to look like in Pioneer 20180203 with the same savefiles, again at 1920x1080 fullscreen (image cropped):
image

NB: the truncation gets more pronounced the lower the resolution.

### Expected behaviour
It would be nice to have a bit more horizontal space to display more of the file name and date.

### Steps to reproduce
Load game (with some saves present)

Pioneer master (6fd8f02f69f51ed772e6d0f6888491feba60dbf6) - Release x64
Pioneer release (20190203)

__opengl.txt__
[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/3618762/opengl.txt)

The-EG:
Resized the Load/Save game window to something that looks a bit more reasonable. Also resized the buttons and updated both sizing to use ui.rescaleUI.

Removed bindingButtonSize  and mainButtonSize since they don't seem to be used in this module (I think these are leftovers from when Save/Load was in the same module as main menu).

image

fixes #4674

### DIFF
--- a/data/pigui/modules/saveloadgame.lua
+++ b/data/pigui/modules/saveloadgame.lua
@@ -21,9 +21,7 @@ local icons = ui.theme.icons
 local pionillium = ui.fonts.pionillium
 local popupOpened = false
 
-local mainButtonSize = Vector2(40,40) * (ui.screenHeight / 1200)
-local optionButtonSize = Vector2(125,40) * (ui.screenHeight / 1200)
-local bindingButtonSize = Vector2(135,25) * (ui.screenHeight / 1200)
+local optionButtonSize = ui.rescaleUI(Vector2(150,25), Vector2(1920,1080))
 local mainButtonFramePadding = 3
 
 local saveFileCache = {}
@@ -129,6 +127,7 @@ ui.saveLoadWindow = ModalWindow.New("LoadGame", function()
 
 	showSaveFiles()
 end, function (self, drawPopupFn)
+	ui.setNextWindowSize(ui.rescaleUI(Vector2(640, 400), Vector2(1920,1080)), "Always")
 	ui.setNextWindowPosCenter('Always')
 	ui.withStyleColorsAndVars({PopupBg = Color(20, 20, 80, 230)}, {WindowBorderSize = 1}, drawPopupFn)
 end)


data/pigui/modules/saveloadgame.lua
```diff
@@ -129,6 +127,7 @@ ui.saveLoadWindow = ModalWindow.New("LoadGame", function()
 
 	showSaveFiles()
 end, function (self, drawPopupFn)
+	ui.setNextWindowSize(ui.rescaleUI(Vector2(350, 400), Vector2(1920,1080)), "Always")
```
Web-eWorks, line 130:
```suggestion
	ui.setNextWindowSize(ui.rescaleUI(Vector2(640, 400), Vector2(1920,1080)), "Always")
```

I'd prefer the save/load dialog be wider than it is tall as a point of personal preference. This value is chosen for being golden-rule proportionate to 400, but is otherwise arbitrary.

Looks like a very straightforward PR otherwise, thanks! The ideal goal is to redesign this dialog so it fits our UI theming a little better, but until vakhoir comes back or I can drag myself out of the engine internals, this is a very acceptable solution. :smile:

The-EG: I'm not opposed to that, but it will change the button layout (they'll be grouped to the left), so I'll have to tweak them a bit too I suspect. I'll look at it.

The-EG:
More wide than tall now, per @Web-eWorks preference. I think the buttons/layout is still blah, but if it's going to/needs to be re-written then there's not much point in playing with it now...this was meant more for a quick QOL update.

Web-eWorks:
Regarding button layout, it might/should be possible to move the buttons inside the columns and use PushItemWidth to make them lay out using the full width of the column, if you're interested in another quick fix.

Otherwise, I'm thinking we'll redesign it to more closely match the style of the main menu, pending @nozmajner's thoughts, of course.

Thank you for the effort on this, I'll get it merged this evening if you're not interested in adjusting it further :smile:

PR #4916, opened 2020-07-16
Recon/combat missions require radar

WKFO:
Recon/combat missions can no longer be accepted unless you have a radar.

### DIFF
--- a/data/modules/Combat/Combat.lua
+++ b/data/modules/Combat/Combat.lua
@@ -140,36 +140,34 @@ local onChat = function (form, ref, option)
 		end
 
 	elseif option == 5 then
-		if Game.player:GetEquip('radar', 1) then
-			form:RemoveAdvertOnClose()
-			ads[ref] = nil
-			local mission = {
-				type        = "Combat",
-				client      = ad.client,
-				faction     = Game.system.faction.id,
-				org         = ad.org,
-				location    = ad.location,
-				rendezvous  = ad.rendezvous,
-				mercenaries = {},
-				introtext   = ad.introtext,
-				flavour     = ad.flavour,
-				in_progress = false,
-				complete    = false,
-				risk        = ad.risk,
-				dedication  = ad.dedication,
-				reward      = ad.reward,
-				bonus       = -(math.ceil(ad.dedication * 3) - 1),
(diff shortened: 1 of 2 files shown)


impaktor:
I don't thik I've ever played this mission type, is it impossible (or very very hard) to complete without a radar?

WKFO:
It is hard (not impossible though), and I think, for the most part, detecting hostile ships is what makes a recon mission a recon mission after all. 

In search and rescue missions you can get away with not having a radar because you already know the ship's last location. I don't think you should be sent without a radar to a combat mission though.

Another point to note (for gameplay), when you don't have a radar it doesn't trigger alerts when there is a ship nearby, so the radar-less player can timewarp themselves out of potential trouble, unlike those who do have radars installed.

impaktor:
I wonder if the message could be more "in character", like "Sorry, I don't think I want to give this to a pilot without even a radar", or something similar?

WKFO:
Yes, that sounds better. I am open for suggestions if anyone wants to add flavours.

impaktor:
I vote "no" on flavours, this is not a message the player will be pestered with every other minute, and adding flavours to what basically is a (mission) error message feels like over-engineering. But please feel free to re-word my previous suggestion.

data/modules/Combat/Combat.lua
```diff
@@ -140,29 +140,35 @@ local onChat = function (form, ref, option)
...
-			risk        = ad.risk,
-			dedication  = ad.dedication,
-			reward      = ad.reward,
-			bonus       = -(math.ceil(ad.dedication * 3) - 1),
-			due         = ad.due,
-		}
-		table.insert(missions,Mission.New(mission))
-		form:SetMessage(l["ACCEPTED_" .. Engine.rand:Integer(1, getNumberOfFlavours("ACCEPTED"))])
-		return
+		if Game.player:GetEquip('radar', 1) then
```
Web-eWorks, line 143: Rather than indenting the entire block, I'd prefer you simply invert the if statement to check if the player *doesn't* have a radar and early out, leaving the block at the existing indentation.

WKFO: I hope I got it right

Issue #4923, opened 2020-07-19
Missile: yield is hardcoded


PtrMan:
Missile.cpp hardcodes the yield of the missile.
>	const double damageRadius = 200.0;
>	const double kgDamage = 10000.0;

The constructor has a power parameter (which is stored in m_power), but it is just used for ECM jamming. I would expect that there are different parameters for the yield, jamming, etc. The radius can get computed with a simple formula by the yield, this is realistic in most environment (space and in atmosphere combat), the radius should actually depend on the density of the atmosphere, type of explosion and yield.

__My pioneer version (and OS):__
20200718

Web-eWorks:
Thank you for the contribution! Additional terms for atmospheric density et. al. are at your discretion to open a new pull request - at this time, missile damage is "close enough" that I'm satisfied with where it's at.

PR #4924, opened 2020-07-19
Missile: convert yield to radius

PtrMan:
convert yield of the missile explosion to radius.
This addresses the radius computation problem of #4923 .


PtrMan:
Hm the damage falloff calculation is still not realistic, I'll work out a realistic formula soon, either in this PR or in another one.

PR #4927, opened 2020-07-20
Missile.cpp: yield based damage

PtrMan:
Missiles use with this change only the yield. The received damage computation is now more physically correct/realistic.

This addresses some of the computation of #4923 .

It basically allows warheads with a wide range of choices for yield and type of explosion (conventional explosive, (thermo)nuclear)

### DIFF
--- a/src/Missile.cpp
+++ b/src/Missile.cpp
@@ -194,24 +194,25 @@ void Missile::Explode()
 	Space::BodyNearList nearby = Pi::game->GetSpace()->GetBodiesMaybeNear(this, queryRadius);
 	for (Body *body : nearby) {
 		if (body->GetFrame() != GetFrame()) continue;
-		double dist = (body->GetPosition() - GetPosition()).Length(); // distance from explosion in meter
+		const double dist = (body->GetPosition() - GetPosition()).Length(); // distance from explosion in meter
 
-		double targetRadius = 6.0; // radius of the hit target in meter
+		const double targetRadius = body->GetPhysRadius(); // radius of the hit target in meter
 
-		double areaSphere = calcAreaSphere(dist);
-		double crossSectionTarget = calcAreaCircle(targetRadius);
+		const double areaSphere = calcAreaSphere(dist);
+		const double crossSectionTarget = calcAreaCircle(targetRadius);
 		double ratioArea = crossSectionTarget / areaSphere; // compute ratio of areas to know how much energy was transfered to target
 		ratioArea = std::min(ratioArea, 1.0); // we must limit received energy to finite amount
 
-		double jReceivedEnergy = ratioArea * jYield; // compute received energy by blast
+		const double jReceivedEnergy = ratioArea * jYield; // compute received energy by blast
 
+		//Output("target.r %f\n", targetRadius);
 		//Output("dist %f\n", dist);
 		//Output("area sphere %f\n", areaSphere);
 		//Output("crossSectionTarget %f\n", crossSectionTarget);
 		//Output("ratio %f\n", ratioArea);
 		//Output("received energy %f\n", jReceivedEnergy);
 
-		double kgDamage = jReceivedEnergy * 0.000022; // received energy back to damage in pioneer "kg" unit
+		double kgDamage = jReceivedEnergy * 0.000016; // received energy back to damage in pioneer "kg" unit
 		//Output("dmg %f\n", kgDamage);
 
 		body->OnDamage(m_owner, kgDamage, dummy);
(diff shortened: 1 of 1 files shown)


src/Missile.cpp
```diff
@@ -172,24 +172,51 @@ bool Missile::OnDamage(Object *attacker, float kgDamage, const CollisionContact
...
-		double dist = (body->GetPosition() - GetPosition()).Length();
-		if (dist < damageRadius) {
-			// linear damage decay with distance
-			body->OnDamage(m_owner, kgDamage * (damageRadius - dist) / damageRadius, dummy);
-			if (body->IsType(Object::SHIP))
-				LuaEvent::Queue("onShipHit", dynamic_cast<Ship *>(body), m_owner);
-		}
+		double dist = (body->GetPosition() - GetPosition()).Length(); // distance from explosion in meter
+
+		double targetRadius = 6.0; // radius of the hit target in meter
```
fluffyfreak, line 199: You can get the radius of the body from `body->GetRadius()` I think

src/Missile.cpp
```diff
@@ -172,24 +172,51 @@ bool Missile::OnDamage(Object *attacker, float kgDamage, const CollisionContact
...
 	Pi::game->GetSpace()->KillBody(this);
 
-	const double damageRadius = 200.0;
-	const double kgDamage = 10000.0;
+	double kgYield = 200.0; // yield of the used explosive in kg
+	const double jEnergyPerKgExplosive = 4.184e6; // energy density of explosive, here it's TNT
+
+	const double jYield = kgYield * jEnergyPerKgExplosive; // how much energy was converted in the explosion?
 
+	double queryRadius = 2000.0; // hardcoded to 2 km, this is sufficient for most explosions
```
fluffyfreak, line 192: Why such a large distance compared to the old one?

PtrMan: This is just the query distance, not the "real" explosion distance. There is no real explosion distance due to the nature of an explosion in empty space. The spatial data structure which is queried has to know a maximum query distance.
I've chosen a relatively large distance to allow for a bigger yield without modifying the queryDistance too.
Note here that this doesn't have any effect on the performance because the query is done once and because space is relatively empty anyways. Not an issue in my opinion here.

src/Missile.cpp
PtrMan: thx, I'll use   it and retune the variable to get roughtly the same damage as the old code.

PtrMan:
I am done here for this PR.

The computation is still not as realistic as it may get. I think the transfered energy depends on the solid angle of the exposed surface of the hit target. The computation to approximate it better isn't that much more complicated.
It converges to roughly the same result with a increased distance, that's all what matters to me with this PR.

impaktor:
1. If this fixes the issue, please edit your first post to use [keywords](https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords) that close it if this is merged.

2. If this sets variables, like damage, radius, etc. of warhead, those aught to be set in data/libs/equipment.lua, rather than hardcoding in C++ side

PtrMan:
2) the damage was hardcoded in the original code too, can I do this in another PR after this one?

src/Missile.cpp
laarmen: May I suggest adding a comment to that effect?

```suggestion
	// The radius in which to search for possible recipients of the damages
	// Hardcoded to 2 km, this should be sufficient for most explosions
	const double queryRadius = 2000.0;
```

impaktor:
> 2. the damage was hardcoded in the original code too, can I do this in another PR after this one?

Sure, I withdraw my objection.

Web-eWorks:
I'm still (unfortunately) unable to chime in further, but I'll echo @impaktor's comment that we'd love to see this exposed to Lua and the equipment system. If you have any suggestions or ideas for how to improve that system that come up as you work on it, I'd love to hear them - not sure (and don't have time to check right now) if there's an open issue, but I'd definitely like to see equipment capabilities gain a bit more structure and flexibility than simply lumping them all into the Ship's properties.

fluffyfreak:
Is this just waiting on clang formatting to be fixed?

PtrMan:
Yes

PtrMan:
Closes #4923

impaktor:
@PtrMan we're about to do a release soonish (in a few days? a week?), so if you could 1. rebase to master, 2. and fix the clang complaint, then we could merge this, assuming @fluffyfreak is OK, which he seemed to be.

PtrMan:
Maybe I'll do it, I don't care if it is in this release or a future one.

Web-eWorks:
This PR has been manually merged into master in commit 3b4b43677571e647265f597396e740c5bf69dadd. Thank you for proposing it!

PR #4928, opened 2020-07-20
don't decay damage of projectile

PtrMan:
This is unrealistic thus I think that this is better. There is no friction in space :) .
This allows for long ranged projectile based weapons over like 10's to 100 km, which is the mode where "real" projectile based space combat will take place.

### DIFF
--- a/src/Projectile.cpp
+++ b/src/Projectile.cpp
@@ -210,9 +210,7 @@ void Projectile::TimeStepUpdate(const float timeStep)
 /* In hull kg */
 float Projectile::GetDamage() const
 {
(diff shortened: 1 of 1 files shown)


WKFO:
I thought we were launching charged particles that dissipate as they travel, not bullets.

nozmajner:
As WKFO is saying, the base assumption is that these are plasma cannons or something similar, not kinetic projectile.  
There were lots of discussion about ovehauling combat, with multiple types of weapons and such. 
So known, and you could say it will be fixed, if considered a bug at all.

Issue #4926, opened 2020-07-20
Star System names are not unique


PtrMan:
### Observed behaviour
Names of star systems are not unique.
![image](https://user-images.githubusercontent.com/1067920/87897714-7ce61480-ca4c-11ea-8660-b78040f683fc.png)

### Expected behaviour
Unique names of star systems in the system view

### Steps to reproduce

__My pioneer version (and OS):__
git 20200718 linux

Gliese852:
@impaktor will like it.

PtrMan:
heh found the other issue #4866 , it's funny that the screenshots look similar...

impaktor:
Please see my comments in the other issue, and please check issue tracker for existing bugs, and please close when you discover duplicates.

PR #4930, opened 2020-07-20
generate more unique sector names

PtrMan:
Adresses issues #4866 #4926 .

* name depends on sector index to add a constraint that close same names are less likely
* name is a bit longer to lower the probability of the same name existing at all

### DIFF
--- a/src/galaxy/SectorGenerator.cpp
+++ b/src/galaxy/SectorGenerator.cpp
@@ -114,7 +114,7 @@ const std::string SectorRandomSystemsGenerator::GenName(RefCountedPtr<Galaxy> ga
 		for (int i = 0; i < len; i++) {
 			name += sys_names[rng.Int32(0, SYS_NAME_FRAGS - 1)];
 		}
-		name += sys_names[SYS_NAME_FRAGS + ((sx * 23 + sy * 7) % SYS_NAME_FRAGS)]; // make name more unique, note that prime numbers are used to lower chance of collisions
+		name += sys_names[(SYS_NAME_FRAGS + ((sx * 23 + sy * 7) % SYS_NAME_FRAGS)) % SYS_NAME_FRAGS]; // make name more unique, note that prime numbers are used to lower chance of collisions
 		name[0] = toupper(name[0]);
 		return name;
 	} else if (weight < 800) {
--- a/src/galaxy/SectorGenerator.cpp
+++ b/src/galaxy/SectorGenerator.cpp
@@ -114,6 +114,7 @@ const std::string SectorRandomSystemsGenerator::GenName(RefCountedPtr<Galaxy> ga
 		for (int i = 0; i < len; i++) {
 			name += sys_names[rng.Int32(0, SYS_NAME_FRAGS - 1)];
 		}
+		name += sys_names[(sx*23+sy*7) % SYS_NAME_FRAGS]; // make name more unique, note that prime numbers are used to lower chance of collisions
 		name[0] = toupper(name[0]);
 		return name;
 	} else if (weight < 800) {
--- a/src/galaxy/SectorGenerator.cpp
+++ b/src/galaxy/SectorGenerator.cpp
@@ -114,7 +114,7 @@ const std::string SectorRandomSystemsGenerator::GenName(RefCountedPtr<Galaxy> ga
 		for (int i = 0; i < len; i++) {
 			name += sys_names[rng.Int32(0, SYS_NAME_FRAGS - 1)];
 		}
-		name += sys_names[(sx*23+sy*7) % SYS_NAME_FRAGS]; // make name more unique, note that prime numbers are used to lower chance of collisions
+		name += sys_names[(sx * 23 + sy * 7) % SYS_NAME_FRAGS]; // make name more unique, note that prime numbers are used to lower chance of collisions
 		name[0] = toupper(name[0]);
 		return name;
 	} else if (weight < 800) {
--- a/src/galaxy/SectorGenerator.cpp
+++ b/src/galaxy/SectorGenerator.cpp
@@ -114,7 +114,7 @@ const std::string SectorRandomSystemsGenerator::GenName(RefCountedPtr<Galaxy> ga
 		for (int i = 0; i < len; i++) {
 			name += sys_names[rng.Int32(0, SYS_NAME_FRAGS - 1)];
 		}
-		name += sys_names[(sx * 23 + sy * 7) % SYS_NAME_FRAGS]; // make name more unique, note that prime numbers are used to lower chance of collisions
+		name += sys_names[SYS_NAME_FRAGS + ((sx * 23 + sy * 7) % SYS_NAME_FRAGS)]; // make name more unique, note that prime numbers are used to lower chance of collisions
 		name[0] = toupper(name[0]);
 		return name;
 	} else if (weight < 800) {


nozmajner:
Could you post some exampes? Even a screenshot if relevant.

impaktor:
1. Not sure I see how sector coordinates solves anything? It's still random, you will thus still have events with same outcome.
2. Making the names longer is, in my estimation, not a fix
3. As I've previously noted: I don't know where people get the notion of names have to be unique. In my class there were three "Eric", There are (at least in Scandinavian languages) two countries named "Congo" (and they're right next to each other!), there are two Montenegro, there are infinite many cities in US called "Springfield", there are (technically) two "China", there are two world leading universities both located in "Cambridge",  bla bla bla.

However, I will concede that if they're right next to each other, one could add a modifier, like "Congo Kinshasa" and "Congo Brazzaville", to distinguish them (which is what they're called in Scandinavian languages, google tells me english speaking world calls one (both?) "republic of Congo"), so maybe "New Earth alpha" and "New Earth beta", or similar.

PtrMan:
An example where the old code would generate a collision but the new code doesn't:
Here the prefix is "ack-ack", so the old code would just generate two systems with "ackack", maybe they are close to each other too, no idea.
image

PtrMan:
>However, I will concede that if they're right next to each other, one could add a modifier, like "Congo Kinshasa" and "Congo Brazzaville", to distinguish them (which is what they're called in Scandinavian languages, google tells me english speaking world calls one (both?) "republic of Congo"), so maybe "New Earth alpha" and "New Earth beta", or similar.

>Making the names longer is, in my estimation, not a fix
It lowers the probability by the length of the array in this case. Hitting something with 3% may be better than hitting it with 100% :D

The code does something similar with the suffix...

PtrMan:
@impaktor the script (without parameters) says that it passes it.

Gliese852:
@PtrMan 
If interested, you can cherry-pick this commit
https://github.com/Gliese852/pioneer/commit/dd40e682db2e8f495a8b2f9a486714208d088c83
Using this, you can count the number of duplicate names in a given cube.

The-EG:
I've come across duplicate system names while playing...I generally take the same as @impaktor , but I also agree this is somewhat confusing when you get multiples in a small area, such as Quso in this case (I've created a HJ route between them all):
image
 
They aren't directly adjacent, but they are close enough that there's 3 of them near the system I'm in (Anti). 
I think a more straightforward solution (for my case) may be to include the sector route in the search window results (So Quso (5, 8, 5), Quso (2, 12, 7), & Quso (-1, 7, 6)). That obviously won't do anything for the system name itself, but would make distinguishing them on the sector map easier.

Gliese852:
I have collected statistics for sectors in (+-50, +-50, +-50) (~1000000 sectors):

**Original generator:**
```
Top 10 system names:
1. Anda : 1629
2. Greex : 881
3. Caneth : 863
4. Phiho : 863
5. Urve : 863
6. Cemi : 862
7. Arho : 859
8. Enlia : 858
9. Iove : 856
10. Ackex : 854
1987754 systems total.
```

**With this patch:**
```
Top 10 system names:
1. Andaa : 62
2. Andaan : 61
3. Exanda : 61
4. Aanda : 60
5. Andaay : 60
6. Andaur : 60
7. Andaho : 57
8. Andaio : 57
9. Andami : 57
10. Andaso : 57
1987754 systems total.
```

PtrMan:
@Gliese852 wow good work! Looks like I introduced a bug because the probability of collisions should be lower.

I should fix the computation and check with your LUA script.

PtrMan:
>I think a more straightforward solution (for my case) may be to include the sector location in the search window results (So Quso (5, 8, 5), Quso (2, 12, 7), & Quso (-1, 7, 6)).

This would be fine to me, then one can remember or write down the sector coordinates for trade runs etc.

The-EG:
> @impaktor the script (without parameters) says that it passes it.
.clang-format mentions it was based on clang 6.0.1, so if your local version is different it may handle things a bit differently.

I know the extension I use for Visual Studio gets a couple of things wrong now and again (usually whitespace related) and I think it's based on version 10.0. It also gets super picky with aligning trailing comments (it seems to want a combination of tabs and spaces sometimes).

PtrMan:
sorry, looks like it's doing nothing.

>[r0b3@toshi pioneer]$ time ./scripts/clang-format.sh 
>Checking files:
>Files in this commit comply with the clang-format rules.
>
>real	0m0,015s
>user	0m0,011s
>sys	0m0,009s

>[r0b3@toshi pioneer]$ clang --version
>clang version 10.0.0 
>Target: x86_64-pc-linux-gnu
>Thread model: posix
>InstalledDir: /usr/bin

but this is a different issue :/

Gliese852:
@PtrMan clang-format.sh checks **staged** files.

PtrMan:
Well this makes sense :D

Web-eWorks:
Yes, for your use you'd likely want to invoke `FORMAT_BASE=master ./autoformat` or `FORMAT_BASE=HEAD~1 ./autoformat` - there's little difference between the `autoformat` script and the `scripts/clang-format.sh` script other than the latter prompts you to automatically apply the issues in your files.

Before I can sign off on this PR I'd like to see some effort put into ensuring system names don't degrade in quality - from your example screenshot and the data @Gliese852 was collating, system names in this PR have slipped from very-repetitious-but-readable to a rather nonsensical mismash of difficult-to-pronounce random syllables. This may well be the fault of older code and sample-size, but I'd like to ensure that we keep procedural system names readable and not unilaterally promote all system names to 3+ syllables.

PtrMan:
There is no reason that it won't work with 2+ syllables, it just makes the code more complicated when it needs to check for collisions/matches of adjacent cells. The good news is that the code is not going to be much more complicated.
I'll see what I can do.

orbea:
> I have collected statistics for sectors in (+-50, +-50, +-50) (~1000000 sectors):

From the second list it seems to indicate there is far less diversity in names even if the names technically conflict less. For example there are only two names that do not start with `And` and only one that doesn't start with an `A`. Is this consistent?

PtrMan:
Note that the code had a stupid out of array access bug (fixed now with the strange looking double modulo code to handle negative indices).

laarmen:
FWIW I'd prefer the sector coordinate "solution" (which isn't a full solution *unless* we know that system names are unique within their sectors ?) to the suffix, which has exactly the same problem regarding uniqueness within the sector. As pointed out by @impaktor it's quite common to have the same name used for multiple entities, even within ecosystems with highly available information and relatively "organized" people : name clashed in software development are a quite common occurrence.

I guess it all depends on what the goal is. I feel that working on generating unambiguous names is better for the "efficiency" of the player, whereas working on ways to disambiguate the names when it makes sense is better in a universe building kind of way (and is obviously a harder problem to tackle). Tacking on a suffix from the get go is in the former category, while displaying the sector coordinates (akin to displaying the OR in Portland, OR) falls in the latter to me.

fluffyfreak:
I don't like this solution. It feels like a half-hack which slightly degrades the existing naming scheme and only partially "fixes" the problem.

However I ahte leaving peoples work without something constructive so whilst I'm on my lunchbreak I'll try to do two things.
1. extend the number of SYS_NAME_FRAGS to include more of them which should help avoid repetition.
2. add a second type of fragment name group to provide more variety

I think those are both better approaches and that this PR should probably be closed

impaktor:
The foot came down. Sorry.

PR #4932, opened 2020-07-20
Fix that no icons are added to html-reports of the profiler

Gliese852:
Before
Screenshot_20200720_224556
after
Screenshot_20200720_224649

### What was done
- add profiler icons to `data/icons/profiler`
- add function `CopyDir` to copy all files in a directory, and used it to copy icons to the user's directory
- move the creation of directories and icons from the `class Game` constructor to App startup


impaktor:
Nice. My only comment is I wonder if `data/` is the right place to keep them, as everything there is related to game play, while this feels more like a dev tool related thing. I don't have any better suggestion, atm. though.

Gliese852:
@impaktor I used this directory because everything in it is automatically put into release. Though, if think about it, there is no code for the profiler in the release anyway, so I can move it into `contrib/profiler/icons`, for example?

Web-eWorks:
Finally getting around to this PR, very sorry @Gliese852! Definitely good work finding the icons and setting them up - however I do have some things I'd like to see done a little differently (and that I will be working on, so no worries on your end!)
- CopyDir should be a method in the FileSystem namespace and not in `util.cpp`; I'd like to remove as many things as possible from that file and put them in their correct locations, and copying a directory is something that FileSystem is a natural fit for.
- I agree that data/icons/ isn't a great spot, but something occurs to me - all of the icons are < 1KB, which means that it's extremely simple to base64 encode them and simply bundle them into the profiler's HTML output, completely removing the need to copy them around.

Otherwise thank you very much - I have a branch kicking around where I cleaned up some of the profiler initialization similar to what you did, but never bothered to separate it out and PR it, so thank you for working on this!

Web-eWorks:
Update: got icon embedding to work, will push it to this branch as soon as I'm finished trimming all unused (and now unnecessary) icons from the HTML.

image

PR #4935, opened 2020-07-22
CityOnPlanet.cpp: deoptimization

PtrMan:
This stupid allocation isn't necessary, because not many buildings are allocated anyways.

### DIFF
--- a/src/CityOnPlanet.cpp
+++ b/src/CityOnPlanet.cpp
@@ -52,12 +52,6 @@ void CityOnPlanet::AddStaticGeomsToCollisionSpace()
 		skipMask = 0;
 		break;
 	}
-	Uint32 numVisibleBuildings = 0;
-	for (unsigned int i = 0; i < m_buildings.size(); i++) {
-		if (!(i & skipMask)) {
-			++numVisibleBuildings;
-		}
-	}
 
 	for (unsigned int i = 0; i < m_buildings.size(); i++) {
 		if (i & skipMask) {
--- a/src/CityOnPlanet.cpp
+++ b/src/CityOnPlanet.cpp
@@ -59,8 +59,6 @@ void CityOnPlanet::AddStaticGeomsToCollisionSpace()
 		}
 	}
 
-	// we know how many building we'll be adding, reserve space up front
-	m_enabledBuildings.reserve(numVisibleBuildings);
 	for (unsigned int i = 0; i < m_buildings.size(); i++) {
 		if (i & skipMask) {
 		} else {


laarmen:
Please refrain yourself from using expletives to comment code written by others, as well as placing such judgement without full context. This project has a long history, and optimizations that might have made sense at the time they were added might not be useful today, that does not make them, nor their author, "stupid".

I'm wary of vague statements like this one, be it one way or the other. This micro-optimization *in itself* might or might not be productive, but this patch is only a half measure. By removing the `reserve`call, the code is even *less* optimal as we have a first loop that now computes a number that isn't used. If you want to remove the explicit allocation, please remove its supporting code as well.

As to whether or not this should be merged, I do think that removing the allocation and first loop, combined with a slight rework of the second loop, would improve legibility, which is on itself a good reason to merge. But when claiming things about optimization, the actual situation is rarely as clear-cut as we'd like, and it's always better to provide hard numbers to back up your claims.

PtrMan:
> optimizations that might have made sense at the time
I am sorry, saving maybe 10us with 7 additional lines which is triggered once every 5 minutes didn't make any sense 10 or 20 years ago.
Yes it's hard to estimate how much time is spent when and where, but I can do it statically here :D .
The code is adding absolutly nothing, it's just bloat, and I just care about what the code does and if it's maintainable etc. not about the past people who wrote the code 10 years ago.

It is a de-optiomization, a de-optimization is not an optimization (for speed).

PtrMan:
I am playing JIT here doing deoptimization :) .

laarmen:
My point is *not* about the code. It's about the communication.

Software engineering is as much about communicating with other human beings than trying to make a computer do stuff, and that's even more valid in the context of a Free Software project. Stupidity is an attribute of sentient beings. One cannot be stupid if one couldn't be intelligent. Judging code "stupid" conveys a judgment of its author, even if it wasn't your intention.

Note that said author is probably not around anymore (I haven't seen them in quite a while on GH), and would probably not have been hurt by that, but you didn't know that. And this is a public forum. Any junior developer looking for a place to contribute to gain some experience might stumble upon this PR, where they basically see some code they would have written being removed with a free insult and a vague comment.

Now, even if you didn't mean to insult anyone and the comment was only directed at the code, don't automatically assume it was written for the context in which it is now. For instance you're assuming that it was triggered once every 5mn when it was written. Perhaps the collision space was once dynamically regenerated every frame ? That's unlikely, but actually possible, we've done worse things.

May I suggest a commit log along those lines for this PR?

```
CityOnPlanet: remove optimization

Remove the pre-allocation of m_enableBuildings as its code isn't really readable and its gains are small. The method is run once when instanciating a city, and the number of buildings is usually small enough no to have that big of an impact.
```

This provides the reasons behind the removal (readability), and the context for it (rarely run). That way, in 7 years when somebody wonders why this array isn't pre-allocated, they'll browse the git logs for the method and have a clear answer. :-)

PtrMan:
>Stupidity is an attribute of sentient beings.
Your wrong, current AI is pretty stupid and will be in our lifetime.

>Judging code "stupid" conveys a judgment of its author, even if it wasn't
your intention.
If you say so, I just care if a code follows a requirement or
specification. Code can be stupid even without a author. Who has written
the weights of neural networks? No one. It can still do pretty stupid stuff
without being handcrafted by a human.

>don't automatically assume it was written for the context in which it is
now.

What about flying elephants :) Code can always run in any context. Not a
good point in my opinion.

I'll improve the message.

Simon Chopin <notifications@github.com> schrieb am Mi., 22. Juli 2020,
09:52:

PtrMan:
Why should I care about fictitious people doing stuff that's unlikely, I have bigger problems now :D

PtrMan:
I am not new to OSS dev or engine dev etc. This is a bad trade for anyone's time, I just want to remove 7 lines :D

PtrMan:
>That way, in 7 years when somebody wonders why this array isn't pre-allocated, they'll browse the git logs for the method and have a clear answer. :-)

Sounds pretty unlikely to me, but one in a billion can still happen, right :D

fluffyfreak:
Actually I wrote that code, and maybe it could do with a comment to explain why I did it.
It's iterating over sparse data that can be anywhere from 1/4 to the full number of visible buildings so reserving the full size doesn't make sense.

Also it's missing the point that this can happen a lot. We don't by default have many cities on a planet by we do have **mods** that add thousands of cities to planets and avoiding reallocation does have a noticeable affect on performance on lower end machines.

So you're removing something without profiling it or asking why it might be like that?

PtrMan:
This is ridiculous :D . This bloat even doesn't make sense if it allocates
1000 cities in a frame. Let's say that a single malloc takes 100ns and
there are 100 buildings. Then allocating all with the current code takes
100us. Let's take the code here, that takes 10ms. Is this speedup worth the
code bloat? Its not in my opinion. Well its a codebase guarded by you so
you make the decisions. :D
Why should I use a profiler to profile something statically obvious?
Doesn't compute to me, good luck with the code bloat here.

Andrew Copland <notifications@github.com> schrieb am Mi., 22. Juli 2020,
11:16:

PR #4938, opened 2020-07-22
Add function GalaxyStats to the LuaEngine.cpp

Gliese852:
To collect statistics on star systems in a given area.
Screenshot_20200722_215149

It does not get in the way, and may even be useful.

example using from lua console:
```
require 'Engine'.GalaxyStats(0,0,0,5, "CountSystemNames CountSystems CountPopulation")
```
This will output statistics for a cube with a center 0,0,0, and a "radius" of 5 (in sectors)
the content of statistics is determined by the names of "processors", at the moment they are available 3. Each processor receives a link to each system in given cube, and generates a report.

### DIFF
--- a/src/lua/LuaDev.cpp
+++ b/src/lua/LuaDev.cpp
@@ -6,11 +6,163 @@
 #include "LuaObject.h"
 #include "Pi.h"
 #include "WorldView.h"
+#include <sstream>
 
 /*
  * Lua commands used in development & debugging
  * Everything here is subject to rapid changes
  */
+
+/*
+ * Method: GalaxyStats
+ *
+ * Output to stdout and lua console statictics in the given cube. Start the game before using!
+ *
+ * using:
+ * > GalaxyStats(centerX, centerY, centerZ, radius of cube, processors)
+ *
+ * example using from console:
+ * > require 'Dev'.GalaxyStats(0,0,0,5, "CountSystemNames CountSystems CountPopulation")
+ *
+ * Parameters:
+ *   centerX, centerY, centerZ - integer, coordinates of center of the cube, 0, 0, 0 = Sol
+ *   radius - integer - distance in sectors from center to edge of the cube
+ *   processors - string - Processor names separated by spaces
+ *     available processor names:
+ *       CountSystems
+ *       CountSystemNames
+ *       CountPopulation
+ *
+ * Availability:
+ *
+ *   2020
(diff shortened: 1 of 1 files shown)


PtrMan:
You may call it a Query instead of a Processor. All tasks are queries...

fluffyfreak:
This has been useful with the name generation testing!

impaktor:
I wonder if this should not be in the Ctrl+I debug window we already have? There is already a tab there for "Galaxy stat" (but currently only populated if using HYG).

I don't know if that's a lot more work, either way: yay for more debug!

Gliese852:
@impaktor Agree, it would be better with UI, but imho this is not the kind of information that needs to be updated every frame and constantly kept in front of your eyes. Therefore, I do not consider it vital.

Web-eWorks:
I heavily support adding it to the CTRL+I menu - I'll look at adding support for Lua to add items to those menus over the next few days. Having more dev tools is great, having them well integrated for developers to be able to easily use them is even better!

Gliese852:
@Web-eWorks imho UI != more easy to use. not that flexible. Do you use git through UI?

impaktor:
For those not on IRC, there's some effort into getting this into the Ctrl+I menu, thus this is on (temporary) hold

Web-eWorks:
Status update: adding this to the devtools menu is currently on hold due to a few issues with appending to tab items from Lua, as well as the galaxy stats mechanism would need to be reworked to return the information about local galaxies.

I have a few changes I'd like to request before I can give this PR my approval - @Gliese852 I'd prefer that you separate the GalaxyStats method and associated classes out into its own file given the nature of its implementation, and register it in the lua `debug` global table instead of in the `Engine` class table. Until the method is "well integrated" (e.g. it returns a string instead of dumping to console, and can be extended more easily) I don't think lumping it into the LuaEngine file improves our code quality. Also, for something that is very clearly a debug functionality, I don't think that Engine is the correct anchor point for it anyways.

Gliese852:
@Web-eWorks moved the function to `LuaDev.cpp`. Now instead of outputting to the console it returns a string. Not sure about registering in the `debug` table, since the function is already available through the `require 'Dev'`.

Web-eWorks:
Thank you @Gliese852 - when I wrote the original message I had forgotten LuaDev was actually a thing that we used.

Issue #4939, opened 2020-07-26
Build failure with USE_SYSTEM_LIBLUA


orbea:
### Observed behaviour

The build with `-DUSE_SYSTEM_LIBLUA=ON` fails.

```
[16/432] Building CXX object contrib/imgui/CMakeFiles/imgui.dir/examples/imgui_impl_opengl3.cpp.o
samu: job failed: /usr/local/bin/clang++  -DFMT_LOCALE -DGLEW_STATIC -I. -I/tmp/pioneer -I/usr/include/lua5.2 -I/tmp/pioneer/src -I/tmp/pioneer/contrib -I/tmp/pioneer/contrib/fmt/include -I/usr/include/assimp -I/usr/include/freetype2 -I/usr/include/SDL2 -I/usr/include/sigc++-2.0 -I/usr/lib64/sigc++-2.0/include -I/tmp/pioneer/contrib/glew -O2 -g -DNDEBUG   -std=gnu++11 -MD -MT CMakeFiles/pioneer.dir/src/main.cpp.o -MF CMakeFiles/pioneer.dir/src/main.cpp.o.d -o CMakeFiles/pioneer.dir/src/main.cpp.o -c /tmp/pioneer/src/main.cpp
In file included from /tmp/pioneer/src/main.cpp:5:
In file included from /tmp/pioneer/src/ModelViewer.h:18:
In file included from /tmp/pioneer/src/ui/Context.h:45:
In file included from /tmp/pioneer/src/ui/Lua.h:8:
In file included from /tmp/pioneer/src/lua/LuaObject.h:9:
/tmp/pioneer/src/lua/LuaMetaType.h:11:10: fatal error: 'src/lua.h' file not found
#include "src/lua.h"
         ^~~~~~~~~~~
1 error generated.
samu: subcommand failed
```

### Expected behaviour

The build with `-DUSE_SYSTEM_LIBLUA=ON` should work.

### Steps to reproduce

Build pioneer with `-DUSE_SYSTEM_LIBLUA=ON` and lua52 installed system wide.

I bisected this.
```
e80d11ee18165127c8710c3bd17acb168726075a is the first bad commit
commit e80d11ee18165127c8710c3bd17acb168726075a
Author: Webster Sheets <webster@web-eworks.com>
Date:   Sat May 9 05:48:42 2020 -0400

    LuaMetaType + LuaObject 2.0
    
    Add a new system to bind member functions and data types to lua without
[...]

PR #4940, opened 2020-07-26
Fix the build with USE_SYSTEM_LIBLUA.

orbea:
Fixes https://github.com/pioneerspacesim/pioneer/issues/4939

This header should not be included here and does not seem to be needed. It was probably accidentally included in commit 
https://github.com/pioneerspacesim/pioneer/commit/e80d11ee18165127c8710c3bd17acb168726075a.


Web-eWorks:
That's the fault of my IDE, it loves to do IWYU automatically without telling me. Good catch!

PR #4941, opened 2020-08-02
Debug Tooling, Worldview Refactor

Web-eWorks:
Most of this is described in #4908, but the bits that aren't are as follows:

#### Matrix4x4 refactor
<s>This one's a bit painful, it refactors matrix construction and adds a "shortcut" symmetric projection matrix function for future work on Screen::Project. However, due to the way it's written, I had to link in `std::array` to get "easy" initializer list construction with automatic double conversion.</s>
Struck this, going to rethink the implementation a bit.

#### Debug Tooling Improvements v2
Added extension points to the PerfInfo window, making it a general debug tool interface. Refactored the debug ship spawner to replace the old hardcoded CTRL+F12 spawner (incidentally, only CTRL+F12 works now, the rest of the F12 shortcuts are deprecated in lieu of a *much* easier to use interface).

Also added a Reload Shaders button to the renderer debug info to complement the F11 shortcut.

Will be merging this soon, as it's blocking on getting #4938 merged in. Would appreciate a quick sanity check review (especially from the Windows side of things, CC @The-EG @fluffyfreak ) to ensure I haven't fouled up MSVC's special heart for STL bugs.


Web-eWorks:
CC @Gliese852 as this changes a few sectorview things and I'd like his feedback and go/no-go on it!

The-EG:
Builds and runs with MSVC (both the solution and cmake). Didn't do any actual testing other than seeing it run...

Gliese852:
@Web-eWorks go on it! (regarding maps)

Web-eWorks:
Removed the matrix4x4 changes as they need some more time to cook, but otherwise merging this now!

PR #4943, opened 2020-08-12
DynamicBody: make CalcAtmosphericForce method protected in class hier…

starling13:

### DIFF
--- a/src/DynamicBody.h
+++ b/src/DynamicBody.h
@@ -36,7 +36,6 @@ public:
 	bool IsMoving() const { return m_isMoving; }
 	virtual double GetMass() const override { return m_mass; } // XXX don't override this
 	virtual void TimeStepUpdate(const float timeStep) override;
-	virtual vector3d CalcAtmosphericForce() const;
 	double CalcAtmosphericDrag(double velSqr, double area, double coeff) const;
 	void CalcExternalForce();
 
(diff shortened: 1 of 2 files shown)


Web-eWorks:
@starling13 it helps if you add a description to your pull request explaining exactly *why* we should accept the PR and merge your changes into code. Even the de facto 'core developers' who directly merge follow this model (albeit sometimes in an issue or in commit descriptions), and for a change as trivial as this I'd appreciate a little more information about why you think this was something necessary to fix.

impaktor:
1. Something small like this is good for initiating a first time contributor, like in this case, so this PR also serves that purpose.
2. I don't know the code, so either way, we should only merge this if it's an improvement, even if it's an infinitesmally small one, for purpose of point 1 above.

Web-eWorks:
Don't get me wrong, I think it's a great first pull request, but I'd prefer to encourage the habit of describing pull requests rather than simply opening them with no context or description - it increases the review burden unnecessarily, and a pull request description is the sort of "standard of minimum effort" that we apply to all contributors. It doesn't have to be War and Peace, but even a "hey, I'm new here, this seemed like a good first contribution" is extremely helpful.

starling13:
Agreed, some explanations always better then its absence.
I hope to do some humble improvements in the flight model and noticed some protected by design methods are public while dig in the code.

Web-eWorks:
Looking good, thank you for adding a description!

PR #4944, opened 2020-08-17
Load Economy Data from JSON

Web-eWorks:
## Load commodity and economy data from JSON

This PR completely rewrites the existing Economy system to load all relevant data about commodities from JSON, enabling modders to extend the game with new commodities, different commodity generation weights, and eventually, entirely new economies![1]

[1]: this message not contractual and will take quite a bit of work to implement - our current system generation setup doesn't quite understand what an "economy" really is, and economies have pretty much no impact on commodity pricing. This is something for the future when individual markets have their own economy types and intra-system trade is possible.

Additionally, this removes quite a few lines of duplicated commodity data between Equipment.lua and Economy.h, which should result in this PR coming... somewhat close to breaking even in terms of lines changed.
(No, I don't believe that while I'm writing it, and neither should you.)

The real upside here is that a lot of previously-hardcoded information about system economies is more data-driven and *far* less complex to reason about on the C++ side of things - what used to be a 100-line list of commodities is now a one-line array and 200 lines of loader code. (Yes, I know, but it's way easier to *use* that array now, and the generation code is much better for the change.)

I've made an effort to find and fix as many bugs as possible and ensure it operates as close to "vanilla" behavior as possible, but I need more eyes on the ball before I'm comfortable merging this in.

I've left commit history relatively as it was written, and included impaktor's (slightly modified) commodity debug tab to ensure that system generation is functioning as intended.

I'd appreciate it if all interested parties could try this branch and just do a sanity check during normal gameplay - it may or may not be backwards compatible with old saves, but I'm going to assume it is until proven otherwise.

## TODO:

- [x] Fix bonkers indentation in `StarSystemGenerator.cpp`
- [x] Remove fixed FIXMEs
- [x] Fix / utilize system descriptions being a translation key (might need C++ side l10n work, which isn't ideal)

﻿


Web-eWorks:
CC @Gliese852 @impaktor @fluffyfreak @The-EG - I'd appreciate your eyes on this PR, as it's been about five months in the oven and that's a long time to kick up bugs.

impaktor:
Nice to see a big project coming to fruition!

Now for the criticism ;)


> ...enabling modders to extend the game with new commodities, different commodity generation weights, and eventually, entirely new economies

So this would be stuff that's hooked into the economy, and not just the cargo hold (as is currently possible with custom cargo). 

Should not the commodity icon be set from json as well then?

To me, it feels strange to have illegal in its own file, since what is illegal varies from system and faction, it's something set by the judicial system, rather than the economic system. Does it even serve a purpose to split the commodities into diffferent json files? I think common items could also be illegal at some place, like "Amish" banning computers & robots, or liberterian world allowing everything. I'd prefer if that line would be more blurred. I.e. drugs could go into agriculture (or a future tourist economy?), and guns into industrial.

It would be good to have @ecraven and @laarmen (Now comming full circle!) pitch in with thoughts on this PR.

laarmen:
My first thought *from a user perspective* : JSON is a PITA to hand-edit (no comments, quotes everywhere, no trailing comma, mandatory uneasy-to-reach (on my layout) glyphs (`{}[]`). I **know** why this uses JSON, no need to introduce yet another dependency, but at least this downside is spelled out.

And yes, despite this, this is an improvement over the n-e situation for non-developers contributions, as was n-e over the previous, full-C++ implementation. Yay!

I'll leave inline comments when looking at the code, but the overall PR description sounds good.

data/libs/Commodities.lua
```diff
@@ -0,0 +1,63 @@
+-- Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local Economy = require 'Economy'
+
+local commodities = Economy.GetCommodities()
+local economies = Economy.GetEconomies()
+
+local EquipType = _G.EquipType
```
laarmen, line 9: Why is it necessary to use a global table here instead of simply requiring EquipType ?

data/libs/Commodities.lua
```diff
@@ -0,0 +1,63 @@
...
+	hand_weapons = "Hand_weapons",
+	air_processors = "Air_processors",
+	farm_machinery = "Farm_machinery",
+	mining_machinery = "Mining_machinery",
+	battle_weapons = "Battle_weapons",
+	industrial_machinery = "Industrial_machinery"
+}
+
+-- TODO: don't create a separate EquipType for each commoditity,
+-- instead store cargo separately from ship equipment
```
laarmen, line 51: Ah, that old can of worms... I hadn't the patience of doing so back then as we had at least one prominent mod (Scout+) that would have had a lot of breakage over this. How is the mod situation nowadays ?

data/libs/SpaceStation.lua
```diff
@@ -104,7 +104,7 @@ function SpaceStation:GetEquipmentPrice (e)
 	if equipmentPrice[self][e] then
 		return equipmentPrice[self][e]
 	end
-	local mul = e:IsValidSlot("cargo") and ((100 + Game.system:GetCommodityBasePriceAlterations(e)) / 100) or 1
+	local mul = e:IsValidSlot("cargo") and ((100 + Game.system:GetCommodityBasePriceAlterations(e.name)) / 100) or 1
```
laarmen, line 107: This comment is NOT to be addressed for this PR, and is something that could be solved in an unrelated, newcomer-friendly PR:

The `name` API for the equipment instances is misleading. I'd expect a name to be a string that is displayable on screen, whereas this value is more of an ID. It's just a small papercut in the many pitfalls of our codebase, but it's easily solvable.

src/Pi.cpp
```diff
@@ -912,6 +927,19 @@ void Pi::App::PostUpdate()
...
+	if (Pi::doProfileOne || (Pi::doProfileSlow && (GetFrameTime() > 0.1))) { // slow: < ~10fps
+		Pi::doProfileOne = false;
+		if (!Pi::profileOnePath.empty()) {
+			Profiler::dumphtml(Pi::profileOnePath.c_str());
+			Pi::profileOnePath.clear();
+		} else {
+			Profiler::dumphtml(Pi::profilerPath.c_str());
+		}
+	}
+#endif
```
laarmen, line 942: Are the profiling-related patches meant to go in within this PR or is it just to make the development easier but will go away in the non-draft version?

If the former, would it be possible to move them in a separate PR if that's not too much work ? At a glance the changes don't seem coupled with the equipment/economy stuff, but I might be wrong.

src/galaxy/StarSystem.cpp
```diff
@@ -109,35 +110,35 @@ const Color StarSystem::starRealColors[] = {
...
+	100.0,	  // B0
+	200.0,	  // O5
+	1000.0,	  // M0 Giant
+	2000.0,	  // K0 Giant
+	4000.0,	  // G0 Giant
+	6000.0,	  // F0 Giant
+	8000.0,	  // A0 Giant
+	9000.0,	  // B0 Giant
+	12000.0,  // O5 Giant
+	12000.0,  // M0 Super Giant
```
laarmen, line 136: On Github this looks unaligned with the above data. @impaktor didn't you use to have a trick to make the GH code view behave sensibly ?

impaktor:
> How is the mod situation nowadays ?

There's no modding community anymore that I'm aware of, so I say we need not worry too much about that, until someone complains, in which case we can help them be pioneer anno 2020 compliant.

> On Github this looks unaligned with the above data. @impaktor didn't you use to have a trick to make the GH code view behave sensibly ?

GH should use the definitions in [.editorconfig](https://github.com/pioneerspacesim/pioneer/blob/master/.editorconfig)

The-EG:
> To me, it feels strange to have illegal in its own file, since what is illegal varies from system and faction, it's something set by the judicial system, rather than the economic system. 

I had the same thought, and it may be confusing since many items in the other files are commonly illegal, like animal meat. Also, probably to the same point, slaves are in both illegal and agricultural. I think it would be more intuitive to put everything into a category and let the system decide what is illegal and not.

That being said, I'll try to get this compiled and run my precious metals<->air handlers route that I found and see how it compares! (among other things)

impaktor:
@laarmen: I'd like to append to my answer:

[github cheat sheet](https://github.com/tiimgreen/github-cheat-sheet#adjust-tab-space) says you can append `?ts=2` or `?ts=4` to the URL to change the tab-size, up to 12.

data/libs/Commodities.lua
Web-eWorks: There's a dependency hell going on here where Commodity.lua needs access to EquipType, but the rest of Equipment.lua needs access to the commodity data, because everything is interlinked. The global assignment is just a quick hack - the solution is to add a "preregistration" mechanism to the `package` table so that circular dependencies aren't a problem. I've built one of those before and it makes a *huge* difference for complex lua projects.

I might merge the contents of Commodity.lua back into Equipment.lua, but this was the first step in separating commodities from equipment so I kinda want to keep it that way.

data/libs/Commodities.lua
Web-eWorks: In my eyes, Pioneer mods don't exist right now. I haven't seen a recent mod (other than maybe one personal hobby project) be announced for the entire time I've been contributing, and all of the mods I can find are at least 4 years old. I'd rather move forward and break old mods than stay in this limbo of massively inflated savefiles and awkward cargo handling. (No offense meant to you, the equipment system is a massive step up from what was previous! It's just... not a good fit for cargo.)

If someone wants to upgrade existing mods, absolutely more power to them and I'd be more than happy to support them, but until that time I'm not going to let old and inactive mods be an excuse for leaving things as the status quo.

data/libs/SpaceStation.lua
Web-eWorks: Absolutely! This is a 1am hack to get things working - I'm not sure about the purpose of the `volatile` sub-object, but I needed to be able to pass around commodity names (and this isn't meant to be the final iteration of commodities anyways, `name` just so happened to be an unused key to attach this datum to).

src/Pi.cpp
Web-eWorks: Profiling will probably be committed directly to master, yeah. It was work that was kicking around in the branch and I didn't want to separate it out yesterday. I don't think this is particularly large enough for a separate PR (especially given that we just merged in a bunch of profiler changes), given that the new method is only really used thrice.

src/galaxy/StarSystem.cpp
Web-eWorks: Github's tabstops are weird, they don't respect our editorconfig file. You can set your tab width to 4 IIRC and that fixes things.

It looks perfectly fine in a 4-TAB text editor, so I'm going to call it an acceptable weirdness.

Web-eWorks:
So I do agree about the illegal file. It's definitely not the perfect way to do it, and I'd ideally like to make that the purview of a faction, where each faction can set commodity legality in their systems.

*However*, I'm pretty much porting the existing code directly, and the existing code literally doesn't care about factions or anything else, and it definitely doesn't use the judicial system. It's just a hardcoded list of commodities in StarSystemGenerator. The illegals file and consumables file is actually a step up now, because you can tweak the chances (or consumption) without having to recompile the entire 1500 line generator. So until someone (probably myself) redesigns the economic generation code, the illegals file stays.

Now the icon names. This is something I don't think is a problem - it's the UI's purview to map icons to commodities, not the other way round. The list of icon names is only there because I didn't feel like tracking down everything that used those icons and renaming them to exactly match the commodity names, but that's the eventual goal is to completely remove that list.

Thank you all for the wonderful feedback so far, you've definitely given me some things to think about for the future!

The-EG:
> it may or may not be backwards compatible with old saves

My prior save worked fine. 

I did my trade Quso<->Anti trade route, and everything seems normal commodity wise. The trade computer is still doing the trade analysis properly, illegal goods seem to be where they should be...
If I get more time I'll try to play around some more.

src/galaxy/Economy.cpp
```diff
@@ -2,108 +2,370 @@
...
+		const Json &inputs = j["inputs"];
+		if (inputs.is_array() && inputs.size() > 0) {
+			for (size_t idx = 0; idx < inputs.size() && idx < CommodityInfo::MAX_ECON_INPUTS; idx++) {
+				const Json &input = inputs[idx];
+
+				std::string name;
+				fixed amount = fixed(1, 1);
+
+				if (input.is_array()) {
+					name = input[0];
```
The-EG, line 86: This and line 89 fail to compile in MSVC:
> E:\Games\pioneer-dev\the-eg\pioneer\src\galaxy\Economy.cpp(86): error C2593: 'operator =' is ambiguous

The-EG: `name = input[0].get<std::string>()` and `name = input.get<std::string>()` instead seems to work, but I'm not familiar with this JSON library or what's going on here.

The-EG:
> it may or may not be backwards compatible with old saves

Going to have to go with 'not.' My old save loads without error, but as soon as I try to open the station window I get:
image

It seems to work fine with a new game. I'll have to recreate my save with the console and otherwise play around a bit to see how it's actually working. That reminds me, I had the idea to write a script that would 'recreate' a save (basically put you in the same location, ship, money, etc)...

craigo-:
> That reminds me, I had the idea to write a script that would 'recreate' a save (basically put you in the same location, ship, money, etc)...

Ooh, I'd be interested in that.

src/galaxy/Economy.cpp
Web-eWorks: Yup, that's my bad. It's a limitation of the json library (and C++ implicit conversion operator overloading in general) - I'll fix it in the cleanup.

Web-eWorks:
And to add an additional comment about illegality - the `illegal` file doesn't override what factions set for commodity legality - it's used as the "default" legality of a commodity. I'll likely move it into commodity data instead (once I figure out how I'm going to handle having custom data defined in commodities for mods to consume), as it doesn't make sense to have that default parameter in its own file.

danielrogowski:
> My first thought _from a user perspective_ : JSON is a PITA to hand-edit (no comments, quotes everywhere, no trailing comma, mandatory uneasy-to-reach (on my layout) glyphs (`{}[]`). I **know** why this uses JSON, no need to introduce yet another dependency, but at least this downside is spelled out.

Could YAML be an easier alternative to JSON? If so, imho this would apply to all data files, not just for the commodities.

Web-eWorks:
> Could YAML be an easier alternative to JSON? If so, imho this would apply to all data files, not just for the commodities.

@danielrogowski I have considered using YAML and while it's attractive from a user's perspective; we're depending upon several features that depend on the JSON spec, including the ability to merge-patch JSON files which allows mods *far* more flexibility than before. I'll keep YAML on the back burner, but I don't have any serious plans to switch to it for mod files anytime soon.

Web-eWorks:
Finally, I think it's close to done. Would appreciate as much help as possible debugging this; all of the hard work should be done, but just poke around and let me know if you find any issues, missing translations, etc.

Web-eWorks:
Looking into @The-EG's issue, it looks like old saves serialize equipment objects in their entirety so commodity items don't have the correct fields. At the moment, I'm going to mark this as a savegame bump, but I'll put some more thought into the issue and see if I can find a work-around.

Web-eWorks:
As an extra-fun change, now saved cargo items are collapsed to only save one table per cargo type per ship instead of 4,000+ (one for each item) for the larger bulk ships. This change will be cherry-picked into another PR but I'd like some testing first! :smile:

data/pigui/libs/icons.lua
```diff
@@ -14,6 +12,20 @@ local function get_wide_icon_tex_coords(icon)
 	return Vector2(rem / count, quot/count), Vector2((rem+2) / count, (quot+1)/count)
 end
 
+local icons_texture_small = pigui:LoadTextureFromSVG(pigui.DataDirPath({"icons", "icons.svg"}), 16 * 24, 16 * 24)
+local icons_texture_med = pigui:LoadTextureFromSVG(pigui.DataDirPath({"icons", "icons.svg"}), 16 * 32, 16 * 32)
+local icons_texture_large = pigui:LoadTextureFromSVG(pigui.DataDirPath({"icons", "icons.svg"}), 16 * 64, 16 * 64)
+
+function ui.get_icons_texture(size)
```
Gliese852, line 19: Let say, size = 22. Why scaling from 24 pixels is better than from 64?

Web-eWorks: Moire patterns! More specifically, these icons are not created with mipmaps, so scaling from 64px to 22px causes a much more jagged and incorrect icon than scaling from 24px to 22px. You can see this with the navpoint icon and certain icons in the worldview - they retain much more of their shape when rasterized at 24px instead of at 64px and having the rasterized icon scaled down to <16x16.

Web-eWorks: It's definitely not perfect; some of our icons were never intended to display at less than 64px, but for text icons and other buttons it makes a small but noticeable difference. The first instinct is to think that the icon is more blurry when rendered from a smaller buffer, but when you go back the sharp, jagged edges and downsampling artifacts are *very* noticeable.

Web-eWorks:
Finally removed the `illegal.json` file and reworked how commodity legality is handled - factions which do not bother to specify custom illegality information inherit the default legality of commodities. In so doing I trimmed the PR back a little - I think it's finally ready for review!

Web-eWorks:
Since most of the code hasn't changed since the initial review (and I've had some smaller reviews on the Lua side) I'm planning on merging this later today. If you'd like to throw a sabot in the works, speak now or forever hold your peace!

PR #4947, opened 2020-08-28
Cleanup Renderer API to Remove Fixed-Function State

Web-eWorks:
Matrix state ~~must~~ did die!

Remove all OpenGL 2.x like matrix state methods from the renderer API, now the matrix state is the responsibility of the caller via MatrixTicket / StateTicket.

Still TODO is improving the viewport state tracking API, although that should be rather simple to fix via promoting the Viewport struct to the top-level Renderer API.

Needs a little dead-code-elimination because I wrote this between the hours of 3 and 4 AM, but should produce identical results to the previous system.

Ping @fluffyfreak - I recognize that `XXX` comment annotation and this was actually surprisingly simple to do, so there's gotta be a bug in it somewhere, right? *Right?*

### DIFF
--- a/src/core/GuiApplication.cpp
+++ b/src/core/GuiApplication.cpp
@@ -25,7 +25,7 @@ void GuiApplication::BeginFrame()
 	m_renderer->SetRenderTarget(m_renderTarget);
 #endif
 	// TODO: render target size
-	m_renderer->SetViewport(0, 0, Graphics::GetScreenWidth(), Graphics::GetScreenHeight());
+	m_renderer->SetViewport({ 0, 0, Graphics::GetScreenWidth(), Graphics::GetScreenHeight() });
 	m_renderer->BeginFrame();
 }
 
--- a/src/graphics/Renderer.h
+++ b/src/graphics/Renderer.h
@@ -79,10 +79,10 @@ namespace Graphics {
 		virtual bool ClearDepthBuffer() = 0;
 		virtual bool SetClearColor(const Color &c) = 0;
 
-		virtual bool SetViewport(int x, int y, int width, int height) = 0;
+		virtual bool SetViewport(Viewport vp) = 0;
+		virtual Viewport GetViewport() const = 0;
 
 		//set the model view matrix
-		[[deprecated]] virtual bool SetTransform(const matrix4x4d &m) = 0;
 		virtual bool SetTransform(const matrix4x4f &m) = 0;
 		virtual matrix4x4f GetTransform() const = 0;
 
@@ -144,9 +144,6 @@ namespace Graphics {
 
 		virtual bool ReloadShaders() = 0;
 
-		/// XXX at least use a custom structure you heathens
-		[[deprecated]] virtual void GetCurrentViewport(Sint32 *vp) const = 0;
-
 		// take a ticket representing the current renderer state. when the ticket
 		// is deleted, the renderer state is restored
 		// XXX state must die
@@ -156,6 +153,7 @@ namespace Graphics {
 				m_renderer(r)
 			{
 				m_renderer->PushState();
+				m_storedVP = m_renderer->GetViewport();
 				m_storedProj = m_renderer->GetProjection();
 				m_storedMV = m_renderer->GetTransform();
 			}
@@ -163,6 +161,7 @@ namespace Graphics {
 			virtual ~StateTicket()
 			{
 				m_renderer->PopState();
+				m_renderer->SetViewport(m_storedVP);
 				m_renderer->SetTransform(m_storedMV);
 				m_renderer->SetProjection(m_storedProj);
 			}
@@ -174,6 +173,7 @@ namespace Graphics {
 			Renderer *m_renderer;
 			matrix4x4f m_storedProj;
 			matrix4x4f m_storedMV;
+			Viewport m_storedVP;
 		};
 
 		// Temporarily save the current transform matrix to do non-destructive drawing.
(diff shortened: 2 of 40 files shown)


Web-eWorks:
For the record, I'm not 100% happy with this solution as we're still keeping some state in the renderer, but it's still miles better than retaining unnecessary fixed-function semantics in our API. I'll revisit this when I get around to replacing individual uniforms with UBOs, as that will provide a decent opportunity to remove this kind of state "for free".

src/graphics/opengl/RendererGL.cpp
```diff
@@ -603,11 +597,19 @@ namespace Graphics {
 	{
 		PROFILE_SCOPED()
 		//same as above
-		SetMatrixMode(MatrixMode::MODELVIEW);
-		LoadMatrix(m);
+
+		// SetMatrixMode(MatrixMode::MODELVIEW);
```
fluffyfreak, line 601: Remove instead of comment out

src/graphics/opengl/RendererGL.cpp
```diff
@@ -696,7 +706,8 @@ namespace Graphics {
 
 	void RendererOGL::SetMaterialShaderTransforms(Material *m)
 	{
-		m->SetCommonUniforms(m_modelViewStack.top(), m_projectionStack.top());
+		//m->SetCommonUniforms(m_modelViewStack.top(), m_projectionStack.top());
```
fluffyfreak, line 709: Remove instead of comment out

src/graphics/Renderer.h
```diff
@@ -145,59 +144,65 @@ namespace Graphics {
...
-		virtual void Translate(const float x, const float y, const float z) = 0;
-		virtual void Scale(const float x, const float y, const float z) = 0;
+		/// XXX at least use a custom structure you heathens
+		[[deprecated]] virtual void GetCurrentViewport(Sint32 *vp) const = 0;
 
 		// take a ticket representing the current renderer state. when the ticket
 		// is deleted, the renderer state is restored
 		// XXX state must die
-		class StateTicket {
+		class [[deprecated]] StateTicket
```
fluffyfreak, line 153: Deprecated?

fluffyfreak:
It looks good and I agree with the direction 👍
When I wrote the Renderer state tracking stuff it was to avoid the OpenGL calls querying state, which are a performance killer especially on ATi/AMD drivers. 
It was never really with a view to refactoring it like this so this is good progress!

src/graphics/opengl/RendererGL.cpp
Web-eWorks: This is that dead code elimination I was talking about - forgot to clean this up, but I still have one more commit to go, I'll clean it up then.

src/graphics/Renderer.h
Web-eWorks: I deprecated the existing methods (to see the extent of their use) before removing them, this is an unintentional slip.

Web-eWorks:
While I'm at it, I'm adding explicit type conversion between the `float` and `double` versions of `matrix4x4` and completely removing the "helper function" that was previously involved. I'll also be removing the `double` overload of `Renderer::SetTransform` and fixing things up so they communicate better to the reader that there's a loss of precision involved in passing double-based matricies to the renderer.

Web-eWorks:
Fixed up Viewport handling, removed the `matrix4x4d` overload of `Renderer::SetTransform` (it makes it a little more messy when calling, but also explicitly clarifies that you're losing precision), and ran some clang-format over files that have slipped out of format compliance.

This also has the side-effect of majorly cleaning up `GasGiantJobs.cpp`, as we no longer need twelve lines to enter/leave orthographic state and we also no longer stomp renderer state!

Web-eWorks:
@fluffyfreak let me know if you want to review the new changes - if you're swamped currently I'm good to go ahead and merge, but I wanted to make sure I got your sign-off first!

src/core/GuiApplication.cpp
```diff
@@ -25,7 +25,7 @@ void GuiApplication::BeginFrame()
 	m_renderer->SetRenderTarget(m_renderTarget);
 #endif
 	// TODO: render target size
-	m_renderer->SetViewport(0, 0, Graphics::GetScreenWidth(), Graphics::GetScreenHeight());
+	m_renderer->SetViewport({ 0, 0, Graphics::GetScreenWidth(), Graphics::GetScreenHeight() });
```
fluffyfreak, line 28: I need to test this kind of initialisation with Visual Studio builds, and non GCC builds basically.

fluffyfreak:
I'm all good with this conceptually, I just need to check that it builds in Visual Studio.
I don't have time to do so tonight, but hopefully will tomorrow.

src/core/GuiApplication.cpp
Web-eWorks: This *should* be standard-compliant list-initialization, but don't quote me on that. Appveyor didn't build because there's a merge conflict but I don't think it had problems with it.

Returning a list-initialized object (and this particular call) is a form of `copy-list-initialization`, which as far as I can tell is directly standard-compliant according to [the C++ reference](https://en.cppreference.com/w/cpp/language/list_initialization).

laarmen:
Yeah I don't think Andrew had any doubt about the code being standard
compliant, but more about VS knowing about the standard ;-)

That said, they've improved dramatically since the days of C++11.

src/core/GuiApplication.cpp
fluffyfreak: Yeah this isn't your fault or a C++ standards issue 😉 it's a does-VS-conform-to-the-C++-standards issue, which with this particular list initialisation I have seen it fail before, so I want to double check it before it's merged

fluffyfreak (approved):
This is all good from my perspective 👍

PR #4955, opened 2020-09-07
Camera control

zesterer:
Various improvements to the game's camera


src/ship/CameraController.cpp
```diff
@@ -219,29 +219,21 @@ void ExternalCameraController::Update()
...
-		m_smoothed_ship_orient = Quaternionf::lerp(
-			m_smoothed_ship_orient,
-			Quaternionf::FromMatrix3x3(GetShip()->GetInterpOrient()),
-			lerp_factor)
-									 .Normalized();
-		const matrix3x3d &smoothed_m = m_smoothed_ship_orient.ToMatrix3x3<double>();
+	// This lerping factor feels nice and scales non-linearly with larger (and slower-turning) ships
+	float lerp_factor = (2.5 + GetShip()->GetClipRadius() * 0.05) * Pi::GetFrameTime();
+	// Smooth the ship orientation by lerping toward the current (frame-invariant) orientation
+	Quaternionf frame_quat = Quaternionf::FromMatrix3x3(Frame::GetFrame(GetShip()->GetFrame())->GetOrient());
```
Web-eWorks, line 227: This line should be using `GetOrientRelTo(Pi::game->GetSpace()->GetRootFrame())` - it's very possible there are several frames in between the ship's frame and the root frame, which will make this calculation incorrect.

In the case that the root frame is the parent of the ship's frame, this is the same as `GetOrient()` so the frame-invariant behavior is preserved.

Web-eWorks (changes requested):
Looking good, just a slight issue where your frame-invariance isn't actually preserved in complex systems.
I'd appreciate it if you could confirm that you're not seeing any terrain jitter after changing frames multiple times and going on a short interplanetary flight!

src/ship/CameraController.cpp
zesterer: Changed!

Web-eWorks (approved):

PR #4956, opened 2020-09-15
Different tech levels for the thruster upgrades

robothauler:
Feels more realistic when the upgrades are not visible at all tech levels.

### DIFF
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -785,19 +785,19 @@ misc.hull_autorepair = EquipType.New({
 })
 misc.thrusters_basic = EquipType.New({
 	l10n_key="THRUSTERS_BASIC", slots="thruster", price=3000,
-	tech_level=1,
+	tech_level=5,
 	capabilities={mass=0, thruster_power=1}, purchasable=true,
 	icon_name="thrusters_basic"
 })
 misc.thrusters_medium = EquipType.New({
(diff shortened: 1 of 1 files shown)


fluffyfreak:
🤔 not sure about this, the reason I did it was because it's a bit like modding your car. Something open to all levels.
But I'm willing to hear what others thinks.

impaktor:
I'm positive, I'm all for differentiating space stations. Could even have minimum tech level x > 1.

By @fluffyfreak's description, it sounds more like it shold be something offered by the shipyard/ship repair shop, rather than the equipment shop?

nozmajner:
@impaktor Moving them to the ship repair shop sounds apt to me.

impaktor:
I don't want to merge this against @fluffyfreak's wishes, so it's up to him, I guess, to give thumbs up on this.

Web-eWorks:
To be honest, this is kind of a no-brainer change; if you want top-of-the-line high-performance parts it's not like you can just walk into your local AutoZone and pick them up off the shelf - you usually have to order them from the manufacturer, which gets more difficult in an interstellar society.

Given that ship equipment will be getting reworked sooner or later I wouldn't worry about moving them to a different station screen, and I'm personally all for giving them a (low) minimum tech level.

fluffyfreak:
@impaktor I'm not precious about it, people seem in favour of using the tech levels. Go for it 👍

PR #4957, opened 2020-09-15
Better container for bodies in space

Gliese852:
~~Add a `bunch` class, which is essentially a vector, but removal from the middle occurs in O(1), thanks to the swap-delete idiom. But this changes the order of the elements.~~
Remake the container for bodies in space from `std::list` to `std::vector`, and removing an element from the middle is done by moving the last element to its place.

### DIFF
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -157,8 +157,8 @@ Space::Space(Game *game, RefCountedPtr<Galaxy> galaxy, const Json &jsonObj, doub
 Space::~Space()
 {
 	UpdateBodies(); // make sure anything waiting to be removed gets removed before we go and kill everything else
-	for (std::list<Body *>::iterator i = m_bodies.begin(); i != m_bodies.end(); ++i)
-		KillBody(*i);
+	for (Body *body : m_bodies)
+		KillBody(body);
 	UpdateBodies();
 	Frame::DeleteFrames();
 }
@@ -278,7 +278,7 @@ void Space::RemoveBody(Body *b)
 #ifndef NDEBUG
 	assert(!m_processingFinalizationQueue);
 #endif
-	m_removeBodies.push_back(b);
+	m_assignedBodies.emplace_back(b, BodyAssignation::REMOVE);
 }
 
 void Space::KillBody(Body *b)
@@ -296,7 +296,7 @@ void Space::KillBody(Body *b)
 		// it still collides, moves, etc. better to just snapshot its position
 		// elsewhere
 		if (b != Pi::player)
-			m_killBodies.push_back(b);
+			m_assignedBodies.emplace_back(b, BodyAssignation::KILL);
 	}
 }
 
@@ -367,13 +367,13 @@ Body *Space::FindNearestTo(const Body *b, Object::Type t) const
 {
 	Body *nearest = 0;
 	double dist = FLT_MAX;
-	for (std::list<Body *>::const_iterator i = m_bodies.begin(); i != m_bodies.end(); ++i) {
-		if ((*i)->IsDead()) continue;
-		if ((*i)->IsType(t)) {
-			double d = (*i)->GetPositionRelTo(b).Length();
+	for (Body *const body : m_bodies) {
+		if (body->IsDead()) continue;
+		if (body->IsType(t)) {
+			double d = body->GetPositionRelTo(b).Length();
 			if (d < dist) {
 				dist = d;
-				nearest = *i;
+				nearest = body;
 			}
 		}
 	}
@@ -1002,23 +1002,26 @@ void Space::UpdateBodies()
 	m_processingFinalizationQueue = true;
 #endif
 
-	for (Body *rmb : m_removeBodies) {
-		rmb->SetFrame(FrameId::Invalid);
-		for (Body *b : m_bodies)
-			b->NotifyRemoved(rmb);
-		if (Pi::GetView()) Pi::game->GetSystemView()->BodyInaccessible(rmb);
-		m_bodies.remove(rmb);
+	// removing or deleting bodies from space
+	for (const auto &b : m_assignedBodies) {
+		auto remove_iterator = m_bodies.end();
+		for (auto it = m_bodies.begin(); it != m_bodies.end(); ++it) {
+			if (*it != b.first)
+				(*it)->NotifyRemoved(b.first);
+			else
+				remove_iterator = it;
+		}
+		if (remove_iterator != m_bodies.end()) {
+			*remove_iterator = m_bodies.back();
+			m_bodies.pop_back();
+			if (b.second == BodyAssignation::KILL)
+				delete b.first;
+			else
+				b.first->SetFrame(FrameId::Invalid);
+		}
 	}
-	m_removeBodies.clear();
 
-	for (Body *killb : m_killBodies) {
-		for (Body *b : m_bodies)
-			b->NotifyRemoved(killb);
-		if (Pi::GetView()) Pi::game->GetSystemView()->BodyInaccessible(killb);
-		m_bodies.remove(killb);
-		delete killb;
-	}
-	m_killBodies.clear();
+	m_assignedBodies.clear();
 
 #ifndef NDEBUG
 	m_processingFinalizationQueue = false;
--- a/src/Space.h
+++ b/src/Space.h
@@ -11,7 +11,6 @@
 #include "RefCounted.h"
 #include "galaxy/StarSystem.h"
 #include "vector3.h"
-#include <list>
 
 class Body;
 class Frame;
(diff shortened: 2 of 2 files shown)


Web-eWorks:
I apologize that I've taken so long to get to this @Gliese852; I've been checking IRC but somehow forgot to actually look at Github.

I'm going to play devil's advocate here and ask: do we really need to implement our own container? Can't we just make the body management code do swap-and-pop in (all of) two sites where it actually removes something from the containers? I definitely appreciate what you've done here implementing a minimal container, but I'd much rather just implement the needed idiom in a private function or in-situ to reduce cognitive complexity.

Otherwise, it's great that we're consolidating a list of pointers to a vector of pointers, that's definitely much much more efficient as both space and allocations go; if we need to preserve an order invariant, it should be extremely trivial to re-sort the main body list on frames where we've seen deletions.

Gliese852:
@Web-eWorks no problem, your path was one of the two alternatives I chose between, and it seemed more interesting to me to implement the container. Therefore, I do not insist, but I want to point out something to justify my approach:
- the logic and order of working with the container is concentrated in one place, it is made clear that this is an unsorted container that can simply contain objects
- the number of places for using this idiom can grow
- you can implement additional functions to work with this container, for example `remove_if` or even `move_to_if(predicate, another_container)`

But without much experience, it is difficult for me to understand how valuable it is in the long run, so I do not insist at all, I'll redo it.

P.S. I can't imagine a case for already sorted vector, where `swap_delete+sort` if faster than just `erase`.

Web-eWorks:
> But without much experience, it is difficult for me to understand how valuable it is in the long run, so I do not insist at all, I'll redo it.

I definitely applaud you for writing it; I've implemented a few std-like containers and adapters and they're great for gaining a better understanding of data structures!

In my eyes it doesn't really justify the added maintenance overhead when it's "just a vector but it swaps before erase()"; I'd have no problems if e.g. this were a lockless queue or an octree spatial querying structure - when the need for a new container class is apparent - but to be honest I don't think the problem is large enough. Similarly, I've written (twice now) a reverse-iterator adapter for a range-for loop and then realized that it's going to be used about four times in the entire codebase and doesn't justify the 50 lines it takes, even if it *does* make iterating a container backwards in a single statement so much easier.

As much as I hate to sound like I agree with [this article](https://realtimecollisiondetection.net/blog/?p=44), swap-and-delete is one of those extremely simple patterns that I don't think warrants a new name; much like Singletons are the "cool" and misunderstood form of a global variable.

Ultimately, what it comes down to is it's another 112 lines of code that we have to maintain and that contributors will have to get up to speed on, the name is even less intuitive than `std::vector` (no offense to you or the standards committee!), and the only thing that's actually unique in that file is the implementation of the `remove()` method. If that was harsh, I apologize; it's been a rough day, but that's my analysis of it.

> P.S. I can't imagine a case for already sorted vector, where swap_delete+sort if faster than just erase.

In the case which we have multiple bodies to delete each frame (potentially with a worst-case of 50% or more), using `std::vector::erase()` will invoke a `memmove()` equivalent on the entire right-half of the array *once for each deleted body*, whereas swapping, deleting, and then sorting at the end (ideally with a sort algorithm that's intended for mostly-sorted arrays) is far cheaper than thrashing the DMA controller and cache with every deletion.

src/Space.cpp
```diff
@@ -157,7 +157,7 @@ Space::Space(Game *game, RefCountedPtr<Galaxy> galaxy, const Json &jsonObj, doub
 Space::~Space()
 {
 	UpdateBodies(); // make sure anything waiting to be removed gets removed before we go and kill everything else
-	for (std::list<Body *>::iterator i = m_bodies.begin(); i != m_bodies.end(); ++i)
+	for (std::vector<Body *>::iterator i = m_bodies.begin(); i != m_bodies.end(); ++i)
 		KillBody(*i);
```
Web-eWorks, line 161:
```suggestion
	for (Body *body : m_bodies)
		KillBody(body);
```
I'd recommend using a range-for loop here instead; it saves a fair amount of typing and is pretty universally-accepted syntax.

src/Space.cpp
```diff
@@ -367,7 +367,7 @@ Body *Space::FindNearestTo(const Body *b, Object::Type t) const
 {
 	Body *nearest = 0;
 	double dist = FLT_MAX;
-	for (std::list<Body *>::const_iterator i = m_bodies.begin(); i != m_bodies.end(); ++i) {
+	for (std::vector<Body *>::const_iterator i = m_bodies.begin(); i != m_bodies.end(); ++i) {
```
Web-eWorks, line 370: Same thing here, you can replicate `const_iterator` behavior by using `for (const Body *body : m_bodies)` (incidentally, if you're using a const_iterator in a normal for-loop you want `vector.cbegin()`/`vector.cend()` instead of the normal begin/end pair.)

src/Space.cpp
```diff
@@ -1002,23 +1002,40 @@ void Space::UpdateBodies()
...
-
-	for (Body *killb : m_killBodies) {
-		for (Body *b : m_bodies)
-			b->NotifyRemoved(killb);
-		if (Pi::GetView()) Pi::game->GetSystemView()->BodyInaccessible(killb);
-		m_bodies.remove(killb);
-		delete killb;
+	// removing or deleting bodies from space in one pass
+	// using integer index instead of iterator, because pop_back() invalidates the iterator to the last element
+	for (uint32_t i = 0; i < m_bodies.size();) {
```
Web-eWorks, line 1007: I'm not sure this algorithm is optimal. The previous version was O(M * N), whereas this one seems like it's O(M * N log N), degrading to O(M log M * N<sup>2</sup>). This discrepancy mostly comes from the need to iterate the main bodies list, then do a linear search (protip: `std::binary_search` is useful for logarithmic-time searches of sorted vectors) of, in the worst case, both the removed and killed bodies list, and then do another full iteration of the main bodies list.

I think a much better way to do it is to store a `uint32_t` flag in a single list of bodies to remove, iterate the removed list once to notify bodies and delete, and then clear the removed list. That should (technically) be the most efficient method short of maintaining a separate list of bodies to receive delete notifications; which would drop it to O(M * log N).

src/Space.h
```diff
@@ -106,11 +106,11 @@ class Space {
...
 
 	// all the bodies we know about
-	std::list<Body *> m_bodies;
+	std::vector<Body *> m_bodies;
 
 	// bodies that were removed/killed this timestep and need pruning at the end
-	std::list<Body *> m_removeBodies;
-	std::list<Body *> m_killBodies;
+	std::vector<Body *> m_removeBodies;
+	std::vector<Body *> m_killBodies;
```
Web-eWorks, line 113: I'd recommend removing these two and replacing them with a single `std::vector<std::pair<Body *, uint32_t>>` list, and storing whether the body should be removed or killed as a flag in that integer. IIRC not many bodies actually need to go on the remove-but-not-kill list (possibly the other way around), and for efficiency reasons it's much more performant to only have one list of bodies to remove.

Web-eWorks (changes requested):
Thank you for working on this @Gliese852! I've left some feedback and outlined a few problem areas (yet more work, I know) but overall I appreciate the change!

src/Space.cpp
Gliese852: @Web-eWorks  it seems to me that since the notification loop is performed only once for object to be deleted, (after `std::find`, not during), the complexity of this algorithm is O(N * M + M * N) -> O(N * M) ?

Web-eWorks: You do have a point, it's closer to O(N + M * N) than it is to O(M * N log N). It's still a more expensive version of the algorithm especially at a lower number of deleted bodies due to needing an additional loop over all elements in the bodies list and then a linear search in all containers of deleted bodies, whereas the prior version at least didn't incur two `std::find`s for every live body in the space.

Don't get me wrong, I appreciate the work you've gone to in implementing this! I'd just like to be sure we're using the most efficient and effective algorithms when the PR is intended to use a more efficient and effective data container...

Web-eWorks:
@Gliese852 Thanks! Other than the two instances where you could use a range-for instead of an iterator loop, this PR looks good to me!

What's the state of testing on this PR? Have you noticed any issues with body iteration order? I don't *think* there would be any...

src/Space.cpp
Gliese852: done!

src/Space.cpp
Gliese852: done!

Gliese852:
@Web-eWorks I tested a little, everything seems to work

Web-eWorks (approved):

Web-eWorks:
Will give this a bit further testing but the code looks good to me!

impaktor:
> Will give this a bit further testing but the code looks good to me!

Bah, player can test it, if code is fine, and no obvious crash, we can merge and see what happens :) 
(assuming this frees up time for the coders to code more, then I think player's can test it for us)

Web-eWorks:
That's my bad; I said I'd test it and then I got busy both in real life and with "ooh shiny" things in the codebase. I think this can merge, as long as Gliese hasn't noticed any issues in his testing I'm good to pull the trigger.

PR #4958, opened 2020-09-18
Ship-specific atmospheric pressure limits

WKFO:
The relevant code for ship-specific atmospheric pressure limit has already existed in Ship.cpp (see line 1235) but this feature was not implemented for some reason and every ship defaulted to 10atm because Ship.cpp told so.

With this PR, all ships have their own external pressure limits, which are now shown in ship market (with and without shield) and ship info (with shield if installed) view.

- [x] Set custom pressure values for ships
- [X] Display relevant info in ship market and ship info
- [x] Display the effect of atmospheric shield (e.g. if base = 8atm, shielded = 25atm; it will be shown as **8(+17) atm**)
- [Not doing this anymore] Set flight gauge maximum value to player ship max value (as opposed to constant 15 atm)
- [x] Make autopilot reject docking order if planet surface pressure is too high (extreme case scenario, but why not have a fail-safe)

 ̷B̷e̷f̷o̷r̷e̷ ̷y̷o̷u̷ ̷s̷a̷y̷,̷ ̷I̷ ̷w̷i̷l̷l̷ ̷f̷i̷x̷ ̷t̷h̷o̷s̷e̷ ̷t̷a̷b̷s̷-̷a̷n̷d̷-̷s̷p̷a̷c̷e̷s̷ ̷m̷e̷s̷s̷ ̷i̷n̷ ̷s̷h̷i̷p̷ ̷J̷S̷O̷N̷ ̷f̷i̷l̷e̷s̷.̷ ̷T̷o̷m̷o̷r̷r̷o̷w̷.̷ (Done)

I usually read the wiki while doing this, but let's ping @nozmajner for wisdom on balancing the values.


nozmajner:
Nice!
Oh my, so the rebalance is more and more in order?
I could do a pass somewhere this month, if you think the time is ripe. (The Sino overhaul will take more time anyway)
I'd really like to try a more nerfed approach, capping Gs at 4-5 tops. To see how it handles trough a couple of releases, then evaluate.

WKFO:
I was mainly thinking of atmospheric pressure limits for this PR but since I already went down this rabbit hole and nozmajner is also into it, maybe we can handle the whole ship stats rebalance in a follow-up PR. I don't see why not.

I fixed the tabs/spaces.

WKFO:
Whenever I try to use a function to set the max value of the pressure gauge everything just disappears for some reason, so that might not happen...

WKFO:
Autopilot should now refuse docking to ground starports if the atmospheric pressure there is too high. I only did minimal testing so far. The clang format shows red, I will look at it later.

Web-eWorks:
@nozmajner if you're doing a full rebalance pass, make sure to set up drag coefficients properly for all ships, only a few of them have them set from the original atmo flight PR; the rest are "perfect spheres".

nozmajner:
Right, thanks for the reminder. Those aren't in my spreadsheet yet.

WKFO:
I was asked about the status of this PR a few days ago, but I also want to get nozmajner's opinion since he knows more about ships than a lot of us; so @nozmajner ,would you rather;
a) hold this PR until rebalanced values are chosen (at least for the pressure limits)
b) consider this as a feature implementation and leave picking the values on a big "complete ship stats rebalance" PR?

nozmajner:
I'd say don't wait around for the rebalance. I was thinking about the values the past couple of days, but I need more time for the rebalance.

WKFO:
If the code looks fine, I say this is ready for merge.

impaktor:
Hold onto your hats, ladies and gentlemen

impaktor:
Hold onto your hats, ladies and gentlemen

PR #4959, opened 2020-09-23
Clean up PiGUI Lua

The-EG:
This has come about after some discussion in IRC about the pigui.lua module needing to be split into separate modules for organizational and maintenance reasons.

This is pretty big, but there should be no change in functionality as a result. All Lua using pigui should still use `local ui = require 'pigui'`.  Anything still using `local ui = import 'pigui/pigui.lua'` will not work (that file no longer exists), but I have changed all ~the ones I have found so far~ existing occurrences.

data/pigui/pigui.lua is now contained in several modules:
  - data/pigui/libs/forwarded.lua - functions directly forwarded from C++
  - data/pigui/libs/wrappers.lua - convenience wrappers (mostly)
  - data/pigui/libs/baseui.lua - a base set of the ui functions (forwarded + wrappers + basic shared config)
  - data/pigui/libs/buttons.lua - button convenience functions
  - data/pigui/libs/icons.lua - icon related stuff
  - data/pigui/libs/text.lua - text display related
  - data/pigui/libs/gauge.lua - gauge convenience functions
  - data/pigui/libs/radial-menu.lua - radial menu convenience functions

In addition to organization, I've taken a shot at documenting most of the functions defined in the above modules. I did not document the forwarded functions, as those should be documented where they are defined on the C side.

This is still in draft as I need to finish up documenting some of the functions (text.lua), there are some functions that I think can be removed (currently commented out) and I want to do some more testing to ensure this doesn't break the UI anywhere.


Web-eWorks:
Looking good, this is something that we've needed for a while but I've been too overwhelmed with engine internals to make major Lua-side changes.

> In addition to organization, I've taken a shot at documenting most of the functions defined in the above modules. I did not document the forwarded functions, as those should be documented where they are defined on the C side.

Thank you very much! This is amazing and way beyond what I was asking for; thank you!

If you still have energy after documenting all of those functions, could I ask you to take a look at the `import()` warnings logged in output.txt and clean up the ones in the `pigui/` folder? I had a branch that cleaned some of them, but I don't know if it ever got merged and is likely horribly out of date.

Feel free to say no, this on its own is a great candidate for merge!

The-EG:
> If you still have energy after documenting all of those functions, could I ask you to take a look at the `import()` warnings logged in output.txt and clean up the ones in the `pigui/` folder? I had a branch that cleaned some of them, but I don't know if it ever got merged and is likely horribly out of date.

Sure, that's actually pretty easy, I think I had done half of them already since I had to fix any 'import "pigui/pigui.lua' already.

Web-eWorks:
@The-EG I'd like to merge this very shortly after #4963 to take advantage of the new support for loading theme files from lua.

I'd like there to be (at least) one file in this PR that only depends on `Engine.pigui` (and possibly not even that; I'm considering reworking where to hang the PiGui object from and whether it needs to be an actual object or just a namespace) that handles creating the `ui` object and loading and registering themes (plus whatever low-level work can be accomplished without any other dependencies).

This restriction comes from the fact that themes need to be loaded before the `Lua::InitModules()` step to properly style the loading screen, so the only Lua types and objects registered from C++ at that time are those in `PiGui::Lua::Init()` and `Lua::InitMath()` from #4963.

The-EG:
> I'd like there to be (at least) one file in this PR that only depends on `Engine.pigui`

How many modules that depend on `Engine.pigui` or where ever the c++ side is exposed, will depend on how many of the 'base' functions are exposed. Multiple files in libs/ currently depend on Engine.pigui because they use functions not exposed in /libs/forwarded.lua, generally because they are wrapping functionality there. 

The theme should be fine...there's only a single place that the theme is set now, in libs/baseui.lua, and everything else just references ui.theme. Once the theme is being handled elsewhere it can just be forwarded like the rest.

Web-eWorks:
Now that #4963 is merged, I think the only blocker on this PR is carrying over the `ui.registerTheme` function from master into the newly-split files; that seems to be the only thing that Git is complaining about when I tested a rebase.

Let me know when you're ready for review; I know you wanted to continue documenting more functions and (presumably) clean up the rest of the uses of `import`, I'll be ready to merge whenever you are.

The-EG:
> Now that #4963 is merged, I think the only blocker on this PR is carrying over the `ui.registerTheme` function from master into the newly-split files; that seems to be the only thing that Git is complaining about when I tested a rebase.

I put ui.registerTheme into libs/wrappers and the call to ui.registerTheme('default', defaultTheme) into libs/baseui.lua (which is also where defaultTheme is defined now.

Note, ui.theme is still defined, and presumably used (as the value of defaultTheme).

> Let me know when you're ready for review; I know you wanted to continue documenting more functions and (presumably) clean up the rest of the uses of `import`, I'll be ready to merge whenever you are.

It's probably a good time for review. There should be no imports left under data/pigui. 

I'd like to document more, but this weekend has sapped my motivation, so if there's interest to get this in sooner or later I can always come back and finish up the documentation later. My main motivation behind the documentation was making it easier for others to move things to PiGui without having to delve into the C++ side to figure out what functions do. As-is, it should be better than it was in that respect.

data/pigui/libs/wrappers.lua
```diff
@@ -0,0 +1,758 @@
...
+--local function maybeSetTooltip(tooltip)
+--	if not Game.player:IsMouseActive() then
+--		pigui.SetTooltip(tooltip)
+--	end
+--end
+
+function ui.registerTheme(name, theme)
+	assert(type(theme) == "table", "UI Themes must be table values!")
+	pigui.handlers[name] = theme
+end
```
Web-eWorks, line 758: This should go in the same file with `ui.registerHandler()` - they serve the same purpose.

data/pigui/libs/text.lua
```diff
@@ -0,0 +1,260 @@
...
+end
+
+ui.addStyledText = function(position, anchor_horizontal, anchor_vertical, text, color, font, tooltip, bg_color)
+	-- addStyledText aligns to upper left
+	local size = Vector2(0, 0)
+	ui.withFont(font.name, font.size, function()
+								size = pigui.CalcTextSize(text)
+								local vert
+								if anchor_vertical == ui.anchor.baseline then
+									vert = nil
```
Web-eWorks, line 234: While we're at it, I'd appreciate it if you could fixup the various whitespace issues; we'd like to strive for a four-wide physical-TAB-only indentation. New scopes should introduce a single level of indentation, these alignment issues are left over from a previous contributor's emacs settings.

data/pigui/libs/wrappers.lua
```diff
@@ -0,0 +1,758 @@
...
+	end
+
+	pigui.BeginChild(id, size, flags)
+	fun()
+	pigui.EndChild()
+end
+
+function ui.registerHandler(name, fun)
+	pigui.handlers[name] = fun
+end
```
Web-eWorks, line 262: This function should ideally go in `baseui.lua`.

Web-eWorks (changes requested):
I think I'd ideally like `pigui/libs/baseui.lua` to be moved to `pigui/baseui.lua` so it's a little easier to reference as an alternative to the "full" implementation; in contexts where it's important to load the minimum set of UI features before C++-side Lua modules are registered (e.g. when displaying the init screen) I'd like to be able to `require 'pigui.baseui'` instead of having to delve into internal libraries.

Otherwise just some minor organizational and stylistic changes requested; now is probably the best time to fix all of the indentation issues we had in `pigui.lua`, so if you still have motivation I'd like to do at least a cursory pass over the worst offenders.

data/pigui/libs/wrappers.lua
The-EG: Ok, both this and `ui.registerHandler` are now in baseui.lua.

data/pigui/libs/text.lua
The-EG: That's now cleaned up and I've fixed the other files in libs/. I'm hesitant to make massive changes to views/ or modules/ due to how it could affect any in-progress development. On a glance through, I didn't see anything as bad as the above, but there's probably some spaces vs. tabs still there in places.

data/pigui/libs/wrappers.lua
The-EG: Moved to baseui.lua with `ui.registerTheme`.

The-EG:
In addition to the review comments, I've put ui.maybeSetTooltip into libs/wrappers. This is used by a few things (libs/icons & libs/text) and ui.setTooltip was set to be `= maybeSetTooltip` instead of the base function.
I'd like to just expose one or the other, but they've both been used throughout the UI, so that will require some additional, broader cleanup.

Because of that function being included in wrappers, it now also brings in `require 'Game'`. That's because `ui.maybeSetTooltip` checks if the player is using the mouse to navigate and only shows the tooltip if not.

The-EG:
> I think I'd ideally like pigui/libs/baseui.lua to be moved to pigui/baseui.lua so it's a little easier to reference as an alternative to the "full" implementation; 

Oops, I missed this when going through things. That's now done too. 

The only thing that's keeping baseui.lua from being loaded before everything else, excluding Engine or where ever PiGui is exposed of course,  is `ui.maybeSetTooltip` in wrappers.lua. It requires `Game.player:IsMouseActive()`.

The-EG:
Also, I've opened up every screen, hovered over, clicked, right clicked, etc. everything I could think of...flown around, used the auto-pilot etc. 

I've already fixed any issues that have come up, and everything appears to be working properly. But if there's an edge or weird cases to test, let me know.

data/pigui/baseui.lua
```diff
@@ -0,0 +1,71 @@
...
+ui.anchor = { left = 1, right = 2, center = 3, top = 4, bottom = 5, baseline = 6 }
+
+
+function ui.get_icon_tex_coords(icon)
+	assert(icon, "no icon given")
+	local count = 16.0 -- icons per row/column
+	local rem = math.floor(icon % count)
+	local quot = math.floor(icon / count)
+	return Vector2(rem / count, quot/count), Vector2((rem+1) / count, (quot+1)/count)
+end
```
Web-eWorks, line 31: This file is using two-space indentation; if you could convert it to one-tab that'd be excellent!

data/pigui/libs/buttons.lua
```diff
@@ -0,0 +1,141 @@
...
+--
+--   boolean - true if button was clicked
+--
+function ui.imageButton(icon, size, frame_padding, bg_color, tint_color, tooltip)
+	local uv0, uv1 = get_icon_tex_coords(icon)
+	ui.withID(tooltip, function()
+		local res = pigui.ImageButton(ui.icons_texture, size, uv0, uv1, frame_padding, bg_color, tint_color)
+	end)
+	return res
+end
```
Web-eWorks, line 36: As is this file!

Web-eWorks (changes requested):
Looks like two files slipped through the cracks; otherwise it looks good to me! `maybeSetTooltip` is an acceptable casualty; I'll take a crack at it later.

EDIT: okay, it shows as two-space indentation in the Changed Files view but one-tab here in the Requested Changes view...

data/pigui/baseui.lua
The-EG: Weird, it definitely has tabs (CR, LF, TAB):
image

data/pigui/libs/buttons.lua
The-EG: Same!
image

data/pigui/baseui.lua
Web-eWorks: Alright, that's on Github then. Thank you for looking into it so closely!

Web-eWorks (approved):

The-EG:
> Looks like two files slipped through the cracks; otherwise it looks good to me! `maybeSetTooltip` is an acceptable casualty; I'll take a crack at it later.
> 
> EDIT: okay, it shows as two-space indentation in the Changed Files view but one-tab here in the Requested Changes view...

Those two are definitely tabs (see above). Not sure why GitHub is showing spaces like that...
I see there are a couple that _look_ like they aren't 2 space indented in the Changed Files...those may be 4 spaces. If so I'll fix them.

Web-eWorks:
If there are any that slipped through the cracks feel free to put them all in a single commit and push it directly to master. Otherwise, I think this PR is good to merge!

PR #4960, opened 2020-09-23
Show lawlessness

danielrogowski:
Adds a "lawlessness" indicator to the Demographics screen (without the red box of course):
screenshot_lawlessness

I know it wasn't something viewable in Frontier, but I think it's really useful info for Newbies (like me). Additionally I think from an immersive view, such information would be available (like criminal statistics in the real world).

The strings are only provided in the english core lang file. I didn't find any contribution rules regarding this topic.

I adhered to the contribution as best I could. If there's something amiss, please feel free to reach out to me and I will fix it.


impaktor:
Don't we show lawlessness in sector view anymore?

We ussed, to so I assume there should all ready be strings somewhere?

Not sure about the wording, "Very litle" feels strange.

The-EG:
> Don't we show lawlessness in sector view anymore?

Not currently. It's may be in some of the mockups. I do think it makes more sense for the demographics and system info to be in the sector map eventually.

danielrogowski:
> > Don't we show lawlessness in sector view anymore?
> 
> Not currently. It may be in some of the mockups. I do think it makes more sense for the demographics and system info to be in the sector map eventually.

Sounds reasonable.

The-EG:
> The strings are only provided in the english core lang file. I didn't find any contribution rules regarding this topic.

Providing the English strings within the PR is the correct approach. Translations are handled through Transifex and get automatically merged back in later: https://pioneerwiki.com/wiki/Translations

I think every time someone touches an older part of the UI, we end up talking about _which_ lang module is used...There's definitely still UI related stuff in core but it would probably be better to add any new UI stuff to ui-core.

danielrogowski:
Is there a plan to completely remove Demographics tab?

As long as it's in place, it is a good idea to display something like "lawlessness" or "danger level" or "criminal activity" or something like that there. Or isn't it?

impaktor:
Yes, we want the information, on system level, but this screen is scheduled for removal, which is why we might re-consider it

Web-eWorks:
The SystemInfo view is intended to be merged into the SystemMap view as an alternate display mode and some of the information also displayed in the SectorMap view. However, there's a fair amount of work still needed (and a few things it's waiting for like #4944) before I'm able to commit to that work.

For now I think this change is small enough that we can merge it in; we'll need a way to move translation strings between translation resources eventually (yeah I know, on the backlog it goes) so I don't have any qualms about it being in the wrong module *right now*... we have plenty of those already.

This gets a conditional thumbs-up; I'm not going to be able to test it for a few days but if someone wants to sign-off on it you've got my go-ahead to merge.

impaktor:
I'm not ready to let this through just yet. I do want to show lawlessness (like we used to), but the following issues irk me:

1. I don't like the phrasing, there are too many, and some are strange e.g. "A little"? Current table summarizes this PR:

| Variable            | Description         |  value |
|---------------------|---------------------|--------|
| ALMOST_NON_EXISTENT | Almost non-existent | < 0.01 |
| VERY_SAFE           | Very little         |   0.05 |
| SAFE                | A little            |    0.1 |
| MODERATE            | Moderate            |   0.25 |
| HIGH                | High                |    0.4 |
| VERY_HIGH           | Very High           |    0.6 |
| EXTREME             | Rampant             |  > 0.6 |

2. I think the intervals are too many and of un-even resolution. E.g. is it of interest for the player to resolve the difference between 0.01 and 0.05? (perhaps something like, e.g.: 0%, 33%, 66%, 100% ?)

3. Question is what these levels will entail? Ideally, there should be modifications in the pirate and patrol scripts, so no pirates if e.g. fully safe, and no police patrols if fully unsafe? Perhaps then the special cases of identical 0 and identical 1 have some own term, assuming some other game play follows.

4. You also need to provide `description` string to translators, e.g. "Describes lawlessness of a system"

5. squash commits?

The-EG:
In regards to the, it may help to consider it's also somewhat coupled to the government type and economy type. Each type has a base value for lawlessness, which, I'm guessing, procedurally created systems will then modify with a random seed:

| Gov. Type | Base Lawlessness 
|:-|-:
| No Central Gov. | 100% 
| SolFed Colony | 30%
| SolFed Democracy | 15%
| Imperial Rule | 15%
| Social Democracy | 20%
| Liberal Democracy | 25%
| Corporate | 40%
| Military Dictatorship (capitalist - 1) | 40%
| Military Dictatorship (capitalist - 2) | 25%
| Military Dictatorship (mixed - 1) | 25%
| Military Dictatorship (mixed -2 ) | 40%
| Communist | 25%
| Plutocratic Dictatorship | 45%
| Violent Anarchy | 90%

So instead of 'very little,' maybe something getting at: 'organized police force and judicial system'. And to the other end of the spectrum of 'unorganized local militias.'

Gliese852:
I personally like the simple system in Eve online - 1.0 is very safe, 0.0 is absolutely dangerous

impaktor:
I think the general idea of this PR is good:
1. we should show lawlessness (like we used to) - note: on system level
2. however, we should not show it as the underlying engine's float representation `∈ [0,1]`, instead:
3. show it as strings that quantify it, but not in too many steps, I like what @The-EG suggests. 

That said, what I don't like about the current PR is the wording, and the number of levels. Sorry to expand upon the intention on this PR.

Ideally, player should not a difference between the different lawlessness-levels. (this actually raises the question if it should be discrete in the engine as well, but that's for another day)

impaktor:
Honestly, I think we should, as first iteration on this, just go with something very simple, not too many levels, and have each be very clear, also, able to infer their order, without having seen all available labels for different lawlessness. That leads me to suggest that it takes the following:
```
lawlessness: LOW, MEDIUM, HIGH
```
(possibly identical 0 and identical 1 should be "special", but that might be more than we want to do in this PR.)

Web-eWorks:
The SystemInfoView has been removed and the pertinent information is being merged into the SectorMap / SystemMap views. Please feel free to port the code of this PR over to the new UI screens and re-open, but I'm unfortunately closing this PR as it's no longer applicable.

PR #4963, opened 2020-09-26
Collapse PiGui namespaces, load theme from Lua

Web-eWorks:
Collapse all differences between the PiGui and PiGUI namespaces, reorganizing files as needed. This will cause a bit of code churn but it's been needed for a while.

Cleaned up a few deprecated style var names while I was at it, as well as adding a few new ones that were originally overlooked.

Also implemented loading ImGui themes from lua files (piggybacking the existing theme file structure), removing the need for a hard-coded style function in PiGui.cpp.

This loading needs a bit more thought before it's fully baked, but I wanted to make sure all of the other changes in this PR are well tested while I noodle on it.

@The-EG if you could ensure this compiles and launches on Windows I'd appreciate it; I know MSVC has some issues with files of the same name but different directories gumming up the build.


The-EG:
Windows cmake works, the VS2019 solution does not. There may actually be changes from earlier merges that broke it too, so I'll fix those first if so and then this one.

The-EG:
Not as simple as I was hoping, but still easy enough. There was another object name clash:
https://github.com/The-EG/pioneer/tree/sturnclaw-pigui

Specifically: https://github.com/The-EG/pioneer/commit/09646df2f772e2682158cf732123ec4b9b8c6267

Web-eWorks:
Hmm... is there a reason there are so many `Condition=` variations for the same object file? Is that just VS being VS or could it be defined unconditionally? It doesn't really matter that much, just something that stuck out to me. Thanks @The-EG!

The-EG:
> Hmm... is there a reason there are so many `Condition=` variations for the same object file?

VS compiles everything all sources for a project into the same directory. src/Lua/LuaPiGui.cpp and src/pigui/LuaPiGui.cpp both end up being compiled to <arch>/<config>/LuaPiGui.obj. 

I haven't found a nice way of telling it not to do that (by use of a variable or something). Specifying a directory for one of the files (pigui/LuaPiGUI.obj) is the most straightforward way I found around the issue. There's multiple lines because it has to be done for every build configuration (Debug, Release, PreRelease, Profile) and architecture combination (x64, x86).

Issue #4962, opened 2020-09-24
Current official PSS version (20200203) for 32-bit Linux (and Windows) systems


gerrygk:
Hey all,

lately I was looking for a current official PSS version (20200203) for Linux x86 / 32-bit systems. The only one I found was an older version from 2019 on Flathub (not interesting for me).

It seems to me that the newest official PSS versions for all 32-bit OS's are no more being offered at the Pioneer website for some time now. This could exclude a whole bunch of users from trying and evaluating the newest development results (at least for the Linux community), what would be a real pity.

Maybe it's just a matter of effort, but I would appreciate this versions to be back in the download list again.

What's your opinion on that?

impaktor:
If you're running a very old computer, chances are it won'd support our OpenGL 3.1

The-EG:
Also #4788 .

I think the issue being raised here is that 32bit builds aren't deployed when a new release is built. I suspect that they aren't built as a part of the appveyor/travis process. Compiling 32bit should definitely be working now, with the exception of Windows when using the cmake system (the VS2019 solution does work), but do we want to also provide pre-built 32bit binaries?

p.s. Fluffyfreak, you can now say I told you so!

impaktor:
If anyone can make 32 bit build work, and artifact be made, and ideally, uploaded/linked somewhere, I don't think anyone will object.

gerrygk:
As the source code for v2020-02-03 is still available here, I could give it a try. (I once succeeded when building a Linux 64-bit specimen after February 2020.) I remember having seen a (hopefully up-to-date) list of programs, libraries etc. necessary therefor, but couldn't remember where.

impaktor:
@gerrygk compiling for 32 bit computer from current master should be possible, if not paste error.

gerrygk:
Today I tried the compilation. Unfortunately, it stopped with the following error message:

**********************************************

In file included from /home/gerry/Builds/pioneer-20200203/src/CollMesh.cpp:6:0:
/home/gerry/Builds/pioneer-20200203/src/scenegraph/Serializer.h:29:2: error: static assertion failed: Vector2d is padded differently on this platform and will not serialize properly.
  static_assert((sizeof(vector2d) == 16 && alignof(vector2d) == 8), "Vector2d is padded differently on this platform and will not serialize properly.");
  ^~~~~~~~~~~~~
[...]

PR #4964, opened 2020-09-27
Fix VS2019 Solution

The-EG:
image

Multiple files moved between src/ <-> src/ship/
src/View.cpp and src/pigui/View.cpp were being compiled to view.obj. This caused some weird linking issues: symbols from one would always be missing, but which one depended on what file was compiled last. 
Changed src/pigui/View.cpp to compile to pigui/view.obj


Web-eWorks:
Thank you @The-EG for the timely fix! I'm afraid it'll keep breaking...

The-EG:
> Thank you @The-EG for the timely fix! I'm afraid it'll keep breaking...

Yes, it will. I think I'll try to get 32bit building working for the cmake windows build (#4962), and at that point we can probably really think about using only cmake.

PR #4966, opened 2020-09-29
Define ui.oneOverSqrtTwo (fixes 4965)

### LINKED ISSUE #4965: 2D radar does not work anymore
### Observed behaviour
Game stops with error message if player switches to 2D radar.

Fatal: [string "[T] @pigui/modules/radar.lua"]:63: attempt to perform arithmetic on local 'l' (a nil value)
stack traceback:
        [string "[T] @pigui/modules/radar.lua"]:63: in function 'display2DRadar'
        [string "[T] @pigui/modules/radar.lua"]:142: in function 'fun'
        [string "[T] @pigui/views/game.lua"]:107: in function 'callModules'
        [string "[T] @pigui/views/game.lua"]:286: in function 'fun'
        [string "[T] @pigui/libs/wrappers.lua"]:138: in function 'window'
        [string "[T] @pigui/views/game.lua"]:276: in function 'fun'
        [string "[T] @pigui/libs/wrappers.lua"]:534: in function 'withStyleColors'
        [string "[T] @pigui/views/game.lua"]:275: in function <[string "[T] @pigui/views/game.lua"]:267>Fatal: Frame instance deletion outside 'DeleteFrame' [0]

### Expected behaviour
No game crash

### Steps to reproduce
Start a new game
Right click on the radar
Select 2D radar

current master (GNU/Linux)

The-EG:
Looks like I missed a variable being defined when I split the PiGui Lua modules apart, breaking 2D radar. 

`ui.oneOverSqrtTwo` and `ui.standardGravity` are now defined again in pigui/baseui.lua.

Fixes #4965

### DIFF
--- a/data/pigui/baseui.lua
+++ b/data/pigui/baseui.lua
@@ -13,6 +13,8 @@ local two_pi = pi * 2
 local standard_gravity = 9.80665
 local one_over_sqrt_two = 1 / math.sqrt(2)
 
+ui.oneOverSqrtTwo = one_over_sqrt_two
+ui.standardGravity = standard_gravity
 ui.theme = defaultTheme
 ui.twoPi = two_pi
 ui.pi_2 = pi_2
--- a/data/pigui/libs/forwarded.lua
+++ b/data/pigui/libs/forwarded.lua
@@ -50,7 +50,6 @@ ui.getTextLineHeight = pigui.GetTextLineHeight
 ui.getTextLineHeightWithSpacing = pigui.GetTextLineHeightWithSpacing
 ui.lowThrustButton = pigui.LowThrustButton
 ui.thrustIndicator = pigui.ThrustIndicator
-ui.oneOverSqrtTwo = one_over_sqrt_two
 ui.isMouseClicked = pigui.IsMouseClicked
 ui.isMouseDown = pigui.IsMouseDown
 ui.getMousePos = pigui.GetMousePos


Gliese852:
note: https://github.com/pioneerspacesim/pioneer/blob/a3d3f5c1f09f897276be4b2437246a2b5b0631e2/data/pigui/libs/forwarded.lua#L53

The-EG:
> note:
> 
> https://github.com/pioneerspacesim/pioneer/blob/a3d3f5c1f09f897276be4b2437246a2b5b0631e2/data/pigui/libs/forwarded.lua#L53

Thanks, that should actually be removed from forwarded.lua then. I thought I had gotten them all, but this one slipped through. It doesn't exist yet when forwarded is loaded, so it's just nil at that point, until it's defined in baseui.lua.

fluffyfreak (approved):

impaktor:
@The-EG do the honor of using the web interface to merge/rebase: I suggest using the drop-down menu next to the green merg button and select **rebase and merge**, that way we don't need a merge commit for a PR with single commit. I'd also skip writing anything in Changelog.txt, since no one will be aware of this bug, introduced a few hours ago.

Gliese852:
can't believe things like this aren't optimized (I mean calculations over constants)

Web-eWorks:
Because Lua is a dynamic language, the amount of optimization the compiler can do is severely limited. Any call to `math.sqrt` cannot be optimized because you may well be using a totally different math library that does something completely different than a standard sqrt with that function. There *might* be a limited amount of CSE (constant subexpression elimination, depends on the Lua implementation; LuaJIT does a lot more, PUC-Rio probably doesn't do that to keep the implementation manageable) so you'd want to write that variable as `1 / 2^(1/2)` to avoid the call to `sqrt`.

Gliese852:
@Web-eWorks thanks a lot for the explanation!

PR #4967, opened 2020-10-03
Move Economy & Trade -view to pigui

impaktor:
## Thoughts
Working with this, I think there's a lot of stuff that's overlapping with "ship info". I'd suggest "ship info" only show the static info on that ship's stats, and equipment, and this "economy & trade view" shows all the stuff about cargo, fuel, interna-tank mass, and range. Possibly, this could be outside the scope of this PR.

Also, I had a wrestling match against the colorselectbutton since, in my estimation, it makes no sense, in how it tints the button, but for now, it's some kind of deactivated when landed (one cannot jettison when landed). I got the feeling I was using the wrong function, since it's not doing what I want, which is what the name suggestes it's supposed to do. Theme-issues.

## Bug exposed in master?
I remember playing with this in june/july, and finding some inconsistency  in how we define "cargo space" relative internal fuel mass, but I'm not sure I notice what it was. I.e. (back then) I thought the "Used" cargo space displayed in the cargo gauge bar was wrong.

## What remains
At the moment I'm not quite sure how to do the first two in this list:
- [ ] Move jettison-buttons further right?
- [x] Fix there not being any margin to the left
- [x] Clean up / remove commented out code experiments
- [ ] Tool tip doesn't seem to work?
- [x] Settle on what info should be displayed / how?
  - should progress gauges be shown (at all? / or like now, clustered under "summary")?
  - should we show cash?
- [x] squash some commits
 
2020-11-10-092015_1600x900_scrot

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -527,6 +527,10 @@
     "description": "Hyperjump planner, referring to star system target",
     "message": "Final Target:"
   },
+  "FINANCE": {
+    "description": "Financial statement of player, e.g. cash",
+    "message": "Finance"
+  },
   "FLIGHTLOG_DOCKING": {
     "description": "Used in Flight Log, to indicate player's position. 'primary_info' is a station name",
     "message": "Docking at {primary_info}"
--- a/data/libs/EquipSet.lua
+++ b/data/libs/EquipSet.lua
@@ -51,9 +51,9 @@ function EquipSet:AddListener(listener)
 	listeners[self] = listener
 end
 
-function EquipSet:CallListener()
+function EquipSet:CallListener(slot)
 	if listeners[self] then
-		listeners[self]()
+		listeners[self](slot)
 	end
 end
 
@@ -213,7 +213,7 @@ function EquipSet:__TriggerCallbacks(ship, slot)
 	else
 		ship:setprop("totalCargo", math.min(self.slots.cargo.__limit, self.slots.cargo.__occupied+ship.freeCapacity))
 	end
-	self:CallListener()
+	self:CallListener(slot)
 end
 
 -- Method: __Remove_NoCheck (PRIVATE)
--- a/data/pigui/modules/info-view/03-econ-trade.lua
+++ b/data/pigui/modules/info-view/03-econ-trade.lua
@@ -18,6 +18,9 @@ local gray = colors.grey
 
 local iconSize = Vector2(28, 28) * (ui.screenHeight / 1200)
 local buttonSpaceSize = iconSize
+-- FIXME: need to manually set itemSpacing to be able to properly size columns
+-- Need a style-var query system for best effect
+local itemSpacing = ui.rescaleUI(Vector2(6, 12), Vector2(1600, 900))
 
 local shipDef
 local hyperdrive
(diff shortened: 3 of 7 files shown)


WKFO:
I have no strong opinions on progress gauges, but cash should be shown. Isn't it the only place where we can see cash in-fight?

The-EG:
As a straight conversion of what's there, this is good.

But, I feel like there's a struggle to find a way to fill up the more of the screen on some of these windows (not just this one). Do they really need to be dedicated full screen displays?

As far as gauges or not go: I'd approach it like any data analysis visualization. Why am I opening this screen in the first place, and when I do what's the question(s) I'm trying to answer?

For an example:
I'm making a multiple jump trip with a full cargo hold, which means I don't have enough room to also carry enough hydrogen for the jumps. I'll need to pump fuel down after each jump to make the next one.
So, why am I opening this window? To pump fuel  down to the cargo hold to make another hyperjump.
What questions do I need to answer?
 - How much fuel do I currently have in the fuel tanks?
   Do I need to stop in a system along the way to refuel? Displaying the remaining Delta-v is interesting here, if I had a way of knowing how much I'd need to make the final leg of the trip (ie. navigating to and landing at a station) this value could be _really_ useful. As is, a percentage is definitely needed, I probably don't want to go under 50%.
 - How much fuel do I need to make the next jump?
   I actually have to flip between windows to get this currently...or I'll just fill any empty space with hydrogen and hope I didn't misjudge how much space I'd need.
 - How much fuel do I have currently in cargo? (I may need 2T, but already still have 1T left in cargo)
   Related to the above, but how much do I need to move...
 - How do I move the fuel?
   Fairly obviously, the buttons. 

Going through that, I'd argue that the 'pump down,' 'refuel,' fuel and hyperjump fuel info/gauges are all related and could actually be grouped visually. I think a grouping like that would also help make the interface more intuitive as well. If we want gauges, having the fuel gauge where the current fuel info is (above the pump down, refuel buttons) and the hyperjump fuel gauge below the buttons could give a nice visual que of what's going on.

impaktor:
These are the dynamic variables, that I thin aught to be shown together with pump up/down buttons and info on cargo loaded:
shipinfo

(Note: I've with this PR also changed font to medlarge, to make it consistent with the other screens)

Gliese852:
Thanks for the font! What if you also show the fuel use rate?

nozmajner:
I don't think fuel use rate is much useful. Knowing how much time you need to burn up your fuel isn't that important information compared to deltaV.

Gliese852:
@nozmajner deltaV depends on the load of the ship, while this value is constant for the ship, and characterizes the efficiency of the engines, if I understood it correctly. Besides, it seems to me it will add a sense of scale.

nozmajner:
For that Exhaust velocity or ISP is a more standard measure in space flight. Admittedly not that intuitive though.
How even fuel use looks for our ships?

Gliese852:
@nozmajner 
for sinonatrix:
```
   "effective_exhaust_velocity" : 19500000.0,
   "forward_thrust" : 5500000.0,
   "fuel_tank_mass" : 30,
```
since
```
FuelUseRate = forward_thrust / (fuel_tank_mass * effective_exhaust_velocity) * 0.01 
```
use rate = 9.4E5 tonn/sec, I guess -> 338.4 kg/hour?

get it from:

https://github.com/pioneerspacesim/pioneer/blob/d5e69790ce957a5aa69cceb2e86bd57e5f1af9ac/src/ship/Propulsion.cpp#L195

Gliese852:
@nozmajner deltaV does not take into account the fact that you can carry additional fuel in the hold

impaktor:
Testing using icon for jettison (will remove the text version of the button):
2020-11-22-110955_1600x900_scrot

impaktor:
OK, This is what it's look like now:


Here's Nozmajner's mockup:
8g7tah

I'll remove the WIP, as I'm not sure what more there is (other than some squash)

impaktor:
Current state:
2020-11-27-214225_1600x900_scrot

Nozmajner's mockup:
mTptHbs

Web-eWorks:
@impaktor I'm sorry, I just couldn't help myself :stuck_out_tongue:
Fixed a few minor issues I noticed - the window padding is now correct and no longer offset to the right, we rebuild the cargo list only when it changes, and the cargo meter now actually displays the percentage of cargo space and not total equipment space.

Also made the TabView debug reloading interface simpler and fixed a bug with the ImGui stack cleanup handler that wasn't cleaning up fully and was causing C++ assertions when a script errored out. Feel free to squash my commits as needed.

impaktor:
@Web-eWorks thanks! I've squashed my temporary commit, and rebased & merged to master.
Might we expect a PR that will purge master of NewUI later? :)

Web-eWorks:
Indeed you shall!

PR #4831, opened 2020-03-08
Significantly update commodity prices!

impaktor:
_Recommendation: (Please ignore the tables - they're there for the nerds, and make this PR look intimidating, there be plots for the lazy)_

# Introduction

Trading in pioneer has been, more or less broken from the start. "Broken" in the sense that the profit from commodity trading compared to the work input in relation to mission rewards and expenses (ships, equipment) is much too low.

Pioneer is still basically a Frontier clone (but with ships replaced, and galaxy not 2D flattened). Since trade worked fine in Frontier, relative missions and ship/equipment prices, which I've played a lot, I thought the first (or "zeroth") step to improve Pioneer's trade is to model it on Frontier's, to make it at least as good as their system.

# Diagnosing the problem

To first diagnose the problem in Pioneer, I wrote a little [commodity debug](https://github.com/impaktor/pioneer/tree/debug_commodity_prices) module, as I've show in #4828, which, in the screenshot below shows the heart of the current problem: in a radius of 70 ly, the maximum profit per tonne for 2000 inhabited systems scanned is $73 / t, for Precious metals, with a relative profit of 58%. The starter ships in Pioneer have ~7t available, giving a profit of ~$500 per trade, assuming they fly those 64 ly between the min and max price systems. They could just as well take a package delivery for $600 instead.

2020-03-08-083815_1600x900_scrot

A more apt comparison is for trade opportunities in a 20 ly radius (encompassing 23 inhabited systems), where we find $44 / t for Consumer goods between _Epsilon Eridani_ and _Kruger 60_ system, just 4.89 ly apart, with a relative profit of 43%, summing up to a total $300 profit per trade, (given starter ship with 7t capacity):

2020-03-08-083958_1600x900_scrot

## Trade profit compared to Frontier?

In Frontier, one of the most profitable trade routes is the Sol - Barnard's Star run, hauling robots ($868.8 - $566.9) with $300 profit per tonne, with the starter ship that, like Pioneer, has 7t free space (if selling missiles and pulse laser, which aren't needed in those two safe systems), giving $2100 in pure profit single way. In an evening's casual gameplay I made 5-10 trade runs and went from $100 to $14,000 credits.

### Relative price differences

[...]

PR #4970, opened 2020-10-10
Rebalance, the third

nozmajner:
I've did another pass on rebalancing the stats for all ships.
[In mod form](https://github.com/pioneerspacesim/pioneer/files/5428887/rebalance_2020-1023.zip)
[I've used this spreadsheet to calculate the values](https://docs.google.com/spreadsheets/d/1Ibttub5Djgais-F5-Uy2lVMv5Ye14JUpMBgnHAnb-Fo/edit?usp=sharing)
The spreadsheet has a quick way to copy eny of the jsons

I've included a simple price calculator, that prices some features like acceleration, exhaust velocity, mass, cargo space.
Skipjack is included in there too, so I might want to update that PR if this one gets pulled.
I've reduced accelerations quite a bit. The best you can pull off is 4.5g with the Wave. This might be controversial, but I think the lower accelerations in general will make handling more grounded, and combat more enjoyable. Hopefully the sense of mass will be better felt as well. I've added acceleration caps as well, so even at the emptiest, ships won't go up to absurd accelerations (You could go to about 20g before I think. Meat soup anyone?)
Travel times shouldn't differ significantly. They might be a little longer, but by half a day tops.

I've added aerodynamics stats as well, and pressure limit.
I've only explicitly entered equipment slots where they differ from the default.
Also the entries in them are rearranged to a more sane order.

Apart from how they now feel in general, some important testing is needed: 
- How prone the autopilot to crash, both for the player or the tradeships. 
- How much problem it has with navigation?
- Should mission deadlines be updated? (I've botched an urgent courier mission to Jupiter by a small margin, and I suspect I would be able to pull it off if I'd been flying manually. If that's a recurring thing, I think it is good, because now the player has to judge their and their ships abilites before accepting a mission. And for urgent missions, tight margins are good in my opinion)
- Is there any CTD happening? (I had one for the Pumpkinseed from the Eridani start before I explicitly set the default hyperdrive to class 1. Other ships might need that as well)

Issues still:
- [x] No medium fighter/courier/light freighter with a default class 2 hyperdrive - combat missions crash because of this (#4983) 
- [x] Update the default Hyperdrive of the Sinonatrix to class 2, both on market and Sol start (Class one is very slow, deliveries are excessively tough that way) **- OR -** Play with the mission deadlines
- [ ] Check all ships to have decent Hyperspace range with default drive. Mola Mola and Deneb has issues on master


impaktor:
I had a python-script somewhere that converts excel/csv list to our json ship format, I don't know if that's of interest.

nozmajner:
That would be mighty useful

impaktor:
@nozmajner sadly, I can't find it atm. Doesn't seem to be in the branches I've pushed, nor the local ones. Perhaps on my old steam-engine laptop. I'll check later this weekend.

data/ships/lodos.json
```diff
@@ -1,39 +1,54 @@
...
+		"mercenary": true,
+		"pirate": true
+	},
+	"effective_exhaust_velocity": 25000000,
+	"thruster_fuel_use": -1.0,
+	"fuel_tank_mass": 1250,
+	"hyperdrive_class": 0,
+
+	"forward_thrust": 39000000,
+	"forward_acceleration_cap": 13.734,
```
WKFO, line 30: Uhm... @nozmajner , why is Lodos' forwards acceleration cap lower than it's reverse acceleration?

nozmajner:
No probs at all. The sheet works nicely, and I can live with the tiny inconveniences it still has.

nozmajner:
@WKFO I must have missed the Lodos accel cap discrepancy. Will fix.

robothauler:
At the moment this will break some combat missions because no ships for the role "merchant" are available.

nozmajner:
There's a merchant role? Good to know. I'll add some then. Thanks for the heads-up!

robothauler:
Just for the record. Some missions spawn enemy ships only with hyperdriveClass > 0. So, these missions should be adapted afterwards.

Web-eWorks:
I'm noticing all files in this PR are missing a trailing end-of-line. Before we merge this, I'd like to ensure all of the relevant JSON files terminate in a line feed according to what Github likes.

nozmajner:
What is exactly a trailing end-of-line? A new line at the very end? If so, I updated to spreadsheet to add that.
Also added the merchant role.
I'll update the default hyperdrives too, and the mismatched retro and up accelerations tomorrow.

nozmajner:
I've updated the acceleration limits, set up the default hyperdrives, and added the merchant role to the ships that would need it.
I've also streamlined the spreadsheet's exporting part, to fix that little lag that happened when a value was updated, and the exporter expressions redid their thing. Now there's only one exporter cell, and a ship picker for it.

ConfedMajor:
I don't notice any diference
Is it supposed to see a diference in ships? @nozmajner

nozmajner:
@ConfedMajor The stats of ships. Accelerations most importantly, handling.

ConfedMajor:
> The stats of ships. Accelerations most importantly, handling.

@nozmajner Oh, i never noticed something like that, i always use the autopilot.

nozmajner:
@ConfedMajor The autopilot should feel it as well. Not that much likely.

impaktor:
OK, my interest here is mainly the prices of ships, and their relation to #4831, i.e. viability of trade. (And I haven't looked at the prices for the ships in this PR). Since we scaled the trade to be similar to Frontier (because that's what I played, so I know it worked fine), here are some prices in frontier:

- light fighter ships, ~10t free after drive fitted, cost 30k-50k credits
- viper & cobra, have 15t and 25t free after drive fitted, cost 80k-90k credits
- ships with 45t-55t cost about +100k - 150k
- trusty Asp Explorer 75t cost ~190k
- ships with +100t cost more than 200k

Ah, fakk it, I have all the data for the Frontier ships under my fingers, so let's just do a plot (and make @laarmen happy, although I did these in under 5 minutes, so not as fancy as they used to be):

lin

OK, as I suspected, this needs log-log scaling to be approximately straight:

log
 
So, as I've argued a few years ago in a re-balance thread on the forum somewhere:  
- we want a roughly equidistant spacing of ships in log-log to "cover all user cases"
- price and cargo space is roughly linear in log-log (if you squint your eyes and drink some whisky)

So I'm not arguing we should have the exact same prices as in Frontier, but they should be in a ballpark similar-ish, or cheaper because damn kids to day don't have patience to grind (:older_man: (emoticon for old man, apparently). So not off by a very large factor. Yeah, I know, I'm arguing from "what works from player perspective", rather than some complex spreadsheet summing of all the ships variables.

We also see in the plot above some ships are relatively "strangely" priced relative the others, (not monotonically increasing curve), so it's not strange to have some strange prices, ...said the strange man, strangely. 

The more correct way to do this (i.e. compare ship prices with trade prices) would be to jot down free space for each ship after they're fitted with "standard" outfit for a trader (radar, atmo-shield, engine, etc.) and then look at profit for each trade run in each ship, on some profitable trade route given the new commodity prices. 

And as it happens, I have that data for each ship in Frontier for the Sol-Barnard route, (which is very nice way to get going when starting the game), but I'll leave that alone for now.

nozmajner:
Here's how that looks for our ships:

image

impaktor:
The due dates on the missions will have to be adjusted since this PR will slow down the ships. 

We probably need some stats on how long a journey now takes in some "reference" ship in the game, and multiply that by some factor.

I found this plot I made +5 years ago for the dev forum (I think something similar should be done now)

due_date

nozmajner:
Prices:
image

nozmajner:
Updated the PR with the updated stats from the spreadsheet. Including the now merged Skipjack.
I've also updated the cross-section numbers of the Sinonatrix to reflect the overhauled model.

nozmajner:
Updated this with the prices in the sheet, and rebased. Likely I screwed that up a bit, it seems that it wants to delete and add every line in the files, but on the other hand most files were all over the place, so it would happen to them anyway.
Price progression looks like this:
image

I've as discussed with @Gliese852, I've also added the "courier" role to these ships:
- Amphiesma
- Bluenose
- Bowfin
- Kanara civ
- Lunar Shuttle
- Mola Mola
- Natrix
- Pumpkinseed
- Sinonatrix
- Skipjack
- Varada
- Wave
- Xylophis
- As I understand, this role will be used to pick ships that will go around the system locally.

nozmajner:
Sorry for the many commits. The last one is a change suggested by @Gliese852, so the roles are set up as an array.

Web-eWorks:
I'm not going to ask you to redo all of the commits, but none of the ship json files have a line break at the end of the file; it's normally considered good practice to end your files with a new line.

Otherwise, this is looking pretty good; it'll likely get squash-merged to remove the micro-fix commits, but pending a short playtest it gets my thumbs-up!

Web-eWorks:
To amend my prior comment: it doesn't get my thumbs-up yet because the Sinonatrix has a 66t fuel tank for some reason.

A) this makes it really hard to maintain as a starter ship, because a tank costs about 4 in-system delivery missions to refill, and
B) this gives it a fully-laden deltaV of 11,000km/s, far greater than other comparable ships (the highest ship in the courier class has a dV of only 7,500km/s fully laden) or even any ship in the game. A quick test with the spreadsheet suggests a fuel capacity in the 35-40t range is far more appropriate for the ship, yielding a dV in the 7,000-8,500km/s range and better acceleration due to the less extreme weight of the hull.

(I made and reverted some tweaks to your spreadsheet while testing numbers @nozmajner because I didn't realize I still had edit permissions there, sorry!)

nozmajner:
Updated the stats witht the changes @Web-eWorks suggested. Also added an empty line at the end of all files. (Ignore stupid commit message... :D ).
I'm not sure if I'm up to rebasing the branch though, my fork is a bit of a mess already.

Web-eWorks:
Played a bit with the new ships, liking where things are at now. Pumpkinseed feels properly speedy compared to the sinonatrix, and the skipjack's handling has improved a bit. The xylophis seems a little slow for the massive engine on the back, but most likely that's something that will be fixed with the new xylo model later on.

PR #4973, opened 2020-10-14
Fix Flight Log crashes

### LINKED ISSUE #4917: Flight Log crashes when an entry is created while orbiting a system barycenter
### Observed behaviour
If you try to make a custom log entry while orbiting a system barycenter instead of an object, it crashes.

FlightLog.lua:L355 
attempt to index field 'frameBody' (a nil value)

### Expected behaviour
Accept system barycenter (for example Sirius AB) as a valid location

### Steps to reproduce
1) Jump to Sirius.
2) Make sure your frame is the system barycenter, not a body.
3) Try adding an entry to the Custom Log tab.

1827942a4879c1bdf3fd7044395dd2fe30ccccab (Windows 10 x64)

WKFO:
Flight Log no longer crashes the game if you try to create a custom entry while in hyperspace or in orbit around a system barycenter. Fixes #4917 

screenshot1

### DIFF
--- a/data/libs/FlightLog.lua
+++ b/data/libs/FlightLog.lua
@@ -344,30 +344,39 @@ FlightLog = {
 		text = text or ""
 		local location = ""
 		local state = Game.player:GetFlightState()
+		local path = ""
 
 		if state == "DOCKED" then
 			local station = Game.player:GetDockedWith()
 			local parent_body = station.path:GetSystemBody().parent.name
 			location = {station.type, station.label, parent_body}
+			path = Game.system.path
 		elseif state == "DOCKING" or state == "UNDOCKING" then
 			location = {state, Game.player:FindNearestTo("SPACESTATION").label}
+			path = Game.system.path
 		elseif state == "FLYING" then
-			location = {state, Game.player.frameBody.label}
+			if Game.player.frameBody then
+				location = {state, Game.player.frameBody.label}
+			else
+				location = {state, Game.system.name} -- if orbiting a system barycenter, there will be no frame object
+			end
+			path = Game.system.path
 		elseif state == "LANDED" then
+			path = Game.system.path
 			local alt, vspd, lat, long = Game.player:GetGPS()
 			if not (lat and long) then
 				lat, long = "nil", "nil"
 			end
 			location = {state, Game.player:FindNearestTo("PLANET").label, lat, long}
 		elseif state == "JUMPING" or state == "HYPERSPACE" then
-			-- bug: if in hyperspace, there's no Game.system (!)
-			-- local spath, sysname = Game.player:GetHyperspaceDestination()
-			sysname = ""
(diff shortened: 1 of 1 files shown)


WKFO:
Ugh, I seemed to have introduced another bug. Whenever you load a save game, log entries are gone.

data/libs/FlightLog.lua
```diff
@@ -344,30 +344,39 @@ FlightLog = {
...
 			end
 			location = {state, Game.player:FindNearestTo("PLANET").label, lat, long}
 		elseif state == "JUMPING" or state == "HYPERSPACE" then
-			-- bug: if in hyperspace, there's no Game.system (!)
-			-- local spath, sysname = Game.player:GetHyperspaceDestination()
-			sysname = ""
+			--if in hyperspace, there's no Game.system
+			local spath, sysname = Game.player:GetHyperspaceDestination()
+			path = spath
 			location = {state, sysname}
```
impaktor, line 375: Is the `spath` not the same type as the `Game.system.path` or why is the version bump needed?

(I would expect the bump to be needed if unpacking the table is different, but then I'd expect to see edits in the pigui module, that makes use of it, as well).

WKFO: I thought the same way (and I still think the same way), but I couldn't find any explanation for the Flight Log getting erased. I read the whole file over and over again but couldn't find what was wrong, and only bumping up the version number solved it. It still bugs me.

impaktor: I see you pushed again, but not what the change was (just a rebase?). Regarding `location`, looks like it should be a tuple (if indeed that be its lua name), of two strings, so if `sysname` is still a string, that should be fine, I suspect.

I'd merge this in a second if I could understand why the flight log save bump is needed (me not understanding it hints something is wrong in either my brain or the code (although one does not exclude the other)). Granted, I haven't banged my head against a brick wall to understand that particular thingy, yet.

WKFO:
@impaktor I made some more test runs to figure out why the version number bump was needed, and I (un-?)fortunately couldn't retrace the save/load problem. It **DID happen** a few times again, but I found no correlation whatsoever with the version number (or anything at all, really). Reversing the version number to 1 now, as it seems to work like expected.

I would like others to test this PR, using a fresh save file, in various flight situations; and if nothing goes wrong, no problem - it means that the bug was my personal issue. If something does go wrong, however; just increase the version number and try again please.

data/libs/FlightLog.lua
WKFO: Read my last PR comment, resolved.

impaktor:
@WKFO Since the flightlog is not in the last release at all, a save/crash would only be for those:
1. compiling from recent master
2. and saving while in hyperspace(?)

either way, a minor thingy.

PR #4974, opened 2020-10-15
Use an offscreen buffer as the primary render target.

Web-eWorks:
Refactor the renderer to use an offscreen buffer as the primary render target instead of the window surface. Among other benefits, this means that we don't need the windowing system to give us a 24-bit depth MSAA window surface; as long as the GPU is able to render to those textures we'll be fine.

This is required for later postprocessing work; we need to resolve the MSAA before we do image post processing (we'll probably go the tonemap-resolve-inverse-tonemap route as that seems to be pretty standard).

It also lays the groundwork for 32-bit reverse-Z depth buffer (in fact, the clever among you will notice that the renderer is looking for `ARB_clip_control` and creating a `DEPTH_COMPONENT_32F` depth buffer!) This should allow us to drop the logarithmic depth hack, speeding up our shaders somewhat and allowing them to be eligible for the early-Z test again.

Tangentially, this makes resizing windows and changing graphics settings without needing to restart the game much more possible; while changing resolution is going to need more work on the UI layout side of things, changing MSAA is totally technically feasible, just not... implemented, yet.

### DIFF
--- a/src/JobQueue.h
+++ b/src/JobQueue.h
@@ -263,6 +263,7 @@ public:
 	virtual void Order(Job *job)
 	{
 		auto x = m_jobs.insert(m_queue->Queue(job, this));
+		(void)x; // suppress unused variable warning
 		assert(x.second);
 	}
 	virtual void RemoveJob(Job::Handle *handle) { m_jobs.erase(*handle); }
--- a/src/graphics/RenderTarget.h
+++ b/src/graphics/RenderTarget.h
@@ -18,12 +18,13 @@ namespace Graphics {
 	// Specifying a depth format with no allowDepthTexture will create a depth buffer
 	// fixed to this rendertarget
 	struct RenderTargetDesc {
-		RenderTargetDesc(Uint16 _width, Uint16 _height, TextureFormat _colorFormat, TextureFormat _depthFormat, bool _allowDepthTexture) :
+		RenderTargetDesc(Uint16 _width, Uint16 _height, TextureFormat _colorFormat, TextureFormat _depthFormat, bool _allowDepthTexture = false, Uint16 _samples = 0) :
 			width(_width),
 			height(_height),
 			colorFormat(_colorFormat),
 			depthFormat(_depthFormat),
-			allowDepthTexture(_allowDepthTexture)
+			allowDepthTexture(_allowDepthTexture),
+			numSamples(_samples)
 		{}
 
 		const Uint16 width;
@@ -31,6 +32,7 @@ namespace Graphics {
 		const TextureFormat colorFormat;
 		const TextureFormat depthFormat;
 		const bool allowDepthTexture;
+		const Uint16 numSamples;
 	};
 
 	class RenderTarget {
--- a/src/graphics/opengl/GLDebug.h
+++ b/src/graphics/opengl/GLDebug.h
@@ -86,6 +86,9 @@ namespace Graphics {
 			GLuint id, GLenum severity, GLsizei length,
 			const GLchar *message, const void *userParam)
 		{
+			// filter out Type=Other informational messages
+			if (type > GL_DEBUG_TYPE_PERFORMANCE)
+				return;
 			Output("Type: %s, Source: %s, ID: %u, Severity: %s, Message: %s\n",
 				type_to_string(type), source_to_string(source), id,
 				severity_to_string(severity), message);
--- a/src/graphics/opengl/RenderTargetGL.cpp
+++ b/src/graphics/opengl/RenderTargetGL.cpp
@@ -7,34 +7,10 @@
 namespace Graphics {
 	namespace OGL {
 
-		RenderBuffer::RenderBuffer()
-		{
-			glGenRenderbuffers(1, &buffer);
-		}
-
-		RenderBuffer::~RenderBuffer()
-		{
-			glDeleteRenderbuffers(1, &buffer);
-		}
-
-		void RenderBuffer::Bind()
-		{
-			glBindRenderbuffer(GL_RENDERBUFFER, buffer);
-		}
-
-		void RenderBuffer::Unbind()
-		{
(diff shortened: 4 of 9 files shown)


Web-eWorks:
NOTE: this particular change bumps the requested OpenGL version from 3.1 to 3.2, primarily for multisampled textures but also because 3.2 seems to be the "stable child" of the 3.x series. I'll consider bumping it to 3.3 in the future if we need features from GLSL 330, but leaving it here until we switch to Vulkan seems like a decent option.

fluffyfreak:
Comments that come first to mind:

1. Reverse Z isn't quite as good as 32-bit logarithmic depth but it would probably do in a lot of cases
2. Reverse Z might not work at all under [OpenGL](https://outerra.blogspot.com/2012/11/maximizing-depth-buffer-range-and.html) unless using nVidia specific extension `glDepthRangedNV`, this has been added to AMD drivers but what about Intel?Also needs checking on OSX (it won't, ever be exposed there) and Linux.
3. bumping us to 3.2 is probably a breaking change for older OSX whcih is what limits us to GL 3.1 otherwise we'd be on 4.2 by now ;)

Will take a look at the actual code asap

Web-eWorks:
> Comments that come first to mind:
> 
>     1. Reverse Z isn't quite as good as 32-bit logarithmic depth but it would probably do in a lot of cases

In my mind, early-Z discard is more important than logarithmic depth buffer (especially as we move to PBR where each fragment will consume a larger amount of rendering time); an infinite-far-plane 32-bit reverse-Z depth buffer is (according to Kemen) better than the 24bit logarithmic buffer we have now, and means that we get to use some optimizations to keep our rendering time down.

>     2. Reverse Z might not work at all under [OpenGL](https://outerra.blogspot.com/2012/11/maximizing-depth-buffer-range-and.html) unless using nVidia specific extension `glDepthRangedNV`, this has been added to AMD drivers but what about Intel? Also needs checking on OSX (_it won't, ever be exposed there_) and Linux.

https://nlguillemot.wordpress.com/2016/12/07/reversed-z-in-opengl/
https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_clip_control.txt

ARB_clip_control is one of those extensions I expect to be ubiquitous; the hardware has supported it ever since graphics cards were made for DirectX and OpenGL at the same time, but the standards committee in their infinite wisdom dragged their heels for a long time in actually making it an extension. Intel graphics support it ever since Mesa 10.6, Radeon supports it ever since the HD6xxx days according to the compatibility charts I've found.

With the complete deprecation of OpenGL on Mac, I'm not worrying about that side of affairs; we'll need to move to Vulkan through MoltenVK if we want Pioneer to even run on recent Macs.

>     3. bumping us to 3.2 is probably a breaking change for older OSX which is what limits us to GL 3.1 otherwise we'd be on 4.2 by now ;)

Actually not; Macs support specifically GL3.2 and 4.5 IIRC (can't find the link). Actually tracked down some weird misinformation; the forward-compatible flag that we set is completely unneeded for OSX and is totally unused by SDL in the OSX backend, it's a holdover from the internals of GLFW (which the library itself will refuse to give you a context unless you explicitly add the flag).

> 
> Will take a look at the actual code asap

Thanks!

fluffyfreak:
> > ```
> > 3. bumping us to 3.2 is probably a breaking change for older OSX which is what limits us to GL 3.1 otherwise we'd be on 4.2 by now ;)
> > ```
> 
> Actually not; Macs support specifically GL3.2 and 4.5 IIRC (can't find the link). Actually tracked down some weird misinformation; the forward-compatible flag that we set is completely unneeded for OSX and is totally unused by SDL in the OSX backend, it's a holdover from the internals of GLFW (which the library itself will refuse to give you a context unless you explicitly add the flag).
> 
That's what I thought when I implemented it, the setting of it to 3.1 isn't misinformation, it was very painfully won information.
I originally bumped us directly to GL 3.3 and 330, it was OSX that dragged us all the way back down to 3.1, whatever the OSX docs say, they are full of shit :(

That's with SDL and not GLFW btw, on our codebase. I'm willing to get the Mac running again (_I haven't bothered in so long_) to test changes. Hell I'm even willing to drop OSX support entirely!

If no-one from that community is willing to step up and help then why should it hold us back _constantly_?

Web-eWorks:
> That's what I thought when I implemented it, the setting of it to 3.1 isn't misinformation, it was very painfully won information.
> I originally bumped us directly to GL 3.3 and 330, it was OSX that dragged us all the way back down to 3.1, whatever the OSX docs say, they are full of shit :(

It was probably true with older SDL versions, but grep the SDL2.0.8 source code and you'll see those flags are used pretty much nowhere (and absolutely nowhere in the cocoa backend). Mac's docs make very explicit reference to 3.2 in several spots, but I'm sure it wasn't nearly as well supported when Pioneer was first ported to GL3. :(

> That's with SDL and not GLFW btw, on our codebase. I'm willing to get the Mac running again (_I haven't bothered in so long_) to test changes. Hell I'm even willing to drop OSX support entirely!
> 
> If no-one from that community is willing to step up and help then why should it hold us back _constantly_?

Yeah, I think we have maybe one person who intermittently uses Mac; for now I'd say let's proceed under the assumption that GL-on-Mac is broken (by Apple no less!) until we switch to Vulkan. Then MoltenVK can pick up the slack with OSX compatibility and we have a unified standard to write against.

fluffyfreak (approved):
As usual code looks good and clean, can't see any issues with modern systems and it cleans out a lot of cruft 👍
Some of that cruft was for edge cases with older drivers/GPUs but I think we can move on.

Web-eWorks:
Thank you! Will hold on pulling the merge trigger until Reverse-Z is ready to PR, since I'll be re-using the branch.

impaktor:
Thanks for reviewing @fluffyfreak  and thanks @Web-eWorks for doing the work. Very nice to see!

PR #4975, opened 2020-10-16
Use Reverse-Z depth buffer, drop logZ hack

Web-eWorks:
Implement a reverse-Z depth buffer and update all related code to deal with the change in clip-space coordinates from -1..1 to 1..0.

This is the follow-on PR to what was discussed in #4974; we remove `logz.glsl` entirely and stop setting the fragment depth in the fragment shader, which allows the GPU to further optimize fragment shader execution. It should also provide a *slight* precision boost to the depth texture, which translates to a negligible impact.

Among other benefits, this opens the doors to using an "infinite far-plane" projection matrix, which might remove the need to rescale planets during rendering, although I'm not holding my breath on that one.

STURNCLAW TODO:
- [x] Fix PerspectiveFrustum being a 0..1 matrix instead of 1..0
- [x] Fix copyright symbol weirdness in shader files


Web-eWorks:
Since this is all terribly dry technical stuff, have some funny screenshots from while I was updating all of the code to deal with the reversed depth representation...

Screenshot from 2020-10-15 23-21-31
Screenshot from 2020-10-15 22-34-41

Web-eWorks:
CC @Gliese852 since I'm touching your code; I've replaced `Gui::Screen::Project` with a better function hanging off of the Graphics namespace; this means getting a screen-space position of a world object doesn't require `PushToOrtho` anymore and I've collapsed `AddProjected` and `AddNotProjected` as the distinction is now irrelevant.

data/shaders/opengl/geosphere_star.frag
```diff
@@ -1,8 +1,7 @@
-// Copyright � 2008-2020 Pioneer Developers. See AUTHORS.txt for details
+// Copyright � 2008-2020 Pioneer Developers. See AUTHORS.txt for details
```
fluffyfreak, line 1: what going on with the © symbol?

src/graphics/opengl/RendererGL.cpp
```diff
@@ -567,7 +570,8 @@ namespace Graphics {
 
 	bool RendererOGL::SetDepthRange(double znear, double zfar)
 	{
-		glDepthRange(znear, zfar);
+		// XXX since we're using reverse-Z, flip the inputs to this function to avoid breaking old code.
+		glDepthRange(1.0 - zfar, 1.0 - znear);
```
fluffyfreak, line 574: ooh, I'm torn about this decision. Do we just want to make it explicit, possibly by renaming the method and parameters? Tricky. Is it commonly used by non-graphics code? Because if it's just people who are more familiar with the graphics stuff, and know that it's using reverse-Z, then they might double reverse it?

fluffyfreak:
Stop looking up New Hopes skirts! It's rude! :D

Will review and test asap, maybe at lunchtime!

data/shaders/opengl/geosphere_star.frag
Web-eWorks: I honestly have no idea. I didn't touch that line at all. The only thing I can think of is that the file was originally written in utf-16 or something?

fluffyfreak:
I had a few issues getting this to build cleanly, but it worked fine once I'd resolved them.

The Clang format failure is just wrong, it should ignore enum's, or in this case, we should just ignore it.

fluffyfreak (approved):
Code looks good (_as always_) and it functions perfectly as far as I can tell. My testing wasn't exhaustive so we'll need to keep an eye out for strangeness but I think it's good to merge.

fluffyfreak:
@impaktor @nozmajner it'd be good to get more eyes on a change like this since bugs might be subtle and could appear in any rendering subsystem. If anyone else gets a chance to check it by looking at menus, hyperspace, etc etc it'd help to catch any problems.

Web-eWorks:
@fluffyfreak the reason `pigui/View.cpp` was named such is I'm trying to get rid of the actual `View.cpp` as soon as possible and replace it with the PiGui implementation. I suppose it works to rename the file if VS really can't deal with it, but that solution was working for @The-EG, not sure why it wasn't working on your end.

The-EG:
> I suppose it works to rename the file if VS really can't deal with it, but that solution was working for @The-EG, not sure why it wasn't working on your end.

I've been using the CMake build within VS lately, so if the solution broke again I wouldn't have noticed.
The ship/View.cpp and pigui/View.cpp clash should have already been resolved, I manually set on to have a different object file (hackish, but I didn't see another option without renaming the files): see #4964 .

Looking at my PR, it looks like I missed profile and pre-release configurations though, so those may not have been working

fluffyfreak:
That the solution worked wasn't the issue "View" is not a good name for either of the classes, and the filenames should match the class name wherever possible.

I should have caught it when it first happened and just renamed the class+file at the time. That change just makes it a bit better.

nozmajner:
I started testing this. No visual glitch yet, and Jupiter isn't disappearing either. 
I had a crash, but I don't know the reason. The windows wasn't active, since I was writing on IRC when it happened. There was nothing in the output.txt.  I was heading to Marsh High on autopilot. It did not crash the next time I went there. I'll be on the lookout though.

ConfedMajor:
> Here we see the rare "fairy lights" in their natural habitat: shining through heretofore solid walls in blatant violation of the laws of physics... Truly magical!

Are thoose lights floating UIs or something like that?

Web-eWorks:
> Are thoose lights floating UIs or something like that?

They are billboards with emissive textures on them; they don't actually cast any light, but they model the flare from a lightbulb (semi)-decently. In this case, they were still using the old depth method while the station itself was (mostly) using the new depth method, so the lights that should have been hidden behind the geometry were being drawn overtop. :smile:

ConfedMajor:
> They are billboards with emissive textures on them; they don't actually cast any light, but they model the flare from a lightbulb > (semi)-decently. In this case, they were still using the old depth method while the station itself was (mostly) using the new depth method, so the lights that should have been hidden behind the geometry were being drawn overtop. 😄

_Maybe this is useless,_ but in roblox there is a property in SurfaceGUIs that is called AlwaysOnTop:

Enabled:
RobloxScreenShot20201019_223548930

Disabled:
RobloxScreenShot20201019_223543124

@Web-eWorks Maybe you should check if your GUIs has a high ZOffset (in roblox) or if they are on top everytime?

Web-eWorks:
Pioneer's renderer works very differently than Roblox, but thank you anyways for the suggestion!

fluffyfreak:
@Web-eWorks is there technically anything left to do with this PR?

Web-eWorks:
Remove or refactor `matrix4x4::PerspectiveFrustum()`, fix the copyright symbols. That's it!

Web-eWorks:
This PR is good for merge; I'll wait until @fluffyfreak wants to pull the trigger but I've fixed all the remaining bugs I found!

data/pigui/modules/flight-ui/indicators.lua
```diff
@@ -32,7 +32,7 @@ local function displayDirectionalMarkers()
 	local function displayDirectionalMarker(ship_space, icon, showDirection, angle)
 		local screen = Engine.ShipSpaceToScreenSpace(ship_space)
 		local coord = Vector2(screen.x, screen.y)
-		if screen.z <= 1 then
+		if screen.z <= 0 then
```
fluffyfreak, line 35: Should this be `if screen.z <= 0 or screen.z >= 1 then` instead because the valid visible range is 0..1? Or does Lua see the z differently?

fluffyfreak: I may have read that wrong, it just seems strange that it's only visible if it's less than one when the range is 0..1

Web-eWorks: To preserve existing behavior, all of the "project to screen" routines negate the Z coordinate they return; coordinates in the visible Z range are negative, while coordinates behind the camera are positive. It's weird and I'd welcome some improvement in that area, as trying to massage the depth coordinate hasn't yielded good results so far.

src/graphics/opengl/RendererGL.cpp
Web-eWorks: I'd like to totally remove the DepthRange call (or at least make it part of a pipeline state object) as it's a bit abused by the SectorMapView at the moment.

PR #4978, opened 2020-10-21
Input Frame 2.0

### LINKED ISSUE #4915: Configuring controller "View Zoom" does not expand
### Observed behaviour
The very bottom option in configuring controller, "View Zoom" can not be expanded to be altered

### Steps to reproduce

Windows, Git version July 14th, 2020

Web-eWorks:
This is the culmination of about a year's worth of thought and about three months of work spread over six months. This PR completely replaces the existing method of representing input bindings, refactors the Input class into a namespace and the manager class, and introduces a whole host of TODOs, among other nice things.

Fixes #4915.

All code support for multiple-key chord bindings (e.g. CTRL+Z+A or GamepadA + Space + Joy1 Hat3 Up) is implemented, although GUI support for *binding* those chords is currently not-quite-functional. Support for joystick sensitivity and deadzone has been temporarily removed, with the intention to add a separate UI interface for managing those attributes on a per-physical-axis basis.

The interface for actually getting input events to input bindings to trigger signals has been made entirely internal to Input::Manager, finally completing the original work (from 2018) on refactoring the input system. This (in conjunction with the new `REGISTER_INPUT_BINDING` macro to simplify binding registration) significantly reduces the complexity required to add and use input bindings, keeping the process almost entirely contained in the class that is interested in the bindings.

This PR also features first-class support for input 'Frames', including modal frames that block input events from propagating further (e.g. for the Lua console) allowing key bindings to be grouped into logical modes (e.g. scanner mode, flight mode, camera mode) that can be disabled or enabled en-bloc.

While input frames (and receiving binding events in Lua) haven't been exposed to Lua, bindings now have first-class Lua support rather than needing to do string mangling; this has simplified the binding code by a great deal.

I've removed a few unused language strings from LangStrings.inc.h, moved more translation strings into the correct resource to allow `settings-window.lua` to only need the `input-core` resource, and finally, *finally* removed KeyBindings.inc.h!

As a side note, I've also added CMake support to make the build chain use LLVM's LLD linker; it speeds up single-file incremental recompile times by ~20% when enabled. Very much an experimental option for power users, and it only works on linux for right now (CC @The-EG...!)

I'd appreciate *lots* of testing on this one as it's been in the works for a very long time! I'll list known issues here that I intend to fix in a follow-on PR; I'd rather get this in the hands of @Gliese852 and others who are working on gameplay and UX code than leave it to be 100% perfect and slow down the works.

- The console has to be closed with the ESC key and won't re-open until the second press of the toggle keybinding. FIX: rewrite the console to use PiGui instead of newUI.
- Binding key chords in-game currently doesn't work; only the first key in the chord can be defined. FIX: refactor the GUI code to enter 'binding mode' and build a chord based on key press events. This will require exposing a few more signals and defining a stateful API for `settings-window.lua` to consume.

I'll leave this PR open for a week or two before squashing like changes together; if anyone would like to pitch in to fix some of the known bugs or simply ask questions, feel free!


Gliese852:
@Web-eWorks if I'm not mistaken, changes made to non-english json will be overwritten by transifex

The-EG:
MSVC doesn't have the `typeof` macro. It does have `decltype` so I've put a define into libs.h so it will build on Windows now.

The-EG:
> As a side note, I've also added CMake support to make the build chain use LLVM's LLD linker; it speeds up single-file incremental recompile times by ~20% when enabled. Very much an experimental option for power users, and it only works on linux for right now (CC @The-EG...!)

I'm not sure about the linker specifically, but it should be possible to use the LLVM toolchain on the Windows CMake build too. I was playing around with it a bit for another project out of curiousity...at the moment it looks like there's a slash (/) versus dash (-) issues with compiler flags that need to be cleaned up but that's probably it.

Web-eWorks:
@Gliese852 changes to non-english version will be overwriiten, but according to transifex' docs, this should push *new* translations that transifex doesn't know about. I'd rather have a 50% chance of needing to re-translate the strings than have a 100% chance.

@The-EG AFAIK I haven't used typeof at all in the code? It's a GCC-specific extension keyword, so if you found it I'd love to get rid of it and replace it with `decltype`.

src/Input.cpp
```diff
@@ -137,13 +278,225 @@ KeyBindings::AxisBinding *Input::AddAxisBinding(std::string id, BindingGroup *gr
...
+};
+
+void Manager::RebuildInputFrames()
+{
+	// Reset the list of active chords.
+	m_chords.clear();
+	m_activeActions.clear();
+	m_activeAxes.clear();
+
+	for (const auto *frame : reverse_iter<typeof(m_inputFrames)>(m_inputFrames)) {
```
The-EG, line 446: > @The-EG AFAIK I haven't used typeof at all in the code? It's a GCC-specific extension keyword, so if you found it I'd love to get rid of it and replace it with `decltype`.

Here's the one that was causing MSVC to fail.

The-EG:
I'll try to do some more testing (I did get in and try my PS4 controller, which seemed to work) and I need to verify that VS2019 solution is still OK, but the MSVC CMake build is good now (there was only one typeof, now fixed).

The only weird thing I've noticed...my PS4 controller has 2 triggers, which are presented both as axes and buttons. Pioneer picks up the axes fine, but won't let me bind to them as a button. Unfortunately, this could be something with how SDL is getting the events from (I assume) DInput or somewhere in between...but I don't know how to start debugging that.

Web-eWorks:
I'll add some slightly better logging to the joystick code to assist in debugging that. Also while I'm at it I need to consolidate the reverse_iter implementations... yet more work!

impaktor:
I get merge conflict (br.json and systemview.h). I've go too low energy to resolve it now.

Web-eWorks:
Here's something to shake you out of your post-Sunday stupor... this PR is finally a merge candidate!                                                                                                   

I've added support for half-axis mode, added custom sensitivity settings, and restored the deadzone functionality on a per-axis basis - none of this is exposed to the UI sadly, but it is available in the configuration file for anyone to use. (I'll leave the onus of documentation to my future self or maybe impaktor...)

My PS4 controller is binding buttons and axes just fine, so if this doesn't fix your issue @The-EG, I'm thinking it's an SDL-on-Windows bug which I can't do anything about from here.

I'd appreciate a final round of testing (please test, then delete your `config.ini` file and test again) before I merge this in; I'm aiming for Friday at the very latest depending on how busy my schedule gets.

PR #4982, opened 2020-10-23
Update MSVC AppVeyor build and Installer generation

The-EG:
Update the AppVeyor environment to Visual Studio 2019
Change the cmake generator to Ninja (this is what VS now uses by default)
Un-hardcode some paths in pioneer.iss - manual creation of installer now much easier
Create the installer with AppVeyor.
Change output filename of installer to pioneer-(release)-win.exe

So, I had no idea what I was doing when I started this..._but_ if anything, the installer is now much easier to create manually and the output filename is correct. Hopefully the appveyor stuff works too, although I imagine it will need some tweaks.

Still be done: deployment of the artifact to the releases page. This stuff needs to be added to appveyor.yml: https://www.appveyor.com/docs/deployment/github/


The-EG:
Hey, it works!
Ok, it's ugly...I may try to see if I can change the batch files to PowerShell or something. Currently the batch files don't stop on non-terminating errors, so if the build fails, appveyor happily tries to continue on with building the installer...

I switched to batch files because that allows me to properly setup the VS 2019 x64 build environment and use the Ninja CMake generator. I needed to do _that_ so I can first build pioneer, which also copies the thirdparty libraries over, copy those libraries into the build directory and _then_ run the install target.

The libraries have to be copied into the build directory or CMake fixup_bundle can't find them. If I try to do all of that with the 'Visual Studio 2019' generator it restarts the entire build instead of just doing the install portion...

Currently, this builds the installer on every build (so every PR, etc.) Currently, the installer is 303MB, and I remember that we had issues with appveyor running over quota before the 6 month retention policy set it. 
I'm going to disable pushing the installer up as an artifact on the build script and see if I can figure out how to push it up separately when a tag is pushed (or a release is created, etc.).

The-EG:
Some more improvements:
- The build should now exit on an error
- The prepare script is loading in the languages that aren't installed by default
- The installer _should_ now be pushed up to an artifact if the build is triggered by a tag push. I don't have a good way to test this.

The final thing would be getting the artifact to then be deployed to the releases page. But, I'm not familiar enough (yet?) with the GitHub release stuff to know what exactly needs to be done there.

The-EG:
This should be at a point of being useful and mergeable.

At this point, the AppVeyor:
- builds using MSVC (64bit)
- runs modelcompiler -b inplace (the 'install' only uses the compiled models)
- builds the Windows installer
- IF the build was triggered by a tag, it will push the installer up as an artifact

In addition, there is a new build target for Windows: win-installer. 
This can be used to manually generate a Windows installer:

1. Open a MSVC Native Tools Command Prompt
2. CD into Pioneer's source directory
3. `mkdir build`
4. `cd build`
5. `cmake -G Ninja -DCMAKE_INSTALL_PREFIX=install -DPIONEER_DATA_DIR=data -DCMAKE_BUILD_TYPE:STRING=Release ..`
6. `ninja`
7. `cd ..`
8. `modelcompiler.exe -b inplace`
9. `cd build`
10. `ninja install`
11. `ninja win-installer`

Web-eWorks (approved):
Looking very good, thank you @The-EG! Just waiting on configuration to publish artifacts to tagged builds before merge!

PR #4984, opened 2020-10-26
Improve tradeships

Gliese852:
https://user-images.githubusercontent.com/18342621/110258388-6d9fe100-7fb3-11eb-8e02-e8e1c5224d74.mp4

## TODO

- [x] create a pre-calculation of the ship's path according to the specified parameters

- [x] create a function that spawns ships "on the way" with checking for possible collisions at the insertion point

- [x] improve the method of creating merchant ships, using these functions `(TradeShips.lua)`

### Add class for pre-calculating spaceship path
A one-dimensional case is considered, given the parameters of the ship
and the length of the path, a flight plan is calculated, which allows
determining the time, speed, and mass of the ship at any point on the
way. Also added a simple simulator for testing purposes.

### Add function PutShipOnRoute

Function description:

The ship rearranged from its current position to a given body in space, for a given part of the path, as if it were flying in a straight line, consuming fuel and changing speed. The ship's current speed is ignored and is considered to be equal to target's.
This function changes the coordinates, the mass of fuel in the tank, and the speed of the given ship.

This function will be used so that when the game starts, or the player jump into the system, the NPC ships looked as if they had been moving on their routes for a long time.

### Improve Tradeships.lua

Preparations:

- Add method `Ship:GetCargo`
- Add method `Ship:GetDurationForDistance`
- Add method `Ship:Create`
- Add method `ui.treeNode & ui.treePop`
- Add method `LuaObject::CreateInLua`
- Add functions for convenient selection of a random array element
- Add some numerical utilites
- Add a tool to easily draw a table from an array
- Fix `ui.Format.Number`
- Turn off the output when the ship changes frames
- Reduce the detection range of hyperspace clouds
- Remove the call `LuaEvent::Emit` immediately after exiting hyperspace
- Split `Tradeships.lua` into 5 modules

Tradeships.lua

- For current system, calculate all possible routes of tradeships, durations, flows; arrange ships in space in such equilibrium state, as if the have been flying about their business for a long time.

- Calculate the average spawn interval for new ships, and parking intervals, so that the stations are not overcrowded or empty.

- Сreate a tab in the debug window with full information on routes, ships, stations, remote systems.

### Fix some bugs appeared

- Move the drawing of the system map pigui to a separate handler to ensure that it will always be drawn strictly after the `SystemView` object.
- Fix the setting of the `m_colliding` parameter of the ship during deserialization.
### Reduce the flow of tradeships

- reduce the flow to the most popular station in system by 40%.
- reduce the maximum station load by 25%
- reduce the maximum ships limit by 40%

This is to improve the performance. although these parameters are independent, it is better to change them synchronously.  For example, if you just reduce the allowable number of ships, in all systems this number can become the same.  If the load on the station is not reduced, the ships will stay there for too long.

Details on the forum: https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=650


The-EG:
Just a heads up: adding sources below src/ [automatically adds them to the build](https://github.com/pioneerspacesim/pioneer/blob/master/CMakeLists.txt#L139), which I'm guessing you didn't intend for src/ship/PPTester since it has a `main`. Also, it looks like you're missing `#include <algorithm>` for `std::min`.

Gliese852:
@The-EG Thanks! To be honest, I have not yet figured out what to do with `PPTester.cpp`, probably this is a draft for a unit test, if I figure out how to make it properly. Or just throw it away.

impaktor:
@Gliese852 Perhaps it's something we can add to the Ctrl-i menu? (I don't know what it does, but I like in-game debug tools!)

Gliese852:
@impaktor I just can't think of how it can be useful in debugging.  Rather, it can be useful for navigation in space.  With minor modifications it will be possible to calculate the time of arrival at the station, for example.

impaktor:
Yay! The two new Github Actions ran (gcc & clang).

Gliese852:
@impaktor it's strange that the checks passed, locally I get a linking error, because there are two `main` functions

The-EG:
Odd that the checks aren't at least warning about the duplicate main symbols, but it's definitely there. 
The [appveyor build actually picked up the main defined in PPTester.cpp](https://ci.appveyor.com/project/pioneerspacesim/pioneer/builds/36442281#L1117) by the looks of it. That's _supposed_ to be the modelcompiler...I'm guessing GCC and clang are doing something similar.

Gliese852:
I decided to replace a separate test file `PPTester.cpp` with the ability to output a test report directly inside the game.
 if you define the value `PRECALCPATH_TESTMODE`, while calling the functions of this class, a report similar to this will be output:
```
--------------------------------------------------------------------------------
Testing PrecalcPath::setTime()
.
Ship parameters:
EV:             18200000.00
F:               6100000.00
acap:                   inf
mass:             188000.00
fuel:              48000.00
margin:                0.90
.
Path and point parameters:
Stotal:     139533955305.25
V0:                    0.00
time:              96876.48
.
parameter |      calculated |       simulated |   deviation | result
----------|-----------------|-----------------|-------------|-------
Duration  |       126804.42 |       126804.50 |    0.00006% | OK
Vmax      |      2201119.79 |      2201119.26 |    0.00002% | OK
V         |      1080400.80 |      1080399.30 |    0.00014% | OK
fuel      |        16635.80 |        16635.79 |    0.00004% | OK
dist      | 123206922481.00 | 123207023018.29 |    0.00008% | OK
state     |              -1 |              -1 |             | OK
.
Test passed successfully!
--------------------------------------------------------------------------------
```

Gliese852:
now it's even playable

Web-eWorks:
@Gliese852 given that the TradeShips.lua file is now well over 1800 lines of Lua code (even with debug code), I would *highly* recommend that you create a folder for it in `data/modules/` and split the file up into several sub-files. It massively helps readability to have similar functionality grouped into smaller files, and also means we can eventually promote functionality that's common to multiple modules to a top-level library.

src/lua/LuaSpace.cpp
```diff
@@ -150,12 +186,125 @@ static int l_space_spawn_ship(lua_State *l)
 	return 1;
 }
 
+// functions from ShipAiCmd.cpp
+extern int CheckCollision(DynamicBody *dBody, const vector3d &pathdir, double pathdist, const vector3d &tpos, double endvel, double r);
+extern double MaxEffectRad(const Body *body, Propulsion *prop);
+
+/*
+ * Function: PutShipOnTheRoute
```
Web-eWorks, line 194: Generally speaking, in English (and our codebase specifically) you drop articles like 'The' and 'An' - this would be `PutShipOnRoute`; if there's a need to clarify which route, you would add a less context-dependent word (e.g. `PutShipOnDirectRoute`). It's convention to strip articles for brevity and style; it's not `SpawnAShip`, it's `SpawnShip`.

Just something that stood out to me that might not be as obvious to a non-native English speaker. :smile:

src/ship/PrecalcPath.cpp
```diff
@@ -0,0 +1,551 @@
...
+//     1 - acceleration
+//     2 - decceleration
+//     i.e. t1 - acceleration time, S1 - acceleration path, m1 - mass after acceleration e.t.c.
+//     since free flight is very trivial it is not numbered
+//     if no stage is specified, the calculation is the same for both stages
+//
+//   example:
+//     S1_from_t1_norm() - calculate acceleration path from acceleration time, with no acceleration limit
+
+using val = const double &;
```
Web-eWorks, line 40: I'm not sure why you would need to use `const double &` here - double is a scalar value type that is the same size as the implicit pointer that makes up a pass-by-reference function argument.

Were this being compiled in unoptimized mode, it would require extra instruction(s) to take a reference to a value and pass it to a function rather than just copying the double on the stack. When this code compiled in optimized mode, all of these functions are getting inlined anyways; which means there's no actual benefit to this over typedef-ing this to just `double`.

Web-eWorks (changes requested):
A few things stood out to me; I'll wait to review the Lua until it's a little bit more manageable (and I have more time).

src/lua/LuaSpace.cpp
Gliese852: Done!

src/ship/PrecalcPath.cpp
Gliese852: Done!

Gliese852:
> @Gliese852 given that the TradeShips.lua file is now well over 1800 lines of Lua code (even with debug code), I would _highly_ recommend that you create a folder for it in `data/modules/` and split the file up into several sub-files. It massively helps readability to have similar functionality grouped into smaller files, and also means we can eventually promote functionality that's common to multiple modules to a top-level library.

@Web-eWorks divided `Tradeships.lua` into 5 modules, and brought something to the libs

data/modules/TradeShips/Core.lua
```diff
@@ -0,0 +1,159 @@
...
+			inbound - in system and given AIDockWith order
+			docked - currently docked or un/docking
+			outbound - heading away from starport before hyperspacing
+			fleeing - has been attacked and is trying to get away
+			(currently still just following whatever AI order it had)
+			cowering - docked after having been attacked, waiting for
+			           attacker to go away
+			orbit - was unable to dock, heading to or waiting in orbit
+		attacker - what this was last attacked by; Body object
+		chance - used to determine what action to take when attacked; number
```
Web-eWorks, line 45: With large comments like this, please prefer putting them outside of the table initializer braces and reformatting them to read something like:
```lua
--[[
    Property: Core.ships
    ...
--]]
Core.ships = nil
```

It helps with readability and generally makes the file look nicer when doc comments are structured consistently.

data/modules/TradeShips/Debug.lua
```diff
@@ -0,0 +1,333 @@
...
+local e = require 'Equipment'
+local Game = require 'Game'
+local ui = require 'pigui.baseui'
+local utils = require 'utils'
+local Vector2 = _G.Vector2
+
+local Core = require 'modules.TradeShips.Core'
+
+local drawArrayTable = require 'pigui.libs.array-table'.draw
+local addKeys = require 'pigui.libs.array-table'.addKeys
```
Web-eWorks, line 16: I would recommend doing `local arrayTable = require 'pigui.libs.array-table'` and then calling these functions via `arrayTable.draw()` or `arrayTable.addKeys()`; not only does it fit a bit better with existing conventions, but it also logically follows the model of using the local variable as a package namespace.

data/modules/TradeShips/Debug.lua
```diff
@@ -0,0 +1,333 @@
...
+				{ name = "Type",        key = "eq_type",                       string = true },
+				{ name = "Units",       key = "count"                                        },
+				{ name = "Unit's mass", key = "mass",     fnc = format("%dt")                },
+				{ name = "Total",       key = "all_mass", fnc = format("%dt")                }
+			},
+			{ totals = {
+					{ name = "Total:",    all_mass = total_mass            },
+					{ name = "Capacity:", all_mass = capacity              },
+					{ name = "Free:",     all_mass = capacity - total_mass },
+			}})
```
Web-eWorks, line 328: Same thing here, this is doubly-indented. There are several instances of similar cases in this file (and possibly others); I'm only going to point this one out for brevity's sake.

data/modules/TradeShips/Debug.lua
```diff
@@ -0,0 +1,333 @@
...
+				local count = ship:CountEquip(et)
+				if count > 0 then
+					local all_mass = count * et.capabilities.mass
+					table.insert(equipItems, {
+							name = et:GetName(),
+							eq_type = t,
+							count = count,
+							mass = et.capabilities.mass,
+							all_mass = all_mass
+						})
```
Web-eWorks, line 310: This is doubly-indented. In general, only increase indent by one tabstop per new line, regardless of how many 'new block' characters are on the preceeding line (e.g. open-paren, open-brace, etc.)

data/modules/TradeShips/Debug.lua
```diff
@@ -0,0 +1,333 @@
...
+				row.msg and string.match(row.msg, search_text) or
+				row.model and string.match(row.model, search_text)
+			end),{
+				{ name = "Time",    key = "time" , fnc = ui.Format.Datetime                },
+				{ name = "Label",   key = "label",                           string = true },
+				{ name = "Model",   key = "model",                           string = true },
+				{ name = "Message", key = "msg",                             string = true }
+			},{
+				callbacks = {
+					onClick =	function(row)
```
Web-eWorks, line 261: Extra tab between `=` and `function`. There are a few more of these that I'm not going to individually point out.

data/modules/TradeShips/Debug.lua
```diff
@@ -0,0 +1,333 @@
...
+			drawArrayTable("tradeships_log", Core.log, Core.log.iter(search_text ~= "" and function (row)
+				return
+				row.label and string.match(row.label, search_text) or
+				row.msg and string.match(row.msg, search_text) or
+				row.model and string.match(row.model, search_text)
+			end),{
+				{ name = "Time",    key = "time" , fnc = ui.Format.Datetime                },
+				{ name = "Label",   key = "label",                           string = true },
+				{ name = "Model",   key = "model",                           string = true },
+				{ name = "Message", key = "msg",                             string = true }
```
Web-eWorks, line 258: If you're going to align table values, I would recommend aligning the start of the key-value pairs; these are weirdly aligned and in my opinion not worth the extra characters.

Web-eWorks (changes requested):
Slowly but surely making my way through this review. Here's a batch of change requests; I'll dig into the meat of the PR further tomorrow.

data/modules/TradeShips/Events.lua
```diff
@@ -0,0 +1,325 @@
...
+					trader['status'] = 'inbound'
+					trader.ts_error = "attack!"
+				elseif trader.status == 'cowering' then
+					-- already reached starport and docked
+					-- we need to do everything that the ship does after docking
+					onShipDocked(t_ship, trader.starport)
+				end
+				return
+			end
+		end
```
Web-eWorks, line 314: This seems like an unnecessarily expensive function; I would suggest maintaining a cache of attacking ships to test if we need to iterate all tradeships in a system. Making it a table with weak keys would probably be the best route here.

data/modules/TradeShips/Trader.lua
```diff
@@ -0,0 +1,355 @@
...
+	local ship_type = ShipDef[ship.shipId]
+
+	-- add standard equipment
+	ship:AddEquip(e.hyperspace['hyperdrive_'..tostring(ship_type.hyperdriveClass)])
+	if ShipDef[ship.shipId].equipSlotCapacity.atmo_shield > 0 then
+		ship:AddEquip(e.misc.atmospheric_shielding)
+	end
+	ship:AddEquip(e.misc.radar)
+	ship:AddEquip(e.misc.autopilot)
+	ship:AddEquip(e.misc.cargo_life_support)
```
Web-eWorks, line 27: I don't see any provisions here made for cargo capacity. Is the intent that when a ship is full, AddEquip will just silently fail? Or is this made on the assumption that this function will always be operating on ships with enough capacity to hold all of these modules and cargo besides?

data/modules/TradeShips/Trader.lua
```diff
@@ -0,0 +1,355 @@
...
+	if direction == 'import' and #imports == 1 then
+		total = ship:AddEquip(imports[1], empty_space)
+		ship_cargo[imports[1]] = total
+	elseif direction == 'export' and #exports == 1 then
+		total = ship:AddEquip(exports[1], empty_space)
+		ship_cargo[exports[1]] = total
+	elseif (direction == 'import' and #imports > 1) or
+		(direction == 'export' and #exports > 1) then
+
+		-- happens if there was very little space left to begin with (eg small
```
Web-eWorks, line 76: Instead of having all of these branches for import vs export, prefer to assign the list of cargo to a single variable, e.g.:
```lua
local cargoTypes = direction == 'import' and Core.params.imports or Core.params.exports

if #cargoTypes == 1 then
    total = ship:AddEquip(cargoTypes[1], empty_space)
    -- ...
elseif #cargoTypes > 1 then
    local cargo_type = cargoTypes[Engine.rand:Integer(1, #cargoTypes)]
    -- ...
end
```

data/modules/TradeShips/Trader.lua
```diff
@@ -0,0 +1,355 @@
...
+	ship:AIEnterLowOrbit(body)
+	trader['status'] = 'orbit'
+	Core.log:add(ship, 'Ordering orbit of '..body.label)
+end
+
+local getSystem = function (ship)
+	local max_range = ship:GetEquip('engine', 1):GetMaximumRange(ship)
+	if max_range > 30 then
+		max_range = 30
+	end
```
Web-eWorks, line 129: Prefer `max_range = math.min(max_range, 30)`

data/modules/TradeShips/Trader.lua
```diff
@@ -0,0 +1,355 @@
...
+	local min_range = max_range / 2;
+	local systems_in_range = Game.system:GetNearbySystems(min_range)
+	if #systems_in_range == 0 then
+		systems_in_range = Game.system:GetNearbySystems(max_range)
+	end
+	if #systems_in_range == 0 then
+	return nil end
```
Web-eWorks, line 136: This line should either be indented or on the previous line following `then`.

data/modules/TradeShips/Trader.lua
```diff
@@ -0,0 +1,355 @@
...
+		trader.delay = delay
+		trader.fnc = fn_name
+		Trader.callInThisSystem(delay, function()
+			if(not trader_task[fn_name](ship)) then
+				-- after the function ends, remove the assignment
+				-- but we don't do this if the function returned something, because in
+				-- that case it assigned a new task
+				trader.delay = nil
+				trader.fnc = nil
+			end
```
Web-eWorks, line 332: Instead of returning a `true` value to avoid cleanup, I'd recommend statically setting `trader.delay` etc. to nil before calling the function; if the function needs those variables as context you can pass them in from the upvalues as extra parameters. E.g.:
```lua
trader.delay = nil
trader.fnc = nil
trader_task[fn_name](ship, delay, fn_name)
```

Also it seems slightly redundant to be setting those values on the ship, as AFAICT they're only used here, as upvalues, and the copy on the ship object is there for debugging purposes.

Web-eWorks (changes requested):
Yet more changes - I'm sure some of these are pointing at code you didn't write, but it's a good time to do some cleanup of the module anyways.

data/libs/options.lua
```diff
@@ -0,0 +1,91 @@
+-- Copyright © 2008-2021 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local Engine = require 'Engine'
+local options = {}
```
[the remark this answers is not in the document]
Gliese852: Moved to utils.lua

data/libs/utils.lua
```diff
@@ -285,4 +286,62 @@ utils.round = function(x, n)
...
+end
+
+--
+-- Function: utils.asymptote
+--
+-- value = utils.asymptote(x, max_value, equal_to_ratio)
+--
+-- Return:
+--
+--   value - number
```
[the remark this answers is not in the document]
Gliese852: Аdded a description

data/modules/TradeShips/Core.lua
Gliese852: Done!

data/modules/TradeShips/Debug.lua
Gliese852: Done!

data/modules/TradeShips/Debug.lua
Gliese852: Done!

data/modules/TradeShips/Debug.lua
Gliese852: Done!

data/modules/TradeShips/Debug.lua
Gliese852: Done!

data/modules/TradeShips/Debug.lua
Gliese852: Done!

data/modules/TradeShips/Events.lua
Gliese852: Made a cache as a table, where the key is the object of the attacker, and the value is an array of weak links that were added to the traders.

data/modules/TradeShips/Trader.lua
Gliese852: @Web-eWorks it is assumed that this function will always run on an empty ship, It seems to work well for all ships. I think that the equipment of the ship is quite a broad topic, so I most likely did not even touch this function. It seems to me that for each ship model an optimal set of equipment should be designed for different purposes (roles).

data/modules/TradeShips/Trader.lua
Gliese852: Done!

data/modules/TradeShips/Trader.lua
Gliese852: Done!

data/modules/TradeShips/Trader.lua
Gliese852: Done! Indeed, that is much better!

data/modules/TradeShips/Trader.lua
Gliese852: Done!

data/libs/Ship.lua
```diff
@@ -292,6 +292,35 @@ Ship.RemoveEquip = function (self, item, count, slot)
 	return ret
 end
 
+--
+-- Method: GetCargo
+--
+-- Build a table where the keys will be the names of the cargo, and the values
+-- will be their number
```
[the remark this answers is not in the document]
Gliese852: Done!

data/modules/TradeShips/Trader.lua
Gliese852: added an assertion that the cargo hold is empty

Web-eWorks: The idea of a default loadout makes sense and is something I've been wanting to look into for a bit; though the logistics of passing the data around is something that I think needs a bit more thought than trying to tack it on to ShipDef.

I was mostly asking if there were conditions where a small ship might get chosen and potentially only have room for a hyperdrive and a few (<10) tons of cargo; though if there's code above this function that's only selecting ships that are intended to be larger cargo ships then this isn't an issue at all.

data/pigui/baseui.lua
```diff
@@ -25,7 +25,22 @@ ui.pi_4 = pi_4
 ui.pi = pi
 
 ui.anchor = { left = 1, right = 2, center = 3, top = 4, bottom = 5, baseline = 6 }
-
+ui.fullScreenWindowFlags = ui.WindowFlags {"NoTitleBar", "NoResize", "NoMove", "NoInputs", "NoSavedSettings", "NoFocusOnAppearing", "NoBringToFrontOnFocus"}
```
Web-eWorks, line 28: There is an ImGui flag `ImGuiWindowFlags_NoBackground` that would be perfect for this purpose and avoid the need for a `ui.withStyleColors` call. It's not yet exposed, although adding it is just an extra line in the `imguiWindowFlagsTable` in `src/lua/LuaPiGui.cpp`.

src/Body.cpp
```diff
@@ -234,7 +234,7 @@ void Body::UpdateFrame()
 		FrameId parent = frame->GetParent();
 		Frame *newFrame = Frame::GetFrame(parent);
 		if (newFrame) { // don't fall out of root frame
-			Output("%s leaves frame %s\n", GetLabel().c_str(), frame->GetLabel().c_str());
+			// Output("%s leaves frame %s\n", GetLabel().c_str(), frame->GetLabel().c_str());
```
Web-eWorks, line 237: If this is not a performance issue, I'd recommend converting these calls to `Log::Verbose("{} leaves frame{}\n", GetLabel(), frame->GetLabel());`. This way they don't show up in the console, but are still logged for debugging perusal if the log system is set to allow verbose messages in the log file.

src/Frame.h
```diff
@@ -82,7 +82,7 @@ class Frame {
 	bool HasRotFrame() const { return m_flags & FLAG_HAS_ROT; }
 
 	FrameId GetParent() const { return m_parent; }
-	FrameId GetNonRotFrame() { return IsRotFrame() ? m_parent : m_thisId; }
+	FrameId GetNonRotFrame() const { return IsRotFrame() ? m_parent : m_thisId; }
 	FrameId GetRotFrame() { return HasRotFrame() ? m_children.front() : m_thisId; }
```
Web-eWorks, line 86: Is there a reason `GetRotFrame()` isn't marked `const` as well? I'd recommend marking it const if there's nothing preventing that.

src/Game.cpp
```diff
@@ -289,14 +289,6 @@ void Game::TimeStep(float step)
...
 			SwitchToNormalSpace();
 			m_player->EnterSystem();
-			// event "onEnterSystem(player)" is in the event queue after p_player->EnterSystem()
-			// in this callback, for example, fuel is reduced
-			// but event handling has already passed, and will only be in the next frame
-			// it turns out that we are already in the system, but the fuel has not yet been reduced
-			// and then the drawing of the views will go, and inconsistencies may occur,
-			// for example, the sphere of the hyperjump range will be too large
-			// so we forcefully call event handling again
-			LuaEvent::Emit();
```
Web-eWorks, line 299: What was changed to make this comment and call no longer necessary? I don't need it documented in the code, but I'd appreciate an explanation of what resolved this rather long comment about an edge case.

src/lua/LuaObject.h
```diff
@@ -376,6 +405,14 @@ inline void LuaObject<T>::PushToLua(const T &o)
...
 }
 
+template <typename T>
+template <typename... Args>
+inline void LuaObject<T>::CreateInLua(Args &&... args)
+{
+	T *p(new T(std::forward<Args>(args)...));
+	Register(new (LuaObjectBase::Allocate(sizeof(LuaOwnObject<T>))) LuaOwnObject<T>(static_cast<T *>(p)));
+}
+
```
Web-eWorks, line 415: I would *strongly* recommend a comment on this function warning users that `CreateInLua` cannot be mixed with ref-counted pointers; when the Lua object is GC'd, it will delete the underlying object regardless if it is of a refcounted type and has active references pointing to it.

src/lua/LuaShip.cpp
```diff
@@ -660,6 +662,42 @@ static int l_ship_is_ecm_ready(lua_State *l)
 	return 1;
 }
 
+/*
+ * Method: GetDurationForDistance
+ *
+ *   Calculating the duration of the flight of a given ship at a specified distance.
+ *   Assumed that at the beginning and at the end of the travel the speed is 0.
+ *
+ * > duration = ship:GetDurationForDistancel(distance)
```
Web-eWorks, line 671: Very minor nitpick but the code example here has a typo. :smile:

Web-eWorks (changes requested):
Yet another batch of changes; only three files are left to review but I'm out of time to review those right now.

src/Game.cpp
Gliese852: @Web-eWorks The purpose for which the `LuaEvent::Emit()` call was added is described in the comment. Now it turns out that _all tradeships are generated in this call_. Вut in this frame, delta_t is in effect, which was in effect during the hyperjump. In the same frame, the next position of these ships is calculated, based on this _huge_ delta_t. This leads to the fact that in the very first frame the positions of the ships are very displaced from the positions where they were generated, and the speeds turn out to be too high, therefore, some of the ships that are generated near the destination point may even fly by. We didn't notice it before, because tradeships spawned at zero speeds.

I thought that the first problem is not as bad as the second, and in order not to dig into the solution of the first problem, I just accept it. Just if you get out of the hyperjump and you have the sector map turned on, your jump range sphere will be displayed too large in the first frame, it will decrease to the correct size only in the next frame, that's the whole problem.

src/lua/LuaObject.h
Gliese852: @Web-eWorks but only the parameters of the constructor are passed to this function, so it is impossible to simply pass refcounted there, only through the copy-constructor maybe, and if I'm not mistaken, this will lead to an increase in the counter, and everything will be fine?

data/pigui/baseui.lua
Gliese852: Did it for `ui.makeFullScreenHandler`, but did not do it for the "game" handler (`ui.registerHandler("game" ... )`, because almost all modules there override the window flags and the windows become visible

src/Body.cpp
Gliese852: Done!

src/Frame.h
Gliese852: Done!

src/lua/LuaShip.cpp
Gliese852: Done!

Gliese852:
@Web-eWorks probably I will put the changes that I make in separate commits, and then I will squash them all where necessary

src/lua/LuaObject.h
Web-eWorks: However - the object that is generated in lua can be copied around perpetually and passed to other functions that might take a pointer to an object. Our calling convention is to `LuaPull<ClassName *>()`, not `LuaPull<RefCountedPtr<ClassName>>()`, which means that if CreateInLua is called for an object inheriting from `RefCounted` it's possible that it's pushed to Lua as a LuaOwnedObject, pulled as a `ClassName *`, and passed to a function expecting a `RefCountedPtr<ClassName>` (even if such function doesn't *yet* exist). In this case, either there will be a dangling reference or the object stored in LuaOwnObject will be deleted by a ref-count decrement.

I'd specifically like to see a comment that makes this restriction explicit so that a future junior programmer / contributor doesn't need to fully understand the nuances of the lua binding model to be able to make an informed decision about which of the *four* seemingly-equal methods of pushing data to Lua is viable for their specific situation.

data/modules/TradeShips/Flow.lua
```diff
@@ -0,0 +1,535 @@
...
+			Core.ships[ship] = nil
+			removed = removed + 1
+		else
+			trader.status = 'hyperspace'
+		end
+	end
+	-- print('updateTSTable:total:'..total..',removed:'..removed)
+end
+
+return Flow
```
Web-eWorks, line 535: Missing newline at end of file.

data/modules/TradeShips/Flow.lua
```diff
@@ -0,0 +1,535 @@
...
+				if routes_for_ship then
+					trader.route = utils.chooseNormalized(routes_for_ship)
+				else
+				-- TODO but what if there are no ports for it?
+				end
+			end
+		end
+	else
+		Core.ships = {}
+	end
```
Web-eWorks, line 440: Instead of the indentation here, I'd recommend checking for `Core.ships == nil`, creating the table, and moving the loop out of the if statement. That would save one level of indentation, and if the table is newly created will still run no iterations.

```lua
if Core.ships == nil then Core.ships = {} end
for ship, trader in pairs(Core.ships) do
```

data/modules/TradeShips/Flow.lua
```diff
@@ -0,0 +1,535 @@
...
+			ships_in_space = 0
+			total_flow = 0
+			for _, ship_routes in pairs(local_routes) do
+				for _, ship_route in ipairs(ship_routes) do
+					ship_route.flow = flow_per_ship * ship_route.weight
+					total_flow = total_flow + ship_route.flow
+					ship_route.amount = ship_route.flow * ship_route.duration / 3600
+					ships_in_space = ships_in_space + ship_route.amount
+				end
+			end
```
Web-eWorks, line 241: Instead of running a loop twice and breaking halfway through, prefer to move this first half of the loop that calculates `ships_in_space` into a local function (can be defined in the local scope of the function) and call it as needed.

data/modules/TradeShips/Flow.lua
```diff
@@ -0,0 +1,535 @@
...
+	-- now set the flow on the most popular route to MAX_ROUTE_FLOW
+	-- let's collect all data on routes and calculate the most popular route weight
+	local local_routes_map = {}
+	for _,routes_for_ship in pairs(local_routes) do
+		for _, route_for_ship in ipairs(routes_for_ship) do
+			if local_routes_map[route_for_ship.to] then
+				local_routes_map[route_for_ship.to] = local_routes_map[route_for_ship.to] + route_for_ship.weight
+			else
+				local_routes_map[route_for_ship.to] = route_for_ship.weight
+			end
```
Web-eWorks, line 211: Prefer a construct like:
```lua
local current = local_routes_map[route_for_ship.to] or 0
local_routes_map[route_for_ship.to] = current + route_for_ship.weight
```

Web-eWorks (changes requested):

src/lua/LuaObject.h
Gliese852: @Web-eWorks wrote a comment, hopefully ok

src/ship/PrecalcPath.h
```diff
@@ -0,0 +1,66 @@
...
+public:
+	PrecalcPath(
+		const double Stotal,
+		const double V0,
+		const double EV,
+		const double F,
+		const double acap,
+		const double mass,
+		const double fuel,
+		const double margin);
```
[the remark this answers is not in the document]
Gliese852: Done!

src/ship/PrecalcPath.h
```diff
@@ -0,0 +1,66 @@
...
+
+	// set current point of path
+	// by ratio (0.0 .. 1.0) of path completion (in distance)
+	void setSRatio(const double ratio) { setDist(Stotal * ratio); }
+	// by ratio (0.0 .. 1.0) of path completion (in time)
+	void setTRatio(const double ratio) { setTime(getFullTime() * ratio); }
+	// by distance from start of path
+	void setDist(const double S);
+	// by time from start
+	void setTime(const double t);
```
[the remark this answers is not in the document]
Gliese852: Done!

data/modules/TradeShips/Flow.lua
Gliese852: Fixed!

data/modules/TradeShips/Flow.lua
Gliese852: Done!

data/modules/TradeShips/Flow.lua
Gliese852: Done!

data/modules/TradeShips/Flow.lua
Gliese852: Done!

Web-eWorks (approved):
I haven't had a chance to playtest this (or anything else) for a while now, but the code is good to go! Thank you @Gliese852 for working so hard to make the code easy to maintain and understand for contributors (and for putting up with my pedantic style suggestions :smile:) - this PR is quite a big one, but you've done an excellent job with it.

Gliese852:
@Web-eWorks Thank you very much for taking so much of your time to review!
I will squash all tempourary commits soon.

I played for a few hours, mostly courier missions, everything seems to work stably.  
Savegame size with 700 ships in space is about 500 kb.
FPS dropped significantly but your 'shipfest' branch will fix the problem.
If the FPS turns out to be uncomfortable afterwards, it is quite easy to reduce the flow of ships in the master.
There are some non-critical issues with autopilot, but fixes are almost ready in #5127

Gliese852:
Squashed all tempourary commits

fluffyfreak:
Don't worry about the VS failure, I'll fix that up once it's merged.

Web-eWorks:
Building in `RelWithDebInfo` mode, I'm noticing a major frame drop in Sol when turning ship orbit display on in the System View. While I'd like to further improve that particular view's CPU usage, for now I'd suggest dropping the number of tradeships down to ~300 (from 500 in Core.lua) in deference to the members of our community with lower-spec PCs. I'd love to make that number configurable (and be able to bring it back up) but right now I don't think this is *quite* performant enough to merge as the 'default'.

Otherwise, this is an amazing PR and I really love the way the tradeship debug view looks with the new debugger color scheme!
Screenshot from 2021-05-01 19-55-34

impaktor:
What's RelWithDebInfo?

OK, so just drop default number of ships and this will be ready for merge?

Web-eWorks:
> What's RelWithDebInfo?
> 
> OK, so just drop default number of ships and this will be ready for merge?

Release with Debug Info -  it's a default Cmake build target that compiles with optimizations and debug symbols (what we ship in releases).

Yes, just drop the default number of ships until such a time as we're able to optimize some of the remaining bottlenecks in the code, and this PR is ready to merge.

Gliese852:
@Web-eWorks I'll probably make this change in a separate commit to clarify how the flow is set up.

Web-eWorks:
Merged! Thank you for all of the work on this @Gliese852!
@fluffyfreak whenever you've recovered, the VS files need updating for the new files involved in the PR.

Issue #4551, opened 2019-03-04
Search and Rescue crash for deliver crew mission


impaktor:
Copenhagen_Bram reported this on a S&R mission, landing next to a mission ship on Earth for a transfer:

```
error: [string "[T] @libs/Ship.lua"]:922: attempt to get length of field '?' (a nil value)
stack traceback:
        [string "[T] @libs/Ship.lua"]:922: in function 'CrewNumber'
        [string "[T] @modules/SearchRescue/SearchRescue.lua"]:1627: in function 'deliverCrew'
        [string "[T] @modules/SearchRescue/SearchRescue.lua"]:1790: in function <[string "[T] @modules/SearchRescue/SearchRescue.lua"]:1765>
Segmentation fault (core dumped)

```
Just looking at those lines of code, I think this is what I'd expect to happen if the mission is to deliver crew to a mission ship, that is empty. 

- possibly, a mission ship should never be empty, i.e. exclude single seat
  ships from "deliver crew (pilot/doctor)" to mission ship?
  
- Or L922 in lib/Ship.lua shouldn't crash if CrewRoster is empty.

``` lua
Ship.CrewNumber = function (self)
	return #CrewRoster[self]
end
```

ElonSatoshi:
uhh that's actually `pioneer 20190303 (74e38aa37)` compiled with cmake.

I was trying to make an issue but I need to take a bath and manage to load an autosave without segfaulting.

ElonSatoshi:
How to reproduce:

1. 
```
git clone https://github.com/pioneerspacesim/pioneer
cd pioneer
git checkout 74e38aa37
./bootstrap cmake && cmake --build ./build
```

2. Stick this [save file](https://github.com/pioneerspacesim/pioneer/files/2928515/search.and.rescue.json.gz)
into your ~/.pioneer/savefiles folder and try to open it, but be warned that if I try to load a save with this version it segfaults, an issue which I will open after writing this. If you can't open it, play Pioneer and do search and rescue missions and hope you get the same one I got? Maybe there's a thing you can type into the console. Make sure you have an extra passenger bay.

3. Accept the search and rescue mission, fly to the stranded vessel and land next to it, wait for the crew transference.

clausimu:
[...]

PR #4985, opened 2020-10-27
Fix SAR crash where IsPlayer() is falsely called by non-ship body

clausimu:
This fixes the last comment of #4551. When a SAR mission ship is destroyed the script tries to establish if it was the player's fault or not. But IsPlayer() is not defined for bodies other than ships. This should fix this issue. The save-file attached to the bug report runs fine now.

I was not able to reproduce the other comments in this bug report (crash after trying to deliver crew). Maybe this got fixed together with #4601? Also, the attached respective save file does not load, which makes it hard to pinpoint if the problem was solved.

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -1986,7 +1986,7 @@ local onCreateBB = function (station)
 	for _ = 1,num do
 		-- makeAdvert(station, 1, closestplanets)
 		-- makeAdvert(station, 2, closestplanets)
-		-- makeAdvert(station, 3, closestplanets)
+	   -- makeAdvert(station, 3, closestplanets)
 		-- makeAdvert(station, 4, closestplanets)
 		-- makeAdvert(station, 5, closestplanets)
 		-- makeAdvert(station, 6, closestplanets)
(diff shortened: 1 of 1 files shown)


data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -2286,12 +2286,13 @@ end
...
-			else
-				mission.target_destroyed = "BY_ACCIDENT"
-			end
+		   mission.target = nil
+		   mission.target_destroyed = "BY_ACCIDENT"
+		   if attacker:IsDynamic() then
+		      if attacker:IsPlayer() then
+			 mission.target_destroyed = "BY_ACCIDENT"
+		      end
+		   end
```
Web-eWorks, line 2295: Looking good! I have only two things to suggest here:
- First of all, I'd recommend making sure to normalize your indentation, it looks like this is indented with three spaces instead of one tab as we'd prefer.
- Secondly, I'd recommend the use of idiomatic lua and write `if attacker:IsDynamic() and attacker:IsPlayer() then ... end`. Lua does *short-circuit* evaluation, so if the IsDynamic call fails the IsPlayer call will not execute.

And finally, did you mean for the target_destroyed reason to be `BY_PLAYER` instead of `BY_ACCIDENT` inside the if statement?

Web-eWorks (changes requested):

data/modules/SearchRescue/SearchRescue.lua
clausimu: @Web-eWorks - correct on all accounts. Thank you! I (hopefully) fixed the indentation, replaced BY_ACCIDENT with BY_PLAYER, and combined both if statements.

Web-eWorks (approved):

PR #4991, opened 2020-10-31
Decreasing the max star size to 0.3

nozmajner:
I think the bigger star size on the starfield is too excessive, it clashes with the HUD, and isn't really feel like they are brighter stars.
So I've decreased the _brightnessApparentSizeFactor_ from 0.5 to 0.3
(Enlarge the screenshots to see the effect properly:
I think they look much better at 03. They are more subtle, but in a nice way:

Original 0.5:

At 0.4 they are still a bit too much in my opinion

At 0.1, they aren't visible most of the time:

And as a bonus, at 2.0, we turn [Olbers' paradox](https://en.wikipedia.org/wiki/Olbers%27_paradox) on it's head:

### DIFF
--- a/data/configs/Starfield.ini
+++ b/data/configs/Starfield.ini
@@ -17,7 +17,7 @@ medianPosition=0.7
 brightnessPower=3.5
 
 # the bigger this number is the bigger are stars compared to their brightness
-brightnessApparentSizeFactor=0.5
+brightnessApparentSizeFactor=0.3
 
 #the bigger this number is the bigger are stars no matter their brightness
 brightnessApparentSizeOffset=0.1


Web-eWorks (approved):
Approved, with the eventual intention to dynamically adjust star brightness based on exposure / atmosphere scattering.

ConfedMajor:
@Web-eWorks Don't forget star colors, they can create beautiful effects!
RobloxScreenShot20201031_230847024
Also don't forget nebulaes!
RobloxScreenShot20201031_231104940

WKFO:
I say no to nebulae and fancy stuff unless they are actually in the game.

Gliese852:
I am also against beauties, but other galaxies can be drawn, they are unattainable in the game anyway.

nozmajner:
I agree that there should be no such nebulae, unless we actually have them and they are visitable. Especially because most of them, and galaxies for that matter are rahter small on the sky.

ConfedMajor:
> no such nebulae, unless we actually have them and they are visitable

@nozmajner Uh... Hyperspace clouds? 😅

nozmajner:
Hmm, those could sure benefit from a facelift. But there's no way currently to do anything about them without going knee-deep in the code as far as I know.

fluffyfreak:
> Hmm, those could sure benefit from a facelift. But there's no way currently to do anything about them without going knee-deep in the code as far as I know.

If you wanted to change them then do some mockups, or even buildup a model preferrably involving simple geometry and textures. We could put something together without having to write full-screen, post-processed, ray-marched stuff ;)

nozmajner:
I'm not yet sure how I'd approach the looks of it though. I'll think about it.

PR #4993, opened 2020-11-03
Quality of Life Fixes

### LINKED ISSUE #4992: GL_ARB_clip_control is not supported on GeForce 8600M
[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/5471559/opengl.txt)
[output.txt](https://github.com/pioneerspacesim/pioneer/files/5471563/output.txt)
opengl.txt is from July. I dont know how to obtain more recent version. Mayby pioneer was unable to create one now, before the error occured.

If I can remember, last working version was git master from around august. Or from july, if opengl.txt modification date is a proper marker for it.

### Observed behaviour
Loading stops and error message window appears: 
"Fatal: OpenGL extension GL_ARB_clip_control is not supported by your graphics card or graphics driver version.
Please check to see if your GPU driver vendor has an updated driver - or that drivers are installed correctly."

After accepting, game quits.

### Expected behaviour
Should work, as I have opengl 3.1 (supposed to be) still supported and I have set in my config.ini:
`RendererName=Opengl 3.1`

### Steps to reproduce
Compile & run as with instructions.

Git master version, cloned today.
OS: Arch Linux, updated today (except kernel).

Web-eWorks:
This PR is an accumulation of QoL fixes that I've made in the course of finishing several other major features. Among other highlights, these stand out:

- Cargo ships no longer serialize 4000+ copies of the same cargo item, making saved games about 20% smaller. (This hopefully also doesn't eat your cargo on loading, but keep a watchful eye for the Space Kraken noshing on the industrial equipment in your hold...)
- The game's iconography has been spruced up significantly, both with new icons from @nozmajner and new breakpoint-based rendering logic that should make icon rendering less artifact-prone.
- JSON can now serialize fixed-point numbers, rejoice! This change probably shouldn't have been here, but I needed to pad the list for dramatic effect...
- I've reworked all* of the colors in the default theme! They're now defined as semantic aliases for 42 colors pulled from a unified color palette instead of 75+ individual color literals, making the color theme much easier to maintain. I've also made a few common-sense changes to certain parts of the color theme to ensure readability, better identify certain elements, and reduce the scatter-shot approach to color values.

The work on the theme isn't quite complete; as part of further renderer improvements I'll vet our pipeline to ensure that colors are sRGB correct, which I highly doubt they are at this point.

I have one further change coming to fix #4992 before it's ready for merge.

Before: 
image
image

After:
image
image


The-EG (commented):
Overall, I like the color changes. I think they are subtle, but enough that it introduces better contrast and draws a bit more attention to things that need it.

A knit pick though:
The nav target and warning info on the hud are now an orange instead of yellow. I like the orange, but it doesn't match the indication on the radar (for a proximity warning in this case):
image
This is only the case in the 3d planar radar, the 2D radar does display with the updated color. I imagine it's hard coded?

Also, there's now blue:
image

Which seems like it's the current location/frame:
image

But it doesn't seem to be used anywhere else. The current frame info on the HUD is still white.


Edit, I did also test loading my precious metals trading save game...and it didn't seem to have any issues, but my cargo hold was empty at the time. So i'll need to recompile master, create a new save with stuff in the hold, and try again....

Web-eWorks:
Thank you for testing this @The-EG! I didn't even think about old saves being a problem with the cargo bay; I'll have to add a path to load the old cargo items (shouldn't be difficult).

Regarding the planar radar, as far as I'm aware the colors are indeed hardcoded. The system map / radar color inconsistency wasn't intended, insofar as I thought of the two as two separate entities with their own color scheme, but I'll take that feedback under consideration. The blue on the radar is actually the station, and the system map is simply the radar contact for the player's ship.

I may adjust the shade of color used for the player on the system map, as there's no dot on the radar to communicate what the "right" color is. I'd like to rework the system map colors further in another PR, I didn't spend as much time there as I probably should have. I'll leave that for a further sprint that includes displaying surface stations though!

PR #4994, opened 2020-11-03
CMake: Avoid libGLU dependency

pcercuei:
Set the GLEW_NO_GLU build flag to avoid pulling in <gl/glu.h> when
including <gl/glew.h>, as it otherwise adds a dependency we don't
actually need.


The-EG (approved):
This shouldn't be an issue with Windows at all, but I tested compiling it anyway.

Both the cmake and solution built and ran; the fickle MSVC gods are still satisfied...for now...

PR #4995, opened 2020-11-04
Refactor Body inheritance tree to remove Object and CityOnPlanet

### LINKED ISSUE #22: Refactor Body inheritance tree to remove Object and CityOnPlanet
CityOnPlanet should not inherit from class Object. It makes it necessary to have special-case checks for rendering and collision detection. Instead it should be a seperate discrete class, with multiple instances attached to a planet/starport. The render and collision detection can be done as part of the planet draw.

Once thats gone, everything in the Object tree is also a Body, so Object and Body can be merged. Then there is no ambiguity - anything that inherits from Body is something that can exist in the world space.

Web-eWorks:
So yeah, #22 is what I consider low-hanging fruit to knock out in an hour now. I'm a bit in shock to be honest, this was significantly less painful than I thought it would be.

Fixes #22.

This basically turns Object into a simple non-inherited enum-namespace struct, removes CityOnPlanet from the inheritance tree and cleans up some spots that tested for it when they were never going to be passed a CityOnPlanet instance.

The usual bugtesting steps apply; fly around, crash into things, get into a shootout with the police, try a mission or twelve, and if it doesn't crash and you do die horribly then it's working as intended.


impaktor:
Nice!

Ha, this must be the longest open issue that then gets fixed, as @robn opened it +9 years ago.

robn:
Hah! Nice one :laughing:

fluffyfreak:
Nice I'll try to get a look at it today/tonight 👍 

Also, Hi @robn 😀

fluffyfreak (approved):
Had a quick fly around, might be best to wait for more in depth testing from others with time to spare.
However code looks good (_as always_) and initial testing seems to work just fine 👍

impaktor:
With that, I'd say: merge this sucker, and let players test it on master.

Web-eWorks:
Will hold this open until json-commodities and input-frame-2.0 is merged so I can get some more testing. Shouldn't be more than a week, I think.

Gliese852:
Do I understand correctly that the `Object` structure is used only by children of the `Body`?  does not its separate existence now seem strange? (for a person who does not know the background)

Web-eWorks:
@Gliese852 now Object is just a struct that wraps an enum; there is no inheritance anymore. I've only left it alive because I didn't want to re-write every little thing that used `Object::STAR` or `Object::SHIP`.

Gliese852:
I would completely remove the word `Object`, as it is essentially the same as Body, but confusing.  (I don't know how difficult it is)

Web-eWorks:
@Gliese852 good point, I can probably rename it to BodyType or something like that. The entire Object system is going away eventually in favor of an ECS-based system, so I wanted to avoid as much code-churn as possible.

PR #5002, opened 2020-11-08
Draft: Combat mission: Select all ships with a default hyperdrive

### LINKED ISSUE #4983: Combat.lua expects ships to have default hyperdrive class > 1 (ship rebalance)
### Observed behaviour
I finished a recon mission and just when I exit hyperspace to the payment system, game crashes. Consistent. I have a save file below. (For clarification, you can jump to other systems just fine. It is the payment system that is problematic.)

aaa

### Expected behaviour
No crash.

### Steps to reproduce
https://www.dropbox.com/s/sh728r3i83ow5ac/Bugreport?dl=0
Load the save file and click hyperspace.

42b67ec4278638d892561ed75885804b53f16698 + Nozmajner's latest ship rebalance (Windows 10 x64)

robothauler:
Fixes #4983 
Unfortunately, ships with a default hyperdrive 1 sometimes can't reach the destination because of the loaded equipment. So I removed all the equipment too.

### DIFF
--- a/data/modules/Combat/Combat.lua
+++ b/data/modules/Combat/Combat.lua
@@ -427,7 +427,7 @@ local onEnterSystem = function (player)
 			if mission.complete or Game.time > mission.due then
 				local shipdefs = utils.build_array(utils.filter(
 					function (k,def)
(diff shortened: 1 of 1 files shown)


Web-eWorks:
@robothauler these ships are for a rendezvous mission and not supposed to be hostile aggressors, right? If so, then I think this PR is good to merge.

robothauler:
Right, these ships are only for the rendezvous.

PR #5003, opened 2020-11-10
Herding back the Sun and Moon icons who went astray during the latest iconoclast raid.

nozmajner:
I've added back the Star and Moon icons for the HUD. I've accidentally deleted them when I added the icons @Web-eWorks requested. (Silly me forgeting that ctrl+D isn't the same as shift+D in Inkscape.
Turned out there were quite some excessive unused icons lying around far from the image borders, so I removed those as well.
And while at it, I exported the file as optimized svg. In the end it went from 840k to 314k, which is a nice bonus.


The-EG:
data/pigui/themese/default.lua will also need to be updated: `sun = 35,`

The moon icon is ok though...I guess we now have a duplicate of it. It's also in the 8th row, between the ship and 3 exclamation points, which is what the theme is currently using.

nozmajner:
Thanks for the heads-up!

PR #5005, opened 2020-11-12
Silence Inno Setup compressing messages

The-EG:
> [2020-11-11 20:55:14] <sturnclaw> The-EG: if you could look into hiding all of the Compressing: messages from appveyor that'd be appreciated; they're seriously polluting the build log.

Done.


The-EG:
I tried to still get a 'successful' message out of the installer, but it seems like we get unix style instead; no news is good news. If the installer build does fail, it should still produce error messages in the log.

Also, the remaining 'installing' messages are from the cmake 'install' target.

Web-eWorks:
Thank you very much!

PR #5011, opened 2020-11-15
Fall back to an empty resource if translation isn't found

### LINKED ISSUE #5007: Lang doesn't properly fall back to English translation
Just tested master version, commit 9d157761aacfafa9d997260b07955e21d7905726, wondering, if recent substancial changes in language files handling would lead to problems... I wasn't surprised :smile: 

### Observed behaviour
Pioneer stops with error message:

> couldn't read language file 'lang/module-stationtrafficcontrol/pl.json'
> Fatal: Translation of 'module-stationtrafficcontrol' into language 'pl' not found! This should be in 'data/lang/module-stationtrafficcontrol/pl.json'.
> stack traceback:
>         [C]: in ?
>         [C]: in function 'GetResource'
>         [string "[T] @modules/StationTrafficControl/StationTra..."]:13: in main chunk
> 

Wasn't tested with default language (English). In fact, I'm not sure, how to temporarily change language settings on my Linux session, but I hope it's not essential to fix this problem...

Web-eWorks:
This fixes #5007.

- Fix the time accel buttons not working correctly in Polish
- Fix the main menu's weird approach to loading from two lang resources
- Refactor the way we handle missing translations; log a warning and return a placeholder string

### DIFF
--- a/data/pigui/modules/time-window.lua
+++ b/data/pigui/modules/time-window.lua
@@ -48,7 +48,7 @@ local function displayTimeWindow()
 			time = lc.PAUSED
 		end
 		local tooltip = string.interp(lui.HUD_REQUEST_TIME_ACCEL, { time = time })
-		if ui.coloredSelectedIconButton(icons['time_accel_' .. name], button_size, current == name, frame_padding, color, fg_color, tooltip)
+		if ui.coloredSelectedIconButton(icons['time_accel_' .. name], button_size, current == name, frame_padding, color, fg_color, tooltip .. "##time_accel_"..name)
 		or (ui.shiftHeld() and ui.isKeyReleased(key)) then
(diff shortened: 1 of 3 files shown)


impaktor:
If it didn't crash, I wouldn't have woken up to the fact that there was another language resource added to pioneer, so not all crashes are... equal.

Web-eWorks:
> If it didn't crash, I wouldn't have woken up to the fact that there was another language resource added to pioneer, so not all crashes are... equal.

Don't worry, it logs the issue to stderror and the log file as well as... well, showing you the english version.

I could possibly be persuaded to make it open an empty language file instead so that everything reading from it displays `[NO_JSON]TOKEN_NAME` instead of the english translation.

PR #5014, opened 2020-11-16
Update pioneer-thirdparty Windows binaries

The-EG:
**This will need to be merged at the same time as https://github.com/pioneerspacesim/pioneer-thirdparty/pull/30**

Update both CMake and Visual Studio Solution builds to use the new pioneer-thirdparty binaries. 

In addition, add 32bit option to the Windows CMake builds.

edit: I also removed the static build of libsigc++ from the VS Solution and have it dynamically linking again, since that's what the CMake build does, so the two are consistent. Both release and debug should be working properly for both build tools.


fluffyfreak (approved):
Yep go for it

PR #5018, opened 2020-11-21
Fix build with USE_SYSTEM_LIBLUA, again

pcercuei:
Some recent changes did break USE_SYSTEM_LIBLUA, as they included
"src/lua.h" directly instead of <lua.h>.

### DIFF
--- a/src/lua/LuaMetaType.h
+++ b/src/lua/LuaMetaType.h
@@ -8,7 +8,8 @@
 #include "LuaManager.h"
 #include "LuaPushPull.h"
 #include "LuaTable.h"
-#include "src/lua.h"
+
+#include <lua.h>
 
 class LuaMetaTypeBase {
 public:
--- a/src/lua/LuaShip.cpp
+++ b/src/lua/LuaShip.cpp
@@ -16,7 +16,8 @@
 #include "Space.h"
 #include "SpaceStation.h"
 #include "ship/PlayerShipController.h"
-#include "src/lua.h"
+
+#include <lua.h>
 
 /*
  * Class: Ship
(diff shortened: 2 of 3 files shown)


Web-eWorks:
Yup, that's my editor's fault. Dratted vscode-clangd.
Do be advised that this now breaks non-USE_SYSTEM_LIBLUA builds. I'd recommend adding that define to `buildopts.h` and conditionally including, or dropping support for system lua - moving forward we plan to build Lua as C++ for exception support, so we will be unable to use system lua anyways...

The-EG:
Testing this a bit locally, albeit on Windows...It looks like you can completely remove those 3 `#include "src/lua.h" directives. 

Doing that builds fine for me, and I'm guessing the checks here would as well, and it should still work in the case of `-DUSE_SYSTEM_LUA`.

lua.h (the actual Lua one...not Pioneer's) is included already, in LuaUtils.h, as `#include <lua.hpp>`. lua.hpp just wraps `extern "C"` around lua.h, lualib.h and luaxlib.h and should also be shipped with a system packaged Lua.

pcercuei:
`lua.hpp` does include "src/lua.h" etc. though, so it doesn't pick up the system includes. It should include <lua.h> there.

The-EG:
> `lua.hpp` does include "src/lua.h" etc. though, so it doesn't pick up the system includes. It should include <lua.h> there.

But, lua.hpp is being included as a system include [here](https://github.com/pioneerspacesim/pioneer/blob/master/src/lua/LuaUtils.h#L8), so unless `-DUSE_SYSTEM_LUA` is messed up in another way, the systems lua.hpp should be used instead of the contrib one.

pcercuei:
My bad, I thought lua.hpp was a pioneer-only thing.

orbea:
> moving forward we plan to build Lua as C++ for exception support, so we will be unable to use system lua anyways...

Distros may provide a c++ lua in such cases.

PR #5021, opened 2020-11-23
Add "Borderless Window" configuration option

The-EG:
Add an option for running Pioneer in 'borderless window' mode. This will be the same as windowed mode except without window decorations or a border.

This is desirable on multi-monitor systems or when alt-tabbing. I didn't add a 'fullscreen windowed' because that's just this and setting the resolution to the same as the monitor.

Disabled:
image

Enabled:
image

And a new option on the options screen:
image

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -67,6 +67,10 @@
     "description": "",
     "message": "Backward acceleration"
   },
+  "BORDERLESS": {
+    "description": "Option to enable a borderless window",
+    "message": "Borderless Window"
+  },
   "BULLETIN_BOARD": {
     "description": "",
     "message": "Bulletin Board"
--- a/data/pigui/modules/settings-window.lua
+++ b/data/pigui/modules/settings-window.lua
@@ -196,6 +196,7 @@ local function showVideoOptions()
 	local selectedDetail = keyOf(detailLabels,keyOf(detailLevels, Engine.GetPlanetDetailLevel()))-1
 
 	local fullscreen = Engine.GetFullscreen()
+	local borderless = Engine.GetBorderless()
 	local vsync = Engine.GetVSyncEnabled()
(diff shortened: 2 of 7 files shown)


The-EG:
As discussed in chat, another option would be to tie this to the fullscreen option (so either fullscreen, 'fullscreen windowed' or just windowed).

That might be more intuitive since I think that's how most would use this option anyway, but it would require more info and options (like selecting a monitor) since we'll need to know the resolution of the monitor to set in order to get 'full screen.'

Web-eWorks:
@The-EG if you look at the code in (IIRC) Graphics.cpp, you'll see that we already do this to pick the full screen resolution list - it should be a fairly simple matter to use the same code to generate the borderless window list. Regarding which monitor and which size, borderless windows can be moved between monitors easily (win+shift+left/right on most DEs including Windows as I remember) and most games that I know of don't have a dropdown to pick the monitor; you simply pick the resolution of the monitor you want it on and move the window to that monitor.

The-EG:
> most games that I know of don't have a dropdown to pick the monitor;

Out of the games I have installed, more do than don't...including Elite Dangerous:
image

The-EG:
This is now an option between Fullscreen, Window and Fullscreen Window (borderless).
image

 For the moment, the StartFullscreen and borderless options are still used in config.ini. This also changes RendererGL to setup it's viewport to the actual size of the SDL window and not what was requested in config.ini, which in the case of Fullscreen Window could be different.
I suspect that there's probably something in the UI somewhere that will behave badly if those two values diverge too far, but some initial testing didn't get anything to fall out. 

I'll also have to create additional translation strings (where's the mark as WIP button now?).

fluffyfreak:
Should this be draft or do you want to get it merged?

The-EG:
> Should this be draft or do you want to get it merged?

I'm still tinkering with it, although this is a good prod to look at it again...

PR #5025, opened 2020-11-26
Port Missions Display to PiGui

Web-eWorks:
@vakhoir if you're out there you can come back now, the worst is over! I took heed and decided not to redesign the missions display while porting it; this drastically reduced the workload involved. I'd say it took me 6-8 hours to do this, all told.

Some before/after screenshots to show off:
image
image
[Not strictly a "before", but funny nonetheless](https://imgur.com/PTWR8Xa)

After:
image
image
image

Need to fix up the translation strings (why is `GO_BACK` in five different module resources and not ui-core?!) and maybe look into right-aligned wrapped text, but otherwise it's good to go, and that's pretty much everything we had in newUI dead and gone.

### DIFF
--- a/data/lang/module-assassination/en.json
+++ b/data/lang/module-assassination/en.json
@@ -179,10 +179,6 @@
     "description": "",
     "message": "It must be done after {target} leaves {spaceport}. Do not miss this opportunity."
   },
-  "LY": {
-    "description": "",
-    "message": "ly"
-  },
   "OK_AGREED": {
     "description": "",
     "message": "Ok, agreed."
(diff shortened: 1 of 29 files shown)


impaktor:
Sweet! If this is the last of NewUI, I'd imagine there would be a lot on C++ side that could be removed? Basically the NewUI bindings into lua?

Web-eWorks:
> Btw, I liked the use of colours in the crew roster tab, I know if one could use it here as well, e.g. status "active" is green, and maybe "days left: xx" also has some stand out colour? I know, we shouldn't expand the scope of this, just throwing it out there.

@impaktor I know; I wanted to keep this PR simple, and while I could have hacked colors into the display I think there are better ways to do it. I'll revisit it when I implement right-aligned wrapped text.

PR #5028, opened 2020-11-27
Update build badge to github actions

impaktor:
Travis badge is dead - long live github actions.

Waning: I think there's something missing to get this to work properly. Probably obvious to @Web-eWorks who is well versed in GA.


impaktor:
Can be generated like so:

PR #5031, opened 2020-11-28
Add IncrementDrag widget

Gliese852:
ezgif-3-0dc78754ddf9

With its help, you can precisely select the amount of something, either by clicking and moving the mouse, or by clicking on the arrows, or by entering from the keyboard.

Also moved the theme import from baseui.lua to wrappers.lua because I needed to use color there.

### DIFF
--- a/src/lua/LuaPiGui.cpp
+++ b/src/lua/LuaPiGui.cpp
@@ -2584,6 +2584,20 @@ static int l_pigui_drag_int_4(lua_State *l)
 	return 5;
 }
 
+static int l_pigui_increment_drag(lua_State *l)
+{
+	std::string label = LuaPull<std::string>(l, 1);
+	int v = LuaPull<int>(l, 2);
+	int v_min = LuaPull<int>(l, 3);
+	int v_max = LuaPull<int>(l, 4);
+	std::string format = LuaPull<std::string>(l, 5);
+
+	PiGui::IncrementDrag(label, v, v_min, v_max, format);
+
+	LuaPush<int>(l, v);
+	return 1;
+}
+
 static int l_pigui_add_convex_poly_filled(lua_State *l)
 {
 	PROFILE_SCOPED()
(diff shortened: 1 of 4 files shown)


data/pigui/baseui.lua
```diff
@@ -3,9 +3,6 @@ local pigui = Engine.pigui
 local ui = require 'pigui.libs.forwarded'
 require 'pigui.libs.wrappers'
 
-
-local defaultTheme = require 'pigui.themes.default'
-
```
Web-eWorks, line 8: Instead of doing this, please move the theme import to before `local ui =`, as the theme itself does not depend on (and should not be dependent on) the `ui` variable. I'd prefer to keep things like the theme in the baseUI file instead of putting it in the wrappers file (where it should not belong).

data/pigui/libs/wrappers.lua
```diff
@@ -744,4 +746,31 @@ function ui.maybeSetTooltip(tooltip)
...
+--   v_max - int, upper bound
+--   format - optional string, format according to snprintf
+--
+-- Returns:
+--
+--   value - the value that the drag was set to
+--
+function ui.incrementDrag(label, value, v_min, v_max, format)
+	return pigui.IncrementDrag(label, value, v_min, v_max, defaultTheme.styleColors.accent_500, format)
+end
```
Web-eWorks, line 774: Instead of passing a hardcoded color variable, please use `ImGui::GetColorU32` in the C++ widget (with the appropriate ImGui style color) for consistency with the rest of the ImGui widgets - this allows the theme to simply push `FrameBgActive = styleColors.accent_500` instead of having to update every call point. This would also allow the widget definition to be moved to `forwarded.lua` instead of in the wrappers file.

src/lua/LuaPiGui.cpp
```diff
@@ -2549,6 +2549,71 @@ static int l_pigui_drag_int_4(lua_State *l)
 	return 5;
 }
 
+static int l_pigui_increment_drag(lua_State *l)
```
Web-eWorks, line 2552: I'd recommend that the actual widget behavior (especially of this length) should go in `src/pigui/Widgets.cpp`, and this file should only be concerned with parameter transport in and out of Lua.

Web-eWorks (changes requested):
Looking good! There are a few sticking points that I'd like to see improved slightly before merge, but overall I like it!

data/pigui/baseui.lua
Gliese852: Done!

data/pigui/libs/wrappers.lua
Gliese852: Done!

src/lua/LuaPiGui.cpp
Gliese852: Done!

Web-eWorks (approved):

Issue #5030, opened 2020-11-28
Sector Map Auto Route Issues


Web-eWorks:
I'd like to preface this by thanking @The-EG for writing the route planner in the first case, and @fluffyfreak for making it *possible* to route a 25,000ly path; it's because of his work that I'm not making this issue about routing a 200ly path.

Go to Sagittarius A* (-3125, 0, 0) and click Auto Route. The game will then freeze for about five minutes as it considers 187,597 stars for at least 17,000,000,000 (seventeen billion!) iterations. Once it unfreezes, you'll be getting about 1-5 FPS as the sector map then tries to draw 7,000+ links.

In short, we have three problems:

**Route planning is not asynchronous**: plotting a 25,000ly course will take a very long time by UI interaction scales. Even if it's only a few seconds, that's far too long to be blocking the main thread for. This will probably be one of the easier problems to resolve, as we can simply create a new SectorCache and ensure the sector generation code is properly thread-safe. Communication back to the UI will be more difficult but not impossible.

**The route planning algorithm is too slow:** Planning a route is an `O(N^2)` operation, which is a problem when your N is 188,000 stars. This is caused by two things:
- The broadphase is not as tight as it should be; it generates far more sectors than it will ever need and does radius tests that ultimately don't even matter. To make matters worse, this particular route is actually the *best* case for the broadphase - a diagonal route causes the broadphase to be **three orders of magnitude** more expensive. The ideal goal here is to switch to the Bresenham line algorithm to cull entire sectors before they're ever generated.
![Untitled Diagram](https://user-images.githubusercontent.com/4218491/100494039-96d6e880-310b-11eb-8c26-9e3f7bf5f148.png)
- We do somewhere between 17,000,000,000 and 35,000,000,000 iterations, even with the best-case broadphase algorithm. Nodes are unordered in the active set, and we iterate through every single star in the "tube" of stars the broadphase selects regardless of where we are in the tube; to make matters worse, only the stars we've picked for a path are removed from the active set, meaning that we're looking at true N-squared complexity cost. The loop also uses actual vector lengths for all comparisons when squared-length comparisons are possible, so we incur the cost of several square roots per iteration.
I have a few ideas of how to fix this, mostly involving writing the active set in a semi-ordered fashion from the broadphase and shrinking the active set as we move along the "tube" towards the target system.
[...]

PR #5033, opened 2020-11-29
Optimise route rendering

fluffyfreak:
Build all of the lines into a single buffer once, render as necessary and offset using the transform.

Resolves part of #5030 rendering

- https://github.com/pioneerspacesim/pioneer/blob/b5e21686269c7efcadcc2301169147e6f7ca04e7/src/SectorView.cpp#L751 this needs some kind of spacial bucketting!
- Added a sector selection optimisation to avoid generating sectors we then discard `Info: SectorView::AutoRoute, nodes to search = 5253, earlied out sector distance from line: 109870 times.`
- Avoided a bunch of sqrt operations by using the square distance.
- Tried to reserve some of the vectors and other stl structures.

### DIFF
--- a/src/MathUtil.h
+++ b/src/MathUtil.h
@@ -24,6 +24,11 @@ namespace MathUtil {
 	{
 		return t * v2 + (F(1.0) - t) * v1;
 	}
+	template <class T, class F>
+	inline T Lerp(const T& v1, const T& v2, const F t)
+	{
+		return mix(v1, v2, t);
+	}
 
 	inline float Dot(const vector3f &a, const vector3f &b) { return a.x * b.x + a.y * b.y + a.z * b.z; }
 
--- a/src/MathUtil.h
+++ b/src/MathUtil.h
@@ -25,7 +25,7 @@ namespace MathUtil {
 		return t * v2 + (F(1.0) - t) * v1;
 	}
 	template <class T, class F>
-	inline T Lerp(const T& v1, const T& v2, const F t)
+	inline T Lerp(const T &v1, const T &v2, const F t)
 	{
 		return mix(v1, v2, t);
 	}
--- a/src/SectorView.cpp
+++ b/src/SectorView.cpp
@@ -345,30 +345,18 @@ void SectorView::Draw3D()
 	trans.Translate(-(m_pos.x) * Sector::SIZE, -(m_pos.y) * Sector::SIZE, -(m_pos.z) * Sector::SIZE);
 
 	// calculate the player's location (it will be interpolated between systems during a hyperjump)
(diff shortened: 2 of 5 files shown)


fluffyfreak:
I'm thinking that the unvisited possible systems is a column/capsule/hot-dog shape thorugh space, so a simple spatial partitioning scheme would just be distance from the origin along central axis of that. 

Starting at zero at the origin bucket and incrementing by whatever the distance metric used. Each node would have the index of the bucket they're in, and they'd only need to consider their own bucket and bucket+1 for systems.

So for each iteration it could add to the unvisited list it's own bucket & bucket+1's sectors, runs the algorithm, and repeats 🤔
I don't think it's that simple because it's always going by system rather than sector, but something along those lines.
Maybe it adds to the unvisited list everytime it passes into a new bucket? 🤷‍♂️ anyway, something like that should reduce it by a few billion iterations to consider at each step.

impaktor:
```
<sturnclaw> fluffyfreak: here's a simpler idea: instead of using a
minimum-cube of sectors in the broadphase, use the Bresenham line algorithm
to generate the sectors in-order along the start->target line (probably an
antialiased variant of Bresenham to get the fuzzy cylinder). Don't do any
line testing, just push the systems onto the vector. Then when you're doing
the narrowphase, keep track of the current < i < i+n}, where n is some number
(e.g. 24) that's the "reasonable upper bound" of the number of systems in the
number of sectors selected in the "width" of the broadphase [03:29:31] that
also advances to the target (e.g. with a dot product) but you can "jump"
through the list of systems that way and process logN on average the same
time complexity and also fit in O(N) space complexity
```

fluffyfreak:
A quick look at Bresenham or Wu lines didn't make it much simpler when you have to have a variable thickness to the volume of the "line" that you want to calculate. Also it wasn't where the cost was, evaluating all of the sectors takes no time, but generating them does so I eliminated most of that.

It can still be improved with a simple incremental step perhaps 🤔but that's not where the cost is now, the narrow phase is bewilderingly expensive.

fluffyfreak:
@Web-eWorks & @The-EG I'm making this into a 2 part piece, this part that does route rendering, and some other minor broadphase optimisations. However it does not touch the narrow phase at all.

Then either someone else can tackle the narrow phase, since there are some good ideas going around, or I will eventually get too it, but my time is limited so I don't want to delay getting in what I've done so far.

Web-eWorks (approved):
Only nitpick I have is that `m_setupRouteLines` is not quite as descriptive as I'd hope (e.g. `m_rebuildRouteLines`), but that's miniscule compared to the rest of the PR. Thanks!

PR #5036, opened 2020-12-01
Add MSVC build to github-actions CI builds

The-EG:
I think I have the appveyor build working well now, but having all of the builds in one place would be nice, and probably simplify getting the releases automated again too.

This is a first shot at adding MSVC to the github actions builds for PRs.


The-EG:
10 pushes later.....

This appears to be working. I had to remove some languages from the installer because they aren't officially supported by Inno Setup at version 6. For appveyor I was downloading the older files for those languages, but I think now that it may be better to drop them. We can always add them back if/when the language files are updated for 6.0.

The-EG:
In addition to the build, the installer is published to the 'Windows-Artifacts' artifact on *all* successful builds with this change.

Web-eWorks:
@The-EG you'll need to change a CPP file to trigger the Github action - once we confirm it's working, I'll be good to merge this!

The-EG:
> @The-EG you'll need to change a CPP file to trigger the Github action - once we confirm it's working, I'll be good to merge this!

I did, that's what the example build I linked was. But I'll trigger it again so it shows up.

PR #5039, opened 2020-12-02
CMake: Fix modelcompiler not compiling anything

pcercuei:
When models are installed on the system, the modelcompiler would use
these instead of the local ones found in the "data" folder. However,
the *.model files are not installed on the system, so the
modelcompiler would end up not compiling anything.

Fix this by returning only the local "data" folder files in
FileSystemPosix.cpp if the PIONEER_LOCAL_DATA_ONLY environment variable
is set, and set this environment variable in CMake when the
modelcompiler is called.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -367,7 +367,8 @@ if (MODELCOMPILER)
 	# This really shouldn't be done inside the source tree...
 	if (NOT MSVC)
 		add_custom_target(build-models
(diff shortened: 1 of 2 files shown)


fluffyfreak:
This looks good to me

Web-eWorks:
I'm not 100% comfortable with querying environment variables here (instead of command line switches), but it works as a good patch until the modelcompiler tool can be fully refactored.

PR #5042, opened 2020-12-03
Use new icons in Worldview

Web-eWorks:
Final feature PR before feature freeze, I promise! This one fought me trying to get everything properly aligned and working.

Now we use Nozmajner's shiny new icons in the worldview and system view...

image


Gliese852:
Great! Only the overview window size scares me a little, it seems to have increased, how will this work at a lower resolutions?

Web-eWorks:
@Gliese852 I tested it at 800x600 and it works just fine; I've made it expand to cover a greater fraction of the screen at smaller resolutions but I consider that to be acceptable behavior as you're going to have it closed most of the time.

nozmajner:
@Web-eWorks The set has the Libra of Justice icon as well, if you still want it somewhere

Web-eWorks:
Going to merge this without the Libra set as it's having some rendering issues with the cloned half of the icon. Will integrate that for the next release!

PR #5047, opened 2020-12-03
Fix menu music issues

### LINKED ISSUE #5041: Мenu music doesn't stop after starting a new game
### Observed behaviour
when you start a new game, the menu music continues to play. If you finish the game and return to the menu, there is complete silence. If you start the game now, the normal soundtrack starts. Now, if you return to the menu, there is always silence, and when you start the game, the usual music. Menu music never plays again. Until the program restarts completely.

60b80bccb Ubuntu

WKFO:
Fix menu music not stopping when starting a game, and not starting again when returning to main menu.
Fixes #5041.

### DIFF
--- a/data/modules/MusicPlayer.lua
+++ b/data/modules/MusicPlayer.lua
@@ -58,6 +58,11 @@ end
 
 -- handle separate planet/station-specific ambient music
 local playAmbient = function ()
+	local current_song = Music.GetSongName()
+	local current_category = MusicPlayer.getCategoryForSong(current_song)
+	if current_category == "menu" then
+		Music.FadeOut(0.5)
+	end
 	local category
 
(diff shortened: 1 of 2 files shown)


Web-eWorks (changes requested):
I'm not sure why you've changed all occurrences of `playRandomSongFromCategory` to pass an explicit `false` parameter - the implicit `nil` parameter should disable the loop. If this is not the case, please fix the functions in `SoundMusic.cpp` to interpret `nil` as "play once" rather than adding an unneeded parameter to all invocations in `MusicPlayer.lua`.

WKFO:
This PR will only fix #5041. See #5049 for what's left.

Web-eWorks (approved):

PR #5048, opened 2020-12-03
Some map fixes

Gliese852:
- fix segfault in the system map if a centered ship is removed
sysmap2

- reset the system map view if the centered ship is hidden
sysmap1

- if the mission target is a specific systembody, when you press the "set as nav target" button, the sector map is set to the correct star of the multiple system
- shrink the searchbar window in the sector map when nothing is shown on it
secmap


Web-eWorks:
> shrink the searchbar window in the sector map when nothing is shown on it

@Gliese852 while it was somewhat intentional to keep it at full length even when there are no search results, I do like the collapsed look better! I think this is a good solution.

PR #5052, opened 2020-12-06
Scottish placename revision

gunchleoc:
I found this in my repo - de-Englishified some placenames.

I know that Gaelic spelling looks weird to English speakers, so feel free to reject. I think we should fix "Abundância" in any case though.


impaktor:
1. I see it has a bünch öf nön-åscii characters (both on master and current PR), question is if we want that? I'm thinking: we assume players have ascii-compatible input on their keyboards, but can't demand that they will be able to type "Göteborg", rather than Gothenburg, when searching for a place (and those last umlauts we're not me being silly this time).

2. Should we use native language's names? So a future "New Denmark" would be København, and not Copenhagen?

These are open question, I'm just throwing them out there.

gunchleoc:
If searchability is a problem, don't merge this. Usability trumps linguistic correctness, as much as I love my correct spellings.

impaktor:
I don't know what the sane thing to do is here, but I don't think we want non-ascii characters. Other than that, I guess it's up to you to re-name the place names in best taste.

gunchleoc:
Gaelic without accented vowels does not work. I could make a separate branch with those names that don't need them.

Web-eWorks:
At this time we do not support (or have dev knowledge and time to implement) fuzzy matching of non-ASCII characters by their non-diacritic forms. Given the differences in unicode encoding between e.g. combining diacritics vs the extended latin charset, this is something that is likely nontrivial. At such time that we have support for searching non-ASCII characters I'd love to merge this in, but I agree with @gunchleoc and @impaktor that usability comes first in this situation.

gunchleoc:
How about adding an optional search keyword list to these names? Then the player could search for either Göteborg or Goteborg to find Göteborg. More configuration work, less programming work.

impaktor:
Acutally, I see we have surface starport in Sol called `'Mariasurīru'`, which is similar to what is discussed above. So we already have this "problem" in master. I'm not sure what the sane thing is here, if we take the extreme case, like Mandarin, that is impossible for us to type or even read out (correctly or incorrectly), so I think all names must be "readable" to an english proficient person, and able to type it into the game using a US keyboard.

Pioneer does support multiple names (at least for the system name, see `data/systems/custom/18_wolf359.lua` for example)

gunchleoc:
> Pioneer does support multiple names (at least for the system name, see data/systems/custom/18_wolf359.lua for example)

Perfect, let's use that!

Web-eWorks:
We support multiple names for systems themselves but not for individual bodies inside of a system, for historic and performance reasons (additional names add overhead and memory fragmentation). However, given that we don't yet have a search feature for bodies inside of a system I think this PR is good to move forward with - I noticed a few tiny issues that will need to be resolved before merge however.

data/systems/custom/07_1orionis.lua
```diff
@@ -9,19 +9,19 @@ local s = CustomSystem:new('1 Orionis',{ 'STAR_F' }) --Alt name will be "Cluaran
...
 	:short_desc('A proudly independent hisoric colony')
-	:long_desc([["Proud World for Proud People". 
+	:long_desc([["Proud World for Proud People".
 1 Orionis' motto can be traced back to year 2325, as the "Evergreen" -one of the legacy colony ships launched before the era of hyperspace travel- landed on planet Saorsa. Its founding fathers vowed to never live again under Terran oppression,
 This heritage still shapes the system’s values to this day : independence, pride, and freedom.
 1 Orionis acively took part in the War of Hope (2723-2725), as its volunteer forces fought alongside Epsilon Eridani against Sol’s attempt of hegemony. The system is a founding member of the Commonwealth of Independent Worlds and remains politically influent to this day.
-In present day, the systems boasts a strong industry and is the birthplace of many an elite pilot of the Confederation Fleet. 
+In present day, the systems boasts a strong industry and is the birthplace of many an elite pilot of the Confederation Fleet.
 Travellers will also find the citizens are a proud lot : be it of their tradition, their chilly climate, their language, their gas giants...]])
```
Web-eWorks: While you're at it, there are a few issues with the long description (missing 'the', em-dashes formatted as hyphens, trailing commas instead of periods, incorrect spacing around colons, "politically influent" -> "influential", "the systems", etc.) that I'd like to see fixed.

data/systems/custom/07_1orionis.lua
```diff
@@ -95,7 +95,7 @@ local orionisd = CustomSystemBody:new('Neart', 'PLANET_TERRESTRIAL') --Means "st
 	:ice_cover(f(5,1000))
 	:life(f(0,1))
 
-local orionisf = CustomSystemBody:new('Iona', 'PLANET_GAS_GIANT') --Real scottish island
+local orionisf = CustomSystemBody:new('Ì', 'PLANET_GAS_GIANT') --Real scottish island
```
Web-eWorks: The rest of the placenames look good, but I'm going to have to ask that this be converted back to `Iona` - not only is the pronunciation non-obvious for those who don't speak Gaelic, but single-character names for bodies have significant negative usability impact (try finding it in worldview!) and breaks homogeneity with the rest of the bodies in the system.

data/systems/custom/07_1orionis.lua
```diff
@@ -131,7 +131,7 @@ local saorsa = {
 		:ice_cover(f(467,1000))
 		:life(f(45,100)),
 	{
-		CustomSystemBody:new('New Edinburgh', 'STARPORT_SURFACE')
+		CustomSystemBody:new('Dùn Éideann Ùr', 'STARPORT_SURFACE')
```
Web-eWorks: I'm of two minds about this one - I would expect New Edinburgh to be a legitimate placename for this system, but I'm also okay with Dun Eideann. I'll leave that up to your discretion.

Web-eWorks (changes requested):

Web-eWorks:
Also for the record, we prefer not to include commits of the type `merge branch master into PR` - please use `git rebase -i master` and force-push to your topic branch as it keeps the history a lot cleaner when the PR is merged!

data/systems/custom/07_1orionis.lua
gunchleoc: I'll use "Hiort" instead - also an island, longer and no accents :)

data/systems/custom/07_1orionis.lua
gunchleoc: That depends on whether we want to make this a multilingual Scots + Gaelic society, or a monolingual Gaelic society. I'd be fine with either and could research the Scots names for the lowlands-based place names.

gunchleoc:
> Also for the record, we prefer not to include commits of the type `merge branch master into PR` - please use `git rebase -i master` and force-push to your topic branch as it keeps the history a lot cleaner when the PR is merged!

Most of the time, rebasing means that I have to fix the same merge conflict over and over and over again - can you please squash the commits when merging? That's just 1 mouse click and a lot easier to do? This branch will get multiple commits now anyway. Thanks!

data/systems/custom/07_1orionis.lua
gunchleoc: What's the policy for quotes - straight or typographical?

Web-eWorks: At your discretion - for system definitions typographical quotes are definitely an option, though it's more a matter of how much effort you'd like to put in.

data/systems/custom/07_1orionis.lua
Web-eWorks: I think multilingual works a bit better as we don't have a system that's entirely lowland Scots. I'd imagine that even a New Scotland system would be a bit of a cultural melting pot for the various regional Scots languages.

gunchleoc:
This is ready for review again. I have made orionisc Norn and orionisi Scots.

I have also changed the cultural values to something less Braveharty and more in line with the actual cultural values as I perceive them.

thedarklordkitten:
I would just like to point out this reference I linked to in the forum:

https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=638&start=10#p6567

If I get it right, one could have non-ascii character that can be searched for with the "base" ascii corresponding character.
I really wanted to get my hands "dirty" and offer a pull request for this approach, but I needed more time to get acquainted with code and language and I don't have anything (yet). Mostly, I was kind of afraid to rewrap the general search function with this replacement algorithm.
I don't know if the suggestion above can be of any use to more expert coders out there :)

Web-eWorks (approved):
Looking good, thank you for the thoughtful approach to the naming convention!

Issue #5045, opened 2020-12-03
"Earth radius" word order is hard-coded


gunchleoc:
### Observed behaviour

When looking at planet information, I get English word order for "Earth radius" etc.

### Expected behaviour

Ability to translate this with a different word order, like "radius of the Earth". I'd like to put "Earth into the genitive too.

So `<number> + "An Talamh" + "rèideas"  + "(<number>)"` should be encoded as `"{number1} de rèideas na Talmhainn ({number2})`"

English:
 `<number> + "Earth" + "radius"  + "(<number>)"` ->`"{number1} Earth radii ({number2})`"

### Steps to reproduce

Go to the planet info and click on any planet

![Screenshot at 2020-12-03 09-02-50](https://user-images.githubusercontent.com/4095570/100988339-88dbfa00-3547-11eb-8981-760d9435586e.png)

__My pioneer version (and OS):__

cc490837e0c4403ef91cdbfb350d5c0a4df3f88a Linux Mint (Ubuntu 20.04 based), self-compiled release build + `tx pull -l gd`.

gunchleoc:
I'm doing this one myself :)

PR #5053, opened 2020-12-06
Fix grammar in translations for N_WHATEVER* strings

gunchleoc:
Fixes https://github.com/pioneerspacesim/pioneer/issues/5045

Turns out that the word order was not hard-coded but rather a translation mistake. I still need the genitive case though. German also needs this.

| master (wrong) | branch (correct) |
| -------------------- | --------------------- |
| tomadan An Talamh | de thomad na Talmhainn |
| Die Erde Massen | Erdmassen |

Translators can probably hack it with the current implementation, and the German translator did so, but it's very error-prone.

@impaktor


impaktor:
Yeah, we probably want this, as yet another step to untangle assumptions that work in english, but not in other languages. 

Can you squash the last commit into the first one? (If not, we can do it when merging)

gunchleoc:
I don't know how to squash commits without the GitHub button, so please do that.

Web-eWorks:
Saw on IRC that you were waiting on my feedback @impaktor - this one looks good to me!

PR #5059, opened 2020-12-06
BBS advert for soldout commodity

impaktor:
## Intro
In frontier one would often have the habit to check the BBS for special "WANTED" adverts before selling the commodity on the market exchange. Especially if the station had 0t in stock.
2020-12-06-164844_1600x900_scrot
2020-12-06-164855_1600x900_scrot

## About this Pioneer BBS advert
This module selects a major import and with a fairly low probability (for now 10% adds one advert, and zeros out the stock of that commodity in the commodity market).

There's only one flavor because this advert will be rare, so slight rewording of it will not make a difference.

This is in some way similar to #2753, but that one announces where the market is sold out globally, to all neighboring systems, (and offers a larger multiple than x2), where as this advert is as brief as can be.

Advert uses "$" icon for now (could alternatively use the shopping cart icon)
2020-12-06-213818_1600x900_scrot
2020-12-06-213831_1600x900_scrot

## To-Do
Please don't look at the code, it's messy

  - [x] Don't allow selling more than I have
  - [x] Is it normal for other commodities to also be zeroed out? Looks like a bug, zeroing out stock of other major import commodities.
  - [x] Sanitize code & commits.
  - [x] Put strings into language module
  - [ ] Remove debug stuff (e.g. constant spawn)

### DIFF
--- a/data/lang/module-soldout/en.json
+++ b/data/lang/module-soldout/en.json
@@ -1,0 +1,30 @@
+{
+  "AMOUNT": {
+    "description": "The amount (for now in tonne) of goods person wants to buy",
+    "message": "I'll buy anything you have, up to the full {amount}t that I need."
+  },
+  "MESSAGE": {
+    "description": "",
+    "message": "I'm {name}. I need {commodity} urgently, and I'll pay twice the market price ({price})."
+  },
+  "NOT_IN_STOCK": {
+    "description": "Regarding selling more than the player has",
+    "message": "Not enough {commodity} in stock."
+  },
+  "SELL_1": {
+    "description": "Sell one single unit (probably tonne)",
+    "message": "Sell 1"
+  },
+  "SELL_N": {
+    "description": "Sell several units (probably in tonne), i.e. N > 1",
+    "message": "Sell {N}"
+  },
+  "SELL_ALL": {
+    "description": "Sell everything player has of the good/commodity",
+    "message": "Sell all"
+  },
+  "TITLE": {
+    "description": "",
+    "message": "WANTED: {commodity}. Will pay {price} per tonne"
+  }
+}
--- a/.tx/config
+++ b/.tx/config
@@ -111,6 +111,11 @@ file_filter = data/lang/module-secondhand/<lang>.json
 source_file = data/lang/module-secondhand/en.json
 source_lang = en
 
+[pioneer.module-soldout]
+file_filter = data/lang/module-soldout/<lang>.json
+source_file = data/lang/module-soldout/en.json
+source_lang = en
+
 [pioneer.module-stationrefuelling]
 file_filter = data/lang/module-stationrefuelling/<lang>.json
 source_file = data/lang/module-stationrefuelling/en.json
--- a/data/modules/SoldOut/SoldOut.lua
+++ b/data/modules/SoldOut/SoldOut.lua
@@ -1,0 +1,157 @@
+-- Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local Engine = require 'Engine'
+local Lang = require 'Lang'
+local Game = require 'Game'
+local Event = require 'Event'
+local Serializer = require 'Serializer'
+local Character = require 'Character'
+local Format = require 'Format'
+local Equipment = require 'Equipment'
+
+local l = Lang.GetResource("module-soldout")
+
+local ads = {}
+
+local onChat = function (form, ref, option)
+	local ad = ads[ref]
+	-- Maximum amount the player can sell:
+	local max = Game.player:CountEquip(ad.commodity, "cargo")
+	max = ad.amount < max and ad.amount or max
+
+	form:Clear()
+	form:SetTitle(ad.title)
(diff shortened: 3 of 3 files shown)


impaktor:
By the way, in Frontier, when you jumped into a system, you could query each station for list of commodity prices (for a fee of $5, I think), through Comms. Would be cool if we could do the same, and also get stock, so then you might guess which station has one of these adverts, if they have 0t. Just throwing it out there.

nozmajner:
If there's light lag, then it would be odd for you to get the info instantaneously though. Although I guess it would be doable if the info is broadcasted, and that cryptocurrency thing is what the money is. You'd get it instantly, the price deducted, but the recipient will get the money when the signal arrives.

impaktor:
OK, except for debug outputs that should be removed/commented out, this PR is done.

I've re-written it now, so the module knows nothing about what a system imports anymore, it only check for 0 stocked commodities on each station and with 50% chance spawns a "wanted will pay double" advert on the BBS.

I ran this in Sol, with the changes suggested in #5074, (with the much lowered probability to be sold out of imported goods)

The Spaceport.lua reports it made 0 stock for these stations:

| stock | commodity        | price | station          | type         | got an advert in the end |
|-------|------------------|-------|------------------|--------------|--------------------------|
|     0 | Mining machinery |   312 | Mexico City      | MAJOR import | x                        |
|     0 | Carbon ore       |     7 | London           | minor import | x                        |
|     0 | Animal meat      |   125 | Jobs Pad         | minor import | x                        |
|     0 | Plastics         |    36 | Tomm's Sanctuary | MAJOR import |                          |
|     0 | Metal alloys     |    43 | Tomm's Sanctuary | MAJOR import | x                        |
|     0 | Metal alloys     |    43 | Clarke's Station | MAJOR import | x                        |
|     0 | Animal meat      |   125 | Discovery Base   | minor import | x                        |

which resulted in a single advert in each of the stations listed above, including Tomm's (i.e. even though it had sold out on two commodities)

(Now I hope the random number generator, `Engine.rand:Number(0, 1)` isn't in the same repeated state for each BBS station, as acceptance rate seems a bit high for 50% probability)

Web-eWorks:
Two thoughts here: first of all, code doesn't look too terrible, maybe a few more comments needed to explain in the absence of debug messages but otherwise good.

Secondly, I would suggest dropping the `1000` option and setting an explicit maximum number of accepted commodities on this advertisement - I highly doubt this person would be paying double the going price on (especially) expensive commodities with a completely open ended commodity number. I'd rather expect they need a certain number to finish an urgent job and are thus willing to pay a significant incentive to ensure they receive their needed number of commodities. (Also it helps to keep the ad from being too exploitable; this is a 50% profit margin, which is *significantly* better than the best interstellar profit margin by at least 20%.)

impaktor:
> I would suggest dropping the 1000 option and setting an explicit maximum number of accepted commodities 

I've had the same thought, actually. I might not get around to it until next week though.

impaktor:
> I might not get around to it until next week though.

meme

# Changes:
Now adverts have a fix amount they offer to buy. It's determined by the sum of the value of the goods they want to buy, is drawn from a random distribution that is uniform in log-space, as natural numbers typically are, example below:
Reciprocal_Histogram svg
Now, these numbers in the figure above span 0-100, however, mine are 500-10^5 in value (I was too lazy to make my own plot, above is from wikipedia).

# Result
2020-12-18-205532_1600x900_scrot

# Notes
- I thought I saw a nice way to consolidate the last two conditional statements in `OnChat`, however, I kept getting an edge case that kept popping up someplace new in the code once I fixed it at some point
- the "Sell all" option sells everything the player has in stock, or that the buyer wants. Currently, if the player sells the full amount the buyer wants, the advert text is simply changed to "bla bla bla I want 0t", so then the player must press "Hang up" and the advert is removed. Perhaps there should be a "thank you, now I got all {commodity name} I need"

data/lang/module-soldout/en.json
```diff
@@ -0,0 +1,30 @@
+{
+  "AMOUNT": {
+    "description": "The amount (for now in tonne) of goods person wants to buy",
+    "message": "I'll buy anything you have, up to the full {amount}t that I need."
+  },
+  "MESSAGE": {
+    "description": "",
+    "message": "I'm {name}. I need {commodity} urgently, and I'll pay twice the market price ({price})"
```
Web-eWorks, line 8: Missing trailing period.

data/lang/module-soldout/en.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "NOT_IN_STOCK": {
+    "description": "Regarding selling more than the player has",
+    "message": "Not enough in stock"
```
Web-eWorks, line 12: I'd recommend adding a `{commodity}` in here, as well as adding punctuation to the end.

Web-eWorks (approved):
Code looks very clean! I have a few extremely minor nitpicks, but otherwise this is good to go.

impaktor:
@Web-eWorks thanks

Btw, I could point out that this module does not, currently, have the ability to add several adverts for the same commodity, that could be done as a next step, offering different prices, but I think this will do for now.

I'll add this to transifex now as well (just a reminder to myself)

PR #5060, opened 2020-12-08
Show surface starports in the system map

### LINKED ISSUE #5058: System viev: Destinations on ground are not marked
### Observed behaviour

Set a starport on ground as destination, switch to system view: no destination is marked
(but works if the body is the destination)

### Expected behaviour

Same as if an orbital starport is the destination

20201206 on SuSE Linux

Gliese852:
surf-starport
Fixes #5058

### DIFF
--- a/src/SystemView.cpp
+++ b/src/SystemView.cpp
@@ -360,6 +360,26 @@ void SystemView::PutOrbit(Projectable::bases base, RefType *ref, const Orbit *or
 	}
 }
 
+// returns the position of the ground spaceport relative to the center of the planet at the specified time
+static vector3d position_of_surface_starport_relative_to_parent(const SystemBody *starport, double time)
+{
+	const SystemBody *parent = starport->GetParent();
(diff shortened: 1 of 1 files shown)


fluffyfreak:
@Gliese852 well done! There needs to be a little applause emoji 👏oh there is! 😂

Gliese852:
@fluffyfreak this is very nice!

Gliese852:
Corrected something, namely, determining the distance to the center of the planet. If the current system is viewed, the height is taken from the physical body

Bodasey:
Many many thanks, tested this function a bit, can be played as used to, no new bugs found.

impaktor:
People have started playing with this in IRC, so I say this is play/bug-tested for merge. @fluffyfreak you want to glance over the code (32 lines) before we merge?

fluffyfreak (approved):
All looks good to me

Web-eWorks (approved):

PR #5061, opened 2020-12-08
Fix ambient music looping

### LINKED ISSUE #5049: Music loops unintentionally
I intended to fix this in #5047 but I don't have much time to work on it (also MSVC broke), so please fix before release otherwise it will annoy everyone who plays with music on.

### Observed behaviour
All music loops. (ambient space, spacestation etc.) Only the menu music is supposed to.

### Expected behaviour
Don't loop music.

### Steps to reproduce
0) Make sure music is not muted.
1) Start game.

c606f28b56ace4f87ce7174b8253db02a4bc65b3 (Linux Mint 20 Cinnamon x64)

WKFO:
I haven't got the slightest idea why no argument to loop isn't treated as such, but adding a default seems to fix #5049 
It looks like otherwise the related code in SoundMusic.cpp gets a repeat = true (line 53 I think).

¯\_(ツ)_/¯

### DIFF
--- a/src/lua/LuaMusic.cpp
+++ b/src/lua/LuaMusic.cpp
@@ -57,7 +57,7 @@ static int l_music_get_song(lua_State *l)
 /*
  * Method: Play
  *
- * Starts playing a song instantly, on repeat by default.
+ * Starts playing a song instantly.
  *
  * Example:
  *
@@ -66,7 +66,7 @@ static int l_music_get_song(lua_State *l)
  * Parameters:
  *
  *   name - song file name, without data/music/ or file extension
- *   repeat - true or false, default true
+ *   repeat - true or false, default false
  *
  * Availability:
  *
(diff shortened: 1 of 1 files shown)


Web-eWorks:
@WKFO the relevant code is in `src/lua/LuaMusic.cpp` - it uses the pattern
```cpp
bool repeat = true;
if (lua_isboolean(l, 2))
    repeat = lua_toboolean(l, 2) != 0;
```

Which should instead be `bool repeat = LuaPull<bool>(l, 2, false);`, that is, default to a `false` value if no argument is present. Changing all instances of the above resolves the root cause of the issue and also makes the code cleaner!

WKFO:
@Web-eWorks saves me again. Thank you, and I pushed the change.

src/lua/LuaMusic.cpp
```diff
@@ -79,7 +79,7 @@ static int l_music_get_song(lua_State *l)
 static int l_music_play(lua_State *l)
 {
 	const std::string song(luaL_checkstring(l, 1));
-	bool repeat = true;
+	bool repeat = LuaPull<bool>(l, 2, false);
 	if (lua_isboolean(l, 2))
 		repeat = lua_toboolean(l, 2) != 0;
```
Web-eWorks, line 84: This code is now redundant and can be removed... :smile: LuaPull handles pulling the boolean value from Lua.

Web-eWorks (changes requested):

Web-eWorks (approved):

PR #5063, opened 2020-12-09
Localize header strings on ship information screen

gunchleoc:
- Localize table headers
- Fix ragged layout by using text instead of textWrapped. The textWrapped function really needs a width parameter, but we have sufficient space on this screen at 800x600 resolution and I didn't want to dig this deep into it.

Tested with Scottish Gaelic and Hungarian - in my experience, if it fits these 2 languages, it fits.

Before:
pioneer-ship-information-alignment

After:
pioneer-ship-information-alignment-after

### DIFF
--- a/data/pigui/libs/text-table.lua
+++ b/data/pigui/libs/text-table.lua
@@ -18,7 +18,7 @@ function textTable.draw(t)
 			local width, textWidth = ui.getColumnWidth(), ui.calcTextSize(entry[2]).x + 1
 			-- FIXME: add support for ui.getStyleVar("ItemSpacing"). Hardcoding to 8 for now
 			ui.sameLine(math.max(width - textWidth, width / 2 + 8), 0)
-			ui.text(entry[2])
+			ui.textWrapped(entry[2])
 		elseif type(entry) == "string" then
 			ui.spacing()
 			ui.text(entry)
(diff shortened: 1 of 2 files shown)


nozmajner:
"Tested with Scottish Gaelic and Hungarian - in my experience, if it fits these 2 languages, it fits" As a Hungarian, I love this sentence.

Gliese852:
It seems to fix #5060

gunchleoc:
No, it doesn't - it only circumvents the bug for this particular screen.

> As a Hungarian, I love this sentence.

Well, it's true! As a general rule, Gaelic needs more space, but there are some cases where Hungarian is long and Gaelic is short. And Hungarians rock anyway -> hunspell :)

Web-eWorks:
The `textWrapped` function call was added to specifically support the display of mission details on the mission info screen, which rely heavily on tables wrapping text. I definitely agree that the wrapping failure is a problem, but it seems to be endemic to ImGui's text wrapping calculations at our font size and screen resolution.

I would be amenable to a solution that adds a separate table function that draws unwrapped text, but unfortunately the proposed PR does in fact break other display elements in a nontrivial way.

Good catch on those strings though! I occasionally change my language from English and check, but there are so few languages with 100% complete translations it's difficult to tell what's incorrectly hardcoded from what's simply untranslated as of yet.

gunchleoc:
I have undone the textWrapped change and rebased master. I still recommend squashing the commits when merging though.

Web-eWorks:
Thank you, merging!

PR #5068, opened 2020-12-10
Display Background in SystemView

Web-eWorks:
Draw the universe skybox in the SystemView to make it seem less empty.

image

Ideally I'd like to composite a computerized / scanner effect over the background when looking at systems that the player isn't currently in, but I also don't feel like rewriting the entire material system today. For now, I've chosen not to render stars in other systems because we're "stealing" the current system's background object to render in the SystemView.

@Gliese852 brings up a good point on IRC that the plane of the galaxy is aligned with the plane of the ecliptic; as far as I can tell this is universal for all systems in the WorldView and not related to the SystemView's use of the background. I'm choosing to ignore that particular bug and hoping that merging this will motivate someone to fix it.

I've also moved the TransferPlanner to its own file as it doesn't depend on the SystemView and doing so reduces the line count and cognitive complexity of SystemView.cpp dramatically.


impaktor:
Nice!

src/SystemView.cpp
```diff
@@ -509,6 +319,31 @@ void SystemView::Draw3D()
...
+	} else {
+		background->Draw(trans2bg);
+	}
+
+	m_renderer->ClearDepthBuffer();
+
+	// We need to adjust the "far" cutoff plane, so that at high magnifications you can see
+	// distant objects in the background.
+	m_renderer->SetPerspectiveProjection(CAMERA_FOV, m_renderer->GetDisplayAspect(), 1.f, 1000.f * m_zoom * float(AU) + DEFAULT_VIEW_DISTANCE * 2);
+	//TODO add reserve
```
fluffyfreak, line 345: Add reserve?

fluffyfreak (approved):
Only that one question about the TODO comment, otherwise all looks good to me!

PR #5073, opened 2020-12-12
Donate to crank improvements

impaktor:
# Fix: face seed

New pigui chat-form did not support explicitly setting seed, giving new face every time the advert is clicked. Fixed it by having the same charracter for each instance of a BBS. (These adverts are not fix to station over time).

# Fix: crank talking back

The comms that was used before, is rendered to world view, so messages (e.g. "note enough money", or "thanks") sent there are not seen. Update message in chat box instead, making it more of a dialogue. Also, now the advert closes itself when the player has donated.

untitled

# New: Indicate to user what amount to donate

There is currently no way the player to know:

1. How close they are to next jump in reputation status, nor,

2. Difference between different donation organizations (some are more cranks than others)

With this, each donation indicates how much money is needed to gain next level of reputation. Thus, it still obscures directly knowing reputation points and the "force multiplier" factor for the particular donation cause on reputation.

Here is how much is needed to go from current reputation level ("nobody") to next ("inexperienced"):

2020-12-12-200038_1600x900_scrot

### DIFF
--- a/data/modules/DonateToCranks/DonateToCranks.lua
+++ b/data/modules/DonateToCranks/DonateToCranks.lua
@@ -4,7 +4,6 @@
 local Engine = require 'Engine'
 local Lang = require 'Lang'
 local Game = require 'Game'
-local Comms = require 'Comms'
 local Character = require 'Character'
 local Event = require 'Event'
 local Serializer = require 'Serializer'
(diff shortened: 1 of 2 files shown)


data/modules/DonateToCranks/DonateToCranks.lua
```diff
@@ -87,7 +107,7 @@ local onCreateBB = function (station)
 		title    = flavours[n].title,
 		message  = flavours[n].message,
 		station  = station,
-		faceseed = Engine.rand:Integer()
+		charcter = Character.New({armour=false}),
```
Web-eWorks, line 110: `charcter` -> `character`

data/modules/DonateToCranks/DonateToCranks.lua
```diff
@@ -33,46 +34,65 @@ local addReputation = function (money)
...
-		form:AddOption(Format.Money(100000,false), 100000)
 
-		return
-	end
+local onChat = function (form, ref, option)
+	local ad = ads[ref]
+	form:Clear()
+
+	form:SetTitle(ad.title)
+	form:SetFace(ad.charcter)
```
Web-eWorks, line 68: `charcter` -> `character`.
I would prefer to restore deterministic seeds for face generation, but for now this is an acceptable compromise.

Web-eWorks (changes requested):

data/modules/DonateToCranks/DonateToCranks.lua
impaktor: Yes, me too, especially since the Character class documents the functionality, I think. Couldn't figure it out though. Not sure what the best route would be? Overloading the `ChatForm:SetFace(character)` to also take a table with other variables? If you have time at some point we can discuss it

data/modules/DonateToCranks/DonateToCranks.lua
impaktor: fixed.

Web-eWorks (approved):

impaktor:
I was contemplating removing the last commit, as @nozmajner and @Gliese852 didn't like it in its current form (or at all). I completely agree that we should not expose the underlying numbers to the player, (thus reputation isn't shown as a number but as "levels", like "Inexperienced"...), but I know I've never donated a dime in the game, because it's very unclear what it does, if it's price worthy at all, so this was an attempt to fix it, by having the charity suggest how much they think you aught to donate.
(maybe handwave: "we'll consider this a nice donation for someone of your stature, and put the word out of your generosity", compare: we expect Bill Gates to donate more than a poor person).

I'm hoping maybe it could be done in a different/better way, but fundamentally, I think it's a problem with DonateToCranks, that the player doesn't really know what they're buying, thus they don't know how their actions affect the game world, which in my understanding, is bad game design.

@fluffyfreak, as a game designer, do you have ideas / thoughts?

fluffyfreak:
Yes the DonateToCranks has always been opaque. I've never actually done it and don't know why I would.
So the general idea of improvements is good.

Usually with charities and wealthy donators they'll change the amount asked for directly. So a leaflet throught he door is £3 a month, or a chugger (Charity Mugger) in the street will be asking for £1 or spare change, but when they ask Bill Gates they'll be saying "_we need $100 million for XYZ cause_"

impaktor:
One thought, to "hide" it again, is to have the maximum amount you can donate, to be what is needed to "be sure we'll spread the word of your charitable nature", but that also has problems, since we're again making it opaque, and might also look funny in situations where maximum donation amount would be $10, or $100.

PR #5077, opened 2020-12-15
Fix SIGSEGV on switch to system map after HJ

Gliese852:
If you turn on the ships in the system map, turn off the system map and jump into another system, then pointers to the ships from the previous system remain in `m_projected`, which cause the a segfault. Now clear this when switching from view.

### DIFF
--- a/src/SystemView.h
+++ b/src/SystemView.h
@@ -80,10 +80,10 @@ struct Projectable {
 class SystemView : public PiGuiView, public DeleteEmitter {
 public:
 	SystemView(Game *game);
-	virtual ~SystemView();
-	virtual void Update();
-	virtual void Draw3D();
-	// virtual void OnSwitchTo() { Update(); Draw3D(); }
+	~SystemView() override;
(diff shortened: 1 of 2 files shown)


src/SystemView.h
```diff
@@ -80,10 +80,10 @@ struct Projectable {
...
 public:
 	SystemView(Game *game);
-	virtual ~SystemView();
-	virtual void Update();
-	virtual void Draw3D();
-	// virtual void OnSwitchTo() { Update(); Draw3D(); }
+	~SystemView() override;
+	void Update() override;
+	void Draw3D() override;
+	void OnSwitchFrom() override { m_projected.clear(); } // because ships from the previous system may remain after last update
```
Web-eWorks, line 86: I would recommend this go in the .cpp file instead of the header just to keep things clean. SystemView.h is already pretty big.

Gliese852: Done!

Web-eWorks (approved):

PR #5078, opened 2020-12-15
Trigger Build Pioneer github action on pushes to master

The-EG:
This should cause the GitHub actions builds to trigger on all pushes to master, which is how the appveyor and travis builds worked, in addition to any PRs that have cpp/h source changes.

This should get our build badge working again, *and* have the nice side effect of having an unofficial build of master always available: the latest artifacts will be available here -> https://github.com/pioneerspacesim/pioneer/actions?query=workflow%3A%22Build+Pioneer%22+branch%3Amaster

Also, I've turned the 'installing' messages from CMake back on. Since Github actions has separate sections for each step, and creating the installer is one of those steps, the install messages shouldn't be an issue with reviewing the logs. And, it's helpful to see if things are being included/excluded from the installer.

Finally, I also broke the windows installer when I upgraded the windows libraries in pioneer-thirdparty, that is also fixed here.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -4,6 +4,9 @@ name: Build Pioneer
 
 # Controls when the action will run.
 on:
+  push:
+    branches:
+    - master
   pull_request:
     paths:
     - 'src/**.cpp'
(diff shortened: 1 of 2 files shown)


impaktor:
I assume there's not much more to review than to merge, and see if it works?

The-EG:
> I assume there's not much more to review than to merge, and see if it works?

Indeed...so, lets see if it works.

PR #5081, opened 2020-12-19
Port Planar Radar to PiGui

### LINKED ISSUE #5064: Prograde-retrograde vectors shown on screen are calculated in a wrong way
### Observed behaviour
Prograde and retrograde "vectors" behave as if they were point objects in space. Most noticable in external view, but also noticable in internal view (not necessarily, but especially) if you are flying a ship with an offset cockpit, like Sinonatrix.
https://www.youtube.com/watch?v=L40XAnLg8io

### Expected behaviour
Always render velocity and retrograde vectors as if they were at an infinite distance from the ship. (i.e. never see retrograde and prograde vectors at the same time on the same view.)

### Steps to reproduce
1. Fly on external view or at slow speeds in internal view

c5e8502ddf7879f67153d159636d59bac3f6e635 (Win 10 x64)

Web-eWorks:
Fixes #5064. Port the 3d planar radar to pigui and remove all references to the ShipCpanel. This PR also ports the NavTunnels widget to pigui as well.

The radar plane's colors are not currently themeable, as that's a much larger can of worms that I don't want to get into. But overall I think it's a general improvement over the existing radar....

https://imgur.com/5yiht9o


impaktor:
Yay! Also, I assume @The-EG needs to do some magic for visual studio to compensate for the files that have been removed

Web-eWorks:
It's more that I should learn to compile my PRs before committing - I forgot to recompile when I finally removed the ShipCPanel files.

The-EG:
> Yay! Also, I assume @The-EG needs to do some magic for visual studio to compensate for the files that have been removed

Actually, @impaktor is correct. Any time a .cpp file is added, removed or even renamed it breaks the VS solution.

fluffyfreak:
> > Yay! Also, I assume @The-EG needs to do some magic for visual studio to compensate for the files that have been removed
> 
> Actually, @impaktor is correct. Any time a .cpp file is added, removed or even renamed it breaks the VS solution.

This is the joy of VS

impaktor:
Btw, you mentioned choosing the colours for ships/objects in the radar for this PR, but does that mean that the azimuthal radar uses different colours for the same objects? (I can't test this myself right now)

Web-eWorks:
@impaktor I actually chose colors for radar contacts in the Theme PR that was merged a while back, but the C++ planar radar widget historically used hardcoded colors instead of colors defined in Lua. Now, all contact drawing is handled by Lua code, and thus both radars share the exact same contact colors.

I'm going to go ahead with merging this if no one has any issues or further reviews to add... I might have one more PR coming to fix the worst of the UI resolution issues and a few other inconsistencies but I think this is about everything I wanted to get done for release.

PR #5082, opened 2020-12-19
Show import/export in commodity description in commodity market

impaktor:
This should maybe be shown in a separate column (but green/blue icons would look strange here?), and could be placed better,
but it's a start.

2020-12-19-103834_1600x900_scrot

## Open questions 
It feels like the if-else statement to check if something is import/export keeps popping up everywhere (tradeship, spacestion, here, system econ view). Feels like there should be a method that gets that information for us somewhere? 

At least which numbers for price modification should be the threshold for major/minor export import? 
- system econ view sets 4/10 for minor/major
- Tradeship lua: 2/10 for minor/major 
- Spacestation: 2/10 for minor/major
- commodity market (this PR): 2/10


nozmajner:
I think icons wouldn't look much out of place.
market_icons

impaktor:
@nozmajner, yeah, that could work. I assume we still want the text description as in the original (and your) screenshot. Both belt and suspenders!

Web-eWorks (approved):

PR #5083, opened 2020-12-19
Make use of system's alternate names, as original author intended

impaktor:
Following in the wake of #5052, I saw that @SpifforOmnivore has left comments in the system definition what the system's names should be once we have alternative naming scheme in place, which we do, since some time ago. This aims to do the wishes of original author. 

@SpifforOmnivore please have a look and see if it's OK with you.

Also, Cennet is turkish? And Azeri is a language? (ping @WKFO)


WKFO:
@impaktor "Cennet" means "paradise" in Turkish and Azeri. Azeri is just Azerbaijan Turkish, it's very close to Turkey Turkish.

impaktor:
Makes sense to me, I've put it as a separate commit if we want to remove it after all.
(to be squashed otherwise)

Web-eWorks (approved):

impaktor:
btw, interestingly enough, there's no custom definition for Alpha Centauri, so that's up for grabs if anyone want to write some story for it and design its layout.

Since I have the screen shot:
2020-12-19-141822_1600x900_scrot

impaktor:
Zooming in, the other system is Proxima Centauri

2020-12-19-141814_1600x900_scrot

PR #5093, opened 2020-12-25
Implement map view music, improve music selection, add tracks

WKFO:
Implements map view music and adds a few tracks, just like the title suggests.

Map view music is triggered if the already playing track ends while the player is in a map view (sector or system), but fades out and switches back to ambient once player returns to world (flight) view. (This way it doesn't trigger for every short glance at the map, but it's supposed to be playing if you are taking your sweet time admiring the galaxy/plotting routes into hell).

map-core category plays if player is reasonably close to the populated area of the galaxy (starclock.ogg, it's sort of techno-electronica-whatever), map-unexplored plays while the player is far far away from civilization (no music, minimal ambient sounds to fit the feeling of the unexplored ambient tracks).

Finally, also adding 2 tracks to discovery category which I wanted to add for a while (I didn't think those two were worth a PR of their own, so I'm attaching them here.)


WKFO:
A few new changes:

- Combat music now loops as long as the player is in combat.
- Frame change no longer stops combat music either.
- ship-destroyed category is disabled, because it was annoying more than anything. I simply commented out the code so if someone thinks it has been a mistake after this PR is merged, it can be simply uncommented back.
- README.music.txt has been neglected for a while. That has been updated.

WKFO:
Also using this opportunity to add another track.

ConfedMajor:
I would add a track for thoose times that the player stays close or the player orbits a star

WKFO:
I might look into something along those lines, but not for now. It would require some effort I can't provide soon enough.

WKFO:
My initial reservations about a map music in unexplored space has been solved and I came to the conclusion that if there is a difference between world view and map view in populated space, there should be a distinction in unexplored space as well.

WKFO:
I made a slight rearrangement; moved the old, more energetic unexplored map music to map-core, and instead added another, calmer track to map-unexplored.

Web-eWorks:
@WKFO any further changes you'd like to make to this PR, or is it ready for merge? Sorry it's taken so long to get around to merging again!

WKFO:
@Web-eWorks I have nothing planned, it should be ready for merge if you don't see any problems.

PR #5094, opened 2020-12-26
Fix Legal footer always being "clean" in station view

### LINKED ISSUE #2863: Legal status doesn't work.
screenshot-20140412-195440
Legal status is still clean, after just murdering someone right above a starport.

After killing someone I escaped the system via hyperjump. In the next system the police still wanted to kill me. It was quite an exciting landing, with 6 police fighters trying to shoot me dead. I think there should be an easier way to be able to pay the fine...

impaktor:
Me, on April 12th, 2014:

> Yes, I've been meaning to look into this just this weekend.

Well, here it is, fixes #2863 (I'm also including a spelling fix commit)

## Demonstrating this fix below:

2020-12-26-141022_1600x900_scrot
2020-12-26-141112_1600x900_scrot

### DIFF
--- a/data/libs/Player.lua
+++ b/data/libs/Player.lua
@@ -282,6 +282,35 @@ function Player:ClearCrimeRecordHistory (faction)
 end
 
 --
+-- Method: GetLegalStatus
+--
+-- > local legal_status = Game.player:GetLegalStatus(faction)
+--
+-- Parameters:
+--
+--   faction - optional argument, defaults to the faction player is in.
+--
+--   legal_status - one of strings "CLEAN", "OFFENDER", "CRIMINAL", "OUTLAW",
+--	                or "FUGITIVE", mapping to translated strings in lang/ui-core
+--
+function Player:GetLegalStatus (faction)
+	local crimes, fine = self:GetCrimeOutstanding()
+	self:setprop("fine", fine)
+
+	if fine < 1 then
+		return('CLEAN')
+	elseif fine < 5500 then
+		return('OFFENDER')
+	elseif fine < 20000 then
+		return('CRIMINAL')
+	elseif fine < 100000 then
+		return('OUTLAW')
+	else
+		return('FUGITIVE')
+	end
+end
+
+--
 -- Method: GetMoney
 --
 -- Get the player's current money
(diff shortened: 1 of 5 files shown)


craigo-:
Wow, the police take defacing their images pretty seriously 😄

impaktor:
Some extra words:

1. Ideally, something should happen as the player's legal status deteriorates, e.g. bounty hunters start hunting the player. It's open for future contributions.

2. Word of caution, the `setprop("fine", fine)` is half guess-work by me, perhaps it should be placed elsewhere, when the player is given a fine? I know @laarmen tried to explain it to me once, but it's like pearls before the swine, to get biblical. 

3. I got the legal-levels & thresholds from Frontier as noted according to http://www.jades.org/ (I think)

The-EG:
> 2. Word of caution, the `setprop("fine", fine)` is half guess-work by me, perhaps it should be placed elsewhere, when the player is given a fine? I know @laarmen tried to explain it to me once, but it's like pearls before the swine, to get biblical.

Just looking at it, I can't see why that would be needed at all in this case. The fines are stored per faction and it doesn't look like it's ever expected to be stored on the player.

data/libs/Player.lua
```diff
@@ -281,6 +281,28 @@ function Player:ClearCrimeRecordHistory (faction)
...
+	elseif fine < 5500 then
+		return('OFFENDER')
+	elseif fine < 20000 then
+		return('CRIMINAL')
+	elseif fine < 100000 then
+		return('OUTLAW')
+	else
+		return('FUGITIVE')
+	end
+end
```
The-EG, line 304: Looking at this closer, you don't need both calls (`GetCrimeRecord` and `GetCrimeOutstanding`). As is, `fine` will be set to the value of `Game.player:GetCrimeOutstanding()` since that's the last call.

Looking at the two, you want `GetCrimeOutstanding` not `GetCrimeRecord` since the latter returns all crimes, even those that are settled.

impaktor: > Looking at this closer, you don't need both calls 

Ha, yes, it's dead code that survived my initial attempt, good catch.

> and you are already in the player class...

will do.

impaktor:
@The-EG well, the code doesn't work without that line, so that's why I'd rather have some smarter eyes on that line, before merging.

Web-eWorks:
Gonna go ahead and merge this, the stray setprop call is fine and can be moved later if needed.

PR #5095, opened 2020-12-27
Restrict entry to Rondel system

WKFO:
I know the title makes it sound stupid (why would you stop the players from entering a system?), but hear me out. I noticed a red spot on the sector map and found this Haber system in the middle of SolFed territory.
rondel1

```
-- Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details

-- example of a custom system directly specifying a faction
local s = CustomSystem:new('Rondel',{'STAR_M'})
	:faction('Haber Corporation')
	:short_desc('Military Listening Post')
	:long_desc([[A hidden dagger pointed at the heart of the Federation]])
	:seed(1824351)

s:add_to_sector(-1,6,2,v(0.007,0.260,0.060))
```

It looks like someone made this system just as a custom system example (no starports, no defined planets etc.), but I thought it could use more than just a fancy description... some more excitement. Which brings me to this PR:

Upon entering the system, 7 police vehicles appear near the player and warn them to leave. They are quite paranoid too:

- They allow 2 (EDIT: now 10) minutes to ships with civilian hyperdrives, 1 (EDIT: now 5) minute(s) to those with military hyperdrives. (Because military technology raises more suspicion, and they become less tolerant) If you don't leave, you get attacked.
- Jettisoning anything or firing a weapon triggers hostility, even if you don't hit the police.
- Police in this system use 2MW weapons as opposed to 1MW used by police patrols in other systems.

As for the code, it has been almost entirely copied from the PolicePatrol module, and edited.

### DIFF
--- a/data/modules/Rondel/Rondel.lua
+++ b/data/modules/Rondel/Rondel.lua
@@ -12,6 +12,7 @@ local Legal = require 'Legal'
 local Serializer = require 'Serializer'
 local Equipment = require 'Equipment'
 local ShipDef = require 'ShipDef'
+local SystemPath = require 'SystemPath'
 local Timer = require 'Timer'
 
 --local Character = require 'Character'
@@ -25,6 +26,8 @@ local jetissionedCargo = false
 
 local rondel_victory = false -- has the player defeated all Rondel guards once in their career?
 local rondel_prize = false -- whether the player has been rewarded or not
+
+local rondel_syspath = SystemPath.New(-1,6,2,0)
 
 local ads = {}
 
@@ -89,7 +92,7 @@ local attackShip = function (ship)
 end
 
 local onShipDestroyed = function (ship, attacker)
-	if ship:IsPlayer() or attacker == nil or Game.system.name ~= "Rondel" then return end
+	if ship:IsPlayer() or attacker == nil or not Game.system.path:IsSameSystem(rondel_syspath) then return end
 
 	for i = 1, #patrol do
 		if patrol[i] == ship then
@@ -108,7 +111,7 @@ local onShipDestroyed = function (ship, attacker)
 end
 
 local onShipFiring = function (ship)
-	if Game.system.name ~= "Rondel" then return end
+	if not Game.system.path:IsSameSystem(rondel_syspath) then return end
 
 	local police = ShipDef[Game.system.faction.policeShip]
 	if ship.shipId ~= police.id then
@@ -124,7 +127,7 @@ local onShipFiring = function (ship)
 end
 
 local onJettison = function (ship, cargo)
-	if Game.system.name ~= "Rondel" then return end
+	if not Game.system.path:IsSameSystem(rondel_syspath) then return end
 	if not jetissionedCargo and #patrol > 1 then
 		Comms.ImportantMessage(l_rondel.JETTISON_DEFENSIVE_PRECAUTION, patrol[1].label)
 		jetissionedCargo = true
@@ -136,7 +139,7 @@ local onEnterSystem = function (player)
 	if not player:IsPlayer() then return end
 
 	local system = Game.system
-	if system.name ~= "Rondel" then return end
+	if not system.path:IsSameSystem(rondel_syspath) then return end
 
 	local tolerance = 1
 	local hyperdrive = Game.player:GetEquip('engine',1)
--- a/data/modules/Rondel/Rondel.lua
+++ b/data/modules/Rondel/Rondel.lua
@@ -1,0 +1,145 @@
+-- Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local Engine = require 'Engine'
+local Lang = require 'Lang'
+local Game = require 'Game'
+local Space = require 'Space'
+local cargo = require 'Commodities'
+local Comms = require 'Comms'
+local Event = require 'Event'
+local Legal = require 'Legal'
+local Serializer = require 'Serializer'
+local Equipment = require 'Equipment'
+local ShipDef = require 'ShipDef'
+local Timer = require 'Timer'
+
+local l_rondel = Lang.GetResource("module-rondel")
+local l_ui_core = Lang.GetResource("ui-core")
+
+local patrol = {}
+local shipFiring = false
+local jetissionedCargo = false
+
+local attackShip = function (ship)
+	for i = 1, #patrol do
+		patrol[i]:AIKill(ship)
+	end
+end
+
+local onShipDestroyed = function (ship, attacker)
+	if ship:IsPlayer() or attacker == nil or Game.system.name ~= "Rondel" then return end
+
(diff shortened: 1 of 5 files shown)


impaktor:
I like the direction, however, I wonder:
- having the 1 minute vs. 2 minute distinction will not be noticeable to a player that's not reading the source code, I think.
- ...nor the pulse lasers unless they have different colour (do they?). (Maybe lasers? or some weapon that look distinctly different)
- should it even be police? How about "haber military"?
- when we put in a restriction for the player, there should be some way the player can defeat it, and also understand how to (compare with reputation system & kills, player is informed they're lacking in one/both to sign up for a mission). 
- Thus, I think there should be a way to gain access to the system, e.g. a permission-system (like in Frontier), or maybe Military drive makes them think you're one of them? The latter would be trivial compared to the former, to implement.
- You could add a rumour/traveller's tale to the advice module that says something about "Haber has an outpost, but I once sneaked in there by using a military drive, but it was worth it, since I could buy plenty of slaves legally over the market"
- perhaps it should go in the same folder as police patrol? (but still have separate language module?) ping @Web-eWorks, or don't mention "Rondel" explicitly in the module name, as this could be flavoured later on to include some other system as well.

Those are my initial thoughts. I'd stress adding hit to Advice module, and some loophole for player to be granted access. Also, making it into the system, there should be some reward, like cheap slaves. I know slavery is illegal in Haber, but I'd suggest we change it, they're supposed to be "evil", so...

data/modules/Rondel/Rondel.lua
```diff
@@ -0,0 +1,142 @@
+-- Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local Engine = require 'Engine'
+local Lang = require 'Lang'
+local Game = require 'Game'
+local Space = require 'Space'
+local cargo = require 'Commodities'
```
impaktor, line 8: looks to be unused.

WKFO: Ah, I will take care of that. Thanks.

WKFO:
In order, replying to @impaktor ;
- You are probably right.
- 2MW lasers have different color and firing rate.
- Could be.
- I thought about that, but there is nothing in the system yet. Maybe there should be something worth inside first.
- Yep.
- Good idea.
- I had a hard time deciding where to put the files. I was about to create a "custom systems" folder, but then decided against that. I'd appreciate advice.

nozmajner:
Isn't 2 minutes a bit too short? Maybe 5 and 15?

impaktor:
OK, I assumed there was an inhabited system.

If there was, response time could be something with the distance between player and main planet, or similar.

Web-eWorks:
Having not looked at code yet (on mobile), I would recommend making this module completely independent of the actual system. I will look into adding a permit system to the Character framework which should make this easy.

Regarding the gameplay implications,  I'd recommend extending the time to five minutes for all drives and making sure that time acceleration doesn't adversely impact the experience (possibly stop time accel when receiving the message).

Organization wise, I'd recommend adding this to the PolicePatrol module folder and associated language resource - there's no limitation that requires only one lua file per module. (And with the `require '.name'` syntax it's easy to work with.)

data/modules/Rondel/Rondel.lua
WKFO: I guess I happen to be using that, actually.

fluffyfreak:
I ahve nothing to say about he code/lcoation/etc but I do like the idea. Perhaps in future this can be modified for any restricted system which could just be a flag on the system itself.

For now I'd be happy to have it be specific to Rondel, then change it in when the time comes.

WKFO:
Moved Rondel.lua to PolicePatrol directory, increased time to 10 mins for civilian drives, 5 mins for military drives.

WKFO:
Added some rumour/backstory about Rondel to the advice module.

WKFO:
Fun fact, it is humanly possible to defeat all 7 ships (yes, I was actually trying that) and brute force your way into the system with a steady hand, some luck and the somewhat humble ship below:
screenshot-20210102-235535
screenshot-20210102-235519
(Last two ships don't generate messages on death, and yes that's on purpose.)

Originally I was just planning to close off the system for this PR, but since I discovered that 7 police vehicles are not undefeatable, things changed. If someone other than me actually went there and defeated all the guards for nothing in return, I guess they would be pretty pissed. So I added a prize - the next time they land on a spaceport after their victory, a special BBS entry is created.

over_here
dialog

An undercover SolFed agent asks the player for the data about Rondel on their ship's computer and gives them 250,000 credits in return (which I absolutely don't think is a ridiculous amount, given the challenge of fighting 7 ships simultaneously). This is a one time thing per save file and doing the same thing again doesn't give any more credits.

Since it has become more complex than just keeping people away from a system, I moved the module back to its own folder (it is more than a police patrol, still with room for more features). It can be further expanded by adding actual points of interests to the system itself, and later, creating a permit system once there is something worth inside. If more modules for different star systems are created, the folder can be renamed accordingly, to house all custom system modules.

(Not planning any of these for this PR.)

ask

data/modules/Rondel/Rondel.lua
```diff
@@ -0,0 +1,225 @@
...
+end
+
+local attackShip = function (ship)
+	for i = 1, #patrol do
+		patrol[i]:AIKill(ship)
+	end
+end
+
+local onShipDestroyed = function (ship, attacker)
+	if ship:IsPlayer() or attacker == nil or Game.system.name ~= "Rondel" then return end
```
Web-eWorks, line 92: I'd recommend comparing the SystemPath rather than the name of the system; it's entirely possible there will be multiple duplicate systems with the same name when you only meant to refer to one.

data/modules/Rondel/Rondel.lua
```diff
@@ -0,0 +1,225 @@
...
+	if hyperdrive.fuel == cargo.military_fuel then
+		tolerance = 0.5
+	end
+
+	local ship
+	local shipdef = ShipDef[system.faction.policeShip]
+	for i = 1, 7 do
+		ship = Space.SpawnShipNear(shipdef.id, player, 50, 100)
+		ship:SetLabel(l_rondel.HABER_DEFENSE_CRAFT)
+		ship:AddEquip(Equipment.laser.pulsecannon_2mw)
```
Web-eWorks, line 152: Sigh... I'm beginning to realize that we need a lib of some sort to handle spawning ships from templates. This works for now though, no need to change it.

Web-eWorks (changes requested):
Looking good and based on prior discussion I think this is getting ready to merge. I only have one change request, and that is to check for Rondel using a SystemPath instead of a name that can easily be duplicated in procgen.

data/modules/Rondel/Rondel.lua
WKFO: I'm clicking "resolve conversation" so it stops telling me there are changes requested.

Web-eWorks (approved):

PR #5105, opened 2020-12-31
Happy New Year!

impaktor:
MERRY HAPPY 2021

(I also found two pigui files that lacked the copyright header, so added them in the same commit)

__EDIT__ I suspect we also need to update the wiki & forum question for which year the game starts in now

__EDIT2__ Since clang format is unhappy I've split it up int separate commits for now.


fluffyfreak:
@impaktor Happy New Year!

impaktor:
OK, clang-format is acting up. I made a autoformat commit to make it happy, and now it still finds things to complain about. I'm guessing this is due to different versions between my system and the build machine's? (my clang version is 11.0.0).

Also, do we want the format changes clang is suggesting? (see my clang format commit). ping c++-ers: @fluffyfreak and @Web-eWorks. As I said in IRC, why would one want to prepend "} // namespace <namespace>" on closing namespaces? Granted, it's +10 years since I did any serious C++ but I don't remember seeing that being a common practice thing to do.

impaktor:
The clang version on my system is 11.0, the version run by the GH action is [6.0](https://packages.ubuntu.com/bionic/clang-format), and there are changes in the defaults between these versions, that are unspecified in our .clang-format configuration file.

This seems to be an issue that doesn't have a [entirely satisfying solution](https://stackoverflow.com/questions/59395507/how-to-maintain-a-clang-format-file-for-different-clang-format-versions)?

I'm not sure what the best thing to do here is? 
- update our ubuntu version, to more modern one, be closer to clang version 11.0?
- find the exact difference between default in 11.0 and 6.0?

PR #5107, opened 2021-01-02
Generate Lua API Documentation and host on Pages

The-EG:
The goal of this is to get the lua API (src/lua, data/libs, and data/pigui) automatically generated and hosted on github pages, using an additional branch on this repository.

There are a few things going on here:
 - updated the Natural Docs configuration in nd/ to the new 2.1 format
 - add a script (scripts/gen_cpp_lua.sh) that grabs all of the comments from src/lua/\*.cpp and makes them into lua comments in gen/\*.cpp.lua
   - this is done so that Natural Docs will combine functions and attributes for classes implemented in both c++ and lua on the same 'class' page.
 - add a workflow for building the documentation. This workflow:
   - should only trigger when files in src/lua, data/libs or data/pigui change in master
   - does a full checkout of pioneer
   - checks out the gh-pages branch
   - merges master into it
   - downloads Natural Docs 2.1
   - generates the documentation
   - creates a commit in the gh-pages branch and pushes up to origin
 - some updates to comments to get things aggregated and looking a bit better (mostly pigui). More will need to be done, but this is a good start.
 
Once the workflow runs, the gh-pages branch will have the generated documentation in docs/lua. The 'lua' folder is used with the expectation that we may want to use this branch and github pages for other things in the future. With this setup, other web things could be hosted on the same branch, and updated by pushing manually or other workflows.

For reference, the new documentation will look like: https://the-eg.github.io/pioneer-luadoc/lua/
The 'classes' section is what users should use. Unfortunately, the 'files' section can't be removed (easily).

I have this functioning in an isolated repository already: 
 - https://github.com/The-EG/pioneer-luadoc
 - https://github.com/The-EG/pioneer-luadoc/runs/1635380163?check_suite_focus=true

A few things for consideration:
 - The github pages branch will be called 'gh-pages.' This can be anything we want, even 'documentation' or something.
 - I couldn't figure out how to do all of this without doing a full clone of the repository. Maybe there is a better way, but I couldn't figure it out. The default settings for checkout leave the branch detached and wouldn't merge properly...
 - The workflow merges master into gh-pages without rebasing. gh-pages will eventually have a lot of 'merged origin/master' commits. Oh well.
 - The workflow git needs an identity to use for the merge and commits. I chose one...
 - An artifact left over from the existing Natural Docs config: there is a differentiation between methods and functions. I think this was intended to differentiate instance methods (':') and functions ('.') but it's not consistent in the comments and it does some weird things in the documentation. I've left it for now, though.

Things that need to be done outside of this PR:
 - [ ] Create the pages branch (gh-pages) - it may need to have a docs folder added as well
 - [ ] Trigger the new workflow (change a file, etc.)
 - [ ] Set the 'pages' option of this repository (pioneer) to use the 'gh-pages' branch (created above) and 'docs/' folder.
 - [ ] Verify documentation is generated properly
 - [ ] Update URLs to point to new gh pages path: https://pioneerspacesim.github.io/pioneer/lua


.github/workflows/build-luadocs.yml
```diff
@@ -0,0 +1,56 @@
...
+        
+    - name: Setup git identity
+      run: |
+        git config --local user.name "Commander Jameson"
+        git config --local user.email "CmdrJameson@pioneerspacesim.net"
+        
+    - name: Merge master into gh-pages
+      run: |
+        git checkout gh-pages
+        git pull --no-rebase origin master
```
Web-eWorks, line 31: Adding on to what I suggested in IRC, this is what the action would look like using a bare branch - very little would need to change, other than the gh-pages config serving out of the root of the `gh-pages` branch instead of the `docs` subfolder.

```suggestion
    - name: Checkout gh-pages branch
      uses: actions/checkout@v2
      with:
        ref: gh-pages
        path: docs
        token: "CMDR Jameson's Token"
```

Web-eWorks:
Copy-pasting here from IRC:

```
The-EG: getting a chance to read the overview of your gh-pages PR now                                  
A few things that stand out to me - it's usually best to use a "bare" gh-pages branch, that is, one    
that only contains the HTML files. I'd suggest the workflow do a shallow-clone of that branch in a     
subfolder, clear all existing generated files, and copy in the generated documentation HTML and make a 
commit. I've seen setups that force-push so the gh-pages branch is only ever one commit deep,          
though that might be a bit overkill for our needs.                                                     
*clear all existing generated files in the cloned branch, that is                                      
This ensures that we keep the repo size small and the workflow should be very fast to run as it doesn't
need to make duplicates of the code commit history.                                                    
The nice bit about git is that branches can have whatever they want in them; it's trivial to create a  
branch that's totally empty of files to store something that's logically separate from the contents of 
the "main" branch.                                                                                     
in this case, that's the website                                                                       
```

impaktor:
Perhaps it would also possible to add a github "documentation" badge to the README.md, as a final step, sort of like the cherry on the cake(?)

The-EG:
> A few things that stand out to me - it's usually best to use a "bare" gh-pages branch, that is, one
> that only contains the HTML files.

Is there any reason why? I've seen this quite a few times, but never a reason other than 'everyone does it this way.' It doesn't save space, and it would need to be detached from master or it would actually be less efficient.

And, unfortunately I still need all of the code and configuration from master, so it's not saving a full (shallow) checkout there either. If I have a bare detached branch, then I have to checkout both master and the branch.

If we are that worried about how it looks, then it may be better to just have it go to a dedicated repository under pioneerspacesim. At this point, that'd be basically the same.

> I've seen setups that force-push so the gh-pages branch is only ever one commit deep, though that might be a bit overkill for our needs.

See my original description: 
> The 'lua' folder is used with the expectation that we may want to use this branch and github pages for other things in the future. 

Again, we can purely for appearances, but my intent is that this can be used for other workflows and even manually pushing if needed. I suppose that this and any other workflow could just assume it's ok to amend the last commit and force push...but why?

The-EG:
To move this forward I'm going to backout the changes to the github actions and just update the natural docs config and changes to the documentation comments so that we can at least get the documentation looking better.

Then we can decide where/how to host it whenever.

PR #5114, opened 2021-01-05
Add search for the geometrically closest jumpable

### LINKED ISSUE #5112: Game crashes when setting destination as navigation target in the BB
### Observed behaviour
Game crashes when setting destination as navigation target in the BB

### Steps to reproduce
Load the attached save file, go to the BB and chose a "Fearless mercenary wanted" mission to Veurio system. Ask to set destination as navigation target.

Pioneer Screenshot 2021 01 05 - 11 32 59 11
[Pioneer-2.zip](https://github.com/pioneerspacesim/pioneer/files/5768985/Pioneer-2.zip)

__My pioneer version (and OS):__ 20201222-rc1 (Windows 10)

Gliese852:
In case the systembody does not have a jumpable parent.
This can happen if it revolves around several stars at once.
At the moment this situation leads to a segfault.
Fixes #5112

### DIFF
--- a/src/galaxy/SystemBody.cpp
+++ b/src/galaxy/SystemBody.cpp
@@ -3,9 +3,11 @@
 
 #include "SystemBody.h"
 
-#include "Lang.h"
-#include "EnumStrings.h"
 #include "AtmosphereParameters.h"
+#include "EnumStrings.h"
+#include "Game.h"
+#include "Lang.h"
+#include "Pi.h"
 #include "enum_table.h"
 #include "utils.h"
 
@@ -467,7 +469,7 @@ bool SystemBody::IsPlanet() const
 	}
 }
 
-void CollectSystemBodies(SystemBody *sb, std::vector<SystemBody*> &sb_vector)
+void CollectSystemBodies(SystemBody *sb, std::vector<SystemBody *> &sb_vector)
 {
 	for (SystemBody *body : sb->GetChildren()) {
 		sb_vector.push_back(body);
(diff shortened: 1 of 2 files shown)


Gliese852:
Slightly improved implementation.

src/galaxy/SystemBody.cpp
```diff
@@ -575,3 +577,54 @@ void SystemBody::ClearParentAndChildPointers()
...
+
+	//  Now we climbed to the very top of the hierarchy and did not find a jumpable
+	//  parent - we will have to search purely geometrically
+	//  we go through all the bodies of the system, determining the coordinates
+	//  of jumpable objects, and also remember the coordinates of the current
+	//  object
+	//  the result variable now contains the root systembody
+	std::vector<std::pair<SystemBody *, vector3d>> jumpables;
+	vector3d this_position(0.0);
+	std::function<void(SystemBody *, vector3d)> collect_positions = [&](SystemBody *s, vector3d pos) {
```
Web-eWorks, line 599: Instead of using `std::function`, please prefer using `auto collect_positions = []...` - std::function has non-negligible overhead on construction.

Web-eWorks (changes requested):

src/galaxy/SystemBody.cpp
Gliese852: @Web-eWorks the problem is that when using `auto`, a compilation error occurs as I have a recursive call there

Web-eWorks: Ah, got it. Was expecting recursion to be possible, but thanks C++11!

Web-eWorks (approved):

Issue #5113, opened 2021-01-05
Menu music loops in main game


impaktor:
### Observed behaviour
Menu music loops in the main game, according to player on [ssc](https://spacesimcentral.com/community/pioneer/new-pioneer-release-2020-12-22/#post-59004)

Possibly related #5049

__My pioneer version (and OS):__ 20201222-rc1

WKFO:
Funny, I can't reproduce in 20201222-rc1 nor master. It also looks like MusicPlayer.lua in the rc-1's data folder has the fixed version of the module.

impaktor:
Thanks for checking.
Yeah, it would be good if we could get definitive confirmation that this bug still exists.

impaktor:
From the SSC thread:

> I would like to answer your question on GitHub regarding the menu music (if the bug still occurs): Yes, it does, but I am not able to reproduce it specifically. The only thing I notice is that it never happens when I start Pioneer after restarting my computer. Otherwise... I don't know. And since I'm not a coder, I can't come up with any kind of idea what the problem might be. I'm sorry.

another similar bug:

> The lost one from which I think it's not mentioned is that, if you land on earths ground, weather- sound is playing from then on and for... well - ever, even if you launch your ship again. It's like having a whirlpool right behind the pilot's seat...

WKFO:
Ah, I tested it on master and it looks like the planet sound effect does keep playing after launch (until you leave Earth's atmosphere, that is). I still have no idea about what could be causing the music problem though.

**Not** relevant to the bug: speaking of planet sounds, maybe we could use a fade in/out effect for them as well. They start and end a bit too abruptly.

WKFO:
Can anyone else reproduce this?

Web-eWorks:
Closing this as unreproduced / user update error. Please reopen if we have a *concrete* set of reproduction steps against a reasonably recent `master` build.

PR #5115, opened 2021-01-06
Stop planet ambient sound once player gets off the ground

WKFO:
There is not much to explain. For example if you rough-land on Earth's surface it plays some water sound, and we want this sound effect to stop once the player is airborne, not when they fly out of the atmosphere.

Partially f i x e s #5113


impaktor:
thanks for fix

PR #5116, opened 2021-01-09
Blend the radius searched for stars based on desired density.

### LINKED ISSUE #4990: Skybox star density behaves oddly
### Observed behaviour
screenshot-20201031-152401
Above is Sag A* (-3125, 0, 0) with star density at 50 percent.
screenshot-20201031-152547
Above is Sol (0, 0, 0) with star density at 0 percent.

Some part of the sky gets "cut off" (as if someone sliced the galaxy into two using a cosmic knife and took a part of it away). This happens when you get close to the galactic center or change the star density.

### Expected behaviour
Spread stars somewhat homogeneously after density change.

### Steps to reproduce
Play with star density slider.

8b698e1cf5f95d35d0fa4b9f17e8c348fff170e3 (Win 10 x64)

fluffyfreak:
Limit the radius searched for stars and mix it from the minimum to the desired radius depending on the star density.

Fixes #4990

### DIFF
--- a/src/Background.cpp
+++ b/src/Background.cpp
@@ -6,6 +6,7 @@
 #include "FileSystem.h"
 #include "Game.h"
 #include "GameConfig.h"
+#include "MathUtil.h"
 #include "Pi.h"
 #include "Player.h"
 #include "Space.h"
@@ -27,6 +28,8 @@ using namespace Graphics;
 namespace {
 	static const Uint32 BG_STAR_MAX = 500000;
 	static const Uint32 BG_STAR_MIN = 50000;
+	static const float BG_STAR_RADIUS_MIN = 8.0f;
+	static const float BG_STAR_RADIUS_MAX = 500.0f;
 	static RefCountedPtr<Graphics::Texture> s_defaultCubeMap;
 
 	static Uint32 GetNumSkyboxes()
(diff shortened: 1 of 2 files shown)


fluffyfreak:
Also @Web-eWorks I keep being really impressed by how good the Ctrl+i debug/performance display is!

impaktor:
If one sets to 0 background stars, will one still get the "real" stars then?

fluffyfreak:
The minimum was set to 50,000 stars, so it'll work like this:

- Minimum real stars min radius = 8 light years
- Target mac radius for stars is set in inin file to 220 light years
- "0" stars is actualy equal to a multiplier of 0.01 (_this hasn't changed_).
- So you'll get all real stars within a lerp of 8.0 to 220 by 0.01 amount = 10.12 light years.
- After that it'll add in fake stars until it reaches the target number of background stars which will be = 54,500 due to also lerping by 0.01 between 50,000 and 500,000

fluffyfreak:
@impaktor and now I've changed the lerp/mix range to 0.0 to 1.0 👍

Web-eWorks (approved):
Looks nice and clean as always! Haven't had a chance to test it, but it's minimal enough I don't think there will be any bugs to find.

PR #5123, opened 2021-01-19
Improve/fix system map

### LINKED ISSUE #5062: System viev: Ship vanishes when the reference system is an orbital starport
Another detail you may correct if you feel motivated:

### Observed behaviour
The ship vanishes from the system view when the reference system is an orbital starport: (20 km around)

Sysmap_SpacestationReference

Mainview_SpacestationReference

### Expected behaviour

A ship's position should still/(always?) be visible in the system map

20201206 master and Gliese852's version on SuSE Linux

Gliese852:
- add a "static" orbit for bodies that are not moving anywhere
- now bodies in a rotating frame, and also in the space port frame have such "static" orbits
- add display of the orbits of objects moving strictly to the star or from the star
- also fix compile warning for possibly uninitiated variable in `LuaSystemView.cpp`
- also fix an issue where smooth animation sometimes have a hard time catching up with a newly selected object:

https://user-images.githubusercontent.com/18342621/105128153-859ace80-5af3-11eb-89b6-72e05428b993.mp4

Fixes #5062

### DIFF
--- a/src/DynamicBody.cpp
+++ b/src/DynamicBody.cpp
@@ -371,7 +371,12 @@ bool DynamicBody::OnCollision(Body *o, Uint32 flags, double relVel)
 // return parameters for orbit of any body, gives both elliptic and hyperbolic trajectories
 Orbit DynamicBody::ComputeOrbit() const
 {
-	FrameId nrFrameId = Frame::GetFrame(GetFrame())->GetNonRotFrame();
+	auto f = Frame::GetFrame(GetFrame());
+	// if we are in a rotating frame, then dynamic body currently under the
+	// influence of a rotational frame, therefore getting the orbital parameters
+	// is not appropriate, return the orbit as a fixed point
+	if (f->IsRotFrame()) return Orbit::ForStaticBody(GetPosition());
+	FrameId nrFrameId = f->GetId();
 	const Frame *nrFrame = Frame::GetFrame(nrFrameId);
 	const double mass = nrFrame->GetSystemBody()->GetMass();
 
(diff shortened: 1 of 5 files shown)


src/DynamicBody.cpp
```diff
@@ -371,7 +371,10 @@ bool DynamicBody::OnCollision(Body *o, Uint32 flags, double relVel)
 // return parameters for orbit of any body, gives both elliptic and hyperbolic trajectories
 Orbit DynamicBody::ComputeOrbit() const
 {
-	FrameId nrFrameId = Frame::GetFrame(GetFrame())->GetNonRotFrame();
+	auto f = Frame::GetFrame(GetFrame());
+	// if we are in a rotating frame, there can be no orbit
+	if (f->IsRotFrame()) return Orbit::ForStaticBody(GetPosition());
```
Web-eWorks, line 376: If a ship is in a rotating frame it occurs to me that there definitely is an orbit (although for suborbital flight, the periapsis might fall below the surface of the body;) it is simply that the velocity of the ship is defined relative to the rotational velocity of the rotating frame. I'm not sure if our Orbit abstraction is robust enough to handle that case however.

If you could, I'd appreciate it if you could update the comment to mention the specific situation where getting the orbital parameters of a dynamic body currently under the influence of a rotational frame is not appropriate.

Bodasey:
I have tested this version for an hour - i'd say FIX CONFIRMED

src/DynamicBody.cpp
Gliese852: Sure!

impaktor:
I assume this can be merged. If not, speak up. If someone/@Gliese852 needs it in master now/soon: speak up.

src/DynamicBody.cpp
Web-eWorks: Apologies, I was intending more that you should iterate *why* being in a rotating frame makes it inappropriate to retrieve orbital parameters for a body. Don't worry about it for now, I think this is good as is.

Web-eWorks (approved):

PR #5124, opened 2021-01-21
General Improvements to star field

tsoj:
screenshot1
Star density at 25%, star size at 70%¹
________________________________________________
screenshot2
Star density at 100%, star size at 70%¹
________________________________________________
screenshot2
Star density at 50%, star size at 100%¹
________________________________________________
¹ Edited: I decreased the maximum possible star size by half, before it was at 35%, 35%, 50% see 302e8b01ce77276b30973f2600aa1dd72d38d366

 - Added a skybox based on [NASA photographs](https://svs.gsfc.nasa.gov/4851), it has pretty much no license (https://www.nasa.gov/multimedia/guidelines/index.html), but I am no lawyer.  
This skybox will of course not be realistic once you move away from space near our own solar system, but I personally think that this is no big deal, the old version wasn't better in that regard.  
I made two skybox versions, one exactly like the original photo (`skybox.dds`), the other without stars (`skyboxNoStars.dds`, default), because I thought that the stars should be generated. 
- I removed unnecessary parameters for the star field generation:
  - `brightnessApparentSizeFactor` (star size) is now a configurable parameter that can also be changed with a slider in the option-menu
  - `brightnessApparentSizeOffset`, `brightnessColorFactor`, `brightnessColorOffset` made no real difference in appearance 
  - `brightnessPower` is to an extent the same as `brightnessApparentSizeFactor`, but I am unsure whether it should be a configurable parameter, and if yes where? In `GameConifg.cpp` or rather like before in `Starfield.ini`. Same with `medianPosition`
  - `visibleRadiusLy` is now computed fully based on the selected star density
- Fixed hyperspace animation from [messy](https://i.imgur.com/AS1YEI8.png) to [less messy](https://i.imgur.com/EvhMTa8.png).
- The intro/main-menu background now has generated galaxy stars instead of fake ones.² This not only looks nicer, but it is now possible to see the result of the star field density and star size options without having to start a game.

TODO:
- ² The generated stars for the intro background create some performance penalty at start up, I personally would say it is acceptable. Also, I changed some code bits where I am not sure if it has any side effects


impaktor:
You only need to edit en.json, so please remove all other {lang}.json files.

src/Background.cpp
```diff
@@ -27,7 +27,7 @@ using namespace Graphics;
 
 namespace {
 	static const Uint32 BG_STAR_MAX = 500000;
-	static const Uint32 BG_STAR_MIN = 50000;
+	static const Uint32 BG_STAR_MIN = 1;
```
fluffyfreak, line 30: The reason this was 50,000 was so that there are always some stars generated. With it at 1 you could literally have a starfield with just 1 star.

nozmajner:
Looks nice!
I'd say don't up the star size to 50% It starts to clash with the UI elements. That was the reason I decreased it.

src/Background.cpp
```diff
@@ -259,19 +253,30 @@ namespace Background {
...
+		// TODO: find a nicer solution for main menu
+
 		//fill the array
 		Uint32 num = 0;
 		if (space != nullptr && galaxy.Valid() && space->GetStarSystem() != nullptr) {
 			const SystemPath current = space->GetStarSystem()->GetPath();
 
-			const double size = 1.0;
-			const Sint32 visibleRadius = MathUtil::mix(BG_STAR_RADIUS_MIN, Clamp(m_visibleRadiusLy, BG_STAR_RADIUS_MIN, BG_STAR_RADIUS_MAX), Pi::GetAmountBackgroundStars()); // lyrs
+			constexpr float starsPerSector = 2.0; /* experimentally determined (already ajusted for that we are
```
fluffyfreak, line 264: The potential sectors used are selected in a square, because that was easy, but are then rejected by distance leaving only stars within a spherical region.

fluffyfreak: Also why pick only 2 stars per sectors when we have many more?

tsoj: The problem I encountered was that if `BG_STAR_RADIUS_MAX` was increased we'd run again in the issue that only half of the sky has stars. I also found that we can increase the visibleRadius far over the hard-coded 220 LY when selecting the maximum density, which I thought is nicer than filling up the rest with fake stars.

My idea was that we want to know, how big we need to make the `visibleRadius` to get approximately the number of stars we want (`NUM_BG_STARS`).

4/3 * pi * visibleRadius^3 = numSectors * sectorSize^3, i.e. I want the sum of the volume of all sectors be approximately the same as the volume of the sphere with radius `visibleRadius`. We can the numSectors by experimentally checking how many stars on average are in one sector. And I found that on average there are 2 stars in one sector (or rather 2 solar systems per sector). If we'd change the size of a sector this would break that's why I'm not sure if it's actually ok.

src/Background.cpp
tsoj: maybe there's someone who wants a pitch black sky :)

tsoj:
> You only need to edit en.json, so please remove all other {lang}.json files.

Won't they then also get deleted here if this pull request would merge?

The-EG:
> Won't they then also get deleted here if this pull request would merge?

Don't remove the entire files, just your changes to them. In other words, just edit the en.json file and leave the rest unchanged.

src/Background.cpp
fluffyfreak: Ah no, the fake stars are there because we never want to run the sector generator too much. It's incredibly expensive, so the populated stars are really only a shell close to the player, and anything further away is just ... padding to make it look more believable.

An extension to this idea is to eventually have the stars within jump distance be displayed on the night sky somehow, so again, that's only going to be stars close enough that the ship can jump too them.

tsoj: Oh ok, didn't think about that. What about if I cap the visible radius after calculating to BG_STAR_RADIUS_MAX or a parameter that can be changed? To be honest it takes on my notebook only half a second to generate the star field at maximum density, but this way will be easy to change it if its performance becomes more important.

Web-eWorks:
@tsoj your `credit_for_skybox_source` file should be merged into `AUTHORS.txt` at the root of this repository. At the bare minimum if you think it's important enough to keep it separate, it needs a file extension - Pioneer is cross-platform and not all platforms work correctly with extension-less files.

nozmajner:
Checked the skybox in-game, looks nice. 
Shouldn't it be named default.dds?

tsoj:
> Shouldn't it be named default.dds?

@nozmajner  That's how it was named before yes, I used the more descriptive name to make it easier for myself. But it doesn't really matter, if it is better I can rename it to `default.dds`

I think it would be generally a good idea to make the skybox filename not hard coded, at least using come config file or even making it an in-game option, but I think that can be done in another pull request.

nozmajner:
I'm fine with the name, if the game picks it up.

impaktor:
Not having read the code, but if it has a new name, the original could stay, as optional choice, or?

tsoj:
I'll add it. However, this doesn't make much sense at the moment, as the filename is hardcoded.

impaktor:
OK, hold off on any changes, I was just trying out the idea.

tsoj:
Okay, I think I fixed the conflicts.

Web-eWorks:
@tsoj sorry this has taken so long to get reviewed and merged. I'd appreciate it if you could do before-and-after profiles of the time taken by the background setup so we can measure what kind of performance impact this has, given that you mentioned it was slower than the current code.
The profile points should be there already in Starfield::Fill, and if you add `ProfilerZoneOutput=1` to your config.ini it will generate JSON-format trace files in the `profiler/NewGame` user data directory when you start a new game - those can be loaded in [Speedscope](https://speedscope.app) and they'll give you a visual representation of what time is being taken by what step in the new-game startup phase, including background creation.

tsoj:
@Web-eWorks 
Before it takes around 270 ms for `Starfield::Fill` and now at the highest possible star density (1.0) it takes 600 ms (on my notebook with an i5-8250U). When the star density is set to 0.5 (the current default) it takes 300 ms.

What I could add is another option for the number of real stars that is independent of the star density, which could then be used to adjust the performance. The rest of the stars that would need to be added to reach the chosen star density would be fake stars (randomly generated). Though, I am wondering if that would be really necessary for this kind of performance penalty.

I didn't do proper benchmarks (I had other programs running at the same time, and I didn't run all version of pioneer at the same time), but I think these numbers are roughly correct. For all tests, I did the same thing in-game: launched (a new game) from Mars, flew a bit around, jumped to another system and quit.

Web-eWorks:
> Before it takes around 270 ms for Starfield::Fill and now at the highest possible star density (1.0) it takes 600 ms (on my notebook with an i5-8250U). When the star density is set to 0.5 (the current default) it takes 300 ms.

Excellent, thank you! Wanted to get a "real world" example of what kind of performance difference we're talking about here; looks like it's negligible enough to not be a concern.

Regarding building with profiler, `./bootstrap -DPROFILER_ENABLED=1` is the way to go; no source code changes are required.

Web-eWorks:
Merged! Thank you for the work and the patience with the review process!

PR #5126, opened 2021-01-25
Move the surface starport's system bodies on loading

Gliese852:
Sometimes surface ports move when terrain is generated because it is not suitable for their original position.  When serialized, the physical body saves its changed position, but the system body does not.
Therefore, upon subsequent loading, the position of the system body and the physical body may not coincide:

image
(In the picture, the player's ship has landed on Cydonia)

In this commit, the same mechanism is launched as during the initial generation of the relief, which displaces the system bodies to the right position.

### DIFF
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -28,6 +28,131 @@
 
 //#define DEBUG_CACHE
 
+static void RelocateStarportIfNecessary(SystemBody *sbody, Planet *planet, vector3d &pos, matrix3x3d &rot, const std::vector<vector3d> &prevPositions)
+{
+	const double radius = planet->GetSystemBody()->GetRadius();
+
+	// suggested position
+	rot = sbody->GetOrbit().GetPlane();
+	pos = rot * vector3d(0, 1, 0);
+
+	// Check if height varies too much around the starport center
+	// by sampling 6 points around it. try upto 100 new positions randomly until a match is found
+	// this is not guaranteed to find a match but greatly increases the chancessteroids which are not too steep.
+
+	bool variationWithinLimits = true;
+	double bestVariation = 1e10; // any high value
+	matrix3x3d rotNotUnderwaterWithLeastVariation = rot;
+	vector3d posNotUnderwaterWithLeastVariation = pos;
+	const double heightVariationCheckThreshold = 0.008;					 // max variation to radius radius ratio to check for local slope, ganymede is around 0.01
+	const double terrainHeightVariation = planet->GetMaxFeatureRadius(); //in radii
+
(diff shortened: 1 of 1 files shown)


impaktor:
@fluffyfreak I think you've worked a lot on that code, you have any thoughts, before we/I merge?

src/Space.cpp
```diff
@@ -149,6 +151,28 @@ Space::Space(Game *game, RefCountedPtr<Galaxy> galaxy, const Json &jsonObj, doub
 	for (Body *b : m_bodies)
 		b->PostLoadFixup(this);
 
+	// some spaceports could be moved, now their physical bodies were loaded from
+	// json, with offsets, now we should move the system bodies also so that
+	// there is no mismatch
+	for (auto sbody : m_starSystem->GetBodies())
```
fluffyfreak, line 157: Style point but please use brackets on multi-line for loops

src/Space.cpp
```diff
@@ -108,6 +108,8 @@ Space::Space(Game *game, RefCountedPtr<Galaxy> galaxy, const SystemPath &path, S
 	GenSectorCache(galaxy, &path);
 }
 
+static void RelocateStarportIfNecessary(SystemBody *sbody, Planet *planet, vector3d &pos, matrix3x3d &rot, const std::vector<vector3d> &prevPositions);
```
fluffyfreak, line 111: Style point, please move this to the top of the file so that all external function declarations are grouped up there, or include it's definition if it's already exposed somewhere else.

fluffyfreak (changes requested):
Couple of minor style changes but otherwise it looks good

fluffyfreak:
I'd really like it if we did modify the terrain for cities, rather than moving the cities but 🤷‍♂️

Web-eWorks:
> I'd really like it if we did modify the terrain for cities, rather than moving the cities but man_shrugging obviously i'm NOT asking for that in this PR rofl just saying it out loud

One thing at a time, grasshopper... but in all seriousness that's something I've been doing some thinking on, about being able to merge a "terrain impostor" around cities that overrides the terrain calculations. Would likely also entail generating terrain patches as textures or byte arrays instead of running the fractal every single time for every micro-step we do in the physics code... should also make querying normals and slope data much easier and faster.

...sadly it's going to have to wait until I have a number of free weekends to deal with the more outstanding projects I have in flight for Pioneer.

src/Space.cpp
Gliese852: Done!

src/Space.cpp
Gliese852: This function is defined in the same file, just below, I didn't want to move it, now I just moved the whole function to the beginning of the file, will it go?

fluffyfreak (approved):

PR #5127, opened 2021-01-26
Fix that sometimes ships get stuck the moment they start to dock

Gliese852:
- Fix that sometimes ships get stuck the moment they start to dock.
- ~~Fix that during a long free flight the autopilot constantly corrects the course towards the target and this leads to unnecessary fuel consumption.~~ **Сan't reproduce**
- ~~Add a simple queue for docking with the station in order to somewhat reduce the probability of a collision of ships~~ **It didn't work out that easy, let's postpone**

~~Also in order to find out which ships are in a given frame, I implemented the `Frame::GetBodiesInside` function, and for its work - the `CollisionSpace::GetGeoms` function.~~

### DIFF
--- a/src/ShipAICmd.cpp
+++ b/src/ShipAICmd.cpp
@@ -1389,6 +1389,10 @@ bool AICmdDock::TimeStepUpdate()
 		if (m_state != eDockGetDataEnd) {
 			m_dockpos = dockpos.pos;
 		}
+		// we are already near the station, what could go wrong?
+		// set the fuel reserve to 0, since the fuel could become lower than the
+		// current reserve during the flight, and the ship will not fly anywhere
+		m_prop->SetFuelReserve(0.0);
 
 		m_dockdir = dockpos.zaxis.Normalized();
 		m_dockupdir = dockpos.yaxis.Normalized(); // don't trust these enough


Gliese852:
sorry for the force pushes, minor alterations

Gliese852:
Turned into a draft, because the queue does not work well at large time accelerations

src/ShipAICmd.cpp
```diff
@@ -1281,24 +1305,60 @@ bool AICmdDock::TimeStepUpdate()
...
-		m_dockdir = dockpos.zaxis.Normalized();
-		m_dockupdir = dockpos.yaxis.Normalized(); // don't trust these enough
-		if (type->IsOrbitalStation()) {
-			m_dockupdir = -m_dockupdir;
+		const SpaceStationType *type = m_target->GetStationType();
+		// calculate which we are in the queue.
+		int queue_pos = 0;
+		// since ground stations have a different path for each pad, this queue is
+		// not needed there
+		if (type->IsOrbitalStation())
```
fluffyfreak, line 1318: No single-line if/for statements please, always use brackets especially with something like this.

fluffyfreak:
What happens to the queue at large time accelerations?
And if we're under high time acceleration can we skip the proper docking and just swtich from arrings straight to docked?

src/ShipAICmd.cpp
Web-eWorks: We have a decent amount of single-body-line if statements without brackets, though I agree in this case that if the body of the if statement is greater than one line it should have brackets.

Gliese852:
@fluffyfreak 

> What happens to the queue at large time accelerations?

I noticed that with large accelerations of time, the queue begins to grow, up to dozens of ships, although with normal acceleration, there are a maximum of 2 ships in the queue (we have a very small flow)

> And if we're under high time acceleration can we skip the proper docking and just swtich from arrings straight to docked?

yes, we will probably do so, although in the future I would like to see ships in the queue for which there is still no room at the station.

Bodasey:
@Gliese852 If it's easy to separate the bugfix part from the new feature/improvement part, i please you to do as fluffyfreak proposed. I propose this in general to avoid annoying extra efforts and clarifications when you hard test a "redesigned internal code" branch and you first find a lot of things who are heritage from master branch.

First test: Superior conjunctions of the destination during the flight are done better (the ship now passes in front of the star and not behind any more), but you can still run out of fuel. The autopilot should realize them earlier? Get a better idea of the destination's position at an expected time of arrival?

Gliese852:
All that was left of my PR was just one line, although it is quite useful.

fluffyfreak (approved):

PR #5128, opened 2021-01-26
Modify GetAtmosphericState() to accept any two bodies

WKFO:
...instead of just the player. So you can calculate the pressure around any object in atmosphere, not just the player ship.
EDIT: The main motivation for this PR is that it fixes the traffic control giving wrong pressure data.


WKFO:
I apparently forgot to revert that lua_pushnil thingy, did that just now.

Web-eWorks (approved):
Straightforward, simple, and best of all, correct. Thank you for fixing this!

PR #5129, opened 2021-01-29
Fix death music not playing over tombstone screen

WKFO:
While trying to prevent sounds from playing after the game was ended, Pi.cpp accidentally(?) stopped music player from playing as well, which meant no death music would play as the tombstone spinner appeared. Now it works as intended; sound effects stop, but music is not stopped until death view is closed and the program returns to main menu (at which point menu music starts).

I made this a separate PR from #5093 so that only the fix can be merged if one desires.

### DIFF
--- a/src/sound/Sound.cpp
+++ b/src/sound/Sound.cpp
@@ -468,6 +468,17 @@ namespace Sound {
 		SDL_UnlockAudioDevice(m_audioDevice);
 	}
 
+	void DestroyAllEventsExceptMusic()
+	{
+		/* silence any sound events EXCEPT music
+		   which are on wavstream[0] and [1] */
+		SDL_LockAudioDevice(m_audioDevice);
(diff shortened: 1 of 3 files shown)


src/sound/Sound.cpp
```diff
@@ -468,6 +468,17 @@ namespace Sound {
 		SDL_UnlockAudioDevice(m_audioDevice);
 	}
 
+	void DestroyAllEventsExceptMusic()
+	{
+		/* silence any sound events EXCEPT music
+		   which are on wawstream[1] and [2] */
```
WKFO, line 474: I meant 0 and 1, will fix in a second...

WKFO:
Force-pushed a typo fix on the comment.

Web-eWorks (approved):
The solution is a bit rough around the edges, but that's just our Sound API. Another thing for another day... Looking good, WKFO! Sorry this took so long to do.

PR #5131, opened 2021-01-31
Updating the orientation of the docking tags. - No more backwards docking

nozmajner:
I've played around a bit with the docking tags and found the proper orientation. No more ass-backwards docking or inexplicable rotating around. 
For posterity's sake, this is the orientation of the tags:
station_docking_orientation
It's from top view in Blender, since you need to rotate all the tags 90° on the X axis for them to line up. The Upper-center single tag is the entry indicator, its Z points inwards. All other locator Zs point the opposite direction

For some reason with this setup, the docking system picks one of the small pads for the Sinonatrix (even the Deneb docks there). Which does makes sense. Although there's less detail on that part of the station. 
screenshot-20210131-100500

The downside is that now on the large station the ship goes through the whole station quite fast on rails.
Exit direction stays proper, you don't turn around.


impaktor:
Perhaps this issue should be linked from somewhere on the wiki (if not already done)? Maybe in Modelling?

fluffyfreak:
You beauty! Well done @nozmajner

nozmajner:
@impaktor Good point! Added it.

Bodasey:
Looks nice - and this branch successfully loads games saved with the stable version also when a ship is approaching an orbital starport when saving.
To link and close?
#3344

impaktor:
If this is done we could merge it, and make master branch "unstable"/"testing" state again.

nozmajner:
I think it's done

Web-eWorks:
Good to merge - we had a successful release, so feel free to start merging wild-eyed broken PRs again!

impaktor:
Merge untested. Let's live dangerously again!

PR #5132, opened 2021-01-31
Update Quickstart.txt

WKFO:
Someone WILL most likely read Quicksave.txt and try to learn the game from there, so this is a miserable attempt to reduce the amount of "Why is X not working?" complaints a little. I never use default controls so if there is something wrong there, someone might want to take a look at that.

### DIFF
--- a/Quickstart.txt
+++ b/Quickstart.txt
@@ -3,16 +3,15 @@ Getting started with Pioneer
 
 The principles of the game are largely the same as those of Frontier: Elite 2.
 
-You can buy and sell on the commodities market, take missions on the bulletin
-board, buy and upgrade ships in the shipyard, when docked. Do all of this with
-the mouse.
-
-The eight main console buttons are activated by the F1-F8 keys or the mouse.
+You can buy and sell on the commodities market, take missions or interact with
+people on the bulletin board, buy new ships in the ship market, install or
+remove equipment on the equipment market, repair your ship or contact local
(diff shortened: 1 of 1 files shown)


Quickstart.txt
```diff
@@ -66,21 +65,26 @@ button.
 Use the cursor keys to move the camera in external and sidereal view. In
 sidereal view, use ',' (comma) and '.' (period) to roll the camera in place.
 
-To jump to another system, use the map (F2) and click on another system. Systems
-with ports are labelled in blue. You need enough Hydrogen to get there, and it
-must be close enough for your ship. F7 will start the hyperdrive.
+To jump to another system, use the map (F2) and click on another system. The
+map will provide you with info about the system, such as population, faction
```
nozmajner, line 69: You also need to add it to your route. That caused confusion a couple of times if I remember correctly.

WKFO: Ah, true. Good find.

WKFO: I changed it, you may check if it is good.

nozmajner: Sounds good.

impaktor:
Good initiative. 

I note the linked wiki page, does not in turn link to any of these very useful pages:
https://pioneerwiki.com/wiki/Basic_flight
https://pioneerwiki.com/wiki/Flight_UI

I don't know if we want to minimize the number of wiki links in the quickstart? Or fi those three articles should reference each other?

PR #5141, opened 2021-02-05
Support PowerPC 64 platform

runlevel5:
## Goals

* Compile the game sucessfully on PPC64LE

## Notes

There are quirks with graphics when the game starts which I will see if I could fix them or not in a separate PR

### DIFF
--- a/contrib/profiler/Profiler.h
+++ b/contrib/profiler/Profiler.h
@@ -8,7 +8,7 @@
 #define __PROFILER_FULL_TYPE_EXPANSION__
 
 //#define USE_CHRONO
-#if !defined(USE_CHRONO) && (defined(__arm__) || defined(__aarch64__) || defined(_M_AMD64) || defined(_WIN64) || defined(_M_X64))
+#if !defined(USE_CHRONO) && (defined(__arm__) || defined(__aarch64__) || defined(__PPC64__) || defined(_M_AMD64) || defined(_WIN64) || defined(_M_X64))
 // this isn't optional for __arm__ or x64 builds
 #define USE_CHRONO
 #endif
--- a/contrib/miniz/miniz.h
+++ b/contrib/miniz/miniz.h
@@ -195,7 +195,7 @@
 #define MINIZ_USE_UNALIGNED_LOADS_AND_STORES 1
 #endif
 
-#if defined(_M_X64) || defined(_WIN64) || defined(__MINGW64__) || defined(_LP64) || defined(__LP64__) || defined(__ia64__) || defined(__x86_64__)
+#if defined(_M_X64) || defined(_WIN64) || defined(__MINGW64__) || defined(_LP64) || defined(__LP64__) || defined(__ia64__) || defined(__x86_64__) || defined(__PPC64__)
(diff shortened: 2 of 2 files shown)


contrib/profiler/Profiler.h
```diff
@@ -8,7 +8,7 @@
 #define __PROFILER_FULL_TYPE_EXPANSION__
 
 //#define USE_CHRONO
-#if !defined(USE_CHRONO) && (defined(__arm__) || defined(__aarch64__) || defined(_M_AMD64) || defined(_WIN64) || defined(_M_X64))
+#if !defined(USE_CHRONO) && (defined(__arm__) || defined(__aarch64__) || defined(__PPC64__) || defined(_M_AMD64) || defined(_WIN64) || defined(_M_X64))
```
runlevel5, line 11: This change resolve the compilation error:

```
error: invalid output constraint '=a' in asm
     asm volatile("rdtsc" : "=a" (__a), "=d" (__d));
```

fluffyfreak:
This is... interesting. What PPC64LE machine are you using? I couldn't even find where you could buy one!

runlevel5:
@fluffyfreak I am using the Raptor Computing Black Bird, please check out https://www.raptorcs.com/BB/

runlevel5:
FYI TravisCI does offer agents in PPC64LE

fluffyfreak:
Interesting, I can't justify buying one myself but good luck getting it to run!

impaktor:
@fluffyfreak any reason we wouldn't want to merge this? Looks ("mostly") harmless to me.

laarmen:
It adds a new configuration to support, and thus, potential brokenness.
It's never fun to have potential users come in and saying "oh it looked
like X is supported, but I tried to run it and it failed". If the currently
"supported" platforms were all in a good shape it'd be fine, but I think we
can simply look at how often the OSX builds are silently broken.

I'm not saying we should absolutely not merge this, I'm just answering the
question of why you might want not to :-).

fluffyfreak:
It looks safe to merge, can't impact on anything existing, and means that runlevel5 can dig into their other issues.
I say merge it, I was going to earlier but I'm working.

fluffyfreak (approved):

Web-eWorks:
> There are quirks with graphics when the game starts which I will see if I could fix them or not in a separate PR

That's likely due to `scenegraph/Serializer.h` - it assumes hardware encoding is entirely little-endian for performance reasons (and because I couldn't find a decent cross-platform way to handle endianness and specifically floating-point value endianness).

This PR looks good however, good luck on getting Pioneer to work on PPC!

runlevel5:
> That's likely due to scenegraph/Serializer.h - it assumes hardware encoding is entirely little-endian for performance reasons (and because I couldn't find a decent cross-platform way to handle endianness and specifically floating-point value endianness).

I would try to get it up running with ppc64le because the industry has moved away from the big endian.

runlevel5:
Good news, game runs perfectly 

Screenshot from 2021-02-26 02-16-29
Screenshot from 2021-02-26 02-16-39
Screenshot from 2021-02-26 02-18-23

PR #5151, opened 2021-03-10
Build with C++17, remove nonstd::string_view

Web-eWorks:
It's finally time - I got fed up with how difficult it is to implement compile-time hashing in C++11, so I might as well merge this branch in.

I make no guarantees this runs on linux or compiles on MSVC; the branch has been sitting for two months but it worked then. Now I'll go back to rewriting the renderer... *mutters into the wind*

This should be good to go, though I'd prefer if fluffy, TheEG et. al. can take a look and make sure it builds properly on our different development environments. I haven't updated any documentation, but this will require at least Clang 5 / GCC 7 to build, potentially even higher versions as we add more C++17 libraries.


Web-eWorks:
Will merge changes from #5156 into this branch so it can be merged sometime over the weekend.

fluffyfreak:
I was trying to review/test it and applying all those changes thinking: "_I'm sure that I've done this before_" 🤣

I'll wait until you've merged them over instead of redoing them, ugh I am so tired tonight.

Web-eWorks:
Changes have been merged, the VS solution needs touchup to compile as c++17, but otherwise good to go.

fluffyfreak:
@Web-eWorks just pushed the VS2019 fixes directly. So it should be good to go whenever you like 👍

Web-eWorks:
> @Web-eWorks just pushed the VS2019 fixes directly. So it should be good to go whenever you like +1

Excellent! Will merge once CI finishes running.

impaktor:
@fluffyfreak we / @pcercuei and @The-EG were talking about updating the cmake and build scripts such that VS files will be nuked for good. My understanding is, after this update, it will be a one-click change needed in VS, then it's good to go. Just wanted to give you a heads up, and an opportunity to throw a fit :)

fluffyfreak:
> @fluffyfreak we / @pcercuei and @The-EG were talking about updating the cmake and build scripts such that VS files will be nuked for good. My understanding is, after this update, it will be a one-click change needed in VS, then it's good to go. Just wanted to give you a heads up, and an opportunity to throw a fit :)

@impaktor Sure, propose it via a PR I guess and give me a chance to test it out. I've had issues with the CMake builds before and never understood modifying them.

PR #5156, opened 2021-03-25
Renderer Rewrite

### LINKED ISSUE #4969: Shields visible on Main Menu after dying in combat
### Observed behaviour
The shields show up on the model spinner in the Main Menu after you die in combat, and had shields. Reported by Confed_ on Discord.
shield
unknown

### Expected behaviour
Shields shouldn't show up in the Main Menu

### Steps to reproduce
Start a new game, buy some shields, let them charge, take off and piss off the police by shooting around a bunch. Then wait until they shoot you to pieces, and the game returns to the Main Menu. Shields show up, and not only for the ship you had. 

Master from 2020 Oct. 1, Happens on 20200202 release as well.
Windows 10, 6 bit.

Web-eWorks:
**EDIT:** Tracking remaining implementation items here:

- [x] Add support for "partial buffer draws", to allow coalescing vertex data to a single buffer and dispatching multiple draws from it
- [x] Add support for scissor state to RenderStateDesc
- [x] Remove full RenderStateDesc from Material class, replace with RSD hash value
- [x] Fix use-after-free when resetting planet graphical detail level (deletion of active MeshObject).
- [x] Add Position2D vertex attribute (for imgui)
- [x] Draw ImGui using Renderer API
- [x] Promote UniformBuffer to public API
- [x] Add public Renderer interface for creating static UniformBuffers ~~and allocating per-frame UBO regions~~
- [x] Refactor Shields class to not use a global uniform buffer (causes issues with reading undefined data when drawing shields) 
- ~~Make linear [uniform, vertex] buffer allocation support generic based on a templated class~~

Fixes #4969.

-----

To make up for the amount of code involved in this PR, have a blog post:

Starting in November, I wanted to convert our shaders to use Uniform Buffer Objects instead of "a fistful of uniforms" for forwards compatibility with Vulkan (which doesn't have the concept of standalone uniforms.) As it turned out, trying to use UBOs with the rendering architecture we had was pretty bad on performance, due to the need to update a subrange of the buffer object once for every single draw. I dropped from 320 FPS in a test scene to about 180 FPS. This didn't sit well with me, given that the goal of this effort was to improve performance as well as forwards compatibility.

So I slept on the problem for a few weeks. In early December, I picked the torch back up, and made some quality of life changes to the renderer internals; things like de-duplicating code that built vertex attribute descriptors, and avoiding some unnecessary GL state changes related to vertex array objects.

I also ripped out the oldUI (Gui) system in its entirety for crimes against all GPUkind; with it gone, making radical changes to the renderer suddenly became a lot more feasible. Once that was gone, it was time for RenderState to make its final bow, and the MeshObject supplanted passing vertex+index buffers to drawing commands (in hindsight, this turned out to be the best possible change I could have made).

With RenderState no longer mutable during a frame, and MeshObjects to simplify passing data around to rendering functions, I was running out of ways to approach the problem. The main issue was now the plethora of custom Material subclasses, each with their own data layout and custom code to bind uniform data to shaders just before drawing. It made any kind of whole-frame reasoning impossible (thus the need to update the buffer object with each draw), and involved a *lot* of duplicated boilerplate before any programmer, junior or senior, could think about adding a new material for visual effects.

So I took a break for a few months, and busied myself with my day job and reading every scrap of information about rendering architecture that I could get my hands on. I knew I had to get rid of the hard-coded material types *somehow*, but I didn't know how I was going to do that. Then, after a few good articles and months of thinking, I struck upon a solution: what if every data and texture binding point was just referred to by a name and an offset, and not a C++ struct field?

Thus, the new Shader system was born. The existing Shader struct (that was used to load a single GLSL file for a shader stage) got stabbed in the back and the name stolen to refer to a new breed of Material->OpenGL interface. Materials are now written in `<name>.shaderdef` files, with a centralized pipeline for reading those files, storing data for textures, buffers, and per-draw constants, passing all of that data to the graphics API, and early support for generating and loading variants of shaders entirely at runtime based on renderer state.

No longer did `LitMultiMaterial` need to manually maintain a cache of Program objects for different numbers of stars in the scene; in fact, I made it a personal priority to completely eradicate every single subclass of `OGL::Material` and replace them with the generic ShaderDef system. The end result is far more flexible, slightly less LOC (that's my story and I'm sticking to it!), and makes writing a new material to draw custom effects for the map views as simple as authoring an extremely simple self-documenting data file and using the new material's name in the map view code.

No custom renderer-API specific subclassing needed, no need for a new entry in the `Graphics::EffectType` enum (though that's still used in a few cases to handle shader variants), and since all of the information about a material is just data coming from a file now, the engine has *free reflection information*.

This would prove to be useful, because I still hadn't solved the problem of writing buffer updates to the GPU once per draw, every single draw from here 'till doomsday. There was enough code modifying material parameters, drawing with the material, modifying the material parameters again, drawing again, etc. that I couldn't simply just chuck a tuple of `{ mesh, material, drawcall }` in a big list to run later once the uniform buffer data had been sent to the GPU.

I needed a way to save all of the state related to the drawcall *somewhere* that game code wasn't going to stomp on it, and then store it until all of the updates to the uniform buffer data had been gathered and sent to the GPU in one big batch. After some thought, the solution turned out to be pretty simple: just allocate a chunk of memory alongside the drawcall itself, and use the reflection information that I already had to pack all of the uniforms and texture bindings in that chunk of memory, then unpack them when it came time to execute the drawcall.

Once this was done (and the few bugs related to initialization and flushing of command lists ironed out), I was back up to 290 FPS. Just from caching all of the uniform data in a CPU-side buffer and sending it all at once, I had made back most of the performance I had lost from the initial implementation. But it still wasn't perfect, and there was still that matter of 30 FPS lost...

Then enters RenderDoc and Nvidia's Nsight profiler; once I had dug into the event trace for a frame, it became clear to me that I was still simply sending way too many OpenGL commands per draw; most of which I didn't even need to do. So I wrote a state cache class, and made it responsible for modifying the OpenGL state only when it actually needed to be changed.

After that, I was at 350+ FPS in the test scene! That's a good 30 FPS faster than our current way of simply throwing a bunch of data at the GPU every draw and letting the driver optimize it all behind the scenes, which I will take as a massive win. It's also had the side effect of reducing the amount of boilerplate the game code needs to write to send data to the renderer, and generally moving towards a more modern graphics API like Vulkan.

/end blog post

Now on to the technical details:
- This PR is a massive one. I don't even know what the diff is, but it's 2000+ lines of new rendering code, and at least another 2k of API and callsite changes.
- It's not suitable to be merged until the SystemInfoView and label rendering in the SectorView have been ported to PiGui (due to the removal of oldUI). I'm marking it RFC (request for comments) until that time.
- It requires a new OpenGL extension (ARB_vertex_attrib_binding, which is supported on 75%+ of GPUs "in the wild" according to [gpuinfo](https://opengl.gpuinfo.org/listreports.php?extension=GL_ARB_vertex_attrib_binding)).
- It's fairly well documented; I wrote a mini-blogpost in `src/graphics/Material.h` to explain the new material system, but I'm sure there are a few spots where I've written new code that didn't get commented adequately.
- You like C++17? I like C++17! We all like C++17!
- I'm truly sorry @fluffyfreak, but I think you're the only one left who can review the graphics code.

I'll try to pull a few minor fixes (shader normal normalization and infinite projection matrix) out to PR against master in a mergeable timeframe, but honestly I'm pretty much done with new features on this branch and I've got to focus on my day job again.

Some of the early commits during the initial implementation and later commits during the cleanup phase can be squashed together; but the meat of this is organized with one or two commits per feature to keep it at least slightly reviewable.

Feel free to review this in bite-sized pieces, but I'd appreciate as many people as possible pulling this locally and testing it; I've tested it on a few saves in different scenarios but it's entirely possible I've missed a segfault or OpenGL error somewhere. Remember to turn on `EnableGLDebug` in your config; this will log OpenGL issues you run across.

OpenGL log issues that say "recompiling shader <X> based on GL state" are fine to ignore; those aren't going to go away without further work on OpenGL program creation that I don't have the time or energy for.

Any questions, comments, etc. feel free to drop them below!


src/Background.cpp
```diff
@@ -85,6 +87,7 @@ namespace Background {
 	{
 	}
 
+	static size_t s_texture0Name = Graphics::Renderer::GetName("texture0");
```
Web-eWorks, line 90: This was intended to be a C++17 `constexpr` string hashing function inline instead of storing name hashes as static variables; I started writing this under C++11 and never found a good hash library to integrate. Work for the future.

data/shaders/opengl/sgaussian.glsl
```diff
@@ -0,0 +1,35 @@
+// Copyright © 2008-2021 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+// Code adapted from https://therealmjp.github.io/posts/sg-series-part-2-spherical-gaussians-101/
+
+struct SGaussian {
```
Web-eWorks, line 6: I am a little unsure as to which license to use for this code; I could not find a license statement on the blog so I took the examples to be public domain / uncopyrighted and considered the GLSL version to be enough of a derivative work to qualify for the Pioneer copyright.

Web-eWorks:
Right... there are some issues due to MSVC weirdness (`std::basic_string_view<char>::const_iterator` isn't an actual pointer? How nice.) and at least one issue where I build with profiler enabled and the build machine doesn't, causing some bad code to be missed. I'm going to deal with those... later.

impaktor:
That's quite some work you've pulled off, I gather. Was an interesting read, even for me, who don't know anything about rendering.

> It requires a new OpenGL extension

Then we might do one (several) more releases prior to this being merged, so people with old computers can use those.

> I'm truly sorry @fluffyfreak, but I think you're the only one left who can review the graphics code.

Perhaps someone in the old team (@robn, @johnbartholomew, @Brianetta, @laarmen) still cares to throw out general thoughts and opinions, if so, here's your chance!

> Some of the early commits during the initial implementation and later commits during the cleanup phase can be squashed together;

I'm not going to git-fascist this one :)

> Feel free to review this in bite-sized pieces, but I'd appreciate as many people as possible pulling this locally and testing it;
> Remember to turn on `EnableGLDebug` in your config

I just found out I have next Thursday & Friday free, for Easter. I'll play with it then, I hope

I applaud your effort!

robn:
Great read. I am long enough out of the game that I don't think I remember enough Pioneer and OpenGL generally to be able to review this. But your approach sounds right, and you've got rid of a few warts that I do remember, and well, its pretty hard to argue with results!

> So I took a break for a few months, and busied myself with my day job and reading every scrap of information about rendering architecture that I could get my hands on.

Oh boy, I remember that period of my life...

:D

fluffyfreak:
Wow, well now I know what I'm ding this weekend! 🤣

I'll dig in starting tomorrow and try to see if it builds in MSVC various ways, this might be the final straw for the manual maintenance of the project files.

Web-eWorks:
> Then we might do one (several) more releases prior to this being merged, so people with old computers can use those.

@impaktor this extension is old enough that I'm 99% certain it's not going to actually prevent anyone from playing Pioneer; the extensions needed for floating-point depth buffers and the reverse-Z implementation are actually supported by less devices than support this new extension. However, I'm amenable to adding a check in an interim release (possibly when #5157 is merged) that warns users if their hardware will not support it.

> I just found out I have next Thursday & Friday free, for Easter. I'll play with it then, I hope

Enjoy your Easter, no rush! There aren't any major end-user changes in here, just a whole lot of quality-of-life improvements for developers. :smile:

> Great read. I am long enough out of the game that I don't think I remember enough Pioneer and OpenGL generally to be able to review this. But your approach sounds right, and you've got rid of a few warts that I do remember, and well, its pretty hard to argue with results!

@robn No worries, thank you for popping in and having a look! This effort stands on the shoulders of giants here, and even though I horribly defamed oldUI (seriously, an individual vertex buffer upload and draw call for each side of a button's indent?), none of this would be possible without the efforts of those who worked on the codebase long before I even knew about it.

> Oh boy, I remember that period of my life...

Heh, I wouldn't trade it for the world. I feel like I've learned so much more than just how to structure a material system from all of the research I did on this project.

> Wow, well now I know what I'm ding this weekend!
> I'll dig in starting tomorrow and try to see if it builds in MSVC various ways, this might be the final straw for the manual maintenance of the project files.

@fluffyfreak Thanks man! MSVC is mostly just differences in the standard library; I wasn't aware that treating the string_view iterator as a pointer was out-of-spec, but I don't really have the energy to go mess with this.

A lot of this stuff is going to translate (eventually) into improvements to the terrain texturing WIP PR, as I'll be able to experiment much more easily with sending better data... and finally moving the depth fog out of the terrain shader and putting it in an atmosphere post-process shader instead.

fluffyfreak:
@Web-eWorks hopefully you can see it but I created a PR with my changes from loading it in vs2019
https://github.com/Web-eWorks/pioneer/pull/1

Mostly just cleanup and making it work with vs2019.

fluffyfreak:
When it's shutting down and release references it seem like the/a material is releasing a lot of times so it asserts. Might need checking out.

Web-eWorks:
> When it's shutting down and release references it seem like the/a material is releasing a lot of times so it asserts. Might need checking out.

I uncommented the code that deletes shader programs on shutdown, I have a feeling that's part of the issue. Pioneer in general has a lot of shutdown issues, though I will admit that I've introduced one or two. Will look into that.

Web-eWorks:
After the above commit, I haven't been able to encounter a crash when shutting down; looks like just a free-after-program-shutdown issue. This is only a monkeypatch on the real problem of *all of the static state that we use everywhere*, but that has far deeper architectural concerns than I really want to get into with this PR.

Web-eWorks:
Marking this as work-in-progress as the feedback I've gotten has been generally positive. Still have a few more priority things I need to do before opening it up to full review, but it's getting there.

- [ ] Add support for "partial buffer draws", to allow coalescing vertex data to a single buffer and dispatching multiple draws from it
- [ ] Add support for stencil buffer to RenderStateDesc
- [ ] Remove full RenderStateDesc from Material class, replace with RSD hash value
- [ ] Add Position2D vertex attribute (for imgui)
- [ ] Promote UniformBuffer to public API
- [ ] Make linear [uniform, vertex] buffer allocation support generic based on a templated class
- [ ] Refactor Shields class to not use a global uniform buffer (causes issues with reading undefined data when drawing shields)

Web-eWorks:
After a bit of thought and hacking in support for 16-bit index buffers, I've managed to port the ImGui rendering code to our Renderer API instead of issuing raw OpenGL commands. Not only is the resulting code much smaller, it's also slightly more efficient about buffer uploads *and* it's easier to reason about and extend in the future if we need to.

In the process, I found and fixed a small issue where we were never deleting old versions of font textures when we recreated them, which would cause a small VRAM leak when playing with non-Latin languages as the font texture has new glyphs added to it.

pcercuei:
Working on a small handheld with a tiny GPU... I concur that batching draw calls is the key to a good performance. Good job!

Web-eWorks:
I... appear to have actually managed to implement everything I had outstanding on my TODO list that touched the internals of the renderer. I'm still slightly in shock, but I'm going to go ahead and open this up to wider review for @fluffyfreak and any other interested parties to start reviewing the code quality and testing this against "real world" hardware.

Over the course of the next few weeks I'm going to try to rebase this against master and clean up the few redundant commits in the history; otherwise, I consider this PR to be "done" insofar as its original scope was intended. There is a little bit more work needed to refactor Shields and port the SectorView labels to pigui, but with the addition of another topic branch porting the SystemInfoView, there is only a small feature-level difference between this PR and `master`.

Web-eWorks:
If it is worthwhile, I can separate out the series of commits that deletes oldUI into a separate PR, to get that in earlier and better facilitate review of this PR. I leave that at the discretion of anyone who would like to tackle implementing SectorView labels.

Web-eWorks:
All compilation errors except the issue caused by missing files in the VS solution have been fixed; this branch should (now) be fully testable on both Windows and Linux. I'm going to elect to *ignore* the clang-format error, as it's frankly wrong and I haven't found the magic incantation to make clang-format do what the manual says it's should be doing given the config file it's using.

fluffyfreak:
I will attempt to review this monster 288 file PR... well, by Monday, probably best to set a deadline 🤣

src/graphics/opengl/CommandBufferGL.h
```diff
@@ -0,0 +1,135 @@
...
+
+			bool IsEmpty() const { return m_drawCmds.empty(); }
+			void Reset();
+
+		private:
+			friend class Graphics::RendererOGL;
+			CommandList(Graphics::RendererOGL *r) :
+				m_renderer(r)
+			{
+				m_drawCmds.reserve(32);
```
Gliese852, line 91: I wonder why 32?

Web-eWorks: STL vectors typically allocate in increasing powers of two. Reserving 8 elements is usually considered the "magic number" to prevent spurious allocations without also reserving too much memory, but generally speaking I expect *at least* 32 commands to be written to a command list (to make up for the cost of submitting the command list). This could honestly be bumped up higher, as all 700+ draw commands are recorded in the same command list, but I eventually want to separate the scene draw into multiple command lists so it stays at 32 for now.

impaktor: I suppose stuff like that can warrant a line or two as comment in the code.

Gliese852: @impaktor imhi it doesn't seem to matter much, because after the very first frame the capacity of this vector will greatly increase and will no longer decrease, if I'm not mistaken

fluffyfreak:
Getting the MSVC files updated was just adding 2 or 3 files to the project. 
Building just took including `Color.h` in `VertexBuffer.cpp`.
Getting it to run is proving trickier as it fails a lot of asserts during initialisation.

Think these issues are mostly because it's not been built/run with asserts enabled.

fluffyfreak:
@Web-eWorks Ok, GitHub is being weird and won't let me PR my changes:
Visual Studio project changes:
https://github.com/pioneerspacesim/pioneer/commit/8eb76e2c5eb297b9190c7297c18792c39785fda2
Compilation and some runtime fixes:
https://github.com/pioneerspacesim/pioneer/commit/dbbb81fe96b533b769da4f8cbb61f6071819e6dd

I had to comment out a bunch of asserts in `matrix4x4.h` to get it to run because the asserts are probably wrong now that we use inverse-z.

However even with that, or with running in Release, the game loads to the start screen and there is no UI. No loading bar or text either.

Web-eWorks:
That's strange, I could have sworn I built it in debug mode at least a few times during development and didn't hit any asserts. I'll look into the imgui issue though, there shouldn't be a code path that completely avoids rendering it. If you want to toss some breakpoints in, the new imgui rendering code is in src/pigui; my best guess is that something's not fully configured on windows and imgui draw commands are failing the coarse clipping test.

fluffyfreak:
It appears to be running the `InstanceRenderer::RenderDrawData` but I'm just not seeing the drawcalls in RenderDoc and nothing appears onscreen. It's quite weird.

I do see the IMGUI mouse cursor.

Web-eWorks:
The imgui cursor is actually the system cursor! We purposefully use SDL facilities to use the OS cursor in our style to avoid the pitfalls of a software cursor.

I'd appreciate it if you step through the RenderDrawData function and let me know where the if statements are failing; it's copied pretty much verbatim from the existing OpenGL implementation and *shouldn't* be breaking like this.

fluffyfreak:
I've stepped through the `FlushCommandBuffers` a few times and it all seems fine 🤷‍♂️ it doesn't appear to be "failing" in any conventional sense. I have to head out now but I'll get back to debugging it asap.

Web-eWorks:
@fluffyfreak good catch, I thought I fixed all of the errors with the 16-bit index buffer "feature" but it looks like some others slipped through. If it's adding draw commands to the command buffer properly in RenderDrawData, then I'm not sure what's going on to prevent those commands from being sent to OpenGL.

Web-eWorks:
One thing I noticed with RenderDoc is that ImGui draw calls sometimes don't appear to be affecting the output buffer in RenderDoc, but actually are taking effect. The only way to check if the drawcalls are being correctly generated is to check which input texture is being used; if it's an alpha-tested font texture then they're being correctly generated and not showing up on screen for some reason.

Web-eWorks:
> I've stepped through the `FlushCommandBuffers` a few times and it all seems fine man_shrugging it doesn't appear to be "failing" in any conventional sense. I have to head out now but I'll get back to debugging it asap.

Found the problem; your code defines POSITION2D as attrib 5, although it's meant to share slot 0 with the POSITION attribute. This is why you don't see anything; you inadvertently fed (0,0) to all vertex positions! :smile:

fluffyfreak:
The SystemOverview doesn't seem to render? 
Otherwise looking good from testing so far 👍

Web-eWorks:
> The SystemOverview doesn't seem to render?
> Otherwise looking good from testing so far +1

SystemOverview was purposefully removed; the tab item is leftover and removed in a follow-on branch that reimplements it as an atlas view mode for the SystemView.

src/graphics/opengl/RendererGL.cpp
```diff
@@ -1202,17 +1074,50 @@ namespace Graphics {
...
+
+	OGL::UniformLinearBuffer *RendererOGL::GetDrawUniformBuffer(Uint32 size)
 	{
-		// empty since viewport handling is now external, evaluate if renderer will need to save any custom state
+		for (auto &buffer : m_drawUniformBuffers)
+			if (buffer->FreeSize() >= size)
+				return buffer.get();
+
+		auto *buffer = new OGL::UniformLinearBuffer(1 << 20);
+		buffer->IncRefCount();
```
Gliese852, line 1113: Why not do
`std::vector<RefCountedPtr<OGL::UniformLinearBuffer>> m_drawUniformBuffers` ?
It seems that the entire overhead (I mean `mutable std::atomic<int> m_refCount`) is still with us.
But at least you don't have to manually increase the smart pointer germ counter through the raw pointer.

Web-eWorks: Good catch with the update; I didn't get a chance to comment earlier, but this is in fact a left-over "optimization" from an earlier version when Materials stored `RefCountedPtr<UniformBuffer>` instead of a raw pointer.
Ownership of the linear UBO should stay entirely with the Renderer (hence the API using raw pointers), but the ref count increment was to prevent it from being accidentally deleted by Material constructing a RefCountedPtr from a raw pointer, as is the general paradigm with objects returned from the Renderer interface.
This IncRefCount() line can *most likely* be deleted without impacting the code at all, but I'd have to swap over to the branch and trace it just to be sure.

Web-eWorks:
Cherry-picked in an ImGui upgrade to resolve issues @Gliese852 has been encountering in rewriting the SectorView labels.

Web-eWorks:
With this last commit, this branch should be in good shape for a merge. I've refactored the Shields class to no longer have a static global data structure, and instead cache a material instance for each copy of the Shields object. This avoids several cases where the shield material would read from undefined data, and directly fixes #4969.

Going to hold off on merging until #5227 is in and a release has been cut, but this should otherwise be good to go.

fluffyfreak:
For some reason @Web-eWorks your repository doesn't show up as a fork of Pioneer so I can't PR this commit to your branch https://github.com/fluffyfreak/pioneer/commit/750951bf4a963fa9f1283b9df386365f1a73b418

Web-eWorks:
Roger that, will pull it in.

PR #5157, opened 2021-03-27
Fix GasGiant clipping and bright spots on models

### LINKED ISSUE #5145: Gas Giants Do Not Render Properly
### Observed behaviour
Gas giants will completely disappear when viewed at the center of the screen. They are only visible when they are viewed at the edge of the screen. Basically, the entire gas giant can never be viewed because when the player attempts to bring the whole body into view, the gas giant will disappear entirely. Here is a screenshot that illustrates this behavior.

Disappearing Gas Giant

### Expected behaviour
Gas giants should not disappear at all. 

### Steps to reproduce
Visit any gas giant. 

ver 20210203 (142693b), Ubuntu 20.04 LTS
OpenGL.txt file: https://pastebin.com/g27cQT78

Web-eWorks:
As promised, this was pulled out of #5156 to be applied to master. This PR contains a fix for planets and gas giants clipping out of the view, and also a fix for the unreported issue of persistent bright-spot acne on our models, most noticeable on the fuselage of the Pumpkinseed (but prevalent pretty much everywhere).

Closes #5145.

These screenshots are identical loads of the same save (including camera position/zoom), showing the difference with this change applied.

## Before:
Screenshot from 2021-03-26 22-37-23
Screenshot from 2021-03-26 22-40-22

## After:
Screenshot from 2021-03-26 22-37-56
Screenshot from 2021-03-26 22-41-00

### DIFF
--- a/data/shaders/opengl/multi.frag
+++ b/data/shaders/opengl/multi.frag
@@ -88,6 +88,8 @@ void main(void)
 #else
 	vec3 vNormal = normal;
 #endif
+	vNormal = normalize(vNormal);
+
 	//ambient only make sense with lighting
 	vec4 light = scene.ambient;
 	vec4 specular = vec4(0.0);
--- a/src/Camera.cpp
+++ b/src/Camera.cpp
@@ -72,7 +72,7 @@ void CameraContext::EndFrame()
 
 void CameraContext::ApplyDrawTransforms(Graphics::Renderer *r)
 {
-	r->SetPerspectiveProjection(m_fovAng, m_width / m_height, m_zNear, m_zFar);
+	r->SetProjection(matrix4x4f::InfinitePerspectiveMatrix(DEG2RAD(m_fovAng), m_width / m_height, m_zNear));
 	r->SetTransform(matrix4x4f::Identity());
 }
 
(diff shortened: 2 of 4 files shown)


data/shaders/opengl/multi.frag
```diff
@@ -88,6 +88,8 @@ void main(void)
 #else
 	vec3 vNormal = normal;
 #endif
+	vNormal = normalize(vNormal);
```
fluffyfreak, line 91: Damn, all this time and we just weren't normalizing the normals 🤦‍♂️ ugh, I must have seent aht hundreds of times!

fluffyfreak (approved):
Good fixes 👍

data/shaders/opengl/multi.frag
Web-eWorks: Yeah, I noticed when I was turning off the lights to test spherical gaussian indirect (and yes, they're in the other branch for you to play with :smile:) that most of the light acne disappeared; for the longest time I thought it was just floating point inaccuracy with vertex interpolation, but it turned out to just be out-of-range normals. :facepalm:

PR #5164, opened 2021-04-18
Negotiate amount of cargo

robothauler:
Requested by @impaktor some time ago for the cargo run module.
Player with smaller ships can now choose a smaller amount of cargo. The advertisement remain in the BB with the rest of the cargo.

### DIFF
--- a/data/lang/module-cargorun/en.json
+++ b/data/lang/module-cargorun/en.json
@@ -135,6 +135,18 @@
     "description": "",
     "message": "Come back when you're more experienced."
   },
+  "DENY_AGAIN_1": {
+    "description": "",
+    "message": "Sorry, we already have a deal. However, I would be happy to do business with you again after a successful delivery."
+  },
+  "DENY_AGAIN_2": {
+    "description": "",
+    "message": "Excuse me sir, please complete our existing contract first. Then we can negotiate another delivery."
+  },
+  "DENY_AGAIN_3": {
+    "description": "",
+    "message": "Sir, please deliver my cargo first. Then I will do more business with you."
+  },
   "DIAMONDS": {
     "description": "custom cargo",
     "message": "Diamonds"
@@ -199,6 +211,10 @@
     "description": "custom cargo",
     "message": "Furniture"
   },
+  "GO_BACK": {
+    "description": "",
+    "message": "Go back."
+  },
   "GREENHOUSES": {
     "description": "custom cargo",
     "message": "Greenhouses"
@@ -213,35 +229,35 @@
   },
   "HOWMUCH_1": {
     "description": "regarding mass, always more than 1t",
-    "message": "{amount} metric tons."
+    "message": "Yes it is. The total amount is {amount} metric tons."
   },
   "HOWMUCH_2": {
     "description": "regarding mass, always more than 1t",
-    "message": "{amount}t"
+    "message": "Yes, we can split the cargo. I have {amount}t in total."
   },
   "HOWMUCH_3": {
     "description": "regarding mass, always more than 1t",
-    "message": "The mass is {amount}t."
+    "message": "Yes. The entire mass is {amount}t."
   },
-  "HOWMUCH_4": {
+  "HOWMUCH_NO_1": {
     "description": "regarding mass, always more than 1t",
-    "message": "There are {amount}t."
+    "message": "No, sorry. It is not possible to split the cargo. You have to transport the whole {amount} tons at once."
   },
   "HOWMUCH_SINGULAR_1": {
     "description": "This string is only used for 1 metric tons (singular)",
-    "message": "It's only {amount}t."
+    "message": "No it is not. It's only {amount}t."
   },
   "HOWMUCH_WHOLESALER_1": {
     "description": "",
-    "message": "There are {amount}t of {cargoname}."
+    "message": "There are {amount}t of {cargoname}. I can offer the following freight runs:"
   },
   "HOWMUCH_WHOLESALER_2": {
     "description": "",
-    "message": "{amount}t {cargoname}"
+    "message": "{amount}t {cargoname} can be divided as follows:"
   },
   "HOW_MUCH_MASS": {
     "description": "",
-    "message": "How much mass?"
+    "message": "Is the amount of cargo negotiable?"
   },
   "HOW_SOON_MUST_IT_BE_DELIVERED": {
     "description": "",
@@ -335,9 +351,17 @@
     "description": "custom cargo",
     "message": "Neptunium"
   },
+  "OFFER": {
+    "description": "",
+    "message": "Transport {amount}t for {reward}"
+  },
   "OK_AGREED": {
     "description": "",
     "message": "OK, agreed."
+  },
+  "OK_WE_AGREE": {
+    "description": "",
+    "message": "OK, we agree on {amount}t for {reward}."
   },
   "PICKUP_FROM": {
     "description": "from a location",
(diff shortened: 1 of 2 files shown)


data/lang/module-cargorun/en.json
```diff
@@ -213,35 +217,35 @@
...
   },
   "HOWMUCH_3": {
     "description": "regarding mass, always more than 1t",
-    "message": "The mass is {amount}t."
+    "message": "Yes. The entire mass is {amount}t."
   },
   "HOWMUCH_4": {
     "description": "regarding mass, always more than 1t",
-    "message": "There are {amount}t."
+    "message": "There are {amount}t. We can split the cargo run if you want."
```
impaktor, line 232: Just a personal opinion, I think HOWMUCH_4 and HOWMUCH_2 are basically the same, i.e. a player seeing two adverts, with each flavor will probably not notice they're technically different. (Generally, I'd opt for having fewer flavours if they're this similar)
Action: none required, I just felt like pointing out my view (which could be wrong) on this subject.

data/lang/module-cargorun/en.json
```diff
@@ -213,35 +217,35 @@
...
   },
   "HOWMUCH_WHOLESALER_2": {
     "description": "",
-    "message": "{amount}t {cargoname}"
+    "message": "{amount}t {cargoname} can be divided as follows:"
   },
   "HOW_MUCH_MASS": {
     "description": "",
-    "message": "How much mass?"
+    "message": "Is the amount of cargo negotiable?"
```
impaktor, line 248: I haven't playtested, and maybe this is needed for the code to enter a menu system, but if not, one could simply have the player, instead of asking to negotiate, make a counter-deal right away: 
"Unfortunately I can only take:
- [x] 10t
- [ ] 100t 
- [ ] 1000t"

data/modules/CargoRun/CargoRun.lua
```diff
@@ -367,37 +372,53 @@ local onChat = function (form, ref, option)
...
-				howmuch = string.interp(l["HOWMUCH_" .. Engine.rand:Integer(1,getNumberOfFlavours("HOWMUCH"))],
-					{amount = ad.amount})
+			if ad.wholesaler then
+				howmuch = string.interp(l["HOWMUCH_WHOLESALER_" .. Engine.rand:Integer(1, getNumberOfFlavours("HOWMUCH_WHOLESALER"))], {
+					amount    = ad.amount,
+					cargoname = ad.cargotype:GetName()})
 			else
-				howmuch = string.interp(l["HOWMUCH_SINGULAR_" .. Engine.rand:Integer(1,getNumberOfFlavours("HOWMUCH_SINGULAR"))],
+				howmuch = string.interp(l["HOWMUCH_" .. Engine.rand:Integer(1,getNumberOfFlavours("HOWMUCH"))],
 					{amount = ad.amount})
```
impaktor, line 385: In general: using the same string twice "HOWMUCH_" and "HOWMUCH", I'm half thinking if this might be made more generic, but I'm not suggesting any change. Just thinking out loud.

impaktor (approved):
Cool. I forgot about this PR (sorry).
I haven't tested it, but I trust it's good to go. 
My comments are just me thinking out loud, if they inspire you to want to make any changes, then notify us so we don't merge it before you've had a chance to update the PR. If not, we'll merge it as it is, next time someone with merge-power feels like merging PRs.

Thanks!

data/lang/module-cargorun/en.json
robothauler: Yes, sounds very similar. I will remove one.

data/lang/module-cargorun/en.json
robothauler: Yes, it is needed to enter the menu with the amounts.

impaktor:
I'm just trying this out now.

I wonder if it should be more clearly worded that payment scales linearly (baring round off) with mass? It's easy to see in the below example since it's multiples of $25, which is easy to do head math on. 

2021-05-23-140302_1600x900_scrot

But perhaps one would like the text to read: "Yes we can split the cargo. I have 9t in total, I'll pay $25 for each tonne you can transport for me"

This is an open question, as it's not obvious how much we should spell out to the player. Perhaps we _want_ them to do some thinking and actually consider if it's a good deal or bad deal to take a fraction of the total mass?

Also:
When clicking "transport Xt for $Y", you get the below:

2021-05-23-141255_1600x900_scrot

I assume the "deal" is sealed when I again click "OK, agreed", I wonder if she should summarize the deal a bit more: 
"OK, agreed: you will take Xt for $Y?"

I haven't looked at the code in detail, but they only refuse to split the cargo when it's 1t? If so, there could be those who prefer a "fill or kill" deal, i.e. you either take it all or none of it. I'm just throwing it out there.

Anyway, those are my thoughts when I tried it out, what do you/others think?
We could merge this as is, it depends on if @robothauler agrees/want to work on any of the above points.

robothauler:
To be honest, after a few days of playing around, I'm no longer satisfied with this PR. So I also think this should be improved.

impaktor:
What are you dissatisfied with? Overall I think it's good. One thought I had was if it's possible to not re-create a new advert for the remaining goods left to be shipped, but if it's possible to just update the existing advert, i.e. pick out the relevant parts of the table with the advert info and update, so just lower the remaining cargo to be shipped?

Anyway, I also don't want to stall this from going in, I've been much too slow to review this as it is.

robothauler:
> What are you dissatisfied with? Overall I think it's good. One thought I had was if it's possible to not re-create a new advert for the remaining goods left to be shipped, but if it's possible to just update the existing advert, i.e. pick out the relevant parts of the table with the advert info and update, so just lower the remaining cargo to be shipped?

The recreation with the remaining cargo works. But there are still some oddities.
- The player can gain more reputation by clicking 10 times 1t for a 10t advert
- e.g. the customer says: 1, 2, 4, 6, 9t for selection as in your example above. I have 8t available on my ship. So I click 6t and then 2t. Why not 8t at once?
- The cash difference between a 2t delivery advert and a 10t delivery advert with a 2t selection is too big IMHO

> Anyway, I also don't want to stall this from going in, I've been much too slow to review this as it is.

I also have no problem with a later improvement after merge.

impaktor:
> The player can gain more reputation by clicking 10 times 1t for a 10t advert

Good point. You could add an additional attribute to the advert "denyPlayer = false" or similar, that flips to true if the player takes the mission with "negotiated cargo", and flip it to "true" for the new advert with the remaining cargo, then deny the player snarkily: "you again? We already have a contract, and I don't want to sign more with you flipp flopping", (but differently worded). Or just check that variable when assigning reputation, so only the first version of the advert gives reputation, and the other give 0 reputation.

robothauler:
Feels much better now. I did not touch the reward. I have changed my mind. That should be the advantage of a freighter compared to a courier or fighter where cargo mass is not so important.

impaktor:
Tested it, seems to work as advertised.

For future: could be good to have some boolean `will_split_cargo` variable set to an advert, as some might only want to deal with a single transporter, and as @jimishol pointed out, it gives an incentive to have a larger ship.
But that's for another day/PR.

Thanks for the work!

impaktor (commented):
Tested it, seems to work as advertised.

For future: could be good to have some boolean will_split_cargo variable set to an advert, as some might only want to deal with a single transporter, and as @jimishol pointed out, it gives an incentive to have a larger ship.
But that's for another day/PR.

Thanks for the work!

impaktor:
@robothauler Feedback from the [player forum](https://spacesimcentral.com/community/pioneer/graphics-issues/#post-60225):

> I do indeed have a first feedback for you regarding the new feature (negotiating the quantity to be transported). Because the option 'how much to transport' is no longer available, you now have to use the option 'negotiate' to find out how much it is at all. Doesn't make much sense in my opinion. It would be better if it was the same as before and you put the option of negotiating in there?

PR #5166, opened 2021-04-29
Improve performance with many ships in one system

Web-eWorks:
This is primarily merging in the work I did to make the tradeships PR (#4984) feasible from a performance perspective. I've added in a few other commits from unmerged branches that made sense and cleaned up the codebase a little bit.

This prevents model animation interpolation from being run when active animations aren't being played, which was the number one performance hog with many ships in a system. It also cleans up a fair amount of C++-to-Lua overhead that was being run every `StaticUpdate` tick.

This PR also removes unneeded code in the TabView and remembers the last tab the user was on in the Comms and Info windows.

### DIFF
--- a/contrib/profiler/Profiler.cpp
+++ b/contrib/profiler/Profiler.cpp
@@ -872,6 +872,7 @@ namespace Profiler {
 	};
 
 	u64 globalStart = Timer::getticks();
+	u64 globalClockStart = Clock::getticks();
 	GlobalThreadList threads = { NULL, {0} };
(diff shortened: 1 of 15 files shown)


Gliese852:
Did you mean #4984?

fluffyfreak:
I was going to ask about this as I was looking at it the other day 👍

fluffyfreak (approved):
Good looking code as always 👍
Tested on Windows using VS2019, saving & loading, hyperspacing etc, all works good.

PR #5167, opened 2021-05-02
Warn about loading dead bodies, but handle safely.

fluffyfreak:
Save games could fail to load due to the body index being written as `(Uint32)-1`.
Handle getting an out-of-range / SDL_MAX_UINT32 value when trying to find a body by returning `nullptr` instead.

This allows savegames to load despite this issue.

The issue itself is caused by the `AICmdFlyAround` having a pointer to a `Body *m_obstructor` that it does not own and is not refcounted. This can be destroyed before/whilst saving, which fails to find the body (_'cos it's gone_), the GetIndexForBody returns `(Uint32)-1`, and upon loading it crashes.

I would like to have fixed the actual cause by this COVID19 vaccination has hit me like a bus and I feel like I've got flu.
So posting this as-is for possible merging, comments, and suggestions. Either other can fix it, or it can wait until my brain works again.

### DIFF
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -350,6 +350,10 @@ Body *Space::GetBodyByIndex(Uint32 idx) const
 {
 	assert(m_bodyIndexValid);
 	assert(m_bodyIndex.size() > idx);
+	if (idx == SDL_MAX_UINT32 || m_bodyIndex.size() <= idx) {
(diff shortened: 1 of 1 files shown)


Web-eWorks (approved):
I would prefer returning `INT32_MAX` (and in general avoiding all SDL int-types in the codebase) *but* this is good to merge now and we can convert it to standard types at a later date.

PR #5168, opened 2021-05-03
Add zone-based profiling support

Web-eWorks:
"Congratulations, you're a real profiler now!"

This PR adds support for capturing profile zone start/end events, storing them in a buffer, and writing them to disk in Speedscope Json format for inspection in [Speedscope](https://speedscope.app). This is the "missing link" in our profiling capabilities, the ability to dig into the sequential nature of what's happening in a frame and visually see what functions are taking significant amounts of time.

Screenshot from 2021-05-02 23-14-46

In the process of doing this, I've removed appx. 2000 'unhelpful' profiling calls from the startup phase; these are functions that on average consume <1 microsecond per call and overall have very little contribution to the overall time. I've also dug in a bit and removed 4000+ of these small profiling calls from the average in-game frame capture, which helps to not only reduce overhead (even though the overhead from a profiler enter/exit pair is on the order of 100ns) but also keeps the file size of the resulting profile reports and event data smaller.

Eventually I'd like to run the zone profiler always-on in the background to be able to dynamically inspect data about multiple threads across the last few seconds, but at the moment the worst-case frame (when physics is running) is ~7000 enter-exit pairs per frame. The default thread buffer size of 4MB can only hold 18 frames of history at that size, which is an indication that I'll either need to move the application code causing those events off the main thread into worker threads (divide the storage requirements), improve the size of profiler events, reduce the number of profiler event calls during "normal" conditions (I can easily drop another 4000 semi-useful profiles from the physics system, see the 1000s of tiny profile zones under Geom::Collide), or most likely all three.

~~**Future-compatibility note**: the zone-based profiling uses the "packed pointers" trick made possible by the AMD64 architecture only using 47 of the 64 bits in a pointer; I've added in a (hopefully) complete macro to detect incompatible platforms (basically anything that's not x86 / amd64); if someone actually intends to port Pioneer to an incompatible platform (e.g. PPC64LE) an alternative struct definition for profile event data will need to be added to Profiler.cpp.~~

The above is no longer accurate; this profiler implementation should be fully portable across our prospective platforms!

### DIFF
--- a/contrib/profiler/Profiler.cpp
+++ b/contrib/profiler/Profiler.cpp
@@ -303,33 +303,29 @@ namespace Profiler {
 	=============
 	*/
 
-#if !defined(__LP64__) && !defined(__ILP32__) && !defined(_M_AMD64) && !defined(__i386__) && !defined(_M_IX86)
-	#error "Unknown processor architecture detected; zone-based profiling cannot work on this architecture"
-#endif
-
 	enum ZoneType : u8 {
 		ZoneEnter = 0,
 		ZoneExit
 	};
 
-	// Exploiting the fact that 64-bit pointers on AMD64 are restricted to the
-	// low 47 bits, with bit 48 and above copies of bit 47. This means we can
-	// stuff all needed metadata in the high 16 bits.
+	// We store relative time since epoch (generally, last profiler reset) in
+	// 56 bits, and the type of the event in the remaining 8 bits. This
+	// provides approximately 833 days of runtime before the time value is
+	// exhausted; generally speaking you will run out of memory and disk space
+	// well before that limit.
 	struct Zone {
-		u64 data : 48;
-		u64 _unused : 8;
+		u64 data;
+		u64 time : 56;
 		u64 type : 8;
-		u64 time;
 
 		Zone(ZoneType _t, void *_d, u64 _time) :
-			data((uintptr_t)(_d) & 0x0000FFFFFFFFFFFF),
-			type(_t),
-			time(_time)
+			data(u64(_d)),
+			time(_time),
+			type(_t)
 		{}
 
-		// recover an object pointer from the 48 bits of this value
 		template<typename T = void>
-		T *ptr() const { return (T *)(uintptr_t)(data | (data & (1UL << 47) ? 0xFFFF000000000000 : 0) ); }
+		T *ptr() const { return reinterpret_cast<T *>((void *)data); }
 
 		const char *str() const { return ptr<const char>(); }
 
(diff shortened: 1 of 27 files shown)


contrib/profiler/Profiler.cpp
```diff
@@ -289,6 +297,130 @@ namespace Profiler {
...
+#if !defined(__LP64__) && !defined(__ILP32__) && !defined(_M_AMD64) && !defined(__i386__) && !defined(_M_IX86)
+	#error "Unknown processor architecture detected; zone-based profiling cannot work on this architecture"
+#endif
+
+	enum ZoneType : u8 {
+		ZoneEnter = 0,
+		ZoneExit
+	};
+
+	// Exploiting the fact that 64-bit pointers on AMD64 are restricted to the
```
fluffyfreak, line 315: Is there a benefit to this that's worth it not working on other platforms?
If it's just storing an extra 16-bits of data then an extra `Uint16` doesn't seem like much overhead

impaktor:
> ...trick made possible by the AMD64 architecture only using 47 of the 64 bits in a pointer; I've added in a (hopefully) complete macro to detect incompatible platforms (basically anything that's not x86 / amd64)

Perhaps something @runlevel5 would be interested in, like PPC64LE from #5141?

contrib/profiler/Profiler.cpp
Web-eWorks: There are several benefits - first of all, this allows me to store a pointer, an index, and a 64 bit time value in 16 bytes; multiplied by 26,000 (the average number of profile zones in a startup trace), that's 400KB of data. Doing it this way also preserves the 8-byte alignment of the type, which I consider to be more important than breaking alignment to add an extra 16 bits of data storage to the struct. Were I to unpack the event index, I'd need to bump the size of the struct to 24 bytes, making that same startup trace take 600KB of the nominally fixed-size memory buffer.

I only really need 3 bits of the pointer even with future expansion in mind; I've reserved the upper 16 bits however because they're unused on our main target platforms. As part of future work I can look into storing time-since-startup instead of time-since-epoch in the time value and moving these bits into that value instead of the pointer which should sidestep the issue; for the initial implementation I wanted to use the known target of 48-bit pointers rather than guesswork as to which bits in the rdtsc output are likely to be unused.

Web-eWorks: Ultimately I managed to avoid this entirely by moving the epoch calculations into event recording instead of during the serialization step. Now the Zone structure stores a full-size pointer data value, and packs epoch-relative time into 56 bits and the zone type into 8 bits. This provides some room for future expansion, as I highly doubt anyone has enough disk space for the 833 days of runtime that 56 bits makes possible.

Web-eWorks:
Updated this branch to be portable across platforms and no longer use the pointer trick. I'm going to consider this a candidate for merge as I'd like to be able to use it to further dig into startup times for #4951.

PR #5197, opened 2021-05-25
Switch to Libera.Chat in Readme.md

Web-eWorks:
With the sale and subsequent takeover of freenode, the development team no longer uses that IRC network for collaboration. Update links to point to libera.chat instead.

As an additional note, http://pioneerspacesim.net/irc needs to be updated to redirect to a KiwiIRC join link for the new irc channel instead of pointing to freenode's webchat app.

### DIFF
--- a/COMPILING.txt
+++ b/COMPILING.txt
@@ -35,7 +35,7 @@ builds and is otherwise not actively maintained. Read the section below for
 details on using it for your platform.
 
 If you're having trouble compiling, please ask in #pioneer on
-irc.freenode.net. Most users are available to answer during CET daytime, but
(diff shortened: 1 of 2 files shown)


impaktor:
I don't know if we formally have changed yet, but COMPILING.txt also mentions freenode.

wiki and https://github.com/pioneerspacesim/pioneer-hugo also need updating.

Brianetta:
I just kind of assumed that that migration would take place.

impaktor:
@Web-eWorks Could you also please add an edit of COMPILING.txt, and then we can merge this ASAP.

Issue #5200, opened 2021-05-28
Floating point exception while cruising to destination


tropicalpioneer:
### Observed behaviour
While travelling to destination, the game crashes with output:

```
Game::LoadGame('bluenose mission')
savefile version: 87
Creating new galaxy generator 'legacy' version 1
Clearing and re-using previous Galaxy object
StarSystemCache: misses: 0, slave hits: 0, master hits: 0
SectorCache: misses: 101, slave hits: 0, master hits: 1
Stars picked from galaxy: 183416
Final stars number: 500000
Hosobi A:
    height fractal: Ellipsoid
    colour fractal: StarG
    seed: 2097508729
Hosobi A a:
    height fractal: MountainsVolcano
    colour fractal: Desert
    seed: 139585229
Hosobi B:
    height fractal: Ellipsoid
    colour fractal: StarG
    seed: 2473717441
Hosobi B a:
    height fractal: HillsCraters2
    colour fractal: Rock
    seed: 4286993555
Hosobi B b:
    height fractal: RuggedDesert
    colour fractal: Desert
    seed: 2898156491
Hosobi B c:
    height fractal: Barren Rock 2
    colour fractal: Rock
    seed: 666147143
Hosobi A,B a 1:
    height fractal: Asteroid4
    colour fractal: Asteroid
    seed: 2851529373
Hosobi A,B a 2:
    height fractal: Barren Rock 3
    colour fractal: Rock
    seed: 4281332000
Hosobi A,B b 1:
    height fractal: Barren Rock 3
    colour fractal: Rock
    seed: 2986009267
Hosobi A,B b 2:
    height fractal: MountainsCraters2
    colour fractal: Rock
    seed: 3369114528
Hosobi A,B b 3:
    height fractal: MountainsCraters2
    colour fractal: Rock
    seed: 2489344644
Hosobi A,B b 4:
    height fractal: MountainsCraters2
    colour fractal: Rock
    seed: 694969596
Hosobi A,B b 5:
    height fractal: MountainsCraters2
    colour fractal: Rock
    seed: 1488627882
Stars picked from galaxy: 183416
Final stars number: 500000
Color Fractal name: GGSaturn
Color Fractal name: GGSaturn
Floating point exception
````

### Steps to reproduce
load game
[issue floating point exception.zip](https://github.com/pioneerspacesim/pioneer/files/6561635/issue.floating.point.exception.zip)
fastforward time
crash

__My pioneer version (and OS):__
[...]

PR #5204, opened 2021-06-03
Fix the floating point crash

Damjan94:
deals with the issue #5200

I say "deals" because I have no idea why the value was that large in the first place...

### DIFF
--- a/src/perlin.cpp
+++ b/src/perlin.cpp
@@ -53,7 +53,7 @@
  * Eliot
  */
 
-inline int fastfloor(const double x) { return int(x > 0 ? x : x - 1); }
+inline int fastfloor(const double x) { return long(x > 0 ? x : x - 1); }
 
 //static double dot( const int* g, const double x, const double y ) { return g[0]*x + g[1]*y; }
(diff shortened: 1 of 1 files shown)


Damjan94:
screenshots of the sips that were causing the crash
image

image

image


Doesn't seem to have any relevance though....

Web-eWorks:
It's an issue with calculating terrain height, but I find it interesting that it causes an fpexcept crash on your computer; I personally have never had that happen. What hardware are you running, by chance?

Ideally the fix is to adjust terrain generation to avoid ridiculous values like that, but I suppose the long->int truncation is an acceptable patch for right now.

Damjan94:
I use amd fx 8350
are you not getting a crash when fast forwarding the save file(before this "cast to long" patch)? I got it roughly 80% of the time...

Web-eWorks:
> I use amd fx 8350
> are you not getting a crash when fast forwarding the save file(before this "cast to long" patch)? I got it roughly 80% of the time...

No! I've actually never gotten a crash from that, no matter how I build the repo, although I'm on an Intel i5 4690k. Nor have I seen this reported before, which makes me think it's something to do with your specific processor generation.

However, I'm happy to pull this in as it shouldn't cause any major differences in behavior on other unaffected platforms.

PR #5206, opened 2021-06-06
Fix build with WITH_SYSTEM_LIBLUA=ON

pcercuei:
Don't depend on hardcoded paths to the Lua library in contrib/.


The-EG:
Maybe having an action to build scenarios like this would be useful?

It wouldn't keep breaking changes from happening necessarily, but it would point them out. This is why I made the build-vs-solution action.

pcercuei:
Looks like I broke the WITH_SYSTEM_LIBLUA=OFF case, so don't merge now.

@The-EG Yes, that would be great. This issue happened before and if there is no way to detect these issues early, they will keep happening.

pcercuei:
OK, problems fixed.

PR #5207, opened 2021-06-06
Add build check for additional build options

The-EG:
An extra action to test different build scenarios, such as the one mentioned in #5206 . Some of these are fairly easy to break without realizing it, this will help identify when it happens sooner.

Currently just -DUSE_SYSTEM_LIBLUA=ON, but others can be added in the future

### DIFF
--- a/.github/workflows/additional-build-checks.yml
+++ b/.github/workflows/additional-build-checks.yml
@@ -1,0 +1,44 @@
+# Test all pull requests to ensure they build
+
+name: Test Additional Build Options
+
+# Controls when the action will run.
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    paths:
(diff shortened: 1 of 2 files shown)


The-EG:
@pcercuei Does this look like the proper usage? https://github.com/pioneerspacesim/pioneer/pull/5207/checks?check_run_id=2758380736#step:4:1

Any other flags that would be recommended?

*note*: I'll clean up the commits and remove the change to the cpp file before it's ready to merge.

pcercuei:
@The-EG you can add `-DUSE_SYSTEM_LIBGLEW=ON` too, two birds with one stone.

impaktor:
Should this not have been merged?
Just wonder why it was closed.

PR #5209, opened 2021-06-10
.vscode: Add launch settings for VS Code

pcercuei:
With this, it is now possible to launch Pioneer directly from VS Code on
Windows by simply pressing F5.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,5 +84,4 @@ result
 # Don't track C++ language server cache files.
 .ccls-cache
 .clangd/
-.vscode/
 *.code-workspace
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "(Windows) Lancher",
+            "type": "cppvsdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/${command:cmake.buildType}/pioneer.exe",
+            "args": [],
+            "stopAtEntry": false,
+            "cwd": "${workspaceFolder}",
+            "environment": [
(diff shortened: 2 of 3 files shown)


Web-eWorks:
I like this idea on the surface, but I'm very hesitant to merge it if there's a possibility it will cause issues or confusion for other contributors. For example, I use VScode as well and have a decent amount of settings related to my personal workflow in the VScode config files; tracking these files in source control has the potential to turn into a constant stream of overwrite commits when submitting otherwise unrelated work.

I'd much prefer a setup script that creates the initial VScode files for a specific platform and leave these files untracked; I think this would be much more flexible and allow developers to tailor their launch options etc. to their liking without having to constantly preserve them from being overwritten.

impaktor:
If we don't want it in master, it could be offered as a downloadable on the wiki under development section, or somewhere else?

pcercuei:
@Web-eWorks well, if your local `launch.json` are useful, why not add them here? Having several launch options is better than having none.

impaktor:
So in summary: Either we close (and potentially add note to [dev docs](https://dev.pioneerspacesim.net/contribute/development-tools)), or add more useful stuff for pioneer dev in `launch.json`. It's up to @Web-eWorks . I see the point in not wanting local launch.json clobbered (Does VSCode support mulitple launch<n>.json files?).

Web-eWorks:
I consider this PR to be superseded with the merge of #5626. VSCode implicitly merges launch definition lists from the `.code-workspace` file as well as `launch.json`, but defaults to creating new configurations in the latter. That makes the `launch.json` file a better fit to be excluded from source control as it's for user-specific configurations which we *do not* want to be checking into source control.

If @pcercuei can determine the above configurations are still relevant with the changes to CMake management in VSCode, I'd welcome a PR to add them to the launch section of the optional `pioneer-default.code-workspace` file we now provide as a "quickstart" method for VSCode users. The quickstart method is fully documented in `COMPILING.txt` - you're welcome to copy / document in the dev docs as well.

PR #5210, opened 2021-06-10
Mark Pioneer as DPI-aware on Windows

pcercuei:
Without this change, Windows would automatically apply a scaling factor
for Pioneer which prevented the game from using the full resolution.

Mark Pioneer as DPI-aware so that Windows won't mess in our business.

### DIFF
--- a/src/core/OS.h
+++ b/src/core/OS.h
@@ -33,6 +33,9 @@ namespace OS {
 	// Open the Explorer/Finder/etc
 	bool SupportsFolderBrowser();
 	void OpenUserFolderBrowser();
+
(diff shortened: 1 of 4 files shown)


Web-eWorks (changes requested):
Instead of #including windows.h in `main.cpp`, I'd recommend adding a new method to the `OS` namespace - e.g. `SetDPIAware` that correctly executes SetProcessDPIAware on Windows and has stubs for the Posix and OSX implementations for platform-specific handling.

pcercuei:
@Web-eWorks updated.

Web-eWorks (approved):

Issue #5220, opened 2021-06-25
Fast sequential hyperjumps may cause crash if police is scanning your cargo


jimishol:
This is how crash happened
1. jump to system  
2. immediately get scanned by police 
3. during scanning jump to new system already set in hyperjump planner
result of step 3 was a crash.
 
the output.txt says nothing. it is difficult to reproduce and all i have is what i see on terminal now, I post last lines

Struve 2398 B d:
    height fractal: HillsNormal
    colour fractal: Rock
    seed: 2549709410
Color Fractal name: GGSaturn
WARNING: ship varada discarded, because there are no hyperspace routes available for it
WARNING: ship deneb discarded, because there are no hyperspace routes available for it
DEBUG: Jumps since warranty: 0
Warranty expires: 20:30:44 23 Ιούλ 3222
0 clouds brought over
Stars picked from galaxy: 0
Final stars number: 140000
Started hyperspacing...
Fatal: [string "[T] @modules/PolicePatrol/PolicePatrol.lua"]:36: attempt to index field 'system' (a nil value)
stack traceback:
        [string "[T] @modules/PolicePatrol/PolicePatrol.lua"]:36: in function 'hasIllegalGoods'
        [string "[T] @modules/PolicePatrol/PolicePatrol.lua"]:159: in function <[string "[T] @modules/PolicePatrol/PolicePatrol.lua"]:157>

As in IRC suggested 'shouldn't crash even if jumping fast. That should be part of playing mechanics: "jump before police finish scanning".'

I will try to reproduce. I fast jumped many times and never experienced crashes. Today i decided to start new game on Barnard's star, i just bought my first hyperdrive 1 and it was my first hyperdrive jump sequence.

After closing windows output.txt filled up. So there ii is https://susepaste.org/36991731

Gliese852:
@jimishol I think the problem is that activating the hyperdrive at the time of scanning leads to the collapse of the universe

jimishol:
On my first RP request, because of my ignorance of how RP works, i unwillingly disturbed the peace of master branch multiple times for something simple. Because of my knowledge, that limits me only on trivial stuff, if you agree with my solution, please make a proper PR.
The crash can happen statistically few times, certainly below 50% on drivers class 1 (that probability decrease as class gets bigger), with zero probability for class 4  and above. That means it can happen only on early few gameplay hours because in the first few hours of play the player gets military or standard hyperdrive of class 4.

The crash can happen when police scan duration

[...]

PR #5222, opened 2021-06-27
RFC: Safe LuaObject:ApplyAt API

laarmen:
This is an RFC API aiming to mitigate the safety problems caused by
Timer:CallAt as seen in plenty of issues, the latest of which is #5220.
The fundamental issue with the CallAt API is that the environment in
which the closure is called can be drastically different to the one in
which it was defined.

This new API aims to guarantee the caller that at least one object must
still be valid when calling the closure, by storing it out of band and
having the engine check its validity before calling the callback.

This would allow some simplification in existing correct callbacks, e.g.
in the assassination module, we could turn this

    Timer:CallAt(mission.due, function () if mission.ship:exists() then mission.ship:Undock()

into

    mission.ship:CallAt(mission.due, Ship.Undock)

Thoughts ?


Gliese852:
@laarmen what if the second argument is a method not from the 'Ship' table, but from another table (Space, for example)?  Will the object be passed there as self?

laarmen:
I think so. My Lua memories are a bit fuzzy, but IIRC the whole "method" syntax is purely syntactical sugar.

When I wrote the example, I wasn't planning on it being reduced to directly passing a method, it was just a happy coincidence. You could also write it as follows.

```lua
mission.ship:CallAt(mission.due, function (s) s.Undock() end)
```

Gliese852:
@laarmen 
Also had a thought, maybe it can be done like this:
```
mission.ship:CallAt(mission.due, 'Undock')
```
imho in that case it will even be easy to serialize, although it only becomes usable for this object.

laarmen:
Serialization is out of scope. This API is meant to reduce the amount of boilerplate necessary for delayed calls in the Lua scripts. Such calls are not currently serialized (unless I'm missing something?).

I am vehemently against using strings instead of actual functions. See https://devcards.io/stringly-typed for more details on this. If you want to ensure you're using the correct method for your object, you can simply write

```lua
mission.ship:CallAt(mission.due, mission.ship.Undock)
-- or
mission.chip:CallAt(mission.due, function(s) s:Undock() end)
```

Web-eWorks:
I like this idea, though I think on its own it merely reduces to `Timer.CallAt(mission.due, wrapperFn(mission.ship, Ship.Undock))`; this isn't something that would require an entirely separate API, merely a little extra logic in a wrapper function to check if the object exists.

If we're going to go the route of adding additional API specifically for this type of pattern, I think we should look more closely at LuaEvent as well - it's a common pattern to write a function preamble in event handlers that tests if a specific argument a) still exists, and/or b) is a specific object or class of objects. This seems like a similar requirement to that proposed for Timer, and if an API pattern can be found that covers both use-cases it would certainly improve the code quality significantly.

PR #5225, opened 2021-07-06
Update to AUTHORS.txt File

RacerBG:
I don't know if this is the right way to do it but basically I'm adding myself via this PR to the AUTHORS.txt file. I'm doing this because I feel that the Bulgarian translation is in good shape right now (for clarity reasons I'm the translator in question). :smiley: 

From the Wiki:
> If you're a new translator and you'd like your name included in AUTHORS.txt, please let us know! 

Regards,
Georgi


The-EG:
Rather than deleting the entire file and replacing it, I'd recommend just modify the existing file to add yourself (you should be able to do that within the GitHub web interface even).

RacerBG:
> 
> 
> Rather than deleting the entire file and replacing it, I'd recommend just modify the existing file to add yourself (you should be able to do that within the GitHub web interface even).

This method would add an empty line to the end of the file. Otherwise, you are correct.

impaktor:
It should just edit one line not "+253 −252". Fixed it with a separate push. Thanks for the translations!

PR #5227, opened 2021-07-06
Improve automatic star selection in the sector map

Gliese852:
During manual movement in the sector map, when the option is enabled, the closest star to the map viewing position is automatically selected. But the search is performed only in the sector containing the original position, so the wrong star may be highlighted. This commit fixes this defect, now the search is performed in an increasing (if necessary) range, and the really closest star is found.

Also, fixed that the star was not automatically selected when animating "by inertia", but only when a key was pressed.

### DIFF
--- a/src/SectorView.cpp
+++ b/src/SectorView.cpp
@@ -60,7 +60,6 @@ enum Side {
 };
 
 using box_t = std::array<int, SIDES>;
-using cache_t = SectorCache::Slave *;
 
 static bool in_box(int x, int y, int z, const box_t &box)
 {
@@ -81,7 +80,7 @@ static int box_max_size(const box_t &box)
 // targetSector - coordinates of the sector in which we are looking for the nearest system
 // min_dist_sq - here we will accumulate the minimum distance
 // best_sys - here we will accumulate the nearest system
-static void try_sector_for_nearer_system(cache_t cache, const vector3f &crd, const SystemPath &originSector, const SystemPath &targetSector, float &min_dist_sq, SystemPath &best_sys)
+static void try_sector_for_nearer_system(SectorCache::Slave *cache, const vector3f &crd, const SystemPath &originSector, const SystemPath &targetSector, float &min_dist_sq, SystemPath &best_sys)
 {
 	RefCountedPtr<Sector> ps = cache->GetCached(targetSector);
 	vector3f base(
@@ -89,15 +88,13 @@ static void try_sector_for_nearer_system(cache_t cache, const vector3f &crd, con
 		targetSector.sectorY - originSector.sectorY,
 		targetSector.sectorZ - originSector.sectorZ);
 	base *= Sector::SIZE;
-	if (!ps->m_systems.empty()) {
-		for (unsigned int i = 0; i < ps->m_systems.size(); i++) {
-			Sector::System *ss = &ps->m_systems[i];
-			vector3f dx = crd - (ss->GetPosition() + base);
-			float dist = dx.LengthSqr();
-			if (dist < min_dist_sq) {
-				min_dist_sq = dist;
-				best_sys = ss->GetPath();
-			}
+	for (unsigned int i = 0; i < ps->m_systems.size(); i++) {
+		Sector::System *ss = &ps->m_systems[i];
+		vector3f dx = crd - (ss->GetPosition() + base);
+		float dist = dx.LengthSqr();
+		if (dist < min_dist_sq) {
+			min_dist_sq = dist;
+			best_sys = ss->GetPath();
 		}
 	}
 }
@@ -111,7 +108,7 @@ static void try_sector_for_nearer_system(cache_t cache, const vector3f &crd, con
 // prevSearchBox - we shouldn't search inside this box (because already searched)
 // min_dist_sq - here we will accumulate the minimum distance
 // best_sys - here we will accumulate the nearest system
-static bool search_nearest_in_box(cache_t cache, const SystemPath &sectorPos, const vector3f &crd, box_t &searchBox, box_t &prevSearchBox, float &min_dist_sq, SystemPath &best_sys)
+static bool search_nearest_in_box(SectorCache::Slave *cache, const SystemPath &sectorPos, const vector3f &crd, box_t &searchBox, box_t &prevSearchBox, float &min_dist_sq, SystemPath &best_sys)
 {
 	for (int x = searchBox[XMIN]; x < searchBox[XMAX]; ++x)
 		for (int y = searchBox[YMIN]; y < searchBox[YMAX]; ++y)
@@ -128,7 +125,7 @@ static bool search_nearest_in_box(cache_t cache, const SystemPath &sectorPos, co
 // pos -  absolute coordinates of a point (in sectors, float)
 // returns the found SystemPath, (system closest to the given position)
 // if returned !SystemPath.IsSystemPath(), it means that nothing was found
-static SystemPath nearest_system_to_pos(cache_t cache, const vector3f &pos)
+static SystemPath nearest_system_to_pos(SectorCache::Slave *cache, const vector3f &pos)
 {
 	// maximum size in sectors of the search box
 	const int MAX_SEARCH_BOX = 10;
@@ -151,12 +148,13 @@ static SystemPath nearest_system_to_pos(cache_t cache, const vector3f &pos)
 	float min_dist_sq = FLT_MAX;
 	// here we will accumulate the nearest system (this will be the result of the function)
 	SystemPath best_sys;
-	// the first iteration of the search - only in the sector where the base point is located
-	box_t searchBox = originBox;
-	++searchBox[XMAX];
-	++searchBox[YMAX];
-	++searchBox[ZMAX];
 
+	// the first iteration of the search - only in the sector where the base point is located
+	box_t searchBox = {
+		sectorPos.sectorX, sectorPos.sectorX + 1,
+		sectorPos.sectorY, sectorPos.sectorY + 1,
+		sectorPos.sectorZ, sectorPos.sectorZ + 1
+	};
 	// scan sectors, and expand the search box until we find at least one nearest system
 	while (!search_nearest_in_box(cache, sectorPos, crd, searchBox, prevSearchBox, min_dist_sq, best_sys)) {
 		// expand the search box
@@ -172,9 +170,8 @@ static SystemPath nearest_system_to_pos(cache_t cache, const vector3f &pos)
 	// edges of the search box, whether they are closer than the found star
 	bool search_complete = true;
 	for (int side = 0; side < SIDES; ++side) {
-		float side_crd = (searchBox[side] - originBox[side]) * Sector::SIZE;
-		float dist_to_edge_sq = pow(posBox[side] - side_crd, 2);
-		if (dist_to_edge_sq < min_dist_sq) {
+		float dist_to_edge = posBox[side] - (searchBox[side] - originBox[side]) * Sector::SIZE;
+		if (dist_to_edge * dist_to_edge < min_dist_sq) {
 			// the origin is closer to this edge of the search box than to the nearest found - we expand in this direction
 			searchBox[side] += enlarge[side];
 			search_complete = false;
--- a/src/SectorView.cpp
+++ b/src/SectorView.cpp
@@ -79,9 +79,9 @@ static int box_max_size(const box_t &box)
 // crd - coordinates of the origin point in the sector, l.y.
 // originSector - coordinates of the sector containing the origin point
 // targetSector - coordinates of the sector in which we are looking for the nearest system
-// min_dist - here we will accumulate the minimum distance
+// min_dist_sq - here we will accumulate the minimum distance
 // best_sys - here we will accumulate the nearest system
-static void try_sector_for_nearer_system(cache_t cache, const vector3f &crd, const SystemPath &originSector, const SystemPath &targetSector, float &min_dist, SystemPath &best_sys)
+static void try_sector_for_nearer_system(cache_t cache, const vector3f &crd, const SystemPath &originSector, const SystemPath &targetSector, float &min_dist_sq, SystemPath &best_sys)
 {
 	RefCountedPtr<Sector> ps = cache->GetCached(targetSector);
 	vector3f base(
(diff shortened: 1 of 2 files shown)


Web-eWorks (changes requested):
At great length (and with no shortage of internal deliberation), I'm going to have to ask that this code be revised before it's merged.

Though it's certainly a valid way to solve the problem, the inline type declarations of an ad-hoc integer AABB type (and the associated functions) make the code difficult to read, and the expanding-search implementation is far more complicated than necessary for the limited scope of its usage.

I would like to see an implementation that uses the existing AABB type to search *at most* the 3x3 cube of sectors surrounding the current sector, returning no selection if there are in fact no systems there. A lot of the logic I'm reading in the current implementation - to me, at least - seems to ignore the principle of YAGNI: "you aren't going to need it". I'm laying out a few points as review comments that would need to be changed whether or not this implementation was approved, with the intent of advice and not attacking your code directly.

I know this is a bit hypocritical of me to deny merging code that I myself asked you to write, but it is my firmly-held opinion that merging this code as-is would not have a positive effect on the codebase - it doesn't follow the idioms or types of the surrounding code, and it took me a good five minutes to begin to understand the control flow and especially the math in the hand-rolled AABB type.

That was a whole lot of negatives without many positives, so for what I do like about this code:
- It selects systems by closest distance, the most critical part of the PR!
- It's extremely well commented, to the point where I almost don't need to read the actual C++ to understand the ins and outs of the function. (Although requiring comments to know what's going on in a function is considered by some to be a code smell, I personally appreciate comments that explain both the what and the *why* of a function's implementation.)
- Functions and variables are well-named and make sense when read in English.

src/SectorView.cpp
```diff
@@ -47,6 +47,128 @@ enum DetailSelection {
...
+	box_t prevSearchBox = originBox;
+	// here we will accumulate the minimum distance
+	float min_dist = FLT_MAX;
+	// here we will accumulate the nearest system (this will be the result of the function)
+	SystemPath best_sys;
+	// the first iteration of the search - only in the sector where the base point is located
+	box_t searchBox = originBox;
+	++searchBox[XMAX];
+	++searchBox[YMAX];
+	++searchBox[ZMAX];
```
Web-eWorks, line 137: This section in particular catches my eye as the first thing that would need improvement; instead of manually incrementing the maximum values of the box, I'd recommend changing the for-loop condition to `<=` instead; not only does it cause significant confusion about whether a box with `min = (1,1,1); max = (2,2,2)` has size==1 or size==2, it also generally is contradictory to existing AABB logic in the codebase which uses bounds tests of `p >= min && p <= max`.

src/SectorView.cpp
```diff
@@ -47,6 +47,128 @@ enum DetailSelection {
...
+			for (int side = 0; side < SIDES; ++side) {
+				float side_crd = (searchBox[side] - originBox[side]) * Sector::SIZE;
+				float dist_to_edge = pow(posBox[side] - side_crd, 2);
+				if (dist_to_edge < min_dist) {
+					// the system is closer to this edge of the search box than the nearest found - we expand in this direction
+					searchBox[side] += enlarge[side];
+					search_complete = false;
+				}
+			}
+		}
```
Web-eWorks, line 166: I am very confused about the logic in this function; to start, `dist_to_edge` should critically mention that it is *squared* distance, as the actual LengthSqr call is squirreled away in the `try_sector_for_nearer_system` function and thus the logical assumption is that this is simply Euclidean distance, leading the `pow(..., 2)` call to misdirect the reader.

Additionally, the need (if any) to expand the *search box* to encompass a system found by *searching with that search box* is extremely counter-intuitive logic, and I find myself completely questioning its validity and existence. After a good few minutes of thinking, the only case I can derive is where the search position originates at the very extreme edge of a sector's bounds, a system is found N sectors away, and a second, nearer system is located N+1 sectors away from the originating sector. Even so, this is very non-obvious behavior not explained by the comments, and something that I would highly recommend be replaced by a search with a fixed radius.

src/SectorView.cpp
```diff
@@ -47,6 +47,128 @@ enum DetailSelection {
 static const float ZOOM_SPEED = 15;
 static const float WHEEL_SENSITIVITY = .03f; // Should be a variable in user settings.
 
+#define FFRAC(_x) ((_x)-floor(_x))
```
Web-eWorks, line 50: If you're going to move the definition of FFRAC to the top of the file, please remove it from its previous location :D Makes reviewing the file much easier, as it doesn't look like it was invented from whole cloth.

src/SectorView.cpp
```diff
@@ -47,6 +47,128 @@ enum DetailSelection {
...
+// best_sys - here we will accumulate the nearest system
+static void try_sector_for_nearer_system(cache_t cache, const vector3f &crd, const SystemPath &originSector, const SystemPath &targetSector, float &min_dist, SystemPath &best_sys)
+{
+	RefCountedPtr<Sector> ps = cache->GetCached(targetSector);
+	vector3f base(
+		targetSector.sectorX - originSector.sectorX,
+		targetSector.sectorY - originSector.sectorY,
+		targetSector.sectorZ - originSector.sectorZ);
+	base *= Sector::SIZE;
+	if (ps->m_systems.size()) {
```
Web-eWorks, line 92: This if statement is redundant, as a for loop operating in the range `[0 .. m_systems.size)` will run zero iterations. The conditional in a for loop is always run before the loop body begins execution, no matter how many iterations have been processed.

Web-eWorks:
It occurs to me (perhaps a bit late) that I may not have clearly communicated what I originally envisioned the selection behavior functioning as. I'd envisioned the selection functioning with a maximum radius that does a single-pass search over sectors within that radius (perhaps `max.x = center.x + ceil(radius / SECTOR_SIZE)`), either returning the closest system within that radius or returning no selection if the player has found a dense enough region of space.

You've certainly gone above and beyond that requested behavior, and I'm of two minds as to whether an expanding selection radius until a system is found is a good idea - my thought is that it could potentially select systems offscreen if density in the viewed region of space is low enough. The current behavior is certainly acceptable, although it would require some cleanup and clarification of control flow before I'm comfortable merging it.

src/SectorView.cpp
Gliese852: @Web-eWorks of course, here only the squares of distances are used, apparently I underestimated the "unusualness" of this.
About the need to expand the search box - for example, here is a search in a 3x3 box:

artflow_202107161139

Arrows indicate places where unaccounted points may be located. I think this is not a very extreme case, given that on average there can be 2-3 stars in the sector. Moreover, in three dimensions the situation is even more dramatic, because the diagonal of the cube is longer than the diagonal of the square.

Web-eWorks: After another 20 minutes I realized that was what this code was taking into account, although to do so I had to disregard the comments entirely and rethink the entire logic flow of that sub-branch. I see what's going on there, although I'd still recommend side-stepping it with a fixed-radius search (or at least clarifying variable names and comments to explicitly mention that's the edge case you're handling).

src/SectorView.cpp
Gliese852: @Web-eWorks This is because otherwise the bottom and top edges will have different semantics (?), I mean you have to write different comparison functions for the min and max edges. That is, if you follow your advice, a box of size 1 will be `min (3,3,3) max (3,3,3)`, the point with absolute coordinates `(3.5 3.5 3.5)` will be inside this box, although its coordinates will be greater than the maximum values of the box, therefore, comparing with the upper bound, you will need to add 1.

I don't understand what the logic of our AABB has to do with it, because there are real numbers, and for them `>` and `>=` are almost the same.

src/SectorView.cpp
Gliese852: @Web-eWorks Please tell me which comment misled you, because it seems to me that the comments describe exactly what is happening in the code. Therefore, it is very important to see the outside view.

It seemed to me that such a problem arises when searching for the nearest in any spatial index, be it a quad-tree, or a kd-tree (or in a simple grid, as in our case), when the found nearest in a given branch may not actually be the nearest. And we need to search in neighboring branches that are closer to us than the nearest found.
About searching in a fixed sphere, I will answer elsewhere.

Gliese852:
@Web-eWorks 
> the inline type declarations of an ad-hoc integer AABB type

But it's not a type (from the certain point of view©), it's just an array, I can remove `using`

> I would like to see an implementation that uses the existing AABB

I didn't use this structure for several reasons:
1. It contains real numbers, but I need integers (mostly), and I am suspicious of checking real numbers for equality.
2. I have not seen examples of using AABB with sector coordinates in the code, it seems that this is meant to store the dimensions of a certain model.
3. AFAIK I cannot iterate over all the elements of this structure as an array (without modifications), and instead of one loop on all sides of the box, I have to write 6 approximately identical pieces of code (I mean, I can't think of any other way to achieve this yet)

> to search at most the 3x3 cube of sectors surrounding the current sector

If you just blindly do this, you have to check 27 sectors, and discard all results that are further than 1 sector (8 ly), Otherwise, you cannot guarantee that you did not miss a closer point.

About searching in a sphere of fixed radius, this approach seems to be quite good, if first select all sectors that exactly fall into this sphere, and then discard the results that do not fall into this sphere. An important question - what is this radius, and how many sectors we are must to scan, to find that the nearest star is 1 light year away from the cursor pos?

My algorithm will scan 1 or 2 sectors in most cases  (8 in very most cases). But I really don't know how important it is.

src/SectorView.cpp
Gliese852: Thanks, accidentally missed, corrected, thanks, inadvertently missed, corrected!

src/SectorView.cpp
Gliese852: Done!

src/SectorView.cpp
Web-eWorks: Hmm... I see your point now that I understand what your actual bounds test does. In any case, I'd recommend constructing the max values of the search box directly as `x + 1, y + 1` etc. instead of doing it in this manner.

src/SectorView.cpp
Web-eWorks: Yes, the comments mention that you're expanding towards the found system rather than away; I'd appreciate the comments explicitly mentioning that this covers the edge case in which the distance to the search box edge is less than the diagonal distance to the found system.

Web-eWorks:
> AFAIK I cannot iterate over all the elements of this structure as an array (without modifications), and instead of one loop on all sides of the box, I have to write 6 approximately identical pieces of code (I mean, I can't think of any other way to achieve this yet)

I withdraw the request to use the existing AABB now that I see how you're using this capability for optimization purposes, especially the distance-to-side calculation. I'd emphasize the need for comments that clarify the nature of that algorithm, because otherwise a lot of the written code seems to be overengineered and needlessly complicated.

> If you just blindly do this, you have to check 27 sectors, and discard all results that are further than 1 sector (8 ly), Otherwise, you cannot guarantee that you did not miss a closer point.

If the sector the "focal point" resides in has no systems (a fairly common occurrence around Sol, I counted at least 4-5 empty sectors within 6-8 sectors and your video pointed out it happens a lot more than even I thought), you're already skipping straight to searching 27 sectors in the cube around the focal point. I don't think that this search, performed only one or a few times per frame is going to cause a particular performance issue by iterating through ~64 systems per test. Indeed, this provides a better upper-bound in low density regions of space, and also resolves the case where the closest system may well be behind the camera instead of the visible but slightly further distant system the user expects it to select.

> About searching in a sphere of fixed radius, this approach seems to be quite good, if first select all sectors that exactly fall into this sphere, and then discard the results that do not fall into this sphere. An important question - what is this radius, and how many sectors we are must to scan, to find that the nearest star is 1 light year away from the cursor pos?

My suggestion was to merely search an 8ly radius (1 sector width) around the cursor, clearing the selection if nothing was found. In this way, only a 3x3 cube needs be searched, even if the focal point lies near or on the edge of the current sector. However, if this search function is meant to be used as a more general "find system closest to point" for more than just selecting systems based on the user's camera position, I see why it is implemented as such, though I'd request that it be more well-integrated with the SectorMap API.

> About selecting the offscreen star - now the maximum size of the search box is 20 sectors, const int MAX_SEARCH_DIST = 20; this is a configurable parameter. Perhaps we can even change it, depending on how many sectors are visible on the screen.

If this is the maximum "diameter" of the search box (rather than the half-extent of any given axis), I'd recommend dropping it to 12-16 instead of 20 - 9100 sectors is a little steep as an upper-bound worst case.

> But it's not a type (from the certain point of view©), it's just an array, I can remove using

I will leave a review comment about the problematic usage of `using` - this one was more an issue that the `std::array<int, SIDES>` (with or without the `using` statement) was a red-flag to my sense of "code smell", it indicates that the following code tried to re-invent the wheel rather than doing something sensible but boring.

It's a bit of a humorous idiom among software engineers, but if a piece of code seems 'clever' when you're writing it, that's usually a big red flag that it will wind up incomprehensible and a pain to maintain by others (to say nothing of yourself). Doesn't always work out that way, but anytime you think you're writing smart or clever code it's like a red sky in the morning - coders should take warning.

In this case, making the code comprehensible again is mostly a matter of fixing the comments to no longer be misleading and addressing a few issues I'll leave in review comments,

src/SectorView.cpp
```diff
@@ -47,6 +47,145 @@ enum DetailSelection {
...
+	XMAX,
+	YMIN,
+	YMAX,
+	ZMIN,
+	ZMAX,
+	SIDES
+};
+
+using box_t = std::array<int, SIDES>;
+using cache_t = SectorCache::Slave *;
```
Web-eWorks, line 63: This is a problem. `cache_t` hides the fact that this is a pointer type to an already typedef-ed type alias (`SectorCache` being a wrapper typedef for a long template). It's only used in three functions, and the actual type name is fairly short. I'm sorry I didn't catch it in the initial review, I was focusing on the box_t above it. In general, there's no need to typedef this, especially not to a "C-style" typename - we're not writing kernel-style opaque C interfaces here where a `cache_t` is a pointer to an opaque anonymous struct typedef that is never defined in the interface.

src/SectorView.cpp
Web-eWorks: > apparently I underestimated the "unusualness" of this

Generally speaking, if it's not immediately apparent what distance metric you're using (e.g. a call to LengthSqr within the same function) it's better to either comment it or make it part of the variable name. Using `min_dist_sq` significantly reduces the ambiguity of the function, and overall improves the immediate readability of the code.

I'm personally trying to adopt the philosophy that the larger the scope a variable has, the longer and more descriptive its name should be. Loop indices can be a single letter, function-local variables and parameters one to two words, variables that are passed around to multiple functions more than one word, etc. If you have to read back to a variable's definition to know the semantics of the contents, you've written the equivalent of a rude note to the next guy who has to work on the code.

Aside from all that, I'm requesting one more change here - instead of `pow(v, 2)`, please prefer `v * v` - unless you're using a non-integer power argument, you can simply compose a square or cube exponentiation and it winds up faster in debug builds as it doesn't call into the fairly complicated exponentiation function.

src/SectorView.cpp
```diff
@@ -47,6 +47,145 @@ enum DetailSelection {
...
+// best_sys - here we will accumulate the nearest system
+static void try_sector_for_nearer_system(cache_t cache, const vector3f &crd, const SystemPath &originSector, const SystemPath &targetSector, float &min_dist_sq, SystemPath &best_sys)
+{
+	RefCountedPtr<Sector> ps = cache->GetCached(targetSector);
+	vector3f base(
+		targetSector.sectorX - originSector.sectorX,
+		targetSector.sectorY - originSector.sectorY,
+		targetSector.sectorZ - originSector.sectorZ);
+	base *= Sector::SIZE;
+	if (!ps->m_systems.empty()) {
```
Web-eWorks, line 92: I see you implemented one half of the prior review comment - with `i == 0 && ps->m_systems.empty()`, the loop condition `i < ps->m_systems.size()` will always return false and the loop will not be executed, making this test here completely redundant.

src/SectorView.cpp
Gliese852: Done!

src/SectorView.cpp
Gliese852: Done!

src/SectorView.cpp
Gliese852: Done!

src/SectorView.cpp
Gliese852: Done!

Web-eWorks (approved):

Web-eWorks:
Thank you @Gliese852 for all of your hard work on this PR and for putting up with my detailed and sometimes pedantic reviews! I'll try to cut a release a little later today once I've built everything locally and it all works!

PR #5228, opened 2021-07-06
Restore the "How much mass" question

robothauler:
Improvement of #5164
This restores the "How much mass" question and moves the "Is the amount of cargo negotiable" button below that.

image


impaktor:
Cool, there'll be a release this week, I'll aim to include this

PR #5229, opened 2021-07-17
Fixing the normals issue of the DSMiner

nozmajner:
There was some kind of normals issue on the high lod of the DSMiner. Not sure how it got there, or when. It was an eyesore, so I've figured I fix it since we are having a release soon.
Before:
dsminer_1
After:
dsminer_2


Gliese852:
maybe it should have been that concave?  (or convex)

Web-eWorks (approved):
This gets my thumbs-up - the old normals looked very ugly, and if it needed to be concave it would have been much better to simply model it in or bake it to the normal map instead of messing with vertex normals.

PR #5240, opened 2021-08-15
Refactor Lua Serialization

Web-eWorks:
This PR removes the old text-based Lua pickling and serialization entirely, and fixes a long-running bug where an attempt to serialize out a non-LuaWrappable userdata object was impossible and would cause a segfault. Now, object serialization operates in a type-erased fashion, allowing userdata objects of all sorts to be serialized to Json from a Lua context with only a single function call to register the serialization hooks.

This PR also includes several major improvements to the LuaMetaType framework for efficient Lua binding of C++ objects - now it supports binding generic C++ functions (like ImGui methods), as well as directly registering meta-methods in several different ways and first-class support for registering constructors. The internals of the binding framework have been slightly refactored as well to make them more readable and better communicate the use of the different wrapper function templates.

Additionally, LuaVector, LuaVector2, and LuaColor have been ported to the LuaMetaType binding system and serialization support has been added to each class.

### DIFF
--- a/src/Color.h
+++ b/src/Color.h
@@ -124,17 +124,27 @@ struct Color4ub {
 	static Color4ub FromLuaTable(lua_State *l, int idx);
 
 	Uint8 GetLuminance() const;
-	void Shade(float factor)
+	Color4ub Shade(float factor)
 	{
-		r = static_cast<Uint8>(r * (1.0f - factor));
-		g = static_cast<Uint8>(g * (1.0f - factor));
-		b = static_cast<Uint8>(b * (1.0f - factor));
+		Color4ub out = *this;
+		out.r = static_cast<Uint8>(r * (1.0f - factor));
+		out.g = static_cast<Uint8>(g * (1.0f - factor));
+		out.b = static_cast<Uint8>(b * (1.0f - factor));
+		return out;
 	}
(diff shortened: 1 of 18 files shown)


impaktor:
Is this save bumping?

As for metatype stuff, I can read dumb code, but I suspect the meta type stuff is too smart for me to read/use. Anyway, would need examples, as existing code in the lua/Lua*cpp files is what I use when I expose c++ functions to lua-side.

Web-eWorks:
This does not bump savefile versions - it's backwards compatible with existing saves, although it technically generates saves that cannot be opened by previous versions of the game using the same savefile number. Might be worth bumping saves at next release, but it's not required.

The LuaMetaType code should actually be slightly easier to work with in a copy-paste manner compared to old code, as there's less boilerplate to write and there's only a little to remember when binding stuff to Lua - it's far more flexible too. That being said, it's still the evil < template metaprogramming > so I don't hold your reticence to touch it against you!

impaktor:
I suggest we hold of on save bumping, until next release, I'm sure something will motivate us to bump, and merge this as is.

Web-eWorks:
Merging this as it's been ready for quite some time.

PR #5247, opened 2021-08-25
Draw sector map labels with ImGui.

Gliese852:
Add polymorphic type SectorView::Label, for displaying different types of labels.
Implement two types of labels - for stars and for factions.
Add highlight colors to the theme.

ezgif-6-f06bb82c43b3

### DIFF
--- a/src/SectorView.cpp
+++ b/src/SectorView.cpp
@@ -43,6 +43,242 @@ static const float FAR_THRESHOLD = 7.5f;
 static const float FAR_LIMIT = 36.f;
 static const float FAR_MAX = 46.f;
 
+class SectorView::Label {
+public:
+	virtual ~Label() {}
+	// check if the cursor has hit the label, and if so, also save the boxes to select the text
+	// should run inside imgui context
+	virtual bool Hovered(const ImVec2 &p) = 0;
+	virtual void Draw(ImDrawList &dl, bool hovered = false) = 0;
+	virtual void OnClick() = 0;
+	float Depth() const { return pos.z; }
+	Label(Labels &host, const vector3f &pos, ImU32 color) :
+		pos(pos), color(color), host(host){};
+
+protected:
+	vector3f pos;
+	ImU32 color;
+	Labels &host;
+};
+
+// the Label type must be complete for this constructor to work, because we
+// have a unique_ptr<Label> here
+SectorView::Labels::Labels(SectorView &view) :
+	view(view) {}
+
+// helper functions to detect when the cursor hits the label
+static bool inbox(ImRect b, ImVec2 p)
+{
+	return p.x > b.Min.x && p.x < b.Max.x && p.y > b.Min.y && p.y < b.Max.y;
+}
+
+static bool incircle(ImVec2 c, float r, ImVec2 p)
+{
+	return (p.x - c.x) * (p.x - c.x) + (p.y - c.y) * (p.y - c.y) < r * r;
+}
+
+static bool indiamond(ImVec2 c, float r, ImVec2 p)
+{
+	float dx = p.x - c.x;
+	float dy1 = p.y - (c.y - r);
+	float dy2 = p.y - (c.y + r);
+	return dy1 > dx && dy1 > -dx && dy2 < dx && dy2 < -dx;
+}
+
+class SectorView::StarLabel : public SectorView::Label {
+
+	std::string name;
+	ImVec2 namePos;
+	// screen radius of the star
+	float radius;
+	// the height of the text of each label is different
+	float textHeight;
+	float scaledGap;
+	SystemPath path;
+
+	//  how much we will increase the original font so that it is less blurry for heavily enlarged labels
+	static constexpr float fontScale = 5.f;
+
+public:
+	StarLabel(Labels &host, const vector3f &pos, const Color &color_, const std::string &name, const float radius, const SystemPath &path) :
+		Label(host, pos, 0),
+		name(name),
+		radius(radius),
+		path(path)
+	{
+		ImFont *font = host.starLabelFont;
+		// we want to shade more distant labels, and make the closer ones transparent
+		const float far = -0.001;
+		// at a given depth, we see a label with a nominal font size
+		const float mid = -0.03;
+		const float near = -0.08;
+		// the font may not exist in the very first frame
+		const float fsize = font ? font->FontSize / fontScale : 15.f;
+		textHeight = fsize / mid * pos.z;
+		scaledGap = host.gap / mid * pos.z;
+		const int alpha = pos.z > near ? 255 : std::max(int(near / pos.z * 255), 0);
+		const float shad = pos.z < mid ? 1.0f : std::max((far - pos.z) / (far - mid), .0f);
+		namePos = ImVec2(pos.x + radius + scaledGap + scaledGap, pos.y - textHeight / 2);
+		color = IM_COL32(color_.r * shad, color_.g * shad, color_.b * shad, alpha);
+	};
+
+	bool Hovered(const ImVec2 &p) override
+	{
+		// something we can weed out right away
+		if (p.x < pos.x - radius) return false;
+		if (p.y < std::min(pos.y - radius, namePos.y)) return false;
+		if (p.y > std::max(pos.y + radius, namePos.y + textHeight)) return false;
+		ImFont *font = host.starLabelFont;
+		ImGui::PushFont(font);
+		ImVec2 ts = ImGui::CalcTextSize(name.c_str());
+		ImGui::PopFont();
+		// scale the size
+		ts.x = textHeight / ts.y * ts.x;
+		ts.y = textHeight;
+		// sift out by last dimension
+		if (p.x > namePos.x + ts.x + scaledGap) return false;
+		// so, the cursor is somewhere in the dimensions of this label
+		if (incircle(ImVec2(pos.x, pos.y), radius, p) ||
+			inbox(ImRect(namePos.x - scaledGap, namePos.y - scaledGap, namePos.x + ts.x + scaledGap, namePos.y + font->FontSize + scaledGap), p)) {
+			// so the cursor really covers this label
+			// save the boxes for further highlighting the label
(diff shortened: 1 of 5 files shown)


src/SectorView.cpp
```diff
@@ -43,6 +43,236 @@ static const float FAR_THRESHOLD = 7.5f;
...
+	static constexpr float fontScale = 5.f;
+
+public:
+	StarLabel(SectorView &host, const vector3f &pos, const Color &color_, const std::string &name, const float radius, const SystemPath &path) :
+		Label(host, pos, 0),
+		name(name),
+		radius(radius),
+		path(path)
+	{
+		ImFont *&font = host.m_labels.fonts.starLabel;
```
Web-eWorks, line 103: It's not necessary (and honestly just confusing) to take a reference to a pointer variable that you're using in read-only fashion; this should just be `ImFont *`.

src/SectorView.cpp
```diff
@@ -507,6 +731,65 @@ void SectorView::Draw3D()
...
+				if (ImGui::IsMouseReleased(0)) label.OnClick();
+				break;
+			}
+		}
+	}
+
+	// draw labels
+	for (size_t i = 0; i < m_labels.array.size(); ++i) {
+		Label &label = *m_labels.array[i];
+		if (i != hovered_i) {
```
Web-eWorks, line 779: Rather than this extra control flow here, I'd recommend at the start of the for-loop just writing:
```c++
if (i == hovered_i)
    continue;
```

This simplifies control flow and makes it easier on the reader to understand what's going on.

src/SectorView.cpp
```diff
@@ -507,6 +731,65 @@ void SectorView::Draw3D()
...
+{
+	m_labels.fontName = fontName;
+	m_labels.fontSize = fontSize;
+	m_labels.gap = gap;
+	m_labels.colors.highlight = IM_COL32(highlight.r, highlight.g, highlight.b, highlight.a);
+	m_labels.colors.shade = IM_COL32(shade.r, shade.g, shade.b, shade.a);
+}
+
+// function runs from lua, inside imgui context
+void SectorView::DrawLabels()
```
Web-eWorks, line 747: Instead of Lua calling into this function, prefer calling this from c++ under the `DrawPiGui` virtual function and let Lua call in to only enable/disable drawing of labels. This is mostly for issues with Lua function continuations and to allow us to further harden against Lua errors - as much as possible should be event driven and I want to avoid large, lengthy functions in C++ executing inside of a lua callstack context if possible.

Additionally, there's room to go the extra mile and convert the SectorMap to run its rendering on the appropriate handler instead of the `game` pigui handler - this now works for all PiGuiViews.

src/SectorView.cpp
```diff
@@ -507,6 +731,65 @@ void SectorView::Draw3D()
...
+			m_drawList->CmdBuffer.back().PrimDepth = -label.Depth() * 1.01;
+			label.Draw(*m_drawList);
+		}
+	}
+
+	// draw the label over which the cursor is hovering the most recent
+	if (hovered_i != NOT_FOUND) {
+		Label &hovered = *m_labels.array[hovered_i];
+		m_drawList->AddDrawCmd();
+		m_drawList->CmdBuffer.back().PrimDepth = 1.0f;
```
Web-eWorks, line 790: If I correctly remember how this is implemented, a PrimDepth of 1.0 is the very back of the scene and all UI elements are drawn with a depth of 0. I could be misremembering this, I'd have to dig back and check the commit.

Web-eWorks (changes requested):
Some minor changes, some major changes. I haven't reviewed absolutely everything, but this should be most of the major change-required areas.

If I had to sum up the entire review in one useless phrase it would be "prefer writing idiomatic code" - use our types and functions, don't nest anonymous structs for namespacing (and as I write that I realize my input code is guilty), and try to minimize control flow scopes - rather than create a new block wrapped in {}, see if you can invert the logic to avoid the extra indent and new block.

Web-eWorks:
Thanks for poking me @fluffyfreak - I've been chipping away at this review for a few days now, just haven't had the time or energy to finish it :smile:

fluffyfreak:
@Web-eWorks wasn't a poke just found I had a few minutes for reviews and "filing" work 😁

src/SectorView.h
```diff
@@ -206,6 +209,33 @@ class SectorView : public PiGuiView, public DeleteEmitter {
 	Graphics::Drawables::Lines m_sectorlines;
 	Graphics::Drawables::Lines m_routeLines;
 	Graphics::Drawables::Points m_farstarsPoints;
+
+	class Label;
+	class StarLabel;
+	class FactionLabel;
+	struct {
```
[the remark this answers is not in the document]
Gliese852: Done!

src/SectorView.h
```diff
@@ -206,6 +209,33 @@ class SectorView : public PiGuiView, public DeleteEmitter {
...
+		struct {
+			ImFont *starLabel = nullptr;
+			ImFont *factionHome = nullptr;
+			ImFont *factionName = nullptr;
+		} fonts;
+		struct {
+			using box2d = std::array<float, 4>;
+			box2d starLabel;
+			box2d factionHome;
+			box2d factionName;
```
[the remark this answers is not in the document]
Gliese852: Done!

src/SectorView.cpp
```diff
@@ -43,6 +43,236 @@ static const float FAR_THRESHOLD = 7.5f;
...
+	vector3f pos;
+	ImU32 color;
+	SectorView &host;
+};
+
+// helper functions to detect when the cursor hits the label
+static bool inbox(float xmin, float xmax, float ymin, float ymax, float x, float y)
+{
+	return x > xmin && x < xmax && y > ymin && y < ymax;
+}
```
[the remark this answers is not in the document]
Gliese852: Done!

src/SectorView.cpp
```diff
@@ -43,6 +43,236 @@ static const float FAR_THRESHOLD = 7.5f;
...
+
+static bool incircle(float cx, float cy, float r, float x, float y)
+{
+	return (x - cx) * (x - cx) + (y - cy) * (y - cy) < r * r;
+}
```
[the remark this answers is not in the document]
Gliese852: Done!

src/SectorView.cpp
```diff
@@ -43,6 +43,236 @@ static const float FAR_THRESHOLD = 7.5f;
...
+	float dx = x - cx;
+	float dy1 = y - (cy - rad);
+	float dy2 = y - (cy + rad);
+	return dy1 > dx && dy1 > -dx && dy2 < dx && dy2 < -dx;
+}
+
+class SectorView::StarLabel : public SectorView::Label {
+
+	std::string name;
+	std::array<float, 2> namePos;
```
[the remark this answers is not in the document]
Gliese852: Done!

src/SectorView.cpp
Gliese852: Done, where this pointer is not supposed to change

src/SectorView.cpp
Gliese852: Done!

src/SectorView.cpp
Gliese852: Done!

src/SectorView.cpp
Gliese852: I have empirically established that with such depth, the object seems to be drawn on top of all.

Gliese852:
Force-pushed because I forgot to check with clang.

src/SectorView.cpp
Web-eWorks: Looking good, my only suggestion is to allow pigui to control whether or not labels are being drawn by setting a variable (to fully replicate previous behavior)

Gliese852: Done!

src/SectorView.cpp
```diff
@@ -43,6 +43,237 @@ static const float FAR_THRESHOLD = 7.5f;
 static const float FAR_LIMIT = 36.f;
 static const float FAR_MAX = 46.f;
 
+class SectorView::Label {
```
Gliese852, line 46: @Web-eWorks oh, should I make the base class destructor virtual?

impaktor:
Looks like there's one change requested left? (and fixes to be squashed.)

src/SectorView.cpp
Web-eWorks: Most likely yes, if you're calling `delete` on a pointer to the base type. `std::unique_ptr<Label>` initialized with a StarLabel object pointer will need a virtual destructor for example.

src/lua/LuaSectorView.cpp
```diff
@@ -39,6 +40,11 @@ void LuaObject<SectorView>::RegisterClass()
 		.AddFunction("SetHyperspaceTarget", &SectorView::SetHyperspaceTarget)
 		.AddFunction("ResetHyperspaceTarget", &SectorView::ResetHyperspaceTarget)
 		.AddFunction("IsCenteredOn", &SectorView::IsCenteredOn)
+		.AddFunction("SetLabelParams", [](lua_State *l, SectorView *sv) {
+			sv->SetLabelParams(LuaPull<std::string>(l, 2), LuaPull<int>(l, 3), LuaPull<float>(l, 4), *LuaColor::CheckFromLua(l, 5), *LuaColor::CheckFromLua(l, 6));
+			return 0;
+		})
```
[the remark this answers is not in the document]
Gliese852: @Web-eWorks Strange, apparently the language server quietly added the required header `LuaColor.h`, but I didn't double-check it then, but now everything works, there is the necessary overload. So I removed the lambda!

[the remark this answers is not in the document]
Web-eWorks: Excellent! Sometimes I wish I could keep the language list from adding headers (src/lua.hpp for one), but it's pretty helpful to manage dependencies like this.

Web-eWorks (approved):
Looking good @Gliese852! Thanks for putting up with all of my stylistic and control flow pedantry as usual!

Gliese852:
Squashed all the fix-commits

PR #5249, opened 2021-08-27
Increase and limit the height of the BvhTree (for collision geoms)

### LINKED ISSUE #5245: Crash issue
### Observed behaviour
A consistent crash when you get close to OPLI Contemptation (about 0.06 AU), output.txt suggests it might be because of the AI traffic but I honestly have no idea

### Steps to reproduce
Load the save, put time compression on 10000x

pioneer-20210723 win10 x64

[savefile.zip](https://github.com/pioneerspacesim/pioneer/files/7025234/savefile.zip)
[config and output.txt.zip](https://github.com/pioneerspacesim/pioneer/files/7025238/config.and.output.txt.zip)

Gliese852:
The height of that tree can fluctuate very much, starting from log2(n) and ending with n (n is the number of objects in space).

When traversing the tree, an array (for the stack) is used, the size of which is fixed, because we do not want dynamic allocations in this case, but must be at least the height of the tree.

We can limit the height of the tree, since a leaf can contain multiple objects.

It seems the height of the tree in the game almost never exceed 20.

So I took the number 38 (with a margin), it seems almost never the height of the tree will be limited, and at the same time we are insured against UB in all cases.

Fixes #5245 

ADD:

I would like to add that I checked the program's work with the height limit of 5, everything seems to work stably, while the tree height (excluding the root) does not exceed 5, and the value of the 'stack' array index does not overflow.

### DIFF
--- a/src/collider/CollisionSpace.cpp
+++ b/src/collider/CollisionSpace.cpp
@@ -59,6 +59,7 @@ public:
 	BvhNode *m_nodesAlloc;
 	int m_nodesAllocPos;
 	int m_nodesAllocMax;
+	static constexpr int MAX_TREE_HEIGHT = 38;
 
 	BvhNode *AllocNode()
 	{
@@ -77,9 +78,10 @@ public:
 	void CollideGeom(Geom *, const Aabb &, int minMailboxValue, void (*callback)(CollisionContact *));
 
 private:
-	void BuildNode(BvhNode *node, const std::list<Geom *> &a_geoms, int &outGeomPos);
+	void BuildNode(BvhNode *node, const std::list<Geom *> &a_geoms, int &outGeomPos, int maxHeight);
 
-	void FreeAll() {
+	void FreeAll()
+	{
 		if (m_geoms) delete[] m_geoms;
 		m_geoms = nullptr;
 		if (m_nodesAlloc) delete[] m_nodesAlloc;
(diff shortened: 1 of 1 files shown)


Gliese852:
I got interested, and I tried to visualize this tree, and it turned out that in this save there are two ships that are very far from the center of the system. Apparently, this is some kind of error in generating the initial position of the ship. Because of these ships, the volume tree becomes 2 higher than it could be without them, what contributed to the height overflow at a critical moment.

https://user-images.githubusercontent.com/18342621/131211055-bc90caee-757f-42a6-8d3e-cab5b4d8b8b1.mp4

fluffyfreak:
That's a nice visualisation too, pelase add it as a PR sometime!

fluffyfreak (approved):
Looks great to me. Any reason for picking 38 as the max depth and not just going all the way to 64?

Gliese852:
@fluffyfreak I chose the number 38 because `three eight` - `tree height`. And with such a height, a tree can fit ~275 billion objects, it seemed to me that was enough.

fluffyfreak:
Ah yes, that should suffice 🤣

Gliese852:
@fluffyfreak 
> That's a nice visualisation too, please add it as a PR sometime!

Sure, I have a few more ideas for visualizations in the system map, probably I will someday combine them and do a PR.
(although in this current case I had to drag some members from the private section to the public, in the `CollisionSpace` class).

Web-eWorks (approved):
+1 for BVH tree visualization, anything that gives us better debug is a good thing (and it'll aid me in rewriting the BVH tree internals later). I'd like a comment explaining the choice of 38 as a tree height next to the declaration, as it will prevent future maintainers from having to work their way all the way back to this PR, but it's not a priority and certainly not something to block merge over.

Gliese852:
@Web-eWorks added comments as I like to write comments.

impaktor:
Good to merge? Last commit seems to indicate it wants to be squashed...

PR #5250, opened 2021-08-30
Fix PutShipOnRoute function

Gliese852:
After placing the ship on the route, the possibility of a collision is checked.  In case of a possible collision, its position is changed to avoid this. It turns out that this calculation worked incorrectly, because the `FromVectors` function creates a matrix using vectors as
columns, and in the further calculation it is assumed that vectors are rows.

Also removed unnecessary `Frame` object retrieval by `FrameId`.

This video shows that now everything is working as it should (at least in the video). The dangerous trajectory that was originally shown in red, the new safe trajectory is shown in blue, and the alternative trajectory (also safe, but on the other side of the obstructor) in gray.

https://user-images.githubusercontent.com/18342621/131369271-f4b4a4fb-f1df-49ea-8e7b-0c0929afd538.mp4

Two cases are shown - when the ship is out of the target frame, and when it is in the same frame with the target. In the first case, the trajectory turns so as not to fall into a dangerous proximity, and in the second case, the ship moves to the normal to the planet's surface.

### DIFF
--- a/src/lua/LuaSpace.cpp
+++ b/src/lua/LuaSpace.cpp
@@ -267,7 +267,7 @@ static int l_space_put_ship_on_route(lua_State *l)
 	const int coll = CheckCollision(ship, reldir, targdist, targpos, 0, erad);
 	if (coll) {
 		// need to correct positon, to avoid collision
-		if (Frame::GetFrame(shipFrame->GetNonRotFrame()) != Frame::GetFrame(targFrame->GetNonRotFrame()) && targpos.Length() > erad) {
+		if (shipFrame->GetNonRotFrame() != targFrame->GetNonRotFrame() && targpos.Length() > erad) {
 			// the ship is not in the target's frame or target is above the effective radius of obstructor - rotate the ship's position
 			// around the target position, so that the obstructor's "effective radius" does not cross the path
 			// direction obstructor -> target
@@ -276,8 +276,9 @@ static int l_space_put_ship_on_route(lua_State *l)
 			const vector3d y = z.Cross(shippos).NormalizedSafe();
 			// just the third axis of this basis
 			const vector3d x = y.Cross(z);
+
 			// this is the basis in which the position of the ship will rotate
-			const matrix3x3d corrCS = matrix3x3d::FromVectors(x, y, z);
+			const matrix3x3d corrCS = matrix3x3d::FromVectors(x, y, z).Transpose();
 			const double len = targpos.Length();
 			// two possible positions of the ship, when flying around the obstructor to the right or left
 			// rotate (in the given basis) the direction from the target to the obstructor, so that it passes tangentially to the obstructor
--- a/src/lua/LuaSpace.cpp
+++ b/src/lua/LuaSpace.cpp
@@ -253,22 +253,19 @@ static int l_space_put_ship_on_route(lua_State *l)
 	ship->SetFuel((0.001 * pp.getMass() - ss.static_mass) / st->fuelTankMass);
 
 	ship->UpdateFrame();
-	Frame *shipFrame = Frame::GetFrame(ship->GetFrame());
-	Frame *targFrame = Frame::GetFrame(targetbody->GetFrame());
-
 	// check for collision at spawn position
 	const vector3d shippos = ship->GetPosition();
 	const vector3d targpos = targetbody->GetPositionRelTo(ship->GetFrame());
 	const vector3d relpos = targpos - shippos;
 	const vector3d reldir = relpos.NormalizedSafe();
 	const double targdist = relpos.Length();
-	Body *body = shipFrame->GetBody();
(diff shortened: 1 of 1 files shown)


fluffyfreak (approved):
Nice work, and I love a good debug visualisation video!

impaktor:
Will merge within 24 h if no one objects

src/lua/LuaSpace.cpp
```diff
@@ -276,8 +276,9 @@ static int l_space_put_ship_on_route(lua_State *l)
 			const vector3d y = z.Cross(shippos).NormalizedSafe();
 			// just the third axis of this basis
 			const vector3d x = y.Cross(z);
+
 			// this is the basis in which the position of the ship will rotate
-			const matrix3x3d corrCS = matrix3x3d::FromVectors(x, y, z);
+			const matrix3x3d corrCS = matrix3x3d::FromVectors(x, y, z).Transpose();
```
Web-eWorks, line 281: I'd appreciate a comment here denoting the fact that FromVectors is returning a matrix in the wrong (columnar) layout. Otherwise, this is good to merge.

Gliese852: @Web-eWorks to be honest, I find it difficult to say whether this is correct or not, the question is, what does the action "construct a matrix3x3 from vector3s" mean?

Web-eWorks: Matrix3x3::FromVectors (should, as an API contract) create a row-major orthonormal rotation matrix from three basis vectors defining the directions of the cartesian X, Y, Z axes under the transformation of the resulting matrix as long as all three vectors are unit-length and perpendicular to each other.

In this regard, returning a column-major matrix is invalid behavior, although given that transposition == inversion for affine transformation matrices, it's entirely possible that it has been miscommunicated and the function is returning the right matrix but the operation with the resulting matrix is required to be an inverse transformation.

Web-eWorks: Looking at the code, it appears that the matrix is being constructed in the correct order; what is happening here is that FromVectors is creating a matrix that takes a coordinate system and rotates it into the 'world' coordinate system, but the code wants to perform calculations on world positions in the simplified coordinate system. As such, an inverse transformation is required to bring the world coordinates (targpos) into the coordinate system, and then the result in the simplified coordinate space is transformed back into world space (by an inverse-inverse transformation, which is the same as transforming by the result of FromVectors directly).

src/lua/LuaSpace.cpp
```diff
@@ -267,7 +267,7 @@ static int l_space_put_ship_on_route(lua_State *l)
 	const int coll = CheckCollision(ship, reldir, targdist, targpos, 0, erad);
 	if (coll) {
 		// need to correct positon, to avoid collision
-		if (Frame::GetFrame(shipFrame->GetNonRotFrame()) != Frame::GetFrame(targFrame->GetNonRotFrame()) && targpos.Length() > erad) {
+		if (shipFrame->GetNonRotFrame() != targFrame->GetNonRotFrame() && targpos.Length() > erad) {
 			// the ship is not in the target's frame or target is above the effective radius of obstructor - rotate the ship's position
 			// around the target position, so that the obstructor's "effective radius" does not cross the path
 			// direction obstructor -> target
```
Web-eWorks, line 273: Interesting dillemma here, the code checks that we are both not in the same frame and further than the effective radius, however the comment is saying that this is an 'or' condition - what happens if we are in the same frame but beyond effective radius?

src/lua/LuaSpace.cpp
Web-eWorks: This could be rewritten (without the initial .Transpose() call) using inverse-multiplication syntax, although I don't particularly recommend it:
```cpp
const vector3d safe1 = corrCS * (matrix3x3d::RotateY(+asin(erad / len)) * (-targpos * corrCS)).Normalized() * targdist;
```

src/lua/LuaSpace.cpp
Gliese852: @Web-eWorks Really weird. After paradoxically long deliberation, it seems to me that we need to change the comment to "and", because we want to enter the sphere of effective radius of obstructor only perpendicular to it, even if for some reason the target is not in the obstructor frame. (perhaps it could be an orbital station, in a very very low orbit?)

Gliese852: @Web-eWorks if we are in the same frame but the target is beyond effective radius, like in any other case, when the target is beyond the effective radius, we can calculate the tangent, so option 1 can be used. I think I should remove the frame check at all.

Web-eWorks: Feel free - I'm not particularly familiar with this algorithm, but that inconsistency stood out.

Gliese852: @Web-eWorks done!

PR #5253, opened 2021-09-02
unit fix

joonicks:
simple G --> g

### DIFF
--- a/data/lang/core/en.json
+++ b/data/lang/core/en.json
@@ -916,8 +916,8 @@
     "message": "."
   },
   "NUMBER_G": {
-    "description": "",
-    "message": "%acceleration{f.1} G"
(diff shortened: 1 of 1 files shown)


impaktor:
I'm always very hesitant to merge changes to en.json files (unless just uploaded), since it will trigger a cascade of changes in all translation files.
@Web-eWorks didn't you try to also do similar edit in every foo/*json file, and that actually updated the language sources?

Web-eWorks:
I can't confirm if it prevented a transifex push, but I did use the translation script to propagate deletion of an entry across all language files in a recent PR. Currently, I don't see the need to invalidate all translations of this constant to change the case in english at this time.

joonicks:
... a system that discourages fixes seems like a flawed system ...

PR #5255, opened 2021-09-04
make jump time depend on equipment

joonicks:
preserves jump time for all existing civilian drives but reduces jump duration by 22% for military drives

allows for jump times to be tweaked on a per-drive basis


Web-eWorks (approved):
I'm good with this change, though I intend to revisit the underlying math in both route duration and range calculation to adjust overall balance of hyperdrives.

joonicks:
this one has been merged into #5270, both substantially alter "Equipment.lua"

Issue #4688, opened 2019-10-07
commslog does not move to last entry automatically


Cyborgscode:
the commslog slider does not automove to the bottom end of the message list,
if a new entry is done and the list already fills the full window.

zonkmachine:
I'm thinking it would be elegant to fill the log from the top. That would fix this issue.

As a proof of concept. Change this:
https://github.com/pioneerspacesim/pioneer/blob/db12012b723a80d0c21c605ab6dc645a8d15f694/data/libs/Game.lua#L5

to:
`table.insert(Game.comms_log_lines, 1, { text=text, time=Game.time, sender=sender, priority = priority or 'normal'})`

This isn't enough though as it eventually crashes with:
`Fatal: [string "[T] @pigui/modules/comms.lua"]:68: attempt to index local 'last' (a nil value)`

zonkmachine:
I found a working solution in `/data/pigui/modules/comms.lua`.
 
https://github.com/pioneerspacesim/pioneer/blob/master/data/pigui/modules/comms.lua#L82
```diff
-  for k,v in pairs(utils.reverse(utils.take(lines, commsLinesToShow))) do
+  for k,v in pairs(utils.take(lines, commsLinesToShow)) do
```

https://github.com/pioneerspacesim/pioneer/blob/master/data/pigui/modules/comms.lua#L98
```diff
-  for k,v in pairs(lines) do
+  for k,v in pairs(utils.reverse(lines)) do
```

I can pull that if you like.

PR #5256, opened 2021-09-05
Commsfixes

zonkmachine:
Two commits concerning the comms terminal.
 
 * When new messages are pushed to the comms terminal they are eventually presented hidden after the old messages. Solved by showing the new messages on top. Pizzazz!
Fixes https://github.com/pioneerspacesim/pioneer/issues/4688

 * When the small window reaches its five maximum line a scrollbar appears to the right. Scrollbar removed.

Pictures of the comms log messages propagating top to bottom and with scrollbar with five messages.
test123
scrollbar

### DIFF
--- a/data/pigui/modules/comms.lua
+++ b/data/pigui/modules/comms.lua
@@ -53,7 +53,7 @@ local function displayCommsLog()
 										ui.setNextWindowSize(aux , "Always")
 										aux = Vector2(mainButtonSize.x + 2 * mainButtonFramePadding + 15, 10)
 										ui.setNextWindowPos(aux , "Always")
-										ui.window("ShortCommsLog", {"NoTitleBar", "NoResize", "NoFocusOnAppearing", "NoBringToFrontOnFocus"},
+										ui.window("ShortCommsLog", {"NoTitleBar", "NoResize", "NoFocusOnAppearing", "NoBringToFrontOnFocus", "NoScrollbar"},
 															function()
 																local last = nil
 																local rep = 0
--- a/data/pigui/modules/comms.lua
+++ b/data/pigui/modules/comms.lua
@@ -79,7 +79,7 @@ local function displayCommsLog()
 																	table.insert(lines, 1, { sender = last.sender, text = last.text .. ((rep > 1) and (' x ' .. rep) or ''), priority = last.priority })
 																end
 																ui.pushTextWrapPos(ui.screenWidth/4 - 20)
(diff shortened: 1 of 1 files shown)


Web-eWorks (approved):

impaktor:
1. I feel I'd like to get some community reaction on this, before merge.
2. is it possible to scroll the log without the scroll bar?

zonkmachine:
>     1. I feel I'd like to get some community reaction on this, before merge.
>

:+1: 

>     2. is it possible to scroll the log without the scroll bar?

Yes. You can scroll it with the wheel, but I actually didn't think there was any real reason for the scrollbar to appear. I was wrong. The text can of course be longer than one line and then it would be useful. I can drop that change.

zonkmachine:
This earlier code for scrollbar position is commented out. I've played along with it and it looks like it's working but I still think new messages on top is the way to go. Remove?

https://github.com/pioneerspacesim/pioneer/blob/cc7e945d628d5467e0e131d898cc4f86f5405ecb/data/pigui/modules/comms.lua#L101-L105

> The text can of course be longer than one line and then it would be useful. I can drop that change.

I've tested it some more and I still think removing the scrollbar is better. It can still be scrolled with the mouse wheel and the text isn't persistent anyway.

Web-eWorks (approved):
I agree that new messages on top and removing the scrollbar is better - if the player wants to see the whole message, there's a button right there to open the commslog window and scroll the entire thing. Overall this is a very short and clean PR, thank you @zonkmachine!

If you want to do some future work in another PR, you could add a new messages count to the commslog button - count the number of messages that have appeared since the comms log was last opened, set the imgui cursor pos to somewhere in the top right of the comms log button, and `ui.text()` the number. That might resolve any questions that messages could be potentially missed by the player.

zonkmachine:
> If you want to do some future work in another PR, you could add a new messages count to the commslog button - count the number of messages that have appeared since the comms log was last opened, set the imgui cursor pos to somewhere in the top right of the comms log button, and `ui.text()` the number. That might resolve any questions that messages could be potentially missed by the player.

I'll keep that in mind for future exploration. 

I haven't actually played around that much with Pioneer but it looks like most of the comms messages, at least at this point, are one liners and also pretty scarce in between. On my screen the small log is 6 lines long so the chance of it actually overflowing isn't that large. I could make it slightly larger to further reduce this chance. Today the size is `ui.screenHeight / 8`

zonkmachine:
OK. I think I'm done with comms log for this PR.

PR #5257, opened 2021-09-05
unbloated best-match code

joonicks:
instead building lists and sorting, just keep track of best match


Web-eWorks (approved):
Appreciate this, the original intent was to have more complicated economy selection logic but this is a perfectly fine algorithm even in that case!

PR #5259, opened 2021-09-06
Replace obsolete imgui functions

Jeinzi:
Hello! :)
- AddBezierCurve has been renamed to AddBezierCubic in [4d8e839dd](https://github.com/ocornut/imgui/commit/4d8e839ddf3786f11390002b86218419311843fb)
- ImGuiInputTextFlags_AlwaysInsertMode has been renamed to ImGuiInputTextFlags_AlwaysOverwrite in [287bd9b98](https://github.com/ocornut/imgui/commit/287bd9b98488d5f7bc0f730904af7f0c842971eb)

### DIFF
--- a/src/lua/LuaPiGui.cpp
+++ b/src/lua/LuaPiGui.cpp
@@ -181,7 +181,7 @@ static LuaFlags<ImGuiInputTextFlags_> imguiInputTextFlagsTable = {
 	{ "AllowTabInput", ImGuiInputTextFlags_AllowTabInput },
 	{ "CtrlEnterForNewLine", ImGuiInputTextFlags_CtrlEnterForNewLine },
 	{ "NoHorizontalScroll", ImGuiInputTextFlags_NoHorizontalScroll },
-	{ "AlwaysInsertMode", ImGuiInputTextFlags_AlwaysOverwrite },
+	{ "AlwaysOverwrite", ImGuiInputTextFlags_AlwaysOverwrite },
 	{ "ReadOnly", ImGuiInputTextFlags_ReadOnly },
 	{ "Password", ImGuiInputTextFlags_Password }
 };
--- a/src/lua/LuaPiGui.cpp
+++ b/src/lua/LuaPiGui.cpp
@@ -181,7 +181,7 @@ static LuaFlags<ImGuiInputTextFlags_> imguiInputTextFlagsTable = {
 	{ "AllowTabInput", ImGuiInputTextFlags_AllowTabInput },
 	{ "CtrlEnterForNewLine", ImGuiInputTextFlags_CtrlEnterForNewLine },
 	{ "NoHorizontalScroll", ImGuiInputTextFlags_NoHorizontalScroll },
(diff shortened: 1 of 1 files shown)


src/lua/LuaPiGui.cpp
```diff
@@ -1235,7 +1235,7 @@ static int l_pigui_add_bezier_curve(lua_State *l)
 	ImColor color = LuaPull<ImColor>(l, 5);
 	float thickness = LuaPull<double>(l, 6);
 	int num_segments = LuaPull<int>(l, 7);
-	draw_list->AddBezierCurve(a, c0, c1, b, color, thickness, num_segments);
+	draw_list->AddBezierCubic(a, c0, c1, b, color, thickness, num_segments);
```
fluffyfreak, line 1238: In this case Bezier Curve is the correct name, Cubic and Bezier curves are different things

Jeinzi: The curves added here are in fact cubic Bezier curves, in contrast to for example quadratic Bezier curves - the recent naming changes in imgui reflect that and make the difference clearer.
If you look at [line 2491 in imgui.h](https://github.com/ocornut/imgui/blob/607ad8c949a5411f4ce069e8e0f8ae773718e7ec/imgui.h#L2491), you can see that AddBezierCurve() is nothing more than a redirect to AddBezierCubic() and will be removed entirely in the near future. See also the linked commits above.

fluffyfreak: Ah ok

src/lua/LuaPiGui.cpp
```diff
@@ -181,7 +181,7 @@ static LuaFlags<ImGuiInputTextFlags_> imguiInputTextFlagsTable = {
 	{ "AllowTabInput", ImGuiInputTextFlags_AllowTabInput },
 	{ "CtrlEnterForNewLine", ImGuiInputTextFlags_CtrlEnterForNewLine },
 	{ "NoHorizontalScroll", ImGuiInputTextFlags_NoHorizontalScroll },
-	{ "AlwaysInsertMode", ImGuiInputTextFlags_AlwaysInsertMode },
+	{ "AlwaysInsertMode", ImGuiInputTextFlags_AlwaysOverwrite },
```
Web-eWorks, line 184: I'd appreciate it if you can Ctrl+F over the files in the data/ directory for usage of the AlwaysInsertMode name - if there are instances, please replace them with AlwaysOverwrite, otherwise please rename the string as well to match!

Web-eWorks (changes requested):

src/lua/LuaPiGui.cpp
Jeinzi: There were no matches for "AlwaysInsertMode" in data/ so I just renamed the string in LuaPiGui.cpp :)

Web-eWorks: Thanks! Just standard procedure when changing string script constants.

Web-eWorks (approved):

PR #5260, opened 2021-09-07
WIP warning redux

joonicks:
try to reduce warning spam during compilation

"WIP": while pragma messages are in place to see where warning have been fixed

if PR is accepted the final step would be to clean out all pragma messages

### DIFF
--- a/src/SystemView.cpp
+++ b/src/SystemView.cpp
@@ -486,7 +486,8 @@ void SystemView::Update()
 	AnimationCurves::Approach(m_rot_y, m_rot_y_to, ft);
 
 	// to capture mouse when button was pressed and release when released
-	if (Pi::input->MouseButtonState(SDL_BUTTON_MIDDLE) != m_rotateWithMouseButton) {
+#pragma message("FIX: warning comparing int with bool")
+	if (bool(Pi::input->MouseButtonState(SDL_BUTTON_MIDDLE)) != m_rotateWithMouseButton) {
 		m_rotateWithMouseButton = !m_rotateWithMouseButton;
 		Pi::input->SetCapturingMouse(m_rotateWithMouseButton);
 	}
--- a/src/collider/GeomTree.cpp
+++ b/src/collider/GeomTree.cpp
@@ -7,7 +7,10 @@
 #include "Weld.h"
 #include "scenegraph/Serializer.h"
 
+#pragma message("FIX: warning unknown pragma")
+#ifdef __GNUC__
 #pragma GCC optimize("O3")
+#endif /* __GNUC__ */
 
 GeomTree::~GeomTree()
 {
@@ -57,7 +60,8 @@ GeomTree::GeomTree(const int numVerts, const int numTris, const std::vector<vect
 		std::vector<Uint32> xrefs;
 		nv::Weld<vector3f> weld;
 		weld(m_vertices, xrefs);
-		m_numVertices = m_vertices.size();
+#pragma message("FIX: warning of data loss (x64)")
+		m_numVertices = 0xFFFFFFFF & m_vertices.size();
 
 		//Output("---   %d vertices welded\n", count - newCount);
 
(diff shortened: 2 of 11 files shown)


Web-eWorks (changes requested):
In all cases, please do not use bitwise-and with a literal constant, instead please prefer `uint32_t(<expr>)` (or as needed) as a functional-style cast instead.

joonicks:
Ive been debating with myself which one makes more sense. The resulting code should be identical but the anding version makes more explicit that bit are truncated, a simple typecast looks neater but later on it may be overlooked that it actually truncates bits.

Web-eWorks:
The lack of visual clarity far outweighs any semantic benefit that might be gained by highlighting the fact that we're truncating the array size to *only* four billion entries. Please use the functional-style cast, explicit bit constants add visual clutter and make the code harder to read for no benefit in this case.

src/SystemView.cpp
```diff
@@ -197,10 +197,10 @@ static vector3d position_of_surface_starport_relative_to_parent(const SystemBody
...
 		// we need the distance to the center of the planet
 		(Pi::game->IsNormalSpace() && Pi::game->GetSpace()->GetStarSystem()->GetPath().IsSameSystem(Pi::game->GetSectorView()->GetSelected()) ?
-				// if we look at the current system, the relief is known, we take the height from the physical body
-				Pi::game->GetSpace()->FindBodyForPath(&(starport->GetPath()))->GetPosition().Length() :
-				// if the remote system - take the radius of the planet
-				parent->GetRadius());
+				  // if we look at the current system, the relief is known, we take the height from the physical body
+				  Pi::game->GetSpace()->FindBodyForPath(&(starport->GetPath()))->GetPosition().Length() :
+				  // if the remote system - take the radius of the planet
+				  parent->GetRadius());
```
Web-eWorks, line 203: This particular indenting can be ignored and is caused by an upstream malfunction in clang-format. Any ternary-operator related indentation other than a single tab is not part of the Pioneer style, and the output from clang-format is wrong. No action is required on your part, just keep that in mind in the future.

joonicks: I didnt touch those lines, I think msvs did something "clever" on its own

src/galaxy/SystemPath.cpp
```diff
@@ -27,9 +27,18 @@ static int ParseInt(const std::string &str)
...
+#pragma message("FIX: warning of unused variable")
+#pragma warning(push)
+#pragma warning(disable : 4101)
+#endif /* __GNUC__ */
 	} catch (const std::invalid_argument &e) {
 		throw SystemPath::ParseFailure();
 	}
+#ifndef __GNUC__
+#pragma warning(pop)
+#endif /* __GNUC__ */
```
Web-eWorks, line 41: Instead of the ugly preprocessor check to disable a warning number, please just remove the variable name or cast it to void; e.g. `catch (const std::invalid_argument &)`.

src/collider/GeomTree.cpp
```diff
@@ -7,7 +7,10 @@
 #include "Weld.h"
 #include "scenegraph/Serializer.h"
 
+#pragma message("FIX: warning unknown pragma")
+#ifdef __GNUC__
 #pragma GCC optimize("O3")
+#endif /* __GNUC__ */
```
Web-eWorks, line 13: Would appreciate if this could have an MSVC branch with MSVC's optimize pragma equivalent - it's fairly important this file is optimized even in debug builds.

Web-eWorks (changes requested):
I appreciate the effort in fixing all of these warning locations - there are a few tweaks related to the preprocessor that I'd like to see made, but once those are done I'm happy to merge this (with the debug messages removed, of course).

src/galaxy/SystemPath.cpp
joonicks: Im a C++ noob, now I know how to

PR #5262, opened 2021-09-13
fix crash from selling when amount=0

### LINKED ISSUE #5251: Soldout module allows to sell more than a trader needs causing fatal error
Today i saw a WANTED computers, where computer stock in station was zero. I did not opened sell option. I  traveled to nearby station and bought 2 computers because i have only Xylophis. Got back and forcefully i sold 1 computer. Upon selling the 2nd computer gamed crashed with 
Fatal: [string "[T] @modules/SoldOut/SoldOut.lua"]:20: attempt to index local 'ad' (a nil value)
It seems that trader needs only 1 computer and when i try to sell a 2nd game crashes
[output.txt](https://github.com/pioneerspacesim/pioneer/files/7078900/output.txt)
_Originally posted by @jimishol in https://github.com/pioneerspacesim/pioneer/issues/5221#issuecomment-908585304_

jimishol:
Fixes #5251 by closing the form, so player can not sell when amount=0.

### DIFF
--- a/data/modules/SoldOut/SoldOut.lua
+++ b/data/modules/SoldOut/SoldOut.lua
@@ -61,6 +61,7 @@ local onChat = function (form, ref, option)
 	if ad.amount == 0 then
 		form:RemoveAdvertOnClose()
 		ads[ref] = nil
+		form:Close()
 	end
 
 	return


impaktor:
I could edit your post to be correct, but better to teach: Please use word "fixes #5251" in the body of your post (above), and change title to something like "fix crash from something something" (but better than I suggested here), we don't put URLs in  the title.

PR #5264, opened 2021-09-16
Constellations appear in background sky.

jimishol:
By default constellations do not appear in background sky. May be constellation are not visible by ships in space because there is no atmosphere to filter out less bright stars. However it seems more amusing for player to see a more familiar background sky with constellations.  
**By default background sky looks like** 
no_constellations_100
Reducing, in options, the AmountOfBackgroundStars to 0.18 the sky looks like
no_constellations_18
By the present brightnessPower value in Starfield.ini is increased, while the product brightnessPower * brightnessApparentSizeFactor kept equal to the current configuration's one. 
visibleRadiusLy is very sensitive to visualization effects. By trail and error a value of 240 gave best results. By altering only Starfield.ini leaves AmountOfBackgroundStars=1 and background sky will look like
constellations_100
By altering the default AmountOfBackgroundStars to 0.18 **the default background sky will look like**
constellations_18
and player can alter it up to 100 to get the previous image if last default is not the one he likes.

_Unfortunately the above images do not do justice to the real picture on screen. Fortunately these minor changes can be done very easily by hand, so that their effect can be checked on the run._


nozmajner:
I think the last example doesn't work too nicely. Those larger starts feel very out of place, and not like a speck of bright light at all. That was the reason I muted them back half a year or so ago.

jimishol:
_nozmajner, I assume you checked the values on the run and not from the above pictures only._

Yes, that was my initial thoughts too and I was trying to lower the brightness. Till, from my porch, i looked up in the sky  more carefully. Stars in Greece sky feel very out of space and are not just a speck of bright light at all. And, unless we live in a simulation, the stars I see are real ones. 
The fact is that the perspective of what one sees as constellations is subjective. Other sky is natural for a country living player, other is for a city living one and probably both will disagree with astronauts' perspective. 
With Starfield.ini that i propose, if you put dense of stars at options to 100, you effectively remove the effect.
Most important for me is to move the default value of dense of stars at options way down from 100. Thus player will know that he can have that effect and may try either to remove it my increasing the dense of stars or even  to alter values in Starfield.ini, in a try to catch the perfect for him. While the default remains to 100, it is possible that he will never realize that he can do something to get a similar effect because he will miss the fact that whatever he do with Starfield.ini he will see no constellations unless he decrease generously the dense of stars in options too.

_In our open source world, if a player can have a choice, what's the point of taking it away._ For me, not even for the sake of the accuracy of the simulation.

nozmajner:
I haven't yet tried it, I don't have the facilities to build right now, and the Amount of BG stars option seems to govern that, so I would need to compile.
And my problem is not with the stars looking a bit bigger and brighter, but how those bigger stars are rendered right now. They looked more like dots from the HUD than bright points of light on the night sky. And the last image shows exactly that.
I don't know if that could be fixed from the art side of things, or it is a rendering problem though.

And I'm certainly not against adding more choice to the player. (Might even be a good idea to expose these settings in the UI, even if it is just in the debug window)

tsoj:
I feel this is related to #5124, there I cleaned up a few unnecessary options for the background stars and also added in-game UI options to change the parameters. (Unfortunately it has some conflicts, and I don't know how to fix them).

@nozmajner the stars are basically just [an image](https://github.com/pioneerspacesim/pioneer/blob/3e3274820dc0af8dd37323d5f7e72a8ed21861c4/data/textures/star_point_2.png) rendered at different sizes. At the time of making the change, I thought that star_point_2.png looked better than star_point.png (the old image for BG stars). But what should end up in a release is up to you. Changing this won't need a recompile, just rename the files. The current filename that the game will pick up is "star_point_2.png" I think.

nozmajner:
Thanks! I'll check

jimishol:
@tsoj very nice #5124. I had very hard time as player when I was editing the Starfield.ini and was ready to quit, because i didn't realized in time that the options' parameter was so closely related to them. My experience as player is that your choices in your PR is very good and should not  get the rest of the parametrs in options back. They will not help the players and are unexpectedly sensitive on what appears on screen, even the visibleRadiusLy.
As a matter of a player's taste, i checked  "star_point.png". On my screen, I prefer yours "star_point_2.png", by far i think.

Web-eWorks:
@jimishol when #5124 is merged, will this PR still need to be considered and merged? Given the changes to the config file and the discussion here, I'm getting the impression that the mentioned PR supersedes this one.

jimishol:
> @jimishol when #5124 is merged, will this PR still need to be considered and merged? Given the changes to the config file and the discussion here, I'm getting the impression that the mentioned PR supersedes this one.

@Web-eWorks my PR tries to improve the default initial values and is about the first impression a plays gets from his screen before he start to alter his available options. That is why it includes no code at all. #5124, as I understand it, has do  with the available options to the player and not the initial values of parameters to be altered. So, it alters only lua code and the ini or default values. 
So, I believe the two PRs are totally foreign (as sets have no common element)

jimishol:
Sorry for the inconviniance. When I wrote my reply to @tsoj I had the false impression that his PR was already merged and I was playing in it. As it is not merged and it doesn't remove density of stars in options (the only parameter from options that my PR uses) I have no reasons to disagree with it.  I am happy with current state's options and since density of stars remains I wonder what options #5124 removes.

jimishol:
**After last master update i altered this PR.** 

Screenshots do not help to decide about the defaults of these two parameters. Uploaded png images are not exact as in screen during play.
Values are sensitive. The question to be answered is 'what is the best sky background to appear when players, just after installation, begins their very first game'. I filled in commit my suggestions. The matter is highly subjective.

Please, it is easy, check and propose your values for these parameters or inform that you like the current default ones. Decide and close this PR. (It was not easy for me the alteration of it, so, i will not insist and alter it again.)

impaktor:
After testing, we've decided to close this, thanks for the suggested code, though.

PR #5267, opened 2021-09-18
faction motivations

joonicks:
what motivates a faction to take actions

allow arbitrary string and let lua sort out what the effects are

### DIFF
--- a/data/factions/000_Independent.lua
+++ b/data/factions/000_Independent.lua
@@ -4,6 +4,7 @@
 local f = Faction:new('Independent')
 	:description_short('Worlds without a group allegience')
 	:description('Free and independent worlds, self governing and either self reliant or striving for it.')
+	:motivation('mixed')
 	:foundingDate(2900)
 	:expansionRate(2)
 	:military_name('Independent Space Fleet')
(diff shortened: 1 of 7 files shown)


impaktor:
This looks to add a variable that then isn't used by anything?

Gliese852:
My purely personal opinion is that a faction (thousands of worlds) cannot be so integral and homogeneous that its motivation can be clearly identified.  It seems to me again that there is a discrepancy in scale.

joonicks:
> 
> 
> This looks to add a variable that then isn't used by anything?

Id use it in my 3x factions, Julians set to science and giving extra rewards to exploration guild, also Druello pirates to piracy and double amount of pirates in their systems

Web-eWorks:
I've been thinking about the best way to pass arbitrary data to mods through an optimized C++ structure like this (or an economy def, or the ShipType struct) without having to hardcode in extra string storage for items that won't be used by mods.

I don't have a solid solution for that yet, but I will make vague noises in support of this PR on the grounds that having a faction 'ethos' attribute will give us more opportunities to vary how the player perceives different factions and overall allow more possibility for gameplay depth. Not sure this particular PR is the "correct" way to do it yet, but it's a pretty clean implementation and as long as the variable is clearly communicated as experimental and subject to change I have no objection to merging it.

impaktor:
If this variable is something that's needed in some new future feature, then I think this commit belongs in that branch, or else we're merging unused code.

joonicks:
after considering, I admit this was a bit rushed and doesnt quite function as I was hoping it would

I'll withdraw the PR but save the code. Im still trying to find ways of making docking to new stations and travel to new starsystems less of a "yet another" experience.

PR #5268, opened 2021-09-19
More debug candy 

impaktor:
## Description
Adding two scripts I often used in the past.
- One I wrote for investigating the commodity prices that I showed of in PR #4831. I think this is probably useful to others, as commodity rebalance will be tweaked more by others.
- The second "RPG", is for adjusting crime, reputation, money, and such. It's a bit wonky, since it's the first Imgui thingy I ever wrote so it has sliders and other sub-optimal input GUI because I was playing with it, and it's what worked. I'm not super motivated to spend time to perfect it, perhaps someone else can make the layout better.

2021-09-19-094418_1600x900_scrot

2021-09-19-094337_1600x900_scrot

## Issues to discuss / address
1. I created a generic modules/Debug folder, and moved @Web-eWorks's ship spawned to there (together with the two new scripts/tabs I'm adding in this PR)
2. Should folder change name? `Debug` is in .gitignore, so added an exception for modules/Debug
3. Should @Gliese852's tradeship debug script be moved to Debug folder?

## Future
My plan was/is to take the code from commodity debug tab and have some advert / person in future bar, where experienced trader with eye-patch be willing to sell info on profitable trade routes, which this script already finds.


Web-eWorks:
> 1. I created a generic modules/Debug folder, and moved @Web-eWorks's ship spawned to there (together with the two new scripts/tabs I'm adding in this PR)

Looks good to me, I appreciate having the little one-off scripts in that folder.

> 2. Should folder change name? Debug is in .gitignore, so added an exception for modules/Debug

No, if anything that gitignore line should be re-evaluated and possibly removed - should only be a top-level Debug folder or possibly a holdover from the automake build system.

> 3. Should @Gliese852's tradeship debug script be moved to Debug folder?

I'd say probably not, as the debug script is a topical part of a much larger system. It makes more sense to keep that code with the system that it's debugging as a complete module.

Web-eWorks:
Looking at the screenshots, I think I might have to bump up plans on a dedicated debug/editor UI sooner than I thought; we're starting to have debug widgets that need their own panel rather than the glorified stats-box that is the current debug tools window.

impaktor:
@Web-eWorks Any actionable objections on this PR?

...possibly I could squash the two new scripts commits into one. 

Else, I'm considering this ready.

Issue #5221, opened 2021-06-25
Game crashes on arriving on system with News Event (resurrected)


jimishol:
I commented in https://github.com/pioneerspacesim/pioneer/issues/5173
I am afraid the bug resurrected.

https://drive.google.com/file/d/1kI1V0WHwCAngIsHYLxtgccMXcDOnBUZj/view?usp=sharing
Load the above file save
Read NEWS about Gliese 644
Full fuel and buy 1t for hyperdrive
Buy 2t of grain, Request launch and jump to Gliese 644 system.
Land to any station. Upon docking the output.txt must be similar to the following
https://susepaste.org/24168872

From the above save file, game crashed 3 times in a row, so i assume it crashes always.
If you do not open the news then all is ok and you get the 10xprice benefit, despite you didn't read the newspaper.

Trying to write the present, i played the exact above save_02 file and 4 times in a row the bug had gone. Game proceeded as expected. I decided not to fill new issue and continue my game. This time i loaded save_03 that created after game continuation from save_02 and before landing on stone_pad (i think, it is the nearest station), Upon docking game crashed. I reloaded save_03 and tried to dock to the same station just for the confirmation of the bug. The game continued without bug and crash. The randomness of crash makes me speechless. As soon as i manage to get the link of save_03 i will edit the present post with the public link.
Thank you.

The link to the above save_03 is https://drive.google.com/file/d/1X-4CvAIy3cqQpCvJ1Zqpwl-GQGX7_4AE/view?usp=sharing

jimishol:
Today i saw a WANTED computers, where computer stock in station was zero. I did not opened sell option. I  traveled to nearby station and bought 2 computers because i have only Xylophis. Got back and forcefully i sold 1 computer. Upon selling the 2nd computer gamed crashed with 
Fatal: [string "[T] @modules/SoldOut/SoldOut.lua"]:20: attempt to index local 'ad' (a nil value)

Web-eWorks:
Fixed in release 20220203.

PR #5275, opened 2021-09-21
Remove news, once player docks in their system.

jimishol:
It is a way to prevent player to pick NEWS from a far away system and when
he gets there to start trading back and forth, multiple times, to a near,
few ly away, system exploiting the same news, that remains active because
they was far away when they were picked. A far away news event can last
several, up to 5 game months. An exploit that will eventually be combined
with SoldOut benefits can offer tremendous profit.

### DIFF
--- a/data/modules/NewsEventCommodity/NewsEventCommodity.lua
+++ b/data/modules/NewsEventCommodity/NewsEventCommodity.lua
@@ -378,6 +378,7 @@ local onShipDocked = function (ship, station)
 			-- print("--- NewsEvent: cargo:", cargo_item:GetName(), "price:", newPrice, "stock:", newStockChange)
 			station:SetEquipmentPrice(cargo_item, newPrice)
 			station:AddEquipmentStock(cargo_item, newStockChange)
+			table.remove(news, i)
 		end
 	end
 end


jimishol:
Please close this PR. I think it introduces the exact failure of issue #5221. Even worse the workaround of issue is not working with that. I am sorry for the bad fix. (The workaround is 'just before player docks to NEWS' destination he must save, end the game BUT not quit -else issue continues- and reload). 

The exploit can exist, though.

Web-eWorks:
@jimishol there is a button right next to "comment" that lets you close the pull request and not need to ask someone else to do it :smile:

impaktor:
I'm not sure what this aims to fix?

Making multiple (within reason) trade runs is a feature, it's not intended that the event just stops as soon as the player has made one trade. The oil crisis in the 70s lasted a while. I'm not saying it's perferctly balanced, but it's supposed to be a quite rare event, not something that's on every BBS. This is like finding (many) quarters in the sofa.

If player is lucky to have a sold out advert at the same time, that's also a feature, I'd think, but if it happens too often, it should be tweaked. 

If news event is still advertised as ongoing on the BBS, when it's not, then that's a bug.

PR #5277, opened 2021-09-24
Illegal commodities nominal price x4

jimishol:
When illegal commodities get in SoldOut module, double their nominal
price makes no difference as black markets sells already at double
that price.
We double the doubled price when commodity is illegal.


impaktor:
Looks fine to me.
I assume you've play tested it.

One possibly could have introduced a `local isIllegal = Game:....` boolean variable a few lines up, to make code more readable, as negation plus double ternary operations, but meh.

jimishol:
It worked once at 4x price. Did not wait for a second illegal soldout. I assumed if it worked once, will repeat be OK on next ones.

PR #5279, opened 2021-09-24
doubleclick pause for options

joonicks:
accessibility: open options by mouse actions

instead of needing keyboard + mouse

### DIFF
--- a/data/pigui/modules/time-window.lua
+++ b/data/pigui/modules/time-window.lua
@@ -47,7 +47,7 @@ local function displayTimeWindow()
 			Game.SetTimeAcceleration(name, ui.ctrlHeld() or ui.isMouseDown(1))
 		end
 		-- isItemHovered is true for ALL the buttons
-		if ui.isMouseDoubleClicked(0) and ui.isItemHovered(0) and (name == "paused") then
+		if ui.isItemHovered(0) and ui.isMouseDoubleClicked(0) and (name == "paused") then
 			ui.optionsWindow:open()
 		end
 		ui.sameLine()
--- a/data/pigui/modules/time-window.lua
+++ b/data/pigui/modules/time-window.lua
@@ -23,7 +23,6 @@ local button_size = Vector2(32,32) * (ui.screenHeight / 1200)
 local frame_padding = 3
 local bg_color = colors.buttonBlue
 local fg_color = colors.white
-local mTicks = 0
 
 local function displayTimeWindow()
 	-- HACK: Don't display the time window if we're in a bespoke view
@@ -48,13 +47,8 @@ local function displayTimeWindow()
 			Game.SetTimeAcceleration(name, ui.ctrlHeld() or ui.isMouseDown(1))
 		end
 		-- isItemHovered is true for ALL the buttons
-		if (name == "paused") and ui.isItemHovered(0) then
-			if ui.isMouseDown(0) and (200 >= (Engine.ticks - mTicks)) then
-				ui.optionsWindow:open()
-			end
-			if ui.isMouseReleased(0) then
-				mTicks = Engine.ticks
-			end
+		if ui.isItemHovered(0) and ui.isMouseDoubleClicked(0) and (name == "paused") then
+			ui.optionsWindow:open()
 		end
 		ui.sameLine()
 	end
--- a/data/pigui/modules/time-window.lua
+++ b/data/pigui/modules/time-window.lua
@@ -23,6 +23,7 @@ local button_size = Vector2(32,32) * (ui.screenHeight / 1200)
 local frame_padding = 3
 local bg_color = colors.buttonBlue
 local fg_color = colors.white
+local pButt = Vector2(frame_padding * 2 + 32 + 15, ui.screenHeight - (frame_padding * 2 + 32 + 20))
 local lastPause = 0
 
 local function displayTimeWindow()
@@ -46,11 +47,14 @@ local function displayTimeWindow()
 		local time = name
 		if (name == "paused") then
 			time = lc.PAUSED
-			if ui.isMouseDown(0) and (200 >= (Engine.ticks - lastPause)) then
-				ui.optionsWindow:open()
-			end
-			if ui.isMouseReleased(0) then
-				lastPause = Engine.ticks
+			local mpos = ui.getMousePos()
+			if mpos.x < pButt.x and mpos.y > pButt.y then
+				if ui.isMouseDown(0) and (200 >= (Engine.ticks - lastPause)) then
+					ui.optionsWindow:open()
+				end
+				if ui.isMouseReleased(0) then
+					lastPause = Engine.ticks
+				end
 			end
 		end
 		local tooltip = string.interp(lui.HUD_REQUEST_TIME_ACCEL, { time = time })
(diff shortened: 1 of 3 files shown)


data/pigui/modules/time-window.lua
```diff
@@ -42,10 +42,16 @@ local function displayTimeWindow()
...
-		if time == "paused" then
+		local time = name
+		if (name == "paused") then
 			time = lc.PAUSED
+			if ui.isMouseDown(0) and (200 >= (Engine.ticks - lastPause)) then
+				ui.optionsWindow:open()
+			end
+			if ui.isMouseReleased(0) then
+				lastPause = Engine.ticks
+			end
```
Web-eWorks, line 54: Please prefer the use of `ui.isMouseDoubleClicked(0)` (refer to pigui/libs/forwarded.lua for precise naming) instead of manually tracking double-click time.

Web-eWorks (changes requested):

data/pigui/modules/time-window.lua
joonicks: done

(it didnt work the first time I tried it, probably because I used (1) as params)

data/pigui/modules/time-window.lua
```diff
@@ -42,10 +42,14 @@ local function displayTimeWindow()
...
 			color = colors.white
 		end
-		local time = name
 		-- translate only paused, the rest can stay
-		if time == "paused" then
+		local time = name
+		if (name == "paused") then
 			time = lc.PAUSED
+			local mpos = ui.getMousePos()
+			if mpos.x < pButt.x and mpos.y > pButt.y and ui.isMouseDoubleClicked(0) then
```
Web-eWorks, line 50: Please use `ui.isItemHovered()` and move this test to after the `ui.coloredSelectedIconButton(...)` line (line 59).

Web-eWorks: isItemHovered and the related family of isItem* functions operate on the interaction box of the last submitted item, which means they have to be run after the button is actually drawn; however, they remove the need to manually calculate and track the button interaction rect.

joonicks: tried without matching x, all buttons triggered

Web-eWorks: That's because this function is run for each button - if you want the code to only run for the pause button, you need to replace the term that's calculating the rect with a term that checks for `time == paused`.

impaktor:
Is pause button the "right place". Doesn't pause usually toggle pause/unpause?
I know we also have the play button, but just thinking about what's players expect.

Alternative solution:
- click pixel 0,0 / (top left corner)
- right click pause
- right click somewhere else?
- pause automatically opens options menu? (So one doesn't do in-game actions when paused, as that could be considered cheating, or unrealistic)

@nozmajner thoughts?

nozmajner:
It was on the pause button in Frontier as far as I can remember. Or rather it turned into a stop button that would bring the settings up. Which made sense to me back then.
In our case pause isn't toggling play. Which I don't see as a problem.
I think pause alone shouldn't bring up the settings. If we want to keep the player from doing any interaction, then maybe gray up the interface stuff that wouldn't work? And for screenshots, the ability to move around the camera, and a menu not obscuring the view can be very handy I think.

joonicks:
altered as requested

@impaktor pause button does not toggle (in pioneer)

enforcing options to be open always on pause would interfere with screenshots

impaktor:
Have the requested changes by @Web-eWorks been addressed? 
If so, I assume this is ready for merge?

joonicks:
@impaktor the changes have been made but now for some reason "IsMouseDoubleClicked" doesnt exist on my copy of pioneer. im baffled.

joonicks:
@impaktor simply reversing the order made it exist again. pigui is a fickle beast!

zonkmachine:
>  "IsMouseDoubleClicked" doesnt exist on my copy of pioneer. im baffled.

I get `Fatal: [string "[T] @pigui/modules/time-window.lua"]:50: attempt to call field 'isMouseDoubleClicked' (a nil value)` on all versions since d06b656aa8a218ad39602fc37d6d45ff08113383. I just assume the chance the error is on my side is pretty big but I can't make this work.

joonicks:
@zonkmachine is that with patch "unknown issue" of this patch?

zonkmachine:
Yes, with that one too, but the issue shows up in one way or the other since the introduction of `IsMouseDoubleClicked`

joonicks:
changed it back to manual doubleclick

zonkmachine:
OK, I think I found it. There is not `isMouseDoubleClicked` in `a/src/lua/LuaPiGui.cpp` and `/data/pigui/libs/forwarded.lua` so it simply hasn't been introduced yet. I tested with this code and it seem to work.


```cpp
diff --git a/data/pigui/libs/forwarded.lua b/data/pigui/libs/forwarded.lua
index aca3706ef..af377f37d 100644
--- a/data/pigui/libs/forwarded.lua
+++ b/data/pigui/libs/forwarded.lua
@@ -54,6 +54,7 @@ ui.getTextLineHeightWithSpacing = pigui.GetTextLineHeightWithSpacing
 ui.lowThrustButton = pigui.LowThrustButton
 ui.thrustIndicator = pigui.ThrustIndicator
 ui.isMouseClicked = pigui.IsMouseClicked
+ui.isMouseDoubleClicked = pigui.IsMouseDoubleClicked
 ui.isMouseDown = pigui.IsMouseDown
 ui.getMousePos = pigui.GetMousePos
 ui.getMouseWheel = pigui.GetMouseWheel
diff --git a/src/lua/LuaPiGui.cpp b/src/lua/LuaPiGui.cpp
index 42d26744e..6dc982ec2 100644
--- a/src/lua/LuaPiGui.cpp
+++ b/src/lua/LuaPiGui.cpp
@@ -1517,6 +1517,14 @@ static int l_pigui_is_mouse_clicked(lua_State *l)
        return 1;
 }
 
+static int l_pigui_is_mouse_doubleclicked(lua_State *l)
+{
+       PROFILE_SCOPED()
+       int button = LuaPull<int>(l, 1);
+       LuaPush(l, ImGui::IsMouseDoubleClicked(button));
+       return 1;
+}
+
 static int l_pigui_push_font(lua_State *l)
 {
        PROFILE_SCOPED()
@@ -2875,6 +2883,7 @@ void LuaObject<PiGui::Instance>::RegisterClass()
                { "PopID", l_pigui_pop_id },
                { "IsMouseReleased", l_pigui_is_mouse_released },
                { "IsMouseClicked", l_pigui_is_mouse_clicked },
+               { "IsMouseDoubleClicked", l_pigui_is_mouse_doubleclicked },
                { "IsMouseDown", l_pigui_is_mouse_down },
                { "IsMouseHoveringRect", l_pigui_is_mouse_hovering_rect },
                { "IsWindowHovered", l_pigui_is_window_hovered },
(END)

```

joonicks:
that doesnt explain why it DID work in some cases even before it was exposed, a real mystery

joonicks:
hopefully thats the last time this patch needs to be revisited

Web-eWorks (approved):
Code looks good to me, thanks for the many iterations!

joonicks:
Thanks to @zonkmachine too for testing and fixing

PR #5286, opened 2021-09-30
More transparent comm drop down window.

jimishol:
As player, most of the time i have system overview window open.
In current state, even if window is transparent with opacity=0.75,
i have the annoying feeling that something i miss from background.
With opacity=0.10 this feeling is much reduced.
In the rare case that background is so bright that i cannot see window's
informations, a hover of mouse upon it, as it uses other background,
reveals the missing information.

opacity_10
opacity_75

### DIFF
--- a/data/pigui/themes/default.lua
+++ b/data/pigui/themes/default.lua
@@ -102,7 +102,7 @@ theme.colors = {
 	blueFrame				= styleColors.accent_100:shade(0.50),
 	tableHighlight			= styleColors.primary_700,
 	tableSelection			= styleColors.accent_300:shade(0.40),
-	commsWindowBackground	= styleColors.primary_800:opacity(0.75),
+	commsWindowBackground	= styleColors.primary_800:opacity(0.10),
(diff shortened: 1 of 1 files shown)


joonicks:
opacity should really be a configurable setting. im not comfortable with opacity=0.1

nozmajner:
The 0.1 one feels too transparent to me as well.
How about 0.35:
screenshot-20210930-182543

I agree that ultimately this should be configurable.

jimishol:
Of course i am not able to do it a configurable option. As player, i would like editable themes for this and saving profiles.
This PR is about the default setting. That means the first impression a player gets by the default value. It is very easy for anyone to check several values. Every value have cons and pros. Preference change between dark and light backgrounds and one must count how often happens the dark and light backgrounds and at what background the information in windows is truly needed. 
Why not write down our preferences and merge the average as default? Till now i understand that you prefer to diverge well down than the default 0.75.
From my tries, during game play, i prefer 0.10. _(i have darker, less milky, background than you, because i see the constellations from mine previous PR)_

PR #5272, opened 2021-09-21
Progressive trading

jimishol:
**Progressive trading. The power of denominator's power.**

**Introduction**
By adding an exponential in ‘rn’ of updateEquipmentStock function in SpaceStation.lua function, ships’ average maximum profit is defined. In order, that ship, to be able to earn that profit from trading a particular commodity, it must find it in stock that his maximum equals to ship’s free available cargo.  From that, the added exponential, defines the price of that particular commodity. In that manner, that exponential, as the ultimate constant of the economy of the pioneer game, potentially quantifies every aspect that is related to economy.

_Current state’s trading system is flat_ because of player’s dominating best strategy.
He finds a system that major exports the most expensive commodity and trade around it. He never has to alter commodity. He never has to alter system. He never has to alter his ship by a larger one, if his current ship is able to  trade about 144000 credits of that commodity. The only reason to do any of the above is because he is boring to repeat the same routes with the same ship and trade the same commodity. In the same time, he likes the screen images of game flights, wanting to continue playing it. However, the boring element should not be involved at all as an element of player’s strategy. _I cry this out from personal experience because I am a pure player that just try to improve his playing experience._

There is no reason for developers to spent their time and efforts to alter commodity prices (including that of precious metals), to alter ship or equipment prices, to seek for the best prices for them, to copy commodity prices from other games, to compare prices with other games, to try to lure players by undetectable by players raise of stock or prices by 1%, 2%, 5% or whatever. Nor to blame for their failures the large variation of stocks or the absence of export on some commodities, trying to decrease it. 
_The above dominating player’s strategy will be unbeatable in all cases._
Neither seems right to alter missions’ rewards to compensate the boring trading. Missions will either dominate or equalize trading profit, so missions, as less boring, will be preferred from trading  and that will erase effectively the trading activity from the game, or missions will be auxiliary to trading and will not help against the boring best strategy.
I analyzed many of that reasons in my pdf article at https://github.com/jimishol/Pioneerspacesim_trading_economy. Anyone wants to read it must keep focus on article’s main declaration, that **_the current code is enough to reveal the present, but hidden, relations between commodity prices and stock, ship profits and their prices._**
By the realization of the effects of adding the exponential, along with the relationships it reveals, the way is open for developers to fine adjust these relationships at will. There is no point to copy prices from other games. Neither other games will accomplish anything copying prices from pioneer. The source of success is not in prices. It is in the existence of the exponential and the relationships it reveals.

[...]

PR #5291, opened 2021-10-03
Make (Bulk) Trading Profitable Again + Station Restocking

### LINKED ISSUE #5149: Station Menus Ocassionally Unusable
### Observed behaviour
Occasionally, some of the station menus will disappear and not be usable. The following workarounds were attempted without success.
1. Leave and redock at the same station.
2. Restart the game.
3. Go to another station within the same system.
4. Load an earlier game save where the player is docked in another system. 

There seem to be two ways around this bug. One is to jump to different systems until the station menus appear again. However, this can be difficult because the Lobby and Commodity Market menus are not usable, and therefore fuel cannot be purchased normally. Another is to go to a different station within the system and keep reloading the save until the menus work again. 

This screenshot demonstrates what the station menus look like in their broken state. 
Station Menu Bug

### Expected behaviour
Station menus should always be functional. 

### Steps to reproduce
I'm not sure how to reproduce this given that it seems to happen randomly. 

ver 20210203 (142693b), Ubuntu 20.04 LTS

### LINKED ISSUE #5233: Spacestation.lua attempt to index field '?' (a nil value)
**Observed behaviour**
Loaded my savegame. Went to the already docked station and switched to the commodities or equipment tab and seen the error.

**Expected behaviour**
Obviously the GUI should function without issues.

**Steps to reproduce**
Unknown as I don't know what the cause is. Sounds like someone that understands the code needs to do some digging (I can only read a little bit of powershell and that has nothing to do with a real useful scripting language :> )

**Attachments**

[output.txt](https://github.com/pioneerspacesim/pioneer/files/6889377/output.txt)

Web-eWorks:
## Bulk Trading Changes

This PR addresses a problem very astutely noticed by @jimishol in the economic and trading simulation portion of the game - once the player acquires a ship with ~80t of cargo space, the trading gameplay loop of Pioneer is effectively over. All commodities in the game have the exact same maximum profit for a single trade run, and the only variance is in how much hull space is required to make that exact same amount of money.

This means that once you have found an export/import run for Precious Metals, you have hit the maximum profit ceiling in trading. Sure, you can run a second commodity at the same time in a larger hull, but that costs more fuel, takes more real player time, usually requires a second chain of hyperjumps, and the profit margin is much lower than your primary commodity.

I promised charts, so charts you will have!

image
(Investment and Profit use left axis, Tonnage uses right axis.)

The current paradigm (which I will call "linear stocking") uses a simple formula to determine the amount of commodities that are present at a station when you enter the system. There is some variance in stock from station to station and from system to system based on the price modification from underlying system economy variables, but that variance has comparatively little effect on the primary viable trade route - buy a commodity at a major export station and sell at a minor import station, yielding approximately 30-35% profit.

With linear stocking, the "midgame" of trading ships (that is, medium haulers) is in fact the endgame of trading in Pioneer; you can spend more money on a larger ship, but you will never make any more money hauling a different cargo in a larger hold than you will hauling Precious Metals in a comparatively small hold.

This problem stems from the fact that the current stocking algorithm only takes into account a single factor - the price of the commodity - to cap the maximum profit that can be made in a single run. It ignores the other implicit factors of cargo space (price to acquire the ship), fuel costs (often 3-10x higher in a large bulk ship), and hyperjump range needed to haul larger cargoes.

This PR addresses that problem by introducing a new station stocking algorithm (which I call "exponential stocking"). It's based on the core of #5272, and uses an exponential term in the function that calculates a station's maximum stock to better model the factors of commodity rarity and time-and-resource cost to the player to haul bulk commodities, resulting in a higher credit-value profit from hauling bulk cargo while leaving the actual profit margin of 30-35% entirely untouched. 

image
(Investment and Profit use left axis, Tonnage uses right axis.)

This new algorithm provides a better and more balanced economy that gives both large bulk ships and smaller cargo/courier ships like the Amphiesma and the Sinonatrix a role to play in trading and a natural progression through the game. Precious Metals is now a rare, low-volume commodity, being the best good to haul in a small ship but dropping off in absolute profit once the player has worked their way to a Mola Mola or similar entry-level medium hauler. By the time the player has purchased a Skipjack or larger ship, Robots or even Medicines are the better cargo to haul, giving the player an incentive to change it up and not just run the same route between two systems for the rest of the game until they get bored.

image
You can see the source data for all of these charts [here](https://docs.google.com/spreadsheets/d/1wR9-Pof1Tu69GAPbdOjLraKxaP8ZqrQ9PcczVhd8AwE/edit).

## Dynamic Station Restocking

Now, merely that change alone would solve the stated problem by giving bulk haulers a reason to haul textiles in their massive holds instead of being a glorified courier for Precious Metals, but there's something else that can be done to improve the trading gameplay as well, and something that lays the groundwork for a dynamic, simulated economy.

Stations now no longer re-roll their stock values for every commodity each time you re-enter the system, and instead persistently track the player's purchases and re-stock to a deterministic equilibrium over time. This means that when you've bought a station entirely out of one commodity stock, you can't just jump out and jump back in to refresh that station; you'll have to find another station in the system and buy from that station instead.

Stations will over time recover back to their "equilibrium" stock amounts (being a combination of a very lightweight approximation of NPC trade and commodity production happening in the background and the station interface being a front for multiple trading companies with their own exclusivity contracts and stock reserves) as you jump between systems and trade with other stations, eventually becoming a viable part of the player's trade route again.

The constants behind these two algorithms are currently tuned to allow the player to make one "max profit" run from a station before needing to use another station in the same system, with the station restocking after 3-6 such trade runs. This could be further changed after merge in several directions; either to use a lower equilibrium stock and restock much faster (while the player is on the pad, potentially, though I don't think this has much gameplay utility), or to have a higher equilibrium stock and restock much slower, potentially taking a year or more to get back to equilibrium stock after heavy bulk trading.

However, I don't think tuning in either direction is actually the best way to go, as I've neglected to mention how these new features interact with other potential future changes to the economy - one of the biggest is giving each station its own economy type and making export/import price modifications be relative to a specific station or planet rather than system-wide. This would greatly amplify the impact of limited station stocks as instead of N stations in the system to buy that stock from, you could have as little as 1-2 stations that sell the commodity you want at the right price.

Other changes that could impact and provide even more depth would be dynamic commodity prices based on available stock (buy from the highest-stock station and sell to high-demand); or basing station stock on station population and industry/agriculture production of the station and parent bodies during system generation - these values are currently too random and unpredictable to be used directly.

## General Improvements

And of course, I've saved the best for last - I've rewritten the way SpaceStation.lua caches data about visited stations, adding an actual `visited` cache instead of using the advertisement register. This wound up resolving (or at least mitigating the cause) of the `GetEquipmentStock()` bug that's been causing many crashes and issues over the past few months. This supersedes and will close #5284.

I've also added a `Game.GetStartTime()` function for modules (e.g. hyperdrive breakdown) to use to determine when the loaded game was started; unfortunately it's inaccurate for old saves, but going forward that data will be available.

## Testing Notes

I'd appreciate some eyes on this PR with regards to longer-running games; both to confirm the fix for the GetEquipmentStock() bugs and also to test the progression from Amphiesma to Sinonatrix to Mola Mola and then the "midgame". I don't expect to see any bugs shaking out of the woodwork, but I would like feedback about the experience to see if I need to tweak any of the constants further to make the design more apparent.

-------
Fixes #5149, fixes #5233. Closes #5272, closes #5284.

### DIFF
--- a/data/libs/Game.lua
+++ b/data/libs/Game.lua
@@ -16,7 +16,10 @@ end
 -- Function: GetStartTime()
 --
 -- Returns the zero-offset time in seconds since Jan 1 3200 at which the
--- current game began
+-- current game began.
+--
+-- > local seconds_since_start = Game.time - Game.GetStartTime()
+--
 function Game.GetStartTime()
 	return gameStartTime
 end
--- a/data/libs/Game.lua
+++ b/data/libs/Game.lua
@@ -1,5 +1,9 @@
 local Game = package.core["Game"]
 local Event = require 'Event'
+
+-- Simple way to track the start date of the game until DateTime objects are exposed to lua
+local gameStartTime = 0
+
 Game.comms_log_lines = {}
 Game.AddCommsLogLine = function(text, sender, priority)
 	 table.insert(Game.comms_log_lines, { text=text, time=Game.time, sender=sender, priority = priority or 'normal'})
@@ -9,8 +13,31 @@ Game.GetCommsLines = function()
 	return Game.comms_log_lines
 end
 
+-- Function: GetStartTime()
+--
+-- Returns the zero-offset time in seconds since Jan 1 3200 at which the
+-- current game began
+function Game.GetStartTime()
+	return gameStartTime
+end
+
 Event.Register('onGameStart', function()
 	Game.comms_log_lines = {}
+	gameStartTime = Game.time
 end)
-	
+
+Event.Register('onGameEnd', function()
+	gameStartTime = 0
+end)
+
+local function _serialize()
+	return { startTime = gameStartTime }
+end
+
+local function _deserialize(data)
+	gameStartTime = data.startTime or 0
+end
+
+require 'Serializer':Register('Game', _serialize, _deserialize)
+
 return Game
--- a/data/libs/SpaceStation.lua
+++ b/data/libs/SpaceStation.lua
@@ -90,14 +90,10 @@ local function createStationMarket(station)
 
 	local h2 = Equipment.cargo.hydrogen
 	for key, e in pairs (Equipment.cargo) do
-		if e.purchasable and e.price > 0.0 then
+		if e.purchasable and e.price > 0.0 and e ~= h2 then
 			local rn, targetStock = getStationTargetStock(key, station.seed)
-			if e == h2 then
-				equipmentStock[station][e] = Engine.rand:Integer(rn/4, rn)
-			else
-				storedStation.commodities[key] = targetStock
-				equipmentStock[station][e] = targetStock
-			end
+			storedStation.commodities[key] = targetStock
+			equipmentStock[station][e] = targetStock
 		end
 	end
 end
@@ -126,7 +122,7 @@ local function updateStationMarket (station)
 		local rn, targetStock = getStationTargetStock(key, station.seed)
 
 		for i = 1, math.floor(timeSinceUpdate / kTickDuration) do
-			stock = stock + (randRestock:Number() + randRestock:Number()) / kAvgTicksToRestock * targetStock
+			stock = stock + randRestock:Normal(1, 1) / kAvgTicksToRestock * targetStock
 		end
 		stock = math.min(targetStock, math.ceil(stock))
 
@@ -142,21 +138,12 @@ end
 -- commodity stock information from persistent data
 local function createEquipmentStock (station)
 	assert(station and station:exists())
-	if equipmentStock[station] then return end
+	if equipmentStock[station] then error("Attempt to create station equipment stock twice!") end
 	equipmentStock[station] = {}
 
 	for _,slot in pairs{"laser", "hyperspace", "misc"} do
 		for key, e in pairs(Equipment[slot]) do
 			equipmentStock[station][e] = Engine.rand:Integer(0,100)
-		end
-	end
-
-	for key, e in pairs(Equipment.cargo) do
-		if e.purchasable and e.price < 0.0 then
-			-- "haulaway" commodities
-			-- approximately 1t rubbish for every 10 people
-			local stock = math.floor(station:GetSystemBody().population * 1e8 / math.abs(e.price))
-			equipmentStock[station][e] = Engine.rand:Integer(stock / 10, stock)
 		end
 	end
 
@@ -166,6 +153,20 @@ local function createEquipmentStock (station)
 		for key, stock in pairs(stationMarket[station.path].commodities) do
 			local e = Equipment.cargo[key]
 			equipmentStock[station][e] = stock
+		end
+	end
+
+	local h2 = Equipment.cargo.hydrogen
+	for key, e in pairs(Equipment.cargo) do
+		if e.purchasable and e.price < 0.0 then
+			-- "haulaway" commodities
+			-- approximately 1t rubbish for every 10 people
+			local stock = math.floor(math.max(station:GetSystemBody().population * 1e8, 100) / math.abs(e.price))
+			equipmentStock[station][e] = Engine.rand:Integer(stock / 10, stock)
+		elseif e == h2 then
+			-- make sure we always have enough hydrogen here at the station
+			local rn, targetStock = getStationTargetStock(key, station.seed)
+			equipmentStock[station][e] = Engine.rand:Integer((targetStock + 1)/4, targetStock + 1)
 		end
 	end
 end
@@ -774,7 +775,7 @@ local function createStationData (station)
 	visited[station] = true
 
 	createEquipmentStock(station)
-	createStationMarket(station)
+	if not stationMarket[station.path] then createStationMarket(station) end
 	local shipAdsToSpawn = Engine.rand:Poisson(N_equilibrium(station))
 	addRandomShipAdvert(station, shipAdsToSpawn)
 
--- a/data/libs/SpaceStation.lua
+++ b/data/libs/SpaceStation.lua
@@ -48,21 +48,20 @@ local stationMarket = {}
 -- commodity is an import/export good at this port (based on pricemod)
 local function applyStockPriceMod(maxStock, stock, pricemod)
 	if pricemod > 10 then --major import, low stock
-		return stock - (maxStock*0.10)     -- shifting .10 = 2% chance of 0 stock
+		stock = stock - (maxStock*0.10)     -- shifting .10 = 2% chance of 0 stock
 	elseif pricemod > 4 then --minor import
-		return stock - (maxStock*0.07)     -- shifting .07 = 1% chance of 0 stock
+		stock = stock - (maxStock*0.07)     -- shifting .07 = 1% chance of 0 stock
 	elseif pricemod < -10 then --major export
-		return stock + (maxStock*0.8)
+		stock = stock + (maxStock*0.8)
 	elseif pricemod < -4 then --minor export
-		return stock + (maxStock*0.3)
+		stock = stock + (maxStock*0.3)
 	end
-	return stock
+	return math.max(0.0, math.floor(stock))
 end
 
 -- set commodity stocking based on price adjusted for some rarity curve by exponent
--- math.log() flattens the curve at the low end
 local function getMaxStock(price)
-	return 750000 * math.max(0.1, math.log(price, 10)) / price^1.457
+	return 950000 / price^1.387
 end
 
 -- return the persistent equilibrium stock for a commodity based on a deterministic seed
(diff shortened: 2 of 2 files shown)


jimishol:
### Well done!
by a player.

The GetMaxStockForPrice(price) function means that investment equals to Price * Stock = constant / Price^(a-1).

That gives a huge tool in developers hands. Developers should really find time to play this PR for real and feel the difference it brings in game play. Initial game play is not affected at all. So, take amphiesma SetMoney to 100000, to save time, and try to acquire dsminer as soon as possible. It would be needed about 125 trades or about 12.5 hours of play time. A liitle harder than in current state where it is needed 9.7 hours.  _(Download [Calibration.lua](https://github.com/jimishol/Pioneerspacesim_trading_economy/blob/main/docs/sturnclaw/Calibration.lua), [ship_table.lua](https://github.com/jimishol/Pioneerspacesim_trading_economy/blob/main/docs/sturnclaw/ship_table.lua), [FloydWarshall.lua](https://github.com/jimishol/Pioneerspacesim_trading_economy/blob/main/docs/sturnclaw/FloydWarshall.lua) in same folder and run the last one if you are curious about how i estimated this. This is another example of what can be achieved by this PR)._ Duration of 12.5 hours seems still small for me as player. From personal experience, i propose to try to double by design that duration.

By playing this PR, i welcome the feature of not being profitable to visit same station in row but, i would like motivation to change systems, not just stations in same system. So, i welcome any thought to increase duration of restock. But even that has problem in near Sol systems. Near Sol i find many systems with huge number of stations. So, there is always enough stations to do A1,B1,A2,B2,A3,B3, A4... and so on trips between same pair of systems A, B. For this matter, take a thought if possible to limit number of stations a system can have.

**Very Well Done!!**
Accept this merge and the game will be greatly improved in future, by ways that maybe can not be clearly estimated today.

data/libs/Game.lua
```diff
@@ -9,8 +13,31 @@ Game.GetCommsLines = function()
 	return Game.comms_log_lines
 end
 
+-- Function: GetStartTime()
+--
+-- Returns the zero-offset time in seconds since Jan 1 3200 at which the
+-- current game began
```
impaktor, line 19: Should one explain why player might need it? I.e. a new game doesn't start the timer on 0? Maybe implicitly understood, but my lazy brain would miss it. I.e. an example perhaps? 
```lua
--
-- > local seconds_since_start = Game.time() - Game.GetStartTime()
--
```

Should one also mention that it's _in game_ time?

impaktor (commented):
> Precious Metals is now a rare, low-volume commodity, being the best good to haul in a small ship but dropping off in absolute profit once the player has worked their way to a Mola Mola or similar entry-level medium hauler. By the time the player has purchased a Skipjack or larger ship, Robots or even Medicines are the better cargo to haul, giving the player an incentive to change it up and not just run the same route between two systems for the rest of the game until they get bored.

Sounds very good, and this certainly is an improvement, so it's a step in the right direction. I don't see any downsides with this.

@jimishol & others: some things to playtest:
- Are we still getting "sold out" adverts some times? There's a 50% chance that _if_ market has 0 in stock of some item, then there'll be a "sold out" advert on the BBS.
- That H2 never sold out
- You could increase the `eventProbability` probability and set `minTime=0` in the NewsEvent module, and check that these still work:
  - commodity hinted at in system reported, has either zero or very much in stock, and price is either much higher or much lower, respectively

data/libs/SpaceStation.lua
```diff
@@ -744,6 +882,8 @@ Event.Register("onGameEnd", function ()
 	nextRef = 0
 	equipmentStock = {}
 	equipmentPrice = {}
+	stationMarket = {}
```
Gliese852, line 885: It seems `destroySystem()` has already done this

data/libs/Game.lua
Web-eWorks: I don't consider it high-priority as the name is relatively self-documenting, but I'll add that usage line.

data/libs/SpaceStation.lua
Web-eWorks: I noticed this as well during review; this is leftover cleanup code from an earlier refactor of the file and doesn't really harm anything to clear the table twice. I'll clean it up as a small commit when I get closer to merge.

data/libs/SpaceStation.lua
```diff
@@ -711,15 +838,26 @@ Event.Register("onGameStart", function ()
...
 
 Event.Register("onShipDocked", function (ship, station)
-	if ship ~= Game.player or SpaceStation.adverts[station] then return end
-	createStationMarket(station)
+	if ship ~= Game.player then return end
+
+	if not visited[station] then
+		createStationData(station)
+	else
+		updateStationMarket(station)
```
Gliese852, line 859: hmm, this is updated only when re-docking, it turns out if the player wants to stay longer at the station, he will see that nothing is happening to the market?
and if it then undocks and docks, there will be a sharp change in the market?

data/libs/SpaceStation.lua
```diff
@@ -35,43 +35,155 @@ function SpaceStation:Constructor()
...
+	assert(station and station:exists())
+	if not stationMarket[station.path] then return end
+
+	local storedStation = stationMarket[station.path]
+	local lastStockUpdate = storedStation.lastStockUpdate
+	local timeSinceUpdate = Game.time - lastStockUpdate
+	if timeSinceUpdate <= kTickDuration then return end
+
+	-- make sure the next tick happens at the correct time
+	storedStation.lastStockUpdate = lastStockUpdate + math.floor(timeSinceUpdate / kTickDuration) * kTickDuration
```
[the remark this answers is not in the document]
Gliese852: nevermind, I was stupid

data/libs/SpaceStation.lua
Web-eWorks: There is some possibility to observe a change in the market when redocking, though given that updates happen on a timescale of in-game weeks it takes a non-trivial amount of time even sitting on a pad at max time acceleration to reach even a single update tick, and said single tick has a comparatively small effect on the amount of stock available at the station.

data/libs/SpaceStation.lua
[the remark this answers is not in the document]
Web-eWorks: This line in particular is meant to make sure that updates happen regularly every single week since start, even if the player jumps back into the system a week and a half later and there's half a week of time that isn't being processed.

data/libs/SpaceStation.lua
Web-eWorks: Part of my hope is to discourage this behavior and encourage jumping out of the system or trading multiple commodities instead - the average player will not know the precise details of when the market restocks and at what interval, only that it happens sometime during their travel away from the station.

Additionally, I got entirely sidetracked by the discussion because updateSystem already does this and updates all stations each hour.

impaktor:
As pointed out on IRC the other day, can a market spawn with 0 in stock for some commodity? I.e. have that equilibrium state, or is only possibility for 0t if player buys up the market? 

Asking, since SoldOut advert module doesn't _set_ commodity suply to 0, but instead monitors the market and _then_ spawns advert (with 50% probability) to buy the sold out commodity.

Could the player use this? Buy all available precious metal. Jump out, jump back, and now sell back to someone offering x2 price?

Web-eWorks:
> As pointed out on IRC the other day, can a market spawn with 0 in stock for some commodity? I.e. have that equilibrium state, or is only possibility for 0t if player buys up the market?

Yes. The chance of a commodity spawning at 0 is the same as before, but you get fewer "tries" at it as stock values aren't rerolled every time the player re-enters the system.

> Asking, since SoldOut advert module doesn't set commodity supply to 0, but instead monitors the market and then spawns advert (with 50% probability) to buy the sold out commodity.

You probably won't see as many SoldOuts with these changes, but this does open up the opportunity for a more dynamic market ecosystem (among other things I want to add after this is merged is a separate "demand" tracker in addition to the current supply tracker) that SoldOut can make better use of then just how much of a commodity is in stock at the station.

> Could the player use this? Buy all available precious metal. Jump out, jump back, and now sell back to someone offering x2 price?

They theoretically could, although it's more probable that the commodity would restock to be back above zero when they jump back in. In either case, I consider this internally consistent behavior (SoldOut triggers on zero stock, doesn't care about how it got there); this is something where changing it would require tracking both supply and demand of a given commodity.

data/libs/SpaceStation.lua
Web-eWorks: This function duplicates most of the code from destroySystem() - I'm going to leave it as is for locality's sake.

PR #5292, opened 2021-10-04
debugRPG no crash on end game

joonicks:
if debug is open and you "end game", debugRPG goes boom

for @fluffyfreak


impaktor:
I'm not surprised. I bet there are more of those from the two tabs I merged, e.g. when in hyperspace, or the above case.

PR #5298, opened 2021-10-11
Add compile-time TypeId and string hashing, add unit testing framework

sturnclaw:
Integrate [onqtam/doctest](https://github.com/onqtam/doctest) as a (fast!) unit testing framework and integrate it into the CI passes for both linux-native build tasks. Someone else will need to figure out the incantation to launch the unit test binary on the windows build targets.

Adding new unit tests is as simple as creating a new cpp file in the `src/test` directory and writing as many test cases as are needed to test the code you're working on - I'd like to see the number of tests grow over time and though I won't be requiring it as a merge requirement, PRs that add test cases for the code they're changing get a thumbs-up from me. :smile:

I've finally completed the last step in the render API rewrite and added a compile-time string hashing function for string literals and inlined 90% of the previous `Renderer::GetName()` calls. As a follow on to this, I've added a compile-time type ID generator as a "canonical" version of the two other ad-hoc solutions I've written on separate work branches using a technique that I got from [skypjack's blog](https://skypjack.github.io/2020-03-14-ecs-baf-part-8/). This isn't directly used by anything right this second, but will be a requirement for landing my gun refactor branch.

I've also changed the initialization semantics of vectors and quaternions to always explicitly initialize their values; this cuts down on the number of unanticipated bugs coming from accidentally uninitialized memory for what should be a frankly unnoticeable performance cost - we're not creating giant arrays of vectors and in general it's only an extra 4 8-byte stores at maximum.

### DIFF
--- a/src/Shields.cpp
+++ b/src/Shields.cpp
@@ -105,7 +105,7 @@ void Shields::Init(Graphics::Renderer *renderer)
 	// Create a global shield data buffer containing "nothing" for use when there isn't an active Shields class
 	// attached to the model
 	s_matUniformBuffer.Reset(renderer->CreateUniformBuffer(sizeof(ShieldData), Graphics::BUFFER_USAGE_STATIC));
-	s_matUniformBuffer->BufferData(ShieldData {});
+	s_matUniformBuffer->BufferData(ShieldData{});
 
 	s_matShield.Reset(renderer->CreateMaterial("shield", desc, rsd));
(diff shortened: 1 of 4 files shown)


Web-eWorks:
If there are no outstanding reviews, I'll go ahead and merge this branch sometime today/tomorrow - have a few other branches I need to work on and PR in.

PR #5300, opened 2021-10-19
PropertyMap 2.0

Web-eWorks:
Rewrite the PropertyMap / PropertiedObject implementation to use a C++ native hash map for efficiency and allow the property map data to be persisted across Lua states.

This PR adds a StringName class for immutable hashed strings stored in a per-thread hashmap with minimal copy overhead (it's internally a reference counted pointer). This enables fast string equality tests and lets the Property object fit in 32 bytes (a single cache line). It matches the usecase for strings pulled from Lua fairly well.

Further work could replace quite a number of uses of std::string (e.g. Body labels) in the codebase with StringName to avoid the allocate-on-copy overhead of std::string.

I've added a Property class as a tagged union of common datatypes used by the codebase - it supports vectors, quaternions, colors, mutually-convertible integer/double numbers, strings, and child PropertyMaps. It's intended as a semi-type-erased container for arbitrary data stored in the heap; combined with PropertyMap, its primary usecase is as a "native" datatype for sending JSON data to Lua for mods to use when adding extra keys to the economy data or ship JSON files without needing to add class fields for every possible extra datum that mods might add via JSON-patch files.

I plan to expand it at some point to also support an integer-addressed sub-array of Property objects as well, though I've deferred that until I have a more concrete usecase loading JSON directly from data files.

PropertyMap is a simple insert-only/never-erase open-addressing hash table that exploits the usage patterns of the existing `setprop()` code - property slots are rarely ever used only once, and a property name that is unset is usually set again later. It's meant to have minimal runtime complexity, memory overhead, and still have acceptable lookup performance; even though it's not using a more optimized Robin-Hood hashing scheme it still beats out the existing lua-based implementation in lookup speed by ~20% in my tests.

The StringTable is a thread-local Robin-Hood hashtable to minimize the memory overhead of StringName data - it might be a premature optimization, but later code can add statistic collection to the Reclaim() step that determines how efficient it is at de-duplicating string memory. It uses quiescent space reclamation rather than deallocating blocks of string memory when they're no longer referenced; this allows repeated creation of identical StringNames to reuse string data in the table.

-----

At this point, you're probably asking "what can I do with it?" - the biggest usecase that comes to my mind is allow mods to read custom properties from ShipDef json files without needing to patch the C++ ShipDef object. Another usecase involves the faction lua files setting data that can then be read by game code at runtime when the player enters a system owned by that faction, for example.

I'm sure there are more inventive ways to use this feature as well, but keep in mind that it's not as fast or efficient as using the contained datatypes directly and should mostly be used where you need type-erased basic datums or to support arbitrary mod-authored data along with C++ objects.


src/core/Property.cpp
```diff
@@ -0,0 +1,256 @@
+// Copyright © 2008-2021 Pioneer Developers. See AUTHORS.txt for details
```
impaktor, line 1: Anno 2022

src/core/Property.h
```diff
@@ -0,0 +1,211 @@
+// Copyright © 2008-2021 Pioneer Developers. See AUTHORS.txt for details
```
impaktor, line 1: In the year __2022__, _(people drink soylent green... seriously, the movie takes place in 2022)_

src/core/StringName.cpp
```diff
@@ -0,0 +1,205 @@
+// Copyright © 2008-2021 Pioneer Developers. See AUTHORS.txt for details
```
impaktor, line 1: // Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details

src/core/StringName.h
```diff
@@ -0,0 +1,127 @@
+// Copyright © 2008-2021 Pioneer Developers. See AUTHORS.txt for details
```
impaktor, line 1: // Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details

src/lua/LuaPropertyMap.cpp
```diff
@@ -0,0 +1,117 @@
+// Copyright © 2008-2021 Pioneer Developers. See AUTHORS.txt for details
```
impaktor, line 1: // Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details

src/lua/LuaPropertyMap.h
```diff
@@ -0,0 +1,11 @@
+// Copyright © 2008-2021 Pioneer Developers. See AUTHORS.txt for details
```
impaktor, line 1: // Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details

src/test/TestProperty.cpp
```diff
@@ -0,0 +1,213 @@
+
```
impaktor, line 1: Blaspheme! No Koppywright?!

// Copyright © 2008-2021 Pioneer Developers. See AUTHORS.txt for details

src/test/TestStringName.cpp
```diff
@@ -0,0 +1,184 @@
+
```
impaktor, line 1: // Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details

impaktor (commented):
Small lesson in how to be an annoying retard.

Web-eWorks:
Man, you could have at least waited until I rebased the branch in March when I actually plan on trying to merge the thing! :stuck_out_tongue: 

Seriously though, thanks for finding all of those, I'll fix them at some point. This PR isn't going in the release right yet, so it's low-priority.

laarmen:
Would it be possible to have some more explicit documentation around the PropertyMap implementation? For instance, there's an undocumented class invariant that the underlying vector size is a power of 2, without which pretty much all methods would be broken (due to the `idx &= size - 1` pattern). Also, has there been tests regarding inlining? Compilers are pretty smart with this nowadays.

Otherwise, I'm really liking it, especially how it makes it possible to avoid the need to query Lua in the inner ship methods to know about shields and stuff.

Web-eWorks:
As a completely on-topic remark, let me just say that I'm *very* glad for GCC's verbose error messages, as MSVC has absolutely nothing useful to say about why my code is continuing to not compile. I think I've managed to isolate some or all of the problem (GCC 9.x / MSVC bug with std::variant implicit-type-conversion construction), but... only time and a succeeding CI build will tell. On the upside, this code compiles *perfectly* fine with GCC 11 and clang from both the CI environment and my system.

fluffyfreak:
Cool, I was sailing across the mediterranean sea when this would have pinged me 😂 just catching up on (quite a few) GitHub emails now 👍

Web-eWorks:
Just in time - I was starting to think about removing the VS solution build from our CI configuration since no one uses VS anymore :stuck_out_tongue:

craigo-:
Hey! I do! 😄

Issue #5154, opened 2021-03-19
The Pads goes into the ground


py7hon:
I cant docking because The pads goes into the ground on Canwa system, Canwa B, Anderson's Grave, Fort Alvarez.
[video](https://files.catbox.moe/ttra03.mkv)

__My pioneer version (and OS):__
ver 20210203 (142693b) on: Linux

fluffyfreak:
Took a look at this but so far nothing I've tried has worked. There's a parameter to control the amount of allowed slope under the station but even tightening the tolerance for this doesn't seem to help.

Needs more investigation but on some planets I believe the only thing to do is actually flatten the terrain beneath the cities/stations.

Gliese852:
@fluffyfreak do you think it would be difficult to implement that the terrain heights a read from the mesh of a certain model? Let's say in the station model, there be some mesh, not drawn, but indicating the terrain?

fluffyfreak:
I think it needs something like https://github.com/fluffyfreak/pioneer/tree/RegionBasedTerrainControl which Ae-2222 wrote a loooong time ago, but with some performance tweaks and blending improvements.

Web-eWorks:
@fluffyfreak I noticed the terrain heightfield is sampled continuously, and if you're sampling at a higher resolution than the terrain meshes you can get wildly varying results compared to the actual surface/slope of the terrain. This can lead to false positives when trying to calculate the slope of the terrain under a starport depending on how far distant the sample points are from the actual terrain mesh vertices.

fluffyfreak:
Another location
![image](https://github.com/user-attachments/assets/86c98ded-8bc4-46f5-b660-4282713f2350)
Found by "Coronatrix HQ-0602 pilot" on Discord

Bodasey:
Now / Version 9336ec4 affected: OPLI Contemplation in the Barnard's star system - a spaceport often used when starting new games...

Feature request: refuse assigned docking bay to get one not under ground.

bszlrd:
Made a quick and dirty fix for this specific one: #6283 Changed the seed for the moon to avoid that bumpiness. Oddly enough the station was already relocated automatically to avoid it being underground, so slightly changing the coordinates didn't work.

Bodasey:
multiple / inconsistent surface data (for Deimos) on runtime?

[...]

PR #5302, opened 2021-11-06
Add TaskGraph implementation for mid-frame parallelism

Web-eWorks:
This PR opens up the ability to better take advantage of modern massively-multicore CPUs by adding in a new system to manage worker threads and dispatch work items to them. This replaces the existing AsyncJobQueue implementation (but is backwards-compatible with existing Jobs and provides a compatibility shim) with an implementation that is suitable for both "fire and forget" mid-frame tasks and long-running background tasks that need to be finalized on the main thread.

All threads are now capable of queuing tasks at any time, and there is an explicit API to allow a thread to "go wide" and split computation over all threads including the main thread (`WaitForTaskSet`). "Pinned" tasks are run at the end of the frame on the main thread (for tasks that need access to the OpenGL context or similar externally-synchronized workloads).

Jobs are now used as a low-priority background task system for longer (multi-millisecond) processing and will only be executed if there are no current Tasks to execute available to a worker thread.

I also fixed an issue where the profiler would generate invalid trace files due to the profiler recording the length of the profile before it "closed" the list of profile events, resulting in events being recorded into the trace file with timestamps after the "end" timestamp. (This showed up fairly regularly in heavily threaded code.)

### Testing Notes

I'd like to get this branch tested somewhat widely to confirm there aren't any subtle threading errors or initialization issues; @nozmajner specifically I'd appreciate a test since you managed to catch the last batch of initialization issues when I touched the Application code. You're not looking for any specific problems, just confirming the game runs and hyperspace works correctly etc; if there's a problem you'll most likely get a crash or application deadlock.


Web-eWorks:
Build-VS is failing as expected, added two new files (and another six to contrib/) it doesn't know about. Fluffy can at some point add those to the solution when he next has time to hack on Pioneer.

Bodasey:
Load a game -> click a bit in personal info -> launch request from station lobby -> the ship doesn't move, but the bulletin board is no longer available, as if you fly yet.

press ESC, close menu -> the ship lifts.

And I'm not sure if the autopilot is broken: scoped hydrogen on Jupiter, wanted to fly back, on the first try, I have chosen a target to land, on the second try, I told it to fly back in a low orbit, but both times, the autopilot flew deeper into Jupiter and the ship crashed on it's virtual surface.

fluffyfreak:
Fixed up the VS2019 projects here https://github.com/fluffyfreak/pioneer/commit/2b27096dfda87cf494d80b6031e7df9fd46f1c3c

it won't let me create a PR into your branch for some reason 🤷

Web-eWorks:
@fluffyfreak thanks! I'm still not sure why that's happening, but feel free to push directly to the PR branch in the future - I've cherry-picked in that commit so you don't have to do it now though :smile:

@Bodasey
> Load a game -> click a bit in personal info -> launch request from station lobby -> the ship doesn't move, but the bulletin board is no longer available, as if you fly yet.
> press ESC, close menu -> the ship lifts.

I'm not able to reproduce this at all on a Mars start, but something I'd like you to double confirm is whether the game was paused after you requested the launch - games are loaded in Paused state, but closing the options window puts you back into 1x time mode. I'm guessing that's responsible for the behavior you're seeing here.

> And I'm not sure if the autopilot is broken: scoped hydrogen on Jupiter, wanted to fly back, on the first try, I have chosen a target to land, on the second try, I told it to fly back in a low orbit, but both times, the autopilot flew deeper into Jupiter and the ship crashed on it's virtual surface.

Autopilot bugs like this one are well known issues by now; I'm going to consider this report unrelated to the PR specifically because autopilot navigation and gas giant body parameters aren't affected by any threaded code; the affected areas are primarily music loading on startup, terrain generation on terrestrial planets, and the initial generation of nearby sectors and systems after a game start or hyperjump to a new system.

Bodasey:
Yes, game is in pause when loaded, and keeps being paused when you launch from Station lobby.
(But, in contrary, switches to 1 time speed when you undock in main window - maybe these two should react the same way?)

Bodasey:
#5154 not (completely) fixed, still to see at Mexico City Starport. Not a big Problem here, you can land on partially blocked platforms.

Web-eWorks:
Appreciate the bug hunting Bodasey, but neither of those are actually pertinent to this PR. :smile: They're all known issues that are existing behavior in master branch already - I'm looking specifically for new issues that are caused by this change and cannot be reproduced otherwise.

Web-eWorks:
Going to merge this in during the weekend if no one's managed to find any showstopper bugs - I've added a bit more documentation to the TaskGraph implementation to hopefully make clear any nebulous points and I've not hit any issues in my testing.

Issue #5305, opened 2021-11-10
onCreateBB runs on already created bulletin boards


zonkmachine:
![crewforhire](https://user-images.githubusercontent.com/6368949/141050876-b306898e-3180-4d7d-8224-1af0ef86cc04.png)

### Observed behaviour

Since 721c4e22e89050e3f982b0fbd07ea8c14c339fdd (@Web-eWorks) there is a bug introduced that will leave multiple instances of the 'Crew for hire' ad. I think the multiple ads are a result of the `onCreateBB `event being triggered multiple times for the same station. There was another issue that went away with 721c4e2, where all the ads would disappear intermittently but the 'crew' one. That one did not involve `onCreateBB`.

### Steps to reproduce

Start a new game and observe the 'crew' add.
Restart and open the bulletin board. Repeat until things breaks...

I recommend a print statement to see when `onCreateBB` is running.

impaktor:
@zonkmachine Could you please try this on our current master, as BBS has been revamped now.
(E.g. you can find linux and windows version from 9h ago [here](https://github.com/pioneerspacesim/pioneer/actions/runs/1700377830) )

zonkmachine:
The issue is still there.

impaktor:
OK, thanks. For reference, we're now at: 9adb8fdf9fc4aa170bc520236b759f176b2a5b95

zonkmachine:
I poked the code some more but didn't get much further into the issue.
`createStationData()` is used two times in **data/libs/SpaceStation.lua** . It's the first instance that is being called when this happens.
https://github.com/pioneerspacesim/pioneer/blob/85b77d081bc019bdda4cb6034706041361e37bf0/data/libs/SpaceStation.lua#L843

Web-eWorks:
I am completely unable to reproduce this bug on latest master. No combination of new game / save game and reloading has managed to produce a situation in which onCreateBB is called more than once for a station, and never for a station that already has saved data. If you can add more details (OS, git commit hash, a stack trace of the multiple times it's called through `print(debug.stacktrace())`), I can look into the issue, but I cannot find a way to reproduce it.

zonkmachine:
> a stack trace of the multiple times it's called through `print(debug.stacktrace())`),

I'm not sure I follow here. When I added that print statement after **createStationData(dockedStation)**, I got this the second time I started with `continue game`:

```
[...]

Issue #4810, opened 2020-02-23
Faces change on the BBS


impaktor:
### Observed behaviour
Persistent characters, like FuelClub, and e.g "Carney's Warehouse" on Mars, get different faces each time you start a game.

### Expected behaviour
Same faces.

### Steps to reproduce
1. start new game on Mars
2. Go to Federal Fuel Club and/or "Carney's Warehouse", remember those horrible faces
3. restart the game, repeat.

I believe this migt have been introduced by #4775.

### code that seems to work:
Explicitly sending in the `rand` object to `Character.New()` in Goodstrader seems to solve it, but then I wonder: isn't that what the `faceseed` is for? ping @vakhoir 
```lua
local ad = {
	station  = station,
	flavour  = flavour,
	slogan   = slogan,
	ispolice = ispolice,
	faceseed = r:Integer(),
	trader   = Character.New({title = flavour, armour=false}, r),
}
```

impaktor:
I think I've understood it.

First, fuelclub isn't coded to have persistent faces, so that one isn't a bug, but BreakdownServicing, and DonateToCranks are. The `faceseed` set in goodstrader is never used, but can be passed in as parameter, but then it's called `seed`.

I think I can fix this shortly.

vakhoir:
Go ahead if you want to fix it, but just FYI I'll be going over all the mission modules while working on the Missions View. I didn't want to change them too much for the Bulletin Board, because these 2 views are interdependant. It might be a good idea to open a catch-all ticket for anything Missions/BBS related that comes up. Or I'll open a WIP PR and it can be tagged in any issue related to either Missions or BB.

impaktor:
Sounds like a good idea, however, I'd like to fix this asap, as #4796 makes a case for being able to recognize the honest from the fake goodstrader, which currently is true only by shop-title, not by face.

PR #5306, opened 2021-11-10
Fix mechanic not being persistent

zonkmachine:
The 'Breakdown Servicing' character changes face in between chats. This is because `Character.New()` is called from the `onChat` function and somehow doing it with a persistent seed doesn't give persistent result. I don't know why that doesn't work. Instead I simplified the code a bit by generating the character on `onCreateBB`.

### DIFF
--- a/data/modules/BreakdownServicing/BreakdownServicing.lua
+++ b/data/modules/BreakdownServicing/BreakdownServicing.lua
@@ -62,7 +62,6 @@ for i = 1,#flavours do
 	f.response  = l["FLAVOUR_" .. i-1 .. "_RESPONSE"]
 end
 
-local mechanic = {}
 local ads = {}
 local service_history = {
 	lastdate = 0, -- Default will be overwritten on game start
@@ -188,7 +187,7 @@ end
 local onCreateBB = function (station)
 	local rand = Rand.New(station.seed + seedbump)
 	local n = rand:Integer(1,#flavours)
-	mechanic = Character.New()
+	local mechanic = Character.New()
 
 	local ad = {
 		mechanic = mechanic,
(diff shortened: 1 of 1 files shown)


impaktor:
I'm guessing this might be related: #4810

zonkmachine:
> I'm guessing this might be related: #4810

Yes. PR https://github.com/pioneerspacesim/pioneer/pull/4817 is mentioned in issue #4810 but it doesn't look done yet.
I'll look into the modules discussed in #4810.

zonkmachine:
@impaktor I've tested DonateToCranks and GoodsTrader and they are persistent. However, there is an intermittent problem with onCreateBB being rerun on bulletin boards that are already existing, described in https://github.com/pioneerspacesim/pioneer/issues/5305, and I think this is what's failing here. It's an intermittent issue so it may take some time to trigger it. At the time of #4810, https://github.com/pioneerspacesim/pioneer/issues/5305 would have behaved differently and I haven't seen onCreateBB being triggered on those earlier builds, but I think the core issue would have been the same/same-ish.

zonkmachine:
This PR is ready for review. For the other/similar issues mentioned above I believe they need more troubleshooting and most, if not all, depend on other issues.

impaktor:
Looking at the code, doesn't look like it will be persistent, i.e. same mechanic on same station always? I think that was the intention with the old code. Or are you using the `local rand = Rand.New(station.seed + seedbump)` anywhere?

zonkmachine:
Thanks!

Right, I don't actually know how to fly around in this game and it looks like I actually have to do that now. Will try it... :)

For the code I guess I have to go:

```lua
 local onCreateBB = function (station)
        local rand = Rand.New(station.seed + seedbump)
        local n = rand:Integer(1,#flavours)
-       local mechanic = Character.New()
+       local mechanic = Character.New({seed=rand:Integer()})
```

Issue #5282, opened 2021-09-25
Money Exploit: BBS Fuel Club


joonicks:
1 Become a member
2 Buy X Military fuel from the club
3 Sell X Military fuel to the station market
4 profit
5 Buy X Radioactives from station market
6 Sell X Radioactives to fuel club
7 profit more

Fuel clubs better track what their members sell on the market and enforce terms of service

Bodasey:
Agreed that arbitrage in reality is not possible forever.

Works in the game also with hydrogen (buy at half, sell at full price).

Maybe a solution is that Hydrogen from fuel corporations can not be sold back to markets and is burnt first, or corporations sell deuterium who has other properties than hydrogen as fuel.

PR #5311, opened 2021-12-22
FuelClub: Refuel internal fuel tank only once a day

robothauler:
Members get refueled every time and can sell the internal fuel over and over again.
This PR prevents that and limits the refilling to once a day after takeoff.

### DIFF
--- a/data/lang/module-fuelclub/en.json
+++ b/data/lang/module-fuelclub/en.json
@@ -73,7 +73,7 @@
   },
   "LIST_BENEFITS_FUEL_TANK": {
     "description": "",
-    "message": "Fuel tank refilling, where necessary"
+    "message": "Fuel tank refilling, once per day"
   },
   "LIST_BENEFITS_JOIN": {
     "description": "",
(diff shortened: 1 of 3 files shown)


impaktor:
I think there should also be edits to the language file, so when you ask for the rules/contract there's a "we'll refuel you once per day", or similar, somewhere, where it's fitting.

impaktor:
If anyone has play tested this / tried to get it to break, please let us know.

I assume we're aiming for merging stuff for a February 3rd release again...

Bodasey:
played a little bit, found this:

1. "refuel full" your tank in the station's lobby
2. enter fuel foundation (or another fuel organization)
3. pump down fuel to freight room
4. enter fuel foundation again

=> no refill, the first call of the foundation has blocked the function yet, although no fuel has been transferred.

I assume the same behave when you come back from fuel scooping and like to sell hydrogen to the foundation (fuel scooping always fails in this branch from code heritage). It will fill the missing 0.xxx tons when you first enter the fuel foundation as you can only transfer integer amounts to the tank.

I thought further to block automatic refuel function if a tank has less than 3% / 3t or so capacity left, but then you maybe want to force refueling when your tank is small, so you need a button for this...

...but if you add a button there anyway, you can directly call it "free tank refill for today" and skip the automatic background function.

Bodasey:
fine work, nothing found

(maybe a one-time cheat when a game from certain earlier versions is loaded, but once you save in this version, it's safe, i'd say there is no need for more code to prevent this single event)

What about #5282? There is still some possibility for arbitrage, but the quantities are very limited now. The clubs still don't accept taking back more waste than military fuel sold, i tested this also.

PR #5323, opened 2022-01-15
Xylophis overhaul

nozmajner:
This tiny one is ready to fly!
xylophis_overhaul_023
- Has three lods: 27k, 4.7k, 2.3k
- Collision mesh has 224 tri
- Has a shield mesh (seems like shields aren't showing on master, neither in-game or in the modelviewer)

It is a tiny little van-sized ship that leans into that visual intentionally. The scale is pretty much that as well. (I'm not planning on modeling the other ships as close to their inspirations as this one, rest asured. But I think this ship is a fun little craft :) )
It has the same docking collar as the escape capsule (size 0).  Has a tag_collar_1 tag for future use.

Diffuse, pattern and normal textures are 2048x1024, spec and glow are half that size. The internals only have a 512x512 diffuse.
Has 7 patterns, but in practical sense  they are variations of two, and a full-color one. One is the A-Team pattern, the other has the bottom half differently colored, and there could be a stripe where they meet.
All textures are dds, except for the normals, to avoid any artefacts. **And on that note: it seems to me that the normals aren't working neither in the summer release nor current master.** Hard to tell if I'm right, but the ships definitelly look flat.

I did not touch any of the stats. I modeled it to be able to provide the same amount of capacities, and it is roughly the same size as well.
xylophis_overhaul_027

xylophis_overhaul_024
xylophis_overhaul_025
xylophis_overhaul_028
Has glowy bits :)

The S0 docking collar means that it fits into the inner side of the S1 collar, if the petals are retracted:
xylophis_overhaul_029

---------

xylophis_overhaul_026
screenshot-20220115-173208
screenshot-20220115-173221
screenshot-20220115-173234
screenshot-20220114-175845


ARGHouse:
Oooo space van, I like how it still has that blocky look.

Bodasey:
Good and a lot of work for sure, it's really a pity some cannot enjoy it...

now errors with libassimp3 / assimp-devel3, while libassimp5 / assimp-devel 5 is all right:
(but I know you reckon with me testing this ;-)

models/ships/xylophis/xylophis_hi.dae:
Couldn't load file
Could not find model: xylophis
CreateCollisionMesh for : (error)

xylophis_equipment

Xylophis_orbitalstarport

fluffyfreak:
Have space van, will go space camping! 😁👍

Web-eWorks:
The cockpit view is quite smashing as well!

If there are no further changes needing to be made to the model, this has my thumbs-up to go ahead and merge.

image

PR #5325, opened 2022-01-16
Make "Travellers Advice" module work with new revamped BBS

impaktor:
## Seems to work 
2022-01-16-203812_1600x900_scrot

In screenshot above, I've made a for-loop over all advises, just for demonstration purpose.

## If it doesn't work for testers, well then:

 @Web-eWorks was there any other module that I should fix? I see you already fixed the SoldOut module.


Web-eWorks:
I don't *remember* fixing the SoldOut module, so just double-check it's working please - I wasn't able to test it. Otherwise, this should be it!

PR #5333, opened 2022-01-28
Slight fix/improve atlas-view

Gliese852:
https://user-images.githubusercontent.com/18342621/151628791-8cb893e5-f909-4324-91ff-b471011a7ed7.mp4

- show icons for asteroids and orbitals only
- show switch button to atlas-view in top row
- don't show grid, L-points and ships buttons in atlas mode
- fix mouse panning and zooming
- add a button to hide the overview window
- correct overview-window size in worldview
- fix that when you click on the planet if the ship is centered, the view is reset
- show ground stations around the planet when hovering or clicking on it
- generate only one popup (and not for each object)

### DIFF
--- a/data/pigui/modules/system-view-ui.lua
+++ b/data/pigui/modules/system-view-ui.lua
@@ -162,6 +162,7 @@ end
 local Windows = {
 	systemName = layout.NewWindow("SystemMapSystemName"),
 	systemOverview = layout.NewWindow("SystemMapOverview"),
+	systemOverviewHidden = layout.NewWindow("SystemOverviewHidden"),
 	objectInfo = layout.NewWindow("SystemMapObjectIngo"),
 	edgeButtons = layout.NewWindow("SystemMapEdgeButtons"),
 	orbitPlanner = layout.NewWindow("SystemMapOrbitPlanner"),
@@ -173,6 +174,7 @@ local systemViewLayout = layout.New(Windows)
 systemViewLayout.mainFont = winfont
 
 local systemOverviewWidget = require 'pigui.modules.system-overview-window'.New()
+systemOverviewWidget.visible = true
 
 function systemOverviewWidget:onBodySelected(sBody)
 	systemView:SetSelectedObject(Projectable.OBJECT, Projectable.SYSTEMBODY, sBody)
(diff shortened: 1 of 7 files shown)


Web-eWorks:
Thoughts about the ground stations display: ideally there is some sort of single "more..." icon displayed when the display is zoomed out, and the user must zoom in to show the full ring of icons. Additionally it'd be nice to have the station icons scale with the zoom level, but that's a very trivial request and feel free to disregard it for this cycle.

It occurs to me that the system overview toggle button could be moved to the right hand sidebar with the rest of the window controls (and possibly the "Switch to Atlas View" button there removed), but I'm not certain whether that's easier to find and remember for the end user (probably not).

data/pigui/modules/system-view-ui.lua
```diff
@@ -408,6 +440,102 @@ function Windows.unexplored.Show()
...
+	else
+		return nil
+	end
+end
+
+local popup = {}
+
+function popup.make()
+	if not popup.enabled then return end
+	ui.popup("system-view-ui-popup", function()
```
Web-eWorks, line 507: Since you're hardcoding the popup ID here, instead of calling `popup.enabled = false` later down the line I'd recommend calling `ui.closeCurrentPopup` directly - the draw function will not be called if the popup itself is not open.

data/pigui/modules/system-view-ui.lua
```diff
@@ -408,6 +440,102 @@ function Windows.unexplored.Show()
...
+local function displaySurfPorts(center, radius, ports, mouseover)
+	local nports = #ports
+	if nports == 0 then return nil end
+	local icon_size = bodyIconSize.x
+	local two_pi = math.pi * 2
+	-- pretty rough estimate
+	local min_rad = icon_size * nports / two_pi
+	radius = math.max(icon_size, min_rad, radius - icon_size / 2)
+
+	local start_ang = 225 * two_pi / 360
```
Web-eWorks, line 470: You can also use `math.deg2rad()` here instead of manually doing the degrees->radians conversion.

Web-eWorks (approved):
Overall looks good, a few minor corrections but we can make it pretty later.

data/pigui/modules/system-view-ui.lua
Gliese852: Indeed! For some reason I thought that the function passed to `ui.popup` would be executed every frame in `popup.make()`, but of course this does not happen, so you don't even need to call anything, the popup itself closes when you click in it.

data/pigui/modules/system-view-ui.lua
Gliese852: Done!

Gliese852:
Fixed a few things, added a few things
demo at 1280x720

https://user-images.githubusercontent.com/18342621/151678546-89bb2445-6f5b-4caa-a0f8-75b5b5613474.mp4

- use more correct icons for ground stations
- show "plus" icon if zoom is too small
- zoom to mouse cursor position

PR #5334, opened 2022-01-30
System map fixes after play-testing

Gliese852:
- fix the assertion when hyperjumping with the system map enabled
- fix that sometimes a body with a radial orbit disappears from the system map

### DIFF
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -541,7 +541,7 @@ Body *Space::FindNearestTo(const Body *b, ObjectType t) const
 
 Body *Space::FindBodyForPath(const SystemPath *path) const
 {
-	if (!path->IsSameSystem(m_starSystem->GetPath()))
+	if (!m_game->IsNormalSpace() || !path->IsSameSystem(m_starSystem->GetPath()))
 		return nullptr;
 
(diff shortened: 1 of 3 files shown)


src/Space.cpp
```diff
@@ -541,6 +541,7 @@ Body *Space::FindNearestTo(const Body *b, ObjectType t) const
 
 Body *Space::FindBodyForPath(const SystemPath *path) const
 {
+	if (!m_game->IsNormalSpace()) return nullptr;
 	if (!path->IsSameSystem(m_starSystem->GetPath()))
```
Web-eWorks, line 545: Would suggest making this part of the same if-statement for visual clarity, otherwise looks good!

```suggestion
	if (!m_game->IsNormalSpace() || !path->IsSameSystem(m_starSystem->GetPath()))
```

Web-eWorks (approved):

src/Space.cpp
Gliese852: Fixed, force-pushed, because the changes are quite simple. Also renamed 'linear orbit' to 'radial orbit' so that others also understand what I mean.

PR #5340, opened 2022-02-10
[Loader] Add assimp version information and error string when loading a model fails

ecraven:


src/scenegraph/Loader.cpp
```diff
@@ -340,8 +342,17 @@ namespace SceneGraph {
...
+		if (!scene) {
+			std::ostringstream version;
+			version << std::to_string(aiGetVersionMajor()) << "." << std::to_string(aiGetVersionMinor());
+			// Assimp 3.1.1 doesn't have this, add it back in at some point:
+			// version << "." << std::to_string(aiGetVersionPatch());
+			std::string str1 = "Assimp (";
+			str1.append(version.str());
+			str1.append(") importer error: ");
+			str1.append(importer.GetErrorString());
+			throw LoadingError(str1.c_str());
```
Web-eWorks, line 354: I'd recommend directly formatting the string like so:
```c++
std::string err = fmt::format("Assimp {}.{} importer error: {}\n",
    aiGetVersionMajor(), aiGetVersionMinor(), importer.GetErrorString());
```

This assumes that LoadingError makes a copy of the string data when created, or else the underlying string data will be freed when the exception is thrown.

Web-eWorks (changes requested):

Web-eWorks (approved):

PR #5345, opened 2022-02-16
Add riscv support

pkubaj:
1. Add `__riscv` to `USE_CHRONO` check, like other architectures.
2. Merge https://github.com/max0x7ba/atomic_queue/commit/4e559dabe28e57ee27cb45c8297e1e387beed1d3

### DIFF
--- a/contrib/atomic_queue/defs.h
+++ b/contrib/atomic_queue/defs.h
@@ -47,6 +47,13 @@ namespace atomic_queue {
 constexpr int CACHE_LINE_SIZE = 256; // TODO: Review that this is the correct value.
 static inline void spin_loop_pause() noexcept {} // TODO: Find the right instruction to use here, if any.
 } // namespace atomic_queue
+#elif defined(__riscv)
+namespace atomic_queue {
+constexpr int CACHE_LINE_SIZE = 64;
+static inline void spin_loop_pause() noexcept {
+    asm volatile ("nop");
+}
+} // namespace atomic_queue
 #else
 #warning "Unknown CPU architecture. Using L1 cache line size of 64 bytes and no spinloop pause instruction."
 namespace atomic_queue {
--- a/contrib/profiler/Profiler.h
+++ b/contrib/profiler/Profiler.h
@@ -8,7 +8,7 @@
 #define __PROFILER_FULL_TYPE_EXPANSION__
 
 //#define USE_CHRONO
-#if !defined(USE_CHRONO) && (defined(__arm__) || defined(__aarch64__) || defined(__PPC64__) || defined(_M_AMD64) || defined(_WIN64) || defined(_M_X64))
+#if !defined(USE_CHRONO) && (defined(__arm__) || defined(__aarch64__) || defined(__PPC64__) || defined(_M_AMD64) || defined(_WIN64) || defined(_M_X64) || defined(__riscv))
 // this isn't optional for __arm__ or x64 builds
 #define USE_CHRONO
 #endif


impaktor:
Does pioneer run on RISCV?

pcercuei:
I was really wondering why RISC-V was actually needed to be enabled. Pioneer should just be architecture-independent.

And ugggrrrhhhh... that atomic_queue. I would like to hear a very good reason about why this is used. Doing spinlocks in userspace is a terrible idea and this is no exception. It just doesn't work on regular non-realtime operating systems.

Chances are that it's more harmful to the performance than a lock-based queue would be.

Gliese852:
@pcercuei see #5302

Web-eWorks:
> And ugggrrrhhhh... that atomic_queue. I would like to hear a very good reason about why this is used. Doing spinlocks in userspace is a terrible idea and this is no exception. It just doesn't work on regular non-realtime operating systems.

Because it is more important that the thread *does not sleep* when there is possible work available for it than that it uses "nice" OS primitives. In game code contexts, spinning on a semaphore is still more efficient (broadly speaking) than a kernel-mode call and context switch in the case where work is likely to come in during the duration of the spin. A sleep/context switch results in the worker thread losing up to 2ms of the 16ms target frame time in my observed benchmarks.

You're more than welcome to look at the [library benchmarks](https://max0x7ba.github.io/atomic_queue/html/benchmarks.html) as well, although as a note there were some OS scheduling configuration concessions made in the interest of achieving reliable benchmark results.

Also in more practical terms, it's a very small library using portable code that performs *well enough* on both Windows (where spinlocks are significantly faster than mutexes) and Linux (where they're on par or slightly worse and other side-effects of spin-lock are more important) rather than depending on OS-specific performance characteristics and manually writing and maintaining a mutex-based queue.

Additionally, I'd direct you to [this article](https://probablydance.com/2019/12/30/measuring-mutexes-spinlocks-and-how-bad-the-linux-scheduler-really-is/) and its benchmark findings; for our usecase the *most important* aspect is that the lock is free most of the time, and most threads acquire the lock only to transfer ownership of a work item handle which they then spend 95% of their runtime working on and not asserting a lock on the spinlock. In other words, idle time is more important than longest-wait time, which is a situation in which spinlocks actually do perform on par (or in overall duration potentially up to 30% better than) OS mutexes on the Linux kernel scheduler.

pcercuei:
@Web-eWorks the article you linked has been debunked many times. Please read the whole thread and in particular Linus Torvalds' answer to the blog post: https://www.realworldtech.com/forum/?threadid=189711&curpostid=189723

TL/DR: userspace spinlock is only useful in realtime systems where you can stop the scheduler from scheduling you in the middle of the critical section. For general-purpose OSes it's a shit fest.

Web-eWorks:
A subject matter "expert" (ignoring the disclaimer at the end that everyone gets the subject at hand wrong multiple times) writing a blog post asserting their *opinion* that the test case is flawed or incorrect does nothing to invalidate the results without any measured data to back that up. There's a reason actual scientific debunking involves a separate study with measured results to prove the first study is invalid.

While I have read the thread before and do happen to agree with Torvalds that spinlocks definitely have the same problem as every other synchronization primitive with the owning thread being preempted while holding the lock, I've yet to see any data from a supposed "correct" test case that shows the impact (and likelihood) of the thread being scheduled out, especially in a scenario where the duration of the lock is on the order of 10s of cycles, as it is in our usecase.

And finally, none of this matters at all, because we use the atomic_queue to pass around pointers to job records, which means the "critical section" is *a single atomic compare-exchange*. If I'm remembering the details of x86 and ARM atomics correctly, that means it is architecturally impossible for a worker thread to be pre-empted in the middle of a critical section inside of atomic_queue, so the entire point is moot.

And that in particular is why I chose this library rather than try to hack together my own locking queue based on std::mutex.

Web-eWorks (approved):
Either way, we've gotten quite off topic here, with apologies to @pkubaj. I appreciate you adding in the upstream patch to atomic_queue in this PR; adding additional special-casing to the profiler's USE_CHRONO check is fine, but I'll most likely follow up this PR with another one that removes the USE_CHRONO check entirely given that 32-bit x86 is such a sufficiently small marketshare it's no longer worth the complexity of having a specific asm carve-out for it.

pcercuei:
> A subject matter "expert" (ignoring the disclaimer at the end that everyone gets the subject at hand wrong multiple times) writing a blog post asserting their _opinion_ that the test case is flawed or incorrect does nothing to invalidate the results without any measured data to back that up. There's a reason actual scientific debunking involves a separate study with measured results to prove the first study is invalid.

That's not true. To debunk a scientific study it's enough to prove that the process was flawed. And he has pretty solid arguments against the way the benchmark was done.

> While I have read the thread before and do happen to agree with Torvalds that spinlocks definitely have the same problem as every other synchronization primitive with the owning thread being preempted while holding the lock,

No, and that's the whole point. With futexes the scheduler knows what thread is blocking the lock, and what thread is waiting on the lock. With spinlocks, the scheduler has no clue what's going on.

> I've yet to see any data from a supposed "correct" test case that shows the impact (and likelihood) of the thread being scheduled out, especially in a scenario where the duration of the lock is on the order of 10s of cycles, as it is in our usecase.
 
> And finally, none of this matters at all, because we use the atomic_queue to pass around pointers to job records, which means the "critical section" is _a single atomic compare-exchange_. If I'm remembering the details of x86 and ARM atomics correctly, that means it is architecturally impossible for a worker thread to be pre-empted in the middle of a critical section inside of atomic_queue, so the entire point is moot.

Yes, atomics are fine.

> And that in particular is why I chose this library rather than try to hack together my own locking queue based on std::mutex.

That's a perfectly sound reasoning. An atomic queue based on atomic primitives sounds fantastic, I just question their use of spinlocks for the sake of being fully "lockless". The funniest part is that when a given atomic primitive is not available on a given CPU, GCC will use regular locks anyway...

Anyway, rant over. I'm done hijacking the topic, sorry about that.

PR #5370, opened 2022-05-07
Sinonatrix cockpit & visible thruster plumes

nozmajner:
I've finally finished the cockpit for the Sinonatrix:
sinonatrix_cockpit_018
sinonatrix_cockpit_019
screenshot-20220507-173304
screenshot-20220507-173307
screenshot-20220507-173224

[Sketchfab model](https://sketchfab.com/3d-models/sinonatrix-cockpit-3d8e9c08a07042c4964e9e826f407d92)

The theme is retro computing, with lots of plastic parts and old periferials. Be sure to check out the keyboard layout! :)
Not much visible, but I placed a screen-window downwards where the pedals are. Not the nicest panorama, but might help a tiny bit with landing. Figured it would be better than pitch-blackness.
(Also I noticed that seeing the ships hull provides a bit of helpful reference during landing)

It is a bit heavy with detail with 124k tris, but it shouldn't be too much of a problem, since there's only one visible at any time.
It has a 4k diff and normals texture, half-sized spec and quarter-sized glow. 1k and halfk for the screen material. All textures except the normals are in DDS. I did not generate mipmaps, since you will only see it at this scale.

I've placed it in the Sinonatrix model directory to be able to use its texture for the hull. But I'd rather put it in the models/cockpits I created with the default cockpit later on, but that would require the ability to reference images/models from another directory.

And I noticed that it included my gitignore commit here as well, even though I wasn't on this branch when I did that. Sorry about that.

Edit:
Thanks to @Gliese852 now it is possible to include thruster tags in the cockpit model. [Behold](https://user-images.githubusercontent.com/4182678/167301341-ff6003dd-dcd4-4622-bf24-a4347a6ce93d.mp4):
image
This is included in the PR


nozmajner:
Not sure what's that latest commit, hopefully it is straightforward to squash the whole thing upon merge.
The one before it fixes a tiny mesh error:
image

PR #5371, opened 2022-05-14
Improve ship control

### LINKED ISSUE #5379: Crash when saving a game if a new game was relaunched.
### Steps to reproduce

- Start a new game.
- Exit to menu.
- Start a new game again.
- Save the game.

### Backtrace:
```
Thread 1 "pioneer" received signal SIGSEGV, Segmentation fault.
0x00005555559d8c3d in PropertyMap::GetRef (this=this@entry=0x555557137fe8, hash=1002329533) at ...pioneer/src/core/Property.cpp:185
185			uint32_t probed_key = m_keys[idx];

#0  0x00005555559d8c3d in PropertyMap::GetRef (this=this@entry=0x555557137fe8, hash=1002329533) at .../pioneer/src/core/Property.cpp:185
#1  0x00005555558a2dd9 in PropertyMap::Get (str=..., this=0x555557137fe8) at .../pioneer/src/core/Property.h:221
#2  l_index (l=0x555555f13510) at .../pioneer/src/lua/LuaMetaType.cpp:116
#3  0x00005555558f3b36 in luaD_precall (L=L@entry=0x555555f13510, func=<optimized out>, nresults=1) at .../pioneer/contrib/lua/src/ldo.c:318
#4  0x00005555558f3ed8 in luaD_call (L=L@entry=0x555555f13510, func=<optimized out>, nResults=nResults@entry=1, allowyield=<optimized out>) at .../pioneer/contrib/lua/src/ldo.c:394
#5  0x0000555555900d3a in callTM (L=L@entry=0x555555f13510, p1=p1@entry=0x5555682816b0, p2=p2@entry=0x55555695e520, p3=p3@entry=0x5555682816d0, hasres=1, f=<optimized out>, f=<optimized out>) at .../pioneer/contrib/lua/src/lvm.c:102
#6  0x000055555590104c in luaV_gettable (L=L@entry=0x555555f13510, t=0x5555682816b0, key=0x55555695e520, val=val@entry=0x5555682816d0) at .../pioneer/contrib/lua/src/lvm.c:127
#7  0x0000555555902700 in luaV_execute (L=L@entry=0x555555f13510) at .../pioneer/contrib/lua/src/lvm.c:588
#8  0x00005555558f3ebe in luaD_call (L=0x555555f13510, func=<optimized out>, nResults=<optimized out>, allowyield=<optimized out>) at .../pioneer/contrib/lua/src/ldo.c:395
#9  0x00005555558f340f in luaD_rawrunprotected (L=L@entry=0x555555f13510, f=f@entry=0x5555558eada0 <f_call>, ud=ud@entry=0x7fffffffd440) at .../pioneer/contrib/lua/src/ldo.c:131
#10 0x00005555558f41af in luaD_pcall (L=L@entry=0x555555f13510, func=func@entry=0x5555558eada0 <f_call>, u=u@entry=0x7fffffffd440, old_top=368, ef=<optimized out>) at .../pioneer/contrib/lua/src/ldo.c:595
#11 0x00005555558ecf05 in lua_pcallk (L=L@entry=0x555555f13510, nargs=nargs@entry=0, nresults=nresults@entry=1, errfunc=errfunc@entry=22, ctx=ctx@entry=0, k=k@entry=0x0) at .../pioneer/contrib/lua/src/lapi.c:949
#12 0x00005555558ea546 in pi_lua_protected_call (L=L@entry=0x555555f13510, nargs=nargs@entry=0, nresults=nresults@entry=1) at .../pioneer/src/lua/core/Sandbox.cpp:177
#13 0x00005555558c95d1 in LuaSerializer::ToJson (this=<optimized out>, jsonObj=...) at .../pioneer/src/lua/LuaSerializer.cpp:370
#14 0x0000555555798eba in Game::ToJson (this=this@entry=0x555563bcb7d0, jsonObj=...) at .../pioneer/src/Game.cpp:230
#15 0x0000555555799b58 in Game::SaveGame (filename="_quicksave", game=0x555563bcb7d0) at .../pioneer/src/Game.cpp:916
#16 0x0000555555616daf in Pi::HandleKeyDown (key=0x7fffffffdb50) at .../pioneer/src/Pi.cpp:796
#17 0x0000555555619b99 in sigc::pointer_functor1<SDL_Keysym*, void>::operator() (_A_a1=<optimized out>, this=<optimized out>) at /usr/include/sigc++-2.0/sigc++/functors/ptr_fun.h:111
#18 sigc::adaptor_functor<sigc::pointer_functor1<SDL_Keysym*, void> >::operator()<SDL_Keysym* const&> (_A_arg1=<optimized out>, this=<optimized out>) at /usr/include/sigc++-2.0/sigc++/adaptors/adaptor_trait.h:89
#19 sigc::internal::slot_call<sigc::pointer_functor1<SDL_Keysym*, void>, void, SDL_Keysym*>::call_it (rep=<optimized out>, a_#0=<optimized out>) at /usr/include/sigc++-2.0/sigc++/functors/slot.h:429
#20 0x00005555555dc549 in sigc::internal::signal_emit1<void, SDL_Keysym*, sigc::nil>::emit (impl=0x555556209df0, _A_a1=@0x7fffffffda50: 0x7fffffffdb50) at /usr/include/sigc++-2.0/sigc++/signal.h:1045
#21 0x00005555555d97b4 in sigc::signal1<void, SDL_Keysym*, sigc::nil>::emit (_A_a1=@0x7fffffffda50: 0x7fffffffdb50, this=0x55555616a780) at /usr/include/sigc++-2.0/sigc++/signal.h:2955
#22 Input::Manager::HandleSDLEvent (this=0x55555616a780, event=...) at .../pioneer/src/Input.cpp:600
#23 0x00005555555c38c2 in GuiApplication::PollEvents (this=0x555555e71af0) at .../pioneer/src/core/GuiApplication.cpp:149
#24 0x0000555555616105 in GameLoop::Update (this=0x555555e71dd0, deltaTime=0.0781602859) at .../pioneer/src/Pi.cpp:996
#25 0x00005555555c263a in Application::Run (this=0x555555e71af0) at .../pioneer/src/core/Application.cpp:191
#26 0x00005555555aaf75 in main (argc=1, argv=0x7fffffffe028) at .../pioneer/src/main.cpp:186
```
Lua stacktrace (?)
```
(gdb) call pi_lua_stacktrace(l)
[C](-1): __index
[string "[T] @libs/Ship.lua"](904): ?
```

Bug wound up after May 9.

9aa6e1563b136600382abc966c4be33f14778d14

Gliese852:
https://user-images.githubusercontent.com/18342621/168445666-df6de4c5-ba58-4a3d-a531-29681b64516b.mp4

Turned out quite voluminous, although the original goal was quite simple - the ship should automatically adjust its position so as to remain stationary, or move in a straight line inside a rotating orbital, as a preparatory stage for a comfortable landing in manual (assisted) mode. Some preparations have also been made related to the input system, improving the behavior of radial menus, some calculations in propulsion. In addition, there are several commits that are almost not related to the main topic, but somehow got in the way. The main commit related to the upgrade of the ship's control system is the most recent one. (UPD:  order messed up after force-push, I mean "Add cruise direction, follow modes, speed limiter") Unfortunately, it was not possible to divide it into components, everything is quite strongly connected.

Overview:

- rename `SetSpeed` to `Cruise`, `SetSpeedTarget` to `FollowTarget`

- split cruise mode in to 2 directions - `CRUISE_FWD` and `CRUISE_UP`. Differ in which axis the flight speed is fixed. (CRUISE_FWD(-z) and CRUISE_UP(+y) respectively)

- cruise directions automatically switch when lowering / raising the chassis.

- when you press W/S, R/F, respectively, cruise speed is set to the value with which the ship is actually moving, depending on the cruise direction ( `CRUISE_FWD` / `CRUISE_UP` ) - comfortable to fly at low speeds

- introduce a new type of flight assistance - "follow". There are 2 modes: `FOLLOW_POS` when the ship maintains a position relative to the target, in the coordinate system of the main frame.  and `FOLLOW_ORI` in which the ship maintains the relative position and orientation in the followed object's coordinate system. `FOLLOW_ORI`, enabled in manual mode, just copies the rotation of the followed body

- add speed limiter - a device that does not allow cruise speed to be set above some limit. In cruise mode it limits the setting of the cruise speed, in manual mode and fixheading modes it limits the actual speed of the ship

UI:

- add speed limiter widget, remove 'flight control' button
- add 'flight assist' button, also moved functionality from 'flight control'  button to it
- move the radial menu of fixheading modes to the flight assist button
- radial menu can now be used from keyboard or joystick

A more detailed description is available in the commits notes.

Also fixes #5379


nozmajner:
We talked about this constantly, and I also tested it from time to time. It will improve flight quite a lot, and has lots of quality of life features. @Gliese852 did a great work with this!

data/pigui/modules/autopilot-window.lua
```diff
@@ -58,101 +70,187 @@ local function button_undock()
...
-				button_undock()
-				ui.sameLine()
-				button_flight_control()
+				ui.setCursorPos(ui.getCursorPos() + Vector2(0, shift))
+				if button_hyperspace() then ui.sameLine() end
+				if button_undock() then ui.sameLine() end
+				speed_limiter.show()
 			end -- current_view == "world"
 		end)
+	dbg.drawLog(ui)
```
Web-eWorks, line 250: Also left over from debugging and causes errors.

data/lang/input-core/en.json
```diff
@@ -51,6 +51,14 @@
     "description": "Descriptive name for the DecreaseTimeAcceleration action.",
     "message": "Decrease Time Acceleration"
   },
+  "BIND_FIXHEADING_RADIAL": {
+    "description": "Descriptive name for the action binding.",
+    "message": "Activate fixheading radial menu"
```
Web-eWorks, line 56:
```suggestion
    "message": "Open Attitude Hold Radial Menu"
```

data/lang/input-core/en.json
```diff
@@ -51,6 +51,14 @@
...
+  },
+  "BIND_FLIGHT_ASSIST_RADIAL": {
+    "description": "Descriptive name for the action binding.",
+    "message": "Activate flight assist radial menu"
```
Web-eWorks, line 60:
```suggestion
    "message": "Open Flight Assist Radial Menu"
```

data/lang/input-core/en.json
```diff
@@ -111,6 +119,14 @@
     "description": "Descriptive name for the PrimaryFire action.",
     "message": "Primary Fire"
   },
+  "BIND_RADIAL_HORIZONTAL_SELECTION": {
+    "description": "Input axis name to select in the radial menu.",
+    "message": "Selection in radial menu horizontally"
```
Web-eWorks, line 124:
```suggestion
    "message": "Radial Menu Left/Right"
```

data/lang/input-core/en.json
```diff
@@ -111,6 +119,14 @@
...
+  },
+  "BIND_RADIAL_VERTICAL_SELECTION": {
+    "description": "Input axis name to select in the radial menu.",
+    "message": "Selection in radial menu vertically"
```
Web-eWorks, line 128:
```suggestion
    "message": "Radial Menu Up/Down"
```

Web-eWorks (changes requested):
Overall, I'm *very* happy with the contents of this PR; the new flight controls are very welcome and the UI is *extremely* slick and polished. I love what you've done with the UI animations, although there are some new bugs arising from all of the UI changes.

Hotplugging controllers works beautifully by the way, it saw my controller immediately after I connected it with Bluetooth.

This batch of review comments only deals with the first six files in the PR - I've suggested changes to some translations to better match existing strings, and done a fairly detailed review on some of the library code. I will be continuing to review the PR as I have time - between the comments and bugs I've found there should be plenty here to work on.

Now for the bugs (all experienced from multiple Mars starts, after removing two references to the `mydebug` module so the game runs):
- I can no longer select bodies in the world by clicking on their labels/icons, nor summon a radial menu by right-clicking in the world on a body label. Selection of bodies through the overview window is the only working option at current.
- Summoning a radial menu by right-clicking a body in the overview window works, but attempting to execute any of the actions in the radial menu fails silently. For example, I cannot request docking permission from Cydonia and can no longer redock to the station. The flight assist and attitude hold radial menus work as expected however.
- The speed limiter is a bit unclear as to its function, especially when using forward-cruise flight assist. I'd recommend adding a tooltip "Set maximum speed limit" to the slider drag area itself to make it more clear that it's not changing the cruise speed but limiting the maximum speed of the vessel (and also making it clear that it's an actual slider and not a text entry field).
- Related to above, the speed limiter slider does not adequately communicate it's an actual slider. I'd recommend adding a left and right arrow to the outer edges of the slider (`< 0.12km/s >`, either in the slider drag area or just outside of it) so the user is more able to see that it can be dragged.
- If you hold W, drag the speed limiter slider, release W while you're dragging, and then release the slider, the game behaves as if W is still held until you next press and release W. I believe this has something to do with ImGui's `WantCaptureKeyboard` flag being set even though mouse-dragging the slider is not a text input action. This is somewhat a problem, because it's not terribly uncommon to adjust the speed limiter and hold W to set your cruise speed up to the speed limiter.
- Radial Menu Left/Right axis binding has an inverted effect - pushing the axis left (or pressing the Positive keybind) selects the right-hand side of the radial menu; this is backwards to both existing axis conventions and the physical axis directions reported by my test hardware (PS4 controller).
- The Radial Menu Left/Right and Radial Menu Up/Down keys should *probably* be bound to the arrow keys by default, but this is very low priority. I'd like them to be on WASD for maximum user-friendliness, but that will require a further refactor of the InputFrame paradigm.

In closing, it feels to me that this PR really makes a phenomenal improvement to the control of spaceships and the user-friendliness of the UI; I'm very happy with what I'm seeing and I thank you for taking up the torch with the input binding UI and finishing the remaining pain points with the new control binding system!

data/libs/bind-manager.lua
```diff
@@ -0,0 +1,225 @@
...
+--   id - number, KeyBinding.joystick
+--
+-- Returns:
+--
+--   a string, "Joy1","Joy2" etc.
+--
+function bindManager.joyAcronym(id)
+	if id == 65535 then return "DISABLED" end
+	return "Joy" .. tostring(id)
+end
```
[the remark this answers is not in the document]
Web-eWorks: Indeed, after compiling and testing the game, I did not see a place where the `DISABLED` text is actually used, so this function can likely be removed.

data/pigui/modules/autopilot-window.lua
Gliese852: Thanks, removed!

data/pigui/modules/flight-ui/reticule.lua
```diff
@@ -33,6 +41,11 @@ local player = nil
 -- this should go into HUD settings
 local showNavigationalNumbers = true
 
+-- to interact with actions and axes in the input system
+local bindings = {}
+bindings.assistRadial = bindManager.registerAction('BindFlightAssistRadial')
+bindings.fixheadingRadial = bindManager.registerAction('BindFixheadingRadial')
```
Web-eWorks, line 47: Seems a little strange to create the table and then assign into it when you don't need to back-reference the table; I'd prefer this form:

```suggestion
local bindings = {
    assistRadial = bindManager.registerAction('BindFlightAssistRadial'),
    fixheadingRadial = bindManager.registerAction('BindFixheadingRadial')
}
```

data/pigui/modules/flight-ui/reticule.lua
```diff
@@ -420,23 +461,289 @@ local function displayManeuverData(radius)
...
+	local button_size = flightAssistButtonSize
+	local tooltip = ""
+	local back_color = colors.lightBlueBackground:opacity(0.3)
+	-- info:
+	local flightstate = player:GetFlightState()
+	local flightcontrolstate = player:GetFlightControlState()
+	local followtarget = player:GetFollowTarget()
+	local navtarget = player:GetNavTarget()
+	local cmbtarget = player:GetCombatTarget()
+	local frmtarget = player.frameBody
```
Web-eWorks, line 517: Is `player.frameBody` an existing bound variable from before this PR? If it's not, it feels a bit weird to have it here as an bound variable instead of a Get*() function.

data/pigui/modules/flight-ui/reticule.lua
```diff
@@ -420,23 +461,289 @@ local function displayManeuverData(radius)
...
+		main_icon = cruisedir == "CRUISE_FWD" and icons.cruise_fwd or icons.cruise_up
+		tooltip = cruisedir == "CRUISE_FWD" and lui.HUD_BUTTON_CRUISE_FORWARD or lui.HUD_BUTTON_CRUISE_UP
+	else
+		main_icon = flightstate_info[flightcontrolstate].icon
+		tooltip = flightstate_info[flightcontrolstate].tooltip
+	end
+
+	-- background
+	ui.addIcon(pos, icons.follow_fill, back_color, Vector2(button_size, button_size), ui.anchor.center, ui.anchor.center, tooltip)
+	ui.addIcon(pos, icons.follow_edge, back_color, Vector2(button_size, button_size), ui.anchor.center, ui.anchor.center, tooltip)
```
Web-eWorks, line 541: Is there a reason this is drawn using multiple icons composed on top of each other instead of using `ui.addCircle` + `ui.addCircleFilled`? (Or even just a single icon?) That would save two slots in the icon atlas, at the expense of generating some extra vertex data. It's not a required optimization by any means, just a bit weird seeing it here.

data/pigui/modules/flight-ui/reticule.lua
```diff
@@ -420,23 +461,289 @@ local function displayManeuverData(radius)
...
+
+		-- up/down icons
+		if flightcontrolstate == "CONTROL_MANUAL" or flightcontrolstate == "CONTROL_FIXSPEED" then
+			icon_up.icon = icons.circ_cruise_fwd
+			icon_up.tooltip = lui.HUD_BUTTON_CRUISE_FORWARD
+			icon_up.color = color_active
+			icon_up.action = function(_)
+				player:SetCruiseDirection("CRUISE_FWD")
+				player:SetFlightControlState("CONTROL_FIXSPEED")
+			end
```
Web-eWorks, line 573: Do all of these actions need to be generated inside the function? This seems like a mostly-invariant table that could take a reference to the player through the unused argument and be defined in a top-level table. Handling the different permutations of the radial menu could be done the same way, just using references to the different radial items when generating the radial table. It's not a major performance concern to have this code here, but it does feel a bit messy.

data/pigui/modules/settings-window.lua
```diff
@@ -311,45 +273,57 @@ local function showVideoOptions()
...
 	local info = keyCaptureBind
 
-	ui.text(localize_binding_id(info.id))
+	ui.text(bindManager.localizeBindingId(info.id))
 	ui.text(lui.PRESS_A_KEY_OR_CONTROLLER_BUTTON)
 
 	if info.type == 'Action' then
 		local desc = keyCaptureNum == 1 and info.binding or info.binding2
-		ui.text(desc.enabled and get_chord_desc(desc) or '<None>')
+		ui.text(desc.enabled and bindManager.getChordDesc(desc) or '<None>')
```
Web-eWorks, line 285: The `<None>` text should probably be localized - that's my bad for forgetting to do it when I introduced the bindings window.

data/pigui/modules/settings-window.lua
```diff
@@ -490,10 +464,25 @@ local function showControlsOptions()
...
+	-- list all the joysticks
+	local joystick_count = Input.GetJoystickCount()
+	if joystick_count > 0 then
+		ui.separator()
+		ui.text(linput.JOYSTICKS .. ":")
+		for id = 0, joystick_count - 1 do
+			ui.text("  " .. bindManager.joyAcronym(id) .. ":")
+			ui.sameLine()
+			ui.withStyleColors({["Text"] = ui.theme.colors.grey}, function()
+				local status = Input.IsJoystickConnected(id) and ", " .. linput.CONNECTED or ", " .. linput.NOT_CONNECTED
```
Web-eWorks, line 476: Instead of having the string concatenation on this line, I'd move it to the next line as part of `.. status`.

data/pigui/views/game.lua
```diff
@@ -118,35 +90,30 @@ local function displayOnScreenObjects()
...
-						-- also set setspeed target on ctrl-click
-						player:SetSetSpeedTarget(mainBody)
-					elseif not player:GetSetSpeed() then
-						-- clear setspeed target if not in setspeed mode
-						player:SetSetSpeedTarget(nil)
+						-- also set follow target on ctrl-click
+						player:SetFollowTarget(mainBody)
+					elseif not player:GetCruiseSpeed() then
+						-- clear follow target if not in cruise mode XXX???
+						-- player:SetFollowTarget(nil)
```
Web-eWorks, line 116: I'd like to see this branch removed or the comment resolved before merge for cleanliness.

data/pigui/views/game.lua
```diff
@@ -168,11 +135,11 @@ local function displayOnScreenObjects()
...
-						-- also set setspeed target on ctrl-click
-						player:SetSetSpeedTarget(b)
-					elseif not player:GetSetSpeed() then
-						-- clear setspeed target if not in setspeed mode
-						player:SetSetSpeedTarget(nil)
+						-- also set follow-target target on ctrl-click
+						player:SetFollowTarget(b)
+					elseif not player:GetCruiseSpeed() then
+						-- clear follow target if not in cruise mode
+						-- player:SetFollowTarget(nil)
```
Web-eWorks, line 142: Same with this branch.

Web-eWorks (commented):
Another batch of review comments - I'm very happy with the overall cleanliness of the code! Thanks for working on this PR and I appreciate your dedication towards improving the Pioneer codebase.

data/libs/bind-manager.lua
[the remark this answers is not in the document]
Gliese852: @Web-eWorks the fact is that this function is used in two places - in the list of joysticks (in settings-window), and in `getBindingDesc`. Removing the function will cause some code duplication, and the convention of naming the joystick "Joy" will no longer be formalized in the code.
As I can see, the `Input::JoystickFromGUID` function can return -1, which will be converted to 65535 for `uint16_t`, so without the function,  `Joy65535` will be output.  This can happen if the binding string contains a GUID that does not have a description. Most likely, this can only happen as a result of manual editing of the config.

data/libs/bind-manager.lua
```diff
@@ -0,0 +1,225 @@
...
+
+	return str
+end
+
+--
+-- Function: bindManager.localizeBindingId
+--
+-- Convert an axis or action binding style ID to a translation resource
+-- identifier, and get by this identifier a localized string from the
+-- "input-core" module.
```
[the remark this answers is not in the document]
Gliese852: Done!

data/libs/bind-manager.lua
```diff
@@ -0,0 +1,225 @@
...
+--
+-- action_name - string, action binding ID
+--
+-- Returns:
+--
+--   table:
+--     action - a LuaInputAction binding (see GetBindingPages)
+--     tooltip - localized string
+--
+function bindManager.registerAction(id)
```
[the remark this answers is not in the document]
Gliese852: Done!

data/libs/bind-manager.lua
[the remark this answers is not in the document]
Web-eWorks: Alright, in this case I'd recommend amending the function to pull the format strings from the translation resource anyways. It's cleaner and allows the game to be more fully translated in languages where it makes sense (e.g. character-based languages like Chinese or Japanese).

src/WorldView.cpp
```diff
@@ -40,13 +41,23 @@ REGISTER_INPUT_BINDING(WorldView)
 	input->AddActionBinding("BindToggleHudMode", group, Action({ SDLK_TAB }));
 	input->AddActionBinding("BindIncreaseTimeAcceleration", group, Action({ SDLK_PAGEUP }));
 	input->AddActionBinding("BindDecreaseTimeAcceleration", group, Action({ SDLK_PAGEDOWN }));
+	// universal axes for selecting an item from a radial menu
+	input->AddAxisBinding("BindRadialHorizontalSelection", group, Axis({}, { SDLK_LEFT }, { SDLK_RIGHT }));
+	input->AddAxisBinding("BindRadialVerticalSelection", group, Axis({}, { SDLK_UP }, { SDLK_DOWN }));
```
Web-eWorks, line 46: These two axes eventually shouldn't be in WorldView as they're more generic user-interaction bindings, but this is the best solution for the moment so no action is required.

src/lua/LuaInput.cpp
```diff
@@ -321,6 +362,178 @@ static int l_input_set_joystick_enabled(lua_State *l)
...
+ *   nothing
+ *
+ */
+static int l_input_action_on_press(lua_State *l)
+{
+	auto action = LuaObject<LuaInputAction>::CheckFromLua(1)->getAction();
+	action->binding.m_queuedEvents |= 1;
+	action->binding.m_active = true;
+	return 0;
+}
```
Web-eWorks, line 394: I think this (and most of the remaining action/axis manipulation functions added in this PR) should be a member function on the action object itself; e.g. `binding.action:OnPress()` or `binding.action:IsActive()`. While I do like how easy it is to search for `Input.IsActionActive`, it's unwieldy to type and logically the effect of these functions is limited to the action/axis object itself so it is a good fit for a member function.

src/lua/LuaInput.cpp
```diff
@@ -321,6 +362,178 @@ static int l_input_set_joystick_enabled(lua_State *l)
...
+	auto action = LuaObject<LuaInputAction>::CheckFromLua(1)->getAction();
+	// the action itself becomes active after dispatching, so we will check the
+	// corresponding bindings
+	bool active = action->binding.IsActive() || action->binding2.IsActive();
+	LuaPush(l, active);
+	return 1;
+}
+
+/*
+ * Function: IsActionActivated
```
Web-eWorks, line 457: Instead of `IsActionActivated` vs `IsActionActive`, I'd recommend `IsActionJustActive`, mimicking the ImGui naming convention.

src/lua/LuaPiGui.cpp
```diff
@@ -1103,13 +1106,15 @@ static int l_pigui_get_axisbinding(lua_State *l)
...
-	for (const auto &js : joysticks) {
-		const auto &axes = js.second.axes;
+	for (unsigned id = 0; id < joysticks.size(); ++id) {
+		const auto &axes = joysticks[id].axes;
 		for (size_t axis_num = 0; axis_num < axes.size(); axis_num++) {
 			float val = axes[axis_num].value;
-			if (std::abs(val) > 0.25) {
+			// we will cut off the possible joystick noise, as well as the fact that
+			// the trigger in a relaxed position means -1.0
+			if (std::abs(val) > 0.25 && std::abs(val) < 0.9) {
```
Web-eWorks, line 1115: I really don't like this hard-coded hack to avoid triggers constantly binding. I'd much prefer if the input system had some tracking for "at-rest" axes and only evaluate ones that have changed their values during this frame. That being said, a change of that size is out of scope for this PR, so I'd appreciate it if you'd add a `TODO:` to this comment for later work.

src/lua/LuaPiGui.cpp
```diff
@@ -1156,49 +1168,102 @@ static InputBindings::KeyBinding get_joy_button()
...
+		}
 	}
 
-	if (!binding.Enabled()) {
+	// mouse
+	if (bind_count < MAX_BIND_KEYS) {
+		for (int i = 0; i < 5; i++) {
+			if (ImGui::GetIO().MouseDown[i]) {
+				// imgui mouse button numbering different from sdl
+				const uint8_t imgui_to_sdl[] = { 1, 3, 2, 4, 5 };
```
Web-eWorks, line 1214: Instead of an explicit array here, you should be able to directly use `i + 1` to get the SDL key binding index.

src/lua/LuaPlayer.cpp
```diff
@@ -686,6 +687,167 @@ static int l_player_is_hyperspace_active(lua_State *l)
...
+ *
+ *   string - 'CRUISE_UP' or 'CRUISE_FWD'
+ *
+ */
+static int l_player_get_cruise_direction(lua_State *l)
+{
+	Player *player = LuaObject<Player>::CheckFromLua(1);
+	auto mode = player->GetPlayerController()->GetCruiseDirection();
+	auto str = EnumStrings::GetString("CruiseDirection", mode);
+	assert(str && "Wrong get cruise direction");
```
Web-eWorks, line 703: Instead of using `assert` in these functions, please use `luaL_error()` instead - it's nicer error handling and it's one less program crash I have to remove when making Lua code hot-reloadable and fault tolerant.

src/ship/PlayerShipController.cpp
```diff
@@ -9,13 +9,138 @@
...
+// vector3d (0.0) means we want zero speed, for example, so I made optional
+// so that the system can say that it does not want to touch this parameter at all
+struct PlayerShipController::TotalDesiredAction {
+	std::optional<vector3d> velLinear = {};	 // in ship's frame
+	std::optional<vector3d> velAngular = {}; // in ship's local space
+	std::optional<vector3d> thrLinear = {};
+	std::optional<vector3d> thrAngular = {};
+	vector3d angPower = vector3d(1.0);
+	vector3d linPower = vector3d(1.0);
+};
```
Web-eWorks, line 34: Instead of using `std::optional` here which wastes an extra 7 bytes for each vector3d member, this struct can be collapsed into four vectors and two bools. std::optional is a good fit for function parameters and return types where you need to bundle both value and state into a single return type, but for general-purpose data members it is less useful; in this case I think it's probably a better fit to think of the velocity-vs-thrust distinction as two exclusive modes rather than two optional values.

```suggestion
struct PlayerShipController::TotalDesiredAction {
	vector3d desiredLinear = {};	 // in ship's frame
	vector3d desiredAngular = {}; // in ship's local space
	vector3d angPower = vector3d(1.0);
	vector3d linPower = vector3d(1.0);
	bool desireLinVel = false;
	bool desireAngVel = false;
};
```

src/ship/PlayerShipController.cpp
```diff
@@ -149,13 +289,224 @@ void PlayerShipController::PostLoadFixup(Space *space)
...
+void PlayerShipController::ApplyTotalAction(const TotalDesiredAction &params)
+{
+	// setting the desired speed has priority over the low-level thrust setting
+	if (params.velLinear) {
+		m_ship->AIMatchVel(*params.velLinear, params.linPower);
+	} else if (params.thrLinear) {
+		m_ship->GetPropulsion()->SetLinThrusterState(*params.thrLinear);
+	} else {
+		m_ship->ClearThrusterState();
+	}
```
Web-eWorks, line 484: If the TotalDesiredAction is rewritten as above, this block becomes:

```suggestion
	// setting the desired speed has priority over the low-level thrust setting
	if (params.desireLinVel) {
		m_ship->AIMatchVel(params.desiredLinear, params.linPower);
	} else {
		m_ship->GetPropulsion()->SetLinThrusterState(params.desiredLinear);
	}
```

src/ship/PlayerShipController.cpp
```diff
@@ -227,16 +579,11 @@ void PlayerShipController::StaticUpdate(const float timeStep)
...
 	}
-
-	//call autopilot AI, if active (also applies to set speed and heading lock modes)
-	OS::EnableFPE();
-	m_ship->AITimeStep(timeStep);
-	OS::DisableFPE();
 }
 
-void PlayerShipController::CheckControlsLock()
+bool PlayerShipController::IsControlsLocked()
```
Web-eWorks, line 584: Minor nitpick - the rules of English say that this function should be named `AreControlsLocked`, as Controls is a plural noun.

src/ship/PlayerShipController.h
```diff
@@ -5,6 +5,7 @@
 
 #include "Input.h"
 #include "ShipController.h"
+#include <optional>
```
Web-eWorks, line 8: And again, to reiterate my original comment from the first part of the review, I think std::optional is not an optimal fit for the class variable, as its state is handled nicely by the presence of `m_followTarget`.

Web-eWorks (changes requested):
Alright, I think that's everything here in this PR that I've taken at least a cursory look over. Plenty of feedback for you to look over, but I don't expect to need any major changes beyond those called out in the individual review comments. I'll do another round of testing this weekend and mark off the issues from my original review that are fixed then.

src/ship/PlayerShipController.h
```diff
@@ -76,39 +94,60 @@ class PlayerShipController : public ShipController {
...
-	Body *m_setSpeedTarget;
-	bool m_controlsLocked;
+	Body *m_followTarget;
 	bool m_invertMouse; // used for rear view, *not* for invert Y-axis option (which is Pi::input->IsMouseYInvert)
 	bool m_mouseActive;
 	bool m_disableMouseFacing;
 	bool m_rotationDamping;
+	bool m_stickySpeedKey = false; // helps cruise speed sticks to 0 when it crosses it
+	matrix3x3d m_followTargetPrevOrient;
+	std::optional<vector3d> m_followTargetPrevVel;
```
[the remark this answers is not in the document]
Gliese852: @Web-eWorks the fact is that the vector {0,0,0} means zero speed and i need to show no speed data. I.e. the usual problem with creating a nullable from a non-nullable, which seems to be the vector.
I need to calculate the actual acceleration, for this I need to know 2 speeds, in the previous and in the current frame.

data/pigui/modules/flight-ui/reticule.lua
Gliese852: image
The fact is that in the follow mode even 3 icons are superimposed there, and I would not want to adjust the diameters and thicknesses of the circles, all this is provided in inkscape.

data/pigui/modules/flight-ui/reticule.lua
Gliese852: Judging by the code, this attribute is 11 years old.

data/pigui/modules/flight-ui/reticule.lua
Gliese852: Do I understand correctly, you propose that this function of 2 lines be taken out somewhere outside, and wrapped in some name?
And in this case, when the function is one line?
```
fix_icon.action = function(_) player:SetFlightControlState("CONTROL_FIXSPEED") end
```

Web-eWorks: I'm mentioning more that the entire `icon_up` table be defined once at file scope and referenced in the function to build the table of buttons for the radial menu. As far as I can tell, there are a total of six radial menu buttons in use here, none of which actually have internals that need to be generated at runtime. Only the selection of which buttons to display changes between invocations of this radial menu.

data/pigui/modules/flight-ui/reticule.lua
Web-eWorks: Ah, I just haven't seen it in a while. No action required then! :smile:

data/libs/bind-manager.lua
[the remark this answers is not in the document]
Gliese852: Done!

data/pigui/libs/radial-menu.lua
```diff
@@ -2,26 +2,29 @@
...
 local radialMenuSize = 10
 local radialMenuIconSize = Vector2(radialMenuSize)
 local radialMenuTarget = nil
 local radialMenuMouseButton = 1
 local radialMenuActions = {}
-local radialMenuMousePos = nil
+local radialMenuActionBinding = nil
+local radialMenuPadding = 0
+local defaultRadialMenuIconSize = ui.theme.styles.MainButtonSize.x
+local defaultRadialMenuPadding = ui.theme.styles.MainButtonPadding
```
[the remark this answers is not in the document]
Gliese852: Done!

src/ship/PlayerShipController.h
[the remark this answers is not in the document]
Gliese852: In the end, it became clear to me, now it's better, thank you!

data/pigui/libs/text.lua
```diff
@@ -356,16 +359,25 @@ ui.addStyledText = function(position, anchor_horizontal, anchor_vertical, text,
...
+		if anchor_vertical == ui.anchor.baseline then
+			vert = nil
+		else
+			vert = anchor_vertical
+		end
+		position = ui.calcTextAlignment(position, size, anchor_horizontal, vert) -- ignore vertical if baseline
+		if anchor_vertical == ui.anchor.baseline then
+			position.y = position.y - font.offset
+		end
+
```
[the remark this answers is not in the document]
Gliese852: Put it into `ui.calcTextAlignment`.

data/pigui/modules/flight-ui/reticule.lua
Gliese852: Done!

data/pigui/modules/settings-window.lua
Gliese852: Done!

data/pigui/modules/settings-window.lua
Gliese852: Done!

data/pigui/views/game.lua
Gliese852: Done!

data/pigui/views/game.lua
Gliese852: Done!

data/lang/core/en.json
```diff
@@ -871,6 +871,10 @@
     "description": "",
     "message": "Nerve Gas"
   },
+  "NONE": {
+    "description": "To indicate the absence of something",
+    "message": "<None>"
+  },
```
Web-eWorks, line 877: Because this has the < brackets > around it, I think this would be a better fit for the `input-core` or even `ui-core` resource - or even apply the brackets in Lua to allow this string to be reused in other contexts where the brackets are not appropriate.

src/Input.cpp
```diff
@@ -119,48 +168,43 @@ void Input::InitJoysticks(IniConfig *config)
 	Output("Initializing joystick subsystem.\n");
 
 	for (int n = 0; n < joy_count; n++) {
-		JoystickInfo state;
 
-		state.joystick = SDL_JoystickOpen(n);
-		if (!state.joystick) {
+		auto joystick = SDL_JoystickOpen(n);
```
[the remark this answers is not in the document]
Web-eWorks: And yes, I'm aware I'm the one who introduced the use of `auto` in this file... admittedly before we had guidelines for the use of `auto`, but that's not a good excuse.

Web-eWorks:
On my test screen resolution (1440x900), the left/right arrows for the speed limiter are drawn out-of-line with the text and the speed limiter area itself.
image

data/pigui/modules/flight-ui/reticule.lua
Gliese852: @Web-eWorks Ah, now I understand, thanks! What to do with dynamic tooltips?
https://github.com/Gliese852/pioneer/blob/838d73017c5852c1b2e9e13632955e746cd3b30c/data/pigui/modules/flight-ui/reticule.lua#L620

src/Input.cpp
```diff
@@ -79,15 +81,22 @@ int Input::JoystickFromGUIDString(const char *guid)
...
 		}
 	}
 	return -1;
 }
 
+int Input::JoystickFromID(SDL_JoystickID id)
+{
+	assert(id >= 0 && "this ID means no joystick");
+	assert(static_cast<unsigned>(id) < m_joyIDs.size() && "there is no internal ID for this SDL ID");
+	return m_joyIDs[id];
```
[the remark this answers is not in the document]
Gliese852: @Web-eWorks
The fact is that values that do not satisfy the assertion are invalid states, a normally working program simply should not call this function with invalid parameters. Also I would like to note that the result of this function in 6 places is directly used to index the array, you suggest to write check in each of these places?
```
Input.cpp:
629: if (!GetJoysticks()[Input::JoystickFromID(event.jaxis.which)].joystick)
631: auto &axis = GetJoysticks()[Input::JoystickFromID(event.jaxis.which)].axes[event.jaxis.axis];
640: if (!GetJoysticks()[Input::JoystickFromID(event.jaxis.which)].joystick)
642: GetJoysticks()[Input::JoystickFromID(event.jbutton.which)].buttons[event.jbutton.button] = event.jbutton.state != 0;
645: if (!GetJoysticks()[Input::JoystickFromID(event.jaxis.which)].joystick)
647: GetJoysticks()[Input::JoystickFromID(event.jhat.which)].hats[event.jhat.hat] = event.jhat.value;
```

src/lua/LuaPiGui.cpp
Gliese852: @Web-eWorks There is also a problem with "value jitter"

src/lua/LuaPiGui.cpp
Gliese852: @Web-eWorks note that the numbers in the array are not exactly in order. `i + 1` will confuse the middle and right mouse buttons.

src/ship/PlayerShipController.cpp
Gliese852: @Web-eWorks when to call `ClearThrusterState`? this is not exactly the same as `SetLinThrusterState({})`.

data/pigui/modules/flight-ui/reticule.lua
Web-eWorks: Hmm... I overlooked that you were dynamically generating tooltips for two of the radial buttons. You *could* just change the tooltip field of the static table right before showing the radial menu, although that's not a particularly clean solution.

Probably the best solution would be to still move the invariant radial button tables to file scope and only generate the dynamic tooltip buttons in the function itself; that would still reduce runtime complexity but maintain flexibility. Alternatively you could avoid dynamic tooltips entirely, as the format of `Mars: Not Available` doesn't actually tell the player anything about what the button is supposed to do.

src/Input.cpp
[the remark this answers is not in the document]
Web-eWorks: Exactly half of the places that call this function are calling it redundantly to retrieve a value that hasn't changed since the first time it was called, so that can be cleaned up eventually. :smile:

As this is public API that has no guarantee whether the caller intends an invalid joystick ID to cause a program crash, I'd *vastly* prefer that it return an invalid value. The existing callsites do not need to check the return value as they have the external precondition that the joystick IDs are valid, as they're coming from SDL for joysticks that the program has explicitly opened and added to the array.

On the other hand, if this is made a private function and not public API I don't really see the value of having any checks at all here if it's not going to be called from outside the Input class, as we're able to completely enforce the precondtion of a valid SDL_JoystickID externally to the function and don't need the overhead of an assert.

data/pigui/modules/flight-ui/reticule.lua
Gliese852: @Web-eWorks I still think that specifying the object for the action is important, because the player can have many different types of targets (combat, navigation, autopilot, frame) and it would be nice to let him know which target he will follow. Therefore, I am for the first option - partially move it in a static table.
Perhaps I should also change the phrase "Not available" to "Follow orientation not available" and "Follow position not available" respectively.

src/ship/PlayerShipController.cpp
Web-eWorks: ClearThrusterState has the exact same effect as calling SetLinThrusterState({0, 0, 0}), it's just a comparative "fast path" due to not needing to clamp and scale the thrust values. Given that this is called ~once per frame, I don't see any issues with taking the slow path here, and it can be optimized inside of SetLinThrusterState.

Gliese852: @Web-eWorks not exactly - this is done so that the ship maintains thrust during takeoff for some time
https://github.com/pioneerspacesim/pioneer/blob/6ca9518f400619704c3fea49a4abb0d76b69c6a3/src/Ship.h#L91-L95

Web-eWorks: ...so *that's* where the launch thrust is handled. That really should be moved to an explicit routine somewhere instead of stuffed into `ClearThrusterState`, as this is one of the jankier hacks in the flight control system in my opinion. (For example, apply some forward thrust during the launch timeout - it should clear that forward thrust and continue launching upwards.)

src/Input.cpp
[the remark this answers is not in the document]
Web-eWorks: Having argued the point to death on IRC, I withdraw my recommendation as at current there's nothing using it where the assert would be particularly harmful.

src/lua/LuaPiGui.cpp
Web-eWorks: This is what I get for reviewing at 3am, sorry. Ignore this one!

data/lang/core/en.json
Gliese852: @Web-eWorks There is also `<unknown>` in that file, so I thought to put this string there too.

data/pigui/modules/flight-ui/reticule.lua
Gliese852: I moved something, I hope it got better. Only today I realized that this code is executed once, at the moment of clicking, so the difference in performance seems negligible.

src/Input.cpp
```diff
@@ -110,6 +119,46 @@ static void loadAxisConfig(const std::string &str, Input::JoystickInfo::Axis &ou
...
+			i, state.axes[i].deadzone, state.axes[i].curve, state.axes[i].zeroToOne);
+	}
+	return state;
+}
+
+void Input::AddJoystickID(SDL_JoystickID sdl_id, unsigned internal_id)
+{
+	assert(sdl_id < 1000 && "SDL IDs seem to be big enough to use vector as map");
+	assert(sdl_id >= 0 && "attempt to add absent ID");
+	if (m_joyIDs.size() <= static_cast<unsigned>(sdl_id)) m_joyIDs.resize(sdl_id + 1);
```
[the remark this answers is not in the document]
Gliese852: Done!

src/Input.cpp
```diff
@@ -110,6 +119,46 @@ static void loadAxisConfig(const std::string &str, Input::JoystickInfo::Axis &ou
...
+		}
+
+		loadAxisConfig(config->String(joystickName, axisName, ""), state.axes[i]);
+		Output("  - axis %ld: deadzone %.2f, curve: %.2f, half-axis mode: %b\n",
+			i, state.axes[i].deadzone, state.axes[i].curve, state.axes[i].zeroToOne);
+	}
+	return state;
+}
+
+void Input::AddJoystickID(SDL_JoystickID sdl_id, unsigned internal_id)
```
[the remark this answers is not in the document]
Gliese852: Done!

src/Input.cpp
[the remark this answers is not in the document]
Gliese852: Done!

src/lua/LuaPlayer.cpp
Gliese852: Done! By the way, these assertions helped me a lot earlier.

src/lua/LuaInput.cpp
Gliese852: Done! It's nice that this change allowed to remove the `require 'Input'` from several modules.

src/lua/LuaInput.cpp
Gliese852: Done! however, this has been moved to class member-function `IsActive` and `IsJustActive` due to the previous item.

src/ship/PlayerShipController.h
Gliese852: Removed!

Gliese852:
@Web-eWorks 
> On my test screen resolution (1440x900), the left/right arrows for the speed limiter are drawn out-of-line with the text and the speed limiter area itself.

Oh, sorry, it should be ok now!

src/ship/PlayerShipController.cpp
Gliese852: @Web-eWorks could you please describe in more detail how to implement this?

Gliese852: @Web-eWorks Oh, are you suggesting storing either thrust or velocity in the same variable? Does it bother you that these are different units of measurement?

data/pigui/modules/flight-ui/reticule.lua
Web-eWorks: Yes, I'm not entirely concerned about the performance cost of the function (although every bit helps with immediate-mode UI), but more about the source-code complexity of the function; less code being executed at runtime means it's easier to debug and understand what's going on in the function. I'm happy with the updated function, thank you!

src/lua/LuaPlayer.cpp
```diff
@@ -700,7 +700,8 @@ static int l_player_get_cruise_direction(lua_State *l)
 	Player *player = LuaObject<Player>::CheckFromLua(1);
 	auto mode = player->GetPlayerController()->GetCruiseDirection();
 	auto str = EnumStrings::GetString("CruiseDirection", mode);
-	assert(str && "Wrong get cruise direction");
+	if (!str)
+		luaL_error(l, "Wrong get cruise direction");
```
Web-eWorks, line 704: This message (in `GetCruiseDirection()`) may be acceptable to keep as an assert, as Lua context is unable to cause an error by supplying an invalid parameter.

src/lua/LuaPlayer.cpp
```diff
@@ -721,7 +722,8 @@ static int l_Player_set_cruise_direction(lua_State *l)
 {
 	Player *player = LuaObject<Player>::CheckFromLua(1);
 	auto value = EnumStrings::GetValue("CruiseDirection", LuaPull<const char *>(l, 2));
-	assert(value != -1 && "Wrong set cruise direction");
+	if (value == -1)
+		luaL_error(l, "Wrong set cruise direction");
```
Web-eWorks, line 726: As a minor nitpick because we do not (yet) automatically add context information to error messages, I'd appreciate if this message was formatted like so (pseudocode):
```suggestion
		luaL_error(l, "Player:SetCruiseDirection(): invalid cruise direction '%s' specified\n", mode);
```

data/lang/ui-core/en.json
```diff
@@ -951,10 +951,26 @@
...
     "message": "{targetname}: Follow orientation"
   },
+  "HUD_FOLLOW_ORIENTATION_NOT_AVAILABLE": {
+    "description": "Tooltip for follow mode 'orientation' button, when it's not available for the current object.",
+    "message": "{targetname}: Follow orientation is not available."
+  },
+  "HUD_FOLLOW_ORIENTATION_NOT_AVAILABLE_TOO_FAR": {
+    "description": "Tooltip for follow mode 'orientation' button, when it's not available for the current object, because it's too far.",
+    "message": "{targetname}: Follow orientation is not available - too far."
+  },
```
Web-eWorks, line 961: These strings are a little bit long for optimal tooltips; I'd almost recommend trying to plumb in support for custom tooltip functions to add multiline/colored tooltip strings (e.g. "{targetname}: Follow Orientation" \<newline\> "Unavailable - Too Far"). That's outside the scope of this PR for right now though, so just something to keep as an idea for future improvements.

src/ship/PlayerShipController.cpp
Web-eWorks: The different units of measurement are mutually exclusive as far as I can tell - linear and angular motion are individually either used in match-velocity or total-thrust mode but never as thrust and match-velocity at the same time. Reusing the same variable for both modes makes sense, as the actual representation of the value does not change even if the interpretation does.

data/lang/core/en.json
Web-eWorks: That works fine by me then.

src/ship/PlayerShipController.cpp
Gliese852: @Web-eWorks yes, it will work, although now you will have to either check or assume the "type" of the data at the current point of execution each time you access this fields.

Web-eWorks: As far as I was able to see looking through the relevant code, most functions using this datastructure are already implicitly assuming the entire datastructure is being used for one or the other, as the way the optionals are evaluated to set the thruster state precludes any use of the datastructure for both match-vel and total-thrust modes for the same component at the same time.

Gliese852: Done, it's still nice that we managed to get rid of optional!

src/ship/PlayerShipController.cpp
Gliese852: Done!

src/ship/PlayerShipController.cpp
```diff
@@ -516,11 +502,14 @@ void PlayerShipController::StaticUpdate(const float timeStep)
 			ApplyTotalAction(act);
 			break;
 		case CONTROL_MANUAL:
+			// maintain the thrust set by the station when undocking while the timeout lasts
+			// but only in manual mode
+			if (m_ship->m_launchLockTimeout > 0) return;
```
Web-eWorks, line 507: Just as a confirmation, have you tested that AI ships are still able to launch and fly a safe distance above the pad before rotating and hyperjumping?

Web-eWorks (approved):
Overall I'm extremely happy with the quality of this PR - I haven't gotten a chance to fully put it through its paces and fly a full mission or long save with it yet, but any bugs that arise in the long term should be easy to handle. Well done @Gliese852!

Gliese852:
@impaktor yes I plan to move all the fixes to their respective original commits and do a force-push

impaktor:
Any reason not to merge now?

Gliese852:
> Any reason not to merge now?

@impaktor  after playtesting there were some bugs, I think they should be fixed first.

src/Input.cpp
```diff
@@ -140,6 +140,7 @@ static JoystickInfo loadJoystick(SDL_Joystick *joystick, IniConfig *config)
 	for (size_t i = 0; i < state.axes.size(); i++) {
 		std::string axisName = "Axis" + std::to_string(i);
 		if (!config->HasEntry(joystickName, axisName)) {
+			state.axes[i].deadzone = 0.2;
```
Web-eWorks, line 143: I'd prefer if this deadzone is set to `0.05` or similar, a 20% deadzone is extremely aggressive for anything that's not a thumbstick and the OS usually has a pretty large deadzone for most hardware. A 5% deadzone will filter out issues with sticks being almost centered but still reporting a non-zero value, and be a sane default for users who do not know about the ability to change deadzones in the config file.

Web-eWorks (changes requested):

src/Input.cpp
Gliese852: Done! And ready to merge!

Web-eWorks (approved):
Well done @Gliese852! I'm not sure if this is your biggest PR to be merged yet in terms of code changes, but it's certainly the cleanest and most well-anticipated set of changes merged this year! Thanks for bearing with all of my review changes, and for being willing to tackle this work in the first place!

PR #5372, opened 2022-05-20
>20 MB of Memory Leaks

Jeinzi:
I ran Valgrind on the current version of Pioneer and spotted a few memory leaks. I tested by starting the game on Mars, launching the ship and quitting again.

Category|Before|After
-|-|-
definitely lost|9,115,591 bytes in 2,570 blocks|4,636 bytes in 495 blocks
indirectly lost|12,183,945 bytes in 38,957 blocks|58,829 bytes in 636 blocks
possibly lost|6,610,824 bytes in 604 blocks|96 bytes in 1 blocks
still reachable|371,281 bytes in 3,372 blocks|346,565 bytes in 3,368 blocks
suppressed|0 bytes in 0 blocks|0 bytes in 0 blocks

### DIFF
--- a/src/Player.cpp
+++ b/src/Player.cpp
@@ -79,7 +79,7 @@ void Player::SaveToJson(Json &jsonObj, Space *space)
 
 void Player::InitCockpit()
 {
-	m_cockpit.release();
+	delete m_cockpit.release();
 	if (!Pi::config->Int("EnableCockpit"))
 		return;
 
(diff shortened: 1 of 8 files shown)


impaktor:
Nice!

Web-eWorks (approved):
Good catches, thank you! 20 MB is not a horribly large amount of memory compared to the total working set of ~1GB, but it's good to close these leaks.

PR #5373, opened 2022-05-22
Xylophis cockpit

nozmajner:
This one is a two-seater, and right-side drive (I want to make all OPLI ships right-siders, SolFed being imperialist)
[Sketchfab model](https://sketchfab.com/3d-models/xylophis-cockpit-2662e4f4dc0142059f9424edeec7730c)
xylophis_cockpit-009
xylophis_cockpit-010
xylophis_cockpit-011
xylophis_cockpit-012
In_game:
screenshot-20220522-152033

It is a rather heavy mesh at 190k tris, but I'm hoping to reuse it partially for the larger OPLI ships.
Textures are one 4k diff and normals. Spec is 2k, glow is 256*. Screens are 1k, with proportionally smaller other channels, and no normals. All textures except the normals are DXT5 dds-es.
I've put it to models/cockpits/ since it does not use the ship mesh at all.


Web-eWorks:
My only concern with the cockpit mesh is that the pillar on the right side takes up a non-negligible amount of screen space - I think if both seats were moved towards the center of the cockpit by just a little bit it would result in a much more enjoyable experience.

nozmajner:
This is as much as I can move the seat, no more room at the right of the seat:
screenshot-20220522-213036
Also I've updated the camera tag in the ship itself for the right-seater arrangement

PR #5378, opened 2022-05-27
Fix docked music not playing

### LINKED ISSUE #5362: Wrong music plays when loading saved game
### Observed behaviour
When loading a saved game in which the player is docked, the game plays music from the "near-spacestation" or "near-planet" folder (or "space" if none exists in those folders) instead of from the "docked" folder. This is not the case when starting a new game, where it does in fact play from the "docked" folder.

### Expected behaviour
The game should check if the player is docked upon loading a saved game, then play music from the "docked" folder.

### Steps to reproduce
Place unique music files into "near-spacestation" and/or "near-planet" as well as "docked" folders. Save a game while docked, then exit the game and load the saved game.

Windows 10 customized build, but nothing which would have caused this (some things trying to mitigate it unsuccessfully...).

WKFO:
Fixes #5362  hopefully.
Ping @Keeper1st .

### DIFF
--- a/data/modules/MusicPlayer.lua
+++ b/data/modules/MusicPlayer.lua
@@ -130,14 +130,24 @@ end
 
 Event.Register("onGameStart", function ()
 	MusicPlayer.rebuildSongList()
-	playAmbient()
+	if Game.player:GetDockedWith() and music["docked"] then
+		MusicPlayer.playRandomSongFromCategory("docked")
+	else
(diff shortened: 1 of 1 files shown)


Keeper1st:
Yep, it works! Now when loading a saved game where you're docked, the docked music plays instead of the near-planet or near-spacestation music.

PR #5380, opened 2022-06-12
Add Chrome-Tracing output support to profiler

Web-eWorks:
In order to support better profiling visualization, I've added support for outputting chrome://tracing format event traces from the profiler - this is a significantly more wasteful format than speedscope for profiling complex applications with many event zones, but it's a format that *many* visualization tools speak or can import to their own format. Chrome-tracing JSON output can be enabled with the `ProfilerTraceOutput=1` option in the config file, and is mutually exclusive with the regular Speedscope JSON output.

I've also removed a few more unnecessary / wasteful profile points which significantly bloat profiling for little gain - the methods they're profiling are very small and optimized, and potentially called hundreds of thousands of times in some traces.

### DIFF
--- a/contrib/profiler/Profiler.cpp
+++ b/contrib/profiler/Profiler.cpp
@@ -1005,6 +1005,23 @@ namespace Profiler {
 
 	/*
 		Thread Dumping
+
+		Each profile data format is handled by a separate dumper struct
+		implementation.
+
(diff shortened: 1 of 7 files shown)


impaktor:
I assume we should put some text about it here: https://pioneerwiki.com/wiki/Development_Tools#Profiling

PR #5381, opened 2022-06-12
Improve SVG rasterization and Starfield::Fill performance

Web-eWorks:
I'm starting to take advantage of the capabilities afforded by the new task graph system - I've rewritten the SVG image loading to run rasterization and image generation on a separate thread, and defer uploading the resulting image data to the main thread at the next processing point. This can result in a few frames where the texture the image is rasterized to contains only the clear-color, but I consider that to be a feature, not a bug. This removes a main-thread stall while loading Lua scripts at program startup, and will reduce or remove stalls if future code needs to load (or reload) SVG files at runtime.

I've also rewritten the way filling background stars works when starting a new game - this involves a very slight change to the algorithm to sum the thread-local median of stars rather than the global median of stars, but allows up to a potential 8x speedup based on the number of threads available. This change has a very perceptible impact on the hitch observed when starting a new game.

### DIFF
--- a/src/Background.cpp
+++ b/src/Background.cpp
@@ -11,7 +11,10 @@
 #include "Player.h"
 #include "Space.h"
 #include "StringF.h"
+
 #include "galaxy/StarSystem.h"
+#include "galaxy/GalaxyGenerator.h"
+
 #include "graphics/Graphics.h"
 #include "graphics/RenderState.h"
 #include "graphics/TextureBuilder.h"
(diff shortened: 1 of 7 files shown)


impaktor:
Could the task graph system also be used to speed up how soon we can use sound when starting pioneer?

Web-eWorks:
Yes, in fact we already do that to some extent using the job system - Pioneer currently loads music and sound files on two separate threads as part of a load step, which is something I could refactor to use the task system for. However, loading sound files is something that's mostly disk-bound rather than compute-bound, so I do not expect further refactors in that area to provide a major improvement.

The sound-loading step is pretty independent of the rest of the loading system, so it could be fairly easily moved earlier during the load process.

Issue #4803, opened 2020-02-22
New game may give you zero equipment


pcercuei:
### Observed behaviour

It's possible to start a new game at Barnard's Star (and probably the others as well) without any equipment given to the player.

### Expected behaviour

The player should start with the atmospheric shielding, the radar, and the autopilot.

### Steps to reproduce

1. Create a new game at Barnard's Star.
2. Sell your equipment.
3. Save the game.
4. Return to the main menu.
5. Load the game.
6. Return to the main menu.
7. Create a new game at Barnard's Star. Now you don't have any equipment.

__My pioneer version (and OS):__
pioneer master (e25517163) on Debian x64.

craigo-:
The symptoms and steps to reproduce sound eerily similar to #4675

WKFO:
Happened to me recently.

- I sell some of my equipment in-game.

- I close the game and return to main menu.

- I start a new game on Mars. My ship only has a hyperdrive.

Another-MrZ:
I had same experience, but slightly different.
If you start a new game afresh (difficult start), you get your Xylophis with all the equipment that appear in the shop list as sellable.
But if you chose different settings, then restart, or if you sell your equipment and restart, you still get all the needed equipment, but it does not show up in the shop list neither in the Ship equipment list.
I started a game this way: I still had autopilot, atmospheric shielding and radar, but they would not show up and could not be sold.

joonicks:
ship equipment isnt stored in the C++ domain, its stored in the Lua domain... isnt Lua being reset fully between games?

maxcrystal:
I have the same bug. New game starts without any equipment visible anywhere (except hyperdrive).

impaktor:
@maxcrystal thanks for reporting many bugs, if you find any way to reproduce this, or report if it's random (how often?), or always only hyperdrive, then please fill us in with any more info you find.

This bug was also reported on [SSC](https://spacesimcentral.com/community/pioneer/new-pioneer-release-2020-12-22/#post-59020) for the release candidate 1 (20201222-rc1)

Web-eWorks:
Closing as addressed by #5734 - please re-open if still valid in the upcoming release.

PR #5384, opened 2022-06-15
Add BodyComponent system

### LINKED ISSUE #5385: Missing Reticule
### Observed behaviour
After a few seconds/minutes in flight, the reticule disappears

Warning: [string "[T] @pigui/libs/text.lua"]:365: bad argument #1 to 'CalcTextSize' (string expected, got nil)
stack traceback:
        [C]: in function 'CalcTextSize'
        [string "[T] @pigui/libs/text.lua"]:365: in function 'fun'
        [string "[T] @pigui/libs/wrappers.lua"]:557: in function 'withFont'
        [string "[T] @pigui/libs/text.lua"]:364: in function 'addStyledText'
        [string "[T] @pigui/modules/flight-ui/reticule.lua"]:399: in function 'displayFrameData'
        [string "[T] @pigui/modules/flight-ui/reticule.lua"]:829: in function 'displayReticule'
        [string "[T] @pigui/modules/flight-ui/reticule.lua"]:839: in function <[string "[T] @pigui/modules/flight-ui/reticule.lua"]:835>
        [C]: in function 'xpcall'
        [string "[T] @pigui/libs/wrappers.lua"]:34: in function 'pcall'
        [string "[T] @pigui/views/game.lua"]:168: in function 'drawGameModules'
        [string "[T] @pigui/views/game.lua"]:195: in function 'fun'
        [string "[T] @pigui/libs/wrappers.lua"]:90: in function 'window'
        [string "[T] @pigui/views/game.lua"]:191: in function 'fun'
        [string "[T] @pigui/libs/wrappers.lua"]:597: in function 'withStyleColors'
        [string "[T] @pigui/views/game.lua"]:190: in function <[string "[T] @pigui/views/game.lua"]:177>

### Expected behaviour
An always visible reticule

### Steps to reproduce
It's hard to catch. Sometimes it disappears after a few seconds, sometimes after minutes.

Start a new game
Takeoff
Jump to a system
Land

Current master
Arch Linux

image

Web-eWorks:
Fixes #5385.

This PR introduces something I've had sitting in a work branch for several months now: a move towards a composition-based object model rather than a strict inheritance model.

I've added fully-serialized C++ components that can be queried and composed at runtime; other than a need to register a serializer in some sort of centralized fashion the composition functionality is usable at any time by any piece of code without needing to know the concrete type of the body object it is operating on. The ship FixedGuns and Propulsion variables have been migrated to use this new system, although some further work is required to handle querying these components from Lua in a sane manner.

I've also added the ability for Lua to associate arbitrary tables with Body objects as components - also fully serialized - and query them at runtime. This allows for more fully-featured data ownership compared to using a PropertyMap or maintaining a separate lookup table keyed by the object. I intend to migrate the EquipSet functionality to use this LuaComponent feature at some point, though that will likely involve a refactor or rewrite of the equipment handling at the same time.

There is not currently any event handling for LuaComponents when the owning Body object is deleted, though that's something that can be implemented fairly easily if needed. Persisting LuaComponents across body destruction (e.g. when leaving and re-entering a system) is something that will still need to be done manually at a higher level of abstraction.

I know this PR is heavy on implementation and light on use-cases, but it makes moving code to a modular, composition-based approach much easier moving forwards and is already used internally by several of my work branches.

The API related to these two features is quite simple; for C++ objects you need to register a component type, optionally register its serializer, and then just add a new component to the body:

```c++
struct MyComponent {
    MyComponent();
    void SaveToJson(Json &out, Space *space);
    void LoadFromJson(const Json &obj, Space *space);
};

void SomeRegisterFunc() {
    BodyComponentDB::RegisterComponent<MyComponent>("MyComponent");
    BodyComponentDB::RegisterSerializer<MyComponent>();
}

void InitSomeBody(Body *body) {
    // Serialization and creation of the component when loading a save is automatic
    MyComponent *comp = body->AddComponent<MyComponent>();
}
```

Lua components are similar; simply make a new table and call SetComponent on a body with the value.
```lua
local TestComponent = utils.inherits(nil, "TestComponent")

Serializer:RegisterClass("TestComponent", TestComponent)

function someTestFunc(body)
    body:SetComponent("TestComponent", TestComponent.New())

    -- This value will be serialized with the body and loaded via TestComponent.Unserialize
    body:GetComponent("TestComponent").someField = "test"
end
```

### DIFF
--- a/data/modules/Debug/TestComponent.lua
+++ b/data/modules/Debug/TestComponent.lua
@@ -1,0 +1,40 @@
+-- Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+-- This file tests adding and retrieving Lua BodyComponents to the player,
+-- and is used to test serialization of body components.
+
+local Event = require 'Event'
+local Game = require 'Game'
+local Serializer = require 'Serializer'
+local utils = require 'utils'
+
+local TestComponent = utils.inherits(nil, 'TestComponent')
+
+function TestComponent.New()
+	local self = setmetatable({}, TestComponent.meta)
+
+	self.testStr = "I am a test string"
+	self.testNumber = 1234
+
+	return self
+end
+
+--[[
+Serializer:RegisterClass('TestComponent', TestComponent)
+
+Event.Register('onGameStart', function()
+	print('TestComponent #onGameStart')
+	local comp = Game.player:GetComponent('TestComponent')
+
+	if comp then
+		print("Loaded from save file, test component:")
+		utils.print_r(comp)
+	else
+		print("Setting player component")
+		comp = TestComponent.New()
+		Game.player:SetComponent('TestComponent', comp)
+		utils.print_r(Game.player:GetComponent('TestComponent'))
+	end
+end)
+--]]
--- a/src/BodyComponent.h
+++ b/src/BodyComponent.h
@@ -41,9 +41,11 @@ public:
 
 	// Polymorphic interface to support generic deletion operations
 	struct PoolBase {
-		PoolBase(size_t i) :
-			componentIndex(i) {}
+		PoolBase(size_t index, size_t type) :
+			componentIndex(index),
+			componentType(type) {}
 		size_t componentIndex = 0;
+		size_t componentType = 0;
 		SerializerBase *serializer = nullptr;
 
 		virtual void deleteComponent(Body *body) = 0;
--- a/src/BodyComponent.h
+++ b/src/BodyComponent.h
@@ -33,7 +33,7 @@ public:
 
 		std::string typeName;
 		virtual void toJson(const Body *body, Json &obj, Space *space) = 0;
-		virtual BodyComponent *fromJson(Body *body, const Json &obj, Space *space) = 0;
+		virtual void fromJson(Body *body, const Json &obj, Space *space) = 0;
 	};
 
 	// Polymorphic interface to support generic deletion operations
@@ -76,6 +76,10 @@ public:
 
 	// Type-specific serialization implementation. Delegates to the type's
 	// internal serialization methods and provides the needed glue code.
+	//
+	// The Component::LoadFromJson method will be called after the component
+	// is constructed and added to the body, and may potentially have defaults
+	// set by the owning Body before it is deserialized.
 	template <typename T>
 	struct Serializer final : public SerializerBase {
 		Serializer(std::string name, Pool<T> *pool) :
@@ -89,12 +93,10 @@ public:
 			pool->get(body)->SaveToJson(obj, space);
 		}
 
-		virtual BodyComponent *fromJson(Body *body, const Json &obj, Space *space) override
+		virtual void fromJson(Body *body, const Json &obj, Space *space) override
 		{
-			pool->deleteComponent(body);
 			auto *component = pool->newComponent(body);
 			component->LoadFromJson(obj, space);
-			return component;
 		}
 	};
 
--- a/src/BodyComponent.h
+++ b/src/BodyComponent.h
@@ -1,0 +1,142 @@
+// Copyright © 2008-2020 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+#pragma once
+
+#include "JsonFwd.h"
+#include "core/TypeId.h"
+
+#include <cassert>
+#include <cstddef>
+#include <map>
+#include <vector>
+
+class Body;
+class Space;
+class BodyComponent {};
+
+/*
+	BodyComponentDB provides a simple interface to support dynamic composition
+	of game objects.
+
+	It is intended for use as a transitional interface to facilitate the
+	migration of Pioneer's inheritance-hierarchy based object model to an ECS-
+	style system, and thus is not and never will be a permanent solution.
+
+	Please direct all concerns about speed and efficiency to /dev/null
+	Improvements to compile times are welcomed wherever possible.
+*/
+class BodyComponentDB {
+public:
+	// Polymorphic interface to support generic serialization operations
+	// This functionality is separated to facilitate components that do not wish
+	// to be serialized.
+	struct SerializerBase {
+		SerializerBase(std::string name) :
+			typeName(name) {}
+		std::string typeName;
+		virtual void toJson(const Body *body, Json &obj, Space *space) = 0;
+		virtual BodyComponent *fromJson(Body *body, const Json &obj, Space *space) = 0;
+	};
+
+	// Polymorphic interface to support generic deletion operations
+	struct PoolBase {
+		PoolBase(size_t i) :
+			componentIndex(i) {}
+		size_t componentIndex = 0;
+		SerializerBase *serializer = nullptr;
+
+		virtual void deleteComponent(Body *body) = 0;
+	};
+
+	template <typename T>
+	struct Serializer;
+
+	// Type-specific component pool; uses std::map as a backing store.
+	// This is not meant to be particularly performant, merely to transition API usage.
+	template <typename T>
+	struct Pool final : public PoolBase {
+		using PoolBase::PoolBase;
+
+		virtual void deleteComponent(Body *body) override { m_components.erase(body); }
+		// Create a new component, or return the existing one.
+		T *newComponent(const Body *body) { return &m_components[body]; }
+		// Assert that a component exists for this body and return it
+		T *get(const Body *body) { return &m_components.at(body); }
+
+	private:
+		template <typename U>
+		friend struct BodyComponentDB::Serializer;
+		std::map<const Body *, T> m_components;
+	};
+
+	// Type-specific serialization implementation. Delegates to the type's
+	// internal serialization methods and provides the needed glue code.
+	template <typename T>
+	struct Serializer final : public SerializerBase {
+		Serializer(std::string name, Pool<T> *pool) :
+			SerializerBase(name),
+			pool(pool)
+		{}
+		Pool<T> *pool;
+
+		virtual void toJson(const Body *body, Json &obj, Space *space) override
+		{
+			pool->get(body)->SaveToJson(obj, space);
+		}
+
+		virtual BodyComponent *fromJson(Body *body, const Json &obj, Space *space) override
+		{
+			pool->deleteComponent(body);
+			auto *component = pool->newComponent(body);
+			component->LoadFromJson(obj, space);
+			return component;
+		}
+	};
+
+	// Returns (and creates) a type-specific pool.
+	template <typename T>
+	static Pool<T> *GetComponentType()
+	{
+		auto iter = m_componentPools.find(TypeId<T>::Get());
+		if (iter == m_componentPools.end()) {
+			iter = m_componentPools.emplace(TypeId<T>::Get(), new Pool<T>(m_componentIdx++, TypeId<T>::Get())).first;
+			m_componentTypes.push_back(TypeId<T>::Get());
+		}
+
+		return static_cast<Pool<T> *>(iter->second.get());
+	}
+
+	// Returns (if present) the polymorphic interface to component associated with the given index
+	// This differs from the type-ID and is volatile between program restarts
+	static PoolBase *GetComponentType(size_t componentIndex) { return m_componentPools.at(m_componentTypes[componentIndex]).get(); }
+
+	// Register a serializer for the given type.
+	template <typename T>
+	static bool RegisterSerializer(std::string typeName)
+	{
+		assert(!m_componentSerializers.count(typeName));
+		SerializerBase *serial = new Serializer<T>(typeName, GetComponentType<T>());
+		m_componentSerializers.emplace(typeName, serial);
+		GetComponentType<T>()->serializer = serial;
+		return true;
+	}
+
+	// Returns a pointer to the registered Serializer instance for a type identified by the given name, or nullptr.
+	// To retrieve the serializer instance for a given type index, use GetComponentType(idx)->serializer
+	// or GetComponentType<T>()->serializer.
+	static SerializerBase *GetSerializer(const std::string &typeName)
+	{
+		auto iter = m_componentSerializers.find(typeName);
+		if (iter != m_componentSerializers.end())
+			return iter->second.get();
+
+		return nullptr;
+	}
+
+private:
+	static std::map<size_t, std::unique_ptr<PoolBase>> m_componentPools;
+	static std::map<std::string, std::unique_ptr<SerializerBase>> m_componentSerializers;
+	static std::vector<size_t> m_componentTypes;
+	static size_t m_componentIdx;
+};
--- a/src/BodyComponent.h
+++ b/src/BodyComponent.h
@@ -17,14 +17,9 @@ class BodyComponent {};
 
 /*
 	BodyComponentDB provides a simple interface to support dynamic composition
-	of game objects.
-
-	It is intended for use as a transitional interface to facilitate the
-	migration of Pioneer's inheritance-hierarchy based object model to an ECS-
-	style system, and thus is not and never will be a permanent solution.
-
-	Please direct all concerns about speed and efficiency to /dev/null
-	Improvements to compile times are welcomed wherever possible.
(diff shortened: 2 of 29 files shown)


data/modules/Debug/TestComponent.lua
```diff
@@ -0,0 +1,40 @@
+-- Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details
```
Gliese852, line 1: is this file not completed?

Web-eWorks: This file was committed both to provide an example of the API and also to act as a unittest to verify the system does in fact work. I've left it present but commented out as it's not required to run the test every time we start a new game.

Gliese852: @Web-eWorks ok, why the file is not commented out completely, it seems that what is not commented out still does nothing?
Also, maybe it would be good to leave a mark like "uncomment for test", for code readers of my level.

impaktor: I'm happy with the way this file is, that only the code that makes it "live" is commented out.

src/BodyComponent.h
```diff
@@ -0,0 +1,212 @@
...
+	// Multiple serializers can be registered for the given type and used while
+	// loading for backwards compatibility, however only the last-registered
+	// serializer will be used when serializing to JSON
+	template <typename T>
+	static bool RegisterSerializer(std::string typeName)
+	{
+		assert(!m_componentSerializers.count(typeName));
+		Pool<T> *pool = GetComponentType<T>();
+
+		SerializerBase *serial = new Serializer<T>(typeName, pool);
```
Gliese852, line 186: It seems this memory is not deleted anywhere?

Web-eWorks: Pools are never unregistered - they're created on first use and live for the entire program lifetime. I need to migrate them to being explicitly created in a startup function for thread safety however...

src/BodyComponent.h
```diff
@@ -0,0 +1,212 @@
...
+		pool->serializer = serial;
+
+		m_componentSerializers.emplace(typeName, serial);
```
Gliese852, line 189: Oh no, this is put in `unique_ptr`, but it is also put as a raw pointer in the `pool->serializer`?

src/BodyComponent.h
```diff
@@ -0,0 +1,212 @@
...
+
+	// Register a component type to be queryable from Lua with body:GetComponent()
+	template <typename T>
+	static bool RegisterLuaInterface(std::string typeName)
+	{
+		Pool<T> *pool = GetComponentType<T>();
+		if (pool->luaInterface)
+			return false;
+
+		pool->luaInterface = new LuaInterface<T>(pool);
```
Gliese852, line 171: But this memory is definitely not deleted anywhere?

src/BodyComponent.h
Web-eWorks: The unique_ptr is just so the memory is automatically cleaned up at program termination - as far as the system is concerned this can never be a dangling pointer.

src/BodyComponent.h
```diff
@@ -0,0 +1,212 @@
...
+			return iter->second.get();
+
+		return nullptr;
+	}
+
+private:
+	static std::map<size_t, std::unique_ptr<PoolBase>> m_componentPools;
+	static std::map<std::string, std::unique_ptr<SerializerBase>> m_componentSerializers;
+	static std::map<std::string, PoolBase *> m_componentNames;
+	static std::vector<PoolBase *> m_componentTypes;
```
Gliese852, line 210: Maybe it's worth explicitly limiting the size of this array to 64 somewhere?

src/BodyComponent.h
Web-eWorks: This is indeed not deleted anywhere - it is technically leaked at program termination but I can add a destructor to PoolBase that deletes the interface if present.

src/BodyComponent.h
Web-eWorks: Yes - when I move to explicit initialization (before merging this PR) I'll add an assert when trying to register components that ensures we don't go above 64 registered components.

src/BodyComponent.h
Web-eWorks: And specifically, the use of unique_ptr is because there is a many-to-one relationship between serializers and pools; you can register multiple serializers to handle the case in which a component is migrated to a different name or different layout (although there's not currently a good way to implement that at current) so there may be multiple serializers but only one referenced from the pool itself.

src/lua/LuaShip.cpp
```diff
@@ -1146,7 +1146,7 @@ static int l_ship_get_gun_temperature(lua_State *l)
 {
 	Ship *s = LuaObject<Ship>::CheckFromLua(1);
 	int gun = luaL_checkinteger(l, 2);
-	LuaPush(l, s->GetGunTemperature(gun));
+	LuaPush(l, s->GetFixedGuns()->GetGunTemperature(gun));
```
Gliese852, line 1149: why not `GetComponent` ?

Web-eWorks: I was too lazy to add a commit to change this specific occurrence - will clean that up.

src/ShipAICmd.cpp
```diff
@@ -722,8 +723,8 @@ static vector3d GenerateTangent(DynamicBody *dBody, FrameId targframeId, const v
 //4 - probable path intercept
 int CheckCollision(DynamicBody *dBody, const vector3d &pathdir, double pathdist, const vector3d &tpos, double endvel, double r)
 {
-	if (!dBody->Have(DynamicBody::PROPULSION)) return 0;
-	Propulsion *prop = dBody->GetPropulsion();
+	if (!dBody->HasComponent<Propulsion>()) return 0;
```
Gliese852, line 726: It seems that in about all cases it will be `true`, which means we will almost always do the same thing twice. Maybe just return 0 if `GetComponent` returned `nullptr`?

Web-eWorks: Because the underlying container is a `std::map`, GetComponent will throw an exception if the body does not have the specified component. HasComponent is pretty fast, as it's a single hashtable lookup and a bitset index.

I may eventually move to a model where GetComponent returns `nullptr` if the underlying datastructure is amenable to it.

Web-eWorks:
This branch resolves (in a trivial one-liner that could be split into a separate PR admittedly) a very long-running bug with the LuaObject system that was causing lua handles for C++ body objects to incorrectly deregister other valid handles when they were garbage-collected. It resolves the error responsible for the missing reticle on master branch, and may very possibly resolve a *number* of other heisenbugs we've had with Lua not quite behaving correctly.

In my testing this (or the PropertyMap2.0 PR) seems to have resolved #4803, but I'd appreciate a hefty amount of testing as the regression surface for this level of change is quite large.

src/ShipAICmd.cpp
Gliese852: but no, `GetComponent` does a check identical to `HasComponent` before accessing.

src/lua/LuaBody.cpp
```diff
@@ -118,6 +118,55 @@ static int l_body_attr_path(lua_State *l)
...
+	Body *b = LuaObject<Body>::CheckFromLua(1);
+	std::string componentName = luaL_checkstring(l, 2);
+
+	BodyComponentDB::PoolBase *pool = BodyComponentDB::GetComponentType(componentName);
+	if (pool) {
+		return luaL_error(l, "Cannot set C++ body component '%s' to a lua value.");
+	}
+
+	if (componentName == "__properties") {
+		return luaL_error(l, "'__properties' is a reserved name for body components.");
```
Gliese852, line 155: if I'm reading the code correctly, you don't put the components in the `__properties` subtable.

src/ShipAICmd.cpp
Web-eWorks: You are indeed correct, my apologies. It's been so long since I initially wrote this all the little details have slipped from my mind.

src/lua/LuaBody.cpp
Web-eWorks: So, before PropertyMap 2.0 the uservalue on a LuaObject was the properties table directly. We no longer use a Lua table to hold properties, and instead all properties are stored in a PropertyMap on the C++ side. The uservalue for a LuaObject is now used to store LuaComponent tables, with the __properties key reserved to hold a lightuserdata pointer to the PropertyMap for the object. This allows quickly querying properties without needing to do a dynamic cast, because the same index metamethod is used for everything from LuaVectors to Ships.

In sum, there is no __properties subtable; it's not even a table at all.

Gliese852: oops really it's not a subtable.  But my complaint remains - you don't store components in an object named "__properties", but write that it is reserved for components. Or am I mistranslating the meaning?

Web-eWorks: Hmm, good point - I wrote that when I was too tired for good english - it's a reserved name that lua components are not allowed to use, because otherwise they could inadvertently corrupt the property map pointer and cause difficult-to-debug downstream errors.

src/lua/LuaBody.cpp
```diff
@@ -118,6 +118,55 @@ static int l_body_attr_path(lua_State *l)
...
+	lua_getuservalue(l, 1);
+	lua_pushvalue(l, 2);
+	lua_rawget(l, -2);
+
+	return 1;
+}
+
+static int l_body_set_component(lua_State *l)
+{
+	Body *b = LuaObject<Body>::CheckFromLua(1);
```
Gliese852, line 146: Unused variable.

src/lua/LuaBodyComponent.h
```diff
@@ -0,0 +1,35 @@
+// Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details
```
Gliese852, line 1: It seems this file is not included anywhere.

src/lua/LuaObject.cpp
```diff
@@ -706,6 +732,96 @@ void LuaObjectBase::RegisterSerializer(const char *type, SerializerPair pair)
...
+	lua_State *l = Lua::manager->GetLuaState();
+	LUA_DEBUG_START(l);
+
+	luaL_getsubtable(l, LUA_REGISTRYINDEX, "LuaObjectRegistry");
+
+	// Get the LuaObject for the given LuaWrappable passed in
+	lua_pushlightuserdata(l, object);
+	lua_rawget(l, -2); // Registry, LuaObject
+
+	if (lua_isnil(l, -1)) {
```
Gliese852, line 746: I wonder if it can happen that the object does not have registration? Is this allowed?

Web-eWorks: It is *technically* impossible for a Body that has been pushed to lua to not be present in the registry, as the only time a Body-derived LuaObject is removed from the registry is after the underlying Body has been deleted and the LuaObject is orphaned.

However, we call SerializeComponents on all Body objects during the save step, so we need to handle the case in which we're serializing a body that didn't get pushed to Lua during its lifetime.

src/lua/LuaBodyComponent.h
Web-eWorks: Correct - none of the existing components have a Lua binding so this is currently unused. It should be included by whatever file calls `BodyComponentDB::RegisterLuaInterface<T>(typeName)`. This file is separate to avoid dragging in the LuaObject dependency into the BodyComponent header (and thus into most of our C++ files).

src/lua/LuaBody.cpp
Web-eWorks: "Not a bug" - the CheckFromLua needs to validate the first parameter is the right userdata. This is a common pattern in LuaObject bindings.

src/BodyComponent.h
Web-eWorks: Added an assert to prevent more than 64 components from being registered.

src/BodyComponent.h
Web-eWorks: PoolBase deletes the luaInterface if present now.

src/BodyComponent.h
Web-eWorks: Added a comment to denote the pointed-to object's lifetime.

data/modules/Debug/TestComponent.lua
Web-eWorks: Leaving this file as is, as the uncommented portions are no-op and it's clear enough from the header comment why the file exists as far as I'm concerned.

src/lua/LuaBody.cpp
Web-eWorks: Revised the error message, should be clearer now!

src/lua/LuaObject.cpp
```diff
@@ -562,23 +560,52 @@ void LuaObjectBase::Register(LuaObjectBase *lo)
...
 
-void LuaObjectBase::Deregister(LuaObjectBase *lo)
+void LuaObjectBase::RegisterPersistent(LuaObjectBase *lo)
 {
-	LuaWrappable *o = lo->GetObject();
-
 	lua_State *l = Lua::manager->GetLuaState();
+	LUA_DEBUG_START(l); // lo userdata
+
+	Register(lo);
```
Gliese852, line 568: It seems to me that the fact that the object gets into both registers, and then from one of them is deleted by chance - is not very sane. I don't know, maybe do something like `Register(l, "LuaObjectPersistentRegistry")` ?

src/lua/LuaBody.cpp
Gliese852: It seems that this is the first unused variable in Pioneer code that I and my static analyzer see, an he is unhappy. I have looked at all occurrences of `CheckFromLua`, the variable is always used afterwards. You don't have to put the result in a variable if you don't need it later, and you don't have to suffer a compile-time warning.

src/lua/LuaObject.cpp
Web-eWorks: The persistent registry just exists to prevent the LuaObject from being garbage collected when it's no longer referenced directly by Lua - we still have to register in the transient registry to allow the rest of the LuaObject system to work properly. If I could make it work in a single registry table I'd be happy, but the need to support RefCounted / GC'd objects as well as persistent C++ owned objects means it has to be done this way.

The object will not be removed from the transient registry while a reference is held in the persistent registry - the only time something is deleted from the transient registry by chance is during a garbage collection cycle when a LuaSharedObject / LuaCopyObject is cleaned up due to a lack of references.

Gliese852: oops, I really misunderstood how it works, sorry!

src/lua/LuaBody.cpp
Web-eWorks: Well, if it's for the sake of your static analyzer I suppose I can just discard the return value. :smile:

src/lua/LuaSerializer.cpp
```diff
@@ -435,6 +439,58 @@ void LuaSerializer::FromJson(const Json &jsonObj)
 	LUA_DEBUG_END(l, 0);
 }
 
+void LuaSerializer::SaveComponents(Json &jsonObj, Space *space)
+{
+	Json &bodies = jsonObj["space"]["bodies"];
+	if (!bodies.is_array())
+		return;
+
+	// Note: this loop relies on the ordering and contents of Space::m_bodies not changing
```
Gliese852, line 460: This is a dark comment.

Web-eWorks: Indeed it is. I'd like a more generic solution that serializes Body references similar to how lua table references are serialized, but the lua serialization is a general pile of order-dependent kludge and nothing can solve that. (No offense to the original author of LuaSerializer is meant, it's very brilliant for the domain of the problem it needs to solve and there's really no way around the ordering issues)

impaktor: > This is a dark comment.

Makes me think of this poem:

_I have seen the dark universe yawning,
Where the black planets roll without aim;
Where they roll in their horror unheeded, without knowledge or lustre or name._

Web-eWorks:
Unless any significant bugs have been found with this branch (please test it, people!) I will be merging this over the weekend and introducing a new PR that refactors cargo handling into a brand-new LuaComponent and finally (finally!) separates it from EquipSet and the Cargo-as-Equipment paradigm.

I think I've ironed out all of the ordering bugs and gotchas related to this new system as a result of developing that branch, so I'm pretty confident that this is ready to merge and doesn't have any issues remaining.

PR #5386, opened 2022-06-17
Better Grid Rendering

Web-eWorks:
They say a picture is worth 1000 words... I've improved the grid rendering in the ModelViewer to support infinite zoom level-of-detail, different grid colors for primary and secondary grid increments, arbitrary line width, and smooth antialiasing. I've also implemented a spherical grid for rendering planetary surface information in the SystemMap.

I'll leave you with these pictures and turn it over to @Gliese852 to use the new grid rendering in the SystemMap:

### DIFF
--- a/data/shaders/opengl/grid.frag
+++ b/data/shaders/opengl/grid.frag
@@ -1,0 +1,74 @@
+// Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+// Adapted from https://ourmachinery.com/post/borderland-between-rendering-and-editor-part-1/
+// Credit to Tobias Persson for providing an example implementation.
+
+#include "attributes.glsl"
+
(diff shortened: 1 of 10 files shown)


data/shaders/opengl/grid.frag
```diff
@@ -0,0 +1,74 @@
...
+{
+	vec2 cover = 1 - abs(clamp(mod(uv + line_width * 0.5, cell_size) / line_width, 0, 1) * 2 - 1);
+	return max(cover.x, cover.y);
+}
+
+void main(void)
+{
+	// Calculate the screen-space derivatives of the UV coordinates
+	vec4 derivatives = vec4(dFdx(v_uv0), dFdy(v_uv0));
+	vec2 dudv = vec2(length(derivatives.xz), length(derivatives.yw));
```
Gliese852, line 38: are this two lines similar to the expression:
`vec2 dudv = vec2(length(dFdx(v_uv0)), length(dFdy(v_uv0)));`

PR #5388, opened 2022-06-26
Improve Lua documentation, add static type analysis definition files

Web-eWorks:
At the great expense of possibly further reducing the lifetime of my keyboard from all of the furious typing, I've added a significant amount of documentation to the `utils` Lua module and added type information "meta files" for several foundational Lua classes defined in C++.

I've also added a helper `utils.class` function which normalizes the LuaObject constructor paradigm with lua-defined classes for less confusion; and refactored the `utils.take` and `utils.reverse` methods to avoid table copies and properly carry out their function.

The new type-information files live in the `data/meta` directory and are simply specially-formatted Lua files containing function and class prototype declarations marked up with EmmyLua syntax that a lua language server can understand. These files are tailored specifically for [sumneko/lua-language-server](https://github.com/sumneko/lua-language-server) as most of our developers use this implementation and it is the most fully-featured language server that is available in multiple IDEs.

Adding new type information is simple; create a new Lua file in the `data/meta` directory and add a `---@meta` comment near the top of the file. Further lua code and documentation will be read by the lua language server and used to populate the type information available during static analysis, but will not be executed by the game at runtime, allowing fields and methods on C++ userdata and other type information that would be invisible to the language server's static analysis to be discovered and used.

I have not documented our entire class hierarchy due to time constraints, as this was originally intended to be a small PR merged separately before the cargo component refactor. I would highly appreciate additional effort from other developers in completing this type documentation - it's quite simple to add this information by simply reading the C++ sources in `src/lua`.

Further documentation efforts would also be appreciated to mark up types in the Lua game files where they're unable to be inferred - e.g. in `onShipXXXX` event handlers, the `ship` parameter of the handler will need to be explicitly declared with `---@param ship Ship`; that is a situation where the type of the parameter is simply unable to be inferred, even though it is strongly known from the C++ side.

To make use of these type information files in VS Code, add an entry to the Lua > Workspace > Library setting (in the UI or in the `settings` object of your workspace file) with the path from the workspace root folder to the `data/meta` folder, like so:
```json
"Lua.workspace.library": [
    "data/meta"
]
```

For developers not using VS Code, consult the documentation for your specific language server adapter plugin to properly set the `workspace.library` configuration value.

**NOTE:** this PR will only remain open for a very short time (less than 24hrs) to gather feedback, as it is a required component to merge the cargo component refactor branch.

### DIFF
--- a/data/meta/Body.lua
+++ b/data/meta/Body.lua
@@ -1,0 +1,144 @@
+-- Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+-- This file implements type information about C++ classes for Lua static analysis
+
+---@meta
+
+---
+---@class Body
+---
+---@field label string
+---@field seed integer
+---
+--- The <SystemPath> that points to this body.
+--- Only available for non-dynamic bodies.
+---@field path? SystemPath
+---
+--- The type of the body, as a <Constants.BodyType> constant.
+---@field type? BodyType
+---
+--- The supertype of the body, as a <Constants.BodySuperType> constant
+---@field superType? BodySuperType
+---
+--- The non-dynamic body attached to the frame this dynamic body is in.
+--- Only available for dynamic bodies.
+---@field frameBody? Body
+---
+--- Whether the frame this body is in is a rotating frame.
+--- Only available for dynamic bodies.
+---@field frameRotating? boolean
+---
+
+---@class Body
+local Body = {}
+
+--- Get a C++ or Lua component object from the body if present.
+---
+--- Note: the caller should check the return value if there is a possibility if
+--- the component may not be present on this Body.
+---@generic T
+---@param name `T`
+---@return T
+function Body:GetComponent(name) end
+
+--- Set or clear the given Lua component on the body to the passed object.
+---@generic T
+---@param name `T`
+---@param value T|nil
+function Body:SetComponent(name, value) end
+
+--- Determine if the body is a dynamic body.
+---@return boolean
+function Body:IsDynamic() end
+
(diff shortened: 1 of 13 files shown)


data/meta/Body.lua
```diff
@@ -0,0 +1,144 @@
...
+--- The non-dynamic body attached to the frame this dynamic body is in.
+--- Only available for dynamic bodies.
+---@field frameBody? Body
+---
+--- Whether the frame this body is in is a rotating frame.
+--- Only available for dynamic bodies.
+---@field frameRotating? boolean
+---
+
+---@class Body
```
Gliese852, line 33: `[Lua Diagnostics. duplicate-doc-class] [W] Duplicate defined class.`

data/meta/Game.lua
```diff
@@ -0,0 +1,101 @@
...
+-- This file implements type information about C++ classes for Lua static analysis
+-- TODO: this file is partially type-complete, please expand it as more types are added.
+
+---@meta
+
+---@class Game
+---
+---@field player Body #TODO: add Ship / Player to type tree
+---@field system StarSystem? The system the game is currently playing in or nil when in hyperspace.
+---@field systemView unknown #TODO: add type info for SystemView interface
```
Gliese852, line 13: On the word `unknown`:
`[Lua Diagnostics. undefined-doc-name] [W] Undefined type or alias.`

data/meta/Body.lua
Web-eWorks: Based on conversation in IRC, this appears to be caused by an old version of lua-language-server, as this diagnostic string is no longer present in the language server sources.

Web-eWorks:
Above issues were resolved by updating the language server version in question to something a little bit more recent than Jan 2021 :smile:

PR #5391, opened 2022-07-02
Improve error reporting and debug printing utilities

Web-eWorks:
This PR adds a few utility functions to vastly improve the quality of debugging information we're able to collect in the event of a Lua error or crash.

I've added `debug.dumpstack(level)` which performs a verbose trace through the callstack, recursively dumping the contents of parameters and locals in each stack frame to a string. This allows the state of the program at the point of the error to be inspected to attempt to determine what is responsible for the crash, as well as providing a little more context as to what events lead to the crash in question.

The `ui.pcall` utility has been updated to automatically log a stack dump if an error was triggered. A short callstack will still be displayed in the UI where appropriate, but the full stack dump will be written to `output.txt`.

C++ panic and pcall error handlers also now log stack dumps when encountering a fatal error.

I've changed the 'trusted file' indicator string used when loading Lua chunks to `@[T]` instead of `[T] @`, which matches Lua's assumptions about whether a chunk name is a filepath or a raw string. I've updated `debug.stacktrace()` with a new implementation that respects this change.

I've also made some utility changes to the Lua-scope logging tools:
- Add Lua `logVerbose()` function to complement `logWarning()`.
- Update our `print()` implementation to write to the log as well as the console / stdout.
- Rewrite `utils.print_r()` to print a single string instead of 100s of tiny strings.
- The log functions now flush the log file on write, to prevent log messages from being lost or truncated in event of a crash.

### DIFF
--- a/data/pigui/views/game.lua
+++ b/data/pigui/views/game.lua
@@ -168,7 +168,7 @@ local function drawGameModules(delta_t)
 			local ok, err = ui.pcall(module.draw, module, delta_t)
 			if not ok then
 				module.disabled = true
-				logWarning(err)
+				-- TODO: visually notify the user of an error
 			end
 		end
 	end
--- a/data/libs/utils.lua
+++ b/data/libs/utils.lua
@@ -336,37 +336,49 @@ end
 --
 utils.print_r = function(t)
 	local print_r_cache={}
-	local function sub_print_r(t,indent,rec_guard)
+	local print_str_cache = {}
+
+	local write = function(indent, str, ...)
+		table.insert(print_str_cache, string.rep(" ", indent) .. string.format(str, ...))
+	end
+
+	local function sub_print_r(t, indent)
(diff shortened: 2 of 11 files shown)


data/pigui/views/game.lua
```diff
@@ -168,7 +168,7 @@ local function drawGameModules(delta_t)
 			local ok, err = ui.pcall(module.draw, module, delta_t)
 			if not ok then
 				module.disabled = true
-				logWarning(err)
+				-- TODO: visually notify the user of an error
```
impaktor, line 171: To be fixed prior to merge?

Web-eWorks: No - this will be resolved by adding a queue for error notices and displaying a popup/notification to the user in-UI. Well beyond the scope of this change, which increases the amount of information present in an error log.

Issue #4392, opened 2018-08-11
FreeDesktop spec for user folder should be respected on Linux


pcercuei:
### Observed behaviour

Pioneer saves to ```~/.pioneer```

### Expected behaviour

Pioneer saves to ```$XDG_DATA_HOME/pioneer``` if that variable is set, else ```~/.local/share/pioneer```

### Steps to reproduce

__My pioneer version (and OS):__

git master on Debian GNU/Linux

johnbartholomew:
It's actually a little more complicated than that, because the XDG spec separates different types of user data:
- config files in $XDG_CONFIG_HOME
- "data" files in $XDG_DATA_HOME
- cache files in $XDG_CACHE_HOME

And lookups for data and config are supposed to look in multiple directories as controlled by $XDG_CONFIG_DIRS and $XDG_DATA_DIRS.

I don't think we have any cache type files (we used to; some model data got processed and cached IIRC) any more. We certainly have config. We also have:
- log files
- save files -- is that config?
- mods -- that's almost certainly data
- screenshots -- config?

Migrating existing users... eh, probably the best thing to do there is to check for the existence of ~/.pioneer/ and add it to the necessary search paths, but only write to the new directories. Maybe write a CONFIG_FILES_HAVE_MOVED file into ~/.pioneer/ if it exists, if we're feeling fancy.

Or easier just ignore ~/.pioneer/ completely, only use the new dirs, and accept that people will have to do some manual file moves. Document it somewhere.

impaktor:
> Migrating existing users... eh, probably the best thing to do there is to check for the existence of ~/.pioneer/ and add it to the necessary search paths, but only write to the new directories.

Sounds like the best option, if we're to do this switch.

> Or easier just ignore ~/.pioneer/ completely, only use the new dirs, and accept that people will have to do some manual file moves.

That sounds like it will cause some chaos in the player base. If we even have one?

> Document it somewhere.

I imagine most people don't read documentation

pcercuei:
I think config.ini and imgui.ini should go to ```${XDG_CONFIG_HOME}/pioneer```; the other files (mods, savestates) should go to ```${XDG_DATA_HOME}/pioneer```.

Note that due to confinement, the Flatpak-packaged Pioneer won't be able to see the current savestates anyway, so it's a perfect timing for the change.

Maybe we can move the folder with some light scripting? In the top-level "pioneer" script for instance, when Pioneer is built from the source.

[...]

PR #5400, opened 2022-09-05
Initial support for the XDG folder specification

Jummit:
Initial stab at #4392

Instead of splitting up the contents of the user folder to fully support the XDG spec, this pull request just puts everything in XDG_DATA_HOME if no .pioneers folder exists.


src/posix/FileSystemPosix.cpp
```diff
@@ -57,7 +57,20 @@ namespace FileSystem {
...
+		struct stat info;
+		stat((path + ".pioneer").c_str(), &info);
+		if (S_ISDIR(info.st_mode)) {
+			// Check for legacy pioneer directory.
+			path += ".pioneer";
+		} else {
+			char *state_home = getenv("XDG_STATE_HOME");
+			if (state_home == NULL) {
+				path += state_home;
+				path += "/pioneer";
```
Web-eWorks, line 69: I believe this check should be for `state_home != NULL`.

pcercuei:
Your commit message says XDG_DATA_HOME, the commit itself says XDG_STATE_HOME, so that's already a problem.

The spec says that XDG_STATE_HOME is for data that is "not important", I can't really see the savestates falling into this category. So XDG_DATA_HOME would make more sense.

Jummit:
> Your commit message says XDG_DATA_HOME, the commit itself says XDG_STATE_HOME, so that's already a problem.
> 
> The spec says that XDG_STATE_HOME is for data that is "not important", I can't really see the savestates falling into this category. So XDG_DATA_HOME would make more sense.

Sorry for that oversight, it's fixed now.

src/posix/FileSystemPosix.cpp
```diff
@@ -57,7 +57,20 @@ namespace FileSystem {
...
+		struct stat info;
+		stat((path + ".pioneer").c_str(), &info);
+		if (S_ISDIR(info.st_mode)) {
+			// Check for legacy pioneer directory.
+			path += ".pioneer";
+		} else {
+			char *state_home = getenv("XDG_DATA_HOME");
+			if (state_home != NULL) {
+				path += state_home;
+				path += "/pioneer";
```
Web-eWorks, line 69: According to what I can find, `XDG_DATA_HOME` is intended to be an absolute path, defaulting to `$HOME/.local/share`. This form would result in `$HOME/$HOME/.local/share` which would result in unintended directory nesting at best and a malformed path at worst.

Additionally, this code does not handle the (possible, albeit improbable) case where XDG_DATA_HOME is set to an empty string (e.g. by a user-specific config) and `getenv()` returns a valid string pointer with length 0.

Other than this issue, I am happy with this change!

Web-eWorks (changes requested):

Web-eWorks:
Sorry about the very long wait time on getting this PR reviewed and approved. When the above changes are implemented, I believe this PR is ready to be merged!

Jummit:
> Sorry about the very long wait time on getting this PR reviewed and approved. When the above changes are implemented, I believe this PR is ready to be merged!

No problem! I've implemented the mentioned changes, let me know if there are any problems, I haven't used C++ in a long time.

src/posix/FileSystemPosix.cpp
```diff
@@ -57,7 +57,20 @@ namespace FileSystem {
...
 #else
-		path += ".pioneer";
+		struct stat info;
+		stat((path + ".pioneer").c_str(), &info);
+		if (S_ISDIR(info.st_mode)) {
+			// Check for legacy pioneer directory.
+			path += ".pioneer";
+		} else {
+			char *data_home = getenv("XDG_DATA_HOME");
+			if (data_home == NULL || strcmp(data_home, "")) {
```
Web-eWorks, line 67: `strcmp` returns 0 when two strings are equal, it's a holdover from the C API. As written, this branch will be taken when `data_home` is anything other than an empty string.

Jummit: Yep, this was the sort of thing I was worried about. It's fixed now.

Web-eWorks (approved):

PR #5407, opened 2022-10-16
Fix #5406: broken scoop mission

### LINKED ISSUE #5406: Scoop missions do not work properly because onCargoDestroyed event is not emitted when cargo is scooped.
### Observed behaviour
If player scoops navigational target for the mission (like rescue capsule) and then tries to press the button "Set destination as navigation target" in the mission details then the mission crashes.
The mission code relies on onCargoDestroyed event to update the list of available debris to collect. If it is not called then mission code tries to return for navigation an object that does not exist anymore.
### Expected behaviour
No mission exception.

### Steps to reproduce
1. Start a scoop mission.
2. Scoop the "navigation target" object.
3. Press the "Set destination as navigation target" in the mission details

master as of 16 Oct 2022 (Debian)

azieba:
Fixes #5406

### DIFF
--- a/src/CargoBody.cpp
+++ b/src/CargoBody.cpp
@@ -127,7 +127,7 @@ bool CargoBody::OnCollision(Body *b, Uint32 flags, double relVel)
 {
 	// ignore collision if its about to be scooped
 	if (b->IsType(ObjectType::SHIP)) {
-		int cargoscoop_cap = Properties().Get("cargo_scoop_cap");
+		int cargoscoop_cap = b->Properties().Get("cargo_scoop_cap");
 		if (cargoscoop_cap > 0) {
 			LuaEvent::Queue("onCargoDestroyed", this);
 			return true;


Web-eWorks (approved):
Good catch, thanks!

impaktor:
@azieba Thanks for contribution. If you want to be in [AUTHORS.txt](https://github.com/pioneerspacesim/pioneer/blob/master/AUTHORS.txt), let us know what name to put.

azieba:
> @azieba Thanks for contribution. If you want to be in [AUTHORS.txt](https://github.com/pioneerspacesim/pioneer/blob/master/AUTHORS.txt), let us know what name to put.

That would be very generous of you to add me to the Authors.txt after committing just 3 characters to the code :)
I do not mind if you use my real name. 
BR,
Andrzej Zięba

impaktor:
Done. 

We sometimes forget to add new contributors, but there's no lower code contribution limit for being in the authors file, anything above 0 lines. (Translations (and comments?) don't count, I think).

> That would be very generous of you to add me to the Authors.txt after committing just 3 characters to the code

Perhaps I have an ulterior motive, in getting you to submit more pull requests...

PR #5410, opened 2022-10-20
Fixes #5403 kill rotation killing HUD and bringing back possibility to turn rotation damping off

### LINKED ISSUE #5403: Kill rotation kills the HUD and not the rotation
### Observed behaviour
When 'kill rotation' is used, the white circle, including distance displays, delta-v bar and the braking distance bar disappears. Cycling HUD modes with TAB does not restore the HUD. Also, rotation damping is broken and is always on.

screen1
screen2

### Expected behaviour
All HUD elements remain visible. Player is able to disable rotation damping.

### Steps to reproduce
1. Press kill rotation button in flight.

8dabc676cbc530ae605af3b054f18637526fb7bd, Win11 x64

azieba:
First issue was undefined icon for kill rotation in lua code for the HUD reticule. The result was assertion and killing the HUD permanently.
The second thing was that handling of rotation damping on/off was completely removed with some recent refactoring. I brought it back using logic from 20220203.

### DIFF
--- a/src/Ship.h
+++ b/src/Ship.h
@@ -343,7 +343,10 @@ public:
 	bool AIMatchVel(const vector3d &vel, const vector3d &powerLimit = vector3d(1.0)) { return m_propulsion->AIMatchVel(vel, powerLimit); }
 	double AIFaceDirection(const vector3d &dir, double av = 0) { return m_propulsion->AIFaceDirection(dir, av); }
 	void SetThrusterState(int axis, double level) { return m_propulsion->SetLinThrusterState(axis, level); }
-	void AIMatchAngVelObjSpace(const vector3d &desiredAngVel, const vector3d &powerLimit = vector3d(1.0)) { return m_propulsion->AIMatchAngVelObjSpace(desiredAngVel, powerLimit); }
+	void AIMatchAngVelObjSpace(const vector3d &desiredAngVel, const vector3d &powerLimit = vector3d(1.0), bool ignoreZeroValues = false)
+	{
+		m_propulsion->AIMatchAngVelObjSpace(desiredAngVel, powerLimit, ignoreZeroValues);
+	}
 };
 
 #endif /* _SHIP_H */
--- a/src/ship/PlayerShipController.cpp
+++ b/src/ship/PlayerShipController.cpp
@@ -21,8 +21,6 @@
 #include <algorithm>
 
 // so that we can summarize the controller action from all subsystems
-// vector3d (0.0) means we want zero speed, for example, so I made optional
-// so that the system can say that it does not want to touch this parameter at all
 struct PlayerShipController::TotalDesiredAction {
 	vector3d desiredLinear = {};  // in ship's frame
 	vector3d desiredAngular = {}; // in ship's local space
@@ -30,8 +28,9 @@ struct PlayerShipController::TotalDesiredAction {
 	vector3d linPower = vector3d(1.0);
 	// true means that desiredLinear contains speed, false means thrust
 	bool desireLinVel = false;
-	// true means that desiredAngular contains speed, false means thrust
-	bool desireAngVel = false;
+	// false means that zero desireAngVel components are to be ignored 
+	// used for rotation damping off
+	bool desireAngVelFull = false;
 };
 
 // static functions with access to private class members
@@ -284,8 +283,8 @@ void PlayerShipController::FlightAssist(const float timeStep, TotalDesiredAction
 	outParams.desireLinVel = true;
 	outParams.desiredLinear = {};
 	// someone could touch the angular speed already, or not
-	if (!outParams.desireAngVel) {
-		outParams.desireAngVel = true;
+	if (!outParams.desireAngVelFull) {
+		outParams.desireAngVelFull = true;
 		outParams.desiredAngular = {};
 	}
 	vector3d &lin = outParams.desiredLinear;
@@ -468,11 +467,7 @@ void PlayerShipController::ApplyTotalAction(const TotalDesiredAction &params)
 		m_ship->m_launchLockTimeout = 0;
 	}
 
-	if (params.desireAngVel) {
-		m_ship->AIMatchAngVelObjSpace(params.desiredAngular, params.angPower);
-	} else {
-		m_ship->GetPropulsion()->SetAngThrusterState(params.desiredAngular);
-	}
+	m_ship->AIMatchAngVelObjSpace(params.desiredAngular, params.angPower, !params.desireAngVelFull);
 }
 
 void PlayerShipController::StaticUpdate(const float timeStep)
@@ -498,19 +493,22 @@ void PlayerShipController::StaticUpdate(const float timeStep)
 	if (m_ship->GetFlightState() == Ship::FLYING) {
 		switch (m_flightControlState) {
 		case CONTROL_FIXSPEED:
-			PollControls(timeStep, true, mouseMotion, act);
+			PollControls(timeStep, mouseMotion, act);
 			FlightAssist(timeStep, act);
 			ApplyTotalAction(act);
 			break;
 		case CONTROL_MANUAL:
 			// maintain the thrust set by the station when undocking while the timeout lasts
 			// but only in manual mode
-			PollControls(timeStep, true, mouseMotion, act);
+			PollControls(timeStep, mouseMotion, act);
 			Util::LimitActualSpeed(*this, timeStep, act);
 			// synchronize the rotation in 'follow orient' mode
 			if (m_followTarget && m_followMode == FOLLOW_ORI) {
 				act.desiredAngular += m_followTarget->GetAngVelocity() * Pi::player->GetOrient();
 			}
+			//m_followMode == FOLLOW_ORI is when we try to match rotation speeds with a target
+			//so all axes should be matched even with rotation damping off 
+			act.desireAngVelFull = m_rotationDamping || (m_followTarget && m_followMode == FOLLOW_ORI);
 			ApplyTotalAction(act);
 			break;
 		case CONTROL_FIXHEADING_FORWARD:
@@ -520,7 +518,7 @@ void PlayerShipController::StaticUpdate(const float timeStep)
 		case CONTROL_FIXHEADING_RADIALLY_INWARD:
 		case CONTROL_FIXHEADING_RADIALLY_OUTWARD:
 		case CONTROL_FIXHEADING_KILLROT:
-			PollControls(timeStep, true, mouseMotion, act);
+			PollControls(timeStep, mouseMotion, act);
 			Util::LimitActualSpeed(*this, timeStep, act);
 			v = m_ship->GetVelocity().NormalizedSafe();
 			if (m_flightControlState == CONTROL_FIXHEADING_BACKWARD ||
@@ -597,14 +595,14 @@ static double clipmouse(double cur, double inp)
 	return inp;
 }
 
-void PlayerShipController::PollControls(const float timeStep, const bool force_rotation_damping, int *mouseMotion, TotalDesiredAction &outParams)
+void PlayerShipController::PollControls(const float timeStep, int *mouseMotion, TotalDesiredAction &outParams)
 {
 	if (AreControlsLocked()) return;
 
 	const float thrustPower = (InputBindings.thrustLowPower->IsActive() ? m_lowThrustPower : 1.0f);
 	outParams.angPower.x = outParams.angPower.y = outParams.angPower.z = thrustPower;
 	// at the moment we do not work with angle thrusters directly at all
-	outParams.desireAngVel = true;
+	outParams.desireAngVelFull = true;
 
 	if (IsAnyLinearThrusterKeyDown()) {
 		if (InputBindings.thrustForward->IsActive())
@@ -688,6 +686,7 @@ void PlayerShipController::PollControls(const float timeStep, const bool force_r
 	if (InputBindings.killRot->IsActive()) SetFlightControlState(CONTROL_FIXHEADING_KILLROT);
 
 	outParams.desiredAngular = wantAngVel;
+
 	if (IsAnyAngularThrusterKeyDown()) {
 		// rotation speed limit identical to thrust power (almost)
 		// and for 100% thrust speed limit is 1 rad/s, but
--- a/src/ship/Propulsion.h
+++ b/src/ship/Propulsion.h
@@ -116,7 +116,7 @@ public:
 	bool AIMatchVel(const vector3d &vel, const vector3d &powerLimit = vector3d(1.0));
 	bool AIChangeVelBy(const vector3d &diffvel, const vector3d &powerLimit = vector3d(1.0)); // acts in object space
 	vector3d AIChangeVelDir(const vector3d &diffvel);										 // object space, maintain direction
-	void AIMatchAngVelObjSpace(const vector3d &angvel, const vector3d &powerLimit = vector3d(1.0));
+	void AIMatchAngVelObjSpace(const vector3d &angvel, const vector3d &powerLimit = vector3d(1.0), bool ignoreZeroValues = false);
 	double AIFaceUpdir(const vector3d &updir, double av = 0);
 	double AIFaceDirection(const vector3d &dir, double av = 0);
 	vector3d AIGetLeadDir(const Body *target, const vector3d &targaccel, double projspeed);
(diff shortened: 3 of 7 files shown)


src/ship/PlayerShipController.cpp
```diff
@@ -505,7 +505,7 @@ void PlayerShipController::StaticUpdate(const float timeStep)
 		case CONTROL_MANUAL:
 			// maintain the thrust set by the station when undocking while the timeout lasts
 			// but only in manual mode
-			PollControls(timeStep, true, mouseMotion, act);
+			PollControls(timeStep, false, mouseMotion, act);
 			Util::LimitActualSpeed(*this, timeStep, act);
 			// synchronize the rotation in 'follow orient' mode
 			if (m_followTarget && m_followMode == FOLLOW_ORI) {
```
Gliese852, line 511: @azieba thanks for the fix, indeed, it can also be done this way! however, if you have follow orientation enabled, it can, on the contrary, spin the ship here, which can be disabled as follows:
```suggestion
			if (m_followTarget && m_followMode == FOLLOW_ORI && m_rotationDamping) {
```

src/ship/PlayerShipController.cpp
```diff
@@ -688,6 +686,8 @@ void PlayerShipController::PollControls(const float timeStep, const bool force_r
 	if (InputBindings.killRot->IsActive()) SetFlightControlState(CONTROL_FIXHEADING_KILLROT);
 
 	outParams.desiredAngular = wantAngVel;
+	outParams.desireAngVel = force_rotation_damping || m_rotationDamping || wantAngVel.Length() >= 0.001;
```
Gliese852, line 689: In fact `desireAngVel` is a runtime type specifier for `desiredAngular`, it seems worth explaining here why we can so easily convert a speed request to a thrust request. Also, the question is, is the desired behavior obtained? Should we limit the rotation speed if we have rotation damping disabled?

Gliese852:
It also would be nice to test this with a joystick.

src/ship/PlayerShipController.cpp
Keeper1st: I've always assumed that the maximum rotational speed with dampening on is based on ship and pilot structural integrity, so it would make sense to have a limit with dampening off as well. If people really are opposed to it, there could be some sort of "safeties off" mode one could apply. There should be some kind of damage to the ship done for rotating beyond tolerances with safeties off, in that case!

src/ship/PlayerShipController.cpp
azieba: This might potentially turn off rotation to the followed target when dumping is off. You are right that there is something wrong here though. I think that the manual mode with fallow target on is kind of semiautomatic mode that should have the dumping turned on like with fixspeed. I will append the PR with a change for that.

src/ship/PlayerShipController.cpp
azieba: The limit for rotation speed is not disabled with rotation damping off. In fact the change is only effective if the player is not pressing any rotation button or in general if rotation requested is close to zero. I tested with rotation damping turned off and I was unable to spin the ship faster then with damping off.
When ship hits something with rotation damping of it could spin faster then limit though, but that should be expected.
From my point of view the desired behavior is obtained.

azieba:
I do not have a joystick so if anybody else with one could test it would be great.

src/ship/PlayerShipController.cpp
azieba: > it seems worth explaining here why we can so easily convert a speed request to a thrust request

I'm not sure I understand what you expect. The desireAngVel is set to false only if we have 0 requested  rotation from inputs this translates to 0 thruster request which is correct in case of rotation damping off.
Do you want me to add a comment about that?

impaktor:
I've probably asked before, but do we need the appveyor build for anything? The build-msvc does the windows build, now, by the looks of it.

What's the difference between bulld-vs-solution, compared to build-msvc?

Ping @fluffyfreak / @Web-eWorks

azieba:
> What's the difference between bulld-vs-solution, compared to build-msvc?

It looks like build-msvc uses cmake and bulld-vs-solution uses msbuild which in turn uses vcxproj files with list of files to compile that is out of sync with actual files in the repository.
I'm not sure but this bulld-vs-solution version is something different then what is described in COMPILING.txt: 1.4 Windows - CMake and MSVC. The file describes using pioneer/CMakeLists.txt instead. So this build needs updating, I guess, and the list of files need to be cleaned up possibly by someone using msvc on windows.

src/ship/PlayerShipController.cpp
Gliese852: @azieba Yeah, something like that, I just personally think it's not quite obvious from the code, and variable names are not conducive to easy understanding.

src/ship/PlayerShipController.cpp
```diff
@@ -505,7 +505,8 @@ void PlayerShipController::StaticUpdate(const float timeStep)
 		case CONTROL_MANUAL:
 			// maintain the thrust set by the station when undocking while the timeout lasts
 			// but only in manual mode
-			PollControls(timeStep, false, mouseMotion, act);
+			//Rotation damping forced for follow mode
+			PollControls(timeStep, m_followTarget && m_followMode == FOLLOW_ORI, mouseMotion, act);
```
Gliese852, line 509: It turns out strange - you click on the rotation damping button, it is activated, but does not work. Maybe it's better to block it somehow?

Gliese852:
@azieba I would like to draw attention to one more thing - the slightest movement of the mouse with the RMP pressed stops any rotation of the ship.

azieba:
> the slightest movement of the mouse with the RMB pressed stops any rotation of the ship.

Oh yes this is unfortunately a result of the design of the code. The PollControls functions does not have an access to the current rotation vector so when you input any angle request by key or even slight mouse movement even for one axis you automatically request 0 angle movement for all the other axis.
It could be also solved by using separate request flags for all the axis. Anyway I did not want to make this bugfix overly complicated, just make it back to the way it was before it was broken.

src/ship/PlayerShipController.cpp
azieba: As strange as rotation damping in autopilot mode when damping is off. The autopilot mode is different from manual follow target roughly by automatically applied linear thrust.
I guess it is a subject for improvement of the lua flight gui to grey out the rotation button in automatic modes, but do we want such thing in this PR?

Gliese852:
@azieba 
> The PollControls function does not have an access to the current rotation vector

`m_ship->GetAngVelocity()`

> We could have a normal free spin around some major axis, but as soon as we add a spin to another axis things start to get funny. 

in fact, nothing terrible happens, the rotation vectors add up perfectly with each other - just add all the vectors, and then shorten the resulting vector so that it is no longer than the maximum rotation speed (currently it is 1 radian/s, for convenience).

> So this means that in context of rotation pioneer currently treats all ships as perfect spheres with even mass distribution.

Rotation always occurs around some constant center of mass, so modeling an object in a void as a sphere seems to be quite tolerable. 

P.S. In my opinion, in the case of disabling rotation damping, it is better to simply switch the rotational part of the control to "thrust mode", although I admit that this is not the only neither the absolutely best option.

azieba:
> P.S. In my opinion, in the case of disabling rotation damping, it is better to simply switch the rotational part of the control to "thrust mode", although I admit that this is not the only neither the absolutely best option.

Yes I agree. It would be consistent with how manual control works for linear acceleration. I will try to make it that way.

Web-eWorks:
@impaktor
> I've probably asked before, but do we need the appveyor build for anything? The build-msvc does the windows build, now, by the looks of it.

Appveyor is redundant, we no longer need it as it is fully covered by the github-ci builds.

The `build-vs-solution` action uses MSBuild (AFAIK) to build a manually-maintained visual studio solution (which hasn't been updated in some months), whereas `build-msvc` uses CMake to generate a Ninja-based build toolchain that sidesteps MSBuild and a Visual Studio solution entirely (though it can also generate solutions for end-user development).

---

I'll read through the rest of the actual discussion and comment as appropriate when I have further time, but it's good to see new contributors willing to invest this much time in discussing a feature and ensuring it functions correctly - bravo @azieba!

azieba:
I made a bit of improvement. The idea is that the when rotation damping is off we should get the same behavior of thrusters as with damping, except the axes that are not affected at the moment - those should be quiet.
So before when damping was still working any change requested to rotation even on one axis was breaking the rotation on all axes. Now when player steers on one or two axes the ship will continue rotating around remaining axis.
Bits of unused code removed.
I also changed option -fdiagnostics-color=auto for gcc compilation. This is to remove terminal escape codes for coloring for compilation output for gvim - it prevents lots of garbage for me. I hope nobody minds this change. The coloring is still present for build in terminal that supports it.

src/Ship.h
```diff
@@ -343,7 +343,8 @@ class Ship : public DynamicBody {
 	bool AIMatchVel(const vector3d &vel, const vector3d &powerLimit = vector3d(1.0)) { return m_propulsion->AIMatchVel(vel, powerLimit); }
 	double AIFaceDirection(const vector3d &dir, double av = 0) { return m_propulsion->AIFaceDirection(dir, av); }
 	void SetThrusterState(int axis, double level) { return m_propulsion->SetLinThrusterState(axis, level); }
-	void AIMatchAngVelObjSpace(const vector3d &desiredAngVel, const vector3d &powerLimit = vector3d(1.0)) { return m_propulsion->AIMatchAngVelObjSpace(desiredAngVel, powerLimit); }
+	void AIMatchAngVelObjSpace(const vector3d &desiredAngVel, const vector3d &powerLimit = vector3d(1.0), bool allAxes = true) {
+		return m_propulsion->AIMatchAngVelObjSpace(desiredAngVel, powerLimit, allAxes); }
```
Gliese852, line 347: Рerhaps return is not needed here; and style nitpicking - as a rule, if the function definition does not fit on one line, curly braces are placed on separate lines

src/ship/PlayerShipController.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details
```
Gliese852, line 1: Got lost.

src/ship/PlayerShipController.cpp
```diff
@@ -30,8 +30,9 @@ struct PlayerShipController::TotalDesiredAction {
 	vector3d linPower = vector3d(1.0);
 	// true means that desiredLinear contains speed, false means thrust
 	bool desireLinVel = false;
-	// true means that desiredAngular contains speed, false means thrust
```
Gliese852, line 33: If not difficult, please correct the comment above `struct PlayerShipController::TotalDesiredAction`, about optional - that was the original idea.

src/ship/PlayerShipController.cpp
```diff
@@ -498,19 +495,22 @@ void PlayerShipController::StaticUpdate(const float timeStep)
...
 			// but only in manual mode
-			PollControls(timeStep, true, mouseMotion, act);
+			PollControls(timeStep, mouseMotion, act);
 			Util::LimitActualSpeed(*this, timeStep, act);
 			// synchronize the rotation in 'follow orient' mode
 			if (m_followTarget && m_followMode == FOLLOW_ORI) {
 				act.desiredAngular += m_followTarget->GetAngVelocity() * Pi::player->GetOrient();
 			}
+			//Follow mode as I understand is when we try to match rotation speeds with a target
+			//so all axes should be matched even with rotation damping off
```
Gliese852, line 512: I would like to add that the ship not only corrects its orientation, but also corrects its position in order to remain "stationary" relative to the turning target.
Therefore, I personally believe that in this case the rotation can be "released", but it is also possible not to do so.

src/ship/Propulsion.h
```diff
@@ -116,7 +116,7 @@ class Propulsion : public RefCounted {
 	bool AIMatchVel(const vector3d &vel, const vector3d &powerLimit = vector3d(1.0));
 	bool AIChangeVelBy(const vector3d &diffvel, const vector3d &powerLimit = vector3d(1.0)); // acts in object space
 	vector3d AIChangeVelDir(const vector3d &diffvel);										 // object space, maintain direction
-	void AIMatchAngVelObjSpace(const vector3d &angvel, const vector3d &powerLimit = vector3d(1.0));
+	void AIMatchAngVelObjSpace(const vector3d &angvel, const vector3d &powerLimit = vector3d(1.0), bool allAxes = true);
```
Gliese852, line 119: Just a thought, I personally would like a more descriptive name for this bool, like `ignoreZeroValues` or something like that.

Gliese852:
@azieba Looks good in my opinion!  I hope I can get someone to test this with a joystick. I suspect that with a smooth release of the stick, the rotation speed will decrease.

src/ship/PlayerShipController.cpp
azieba: I'm not sure about this comment above stuct definition, do you think the two lines should be removed? : 
// vector3d (0.0) means we want zero speed, for example, so I made optional
// so that the system can say that it does not want to touch this parameter at all

Web-eWorks:
I will try to build and test this PR after work this evening if I happen to have time; otherwise it may have to be on the weekend.

src/ship/PlayerShipController.cpp
azieba: If you fly out of a space station and turn on the "fallow orientation" mode you will match the rotation with the space station but the ship will be still drifting out into space and you will be in this FALLOW_ORI mode internally. With rotation dampening off I think we want the same behavior.
There is another mode FALLOW_POS that is maybe what you describe.

src/ship/PlayerShipController.cpp
azieba: my bad, I will correct it

src/ship/Propulsion.h
azieba: OK

azieba:
> I suspect that with a smooth release of the stick, the rotation speed will decrease.
Yes I think it will be the case. The question is if we would like it to work differently with rotation damping off?

src/Ship.h
azieba: OK

Web-eWorks:
Having tested this with a joystick and found an unrelated issue, the only showstopper for this change is that the default joystick deadzone of `0.0` causes the ship to continue rotating even with rotation damping enabled due to a very small deflection being reported from the joystick. That's a config / input handling change though, so it just needs to be done before this update ships.

I'm generally okay of the behavior of the ship with rotation damping off, though the rotation speed of a sinonatrix is *very* sensitive with a joystick - something that can be mitigated with a GUI way to set per-axis curves, admittedly.

src/ship/PlayerShipController.cpp
Gliese852: Yes, that will be enough

src/ship/PlayerShipController.cpp
azieba: Ok, if you do not mind then I will leave it as it is now. Rotation damping off is a minor thing in my opinion.

Web-eWorks:
> Looks like it's time for you to update your config - now deadzone is set to 10% by default.

Thanks, will do that! My config file has probably been around since 2019-ish so it's a good idea to reset it back to defaults!

Web-eWorks:
@azieba I'm generally happy with the changes made here other than that it's a bit difficult to get a ship to actually *stop* rotating completely with a joystick when rotation dampening is off - a minor amount of dampening when rotational velocity is very close to zero would be appreciated from a user experience perspective, as otherwise your velocity becomes an annoying asymptote.

@Gliese852 I will leave further sign-off and PR approval / merge in your hands.

azieba:
@Web-eWorks I suppose you tested with joystick dead zone. I bet this makes stopping rotation with dampening off annoying :)
But I would say that flying without rotation dampening is supposed to be annoying. Having the ability to turn it off serves imo one purpose only - to make players realize why rot damping is so nice to have and should not be turned off.
Furthermore there is a key to kill the rotation in your case - the one that was the origin of this bug - now it works and you can try it out, it does not kill the flight interface instead.

Web-eWorks:
> Furthermore there is a key to kill the rotation in your case - the one that was the origin of this bug - now it works and you can try it out, it does not kill the flight interface instead.

I in fact forgot that was a key that existed, thanks for reminding me! I withdraw my suggestion as that adequately covers the feature requirement.

impaktor:
I've removed AppVeyor, so we should not see it upon any new push to the PR, (I hope).

@Web-eWorks This PR is ready for merge, then?

Web-eWorks (approved):

Web-eWorks:
> This PR is ready for merge, then?

If @Gliese852 has no further comments/concerns then I'm good to merge this PR.

Gliese852:
@azieba You have several fix and merge commits, it usually doesn't get merged into master like this. Will you improve the commit history, or should I do a squash-merge?

azieba:
> Will you improve the commit history, or should I do a squash-merge?

I have squashed all my changes.

PR #5412, opened 2022-10-29
Fix animation import issues with assimp >= 5.1.0

Web-eWorks:
Assimp 5.1.0 changed the `mTicksPerSecond` field for Collada animations from `1.0` to `1000.0`, breaking some assumptions made by the Pioneer model loader.

This PR resolves that, enables support for importing GLTF files with animations, and refactors Pioneer's to-console and to-file logging to provide more information about the source of messages and to correctly handle log messages that do not contain a trailing newline (finally!).

NOTE: to avoid having to add significant amounts of extra code to ensure imports function correctly across three major versions of assimp, we now require assimp v5.0.1 or greater as a compile-time and runtime dependency. This should only be an issue on Ubuntu 18.04 LTS, where you will need to find some way to install libassimp5 as it's not present in the package repositories provided by Canonical.

NOTE: this PR changes the maximum logged level from `Verbose` to `Debug` to avoid completely spamming the log with unneeded information - at the moment, logging `Verbose` messages to file or the console requires a code change. This will be replaced in a separate PR with a command-line argument to specify the wanted log-level provided by the game.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,7 +219,7 @@ else (MSVC)
 	pkg_check_modules(SDL2 REQUIRED sdl2)
 	pkg_check_modules(SDL2_IMAGE REQUIRED SDL2_image)
 
-	pkg_check_modules(ASSIMP REQUIRED assimp>=5.0.1)
+	pkg_check_modules(ASSIMP REQUIRED assimp>=5.0)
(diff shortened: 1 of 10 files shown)


impaktor:
Nice!

maybe update COMPILING.txt, as it still says our requirement is:

    libassimp-dev >= 3.2

PR #5417, opened 2022-11-07
Fixes #5360 - lead calculation changes

### LINKED ISSUE #5360: Beam weapons show target leading indicator despite being hitscan
### Observed behaviour
Beam lasers show a target leading indicator like that of pulse lasers, which is completely wrong - since they are hitscan, one should instead aim directly at the enemy.
Moreover, the target leading indicator has aim assist. This is not only further misleading, but also makes it impossible to hit the enemy if the target leading indicator is close enough to their ship - the aim assist will steal the shots, even if you are aiming directly at the target.

### Expected behaviour
Beam lasers either shouldn't have a leading indicator at all, or, better, have it exactly align with the target.

### Steps to reproduce
Use a beam laser, such as the `1MW beam laser`, on an enemy, and compare results from shooting directly vs on the leading indicator.

### Notes
The likely cause of this bug is that beam weapons [still have](https://github.com/pioneerspacesim/pioneer/blob/master/data/libs/Equipment.lua#L331) a `speed` attribute - exactly equal to that of pulse lasers in fact, 1km/s:
```lua
laser.beamlaser_1mw = LaserType.New({
	l10n_key="BEAMLASER_1MW", price=2400, capabilities={mass=3},
	slots = {"laser_front", "laser_rear"}, laser_stats = {
		lifespan=8, speed=1000, damage=1500, rechargeTime=0.25, length=10000,
```
 On beam lasers, [that attribute is not used for shooting](https://github.com/fluffyfreak/pioneer/blob/master/src/FixedGuns.cpp#L202-L207):
 ```cpp
		if (m_gun[num].projData.beam) {
			Beam::Add(b, m_gun[num].projData, pos, b->GetVelocity(), dir);
		} else {
			const vector3d dirVel = m_gun[num].projData.speed * dir;
			Projectile::Add(b, m_gun[num].projData, pos, b->GetVelocity(), dirVel);
		}
 ```
It is however [still used for target leading](https://github.com/fluffyfreak/pioneer/blob/master/src/FixedGuns.cpp#L250-L251):
```cpp
	// calculate firing solution and relative velocity along our z axis
	double projspeed = m_gun[num].projData.speed;
```

As such, perhaps setting the `speed` of beam weapon projectiles to something appropriately high, like `299792458`, would fix it.

Pioneer version: 2022-02-03 Release (9c06c9f)
OS: Linux Mint 20.3 Cinnamon

azieba:
1. Removal of lead calculation for beam weapons. They become deadly!
2. I tried to improve the lead calculation a bit. I have a save file where I have 0 speed relative to a trader ship that is 2.3 km away and unable to hit it because the target is accelerating. Adjustment for target acceleration helps with this.
3. There was this FIXME for doubled lead angle. The problem was that the Fire function was rotating shooting direction that was already rotated giving twice the needed angle..I think. Anyway the additional transformation in the Fire function was unnecessary.

### DIFF
--- a/src/FixedGuns.cpp
+++ b/src/FixedGuns.cpp
@@ -194,12 +194,11 @@ bool FixedGuns::Fire(int num, Body *b)
 	m_temperature_stat[num] += m_gun[num].temp_heat_rate;
 	m_recharge_stat[num] = m_gun[num].recharge;
 
-	const matrix3x3d leadRot = m_shouldUseLeadCalc ? Quaterniond(m_currentLeadDir).ToMatrix3x3<double>() : matrix3x3d::Identity();
-
 	const int maxBarrels = std::min(size_t(m_gun[num].dual ? 2 : 1), m_gun[num].locs.size());
 
 	for (int iBarrel = 0; iBarrel < maxBarrels; iBarrel++) {
-		const vector3d dir = (b->GetOrient() * leadRot * vector3d(m_gun[num].locs[iBarrel].dir)).Normalized();
+		//m_currentLeadDir already points in direction to shoot - corrected with aim assist if near the crosshair
+		const vector3d dir = b->GetOrient() * (m_shouldUseLeadCalc ? m_currentLeadDir : vector3d(m_gun[num].locs[iBarrel].dir));
 		const vector3d pos = b->GetOrient() * vector3d(m_gun[num].locs[iBarrel].pos) + b->GetPosition();
 
 		if (m_gun[num].projData.beam) {
@@ -242,6 +241,7 @@ void FixedGuns::UpdateLead(float timeStep, int num, Body *ship, Body *target)
 	assert(num < GUNMOUNT_MAX);
 	const vector3d forwardVector = num == GUN_REAR ? vector3d(0, 0, 1) : vector3d(0, 0, -1);
 	m_targetLeadPos = forwardVector;
+	m_firingSolutionOk = false;
 
 	if (!target) {
 		m_currentLeadDir = forwardVector;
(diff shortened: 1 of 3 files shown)


impaktor:
Why do you need to merge with master? If your branch needs updating, due to conflict, it's better to rebase, me thinks.

azieba:
> Why do you need to merge with master? If your branch needs updating, due to conflict, it's better to rebase, me thinks.

I had branch with the same name that I deleted. When I created it again it was outdated. I'm not very proficient in git magic. I'll try to remember that I can rebase. Thanks!

impaktor:
Sometimes when I'm in that situation, I make note of the hash of the commit I want to save, and git reset --hard the branch to latest master, then git cherry-pick the commit I want, (it's still in reflog, and not lost by the git reset --hard). Although, reset --hard, should be used by caution, as anything not checked in will be lost.

src/FixedGuns.cpp
```diff
@@ -255,21 +254,66 @@ void FixedGuns::UpdateLead(float timeStep, int num, Body *ship, Body *target)
...
+			if (l < 0) {
+				//no positive solution
+				m_currentLeadDir = forwardVector;
+				return;
+			} else {
+				//t = 1/l
+				//This is an exact solution as opposed to 2 step aproximation used before
+				//It does not improve the accuracy as expected though.
+				//If the target is accelerating and is far enough then this aim assist will
+				//actually make sure that it is mpossible to hit..
```
Web-eWorks, line 298: I'd be interested in a bit more data on the perceptual or actual measured increase in accuracy between the two methods, as the comment here seems to imply both have about the same effective accuracy when considered without the acceleration term.

src/FixedGuns.cpp
```diff
@@ -194,12 +194,11 @@ bool FixedGuns::Fire(int num, Body *b)
 	m_temperature_stat[num] += m_gun[num].temp_heat_rate;
 	m_recharge_stat[num] = m_gun[num].recharge;
 
-	const matrix3x3d leadRot = m_shouldUseLeadCalc ? Quaterniond(m_currentLeadDir).ToMatrix3x3<double>() : matrix3x3d::Identity();
-
```
Web-eWorks, line 198: I believe this has been fixed in my gun-manager rewrite branch already, but the construction of the quaternion here was definitely responsible for some part of the inaccuracy - I wasn't quite up to speed on quat semantics when I wrote the lead code originally :smile:

Web-eWorks (changes requested):
Overall I like these changes, and I definitely welcome an exact solution to the lead problem rather than the approximation we had before! I've mentioned a few things that probably should be cleaned up before merge, but other than those the PR is very clean.

src/FixedGuns.cpp
azieba: I do not understand those quaternion transformations at all. Can you maybe point me to some documentation about those?

src/FixedGuns.cpp
azieba: Well I tried to solve this without approximation, lets say, for fun. Theoretically it should be better but in practice there was no noticeable difference. This is why I left the approximation method to decide during review which to choose. I do not think that the exact method is heavier computationalylly so maybe it could stay.

The adjustment for acceleration made noticeable difference. If a target flies away from you and is accelerating then you will get reliable hits. The good thing about the exact method is that it gave me the time to use for the acc adjustment.

src/FixedGuns.cpp
```diff
@@ -255,21 +254,66 @@ void FixedGuns::UpdateLead(float timeStep, int num, Body *ship, Body *target)
...
+			m_targetLeadPos = targpos;
+		} else {
+			const vector3d targvel = target->GetVelocityRelTo(ship) * ship->GetOrient();
+
+#if 0
+			vector3d leadpos = targpos + targvel * (targpos.Length() / projspeed);
+			leadpos = targpos + targvel * (leadpos.Length() / projspeed); // second order approx
+
+			m_targetLeadPos = leadpos;
+#else
```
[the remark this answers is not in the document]
azieba: done

src/FixedGuns.cpp
Web-eWorks: There's precious little documentation in the Pioneer codebase, but I highly recommend referencing [EuclideanSpace](https://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/) for a thorough understanding of quaternions at both a mathematic level and an applicational level. The issue in question was I was trying to construct a rotation quaternion from a normalized unit vector without any reference as to which axis should be pointing in that direction (nor the roll of the resulting rotation), which was not what the `Quaternion(vector3d)` constructor actually does. :smile:

src/FixedGuns.cpp
azieba: So the approximation method used sqrt also to calculate length of the distance vector twice, but I suppose if the compiler is not too dumb it would only be done once.
I'll try to get rid of those divisions then. Good to know about the difference in complexities.

src/FixedGuns.cpp
azieba: Thnx for the link!

src/FixedGuns.cpp
Web-eWorks: Hah, good eyes. I can only apologize for being very tired when I wrote that, this is indeed one `sqrt()` cheaper than the prior solution, and the current iteration of the code has the same number of divisions. I'm satisfied with the current state of the code; I don't think there's any further optimization gain here that would be worth the time investment to realize (if at all).

The overall performance cost of this code is quite small for a single iteration (I'd guess around 1-2 microseconds at most), but in practice I'd like to be able to run lead computations on AI turrets, fixed/gimballed ship weapons, etc. for at least 100 (if not closer to 1000) individual weapons per frame in the future, which requires the individual iteration cost to be quite small indeed. :smile:

Web-eWorks (approved):
Unless you have further changes you'd like to make to the code (or I'm missing some part that doesn't compile), I consider this PR to be in a good state to merge, pending some testing effort I'm unable to provide at this time.

As an aside, we (the "core" development team) generally only start a PR as a draft when it's in the work-in-progress stage and not anywhere near the final form; once a PR has matured enough that we're ready to review and consider it for merge we usually don't convert it back to draft unless there are *major* architectural or fundamental changes required that would essentially "send it back to the drawing board" and require a redesign or more effort than the author is able to contribute.

All that to say, this PR is in a very good state and doesn't need to be sent back to the draft folder at all!

Web-eWorks:
Landed ships do indeed have collision switched off, as part of an optimization effort by @Gliese852 from the last update.

Gliese852:
@Web-eWorks 
> Landed ships do indeed have collision switched off, as part of an optimization effort by @Gliese852 from the last update.

It's been off since time immemorial, I almost didn't touch that code. Maybe I fixed some bug related, but I don’t remember if it got into the master.

src/FixedGuns.cpp
azieba: > AI turrets, fixed/gimballed ship weapons, etc. for at least 100 (if not closer to 1000) individual weapons per frame in the future

Wow. Are we gonna have battleships in the game?

azieba:
I have checked. Landed ships have 0 force so no need to correct.  The switched off collisions confused me I think.

I am also wondering if some averaging of the target acceleration would be a good idea.

src/FixedGuns.cpp
Web-eWorks: Eventually stations will need self-defense weapons, and there has long been a plan for large (2km+) "bulk ships"... including military/naval variants of such. You can imagine how many guns would by design be present on such installations!

PR #5419, opened 2022-11-12
Fix life support

azieba:
Variable in C++ was different then the equivalent in lua, then the name from c++ was used in lua. To prevent such mistakes I made them the same.


Web-eWorks (approved):
Good catch, thank you!

impaktor:
Nice!
Is this reported as a bug anywhere, or you found it now?

azieba:
> Is this reported as a bug anywhere, or you found it now?

I found it myself. I tried to look for a bug about it but haven't found one. I think it is recent.

PR #4913, opened 2020-07-12
Update CityOnPlanet.cpp

joonicks:
Optimized to use bitfield instead of a byte array for occupancy (5kb vs 40kb).
City size is calculated different and slightly randomized (based on city seed).
Uses precalculated distance-from-center instead of doing the math realtime.
Largest cities on planets with population of 1Billion+.
Mars is a 1B+ planet so Cydonia is now huge. Same with New Hope.

Fixes #4802, fixes #5198.

PR #5430, opened 2022-11-27
Rewritten and Improved City Generation for Spaceports

Web-eWorks:
This PR finally finishes implementing the work that joonicks started a very long time ago. It reworks Pioneer's city generation code to be faster, support more natural and intentional building placement, massively improves the code quality and maintainability, allows many parameters to be data-driven, and fixes several major bugs with city generation.

I've included a sample default configuration, with plenty of comments in the code explaining how the algorithm works and what effect the various inputs in the configuration have on the generated result. Feel free to tweak the config further or add entirely new configuration files - when a city is generated it will pick a random configuration from the `data/configs/buildings/` directory.

The default city configuration creates cities anywhere from a few hundred meters wide on sparsely populated planets to ~30km^2 on the most populated worlds (e.g. Earth). This is done at roughly interactive framerates (generating a city causes at most a ~50ms stall), though there's future potential to run the entire generation process asynchronously on a background thread and refrain from rendering the city until it is finished.

There is still plenty of room for optimization in city rendering; that code is mostly unchanged and requires a deeper refactor into the Scenegraph instanced rendering code to achieve a performance gain. The CPU-side cost of rendering 12-20k buildings is very negligible though, coming in at sub-millisecond times on my machine. The existing city detail quality setting can also help with that.

This PR should fix issues where starting a game at New Hope, leaving to the main menu, and then starting one on Mars without closing the game would cause a completely different city layout than if you started directly on Mars.

I've also updated the logging with a new config variable to control printing verbose log messages and added a new debug tool capability to teleport the player's ship to a docked state at the targeted space station. The debug tool is available on the Ship Spawner menu right now, though a refactor of the debug tool UI in general is needed at some point.

Enough from me, have some screenshots!

Itzalean on New Hope:
image

Cydonia on Mars:
image

London, England, Earth: (ever so slightly in the wrong place)
image
image

Closes #4913.

### DIFF
--- a/src/CityOnPlanet.cpp
+++ b/src/CityOnPlanet.cpp
@@ -189,13 +189,17 @@ void CityOnPlanet::LoadCityFlavour(const FileSystem::FileInfo &file)
 		radius.density = item.value("density", 1.0);
 
 		if (!cityType.sizeDefs.empty() && radius.population > cityType.sizeDefs.back().population) {
-			Log::Warning("\tSize definition {} has higher population threshold than previous ({}b vs. {}b); ignoring",
+			Log::Warning("\tSize definition {} has higher population threshold than previous ({}b vs. {}b).\n" \
+				"Size defs should be in descending order.",
 				numSizeDefs, radius.population, cityType.sizeDefs.back().population);
-			continue;
 		}
 
 		cityType.sizeDefs.emplace_back(std::move(radius));
 	}
+
+	std::sort(cityType.sizeDefs.begin(), cityType.sizeDefs.end(), [](auto &a, auto &b) -> bool {
+		return a.population > b.population;
+	});
 
 	Log::Verbose("\tLoaded {} city size definitions", numSizeDefs);
 
@@ -350,10 +354,10 @@ void CityOnPlanet::Generate(SpaceStation *station)
 		m_body->GetPopulation(), m_citySize, m_cityRadius);
 
 
-	// ensure divisible by 8 bytes for fast 64-bit lookup
-	if (m_gridPitch & 7) {
-		m_gridPitch = (m_gridPitch + 7) & ~uint32_t(7);
-	}
+	// reserve space off the 'edge' of the grid for fast 64-bit lookup
+	// prevents reading or writing unallocated memory on grid edges
+	m_gridPitch = (m_gridPitch + 7) & ~uint32_t(7);
+
 	m_gridLen = m_gridPitch * m_citySize;
 
 	m_gridBitset.reset(new uint8_t[m_gridLen]);
--- a/src/CityOnPlanet.cpp
+++ b/src/CityOnPlanet.cpp
@@ -135,6 +135,8 @@ void CityOnPlanet::LoadBuildingType(std::string_view key, const Json &buildingDe
 		out.buildingKind = SectorKind::Monument;
 	else if (kind == "habitat")
 		out.buildingKind = SectorKind::Habitat;
+	else if (kind == "frontier")
+		out.buildingKind = SectorKind::Frontier;
 
 	std::string idleName = buildingDef.value("idle", "");
 	if (!idleName.empty()) {
@@ -487,8 +489,8 @@ void CityOnPlanet::Generate(SpaceStation *station)
 					// 1.0 at city center and 0.2 at outskirts
 					rarity *= 1.0 - distribution * 0.8;
 				} else if (buildingType->buildingKind == SectorKind::Frontier) {
-					// extremely rare at city center and 0.8 at outskirts
-					rarity *= (distribution * distribution) * 0.8;
+					// extremely rare at city center and 1.5 at outskirts
+					rarity *= (distribution * distribution) * 1.5;
 				}
 
 				// roll to see if the building spawns here or not
--- a/src/CityOnPlanet.cpp
+++ b/src/CityOnPlanet.cpp
@@ -261,17 +261,6 @@ void CityOnPlanet::Uninit()
 	s_debugMat.reset();
 }
 
-// Need a reliable way to sort the models rather than using their address in memory we use their name which should be unique.
-bool setcomp(SceneGraph::Model *mlhs, SceneGraph::Model *mrhs) { return mlhs->GetName() < mrhs->GetName(); }
-bool (*fn_pt)(SceneGraph::Model *mlhs, SceneGraph::Model *mrhs) = setcomp;
-
-struct ModelNameComparator {
-	bool operator()(const SceneGraph::Model *lhs, const SceneGraph::Model *rhs) const
-	{
-		return lhs->GetName() < rhs->GetName();
-	}
-};
-
 // FIXME: this is a horrible idea as it directly mutates models in the ModelCache
 // and forces all instances of a building to have the exact same color.
 // Color data should be supplied via an instance buffer and managed as part of a building instance.
@@ -315,13 +304,10 @@ CityOnPlanet::CityOnPlanet(Planet *planet, SpaceStation *station, const Uint32 s
 	m_body = planet->GetSystemBody();
 	m_planet = planet;
 	m_frame = planet->GetFrame();
-	m_detailLevel = Pi::detail.cities;
 
 	m_rand.seed(seed);
 
 	if (s_cityFlavours.empty()) {
-		// initialize needed variables to render nothing
-		AddStaticGeomsToCollisionSpace();
 		return; // no buildings available to generate, we already logged an error about this during startup
 	}
 
@@ -366,7 +352,7 @@ void CityOnPlanet::Generate(SpaceStation *station)
 
 	// ensure divisible by 8 bytes for fast 64-bit lookup
 	if (m_gridPitch & 7) {
-		m_gridPitch = (m_gridPitch + 7) & ~7;
+		m_gridPitch = (m_gridPitch + 7) & ~uint32_t(7);
 	}
 	m_gridLen = m_gridPitch * m_citySize;
 
@@ -381,7 +367,8 @@ void CityOnPlanet::Generate(SpaceStation *station)
 	vector3d incZ = station->GetOrient().VectorZ() * CELLSIZE;
 
 	// top-left corner of the grid is -X, -Z relative to station position at center
-	m_gridOrigin = station->GetPosition() - incX * cityExtents - incZ * cityExtents;
+	// offset origin by half-grid to ensure grid centers are aligned with the station model
+	m_gridOrigin = station->GetPosition() - incX * (cityExtents + 0.5) - incZ * (cityExtents + 0.5);
 
 	// Setup the station somewhere in the city (defaults to center for now)
 	const uint32_t stationPos[2] = {
@@ -390,7 +377,7 @@ void CityOnPlanet::Generate(SpaceStation *station)
 	};
 
 	// Reserve space for the spacestation
-	SetGridOccupancy(stationPos[0], stationPos[1], stationSize, true);
+	SetGridOccupancy(stationPos[0], stationPos[1], stationSize);
 
 	Log::Verbose("\tCityOnPlanet: Station {} placed at grid {}:{} with extents {}x{}",
 		station->GetModel()->GetName(), stationPos[0], stationPos[1], stationSize[0], stationSize[1]);
@@ -508,7 +495,7 @@ void CityOnPlanet::Generate(SpaceStation *station)
 				continue;
 			}
 
-			SetGridOccupancy(x, y, buildingType->cellSize, true);
+			SetGridOccupancy(x, y, buildingType->cellSize);
 
 			// rotate the building to face a random direction
 			const int32_t orient = m_rand.Int32(4);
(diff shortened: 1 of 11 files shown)


impaktor:
That looks way better than what we currently have, very nice!

> The CPU-side cost of rendering 12-20k buildings is very negligible though, coming in at sub-millisecond times on my machine. 

What would it take before? I guess it could only render buildings in the hundreds?

Does scaling down cities, from the options menu, for those with poor hardware still work?

Web-eWorks:
> What would it take before? I guess it could only render buildings in the hundreds?

It's still the same computational cost, but with 10,000s of instances instead of 500-1000. 1ms to draw an entire city certainly isn't bad, but there's a *lot* of work being done on the CPU that is already being done (with no effect) on the GPU - unfortunately it's the result of a poor design in the SceneGraph and I won't be touching that for a little while.

> Does scaling down cities, from the options menu, for those with poor hardware still work?

Absolutely, it reduces the number of drawn buildings which linearly scales the computational cost. I should mention the numbers I was quoting are on a roughly "stock" Ryzen 7 5800x, so I expect players to need to use a lower quality setting on older hardware.

Gliese852:
@Web-eWorks It will be very interesting to explore it! So far I can see that the lines in commit-messages are too long.

Web-eWorks:
Hah, that's something I got used to doing at work - they'll be auto-wrapped by the display software anyways (Github, Gitlab, the git client) so it's not worth it to me to manually wrap a sentence at 70-something characters.

src/CityOnPlanet.cpp
```diff
@@ -85,69 +86,179 @@ void CityOnPlanet::RemoveStaticGeomsFromCollisionSpace()
...
 }
 
 void CityOnPlanet::Uninit()
 {
-	delete[] s_buildingList.buildings;
+	s_cityFlavours.clear();
+	s_debugMat.reset();
 }
 
 // Need a reliable way to sort the models rather than using their address in memory we use their name which should be unique.
```
Gliese852, line 264: the next two lines don't seem to be used anywhere

Web-eWorks: Not only are those two functions unused but I think the ModelNameComparator struct is also unused. Thanks for catching that!

src/CityOnPlanet.cpp
```diff
@@ -347,25 +679,32 @@ void CityOnPlanet::Render(Graphics::Renderer *r, const Graphics::Frustum &frustu
...
+	uint32_t numBuildings = m_cityType->buildingTypes.size();
+
+	std::vector<uint32_t> instCount;
 	std::vector<std::vector<matrix4x4f>> transform;
-	instCount.resize(s_buildingList.numBuildings);
-	transform.resize(s_buildingList.numBuildings);
-	memset(&instCount[0], 0, sizeof(Uint32) * s_buildingList.numBuildings);
-	for (Uint32 i = 0; i < s_buildingList.numBuildings; i++) {
+
+	instCount.resize(numBuildings, 0);
```
Gliese852, line 700: This is also unused

src/CityOnPlanet.cpp
```diff
@@ -347,25 +679,32 @@ void CityOnPlanet::Render(Graphics::Renderer *r, const Graphics::Frustum &frustu
...
+	transform.resize(numBuildings);
+
+	for (uint32_t i = 0; i < numBuildings; i++) {
 		transform[i].reserve(m_buildingCounts[i]);
 	}
 
-	for (std::vector<BuildingDef>::const_iterator iter = m_enabledBuildings.begin(), itEND = m_enabledBuildings.end(); iter != itEND; ++iter) {
+	for (std::vector<BuildingInstance>::const_iterator iter = m_enabledBuildings.begin(), itEND = m_enabledBuildings.end(); iter != itEND; ++iter) {
```
Gliese852, line 707: It seems that nothing prevents us from using range-based for here, and not all that.

Web-eWorks: Indeed, I didn't (yet) do a full pass on the code to convert it all to a new style. Will do that before merge.

src/CityOnPlanet.cpp
```diff
@@ -202,123 +312,345 @@ CityOnPlanet::~CityOnPlanet()
...
-	m_buildings.clear();
-	m_buildings.reserve(DEFAULT_NUM_BUILDINGS);
+	m_rand.seed(seed);
 
-	const Aabb &aabb = station->GetAabb();
-	const matrix4x4d &m = station->GetOrient();
-	const vector3d p = station->GetPosition();
+	if (s_cityFlavours.empty()) {
+		// initialize needed variables to render nothing
+		AddStaticGeomsToCollisionSpace();
```
Gliese852, line 324: This crashes if you remove `buildings/default.json`
At least because `m_cityType` points nowhere yet
And it is not clear what variables are initialized there, if we have everything 0.

src/CityOnPlanet.cpp
```diff
@@ -202,123 +312,345 @@ CityOnPlanet::~CityOnPlanet()
 CityOnPlanet::CityOnPlanet(Planet *planet, SpaceStation *station, const Uint32 seed)
 {
 	// beware, these are not used in this function, but are used in subroutines!
+	m_body = planet->GetSystemBody();
 	m_planet = planet;
 	m_frame = planet->GetFrame();
 	m_detailLevel = Pi::detail.cities;
```
Gliese852, line 318: we will do it again in `AddStaticGeomsToCollisionSpace`

src/CityOnPlanet.cpp
```diff
@@ -202,123 +312,345 @@ CityOnPlanet::~CityOnPlanet()
...
+			continue;
+
+		m_cityRadius = radii.baseSize + (atmo * radii.atmoSize) + (m_rand.Double() * radii.randomSize);
+		m_cityDensity = radii.density * 0.8;
+		break;
+	}
+}
+
+#ifdef __BIG_ENDIAN__
+#error "CityOnPlanet does not (yet) support big-endian architectures"
```
Gliese852, line 598: bit games lead to `#ifndef`'s and incompatibilities

src/CityOnPlanet.cpp
```diff
@@ -85,69 +86,179 @@ void CityOnPlanet::RemoveStaticGeomsFromCollisionSpace()
...
+		numSizeDefs++;
+
+		radius.population = item.value("population", 0.0);
+		radius.baseSize = item.value("baseSize", 0.0);
+		radius.atmoSize = item.value("atmoSize", 0.0);
+		radius.randomSize = item.value("randomSize", 0.0);
+		radius.density = item.value("density", 1.0);
+
+		if (!cityType.sizeDefs.empty() && radius.population > cityType.sizeDefs.back().population) {
+			Log::Warning("\tSize definition {} has higher population threshold than previous ({}b vs. {}b); ignoring",
```
Gliese852, line 192: Why not just sort it later?
Оr at least write a warning in `default.json` I think.

src/CityOnPlanet.cpp
```diff
@@ -202,123 +312,345 @@ CityOnPlanet::~CityOnPlanet()
...
+	m_gridPitch = std::ceil(m_citySize / 8.0);
+
+	Log::Verbose("Generating City for spacestation {}", station->GetSystemBody()->GetName());
+	Log::Verbose("\tpopulation: {0} size {1}x{1}c (radius {2})",
+		m_body->GetPopulation(), m_citySize, m_cityRadius);
+
+
+	// ensure divisible by 8 bytes for fast 64-bit lookup
+	if (m_gridPitch & 7) {
+		m_gridPitch = (m_gridPitch + 7) & ~7;
```
Gliese852, line 369: It seems to me that 7 is not a `uint32_t` literal, and especially ~7 is worrisome.

Gliese852: Maybe even  `auto rem = m_gridPitch % 8; if (rem != 0) { m_gridPitch += 8 - rem; }`

Web-eWorks: This is the general form of a round-to-power-of-two expression in C/C++; using modulo introduces additional division operations when all that is needed is an addition and two binary and operations. I'll look at explicitly specifying the type of the literal, though.

src/CityOnPlanet.cpp
```diff
@@ -202,123 +312,345 @@ CityOnPlanet::~CityOnPlanet()
...
 	}
+	m_gridLen = m_gridPitch * m_citySize;
+
+	m_gridBitset.reset(new uint8_t[m_gridLen]);
+	std::memset(m_gridBitset.get(), 0, m_gridLen);
```
Gliese852, line 374: now I don't understand why we don't use `std::vector`

src/CityOnPlanet.cpp
```diff
@@ -202,123 +312,345 @@ CityOnPlanet::~CityOnPlanet()
...
+
+	// Setup the station model position relative to the city grid
+	// ==========================================
+
+	// get the increment vectors to multiply X/Z grid coordinates by
+	vector3d incX = station->GetOrient().VectorX() * CELLSIZE;
+	vector3d incZ = station->GetOrient().VectorZ() * CELLSIZE;
+
+	// top-left corner of the grid is -X, -Z relative to station position at center
+	m_gridOrigin = station->GetPosition() - incX * cityExtents - incZ * cityExtents;
```
Gliese852, line 384: On pure intuition, I rewrote it like this:
`m_gridOrigin = station->GetPosition() - incX * (cityExtents + 0.5) - incZ * (cityExtents + 0.5);`
before
image

after
image

src/CityOnPlanet.cpp
Web-eWorks: The console and log file are a good enough warning as far as I'm considered, though I do eventually want to capture the warning messages and display them in the menu at startup... I've added this constraint here somewhat arbitrarily during development to avoid complicating things too much, but it's much better to keep the out-of-order declaration and sort. Will address that.

src/CityOnPlanet.cpp
Web-eWorks: std::vector is a good fit for resizable arrays or piece-wise construction of contiguous lists. This is a single allocate-once slab of memory; `std::vector<bool>` is a nightmare to work with, `std::bitset` requires you to know the size of your allocation up-front, and given that we're reinterpreting the memory to enable aligned addressing anyways, just a raw memory pointer is good enough for this.

All of the access to m_gridBitset is hidden behind helper functions anyways, so a std::vector would not improve the code in any significant fashion other than RAII deletion of the memory which std::unique_ptr already does.

src/CityOnPlanet.cpp
Web-eWorks: Good catch, thank you! That subtle offset always bugged me.

src/CityOnPlanet.cpp
```diff
@@ -202,123 +312,345 @@ CityOnPlanet::~CityOnPlanet()
...
+{
+	uint32_t maskOffset = x & CityOnPlanet::CELLMASK;
+	uint64_t bitmask = uint64_t((1 << xsize) - 1) << maskOffset;
+	// note: need crossplatform bswap_64 here for big-endian compat
+	// we should be able to swap the bitmask rather than the in-memory bytes and have the same effect
+
+	return bitmask;
+}
+
+void CityOnPlanet::SetGridOccupancy(uint32_t x, uint32_t y, const uint8_t size[2], bool occupied)
```
Gliese852, line 612: `bool occupied` is not used.

src/CityOnPlanet.cpp
Web-eWorks: I'm not really a fan of the way that assignment is hidden in AddStaticGeoms, but the whole thing is not the best way to handle it anyways. I'll get rid of the duplicated assign here, it's really not necessary.

src/CityOnPlanet.cpp
Web-eWorks: I had originally intended to allow both setting and clearing of grid bits, but never actually used it. A branch inside the inner loop would be counterproductive anyways, so thanks for noticing!

src/CityOnPlanet.cpp
Web-eWorks: This should be fixed now, I hope. I'm not going to invest too much effort into it, as the case of having no building definitions at all usually means the game can't find the data directory and that causes a crash/error much earlier.

src/CityOnPlanet.cpp
```diff
@@ -202,129 +305,354 @@ CityOnPlanet::~CityOnPlanet()
...
+void CityOnPlanet::SetGridOccupancy(uint32_t x, uint32_t y, const uint8_t size[2])
+{
+	if (x + size[0] > m_citySize || y + size[1] > m_citySize)
+		return; // footprint would be off-grid, prevent writing outside the bitset
+
+	// City grid is stored as lsb-first bitset with one bit per grid cell;
+	// bit 1 of byte 1 is grid 0, while bit 8 of byte 4 is grid 31
+	// Support maximum size of 32 cells for a single building to enable fast 64-bit aligned operations
+
+	uint8_t *dataPtr = m_gridBitset.get() + (x & ~CELLMASK) / 8;
```
Gliese852, line 612: Oh, by the way, why not just `>> 3` ?

Web-eWorks: Compiler should fold it to `>> 3` anyways, and this is cleaner and easier to understand for future contributors who aren't versed in the dark art of C bitops.

PR #5435, opened 2022-12-04
Fixes #5354 - autoroute to wrong body in multiple star system.

### LINKED ISSUE #5354: Setting a course with "Auto Route" to a multiple star system will default to the "A" star.
### Observed behaviour

Even if you choose another star in a multiple system (by clicking on it on the system listing or by selecting it in the bulletin board interface), clicking on "Auto Route" will add the first star in the system as the destination.

### Expected behaviour

"Auto Route" should plan a route to the chosen star in a multiple system.

### Steps to reproduce

- Start at Mars
- Switch to sector map
- Choose "Alpha Centauri" from the map
- Click "Alpha Centauri B" from the "Binary system" submenu on the right
- Click "Auto Route"
- The last jump in the route will lead to "Alpha Centauri A"

Adding jumps one by one by clicking on the "Add Jump" button doesn't trigger this behaviour, i.e. the chosen star is added to the jump list.

20220203 Win 7

azieba:
Easy fix for minor issue.

Fixes #5354.

### DIFF
--- a/src/SectorView.cpp
+++ b/src/SectorView.cpp
@@ -1197,6 +1197,8 @@ const std::string SectorView::AutoRoute(const SystemPath &start, const SystemPat
 			outRoute.push_back(m_galaxy->GetStarSystem(nodes[u])->GetStars()[0]->GetPath());
 			u = path_prev[u];
 		}
+		//End at given body in multistar systems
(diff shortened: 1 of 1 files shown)


impaktor:
(because I can't click link when it's only in headline) #5354

Web-eWorks (approved):

PR #5436, opened 2022-12-05
Fix hostiles running away.

azieba:
Now this is some funny bug. For some time I thought that the ships spawning in the combat missions were running away from me because I bought this big gun.
The truth is that when a ship gets AI command kill, if it is further then 16 km away it executes first the AICmdFlyTo and this command completes only on specific condition that the ship arrives at a point 15 km in front of target. This condition is almost never fulfilled if the player is actively perusing the hostile ship.
Additionally when the player is closer then 15km to the hostile, it still tries to fly to this point 15 km away from the player. Effectively it looks like it is running, always facing away.
I have attached a save file with a hostile Natrix in front of me. If you load it will be running away, unless you stop and wait for it to fly to this 15km point from you and only then it will return with guns blazing.

At first I tried to refactor the autopilot code not to use this 15km point and just use a distance condition. It was fine, but I noticed some performance issue. The AI ships started to behave somehow differently and the number of them started to increase with time. Maybe they could not land or stopped crashing.. I do not know.

So instead I decided to add this distance condition in the AICmdKill and to terminate the AICmdFlyTo from there.
I also removed numerous target distance calculations. Only one remains.

If you load the save game with the fix, the hostile will immediately turn around and start shooting.

[test.gz](https://github.com/pioneerspacesim/pioneer/files/10158483/test.gz)

### DIFF
--- a/src/ShipAICmd.cpp
+++ b/src/ShipAICmd.cpp
@@ -381,11 +381,22 @@ bool AICmdKill::TimeStepUpdate()
 		return false;
 	}
 
-	if (!ProcessChild()) return false;
 	if (!m_target || m_target->IsDead()) return true;
 
+	const vector3d targpos = m_target->GetPositionRelTo(m_dBody);
+	double dist = targpos.LengthSqr();
+
+	//Autopilot leads to a point VICINITY_MIN in front of target
+	//If terget (Player) is manouvering this point might never be reached..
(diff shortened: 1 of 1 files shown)


impaktor:
Ha, that was interesing! What you describe sounds familiar, and makes sense

PR #5439, opened 2022-12-10
Fixes #5343  Missing info

### LINKED ISSUE #5343: Missing info
Two things that should **definitely** be shown:

- Maximum number of passenger cabins in the "Ship information" tab AND when buying ships;
- Individual Stations/Spaceports Tech Levels in System Overview (we had it in Elite, come on! I don't want to have to dock at every nearby station to know if they have the hardware I'm looking for!).

Keep up with your excellent work!

Ver. Win 7 20220203

azieba:
Fixes #5343 - missing ship cabin capacity info and space station tech level.

I have only realized that @Web-eWorks is already assigned to this issue when I was about to create the PR, sorry about that..

The change for the ship cabin capacity was quite easy but I need some feedback if it is ok as it is or some rephrasing is needed.

With tech level it was a bit more complicated. Tech level was calculated only when real SpaceStation body was created and is not know in SystemBody objects. SystemBody knows only seed that can be used to calculate the techLevel but it is rather complex operation including RNG.
The lua windows work somewhat surprising to me, their contents are recalculated each time frame. So recalculating tech level of selected body each frame did not seem to be a good idea. So I added a state for the info window with data to display updated only if the selected body changes.

I also added a cmake target for calling scan_enums.py with update in data/meta/Constants.lua after running it.

And one more interesting thing. SolFed homeworld path was pointing to Shanghai instead of Earth. Was it intentional? I hope I do not offend any Chinese people with this correction.

### DIFF
--- a/data/pigui/modules/system-view-ui.lua
+++ b/data/pigui/modules/system-view-ui.lua
@@ -7,6 +7,7 @@ local Event = require 'Event'
 local Lang = require 'Lang'
 local ui = require 'pigui'
 local Format = require 'Format'
+local SpaceStation = require 'SpaceStation'
 local Constants = _G.Constants
 
 local Vector2 = _G.Vector2
@@ -711,12 +712,15 @@ function Windows.objectInfo.ShouldShow()
 	return true
 end
 
-function Windows.objectInfo.Show()
+function Windows.objectInfo:Show()
 	local obj = systemView:GetSelectedObject()
-
 	local isSystemBody = obj.base == Projectable.SYSTEMBODY
 	local body = obj.ref
 
+	--FIXME there is some flickering when changing from one info to another
+	--which has different lenght. If the new one is shorter then the first header drawing
+	--seems to be too high (accodring to positioning relative to old info).
+	--Probably only durring the second drawing the position is ok. The window is anchored at bottom 
 	textIcon(getBodyIcon(obj))
 	ui.text(isSystemBody and body.name or body.label)
 	ui.spacing()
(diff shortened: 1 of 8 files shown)


impaktor:
> And one more interesting thing. SolFed homeworld path was pointing to Shanghai instead of Earth. Was it intentional? I hope I do not offend any Chinese people with this correction.

Didn't @fluffyfreak move the SolFed home world to Mars? Or that was just the starting location, now that I think of it. Anyway, I'm not sure what the "pioneer lore" says on this, as long as it is consistent (e.g. either change wiki or change code). I don't think that that part of the lore is used in any missions / flavors.

azieba:
> Didn't @fluffyfreak move the SolFed home world to Mars?

The body index was set to be 4 like maybe intended for Mars, but Mars has much higher index like 16 or something like that.

data/pigui/modules/system-view-ui.lua
```diff
@@ -730,66 +734,78 @@ function Windows.objectInfo.Show()
...
+			-- more info here.
+			-- This is what should be inserted:
+			table.insert(data, { name = luc.SHIP_TYPE, value = body:GetShipType() })
+			if player:GetEquipCountOccupied('target_scanner') > 0 or player:GetEquipCountOccupied('advanced_target_scanner') > 0 then
+				local hd = body:GetEquip("engine", 1)
+				table.insert(data, { name = luc.HYPERDRIVE, value = hd and hd:GetName() or lc.NO_HYPERDRIVE })
+				table.insert(data, { name = luc.MASS, value = Format.MassTonnes(body:GetStats().staticMass) })
+				table.insert(data, { name = luc.CARGO, value = Format.MassTonnes(body:GetStats().usedCargo) })
+			end
+		else
```
Web-eWorks, line 798: This section (computing data on a ship target) should not be cached and instead recomputed every update. The values displayed here in the future are expected to change unpredictably (e.g. velocity, signal strength, etc.) and it's easier to pay the small cost of recomputing than to try to invalidate the cached data only when changed.

Web-eWorks (changes requested):
> I have only realized that @Web-eWorks is already assigned to this issue when I was about to create the PR, sorry about that..

No worries, I'm assigned to a few different issues but rarely have the time to work on all of them. This was a good one to take over!

> The lua windows work somewhat surprising to me, their contents are recalculated each time frame. 

It's not immediately visible in the development resources we have, but pigui is an immediate-mode UI so it is indeed redrawn and recalculated each frame. It's also why there are a number of places in the code where we do a lot of manual size calculation etc. because it doesn't use a multi-pass architecture over predefined primitives to perform "automatic" layouts.

> And one more interesting thing. SolFed homeworld path was pointing to Shanghai instead of Earth. Was it intentional? I hope I do not offend any Chinese people with this correction.

The homeworld is intended to point to a planet, not a starport so this was an accident. As @impaktor mentioned, that was probably intended to be Mars and not Shanghai.

Apologies on the long delay on reviewing your other PRs, I will hopefully be able to test and merge all of them (including this one) in the next few days. I've left one change request here that I'd like to see addressed before merge, and pending testing I'm otherwise happy with the code.

data/pigui/modules/system-view-ui.lua
azieba: Yes I understand immediate-mode UI is intended for such stuff. I have made the adjustment.

azieba:
> The homeworld is intended to point to a planet, not a starport so this was an accident. As @impaktor mentioned, that was probably intended to be Mars and not Shanghai.

I have corrected it to be Mars now.

azieba:
@Web-eWorks I think I have implemented the requested change. Is there anything else?

nozmajner:
An important note… so we don't forget: @azieba  should be included in contributors.txt before the new release is out. (if not already)
How would you like to be in there? Github handle, full name, other nickname for example?

azieba:
> An important note, so we don't forget: @azieba should be included in contributors.txt before the new release is out. (if not already)

I'm already in the AUTHORS.txt, but thnx again :smiley:

Web-eWorks:
Thanks for the ping @azieba - I've had a very difficult week last week and unfortunately didn't get a chance to review your outstanding PRs. Most of your currently open PRs are ready to be merged in my eyes, but I haven't gotten a chance to clear my working tree from the Scout Mission code and test them recently.

This PR in particular I want to personally test before approving/merging, but will attempt to merge other outstanding PRs if I don't get enough time to do that tonight.

Web-eWorks (approved):

PR #5445, opened 2022-12-18
Donate to Cranks - incorporate info from factions

zonkmachine:
Improving immersion in 'Donate to Cranks' by adding some faction info as string variables (faction name and the names of the police and military forces) and some new flavours to make use of them.
I tried to make the string interpolation in just one place but failed at that so it's now done in onCreateBB and onGameStart separately.

crank1

### DIFF
--- a/data/lang/module-donatetocranks/en.json
+++ b/data/lang/module-donatetocranks/en.json
@@ -71,6 +71,54 @@
     "description": "",
     "message": "Liberate your ship! Please choose amount to donate. Happy hacking."
   },
+  "FLAVOUR_6_ADTITLE": {
+    "description": "",
+    "message": "VETERAN DAY!"
+  },
+  "FLAVOUR_6_DESC": {
+    "description": "",
+    "message": "Support the veterans of the {MILITARY}."
+  },
+  "FLAVOUR_6_MESSAGE": {
+    "description": "",
+    "message": "The {MILITARY}. You can always count on them.\nCan they count on you?"
+  },
+  "FLAVOUR_7_ADTITLE": {
+    "description": "",
+    "message": "DO THEY HAVE YOUR SUPPORT?"
+  },
+  "FLAVOUR_7_DESC": {
+    "description": "",
+    "message": "Support the veterans of the {POLICE}."
+  },
+  "FLAVOUR_7_MESSAGE": {
+    "description": "",
+    "message": "The brave men and women of the {POLICE} who is working day and night to keep us safe!"
+  },
+  "FLAVOUR_8_ADTITLE": {
+    "description": "",
+    "message": "SUPPORT OUR VETERANS!"
+  },
+  "FLAVOUR_8_DESC": {
+    "description": "",
+    "message": "Support the veterans of our armed forces."
+  },
+  "FLAVOUR_8_MESSAGE": {
+    "description": "",
+    "message": "The brave soldiers of the {MILITARY} places themselves in harm's way for you!"
+  },
+  "FLAVOUR_9_ADTITLE": {
+    "description": "",
+    "message": "END POWERTY!"
+  },
+  "FLAVOUR_9_DESC": {
+    "description": "",
+    "message": "Support the fight against powerty in the {FACTION}."
+  },
+  "FLAVOUR_9_MESSAGE": {
+    "description": "",
+    "message": "We empower families in the {FACTION} to create a better future for all."
+  },
   "SALES_PITCH": {
     "description": "",
     "message": "Your reputation would benefit by donating {cash} to our cause, but any donation would help."
--- a/data/lang/module-donatetocranks/en.json
+++ b/data/lang/module-donatetocranks/en.json
@@ -109,7 +109,7 @@
   },
   "FLAVOUR_9_ADTITLE": {
     "description": "",
-    "message": "END POWERTY!"
+    "message": "END POVERTY!"
   },
   "FLAVOUR_9_DESC": {
     "description": "",
--- a/data/lang/module-donatetocranks/en.json
+++ b/data/lang/module-donatetocranks/en.json
@@ -73,7 +73,7 @@
   },
   "FLAVOUR_6_ADTITLE": {
     "description": "",
-    "message": "VETERAN DAY!"
+    "message": "VETERANS DAY!"
   },
   "FLAVOUR_6_DESC": {
     "description": "",
--- a/data/lang/module-donatetocranks/en.json
+++ b/data/lang/module-donatetocranks/en.json
@@ -113,7 +113,7 @@
   },
   "FLAVOUR_9_DESC": {
     "description": "",
-    "message": "Support the fight against powerty in the {FACTION}."
+    "message": "Support the fight against poverty in the {FACTION}."
   },
   "FLAVOUR_9_MESSAGE": {
     "description": "",
--- a/data/lang/module-donatetocranks/en.json
+++ b/data/lang/module-donatetocranks/en.json
@@ -105,7 +105,7 @@
   },
   "FLAVOUR_8_MESSAGE": {
     "description": "",
-    "message": "The brave soldiers of the {MILITARY} places themselves in harm's way for you!"
+    "message": "The brave soldiers of the {MILITARY} place themselves in harm's way for you!"
   },
   "FLAVOUR_9_ADTITLE": {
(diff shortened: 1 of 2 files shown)


Web-eWorks:
The easiest way to handle the interpolation is to store the `stringVariables` table in the `ad` object in onCreateBB and do the interpolation using `ad.stringVariables` in both locations rather than reconstructing the `stringVariables` table. It's also best to avoid storing the final interpolated text in the `ad` object, as that's serialized to disk and increases the size of the savefile. (Technically this also applies to the `stringVariables` object, if you want to go the extra mile and reconstruct it each time the advert is placed.)

I'll need to do a further pass suggesting corrections to the grammar / structure of the ad titles later this week but overall I like the idea of this PR!

zonkmachine:
> I'll need to do a further pass suggesting corrections to the grammar / structure of the ad titles later this week but overall I like the idea of this PR!

Yes. I'm not an author, that's for sure. ;P


> The easiest way to handle the interpolation is to store the `stringVariables` table in the `ad` object in onCreateBB and do the interpolation using `ad.stringVariables` in both locations rather than reconstructing the `stringVariables` table.

OK. I think I managed to come up with something that works. I also added a {SYSTEM} variable.

Screenshot at 2022-12-18 23-41-44

data/modules/DonateToCranks/DonateToCranks.lua
```diff
@@ -106,18 +106,25 @@ local onCreateBB = function (station)
...
+	local military = Game.system.faction.militaryName
+	local police = Game.system.faction.policeName
+	local system = Game.system.name
+	local stringVariables = {SYSTEM = system, FACTION = faction, MILITARY = military, POLICE = police}
 	local ad = {
 		modifier = n == 6 and 1.5 or 1.0, -- donating to FOSS is twice as good
-		title    = flavours[n].desc,
+		title    = flavours[n].title,
+		desc     = flavours[n].desc,
 		message  = flavours[n].message,
```
Web-eWorks, line 118: It's actually easier here to store *only* `n` in the ad data and index `flavours` when placing the advertisement or generally displaying translated text to the user. :smile: The list of flavor text is not expected to change, and even when the mission is being adjusted like you're doing here it's usually only to add new flavours.

This is actually done down in onGameStart I believe; I must have simply missed cleaning it up here in onCreateBB.

zonkmachine: OK. Code updated... :whale:

Web-eWorks: Looking good now!

data/lang/module-donatetocranks/en.json
```diff
@@ -71,6 +71,54 @@
...
+    "description": "",
+    "message": "DO THEY HAVE YOUR SUPPORT?"
+  },
+  "FLAVOUR_7_DESC": {
+    "description": "",
+    "message": "Support the veterans of the {POLICE}."
+  },
+  "FLAVOUR_7_MESSAGE": {
+    "description": "",
+    "message": "Those brave men and women of the {POLICE} who has dedicated their lives to keeping {SYSTEM} safe!"
```
impaktor, line 96: I think men & women are plural, shouldn't it be "who _have_ dedicated"? (They _have_, he she it _has_).

zonkmachine: Fixed!

data/lang/module-donatetocranks/en.json
```diff
@@ -71,6 +71,54 @@
     "description": "",
     "message": "Liberate your ship! Please choose amount to donate. Happy hacking."
   },
+  "FLAVOUR_6_ADTITLE": {
+    "description": "",
+    "message": "VETERAN DAY!"
```
Web-eWorks, line 76:
```suggestion
    "message": "VETERANS DAY!"
```

data/lang/module-donatetocranks/en.json
```diff
@@ -71,6 +71,54 @@
...
+    "description": "",
+    "message": "DO THEY HAVE YOUR SUPPORT?"
+  },
+  "FLAVOUR_7_DESC": {
+    "description": "",
+    "message": "Support the veterans of the {POLICE}."
+  },
+  "FLAVOUR_7_MESSAGE": {
+    "description": "",
+    "message": "Those brave men and women of the {POLICE} who have dedicated their lives to keeping {SYSTEM} safe!"
```
Web-eWorks, line 96: This isn't a sentence as written, there are only linking verbs and no primary verb. I've added a call to action here, feel free to tweak it if you think it could be better.

```suggestion
    "message": "Those brave men and women of the {POLICE} who have dedicated their lives to keeping {SYSTEM} safe need your support!"
```

data/lang/module-donatetocranks/en.json
```diff
@@ -71,6 +71,54 @@
...
+    "description": "",
+    "message": "SUPPORT OUR VETERANS!"
+  },
+  "FLAVOUR_8_DESC": {
+    "description": "",
+    "message": "Support the veterans of our armed forces."
+  },
+  "FLAVOUR_8_MESSAGE": {
+    "description": "",
+    "message": "The brave soldiers of the {MILITARY} places themselves in harm's way for you!"
```
Web-eWorks, line 108:
```suggestion
    "message": "The brave soldiers of the {MILITARY} place themselves in harm's way for you!"
```

data/lang/module-donatetocranks/en.json
```diff
@@ -71,6 +71,54 @@
...
+  },
+  "FLAVOUR_9_ADTITLE": {
+    "description": "",
+    "message": "END POWERTY!"
```
Web-eWorks, line 112: Unless this is some reference to something I'm not familiar with, this should be "poverty".
```suggestion
    "message": "END POVERTY!"
```

data/lang/module-donatetocranks/en.json
```diff
@@ -71,6 +71,54 @@
...
+  },
+  "FLAVOUR_9_DESC": {
+    "description": "",
+    "message": "Support the fight against powerty in the {FACTION}."
```
Web-eWorks, line 116:
```suggestion
    "message": "Support the fight against poverty in the {FACTION}."
```

Web-eWorks (changes requested):
I've left suggestions with various typo fixes, you should be able to go through and accept them all + commit the changes from the web interface if you have no objections.

Other than the grammar improvements, I think this PR is good to land!

data/lang/module-donatetocranks/en.json
zonkmachine: !

zonkmachine:
> I've left suggestions with various typo fixes,

Ouch! That was worse than I expected.

> Other than the grammar improvements, I think this PR is good to land!

OK. I've been looking into more stuff in `'Donate to Cranks'` but I think I'm good for now. One caveat! I haven't actually tested this in other systems than the two available from the start menu. For instance, I don't know if `families in the {FACTION}...` works in. I'm looking into ways to hack myself to other stations. I have no intention to actually learn how to fly those things...
In any case, I think this should be tested more before merging.

zonkmachine:
`dedicated their lives to keeping {SYSTEM} safe need your support!`

{SYSTEM} is the only variable I'm worried about. I don't know if `Game.system.name` is useful always.

Web-eWorks:
`Game.system.name` will always be the translated string that's shown on the map as the system name. Given that it can be randomly generated quite often, there's very little further processing that can be done. The only thing you have to be wary of is `Game.system` is a nil value while in hyperspace.

zonkmachine:
OK. So just merge and see if it works for our explorers?

Web-eWorks (approved):
Yes, at this point if you've seen no issues after testing locally, I'm fine to merge this and sort out any further bugs if they arise.

zonkmachine:
It's a go!

zonkmachine:
@Web-eWorks  Right, it's much cleaner to use the github builtin squash merge option. I always assume it because my PR tend to start out clean and then deteriorate exponentially. (for the record. If it's me, you don't have to ask to squash a PR).

Web-eWorks:
Y'know, I didn't even bother to look at the git history before merge... too much on the brain. It's not too bad, just some extra commits in the history.

impaktor:
Ideally, these commits should have been squashed.

PR #5446, opened 2022-12-20
Scoop and Search and Rescue mission fixes

azieba:
I. Fix for bug in scoop mission. All non tradable (custom) cargo was going to illegal goods list resulting in almost all junk to scoop.

II. Various Fixes and adjustments for Search & Rescue mission:
1. Added a check to make sure that the stranded ship has enough UP thruster power to take off. Fun fact: Natrix with half of fuel tank is not able to take off from 1G planet.
2. Revised passenger cabin availability check. My previous change for this was wrong - would not compile in c++ but in lua was ok as it was never called.
3. Target ship building function revised to produce viable ships more reliably. E.g. before a ship with 0 cargo space could be produced - it couldn't be refueled as 1t of fuel goes to cargo then from cargo to fuel tank.
4. Hyperjump delay of discarded ships changed from 5 to 30 s. Now they jump at some reasonable altitude.
5. Due date adjustments. If the mission was to other planet and it was close one (all the travel accelerating/decelerating)  it was almost impossible to complete in time. On the other hand if the planet was far away then the time was ridiculously long. Also added doubling the time if the completion of the mission requires 2 way trip.
6. Increased the fuel amount requested by ships stranded on other planets to 0.5. With 0.2 of fuel tank I have observed the ship run out of fuel again before reaching station. 
7. Corrected the conShipDocked function. The end mission condition was wrong. Also this function was discarding random landing ships instead of the target ship.

I have been testing the missions with corrections for some time, but there is so many possibilities...

### DIFF
--- a/data/libs/Game.lua
+++ b/data/libs/Game.lua
@@ -6,7 +6,10 @@ local gameStartTime = 0.0
 
 Game.comms_log_lines = {}
 Game.AddCommsLogLine = function(text, sender, priority)
-	 table.insert(Game.comms_log_lines, { text=text, time=Game.time, sender=sender, priority = priority or 'normal'})
+	table.insert(Game.comms_log_lines, { text=text, time=Game.time, sender=sender, priority = priority or 'normal'})
+	if type(priority) == "number" and priority > 0 then
+		Game.SetTimeAcceleration("1x")
+	end 
 end
 
 Game.GetCommsLines = function()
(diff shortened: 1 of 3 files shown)


azieba:
And one more thing. I find it annoying that important comm messages do not stop the time acceleration, so I also included change for that. I wonder if there is some rationale for having it differently.

Web-eWorks (approved):

nozmajner:
Thanks for the fixes!
Important messages did stop time acceleration in the past I think. Might have fallen through the crack during the UI rewrite.

PR #5449, opened 2022-12-21
Fix HUD disappearing a second time

### LINKED ISSUE #5448: Missing crosshair and target square
If accepting a local search & rescue mission and clicking on "Set destination as navigation target" the crosshair and the square disappears.

        [T] pigui/modules/flight-ui/indicators.lua:121: attempt to index field 'path' (a nil value)

Current master branch

WKFO:
fixes #5448
Turns out when I added the orbital stations into the code I reordered the if conditions in a way I shouldn't have.
(Not all navTarget's have 'path's, but if the navTarget is a station, it and it's parent should both have 'path's.)

@robothauler if you have a save file or you encounter a similar situation, maybe you could try reproducing the error with this correction?

### DIFF
--- a/data/pigui/modules/flight-ui/indicators.lua
+++ b/data/pigui/modules/flight-ui/indicators.lua
@@ -118,16 +118,16 @@ local function displayNavTargetIndicator(navTarget)
 	-- checks if the nav target is a starport behind (on the other side of) the player's framebody
 	local isaPortBehindPlanet = false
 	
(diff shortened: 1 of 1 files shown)


azieba:
@WKFO If you need to generate a local search & rescue mission to test, uncomment [here](https://github.com/pioneerspacesim/pioneer/blob/ebd721998396796b2d3180f1cb828f69faf1dec3/data/modules/SearchRescue/SearchRescue.lua#L2022) and enter the game in a spaceport and accelerate the time. Flavor 2 and 3 are local rescue missions.

robothauler:
@WKFO I tested with ships and cargo containers. It works!

PR #5460, opened 2022-12-26
Add cockpit headtracking, mouse zoom and smoothing

Web-eWorks:
This PR has been a long time in the making. The original commit dates back to May 2021, and I've finally unearthed it from the annals of my git history and dusted it off.

I've added support for mouse smoothing and scroll-wheel zooming to the internal (cockpit) camera. Now if you want to see what that thing in the distance is, you no longer have to use the sidereal camera mode but can simply just zoom in on it from your cockpit. This pairs extremely well with the next feature, which is...

Headtracking! (Finally!) - I've added proper UDP headtracking support which can listen to a configurable host/port to support any headtracking source that outputs the FlightGear UDP protocol. This is primarily designed to support OpenTrack, though if you have another software that outputs the right data then you should be able to use that too.

https://youtu.be/MuGB3oN43Kc

I've also added the ability for users to customize their mouse aim sensitivity when flying - I found that the default sensitivity was *way* too high for my setup, and a value of 0.3 works really well.

None of these values have UI configuration available at this time - I don't think it'd be a large task to implement, but I'd rather change how the lua->engine config interface works than copy around another 6-8 boilerplate functions.

I'd appreciate some double-checking from those on Windows regarding the `SDL2_net` dependency - I'll add it to `pioneer-thirdparty` if needed, but I develop primarily and solely on Linux.


nozmajner:
Yupp, it won't build on windows. Tried with the cmake solution in VS

Web-eWorks:
@fluffyfreak I'd really like to ship this PR in the upcoming Pioneer Day update, but there are Windows build issues due to a new `SDL_Net` dependency - if you're available, I'd appreciate getting a built copy of that library added to the `pioneer-thirdparty` repo (the source should be in there already) to unblock this PR. Ideally no later than Sunday the 15th...

nozmajner:
Progress report:
I've managed to include the sdl_net.dll-s from their [binary release](https://github.com/libsdl-org/SDL_net/releases) and updated the pioneer/msvc-defaults.cmake with
`set(SDL2_NET_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/include)`
`set(SDL2_NET_LIBRARIES ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${MSVC_ARCH}/vs2019/SDL2_net.lib)`
as per @sturnclaw's instructions. It now runs (and is quite nice), but throws an sdl_net error which yanks you out of the game window:
image
It also logs it in [output.txt](https://github.com/pioneerspacesim/pioneer/files/10487557/output.txt)

Web-eWorks:
@nozmajner I have completely replaced the SDL_Net dependency with NanoSockets instead - the code should now compile without any external dependencies on Windows and (hopefully) prevent any errors when opening the socket now. I'd appreciate you (and anyone else reading this) giving the latest build a try and ensuring there are no platform-specific issues.

nozmajner:
Roger roger!

nozmajner:
@Web-eWorks Builds with cmake on windows, in VS and runs without a hitch.
I'm afraid now people can actually read my stupid joke in the default cockpit:
image
"Yes, this is indeed a fire extinguisher"

impaktor:
I'm all for stupid jokes!

Web-eWorks:
Going to merge this PR even though the VS solution is failing. Unless @fluffyfreak or others are able to update that solution before Pioneer day (it's mostly complaining about the Headtracker files and nanosockets source being missing from the sln) we'll be deprecating that and moving on to a purely CMake based setup.

craigo-:
On Windows here. I haven't compiled in a while but have previously used both the VS solution and CMake successfully. Makes sense to me to standardise on a single build strategy.

Keeper1st:
I just compiled using the CMake with MSVC. Successful build, though it doesn't show -x64 at the end of the file name when x64 was selected. Connecting my opentrack headtracker to FlightGear mode, nothing happens. I've never used this mode before so I'm not sure it even works, or maybe I need to install something else for it to work. I haven't tried FlightGear since before there was any headtracking support. (Of course, I've been using headtracking in joystick emulation mode with Pioneer for a couple years now, so even if this FlightGear mode support doesn't work, I'm fine going back to stick emulation.)

Web-eWorks:
> Connecting my opentrack headtracker to FlightGear mode, nothing happens. I've never used this mode before so I'm not sure it even works, or maybe I need to install something else for it to work.

Double check in opentrack that you're outputting packets to `127.0.0.1` port 4242 (the port Pioneer listens on by default). If that doesn't work, check your firewall settings and/or change the IP address in opentrack to the local network IP of your computer instead. You can also adjust the listen port used for headtracking via Pioneer's config file.

Keeper1st:
Ah, I had to use not the FlightGear mode but UDP Receiver, and change the settings to that 127.0.0.1 with port 4242. It works. Pretty smooth! I'll have to see if I can get a nice dead zone in the middle so I can have the camera straight ahead when I need it. The joystick emulation setup I've been using has a deadzone setup in Pioneer's joystick profiles. It does make looking around a little wonky as there's a pause in the middle of both X and Y axes, though. Maybe I'll just get used to no-deadzone flying and paying attention to the crosshair!

Web-eWorks:
> Ah, I had to use not the FlightGear mode but UDP Receiver, and change the settings to that 127.0.0.1 with port 4242. It works.

Ah, my mistake - the UDP output mode uses the FlightGear UDP packet format for the data format it sends, so I've just referred to it as "the FlightGear protocol".

> Pretty smooth! I'll have to see if I can get a nice dead zone in the middle so I can have the camera straight ahead when I need it. The joystick emulation setup I've been using has a deadzone setup in Pioneer's joystick profiles. It does make looking around a little wonky as there's a pause in the middle of both X and Y axes, though. Maybe I'll just get used to no-deadzone flying and paying attention to the crosshair!

If you want, you can setup a button to disable headtracking output in Opentrack (called "zero output" IIRC) which acts as a toggleable 100% deadzone for when you don't want to be looking around in the cockpit. If you're just looking for something which centers the view when looking relatively forwards... I'd honestly recommend trying to work without the deadzone; I've not had any issues with the view feeling offset in other games or Pioneer without any deadzone. That being said, you can definitely adjust your pitch/yaw output curves in Opentrack to put a big zero-output deadzone between 0 and ±5 degrees or so. :smile:

fluffyfreak:
I wonder if there's a way to use this for VR 🤔 

Not the rendering side obviously, but I had a (VERY) old branch which did that and it wasn't as tough as I'd imagined.

Trickiest part was the UI

Web-eWorks:
This code is separated out from a larger effort to replace the cockpit code with a full 6-DOF cockpit which isn't moving the ship around the player to indicate motion, as well as some nice things like clickable buttons and screens in the cockpit... when that code lands, it could be quite compatible with VR headtracking.

PR #5462, opened 2022-12-30
Coronatrix Courier (Amphiesma remodel)

nozmajner:
This is the remodeling of the Amphiesma, renamed to Coronatrix.
afronatrix_019

[The model on Sketchfab](https://sketchfab.com/3d-models/coronatrix-light-courier-43b06a832e954c4182434ced57e91015)
A small courier ship that's leaning more tovards combat then evasion, the little sister of the Sinonatrix. Has a bit better acceleration, but less deltaV. And ample space for missiles. Thanks to @sturnclaw for the ship stats and the name suggestion (sorry, I forgot to credit the commit).
The third commit removes the Amphiesma (I have plans with that name though, I want to turn it into a Sino-sized freighter)
I suspect this means it can break saves, but I'm quite sure it's far from being the only one in this cycle. :P

afronatrix_018

It has the same docking collar as the Sinonatrix, S1 sized. Eventually it will be able to have an S2 gun and an ample ammount of missiles. 

High lod has 35k tris, mid has 10k, low has 6k. Has shield and collision mesh as well.
The fuselage textures are 2048x1024, spec half sized. The engine and cockpit texture is 512x512, spec also half-size. All textures are DTX1 DDS-es, except the normals, which are kept as pngs.
No patterns yet, I wanted to put it out to prying eyest to chatch any errors I missed. I'll be doing the patterns soon.
I'm planning on making a police version from this one too.

I've also adapted the Sinonatrix cockpit for it (I built the ship this way from the get-go). This is the reason the second commit copies the Coros textures into the Sinonatrix directory as well, along with the new cockpit mesh. I'm doing it this way to avoid duplicating 20mb cockpit texture (The Coros textures are about 3mb). Which is a good opportunity to request the feature of being able to reference textures or meshes from another folder, via relative paths for example. Would be quite helpful for avoiding these kinds of redundancies.

image

I've included patterns created by @Keeper1st:
pattern4
pattern2
pattern3

I did a police version as well:
image

#### Some additional renders and screenshots:
afronatrix_019
afronatrix_020
afronatrix_021
afronatrix_022

screenshot-20221230-112312
screenshot-20221230-112352
screenshot-20221230-112403
screenshot-20221230-112502


nozmajner:
Aand I've noticed that there's a fourth commit touching .gitignore. I hope it can be removed upon merge, my git-fu is still not spectacular

Web-eWorks:
> I suspect this means it can break saves, but I'm quite sure it's far from being the only one in this cycle. :P

This patch is definitely a mandatory savebump for many reasons, so don't worry about that!

> except the normals, which are kept as pngs.

This is not technically a good practice, as lossless/uncompressed textures consume much more VRAM than even high-quality DDS encodings (even with comparable disk sizes). That being said, we don't have a good normal map texture pipeline at this time so we'll have to revisit the matter in the future when I get enough time to continue work on the renderer.

> Which is a good opportunity to request the feature of being able to reference textures or meshes from another folder, via relative paths for example. Would be quite helpful for avoiding these kinds of redundancies.

As mentioned on IRC, this was ninja'd in #5459 (merged as 3fde458cb0f6b80d348d1b0f8b55dd987fdea569) :smile:

---

On the topic of gameplay, I'm considering making the Coronatrix the default starting ship for one of our starts - Hard mode is fine with the Xylophis, but the Sinonatrix is pretty much the best multi-role ship in the "small ship" class and there's nothing really to upgrade to other than a Mola Mola or Skipjack beyond it. All other ships in the relative price range are either direct downgrades in most aspects or a complete side-grade that sacrifices most of the capacity and fuel efficiency for extra passenger cabins or missile slots.

With the Coronatrix as a starting ship, you can sidegrade to a Bowfin for slightly improved combat performance, side-grade to a Pumpkinseed for extremely punchy acceleration, or upgrade to a Wave, Sinonatrix, Mola Mola, or Skipjack depending on what kind of gameplay you want to do. In my mind, this might actually make it a better starting ship for "easy" than a Sinonatrix, as the fuel tank and associated refueling cost is also smaller, and the investment for a ship upgrade is more reasonable relative to the available mission rewards.

nozmajner:
Making it into the Sol default sounds good.
I'll update the cockpit and move everything into its proper place, thanks for the relative paths feature!

nozmajner:
I've moved around the cockpit related files to models/cockpits/opli_courier_cockpits, and updated the hull textures to use relative paths in them.
I've also changed the Sol starting ship to Coronatrix, as we talked about it on irc, that it would give more routs for up and sidegrading than the Sinonatrix, which is a quite good all-rounder already.
But I messed up my commits a bit there. Can somebody help me with that (or cherry-pick upon merge, if possible?)

nozmajner:
Alright, this should be good to go now. (Sorry for the "colorful"  history)
I've added patterns made by @Keeper1st and also a police version of the ship. (Not sure if it will be picked up by any of the factions though)

Keeper1st:
Unfortunately my internet went out right after I sent the initial pattern ideas and remained out for a few days. I've sent him some updated/improved patterns so hopefully they'll get put in instead of those initial drafts. There's still an issue with the flaps above the docking collar getting textured from a painted part of the ship's body. Hopefully that can be fixed.

Web-eWorks:
@nozmajner Branch has been rebased and git history cleaned up. I suspect (at least) patterns 1-4 are missing mipmaps, as they're quite pixelated even with MSAA on.

Unsure if you have further changes to merge in - I will be making some changes to stats (in this branch) and equipment (elsewhere) to better fit the Coronatrix; the current capacity numbers need a slight adjustment to correctly allow mounting a gun and missiles on the craft.

nozmajner:
Thanks for the fix!
I suspect I won't have the time to add in the pattern fixed right now, I'll do a separate PR for them. Thanks for those as well, @Keeper1st 
They do have mipmaps, not sure why they are pixelated.

Keeper1st:
That pixelation is being caused by the RGB layers having been changed when you converted them to DDS, so instead of overlap of each patterned area which is smoother by the Alpha, it's a sudden drop-off and you're getting colour bleed on the edges to boot. I'll save them as DDS myself and send them to you (still haven't figured out how to set up Git to do this stuff myself). I'll reduce the texture size to what you did, though.

Keeper1st:
Here are the patterns fixed. Zoom in up close! http://www.keeper1st.com/pioneer/coropats.zip

PR #5464, opened 2022-12-30
Dev forum url moved

impaktor:


impaktor:
There are two comments that link to forum posts, maybe they should be updated as well....
```
data/culture/gd.lua:-- source: http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=466
src/RandomColor.h:// Originally pointed out to me by RobN in a forum thread (http://pioneerspacesim.net/forum/viewtopic.php?f=3&t=221)
```

Web-eWorks:
Agreed, as much as possible we should rewrite forum links to point to `https://forum.pioneerspacesim.net`.

PR #5467, opened 2023-01-05
Advice

zonkmachine:
* Don't save strings in ads{}. Just save the random number that selects the flavour.
* ~~Add a new advice~~

**If something sounds too good to be true, it probably is.**
The string is one of my favourite sayings and I think it fits in rather well among the other advice. There is also a news event that 
could be tailored to this advice. News event no.19 currently makes the demand on precious metals to go down.
https://github.com/pioneerspacesim/pioneer/blob/38a0156289f51d9f898e848b1734b5bf822ed886/data/modules/NewsEventCommodity/NewsEventCommodity.lua#L101-L102
The news concerns permutation of metals, 'any solid into gold or platinum', and the result is a drop in the price of precious metal. I think it would be better to see the precious metals drop just marginally or not at all and treat the news as a baseless rumor.

```
"message": "NEWS {date}: It is raining gold in the {system} system."
"message": "A man in the {system} system ({sectorx}, {sectory}, {sectorz}) has inherited an invention
from his late uncle, allowing him to transmute any solid into gold and platinum. Economists expect a
destabilization of the economy in the system, unless the machine breaks down soon."
```

https://github.com/pioneerspacesim/pioneer/blob/master/data/lang/module-newseventcommodity/en.json#L98-L104

### DIFF
--- a/data/modules/Advice/Advice.lua
+++ b/data/modules/Advice/Advice.lua
@@ -72,8 +72,8 @@ end
 local onChat = function (form, ref, option)
 	local ad = ads[ref]
 	form:Clear()
-	form:SetTitle(ad.id)
-	form:SetMessage(ad.bodytext)
+	form:SetTitle(flavours[ad.n].ID)
+	form:SetMessage(flavours[ad.n].bodytext)
 end
 
 local onDelete = function (ref)
@@ -86,19 +86,16 @@ local onCreateBB = function (station)
 	local n = rand:Integer(1, #flavours)
 
 	local ad = {
-		id = flavours[n].ID,
-		headline = flavours[n].ID,
-		description = flavours[n].headline,
-		bodytext = flavours[n].bodytext,
-		station  = station
+		station = station,
+		n = n
 	}
 
 	-- only create one per BBS, with advice_probability
 	local ref
 	if rand:Number() < advice_probability then
 		ref = station:AddAdvert({
-			title       = ad.headline,
-			description = ad.description,
+			title       = flavours[n].ID,
+			description = flavours[n].headline,
 			icon        = "advice",
 			onChat      = onChat,
 			onDelete    = onDelete})
@@ -116,8 +113,8 @@ local onGameStart = function ()
 
 	for k,ad in pairs(loaded_data.ads or {}) do
 		local ref = ad.station:AddAdvert({
-			title       = ad.headline,
-			description = ad.description,
+			title       = flavours[ad.n].ID,
+			description = flavours[ad.n].headline,
 			icon        = "advice",
 			onChat      = onChat,
 			onDelete    = onDelete})


impaktor:
Not sure I follow. News event should make good buy or sell opportunities. If news is "it's raining gold", then that should indicate the player should travel there with empty (big!) cargo hold, and buy precious metals like crazy. 

If the advice is that this is "too good to be true", then they shouldn't go there? 

The idea with the Advice module is to be a stand in for our lack of tutorial, it's a bit of A to the Q in the FAQ, because we constantly see the same issues from players, like getting stranded in uninhabited system, or targeting the wrong star in a binary system. So with that in mind, this advice doesn't really translate to a specific action in pioneer, as I can see.

Or am I misunderstanding something?

If News event 19 is too absurdley profitable, then it can be tweaked (keep in mind, I suspect usually, the player only has room for a few tonnes of goods, or not enough money to fill the cargo hold). If it's not profitable at all, then it's a bug.

zonkmachine:
> If news is "it's raining gold", then that should indicate the player should travel there with empty (big!) cargo hold, and buy precious metals like crazy.

And my point here is that it probably isn't raining gold because the invention mentioned is not sane scientifically. Not all news is correct in the world so some could be duds in the game. Maybe that isn't necessary here in Pioneer but that's the way I see it.

 > If the advice is that this is "too good to be true", then they shouldn't go there?

Something like that. I was going to bring it up in the forum or on irc but I couldn't find the right words, so I just added it to this pull request instead.

> this advice doesn't really translate to a specific action in pioneer, as I can see.

I guess not. 

Also, the other part isn't super important either. After the donatetocranks pr where I[ refactored the code similarly ](https://pioneerwiki.com/wiki/Surviving_a_reload#Don.27t_serialize_strings)to not save strings. I proceeded to update the scripting tutorial to use strings the same way. To make sure I got the ropes of it I tested it on the Advice module and committed it when I was satisfied it worked. It's a tiny refactoring that I don't think is super important. I am happy to omit the new advice from this PR and send it back to the drawing board.

impaktor:
> `"message": "NEWS {date}: It is raining gold in the {system} system."` 

I think I meant this more figuratively, as "very lucrative opportunity". One could put quotation marks on the "raining gold" part, if that's what's confusing. 

Yeah, the refactor is good (I haven't looked at the code, but I assume it's tested & works).

Web-eWorks:
So. Missions in Pioneer are usually "played straight", because the player has no context to determine whether a mission description is purposefully lying to the player or actually providing useful information. Not only that, the news event mission has a fairly low chance of appearing and is meant to provide a reward for players who are paying attention to the bulletin boards, which means that the mission descriptions are written to explain actual, tangible modifications to the game's economic state. Breaking that pattern for one specific news event flavour simply for the sake of "but science" is not a change I see myself willing to merge.

If the mission description doesn't make any sense, you can certainly propose a replacement description. However, I don't have a problem with the current description. Yes, there's a glut of precious metals on the market that rumors attribute to an unscientific cause, but those are simply rumors - it's much more likely to be a temporary scam to drive down the price of gold so an unscrupulous buyer can acquire the metals. If the player gets there before the scam is found out, they can profit by snapping up some of the surplus on the market. If they get there too late, there's room in the future to have a follow-on BBS article revealing it was actually a scam.

zonkmachine:
OK. Reverted new advice :)

Web-eWorks:
I wasn't intending to suggest getting rid of the new advice - it was actually quite topical if one considers "gold transmutation" to be a scam in reality... :smile: At your discretion if you want to add it back in however.

zonkmachine:
I think one more round on in the lab will do it just fine. I may bump the forum with a couple of ideas where it could fit in.

Web-eWorks (approved):
Thank you for converting this module to reduce savefile size! This is one of those jobs that isn't critical enough to focus on immediately, but it's a huge load off our backs when someone steps up and takes care of it!

impaktor:
>  wasn't intending to suggest getting rid of the new advice

I think I was. An advice should be able to be interpreted to a correct action in the game. In this case, it's very vague, so as to understand it's dealing with News item 19 is almost impossible, and even so, it would lead to the opposite action, i.e. the player dismissing News item 19 as scam, and not bother going there.

@zonkmachine If you want to re-word title of news 19, to indicate it's a rumor, feel free, but then it should still somehow indicate there's a change in the market there: "rumors of transmutating machine turning lead to gold, leads to market collapse" 
(although mentioning the market is slightly too explicit, compared to the other news headlines).

zonkmachine:
> @zonkmachine If you want to re-word title of news 19, to indicate it's a rumor,

I think I just need more time to get a god grasp of the module.

zonkmachine:
I think this can be merged. I'll bump the forum for the rest. Thanks for the feedback!

PR #5472, opened 2023-01-12
Equipment Balance and tweaks

Web-eWorks:
This PR addresses three related but independent balance concerns in the game, all of which were triggered by the recent merge of the Coronatrix and its promotion to a starting ship.

1. I've rebalanced the weight of missiles and unguided rockets to be something a little bit closer to real-life. In 'reality' a 50km-range AIM-120D weighs less than 200kg (and it would land somewhere between S2 (guided missile) / S3 (smart missile) in terms of Pioneer's equipment breakdown), but for this PR I'm primarily focusing on providing weights that tally up to 1t for compatibility with the current equipment system.
- This makes it much more possible to run 10x guided missiles in a Coronatrix without completely running out of equipment space for weapons and shields.
- I've also tweaked the 'capacity consumption' of hyperdrives (by changing their weight) to provide a bit more equipment capacity in smaller ships without eating into the cargo hold too badly. These are by no means 'final values' but do better represent how you'll be able to outfit your ship next patch with the full equipment refactor, so I'm merging this for early feedback purposes.
2. I've massively improved the tracking and offensive capabilities of missiles. Before, a "smart" missile was so dumb it would detonate as soon as anything (including pulse cannon projectiles!) got within 100m of it, thinking the object was a valid target to explode against. This meant that even if a missile did manage to get to the target, it was blowing up as soon as it entered the maximum effective radius, even if it was on a direct collision course.
- Now, missiles try to only explode when they'll do the most damage possible to their target, and they'll only consider certain things to be valid targets. This applies for unguided rockets as well as guided missiles, so an unguided rocket will try to "delay" the fuze until it's closest to its impact target.
- I've also tweaked the damage output of missiles to their target to better match "Real World+" missile technologies - e.g. shaped charges or annular blast-fragmentation warheads focus the destructive potential of a missile within a smaller area than our current missile calculations allow.
- The effects of all of these changes can be seen below - the first missile launched almost completely misses the target, dealing similar damage to the maximum effective damage of a missile before this PR. The follow-up missiles however manage to pass within a few meters of the target and achieve an effective detonation:

https://user-images.githubusercontent.com/4218491/211267587-fe3be9b4-2945-44c8-aece-08e6fcea87d6.mp4

3. I've tweaked the stats of the Sinonatrix to both better reflect its internal volume and equipment space, as well as to make it a better "multi-role" ship in the intermediate space between light fighters and couriers and a full 'medium' ship like a Skipjack. In this regard, the Sinonatrix is now closer to the Mola Mola in both price and flexibility, and should serve the player better as an upgrade from one of the starting ships.
- I've not significantly adjusted the thrust output or EV, so this does reduce the Sinonatrix's straight-line speed and maximum acceleration with a full cargo + equipment load, but on the other hand it massively improves its range with a light lading due to the extra tank volume.


impaktor:
Very cool stuff with the missiles. I wonder if it's desirable / easy to have missile send to comms, the distance they exploded from: "Missile: Reached closest point to target: 103 m. Detonating."

Web-eWorks:
> Very cool stuff with the missiles. I wonder if it's desirable / easy to have missile send to comms, the distance they exploded from: "Missile: Reached closest point to target: 103 m. Detonating."

There are (commented-out) logging statements in the code which provide this information in console; unfortunately wiring this through into Lua (and determining which information is useful, given that a missile can explode within range of multiple targets) is suitably non-trivial such that I'd want a good reason to provide this information before spending the time myself to add it.

impaktor:
The counter says 10, then after first missile, the entire missile indicator disappears? 
mpv-shot0001

Web-eWorks:
That's a bug with the missile equipment display that's probably still an open issue here on Github. Won't be fixed for this release though.

Issue #5133, opened 2021-02-01
Few Translation Quirks


RacerBG:
### Observed behaviour
1. The buttons are not resizing according to the string length.
2. At some places the font size is too big and too small.
3. The mass and radius variables are showing their real names instead of the translation.
3.1. The screenshot from the game seems to have white background for no particular reason.

### Expected behaviour
The buttons should either resize or include text with smaller fonts, the text size at some places should be revised, the mass and radius variables should show their translations and the pictures should have native background colour.

![background and variables](https://user-images.githubusercontent.com/68707872/106519331-ca573a00-64e3-11eb-9cf1-aaefad0ca502.png)
![fonts size](https://user-images.githubusercontent.com/68707872/106519334-cb886700-64e3-11eb-9f24-2ef02807654f.png)
![small button](https://user-images.githubusercontent.com/68707872/106519338-cc20fd80-64e3-11eb-83e8-4512af0f186a.png)

Game: Pioneer RC 2020-12-22
OS: Windows 7 SP1 64 bit
GPU: AMD Radeon R7 250 2 GB

PS: Sorry about not giving more detailed bug report (I could do it tomorrow) and making it all in one place, it's quite late here now... :smile: Also, I hope I still have time to finish the Bulgarian translation of the game, I did my best today but real life and work are taking their toll.

Regards,
Georgi

The-EG:
It looks like it's numbers that are oddly sized...but it's not consistent either.

RacerBG:
> 
> 
> It looks like it's numbers that are oddly sized...but it's not consistent either.

This could be related to my earlier issue #4977 .

The-EG:
> This could be related to my earlier issue #4977 .

Yes, it could be. My first thought is that it may be using the fallback font for those numbers for one reason or another. Or the font we are using just doesn't do well, as you pointed out.

Web-eWorks:
So the background color is an easy issue; the screen is cleared to (0, 0, 0, 0) and we never actually render anything over it with opacity. The fix here would be to either stop saving PNGs with alpha channel or to clear the screen with full alpha coverage.

Fonts are a little harder; supporting intentionally smaller glyphs for display on certain buttons is not nearly as easy as it would sound for a number of reasons, but most notably the fact that it involves rendering separate versions of the glyphs at the smaller size to avoid scaling artifacts. Something we can look into but I wouldn't expect to get fixed soon.

[...]

Issue #4977, opened 2020-10-18
Cyrillic Font - size and style issue


RacerBG:
### **Observed behaviour**
The font style used is not very readable - the spacing looks wrong and the cyrillic letters are too big compared to the original latin letters, while at some places they are too small and at others they are bigger than their allocated space.

Here are a few screenshots (which I don't know why are showing a white background color):

![screenshot-20201018-202926](https://user-images.githubusercontent.com/68707872/96375689-212e4300-1183-11eb-9db3-753e7111ba75.png)
![screenshot-20201018-203018](https://user-images.githubusercontent.com/68707872/96375692-225f7000-1183-11eb-9406-29aab604ab86.png)

### **Expected behaviour**
The font sizes and styles for the cyrillic letters should be changed for better proportions and readability.

### **Steps to reproduce**
Use a language with cyrillic script such as Bulgarian or Russian.

**Pioneer Version:** 2020-02-03
**OS:** Windows 7 SP1 x64

Regards,
Georgi

nozmajner:
Could you suggest a decent alternative font, which is licenced open source? We are a bit in the dark when it comes to cyrillic/non-latin languages.

RacerBG:
> 
> 
> Could you suggest a decent alternative font, which is licenced open source? We are a bit in the dark when it comes to cyrillic/non-latin languages.

It really depends on the person's taste.

A few options are: DejaVu Sans, Droid Sans, Fira Sans, Montserrat, Inconsolata and others. For me, in my opinion, Inconsolata looks better than the others. Of course there are many more listed and I don't know how it will look in game.

A few links:

https://localfonts.eu/freefonts/traditional-cyrillic-free-fonts/

https://localfonts.eu/freefonts/traditional-cyrillic-free-fonts/inconsolata-lgc/

nozmajner:
Thanks! I'll check them out

PR #5475, opened 2023-01-17
Font Consistency and Other Bit-Twiddling

Web-eWorks:
This PR is technically still a bugfix!

This fixes several issues related to font sizing and consistency of font choices in the info and station screens, fixes an overlap between the ScanManager window and the cargo widget, as well as fixing an issue where the comms log did not scale with the rest of the UI, resulting in incredibly tiny comms messages on 4k screens.

I've made a few opinionated changes to UI element layout to fix minor issues and improve readability, like in the Ship Market screen.

I've generally brought the UI elements in line to use the semantic font size breakpoints I've been gradually introducing over the last year - now there are 'details', 'body', 'heading' and 'title' sizes for both Pionillium and Orbiteer, with the various screen code using those breakpoints as appropriate. Some screens use the heading size for content that really should be 'body', simply because the screen is so empty otherwise it looks incredibly out of place. (Those screens would be good candidates for future work or merging into e.g. the Station Lobby tab.)

I'll most likely push one more commit to this branch to adjust the top-level TabView widget and make a few opinionated changes there as well, but this is otherwise complete IMO.

Finally, this PR makes one balance change, which is to increase the trade-in value of ships in good repair to 65% from 50% - this should make our starts (especially the easy one) slightly less difficult to get to the point of upgrading to a better ship.

A picture or two for good measure:

image
image
image
image

And one at 1280x720 showing fonts aren't too small to read :smile:
image

### DIFF
--- a/data/pigui/libs/buttons.lua
+++ b/data/pigui/libs/buttons.lua
@@ -149,13 +149,11 @@ function ui.iconButton(icon, size, tooltip, variant, fg_color, frame_padding, ic
 
 	ui.withID(tooltip, function()
 		ui.withButtonColors(variant, function()
-			pigui.PushID(tooltip)
 			if icon_size then
 				res = pigui.ButtonImageSized(ui.get_icons_texture(size), size, icon_size, uv0, uv1, frame_padding, ui.theme.colors.transparent, fg_color)
 			else
 				res = pigui.ImageButton(ui.get_icons_texture(size), size, uv0, uv1, frame_padding, ui.theme.colors.transparent, fg_color)
 			end
-			pigui.PopID()
 		end)
 	end)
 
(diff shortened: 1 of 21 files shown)


impaktor:
Based on the screen shots, it looks fine.

RacerBG:
Looks very good so far though I would like to know if this PR fixes or improves the cyrillic characters inconsistencies which I reported earlier here #4977 and here #5133.

Web-eWorks:
This does not address font-related issues (incorrect character sizing relative to the rest of the text string, strings clipping off of buttons), but may have a positive effect on some parts of the reported issues due to using a globally smaller font. You're welcome to test the new build and check for yourself though!

Web-eWorks:
Merging as-is, my changes to the Tab View didn't wind up being an improvement sadly.

Web-eWorks:
IIRC that's always been that way on small enough resolutions, and is a side-effect of not being able to specify additional padding next to a scrollbar in ImGui.

PR #5481, opened 2023-01-20
Autopilot tweaks

### LINKED ISSUE #5497: Auto-pilot becomes very slow, need to manually go forward for it to behave normally
### Observed behaviour
The auto-pilot becomes very slow > 135 Km from the target station as well as > 110 Km from the surface (coming at an angle). The doesn't make any sense to me, so I toggle to manual controls and go forward some Kms (probably < 50) then reengage it and it behaves normally. Normally when I want to hover with my ship the thrusters work considerably more, which to me hints of a gravity abnormality. As you can see in a screenshot it took > 19 hrs to travel > 8 Mm on a planet with 0.3g with ~5% fuel.

I know the fuel is low but I've done _plenty_ of missions between 2 stations on, or orbiting, the same planet, and I don't recall ever experiencing anything close to this before.

### Expected behaviour
Coming in--much--faster at the point of the trajectory.

### Steps to reproduce
1. Load gamesave
2. Change time acceleration (auto-pilot docking already set)
3. Keep an eye on the speed for a drastic drops a good distance before the target

20220203-win

The attached file was simply renamed so it would go through the upload extension filter, there was no modification to the file content of my save-file.
screenshot-20230127-033235
screenshot-20230127-033833
screenshot-20230127-045618
screenshot-20230127-053602
[Pioneer-Yoan_Faure-260726-19hrs_8Mm_0.3g.zip](https://github.com/pioneerspacesim/pioneer/files/10518023/Pioneer-Yoan_Faure-260726-19hrs_8Mm_0.3g.zip)

azieba:
I have been trying to fix autopilot crashing into planets/stars and some cases of overshoots. This is the result of my work so far.
Those saves are two examples of situations when autopilot crashes the ship that are now avoidable:
[testPlanet.gz](https://github.com/pioneerspacesim/pioneer/files/10468907/testPlanet.gz)
[testStar.gz](https://github.com/pioneerspacesim/pioneer/files/10468908/testStar.gz)
The main change is moving the CheckSuicide function call that detects the potential crash to AICmdFlyTo instead of AICmdFlyAround. I have changed the calculations in this function as well as the recovery action taken by the autopilot.
Previously in any dangerous situation the recovery was putting rear of the ship in direction of center of the planet and try to accelerate on main thruster killing other speed component at the same time. This was really sub optimal and resulted in many 
ship crashes on autopilot.
The proposed recovery procedure is trying to bend the flght path of the ship to the horizon of the planet while breaking at the same time with the exception of a situation when the ship is below safety radius and the body gravity is greater then ship's available upward thrust. In such case the ship must accelerate to survive.
I have removed speed checks from the CheckCollision function as they are already covered by CheckSuicide.
Quite a few other minor changes to fix autopilot overshoots.

I have also included a generalization of @WKFO code for obstructed target indicators. Now it also works for ships as a nav or combat targets.

Fixes #5497

### DIFF
--- a/src/ShipAICmd.cpp
+++ b/src/ShipAICmd.cpp
@@ -731,7 +731,7 @@ static vector3d GenerateTangent(DynamicBody *dBody, FrameId targframeId, const v
 //2 - unsafe escape from effect radius
 //3 - unsafe entry to effect radius
 //4 - probable path intercept
-int CheckCollision(DynamicBody *dBody, const vector3d &pathdir, double pathdist, const vector3d &tpos, double endvel, double r)
+int CheckCollision(DynamicBody *dBody, const vector3d &pathdir, double pathdist, double tlen, double endvel, double r)
 {
 	Propulsion *prop = dBody->GetComponent<Propulsion>();
 	if (!prop) // This body doesn't have any propulsion to avoid collision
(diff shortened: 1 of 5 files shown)


fluffyfreak:
For testing if something is over the horizon we also have a [HorizonCulling method](https://github.com/pioneerspacesim/pioneer/blob/master/src/Sphere.cpp#L20).

Code looks good to me 👍

src/ShipAICmd.cpp
```diff
@@ -992,9 +1035,47 @@ bool AICmdFlyTo::TimeStepUpdate()
...
 	vector3d relpos = targpos - m_dBody->GetPosition();
+	double targdist = relpos.Length();
+
+	Body* planetNeer = Frame::GetFrame(m_dBody->GetFrame())->GetBody();
+	double safeAlt = MaxEffectRad(planetNeer, m_prop);
+	vector3d obspos = -m_dBody->GetPosition();
+
+	double targetAlt = targpos.Length();
+
+	if ((m_suicideRecovery = CheckSuicide(m_dBody, obspos, planetNeer->GetMass(), safeAlt, targetAlt, m_suicideRecovery))) {
```
Web-eWorks, line 1046: `planetNeer->GetMass()`

This causes an assertion failure due to the Body for a spaceport's frame being the spaceport itself and not the planet. You will need to check `planetNeer->IsType(ObjectType::TERRAINBODY)` before calling GetMass().

(Also, `planetNeer` should probably be `planetNear`.)

Web-eWorks (changes requested):
Other than the mentioned assertion failure, I'm quite happy with this PR! I haven't been able to properly stress-test it, but I haven't seen any obvious failures yet.

From an administrative standpoint, @azieba do you have a presence on IRC for communication purposes? It's not required, but we have been discussing some planned improvements to the autopilot's capabilities to support manual docking and planned maneuver execution if you are interested in chiming in and/or contributing further.

If not, we can discuss here via a tracking issue or on the forums if you are interested in helping to implement this functionality.

azieba:
> From an administrative standpoint, @azieba do you have a presence on IRC for communication purposes?

Which is the main dev communication channel the IRC or the forum?

Web-eWorks:
> Which is the main dev communication channel the IRC or the forum?

IRC is the "primary" developer communication channel (the core developers are all on it and it stays fairly active, and it's where we generally bounce ideas back and forth). The forums are used for more specific long-form discussion to discuss a specific project or concern, for user feature requests, or when we need to write something down for future reference much later.

When planning we generally communicate on IRC and only cross-post the "plan" to the forum if it's significantly large and multiple people have volunteered to help out (e.g. the UI rewrite from many years ago).

The Pioneer IRC channel is accessible at `#pioneer` on Libera.chat via an IRC client, or via a Matrix bridge at `#pioneer:libera.chat`. There are no expectations of your involvement level if you join the IRC, but we'd be happy to have you as part of the discussion!

src/ShipAICmd.cpp
azieba: Corrected

Web-eWorks (approved):

PR #5469, opened 2023-01-09
Planet descriptions

zonkmachine:
Fixes some of the issues described in https://github.com/pioneerspacesim/pioneer/issues/4215

* Ensure the first letter is a capital one. 4215 example:
`"rocky planet with a Argon atmosphere."`
`"Rocky planet with a argon atmosphere."`

* Ensure words that aren't the first one in a sentence doesn't begin with a capital letter.
* Ensure there's only one space between words.
`"Tiny rocky planet with a  Methane atmosphere."`
`"Tiny rocky planet with a methane atmosphere."`

* Fix a/an issue by simply dropping 'a' from 'with a'.
`"Rocky planet with a argon atmosphere."`
`"Rocky planet with argon atmosphere."`

* Make moons moons
`"Titan: Tiny ice planet with tenuous oxygen atmosphere."`
`"Titan: Tiny ice moon with tenuous oxygen atmosphere."`

* Name asteroids around planets moons
`"Amalthea: Asteroid"`
`"Amalthea: Tiny rocky moon"`

europapioneer

PR #5482, opened 2023-01-21
Planet descriptions 2. Dwarf planets, asteroids and atmosphere density tweaks

zonkmachine:
*Continuation from https://github.com/pioneerspacesim/pioneer/pull/5469*
 * Asteroids labelled correctly

 * Dwarf planets shown. 
	Asteroids < Mass of Salacia, at ~0,7% Lunar mass.
	Mass of Salacia < Dwarf planet < Lunar mass.
	Planets > Lunar mass.
.
Yes, this is more spaghetti coding. I'm not ready to take on moving the planet description code into Lua yet. I wanted to see what the Dwarf planets and asteroids would look like in the game, so here we are.

 * Add some ice cover info for known solar system objects.

 * Tweak limit between THICK and VERY_DENSE atmosphere in order
   for Venus and Titan to not have the same label.

asteroid
dwarfplanet2
dwarficewithno
titanatmospere


zonkmachine:
OK. Missing strings fixed. Code and examples updated.

Web-eWorks:
> Asteroids < Mass of Salacia, at ~0,7% Lunar mass.
Mass of Salacia < Dwarf planet < Lunar mass.
Planets > Lunar mass.

Will review further in the afternoon, but wanted to make sure you take something into account - an asteroid is specifically handled as a different BodyType, with moons / dwarf planets falling into an ill-defined category between `ASTEROID`/`PLANET_TERRESTRIAL`.  The breakpoint between an asteroid or terrestrial planet is defined in [StarSystemGenerator.cpp](https://github.com/pioneerspacesim/pioneer/blob/5c7eb62fbf32135f61e3ca7ad44c5a7541394ba0/src/galaxy/StarSystemGenerator.cpp#L742), though beware custom systems can set the body type directly and may not be homogeneous with the procedural generation.

I'm not sure it's required to adjust anything in that file, but it may cause confusion to have the UI reporting an asteroid as a dwarf planet or vice versa, given that there is some existing code (most notably, body icons) specialized around whether a body's type is asteroid or otherwise.

zonkmachine:
> The breakpoint between an asteroid or terrestrial planet is defined in [StarSystemGenerator.cpp](https://github.com/pioneerspacesim/pioneer/blob/5c7eb62fbf32135f61e3ca7ad44c5a7541394ba0/src/galaxy/StarSystemGenerator.cpp#L742),

https://github.com/pioneerspacesim/pioneer/blob/5c7eb62fbf32135f61e3ca7ad44c5a7541394ba0/src/galaxy/StarSystemGenerator.cpp#L742

And I set it again here: https://github.com/pioneerspacesim/pioneer/blob/4620d96f1eebc0ccbbae7d2807ea8e0b63bd2665/src/galaxy/SystemBody.cpp#L245

I guess the values could be put in a variable.

zonkmachine:
I did some more precise computations on the Earth / Salacia mass ratio. The reason Salacia is used is, that it's the biggest object out there who hasn't yet made it to the list of Dwarf planets. If we don't count the margin of error, Salacia is 1/12134 the size of Earth. Round that to some near large number and we get:

```cpp
else if (m_mass > fixed(1, 12000))
```

src/LangStrings.inc.h
```diff
@@ -236,6 +236,12 @@ DECLARE_STRING(PLANET_WITH_SOME_ICE)
 DECLARE_STRING(MOON_WITH_SOME_ICE)
 DECLARE_STRING(ROCKY_PLANET_CONTAINING_COME_LIQUIDS)
 DECLARE_STRING(ROCKY_MOON_CONTAINING_SOME_LIQUIDS)
+DECLARE_STRING(DWARF_PLANET_ICE)
+DECLARE_STRING(DWARF_PLANET_WATER)
+DECLARE_STRING(DWARF_PLANET_CONTAINING_LIQUID_WATER)
+DECLARE_STRING(DWARF_PLANET_WITH_SOME_ICE)
+DECLARE_STRING(DWARF_PLANET_TERRESTRIAL_CONTAINING_SOME_LIQUIDS)
+DECLARE_STRING(DWARF_PLANET_TERRESTRIAL)
```
Web-eWorks, line 244: I don't see JSON language strings defined for `DWARF_PLANET_WATER`, `DWARF_PLANET_CONTAINING_LIQUID_WATER`, `DWARF_PLANET_WITH_SOME_ICE`. Left-overs, or partially-implemented?

data/lang/core/en.json
```diff
@@ -599,10 +599,22 @@
     "description": "",
     "message": "hydrogen atmosphere"
   },
+  "DWARF_PLANET_ICE": {
+    "description": "",
+    "message": "Dwarf planet, ice"
```
Web-eWorks, line 604: This is slightly awkward. Perhaps a better label might be:
```suggestion
    "message": "Dwarf planet, mostly covered in ice"
```

Web-eWorks (changes requested):
I'm seeing a bit of weirdness here but otherwise I'm liking this PR's direction. Technically we're in "feature freeze" right now so I shouldn't be merging this for the upcoming release, but I'll make an exception. :wink:

data/lang/core/en.json
zonkmachine: Yeah, that doesn't look good. I was trying to get the more specific terms, 'Terrestrial Dwarf' and 'Ice Dwarf' into the game. Only, I didn't think they fit as a first string and also, I guessed that they aren't commonly known. Looking into it but your suggestion is good.

src/LangStrings.inc.h
zonkmachine: I think those got lost when I rebased the branch ahead of the push. I'll look into it.

zonkmachine:
> I'm not sure it's required to adjust anything in that file, but it may cause confusion to have the UI reporting an asteroid as a dwarf planet or vice versa, given that there is some existing code (most notably, body icons) specialized around whether a body's type is asteroid or otherwise.

Here's and example. The code that selects the icon in front of the body name, here pluto with a planet symbol, tests for the radius of the body. All the other cases I've worked with has used mass. 

dwarficewithno

src/LangStrings.inc.h
zonkmachine: `DWARF_PLANET_WATER` -> Removed
`DWARF_PLANET_CONTAINING_LIQUID_WATER` -> Implemented
`DWARF_PLANET_WITH_SOME_ICE` -> On hold for now. I think there are enough new strings.

zonkmachine: I also sorted the strings so they are more easy to compare.

zonkmachine:
> though beware custom systems can set the body type directly and may not be homogeneous with the procedural generation.

Heh, I just spent the last hour hunting down an odd bug where planet 'Death' in Barnard's star system would show up as a planet even though it's a wee asteroid by it's mass. :clown_face:

Web-eWorks (approved):
Looks good to me, are there any further changes you have outstanding to make?

zonkmachine:
No other changes planned. I just went over the code and there is only one thing I'm wondering. The string 'Ice dwarf planet'

  "ICE_DWARF_PLANET": {
    "description": "",
    "message": "Ice dwarf planet"

I kind of understand why whoever wrote the strings for the earlier 'planet', used the term 'ice world'. 'Tiny ice world' doesn't sound good to me. 'Tiny' is used much already and I'd like the word 'Dwarf' in there.

zonkmachine:
@Web-eWorks If you're fine with that last string, you can go ahead and commit it. Otherwise I'll look into it further tomorrow.

zonkmachine:
I'm happy with this now. Merge/Squash at will. :rocket:

PR #5495, opened 2023-01-27
Fix uninitialized variables, memory leaks

Web-eWorks:
I've run AddressSanitizer, LeakSanitizer, and ThreadSanitizer on the code (as well as a brief pass with Valgrind) to clean up any possible issues or errors that might have been introduced without notice.

This PR fixes, in total:
- 1 read-past-end bug (in DistanceFieldFont.cpp)
- 2 uninitialized variable accesses (in scenegraph/Model.cpp and ship/ShipViewController.cpp)
- 3 memory leaks, of varying severity between a few hundred bytes and 600kb

### DIFF
--- a/src/FileSystem.h
+++ b/src/FileSystem.h
@@ -168,6 +168,7 @@ namespace FileSystem {
 		}
 		StringRange AsStringRange() const { return StringRange(m_data, m_size); }
 		ByteRange AsByteRange() const { return ByteRange(m_data, m_size); }
+		std::string_view AsStringView() const { return std::string_view(m_data, m_size); }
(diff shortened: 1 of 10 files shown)


impaktor:
Could any of these explain some of the reported bugs on Lua side?
(Just checking if I should add a sub-routine to my brain when hunting reported bugs that says "this might have been fixed in master")

Web-eWorks:
Very unlikely - these are all in C++. The only thing that even comes close to Lua memory was the LuaRef bug, and that was entirely benign other than leaking unused memory.

Issue #5493, opened 2023-01-26
Scan manager does not detect installed sensors


antonspace:
OS = Linux mint

Version = 20230125, latest code, compiled from source.

Scan manager states "no available sensor" even when all 4 scanning sensors are installed in ship (coronatrix), attempted mapping mission, nothing happened.

Deneb also allowed installation of all 4 scanning sensors (bug?) with same result (no available sensor)

Skipjack only allows one scanning sensor to be installed, same result (no available sensor).

Issue #5479, opened 2023-01-18
disturbed docking sequence event


Bodasey:
### Observed behaviour

![Docksequenz](https://user-images.githubusercontent.com/60466586/213206485-328e2aa8-69c3-402e-8a53-bb0d2a6c5297.png)

and this can reproduce closed #4127

[_autosave4.txt](https://github.com/pioneerspacesim/pioneer/files/10447813/_autosave4.txt)

The first entry in "Missions" in this autosave game is one of the two not cleared ones.

### Steps to reproduce

I have no idea why or when this happens
what's this? parse_error.112

[output.txt](https://github.com/pioneerspacesim/pioneer/files/10447948/output.txt)

__My pioneer version (and OS):__
Web-eWorks station economy branch / long time observation

Web-eWorks:
> what's this? parse_error.112

This is a benign error thrown when attempting to interpret a gzipped TAR file as a Pioneer savefile.

> I have no idea why or when this happens

This occurs because the order of evaluation of mission event handlers is non-deterministic at runtime. Thus, one mission completes before the event that triggers autosaves when docking, while the other two complete after an autosave.

There is a potential fix for this that may make it into the upcoming release.

Issue #4123, opened 2017-08-30
Autosave bugs: not properly landed at city port (ship damaging)


Fixer-007:
I've landed without any problems on planet city Cooke (System Ceenex), autosave did its job, yet when I quickly loaded it again - ship appeared sitting on landing pad, yet pressing F4 reveals docking menu and unpausing game causes crashing/tumbling of the ship on landing pad.

This is frequent problem on almost any autosave on any city located on any planets. Very annoying.

Faulty save: [_autosave4.zip](https://github.com/pioneerspacesim/pioneer/files/1267446/_autosave4.zip)

Pioneer 20170827 / win7

Fixer-007:
This is frequent problem on almost any autosave on planet.

Fixer-007:
From what I understand, this situation happens when you undock from city (blast off), almost immediately game gets autosave. Since ship during this time has almost no upward velocity, this causes ship to fall down on landing pad causing tumbling, crashing, etc.

One way to solve this is to add slight delay (1-2 seconds?) before autosave to let ship take some upward speed during blast off.

richardpl:
I can not reproduce this, can you still reproduce this?

Fixer-007:
Will recheck in half a year when will decide what gpu to buy.

fluffyfreak:
@Fixer-007 I've got a GeForce GT 730 SILENT 1024MB lying around you could have for the cost of postage.

Not great, but it'll run Pioneer no problem

Fixer-007:
Nah, I don't want to deal with all that international shipping and customs, so be patient :)

Fixer-007:
Will recheck on Intel HD 2500 ;)

Bodasey:
confirmed with actual master; it's not possible to undock again, undock jump is lost, setting a new target brings the ship back to the game, but the first few seconds is tremor, not move. In space stations, the ship hovers (and luckily leaves out backwards, but may collide?).

Do we really need two saves when visiting a space-/starport?

Web-eWorks:
CC @Gliese852 since you've done/are doing work on space stations and docking code.

PR #5498, opened 2023-01-29
Fixes, bowfin update, terrain rendering fixes

### LINKED ISSUE #5420: User made flight log does not save on closing the game
### Observed behaviour
Writing an entry into the logbook thing under personal info is not preserved when the game exits and does not show up when the game is reloaded

### Expected behaviour
Log is preserved on exit and viewable when reloading the save in a new session

### Steps to reproduce
Open game, start a save, write a diary/logbook entry, save game, close game, reload game

Version: February 2022-02-03
OS: Windows 10 64 bit

### LINKED ISSUE #5458: Wrong wheel up/down icon when in auto-pilot
### Observed behavior
Clicking wheel down when in auto-pilot plays the sound, but icon flickers, then back to what it was.

### Expected behavior
Either:
- clicking wheel button in auto-pilot does nothing, no sound, just "dead", since ship is under AI-control
- if wheels are down, then show correct icon

__version__ I expect this to be reproducible on master 2022-12.

### LINKED ISSUE #5470: Autosave clears bulletin board (for permanent?)
### Observed behaviour / Steps to reproduce

Land on a Star-/Spaceport when autosave is on
End the game
Load the game autosaved when you landed
Check the bulletin board - it's empty!

Even when you return to the port some time in the game later, there are no new entries.

actual master on SuSE Linux 15.4

Web-eWorks:
Unsure how much more bugfixing will be finished before Pioneer Day, so have an early batch. I'll try to get to the other reported issues with the Scout Mission etc.

- Fixes #5420: the global version attribute was being reset on game end, causing invalid data to be saved. The entire flight history log was discarded on load.
- Fixes #5458: added a separate button state for "autopilot control". Not using a separate language string due to time reasons.
- Fixes: #5479: this was caused by order of autosave handler relative to other `onShipDocked` handlers. I've deferred autosave by one frame to ensure all handlers run before saving.
- Fixes: #4123: autosave is now made before undocking, rather than after.
- Fixes #5470: this was caused by the autosave running before BBS / markets are generated for the station. Thus, reloading the autosave yields an empty market.
- Fixes: #5493: reloading the game or buying a new ship didn't update the equipment listener used to detect changes to the ship's loadout.

I've updated the Bowfin to have an additional 50% more equipment capacity as an interim measure before moving to volume-based equipment. This allows it to mount missiles in addition to a shield generator... :upside_down_face: The Bowfin's class has been updated to "Light Fighter" from "Medium Fighter" (as it in no way qualifies for a medium ship).

I've also fixed an unreported bug that caused extremely poor quality terrain in elevated areas. The centroid used to calculate LOD distance to a geopatch was remaining at the "sea level" height of the planet, rather than at the average height of the terrain, causing the game to think the camera was much further from the terrain than it actually is. This makes a *huge* difference in visual terrain quality, but leaves the performance cost about the same as before when the player is in low-lying terrain.

Before:
image

After:
image

### DIFF
--- a/data/libs/Economy.lua
+++ b/data/libs/Economy.lua
@@ -421,7 +421,7 @@ Serializer:Register("Economy",
 		}
 	end,
 	function(data)
-		stationMarket = data.market
+		stationMarket = data.market or {}
 	end
 )
 
--- a/data/libs/FlightLog.lua
+++ b/data/libs/FlightLog.lua
@@ -22,9 +22,6 @@ local FlightLogSystem = {}
 local FlightLogStation = {}
 local FlightLogCustom = {}
 
--- Version for backwards compability
-local version = 1
-
 local FlightLog
 FlightLog = {
 
@@ -379,24 +376,6 @@ FlightLog = {
 			{path, Game.time, Game.player:GetMoney(), location, text})
 	end,
 
---
--- Group: Attributes
---
-
---
--- Attribute: version
---
--- Marks version of flight log, can be used for non-breaking backwards
--- compatibility, if new features are added to FlightLog
---
--- Example:
---
--- > if FlightLog.version > 2 then
--- >     FlightLog.useNewFeature()
--- > end
---
-
-	version = version,
 }
 
 -- LOGGING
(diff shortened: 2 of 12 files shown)


Bodasey:
Maybe you can also easily fix #4123 this way?

Web-eWorks:
That in fact should already be fixed - autosave on takeoff is made before the ship is actually undocked (so if you forget something, you can quickload and buy/sell your commodities).

Bodasey:
#4123 fix confirmed

#5470 fix confirmed

#5479 very probably fixed

Bodasey:
While flying from Earth to Jupiter:

        [T] pigui/modules/info-view/03-econ-trade.lua:71: Cannot call tableSetupColumn() outside of a beginTable() / endTable() pair

Hang with a long LUA error report, said it was written to output.txt, but there's nothing of this.

Web-eWorks:
> Hang with a long LUA error report, said it was written to output.txt, but there's nothing of this.

Without the output.txt I cannot address this issue. If it occurs again, please get the error info from `output.txt` and make a new issue with it (as this is unrelated to this PR).

Issue #3583, opened 2016-01-07
Remove strings in LangStrings.inc.h


walterar:
These strings must be eliminated of LangStrings.inc.h:

-DECLARE_STRING(THREE_KPC_ARM)
-DECLARE_STRING(NORMA_ARM)
-DECLARE_STRING(PERSEUS_ARM)
-DECLARE_STRING(OUTER_ARM)
-DECLARE_STRING(SAGITTARIUS_ARM)
-DECLARE_STRING(SCUTUM_CENTAURUS_ARM)
-DECLARE_STRING(LOCAL_ARM)

PR #5514, opened 2023-01-31
Remove dead links in ui-core

zonkmachine:
Removing dead strings in ui-core.

These strings exists only in the lang files and most of them also don't have any reference to them on searching the Pioneer GitHub site. Some exempts below:

Reference to the last action involving the "Arms of the Milky Way " strings: https://github.com/pioneerspacesim/pioneer/issues/3583

`COMPACT_RADAR` removed here: 04195598386c873a9468a668d3fe0a8b44f7d84f

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -1263,6 +1263,14 @@
     "description": "",
     "message": "Lobby"
   },
+  "LOCAL_ARM": {
+    "description": "Arm of the Milky Way galaxy",
+    "message": "Local arm"
+  },
+  "LOCATED_N_KM_FROM_THE_CENTRE_OF_NAME": {
+    "description": "",
+    "message": "Located {distance}km from the centre of {name}:"
+  },
   "LOCATION": {
     "description": "",
     "message": "Location"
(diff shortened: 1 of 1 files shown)


impaktor:
I see strings from our old econ info / trade analyzer screen, I would have expected those to be used for mouse over tool tip with the new, but seems we have other strings for that.

I also see strings from our old "F8" Galaxy view mode. Are we definitely not going to bring that back, ever?

Web-eWorks:
> Are we definitely not going to bring that back, ever?

Those are the only ones I'm hesitant to delete, as providing a galaxy view when zoomed out far enough in the sector map has been a low-priority planned work item for some time. Unsure if they should remain in ui-core however, or be moved to their own resource or even the core resource.

data/lang/ui-core/en.json
```diff
@@ -1279,14 +1263,6 @@
     "description": "",
     "message": "Lobby"
   },
-  "LOCAL_ARM": {
-    "description": "Arm of the Milky Way galaxy",
-    "message": "Local arm"
-  },
-  "LOCATED_N_KM_FROM_THE_CENTRE_OF_NAME": {
```
impaktor, line 1286: Keep this, and all strings mentioning ARM

zonkmachine: OK

zonkmachine: Fixed!

impaktor:
Feel free to squash those commits & force push

PR #5515, opened 2023-01-31
Fix small errors, adjust orbitscanner balance

### LINKED ISSUE #5508: Mapping mission that Orbital Scanner XKM-1750 is unable to resolve.
### Observed behaviour

Accepted a mission to perform an orbital scan of Gikegaryu F1, flew to vicinity of Gikegaryu F1, then used autopilot to get into a high orbit around the planet. Engaged Orbital Scanner XKM-1750 when the current frame of reference was Gikegaryu F1, scanning started briefly, but as the autopilot neared high orbit, the scanner no longer worked because the "minimum altitude" had been reached.

### Expected behaviour

One would expect the premium scanner would be up for the job, I worked my ass off shuttling green haired humanoids around the galaxy in order to buy this premium piece of kit, where do I get a refund? ;-)

### Steps to reproduce

Load the savefile, you are already in the vicinity of Gikegaryu F1.
[sinonatrix_15.gz](https://github.com/pioneerspacesim/pioneer/files/10539583/sinonatrix_15.gz)
(the attached file was renamed by adding .gz to allow upload).
Use the autopilot to enter a high orbit of the planet.
When close enough to engage the scanner, engage it, making sure the Orbital Scanner XKM-1750 is selected, it should work briefly.
As the ship approaches close to high orbit altitude, the scanner will reach "minimum altitude" and stop working.
The Orbital Scanner XKM-650 does work in this high orbit.

20230129
Linux Mint.
Screenshot_2023-01-30_19-28-42
Screenshot_2023-01-30_19-11-56

### LINKED ISSUE #5511: Launch request denied - missing crew
### Observed behavior
Starting ship is denied take off, due to missing crew - but there' is enough crew according to ship personnel list. Noteworthy: the crew salary is $0. This is a bit of a bummer, since it will hit new players within minutes of trying the game, unless using the "hard" (illegal) take off button.

### Steps to reproduce
- start new game from mars
- F4 for lobby
-  click "request launch" --> will be denied (maybe not always? happened 3/3 times for me now)

### Steps to NOT get the bug:
- start new game from mars
- F3, then check Crew Roster
- F4 for lobby
-  click "request launch" --> works

Should we even have a starting ship that requires crew? If crew has a salary, you have running cost, that a new player might not know how to generate profit fast enough to offset.

__My pioneer version (and OS):__ latest master, e765fd8cfa499ec766a69bec7f13c1d33c5907a8

### LINKED ISSUE #5512: Time acceleration in Bulletin Board leads to negative mission time and possible error.
### Observed behaviour

Using the time acceleration feature in the Bulletin board will result in the mission times running down to zero, then into negative territory. Clicking on another tab, or a mission that still has a positive time will clear out the missions with a negative  time, but clicking on a mission with a negative time will cause an error.

### Expected behaviour

The expected behaviour would be that the missions on the bulletin board would be cleared when the time remaining is less than is practically possible to complete the mission. without having to leave the tab, or take some other action to refresh the page.

### Steps to reproduce

Use time acceleration while remaining in the bulletin board.

20230131
Linux 
Screenshot_2023-01-31_19-22-45
Screenshot_2023-01-31_19-24-05

### LINKED ISSUE #5521: Skipjack crippled by lack of sensor slots.
### Observed behaviour

After upgrading to the Skipjack, in my opinion the best all round ship in the game, it is disappointing to find that this ship can only install either one orbital scanner or one surface scanner which makes it nearly unusable for these type of missions, if an XKM-1750 is already installed and one wishes to undertake a surface scan mission, the loss of $2,200 trade in value almost negates any financial benefit from doing these types of missions. Even the lowly Xylophis is able to install both surface scanners and an orbital scanner. 

### Expected behaviour

One would expect such a fine ship would be on a par with other ships in this regard.

### Steps to reproduce

Buy a Skipjack, give it the best attachments money can buy, then weep as you realise it won't even allow an orbital scanner and surface scanner to be installed at the same time.

20230131
Linux.
Screenshot_2023-02-01_16-29-25

Web-eWorks:
Fixes #5512, by refreshing the BBS if the player tries to click on a negative-time ad. A more in-depth fix will need to be deferred to the next patch.
Fixes #5511, prevents an erroneous overwrite of the CrewRoster table on game start. Also generates enough crew to take off with the starting ship, in case we or others setup a start with a ship requiring multiple crew members.
Fixes #5508, displays orbital distance relative to sea-level-radius of the body rather than body's surface. Provides a true "distance to surface" display.
Fixes #5521, increase Skipjack sensor slots to 3 to make it an "exploration" themed medium courier.

This PR also adjusts the parameters of the expensive orbital scanner to provide a generally faster scan rate than a cheap scanner, and make it a little bit more useful when scanning small-to-medium moons.

### DIFF
--- a/data/libs/Ship.lua
+++ b/data/libs/Ship.lua
@@ -885,10 +885,13 @@ local onGameStart = function ()
 	if loaded_data then
 		CrewRoster = loaded_data
 		Event.Queue('crewAvailable') -- Signal any scripts that depend on initialised crew
-	else
-		CrewRoster = {}
 	end
+
 	loaded_data = nil
+end
+
+local onGameEnd = function()
+	CrewRoster = {}
 end
 
 local serialize = function ()
@@ -940,6 +943,7 @@ end
 Event.Register("onEnterSystem", onEnterSystem)
 Event.Register("onShipDestroyed", onShipDestroyed)
 Event.Register("onGameStart", onGameStart)
+Event.Register("onGameEnd", onGameEnd)
 Event.Register("onShipTypeChanged", onShipTypeChanged)
 Serializer:Register("ShipClass", serialize, unserialize)
 
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -211,7 +211,7 @@ misc.orbitscanner_good = BodyScannerType.New({
 	capabilities={mass=7,sensor=1}, purchasable=true, tech_level=8,
 	max_range=100000000, target_altitude=0, state="HALTED", progress=0,
 	bodyscanner_stats={scan_speed=3, scan_tolerance=0.05},
-	stats={ aperture = 2.2, minAltitude = 1750000, resolution = 11375, orbital = true },
+	stats={ aperture = 2.8, minAltitude = 1750000, resolution = 12375, orbital = true },
 	icon_name="equip_orbit_scanner"
 })
 
--- a/data/pigui/modules/info-view/02-personal-info.lua
+++ b/data/pigui/modules/info-view/02-personal-info.lua
@@ -15,7 +15,7 @@ local orbiteer = ui.fonts.orbiteer
 local colors = ui.theme.colors
 local icons = ui.theme.icons
 
-local textTable = require 'pigui/libs/text-table'
+local textTable = require 'pigui.libs.text-table'
 
 local l = Lang.GetResource("ui-core")
 
--- a/data/pigui/modules/info-view/03-econ-trade.lua
+++ b/data/pigui/modules/info-view/03-econ-trade.lua
@@ -67,7 +67,8 @@ end
 local function cargolist ()
 	local count = cachedCargoList or rebuildCargoList()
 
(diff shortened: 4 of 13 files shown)


impaktor:
Could you also squeeze in a commit for the Q/R keys change as default?

Web-eWorks:
Yes, will do that in the morning. Thanks for the reminder!

data/libs/Ship.lua
```diff
@@ -885,12 +885,15 @@ local onGameStart = function ()
...
 		Event.Queue('crewAvailable') -- Signal any scripts that depend on initialised crew
-	else
-		CrewRoster = {}
 	end
+
 	loaded_data = nil
 end
 
+local onGameEnd = function()
+	CrewRoster = {}
```
Gliese852, line 894: This can now be randomly cleared before auto-save on game exit.

Web-eWorks: I thought the pre-exit autosave was emitted via a separate event queued before `onGameEnd` was added to the event queue now?

PR #5528, opened 2023-02-03
Handle the case when size_t and unsigned int are the same type

### LINKED ISSUE #5527: 2023-02-03 does not build on FreeBSD/i386 due to incorrect overload
```
/work/usr/ports/games/pioneer/work/pioneer-20230203/src/lua/LuaPushPull.h:18:13: error: redefinition of 'pi_lua_generic_push'
inline void pi_lua_generic_push(lua_State *l, size_t value) { lua_pushinteger(l, value); }
            ^
/work/usr/ports/games/pioneer/work/pioneer-20230203/src/lua/LuaPushPull.h:17:13: note: previous definition is here
inline void pi_lua_generic_push(lua_State *l, unsigned int value) { lua_pushinteger(l, value); }
            ^
/work/usr/ports/games/pioneer/work/pioneer-20230203/src/lua/LuaPushPull.h:35:13: error: redefinition of 'pi_lua_generic_pull'
inline void pi_lua_generic_pull(lua_State *l, int index, size_t &out) { out = luaL_checkunsigned(l, index); }
            ^
/work/usr/ports/games/pioneer/work/pioneer-20230203/src/lua/LuaPushPull.h:34:13: note: previous definition is here
inline void pi_lua_generic_pull(lua_State *l, int index, unsigned int &out) { out = luaL_checkunsigned(l, index); }
            ^
2 errors generated.
```

Looks like on FreeBSD/i386 `size_t` and `unsigned int` are the same type, so the overload becomes bogus.

AMDmi3:
Fixes #5527

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,12 @@ if (HAS_FECLEAREXCEPT AND HAS_FEENABLEEXCEPT AND HAS_FEDISABLEEXCEPT)
 	set(HAS_FPE_OPS ON)
 endif()
 
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles("#include <cstddef>\nvoid foo(size_t){}; void foo(unsigned int){};int main(){}" SIZE_T_AND_UNSIGNED_INT_ARE_DIFFERENT_TYPES)
+if (SIZE_T_AND_UNSIGNED_INT_ARE_DIFFERENT_TYPES)
+	add_definitions(-DSIZE_T_AND_UNSIGNED_INT_ARE_DIFFERENT_TYPES)
+endif ()
+
 if (NOT CMAKE_BUILD_TYPE)
 	set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
 		"Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel."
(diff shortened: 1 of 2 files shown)


AMDmi3:
Looks like vs solution needs path to `buildopts.h` added somewhere. Sorry, I cannot fix this bit myself.

Web-eWorks:
Rather than defining this in `buildopts.h` (and thus including buildopts.h into almost every lua-related file), please simply add it as a target-level or global `#define` via CMake's `[target_]add_definitions()` utility.

Web-eWorks:
In general however, we are phasing out support for 32-bit targets for Pioneer. I would not be surprised to find other areas of the code which have compilation or runtime issues on a 32-bit OS.

AMDmi3:
That would be fine too, but if so it would be nice to have it documented somewhere.

AMDmi3:
Done. Still need help with the solution though.

Web-eWorks:
> Done. Still need help with the solution though.

Looking at the individual `.vcxproj` files, the VS solution only provides 64-bit targets. No changes required.

...now that I'm looking at the code with fresh eyes, the "cleanest" solution is to replace `size_t` with `uint64_t` and `unsigned int` with `uint32_t` (something we probably should have done a while ago), as that a) would not need a preprocessor define, and b) more correctly express the intent of the code regardless of processor word size.

I don't want to ask you to rework this PR for a third time though, so if you're satisfied with the current state, we can merge as-is.

AMDmi3:
I have no problem with reworking this once more. I've opened a separate PR #5538,to leave this for reference purposes, if #5538 is OK this may be closed.

Web-eWorks:
Closing in favor of #5538.

PR #5537, opened 2023-02-10
Cleaning up the facegen pictures

zonkmachine:
This PR contains cleaned up facegen pictures. I also included some new sets of clothes. The general idea here is that most of the pictures are not very high quality so it's not really possible to make something spectacular. If that is what you want you need to start from scratch. Howeve,r it's easy to clean up the worst spots and recolor some of the sets of clothes to increase the variation. The latter is what we're doing here.

:question:  I've only used the gimp "Overwrite 'picture' " when I'm done with a picture. Should it be exported in a special format?

The most common error corrected is random noise/spots on the translucent parts of the pictures. The second most common error is pixelization. A part of the picture is simply lower resolution. I think this may have been from working on smaller pictures. I've done this error during the course of work and hit some similar bumps.

Here is a comparison of a hair image with spots on the translucent part, before and after

**Before:**

8b

**After:**

8a

---

New clothes:

I'm happy with 1, 4, 5 and, 6, The rest needs some work.

**Male:**
1 cloth_0_21
2 cloth_0_22  **->**  cloth_0_22

3 cloth_0_23 **->**  cloth_0_23

4 cloth_0_24
5 cloth_0_25
6 cloth_0_26

**Female:**
7 cloth_1_21 **->**  cloth_1_21

8 cloth_1_22

---

Finally some words on the accessories. I recommend only showing accessories on every third or so character. There is only a few accessories and they are in my opinion the weakest link visually of the generated face. Loose some of them and use the others less frequently. In this PR I exchange some of the images for a dummy, blank one... Also, when we're talking about loosing things. The pink fluffy armour is gone in this version. I know you love it but I wanted to try and cope without it. I have no real preferences here and will of course revert those changes if you have better ideas around this.


nozmajner:
I don't think the pink outfit will be missed. And I agree that the accessories are way over the top and stick out. I wouldn't even mind ditching all of them.
If you didn't change the image dimensions, and just overwrote the originals then everything should work properly.
Thanks's for showing some love towards the facegen. It's a sore point for me, but I'm yet tu muster the energy to redo the whole thing.

Web-eWorks:
In my opinion, accessory 6 (the cybernetic eye) should also be removed. It doesn't really fit the visual style of Pioneer, and significant cybernetic augmentations are not mentioned as a major part of the lore. I'd expect what cybernetics do exist to be more minimalist / "baseline human", and employed more as medical technology (e.g. replacement organs) than the typical "borg" cyberpunk portrayal.

I like where these changes are going, though I'd prefer accessory slot 0 be used for an "empty" accessory, as that makes it easier to randomly chose when generating the face whether to display an accessory or not, without needing changes to the facegen-internal rendering.

zonkmachine:
> In my opinion, accessory 6 (the cybernetic eye) should also be removed.

Right, done! I have removed a lot this time. The tattoos and the most low quality accessories. I also added a number of 'empty' acc files to quickly test how it would be to have an accessory one out of three times.

zonkmachine:
OK. I updated the new clothes (pictures updated above) and I think they're pretty much good to go. Just recoloring some of the old ones.

impaktor:
Squash, before merge?

zonkmachine:
> Squash, before merge?

Sure! I think squashing ON merge with the github builtin tool is better though. I squashed this one down to one commit and force pushed.

Web-eWorks:
Thanks for amending AUTHORS.txt! I'm somewhat surprised you weren't already added, that's probably my fault from a while ago.

zonkmachine:
> I'm somewhat surprised you weren't already added,

Yeah, that's because of me really. I was offered to be named among the mighty but felt fixing typos in comments didn't call for it. It then became a habit.

PR #5538, opened 2023-02-11
Convert integer pi_lua_generic_{push,pull} to fixed width types

### LINKED ISSUE #5527: 2023-02-03 does not build on FreeBSD/i386 due to incorrect overload

AMDmi3:
- Convert to fixed width types, which work the same way on 32 and 64 bit archs and cover all variable width types such as int and size_t
- Replace deprecated luaL_checkunsigned with luaL_checkinteger

Fixes #5527

### DIFF
--- a/src/lua/LuaPushPull.h
+++ b/src/lua/LuaPushPull.h
@@ -12,10 +12,10 @@
 #include <tuple>
 
 inline void pi_lua_generic_push(lua_State *l, bool value) { lua_pushboolean(l, value); }
-inline void pi_lua_generic_push(lua_State *l, int value) { lua_pushinteger(l, value); }
+inline void pi_lua_generic_push(lua_State *l, int32_t value) { lua_pushinteger(l, value); }
 inline void pi_lua_generic_push(lua_State *l, int64_t value) { lua_pushinteger(l, value); }
-inline void pi_lua_generic_push(lua_State *l, unsigned int value) { lua_pushinteger(l, value); }
-inline void pi_lua_generic_push(lua_State *l, size_t value) { lua_pushinteger(l, value); }
+inline void pi_lua_generic_push(lua_State *l, uint32_t value) { lua_pushinteger(l, value); }
+inline void pi_lua_generic_push(lua_State *l, uint64_t value) { lua_pushinteger(l, value); }
 inline void pi_lua_generic_push(lua_State *l, double value) { lua_pushnumber(l, value); }
 inline void pi_lua_generic_push(lua_State *l, const char *value) { lua_pushstring(l, value); }
 inline void pi_lua_generic_push(lua_State *l, const std::string &value)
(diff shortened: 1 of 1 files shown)


AMDmi3:
This code calls for adding asserts on value ranges just for the case. I didn't add them though not to complicate the change, but if you think it's a good idea I'll submit them too.

Web-eWorks:
> This code calls for adding asserts on value ranges just for the case. I didn't add them though not to complicate the change, but if you think it's a good idea I'll submit them too.

Lua 5.2 (which we currently use) represents integers via the mantissa of a double-precision float, allowing for integer values up to approximately 2^53 to be stored. Adding runtime asserts to check the range of an input value for all of these functions is not particularly productive (it would represent a non-trivial amount of runtime overhead in already-slow debug builds), as these are intended to be inlined zero-overhead wrappers for the `lua_pushX` API functions.

I may be misunderstanding the context in which you're mentioning asserts though - a `static_assert` that `size_t` is covered by at least one of these function overloads is perfectly fine.

AMDmi3:
No, I've meant actual runtime `assert()`. If not for value ranges, then at least for signs in unsigned pulls. But if debug build performance is really an issue, let's skip this.

Web-eWorks:
> If not for value ranges, then at least for signs in unsigned pulls.

This is an assert I'm quite happy to add, as we have a small proportion of unsigned integers loaded from Lua compared to the common pattern of pulling regular ints. As much as I appreciate the desire for complete correctness in the interface, Pioneer is ultimately a realtime program, and any situations where out-of-range values present a major concern in terms of correctness and what little security risk Pioneer represents are better addressed in upstream code checking for much more stringent callsite-specific limits.

I'd recommend checking the number signedness and calling `luaL_error()` with the appropriate `NDEBUG` macro guard rather than a C-assert; the latter is a recoverable error and makes determining the source of the errant value in bug reports much easier than unrecoverably aborting the program with no way to get the Lua callstack.

Web-eWorks:
Merging, thank you for the contribution!

PR #5543, opened 2023-02-14
fix debian/rules to compile with correct -DPIONEER_DATA_DIR flag

robberer:


Web-eWorks:
Thank you for the fix!

PR #5549, opened 2023-02-17
Purge .gitignore

Gliese852:
It really seems to me that there are a lot of extra lines in this file, and this is somewhat wrong. In my case, two lines are enough for the output of git status to be clean. Maybe something else needs to be added, I don't know.


.gitignore
```diff
@@ -1,92 +1,2 @@
...
-config.log
-config.status
-config.sub
-configure
-depcomp
-install-sh
-missing
-compile
-test-driver
-stamp-h1
```
Web-eWorks, line 65: These are left-overs from in-tree build with autotools and can be removed.

Web-eWorks:
Most of the removed lines are in fact actually necessary, and ignore outputs from e.g. the Doxygen build step or the release process, in addition to a number of lines ignoring per-user project files for Visual Studio, VSCode, and the various other IDEs (e.g. QT creator, etc.). There are a large number of lines in .gitignore from the old in-tree autotools build system which can be removed, but the current set of changes would provide a net negative benefit to a lot of developers.

.gitignore
```diff
@@ -1,92 +1,2 @@
...
-nd/Data/
-doxygen/html
-doxygen/latex
-*.vcxproj.user
-*.user
-*.ncb
-*.suo
-*.sdf
-*.opensdf
-*.o
```
impaktor, line 27: will this not track all build object files?

.gitignore
```diff
@@ -1,92 +1,2 @@
...
-/*.exe
-/*.dll
-/*.pdb
-/*.ilk
-Makefile
-Makefile.in
-.DS_Store
-*.swp
-shipstat.csv
-pioneer
```
impaktor, line 39: should we not ignore the output of a build?

.gitignore
```diff
@@ -1,92 +1,2 @@
...
-*.user
-*.exp
-*.VC.db
-/release
-# Core dumps shouldn't be tracked by git.
-.[0-9][0-9]*
-# Don't track C++ language server cache files.
-.ccls-cache
-.clangd/
-.cache/
```
impaktor, line 87: Why don't we want to ignore cache?

.gitignore
```diff
@@ -1,92 +1,2 @@
...
-.vscode/
```
impaktor, line 88: I imagine people using vscode would want to ignore the config?

impaktor (commented):
What triggered this purge? What's the problem being solved?
(Sorry for not understanding...)

Gliese852:
@impaktor 
> What's the problem being solved?

.gitignore is not up to date, I think.

.gitignore
```diff
@@ -1,92 +1,2 @@
...
-debian/autoreconf*
-debian/files
-debian/*.debhelper.log
-debian/*.substvars
-debian/pioneer*/
-autom4te.cache/
-codedoc/
-nd/Data/
-doxygen/html
-doxygen/latex
```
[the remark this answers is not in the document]
Gliese852: I personally feel more comfortable when I see in the status that this is generated, I admit that this is not the case for everyone.

.gitignore
Gliese852: It should be in `build`.

.gitignore
Gliese852: It should be in `build`.

.gitignore
Gliese852: I admit, although it seems to me that this is a personal workflow - neither `ccls` nor `LuaLS` leave anything in my working tree.

.gitignore
Gliese852: Yes, if it is needed for a large number of developers, it is worth leaving it.

.gitignore
impaktor: Ah, right. We're on CMake now.

.gitignore
impaktor: I have a binary pioneer in my root (that points to build/). It compiles & runs the game.
(Granted, I usually don't use it)

impaktor:
I'm thinking it's natural to gitignore all output from builds (binary, documentation, etc), editor configuration.

Stuff that's no longer being built, should be removed from the gitignore, like .o files in src/

Anyway, that's my thinking, but I'm not the one working with the code much these days...

Gliese852:
I would also like to pay attention to some guides from the gitignore documentation
https://git-scm.com/docs/gitignore

Which file to place a pattern in depends on how the pattern is meant to be used.

- Patterns which should be version-controlled and distributed to other repositories via clone (i.e., files that all developers will want to ignore) should go into a `.gitignore` file.

- Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the `$GIT_DIR/info/exclude` file.

- Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by `core.excludesFile` in the user’s `~/.gitconfig`. Its default value is `$XDG_CONFIG_HOME/git/ignore`. If `$XDG_CONFIG_HOME` is either not set or empty, `$HOME/.config/git/ignore` is used instead.

Gliese852:
do we need debian-specific ignores?
```
debian/autoreconf*
debian/files
debian/*.debhelper.log
debian/*.substvars
debian/pioneer*/
```

Web-eWorks:
I believe so; this would be a question for @robberer however.

In general, every expected "build product", per-user editor configuration file, or other file that we know is produced while editing the code and do not want to have in source control **should** be covered by the .gitignore. The expectation should be that a non-technical user can download the project, open it in a mainstream editor of their choice (Visual Studio, VSCode, vim, emacs, Eclipse, QTCreator, Code::Blocks, etc.), make changes, build the game and any/all associated products, test their changes and blindly run `git add` without any issues at pull-request time.

Editing the repository-local git config file is definitely a good option for esoteric setups, but it presumes a level of Git knowledge that an entry-level contributor is unlikely to possess or learn easily. Thus, a comprehensive .gitignore is both a requirement and a net benefit to the project.

robberer:
> do we need debian-specific ignores?
> 
> ```
> debian/autoreconf*
> debian/files
> debian/*.debhelper.log
> debian/*.substvars
> debian/pioneer*/
> ```

on my debian 12 system using `debuild -b -uc -us -d` in project root, these lines are needed:

```
debian/.debhelper/
debian/pioneer*
debian/tmp/
debian/files
obj-x86_64-linux-gnu/
```

Gliese852:
@robberer i.e. these are some files that you need all the time? do you want to clean it up before the next build, or will you do it manually?

Gliese852:
.gitignore has recently been tidied up, it's probably in a good state now.

Web-eWorks:
I have a little bit of cleanup I'd like to do to it (remove old rules for in-tree builds with autotools) but I think it's in a good state, yes.

PR #5553, opened 2023-02-19
Update all links to the wiki, to point to new (sub)domain

impaktor:
Wiki has moved, we're deprecating old URL.

I've changed some of the wiki links to point to the https://dev.pioneerspacesim.net/ dev docs.

The link in src/scenegraph/Model.h points to a removed wiki page (New Model), I've pointed it to [model system](https://wiki.pioneerspacesim.net/wiki/Model_system) instead. (Is that right?)


impaktor:
Btw, I note our API link is 404: https://pioneerspacesim.net/codedoc/

Do we have a github pipeline for building this, or it's no longer hosted anywhere?

(We have at least one link still pointing to the old site: http://eatenbyagrue.org/f/pioneer/codedoc/ no point in updating this, until we know where we put the new API docs)

Issue #5550, opened 2023-02-17
Multiple empty face accessory items


Gliese852:
### Observed behaviour

https://user-images.githubusercontent.com/18342621/219796197-9625fd26-8ca1-4ab4-a80b-9db5f66f46c2.mp4

As shown in the video, you can click multiple times and nothing changes.
Apparently this is happening because in #5537 a lot of accessories have been removed.

__My pioneer version (and OS):__
https://github.com/pioneerspacesim/pioneer/commit/d25533200fe2acea82d24207209e5e504bd42b5b

Web-eWorks:
Yes, this needs to be replaced with random face generation having a higher chance of "no accessory" compared to the other accessory items. Leaving this open so we don't forget it.

zonkmachine:
Yeah, they were meant to be an approximation of a solution for testing purposes.

PR #5558, opened 2023-02-21
Fixing issue #5550 where empty accessory files makes manual handling erratic

zonkmachine:
Fixes https://github.com/pioneerspacesim/pioneer/issues/5550

Here is a fix for #5550.
The Character faces now have an accessory generated about one time in four. Player character, bbs faces or random button in the face app in the 'Personal Information' window all works. When you step the accessories in the face app they will show up all about the 1st step which is an empty file that was kept so you can intentionally select 'no accessory'.


Web-eWorks (approved):

data/pigui/libs/face.lua
```diff
@@ -36,7 +36,7 @@ local function rerollFaceDesc(oldDesc)
 		FEATURE_NOSE = Engine.rand:Integer() % 2^31,
 		FEATURE_MOUTH = Engine.rand:Integer() % 2^31,
 		FEATURE_HAIRSTYLE = Engine.rand:Integer() % 2^31,
-		FEATURE_ACCESSORIES = Engine.rand:Integer() % 2^31,
+		FEATURE_ACCESSORIES = Engine.rand:Integer() % 4 == 0 and Engine.rand:Integer() % 2^31 or 0, -- accessory on one out of four faces
```
Gliese852, line 39: I guess the `and` branch can also generate an empty accessory?

zonkmachine: Correct. With the current number of accessories every eight png will be invisible.

impaktor:
Did we used to have santa hat for our faces, when computer clock is Christmas, or was it Frontier, or did I dream that?
Either way, would be fun easter egg, when playing during Christmas break.

zonkmachine:
> Did we used to have santa hat for our faces, when computer clock is Christmas, or was it Frontier, or did I dream that?

I just experimented with a santa hat and also a hijab to test the layers of facegen. Are any of those actually in this PR?! I can't find them in here.

Web-eWorks:
Looks good, thanks @zonkmachine!

PR #5560, opened 2023-02-24
Fix + nerf metal creaking sfx

WKFO:
So the reason this squeaky metal sound has been mistriggering was _probably_ (keyword: probably) that I used GetLastForce() which includes the gravity force, which is a body force and shouldn't be added to the metal strain calculation. Also I got rid of the array of 20 last jerk values because that was a safeguard against single-frame thruster spikes of the autopilot and flight assist. Now that we only use atmospheric acceleration, it should no longer be needed.

In addition, I made the triggering acceleration limits quite higher and the volume lower.

Needs playtesting and a sanity check from someone other than me.

### DIFF
--- a/src/Player.cpp
+++ b/src/Player.cpp
@@ -55,7 +55,7 @@ Player::Player(const ShipType::Id &shipId) :
 	InitCockpit();
 	m_fixedGuns->SetShouldUseLeadCalc(true);
 	registerEquipChangeListener(this);
-	m_accel = vector3d(0.0f, 0.0f, 0.0f);
+	m_atmosAccel = vector3d(0.0f, 0.0f, 0.0f);
 }
 
 Player::Player(const Json &jsonObj, Space *space) :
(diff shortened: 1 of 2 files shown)


craigo-:
Thanks for looking at this. I'll give this PR a test run over the next 24 hours or so.

Please remind me - when are we _supposed_ to hear the metallic groaning?

WKFO:
It is supposed to play when the acceleration is high and it changes rapidly, e.g. when you dive into thick atmosphere too fast or when you are going fast in-atmo and you suddenly change your angle of attack.

craigo-:
OK, after a fairly large number of planetary landings in various places, I did not hear any metallic creaking where I believe I probably would have before.

I did a bit of atmospheric flying as well, including mach 2 at 3000ft at Earth with some fairly rugged attitude and velocity changes, and also did not hear the ship creak. Maybe I'm not pushing it hard enough. The ship is a Mola Mola.

impaktor:
@craigo- I just want to say thanks for testing PRs, it's really great with the feedback, and it saves time for developers, so they can do more dev-ing.

Web-eWorks:
3Gs of atmospheric force against the ship is an extremely significant load and may only show up when doing a full atmospheric re-entry with a ship having a very large surface area (e.g. Wave) at *very* high speeds. I'd suggest reducing the required force, but scale the force by `1 - abs(dot(force.Normalized(), ship_forward))` so that the creaking sound effect only triggers when the atmospheric forces are being applied perpendicular to the forward vector.

PR #5569, opened 2023-03-09
Prevent zero length direction vector

### LINKED ISSUE #5516: SIGFPE if flying long enough
### Observed behaviour
```Thread 1 "pioneer" received signal SIGFPE, Arithmetic exception.
0x000055555575f198 in vector3<double>::Normalized (this=<synthetic pointer>) at /home/mcpain/pioneer/src/vector3.h:127
127                     const T l = 1.0f / sqrt(x * x + y * y + z * z);
(gdb) print x
$1 = <optimized out>
(gdb) print y
$2 = <optimized out>
(gdb) print z
$3 = <optimized out>
(gdb) bt
#0  0x000055555575f198 in vector3<double>::Normalized (this=<synthetic pointer>) at /home/mcpain/pioneer/src/vector3.h:127
#1  Propulsion::AIFaceUpdir (this=0x55558bd17ef8, updir=..., av=av@entry=0) at /home/mcpain/pioneer/src/ship/Propulsion.cpp:396
#2  0x00005555556818bc in AICmdFlyTo::TimeStepUpdate (this=0x5555871d4f20) at /home/mcpain/pioneer/src/ShipAICmd.cpp:1141
#3  0x0000555555679ab7 in AICommand::ProcessChild (this=this@entry=0x5555885b2d60) at /home/mcpain/pioneer/src/ShipAICmd.cpp:87
#4  0x000055555567ce3a in AICmdDock::TimeStepUpdate (this=0x5555885b2d60) at /home/mcpain/pioneer/src/ShipAICmd.cpp:1246
#5  0x00005555557ed4f2 in Ship::AITimeStep (this=0x555595507420, timeStep=timeStep@entry=166.666672) at /home/mcpain/pioneer/src/Ship-AI.cpp:37
#6  0x0000555555761702 in ShipController::StaticUpdate (this=0x55557994d6c0, timeStep=166.666672) at /home/mcpain/pioneer/src/ship/ShipController.cpp:11
#7  0x0000555555670587 in Ship::StaticUpdate (this=0x555595507420, timeStep=166.666672) at /home/mcpain/pioneer/src/Ship.cpp:1215
#8  0x000055555568fa33 in Space::TimeStep (this=0x55555db1d800, step=step@entry=166.666672) at /home/mcpain/pioneer/src/Space.cpp:1042
#9  0x00005555557a4ae9 in Game::TimeStep (this=0x5555590e2460, step=166.666672) at /usr/include/c++/12/bits/unique_ptr.h:191
#10 0x000055555563ed92 in GameLoop::Update (this=0x555555be0e70, deltaTime=0.276787758) at /home/mcpain/pioneer/src/Pi.cpp:944
#11 0x00005555555f3c07 in Application::Run (this=0x555555be0c90) at /home/mcpain/pioneer/src/core/Application.cpp:192
#12 0x00005555555d5272 in main (argc=<optimized out>, argv=<optimized out>) at /home/mcpain/pioneer/src/Pi.h:145
```

### Expected behaviour
Pioneer doesn't crash

### Steps to reproduce
1. Start on Mars
2. Take off, remove landing gear
3. Engage vertical thrusters (accelerate away from Mars)
3. Request time warp 10000x
4. Wait until running out of fuel (repeat 10000x as needed)
5. Enjoy a one-way trip away from solar system
6. Eventually, the game crashes. Sometimes before ship runs out of fuel.

ALT Linux, e765fd8cf

Gliese852:
Approximately once every one and a half in-game years (for the Sol system), the following state may occur:

  - some ship is directed towards the planet
  - it has AICmdFlyTo command active
  - its target is on course
  - gravity, respectively, is directed along the the ship, and accelerates it
  - its reverse thrusters are _weaker than gravity_
  - it has zero speed

The combination of these circumstances leads to the fact that both coefficients in the calculation of the head vector are equal to 0. A vector of zero length is passed to Propulsion, where it is normalized and SIGFPE occurs.

Fixes #5516

It would be nice to do a thorough reworking of the autopilot here. For example, to ban ships from flying and stopping with their nose down in gravity, it looks strange. @azieba you worked next to this code not so long ago, I would like to know if you have any plans here?

### DIFF
--- a/src/ShipAICmd.cpp
+++ b/src/ShipAICmd.cpp
@@ -1203,7 +1203,11 @@ bool AICmdFlyTo::TimeStepUpdate()
 	// then flip the ship so we can use our main thrusters to decelerate
 	if (m_state && !is_zero_exact(sdiff) && sdiff < maxdecel * timestep * 60) head = -head;
 	if (!m_state && decel) sidefactor = -sidefactor;
-	head = head * maxdecel + perpdir * sidefactor;
+
+	// check that head does not become zero length
+	if (maxdecel > 0.001 || abs(sidefactor) > 0.001) {
+		head = head * maxdecel + perpdir * sidefactor;
+	}
 
 	// face appropriate direction
 	if (m_state >= 3) {


azieba:
> It would be nice to do a thorough reworking of the autopilot here. For example, to ban ships from flying and stopping with their nose down in gravity, it looks strange. @azieba you worked next to this code not so long ago, I would like to know if you have any plans here?

Does this really happen? Do you have 'how to reproduce' maybe? I would gladly take a look, unless you want to work on it.
I would not rework it entirely though, I am not sure I understand all the concepts in this code.

Gliese852:
@azieba to reproduce the SIGFPE, you can launch on Mars, turn on time acceleration, and skip a year or two. (one year on x10000 is 53 minutes of real time).

I cheated a little to create a situation quickly. If built from this branch, 
https://github.com/Gliese852/pioneer/tree/test-sigfpe
2 buttons appear near the speed limiter, "1" and "2".
You need to start on Mars, press "Undock", "1", "2" and SIGFPE happens.

And the "weird nose-down hover" happens every time you land on autopilot at a surface space station - the ship hangs nose-down at an altitude of 500 m, only then it turns belly down, releases the landing gear and begins to descend onto the pad.

azieba:
> And the "weird nose-down hover" happens every time you land on autopilot at a surface space station - the ship hangs nose-down at an altitude of 500 m, only then it turns belly down, releases the landing gear and begins to descend onto the pad.

Oh, you mean the phase when the ship goes from 15km above the station down to 500m nose down. Well I assumed that this was originally intended, but when I think about it, yes it seems a bit strange.  It might be the case that if the gravity is greater then retro thrusters the autopilot would flip the ship around, but this would be rare. Do you think that the ship should descend using main thrusters in all cases?

Web-eWorks:
> It might be the case that if the gravity is greater then retro thrusters the autopilot would flip the ship around, but this would be rare. Do you think that the ship should descend using main thrusters in all cases?

The ship should always descend with the belly or rear thrusters pointing in the direction of the ship's velocity (ideally the autopilot code queries the ship's Propulsion for the correct descent orientation, for e.g. future tail-sitter craft), depending on the gravity. This is for two reasons - it looks more correct to a human audience, and allows for emergency burns to avoid impacting the planet, especially if the player were to take over the autopilot.

I'd generally expect a descent using rear thrusters on airless worlds, and a descent using belly thrusters on atmospheric worlds, as our craft generally produce the most drag when descending straight up/down. Ships that cannot fight the world's gravity with their downwards thrusters should be incapable of landing on planets (as ships generally do not have landing gear in such a way they could land with the rear thruster facing the pad).

azieba:
> I'd generally expect a descent using rear thrusters on airless worlds, and a descent using belly thrusters on atmospheric worlds, as our craft generally produce the most drag when descending straight up/down. Ships that cannot fight the world's gravity with their belly thrusters should be incapable of landing on planets (as ships generally do not have landing gear in such a way they could land with the rear thruster facing the pad).

This should not be too hard. I'll try to implement it.

azieba (approved):
The fix is fine for me. I just wonder if we should have protection against null dir vectors in the propulsion code?

Web-eWorks:
As long as a vector's length is greater than epsilon, usable direction information (of arguable quality) can be extracted from it - I'm not sure it's worth attempting to "fix" zero-length vectors inside of Propulsion, as there's no useful semantic information left and the calling code is almost always more qualified to determine what the effect should be if the vector is zero-length.

Issue #5387, opened 2022-06-25
Crash during loading with multiple WorkerThreads enabled on Intel(R) Core(TM) i3-2100 CPU @ 3.10GHz


pirogronian:
### Observed behaviour
Crash during loading, always in the same place and manner. 

Console log  with gdb session (last lines):
```
Loading [92%]: PostLoad took 0.03ms
Pioneer loading took 6150.41ms
Creating new galaxy generator 'legacy' version 1
Clearing and re-using previous Galaxy object
StarSystemCache: misses: 0, slave hits: 0, master hits: 0
SectorCache: misses: 101, slave hits: 0, master hits: 1
Created shader starfield (address=0x555556ca8e30)
Stars picked from galaxy: 103983
Generating 21017 random stars
/usr/include/c++/12.1.0/bits/stl_vector.h:1123: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = vector3<float>; _Alloc = std::allocator<vector3<float> >; reference = vector3<float>&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.

Thread 1 "pioneer" received signal SIGABRT, Aborted.
0x00007ffff6e8e36c in ?? () from /usr/lib/libc.so.6
```
Gdb backtrace:
```
(gdb) bt
#0  0x00007ffff6e8e36c in ?? () from /usr/lib/libc.so.6
#1  0x00007ffff6e3e838 in raise () from /usr/lib/libc.so.6
#2  0x00007ffff6e28535 in abort () from /usr/lib/libc.so.6
#3  0x00007ffff72d2002 in std::__glibcxx_assert_fail (file=<optimized out>, line=<optimized out>, function=<optimized out>, condition=<optimized out>)
    at /usr/src/debug/gcc/libstdc++-v3/src/c++11/debug.cc:60
#4  0x00005555555f467e in ?? ()
#5  0x00005555556f6b21 in ?? ()
#6  0x00005555556f6f35 in ?? ()
#7  0x00005555556f7fba in ?? ()
#8  0x000055555572b086 in ?? ()
#9  0x000055555572b259 in ?? ()
#10 0x0000555555600ab9 in ?? ()
#11 0x00005555555bdecc in ?? ()
#12 0x00005555555be069 in ?? ()
#13 0x00005555555a8cb6 in ?? ()
[...]

Issue #5570, opened 2023-03-14
pioneer game crashes on game save


sagitter:
### Observed behaviour
pioneer (cd91559), compiled in Fedora 37, is crashing with

```
Stack trace of thread 80278:
#0  0x00007f89bf4afe5c __pthread_kill_implementation (libc.so.6 + 0x8ce5c)
#1  0x00007f89bf45fa76 raise (libc.so.6 + 0x3ca76)
#2  0x00007f89bf4497fc abort (libc.so.6 + 0x267fc)
#3  0x00007f89bf6d7e80 _ZSt21__glibcxx_assert_failPKciS0_S0_ (libstdc++.so.6 + 0xd7e80)
#4  0x0000556b8a73fef3 _ZN11PropertyMap8iteratorppEv (pioneer + 0x375ef3)
#5  0x0000556b8a74245a _ZN11PropertyMap10SaveToJsonERN8nlohmann10basic_jsonISt3mapSt6vectorNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEblmdSaNS0_14adl_serializerEEE (pioneer + 0x37845a)
#6  0x0000556b8a598c7a _ZN4Body10SaveToJsonERN8nlohmann10basic_jsonISt3mapSt6vectorNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEblmdSaNS0_14adl_serializerEEEP5Space (pioneer + 0x1cec7a)
#7  0x0000556b8a4fcc7f _ZN11TerrainBody10SaveToJsonERN8nlohmann10basic_jsonISt3mapSt6vectorNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEblmdSaNS0_14adl_serializerEEEP5Space (pioneer + 0x132c7f)
#8  0x0000556b8a5ab70f _ZN4Game6ToJsonERN8nlohmann10basic_jsonISt3mapSt6vectorNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEblmdSaNS0_14adl_serializerEEE (pioneer + 0x1e170f)
#9  0x0000556b8a5b0cde _ZN4Game8SaveGameERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPS_ (pioneer + 0x1e6cde)
#10 0x0000556b8a65075f _ZL16l_game_save_gameP9lua_State.lto_priv.0 (pioneer + 0x28675f)
[...]

PR #5574, opened 2023-03-21
Fix out-of-bounds index uncovered by -D_GLIBCXX_ASSERTIONS

yump:
Flatpak and Fedora packages are built with this flag, and it was causing a crash on save. Also [encountered by manolollr](https://github.com/pioneerspacesim/pioneer/issues/5387#issuecomment-1172784841) in #5387.

### DIFF
--- a/src/core/Property.cpp
+++ b/src/core/Property.cpp
@@ -115,7 +115,7 @@ PropertyMap::iterator PropertyMap::iterator::operator++()
 {
 	size_t sz = map->m_keys.size();
 	if (map && sz > idx)
-		while (!map->m_keys[++idx] && idx < sz) {
+		while (++idx < sz && !map->m_keys[idx]) {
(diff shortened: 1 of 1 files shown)


impaktor:
Could #5570 be related to this as well?

yump:
Ah, yeah that's the same stack trace.  I missed it because I searched the bug tracker for "Assertion '__n < this->size()' failed".

yump:
Update: Applying this as a patch to the Fedora 37 package, building locally, and installing, does indeed produce a `/usr/bin/pioneer` that doesn't crash on save.

(This is a separate comment because I don't know whether marking as ready triggers a github notification.)

Web-eWorks (approved):

Web-eWorks:
Thanks for catching that issue!

PR #5575, opened 2023-03-25
Build Lua API docs

impaktor:
Builds lua API using natural docs 2.2 version as discussed on the [dev forum](https://forum.pioneerspacesim.net/viewtopic.php?f=3&t=721)

Seems to work on my fork:
- build pipeline example: https://github.com/impaktor/pioneer/actions/runs/4520007174
- pushes build to here: https://impaktor.github.io/pioneer/
- and also generates zip for download (see same [pipeline](https://github.com/impaktor/pioneer/actions/runs/4520007174) as above), for those who want API docs off line.

I'll put it at https://codedoc.pioneerspacesim.net once merged.

I haven't done any fancy cmake magic, just running it manually (I don't know if @pcercuei has any thoughts on this?)

__To-do__: fix so Attributes are actually included as well....

...also, shall we do something with the doxygen? Perhaps not as urgent, as it doesn't have any heavy dependencies to run (like nd 2.x, which depends on billions of MB for installing mono-devel, and c++ devs have skill to build doxygen on their own, I assume)


pcercuei:
My thoughts are that it's a bit too complicated to have it in CMake... You need to download a ZIP, extract a ZIP, then run naturaldocs.exe with Mono (if you are under Linux, because on Windows or OSX, it's different). To be fair it's not really the CMake script's job to do that. Rather it's NaturalDocs that should provide proper Linux packages.

impaktor:
@pcercuei OK, thanks, then what I have above sounds like the (more) propper way to do it.

pcercuei:
It should just use `mono-runtime`, yeah (provided it's a debian-based OS).

impaktor:
For the record API now lives on: https://codedoc.pioneerspacesim.net and I've cleaned up the `nd/<config>` a bit, some files were deprecated. Not sure if it's possible to add the pioneer logo to the documentation, like we do for doxygen, but this will do for now. The build date of the API docs (shown under the title) should also be updated automatically.

PR #5522, opened 2023-02-01
Show info when the target area is reached

robothauler:
Reduce confusion by always indicating that the target area has been reached.
Closes #5499

Issue #5499, opened 2023-01-29
Armed reconnaissance mission completed without data logging nor destroying


DynV:
### Observed behaviour
I had an armed reconnaissance mission that was completed without ~~scanning~~data logging the planet nor destroying any of the ruthless market competitor.

I know the enemies spawn randomly so I see no use in including the save-file.

**Update 1**
I remember a ship was spawned somewhat away from the planet (100Mm?) and at some point disappeared. Perhaps it was the main target and there was the bug that it crashed into the planet so it completed?

### Expected behaviour
In the comms either get a maximum ~~scanning~~data logging time or after destroying the surrounding ship get a mission completed message, but it was displayed before either appeared.

### Steps to reproduce
Randomly get a mission that behaves the same.

__My pioneer version (and OS):__
20220203-win
![Pioneer-completed_without_scanning_nor_destroying](https://user-images.githubusercontent.com/2823807/215309981-19c9614e-cb5e-4df2-a4a4-ec4f95a10eef.jpg)

impaktor:
Sounds like MISSION_TYPE_2 of the Combat module. (I had to check, because above doesn't sound familiar at all. )
Objective: "Weaken the enemy's strength as much as possible and then jump to safety."
Risk: "The competitor has only minor resources. But he will defend his position if necessary."
I'm very unfamiliar with this module. Whatever "scanning" is done, is not using the scanning equipment we just introduced. Was written by @robothauler (hauler in IRC)

robothauler:
You don't need a scanner for this mission. A radar to detect the enemy ships is sufficient. Sometimes the enemy ship crashes into the Planet and then the mission is accomplished. That may have happened here.

impaktor:
Since we now have surface and orbital scanner, there might be wording in this mission that lets player thing they need those. I'm not sure. Perhaps @DynV want to read through the [mission module lang](https://github.com/pioneerspacesim/pioneer/blob/master/data/lang/module-combat/en.json) and point out if there are any strings that might lead to such confusion between radar and (orbit/surface)-scanner?

DynV:
It was data logging, not scanning. Although... Wouldn't one need instruments to do that, perhaps even a scanner?
![screenshot-20230129-225856](https://user-images.githubusercontent.com/2823807/215384778-8c3f92d7-95f2-418a-976a-7c13f31dc428.png)

DynV:
[...]

PR #5588, opened 2023-05-29
Combat mission explanation about special equipment

robothauler:
An explanation similar to the scout mission of how it works.

Related to #5499 
Addition to #5522

### DIFF
--- a/data/lang/module-combat/en.json
+++ b/data/lang/module-combat/en.json
@@ -381,6 +381,6 @@
   },
   "YOU_NEED_A_RADAR": {
     "description": "",
-    "message": "You need a radar! If you agree, I will program your ship's computer to record the radar data we need. You must hand over this data to us the next time we meet!"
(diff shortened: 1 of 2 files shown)


Web-eWorks (approved):

data/lang/module-combat/en.json
```diff
@@ -374,5 +378,9 @@
   "WILL_I_BE_IN_TROUBLE": {
     "description": "",
     "message": "Will I be in trouble?"
+  },
+  "YOU_NEED_A_RADAR": {
+    "description": "",
+    "message": "You need a radar! If you agree, I will program your ship's computer to record the radar data we need. You must hand over this data to us the next time we meet!"
```
impaktor, line 384: Perhaps nitpick, but feels like one exclamation point too much. How about: 

> Yes, you need a radar! If you agree, I will program your ship's computer to record the radar data we need. The data will be transferred to me next you come back.

PR #5589, opened 2023-06-06
Include ad title in bulletin board search

zonkmachine:
Include ad title in search on the bulletin board.

Motivation: I just found it more practical/correct after interpreting an earlier search result as a wee bit odd.

### DIFF
--- a/data/pigui/modules/station-view/02-bulletinBoard.lua
+++ b/data/pigui/modules/station-view/02-bulletinBoard.lua
@@ -55,12 +55,12 @@ local function refresh()
 	bulletinBoard.items = {}
 
 	for ref,ad in pairs(ads or {}) do
-		if searchText == ""
-			or searchText ~= "" and string.find(
+		if not searchText or searchText == ""
+			or string.find(
 				string.lower(ad.title),
 				string.lower(searchText),
 				1, true)
-			or searchText ~= "" and string.find(
+			or string.find(
 				string.lower(ad.description),
 				string.lower(searchText),
 				1, true)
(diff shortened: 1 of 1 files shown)


data/pigui/modules/station-view/02-bulletinBoard.lua
```diff
@@ -56,6 +56,10 @@ local function refresh()
 
 	for ref,ad in pairs(ads or {}) do
 		if searchText == ""
+			or searchText ~= "" and string.find(
```
zonkmachine, line 59: Isn't `searchText ~= "" and` in this and the next paragraph redundant after the test on line 58?

zonkmachine: I mean you could simplify it as such:

```lua
        for ref,ad in pairs(ads or {}) do
                if searchText == ""
-                       or searchText ~= "" and string.find(
+                       or string.find(
                                string.lower(ad.title),
                                string.lower(searchText),
                                1, true)
-                       or searchText ~= "" and string.find(
+                       or string.find(
                                string.lower(ad.description),
                                string.lower(searchText),
                                1, true)

```

Web-eWorks: @zonkmachine is correct, `searchText ~= ""` is implied by the first condition of the if statement evaluating to false. The only way this breaks is if `nil` is passed as `searchText`, which can be handled by changing the first condition to `if not searchText or searchText == ""`.

Web-eWorks (approved):
Looks good, if the code has been tested then this can be squashed and merged.

impaktor:
If someone has tested this, please holler

Issue #5215, opened 2021-06-16
ECM popup text is not informative


jimishol:
(Not really an issue)
When hovering mouse upon ECM button, popups the 'ECM' string. That is not a plus information concerning on that button and null information should cost zero bytes to produce. So, as a matter of principal, there should not exist a popup at all.
This popup string is defined as last string at 65 line of equipment.lua file, that is:

local size, clicked = iconButton(uiPos, icons[ecm.ecm_type], false, mainIconSize, "ECM", not player:IsECMReady(), mainBackgroundColor, mainForegroundColor, mainHoverColor, mainPressedColor, 'ECM')

If should exist a popup informative text, should not be 'ECM' because the label of button as 'ECM' gives that information already, but something different. My suggestion is to replace last 'ECM' with something like
'Activate', 'Activates electronic countermeasures', 'Electronic CounterMeasures', 'Start ECM', 'Implement ECM' and so on.

zonkmachine:
I suggest "Electronic countermeasure" as it's used in the title of the wikipedia page.
https://en.wikipedia.org/wiki/Electronic_countermeasure

Maybe it should be possible to turn the tooltip off with an empty string if it's not needed? Now you get an empty small box in there. That would be better than repeating the text of the button.

PR #5590, opened 2023-06-11
Mouseover messages for the ECM and the Advanced ECM systems

zonkmachine:
Implement hover over messages for the ECM and ECM Advanced.

ecm3

 * The message is just `Electronic countermeassure` and `Advanced electronic countermeassure`. This harmonizes with the messages for the missiles above which just state the missile types.
 * The hover over message for the missiles capitalize the first letters on all words and this PR doesn't. Which one is preferred?

ecm1

Fixes https://github.com/pioneerspacesim/pioneer/issues/5215

### DIFF
--- a/data/lang/equipment-core/en.json
+++ b/data/lang/equipment-core/en.json
@@ -151,6 +151,10 @@
     "description": "",
     "message": "An electronic countermeasure missile defence system, capable of destroying some homing missiles."
   },
+  "ECM_HOVER_MESSAGE": {
+    "description": "",
+    "message": "Activate Electronic Countermeasures"
+  },
   "EMPTY_SLOT": {
     "description": "Text indicating an empty equipment slot.",
     "message": "EMPTY"
--- a/data/libs/Equipment.lua
+++ b/data/libs/Equipment.lua
@@ -78,12 +78,14 @@ misc.heavy_atmospheric_shielding = EquipType.New({
 misc.ecm_basic = EquipType.New({
 	l10n_key="ECM_BASIC", slots="ecm", price=6000,
 	capabilities={mass=2, ecm_power=2, ecm_recharge=5},
-	purchasable=true, tech_level=9, ecm_type = 'ecm'
+	purchasable=true, tech_level=9, ecm_type = 'ecm',
+	hover_message="ECM_HOVER_MESSAGE"
 })
 misc.ecm_advanced = EquipType.New({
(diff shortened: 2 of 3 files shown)


impaktor:
I suspect capitalization is more in line what we're doing / intend to do. But then @Web-eWorks is the only native speaker here.

zonkmachine:
> I suspect capitalization is more in line what we're doing / intend to do. But then @Web-eWorks is the only native speaker here.

OK, fixed that.

data/lang/equipment-core/en.json
```diff
@@ -151,6 +155,10 @@
     "description": "",
     "message": "An electronic countermeasure missile defence system, capable of destroying some homing missiles."
   },
+  "ECM_BASIC_HOVER_MESSAGE": {
+    "description": "",
+    "message": "Electronic Countermeassure"
```
Web-eWorks, line 160: Properly rendered in English, this would be "Electronic Countermeasure System" or "Electronic Countermeasures" (note the single 's' in countermeasure). As this is a hover message, I'd much prefer "Activate Electronic Countermeasures" as that explains the action rather than the system. This also allows the string to be used when we rework the missile / countermeasures system interface sometime soon:tm:.

Web-eWorks (changes requested):

Web-eWorks (approved):

zonkmachine:
I wonder if the message should be the same for both ECM types. 'Activating Advanced Electronic Countermeasures' sound a bit too much.

Web-eWorks:
Sorry about the response delay. Ideally, we'd differentiate the two types of ECM equipment separately via an icon or some other visual indicator, so I'm generally in favor of using the same "Activate Electronic Countermeasures" tooltip for both. I don't think it makes sense to have basic and advanced ECM generators installed simultaneously and *not* activate both when attempting to defeat a missile, so having separate activation buttons isn't something we'd be aiming for.

zonkmachine:
> so I'm generally in favor of using the same "Activate Electronic Countermeasures" tooltip for both.

Yes, this sounds better. Fixed!

I think I'm happy with the PR now.

nozmajner:
There should be a different icon for both types of ECM (single and double thunder, and as far as I can see, the icon set is set up that way)
image

zonkmachine:
Yes, the icons are showing correctly.

Issue #5581, opened 2023-04-28
1.0g way too common


goat-soup-eater:
### Observed behaviour
Too many planets have exactly 1.0g of gravity

### Expected behaviour
Should be more variation, and not exactly 1.0g

### Steps to reproduce
Check out Zarelium (-12, -1, -2). All 5 main planets have exactly 1.0g

__My pioneer version (and OS):__
v20230203 (e77da96) Windows

zonkmachine:
Since the surface gravity for our solar system seem to work well, and it's fed real and good data, it should be the automatically generated planets that gets suboptimal values for either mass or radius (or both). The planet mass seem to be generated first (I haven't found where).

The line controlling the surface radius is:
https://github.com/pioneerspacesim/pioneer/blob/3fed83c2c660d27d5baa27e6836fb2ace90e9413/src/galaxy/StarSystemGenerator.cpp#L703

Simplified: radius = pow(mass, 0.5));
Radius is just a function of mass with no function for varying density built in.

Surface gravity is computed in SystemBody::CalcSurfaceGravity(), the responsible line is:
https://github.com/pioneerspacesim/pioneer/blob/3fed83c2c660d27d5baa27e6836fb2ace90e9413/src/galaxy/SystemBody.cpp#L597

Now we add the radius equation into the gravity equation.
 return G * mass / pow(pow(mass,0.5), 2); 
 return G * 1    !
 Since the mass is number of earth masses the result will always be 1.0g for planets between 1 and 200 earth masses.

The surface gravity code above, I trace to commit https://github.com/pioneerspacesim/pioneer/commit/528616103eb06aefcebc8e003221a97484010098
@fluffyfreak 

A quick fix could perhaps be to lower the exponent a bit from 0.5 to something like 0.4- 0.45 just to get some variation in there. A more advanced fix would include some randomised mean density. If I'm right above that is.

zonkmachine:
https://github.com/pioneerspacesim/pioneer/blob/3fed83c2c660d27d5baa27e6836fb2ace90e9413/src/galaxy/StarSystemGenerator.cpp#L694



It looks like this file has been moved. I think it's this one:
https://phl.upr.edu/library/labnotes/standard-mass-radius-relation-for-exoplanets

fluffyfreak:
I have been summoned! 😂

Yeah, that was from when I knew things and was smart, so much time has passed since 2014 (_omfg_)

fluffyfreak:
[...]

PR #5592, opened 2023-06-24
Planet density

zonkmachine:
planetinfo

    Randomize radius
    
    * Randomize radius to +/- 20% to give the bodies some difference in surface gravity.
    * Set an absolute limit of 5 g for habitability.
    * Advice on gravity.
    * Always build an orbiting station on inhabited worlds with a larger gravity (large
      currently being defined as over 1.07 g.
    * New Planet info posts: 'Escape velocity' and 'Mean density'.
    * Search and Rescue - Flavour 4 has extra text on dialogue when target planet is more
      than 1.2 g.

Addresses https://github.com/pioneerspacesim/pioneer/issues/5581. A quick way to alleviate the issue with planets between 1 and 200 earth masses all having 1.0g surface gravity.


zonkmachine:
I still don't fly spacecraft so this hasn't been tested in that regard.

Zireael07:
Offtop: I <3 the planet name <3

zonkmachine:
> Offtop: I <3 the planet name <3

I had to google that. You're into football?

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -703,8 +703,11 @@ void StarSystemRandomGenerator::PickPlanetType(SystemBody *sbody, Random &rand)
 		sbody->m_radius = fixed::FromDouble(pow(sbody->GetMassAsFixed().ToDouble(), 0.5));
 	} else {
 		// Anything bigger than 200 EU masses is a Gas Giant or bigger but the density changes to decrease from here on up...
-		sbody->m_radius = fixed::FromDouble(22.6 * (1.0 / pow(sbody->GetMassAsFixed().ToDouble(), double(0.0886))));
+		sbody->m_radius = fixed::FromDouble(22.6 * (1.0 / pow(sbody->GetMassAsFixed().ToDouble(), 0.0886)));
```
Web-eWorks, line 706: Not anything you should address here, but if we're doing planet-generation math in double-precision we might as well drop the fixed-point round-trip - either we have deterministic double-precision IEEE floats or the effort for determinism in the rest of the codebase is worthless because we're not doing the math with deterministic operations.

zonkmachine: OK, reverted. I read that as '0.0886 is a double and double() is casting the 0.0886 to a double which it already is'.

Zireael07:
@zonkmachine Yep! (Pretty much any sport, tbh)

Web-eWorks:
> I still don't fly spacecraft so this hasn't been tested in that regard.

Regarding spaceflight, there is a bit of an implicit dependency on planets with landing zones having <= 1.0g surface gravity in the gameplay balance currently. Increasing the mean gravity of planets with spaceports (and thus mission targets) does mean that several ships currently present in the game will be unable to land or complete missions on some worlds when fully laden.

Ideally, we'd want to follow-up on this effort by introducing some sort of common information display to the BBS interface which notifies the player when a mission would take place on a high-gravity world, or at the very least warn the player if they're flying a space-only craft and the mission involves landing on *any* planet.

I don't think that's a matter that needs to be solved in this PR however, and I'd definitely prefer to tackle it alongside an attempt to redesign the BBS mission dialog interface to follow some of @nozmajner's concepts.

zonkmachine:
As I understand it, since I'm calling rand, the universe now changes. I haven't found a good way around this that isn't hacky beyond what I can stomach. And I can stomach quite a lot. Maybe cheat a bit instead and just modulate the presented gravity for now? Just let the planets size 1 - 200 earth masses stay 1.0 g under the hood but do the computations in system-view-ui.lua and present some variation to the user?

impaktor:
If we're about to change the universe, then we might as well put in some condition that guarantees any High-G planet to have an orbital. It even makes sense, since colonizing a high-G planet might require supporting also low thrust craft, or some other hand waivey thingy.  

(we do similar with Hydrogen, no station can have 0.)

zonkmachine:
A comment on the 'Escape Velocity' and ̈́Mean Density' values. I put them in there to aid in trouble shooting but I think they're good for immersion. I simply like the bling of it. Gravity is pretty unintuitive and this aids in getting a feeling for the body in front of you.

Should the computations maybe go in c++ functions like next to `SystemBody::CalcSurfaceGravity()` instead of `system-view-ui.lua`?

Web-eWorks:
> As I understand it, since I'm calling rand, the universe now changes.

This is expected behavior. In the ideal situation, you add new random rolls after all existing rolls to avoid disrupting other existing values, but it's not required (nor feasible in some cases). The only thing that needs to be done is to have this PR marked as a savegame bump, as it will not have galaxy-level compatibility with the game from before this PR. I'm quite happy to bump for this, and if I'm able to land some of my outstanding branches we'll definitely need a savegame bump anyways.

> Should the computations maybe go in c++ functions like next to SystemBody::CalcSurfaceGravity() instead of system-view-ui.lua?

I'd be fine with that, escape velocity at the very least would be useful to have for autopilot-planning code.

Regarding station generation - you should only add an orbital around a high-gravity world which has a surface starport. Adding an orbital to *every* high-gravity world is probably not a good idea. :smile:

zonkmachine:
> Regarding station generation - you should only add an orbital around a high-gravity world which has a surface starport. Adding an orbital to _every_ high-gravity world is probably not a good idea. 😄

Right, I actually did it general on purpose in order to get bases all over the place while coding this but apparently this part of the code seem to only be concerned with inhabited places so I was a bit disappointed at no bases on the largest planets. Also I selected for g when the value given is m/s2. Oopsie! Anyway, now it seem to work.
Anyway, bases only seem to pop up around already inhabited worlds. The high g worlds should have lesser population.

high-g-planet4
high-g-planet2
high-g-planet

zonkmachine:
> The high g worlds should have lesser population.

I'm looking at fixing this here but I should probably try and limit the PR as the topic could easily be all over the codebase. Maybe  the span of the density can be tweaked a bit but it looks like it's pretty much useful as it is right now. On the low end I've seen planets presented as 'rocky planet' with a mean density of ~2ton/m3 and that is similar to [Ganymede and Callisto](https://en.wikipedia.org/wiki/Galilean_moons#Members) and probably a good low for a rocky planet.

I'm looking at altering the Search and Rescue code to include gravity but not touching any of the other modules right now, for brevity.

Only simple dialogue changes this time like:
Q - `Why don't you do it yourself?`
A - `The planet is 1.4 g.`
A - `We don't have any other ships that can handle the 2 g.`

zonkmachine:
Example mission dialog:

Low - Mid gravity planet
highgravity11

High gravity planet
highgravity12

Web-eWorks:
> I'm looking at altering the Search and Rescue code to include gravity but not touching any of the other modules right now, for brevity.

That's fine, my original thought was to introduce a "mission parameters" UI element into the chat form which would display "legal disclaimers" e.g. warnings to the player that they should expect combat or high-gravity worlds. This concept however is definitely a task for another PR, and if you can fit the gravity warning in naturally then go for it!

zonkmachine:
English help! Which line is best? 
 1 "The gravity on {planet} is {gravity} so please take this into consideration."
 2 "The gravity on {planet} is {gravity}, please take this into consideration."
 3 "The gravity on {planet} is {gravity}. Please take this into consideration."


Time for flight testing?

zonkmachine:
I was thinking a message in the Advice module could be motivated. How about the Jimmy Cliff themed "The harder the g, the harder the fall", "Make sure you know the surface gravity of a planet before you even think of landing on it."

impaktor:
> English help! Which line is best?

2 or 3. Not 1.

Btw, you could look at how "Set as navigation target" is implemented, as it's added to all missions by default, without touching the mission modules, if I remember correctly.

zonkmachine:
> 2 or 3. Not 1.

OK. I settled for 3.



> look at how "Set as navigation target" is implemented, as it's added to all missions by default,

OK, something like "Show object info"? Showing the object info from the 'System Map/System Overview'?

zonkmachine:
targetinfo1

@impaktor Is this something like what you had in mind? (rudimentary implementation)

impaktor:
> @impaktor Is this something like what you had in mind? (rudimentary implementation)

I wasn't thinking of anything in particular, beyond "avoid adding strings to every module".

zonkmachine:
> > @impaktor Is this something like what you had in mind? (rudimentary implementation)
> 
> I wasn't thinking of anything in particular, beyond "avoid adding strings to every module".

OK, I understand, but In reference to `you could look at how "Set as navigation target" is implemented`.
I tried it out above and wonder if it's heading in the right direction?

zonkmachine:
**No population over 5 g**
I looked around the net and there seem to be a consensus that over 5 g is not something we can live in. Technology a thousand years from now not taken into account. I don't know if there are any planets over 5 g that aren't gas giants though. Anyway, there is a hard limit now from population at 5 g.
I think the population should drop on planets after 1.2 g until there is only a few mining machine repair guys, scientists, and perhaps military, left at maybe 3 g.
Maybe add 'grav suits' and 'advanced grav suites' as commodities and perhaps you could buy them for the ship and crew to allow some well paying missions in higher gravity, although I think the gravity stuff should just be something to spice up the game so if references to 'g' becomes overpowering you just bump up the limits a bit and that should be enough.

I'm going to await further review now before continuing on this PR.

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -1369,9 +1372,14 @@ void PopulateStarSystemGenerator::PopulateStage1(SystemBody *sbody, StarSystem::
 			sbody->m_population = fixed(1, 100000) + fixed(starportPopRand.Int32(-1000, 20000), 1000000000);
 			outTotalPop += sbody->m_population;
 		} else if (sbody->GetType() == SystemBody::TYPE_STARPORT_SURFACE) {
-			// give surface spaceports a population between 80000 and 250000
-			sbody->m_population = fixed(1, 10000) + fixed(starportPopRand.Int32(-2000, 15000), 100000000);
-			outTotalPop += sbody->m_population;
+			// No permanent population on gravities larger than about 5 g
+			if (sbody->CalcSurfaceGravity() > 50) {
```
zonkmachine, line 1376: instead of '50' this should be saved in a constant variable somewhere. Same for the hard coded surface gravity limits.

zonkmachine: src/gameconsts.h

Web-eWorks: I generally consider gameconsts appropriate for "facts" about reality like the gravitational constant. This seems like something that'd be better in a galaxy generation related header file, however gameconsts is an acceptable place to put it.

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -705,6 +705,9 @@ void StarSystemRandomGenerator::PickPlanetType(SystemBody *sbody, Random &rand)
 		// Anything bigger than 200 EU masses is a Gas Giant or bigger but the density changes to decrease from here on up...
 		sbody->m_radius = fixed::FromDouble(22.6 * (1.0 / pow(sbody->GetMassAsFixed().ToDouble(), double(0.0886))));
 	}
+	// randomize radius or the surface gravity will be consistently 1.0 g for planets in the 1 - 200 EU range.
+	sbody->m_radius = fixed::FromDouble(sbody->GetRadiusAsFixed().ToDouble() * ( 1.2 - (0.4 * rand.Double()))); // +/-20% radius
+
```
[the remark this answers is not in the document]
zonkmachine: Quick test. It didn't improve anything in the game. I think it could be an interesting idea for a future rewrite but here and now it's just going to make the code messier with no apparent gain.

I stuck this line in there after metallicity was set.

```cpp
	// randomize radius to take metallicity into account
	sbody->m_radius = fixed::FromDouble(sbody->GetRadiusAsFixed().ToDouble() *
 		( 1.1 - (0.2 * sbody->GetMetallicity()))); // +/-10% radius
```

src/galaxy/StarSystemGenerator.cpp
zonkmachine: Now sits in StarSystemGenerator.cpp as it had a bunch of other variables defined in the top and it's only used there.

src/galaxy/StarSystemGenerator.cpp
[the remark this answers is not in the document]
Web-eWorks: If we're going to do that, it'd probably be better to tie it all the way into body formation/distribution; e.g. minor moons / outer planets might be more icy than rocky which would affect their mass and radius together. Either mass + random density could inform metallicity, or pre-calculated metallicity could inform mass + density.

data/lang/module-advice/en.json
```diff
@@ -7,6 +7,14 @@
     "description": "",
     "message": "Know when to leave town"
   },
+  "ADVICE_11_BODYTEXT": {
+    "description": "",
+    "message": "Make sure you know the surface gravity of a planet before you even think of landing on it."
```
zonkmachine, line 12: Usual disclaimer. If your not feelin' it, I'll drop this advice. ;)

src/galaxy/StarSystemGenerator.cpp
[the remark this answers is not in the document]
Zireael07: @Web-eWorks good idea but probably out of scope for this PR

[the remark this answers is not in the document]
zonkmachine: I had to redo the tests above as I had borked the volume/density calculations. The conclusion is the same. It'll take some more consideration to rewrite the planet generation.

> or pre-calculated metallicity could inform mass + density.

This is what I'm curious to look into first. I'm thinking of summing up my thoughts on this in a forum post.

zonkmachine:
I added the info button embryo to the Cargo Run and Deliver Package modules for easier testing. The info is mostly not showing the right body though and the data shown is more or less random planet data. Feedback needed! Maybe drop for a later PR?

zonkmachine:
> I don't know if there are any planets over 5 g that aren't gas giants though.

No. With the current model with +/-20% radius I can change Earth surface gravity between 0.69 and 1.56 g.
Incidentally that may not be totally wrong scientifically: [Surface gravity on many exoplanets similar to that of Earth, study finds](https://www.sciencedaily.com/releases/2017/06/170601095521.htm)
The rocky exoplanet with the highest known surface gravity is [K2-38b](https://en.wikipedia.org/wiki/K2-38b) with maybe over 3 g. The margin of error of the examples I've seen is huge though but 5 g doesn't seem unrealistic, just uncommon.

With +/- 30% the surface gravity would be 0.59 - 2.03 g. That would work if the distribution would have a distribution centering more around 1 g. Like the snippet below calling rand twice. Testing it now and it looks promising.

```cpp
	sbody->m_radius = fixed::FromDouble(sbody->GetRadiusAsFixed().ToDouble() * ( 1.3 - (0.3 * (
          rand.Double() + rand.Double())))); // +/-30% radius
```

zonkmachine:
Ready for review and test. Please note. There are quite a few custom systems in the game and they still have the issue with an overabundance of 1.0 g planets. These needs to be fixed up manually and that this is out of scoop for this PR.

Web-eWorks:
I'm working on a visual editor for custom systems in my available Pioneer time and noticed some issues and inconsistencies with how custom systems are loaded and handled. Unless this branch is a blocker for other efforts of your own, I'll most likely combine efforts and merge this and my branch on similar timeframes to take advantage of the fact that both will require savegame bumps. I'm expecting it to take anywhere between a week to a month to land that branch.

zonkmachine:
> Unless this branch is a blocker for other efforts of your own, I'll most likely combine efforts and merge this and my branch on similar timeframes to take advantage of the fact that both will require savegame bumps. I'm expecting it to take anywhere between a week to a month to land that branch.

Sounds good. It blocks nothing on my side.

src/LangStrings.inc.h
```diff
@@ -300,6 +300,8 @@ DECLARE_STRING(N_EARTH_RADII)
 DECLARE_STRING(N_SOLAR_RADII)
 DECLARE_STRING(SURFACE_TEMPERATURE)
 DECLARE_STRING(SURFACE_GRAVITY)
+DECLARE_STRING(ESCAPE_VELOCITY)
+DECLARE_STRING(MEAN_DENSITY)
```
Web-eWorks, line 304: By the by, you only need to `DECLARE_STRING` here if you're using the string from C++ code as `Lang::ESCAPE_VELOCITY`. Lua has a separate mechanism to load translation strings. I appreciate it being declared for completeness' sake, but I'd like to move description generation entirely into Lua anyways.

Web-eWorks:
> I'm expecting it to take anywhere between a week to a month to land that branch.

Exactly one month later, the editor part of the system editor is finished. I have a few outstanding issues in #5622 before I can merge it, but once those are complete I'll be merging this PR as well. :tada:

src/LangStrings.inc.h
zonkmachine: OK, I think I get it. I grab a random string from that code part, `ORBITAL_PERIOD`, and it seam like it also should not be declared in `src/LangStrings.inc.h`. I can drop the changes I did to this file as they seem pointless. I find deleting things oddly satisfying. :wink:

Web-eWorks: At this juncture, I'd like to leave those strings in. If interested in deleting things, you could port `SystemBody::GetAstroDescription()` over to lua and delete the C++ version and the LangStrings defines... :wink:

To get you started, the lang strings for all of the stars could be handled by this very simple code, as the string versions of the enum so happen to be exactly the same as the translation keys:
```lua
if sbody.superType == "STAR" then
    return lcore[sbody.type]
else
    -- ...
end
```

Web-eWorks (approved):

Web-eWorks:
@zonkmachine if you're able to rebase on master and give this a quick playtest, this PR should be good for merge.

zonkmachine:
> @zonkmachine if you're able to rebase on master and give this a quick playtest, this PR should be good for merge.

OK. Rebased without any problems and I've done a very quick playtest.

PR #4316, opened 2018-03-10
Add Easter egg advert, (shown on every 50th station)

impaktor:
Back in the day, my computer couldn't play Duke Nukem 3D, so I'm making up for it now. As a good nerd, I research all secrets, references and Easter eggs on each level as I play through the game.

On episode 1, level 1, in the bathroom wall, [there's a number](https://en.wikipedia.org/wiki/867-5309/Jenny "Tommy Tutone - 867-5309/Jenny"). Apparently it's "the worlds most famous telephone number" according to Urban dictionary. So why not add it on rare* occasions to pioneer.

But to me, there's [another song](https://en.wikipedia.org/wiki/Dirty_Deeds_Done_Dirt_Cheap_(song) "AC/DC - Dirty deeds done dirt cheap") that mentions a telephone number that fits better in with Pioneer, so I just had to add it, complete with references to the lyrics.

Downside: the player might wrongfully think this is a clue to something in the game, or that they can interact with it some way, but this is just a silly Easter egg.

Anyway, I just thought I'd throw this together and see if it's something we'd like in the game,  this could go the way #4153 did.

_*rare = one station in 50, the advert is "static" to the station, meaning if a station has it it's always the same_

As for BBS icon, I failed to convert the AC/DC lightning bolt to 15x15 pixel image. We can just skip an icon for now.

[youtube: Tommy Tutone- 867-5309 / Jenny](https://www.youtube.com/watch?v=ON56AKnqbog)
[youtube: AC/DC - Dirty Deeds Done Dirt Cheap](https://www.youtube.com/watch?v=26wrvblgGP0)

screenshot-20180310-195210
screenshot-20180310-195203

PR #4153, opened 2017-09-15
The mysterious unknown, highly explosive, substance Covfefe added

impaktor:
Note: Not hooked into the pioneer economy on c++ side,
as this element likes to see itself as an "outsider"

2017-09-15-104119_665x173_scrot

OK I'm not serious about this, just got the idea in my head and had to get it out. Also, prokrastinating!

....so please don't merge

PR #5593, opened 2023-06-26
SecondHand - Disappointing outcome

zonkmachine:
I just devised a method to mildly infuriate someone. Especially effective if you also had to sell the other one of whatever you tried to buy in order to fit a new one in. Chance of occurring, 1 in a 100. You don't really shop that much second hand so this would be really unlikely to happen during a game. :smiling_imp: 

secondhand2

secondhand3

### DIFF
--- a/data/lang/module-secondhand/en.json
+++ b/data/lang/module-secondhand/en.json
@@ -87,6 +87,18 @@
     "description": "Regarding installing a component on the ship",
     "message": "How do I fit it?"
   },
+  "NO_LONGER_AVAILABLE_0": {
+    "description": "If the sentence is hard to translate, just make up an excuse why the item is no longer available.",
+    "message": "Sorry, it has just been sold!"
+  },
+  "NO_LONGER_AVAILABLE_1": {
+    "description": "If the sentence is hard to translate, just make up an excuse why the item is no longer available.",
+    "message": "It seems like my colleague just sold it to another customer. Sorry!"
+  },
+  "NO_LONGER_AVAILABLE_2": {
+    "description": "If the sentence is hard to translate, just make up an excuse why the item is no longer available.",
+    "message": "I've changed my mind, it's not for sale anymore."
+  },
   "REPEAT_OFFER": {
     "description": "",
     "message": "Repeat offer"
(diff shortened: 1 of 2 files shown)


impaktor:
I like it. Gives some more depth.

Not sure it is an "Easter Egg", though, or perhaps it's just me protecting my reputation (#4316, #4153, #2611)

zonkmachine:
> Not sure it is an "Easter Egg", though,

No. Easter Eggs are usually wanted.

zonkmachine:
@impaktor Better?

Web-eWorks (approved):

data/lang/module-secondhand/en.json
```diff
@@ -87,6 +87,14 @@
     "description": "Regarding installing a component on the ship",
     "message": "How do I fit it?"
   },
+  "NO_LONGER_AVAILABLE_0": {
+    "description": "If the sentence is hard to translate, just make up an excuse why the item is no longer available.",
+    "message": "Sorry, it has just been sold!"
+  },
+  "NO_LONGER_AVAILABLE_1": {
+    "description": "If the sentence is hard to translate, just make up an excuse why the item is no longer available.",
+    "message": "It seem like my colleague just sold it to another customer. Sorry!"
```
zonkmachine, line 96: Should it be "It seems like"?

impaktor: Yes. "He, she, it - 's' commts mit", as the Germans are taught in English-class. (Since us poor Scandinavians don't distinguished between have/has, we tend to mess those up far more common than we realize, myself included)

impaktor:
> I guess this is more or less done. Maybe someone has a good line or two to drop in here?

Highly specific text strings look weird when you see them repeatedly (granted, this is 1/100 (for now)), but too generic and the difference will not be noted.

You could just keep it simple, and add a "I've changed my mind, it's not for sale anymore".

zonkmachine:
> You could just keep it simple, and add a "I've changed my mind, it's not for sale anymore".

That's a good one. 

> (granted, this is 1/100 (for now)),

I think it should be a rather exotic thing. 1/100 is probably good.

Ready to merge.

Issue #5544, opened 2023-02-14
Wrong button for rotating in shipspinner


impaktor:
### Observed behaviour

LMB down rotates ship in ship spinner (in personal info tab, and shipmarket)

### Expected behaviour
MMB down rotates

__version__: Current master.

PR #5594, opened 2023-07-12
ModelSpinner - use middle mouse button to spin model

zonkmachine:
Use middle mouse button in model spinner.

Fixes https://github.com/pioneerspacesim/pioneer/issues/5544


src/pigui/ModelSpinner.cpp
```diff
@@ -118,7 +118,7 @@ void ModelSpinner::DrawPiGui()
 
 	const ImGuiIO &io = ImGui::GetIO();
 	bool hovered = ImGui::IsItemHovered();
-	if (hovered && ImGui::IsMouseDown(0)) {
+	if (hovered && ImGui::IsMouseDown(SDL_BUTTON_MIDDLE)) {
```
Gliese852, line 121: SDL and ImGui have different mouse button numbering orders, AFAIK, i.e. even if it works in this case, it is wrong in meaning.

Gliese852:
Also, in my opinion there is a problem with this patch - what to do for people like me - without the middle button (touchpad only) - in other places there is either the ability to twist it with the numpad, or there is a button on the panel for this.

src/pigui/ModelSpinner.cpp
zonkmachine: OK. Got it!

zonkmachine:
> what to do for people like me

Would an option in the settings work? **'No middle button present'**

zonkmachine:
> Would an option in the settings work? **'No middle button present'**

Tried it out and it seem to work well. There is now a setting option that will set the mousebutton to '0' instead. I've been wrestling with the compiler for a couple of hours which is reflected in the variable names. I'll clean that up should this function be wanted.

mousebutton

Gliese852:
@zonkmachine I like the option idea, but I think the option is very global, and it seems that it should somehow affect other places as well - Sector Map, System Map, World View. For example, a person who has a middle button activates this checkbox, but his middle button will still work, which seems odd?

impaktor:
The other MMB rotation screens have a "left click and hold"-button [to rotate], no?

Gliese852:
@impaktor WorldView no.

zonkmachine:
> @zonkmachine I like the option idea, but I think the option is very global, and it seems that it should somehow affect other places as well - Sector Map, System Map, World View. For example, a person who has a middle button activates this checkbox, but his middle button will still work, which seems odd?

Yes, I just picked the ones mentioned in the issue to try it out. I think I got a better grip of the issue now and I've pushed similar changes for the places you mention here. I don't know if there is an instance of rotation that I've missed.

impaktor:
So if no middle mouse button (option is selected), you rotate with some other mouse button? Or how does this work? What's the trade off?

Also, are there no tooltip for the other control options in the settings menu? (If there are, then this should also have, explaining it changes how to rotate ships and maps)

zonkmachine:
> So if no middle mouse button (option is selected), you rotate with some other mouse button? Or how does this work? What's the trade off?

No trade off. If you don't have a middle mouse button you go back to the old way with the left mouse button. Now also implemented so all spin actions work the same, with either middle mouse button or left mouse button.

> Also, are there no tooltip for the other control options in the settings menu?

I haven't seen any for the other options.

impaktor:
But left mouse button does other things in some of the screens, right? Like in the maps? (Pardon me being too lazy to read the source, or start the game or test the PR)

zonkmachine:
> But left mouse button does other things in some of the screens, right?

Yes but we already test for hover on. Pioneer knows if you, for instance in the Sector View where you can grab the scene and spin it, or use the button on the right side and grab that and spin it. Left mouse button both times.

> Pardon me being too lazy to read the source, or start the game or test the PR

I think that's the way do. Grabbing something and manipulating it with a mouse is a complex thing to do. It's easy to see what the PR does when you test it and compare to how it is today.

src/Input.h
```diff
@@ -245,6 +248,7 @@ class Input::Manager {
 
 	bool joystickEnabled;
 	bool mouseYInvert;
+	bool noMiddleMouseButton;
```
zonkmachine, line 251: I adopted to surrounding code here but shouldn't they be member variables, m_mouseYInvert, m_noMiddleMouseButton... etc?

Web-eWorks: Technically yes, they're this way for legacy reasons. Let me review the PR further before I make a decision as to whether it's worth converting them.

src/Input.h
```diff
@@ -194,6 +194,9 @@ class Input::Manager {
 	bool IsMouseYInvert() { return mouseYInvert; }
 	void SetMouseYInvert(bool state);
 
+	bool IsMiddleMouseButton() { return noMiddleMouseButton; }
+	void SetMiddleMouseButton(bool state);
```
Web-eWorks, line 198: Recommend this be `EmulateMiddleMouseButton` instead of `IsMiddleMouseButton`, currently it reads as though the method is querying whether the middle mouse button is down.

src/ship/ShipViewController.cpp
```diff
@@ -272,7 +272,8 @@ void ShipViewController::Update()
 	Pi::input->GetMouseMotion(mouseMotion);
 
 	// external camera mouselook
-	bool mouse_down = Pi::input->MouseButtonState(SDL_BUTTON_MIDDLE);
+	const int mouseButton = (Pi::input->IsMiddleMouseButton() ? SDL_BUTTON_LEFT : SDL_BUTTON_MIDDLE);
+	bool mouse_down = Pi::input->MouseButtonState(mouseButton);
```
Web-eWorks, line 276: I'm unwilling to accept this as-is, given that the left-mouse button is overloaded to fire weapons while rotating the ship (in PlayerShipController IIRC). This means that as soon as the player starts firing in combat, their view is going to rotate wildly with every correction of the mouse while aiming.

Web-eWorks:
> > Would an option in the settings work? **'No middle button present'**
> 
> Tried it out and it seem to work well. There is now a setting option that will set the mousebutton to '0', the old value, instead. I've been wrestling with the compiler for a couple of hours which is reflected in the variable names. I'll clean that up should this function be wanted.
> 
> mousebutton

The "best" way to handle this IMO is to have this checkbox toggle on an additional code-path which looks for e.g. alt-LMB / ctrl-LMB and handles that as a middle-mouse press in all screens. The choice of key should most likely be configurable by the user as a binding, and all callsites which use MMB pan/rotate should be updated for this option.

I believe this is how Mac software usually handles pan/rotate when both left+right mouse button have semantic actions already (though without the configurable modifier key). The advantage to this approach is it allows screens with already-heavily overloaded mouse actions (e.g. WorldView) to support interaction without MMB while not sacrificing the ability of the user to quickly perform important actions in combat or flight.

Web-eWorks (changes requested):
There are several regressions with this option enabled:
- You can no longer left-click on systems in the Sector map to select them.
- Every time you pan in the System map you deselect the current selected object (changes tree visibility in the outliner and hides the info window).
- Same with the System atlas map, panning deselects the currently selected object.
- As mentioned in a separate comment, it becomes impossible to aim while firing in combat.

zonkmachine:
> > Tried it out and it seem to work well. There is now a setting option that will set the mousebutton to '0', the old value, instead. I've been wrestling with the compiler for a couple of hours which is reflected in the variable names. I'll clean that up should this function be wanted.
> > mousebutton
> 
> The "best" way to handle this IMO is to have this checkbox toggle on an additional code-path which looks for e.g. alt-LMB / ctrl-LMB and handles that as a middle-mouse press in all screens. The choice of key should most likely be configurable by the user as a binding, and all callsites which use MMB pan/rotate should be updated for this option.
> 
> I believe this is how Mac software usually handles pan/rotate when both left+right mouse button have semantic actions already (though without the configurable modifier key). The advantage to this approach is it allows screens with already-heavily overloaded mouse actions (e.g. WorldView) to support interaction without MMB while not sacrificing the ability of the user to quickly perform important actions in combat or flight.

I was apparently way less done here than I thought. I'll look into using the alt/ctrl key workaround instead. In any case people with no middle button need a workaround.

> But left mouse button does other things in some of the screens, right?

Now I understand your question.

Thank you all for your reviews.

src/ship/ShipViewController.cpp
Gliese852: This can probably be a problem if you have a mouse without a middle button or wheel, but in the case of a touchpad, it is pretty convenient to shoot with a spacebar.

Web-eWorks: Certainly, but the expectation for most people playing on desktop is they fire with a mouse button click. Spacebar for most people is mentally associated with "go up" rather than "shoot", especially in 6-DOF style games.

impaktor:
I think the aim of this PR is great: to unify MMB for spinning, and an checkbox option for those without MMB, to then use modifier-key + LMB, for instance. But as noted in the three last posts above, there's more to do before this PR is ready for review.

Hope @zonkmachine finds time for this again at some point & rebase to prevent bitrot.

zonkmachine:
> Hope @zonkmachine finds time for this again at some point & rebase to prevent bitrot.

Sorry for the late reply. I'm busy with real world stuff and won't have time to look into this anytime soon. I unfortunately failed to rebase it on current master. If someone want's to pick up this PR they are most welcome.

src/Input.h
zonkmachine: Fixed in 4b49af24bd2dcea54f889c9dcd9f0c83ab3f9f2b

zonkmachine:
> Hope @zonkmachine finds time for this again at some point & rebase to prevent bitrot.

Done. I'm looking into the feedback from the review.

zonkmachine:
>     * As mentioned in a separate comment, it becomes impossible to aim while firing in combat.

Maybe activate this only when landed/docked?

mwerle:
Having just read my way through this (although not tested yet, sorry), how about:

<some modifier> + <some mouse-button (left or right)>  : rotate <whatever>
<middle-mouse-button> : rotate <whatever>

ie, no need for a separate setting; the middle-mouse-button, if a user has one, is simply an additional alternative for the standard modifier+button option.  

I guess Mac users have it even worse in that they don't even have a RMB.. (always found that a bit weird - "let's make it easier for users and remove the confusing right-mouse-button. Oh, hmm, we actually need more button functionality on the mouse so let's make them press a key on the keyboard to emulate a RMB instead..")

Eventually I guess the mouse needs to become a "first class citizen" input device and be configurable in the key bindings just like keys/joysticks can be, with sane defaults. For example, swapping LMB/RMB for left-handed users (unless the framework already takes care of that with OS settings?)

mwerle:
> If you want to adopt this PR, or close it with another solution, please do. My computer broke down and I'm busy with life in general so I may not touch this in along time. 😉

That sucks :(

But yeah, I can try to finish this PR..  I guess the main question to all is : how do we want to unify spinning things in Pioneer? A quick bit of research indicates the MMB is used by most CAD software to rotate. However, they usually don't have a fallback. In this case, we'd like a fallback using, probably, a modifier and the LMB (to cater for Mac users), although I guess it ought to be user-configurable.

So we have 2 types of spinny things - ones in which the LMB already does something (Sector View, World View), and where it doesn't (the various displays allowing the user to spin the ship model). Do we really want to require a modifier for spinning the ship model as well?

Also, some of the views could do with using additional mouse buttons to "do" things, mainly the map screens could use mouse buttons (possibly with modifiers) in order to drag the view, not just spin it.

Summary of current spinny things:

<details>

----

Ship model spinning (New Game, Info View, Station View)

LMB (hold) - spin ship
MMB - nothing
RMB - nothing

---

World View:

MMB - move view
RMB - move ship (rotate)
LMB - fire weapon (if RMB also held down)

----

Sector View:

LMB (click) - select system
MMB (hold) - spin view
RMB - nothing

----

System View:

Orrery:

LMB (click) - select object
MMB (hold) - spin view
RMB - nothing

Details:

LMB (click) - select object
MMB (hold) - drag view
RMB - nothing

----

<details>

bszlrd:
I wouldn't cater to Mac users frankly. Why put limitations in because some manufacturer of overpriced hardware who can't put in the effort to design a proper mouse? (especially since they were instrumental in popularizing it in the olden times) Or for users who won't buy one. Sorry, I'm a bit salty about that, but I really dislike that kind of needless oversimplification, which hurts a lot of users because of a few.
Also, we don't have a mac build for quite a while now, nobody stepped up to try to sort that out, and I kinda suspect that the more time passes, the more technical debt is needed to be fixed for it to happen anyway. 
I'd rather put it on RMB, than having a modifier key, at least not exclusively. RMB and MMB, if we want to cater to cad users. If RMB is going to be spin on every screen, I guess there should be a filtering, so if you rmb on something, it will bring up the RMB menu, and if you do it on an empty space, then it will spin if you drag.

I agree that ideally it is user configurable. But how complicated that would be?

mwerle:
> I wouldn't cater to Mac users frankly. 

Ah, I didn't realise we no longer have Mac builds.. figured Mac was basically Linux from a dev perspective.

> Why put limitations in

I wasn't proposing limitations, just additional functionality.

Ie:

MMB - spin stuff

Modifier + LMB - also spin stuff - useful for people without MMB in gerneral, which just so happens to cater to Mac users as well. No need for a "don't have a MMB" option. And probably much easier to get right than trying to detect if the mouse is currently over a selectable screen object or an "empty" space.

> I agree that ideally it is user configurable. But how complicated that would be?

Not very.. our general input system already has an optional user-configurable secondary binding for every action.

mwerle:
.. ok, so I was looking at unifying all spinning by adding a `Input::Manager::IsMouseRotatePressed()` function (which would then pull the actual configuration from a set of settgings), but, of course, our code-base is never so neat. It turns out the `ModelSpinner` uses `ImGui`, and it seems to be blocking the `Input::Manager` input system..  *le sigh*

I guess for now I can duplicate the `IsMouseRotatePressed()` function for `ImGui` as well, but, longer-term, what is/should be the plan?

mwerle:
Added some fixes and changes and made a new PR as I can't push to the original branch..

Please continue discussion in #5981

zonkmachine:
Closed this one. Superseded by https://github.com/pioneerspacesim/pioneer/pull/5981

Issue #5287, opened 2021-10-01
Some issues around exiting the Lua console


zonkmachine:
### Observed behaviour

I open the Lua console with 'L' and exit it with 'Esc'.

When I 'Escape' out of the Lua Console it takes me to the Settings Menu if I was in the world view or if I'm in another view it brings me to the world view.

### Expected behaviour

I definitely think that Closing the Lua console shouldn't bring up the Settings Menu. I also think that it shouldn't bring you to the world view if you started somewhere else.

I think what's needed here is some more of the same tlc provided by @Web-eWorks in 493e258c52bd1f0a4795ee46954a011d9ada81cd

### Steps to reproduce

Enter the Lua console and exit it with the 'Esc' Key.

__My pioneer version (and OS):__

Ubuntu Mate 21.04 and 21.10
On master at 65c3784f0636b283ec15ba761daa0062e4f628c7

impaktor:
Slightly related, if anyone want some heavy development tools, there's now a [write up](https://pioneerwiki.com/wiki/Development_Tools#Editor_integration) on the wiki by @Gliese852 on how to send lua code to the game from your code editor, to a running pioneer instance.

PR #5597, opened 2023-07-20
Fix issue on quitting Lua console

### LINKED ISSUE #5598: Lua console doesn't handle back keyboard focus nicely
Issue where the Lua Console won't give focus back to the keyboard after exiting.

As mentioned in https://github.com/pioneerspacesim/pioneer/issues/5287

### Observed behaviour and steps to reproduce

Open the Lua console and exit by pressing 'escape'. Open the Lua console again with a keyboard shortcut. You now need to press twice for the action to take place.

on current master (0f417f2810752fa547a92a24ce7c3f0b6b0f9931)
First noticed this on 65c3784f0636b283ec15ba761daa0062e4f628c7 and I suspect it goes back to df10b11def1bba8c157a4cb735f3438d473ff855
@Web-eWorks

zonkmachine:
Fixes https://github.com/pioneerspacesim/pioneer/issues/5287, fixes #5598.

Escaping out of the Lua console on keypress created issues since the key was then released in another window and escape key release action performed. Fixed by exiting Lua console on escape key release instead of press.

### DIFF
--- a/src/lua/LuaConsole.cpp
+++ b/src/lua/LuaConsole.cpp
@@ -69,7 +69,7 @@ REGISTER_INPUT_BINDING(LuaConsole)
 void LuaConsole::SetupBindings()
 {
 	toggleLuaConsole = m_inputFrame.AddAction("BindToggleLuaConsole");
-	toggleLuaConsole->onPressed.connect(sigc::mem_fun(this, &LuaConsole::Toggle));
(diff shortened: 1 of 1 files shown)


Web-eWorks (approved):

Web-eWorks:
Thanks @zonkmachine! Appreciate you digging into the issue!

PR #5599, opened 2023-08-05
Upgrade IMGUI to version 1.89.8

Web-eWorks:
This PR is a simple bog-standard upgrade, with little to no enablement of new features (nothing particularly exciting was implemented between v1.83 -> v1.86 other than the ID stack debug tool). It exists mostly because I'm playing with more advanced IMGUI code in another branch unlikely to see the light of day for a while, but this way the master branch can benefit too.

IMGUI v1.87 introduces a new method for applications to process input, which is a more complex change to the SDL backend than I wish to tackle today. I'll most likely attempt to fully upgrade to v1.89.8 (the current version) in another PR soon.

The most important part of this PR is that I've separated out Pioneer's changes to the IMGUI source into separate git patches stored in-tree, so that future upgrade work can (hopefully) be as simple as overwriting the IMGUI files and applying the patches overtop. Before this change, an IMGUI upgrade involved manually porting over all of the changes we've made to new versions of the library in a very time-consuming and haphazard manner.

I've tested this myself by going through all of the menus and taking a three-minute test flight, but if a brave soul would be interested in doing a more extensive test to see if there are any regressions (e.g. language handling, etc.) I'd appreciate it.

## Updated:

I've further upgraded to ImGui v1.89.8-docking. This pulls in the long-running docking+multiple viewports feature work that's been happening in ImGui for use in my editor-related efforts, as well as several new minor features and a new input/shortcuts system for ImGui. This PR includes no enablement work to turn on those new features or expose them to Lua; that will happen in future PRs so as to keep this one as simple as possible to regression test.

With this PR, we are now up-to-date relative to upstream ImGui and (unless v1.90 includes major breaking changes) future upgrades should be a bit smoother.


impaktor:
Great to hear about the patches!

(Personally, I've smashed my head on getting IMGUI + SDL to be modularized and working, and never fully gorked it)

PR #5600, opened 2023-08-06
changed up and down crosshairs

VorTechnix:
Changed "down" crosshair to a triangle and "up" crosshair to double chevron as discussed with @nozmajner on the discord.

image
image

New crosshairs as seen with Coronatrix.


Web-eWorks:
Judging by the added lines count on this PR, it appears the SVG file has not been stripped of unnecessary editor data before exporting / saving. I believe @nozmajner may remember the process for doing that in Inkscape?

nozmajner:
I think there's an option to save a stripped down svg, but can't recall.

Web-eWorks:
It appears the export size increase may just be caused by an inkscape internal formatting change. Cleaning up the document has no appreciable effect on filesize.

PR #5606, opened 2023-08-23
Model Tag improvements, fix GLTF import

Web-eWorks:
This PR makes model tags a separate object type which is affected and transformed by the model hierarchy in a way that is consistent with the hierarchy of the source file in Blender. This allows tags to be transformed and animated, while still retaining the ability to query their global model-space transform.

I've also made a refactor to how ModelBody lighting is handled, moving responsibility for calculating lighting to the camera/render pass rather than the body object itself. This will be expanded on later with future rendering improvements and is needed for clickable cockpits.

I've also made some minor tweaks to allow the game to load GLTF-format model source files as well as Collada. GLTF is set up so that Blender "text" objects can be imported as Label nodes and the orientation of the node is the same as it appears in Blender.

This branch is a prerequisite for my editor branch and will be merged somewhat quickly.


Web-eWorks:
NOTE: this PR introduces an SGM format bump due to the new Tag node, and as such any cached SGM models in your working tree after merge will need to be regenerated.

impaktor:
> NOTE: this PR introduces an SGM format bump due to the new Tag node, and as such any cached SGM models in your working tree after merge will need to be regenerated.

How will this affect a player who upgrades their pioneer installation? Do we have some way to set configuration to force-rebuild them once? Just thinking: most players probably don't read the release information or instrucitons.

nozmajner:
Aren't the models generated upon build?

impaktor:
@nozmajner last time I checked (which was years ago), @fluffyfreak made is so pioneer builds them first time you start pioneer, if they're missing. I don't remember that having been changed.

Web-eWorks:
We distribute SGM files as part of binary builds of Pioneer, built using CI. For the end-user, they'll just have to delete the old data directory as usual when we make our next release. For developers working from repo, there is a separate facility to generate SGM files in-tree using the model compiler - you'll need to run `find data/ -iname '*.sgm' | xargs rm` or equivalent to remove the old SGM files. SGM files are not generated by the game directly; if they're not present or an incompatible version, the game falls back to loading from model source files instead (e.g. Collada).

PR #5607, opened 2023-08-23
Re-architect System Map to operate on Object Tracks

Web-eWorks:
This PR redesigns the way the System Map generates and consumes Projectables to allow external systems to push (nearly) any kind of sensor contact data into the map to be rendered and inspected as icons. The SystemView is now responsible for providing the game glue code, while the SystemMapViewport is only concerned with visualizing SystemBodies and operating on the abstract interface provided by a list of Projectables.

This has a secondary benefit of being able to integrate the system map into other contexts such as an editor, in addition to improving the technical possibility of adding additional sensor-based gameplay.

There should be no (or almost no) change to the end-user experience of using the SystemView, but the developer experience is much improved.


impaktor:
Sounds fantastic! 
(I know @ecraven used to dream of playing with the sensors, however I believe that was for a gadget in WorldView.)

src/SystemView.cpp
```diff
@@ -778,117 +863,213 @@ void SystemView::Update()
...
-		if ((*s) != Pi::player &&
-			(*s)->GetType() == ObjectType::SHIP) {
+	m_objectTracks.clear();
+	m_orbitTracks.clear();
+	m_projected.clear();
 
-			const auto c = static_cast<Ship *>(*s);
-			m_contacts.push_back(std::make_pair(c, c->ComputeOrbit()));
-		}
+	// TODO: add "true" lower/upper bounds to m_zoomTo / m_zoom
```
impaktor, line 884: should this TODO be fixed?

src/SystemView.cpp
```diff
@@ -778,117 +863,213 @@ void SystemView::Update()
...
+
+	AnimationCurves::Approach(m_rot_x, m_rot_x_to, ft);
+	AnimationCurves::Approach(m_rot_y, m_rot_y_to, ft);
+
+	AnimationCurves::Approach(m_atlasPos.x, m_atlasPosTo.x, ft);
+	AnimationCurves::Approach(m_atlasPos.y, m_atlasPosTo.y, ft);
+
+	// make panning so that the zoom occurs on the mouse cursor
+	if (prevAtlasZoom != m_atlasZoom) {
+		// FIXME The ImGui method one frame out of date
```
impaktor, line 903: one frame off doesn't sound very serious, but perhaps you want to fix or leave as is?

src/SystemView.cpp
```diff
@@ -50,19 +51,245 @@ static const int MAX_TRANSITION_FRAMES = 60;
...
-void SystemView::InputBindings::RegisterBindings()
-{
-	mapViewPitch = AddAxis("BindMapViewPitch");
-	mapViewYaw = AddAxis("BindMapViewYaw");
-	mapViewZoom = AddAxis("BindMapViewZoom");
+namespace {
+	static bool too_near(const vector3f &a, const vector3f &b, const vector2f &gain, float width)
+	{
+		return std::abs(a.x - b.x) < gain.x && std::abs(a.y - b.y) < gain.y
+			// Gliese852:
```
impaktor, line 58: Just checking if we want to reference people in the source? 
(@Gliese852 has been pinged in source)

impaktor (commented):
I liked your high level write up in the PR, I wonder if it should be documented somewhere else? In the head of the header file, for our doxygen?

src/SystemView.cpp
Web-eWorks: This is a low-priority old TODO, will not be fixed in this PR.

src/SystemView.cpp
Web-eWorks: This was just to indicate the original author of the comment/method given that it's more difficult to git blame the line since I did violence to the source file. The depth-based heuristic probably needs a bit more tweaking, and I'd prefer later contributors know where to start looking if they need more details.

Web-eWorks:
I keep forgetting we technically still use doxygen. I'll try to write up a long header comment at some point to document the way the map is intended to be used and any gotchas that might need to be avoided.

impaktor:
> I keep forgetting we technically still use doxygen

I've never seen anyone actually use it. But it's there at least.

impaktor:
Thanks for the documentation strings :+1:

Issue #5237, opened 2021-08-13
Fuel Scoop not functioning


GespenstR:
### Observed behaviour
![screenshot-20210813-131828](https://user-images.githubusercontent.com/88902910/129399940-8400fa83-3bd5-4fbf-871b-8493f881ac08.png)
While flying around NN 4247 b, I tried to get close to the gas giant to get fuel, but despite being withing the parameters stated on the wiki, nothing happened. When I tried to get closer, nothing happened.

### Expected behaviour
The fuel scoop messages would have shown up during flight, and the ship would be stocked up with Hydrogen.

### Steps to reproduce
See save file here:
https://drive.google.com/file/d/1Lo2vJxyTBYQDKNXZ8wxa7a__RuRqlOxW/view?usp=sharing

I placed the ship right next to the planet in question

__My pioneer version (and OS):__
Info: ver 20210723 (0b82c65) on: Linux

joonicks:
your speed needs to be in excess of 1000m/s before it starts scooping, you seem to be just on the edge of that

nozmajner:
I just checked that with a Mola save from discord. It turns out the Mola is struggling to fly at 1km/s in 1 atm pressure. It can be done for very short periods of time though
![screenshot-20210923-220552](https://user-images.githubusercontent.com/4182678/134577241-64b51fb0-15f0-4025-a534-a427150b5589.png)
.

GespenstR:
I think a significant part of the issue was in regards to information about the Fuel Scoop itself. When I opened this ticket, the Wiki states that the requirements were 1km and .10atm. Naturally this lead to confusion when trying the scoop out. This has since been changed to better reflect the current build of the game.

I tried the scoop out again recently, and have found that it seems to work properly at around 1atm rather than the previous .10atm. This also gets you somewhat close the the planet's surface, about 200km above usually.

PR #5609, opened 2023-08-23
Fuel scooping rework

Mc-Pain:
Fuel scooping was:

1) based on random numbers
2) not aware of time acceleration
3) inefficient if rate was too high (which is still unlikely to happen without time acceleration)
4) not working properly (see https://github.com/pioneerspacesim/pioneer/issues/5237)

Instead of this, I suggest:
1) accumulate hydrogen at rate based on speed and pressure
2) replace strict speed/pressure requirement (I think it's likely being able to scoop even at sparse atmospheres even when flying fast enough)

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -1258,8 +1258,11 @@ void Ship::StaticUpdate(const float timeStep)
 				const double dot = vdir.Dot(pdir);
 				if ((m_stats.free_capacity) && (dot > 0.90) && (speed > 100.0) && (density > 0.3)) {
 					const double rate = speed * density * 0.00000333 * double(m_stats.fuel_scoop_cap);
-					if (Pi::rng.Double() < rate) {
-						LuaEvent::Queue("onShipScoopFuel", this, p);
+					m_hydrogenScoopedAccumulator += rate * Pi::game->GetTimeAccelRate();
+					if (m_hydrogenScoopedAccumulator > 1) {
+						const double scoopedTons = floor(m_hydrogenScoopedAccumulator);
+						LuaEvent::Queue("onShipScoopFuel", this, p, scoopedTons);
+						m_hydrogenScoopedAccumulator -= scoopedTons;
 					}
 				}
 			}
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -1265,10 +1265,10 @@ void Ship::StaticUpdate(const float timeStep)
 				 *
 				 * fuel_scoop_cap = area, m^2. rate = kg^2/(m*s^3) = (Pa*kg)/s^2
 				 */
-				const double hydrogen_density = 0.00000333;
+				const double hydrogen_density = 0.0002;
 				if ((m_stats.free_capacity) && (dot > 0.90) && speed_times_density > (100.0 * 0.3)) {
 					const double rate = speed_times_density * hydrogen_density * double(m_stats.fuel_scoop_cap);
-					m_hydrogenScoopedAccumulator += rate * Pi::game->GetTimeAccelRate();
+					m_hydrogenScoopedAccumulator += rate * timeStep;
 					if (m_hydrogenScoopedAccumulator > 1) {
 						const double scoopedTons = floor(m_hydrogenScoopedAccumulator);
 						LuaEvent::Queue("onShipScoopFuel", this, p, scoopedTons);
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -1257,9 +1257,17 @@ void Ship::StaticUpdate(const float timeStep)
 				const vector3d pdir = -GetOrient().VectorZ();
 				const double dot = vdir.Dot(pdir);
 				const double speed_times_density = speed * density;
-				// reference: speed > 100.0, density > 0.3
+				/*
+				 * speed = m/s, density = g/cm^3 -> T/m^3, pressure = Pa -> N/m^2 -> kg/(m*s^2)
+				 * m    T      kg         m*kg^2           kg^2
+				 * - * --- * ----- = 1000 ------- = 1000 -------
+				 * s   m^3   m*s^2        m^4*s^3        m^3*s^3
+				 *
+				 * fuel_scoop_cap = area, m^2. rate = kg^2/(m*s^3) = (Pa*kg)/s^2
+				 */
+				const double hydrogen_density = 0.00000333;
 				if ((m_stats.free_capacity) && (dot > 0.90) && speed_times_density > (100.0 * 0.3)) {
-					const double rate = speed_times_density * 0.00000333 * double(m_stats.fuel_scoop_cap);
+					const double rate = speed_times_density * hydrogen_density * double(m_stats.fuel_scoop_cap);
 					m_hydrogenScoopedAccumulator += rate * Pi::game->GetTimeAccelRate();
 					if (m_hydrogenScoopedAccumulator > 1) {
 						const double scoopedTons = floor(m_hydrogenScoopedAccumulator);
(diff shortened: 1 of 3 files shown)


impaktor:
Thanks for looking into this!

Does this PR fix the issue you link to?

Commit (structure & messages) look nice. Not sure why you would prefer (slow) division over multiplication in the last commit, though?

Mc-Pain:
> Does this PR fix the issue you link to?

Nope, since it's fixed already but there are other things I complained about

> Commit (structure & messages) look nice. Not sure why you would prefer (slow) division over multiplication in the last commit, though?

I'd prefer readability over obscurity.
`a / 300000` looks nicer than `a * 0.00000333`

How much impact slower (?) division would have in this case?
Are we sure if the compiler (gcc 13.2.1 in my case) won't optimize this?

It could replace dividing with multiplying by (1 / divisor) at compile time, and vice versa.

impaktor:
> I will remove this commit unless it is okay to improve readability at cost of speed

he-he, @Web-eWorks gets final say. 

Also, it's not clear to me (from memory when I looked at the code a few hours ago), where that 300000 comes from, should it be a named constant? Doesn't matter much since it was that way from before. I'm just thinking loud.

Mc-Pain:
> Also, it's not clear to me (from memory when I looked at the code a few hours ago), where that 300000 comes from

it comes from d66d2d14e621936529b7cc0208ca2e0eaaab5da6, where 0.00001f became 0.00000333f (scooping capacity for fuel scoop has been tripled so overall rate should remain unchanged)

Mc-Pain:
@Web-eWorks what if I replace the last commit with one replacing 0.00000333f by constant?

```
// used for calculating a fuel scoop rate
const double hydrogen_pressure = 0.00000333f
```

Web-eWorks:
Apologies for the delay in getting back around to this PR - you've probably seen what took up my time.

> what if I replace the last commit with one replacing 0.00000333f by constant?

As long as the constant has a comment indicating the unit conversion performed by employing the constant (i.e. `m/s * kg/m^3 -> kg of h2`) it's a good idea in my eyes. The main thing I'm concerned about here is that it's easy to follow the logical / physical process attempting to be modelled in this algorithm, so future contributors can understand and improve it if needed. Your current version is fairly easy to follow, but could perhaps be expressed in terms of minimum/total flow rate rather than using an intermediate speed*density product to determine whether scooping is happening.

Let me know when you've arrived on a model you're happy with and I'll review it for merge.

Mc-Pain:
Now I'm happy enough, though I was not intended to rework the model completely.

Just the thing about scooping not happening when any of (speed, pressure) is below limits.
Perhaps it would be correct if we set a limit on their product instead and remove randomization of process (I am certain that H2 rate on such gas flow would not be fluctuating much)

Otherwise, imagine if such a place on Earth with sufficiently low oxygen rate appears randomly from natural causes (i.e. not including industrial pollution or closed space with a lot of humans without air vent)

Web-eWorks:
> (I am certain that H2 rate on such gas flow would not be fluctuating much)

> Otherwise, imagine if such a place on Earth with sufficiently low oxygen rate appears randomly from natural causes (i.e. not including industrial pollution or closed space with a lot of humans without air vent)

If I remember correctly, on Earth a disproportionate quantity of O2 is concentrated in the troposphere and other elements behave the same in large gas giants, where there are striated layers of different types/mixtures of gases rather than a 100% homogeneous atmosphere.

That might be a useful thing to model in terms of variable H2 gas flow, if I'm reading your comment correctly. However, that would be a task for another follow-up PR well in the future when we have a more complex and complete atmospheric simulation with multiple atmospheric layers (and an implementation of scanning gameplay for users to become aware of atmospheric composition).

In the context of this PR, a uniform H2 flow rate controlled only by atmospheric density is fine by me.

Finally, something to look at for this or a future PR might be controlling the scoop rate by the `atmosOxidizing` SystemBody parameter, as that is intended to control whether the atmosphere is primarily comprised of hydrogen/ammonia/etc. or oxygen-based compounds. The parameter is not exactly scientific, but that's partially associated with the low number of users in the codebase.

Mc-Pain:
Can I merge this or I should wait for review then?

src/Ship.cpp
```diff
@@ -1256,10 +1256,23 @@ void Ship::StaticUpdate(const float timeStep)
...
+				 * m    T      kg         m*kg^2           kg^2
+				 * - * --- * ----- = 1000 ------- = 1000 -------
+				 * s   m^3   m*s^2        m^4*s^3        m^3*s^3
+				 *
+				 * fuel_scoop_cap = area, m^2. rate = kg^2/(m*s^3) = (Pa*kg)/s^2
+				 */
+				const double hydrogen_density = 0.00000333;
+				if ((m_stats.free_capacity) && (dot > 0.90) && speed_times_density > (100.0 * 0.3)) {
+					const double rate = speed_times_density * hydrogen_density * double(m_stats.fuel_scoop_cap);
+					m_hydrogenScoopedAccumulator += rate * Pi::game->GetTimeAccelRate();
```
Web-eWorks, line 1271: `Pi::game->GetTimeAccelRate()` is not the "full picture" when it comes to making this function update-rate independent - you should multiply by `timeStep` instead, as that is the fraction of simulated time elapsed during the update period. You may need to adjust the rate of fuel scooping, as the function is currently scooping ~60x as fast as it should.

Web-eWorks (changes requested):
Generally, unless a change is a small single-line bugfix we prefer to at least take a cursory glance at the code and test the changed mechanic to ensure it works as intended. This means that if no community members do any testing and report back their findings, testing and review usually has to wait until a developer is finished with their in-flight projects and can "switch gears" to building and reviewing the branch.

I've not yet been able to test this branch, but I can review the code at the moment - I've noticed one issue, but otherwise the code looks quite clean to me.

src/Ship.cpp
Mc-Pain: > as the function is currently scooping ~60x as fast as it should.

I don't actually see why.
Before PR, scooping rate was compared against [0; 1) random number every tick, and if the rate is bigger, a ton of hydrogen is added to your cargo. I kept this rate unchanged.

Did a quick tests:
Scoop at 100 m/s on Saturn (I'll post save later) for 5 minutes.
Atmospheric pressure: ~51.8 atm

Before scooping: 29 tons
After scooping
Old algorithm, try 1: 257 tons (228t scooped)
Old algorithm, try 2: 238 tons (209t scooped)
Old algorithm, try 3: 259 tons (230t scooped)
New algorithm, without 4394d63d3: 239 tons (210t scooped)
New algorithm, without 4394d63d3, x10 acceleration: 239 tons (210t scooped)
New algorithm, with 4394d63d3: 33 tons (4t scooped)

With changes you requested scooping became ~60x times slower

Mc-Pain: Attaching the savefile

[scooping-test.gz](https://github.com/pioneerspacesim/pioneer/files/12642288/scooping-test.gz)

Web-eWorks: I meant that you may need to increase the rate of scooping to (re-)create the intended flow of fuel. The function is currently scooping at a timestep-invariant rate, which will only function correctly so long as the game is updating at a fixed 60 ticks per second. If any change occurs to the frequency of timestep updates, the rate of fuel will change in a manner that is not consistent with the atmospheric density or speed.

The function should be modified to multiply by `timestep`, which you did in https://github.com/pioneerspacesim/pioneer/commit/4394d63d3a3554a0915d1cade9bc4e81a544668f, *and* the rate adjusted upwards by exactly 60x. I specifically called out that it was scooping 60x as fast as you would expect if you tried to do the math by hand using the existing constants in the code.

Mc-Pain: > If any change occurs to the frequency of timestep updates

Is it intended to change?

Web-eWorks: > Is it intended to change?

Assume yes. There's a proposal on IRC right now for adaptive substepping to improve ships crashing into the ground at high time acceleration. Whether or not we intend to change the timestep however, the correct way to express the algorithm is in terms of the timestep value.

Web-eWorks (approved):

Web-eWorks:
Looks good to me, apologies for the merge delay.

Issue #5285, opened 2021-09-29
-startat argument broken


zonkmachine:
### Observed behaviour
Starting with the command line argument `-startat` doesn't work. It fails in StationRefueling.lua

```
Fatal: [string "[T] @modules/StationRefuelling/StationRefuell..."]:34: attempt to compare nil with number
stack traceback:
	[string "[T] @modules/StationRefuelling/StationRefuell..."]:34: in function 'cb'
	[string "[T] @libs/Event.lua"]:34: in function '?'
	[string "[T] @libs/Event.lua"]:185: in function <[string "[T] @libs/Event.lua"]:180>Fatal: Frame instance deletion outside 'DeleteFrame' [0]
corrupted double-linked list
Aborted (core dumped)
```

### Steps to reproduce

Start with the `-startat` argument, either empty to start at Mars or with a systemPath argument.
`./pioneer -startat`

__My pioneer version (and OS):__
OS: Ubuntu Mate, 21.04 and 21.10
I've seen this on the latest master (65c3784f0636b283ec15ba761daa0062e4f628c7) and in release-210203.

zonkmachine:
I don't see a clear connection between Pioneer/Lua, the gdb backtrace and the Valgrind output.

d3rp:
On a debug build (g++-12) the culprit seems to be [UpdateStarSystemCache](https://github.com/pioneerspacesim/pioneer/blob/b193b8e24fdef79a9b04f3f78ba601c19dac529b/src/Space.cpp#L718) and the assert hits a nullptr for `sec`.

Web-eWorks:
I'll take a look at this after I get done with the editor branches I'm working on - I touched startup-related code recently and probably broke something at some point.

zonkmachine:
There are a number of glitches that are related that I find a bit tricky to trouble-shoot without being able to launch a specific system from the command line. It's still possible to modify the start menu and swap with the system you want but some systems seem to not launch when you do it this way. Some examples:

1. You should be able to omit the last two arguments (systemIndex and bodyIndex) to **SystemPath.New()** but this doesn't seem to work when modifying **/data/pigui/modules/new-game-window/class.lua**. Rondel is the only case I've find in the code that doesn't use all five arguments and it works, right? (I haven't tested Rondel myself..).
<img width="657" height="144" alt="Image" src="https://github.com/user-attachments/assets/16c1aa04-8198-4d64-bf61-1d59b00eeb8f" />

[...]

PR #5610, opened 2023-08-25
Disable runtime floating point exceptions

Web-eWorks:
Clang 15+ / GCC 13+ leave us no choice but to disable floating point exceptions. Their auto-vectorization optimizations load invalid values and will cause FPEs even when fp-strict is enabled.

Reference #5601.

@d3rp please confirm this PR fixes the issue on GCC 13.

### DIFF
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -870,6 +870,12 @@ static void OnPlayerDockOrUndock();
 void GameLoop::Start()
 {
 	PROFILE_SCOPED()
+
+	// NOTE: this is here because Clang 15+ and GCC 13+ ignore fp-model when
+	// generating vectorized/optimized code and will happily perform exception-
+	// raising operations on the contents of uninitialized or aliased memory
+	OS::DisableFPE();
+
 	// this is a bit brittle. skank may be forgotten and survive between
 	// games
 	Pi::input->InitGame();
--- a/src/ship/PlayerShipController.cpp
+++ b/src/ship/PlayerShipController.cpp
@@ -494,9 +494,7 @@ void PlayerShipController::StaticUpdate(const float timeStep)
 {
 	// call autopilot AI, if active
 	if (m_ship->AIIsActive()) {
-		OS::EnableFPE();
 		m_ship->AITimeStep(timeStep);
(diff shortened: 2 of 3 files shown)


impaktor:
I guess we could also reference:
https://github.com/llvm/llvm-project/issues/63708

d3rp:
I built this PR on g++-12, g++-13 and clang++ (15.0.7) and all of them could start from Mars without crashing. Crashing was the result of the previous bug. Seems good to me!

There was an intermittent bug on the UI that shows the starting details after pressing "New Game". The starting cash and name were positioned on top of each other, but I couldn't replicate this. It is probably due to the design of drawing stuff in the upper left corner and repositioning them on their supposed positions, as that behavior shows as a quick flicker upon opening the scene (pressing "New Game")

zonkmachine:
> There was an intermittent bug on the UI that shows the starting details after pressing "New Game". The starting cash and name were positioned on top of each other, but I couldn't replicate this

This is fixed in https://github.com/pioneerspacesim/pioneer/pull/5612

d3rp:
Btw, I'm not sure if it's connected or it is a previously creeped in bug, but starting the game with `--start-at` a.k.a. `-sa` will crash on my system.

On a debug build (g++-12) the culprit seems to be [`UpdateStarSystemCache`](https://github.com/pioneerspacesim/pioneer/blob/b193b8e24fdef79a9b04f3f78ba601c19dac529b/src/Space.cpp#L718) and the assert hits a nullptr for `sec`.

As `-sa` skips all intros and jumps straight to Mars, I guess at some point the initiation routine got changed and the calls are in wrong order.

zonkmachine:
> Btw, I'm not sure if it's connected or it is a previously creeped in bug, but starting the game with `--start-at` a.k.a. `-sa` will crash on my system.

Reported here: https://github.com/pioneerspacesim/pioneer/issues/5285 :wink:

d3rp:
> Reported here: https://github.com/pioneerspacesim/pioneer/issues/5285 😉

Thanks - though the cause of the issue is different. I'll paste my findings there and continue on that thread.

PR #5613, opened 2023-08-26
Add Model Viewer Widget

Web-eWorks:
This PR rewrites the old ModelViewer interface into a new, docking-enabled viewport widget which can be used portably as part of a larger editor, and a set of sidebar panels which no longer draw over the viewport.

A picture is worth a thousand words, or so they say. Here's the old interface:

And here's the new interface:

https://github.com/pioneerspacesim/pioneer/assets/4218491/904ad439-1605-490d-9921-a7d7c8cfbeba

The major win here is that the model viewer now renders to an ImGui window instead of the fullscreen backbuffer, which means we can use it in many different contexts, potentially even in a running game.

TODO:
- [x] Enable MSAA when working with render targets
- [x] Use font icons instead of text icons (forgot to cherry-pick the commit)

### DIFF
--- a/data/modules/MissionUtils.lua
+++ b/data/modules/MissionUtils.lua
@@ -1,4 +1,4 @@
--- Copyright © 2008-2022 Pioneer Developers. See AUTHORS.txt for details
+-- Copyright © 2008-2023 Pioneer Developers. See AUTHORS.txt for details
 -- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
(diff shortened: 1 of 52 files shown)


Web-eWorks:
MSAA has been re-enabled and this PR includes a set of bugfixes for rendering-related issues that have cropped up as part of this series of PRs. Pending further testing by interested parties, this PR is now ready for merge.

PR #5614, opened 2023-08-26
Fix docking location tags in Space Station models

Web-eWorks:
As part of #5606, model tags are now fully consistent on import, which means if they appear to be somewhere in Blender, they'll actually be there in game.

This resulted in ships trying to teleport through station walls for two reasons: first, before this change all landing pad and docking tags were pre-rotated in blender to work around the old (broken) tag loading behavior. Secondly, I forgot to update some of the tag transform callsites in SpaceStationType.cpp and it was not correctly reading the model-space position of docking locator and exit tags.

I've also added handling for `collision_port` objects which trigger "docking start" on collision. I believe this may have been a change during manual docking experiments, but for now they should be functional with automatic docking.

Updated models were provided by @Gliese852, and I'd appreciate it if @nozmajner could have a quick look to verify everything works correctly.


nozmajner:
I tried the Barnard's start, and flew to a bunch of stations: Mars High, Torvalds, Jobs, and I did not see any problems.
I noticed that if you enter from the exit side of the station then you aren't captured, only if you hit a pad. Then you get janked to the entrance. But I think that shouldn't be a problem.

PR #5617, opened 2023-08-30
realistic Rayleigh/Mie atmospheric scattering

Mc-Pain:
Experimental shaders for atmospheres on terrestrial planets.

TO-DO list:

- [x] Partial depth calculation: shadow cylinder cuts the far side of ray
- [x] Eclipse handling: use atmosDiffuse when sampling the ray instead of multiplying the overall result
- [x] Use atmospheric density in calculations
- [x] Use star luminosity. We don't want to make atmosphere bright from M-class star thousands AU away as if it was 0.5 AU away
- [ ] Add RGB matrix (For non-Nitrogen atmospheres?)

### DIFF
--- a/data/shaders/opengl/basesphere_uniforms.glsl
+++ b/data/shaders/opengl/basesphere_uniforms.glsl
@@ -32,8 +32,6 @@ layout(std140) uniform BaseSphereData {
 	Eclipse eclipse;
 };
 
-#ifdef FRAGMENT_SHADER
-
 // NOTE: you must include attributes.glsl first!
 
 // Common code to calculate the diffuse light term for a planet's surface
@@ -49,6 +47,8 @@ void CalcPlanetDiffuse(inout vec4 diff, in vec4 color, in vec3 L, in vec3 N, in 
 	float clampedCosine = (nDotVP + 0.5 * clamp(1.0 - nnDotVP * 4.0, 0, 1) * INV_NUM_LIGHTS);
 	diff += color * uneclipsed * 0.5 * clampedCosine;
 }
+
+#ifdef FRAGMENT_SHADER
 
 // Common code to calculate the specular light term for a planet's surface
 // L: light -> surface vector (normalized)
--- a/data/shaders/opengl/basesphere_uniforms.glsl
+++ b/data/shaders/opengl/basesphere_uniforms.glsl
@@ -13,6 +13,8 @@ layout(std140) uniform BaseSphereData {
 	float geosphereAtmosFogDensity;		// TODO documentation
 	float geosphereAtmosInvScaleHeight; // TODO documentation
 	vec4 atmosColor;
+	vec3 coefficientsR;
+	vec3 coefficientsM;
 
 	// Eclipse data
 	Eclipse eclipse;
--- a/data/shaders/opengl/basesphere_uniforms.glsl
+++ b/data/shaders/opengl/basesphere_uniforms.glsl
@@ -4,14 +4,25 @@
 #include "eclipse.glsl"
 
 layout(std140) uniform BaseSphereData {
-	// to keep distances sane we do a nearer, smaller scam. this is how many times
-	// smaller the geosphere has been made
-	vec3 geosphereCenter;				// TODO documentation
-	float geosphereRadius;				// (planet radius)
-	float geosphereInvRadius;			// 1.0 / (planet radius)
-	float geosphereAtmosTopRad;			// in planet radii
-	float geosphereAtmosFogDensity;		// TODO documentation
-	float geosphereAtmosInvScaleHeight; // TODO documentation
+	// To render accurate planets at any distance or scale, almost all
+	// calculations take place in "planet space", with a distance of 1.0
+	// defined as equivalent to the planet's nominal radius (planet radii).
+	//
+	// This allows planets to be rendered in view-space at any scaling factor
+	// desired, to play nicely with the depth buffer and any other content
+	// rendered (e.g. map views)
+	//
+	// The atmosphere simulation approximates Rayleigh scattering by
+	// calculating the optical density of a path through the atmosphere and
+	// assuming a constant in/out scattering factor based on the density
+	// approximation.
+
+	vec3 geosphereCenter;				// view-space center of the planet, in planet radii
+	float geosphereRadius;				// real planet radius, in meters
+	float geosphereInvRadius;			// 1.0 / (view-space radius), converts between view coordinates and planet radii
+	float geosphereAtmosTopRad;			// height of the simulated atmosphere, in planet radii
+	float geosphereAtmosFogDensity;		// atmospheric density scalar
+	float geosphereAtmosInvScaleHeight; // 1.0 / (atmosphere scale height) in planet radii
 	vec4 atmosColor;
 	vec3 coefficientsR;
 	vec3 coefficientsM;
@@ -19,3 +30,52 @@ layout(std140) uniform BaseSphereData {
 	// Eclipse data
 	Eclipse eclipse;
 };
+
+#ifdef FRAGMENT_SHADER
+
+// NOTE: you must include attributes.glsl first!
+
+// Common code to calculate the diffuse light term for a planet's surface
+// L: light -> surface vector (normalized)
+// N: surface normal
+// V: surface position relative to the unit-sphere planet
+void CalcPlanetDiffuse(inout vec4 diff, in Light light, in vec3 L, in vec3 N, in float uneclipsed)
+{
+	float nDotVP  = max(0.0, dot(N, L));
+	float nnDotVP = max(0.0, dot(N, -L));
+
+	//need backlight to increase horizon, attempts to model light propagating towards terminator
+	float clampedCosine = (nDotVP + 0.5 * clamp(1.0 - nnDotVP * 4.0, 0, 1) * INV_NUM_LIGHTS);
+	diff += light.diffuse * uneclipsed * 0.5 * clampedCosine;
+}
+
+// Common code to calculate the specular light term for a planet's surface
+// L: light -> surface vector (normalized)
+// N: surface normal
+// E: eye->surface vector (normalized)
+// power: specular power (blinn-phong)
+void CalcPlanetSpec(inout float spec, in Light light, in vec3 L, in vec3 N, in vec3 E, in float power)
+{
+	//Specular reflection
+	vec3 H = normalize(L - E);
+	//water only for specular
+	spec += pow(max(dot(H, N), 0.0), power) * 0.5 * INV_NUM_LIGHTS;
+}
+
+// E: eye->surface direction
+// scaledPos: position of the pixel in view space, divided by the radius of the geosphere
+void CalcPlanetFogFactor(inout float ldprod, inout float fogFactor, in vec3 E, in vec3 surfacePos, in float dist)
+{
+	// when does the eye ray intersect atmosphere
+	float atmosStart = raySphereIntersect(geosphereCenter, E, geosphereAtmosTopRad).x;
+	float atmosDist = (dist - atmosStart) * geosphereRadius;
+
+	// a&b scaled so length of 1.0 means planet surface.
+	vec3 a = atmosStart * E - geosphereCenter;
+	vec3 b = surfacePos;
+
+	ldprod = AtmosLengthDensityProduct(a, b, atmosColor.w*geosphereAtmosFogDensity, atmosDist, geosphereAtmosInvScaleHeight);
+	fogFactor = clamp(1.5 / exp(ldprod), 0.0, 1.0);
+}
+
+#endif
--- a/data/shaders/opengl/basesphere_uniforms.glsl
+++ b/data/shaders/opengl/basesphere_uniforms.glsl
@@ -24,8 +24,9 @@ layout(std140) uniform BaseSphereData {
 	float geosphereAtmosFogDensity;		// atmospheric density scalar
 	float geosphereAtmosInvScaleHeight; // 1.0 / (atmosphere scale height) in planet radii
 	vec4 atmosColor;
-	vec3 coefficientsR;
-	vec3 coefficientsM;
+	vec3 coefficientsR;			// coefficients for approximating the Rayleigh contribution
+	vec3 coefficientsM;			// coefficients for approximating the Mie contribution
+	vec2 scaleHeight;			// height for (R, M) in km, at which density will be reduced by e
 
 	// Eclipse data
 	Eclipse eclipse;
--- a/data/shaders/opengl/basesphere_uniforms.glsl
+++ b/data/shaders/opengl/basesphere_uniforms.glsl
@@ -39,14 +39,14 @@ layout(std140) uniform BaseSphereData {
 // L: light -> surface vector (normalized)
 // N: surface normal
 // V: surface position relative to the unit-sphere planet
-void CalcPlanetDiffuse(inout vec4 diff, in Light light, in vec3 L, in vec3 N, in float uneclipsed)
+void CalcPlanetDiffuse(inout vec4 diff, in vec4 color, in vec3 L, in vec3 N, in float uneclipsed)
 {
 	float nDotVP  = max(0.0, dot(N, L));
 	float nnDotVP = max(0.0, dot(N, -L));
 
 	//need backlight to increase horizon, attempts to model light propagating towards terminator
 	float clampedCosine = (nDotVP + 0.5 * clamp(1.0 - nnDotVP * 4.0, 0, 1) * INV_NUM_LIGHTS);
-	diff += light.diffuse * uneclipsed * 0.5 * clampedCosine;
+	diff += color * uneclipsed * 0.5 * clampedCosine;
 }
 
 // Common code to calculate the specular light term for a planet's surface
--- a/data/shaders/opengl/rayleigh.glsl
+++ b/data/shaders/opengl/rayleigh.glsl
@@ -6,15 +6,13 @@ float height(const in vec3 orig, const in vec3 center)
 	return height;
 }
 
-void scatter(out float hr, out float hm, const in vec3 orig, const in vec3 center)
+void scatter(out vec2 density, const in vec3 orig, const in vec3 center)
 {
-	float Hr = 7994;
-	float Hm = 1200;
+	vec2 baricStep = vec2(7994.f, 1200.f);
 
 	float height = height(orig, center);
 
-	hr = -height / Hr;
-	hm = -height / Hm;
+	density = -height / baricStep;
 }
 
 void findClosestHeight(out float h, out float t, const in vec3 orig, const in vec3 dir, const in vec3 center)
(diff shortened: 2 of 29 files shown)


nozmajner:
Could you post some screenshots, I'm curious.

Mc-Pain:
Found the old one but shaders has been reworked since.
изображение

I'll post actual screenshots later (this evening?)

impaktor:
Just checking: our eclipse at [Epsilon Eridani](https://wiki.pioneerspacesim.net/wiki/Places_of_interest) still works?
Also, something @Web-eWorks to check on your system-editor branch, if planets and planes and unspecified parameters are being changed.

nozmajner:
How cool is that! Never thought it is missing until I saw what's missing.
If it is a performace hog, maybe it could be useful to tie it to make it togglable in settings?

Mc-Pain:
Origin of this code:
https://www.scratchapixel.com/lessons/procedural-generation-virtual-worlds/simulating-sky/simulating-colors-of-the-sky.html

Mc-Pain:
изображение
Epsilon Eridani, eclipse

Web-eWorks:
Thanks for taking a crack at this! There are a few "problems" which are why I've not already tried to implement this before:
- Obviously, as you've mentioned, performance. Trying to brute-force Rayleigh-Mie is pretty much impossible to do outside of an offline raytracer, which is what that article was originally written for. There are plenty of papers available discussing ways to make it more performant for real-time rendering, but more on that later.
- The reference Rayleigh/Mie coefficients used by most implementations (and this one specifically) are (AFAIK) only appropriate for an oxygenated / earth-like atmosphere and will not produce visually correct results in CO2 / Methane atmospheres or in different atmospheric densities. Without some approximation to generate semi-physically-based Rayleigh / Mie coefficients there's not much point to introducing "realistic" atmospheric scattering.
- Have you verified if "aerial perspective" AKA views from space work correctly? Depending on the implementation of the Nishita model those are sometimes not handled correctly.

If you're interested in addressing the first problem while still implementing the Nishita model, you'll be interested in [this GpuGems2 chapter](https://developer.nvidia.com/gpugems/gpugems2/part-ii-shading-lighting-and-shadows/chapter-16-accurate-atmospheric-scattering) which manages to cram the necessary math down into something a little more performant.

I cannot promise this PR will be merged however, as even if performance improves on desktop our min-spec target is semi-recent laptop/integrated GPUs, and this PR would both need to address the second point above and provide this as an "experimental" toggleable option.

The other reason I can't promise to merge the PR is that we instead intend to implement `A Scalable and Production Ready Sky and Atmosphere Rendering Technique` (by Sebastien Hillaire, described [here](https://sebh.github.io/publications/egsr2020.pdf) and [here](https://blog.selfshadow.com/publications/s2020-shading-course/hillaire/s2020_pbs_hillaire_slides.pdf)) which is a physically-based model designed for real-time game rendering on everything from mobile to high-end PC. Hillaire's model provides better performance and integrates better with other atmospheric effects like cloud rendering. The implementation is not possible until required plumbing work is done in both the render API and scene rendering code sections however, so the usual caveat of "no ETA" applies.

If you are interested in getting some extremely useful code merged, starting with a way to procedurally generate at least semi-accurate Rayleigh/Mie coefficients for our various atmospheric exoplanets would be very worthwhile regardless of which rendering model gets implemented.

Mc-Pain:
There are several reasons why this PR is marked WIP.

> * Obviously, as you've mentioned, performance. Trying to brute-force Rayleigh-Mie is pretty much impossible to do outside of an offline raytracer, which is what that article was originally written for. There are plenty of papers available discussing ways to make it more performant for real-time rendering, but more on that later.

That's one of them.
Looking at the code and trying to figure out what went wrong I became aware of what this model does and how it works. Myself, without looking at the original paper.
The quality of image could be performed via increasing numSamples, but I think it could be easier to solve an integral and get a single solution with given parameters, reducing complexity from O(n^2) to O(n).

> * The reference Rayleigh/Mie coefficients used by most implementations (and this one specifically) are (AFAIK) only appropriate for an oxygenated / earth-like atmosphere and will not produce visually correct results in CO2 / Methane atmospheres or in different atmospheric densities. Without some approximation to generate semi-physically-based Rayleigh / Mie coefficients there's not much point to introducing "realistic" atmospheric scattering.

Atmosphere's color can be used for adjusting these coefficients (does it change between planets with similar composition?)

> Have you verified if "aerial perspective" AKA views from space work correctly? Depending on the implementation of the Nishita model those are sometimes not handled correctly.

That's another one but it might stuck without help of whoever wrote this shader or know what some vars mean and do.

Web-eWorks:
> but I think it could be easier to solve an integral and get a single solution with given parameters, reducing complexity from O(n^2) to O(n).

> That's another one but it might stuck without help of whoever wrote this shader or know what some vars mean and do.

Read this: https://developer.nvidia.com/gpugems/gpugems2/part-ii-shading-lighting-and-shadows/chapter-16-accurate-atmospheric-scattering - it's another implementation of the same atmospheric scattering model using an LUT instead of brute-forcing the integral. You'll have to transcribe the code from Cg if you want to crib off of it, but I recommend implementing it from the article's description instead.

Mc-Pain:
> You'll have to transcribe the code from Cg if you want to crib off of it, but I recommend implementing it from the article's description instead.

Read the whole article. Not understood where's the difference between this and mine (except for the LUP mentioned above)

Mc-Pain:
I am very suspicious of "density function along the vector" being similar to normal distribution.

impaktor:
If we include this, it might even be "too much" to put the on/off in the settings menu, might belong in debug menu instead, for developers / people who want to render a scene for screen shot. just a thought.

Mc-Pain:
Instead, I built my own model for atmospheric density.

Found out this:
1) density from outside atmosphere can be approximated: k*exp(h*b) where h = distance of closest point of the ray to geosphereCenter
2) density from inside could be multiplied by erf(t), where t is parameter from -e to e:
t = -e - no atmosphere along the ray
t = 0 - half of atmosphere along the ray (we're at the closest point)
t = e - everything is along the ray.

Pretty close to each other
pioneer-atmosphere

fluffyfreak:
This is a very exciting development!
There was, long ago, another attempt to do [atmospheric scattering](https://github.com/fluffyfreak/pioneer/tree/scattering) but it was... 😱 12 years ago...

Mc-Pain:
Pushed some optimization but it looks like I messed up some math and scattering looks unexpectedly different.

Could use some fresh look at this

impaktor:
@Mc-Pain Usually, devs don't even have time for their own code, so best to "self review" / leave it to future self.

Mc-Pain:
> @Mc-Pain Usually, devs don't even have time for their own code, so best to "self review" / leave it to future self.

well, I'll try it myself then. Got to my high-end PC, look what happens at very high sample count

Web-eWorks:
I intend to review this PR at some point in the near future, but I've not yet had a chance to get to it.

Mc-Pain:
Looks like it works now as intended...
At close orbits

изображение

nozmajner:
On my rig (Win10, nvidia 3070ti) it seems to run at 60fps at all times.
And I think it looks beautiful. Although on Earth at dusk and dawn the sun is blown out by the glow. Not sure if that's how it supposed to be:

https://github.com/pioneerspacesim/pioneer/assets/4182678/4c780104-bcff-462d-a61f-46d76772de11

Web-eWorks (changes requested):
Overall, I really like what you've done here to implement this model, even if I'm not clear on the details of how you've implemented it mathematically. I still have some reservations about the physical correctness of the model (in terms of incoming / outgoing units and getting "realistic" atmosphere behavior for a variety of atmosphere compositions) and how well it will handle being ported to a high-dynamic-range rendering pipeline, but this PR definitely gets my approval in terms of massively improving the graphical fidelity of Pioneer!

The rest of this review contains a running list of issues I've noticed while testing this PR - while I won't require every one of them to be solved 100% to merge this PR, I think each issue definitely detracts from the visual quality and believability of the effect and would like to see them addressed before merge.

-----

I'm noticing at low / grazing viewing angles the atmosphere well beyond the terminator is illuminated despite the impossibility of light reaching there without first illuminating the higher atmosphere:

image

Sol is well below the horizon here, and the illumination seems to be reasonably correct here:

image

But looking the opposite direction the atmosphere is illuminated where it should otherwise be *well* occluded by the planet.
image

This behavior is not limited to the per-vertex approach and is present in per-pixel mode as well. From further inspection, this appears to be due to the fact that the incident light computation is not handling the effects of the planet self-shadowing and is instead relying on `atmosDiffuse` to darken itself.

Performance seems to be quite good however - the effect in per-pixel mode costs ~1.5ms at 1920x1080 on my 2070 Super and will scale linearly with pixel count.

-----

I am noticing however that while the atmosphere from an aerial perspective looks fairly high-quality, the illumination on terrain and models in the distance is quite jarringly different as the `multi` and `geosphere_terrain` shaders still use the legacy lighting calculations.

Unfortunately, the options there are either to compute incident light and scattering per-pixel for all affected geometry, or to generate per-frame a three-dimensional volume lookup texture computing the optical depth / transmittance term over the view frustum. The first is doable with the existing infrastructure, the second - while much more scalable - will require a potentially major amount of changes to the existing high-level rendering infrastructure to make that LUT available.

-----

It doesn't look like there's any support for varying the Raleigh/Mie `beta` color coefficients yet, which results in a quite unrealistic sky on planets like Mars, which seems to have a much dustier and less blue atmosphere. Even on Earth, there's a very significant green tinge to the edge of the atmosphere at all times of day.

image

Changing the Rayleigh `beta` coefficient to `5.5e-6, 13.0e-6, 22.4e-6` seems to produce a more realistic sky while landed on Earth:

image

I also added a tonemapping step to the end of `rayleigh_accurate.frag` borrowed from https://www.shadertoy.com/view/slSXRW and https://www.shadertoy.com/view/4dBcD1, which gets rid of the "blown-out" sky effect:

```glsl
vec3 jodieReinhardTonemap(vec3 c){
    float l = dot(c, vec3(0.2126, 0.7152, 0.0722));
    vec3 tc=c/(c+1.);
    return mix(c/(l+1.),tc,tc);
}

void main(void)
{
//...

	// 40.0 stands as an intensity scalar to get relatively-correct brightness levels in an LDR pipeline
	specularHighlight *= vec3(40.0);
	// produces a more vibrant / saturated final image
	specularHighlight = pow(specularHighlight, vec3(1.3));

	vec3 final_color = atmosDiffuse.xyz * specularHighlight;
	frag_color = vec4(jodieReinhardTonemap(final_color), 1.0);
}
```

image

There's an alternate version which omits the `pow(specular, vec3(1.3)` and `jodieReinhardTonemap`, instead using `1.0 - exp(-final_color)` as the tonemap for a slightly brighter and more neutral / desaturated image.

```glsl
	specularHighlight *= vec3(40.0);
	vec3 final_color = atmosDiffuse.xyz * specularHighlight;
	frag_color = vec4(1.0 - exp(-final_color), 1.0);
```

image

-----

I'm also noticing the implementation here doesn't handle a light absorption term, which means sunrise and sunset lacks that characteristic "red hue". [This shadertoy](https://www.shadertoy.com/view/wlBXWK) is MIT-licensed and has a fairly clear implementation of atmospheric absorption as well as a good set of coefficients for an earthlike atmosphere if you're interested in adding that to the model.

I've used its Rayleigh coefficients above to produce the more realistic sky color you see in the screenshots.

Web-eWorks:
And how could I forget - part of the color issues are caused by the fact that the blending is taking place in sRGB space instead of linear space!

With an sRGB->Linear conversion of `uLight[i].diffuse` and a matching Linear->sRGB conversion at the end of the pipeline, the intensity scalar can be reduced down to `20.0` and you get much more natural colors - still no red hue shift when the sun is very close to the horizon though.

With the `jodieReinhardTonemap` operator:
image

With the `1.0 - exp(-final_color)` operator and no `pow(specularHighlight, 1.3)`:
image

-----

To better replicate the color grading and look of the sRGB method while blending in linear space, I set the intensity to 20.0, bumped the exponentiation term up to 1.5 from 1.3 and recorded this video (warning: 30MB!):

https://github.com/pioneerspacesim/pioneer/assets/4218491/7a948736-c0fc-48e2-80f0-7a5bf4f51f97

(Music: distant_home_v2_nonlooping.ogg, soon to be added to the Pioneer soundtrack!)

I'd be more than happy to provide a patch or push a commit to this branch implementing these changes if you're interested. With blending happening in linear space, the atmosphere will need some kind of debanding effect but that shouldn't be terribly difficult to implement.

-----

Overall I'm *extremely* happy with the visual effect achieved by this branch now that all of the puzzle pieces are starting to come together. Thank you for persevering with this effort @Mc-Pain!

Mc-Pain:
> I'm noticing at low / grazing viewing angles the atmosphere well beyond the terminator is illuminated despite the impossibility of light reaching there without first illuminating the higher atmosphere:

That definitely looks like being illuminated from behind the planet.
But it is actually illumination from *closer* layers which are still being under direct light.

> But looking the opposite direction the atmosphere is illuminated where it should otherwise be well occluded by the planet.

I suspect the same thing as above. We need to match the colors of atmosphere and terrain to resolve this.

Mc-Pain:
> It doesn't look like there's any support for varying the Raleigh/Mie beta color coefficients yet, which results in a quite unrealistic sky on planets like Mars, which seems to have a much dustier and less blue atmosphere. Even on Earth, there's a very significant green tinge to the edge of the atmosphere at all times of day.

I suggest instead of manipulating `beta` coefficients make them change different colors instead of pure RGB parts.

Web-eWorks:
> I suggest instead of manipulating `beta` coefficients make them change different colors instead of pure RGB parts.

The only way I see for that to make sense is to redefine the wavelength of light that the scattering coefficients are defined in terms of. The `beta` coefficients currently used are defined as the scattering of three extremely specific wavelengths of light (pure red, green, and blue) in specifically an Earth atmosphere (~70 N2, 21% O2, etc...) at sea level.

To represent other planetary atmospheres, we can vary the color of the sunlight and its intensity, but also need to adjust the scattering coefficients of the atmosphere to account for the different composition of gases which have a different effect on the scattering of those specific wavelengths of light.

src/galaxy/SystemBody.cpp
```diff
@@ -262,6 +262,56 @@ bool SystemBody::IsScoopable() const
...
+vector3f SystemBody::GetCoefficients(const double radius, const double atmHeight, const double baricStep) const
+{
+	float k, b, c;
+	k = computeDensity(radius, atmHeight, 0.f, baricStep);
+	b = log(k / computeDensity(radius, atmHeight, 1.f, baricStep));
+
+	float s = 1.f;
+	float sHeight = sqrt(radius*radius + s*s) - radius;
+	float c1 = exp(-sHeight / baricStep) * s;
+	float c2 = k * (error_function(s) - error_function(0));
```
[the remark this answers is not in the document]
Mc-Pain: To handle scattering, we need to do two things:
1) Calculate optical depth along the view ray (from observer to a certain point)
2) Calculate optical depth along the light ray (from point defined above to a light source)

First version of model: take `k=16` uniformly distributed points along the ray, calculate "density density" at each point and multiply by segment length, which involves a nested loop (depth=2)

To get this faster by computing density along the ray in one operation I found out that density function looks pretty like uniform distribution.
By similar digging I found out that overall density along the ray is some exponent.

So the algorithm was:
1) Get total density along the ray:
`total_dens = k*e^(b*h)`
2) Get "density ratio" between given points
`part_dens = total_dens * (erf(t1) - erf(t0))`

src/graphics/Drawables.cpp
```diff
@@ -482,7 +482,7 @@ namespace Graphics {
 
 		Graphics::MeshObject *Icosphere::Generate(Graphics::Renderer *r, int subdivs, float scale, AttributeSet attribs)
 		{
-			subdivs = Clamp(subdivs, 0, 4);
+			subdivs = Clamp(subdivs, 0, 6);
```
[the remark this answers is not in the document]
Mc-Pain: This was agreed to get rid of.

Mc-Pain:
> The only way I see for that to make sense is to redefine the wavelength of light that the scattering coefficients are defined in terms of. The `beta` coefficients currently used are defined as the scattering of three extremely specific wavelengths of light (pure red, green, and blue) in specifically an Earth atmosphere (~70 N2, 21% O2, etc...) at sea level.

Colors could be defined from here:
https://www.shadertoy.com/view/NlGXzz

Web-eWorks:
Other than the aliasing issues on the edge of the terrain (I assume this was taken with AA off...) that first screenshot looks absolutely beautiful.

fluffyfreak:
The planet ring shaders look very self contained, you could maybe do those as a separate PR just to keep this one a little smaller if you wanted too.

Mc-Pain:
> The planet ring shaders look very self contained, you could maybe do those as a separate PR just to keep this one a little smaller if you wanted too.

I'll cherry-pick this then

data/shaders/opengl/basesphere_uniforms.glsl
```diff
@@ -4,16 +4,79 @@
...
+	vec3 geosphereCenter;				// view-space center of the planet, in planet radii
+	float geosphereRadius;				// real planet radius, in meters
+	float geosphereInvRadius;			// 1.0 / (view-space radius), converts between view coordinates and planet radii
+	float geosphereAtmosTopRad;			// height of the simulated atmosphere, in planet radii
+	float geosphereAtmosFogDensity;		// atmospheric density scalar
+	float geosphereAtmosInvScaleHeight; // 1.0 / (atmosphere scale height) in planet radii
 	vec4 atmosColor;
+	vec3 coefficientsR;
+	vec3 coefficientsM;
+	vec2 baricStep;
```
Web-eWorks, line 29: These should be described with comments indicating their purpose and their values.

data/shaders/opengl/rayleigh.glsl
```diff
@@ -0,0 +1,237 @@
...
+void findClosestHeight(out float h, out float t, const in vec3 orig, const in vec3 dir, const in vec3 center)
+{
+	vec3 radiusVector = center - orig;
+	vec3 tangent = dot(dir, radiusVector) * dir;
+	vec3 normal = radiusVector - tangent;
+	h = sqrt(dot(normal, normal));
+	t = dot(tangent, dir);
+}
+
+// error function
```
Web-eWorks, line 30: This should be described in more detail, at minimum something along the lines of "error function to approximate (in/out)-scattering integral over the ray between the sample position and the light source".

data/shaders/opengl/rayleigh.glsl
```diff
@@ -0,0 +1,237 @@
...
+    opticalDepth *= predictDensityIn(radius, atmosphereHeight, h, coefficients.z, t);
+    return opticalDepth;
+}
+
+void skipRay(inout vec2 opticalDepth, const in vec3 dir, const in vec2 boundaries, const in vec3 center)
+{
+	if (boundaries.y == boundaries.x)
+		return;
+
+	int numSamples = 16;
```
Web-eWorks, line 96: I'd recommend lowering the number of samples used by the skip ray as much as possible without compromising the quality of the resulting scattering. In my testing a high sample count here did not significantly contribute to the visual quality of atmospheres.

data/shaders/opengl/rayleigh.glsl
```diff
@@ -0,0 +1,237 @@
...
+	vec2 opticalDepth = vec2(0.f);
+
+	processRay(sumR, sumM, opticalDepth, sunDirection, dir, atmosphere_minus_shadow.xy, center, diffuse, uneclipsed);
+	skipRay(opticalDepth, dir, atmosphere_minus_shadow.yz, center);
+	processRay(sumR, sumM, opticalDepth, sunDirection, dir, atmosphere_minus_shadow.zw, center, diffuse, uneclipsed);
+
+	float mu = dot(dir, sunDirection); // mu in the paper which is the cosine of the angle between the sun direction and the ray direction
+	float phaseR = 3.f / (16.f * 3.141592) * (1 + mu * mu);
+	float g = 0.76f;
+	float phaseM = 3.f / (8.f * 3.141592) * ((1.f - g * g) * (1.f + mu * mu)) / ((2.f + g * g) * pow(1.f + g * g - 2.f * g * mu, 1.5f));
```
Web-eWorks, line 233: If this uses the same Mie phase function as the planet ring PR, please convert this to a call to that function once it is merged.

src/galaxy/SystemBody.cpp
```diff
@@ -262,6 +262,42 @@ bool SystemBody::IsScoopable() const
 	return false;
 }
 
+// all input units are in km
+double SystemBody::ComputeDensity(const double radius, const double atmosphereHeight, const double h, const double baricStep) const
```
Web-eWorks, line 266: I'd strongly prefer this function be annotated with a description of both the math inside as well as its purpose.

src/galaxy/SystemBody.cpp
```diff
@@ -262,6 +262,42 @@ bool SystemBody::IsScoopable() const
...
+                double t = tmin + step * (i + 0.5);
+                double h = sqrt(minHeight*minHeight + t*t) - radius;
+                density += step * exp(-h / baricStep);
+        }
+
+        return density;
+}
+
+// all input units are in km
+vector3f SystemBody::GetCoefficients(const double radius, const double atmHeight, const double baricStep) const
```
Web-eWorks, line 287: I'd strongly prefer this function be annotated with a description of both the math inside as well as its purpose.

Web-eWorks (changes requested):
First of all, thank you for taking the time and effort to implement this as far as you have!

Regarding prior discussion on IRC about this PR's state, I'm generally willing to merge this for the upcoming release in its current condition as an experimental option. While it still requires significant further improvements to match terrain rendering with atmosphere rendering before I consider it a feature-complete replacement for our current atmosphere rendering, those improvements can be added in after its inclusion in the upcoming release.

To that end, I've pushed several commits to add "hot swapping" the used atmosphere shader from the settings menu and to clearly indicate that the Rayleigh/Mie shader is experimental.

I'd like to see the attached comments be addressed before merging, but if you consider this in a good enough state to merge for the upcoming release then I'll sign off on it.

impaktor:
Just checking: Could this PR cause crashes? My main thinking is: if someone opens a bug report, and attach output.txt, will we see they've enabled the "Experimental" option?

data/shaders/opengl/rayleigh.glsl
Mc-Pain: I'll leave 8

data/shaders/opengl/basesphere_uniforms.glsl
```diff
@@ -24,9 +24,9 @@ layout(std140) uniform BaseSphereData {
 	float geosphereAtmosFogDensity;		// atmospheric density scalar
 	float geosphereAtmosInvScaleHeight; // 1.0 / (atmosphere scale height) in planet radii
 	vec4 atmosColor;
-	vec3 coefficientsR;
-	vec3 coefficientsM;
-	vec2 baricStep;
+	vec3 coefficientsR;			// coefficients for approximating the Rayleigh contribution
+	vec3 coefficientsM;			// coefficients for approximating the Mie contribution
+	vec2 baricStep;				// height for (R, M) in km, at which density will be reduced by e
```
Web-eWorks, line 29: Note: in all literature I'm aware of, the distance over which a quantity reduces by a factor of _e_ is called [Scale Height](https://en.wikipedia.org/wiki/Scale_height). If this is in fact the atmospheric scale height for Rayleigh/Mie scattering, I'd recommend this parameter be renamed such in all cases as `baricStep` is unclear.

Web-eWorks:
> Just checking: Could this PR cause crashes? My main thinking is: if someone opens a bug report, and attach output.txt, will we see they've enabled the "Experimental" option?

Very unlikely. This PR mainly makes changes to shaders and the experimental option simply influences which shader is chosen.

Web-eWorks (approved):
@Mc-Pain please go ahead and start squashing related / iterative-work commits down into feature-level commits. Ideally all "rename" and "fix" commits related to the new rayleigh scattering code should be condensed into a small number of commits adding a major feature. Generally unrelated or standalone commits (e.g. "Add experimental label to Rayleigh/Mie scattering option") can be left as-is.

I generally consider this PR as approved, but the commit history will need to be cleaned up before merge. The sooner the better - once the history has been cleaned up I consider this PR good to merge for whoever is able to hit the button first.

Mc-Pain:
image
image
image

fluffyfreak:
Hi @Mc-Pain could you open a new bug report with this image in, where it happened and some info about your setup please?

sturnclaw:
I believe these are work-in-progress screenshots of a future improvement to the shader, not a bug with the current shader.

Mc-Pain:
> I believe these are work-in-progress screenshots of a future improvement to the shader, not a bug with the current shader.

Yes, I had nowhere else to upload those

PR #5619, opened 2023-09-02
Docking with orbital in manual mode

Gliese852:
If you fly in or launch from the orbital station on autopilot - everything works as before, I just improved these animations a little. But if your autopilot is off, you can fly into the station and land manually.

Since the station is rotating, it is recommended to turn on the "Follow orientation" mode, then the ship will compensate for the rotation of the station. Although, you can land completely manually.

In order not to hit anything, when manually starting from the orbital, the "cruise-up" mode is automatically turned on. It seemed to me that it works well in other places too - when launching from a ground station or from the surface - you are slowly ascending at a speed of 2 m/s.

In fact, this is a quite large refactoring of SpaceStation and SpaceStationType, as a result of which it becomes trivial to implement manual docking.

pic

### DIFF
--- a/src/ModelBody.cpp
+++ b/src/ModelBody.cpp
@@ -133,10 +133,17 @@ void ModelBody::SetStatic(bool isStatic)
 void ModelBody::SetColliding(bool colliding)
 {
 	m_colliding = colliding;
-	if (colliding)
+	if (colliding) {
 		m_geom->Enable();
-	else
+		for(auto &g : m_dynGeoms) {
+			g->Enable();
+		}
+	} else {
 		m_geom->Disable();
+		for(auto &g : m_dynGeoms) {
+			g->Disable();
+		}
+	}
 }
 
 void ModelBody::RebuildCollisionMesh()
@@ -152,6 +159,7 @@ void ModelBody::RebuildCollisionMesh()
 
 	//static geom
 	m_geom = new Geom(m_collMesh->GetGeomTree(), GetOrient(), GetPosition(), this);
+	if(!IsColliding()) m_geom->Disable();
 
 	SetPhysRadius(maxRadius);
 
@@ -166,6 +174,7 @@ void ModelBody::RebuildCollisionMesh()
 		SceneGraph::CollisionGeometry *cg = dgf.GetCgForTree(*it);
 		if (cg)
 			cg->SetGeom(dynG);
+		if(!IsColliding()) dynG->Disable();
 		m_dynGeoms.push_back(dynG);
 	}
 
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -897,13 +897,22 @@ void Ship::Blastoff()
 
 	assert(f->GetBody()->IsType(ObjectType::PLANET));
 
-	const double planetRadius = 2.0 + static_cast<Planet *>(f->GetBody())->GetTerrainHeight(up);
-	SetVelocity(vector3d(0, 0, 0));
-	SetAngVelocity(vector3d(0, 0, 0));
-	SetFlightState(FLYING);
+	if (ManualDocking()) {
+		if (!IsType(ObjectType::PLAYER)) {
+			Log::Warning("It wasn't the player's ship that tried to take off without an autopilot!");
+			return;
+		}
+		auto p = static_cast<Player*>(this);
+		p->DoFixspeedTakeoff();
+	} else {
+		const double planetRadius = 2.0 + static_cast<Planet *>(f->GetBody())->GetTerrainHeight(up);
+		SetVelocity(vector3d(0, 0, 0));
+		SetAngVelocity(vector3d(0, 0, 0));
+		SetFlightState(FLYING);
 
-	SetPosition(up * planetRadius - GetAabb().min.y * up);
-	SetThrusterState(1, 1.0); // thrust upwards
+		SetPosition(up * planetRadius - GetAabb().min.y * up);
+		SetThrusterState(1, 1.0); // thrust upwards
+	}
 
 	LuaEvent::Queue("onShipTakeOff", this, f->GetBody());
 }
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -34,6 +34,7 @@
 #include "lua/LuaUtils.h"
 #include "scenegraph/Animation.h"
 #include "scenegraph/Tag.h"
+#include "scenegraph/CollisionGeometry.h"
 #include "ship/PlayerShipController.h"
 
 static const float TONS_HULL_PER_SHIELD = 10.f;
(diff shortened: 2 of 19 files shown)


nozmajner:
I've recorded a short showcase of this feature:
https://www.youtube.com/watch?v=GbELjssyjG0

src/SpaceStationType.h
```diff
@@ -19,14 +19,60 @@ namespace SceneGraph {
...
+
+	DOCK_STAGES_BEGIN,
+
+	CLEARANCE_GRANTED,
+
+	DOCK_ANIMATION_NONE,
+	DOCK_ANIMATION_1,
+	DOCK_ANIMATION_2,
+	DOCK_ANIMATION_3,
+	DOCK_ANIMATION_MAX,
```
Web-eWorks, line 35: Have you considered replacing the hard-coded 4-stage docking animation to instead be a dynamic list of Tags generated from the model file setup? I'd expect that to be significantly more flexible for future changes to the docking procedure (e.g. ships "flying themselves" along the docking path) and also allow e.g. using a "locator beacon" well outside the station for the first approach node.

Web-eWorks:
Apologies I've not touched this for some time - I somehow was under the mistaken impression it was still WIP. If you could rebase against master (just to make sure it works with the physics changes I merged recently), I'll go ahead and do a full review the next time I get a chance.

nozmajner:
Here are the updated collision meshes. Pads aren't overlapping with the volume and port meshes.
[orbital_dockings_collision_non-overlap_fix.zip](https://github.com/pioneerspacesim/pioneer/files/13821197/orbital_dockings_collision_non-overlap_fix.zip)

src/ModelBody.cpp
```diff
@@ -133,10 +133,17 @@ void ModelBody::SetStatic(bool isStatic)
 void ModelBody::SetColliding(bool colliding)
 {
 	m_colliding = colliding;
-	if (colliding)
+	if (colliding) {
 		m_geom->Enable();
-	else
+		for(auto &g : m_dynGeoms) {
+			g->Enable();
+		}
```
Web-eWorks, line 140: I'm curious if this solves the recent batch of issue reports regarding inability to collide with / land on stations. I did make the CollisionSpace code stricter, as it no longer includes disabled Geoms in the BVH construction at all (rather than filtering them out later in the pipeline).

src/Ship.cpp
```diff
@@ -897,13 +897,20 @@ void Ship::Blastoff()
 
 	assert(f->GetBody()->IsType(ObjectType::PLANET));
 
-	const double planetRadius = 2.0 + static_cast<Planet *>(f->GetBody())->GetTerrainHeight(up);
-	SetVelocity(vector3d(0, 0, 0));
-	SetAngVelocity(vector3d(0, 0, 0));
-	SetFlightState(FLYING);
+	if (ManualDocking()) {
+		// no one except the player can take off without the autopilot
+		assert(IsType(ObjectType::PLAYER));
```
Web-eWorks, line 902: I'd strongly prefer this be a runtime check with a soft-fail condition rather than an assertion. If the precondition is that this branch can only be taken by an object of type Player, then the logic should live in a method on Player instead - from a separation-of-concerns perspective at least.

src/SpaceStation.cpp
```diff
@@ -383,72 +409,72 @@ bool SpaceStation::GetDockingClearance(Ship *s)
...
+	shipDocking_t &sd = m_shipDocking[bay];
+	sd.ship = s;
+	sd.stagePos = 0;
+	sd.fromPos = (s->GetPosition() - GetPosition()) * GetOrient(); // station space
+	sd.fromRot = Quaterniond::FromMatrix3x3(GetOrient().Transpose() * s->GetOrient());
+	s->SetFlightState(Ship::DOCKING);
+	s->SetVelocity(vector3d(0.0));
+	s->SetAngVelocity(vector3d(0.0));
+	s->ClearThrusterState();
+	SwitchToStage(bay, (touchOrbitalPad || m_type->NumUndockStages() == 0) ? DockStage::TOUCHDOWN : DockStage::DOCK_ANIMATION_1);
```
Web-eWorks, line 475: Is this block of code worth separating to its own function? If not, I'd recommend adding a slightly more verbose preceeding comment as it's very dense and not fully clear *what* is being done here.

src/SpaceStation.cpp
```diff
@@ -458,132 +484,154 @@ bool SpaceStation::DoShipDamage(Ship *s, Uint32 flags, double relVel)
...
-//
+void SpaceStation::SwitchToStage(Uint32 bay, DockStage stage)
+{
+	shipDocking_t &dt = m_shipDocking[bay];
+
+	switch (stage) {
+
+	case DockStage::TOUCHDOWN:
+		{
+			dt.ship->ClearThrusterState();
```
Web-eWorks, line 495: Off the top of my head, I believe this indentation doesn't match other switch case statements in the codebase.

src/SpaceStationType.h
Gliese852: In fact, the animation of a specific station has an arbitrary number of stages, for example, ground stations have none at all. And it is read from the tags in the model, although there is some hard-coded logic that adds intermediate "locators". This is old code and it works well with current models.
I think it will be easy enough to turn this logic into a separate dynamic list when needed. 
External beacons are stored separately in `DockStage::APPROACH1` and `DockStage::APPROACH2`. I think that conceptually they should be located separately, so that, for example, they do not rotate with the station.

src/ModelBody.cpp
Gliese852: Added these changes to the master and tested #5705 - does not help.

src/Ship.cpp
Gliese852: I just give a warning and return, is it ok? Or is more serious error handling required?

src/SpaceStation.cpp
```diff
@@ -383,72 +409,72 @@ bool SpaceStation::GetDockingClearance(Ship *s)
...
+	}
+	if (bay == -1) {
+		return DoShipDamage(s, flags, relVel); // no permission
+	}
+	if (IsPortLocked(bay)) {
+		return DoShipDamage(s, flags, relVel);
+	}
+	if (m_shipDocking[bay].stage == DockStage::NONE) {
+		return DoShipDamage(s, flags, relVel); // also no clearance?
+	}
```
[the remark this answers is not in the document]
Gliese852: Done!

src/SpaceStation.cpp
Gliese852: Done!

src/SpaceStation.cpp
Gliese852: Since it is not reused anywhere, I did not separate it into a function and just added comments.

src/Ship.cpp
Web-eWorks: That's fine, I'm mostly concerned with it being a "recoverable" error rather than an immediate crash.

impaktor:
Just a reminder to not smash merge until FIXUP commits are rebased. 

_Regards,_
_Your friendly git-fascist_

Web-eWorks (approved):
Approval-ed, pending rebase to appease the resident git-fascist.

Gliese852:
Great! Need to wait for a little more improvement in the collision meshes, related post - https://github.com/pioneerspacesim/pioneer/pull/5619#issuecomment-1875629186, and I will make a rebase.

nozmajner:
@Gliese852 These should be good:
[orbital_dockings_collision_non-overlap_fix_nobulge.zip](https://github.com/pioneerspacesim/pioneer/files/13854663/orbital_dockings_collision_non-overlap_fix_nobulge.zip)
Removed the bulges again, and moved the empties a bit above the pad for good measure.
image

nozmajner:
[orbital_dockings_collision_non-overlap_fix_nobulge2.zip](https://github.com/pioneerspacesim/pioneer/files/13854722/orbital_dockings_collision_non-overlap_fix_nobulge2.zip)

nozmajner:
[station_docking_A.zip](https://github.com/pioneerspacesim/pioneer/files/13854929/station_docking_A.zip)

nozmajner:
[station_docking_A.zip](https://github.com/pioneerspacesim/pioneer/files/13854992/station_docking_A.zip)

Gliese852:
Rebased fixup commits, ready to merge.

PR #5623, opened 2023-09-08
Expand ModManager, add System Editor prerequisites

Web-eWorks:
Catchall PR with useful commits not directly related to the System Editor coming out of that branch.

In brief:
- Cleanup `Background::StarField` so it actually does what it says on the tin.
- Expand `ModManager` functionality to allow querying and disabling mods.
- Added Portable File Dialogs for editor usage, has a very nice API to wrap what is otherwise a horrible platform-specific mess.
- Migrate `autoload.lua` into the "minimal" Lua state configuration (for LuaNameGen and other lightweight Lua utilities to run outside of the main Lua state).
- Replace (most of) our various string-splitting helper functions with a generic interface to iterate string fragments and collate them into a vector-like container.
  - There are several hand-written re-implementations of the same functionality scattered around the code that I have not tracked down and replaced with the SplitString utility. A task for another PR.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -203,6 +203,10 @@ Pioneer includes the following third-party software:
   Copyright (c) 2019 Stanislav Denisov
   Licensed under the MIT license
 
+  Portable File Dialogs
+  Copyright © 2018–2022 Sam Hocevar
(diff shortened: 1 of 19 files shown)


fluffyfreak (approved):
I approve, especially of the background cleanup, but it's all good work!

PR #5624, opened 2023-09-08
Build directly from the CMake configurations on windows with Visual Studio 2019

JonBooth78:
If you follow the updated instructions in COMPILING.txt you should be able to open up visual studio and have it look something like this:

image

You can then select debug/release/profiling configurations, set pioneer.exe as your startup project and build/run/debug


impaktor:
Does @WKFO use Windows?

Web-eWorks (approved):
Looks good to me! The CMakePresets.json should tailor in very nicely with the existing VSCode workflow and allow us to ship an even better "new user experience" when it comes to compiling the project on both Windows and Linux!

fluffyfreak (approved):
Looks good to me!

PR #5628, opened 2023-09-13
Fix for -10% and +10% buttons in station's lobby

### LINKED ISSUE #5585: Incorrect substraction from in stock or demand when using +10% or -10% buttons from station's lobby
### Observed behaviour
Selling or buying fuel in station's lobby using +10% or -10% buttons incorectly substracts it from in stock or demand respectively. For example, when I sell 3t of fuel it will subtract 2t from demand, and when buying 2t of fuel, it will substract 3t from in stock.

### Expected behaviour
Selling fuel and buying fuel should correctly substract corresponding columns.

### Steps to reproduce
1. Go to commodity market;
2. Remember Hydrogen in stock and demand;
3. Go to lobby, watch how much fuel you have now;
4. Sell or buy fuel using +10% or -10% buttons, watch how much it substract or adds it to you ship fuel;
5. Compare this with in stock(buy) or demand(sell).

Latest master build (https://github.com/pioneerspacesim/pioneer/commit/ac06900bc94f28b9a0871b2aba163d769e27373d) (Windows 10 x64)

Max5377:
Fixes #5585

The problem was with this line in `refuelInternalTank` in `data/pigui/modules/station-view/01-lobby.lua`:

`station:AddCommodityStock(Commodities.hydrogen, -math.ceil(mass))`

`math.ceil` is rounding number up, so if this number is negative(ex. -2.3) it will be rounded to -2 because it is nearest value that greater or equal to -2.3. So the solution is to change above mentioned line to:

`local commodityChangeAmount = mass < 0 and math.floor(mass) or math.ceil(mass)
station:AddCommodityStock(Commodities.hydrogen, commodityChangeAmount)`

[Accepted answer by Guffa(albeit it's for javascript, but works the same way)](https://stackoverflow.com/questions/26335076/math-ceil-not-working-with-negative-floats)

### DIFF
--- a/data/pigui/modules/station-view/01-lobby.lua
+++ b/data/pigui/modules/station-view/01-lobby.lua
@@ -128,7 +128,8 @@ local refuelInternalTank = function (delta)
 	end
 
 	Game.player:AddMoney(-total)
-	station:AddCommodityStock(Commodities.hydrogen, -math.ceil(mass))
+	local commodityChangeAmount = mass < 0 and math.floor(mass) or math.ceil(mass)
+	station:AddCommodityStock(Commodities.hydrogen, commodityChangeAmount)
 	Game.player:SetFuelPercent(fuel)
 end
 


impaktor:
Looking good

Just change

> Fix for https://github.com/pioneerspacesim/pioneer/issues/5585

to 

> Fixes #5585

and it will [auto-close](https://github.blog/2013-01-22-closing-issues-via-commit-messages/) the issue when we merge this code.

Max5377:
@impaktor Done

Web-eWorks (approved):

Max5377:
Upon futher testing, I can say that my solution is not good either.
In my code, the station will receive more hydrogen than needed, because number is just rounded up, leaving fractional parts behind. Example:
Coronatrix fuel tank: 23t
10% from this amount: 2.3t

`math.floor(-2.3) = -3, math.ceil(2.3) = 3`

In the original code, it would be a little different:

`math.ceil(-2.3) = -2, math.ceil(2.3) = 3`

So, if there are no fractions in the goods, I don't think this can be correctly counted without them. That's what I think, at least.

Also, in this line:

`station:AddCommodityStock(Commodities.hydrogen, commodityChangeAmount)`

I forgot to add minus to `commodityChangeAmount`, because of this, demand and in stock decrease is mixed up (-10% decreases in stock and +10% decreases demand).

impaktor:
@Max5377 I suggest you start a new branch & PR fixing the minus, and any other issue.

PR #5631, opened 2023-09-18
Fix for glued cockpit

### LINKED ISSUE #5621: Cockpit model occasionally becomes glued to the view at a random angle
At random times the 1st person cockpit model becomes glued to the view at a bad angle: It will always rotate with the view instead of the ship, meaning middle-mouse-button dragging can no longer be used to look around inside it... further more it's badly rotated so you're stuck looking at the chair or window and can't even see outside any more. Fortunately the issue can be resolved by saving and reloading your save.

Screenshot_20230905_024853

I don't yet know what exactly triggers this behavior, only that it happens rarely but does on occasion. I believe someone on Discord suggested it may be related to pausing / unpausing the time.

I observed this behavior both in the newest Pioneer release (20230203) as well as latest Git master. I'm running / compiling it on Linux (Manjaro). AMD graphics card using the default amdgpu driver.

Max5377:
Fixes #5621

This happened, because there was no check in this statement in `void ShipCockpit::Update(const Player *player, float timeStep)`: 
`if (m_icc == nullptr)`,
if current `CameraController` is really `InternalCameraController`. Because of that, this statement in `inline void ShipCockpit::resetInternalCameraController()`:
`m_icc = static_cast<InternalCameraController *>(Pi::game->GetWorldView()->shipView->GetCameraController());`,
would cast current `CameraController` to invalid type if it was not `InternalCameraController`, which is undefined behaviour.
Added check at the start of `Update` if current view is exterior to ensure that current `CameraController` is really `InternalCameraController`.

### DIFF
--- a/src/ShipCockpit.cpp
+++ b/src/ShipCockpit.cpp
@@ -64,13 +64,18 @@ inline void ShipCockpit::resetInternalCameraController()
 
 void ShipCockpit::Update(const Player *player, float timeStep)
 {
-	m_transform = matrix4x4d::Identity();
-
+	//Check if current view is exterior since we don't need to update cockpit
+	//because player can't see it
+	if (Pi::game->GetWorldView()->shipView->IsExteriorView())
+	{
+		return;
+	}
 	if (m_icc == nullptr) {
 		// I don't know where to put this
 		resetInternalCameraController();
 	}
 
+	m_transform = matrix4x4d::Identity();
 	double rotX;
 	double rotY;
 	m_icc->getRots(rotX, rotY);


Web-eWorks:
If there is now no codepath that calls `resetInternalCameraController` while the interior camera controller is inactive, the `dynamic_cast` should not be used. `dynamic_cast` has a significant performance overhead compared to `static_cast` and generally should be avoided in situations where the concrete type of a variable is guaranteed.

Max5377:
Makes sense with a type-checking before casting. I will change this back to `static_cast`.

Web-eWorks:
I have no intentions to further improve ShipCockpit.cpp. It is the last file in the codebase from the Paragon/Jumpdrive codebase (as indicated by the Meteoric Games copyright header), has many significant deficiencies/bugs, and is pending a complete replacement with a "clean-room" implementation of ship cockpit view that's being developed in my clickable cockpits branch.

You are welcome to make any changes to it that you deem necessary to improve performance or reduce unnecessary computation - however given that it will be completely gone in 1-2 releases I'm not going to ask you to spend any further time on it if you don't want to.

Were this another part of the code that isn't in "pending removal" state, I'd say go for it. :smile:

Max5377:
Changed `dynamic_cast` back to `static_cast` and added `IsExteriorView` check at the start of `Update`.

Web-eWorks:
Looks good! I'll test this in a few days in the same batch as your other PRs and (most likely) merge all three at once.

impaktor:
@Max5377 Please let us know if you want to be in [AUTHORS.txt](https://github.com/pioneerspacesim/pioneer/blob/master/AUTHORS.txt)

Max5377:
@impaktor Yes, you can add me in AUTHORS.txt. I write later what to add here, need to think about it.

PR #5632, opened 2023-09-18
Compare specs of your current ship to those in the ship market and highlight changes

JonBooth78:
When you are in the ship market we now color the different specifications and have an indicator arrow next to them to show if the ship being viewed is an improvement or not over the current one:

image

### DIFF
--- a/data/pigui/modules/station-view/04-shipMarket.lua
+++ b/data/pigui/modules/station-view/04-shipMarket.lua
@@ -161,27 +161,53 @@ end
 
 local FormatAndCompareShips = {}
 
-local get_color_and_mod_string_for_compare = function (str, a, b)
+function FormatAndCompareShips:compare_and_draw_column(desc, str, a, b)
 	compare = a-b
-	if compare == 0 then
-		return { str, styleColors.white, " " }
-	elseif compare < 0 then
+	color = styleColors.white
+	indicator = " "
+	if compare < 0 then
+		color = styleColors.warning_300
 		-- note a preceeding 3 per en space as no column padding
 		-- this is unicode point 2004
-		return { str, styleColors.warning_300, " ↓" }
+		indicator = " ↓"
+	elseif compare > 0 then
+		color = styleColors.success_300
+		-- note a preceeding 3 per en space as no column padding
+		-- this is unicode point 2004
+		indicator = " ↑"
+	end
+
+	if self.column == 0 then
+		ui.tableNextRow()
+	end
+
+	ui.tableSetColumnIndex(0 + self.column)
+	ui.textColored(styleColors.gray_300, desc)
+
+	ui.tableSetColumnIndex(1 + self.column)
+	ui.textAlignedColored(str, 1.0, color)
+
+	ui.tableSetColumnIndex(2 + self.column)
+
+	if self.column == 0 then
+		-- this is an em space (unicode point 2003)
+		-- as we don't have and column padding
+		ui.textColored(color,indicator .. " ")
+		self.column = 3
 	else
-		return { str,  styleColors.success_300, " ↑" }
+		ui.textColored(color,indicator)
+		self.column = 0
 	end
 end
 
-function FormatAndCompareShips:get_hyperdrive_tuple()
+function FormatAndCompareShips:draw_hyperdrive_cell(desc)
 	local hyperdrive_str = self.def.hyperdriveClass > 0 and
 		Equipment.hyperspace["hyperdrive_" .. self.def.hyperdriveClass]:GetName() or l.NONE
-	
-	return get_color_and_mod_string_for_compare( hyperdrive_str, self.def.hyperdriveClass, self.b.def.hyperdriveClass )
+
+	self:compare_and_draw_column( desc, hyperdrive_str, self.def.hyperdriveClass, self.b.def.hyperdriveClass )
 end
 
-function FormatAndCompareShips:get_value( key )
+function FormatAndCompareShips:get_value(key)
 	v = self[key]
 	if nil == v then
 		v = self.def[key]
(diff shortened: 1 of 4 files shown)


Web-eWorks (changes requested):
Finally gotten around to reviewing this, been swamped / away from Github for the last little while.

Overall I'm very happy with this PR; it's not quite as compact a solution as I'd prefer in an "ideal world", but it's definitely a very clean and readable solution and not worth spending any further time over for the sake of code golf.

I've made a few suggestions for cleanliness / code smell / maintainability reasons; feel free to ask if you have any questions how to accomplish the needed changes!

data/pigui/modules/station-view/04-shipMarket.lua
```diff
@@ -159,15 +159,143 @@ local function buyShip (mkt, sos)
...
+	if self.column == 0 then
+		-- this is an em space (unicode point 2003)
+		-- as we don't have and column padding
+		ui.textColored(color,indicator .. " ")
+		self.column = 3
+	else
+		ui.textColored(color,indicator)
+		self.column = 0
+	end
+end
```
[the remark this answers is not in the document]
JonBooth78: Happy to change over and use an icon if that's what you'd prefer. 

The problem I have with using the cell padding when defining the table is that it doesn't support different padding per column. Essentially here we've got 2 tables, the first with columns, each with a table with 3 columns inside and I want different cell padding for both. 

Right now we're simulating this with one table with 6 columns. Can we embed one table inside another as this would solve my problem much more neatly and I should then be able to specify the desired table padding separately for each of the tables.  It would though necessitate a different, less natural draw order as I think we'd have to draw the left column in it's entirety first and then the right.

For aesthetics, I want there to be a gap between the two 'major' columns but also want a smaller gap, between the stat and the indicator arrow.  I settled on an em space (a typographic space that is always the same width as the height of the body of a line of text) as the width between the major columns, and 1/3 of the width of the em space (so the 3 per em space) for the gap between the stat and the indicator arrow- and messed up the comment above, where I said en space for that.

I also don't really like using typographic spaces in source code like the above, I'd much rather use something like "UTF.char(2004)" to show exactly what has been chosen, as in the source editor it all just looks like whitespace, however our version of lua doesn't seem to support that and I didn't fancy writing a c util function to enable it - which would be better.

data/pigui/modules/station-view/04-shipMarket.lua
```diff
@@ -159,15 +159,143 @@ local function buyShip (mkt, sos)
...
+	else
+		atmoSlot = string.format("%d atm", self.def.atmosphericPressureLimit)
 	end
+	self:compare_and_draw_column( desc, atmoSlot, self.def.atmosphericPressureLimit, self.b.def.atmosphericPressureLimit )
+end
+
+function FormatAndCompareShips:new(def,b)
+	o = {}
+	setmetatable( o, self )
+	self.__index = self
```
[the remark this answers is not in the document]
JonBooth78: Oh, that's cool, I didn't realize we had any utils for doing that, it's much more readable.

data/pigui/modules/station-view/04-shipMarket.lua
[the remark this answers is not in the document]
JonBooth78: Never mind, I just worked out how to do the columns and have them aesthetically pleasing without all the spacing stuff.

Just add an empty column in between the 2 major ones.

I'll get onto this :-)

data/pigui/modules/station-view/04-shipMarket.lua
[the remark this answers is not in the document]
JonBooth78: Done :-)

data/pigui/modules/station-view/04-shipMarket.lua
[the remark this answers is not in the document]
JonBooth78: Done :-)

data/pigui/modules/station-view/04-shipMarket.lua
```diff
@@ -281,17 +286,14 @@ function FormatAndCompareShips:draw_atmos_pressure_limit_cell(desc)
...
-	o.b = b
+function FormatAndCompareShips:Constructor(def, b)
+	self.column = 0
+	self.emptyMass = def.hullMass + def.fuelTankMass
+	self.fullMass = def.hullMass + def.capacity + def.fuelTankMass
+	self.massAtCapacity = def.hullMass + def.capacity
+	self.cargoCapacity = def.equipSlotCapacity["cargo"]
+	self.def = def
+	self.b = b
 	return o
```
Web-eWorks, line 291: `return o` here has no effect - the new instance (`self` in this function) will be returned from `FormatAndCompareShips.New()`.

data/pigui/modules/station-view/04-shipMarket.lua
```diff
@@ -178,18 +182,14 @@ end
...
-		-- this is unicode point 2004
-		indicator = " ↓"
+		color = ui.theme.colors.shipmarketCompareWorse
+		icon_i = ui.theme.icons.shipmarket_compare_worse
 	elseif compare > 0 then
-		color = styleColors.success_300
-		-- note a preceeding 3 per en space as no column padding
-		-- this is unicode point 2004
-		indicator = " ↑"
+		color = styleColors.success_300 -- ui.theme.colors.shipmarketCompareBetter
```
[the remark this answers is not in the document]
Web-eWorks: I was looking at an old version of the code, my bad!

Web-eWorks:
@nozmajner would you have any time in your schedule to make "better" / "worse" arrow icons for the icon sheet? For a quick-and-dirty option, you could use the export major / import major icons and just remove the outer ring...

nozmajner:
I don't know when I can get around to it. I'd use one of the triangular icons for now instead

nozmajner:
I think I'd use either of these:
image

data/pigui/modules/station-view/04-shipMarket.lua
JonBooth78: Thanks

JonBooth78:
screenshot-20231003-194811
I changed the icon set and we now end up with the above.

nozmajner:
Looks cool!

Web-eWorks (approved):
Looks good to me! I'll give this a test sometime in the next day or so and pull the merge trigger.

Web-eWorks (changes requested):
After having tested it, I have an issue and a feature request:
- The forward acceleration is presented in negative G's - while I know this was probably existing behavior, it's rather nonsensical (and doesn't match the Ship Info screen). It also seems to be incorrectly comparing with the current ship, as a ship with a lower empty forward acceleration than my ship's current forward acceleration is showing up as having "better" acceleration.
- I'd like to see a tooltip when hovering stats that indicates the current value of that stat on your ship, as that would save the player a click into a different info screen every time they want to know exactly *how* much better the new ship's equipment is.

data/pigui/modules/station-view/04-shipMarket.lua
```diff
@@ -159,15 +162,135 @@ local function buyShip (mkt, sos)
...
+
+function FormatAndCompareShips:draw_accel_cell(desc, thrustKey, massKey, multiplier)
+	local accelA = multiplier * self.def.linearThrust[thrustKey] / (-9.81*1000*(self:get_value(massKey)))
+	local accelB = multiplier * self.b.def.linearThrust[thrustKey] / (-9.81*1000*(self.b:get_value(massKey)))
+	self:compare_and_draw_column( desc, Format.AccelG(accelA), accelA, accelB )
+end
+
+function FormatAndCompareShips:draw_deltav_cell(desc, massNumeratorKey, massDenominatorKey)
+	local deltavA = self.def.effectiveExhaustVelocity * math.log( self:get_value(massNumeratorKey), self.b:get_value(massDenominatorKey) )
+	local deltavB = self.b.def.effectiveExhaustVelocity * math.log( self.b:get_value(massNumeratorKey), self.b:get_value(massDenominatorKey) )
```
Web-eWorks, line 244: This deltaV calculation is also incorrect. It should be `math.log( massNumerator / massDenominator )`, with the empty second parameter defaulting to *e*. Lua's log function takes the base of the logarithm in the second parameter, not a divisor.

JonBooth78: All done and completed.  Tooltips are a great addition :-)

JonBooth78:
screenshot-20231008-123501

Here we are with tooltips

data/pigui/modules/station-view/04-shipMarket.lua
```diff
@@ -180,29 +180,43 @@ function FormatAndCompareShips:beginTable()
...
 
+	local new_str = fmt_a and fmt_a( a ) or a
+
 	if compare < 0 then
-		ui.textAlignedColored(str, 1.0, ui.theme.colors.shipmarketCompareWorse)
-		ui.tableSetColumnIndex(2 + self.column)
-		ui.icon( ui.theme.icons.shipmarket_compare_worse, Vector2(ui.getTextLineHeight()), ui.theme.colors.shipmarketCompareWorse)		
-		
+		local old_str = fmt_b and fmt_b( b ) or b
+		ui.withTooltip( old_str, function ()
```
Web-eWorks, line 206: Would prefer if this included a "Current Ship: " prefix to the text to clarify the context of the tooltip.

data/pigui/modules/station-view/04-shipMarket.lua
```diff
@@ -215,69 +229,99 @@ function FormatAndCompareShips:compare_and_draw_column(desc, str, a, b)
...
-function FormatAndCompareShips:draw_accel_cell(desc, thrustKey, massKey, multiplier)
+function FormatAndCompareShips:draw_accel_cell(desc, thrustKey, massKey, multiplier, inverse)
 	local accelA = multiplier * self.def.linearThrust[thrustKey] / (-9.81*1000*(self:get_value(massKey)))
 	local accelB = multiplier * self.b.def.linearThrust[thrustKey] / (-9.81*1000*(self.b:get_value(massKey)))
-	self:compare_and_draw_column( desc, Format.AccelG(accelA), accelA, accelB )
+	if inverse then
+		accelA = -accelA
+		accelB = -accelB
+	end
+	self:compare_and_draw_column( desc, accelA, accelB, Format.AccelG )
```
Web-eWorks, line 260: As mentioned on IRC, this should be `accelA = self.def.linearThrust[thrustKey] / (9.8106 * 1000 * self:get_value(massKey))` with no `multiplier` or `inverse` terms. All accelerations are expressed with positive values, and the current code in the PR performs two inversions of that positive value to render the acceleration value.

data/pigui/modules/station-view/04-shipMarket.lua
```diff
@@ -215,69 +229,99 @@ function FormatAndCompareShips:compare_and_draw_column(desc, str, a, b)
...
+	local function fmt( v )
+		local yes_no = "unknown"
+		if v == 1 then
+			yes_no = l.YES
+		elseif v == 0 then
+			yes_no = l.NO
+		else
+			error("argument to yes_no not 0 or 1")
+		end
+		return yes_no
```
Web-eWorks, line 293: This should be expressed as `function (v) return v == 1 and l.YES or l.NO end`. The internal error bookkeeping overly complicates the purpose and is the wrong place to precondition the inputs; the `error()` means that `"unknown"` will never be returned as part of existing behavior and thus can be dropped.

Web-eWorks (changes requested):
One last batch of mostly stylistic/code-quality nitpicks and this is good to merge, thanks for bearing with the long review process!

As mentioned on IRC, the documentation to `FormatAndCompareShips:compare_and_draw_column` declares `fmt_a` twice; that should be a quick cleanup while you're there.

data/pigui/modules/station-view/04-shipMarket.lua
Web-eWorks: As a note, if you've not already touched the translation system, you'd add a new entry to `data/lang/ui-core/en.json` in the format:
```json
"CURRENT_SHIP": {
    "description": "",
    "message": "Current Ship"
}
```

You'd use that new language string like so:
```lua
-- with string.interp shortcut operator '%'
ui.withTooltip( "{}: {}" % { l.CURRENT_SHIP, old_str }, function() --...
-- the boring (but slightly more performant) way
ui.withTooltip( l.CURRENT_SHIP .. ": " .. old_str, function() --...
```

data/pigui/modules/station-view/04-shipMarket.lua
JonBooth78: Sorry for wasting your time there, I really didn't think about that very much :-(  Still had a slightly sore head from the night before when I did this!

Web-eWorks (approved):
Looks good to me! You can manually clean up the git history, or I can squash-merge all the history into a single commit.

PR #5635, opened 2023-09-28
ScanManager improvements and ship properties fixes

### LINKED ISSUE #5534: Disapearing Fuel after save and load
### Observed behaviour

With the 03/02/23 build, I sometimes lose all the fuel in my cargo after saving and loading a game in space.
Happens randomly around once an hour.

Last time it happened :
- it was with a Coronatrix (30t space, 16t max cargo):
- I started a mission with 23t fuel + 24t equipment + 6t hydrogen in cargo (max cargo of 6t)
- I made a 2t jump
- i checked my ship info and it was as intendeded : 4t hydrogen in cargo

Then save -> quit -> load : now my cargo appears empty and my max cargo is 2t (equipment list still shows 24t of equipments)

### Expected behaviour

Cargo content should not change after save/load.

### Steps to reproduce

Hard to say. When it happens and I have a save just before the jump I can reproduce it again and again.
But it does not happen every time.

See attached saves : 
- Landed 2 - Depart : before take off and jump
- mission 1 : after 2t direct jump to Doligray system & save

__My pioneer version (and OS): last relaese, 03/02/23 on windows 11
[saves.zip](https://github.com/pioneerspacesim/pioneer/files/10689985/saves.zip)

### LINKED ISSUE #5557: Ship info screen shows wrong cargo capacity and usage.
### Observed behaviour
I have 16 tons of cargo in my hold.
The ship info screen says I have 8 tons of cargo capacity of which 0 are used.
image
image

### Expected behaviour
16 tons of cargo in hold, 16 tons cargo capacity, 16 tons used.

### Steps to reproduce

Issue is evident in this save file.
When starting the game I sold off some starter equipment. Not sure if that is related.
I am also wondering if this is the cause of #5554; could a police ship be trying to scan me but the cargo is corrupted?

[vert.zip](https://github.com/pioneerspacesim/pioneer/files/10785808/vert.zip)

e77da96 ; Windows and Steam Deck

### LINKED ISSUE #5577: [Bug] Scans do not resume on game continue
When you close the game or go to main menu during an orbital scan it does not increase progress when you resume the game by clicking the "Continue Game" option in the main menu. I haven't tested this with geographical scan or with resuming from save files but it's probably the same.

In order to resume the scan you have to click on the active scan to move it to pending and then click on it again to move it back to active and then progress will increase again.

Max5377:
**UPDATE**
As suggested by @Web-eWorks, first solution changed to wrap `totalCargo` and `usedCargo` around this check (ex. for `totalCargo`) in `function CargoManager:Constructor(ship)`:
```
if not self.ship:hasprop("totalCargo") then
	ship:setprop("totalCargo", self:GetTotalSpace())
end
```
so this properties will not be reinitialized by `CargoManager`'s constructor, because after saving the game, this properties will be already stored in the ship and will be unserialized during loading the game.
`self.ship:UpdateEquipStats()` was not needed since it's called internally from C++.

While testing more, another bug was found, `maxHyperspaceRange` wasn't properly initialized after loading the game.
This happened, because in this line in `void Ship::UpdateLuaStats()`:

```
	if (hyperclass) {
		std::tie(m_stats.hyperspace_range_max, m_stats.hyperspace_range) =
			LuaObject<Ship>::CallMethod<double, double>(this, "GetHyperspaceRange");
	}
```

`m_stats.hyperspace_range_max`, `m_stats.hyperspace_range` are mixed up since `GetHyperspaceRange` returns `range` and `range_max`, `range` is zero at this time, but will be initialized later. 
Switched them up so `maxHyperspaceRange` will be properly initialized.

### Incorrect display of `cargo space used` after loading the game

Fixes #5557 fixes #5534   
(Maybe more)

The `Ship`'s properties were not assigned new `CargoManager` property values after it was loaded from save.
The most evident case of this is incorrect display of `cargo space used` in `personal info` after loading the save.
This properties is used in a bunch of other places that I listed [here](https://github.com/pioneerspacesim/pioneer/issues/5557#issuecomment-1737314182). 
One of this places is calculating probability of pirates attacking player, so I guess expect more pirate attacks?!
I also want to clarify that this was happening to **every** ship present during save loading.
To fix it, added this lines in `CargoManager:Unserialize`:

```
self.ship:setprop("totalCargo", self:GetTotalSpace())
self.ship:setprop("usedCargo", self.usedCargoSpace)

self.ship:UpdateEquipStats()
```
To assign new values to `Ship`'s properties during `CargoManager` unserialization.

Still don't know if this line is needed:

```
self.ship:UpdateEquipStats()
```
but I leave it for now, just in case, like in `CargoManager:OnShipTypeChanged()`.

### Scanner not starting previous scan after loading the game

Fixes #5577

`ScanManager: StartScanCallback` had no way to determine if the current `ScanManager` was loaded from save when it needed to set up a callback, because field `activeCallback` of `ScanManager` is saved and during unserialization `ScanManager:StartScanCallback` is called which has this check:
```
if updateRate == self.activeCallback then
```
which basically thinks that callback is already setted up for current `ScanManager`, which in reality is not.

Added additional argument to `ScanManager:StartScanCallback` `force` and changed this check to:
```
if updateRate == self.activeCallback and not force then
```
to force through this check if `ScanManager:StartScanCallback` was called from `ScanManager:Unserialize`.

### Orbital scanning stops despite meeting maximum altitude requirements
This was due to an oversight in the `ScanManager`s code, when it did not take into account that player might be too far from the planet, when his frame of reference (`frameBody`) was changing, immediately stopping the scanning process.
Added additional checks when current scan is orbital (ex. immediatly returning from `onFrameChanged` when current scan is orbital) and branching body choice:
```
self.activeScan.orbital and self.activeScan.bodyPath:GetSystemBody().body or self.ship.frameBody
```
**Update**: a small revision of code.
**Update2**: added additional assertions with messages to better describe what happened.

### DIFF
--- a/data/libs/CargoManager.lua
+++ b/data/libs/CargoManager.lua
@@ -27,8 +27,14 @@ function CargoManager:Constructor(ship)
 	self.usedCargoMass = 0
 
 	-- Initialize property variables on owning ship for backwards compatibility
-	ship:setprop("totalCargo", self:GetTotalSpace())
-	ship:setprop("usedCargo", 0)
+	-- don't initialize them if they're already created after first save
+	if not self.ship:hasprop("totalCargo") then
+		ship:setprop("totalCargo", self:GetTotalSpace())
+	end
+	
+	if not self.ship:hasprop("usedCargo") then
+		ship:setprop("usedCargo", 0)
+	end
 
 	-- TODO: stored commodities should be represented as array of { name, count, meta } entries
 	-- to allow for e.g. tracking stolen/scooped cargo, or special mission-related cargoes
@@ -99,7 +105,7 @@ end
 function CargoManager:AddCommodity(type, count)
 	-- TODO: use a cargo volume metric with variable mass instead of fixed 1m^3 == 1t
 	local required_space = (type.mass or 1) * (count or 1)
-
+	
 	if self:GetFreeSpace() < required_space then
 		return false
 	end
@@ -245,6 +251,7 @@ end
 function CargoManager:Unserialize()
 	setmetatable(self, CargoManager.meta)
 	self.listeners = {}
+	
 	return self
 end
 
--- a/data/modules/Scout/ScanManager.lua
+++ b/data/modules/Scout/ScanManager.lua
@@ -221,17 +221,29 @@ function ScanManager:UpdateSensorEquipInfo()
 
 end
 
+---@param ScanData scan
+---@return Body
+function ScanManager:GetScanBodyByType(scan)
+	scan = scan or self.activeScan
+	assert(scan, "ScanManager:GetScanBodyByType: Error: scan was nil")
+	assert(self.scanMap[scan.id], "ScanManager:GetScanBodyByType: Error: scan doesn't belong to current ScanManager")
+	return scan.orbital and scan.bodyPath:GetSystemBody().body or self.ship.frameBody
+end
+
 ---@param body Body
+---@param ScanData scan
 ---@return number altitude
 ---@return number resolution
-function ScanManager:GetBodyState(body)
-	assert(self.activeScan)
+function ScanManager:GetBodyState(body, scan)
+	scan = scan or self.activeScan
+	assert(scan, "ScanManager:GetBodyState: Error: scan was nil")
+	assert(self.scanMap[scan.id], "ScanManager:GetBodyState: Error: scan doesn't belong to current ScanManager")
 
 	local altitude
 	local radius = body:GetSystemBody().radius
 	local sensor = self.activeSensor
 
-	if self.activeScan.orbital then
+	if scan.orbital then
 		-- altitude above sea-level
 		altitude = self.ship:GetPositionRelTo(body):length() - radius
 	else
@@ -261,7 +273,8 @@ function ScanManager:GetState()
 	end
 
 	if self.activeScan then
-		local altitude, resolution = self:GetBodyState(self.ship.frameBody)
+		local body = self:GetScanBodyByType()
+		local altitude, resolution = self:GetBodyState(body)
 		local isInRange = resolution <= self.activeScan.minResolution
 
 		if altitude > self.activeSensor.minAltitude then
@@ -425,11 +438,18 @@ function ScanManager:CanScanBeActivated(id)
 
 	if scan.complete then return false end
 
-	local frameBody = self.ship.frameBody
-	if not frameBody then return false end
+	local body = self:GetScanBodyByType(scan)
+	if not body then return false end
 
-	if frameBody.path ~= scan.bodyPath then
+	if body.path ~= scan.bodyPath then
 		return false
+	end
+	
+	if scan.orbital then
+		local altitude, resolution = self:GetBodyState(body, scan)
+		if(resolution > (scan.minResolution * 1.10)) then
+			return false
+		end
 	end
 
 	if scan.orbital ~= self.activeSensor.orbital then
@@ -490,7 +510,9 @@ end
 ---@package
 function ScanManager:OnEnteredFrame(body)
 	if self.activeScan then
-
+		if self.activeScan.orbital then
+			return
+		end
 		local frameBody = body.frameBody
 
 		if not frameBody or frameBody.path ~= self.activeScan.bodyPath then
@@ -501,22 +523,24 @@ function ScanManager:OnEnteredFrame(body)
 end
 
 -- Start the scanner callback
+---@param force boolean
 ---@package
-function ScanManager:StartScanCallback()
+function ScanManager:StartScanCallback(force)
 	local updateRate = self.activeScan.orbital and ORBITAL_SCAN_UPDATE_RATE or SURFACE_SCAN_UPDATE_RATE
 
 	-- Immediately trigger a scan update for responsiveness
 	self:OnUpdateScan(self.activeScan)
 
 	-- Don't queue a new scan callback if we're already running one at the right frequency
-	if updateRate == self.activeCallback then
+	-- Force this check if this function was called from ScanManager:Unserialize() because
+	-- callback is not yet setted up 
+	if updateRate == self.activeCallback and not force then
 		return
 	end
 
 	self.activeCallback = updateRate
 
 	Timer:CallEvery(updateRate, function()
-
 		-- cancel this callback if the parameters have changed (it's been orphaned)
 		if not self.activeScan or not self.activeSensor or self.activeCallback ~= updateRate then
 			return true
@@ -532,11 +556,11 @@ end
 ---@return boolean cancel
 ---@package
 function ScanManager:OnUpdateScan(scan)
-	local body = self.ship.frameBody
+	local body = self:GetScanBodyByType(scan)
 
 	-- Somehow we don't have a valid body to be scanning
 	if not body then
-		logWarning("ScanManager: owning ship does not have a frameBody to scan!")
+		logWarning("ScanManager: owning ship does not have a system body to scan!")
 		self:ClearActiveScan()
 		return true
 	end
@@ -544,9 +568,18 @@ function ScanManager:OnUpdateScan(scan)
 	local radius = body:GetSystemBody().radius
 	local altitude, resolution = self:GetBodyState(body)
 	local currentScanPos = self.ship:GetPositionRelTo(body):normalized()
+	
+	if scan.orbital then
+		if(resolution > (scan.minResolution * 1.10)) then
+			self:ClearActiveScan()
+			return true
+		end
+	end
 
 	-- Determine if we're currently in range to record scan data
 	local withinParams = resolution <= scan.minResolution and altitude > self.activeSensor.minAltitude
+	
+	
 
 	-- Use great-arc distance to calculate the amount of scan coverage in square meters
 	-- distance = Δσ * radius = arctan( |n1 ⨯ n2| / n1 · n2 ) * radius
@@ -610,7 +643,7 @@ function ScanManager:Unserialize()
 
 	-- Restore the scanning callback on loading a saved game
 	if self.activeCallback then
-		self:StartScanCallback()
+		self:StartScanCallback(true)
 	end
 
 	return self
--- a/data/modules/Scout/ScanDisplay.lua
+++ b/data/modules/Scout/ScanDisplay.lua
@@ -94,8 +94,9 @@ function scanDisplay:drawDebug()
 	minRes = ui.dragFloat("MinRes", minRes, 1.0, 0.1, 100, "%f")
 	coverage = ui.dragFloat("Coverage", coverage, 0.5, 0.5, 100, "%fkm^2")
 
-	if scanMgr:GetActiveScan() then
-		local altitude, resolution = scanMgr:GetBodyState(self.ship.frameBody)
+	local activeScan = scanMgr:GetActiveScan()
+	if activeScan then
+		local altitude, resolution = scanMgr:GetBodyState(scanMgr:GetScanBodyByType())
 
 		ui.text("Current Alt: " .. altitude)
 		ui.text("Current Res: " .. resolution)
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -375,17 +375,17 @@ bool Game::UpdateTimeAccel()
 
 					vector3d toBody = m_player->GetPosition() - b->GetPositionRelTo(m_player->GetFrame());
 					double dist = toBody.Length();
-					double rad = b->GetPhysRadius();
+					double rad = std::max(b->GetPhysRadius(), 10000.0);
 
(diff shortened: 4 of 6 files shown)


Max5377:
Probably going to do more fixes to scanning mission. I have a orbital scanner, which says that maximum scanning altitude is 5.71Mm, but as soon as my frame body changes (~2.3Mm altitude) I can't scan planet. Unless I misunderstood something.

Max5377:
I guess this is how it meant to be:

https://github.com/pioneerspacesim/pioneer/assets/69468517/74ba7107-716c-4da2-ae5f-efb4119a9167

Max5377:
Had an occasion where I need to do two scans from the same planet, but I can do only one scan. Oh well.

Web-eWorks:
Personally I think the constraint of one active scan is acceptable for multiple reasons:
- The probability of getting multiple "compatible" scans of the same body from the same station is intended to be extremely low;
- The point of the scan gameplay loop is to incentivize the player to manually adjust their orbit to the maximum permissible for the contract, to optimize the amount of "real time" they spend scanning the body. With simultaneous scans with differing maximum orbits, the gains from smart maneuvering and manual flight mean less than brute-forcing time acceleration.
- It avoids a pathological "min-maxing" behavior of only taking scan contracts for bodies where you have multiple scans at once. The payout vs. time spent scanning should be the metric that scan missions are balanced around, and that goes completely out-of-control if you can accomplish 2 or 3 scans in the same timeframe it takes to do one (exploiting).

The point of the work being done on Pioneer's economy is to provide consistent rewards to the player for their time investment no matter which gameplay loop they want to try, not have most activities be a pointless grind for pennies and a select few exploits/activities trivialize the rest of the game, as in Elite Dangerous.

Max5377:
Thank you for your opinion. I agree with you on exploiting and min-maxing part. I can also add that performing multiple scans with just one scanner when you have missions that require scanning in different conditions sounds like nonsense in itself. How is this even supposed to be possible?
P.S.: after I wrote this, I had the thought "do multiple scans that require multiple scanners", but this will probably also force player to choose one good strategy or to min-max.

data/modules/Scout/ScanManager.lua
```diff
@@ -425,12 +438,19 @@ function ScanManager:CanScanBeActivated(id)
 
 	if scan.complete then return false end
 
-	local frameBody = self.ship.frameBody
-	if not frameBody then return false end
+	local body = self:GetScanBodyByType(scan)
+	if not body then return false end
```
Web-eWorks, line 442: If the player leaves the frame of influence of the body they are scanning, they are no longer in orbit of that body. They are instead orbiting the parent body and will slowly begin drifting away from the scanned body unless they enter the frame of influence again.

The problem at hand is that the frame of influence of single planets or moons without any children is far smaller than it should be. This is due to Space.cpp#761 using `4 * bodyRadius` instead of calculating the Hill sphere of the body for its correct gravitational frame of influence.

The equation for calculating the Hill radius (in pseudocode) is:

```c
semiMajorAxis * (1.0 - eccentricity) * pow(mass / (3.0 * parent.mass), 1.0 / 3.0)
```

Doing a quick evaluation, using the Hill radius for frame-of-influence should ensure that most if not all bodies are fully scannable while remaining inside their frame of influence, as the Hill radius for small bodies can be 10x or larger than the current frame of influence calculation.

data/libs/CargoManager.lua
```diff
@@ -245,6 +245,12 @@ end
 function CargoManager:Unserialize()
 	setmetatable(self, CargoManager.meta)
 	self.listeners = {}
+	
+	self.ship:setprop("totalCargo", self:GetTotalSpace())
+	self.ship:setprop("usedCargo", self.usedCargoSpace)
+
+	self.ship:UpdateEquipStats()
+
```
Web-eWorks, line 253: This code is only necessary because the CargoManager constructor is overwriting the stored properties of the ship on lines 30-31. Those can be wrapped in `if not self.ship:hasprop("totalSpace")`, which obviates the need for this patch entirely.

There's not really a clean way to handle *not* constructing a default CargoManager before a replacement is de-serialized, so I'd recommend implementing the above solution instead to avoid mutating the ship object entirely.

Web-eWorks (changes requested):
Of the changes here in the PR, I've left two comments here where I'd like to see the solutions amended to better address the underlying issues. Otherwise, this seems fairly sensible and non-invasive; thank you for tackling these issues!

Web-eWorks:
Also, if you're willing, I will push a commit to this PR once you've addressed the above PR feedback that amends the time acceleration limits in Game.cpp to allow higher "default" time acceleration while near bodies, without needing to Ctrl+Click and override time acceleration.

Web-eWorks:
As a minor nitpick for future contributions, we prefer the use of either `hyphen-delimited-lower-case` or `PascalCase` branch names without the use of any special punctuation characters. Adding spaces or punctuation characters makes your branches more difficult to work with in a terminal while performing a review.

data/modules/Scout/ScanManager.lua
Max5377: Do it actually needed since surface scan's are already performed on the low altitude and for orbtial scan's, body borrows from scan's body path and the distance is checked by current resolution:

`function ScanManager:GetScanBodyByType(scan)`
```
return scan.orbital and scan.bodyPath:GetSystemBody().body or self.ship.frameBody
```

`function ScanManager:CanScanBeActivated(id)`
```
if scan.orbital then
        local altitude, resolution = self:GetBodyState(body, scan)
	if(resolution > (scan.minResolution * 1.10)) then
		return false
	end
end
```
?
But feel free to correct me if I misunderstood.

data/libs/CargoManager.lua
Max5377: I tested this, wrapping also helps, I will change it, but I'm still on the fence about `UpdateEquipStats`. I put it here as an equivalent in `CargoManager:OnShipTypeChanged()` and I think it can be safely deleted too, after some testing, but what do you think? Here's the code for it:
```
void Ship::UpdateEquipStats()
{
	PropertyMap &p = Properties();

	m_stats.used_capacity = p.Get("mass_cap");
	m_stats.used_cargo = 0;

	m_stats.free_capacity = m_type->capacity - m_stats.used_capacity;
	m_stats.static_mass = m_stats.used_capacity + m_type->hullMass;

	p.Set("usedCapacity", m_stats.used_capacity);
	p.Set("freeCapacity", m_stats.free_capacity);

	p.Set("totalMass", m_stats.static_mass);
	p.Set("staticMass", m_stats.static_mass);

	float shield_cap = p.Get("shield_cap");
	m_stats.shield_mass = TONS_HULL_PER_SHIELD * shield_cap;
	p.Set("shieldMass", m_stats.shield_mass);

	UpdateFuelStats();
	UpdateGunsStats();

	unsigned int thruster_power_cap = p.Get("thruster_power_cap");
	const double power_mul = m_type->thrusterUpgrades[Clamp(thruster_power_cap, 0U, 3U)];
	m_propulsion->SetThrustPowerMult(power_mul, m_type->linThrust, m_type->angThrust);

	m_stats.hyperspace_range = m_stats.hyperspace_range_max = 0;
	p.Set("hyperspaceRange", m_stats.hyperspace_range);
	p.Set("maxHyperspaceRange", m_stats.hyperspace_range_max);

	m_stats.atmo_shield_cap = p.Get("atmo_shield_cap");
	m_stats.radar_cap = p.Get("radar_cap");
	m_stats.fuel_scoop_cap = p.Get("fuel_scoop_cap");
	m_stats.cargo_life_support_cap = p.Get("cargo_life_support_cap");
	m_stats.hull_autorepair_cap = p.Get("hull_autorepair_cap");
}
```

Max5377:
I think disabling Ctrl+Click can be added as an option in setting's menu like "Always force time acceleration". I can add that if you want, since I already done some changes to `game.lua` and `settings-window.lua` like pause menu now remembers last speed (you probably saw it on IRC channel) and I moved `Input.EnableBindings` and `SetTimeAcceleration` from `settings-window.lua`, so `game.lua` is the only responsible now for setting this and just tracks when `optionsWindow` is closed. It looks like this:
```
ui.registerHandler('game', function(delta_t)
...
       if currentView == "world" and ui.escapeKeyReleased(true) then
	       if not ui.optionsWindow.isOpen then
	       lastTimeAcceleration = Game.GetTimeAcceleration() ~= Game.GetRequestedTimeAcceleration() and        Game.GetRequestedTimeAcceleration() or Game.GetTimeAcceleration()
	Game.SetTimeAcceleration("paused")
		ui.optionsWindow:open()
		Input.EnableBindings(false)
		settingsMenuWasOpened = true
		else
			ui.optionsWindow:close()
		end
	end

	callModules('modal')
		
	if settingsMenuWasOpened and not ui.optionsWindow.isOpen then
		settingsMenuWasOpened = false
		local forceTimeAcceleration = ui.ctrlHeld() and (lastTimeAcceleration ~= "paused")
		Game.SetTimeAcceleration((lastTimeAcceleration == "paused") and "1x" or lastTimeAcceleration, forceTimeAcceleration)
		Input.EnableBindings()
	end
...
end
```
`ui.ctrlHeld()` for `forceTimeAcceleration` doesn't work right now since if you press CTRL and then ESC, it just opens Start Menu on Windows, but will work if you press Close button instead.
Also, spotted another bug while testing `UpdateShipEquipment`, `maxHyperspaceRange` returns zero after save loading for player, but `GetHyperspaceRange` returns normal values. Seems like, `maxHyperspaceRange` is used in displaying ship's info (Hyperspace range max will be zero) and in `DeliverPackage` module:
```
local onCreateBB = function (station)
       ...
       local canHyperspace = Game.player.maxHyperspaceRange > 0
       ...
end
```

data/modules/Scout/ScanManager.lua
Web-eWorks: Yes, this change is needed. I can push it as a new commit to this branch if you're interested; but essentially it modifies how Pioneer's orbital physics simulation works to make it both more physically correct and also more fun for the player.

data/libs/CargoManager.lua
Web-eWorks: `UpdateEquipStats` is no longer needed if you make no changes to ship properties in the unserialize method. IIRC it's also internally called from C++ on loading a saved game.

data/modules/Scout/ScanManager.lua
Max5377: Yes, do it.

data/libs/CargoManager.lua
Max5377: I will do force-push then, to wrap ship properties in the `CargoManager`'s constructor and remove `UpdateEquipStats`. But first I will try to find why `maxHyperspaceRange` is zero for player.

Web-eWorks:
> I think disabling Ctrl+Click can be added as an option in setting's menu like "Always force time acceleration".

I would strongly prefer not to do this.

The Ctrl+Click is intended to override the game's limits as to what's considered a "safe" time acceleration both in-world and from a developer perspective. I'm willing to amend those limits slightly given the optimizations and improvements that have been made since Pioneer first introduced them, however the distinction between forcing a time acceleration setting and requesting one is an important one to have. If clicking on the buttons always forces time acceleration, the player will experience many unintended deaths and other program errors. For an example, in a low orbit around a Titan-sized body at 10,000x time acceleration your orbit will shift very drastically due to numerical error.

I'll address the other comments etc. later once I've gotten a fresh mind to go over them.

Max5377:
`void Ship::UpdateLuaStats()`  in this line:
```
	if (hyperclass) {
		std::tie(m_stats.hyperspace_range_max, m_stats.hyperspace_range) =
			LuaObject<Ship>::CallMethod<double, double>(this, "GetHyperspaceRange");
	}
```
`m_stats.hyperspace_range_max`, `m_stats.hyperspace_range` are mixed up since `GetHyperspaceRange` returns `range` and `range_max`, `range` is zero at this time, but will be initialized later. That's why `maxHyperspaceRange` was zero.

Web-eWorks:
You appear to have a partial manual rebase against master in Ship.cpp - this branch currently doesn't compile because it's missing the change to Ship.h from master. That said, I've tested and pushed the commits I spoke of prior - you should be able to rebase this branch against master in preparation for merge.

Web-eWorks:
Excellent work, did a quick playtest and everything seems to work correctly!

yump:
Thank you so much @Max5377, for finding and fixing the cause of #5534. I had to stop playing the game after spending two weekends trying to wrap my brain around all the things that participate in cargo/mass calculations. I was so stuck on this that I never got around to poking packagers about #5574.

Max5377:
@yump Have fun) Don't forget to also thank @Web-eWorks for giving tip for refining my solution. I don't have Linux, so I can't look why this issue happenes.

PR #5638, opened 2023-10-08
Lua io and FileSystem uri changes

JonBooth78:
I'd like to be able to use lua standard `io` functionality as it's familiar to people who know lua and also easy to find examples and documentation on - and we don't have to build our own version.

However, this needs to be carefully sandboxed so we don't allow scripts and mods to run wild on a users machine.

So this will look to add the `io` module but redact and modify it as appropriate to keep it safe.

Keeping it safe means only allowing access to the user and data folders and the data folder can only have read only access.

Remaining todo list (at the time of writing):
- Unify and standardize the filenames across the API to accept a uri-style schema so you prefix a path with either `user://` or `data://` to indicate where you want to work.
- Do all FileSystem implementation in C++ rather than the additive shim I got to in lua

### DIFF
--- a/src/lua/LuaFileSystem.cpp
+++ b/src/lua/LuaFileSystem.cpp
@@ -7,15 +7,90 @@
 #include "LuaObject.h"
 #include "Pi.h"
 
-/*
- * Interface: FileSystem
- *
- * A global table that provides access to the filesystem.
- *
- * This interface is protected. It can only be used by scripts in Pioneer's
- * data directory. Mods and scripts in the user directory that try to use it
- * will get a Lua error.
- */
+#include "core/StringUtils.h"
+
+struct SplitURI
+{
+	std::string_view full_path;
+	std::string_view scheme;
+	std::string_view rel_path;
+};
+
+SplitURI ParseURIArg(lua_State* l, int numArg)
+{
+	SplitURI res;
+	res.full_path = LuaPull<std::string_view>(l, numArg);
+	size_t sep_pos = res.full_path.find("://");
+	if (sep_pos != std::string_view::npos)
+	{
+		res.scheme = res.full_path.substr(0, sep_pos);
+		sep_pos += 3;
+		res.rel_path = res.full_path.substr(sep_pos);
+	}
+	return res;
+}
+
+struct FileSource
+{
+	FileSystem::FileSource* source;
+	bool					is_read_write;
+};
+
+FileSource DetermineFileSource(std::string_view scheme)
+{
+	FileSource res;
+	if (compare_ci(scheme, "user"))
+	{
+		res.source = &FileSystem::userFiles;
+		res.is_read_write = true;
+	} else if (compare_ci(scheme, "data"))
+	{
+		res.source = &FileSystem::gameDataFiles;
+		res.is_read_write = false;
+	} else
+	{
+		res.source = nullptr;
+		res.is_read_write = false;
+	}
+	return res;
+}
+
+static lua_CFunction l_original_io_open = nullptr;
+
+void LuaFileSystem::register_raw_io_open_function(lua_CFunction open)
+{
+	l_original_io_open = open;
+}
+
+int LuaFileSystem::l_patched_io_open(lua_State* L)
+{
+	SplitURI split_uri = ParseURIArg(L, 1);
+	const FileSource fs = DetermineFileSource(split_uri.scheme);
+	if (!fs.source)
+	{
+		return luaL_error(L, "Path '%s' does not contain a valid scheme, must be user:// or data://", split_uri.full_path.data());
+	}
+
+	std::string_view mode_str = LuaPull<std::string_view>(L, 2, "r");
+	if (mode_str.find_first_of("aw+") != std::string_view::npos && !fs.is_read_write) {
+		return luaL_error(L, "File '%s' is read-only.", split_uri.full_path.data());
+	}
+
+	try
+	{
+		::std::string abs_path = fs.source->Lookup(std::string(split_uri.rel_path)).GetAbsolutePath();
+		LuaPush(L, abs_path);
+		lua_replace(L, 1);
+
+		const int rv = l_original_io_open(L);
+
+		return rv;
+	}
+	catch (const std::invalid_argument&)
+	{
+		return luaL_error(L, "'%s' is not a valid file in its scheme root' - Is the file location within the root?", split_uri.full_path.data());
+	}
+}
 
 static void push_date_time(lua_State *l, const Time::DateTime &dt)
 {
@@ -36,13 +111,12 @@ static void push_date_time(lua_State *l, const Time::DateTime &dt)
 /*
  * Function: ReadDirectory
  *
- * > local files, dirs = FileSystem.ReadDirectory(root, path)
+ * > local files, dirs = FileSystem.ReadDirectory(path)
  *
  * Return a list of files and dirs in the specified directory.
  *
  * Parameters:
  *
- *   root - a <FileSystemRoot> constant for the root of the dir to read from
  *   path - optional. a directory under the root
  *
  * Returns:
@@ -60,61 +134,46 @@ static void push_date_time(lua_State *l, const Time::DateTime &dt)
  */
 static int l_filesystem_read_dir(lua_State *l)
 {
-	LuaFileSystem::Root root = static_cast<LuaFileSystem::Root>(LuaConstants::GetConstantFromArg(l, "FileSystemRoot", 1));
-	std::string path;
-	if (lua_gettop(l) > 1)
-		path = luaL_checkstring(l, 2);
-
-	FileSystem::FileSource *fs = nullptr;
-	switch (root) {
-	case LuaFileSystem::ROOT_USER:
-		fs = &FileSystem::userFiles;
-		break;
-
-	case LuaFileSystem::ROOT_DATA:
-		fs = &FileSystem::gameDataFiles;
-		break;
-
-	default:
-		assert(0); // can't happen
-		return 0;
+	SplitURI split_uri = ParseURIArg(l, 1);
+	const FileSource fs = DetermineFileSource(split_uri.scheme);
+	if (!fs.source)
+	{
+		return luaL_error(l, "Path '%s' does not contain a valid scheme, must be user:// or data://", split_uri.full_path.data());
 	}
 
-	assert(fs);
+	// shame these methods can't take a string_view.
+	std::string rel_path = std::string(split_uri.rel_path);
 
 	{
 		try {
-			const FileSystem::FileInfo &info = fs->Lookup(path);
+			FileSystem::FileInfo info = fs.source->Lookup(rel_path);
 			if (!info.IsDir()) {
-				luaL_error(l, "'%s' is not a directory", path.c_str());
-				return 0;
+				return luaL_error(l, "'%s' is not a directory", split_uri.full_path.data());
 			}
 		} catch (const std::invalid_argument &) {
-			luaL_error(l, "'%s' is not a valid path", path.c_str());
-			return 0;
+			return luaL_error(l, "'%s' is not a valid path", split_uri.full_path.data());
 		}
 	}
-
-	FileSystem::FileEnumerator files(*fs, FileSystem::FileEnumerator::IncludeDirs);
-	files.AddSearchRoot(path);
 
 	lua_newtable(l);
 	int filesTable = lua_gettop(l);
 	lua_newtable(l);
 	int dirsTable = lua_gettop(l);
 
-	for (; !files.Finished(); files.Next()) {
-		const FileSystem::FileInfo &info = files.Current();
+	int dirs_len = 0;
+	int files_len = 0;
 
+	for (const auto &info : fs.source->Enumerate(rel_path, FileSystem::FileEnumerator::IncludeDirs))
+	{
 		lua_newtable(l);
 		pi_lua_settable(l, "name", info.GetName().c_str());
 		push_date_time(l, info.GetModificationTime());
 		lua_setfield(l, -2, "mtime");
 
 		if (info.IsDir())
-			lua_rawseti(l, dirsTable, lua_rawlen(l, dirsTable) + 1);
+			lua_rawseti(l, dirsTable, ++dirs_len);
 		else
-			lua_rawseti(l, filesTable, lua_rawlen(l, filesTable) + 1);
+			lua_rawseti(l, filesTable, ++files_len);
 	}
 
 	return 2;
@@ -146,9 +205,58 @@ static int l_filesystem_join_path(lua_State *l)
 		lua_pushlstring(l, path.c_str(), path.size());
 		return 1;
 	} catch (const std::invalid_argument &) {
-		luaL_error(l, "result is not a valid path");
-		return 0;
+		return luaL_error(l, "result is not a valid path");
 	}
+}
+
+
+/*
+ * Function: MakeDirectory
+ *
+ * > local success = FileSystem.MakeDirectory( dir_name )
+ *
+ * Creating the given directory if it's missing, returning a boolean
+ * indicating success
+ *
+ * Availability:
+ *
+ *   Oct 2023
+ *
+ * Status:
+ *
+ *   experimental
+ */
+static int l_filesystem_make_directory(lua_State *l)
+{
+	SplitURI split_uri = ParseURIArg(l, 1);
+	const FileSource fs = DetermineFileSource(split_uri.scheme);
+	if (!fs.source)
+	{
+		return luaL_error(l, "Path '%s' does not contain a valid scheme, must be user:// or data://", split_uri.full_path.data());
+	}
+
+	if (!fs.is_read_write)
+	{
+		return luaL_error(l, "'%s' does not support file operations that modify it, only things in the user folder do", split_uri.full_path.data());
+	}
+
+	try {
+
+		FileSystem::FileSourceFS& wfs = *static_cast<FileSystem::FileSourceFS*>(fs.source);
+
+		// shame these methods can't take a string_view.
+		std::string rel_path = std::string(split_uri.rel_path);
+		// At the moment, anything with a filesourceFS is also writeable, so the write permission test
+		// above also validates this...
+		wfs.MakeDirectory(rel_path);
+
+		FileSystem::FileInfo f = wfs.Lookup(rel_path);
+
+		lua_pushboolean(l, f.IsDir());
+	} catch (const std::invalid_argument&) {
+		return luaL_error(l, "unable to create directory '%s' the argument is invalid", split_uri.full_path.data());
+	}
+	return 1;
 }
 
 void LuaFileSystem::Register()
@@ -160,6 +268,8 @@ void LuaFileSystem::Register()
 	static const luaL_Reg l_methods[] = {
 		{ "ReadDirectory", l_filesystem_read_dir },
 		{ "JoinPath", l_filesystem_join_path },
+		{ "MakeDirectory", l_filesystem_make_directory },
+		{ "Open", l_patched_io_open },
 		{ 0, 0 }
 	};
 
--- a/src/lua/LuaFileSystem.h
+++ b/src/lua/LuaFileSystem.h
@@ -4,13 +4,13 @@
 #ifndef _LUAFILESYSTEM_H
 #define _LUAFILESYSTEM_H
 
+#include "lua/Lua.h"
+
 namespace LuaFileSystem {
 	void Register();
 
-	enum Root { // <enum scope='LuaFileSystem' name=FileSystemRoot prefix=ROOT_ public>
-		ROOT_USER,
-		ROOT_DATA
-	};
+	void register_raw_io_open_function(lua_CFunction open);
+	int l_patched_io_open(lua_State* L);
 } // namespace LuaFileSystem
 
 #endif
--- a/data/meta/Constants.lua
+++ b/data/meta/Constants.lua
@@ -250,13 +250,6 @@ Constants.DetailLevel = {
 	[5] = "VERY_HIGH",
 }
 
--- A <Constants.FileSystemRoot> string
----@enum FileSystemRoot
-Constants.FileSystemRoot = {
-	[1] = "USER",
-	[2] = "DATA",
-}
-
 -- A <Constants.PiGuiFaceFlags> string
 ---@enum PiGuiFaceFlags
 Constants.PiGuiFaceFlags = {
--- a/data/meta/FileSystem.lua
+++ b/data/meta/FileSystem.lua
@@ -1,0 +1,49 @@
+-- Copyright © 2008-2023 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+-- This file implements type information about C++ classes for Lua static analysis
+-- This is used in FileSyestem.lua whcih then extends that.
+
+---@meta
+
+---@class FileSystem
+local FileSystem = {}
+
+---@param path string The directory to read the contents of.
+---@return string[] files  A list of files as full paths from the root
+---@return string[] dirs   A list of dirs as full paths from the root
+---
+--- Example:
+---  > local files, dirs = FileSystem.ReadDirectory("user://savefiles")
+function FileSystem.ReadDirectory(path) end
+
+--- Join the passed arguments into a path, correctly handling separators and .
+--- and .. special dirs.
+---
+---@param arg string[]  A list of path elements to be joined
+---@return string       The joined path elements
+function FileSystem.JoinPath( ... ) end
+
+---@param dir_name string The name of the folder to create
+---@return boolean Success
+function FileSystem.MakeDirectory( dir_name ) end
+
+--- Wrapper for our patched io.open that ensures files are opened inside the sandbox.
+--- Prefer using this to io.open
+---
+--- Files in the user folder can be read or written to
+--- Files in the data folder are read only.
+---
+---
+---
+--- Example:
+--- > f = FileSystem.Open( "user://my_file.txt", "w" )
+--- > f:write( "file contents" )
+--- > f:close()
+--- 
+---@param filename string   The name of the file to open, must start either user:// or data://
+---@param mode string|nil   The mode to open the file in, defaults to read only. Only user location files can be written
+---@return file             A lua io file
+function FileSystem.Open( filename ) end
+
+return FileSystem
--- a/data/modules/AutoSave/AutoSave.lua
+++ b/data/modules/AutoSave/AutoSave.lua
@@ -6,7 +6,7 @@ local FileSystem = require 'FileSystem'
 local max_autosaves = 9
 
 local function PickNextAutosave()
-	local ok, files, _ = pcall(FileSystem.ReadDirectory, 'USER', 'savefiles')
+	local ok, files, _ = pcall(FileSystem.ReadDirectory, 'user://savefiles')
 	if not ok then
 		print('Error picking auto-save')
 		return '_autosave1'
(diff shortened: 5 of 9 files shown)


src/lua/LuaFileSystem.cpp
```diff
@@ -151,6 +191,42 @@ static int l_filesystem_join_path(lua_State *l)
...
+ * indicating success
+ *
+ * Availability:
+ *
+ *   ????
+ *
+ * Status:
+ *
+ *   experimental
+ */
```
impaktor, line 210: regarding `????` 

I've often wondered if the "availability" and "status" fields are something we really need to specify? Might have been more relevant when we did nightly builds, and alpha-release schedule?

Web-eWorks: At this point I'd use neither; at best the `availability` field helps modders who are writing code against the 'bleeding edge' of Pioneer, but we provide very little in terms of a stable API and there's absolutely no intention for old versions of Pioneer to remain "in use" once a new release is cut. The `status` field is actually more useful to document parts of the API that exist only for backwards compatibility or as hacks, but we should strive to remove those where present.

I do get a kick out of seeing `Status: experimental` on widely-used functions implemented in ~2013 though, it neatly proves the saying "nothing more permanent than a temporary solution".

JonBooth78: lol, yeah, I just put that in as I don't have any clue what to put in there and meant to ask

JonBooth78:
@Web-eWorks, I think I've got the API looking the way it should now.

Not yet happy with the details of the way I've parsed out the URI as I think I'm doing more than I should as some of it's fairly redundant.  Still, I think the broad strokes of the approach is probably OK.

Max5377:
Try to change `FileSystem::FileSourceFS* ParsedFilePath::ValidateWritePermission(lua_State *l) const` to this:
```
if (m_writeableFS == nullptr)
{
	luaL_error(l, "'%s' does not support file operations that modify it, only things in the user folder do", m_luaPath.data());
}
return m_writeableFS;
```
And you're also got a warning in gcc: `warning: catching polymorphic type ‘class std::invalid_argument’ by value [-Wcatch-value=]` in this line `std::string ParsedFilePath::CalculateAbsolutePath(lua_State* l) const`:
`catch (std::invalid_argument e)`
change that to this:
`catch (std::invalid_argument const &e)`
[explanation in the accepted answer](https://stackoverflow.com/questions/62030341/what-is-this-catching-polymorphic-type-x-by-value-wcatch-value)

Web-eWorks:
`ParsedFilePath` is definitely very heavyweight for what it does. I'd split that into two helper functions, one which just does a `find_first_of("://")` to split the string into protocol fragment and relative path, and another which takes a `string_view` of the protocol fragment and returns a lightweight struct containing a `FileSource *` and a bool indicating read/write capability. The `FileSource *` can be converted via `static_cast<FileSourceFS *>`.

Arguably a nicer way to handle this would be to promote the file-write interface in FileSourceFS to virtual method stubs in FileSource, with a query method to check if the FileSource supports writing. Another way would be to keep an internal enum for `DATA` / `USER` and return that alongside the FileSource pointer to check if the given file can be opened read-write.

In general, I'd just gently mention that not everything needs to be a class just because it's written in C++ :smile:

I probably don't need to mention this, but my general advice is to always make sure to iterate in search of the right granularity and boundary to divide the logic into functions and classes. The best APIs are driven by "what do I need to accomplish this task", not "I might need all of these things later in some unforeseen usecase".

src/lua/LuaFileSystem.cpp
```diff
@@ -7,15 +7,112 @@
...
+	m_fs_path.remove_prefix(m_scheme.length());
+}
+
+FileSystem::FileSourceFS* ParsedFilePath::ValidateWritePermission(lua_State *l) const
+{
+	if (m_writeableFS != nullptr)
+	{
+		return m_writeableFS;
+	}
+	luaL_error(l, "'%s' does not support file operations that modify it, only things in the user folder do", m_luaPath.data());
```
Web-eWorks, line 68: By the by, `luaL_error` is a termination point for the function and should be treated as though it has `noreturn` semantics. The following `return` statements will never be reached, and there's a potential for memory leaks if you allocate something on the heap in the function that isn't automatically destroyed before the call to `luaL_error`.

For this reason, I'd highly recommend only calling `luaL_error` directly in a lua CFunction, using the idiom `return luaL_error(...);`.

Web-eWorks (changes requested):

JonBooth78:
Thanks folks.

Some quick notes:
- I think I could have written a better comment above, I don't think the implementation details of `LuaFileSystem.cpp` are ready, I wanted to check we're aligned on how it will look from the lua side so worry I've poorly made use of your time.
- The compile errors/warnings :-( I saw these come through from the autobuild (I think, haven't really looked to deeply) but wasn't too worried about it because of the above.  Annoying that even today we can't get compiler vendors to agree on the wierdness of the stl, templates and so forth.  This might be a good reason for me to re-invest in getting the ming2-gcc toolchain working on windows locally so I can more easily check this kind of thing.
- redaction of io constants, yep, good catch, that's a must.
- I tossed up about if I need to include code paths that can never be reached due to the lua error code, it felt cleaner to have them there but if we have a convention to not worry about it, that's fine too.  I don't hold any real preference.

Now into class vs functions with an associated data type and some personal opinions around the design options, non held strongly enough to be a problem but I'd like to take the opportunity to share.

I realized fairly quickly that the dichotomy of `FileSource` vs `FileSourceFS` is a tricky one and thought about this a bit.

There are a few ways to tackle this, with two extremes.  Either just store a `FileSource` pointer and then use a `dynamic_cast` with the associated gubbins to do the type checking (assuming we even have RTTI available), or essentially do our own version of RTTI with a flag to indicate what type it actually is.

In C++ as a strongly typed language my personal opinion is that the language feature RTTI is not the right way.  I believe they've fixed a lot of the memory bloat and performance issues around it over the years but I still find it an antipattern.

So that leaves us with, at minimum having a some kind of compound object (class/struct) to represent where we got to parsing the incoming URI.  I think there is value in a third one too, the root-relative path, that as we parse the root we can pretty much get it for free and it's going to be needed 99% of the happy path time.

I believe that for convenience and code brevity sake, it makes sense to then have some standard ways to query this, so we write once, use often and it will make the client code (that using it) briefer.

However, one of the main values from doing this is that we can validate and throw Lua errors as appropriate within those validation methods, which flies against the comment above suggesting we would only throw Lua errors from the top level of a Lua method and not for anything called internally.  I wonder if for readability and keeping compilers happy about dead code paths, we could just throw an exception within helper functions; it's still dead code but then our c++ compiler will understand it and it gives some flexibility to allow such helper functions along with the brevity/clarity concerns.

So given those pieces above, we're left with the need for an aggregation of variables  and some methods that act upon them.  I'm currently of the opinion that this concept is one of the key ones behind OO programming and in a language that does support it, one should use it.   This has benefits of familiarity, clarity, brevity and I think quite importantly discoverability, as the user of a class benefits from all the built in goodies of an IDE, should they be using one, e.g. intelisense.  It also helps demark the intent and implementation around the responsibilities of the different components more clearly for any reader and so will help avoid code smells in the future.

The other thing that comes up in the comments threads is that a large amount of this stems from that dichotomy of `FileSource` and `FileSourceFS` and their given responsibilities.  So I thought I'd share my thoughts on that too.

It wasn't until I got a bit through this that I came to the (preliminary and uniformed) opinion that the main difference between those interfaces (from the user of the API's position) is the ability to modify in `FileSourceFS` and only read in `FileSource`.

Making these share an interface and then at run-time determining if features are available also feels to me non-ideal, I think having those separation of concerns and features available at compile/code writing time and front and center are valid and useful.

So what would I propose?  I'd suggest that we should consider two different interfaces.  `FileSource` (as is, although perhaps with the `OpenReadStream` method also supported and then a new interface `FileStore` that adds in those other feature to do with modification.  A source can only be read.  A store is read/write/modify.  However, as we've only got one implementation of `FileStore`, I think it would be fine to simply just have a `FileStoreFS` class (renamed from the `FileSourceFS` one, and not separate/write out the interface at this stage.

That's just my 2c though :-)

Anyway, I'll get back to cleaning up the implementation and see where I get to.

JonBooth78:
Oh, final thought;

I've also spent a chunk of time working in Java, which has led me to pick up some good and bad habits - so appreciate the "not everything needs to be a class" sentiment, it's a very good starting point to challenge when thinking about designs :-)

JonBooth78:
OK, so I've now pushed a version here that uses the class style approach here and an alternative change on top of that over here: https://github.com/pioneerspacesim/pioneer/commit/38b21392624f350f86e81cafde3ab8136d3be812  that uses the other suggested approach.

:-)

Max5377:
Don't forget to do `return luaL_error(...)` like @Web-eWorks said [here](https://github.com/pioneerspacesim/pioneer/pull/5638#discussion_r1350034830).
Also I suggest you to learn about [git commands](https://dev.pioneerspacesim.net/contribute/git-and-github) , or just download Github Desktop (like I'm doing) and experiment here to not produce a lot of commits, especially, for example, if you just fixing some bug or made some typo.

JonBooth78:
Ahh I missed the return directly from the `luaL_error` piece, thanks :-)

I produce lots of commits deliberately :-)  Like my mathematics teacher at school always said "show your working".  Keeping track of the journey is I believe helpful - well it's allowed me to reset and dig myself out of a few fine messes in the past.  It also encourages me to work on a single task at a time and not jump all over the place, something I'm somewhat prone to do. Finally, having a laptop explode on me once, I like to keep stuff backed up offsite!

So my current process is to commit working steps early and often.  If that's distracting/disruptive on a WIP PR I'm happy to oblige though with what I push up and I'll keep the rest to myself.

I agree it's desirable to rebase children/squash when done to tidy up at the end as a lot of the transient ones have little value once the destination is reached.  It's something I really appreciate about the git workflow, although I tend to still create fresh branches, cherry pick in the right things in order, without a commit and then rebuild that way as it feels somehow, more natural and less disruptive and allows me to test each iteration that finally gets presented.  I do end up with many branches to clean up though.

I'd expect once we settle on an acceptable endpoint for this, it can go into master as a single squashed commit. There isn't really a lot of separate pieces that matter too much I don't think.  Overall it's quite a small change.

Off topic; Since I lost access to Beyond Compare a few years back, which I think sets a very high bar I've bounced around through tools.  For git repos, I've settled recently on a combination of command line, SourceTree (old habits from when I first started in on git in anger about 12 years back) and KDiff3 right now.  I really appreciate the 3 way view style, especially when dealing with conflicts, it's the most Beyond Compare like I've found, without forking out for a license.  I don't really code much anymore (shows, lol) so haven't justified shelling out for a personal one, yet.

I do have the GitHub Desktop on my list of things to play with but it seems very locked into GitHub and it's not always the environment I work in so I'm not convinced, but do think it might help my PR workflow and learning the GitHub environment and everything they've bolted on top of git.  Feels like I'm learning heaps right now and adding a new tool to the mix would be a distraction.

To be fair the interface to the tools shouldn't matter so long as one understands sufficiently what's actually going on :-)

Max5377:
Well, I can't speak from a developer's point of view. This was just a tip, from what I managed to learn while working on this project (for almost a month now, time flies). Of course, you can use any tools that you find convenient for yourself.

JonBooth78:
OK, I've just pushed up what I hope is a 'final' solution, as a single commit

Max5377:
`LuaFileSystem::l_patched_io_open(lua_State* L)` you forgotten something:
```
if (!fs.is_read_write)
{
	luaL_error(L, "'%s' does not support file operations that modify it, only things in the user folder do", split_uri.full_path.c_str());
	return 0;
}
```
I'll leave the rest to @Web-eWorks

JonBooth78:
Nah, I just missed it. To be fair the pattern used already in this file was to have the return 0 on the file after the lua error call.

Unfortunately, I just spotted another drawback to this version, over the class system originally presented; not only (due to the way we're parsing the URI), do we have to use a `std::string` which may, or may not do memory allocation and is therefore potentially less efficient than it needs to be, if it fails to validate, we're then leaking any memory it may or may not have allocated :-(  I guess one could make the parsing only check for scheme's 13 code units in length (bytes) so that we could ensure the string allocated would always fit into the small-string optimization; but that's a lot yuck and still copying things onto/off the stack.

I've think I've seen this pattern throughout though so this probably isn't the only place.

src/lua/LuaFileSystem.cpp
```diff
@@ -7,15 +7,107 @@
...
+	std::string		 full_path;
+	std::string_view scheme;
+	std::string_view rel_path;
+};
+
+SplitURI ParseURIArg(lua_State* l, int numArg)
+{
+	SplitURI res;
+	size_t path_size;
+	const char* path_str = luaL_checklstring(l, numArg, &path_size);
```
Web-eWorks, line 23: As long as the underlying string value isn't removed from the stack, this pointer will remain valid. You can store `full_path` as a `string_view` (initialized via ptr + size constructor) as long as the methods using SplitURI don't pop the value from the stack.

src/lua/LuaFileSystem.cpp
```diff
@@ -7,15 +7,107 @@
...
+	if (split_uri.scheme.empty())
+	{
+		return luaL_error(L, "'%s' does not have a valid scheme, must be user:// or data://", split_uri.full_path.c_str());
+	}
+
+	const FileSource fs = DetermineFileSource(split_uri.scheme);
+	if (nullptr == fs.source)
+	{
+		return luaL_error(L, "'%s' does not have a valid scheme, must be user:// or data://", split_uri.full_path.c_str());
+	}
```
Web-eWorks, line 80: This block (due to the way DetermineFileSource is implemented) can be condensed to:

```suggestion
	SplitURI split_uri = ParseURIArg(L, 1);
	const FileSource fs = DetermineFileSource(split_uri.scheme);
	if (nullptr == fs.source)
	{
		return luaL_error(L, "Path '%s' does not contain a valid scheme, must be user:// or data://", split_uri.full_path.c_str());
	}
```

This is arguably better, especially if a `data://` scheme is implied (e.g. for implementing custom data assets in lua) - unsure if an implied scheme is needed for the first implementation however.

src/lua/LuaFileSystem.cpp
```diff
@@ -7,15 +7,107 @@
...
+		if (*c != 'r')
+		{
+			if (!fs.is_read_write)
+			{
+				luaL_error(L, "'%s' does not support file operations that modify it, only things in the user folder do", split_uri.full_path.c_str());
+				return 0;
+			}
+			break;
+		}
+	}
```
Web-eWorks, line 94: Because mode `r+` is read-write (and I'd prefer if a mode of `r` is implied if no argument is passed), this code should most likely be:
```suggestion
	std::string_view mode_str = LuaPull<std::string_view>(L, 2, "r");
	if (mode_str.find_first_of("aw+") != std::string_view::npos && !fs.is_read_write) {
		return luaL_error(L, "File '%s' is read-only.", split_uri.full_path.data());
	}
```

src/lua/LuaFileSystem.cpp
```diff
@@ -7,15 +7,107 @@
...
+
+	try
+	{
+		::std::string abs_path = fs.source->Lookup(std::string(split_uri.rel_path)).GetAbsolutePath();
+		lua_pushlstring(L, abs_path.c_str(), abs_path.length());
```
Web-eWorks, line 99: You can use `LuaPush(L, abs_path)` here instead of explicitly calling `lua_pushlstring`.

Web-eWorks (commented):
Finally getting a chance to sit down and review the code again. Overall I'm happy with the design and architecture - I've left some change requests where there are spots the code could be made less verbose / laborious while keeping the same semantics. Some of the changes can be applied to all functions in the module (for example, validating the URI scheme).

Web-eWorks:
@JonBooth78 Replying now to the discussion in the PR proper:

> I think I could have written a better comment above, I don't think the implementation details of LuaFileSystem.cpp are ready, I wanted to check we're aligned on how it will look from the lua side so worry I've poorly made use of your time.

There's no worry here. Those who are willing to comment are happy to spend their time helping to shape both the implementation and the API.

> assuming we even have RTTI available

We do, though it's used extremely sparingly and I'd like to get rid of it eventually (maybe not until C++26 though...).

> However, one of the main values from doing this is that we can validate and throw Lua errors as appropriate within those validation methods, which flies against the comment above suggesting we would only throw Lua errors from the top level of a Lua method and not for anything called internally. I wonder if for readability and keeping compilers happy about dead code paths, we could just throw an exception within helper functions; it's still dead code but then our c++ compiler will understand it and it gives some flexibility to allow such helper functions along with the brevity/clarity concerns.

I mostly pointed out the fact that lua errors don't behave like you'd expect exceptions to as a pointer against future "gotcha" moments where a more important resource might leak due to the pattern. I'm not concerned about a few dozen bytes of string memory being lost on an error path that should be called very infrequently. Where possible however, allocations should ideally be performed *after* methods that might throw Lua errors. I'd prefer not to use manual exceptions internally; one of the options available to us is to just compile the Lua runtime as c++ using exceptions instead of setjmp for error handling.

This is all part of a larger discussion / effort to make the Lua vm "error-tolerant" so that errors are logged and don't crash the game, and not really something that needs to be solved at this scope / at this time.

> So given those pieces above, we're left with the need for an aggregation of variables and some methods that act upon them. I'm currently of the opinion that this concept is one of the key ones behind OO programming and in a language that does support it, one should use it.

I don't disagree with you by any means. Classes are a very useful feature, but sometimes there's a tendency to cram everything you think you might need into a class and wind up spending a lot more time (both programmer and CPU) on extraneous work that's not needed in 99% of usecase. Another interesting (though more academic) way to look at it is that the collection of data and the method set that acts on that data don't necessarily need to be the *same* entity, which is something I find a lot in the data-oriented-design discipline programming.

> I'd suggest that we should consider two different interfaces. FileSource (as is, although perhaps with the OpenReadStream method also supported and then a new interface FileStore that adds in those other feature to do with modification. A source can only be read. A store is read/write/modify. However, as we've only got one implementation of FileStore, I think it would be fine to simply just have a FileStoreFS class (renamed from the FileSourceFS one, and not separate/write out the interface at this stage.

I think I broadly agree with you on the matter, other than to say that IMO FileSource encapsulates both read/write semantics (as it's the 'file' part that is read/write rather than 'source', and 'store' tends to have more database-like connotations) but quibbling about the name aside that's effectively the same hierarchy we have currently.

> So my current process is to commit working steps early and often. If that's distracting/disruptive on a WIP PR I'm happy to oblige though with what I push up and I'll keep the rest to myself.

This is a good process. I find small topical commits are a good working practice; they make rebase / history cleanup much easier, and there's no issue with them on a WIP PR since they can very easily be squashed down into a single "feature commit" where needed.

I'm also a general advocate that commits should include only changes to a specific system or loose 'area' of the codebase and include a summary of the rationale behind the chosen solution in the commit, any test results garnered from the change, and any additional context behind the *choice* of code in the commit. A commit that meets those guidelines, though it might only be 15 lines changed, has enough reason to stand on its own and not be squashed down. (This also makes it easier to blame/revert/bisect problem changes that snuck through the review process).

> I really appreciate the 3 way view style, especially when dealing with conflicts, it's the most Beyond Compare like I've found, without forking out for a license. I don't really code much anymore (shows, lol) so haven't justified shelling out for a personal one, yet.

Visual Studio Code's "merge editor" mode also does a good job of 3-way diff in my experience; I tend to use VSCode with the Git Graph extension for most of my git history munging. I've not found many other tools that do a good job and don't struggle with large repositories. GitKraken might be worth looking into, though most of its features I believe are locked behind a subscription.

---

I think that's everything I've needed to catch up on. The only other thing I see with this PR that might present an issue is that mod data filesystems can be mounted via FileSourceZip which means the contained files don't actually exist on disk as separate files for the Lua IO facilities to work with. That's not a high-priority issue for this PR specifically, but it will eventually necessitate a migration from the Lua built-in file object methods to our own which can function on packed mod files.

src/lua/LuaFileSystem.cpp
JonBooth78: Ahh, for some reason, I thought `std::string_view` didn't have the `find_first_of` method, which took me down the wrong path - thanks :-)

src/lua/LuaFileSystem.cpp
JonBooth78: Nice, I need to get my head around our LuaPush/Pull stuff.

src/lua/LuaFileSystem.cpp
JonBooth78: I also just realized that `find_first_of` is I believe incorrect here, we should be using `find` as we want to match the entire `://` string, not any char in it.

Web-eWorks: Indeed, `string_view::find()` is probably what we want here - good catch!

src/lua/LuaFileSystem.cpp
```diff
@@ -112,9 +180,9 @@ static int l_filesystem_read_dir(lua_State *l)
 		lua_setfield(l, -2, "mtime");
 
 		if (info.IsDir())
-			lua_rawseti(l, dirsTable, lua_rawlen(l, dirsTable) + 1);
+			lua_rawseti(l, dirsTable, (int)lua_rawlen(l, dirsTable) + 1);
 		else
-			lua_rawseti(l, filesTable, lua_rawlen(l, filesTable) + 1);
+			lua_rawseti(l, filesTable, (int)lua_rawlen(l, filesTable) + 1);
```
Web-eWorks, line 185: Instead of querying `lua_rawlen` every loop iteration, this code should just keep a counter variable (e.g. `int i = 1;`) and increment it every time it appends to the lua table.

Web-eWorks (changes requested):

src/lua/LuaFileSystem.cpp
JonBooth78: Ahh yes, I just did the explicit cast there to remove an annoying compiler warning but didn't really look at the code.

Fixed :-)

src/lua/LuaFileSystem.cpp
```diff
@@ -60,49 +134,37 @@ static void push_date_time(lua_State *l, const Time::DateTime &dt)
...
-			luaL_error(l, "'%s' is not a valid path", path.c_str());
-			return 0;
+			return luaL_error(l, "'%s' is not a valid path", split_uri.full_path.data());
 		}
 	}
 
-	FileSystem::FileEnumerator files(*fs, FileSystem::FileEnumerator::IncludeDirs);
-	files.AddSearchRoot(path);
+	FileSystem::FileEnumerator files(*fs.source, FileSystem::FileEnumerator::IncludeDirs);
+	files.AddSearchRoot(rel_path);
```
Web-eWorks, line 159: This can also be expressed as:

```suggestion
auto files = fs.source->Enumerate(rel_path, FileSystem::FileEnumerator::IncludeDirs);
```

The returned `FileEnumerator` object can also be used with a range-for expression either by name or directly, e.g.:

```c++
for (const auto &info : fs.source->Enumerate(rel_path, FileSystem::FileEnumerator::IncludeDirs)) {
    // ...
}
```

This was implemented fairly recently, so you won't see much code using that pattern (yet) but it's one I'd prefer to move to.
(This uses a temporary adapter object with a reference to the temporary enumerator object, which AFAIK according to temporary lifetime rules will not be destroyed until the end of the range-for statement.)

src/lua/LuaFileSystem.cpp
```diff
@@ -146,9 +208,58 @@ static int l_filesystem_join_path(lua_State *l)
...
+	const FileSource fs = DetermineFileSource(split_uri.scheme);
+	if (nullptr == fs.source)
+	{
+		return luaL_error(l, "Path '%s' does not contain a valid scheme, must be user:// or data://", split_uri.full_path.data());
+	}
+
+	if (false == fs.is_read_write)
+	{
+		return luaL_error(l, "'%s' does not support file operations that modify it, only things in the user folder do", split_uri.full_path.data());
+	}
```
Web-eWorks, line 244: We generally don't use the pattern of `constant == variable` in Pioneer - if there's a concern about the value being overwritten, you could prefer e.g. `!fs.source` or `!fs.is_read_write`. The use of this pattern isn't something that needs to be addressed to merge, just pointing out the de-facto style the rest of the codebase follows.

src/lua/LuaFileSystem.cpp
```diff
@@ -7,15 +7,90 @@
...
+		return luaL_error(L, "File '%s' is read-only.", split_uri.full_path.data());
+	}
+
+	try
+	{
+		::std::string abs_path = fs.source->Lookup(std::string(split_uri.rel_path)).GetAbsolutePath();
+		LuaPush(L, abs_path);
+		lua_replace(L, 1);
+
+		const int rv = l_original_io_open(L);
```
Web-eWorks, line 85: I've most likely mentioned this elsewhere, but since I can't remember - this will fail if the file is provided by a `FileSourceZip`, and the failure may not throw an error (will open an incorrect version of the file) if the FileSourceZip is a mod overwriting a game file.

This will have to be addressed before we add lua code which opens any files from the game data directory. Given the purpose of this PR is to provide flexible access to the user directory however, I won't consider it a requirement for merge.

src/lua/LuaFileSystem.h
```diff
@@ -4,13 +4,16 @@
 #ifndef _LUAFILESYSTEM_H
 #define _LUAFILESYSTEM_H
 
+#include <string>
+
+struct lua_State;
+typedef int (*lua_CFunction) (lua_State* L);
+
```
Web-eWorks, line 11: There's not really a need to forward-declare here; this header isn't included anywhere which isn't also including the main Lua headers. By convention we include the "main" `lua/Lua.h` header to get access to Lua typedefs.

```suggestion
#include "lua/Lua.h"

```

Web-eWorks (approved):
Apologies for the extremely long hang-time here for final review - it's been a very mentally-draining few weeks and what little Pioneer time I've had has been spent on optimization efforts.

I've left several "nitpick" comments, some of which are well outside the scope of this PR and the rest are optional changes for readability or code style which you can make or ignore at your discretion. Regardless of whether those changes are implemented, I consider this PR approved for merge.

src/lua/LuaFileSystem.cpp
JonBooth78: Yeah, I'll have a look at that over time, seems like a fair little project and right down my alley.

src/lua/LuaFileSystem.cpp
JonBooth78: Nice, I hadn't really touched this but I've now moved it to the new idiom

src/lua/LuaFileSystem.cpp
JonBooth78: OK, noted, I'll try and keep to the style.

Will take a while to override my default of never using `!` and always using Yoda conditionals I expect :-)

JonBooth78:
Hey @Web-eWorks , I've just rebased and tidied all this up with a squash, so I think it's ready for you (or anyone else for that matter with permissions) to merge it, once the builds above succeed

src/lua/LuaFileSystem.cpp
```diff
@@ -7,15 +7,90 @@
...
+SplitURI ParseURIArg(lua_State* l, int numArg)
+{
+	SplitURI res;
+	res.full_path = LuaPull<std::string_view>(l, numArg);
+	size_t sep_pos = res.full_path.find("://");
+	if (sep_pos != std::string_view::npos)
+	{
+		res.scheme = res.full_path.substr(0, sep_pos);
+		sep_pos += 3;
+		res.rel_path = res.full_path.substr(sep_pos, res.full_path.size() - sep_pos);
```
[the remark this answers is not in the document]
JonBooth78: done

Web-eWorks:
Thanks! Excellent work on this and thanks for bearing with the long review process. :smile:

PR #5639, opened 2023-10-10
translations/ru: do not translate gravity unit (g)

Mc-Pain:
acceleration unit 'g' is not translated in Russian:
https://ru.wikipedia.org/wiki/%D0%A3%D1%81%D0%BA%D0%BE%D1%80%D0%B5%D0%BD%D0%B8%D0%B5_%D1%81%D0%B2%D0%BE%D0%B1%D0%BE%D0%B4%D0%BD%D0%BE%D0%B3%D0%BE_%D0%BF%D0%B0%D0%B4%D0%B5%D0%BD%D0%B8%D1%8F

### DIFF
--- a/data/lang/core/ru.json
+++ b/data/lang/core/ru.json
@@ -1753,7 +1753,7 @@
   },
   "UNIT_EARTH_GRAVITY": {
     "description": "Acceleration unit: one earth gravity (g)",
-    "message": "г"
+    "message": "g"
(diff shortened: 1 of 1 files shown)


nozmajner:
I think I forgot to tell you that we do translations via transifex, no need for a PR. Pulling from transifex would overwrite it even. Sorry.

PR #5642, opened 2023-10-10
Disable scrolling for tab-view and character info

Max5377:
Added "NoScrollWithMouse" flag for tab-view (personal info and comms upper left corner buttons) and character info (char name, occupation and char's face) so now they can't be scrolled with mouse.
**Edit**: disables this behaviour:

https://github.com/pioneerspacesim/pioneer/assets/69468517/3a678c0b-cdfc-4055-b348-c890b8a61372


Web-eWorks:
Nice catch! This is technically caused by the fact the underlying window isn't providing enough space for its contents, but disabling mouse scrolling is a very valid fix.

Web-eWorks (approved):

PR #5643, opened 2023-10-10
Fixes and changes for functions in LuaBody/Player/Ship

Max5377:
**Update**:
Rewrite of `GetAltitudeRelTo`:
1. This function is moved to body class to support this functionality in C++ code;
2. This function is used in the LUA version of `GetAltitudeRelTo`, `GetGroundPosition` and `GetGPS` for consistency in calculating altitudes;
3. This function supports the choice of which height to calculate (sea-level, above-ground).

Other change:
`GetGPS` is moved to `LuaShip.cpp` to support this function for other ships, not only the player.

Fixes for this issues:

1. `GetAltitudeRelTo` using player pos no matter what first argument is passed;
2. `GetGroundPosition` returns nil when not in the rotating frame of ref;
3. `GetGPS` sometimes return different height than `GetAltitudeRelTo`;
4. Access violation when trying to access `frameBody`, `frameRotation` and call `GetGroundPosition` when body doesn't have frame of ref (ex. when ship is hyperspacing).
**Update2**:
`static bool push_body_to_lua(Body *body)`: 
1. `dynamic_cast` is changed to `static_cast`;
2. case `ModelBody` and `CargoBody` had wrong template type (`Body` and `Star`).
P.S.: Unless, there's some good reason for using templates, it can actually be changed to just `LuaObject<Body>::PushToLua(body)`.
(I checked, haven't seen any oddities).

**Update3**:
Removed space at the start of `l_body_get_altitude_rel_to`. A lot of comments didn't show up for functions, because they had a space between function name. Deleted spaces, so comments will show up when hovering function name.

**Update4**:
1. `GetGPS` is moved to `Ship.lua` to support this function for other ships, not only the player. Implementing this function in C++ is considered unnecessary because of slightly different parameter interface and because of this can be simplified a lot;
2. `GetGroundPosition` also supports the choice of which altitude to calculate (sea level or above terrain).

**Old**:
Fixes for this issues:
1. GetAltitudeRelTo using player pos no matter what first argument is passed;
4. GetGroundPosition returns nil when not in the rotating frame of ref;
5. Access violation for all three functions when body doesn't have frame of ref (ex. when ship is hyperspacing).

Need clarification on this:
1. Is this intentional behavior, that GetAltitudeRelTo and GetGroundPosition returns slightly different values for altitude?
2. Is the second change (deleteing `if (!f->IsRotFrame())`) should be changed back? I deleted that, because with this, when, for example you're on the Moon, altitude will already be `nil` at ~6000m.
**Edit**: to clarify second point, altitude will be different when `IsRotFrame` will be false (basically, when space particles (lines) will start to appear or when not in rotating frame).

### DIFF
--- a/src/Body.h
+++ b/src/Body.h
@@ -43,6 +43,13 @@ enum class ObjectType { // <enum name=PhysicsObjectType scope='ObjectType' publi
 	HYPERSPACECLOUD // <enum skip>
 };
 
+enum class AltitudeType { // <enum name=AltitudeType scope='AltitudeType' public>
+	//SEA_LEVEL if distant, ABOVE_TERRAIN otherwise
+	DEFAULT,
+	SEA_LEVEL,
+	ABOVE_TERRAIN
+};
+
 #define OBJDEF(__thisClass, __parentClass, __TYPE)                                  \
 	static constexpr ObjectType StaticType() { return ObjectType::__TYPE; }         \
 	static constexpr ObjectType SuperType() { return __parentClass::StaticType(); } \
@@ -175,6 +182,7 @@ public:
 	vector3d GetInterpPositionRelTo(FrameId relToId) const;
 	vector3d GetInterpPositionRelTo(const Body *relTo) const;
 	matrix3x3d GetInterpOrientRelTo(FrameId relToId) const;
+	double GetAltitudeRelTo(const Body* relTo, AltitudeType altType = AltitudeType::DEFAULT);
 
 	// should set m_interpolatedTransform to the smoothly interpolated value
 	// (interpolated by 0 <= alpha <=1) between the previous and current physics tick
--- a/src/lua/LuaBody.cpp
+++ b/src/lua/LuaBody.cpp
@@ -102,7 +102,6 @@ static int l_body_attr_seed(lua_State *l)
  *
  *   stable
  */
-
 static int l_body_attr_path(lua_State *l)
 {
 	Body *b = LuaObject<Body>::CheckFromLua(1);
@@ -222,7 +221,6 @@ static int l_body_set_component(lua_State *l)
  *
  *   stable
  */
-
 static int l_body_get_velocity_rel_to(lua_State *l)
 {
 	Body *b = LuaObject<Body>::CheckFromLua(1);
@@ -336,6 +334,7 @@ static int l_body_is_more_important_than(lua_State *l)
 	LuaPush<bool>(l, PiGui::first_body_is_more_important_than(body, other));
 	return 1;
 }
+
 /*
  * Method: GetPositionRelTo
  *
@@ -355,7 +354,6 @@ static int l_body_is_more_important_than(lua_State *l)
  *
  *   stable
  */
-
 static int l_body_get_position_rel_to(lua_State *l)
 {
 	Body *b = LuaObject<Body>::CheckFromLua(1);
@@ -370,11 +368,19 @@ static int l_body_get_position_rel_to(lua_State *l)
  *
  * Get the body's altitude relative to another body
  *
- * > body:GetAltitudeRelTo(otherBody)
- *
  * Parameters:
  *
  *   other - the other body
+ *   (optional) terrainRelative - what altitude to calculate (sea-level(false), above-terrain(true))
+ *   if not specified, sea-level for distant bodies, above-ground otherwise
+ *
+ * Examples:
+ *
+ * > -- Get relative altitude to other body
+ * > body:GetAltitudeRelTo(otherBody)
+ *
+ * > -- Get relative above terrain altitude to other body
+ * > body:GetAltitudeRelTo(otherBody, true)
  *
  * Availability:
  *
@@ -384,28 +390,19 @@ static int l_body_get_position_rel_to(lua_State *l)
  *
  *   stable
  */
-
 static int l_body_get_altitude_rel_to(lua_State *l)
 {
+	Body *b = LuaObject<Body>::CheckFromLua(1);
 	const Body *other = LuaObject<Body>::CheckFromLua(2);
-	vector3d pos = Pi::player->GetPositionRelTo(other);
-	double center_dist = pos.Length();
-	if (other && other->IsType(ObjectType::TERRAINBODY)) {
-		const TerrainBody *terrain = static_cast<const TerrainBody *>(other);
-		vector3d surface_pos = pos.Normalized();
-		double radius = terrain->GetSystemBody()->GetRadius();
-		if (center_dist <= 3.0 * terrain->GetMaxFeatureRadius()) {
-			radius = terrain->GetTerrainHeight(surface_pos);
-		}
-		double altitude = center_dist - radius;
-		if (altitude < 0)
-			altitude = 0;
-		LuaPush(l, altitude);
-		return 1;
-	} else {
-		LuaPush(l, center_dist);
-		return 1;
+	AltitudeType altType = AltitudeType::DEFAULT;
+	if (!lua_isnoneornil(l, 3))
+	{
+		bool terrainRelative = lua_toboolean(l, 3);
+		altType = terrainRelative ? AltitudeType::ABOVE_TERRAIN : AltitudeType::SEA_LEVEL;
 	}
+	double altitude = b->GetAltitudeRelTo(other, altType);
+	LuaPush(l, altitude);
+	return 1;
 }
 
 /*
@@ -493,6 +490,13 @@ static int l_body_attr_frame_body(lua_State *l)
 	}
 
 	Frame *f = Frame::GetFrame(b->GetFrame());
+
+	if (!f)
+	{
+		lua_pushnil(l);
+		return 1;
+	}
+
 	LuaObject<Body>::PushToLua(f->GetBody());
 	return 1;
 }
@@ -522,6 +526,12 @@ static int l_body_attr_frame_rotating(lua_State *l)
 	}
 
 	Frame *f = Frame::GetFrame(b->GetFrame());
+	if (!f)
+	{
+		lua_pushnil(l);
+		return 1;
+	}
+
 	lua_pushboolean(l, f->IsRotFrame());
 	return 1;
 }
@@ -601,10 +611,14 @@ static int l_body_distance_to(lua_State *l)
 /*
  * Method: GetGroundPosition
  *
- * Get latitude, longitude and altitude of a dynamic body close to the ground or nil the body is not a dynamic body
- * or is not close to the ground.
+ * Get latitude, longitude and altitude of frame of ref of a dynamic body or nil if the body is not a dynamic body
  *
  * > latitude, longitude, altitude = body:GetGroundPosition()
+ *
+ * Parameters:
+ *
+ * (optional) terrainRelative - what altitude to calculate (sea-level(false), above-terrain(true))
+ * if not specified, sea-level for distant bodies, above-terrain otherwise
  *
  * Returns:
  *
@@ -619,6 +633,11 @@ static int l_body_distance_to(lua_State *l)
  * > lat = math.rad2deg(lat)
  * > long = math.rad2deg(long)
  *
+ * > -- Get ground position of the player with sea-level altitude
+ * > local lat, long, alt = Game.player:GetGroundPosition(false)
+ * > lat = math.rad2deg(lat)
+ * > long = math.rad2deg(long)
+ *
  * Availability:
  *
  *   July 2013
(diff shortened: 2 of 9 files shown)


Max5377:
@Web-eWorks 
1. Looked a little closer at the description of `GetGroundPosition` and it says: `Get latitude, longitude and altitude of a dynamic body close to the ground or nil (if missing here, small typo) the body is not a dynamic body or is not close to the ground`, but should this really be limited to only close bodies?
(Which creates some issues that I mentioned in the initial comment and to add, for example, this function was used in `TradeShips` module and this was one cause for ships sometimes making illegal jumps. This is also used in `Ships.lua` `IsHyperjumpAllowed` and in `SearchRescue.lua` `randomLatLong`);

2. Although I deleted info about HUD displaying different altitudes at the center(`reticule.lua`) and on the bottom-right(`planetary-info.lua`) when player is not in rotation frame in the initial comment, but I use this info to describe issue and possible solution.
On HUD in the center `GetAltitudeRelTo` is used to calculate altitude to surface of the frame of ref/navTarget, but on the bottom-right `GetGPS` is used.
`GetGPS` is using slightly different way to calculate altitude which is similar to `GetGroundPosition`, but when player is not in rotating frame, resulting altitudes will be different for `GetGPS`/`GetGroundPosition` than `GetAltitudeRelTo`.
My suggestion is to add `GetAltitudeRelTo` to `Body` class (or some static function somewhere else), add calculation from `LuaBody.cpp` `l_body_get_altitude_rel_to` to this method and call this method in
`GetGPS`, `GetGroundPosition` and `l_body_get_altitude_rel_to` when altitude needs to be calculated. This way there will be consistency in getting altitude between all bodies;

3. Returned latitude and longitude is also different for `GetGroundPosition` and `GetGPS`.

These are my thoughts at the moment. I will add comments if I find something more.

Web-eWorks:
> but should this really be limited to only close bodies?

The problem here is the split between "radar altitude" and "sea level altitude" relative to the body. Altitude above terrain features (radar altitude) is only useful for bodies actually near or on trajectory to intersect the terrain, whereas bodies in high orbit shouldn't care about terrain height but use sea-level altitude instead. Not only does this better mimic what other games (e.g. KSP) do with regards to altitude handling, it's *significantly* better for performance because we don't have to thrash the instruction cache (`I$`) evaluating the terrain fractal unless terrain is actually relevant to the body making the query.

> My suggestion is to add GetAltitudeRelTo to Body class (or some static function somewhere else), add calculation from LuaBody.cpp l_body_get_altitude_rel_to to this method and call this method in
GetGPS, GetGroundPosition and l_body_get_altitude_rel_to when altitude needs to be calculated. This way there will be consistency in getting altitude between all bodies;

This is fine. The new method should ideally take a parameter indicating whether to return the altitude relative to terrain or sea-level (sbody radius) - I'll leave that at your discretion how best to ensure existing behavior is preserved with the new API.

> Returned latitude and longitude is also different for GetGroundPosition and GetGPS

This should be fixed, but be wary of existing code that implicitly depends on the old/incorrect values.

Max5377:
Looks like `GetAltitudeRelTo` is already using sBody radius:
```
if (other && other->IsType(ObjectType::TERRAINBODY)) {
		const TerrainBody *terrain = static_cast<const TerrainBody *>(other);
		vector3d surface_pos = pos.Normalized();
		double radius = terrain->GetSystemBody()->GetRadius();
		if (center_dist <= 3.0 * terrain->GetMaxFeatureRadius()) {
			radius = terrain->GetTerrainHeight(surface_pos);
		}
		double altitude = center_dist - radius;
		if (altitude < 0)
			altitude = 0;
		LuaPush(l, altitude);
		return 1;
	}
```

Web-eWorks:
If GetGPS can be replaced by GetGroundPosition with no loss in functionality, then by all means please drop the GetGPS function entirely!

Max5377:
@Web-eWorks I listed all changes in the initial comment. Let me know if you find something unusual.

Max5377:
@Web-eWorks I found the cause for assertion error that I mentioned [here](https://github.com/pioneerspacesim/pioneer/pull/5647#issuecomment-1763326176). `TradeShips` code has nothing to do with it.

When ship is hyperspacing, it's not in space (in `m_bodies` array). 
`void LuaSerializer::SaveComponents(Json &jsonObj, Space *space)`
`void LuaSerializer::LoadComponents(const Json &jsonObj, Space *space)`
This two functions only `save`/`load` `lua` components for bodies that are currently in space.
And since ship is not in space, it's components are not `saved`/`loaded`, making it corrupted.
For now, I just swap `HyperspaceCloud` with it's `Ship` in above mentioned functions and it fixes this behaviour.

But maybe there's a more clean way to do this?

Also, `Ship` stores reference to hyperspace cloud, but it loses it after loading the save. 
But, should `Ship` really store reference to `HyperspaceCloud`(`m_hyperspaceCloud`) since `HyperspaceCloud` itself already stores reference to `Ship`? And I haven't seen it used that much, only when ship is about to go to hyperspace, when it gets created.

src/lua/LuaBody.cpp
```diff
@@ -384,28 +392,26 @@ static int l_body_get_position_rel_to(lua_State *l)
...
+	if (typeString.compare("default"))
+	{
+		if (auto str = altitudeTypesStringMap.find(typeString); str != altitudeTypesStringMap.end())
+		{
+			altType = str->second;
+		}
+		else
+		{
+			return luaL_error(l, "invalid altitude type: %s", typeString.c_str());
 		}
```
Web-eWorks, line 410: Please either use the `EnumStrings` facility to convert a string to an enum value decorated for use with the `scan_enums` script, or (better) use a tri-state boolean (nil, false, true) check here - if `nil`, use AltitudeType::DEFAULT, if `false` use `SEA_LEVEL`, if `true` use `ABOVE_TERRAIN`. There's not really a need to define yet another custom enumeration type for the tiny API call surface at play here, especially when the tri-state semantics map perfectly to the effect of the enumeration.

src/lua/LuaPlayer.cpp
```diff
@@ -361,6 +351,12 @@ static int l_get_remaining_delta_v(lua_State *l)
 	return 1;
 }
 
+static std::map<std::string, Thruster> thrusters_map = {
+	{ "forward", THRUSTER_FORWARD },
+	{ "reverse", THRUSTER_REVERSE },
+	{ "up", THRUSTER_UP },
+};
+
```
Web-eWorks, line 359: If this map isn't used in a new way that requires moving it within the source file, please just leave it in-place - it makes it easier to follow the git history / blame log.

src/lua/LuaBody.cpp
```diff
@@ -375,6 +381,8 @@ static int l_body_get_position_rel_to(lua_State *l)
  * Parameters:
  *
  *   other - the other body
+ *   (optional) altType - what altitude to calculate (sea-level, above-ground)
+ *   if not specified, sea-level for distant bodies, above-ground otherwise
```
Web-eWorks, line 385: If changing the parameters to a LuaBody API function, please also update the function prototype in `data/meta/CoreObject/Body.meta.lua`.

src/lua/LuaShip.cpp
```diff
@@ -1721,6 +1771,7 @@ void LuaObject<Ship>::RegisterClass()
 
 		{ "GetHyperspaceCountdown", l_ship_get_hyperspace_countdown },
 		{ "IsHyperspaceActive", l_ship_is_hyperspace_active },
+		{ "GetGPS", l_get_gps },
```
Web-eWorks, line 1774: If this function isn't going to be used heavily in a follow-up PR in a way that requires it being implemented in C++, I'd highly prefer implementing it in Lua instead. It's used in only two callsites and is effectively a `ship:GetGroundPosition()` and a `ship:GetVelocityRelTo(ship.frameBody):dot(ship:GetPositionRelTo(ship.frameBody):normalized())` paired together. There's no need to duplicate the same code in the name of a slightly different return parameter interface, especially given this limits the use of GetGPS to a ship object when it can be implemented in terms of only LuaBody API calls.

If you're concerned about the performance of the Get*RelTo functions, you can also expose `body.position` and `body.velocity` as new attributes which would simplify the above to `ship.velocity:dot(ship.position:normalized())`.

src/lua/LuaBody.cpp
```diff
@@ -636,22 +654,20 @@ static int l_body_get_ground_position(lua_State *l)
...
 	Body *astro = f->GetBody();
-	if (astro->IsType(ObjectType::TERRAINBODY)) {
-		double radius = static_cast<TerrainBody *>(astro)->GetTerrainHeight(pos.Normalized());
-		double altitude = pos.Length() - radius;
-		lua_pushnumber(l, altitude);
-	} else {
-		lua_pushnil(l);
-	}
+	double altitude = b->GetAltitudeRelTo(astro);
+	lua_pushnumber(l, altitude);
```
Web-eWorks, line 670: This function should take a parameter to pass to `GetAltitudeRelTo` similar to `body:GetAltitudeRelTo()`.

Web-eWorks (changes requested):
I have a few concerns with the API patterns used in this PR, which are explored in more depth in the individual request comments. Otherwise I'm quite happy with the improvements this brings (especially the `dynamic_cast` removal and segfault fixes!)

Thank you for your efforts in fixing these bugs and improving the code quality of Pioneer!

src/Body.h
```diff
@@ -43,6 +43,13 @@ enum class ObjectType { // <enum name=PhysicsObjectType scope='ObjectType' publi
 	HYPERSPACECLOUD // <enum skip>
 };
 
+enum class AltitudeType {
+	//SEA_LEVEL if distant, ABOVE_TERRAIN otherwise
+	DEFAULT,
+	SEA_LEVEL,
+	ABOVE_TERRAIN
+};
```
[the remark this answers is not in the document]
Max5377: Thanks for this script. It was `--pattern="*.h"` though, haven't worked with `--pattern '*.h'`.

src/lua/LuaBody.cpp
Max5377: Used boolean approach.

src/lua/LuaBody.cpp
Max5377: Done. Also, argument name is changed from `altType` to `terrainRelative` here, like suggested in the last comment.

src/lua/LuaShip.cpp
Max5377: I'm not concerned with performance, so I used `ship:GetVelocityRelTo(ship.frameBody):dot(ship:GetPositionRelTo(ship.frameBody):normalized())`

src/Body.h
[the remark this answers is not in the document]
Web-eWorks: Thanks - I use the script about twice a year and promptly forget the invocation until the next time I need it again.

src/lua/LuaBody.cpp
Web-eWorks: As a note for the future - since the enum you previously used here isn't exposed to Lua anymore, it doesn't need to be marked up for use with enum_table/EnumStrings. Having that enum available even though nothing uses it currently is fine, but just a bit messy.

src/lua/LuaBody.cpp
```diff
@@ -630,28 +655,39 @@ static int l_body_distance_to(lua_State *l)
...
+	bool terrainRelative;
+	if (!lua_isnoneornil(l, 2))
+	{
+		if (!pi_lua_strict_pull(l, 2, terrainRelative))
+		{
+			return luaL_error(l,
+				"invalid second argument passed. Should be \"true\" (above-terrain) or \"false\" (sea-level)"
+			);
+		}
+		altType = terrainRelative ? AltitudeType::ABOVE_TERRAIN : AltitudeType::SEA_LEVEL;
```
Web-eWorks, line 668: Note: it's not required that the function strictly validate it's passed a boolean here. A simple `lua_toboolean` maintains the same API contract and is less expensive. This takes advantage of the language's "truthiness" feature and enables more flexible code without loosing correctness.

```suggestion
		altType = lua_toboolean(l, 2) ? AltitudeType::ABOVE_TERRAIN : AltitudeType::SEA_LEVEL;
```

Web-eWorks (approved):
Thanks for addressing the feedback! I consider this PR in a good state to merge, but I've left one new suggestion where the new code is slightly overzealous in my eyes - I don't consider addressing that a requirement to merge however.

src/lua/LuaBody.cpp
Max5377: Changed to `lua_toboolean` instead of `pi_lua_strict_pull`. Performance is a concern here, espesially if this function is intended to be used in some UI later. And, really,  the only thing that prompted me to do first approach with `pi_lua_strict_pull` is to just notify user that he passed the wrong value.
~~I guess this will teach them the basics of implicit type conversion.~~

Web-eWorks: Implicit truthiness is considered to be a very basic concept one learns in the very early stages of writing Lua - I don't think there's a serious concern here about unexpected behavior :smile:

PR #5645, opened 2023-10-10
Fix main menu music not starting to play after exiting back to main menu

Max5377:
Fix for when menu music not started to play after exiting back to main menu.

**Update**: moved `MusicPlayer.playRandomSongFromCategory("menu", true)` from `if else` check.


impaktor:
@WKFO thoughts?

WKFO:
Looks good to me.

Web-eWorks (approved):

PR #5646, opened 2023-10-10
Pause menu sets last time acceleration after closing

Max5377:
The pause menu now sets the last time acceleration that was before it was opened. Its behavior will be the same as when setting up time acceleration without pressing `CTRL`.
Changed behaviour of `Close` button in `settings-window.lua`. Now it will not set time acceleration and not enable input bindings.
**Edit**:

https://github.com/pioneerspacesim/pioneer/assets/69468517/c89d445d-5624-4e12-8cb7-5e4e5421308b

**Update**: deleted bit about forcing time acceleration through `Ctrl+Click` `Close` button.
**Update2**: deleted `settingsMenuWasOpened`. It was not needed.

### DIFF
--- a/data/pigui/views/game.lua
+++ b/data/pigui/views/game.lua
@@ -23,6 +23,8 @@ local colors = ui.theme.colors
 local reticuleCircleRadius = math.min(ui.screenWidth, ui.screenHeight) / 8
 local reticuleCircleThickness = 2.0
 
+local lastTimeAcceleration
+
 -- for modules
 ui.reticuleCircleRadius = reticuleCircleRadius
 ui.reticuleCircleThickness = reticuleCircleThickness
@@ -265,6 +267,17 @@ Event.Register("onGameStart", function()
 	gameView.rightSidebar:Reset()
 end)
 
+Event.Register("onPauseMenuOpen", function()
+	lastTimeAcceleration = Game.GetTimeAcceleration() ~= Game.GetRequestedTimeAcceleration() and Game.GetRequestedTimeAcceleration() or Game.GetTimeAcceleration()
+	Game.SetTimeAcceleration("paused")
+	Input.EnableBindings(false)
+end)
+
+Event.Register("onPauseMenuClosed", function()
+	Game.SetTimeAcceleration((lastTimeAcceleration == "paused") and "1x" or lastTimeAcceleration)
+	Input.EnableBindings()
+end)
+
 ui.registerHandler('game', function(delta_t)
 		-- delta_t is ignored for now
 		gameView.player = Game.player
@@ -285,15 +298,12 @@ ui.registerHandler('game', function(delta_t)
 		-- TODO: dispatch escape key to views and let them handle it
 		if currentView == "world" and ui.escapeKeyReleased(true) then
 			if not ui.optionsWindow.isOpen then
-				Game.SetTimeAcceleration("paused")
 				ui.optionsWindow:open()
-				Input.EnableBindings(false)
+				Event.Queue("onPauseMenuOpen")
 			else
 				ui.optionsWindow:close()
-				if not ui.optionsWindow.isOpen then
-					Game.SetTimeAcceleration("1x")
-					Input.EnableBindings()
-				end
+				Game.SetTimeAcceleration("1x")
+				Event.Queue("onPauseMenuClosed")
 			end
 		end
 
--- a/data/pigui/modules/settings-window.lua
+++ b/data/pigui/modules/settings-window.lua
@@ -2,6 +2,7 @@
 -- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
 local Engine = require 'Engine'
+local Event = require 'Event'
 local Input = require 'Input'
 local Game = require 'Game'
 local Lang = require 'Lang'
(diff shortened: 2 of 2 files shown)


Web-eWorks:
I don't think restoring forced time acceleration is in scope for the pause menu. On resuming from pause, the game should attempt to restore the previous time acceleration according to the game's limits, and ignore the "forced" time acceleration set by the player.

Consider for example a player who hits the ESC key out of desperation because they are about to crash into a planet during forced time acceleration; they would not be able to resume a "safe" time acceleration out of the pause menu due to the need to close the menu and then click on a time acceleration button.

While your code currently addresses this correctly, the ability to Ctrl+Click the "close" button to restore time acceleration is very non-obvious to the player, and I think is better handled by allowing the player to Ctrl+Click the time acceleration buttons after resuming gameplay to select their desired time acceleration override.

Max5377:
@Web-eWorks The original plan was to `Ctrl+Esc` to force time acceleration but right now if you do that, it will just open `Start Menu` on Windows and not close pause menu. I will delete bit about forcing time acceleration through Ctrl+Click `Close` button.

impaktor:
Not directly related to this PR, but I had a thought when reading the discussion:
what if when you hold down Ctrl-key, the time accel button icons shift to red or something, indicating a new mode / layer?
(and similarly for other buttons that change their function when combined with modifier key).

data/pigui/views/game.lua
```diff
@@ -285,19 +287,21 @@ ui.registerHandler('game', function(delta_t)
...
-				end
 			end
 		end
 
 		callModules('modal')
+		
+		if not ui.optionsWindow.isOpen then
+			Game.SetTimeAcceleration((lastTimeAcceleration == "paused") and "1x" or lastTimeAcceleration)
+			Input.EnableBindings()
+		end
```
Web-eWorks, line 304: This block will be called every frame that the options window is closed. The code should either perform a boundary check (was-open to not-open) or the settings window should queue an `onPauseMenuOpen`/`onPauseMenuClosed` event of some sort and this logic be moved inside that event handler.

Web-eWorks (changes requested):

data/pigui/views/game.lua
Max5377: Oof. Now I remember why I added `settingsMenuWasOpened` variable. But I like you solution with event more. I'll see what I can do.

Max5377: On the fence here. While game is paused events are not proceeded. So I can do `Game.SetTimeAcceleration("1x")` before or after `Event.Queue("onPauseMenuClosed")` and this function will be completed but you will see game setting speed to `1x` for brief second. Or I can do `Event._Emit()` but it says in `Event.lua` that it's an internal method that is being called from C++ so I suppose it's not supposed to be called from lua. What do you think is the best approach?

Web-eWorks: Realistically, we need two event queues - one for "world" events that only process while the simulation is being ticked, and one for "game" events like playing music and reacting to the UI. I'd recommend the first solution you proposed; 1x time acceleration for a few frames is not a big deal (and I'd like to smoothly accelerate to the newly-requested time accel value anyways), and it can take advantage of future changes to the Event architecture.

Max5377: Force-pushed with my first approach with `Game.SetTimeAcceleration("1x") `

Web-eWorks: Clicking the "close" button in the Pause menu doesn't resume time acceleration and leaves the game paused. Pressing ESC to close the pause menu does. Otherwise, looks good.

Max5377: Had to include `Event` in `settings-window.lua` and added this bit to `Close` button in the same file:
```
if Game.player then
	Game.SetTimeAcceleration("1x")
	 Event.Queue("onPauseMenuClosed")
end
```
Let this be a temporary solution before two event queues will be implemented in the near future.

Max5377:
Found some oversight. If the game was already paused, `onPauseMenuOpen` will not be executed. Instead, it will be executed when pause menu gets closed and acceleration will be set to `1x`. Then `onPauseMenuClosed` will be queued and also be executed.
```
[11:50:57.116714] Info:     onPauseMenuOpen
[11:50:57.116825] Info:     onPauseMenuClosed
```
Still works fine, but `onPauseMenuOpen` has `Input.EnableBindings(false)`. Comment about this function:
```
// Call EnableBindings() to temporarily disable handling input bindings while
// you're recording a new input binding or are in a modal window.
```
Also, when you press `End game` `onPauseMenuClosed` will not be executed and `Input.EnableBindings()` will not be called.
I still can assign keys and I can't control ship in the pause menu, but can this cause problems in the future?

Web-eWorks:
`EnableBindings()` should probably be called at the beginning of the game and menu lifecycle in `Pi.cpp`. I'd prefer we have a consistent input state entering the menu. You don't need to move back to the old way of handling things if you do the above; I'll look at adding the separate event queue for "game"/"application" events as it will handle both this case and some other workarounds in the codebase.

impaktor:
I've only skimmed the above post-merge comment, but if there's (potential) bug / "fix-me", that needs to be documented, I'd advice to either open an issue, or push a fix-me comment to the code. 

(This issue tracker is filled with "roger that, I'll fix it shortly", then a decade passes. Which is not a criticism, I just prefer to have TO-DOs documented somewhere so they don't get lost, especially a comment in a merged PR).

Web-eWorks:
No worries, sometimes subtle issues like this sneak past the first round of PR review and testing. Feel free to open another PR with the fixes.

Issue #5539, opened 2023-02-12
Many clouds & station full + weapon on sale not installed


DynV:
## Bug 1
Many clouds & station full.

### Observed behaviour
A station both had many hyperspace departure clouds around it and refuses docking permissions as there's seemingly no free docking bay even though the interior seem far from full.
![screenshot-20230211-195818](https://user-images.githubusercontent.com/2823807/218288106-f103e376-bd5b-480e-9e5e-28aa99daed18.png)
![screenshot-20230211-195936](https://user-images.githubusercontent.com/2823807/218288121-6948b976-8e92-46c7-88d9-ca819f47f9ac.png)
![screenshot-20230211-200259](https://user-images.githubusercontent.com/2823807/218288165-df5f1a44-a115-47b5-957d-016832e135ea.png)

### Expected behaviour
The station either be
- full and look full, or
- look sparsely filled and not be full

and for both cases not have this many hyperspace departure clouds.

### Steps to reproduce
1. Load gamesave
2. optionally change time acceleration
3. drop to 1X time when receiving the permission refusal in comms
4. observe.

__My pioneer version (and OS):__
20230203-win

The attached file was simply renamed so it would go through the upload extension filter, there was no modification to the file content of my save-file.
[Pioneer-Yoan_Faure-20230203-230212-Jobs_Pad_full_and_many_clouds.zip](https://github.com/pioneerspacesim/pioneer/files/10715069/Pioneer-Yoan_Faure-20230203-230212-Jobs_Pad_full_and_many_clouds.zip)

**Update 1**
Not as bad but striking
![screenshot-20230219-095608](https://user-images.githubusercontent.com/2823807/219957462-333f4efc-9fed-44ce-8967-d6f8bcbeee53.png)
and _right above_ the target there was 4.

## Bug 2
Weapon on sale not installed.

### Observed behaviour
Continuing from the gamesave which a copy have been uploaded in this post, I had a fight and realized the weapon discharge has a color different than what I had with on my previous small plasma accelerator on a Skipjack on the previous release (20220203-win). I then checked on my equipment and the weapon was the same as before I bought the small plasma (4MW pulse), I then checked my flight log tab station and I did have a money drop close the the price of a small plasma on sale (from London to LA).

### Expected behaviour
Either
- the weapon get installed when the money for it is reduced or
- no weapon is installed and the money stay the same.

### Steps to reproduce
1. Load gamesave
2. notice weapon
[...]

PR #5647, opened 2023-10-13
TradeShips hyperjump code change

Max5377:
This PR is an attempt to fix cluttering hyperclouds around stations and rare illegal jumps almost at the planet's surface. Now `TradeShips` behaviour when undocking looks like this.
For orbital stations:
1. Undock;
2. Fly to limits of current orbital station;
3. `OnAICompleted` event fires, tell ship to fly to current system's star;
4. Assign task to check distance between current orbital station and ship, if distance is large enough, make the jump.

For ground stations;
1. Undock;
2. Tell ship to fly to current system's star;
3. Assign task to check altitude between current orbital station's parent and ship, if altitude is high enough, make the jump.

The distance and altitude will be selected from a random interval of 20,000 - 240,000 meters.
Demonstration:
1. Undocking and jumping from Cydonia

https://github.com/pioneerspacesim/pioneer/assets/69468517/bd73a4d7-0f1f-4607-bc64-b5e843707223

2. Undocking and jumping from Gates Starport
   

https://github.com/pioneerspacesim/pioneer/assets/69468517/eabefebd-fcd8-4e99-8a70-1ec8fba50b99

**Update**: rebase.
**Update2**: `addEquip` is removed from `setPlayerAsTraderDocked` and `setPlayerAsTraderInbound`.
Added check if player is currently docked (`setPlayerAsTraderDocked`) or there's nearest station (`setPlayerAsTraderInbound`).
If not, player will not become trader.
**Update3**: rebase.

### DIFF
--- a/data/modules/TradeShips/Debug.lua
+++ b/data/modules/TradeShips/Debug.lua
@@ -55,7 +55,7 @@ local statuses = {
 }
 
 debugView.registerTab('debug-trade-ships', function()
-	if not Core.ships and not Core.params then return end
+	if not Core.ships and not Core.params or not Game.system then return end
 	if not ui.beginTabItem("Tradeships") then return end
 
 	local function property(key, value)
--- a/data/modules/TradeShips/Events.lua
+++ b/data/modules/TradeShips/Events.lua
@@ -58,7 +58,6 @@ local onEnterSystem = function (ship)
 	elseif Core.ships[ship] ~= nil then
 		local trader = Core.ships[ship]
 		Core.log:add(ship, 'Entered '..Game.system.name..' from '..trader.from_path:GetStarSystem().name)
-
 		if trader.route then
 			ship:AIDockWith(trader.route.to)
 			Core.ships[ship]['starport'] = trader.route.to
@@ -99,23 +98,6 @@ local onLeaveSystem = function (ship)
 	end
 end
 Event.Register("onLeaveSystem", onLeaveSystem)
-
-local onFrameChanged = function (ship)
-	if not ship:isa("Ship") or Core.ships[ship] == nil then return end
-	local trader = Core.ships[ship]
-	Core.log:add(ship, "Entered frame " .. (ship.frameBody and ship.frameBody:GetLabel() or "unknown"))
-
-	if trader.status == 'outbound' then
-		-- the cloud inherits the ship velocity and vector
-		ship:CancelAI()
-		if Trader.getSystemAndJump(ship) ~= 'OK' then
-			ship:AIDockWith(trader.starport)
-			trader['status'] = 'inbound'
-			trader.ts_error = 'cnt_jump_frame'
-		end
-	end
-end
-Event.Register("onFrameChanged", onFrameChanged)
 
 local onShipDocked = function (ship, starport)
 	if Core.ships[ship] == nil then return end
@@ -159,11 +141,10 @@ Event.Register("onShipDocked", onShipDocked)
 
 local onShipUndocked = function (ship, starport)
 	if Core.ships[ship] == nil then return end
-
-	-- fly to the limit of the starport frame
-	ship:AIFlyTo(starport)
-
-	Core.ships[ship]['status'] = 'outbound'
+	local trader = Core.ships[ship]
+	ship:AIEnterLowOrbit(trader.starport:GetSystemBody().system:GetStars()[1].body)
+	Trader.assignTask(ship, Game.time + 10, 'hyperjumpAtDistance')
+	trader['status'] = 'outbound'
 end
 Event.Register("onShipUndocked", onShipUndocked)
 
@@ -171,15 +152,9 @@ local onAICompleted = function (ship, ai_error)
 	if Core.ships[ship] == nil then return end
 	local trader = Core.ships[ship]
 	if ai_error ~= 'NONE' then
-		Core.log:add(ship, 'AICompleted: Error: '..ai_error..' Status: '..trader.status) end
-
-	if trader.status == 'outbound' then
-		if Trader.getSystemAndJump(ship) ~= 'OK' then
-			ship:AIDockWith(trader.starport)
-			trader['status'] = 'inbound'
-			trader.ts_error = 'cnt_jump_aicomp'
-		end
-	elseif trader.status == 'orbit' then
+		Core.log:add(ship, 'AICompleted: Error: '..ai_error..' Status: '..trader.status)
+	end
+	if trader.status == 'orbit' then
 		if ai_error == 'NONE' then
 			trader.ts_error = "wait_6h"
 			Trader.assignTask(ship, Game.time + 21600, 'doRedock')
@@ -258,7 +233,7 @@ local onShipHit = function (ship, attacker)
 		elseif trader.starport and Engine.rand:Number(1) < trader.chance then
 			local distance = ship:DistanceTo(trader.starport)
 			if distance > Core.AU * (2 - trader.chance) then
-				if Trader.getSystemAndJump(ship) then
+				if Trader.getSystemAndJump(ship) == 'OK' then
 					return
 				else
 					trader['no_jump'] = true
--- a/data/modules/TradeShips/Flow.lua
+++ b/data/modules/TradeShips/Flow.lua
@@ -494,6 +494,47 @@ Flow.spawnInitialShips = function()
 	return ships_in_space
 end
 
+Flow.setPlayerAsTraderDocked = function()
+	local ship = Game.player
+	--if player is not a trader
+	if Core.ships[ship] then
+		print("Flow.setPlayerAsTraderDocked: player is already a trader")
+		return
+	end
+	--if player is currently docked
+	if ship.flightState ~= 'DOCKED' then
+		print("Flow.setPlayerAsTraderDocked: can't set player as docked trader when player is not currently docked")
+		return
+	end
+	local dockedStation = ship:GetDockedWith()
+	Core.ships[ship] = { ts_error = "OK", status = 'docked', starport = dockedStation, ship_name = Game.player.shipId }
+	Trader.assignTask(ship, Game.time + utils.deviation(Core.params.port_params[Core.ships[ship].starport].time * 3600, 0.8), 'doUndock')
+end
+
+Flow.setPlayerAsTraderInbound = function()
+	local ship = Game.player
+	--if player is not a trader
+	if Core.ships[ship] then
+		print("Flow.setPlayerAsTraderInbound: player is already a trader")
+		return
+	end
+	-- Space.PutShipOnRoute will teleport player to star's surface when player is docked. We don't want that
+	if ship.flightState ~= 'FLYING' then
+		print("Flow.setPlayerAsTraderInbound: can't set player as inbound trader when player is not currently flying")
+		return
+	end
+	-- if there's any station in the system
+	local nearestStation = ship:FindNearestTo("SPACESTATION")
+	if not nearestStation then
+		print("Flow.setPlayerAsTraderInbound: no nearby station is found to set player as inbound trader")
+		return
+	end
+	Core.ships[ship] = { ts_error = "OK", status = 'inbound', starport = nearestStation, ship_name = Game.player.shipId }
+	Space.PutShipIntoOrbit(ship, Game.system:GetStars()[1].body)
+	Space.PutShipOnRoute(ship, Core.ships[ship].starport, Engine.rand:Number(0.0, 0.999))-- don't generate at the destination
+	ship:AIDockWith(Core.ships[ship].starport)
+end
+
 Flow.run = function()
 	local ship_name = utils.chooseEqual(Core.params.ship_names)
 	local cloud = utils.chooseEqual(Core.params.hyper_routes[ship_name])
--- a/data/modules/TradeShips/Trader.lua
+++ b/data/modules/TradeShips/Trader.lua
@@ -200,10 +200,6 @@ Trader.getSystemAndJump = function (ship)
 	if Core.ships[ship].starport then
 		local body = Space.GetBody(Core.ships[ship].starport.path:GetSystemBody().parent.index)
 		local port = Core.ships[ship].starport
-		-- boost away from the starport before jumping if it is too close
-		if (ship:DistanceTo(port) < 20000) then
-			ship:AIEnterLowOrbit(body)
-		end
 		return jumpToSystem(ship, getSystem(ship))
 	end
 end
@@ -275,6 +271,19 @@ Trader.removeFuel = function (ship, count)
 	return removed
 end
 
+Trader.checkDistBetweenStarport = function (ship)
+	local trader = Core.ships[ship]
+	if not trader then return nil end
+	local distance
+	if trader.starport.type == "STARPORT_ORBITAL" then
+		distance = ship:DistanceTo(trader.starport)
+	else
+		local stationsParent = trader.starport:GetSystemBody().parent.body
+		distance = ship:GetAltitudeRelTo(stationsParent)
+	end
+	return distance >= trader.hyperjumpDist
+end
+
 -- TRADER DEFERRED TASKS
 --
 -- call the passed function in a specified time, checking whether we are still
@@ -292,6 +301,36 @@ local trader_task = {}
 -- made to serialize pending job execution
 -- the task function prototype should be:
 -- function(ship)
+
+trader_task.hyperjumpAtDistance = function(ship)
+	-- the player may have left the system
+	local trader = Core.ships[ship]
+	if not trader then return end
+	if trader.status == "outbound" and trader.ts_error ~= "HIT" then
+		-- if trader is not under attack and started to leave station
+		if trader.starport ~= nil then
+			-- if trader has not started to hyperjump
+			if trader.hyperjumpDist == nil then
+				trader.hyperjumpDist = Engine.rand:Integer(20000, 240000)
+			end
+			if Trader.checkDistBetweenStarport(ship) then
+				-- if distance is large enough attempt to hyperjump
+				local status = Trader.getSystemAndJump(ship)
+				if status ~= 'OK' then
+					ship:CancelAI()
+					ship:AIDockWith(trader.starport)
+					trader['status'] = 'inbound'
+					trader.ts_error = 'cnt_jump_aicomp'
+				end
+				trader.hyperjumpDist = nil
+			else
+				Trader.assignTask(ship, Game.time + 10, 'hyperjumpAtDistance')
+			end
+		end
+	else
+		trader.hyperjumpDist = nil
+	end
+end
 
 trader_task.doUndock = function(ship)
 	-- the player may have left the system or the ship may have already undocked
--- a/src/lua/LuaSpace.cpp
+++ b/src/lua/LuaSpace.cpp
@@ -3,11 +3,12 @@
 
 #include "LuaSpace.h"
 #include "Body.h"
-#include "LuaBody.h"
 #include "CargoBody.h"
+#include "EnumStrings.h"
 #include "Frame.h"
 #include "Game.h"
 #include "HyperspaceCloud.h"
+#include "LuaBody.h"
 #include "LuaManager.h"
 #include "LuaObject.h"
 #include "LuaUtils.h"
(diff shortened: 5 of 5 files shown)


Max5377:
There's still a possibility for this(ex. for Andersen City in Ahonen's Claim in sector Lalande 21185):
Far view:
AndersenCityCloudsFar

Station view:
AndersenCityCloudsStation

But in my opinion it doesn't look this bad as showed here:
https://github.com/pioneerspacesim/pioneer/issues/5539#issuecomment-1437330530
https://github.com/pioneerspacesim/pioneer/issues/5539#issuecomment-1449327723

Max5377:
@Gliese852 
1. There's a possibility of inbound ship coliding with planet's terrain, exploding in the process. Maybe it's better to tell ship to orbit planet first and only then dock with station?
   (autopilot also has this problem);
2. There is still a possibility of undocking ship breaking the rules and making illegal hyperjump when it's under attack
   (Perhaps can be explained that pilot wants to save his live, not caring about rules at the moment);
3. You were right. Ships are jumping at higher distances when game speed is 10000x. I also discovered this behaviour in the original `TradeShips` code too
   (Can't think of any solution for this, unfortenately, unless it's not big of a deal);
4. The second step of behaviour in orbital stations would be skipped if there wasn't a problem with pathfinding when ship is inside orbital station
   (They basically will ignore walls, start to collide with them and possibly exploding in the process);
5. I included `setPlayerAsTraderDocked` and `setPlayerAsTraderInbound` in `Flow` part of `TradeShips` so you can view `TradeShips` AI from player's eyes (used in the demonstration)
   (Let me know if you want to move it from here or delete it completely);
6. Also, right now in my changed code I'm using `GetAltitudeRelTo`, which has a problem of using player pos in calculating altitude, skipping first argument, but I've done [PR](https://github.com/pioneerspacesim/pioneer/pull/5643) that I mentioned above.
Aside from this I haven't discovered any problems.

Max5377:
Forgot to mention that I also had rare case with old code, where when I was jumping to Bernard's Star from Sol and back and going to the Moon after that, I would see hyperspace cloud below planet's terrain. I have a save to demonstrate what I'm talking about:
[CloudBelowTerrain.zip](https://github.com/pioneerspacesim/pioneer/files/12903463/CloudBelowTerrain.zip)

Haven't seen this with my change, but who knows. I hope this will fix this.

Gliese852:
@Max5377 
> 1. There's a possibility of inbound ship coliding with planet's terrain, exploding in the process. Maybe it's better to tell ship to orbit planet first and only then dock with station?
> (autopilot also has this problem);

Ideally, of course, it would be worth solving this problem inside the autopilot, but if such a workaround will reduce the number of victims, I’m for it.

> 2. There is still a possibility of undocking ship breaking the rules and making illegal hyperjump when it's under attack
> (Perhaps can be explained that pilot wants to save his live, not caring about rules at the moment);

I think this is quite correct behavior.

> 3. You were right. Ships are jumping at higher distances when game speed is 10000x. I also discovered this behaviour in the original `TradeShips` code too
> (Can't think of any solution for this, unfortenately, unless it's not big of a deal);

I'm inclined to think that this is not big of a deal.

> 4. The second step of behaviour in orbital stations would be skipped if there wasn't a problem with pathfinding when ship is inside orbital station
> (They basically will ignore walls, start to collide with them and possibly exploding in the process);

It’s strange, it works the same for me, besides, both commands ultimately come down to calling `AICmdFlyTo`, where it turns out that the body is too close to the center of the orbital and the ship tends [to move away normally](https://github.com/pioneerspacesim/pioneer/blob/00b87acd037d14b46db3ffacc14711c7de854686/src/ShipAICmd.cpp#L1112). Maybe you could provide a save file where the difference in behavior is reproduced?

> 5. I included `setPlayerAsTraderDocked` and `setPlayerAsTraderInbound` in `Flow` part of `TradeShips` so you can view `TradeShips` AI from player's eyes (used in the demonstration)
> (Let me know if you want to move it from here or delete it completely);

There are some problems with these functions, as seen in the second video, Probably, you shouldn’t try to `Trader.addEquip`, because it falls with assertion. Overall this seems like it might be quite useful. I would also like to note that undocking on the autopilot already works for the player, so you can give the autopilot a command while on the pad and see what happens.

Max5377:
@Gliese852 
> Ideally, of course, it would be worth solving this problem inside the autopilot, but if such a workaround will reduce the number of victims, I’m for it.

Didn't do anything really. The death rate for `Sol` system is ~50+/-10 ships per week no matter what approach is used.
Also wanted to add that ships are sometimes colliding with each other too (inbound and outbound) and looks like it usually happenes at orbital stations (according to `TradeShips` log). 
Let `Lalande 21185` be a stress-test for fixing ships AI, because the death rate for this system is insane (179 ships in a week first try) and for some reason in this system there's a lot of deaths from the star itself (according to `TradeShips` log too).

> It’s strange, it works the same for me, besides, both commands ultimately come down to calling AICmdFlyTo, where it turns out that the body is too close to the center of the orbital and the ship tends [to move away normally](https://github.com/pioneerspacesim/pioneer/blob/00b87acd037d14b46db3ffacc14711c7de854686/src/ShipAICmd.cpp#L1112). Maybe you could provide a save file where the difference in behavior is reproduced?

My bad. I'll try to show and explain what happenes. And it's completely different for me.

Giving ship command to fly to body, when ship is more to the center of orbital starport, the more chance it will deviate from the course and fly into the wall instead, especially if it rotates backwards beforehand. 
Here's video to demonstrate this:

https://github.com/pioneerspacesim/pioneer/assets/69468517/244e6b91-e981-424a-a09e-db40e5cda600


If ship is in the center of orbital station, he will usually fly upwards and also try to go through wall, no matter if it's moving or not.
Demonstration:

https://github.com/pioneerspacesim/pioneer/assets/69468517/345d3768-4cc0-481d-a660-8bc5b210053c

This happenes with all commands that tell ship to fly to something (`AIFlyTo`, `AIEnterLow`/`Medium`/`HighOrbit`).
You can also experience it by going into big stations (`Jobs Pad`, `Gates Spaceport`) and setting maximum speed. Soon you will hear collision sounds. I will give you the save, but...

I spotted another issue:
1. Start new game;
2. Wait around for some time;
3. Save;
4. Load;
5. Wait around for some time.

There's chance of game failing with this:
AssertionFailed

After diging around more, this happenes at this line in `Events.lua` in `onShipDocked`:

```
local damage = ShipDef[trader.ship_name].hullMass - ship.hullMassLeft
if damage > 0 then
	ship:SetHullPercent()
	Trader.addEquip(ship) <--
end
```

This behaviour is also reproducable in the original `TradeShips` code.

[Ship's behaviour on Orbital stations.zip](https://github.com/pioneerspacesim/pioneer/files/12908995/Ship.s.behaviour.on.Orbital.stations.zip)

> There are some problems with these functions, as seen in the second video, Probably, you shouldn’t try to Trader.addEquip, because it falls with assertion. Overall this seems like it might be quite useful. I would also like to note that undocking on the autopilot already works for the player, so you can give the autopilot a command while on the pad and see what happens.

Yes. It is enough to order the ship to undock after making the player a trader. No point in exposing `trader_task` table in `Trader.lua` (I didn't include this anyway, but showed it in the demonstration). I will also delete `addEquip`. And I don't know about `Space.PutShipOnRoute`. It frequently puts my ship in what looks like the center of the ground station's model, unless I haven't forgot something (or it's intended).
Demonstration (I set `Engine.rand:Number(0.0, 0.999)` to `Engine.rand:Number(0.999, 0.999)` beforehand):
For ground station:

https://github.com/pioneerspacesim/pioneer/assets/69468517/0589ac9f-b83b-41bf-ba43-37e2bc690940

For orbital station:

https://github.com/pioneerspacesim/pioneer/assets/69468517/13b8de9d-901f-460d-a497-d37c145a962f

Gliese852:
@Max5377 have you read the [description of PutShipOnRoute](https://github.com/pioneerspacesim/pioneer/blob/04c611141b623d26ac155cd3f2708756f40101ab/src/lua/LuaSpace.cpp#L194)? I would like to point out that if the distance to the target is short enough, a `t_ratio` of `0.999` will place you almost at the target coordinates.
The only purpose of this function is to arrange the tradeships in the system when the player enters the star system (or starts the new game) as if they had been flying here for a long time already.

>This happenes with all commands that tell ship to fly to something (AIFlyTo, AIEnterLow/Medium/HighOrbit).

Yes, because, as I already wrote, the first thing a ship does is try to get out of the "object’s radius" _along the normal_, that is, along the ray connecting the center of the station and the center of the ship. So, if the ship is lucky and it is exactly on the axis of the station, it will fly through the gate, if it is shifted off the axis, it will crash somewhere.

Max5377:
@Gliese852 Sorry, math is just not my best subject, really. But, I kinda understand what you mean now. Thanks for clarification.
Do you have any suggestions about `TradeShips` code itself or all good?

Gliese852:
@Max5377 Looks like it's still a draft? I see blots in the code (occasionally spaces/tabs at the end of the line), commented out pieces.

Max5377:
@Gliese852 I deleted parts of commented old code that doesn't needed anymore.
Apart from the issue that I mentioned [here](https://github.com/pioneerspacesim/pioneer/pull/5647#issuecomment-1762137010) about `GetAltitudeRelTo`, it should work without any other issues.

data/modules/TradeShips/Events.lua
```diff
@@ -159,26 +142,31 @@ Event.Register("onShipDocked", onShipDocked)
 
 local onShipUndocked = function (ship, starport)
 	if Core.ships[ship] == nil then return end
+	local trader = Core.ships[ship]
+	
+	if trader.starport.type == "STARPORT_ORBITAL" then
+		-- fly to the limit of the starport frame
+		ship:AIFlyTo(starport)
```
Gliese852, line 149: It’s still unclear, is there any point in this stage?

data/modules/TradeShips/Events.lua
```diff
@@ -159,26 +142,31 @@ Event.Register("onShipDocked", onShipDocked)
...
 end
 Event.Register("onShipUndocked", onShipUndocked)
 
 local onAICompleted = function (ship, ai_error)
 	if Core.ships[ship] == nil then return end
 	local trader = Core.ships[ship]
 	if ai_error ~= 'NONE' then
-		Core.log:add(ship, 'AICompleted: Error: '..ai_error..' Status: '..trader.status) end
-
+		Core.log:add(ship, 'AICompleted: Error: '..ai_error..' Status: '..trader.status)
```
Gliese852, line 164: There is an extra space at the end of the line.

data/modules/TradeShips/Flow.lua
```diff
@@ -494,6 +494,25 @@ Flow.spawnInitialShips = function()
...
+		Trader.assignTask(ship, Game.time + utils.deviation(Core.params.port_params[Core.ships[ship].starport].time * 3600, 0.8), 'doUndock')
+	end
+end
+
+Flow.setPlayerAsTraderInbound = function()
+	local ship = Game.player
+	local nearestStation = ship:FindNearestTo("SPACESTATION")
+	if nearestStation then
+		Core.ships[ship] = { ts_error = "OK", status = 'inbound', starport = nearestStation, ship_name = Game.player.shipId }
+		Space.PutShipOnRoute(ship, Core.ships[ship].starport, Engine.rand:Number(0.0, 0.999))-- don't generate at the destination
```
Gliese852, line 511: This function should currently be used immediately after exiting hyperspace - when the ship is in orbit around a star. I think it's worth either teleporting the player to the star, or writing a comment about the applicability of `setPlayerAsTraderInbound`.

data/modules/TradeShips/Trader.lua
```diff
@@ -275,6 +273,21 @@ Trader.removeFuel = function (ship, count)
...
 end
 
+Trader.checkDistBetweenStarport = function (ship)
+	local trader = Core.ships[ship]
+	if trader then
+		local distance
+		if trader.starport.type == "STARPORT_ORBITAL" then
+			distance = ship:DistanceTo(trader.starport)
+		else
+			local stationsParent = Space.GetBody(Core.ships[ship].starport.path:GetSystemBody().parent.index)
```
Gliese852, line 283: might work `stationsParent = trader.starport.frameBody.body`

data/modules/TradeShips/Trader.lua
```diff
@@ -293,6 +306,39 @@ local trader_task = {}
 -- the task function prototype should be:
 -- function(ship)
 
+trader_task.hyperjumpAtDistance = function(ship)
+	-- the player may have left the system
+	if ship:exists() then
+		local trader = Core.ships[ship]
+		if trader then
```
Gliese852, line 313: This is not super important IMHO, but you can get rid of 2 indentations if you exit using the opposite condition, i.e. `if not trader then return end`

data/modules/TradeShips/Events.lua
```diff
@@ -58,7 +58,7 @@ local onEnterSystem = function (ship)
 	elseif Core.ships[ship] ~= nil then
 		local trader = Core.ships[ship]
 		Core.log:add(ship, 'Entered '..Game.system.name..' from '..trader.from_path:GetStarSystem().name)
-
+
```
Gliese852, line 61: Please take a closer look; in many places you have tabs on an empty line - it would be nice to not allow this.

data/modules/TradeShips/Events.lua
Max5377: There's a high chance of ships crashing into walls if given the command to fly to star directly when they are inside station. Let this be for now. This command will still be given, when `AIFlyTo` get completed (~15km from station).

data/modules/TradeShips/Trader.lua
Max5377: `frameBody` only works for dynamic bodies. Station is not dynamic body, sadly.
`LuaBody.cpp`:
```
/*
 * Attribute: frameBody
 *
 * The non-dynamic body attached to the frame this dynamic body is in.
 *
 * Only valid for dynamic <Bodies>. For non-dynamic bodies <frameBody> will be
 * nil.
 ...
```

data/modules/TradeShips/Events.lua
Max5377: Ok, I will try to do this on clear head, to not miss anything.

data/modules/TradeShips/Flow.lua
Max5377: I will try to teleport player to star (or add comment if I fail), but what's so strange in current behaviour to do this step?

data/modules/TradeShips/Events.lua
Gliese852: It would be good to see confirmation of this statement, as I already wrote, this does not reproduce for me - you have the same chance of leaving the orbital no matter what autopilot command you run.

data/modules/TradeShips/Flow.lua
Gliese852: If I understood the question correctly, for example, you described some strange behavior:
> And I don't know about Space.PutShipOnRoute. It frequently puts my ship in what looks like the center of the ground station's model, unless I haven't forgot something (or it's intended).

Max5377: Ah. I readed description for `Space.PutShipOnRoute` more closely and it says:
```
/*
 * Function: PutShipOnRoute
 *
 * The ship rearranged from its current position to a given body in space, for a
 * given part of the path, as if it were flying in a straight line, consuming
 * fuel and changing speed.
```
Basically, if `t_ratio` == 0.999, ship teleports to station's icon as you can see on `PutShipAtPointOfStation_Ground.mp4` video. This happened for me more frequently, because path between ship and station was short and as it says in the description:
**`The ship rearranged from its current position to a given body in space`**
But I already wrote `Space.PutShipIntoOrbit` on my computer, so I'm gonna include it in the next `force-push`.

data/modules/TradeShips/Events.lua
Max5377: There is not much to add, besides what I have already described in this [comment](https://github.com/pioneerspacesim/pioneer/pull/5647#issuecomment-1763326176). Try to fly to the center of `Jobes Pad` (area between landings pads) and give ship, for example, command to fly to current station. There's almost a guarantee ship will not fly to the exit, but will fly upwards to the wall. Or just give this command, when exit is behind ship and it's moving from it. The higher the speed of the ship, the higher the probability that it will simply fly to the wall, not upwards. 
These two behaviors will occur even if there is no obstructor on the way out. I can reproduce this on current release, even on current master.

data/modules/TradeShips/Events.lua
Max5377: Not anymore.

data/modules/TradeShips/Trader.lua
Max5377: Changed to this and also changed the way to get first system's star to:
`trader.starport:GetSystemBody().system:GetStars()[1].body`

data/modules/TradeShips/Trader.lua
Max5377: Done

data/modules/TradeShips/Events.lua
Max5377: There should be no more such errors.
Thanks to @impaktor for suggesting to install plugin for `editorconfig`-file.
P.S.: I use `Notepad++` for `lua` code. This also has built-in feature to trim trailing spaces on save.

data/modules/TradeShips/Flow.lua
Max5377: Added `PutShipIntoOrbit` in `LuaSpace.cpp`. Using this in the `Flow.setPlayerAsTraderInbound`.

Max5377:
@Gliese852 Also pushed fix for `lua` error when you try to hyperspace with `TradeShips` debug tab opened.

data/modules/TradeShips/Events.lua
Gliese852: Please explain why do you give the `AIFlyTo(starport)` command if
> ... give ship, for example, command to fly to current station. There's almost a guarantee ship will not fly to the exit, but will fly upwards to the wall.

Max5377: Decided to try giving command to fly to low orbit of first system's star immediatly when ship undocks once more and I haven't seen much difference in how ship behaves from previous choice, so I pushed change so now ship will do exactly this. When I first tested this, it seemed to me that with this choice everything just gotten worse, so I decided to quickly revert to `AIFlyTo`.
At this point, let's just consider this issues unrelated to `TradeShips` code. Separate issues can be opened for this if needed.

Web-eWorks (changes requested):
I'll generally leave the approval and merge duties to @Gliese852 as this is code he's most familiar with. I apologize for the long delay in contributing my review of this code - I've only noticed one issue that I'd appreciate being addressed before merge, but I otherwise would like to merge this in the next few weeks if there are no outstanding objections.

Thank you for addressing these bugs!

Gliese852:
@Max5377  I think it’s worth doing a rebase again, in the current state I always have a crash at the beginning of the hyperjump.

Max5377:
@Gliese852 Did rebase to current master. Shouldn't crash now.

data/modules/TradeShips/Debug.lua
```diff
@@ -105,8 +105,10 @@ debugView.registerTab('debug-trade-ships', function()
 			property("Total flow", string.format("%.2f ship/hour", Core.params.total_flow))
 			ui.sameLine()
 			property("Last spawn interval", ui.Format.Duration(Core.last_spawn_interval))
-			ui.sameLine()
-			property("Lawlessness", string.format("%.4f", Game.system.lawlessness))
+			if Game.system then
```
Gliese852, line 108: One more point - maybe even disable the entire tab during hyperspace? iI seems that nothing but the log is relevant there anymore (and the log is not particularly relevant).

Max5377: I think there is no harm if the tab stays during hyperspace. It's short sequence anyway. Maybe log should be cleared when player is in hyperspace, because ships that are there are deleted by that point?

Gliese852: How to say, for example, a flight on Coronatrix 36 light years lasts 45 seconds of real time, enough to go to 'Local routes', click somewhere and crash the game. Do we have a good reason to show this tab in hyperspace?

Max5377: `TradeShips` debug tab now will be disabled in the hyperspace.

Gliese852: Thanks!

src/lua/LuaSpace.cpp
```diff
@@ -298,6 +319,62 @@ static int l_space_put_ship_on_route(lua_State *l)
...
+	Body *b = LuaObject<Body>::CheckFromLua(2);
+	const SystemBody* sbody = b->GetSystemBody();
+	if (!sbody)
+	{
+		return luaL_error(l, "the target body doesn't have a system body");
+	}
+	if (!sbody->IsPlanet() && !sbody->IsMoon() && sbody->GetSuperType() != SystemBodyType::SUPERTYPE_STAR)
+	{
+		return luaL_error(l, "the target body is not a planet, moon, or star");
+	}
```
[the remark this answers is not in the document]
Gliese852: @Max5377 looked at your correction, it seems worth tweaking a little more:
```diff
-	if (!sbody->GetMass() && !b->GetPhysRadius()) {
-		return luaL_error(l, "the target body has zero mass and physical radius");
+	if (!sbody->GetMass() || !b->GetPhysRadius()) {
+		return luaL_error(l, "the target body has zero mass or zero physical radius");
```
Otherwise, for example, a body with zero mass but non-zero radius (an orbital) will not be discarded. (If I'm not mistaken and the orbital has zero mass, otherwise we'll have to do some more reworking.)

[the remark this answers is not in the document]
Web-eWorks: I'd recommend separating this into two separate `if` blocks, testing individually for `sbody->GetMass() == 0` and `sbody->GetPhysRadius() == 0`. Both parameters are required to be non-zero to generate a valid orbit.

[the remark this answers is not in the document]
Max5377: @Web-eWorks Did you approach with a little tweaking by borrowing physical radius from body itself, because SystemBody doesn't have physical radius.

Gliese852 (approved):

Web-eWorks (approved):

PR #5648, opened 2023-10-16
Space bodies deferred addition

JonBooth78:
Fix the undefined behaviour when addiing bodies during the update loops that iterate through them

The undefined behaviour comes because when we add an item to the vector, it could have to reallocate to grow the backing storage and therefore the end iterator becomes potentially invalid.

Defer adding bodies until the end of an Space update tick.

A few game related scenarios (emerging from hyperspace, saving a game and loading a game) require this update to be invoked too.

I've been playing with this locally for a while and have yet to find any new issues.

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -68,6 +68,8 @@ Game::Game(const SystemPath &path, const double startDateTime, const char *shipT
 
 	m_space.reset(new Space(this, m_galaxy, path));
 
+	m_space->UpdateBodies();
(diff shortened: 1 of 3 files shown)


Web-eWorks:
Hey, I'd almost forgotten about this one! Good work, the code looks clean and I'll give it a quick spin but I don't anticipate any problems.

Web-eWorks (approved):

PR #5651, opened 2023-10-24
Build with SSE4 / AVX, fix build artifact generation

Web-eWorks:
This PR does a wee refactor of our Linux build scripts to ensure we're correctly generating and uploading release artifacts. It moves the release generation steps into a reusable action file, which is important for the other half of the PR.

This PR enables SSE4.2 support by default for our Linux x64 build type. SSE4.2 has been available and supported in mainstream x86/64 processors for over a decade now, so it makes sense to formally require SSE >= 4.2 for our builds (and allow the compiler to perform automatic optimization and vectorization to take advantage of these instruction sets).

GCC's documentation doesn't specify whether I need to manually add all of the SSE version flags or if `-msse4.2` implies previous SSE versions. If someone is more familiar with the semantics GCC provides I'd appreciate your feedback / review. :smile:

This PR also adds an optional x64 build type with AVX2 enabled. AVX2 has 256-bit-wide SIMD instructions compared to SSE's 128-bit-wide SIMD, and has been supported in most mainstream Intel desktop and laptop CPUs since 2013 (2015 for AMD CPUs). The only reason this hasn't been enabled by default is that there are a number of laptops with Intel Pentium and Celeron "mobile" CPUs which do not provide AVX2 hardware.

Generally, laptops with those "mobile" processors do not have the graphical capability to run Pioneer well, as the integrated GPU is very low-power and they're usually not paired with a discrete GPU of any kind. However, I intend to keep AVX2 an "optional" build type until we've established that it's not affecting our userbase and we can provide a good failure-case UX when a user tries to run Pioneer on a non-AVX2 CPU.

### DIFF
--- a/.github/actions/upload-linux/action.yml
+++ b/.github/actions/upload-linux/action.yml
@@ -38,6 +38,8 @@ runs:
 
     - name: Build Release
       shell: bash
+      env:
+        BUILD_SLUG: ${{ inputs.build_slug }}
       run: ./scripts/package-release.sh
 
     - name: Upload Release Files
--- a/.github/actions/upload-linux/action.yml
+++ b/.github/actions/upload-linux/action.yml
@@ -1,0 +1,49 @@
+name: "Upload Linux Artifacts"
+description: "Manage generating and uploading release artifacts for linux builds"
+
+inputs:
+  artifact_name:
+    description: "Name of generated artifact"
+    required: true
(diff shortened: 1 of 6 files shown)


Web-eWorks:
Yes, I see the build failure... Github Actions is very finicky about certain things. Will come back to this tomorrow.

PR #5654, opened 2023-10-31
Show the system security (lawlessness) in the map sector view

JonBooth78:
In the sector view add a security indication for the system being looked at:

image

I've arbitrarily decided on 7 different security levels and English translations for them, a color scheme and a warning icon for some levels, using a blank icon for safe systems as I couldn't find something I like.  I don't know if we need one?

Note: this lawlessness doesn't do much in the current iteration of the code.  Happy to roll this into a bigger PR when I've added more that hooks into it but feel there is enough here for a discussion.

### DIFF
--- a/data/lang/core/en.json
+++ b/data/lang/core/en.json
@@ -1627,6 +1627,38 @@
     "description": "",
     "message": "System orbit view"
   },
+  "SYSTEM_SECURITY_TOOLTIP": {
+    "description": "Describes how safe or dangerous a star system is",
+    "message": "System security:"
+  },
+  "SYSTEM_SECURITY_1": {
+    "description": "Describes how safe or dangerous a star system is",
+    "message": "Safe"
+  },
+  "SYSTEM_SECURITY_2": {
+    "description": "Describes how safe or dangerous a star system is",
+    "message": "Harmless"
+  },
+  "SYSTEM_SECURITY_3": {
+    "description": "Describes how safe or dangerous a star system is",
+    "message": "Mostly Harmless"
+  },
+  "SYSTEM_SECURITY_4": {
+    "description": "Describes how safe or dangerous a star system is",
+    "message": "Somewhat Dangerous"
+  },
+  "SYSTEM_SECURITY_5": {
+    "description": "Describes how safe or dangerous a star system is",
+    "message": "Dangerous"
+  },
+  "SYSTEM_SECURITY_6": {
+    "description": "Describes how safe or dangerous a star system is",
+    "message": "Very Dangerous"
+  },
+  "SYSTEM_SECURITY_7": {
+    "description": "Describes how safe or dangerous a star system is",
+    "message": "Deadly"
+  },
   "SYSTEM_TYPE": {
     "description": "",
     "message": "System type:"
--- a/data/pigui/views/map-sector-view.lua
+++ b/data/pigui/views/map-sector-view.lua
@@ -18,6 +18,7 @@ local ui = require 'pigui'
 local layout = require 'pigui.libs.window-layout'
 
 local player = nil
+local sytleColors = ui.theme.styleColors
 local colors = ui.theme.colors
 local icons = ui.theme.icons
 
@@ -44,8 +45,10 @@ local draw_out_range_labels = false
 local draw_uninhabited_labels = true
 local automatic_system_selection = true
 
-local function textIcon(icon, tooltip)
-	ui.icon(icon, Vector2(ui.getTextLineHeight()), svColor.FONT, tooltip)
+local function textIcon(icon, tooltip, color)
+	if nil == color then color = svColor.FONT end
+	
+	ui.icon(icon, Vector2(ui.getTextLineHeight()), color, tooltip)
 	ui.sameLine()
 end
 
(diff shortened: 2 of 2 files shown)


impaktor:
I was surprised (a while back) to find we didn't already have this indicator, perhaps I'm thinking of OldUI or maybe even original Frontier. Either way, happy to see this.

> a color scheme and a warning icon for some levels, using a blank icon for safe systems as I couldn't find something I like. I don't know if we need one?

Ping @nozmajner

nozmajner:
How about these icons?
image
Most are from the wpzoom set we already use, some changed or used as a base. 
The "Mostly harmless" level is a towel, to take the Hitchhikers Guide reference a bit further.

JonBooth78:
Nice @nozmajner.  I might already have picked that skull one for 'Deadly' :-)

Are we happy with 7 levels reported to the user (I'm just mapping the 0-1 numeric value), so we could have more or less depending on our preferences.

Web-eWorks:
I think 7 levels is a little much, especially given the lack of distinct verbs to refer to them. I'd prefer 5 - Safe, Mostly Harmless, Dangerous, Very Dangerous, Deadly.

impaktor:
I'd even argue 3 is sufficient, but we really do need "Mostly Harmless", and then we probably need one below, and 3 above that  level.

nozmajner:
Five seems a bit simpler, yeah. I've put them to the empty space in the svg:
[icons.svg](https://github.com/pioneerspacesim/pioneer/assets/4182678/88201e36-f151-49e7-b2fc-9e1a1e449a54)
image
I've included all five, so it is straightforward to change if we want to differentiate the skulls later

I did not update the theme file to add them though.

JonBooth78:
@Web-eWorks or others, how does the asset conditioning pipeline work to import a new icons.svg?  I'm probably missing something but I can't see an obvious way to import it.

Web-eWorks:
> how does the asset conditioning pipeline work to import a new icons.svg? I'm probably missing something but I can't see an obvious way to import it.

Edit in Inkscape / etc., save as `.svg` format, overwrite existing icons.svg. The file is loaded as-is and rasterized on startup. Icon name->index mappings are declared in `pigui/themes/default.lua`.

JonBooth78:
Ahh, thanks, I searched for `icons.svg` and couldn't find it.  Was searching the user folder, not the source tree.  duh.  Apologies.

Web-eWorks (changes requested):
After playtesting this, I have some fundamental objections to the effect of this PR. A system like Wolf 359 shows as "very dangerous", even though it's quite literally a star and a single orbital spacestation, with no in-universe reason for a significant enough amount of pirate traffic to threaten a player who might be in a kitted-out AC33.

The way the danger rating is currently communicated to the player feels very "meta", as if it's an out-of-context message to the player directly about whether they will, with their current ship/loadout, survive jumping to that system or spending a significant amount of time there.

I don't think the single `lawlessness` variable for a system is necessarily a good fit for what's being proposed - the `[safe ... deadly]` range is a better fit for a more direct indicator of hostile (e.g. pirate or aggressive faction) presence or spawn chance, and I think a "security rating" number more akin to EVE Online's system might do a better job at expressing the concept of lawlessness in a way that isn't confusing to the player.

The lawlessness of a system shouldn't be a direct indicator of pirate presence - e.g. we don't want to always spawn a pirate megabase in every single 1.0 lawlessness system - but rather an indicator of how well the system is patrolled or secured by the faction who nominally "owns" it. I think a separate factor - possibly derived from the faction system but still influenced at some level by lawlessness - should be introduced to more directly control the spawn chance of hostile encounters and that factor would be a much better fit for this style of "danger indicator".

It'd also be very confusing for a player to see a system marked Very Dangerous or Deadly still have a bunch of civilian trade ships flying around completely unmolested.

In sum, my feedback is I like the idea, but as currently implemented it communicates a clear and present danger to the player ("you *will* die") for even being in the system, rather than a warning indicator that they're "on their own" if they are unlucky enough to encounter hostile elements.

impaktor:
I haven't tried it, but the critique sounds relevant to me: if a system is dangerous player expects there to be danger. 

(perhaps relevant, but I have a [branch](https://github.com/pioneerspacesim/pioneer/compare/master...impaktor:pioneer:piratesystem) that adds pirate system)

fluffyfreak:
The current objections to merging are good but don't they reflect an underlying problem that we need to fix about how we calculate/present the sucrity rating rather than being a blockers for this PR?

As it is I'd merge this PR, once conflicts are resolved, and then we should fixup the lawlessness calculation, or add in a danger/pirate-chance rating and tie that to this displayed value

impaktor:
## In Summary

Two years since this was opened. In summary, we like the idea, but this PR exposes a flaw in how we model "lawlessness", what I read from comment above is we need a second dimension that models how much piracy/violence there is.

- Lawness/lawlessness: how much / little police patrol and laws are enforced
  - "Can I get away with smuggling?"
  - "Will I get help if attacked?"

- Peace/Violence: how much/little pirates and illegal activity is there?
  - "will I get attacked?"
  - "Can I bounty hunt?"
  - "Safe for unarmed ship?"
 
Pirates are spawned based on vertical axis
Police / customs are spawned based on horizontal axis

pioneer-lawlessness-peace-2D-chart_halfsize

bszlrd:
image

[icons-gavel](https://github.com/user-attachments/assets/63a35711-e926-45d1-86d0-a6afcdd04a88)

sturnclaw:
Merged in #6265, there were non-trivial changes needed to update for the new sector map UI.

PR #5655, opened 2023-10-31
Create a game settings tab in the settings view and add a difficulty …

JonBooth78:
…slider to it

Difficulty is stored 2 ways, once for the Engine as a config value that will apply to any new game you start.

A game itself (so a savefile) will persist it's own difficulty and if it's not available on load, will use the engine configured value.

Within lua code the `Game.difficulty` attribute can be used to determine the current difficulty level, from 0 to 1, easiest to hardest.

Also, separate out settings persistence from closing the window, so now if you open the settings window, change the settings and then save the game they are persisted in the saved game state.

image

Note: This currently has no effect at all on the way the game plays, this will come later, however, I believe there is enough here to have a good conversation on the direction this is taking.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -355,6 +355,10 @@
     "description": "tooltip description in settings menu",
     "message": "Lower gives better performance, at cost of visual aesthetics"
   },
+  "DIFFICULTY": {
+    "description": "Label next to the slider that allows the player to set the difficulty of the game",
+    "message": "Difficulty"
+  },
   "DISABLE_SCREENSHOT_INFO": {
     "description": "",
     "message": "Disable Screenshot Info"
@@ -658,6 +662,10 @@
   "GAME_TIME": {
     "description": "",
     "message": "Game time"
+  },  
+  "GAMEPLAY": {
+    "description": "The tooltip for the settings/options tab for changing gameplay related options",
+    "message": "Gameplay"
   },
   "GENERAL_CREW": {
     "description": "",
--- a/data/meta/CoreObject/Game.meta.lua
+++ b/data/meta/CoreObject/Game.meta.lua
@@ -14,8 +14,7 @@
 ---@field sectorView unknown #TODO: add type info for SectorView interface
 ---@field time number Game time in seconds since Jan. 1 3200
 ---@field paused boolean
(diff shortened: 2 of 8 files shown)


impaktor:
Will the player be informed what will change with added difficulty?

My main concern is that adding difficulty level adds more complexity and difficulty to balancing the game, but if you think you can pull it off, I'm happy.

JonBooth78:
@impaktor, great question.

I didn't really think about it this way until you asked.  Do we want 'one difficulty setting to rule them all', or would we be better off having multiple different sliders for things like
- Number of pirates
- Likelihood of pirate attack
- Equipment level of pirates

etc, etc, and then a few buttons for 'Sightseeing', 'Easy', 'Medium', 'Hard', 'Insane' that sets them to presets?

Re game balance, I kind of feel that so long as we get the balance right for the default settings, then the point of this setting is to allow people to tweak the balance more to their preferences and if they've changed it and find the balance off, they can always tweak back.  I don't know if that's the right perspective though, it's not something I've much thought about.

Also, if that is the case, then perhaps having multiple sliders and presets allows folks even more customization opportunities to set the balance the way they like.  It also allows us to balance that small set of presets.  I guess it depends on how much we want to control the experience vs, how much we want players to be able to experiment.

That second approach would also be great for the developers to find the right balance, rather than trying to get a single slider working...

JonBooth78:
Changing this to a draft, I'm experimenting with something different for now so will revisit later.

mwerle:
From a historical perspective (Elite), players could "choose" their difficulty in-game by the government type of system(s) they visited - Democracy being safest to Anarchy being most dangerous. This was also tied to the commodity/mission markets with less safe systems generally having higher potential profit margins. 

Not sure if this is already factored into Pioneer? I see systems have a large variety of government types and economies.

(in my latest PR I also added a general "Settings" tab to the options window for a new setting.. I should have looked at this PR before doing so!)

From a technical perspective, I would generally never use a float/double to store data if an integer can be used instead. In this case the code is even performing floating point rounding operations to ensure things work correctly. Why not just use an `int` and values from 0..100? I doubt difficulty would ever need to be a fractional part of a percent, and even if, just go from 0..1000, or 0..10000.

PR #5658, opened 2023-11-03
Compute value of g-force in system editor

zonkmachine:
In the System Editor the surface gravity is presented as acceleration  m/s^2.

* Compute value of 'g' as is done in [system-view-ui.lua](https://github.com/pioneerspacesim/pioneer/blob/23b6700a16a788ced5ae7ea11283489e05cfcd0d/data/pigui/modules/system-view-ui.lua#L810)
* Format change - add space between value and 'g'.
* Add the last decimal of standard gravity to it's full 9.80665.

### DIFF
--- a/data/pigui/modules/system-view-ui.lua
+++ b/data/pigui/modules/system-view-ui.lua
@@ -807,7 +807,7 @@ function Windows.objectInfo:Show()
 				{ name = lc.RADIUS, icon = icons.body_radius,
 					value = (not starport) and ui.Format.Distance(body.radius) or nil },
 				{ name = lc.SURFACE_GRAVITY, icon = icons.body_radius,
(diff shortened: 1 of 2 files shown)


Web-eWorks (approved):

zonkmachine:
I added the last digit to the gravity constant. This is certainly not important as it's only used for information purposes, and is abbreviated in the game, but that's the complete constant and I couldn't stand seeing one digit left out. OCD'ish.

Merge at will. Please squash... :)

impaktor:
(technically, G varies quite a bit depending on where on earth you measure it, so no need to specify it with more than 1-2 decimals).

nozmajner:
I agree, one or two decimals is enough in my opinion. We show one on ship screens, and two on the map screen for example.

PR #5660, opened 2023-11-04
Remove transparent backgrounds from screenshots

### LINKED ISSUE #4671: Transparent backgrounds in screenshots
Taking in-game screenshots (Ctrl+PrScr), becomes transparent background (at least when doing it in the PiGUI screens). 

As noted by @vakhoir [here](https://github.com/pioneerspacesim/pioneer/pull/4624#issuecomment-510613586):

> Ok, I almost figured it out. The button coloring was broken by the change in Color.h in c19aab4, and I'm still trying to get to the bottom of it, but the transparent screenshots started when #4626 was merged

Also, there's some unrelated bug with button background coloring?

JonBooth78:
This fixes #4671

Also, somewhat more optimal and safer, given the old code appears to be copying most bytes 3/4 times and then copying 2/3 bytes off the end of the image, if I read it correctly.

As a side effect screenshot images are a bit smaller now too.

### DIFF
--- a/src/graphics/opengl/RendererGL.cpp
+++ b/src/graphics/opengl/RendererGL.cpp
@@ -1190,30 +1190,4 @@ namespace Graphics {
 		return true;
 	}
 
-	bool RendererOGL::FrameGrab(ScreendumpState &sd)
-	{
-		int w, h;
-		SDL_GetWindowSize(m_window, &w, &h);
-		sd.width = w;
-		sd.height = h;
-		sd.bpp = 4; // XXX get from window
-
-		sd.stride = (4 * sd.width);
-
-		sd.pixels.reset(new Uint8[sd.stride * sd.height]);
-
-		glBindFramebuffer(GL_FRAMEBUFFER, 0);
-		glPixelStorei(GL_PACK_ALIGNMENT, 4); // never trust defaults
-		glReadBuffer(GL_FRONT);
-		glReadPixels(0, 0, sd.width, sd.height, GL_RGBA, GL_UNSIGNED_BYTE, sd.pixels.get());
-		glFinish();
-
-		// this might harmlessly error if we're in a single buffered mode,
-		// however in double buffered mode it makes the window in window screens
-		// such as ones that show the ship within a menu (F3) work correctly
-		// as GL_BACK is the default.
-		glReadBuffer(GL_BACK);
-		return true;
-	}
-
 } // namespace Graphics
--- a/src/graphics/opengl/RendererGL.cpp
+++ b/src/graphics/opengl/RendererGL.cpp
@@ -2,8 +2,8 @@
 // Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
 #include "RendererGL.h"
-#include "RefCounted.h"
 #include "MathUtil.h"
+#include "RefCounted.h"
 #include "SDL_video.h"
 #include "StringF.h"
 
@@ -1089,8 +1089,8 @@ namespace Graphics {
 		if (desc.colorFormat != TEXTURE_NONE && desc.depthFormat != TEXTURE_NONE && !rt->CheckStatus()) {
 			GLuint status = glCheckFramebufferStatus(GL_FRAMEBUFFER);
 			Log::Error("Unable to create complete render target. (Error: {})\n"
-				"Does your graphics driver support multisample anti-aliasing?\n"
-				"If this issue persists, try setting AntiAliasingMode=0 in your config file.\n",
+					   "Does your graphics driver support multisample anti-aliasing?\n"
+					   "If this issue persists, try setting AntiAliasingMode=0 in your config file.\n",
 				gl_framebuffer_error_to_string(status));
 		}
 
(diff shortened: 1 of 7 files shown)


JonBooth78:
Not quite sure what to do about the commented out alternative path, or if the optimal path is really required, go for simplicity, leave as it or something else?

Web-eWorks:
I'm not a great fan of the built-in screenshot functionality, but I can see the argument for having it, especially with the existence and quality of screenshot functionality varying wildly between OS brand and distribution. I'd recommend paring down the write_png function to only what's absolutely necessary for taking screenshots / dumping the backbuffer; other uses for the function should either go through a "real" PNG library for e.g. editor asset management or be implemented as an in-process tool for e.g. inspecting intermediate render targets.

If possible, I'd highly recommend all or almost all code for "taking a screenshot" be moved into the PngWriter file and decoupled from the rest of the code; as of the last time I worked on it, we have to link the savegame dump *command line utility* against SDL2_Image because PngWriter is a transitive dependency somewhere and somehow.

JonBooth78:
Hi @Web-eWorks , I'm not quite sure what you're suggesting.

I can see 3 places that use this `"PngWriter.h"` in our codebase
- `Pi.cpp` where it is used to take in-game screenshots, using `write_screenshot`
- `ModelViewer.cpp` where it is used to take screenshots, I assume in that tool, using `write_screenshot`
- `GasGiant.cpp` where it is used to save a png of a gas giant if `DUMP_TO_TEXTURE` is `#define`d.  This directly uses `write_png`.

Are you suggesting moving the fragments of code in the first two cases into `write_screenshot`, consolidating the two functions in that file and removing the third?

And/or are you suggesting we remove the dependency on SDL to save png files?  This could be achieved, for example, by adding a dependency on libpng and does seem a better solution.  This is something I could do, it's been years since I looked at libpng but I'd wager it hasn't changed much.   However, I think that's out of the scope of this PR.

It seems like the `PngWriter.cpp` file is compiled as part of the `pioneer-lib` target.  I'm not sure why this is getting pulled in as a transient dependency of the savegame dump utility unless somehow it's use of something else inside `Pi.cpp` is causing `Pi::HandleKeyDown` to pull it in and the linker hasn't managed to prune it all.  Although why the command line utility would get caught up in all that, I don't know !?!  I do see that if I remove the additional dependency on `SDL2_Image` from the `savegamedump` executable when building with visual studio it links just fine.

Web-eWorks:
> I can see 3 places that use this "PngWriter.h" in our codebase

This is fine - if there's duplicated functionality between the three paths then coalesce them; otherwise it's not a huge exposure. I'd probably suggest removing the GasGiant.cpp codepath entirely, as the need to dump the texture is completely obviated by the in-process texture cache explorer via the Ctrl+I menu.

> And/or are you suggesting we remove the dependency on SDL to save png files?

No, I don't think that's necessary -  I apologize for the wild goose chase I've inadvertently led you down. I have a distinct memory from 3 or 4 years gone of setting up one of the CLI executable targets and it requiring a linkage against SDL_Image for very very non-obvious reasons connected to the screenshot functionality, but it's very possible it's been "fixed at a distance" in the intervening time and I've simply not noticed the knock-on effect.

>  This could be achieved, for example, by adding a dependency on libpng and does seem a better solution.

I'd rather not pull any additional libraries in for the purpose of writing out screenshots on the command line. If we introduce a resource loading system of some sort I'd be happy to revisit the question but I think SDL_Image is a fine (and small) enough dependency for the purpose.

>  I do see that if I remove the additional dependency on SDL2_Image from the savegamedump executable when building with visual studio it links just fine.

Excellent news - if that also builds on the CI machines then I'll consider the whole mess resolved. :smile:

Web-eWorks:
...this would be because I in fact *was* the one to fix everything in the world linking against PngWriter - it used to be linked into `utils.cpp` back when that was still a thing 80% of the codebase depended on.

If interested, this is the commit responsible for fixing the dependency on PngWriter https://github.com/pioneerspacesim/pioneer/commit/6a1945874f93354d160c9fd4deaa8c0f410f886b - it seems I simply forgot to remove the linkage against SDL_Image in the `savegamedump` target.

JonBooth78:
Nice :-)

No worries, I only spent about 15 mins looking at this; no wild geese involved.

I'll remove the `DUMP_TO_TEXTURE` path from the gas giant code, see if I can remove the dependency in `CMakelists.txt` and tidy the rest over the next week or so.

Web-eWorks:
Bump @JonBooth78 - I'd like to merge this ahead of Pioneer Day if your schedule permits making the needed changes in that timeframe :smile:

JonBooth78:
Hi @Web-eWorks , I've done all the tidy up required and then fixed another bug I had (when taking screenshots my F3 view with the ship stopped showing the ship, instead a picture in picture of itself).

I also stripped out a dead function in the renderer that would need to be considered if we ever put video recording back into the game.  I added a comment by the `ifdef`ed out code that might want to call it.  I believe that function was buggy anyway as per my comments in the commit removing it.

src/graphics/opengl/RendererGL.cpp
```diff
@@ -1089,8 +1089,8 @@ namespace Graphics {
 		if (desc.colorFormat != TEXTURE_NONE && desc.depthFormat != TEXTURE_NONE && !rt->CheckStatus()) {
 			GLuint status = glCheckFramebufferStatus(GL_FRAMEBUFFER);
 			Log::Error("Unable to create complete render target. (Error: {})\n"
-				"Does your graphics driver support multisample anti-aliasing?\n"
-				"If this issue persists, try setting AntiAliasingMode=0 in your config file.\n",
+					   "Does your graphics driver support multisample anti-aliasing?\n"
+					   "If this issue persists, try setting AntiAliasingMode=0 in your config file.\n",
```
Web-eWorks, line 1093: Sigh... clang-format constantly tries to squirrel its way into aligning arguments or parameters in an incorrect way. Nothing you need to address here, but I'll have to review the clang-format file _again_, for the Nth time.

Web-eWorks (approved):

PR #5661, opened 2023-11-08
Improve Simplex Noise Performance

Web-eWorks:
For fun yesterday, I ported FastNoise2's Simplex Noise implementation as a competitor to our Simplex `noise()` function taking advantage of SIMD operations. Unfortunately, the overhead required for a SIMD-capable version of the Simplex algorithm almost outweighs the benefit of calculating an additional double-precision noise value per function call.

As part of that work, I found an alternate way to express the truth table in our `noise()` implementation that replaces the complicated if-block with a series of bit operations. This caused our current noise function to be almost as fast as the SIMD version of the algorithm, mostly due to the significantly fewer operations involved in computing the Simplex gradient terms in our existing version.

On my machine, compiling for SSE2, this provides an approximately 20% performance improvement to noise generation functions (33 -> 26 MCycles according to profiling data.) You can see the full table of all the implementations and instruction set levels I tried [here](https://docs.google.com/spreadsheets/d/1_N1cGNaIUIQOQziovTFFBZVs7w87pi9r-gcUFHltCj0/edit?usp=sharing):
- `Noise::Simplex` is a drop-in replacement for `noise()` generating a single value at a time (for apples-to-apples comparison)
- `OctaveNoiseSingle` replaces our `octavenoise()` wrapper with a version that computes multiple octaves in parallel
- The "vectorized" versions are operating in SIMD mode, while scalar versions are using a fallback for platforms without supported vector instructions
- `noise()+` is the code in this PR.

### DIFF
--- a/src/perlin.cpp
+++ b/src/perlin.cpp
@@ -132,57 +132,18 @@ double noise(const vector3d &p)
 	int i1, j1, k1; // Offsets for second corner of simplex in (i,j,k) coords
 	int i2, j2, k2; // Offsets for third corner of simplex in (i,j,k) coords
 
-	if (x0 >= y0) {
-		if (y0 >= z0) {
-			i1 = 1;
-			j1 = 0;
-			k1 = 0;
-			i2 = 1;
-			j2 = 1;
-			k2 = 0;
-		} // X Y Z order
-		else if (x0 >= z0) {
-			i1 = 1;
-			j1 = 0;
-			k1 = 0;
-			i2 = 1;
-			j2 = 0;
-			k2 = 1;
-		} // X Z Y order
-		else {
-			i1 = 0;
-			j1 = 0;
-			k1 = 1;
-			i2 = 1;
-			j2 = 0;
-			k2 = 1;
-		} // Z X Y order
-	} else { // x0<y0
-		if (y0 < z0) {
-			i1 = 0;
-			j1 = 0;
-			k1 = 1;
-			i2 = 0;
-			j2 = 1;
-			k2 = 1;
-		} // Z Y X order
-		else if (x0 < z0) {
-			i1 = 0;
-			j1 = 1;
-			k1 = 0;
-			i2 = 0;
-			j2 = 1;
-			k2 = 1;
-		} // Y Z X order
-		else {
-			i1 = 0;
-			j1 = 1;
-			k1 = 0;
-			i2 = 1;
-			j2 = 1;
-			k2 = 0;
-		} // Y X Z order
-	}
+	int x_ge_y = x0 >= y0;
+	int y_ge_z = y0 >= z0;
+	int x_ge_z = x0 >= z0;
+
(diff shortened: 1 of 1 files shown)


fluffyfreak:
That looks like it should be quicker, I just tried it with a couple of loops of original vs optimised timed against wall clock and get ~20% slower.
Not sure what's happening there as it doesn't make sense. I suspect that it's a VS issue.

I'd be interested in hearing about the other optimisations like parallel Octave etc

Web-eWorks:
> Not sure what's happening there as it doesn't make sense. I suspect that it's a VS issue.

Very baffled there as well - unless MSVC has a very non-optimized codepath for computing boolean negation closely followed by bitwise-and the code should be *at least* as fast as the if-block version and ideally faster.

> I'd be interested in hearing about the other optimisations like parallel Octave etc

Sure - the parallel version I tested used the [FastNoise2 Simplex(seed, x, y, z)](https://github.com/Auburn/FastNoise2/blob/master/include/FastNoise/Generators/Simplex.inl#L56) algorithm, ported to `xsimd` as a flexible SIMD library for use with double-precision floats. You can probably already see some of the differences, but the main difference is the use of a hash function to generate and calculate the gradient dot-product rather than a multi-stage table lookup (because SIMD multi-lookup would be a combinatorial explosion).

This results in a significantly slower implementation when the algorithm is operating on scalar elements (tested with an #ifdef to swap out the SIMD datatype with regular doubles), but obviously the benefit is that it calculates 2 noise values in one call on 128-bit SIMD. I took advantage of this within the existing API by "squashing" the parameters of the repeated `noise()` calls inside of the `octavenoise` variants into an array and computing them 2 or 4 at a time with SSE4/AVX respectively.

This unfortunately didn't result in a significantly better implementation, as the extra overhead to re-arrange the octaves, the cost of "packed" 128/256-bit SIMD ops, and the often narrow octave count (3-6 for most calls) was beaten out by the simplicity of our current noise function. The details of the timing is provided in the table linked in the PR description - while I'm a little suspicious of the vectorized AVX2 implementation's results, AVX2 is still an "optional" target and not worth the effort to target specifically at this time.

While we could certainly refactor the entire terrain height generation API to compute multiple samples in parallel from the top-down, I think the effort spent would be much better served writing a GPU implementation instead to generate terrain data in significantly wider SIMD than the CPU is able to provide. That would most likely involve porting our terrain generation to operate on single-precision floats instead (as doubles are half-rate at best and potentially significantly worse or missing outside of desktop GPUs), so that's a topic for another day (or year).

impaktor:
@fluffyfreak While we have you on the line: rob notified on IRC that Sourceforge will purge our builds. Since you're the other person who have access to sourceforge, and the files there are from back when you were active, I figured you'd might want to know. Is this something we want to preserve somewhere? (where?) One windows build from each year?

(sorry for the off topic. made a [forum post](https://forum.pioneerspacesim.net/viewtopic.php?f=3&t=735) where we can discuss it further)

PR #5662, opened 2023-11-09
Simplify SystemBody::GetAtmPressure

Mc-Pain:
surfaceGravity_g is implicitly divided by and explicitly multiplied by.
Remove it from both numerator and denominator.

L = g/H (specific heat)
gM/RL = gM/R(g/H) = gMH/Rg = MH/R


Web-eWorks (approved):

impaktor:
@Mc-Pain I suspect you're not in AUTHORS.txt. Please let us know if you want us to add you (and if we can use your real/commit name)

PR #5663, opened 2023-11-12
Lua Constant autocomplete support

Web-eWorks:
This PR improves our `scan_enums` script to dump the literal string constants into Lua type information files, providing an additional source of truth when trying to use those enums. It uses the LuaLSP `---@enum` annotation to provide autocomplete for function parameters declared with an enum type (although it will currently also incorrectly suggest e.g. `Constants.BodyType.1` if you don't type a `"` to filter the autocomplete results).

This also updates the `enum_table` file with the changes to the SystemBody enums merged as part of the Json system definition work.


impaktor:
Remind me, it's the [lua-language-server](https://github.com/sumneko/lua-language-server) that we're using, right?

Web-eWorks:
> Remind me, it's the [lua-language-server](https://github.com/sumneko/lua-language-server) that we're using, right?

Correct.

impaktor:
And to make use of this new power, one needs only use LSP in your IDE of choice + run the scan enum's python script?

Web-eWorks:
> And to make use of this new power, one needs only use LSP in your IDE of choice + run the scan enum's python script?

All you need to do is use an LSP adapter to run lua-language-server and configure it to load meta files from `data/meta` - I've already run `scan_enums.py` and we check the meta files into the repository. Details of the configuration vary between IDEs - our "default" VScode workspace is pre-configured.

PR #5664, opened 2023-11-15
Update AppStream destination and filename

reinerh:
the old path is deprecated.
See: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location


Web-eWorks:
CC @pcercuei regarding flatpak etc. packaging requirements.

pcercuei:
I didn't know about that. Is it just the path that changed, or the format as well?

reinerh:
The format is still the same, only other paths are now used.

pcercuei:
See also: https://docs.flatpak.org/en/latest/conventions.html#appdata-files

So LGTM.

PR #5665, opened 2023-11-18
Fix crash in l_body_get_ground_position if there is no body to get po…

JonBooth78:
…sition relative to

Every time I try to hyperspace I was crashing here (jumping from a space station, if that matters).  Well the crash happened in `GetAltitudeRelTo` which was being passed a `null` body as the first argument.

@Max5377 , I think this relates to your earlier PR.

### DIFF
--- a/src/lua/LuaBody.cpp
+++ b/src/lua/LuaBody.cpp
@@ -668,12 +668,18 @@ static int l_body_get_ground_position(lua_State *l)
 		lua_pushnil(l);
 		return 1;
 	}
+	Body* astro = f->GetBody(); // null
+	if (!astro)
+	{
+		lua_pushnil(l);
+		return 1;
+	}
+
 	vector3d pos = b->GetPosition();
 	double latitude = atan2(pos.y, sqrt(pos.x * pos.x + pos.z * pos.z));
 	double longitude = atan2(pos.x, pos.z);
 	lua_pushnumber(l, latitude);
 	lua_pushnumber(l, longitude);
-	Body *astro = f->GetBody();
 	double altitude = b->GetAltitudeRelTo(astro, altType);
 	lua_pushnumber(l, altitude);
 	return 3;
--- a/src/lua/LuaBody.cpp
+++ b/src/lua/LuaBody.cpp
@@ -395,8 +395,7 @@ static int l_body_get_altitude_rel_to(lua_State *l)
 	Body *b = LuaObject<Body>::CheckFromLua(1);
 	const Body *other = LuaObject<Body>::CheckFromLua(2);
 	AltitudeType altType = AltitudeType::DEFAULT;
-	if (!lua_isnoneornil(l, 3))
-	{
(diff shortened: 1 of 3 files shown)


Max5377:
I haven't considered that this can happen when doing `PR`. Thanks for spotting this.
P.S.: Maybe the better way would be to return `null` in `Frame::GetFrame` if no `body` is in this frame?

JonBooth78:
Hey @Max5377 , thanks for looking at this.

I *think* a frame reference with no body attached is fine and so we probably should still return that frame.

I'm not really across all the details here so if you'd like to progress this, how about you create a PR from your branch and then I'll close this one?

Max5377:
@JonBooth78 `cherry-pick` is a command that basically allows you to copy someone else's commit into your branch. https://docs.github.com/en/desktop/managing-commits/cherry-picking-a-commit-in-github-desktop
https://git-scm.com/docs/git-cherry-pick
https://www.atlassian.com/git/tutorials/cherry-pick

You can also give write access to your repo by doing this
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-teams-and-people-with-access-to-your-repository
so I can quickly change your commit.
I would modify this PR directly, but only mainteiners can do that.

I could also do PR to your branch, but I don't know if this would cause merge conflicts, because from the looks you haven't synced your fork in a long time.
I don't really want to do additional PR since, you've already done one that addresses this.

JonBooth78:
Hey @Max5377,I know how to pull more of your changes into this PR and I've done it for now.  I'm just a bit concerned if there are more changes at code review, I might slow things down :-)

If you wanted to do a PR into my repo, you wouldn't want to do it to my `master` as yeah, that's a bit behind (and ahead), instead you'd want to do a PR to the source branch of this PR.  Still that would leave me possibly in the middle.  We'll see how we go.

src/lua/LuaBody.cpp
```diff
@@ -668,12 +668,18 @@ static int l_body_get_ground_position(lua_State *l)
 		lua_pushnil(l);
 		return 1;
 	}
+	Body* astro = f->GetBody(); // null
+	if (!astro)
+	{
+		lua_pushnil(l);
+		return 1;
+	}
```
Web-eWorks, line 676: As a nitpick for both @JonBooth78 and @Max5377 - Pioneer's codebase prefers the use of starting braces on the same line for `if` and `for` statements, as well as class/struct definitions and single-line function definitions. Multi-line function definitions should (assuming I haven't forgotten anything) be the only structure with their beginning brace on a newline.

The clang-format CI check would be complaining at you over this... if it hadn't been silently broken for quite some time and I didn't notice. :neutral_face:

JonBooth78:
Ahh, man sorry @Web-eWorks , tidied those up for you now :-)

I'll also update my VS pro settings to remind me!

Web-eWorks (approved):

PR #5666, opened 2023-11-19
Update the flightlog rebased

JonBooth78:
### User facing changes:

_Note: This replaces / closes #5636 as due to the way I did some merging along the way (since rebased away and hidden in this branch), I could no-longer update that PR.

_Note: Screenshots below have alpha baked into them so the background looks a little off on the export window as that grey is actually transparent so the background actually appears black_

#### Export Collapsed

screenshot-20231119-160052

#### Export Expanded

screenshot-20231119-160057

#### Description

- Consolidate the three tabs of the flightlog into a single page with all entries on it
- Add an options bar to the right hand side to toggle on and off individual types of log entry from the displayed list so you can still see what you’d like
- When pruning the flightlog for space reasons, never remove any elements that have a custom annotation/note added by the user
- Split system flight log entries into a separate arrival and destination entry if there are other entries that come chronologically between them
- Add the ability to order the flightlog oldest to newest as well as the default newest to oldest
- Remove the spurious log on game start of the player docking at their starting station
- Add the option showing the current player stats to the flightlog at the top
- Add the ability to export the currently displayed flightlog to a file, located in the userdata folder based on the player’s name, either as plain text or HTML

### Potentially controversial changes I’m aware of:

- Adding an optional window layout to the tab layout to allow me to add a floating window at the bottom of the screen.  I'm also considering (in the future) to use this to allow a popup when editing a flightlog entry, with a multi-line entry box as although `imgui` doesn't support word-wrapping entry boxes this would feel nicer and mitigate that a bit too.
- This is very much showing my working in the order of commits

### Things I’ve tested:

- Loading existing games
- Saving games and reloading them with the new flightlog format
- Starting a new game
- Jumping between systems
- Landing at spaceports
- Custom flightlog entries and adding custom annotations to other entries

### Things I am considering to doing later:

- Change the backing model such that `system` and `station` events are container events that can contain events that happened when within that system or station.  System events of course will contain station events.
- Buttons to add the location of flightlog entries to the current system navigation route
- Automatic flight log entries when a ship is killed by the player
- Automatic flight log entries when the player changes their ship
- Automatic flight log entries when the player changes the ship equipment
- Automatic flight log entries for commerce, so when the user purchases or sells cargo
- Automatic flight log entries when missions are started/failed/completed
- Automatic flight log entries when a screenshot is taken and some way to view it from the log (as long as it’s still on disc)
- Filters for all of the above
- CSV export


impaktor:
1. @Web-eWorks I assume this is good to merge, as I think that was the state of the original PR this is based from?

2. 

> This is very much showing my working in the order of commits

So rebase, or we can just squash merge, or do we want to keep all commits? E.g I see new functions like `l_pigui_set_next_window_collapsed` and `ui.setNextWindowCollapsed` that could be their own "New function to allow X". I just don't want the full development history in master, as some developers rely on clean / non-messy commit history to understand which feature caused a line to change.

JonBooth78:
@Web-eWorks , I think this is ready for final review, although looks like it'll need rebasing again on master.

JonBooth78:
Rebased. I'm wondering if I'll hit some clang format errors now ?

JonBooth78:
Yeah, hit some clang format errors, I'll have to sort them out later in the week - sorry

JonBooth78:
@Web-eWorks , this should now be ready for review.

data/modules/FlightLog/FlightLogEntries.lua
```diff
@@ -0,0 +1,397 @@
...
+end
+
+---@param entry string A user provided description of the event.
+---If non nil/empty this will cause the entry to be considered custom and not automatically deleted
+---@return nil
+function FlightLogEntry.Base:UpdateEntry( entry )
+
+	if self:IsCustom() then
+		FlightLogEntry.TotalDefaultElements = FlightLogEntry.TotalDefaultElements-1
+	end
```
Web-eWorks, line 111: Note: this seems conceptually messy. I don't think it's the responsibility of the individual entries to track how many entries in the overall flightlog are non-custom, and it means there's quite a bit of code scattered around manipulating this variable.

Instead, I'd personally implement it so that adding and updating entries in the flight log goes through the main FlightLog interface and it's the responsibility of those two API functions (e.g. `FlightLog.AddEntryToLog(entry)` / `FlightLog.UpdateEntryNote(entry, note)`) to track changes to the `TotalDefaultElements` count. This also means that constructing a FlightLogEntry variable and then *not* adding it to the FlightLog doesn't wind up corrupting the overall count.

This is maybe not the traditional Object Oriented way to do it, where you get a reference to the entry itself and then ignore the existence of the owning container afterwards, but I think it's a slightly cleaner API model with significantly less internal bookkeeping required.

data/modules/FlightLog/FlightLogEntries.lua
```diff
@@ -0,0 +1,397 @@
...
+	FlightLogEntry.Base.Constructor( self, Game.time, nil, true )
+end
+
+---@return string The name for this log entry type
+function FlightLogEntry.CurrentStatus:GetLocalizedName()
+	return l.PERSONAL_INFORMATION;
+end
+
+---@param earliest_first boolean set to true if your sort order is to show the earlist first dates
+---@return table<string, string>[] An array of key value pairs, the key being localized and the value being formatted appropriately.
```
Web-eWorks, line 179: Note: `table<string, string>` is an annotation for a key-value table. The type annotation for this would be `{ [1]: string, [2]: string }[]`. (This should be applied to all instances of GetDataPairs.)

data/modules/FlightLog/FlightLogExporter.lua
```diff
@@ -0,0 +1,137 @@
...
+end
+
+function text_formatter:newline()
+	self.file:write( "\n" )
+	return self
+end
+
+function text_formatter:close()
+	self.file:close()
+end
```
Web-eWorks, line 29: Remember to assign `self.file = nil` to avoid keeping a dangling reference to the closed file handle in the Lua GC - otherwise it will remain reachable until the next time the player exports a log.

data/pigui/libs/forwarded.lua
```diff
@@ -22,6 +22,11 @@ ui.getWindowContentSize = pigui.GetWindowContentSize ---@type fun(name: string):
 ui.setNextWindowPos = pigui.SetNextWindowPos ---@type fun(pos: Vector2, cond: string, pivot: Vector2?)
 ui.setNextWindowSize = pigui.SetNextWindowSize ---@type fun(size: Vector2, cond: string)
 ui.setNextWindowSizeConstraints = pigui.SetNextWindowSizeConstraints ---@type fun(min: Vector2, max: Vector2)
+--- Collapse or expand the next window
+---@param collapse boolean|nil If not set or true, collapse the next window, else expand it 
+function ui.setNextWindowCollapsed( collapse )
+    pigui.SetNextWindowCollapsed( collapse )
+end
```
Web-eWorks, line 29: The addition of an extra function call here is unnecessary at best. I strongly recommend defining this as:
```suggestion
--- Collapse or expand the next window
ui.setNextWindowCollapsed = pigui.SetNextWindowCollapsed ---@type fun(collapse: boolean?)
```

We prefer to define all "wrapping functions" which add extra logic in addition to the exported functions in `data/pigui/libs/wrappers.lua` or a topic-specific file.

data/pigui/modules/info-view/06-flightlog.lua
```diff
@@ -37,191 +59,170 @@ local function headerText(title, text, wrap)
...
+				entry:Delete()
+				-- if we were already in edit mode, reset it, or else it carries over to next iteration
+				entering_text = false
+			end
 		end
 
 		ui.nextColumn()
 		ui.separator()
-		ui.spacing()
+		ui.spacing()
```
Web-eWorks, line 136: Trailing whitespace here.

data/pigui/modules/info-view/06-flightlog.lua
```diff
@@ -37,191 +59,170 @@ local function headerText(title, text, wrap)
...
-			-- not next record's. Quick, behind you, velociraptor!
-			goto input
+	local counter = 0
+	for entry in FlightLog:GetLogEntries(getIncludedSet(), nil, earliest_first ) do
+	 	counter = counter + 1
+	
+		 writeLogEntry( entry, formatter, true )
+
+		 if entry:CanHaveEntry() then
+			::input::
```
Web-eWorks, line 114: As I've discussed with you prior, we prefer not to use `goto` unless absolutely necessary. I don't see a surviving `goto` call in the diff - if that's been removed then I'd prefer we also remove the label as well.

src/lua/LuaPiGui.cpp
```diff
@@ -884,6 +884,17 @@ static int l_pigui_set_next_window_pos(lua_State *l)
 	return 0;
 }
 
+static int l_pigui_set_next_window_collapsed(lua_State *l)
+{
+	bool collapsed = true;
+	if (lua_isboolean(l, 1)) {
+		collapsed = lua_toboolean(l, 1);
+	}
```
Web-eWorks, line 892: Note: this should be expressed as `bool collapsed = LuaPull<bool>(l, 1, true);` which does an implicit nil-check and handles the case in which the function stack is smaller than 1 (i.e. when the optional parameter is actually optional).

Web-eWorks (changes requested):
I'm quite happy with the effect and direction of this PR. I do apologize for not getting through the review until now - my schedule has been quite busy and I've accidentally let things slip through the cracks like this review.

I've left (several!) comments where there are things I'd like to see addressed, either as part of this PR directly or in a follow-up PR if you're not confident in being able to address them all in the next ~2 weeks. We'll be entering formal "feature freeze" on the 27th and anything not bugfix/stability related will be put on hold at that point until the release.

Overall I consider this PR to be functional enough to merge as part of the release, but in addition to the nitpicks that comprise most of the comments I've left there are a few maintainability concerns regarding serialization and dataflow that I'd like to see addressed long-term.

data/pigui/modules/info-view/06-flightlog.lua
impaktor: (I _think_ I was the original author to put in `goto` in this code. Thus might be a remnant from my first implementation of it... just to not put blame for such an atrocious act of violence upon poor @JonBooth78 ).

data/modules/FlightLog/FlightLog.lua
```diff
@@ -0,0 +1,326 @@
...
+
+local utils = require 'utils'
+
+local Character = require 'Character'
+
+local FlightLogEntry = require 'modules.FlightLog.FlightLogEntries'
+
+
+---@type boolean|nil  If true then we've just started a game so don't record the first docking callback
+local skip_first_docking = nil
```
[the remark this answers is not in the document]
JonBooth78: Indeed it isn't, removing :-)

data/modules/FlightLog/FlightLog.lua
```diff
@@ -0,0 +1,326 @@
...
+---@param maximum integer|nil              Maximum number of entries to include
+---@param earliest_first boolean|nil       Should the log start with the oldest entry or the most recent
+--- 
+---@return function():FlightLogEntry.Base  An iterator function that when called repeatedly returns the next entry or nil when complete
+---
+--- Example:
+---
+--- > for entry in FlightLog.GetLogEntries( { "Custom", "System", "Station" ) do
+--- >   print( entry.GetType(), entry.entry )
+--- > end
```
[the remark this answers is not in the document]
JonBooth78: OK, that's good to know, this was my first foray into any kind of parameter hinting in Lua and I pretty much learned as I went along so I understand it's probably a bit rough.

@Web-eWorks , if I understand correctly, to do in entirety what you're suggesting here I'd need to have the following:

```
-- Method: GetLogEntries
--
-- Example:
--
-- > for entry in FlightLog.GetLogEntries( { "Custom", "System", "Station" ) do
-- >   print( entry.GetType(), entry.entry )
-- > end
--
-- Parameters:
-- 
--    types         A table where keys are log types to include and the value is a boolean, set to true to include that type
--    maximum   An optional integer, with the maximum number of entries to include
--    earliest_first An optional boolean to say if the log should be ordered in chronological order or reverse chronological order
-- 
-- Returns:
--
--    An iterator function that when called repeatedly returns the next entry or nil when complete
--
--
---@param types table[string,boolean]?     Keys are log types to include, set the boolean to true for the ones you want
---@param maximum integer?                 Maximum number of entries to include
---@param earliest_first boolean?          Should the log start with the oldest entry or the most recent
--- 
---@return fun():FlightLogEntry.Base       An iterator function that when called repeatedly returns the next entry or nil when complete
```

data/pigui/modules/info-view/06-flightlog.lua
JonBooth78: Ahh, yes, sorry; I'd missed the label but have I successfully removed the `goto`

src/lua/LuaPiGui.cpp
JonBooth78: Ahh, nice, I think I hadn't got my head around the `LuaPull` semantics at this time; that's much nicer.

data/modules/FlightLog/FlightLog.lua
```diff
@@ -0,0 +1,326 @@
...
+		table.sort( FlightLogData, sortf )
+
+		CollapseSystemEvents()
+
+	elseif loaded_data and loaded_data.Version > 1 then
+
+		local loader_funcs = {}
+		loader_funcs[FlightLogEntry.System.GetSerializationIndex()] = FlightLogEntry.System.CreateFromSerializationElements
+		loader_funcs[FlightLogEntry.Station.GetSerializationIndex()] = FlightLogEntry.Station.CreateFromSerializationElements
+		loader_funcs[FlightLogEntry.Custom.GetSerializationIndex()] = FlightLogEntry.Custom.CreateFromSerializationElements
```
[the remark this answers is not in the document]
JonBooth78: Yeah, that's far nicer, thanks :-)

data/modules/FlightLog/FlightLog.lua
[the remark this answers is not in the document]
Web-eWorks: I personally don't add descriptions to the `---@param` annotations for public functions which are also documented via NaturalDocs, as that's duplicated effort.

Otherwise that looks fairly correct - although the Natural Docs parameters list needs to be formatted with a `-` to separate the parameter name from the description. We sometimes prefer to include the type of the parameter in the start of the description string to provide extra information to people reading the compiled codedoc:

```
types - [type description], human description
```
e.g.
```
types - optional table<string, boolean>, controls which flight log types should be included in the iterator results
```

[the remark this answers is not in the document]
Web-eWorks: Additionally, `table<string, boolean>` is a key-value table type annotation that will be parsed correctly by the Lua language server. `table[string, boolean]` is a invalid type annotation and the type will be deduced as a general-purpose `table` value followed by a comment string. Square brackets are only used in array-of-type annotations as e.g. `string[]`.

data/modules/FlightLog/FlightLogEntries.lua
```diff
@@ -0,0 +1,397 @@
...
+---@field protected sort_date number Seconds since the epoch in game time, used to sort when loading old style
+---@field protected entry string User entered text associated with the entry
+---@field protected always_custom boolean|nil Is this always treated as a custom entry (so not auto deleted)
+FlightLogEntry.Base = utils.class("FlightLogEntry")
+
+---@return string Description of this type
+function FlightLogEntry.Base:GetType()
+	local full_type = self.Class().meta.class
+	return string.sub( full_type, #"FlightLogEntry."+1, #full_type )
+end
```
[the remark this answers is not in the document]
JonBooth78: Yes, I really didn't like this either; I was hoping to find some way to 'automagically' solve the problem.

In the end I've gone for overloaded methods each returning a constant, only because I'm uncomfortable with my inability to make a variable const and this makes it harder (although far from impossible) to accidentally break.

data/modules/FlightLog/FlightLogEntries.lua
```diff
@@ -0,0 +1,397 @@
...
+
+
+---@return boolean True if this log entry should be considered custom and so not auto deleted
+function FlightLogEntry.Base:IsCustom()
+	if self.always_custom then return true end
+
+	if not self.entry then return false end
+	if self.entry == "" then return false end
+	return true
```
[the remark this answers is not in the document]
JonBooth78: Nice, I think that's pretty neat.  The lack of parenthesis kind of worries me but the ordering I see from the above is very natural.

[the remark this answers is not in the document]
Web-eWorks: The rule of thumb is: `and` has higher precedence than `or`, and comparison operators (e.g. `==`) have higher precedence than both. You don't have to worry about left/right associativity either in 99% of cases, but if you do the Lua manual has a clear section on operator precedence and associativity. :wink:

data/modules/FlightLog/FlightLogEntries.lua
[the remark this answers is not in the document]
Web-eWorks: That's a good argument for making this a function (in addition to the one of maintaining a consistent/uniform API) - I find in practice overwriting class-level variables isn't a significant concern especially with static analysis assistance from a language server, but it's an insignificant performance overhead so safer is better as long as productivity doesn't suffer.

data/pigui/modules/info-view/06-flightlog.lua
```diff
@@ -109,119 +136,92 @@ local function renderCustomLog()
...
-local function displayLog(logFn)
+
+local function displayFilterOptions()
+	ui.spacing()
+	local c
+	local flight_log = true;
+
+	c,include_custom_log = ui.checkbox(l.LOG_CUSTOM, include_custom_log)
+	c,include_station_log = ui.checkbox(l.LOG_STATION, include_station_log)
+	c,include_system_log = ui.checkbox(l.LOG_SYSTEM, include_system_log)
```
Web-eWorks, line 172: Note: this could be simplified to:
```suggestion
	c,includedSet.Custom = ui.checkbox(l.LOG_CUSTOM, includedSet.Custom)
	c,includedSet.Station = ui.checkbox(l.LOG_STATION, includedSet.Station)
	c,includedSet.System = ui.checkbox(l.LOG_SYSTEM, includedSet.System)
```

This would remove the need for the `getIncludedSet` function - simply define an `includedSet = { System = true, ... }` table instead of the `include_*_log` variables currently defined.

Web-eWorks (approved):
Looks good, thanks for addressing the last round of review comments. I've gone over the diff again and only one additional area has stood out to me as something that could be improved.

Before merge, I'd like you to address the `table[key, value] -> table<key, value>` annotation issues (they're not high severity but they are very quick to fix) and you indicated prior you intended to address the NaturalDocs formatting - otherwise, the remaining outstanding issues are an "at your discretion" priority.

I plan to merge this PR sometime Sunday evening/night EST - hopefully that gives you enough time to address the remaining issues. If not, let me know any time and I'll merge sometime next week.

data/modules/FlightLog/FlightLog.lua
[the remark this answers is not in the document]
JonBooth78: OK, this should also be sorted out now

Web-eWorks:
Looks good to me, thanks for addressing that formatting issue. At this point I'd recommend rebasing and squashing together related changes / iterative improvement commits to clean up the history (in the name of appeasing @impaktor, the resident git fascist :smile:).

JonBooth78:
@Web-eWorks following up on your recommendation to tidy up counting elements and I found a bug and another issue I'm tidying. Hope to have it done in the next 3-4 hours

Web-eWorks:
@JonBooth78 bump - ETA on final squash + rebase? I'd like to merge this before #5706.

Web-eWorks:
@JonBooth78 started a new game from the latest commit, initial entry is missing its custom description.

image

Web-eWorks:
Fixed - it was an errant colon in the function declaration for `FlightLog.MakeCustomEntry` :smile:

JonBooth78:
Thanks!

PR #5669, opened 2023-11-21
Improve BVH Broadphase Performance

Web-eWorks:
This PR refactors our CollisionSpace broadphase collision acceleration structure to be significantly more efficient. With no intersections (the usual case for the root frame of a system), the broadphase queries run approximately 3.5x to 4x faster.

I no longer have the performance numbers for this branch due to how many intervening PRs I've needed to merge, but the broadphase cost wasn't a significant performance sink to start with (on the order of 0.3-0.5ms / frame in the Sol system) so don't expect a huge FPS boost - this is mostly attacking low-hanging fruit which would have been a greater problem further down the line.

I've not yet tried to apply this optimization to our narrow-phase BVHTree implementation - I have a different idea which involves using single-precision floats and splitting large Geoms as part of the import step to fit within the precision limits of a regular `float`-based AABB.

The performance gain from this PR is mostly due to completely removing the need for a `std::list<Geom *>` during iteration, reducing the amount of allocations needed per-frame (can re-use the array memory if not significantly increasing the number of geoms or altering the topology), and cramming everything needed for a node into a single cache line.

The (rather abysmal) prior cache inefficiency was the largest factor responsible for the time cost involved in rebuilding and traversing the broadphase BVH, though once bodies start overlapping the narrow-phase collision tri-edge-edge-tri test still significantly dominates the runtime cost.

CC @JonBooth78 to have a gander at the physics code and hopefully improve your familiarity with that area of the codebase. :smile:

### DIFF
--- a/src/Aabb.h
+++ b/src/Aabb.h
@@ -15,14 +15,16 @@ struct Aabb {
 		min(DBL_MAX, DBL_MAX, DBL_MAX),
 		max(-DBL_MAX, -DBL_MAX, -DBL_MAX),
 		radius(0.1)
-	{}
+	{
+	}
 	// Fast constructor for pre-conditioned values
 	// Should only be used when _min < _max for all {x,y,z}
 	Aabb(const vector3d &_min, const vector3d &_max, float rad) :
 		min(_min),
 		max(_max),
 		radius(rad)
-	{}
+	{
+	}
 	void Update(const Aabb &b)
 	{
 		max.x = std::max(max.x, b.max.x);
@@ -69,7 +71,7 @@ struct Aabb {
 };
 
 // SIMD-capable AABB type without additional overhead
-template<typename Vec3>
+template <typename Vec3>
 struct AABB {
 	using Number = typename Vec3::element_type;
 
@@ -100,8 +102,8 @@ struct AABB {
 
 	auto IntersectsRay(const Vec3 &start, const Vec3 &inv_dir, Number dist) const
 	{
-		using std::min;
 		using std::max;
+		using std::min;
 
 		Vec3 l1 = (this->min - start) * inv_dir;
 		Vec3 l2 = (this->max - start) * inv_dir;
--- a/src/Aabb.h
+++ b/src/Aabb.h
@@ -14,6 +14,24 @@ struct Aabb {
 		max(-DBL_MAX, -DBL_MAX, -DBL_MAX),
 		radius(0.1)
 	{}
+	// Fast constructor for pre-conditioned values
+	// Should only be used when _min < _max for all {x,y,z}
+	Aabb(const vector3d &_min, const vector3d &_max, float rad) :
+		min(_min),
+		max(_max),
+		radius(rad)
+	{}
+	void Update(const Aabb &b)
+	{
+		max.x = std::max(max.x, b.max.x);
+		max.y = std::max(max.y, b.max.y);
+		max.z = std::max(max.z, b.max.z);
+		min.x = std::min(min.x, b.min.x);
+		min.y = std::min(min.y, b.min.y);
+		min.z = std::min(min.z, b.min.z);
+		radius = std::max(radius, b.radius);
+	}
+
 	void Update(const vector3d &p)
 	{
 		if (max.x < p.x) max.x = p.x;
@@ -42,9 +60,7 @@ struct Aabb {
 	}
 	bool Intersects(const Aabb &o) const
 	{
-		return (min.x < o.max.x) && (max.x > o.min.x) &&
-			(min.y < o.max.y) && (max.y > o.min.y) &&
-			(min.z < o.max.z) && (max.z > o.min.z);
+		return min < o.max && max > o.min;
 	}
 	// returns maximum point radius, usually smaller than max radius of bounding box
 	double GetRadius() const { return radius; }
--- a/src/Aabb.h
+++ b/src/Aabb.h
@@ -6,6 +6,8 @@
 
 #include "vector3.h"
 
+#include <algorithm> // for std::min / std::max
+
 struct Aabb {
 	vector3d min, max;
 	double radius;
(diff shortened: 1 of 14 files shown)


src/Aabb.h
```diff
@@ -42,12 +62,71 @@ struct Aabb {
...
+
+		double lmin = max(vmin.x, max(vmin.y, vmin.z));
+		double lmax = min(vmax.x, min(vmax.y, vmax.z));
+
+		return ((lmax >= 0.f) & (lmax >= lmin) & (lmin < dist));
+	}
+
+	auto IsIn(Vec3 point) const { return point >= min && point <= max; }
+	auto Intersects(const AABB &rhs) const { return min < rhs.max && max > rhs.min; }
+};
```
JonBooth78, line 128: This type confuses me a little.  We have some things using a `min()` `max()` method that takes two vectors (inside `IntersectsRay`) and other things doing component wise min and max.

Then our `Invalid` method seems to have determined that we're using doubles as our component type for our vectors and yet `0.f` inside our `IntersectsRay` method suggests floats.

Is the intention to proved a SIMD optimized variant of Vector3 that has correct alignment constraints and operator overload functionality?

src/collider/BVHTree.cpp
```diff
@@ -158,3 +160,252 @@ BVHNode *BVHTree::AllocNode()
...
+		m_nodeStack.pop_back();
+
+		if (!node->IsLeaf() && node->kids[0])
+			m_nodeStack.push_back(node->kids[0]);
+		if (!node->IsLeaf() && node->kids[1])
+			m_nodeStack.push_back(node->kids[1]);
+
+		// surface area = 2 * width * depth * height
+		vector3d size = node->aabb.max - node->aabb.min;
+		double area = 2.f * size.x * size.y * size.z;
```
JonBooth78, line 181: Odd that we have a `2.f` being multiplied by a double here, would make more sense for it to just be `2.0`

src/collider/BVHTree.cpp
```diff
@@ -158,3 +160,252 @@ BVHNode *BVHTree::AllocNode()
...
+
+		// Cost function according to https://users.aalto.fi/~laines9/publications/aila2013hpg_paper.pdf Eq. 1
+		if (node->IsLeaf())
+			outSAH += node->numTris * area;
+		else
+			outSAH += 1.2 * area;
+	}
+
+	vector3d rootSize = m_root->aabb.max - m_root->aabb.min;
+	float rootArea = 2.f * rootSize.x * rootSize.y * rootSize.z;
```
JonBooth78, line 191: This calculates a float but does all the calculations along the way in double precision bar the final multiplication by `2.f` (I think).  Is this intended?  I no-longer know on modern hardware what the cost is vs the cost of say casting the min/max bb to float and then doing it.  If you did that you could also cast them to SIMD aligned float vectors for some speed increases, might even find that that if our SIMD float vector 3 has an implicit `w` of 1.0f always the last operation here can be 4 component dot product and even faster.

src/collider/BVHTree.cpp
```diff
@@ -158,3 +160,252 @@ BVHNode *BVHTree::AllocNode()
...
+
+	// Perform 1 / Aroot * ( SAH sums )
+	outSAH /= rootArea;
```
JonBooth78, line 194: So we just converted that all to a float (prev comment) and then are using it as the denominator for double maths?

src/collider/BVHTree.cpp
```diff
@@ -158,3 +160,252 @@ BVHNode *BVHTree::AllocNode()
...
+
+	// Remove the (normalized) SAH cost of the root node from the result
+	// The SAH metric doesn't include the cost of the root node
+	outSAH -= 1.f;
```
JonBooth78, line 198: Again are we floating point or double?

src/collider/BVHTree.cpp
```diff
@@ -158,3 +160,252 @@ BVHNode *BVHTree::AllocNode()
...
+
+	vector3d rootSize = rootNode->aabb.max - rootNode->aabb.min;
+	float rootArea = 2.f * rootSize.x * rootSize.y * rootSize.z;
+
+	// Perform 1 / Aroot * ( SAH sums )
+	outSAH /= rootArea;
+
+	// Remove the (normalized) SAH cost of the root node from the result
+	// The SAH metric doesn't include the cost of the root node
+	outSAH -= 1.f;
```
JonBooth78, line 409: Similar oddness around float/double dichotomy here as above.

I also find it a little odd that our Keys and pivots are floating point vectors when we're dealing in double space too, although probably just fine,

src/Aabb.h
Web-eWorks: The intention for this specific class is to provide a templated, optimized version of the existing Axis-Aligned-Bounding-Box implementation (`Aabb`) which can operate semi-generically on double-precision, single-precision, or packed SIMD vectors, by changing the `Vec3` type provided. E.g. `AABB<vector3<xsimd::batch<float, AVX2>>>` (not exact code) would be an AABB storing eight single-precision float elements "per component" with no extra code compared to the scalar version - it's the responsibility of the calling code to unpack individual elements back to scalar datatypes if necessary.

> Then our Invalid method seems to have determined that we're using doubles as our component type for our vectors and yet 0.f inside our IntersectsRay method suggests floats.

This is a copy-paste error - it should be `Number(0)` instead.

> We have some things using a min() max() method that takes two vectors (inside IntersectsRay) and other things doing component wise min and max.

The min/max on vectors is just a nicer interface compared to manually writing `std::min(a.x, b.x)` etc. everywhere the operation is performed. It uses the same semantics as packed SIMD min/max operations, so it operates per-element and per-component. The component-wise min/max you're seeing in `IntersectsRay` is a "horizontal min/max" (taking the minimum of all elements in a *single* vector) and that's the most optimized way to express it.

...that being said, I see now what you're talking about - I forgot to update the other component-wise calls when I added the min/max overload for our `vector3` type. I also forgot to make the min/max overloads generic enough to use with a vectorized `element_type`. Will fix.

src/collider/BVHTree.cpp
Web-eWorks: Copy-paste / refactoring error, thanks!

src/collider/BVHTree.cpp
Web-eWorks: The surface-area-heuristic calculation function was only implemented for debugging purposes (I used it to validate the correctness of the replacement BVHTree implementation) and was banged together quite quickly. There's no rhyme or reason behind the datatype usage here, other than that it was originally written with `float` and the return value of the function changed afterwards. Will at the least make it consistent and use `double` the whole time.

JonBooth78:
Thanks @Web-eWorks , interesting happy to take a look.  Overall looks very sensible.  There are a few odd places where we do float/double conversions (there was one or two more in `CollisionSpace.cpp` that I'm not sure about in terms of correctness, or if it matters.

I can't see the problem that I've hit when predominantly working in float space and using double literal.  e.g, `float x = y * 2` where y is also a float, will actually promote y to a double, multiply it by 2 and then cast back to a float, massively slower than `float x = y * 2.f`, which is probably what was intended. Or at least this is how compilers used to behave.

src/collider/BVHTree.cpp
Web-eWorks: The keys/pivots are actually completely intended to be in single-precision compared to the actual datatypes - `SingleBVHTree::Build` is designed to try to squeeze out as much precision as possible from the mantissa by normalizing all coordinates over the total bounding volume.

This decision was made because the precision of the sort key/pivot doesn't affect the correctness of computing intersections in the final tree (that's all done in double-precision), but the cache efficiency of the sort step is very important for tree re-build performance as it's an `O(n log n)` traversal of the whole list of entries in the tree.

Web-eWorks:
> I can't see the problem that I've hit when predominantly working in float space and using double literal. e.g, float x = y * 2 where y is also a float, will actually promote y to a double, multiply it by 2 and then cast back to a float, massively slower than float x = y * 2.f, which is probably what was intended. Or at least this is how compilers used to behave.

I'm pretty sure this is how compilers still behave - automatic promotion to larger datatypes is still a thing. I believe the compiler will promote a `2.f` at compile time to a double when working with other double-precision datatypes so there's no speed loss there, but in single-precision code I believe "correctness" requires that speed loss. Will keep a sharper eye out for those kind of issues, thanks!

src/Aabb.h
JonBooth78: Ahh so there is a min/max that returns a scaler of the min/max of a single vector's components and then component-wise min/max too.

I believe most SIMD implementations should accelerate at least the component wise ones to make them single instructions, I'm not sure if we write them component wise though, the compiler will figure that out for us.

src/collider/BVHTree.cpp
JonBooth78: Yeah, I had a feeling this was totally deliberate, interesting.

Ahh yes the whole optimize the precision of the mantissa, been ages since I thought about that - I kind of assumed that for doubles it wouldn't matter so much.  (Last time I remember thinking about this I think was when doing 7e3 math for color channels for XBox360 HDR).

JonBooth78:
I once saw more than an extra 5% framerate on (I think it was a PS2), by the result of sprinkling around some `f`s to make things floating point literals and not do double math like that.  As I recall there was a few right in tight loops of the graphics and collision culling (so very similar areas to this).  So I'm now super pedantic about it :-D

PR #5670, opened 2023-11-22
Improve CI Setup and Performance

Web-eWorks:
This PR refactors some of our CI build scripts and CMake tooling files to support using the `Ninja` generator on Linux instead of Make, which allows us to automatically invoke parallel builds everywhere including our CI runners.

As a result, the CI build should be able to take advantage of all vCores provided to the execution environment, leading to a hopeful build time improvement.

I've also fixed the `clang-format` CI failures by upgrading the runner to a version that provides a newer `clang-format` binary - we were stuck on v11 for CI while developers are on significantly newer versions (my system has v16) and the config files are *not* backwards compatible or well-versioned at all.

For compatibility with VSCode, I've bumped the `CMakePresets.json` version number to v3 - this requires CMake 21 or newer. Older CMake versions will still be able to use the "manual" build workflow on Linux, though CMake is starting to warn about dropping support for the minimum CMake version specified in a lot of our actual `CMakeLists.txt` files so we should re-evaluate those at some point.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -28,6 +28,7 @@ env:
     libassimp-dev
     libsdl2-dev
     libsdl2-image-dev
+    ninja-build
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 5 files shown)


Web-eWorks:
Going to merge even though there's a clang-format failure and run a codebase wide "re-sync" clang-format later on `master` branch once I've merged most of the outstanding PRs to minimize the amount of workflow disruption.

PR #5674, opened 2023-11-26
Delete button for saveloadwindow, make this window less error prone (and some more)

Max5377:
Adds delete button for `saveloadwindow` to allow user to delete any save in `user\\Documents\\Pioneer\\savefiles\\` folder.
For this, adds new static method `Game::DeleteSave(const std::string& filename)` and two new methods in `FileSystemFS` class:
1. `FileSystemFS::RemoveFile(const std::string& relativePath)`, to remove file in the `FileSystemFS::m_root` + `relativePath`. Deletion is restricted to only delete regular files;
2. `FileSystemFS::IsChildOfRoot(const std::string& path)` to check if given path is in the `FileSystemFS::m_root`.
Makes `saveloadwindow` less error prone, especially when trying to `save`/`load`. Will show `MessageBox`  with a text about why `save`/`load` failed.
Before saving will check filename with new `FileSystem::IsValidFilename(const std::string& filename)` if `filename` has forbidden `OS` characters.
Other changes include optimization for `Game::CanLoadGame` (just tries to open file to check it's existence) and `Game::SaveGame` (tries to open file first, if failed, immediatlly throw exception), new `MessageBox` type and little refactoring for `LuaGame.cpp`.
Demonstration (Linux VM, similar functionality in Windows):

https://github.com/pioneerspacesim/pioneer/assets/69468517/5cd6b580-4178-4bdc-bd92-6e7f64f64655

**Update**: 
1. `saveloadwindow` now shows `MessageBox` with confirmation if user really wants to delete save. Next it will show another `MessageBox`, that tells if save deleted successfully or not;
2. Some optimization changes for `FileSystemFS::IsChildOfRoot(const std::string& path)`.


nozmajner:
Nice!
It might be useful to have some kind of confirmation for deleting to avoid any accidental deletes I think.

Max5377:
@nozmajner It has, just not used for now. I want to show `MessageBox` with the error if something goes wrong, but if I show `MessageBox` with confirmation first, `MessageBox` with the error will be just not shown. Added comment in `saveloadwindow.lua` about it:
```
-- XXX Not used for now, because lui.COULD_NOT_DELETE_SAVE has no time to show itself
-- do to creation of it on top of lui.DELETE_SAVE_CONFIRMATION which deletes itself
-- by the time of showing
local function showDeleteConfirmation()
```

Max5377:
@Web-eWorks 

https://github.com/pioneerspacesim/pioneer/assets/69468517/16e3800a-8d84-45b6-8b1b-ff2d8381c407

Web-eWorks:
Looks good! I'd recommend using the next smaller font size breakpoint (with `ui.withFont(<font>, function() ... end)`) and adding some spacing between the text and the Load buttons, but otherwise I think it will work.

Max5377:
@Web-eWorks 
Changed font size to small and added padding between `Load` button and `Save deleted successfully`.
Also added new pigui lib `ui-timer` to create timers that work from internal `ImGui` timer. For example, can be used to show something for a limited amount of time (like with `Save deleted successfully`).
Don't really know if it's good for a solution, but will change it if needed.

data/pigui/libs/ui-timer.lua
```diff
@@ -0,0 +1,35 @@
+local ui = require 'pigui'
```
impaktor, line 1: copyright header.

Max5377: Forgot most important part...

Web-eWorks (approved):
Finally reviewing this now that I have the time available, thank you for the patience.

Code looks good, the new methods in FileSystemPosix/FileSystemWin32 could probably receive another pass at some point for optimization and cleanup (for example, the forbidden characters list on unix is definitely missing a few characters), but overall I think it's fine.

Tested on Linux, if anyone wants to give this a test on Windows before I merge, let me know in the next 24 hours or so.

PR #5676, opened 2023-12-08
Re-add the galaxy dump function

hexagonrecursion:
97828a9ed634b6cb889867e48241df1072e6b810 removed the ability to dump the galaxy data. I needed the data for statistical analysis for balance purposes. Debugging missing global dependencies was quite frustrating, but now we can dump galaxies again.

Note: dumping the galaxy opens and closes a gui window. I am not sure why this happens, but any galaxy dump is better than no galaxy dump.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -9,6 +9,9 @@
 #include "galaxy/GalaxyGenerator.h"
 #include "utils.h"
 #include "versioningInfo.h"
+#include "lua/Lua.h"
+#include "lua/LuaNameGen.h"
+#include "lua/LuaObject.h"
 
 #include <SDL.h>
 #include <cstdio>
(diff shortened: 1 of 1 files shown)


Gliese852:
There is also a [GalaxyStats](https://codedoc.pioneerspacesim.net/#File:src/lua/LuaDev.cpp:GalaxyStats) function.

Gliese852:
You can choose any range, also, everything beyond 90 sectors from the Sol is unexplored.

hexagonrecursion:
The problem is not the range. GalaxyStats appears to only print the summary of the data. It only prints the total population of the selected area. I want the population of each system so that I can make a histogram.

image

Gliese852:
Ah, I understand now. I’ll just note that it’s quite easy to add or modify a "[Processor](https://github.com/pioneerspacesim/pioneer/blob/8d54d501134520eec1ed56810eaacd8ae4b58bad/src/lua/LuaDev.cpp#L101)" that will generate data for the histogram.

impaktor:
So in summary, is this the direction / proper way to implement this? As I see it, it resurrects code we used to have, right?
So this is OK to merge?

hexagonrecursion:
> So in summary, is this the direction / proper way to implement this? As I see it, it resurrects code we used to have, right? So this is OK to merge?

This does indeed resurrect the code we used to have. Most of the code was already there - in the `galaxy->Dump()` method. Only the code in main.cpp was missing.

Web-eWorks (approved):
I'm not very enthused about pulling in Lua implementation details into `main.cpp` but it's not something that cannot be fixed with a later move of the galaxy dump code into e.g. the System Editor.

PR #5686, opened 2023-12-21
Fix installed directory structure on Windows

Web-eWorks:
During the last major change to CMake I accidentally broke the install structure on Windows. This PR resolves that and should make the installation behavior on Windows systems more robust and easier to understand.

I've also included a potential fix for MSVC doing strange things with unicode character escape sequences in C string literals by explicitly specifying that we want the UTF-8 encoding of the unicode character.

CC @nozmajner @pcercuei to give this a test on Windows if able.


pcercuei:
What I see if that you have a install() in the top CMake script, why isn't it in the cmake/InstallPioneer.cmake, below the spot where CMAKE_INSTALL_BINDIR is overriden?

nozmajner:
@Web-eWorks 
The binaries are now put in their proper place, but the data is in a pioneer directory:
image
I've tried this artifact: https://github.com/pioneerspacesim/pioneer/actions/runs/7293314137

Web-eWorks:
That should also be fixed now.

nozmajner:
Confirmed, it works properly now.

PR #5692, opened 2023-12-26
Avoid misaligned read when loading SGM files

Web-eWorks:
Amend the Scenegraph serialization interface to more correctly handle the unaligned reads required by the SGM data structure layout. While the unaligned reads are in practice not an issue, this provides some future proofing against future compiler "improvements" (read: backwards compatibility breakages) and means the code is more portable to architectures like ARM64 which have a hardware requirement of aligned reads.

Closes #5687.

PR #5694, opened 2023-12-26
Fix undefined behaviour in CommandBufferGL.cpp

hexagonrecursion:
By compiling with sanitisers enabled I have detected a number of places where we run into undefined behaviour. The patch for the first is in #5692 (thanks to Web-eWorks). This is a patch for the second issue I detected.

1. `mat->m_pushConstants.get()` is sometimes null.
2. I set a breakpoint in a debugger and it turned out that  `s->GetConstantStorageSize() == 0`
3. [The behaviour of memcpy(dst, src, count) is undefined if either of the pointers is null **even if count == zero**](https://en.cppreference.com/w/cpp/string/byte/memcpy)

Fix: add a null check.

Here is how I compile with sanitisers.
1 Make sure `build/` does not exist (because cmake does not update CXXFLAGS for build directories that have already been populated). Alternatively: build in `build-with-sanitizers/`
2. `CXXFLAGS="-fsanitize=address,pointer-compare,pointer-subtract,undefined" LDFLAGS="$CXXFLAGS" cmake -S . -B build/ && make -j2 -C build/`

### DIFF
--- a/src/graphics/opengl/CommandBufferGL.cpp
+++ b/src/graphics/opengl/CommandBufferGL.cpp
@@ -214,7 +214,9 @@ char *CommandList::SetupMaterialData(OGL::Material *mat)
 	const Shader *s = mat->GetShader();
 
 	char *alloc = AllocDrawData(s);
-	memcpy(alloc, mat->m_pushConstants.get(), s->GetConstantStorageSize());
+	if (mat->m_pushConstants) {
+		memcpy(alloc, mat->m_pushConstants.get(), s->GetConstantStorageSize());
+	}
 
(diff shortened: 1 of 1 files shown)


Web-eWorks (approved):

Web-eWorks:
Apologies for the delay on reviewing these PRs - thanks for digging into it! I'm always glad to see someone take the time to validate the behavior of the compiled code.

impaktor:
@hexagonrecursion Thanks a bunch for the work you've done. Please let us know if you want to be in [AUTHORS.txt](https://github.com/pioneerspacesim/pioneer/blob/master/AUTHORS.txt), and what to put.

Issue #4496, opened 2018-11-26
Prepare a new release and document the process


laarmen:
It's been a while since we've had a new release, and that's a shame. However, our build system has evolved quite a bit, and I've basically been told that we don't know how to do releases anymore.

The process might simply be "tag a commit, wait for defined $people to build on their respective platforms, upload the builds, then do the whole website/twitter/stuff dance", but it still has to be documented.

Oh and I'm not volunteering.

fluffyfreak:
Few reasons we haven't done this, I've been waiting on #4427 to be finished but @pcercuei seems to have been away a lot. It's a real shame as I think it'll be a great addition to the build process.

The CMake stuff has taken a while to work but I think that's now ready. I haven't been following along too closely since I want it to work on Linux (_and OSX?_) before I even try to use it with Visual Studio.

I've got a branch of #4427 and have just tried to update it to master, so I'll give that a go and see if it creates a usable windows package after merging latest into it.

pcercuei:
I'm still here, just travelling around without a computer. I will be back next week.

fluffyfreak:
I have a fork of the branch with latest code, some hacking around in it but it does produce an installer now. Hopefully I'll test tomorrow.

Hope you're having fun @pcercuei, I look forward to you getting back us getting this integrated :)

impaktor:
Do remind me:

1. OSX builds still depend on @salborrelli or @Philbywhizz building and uploading, right?

2. And if we use appveyor for windows builds and github for GNU/Linux build, only OSX will be hosted on sourceforge?

fluffyfreak:
Sort of:

1. OSX has been @Philbywhizz the last few times.
2. Appveyor we'd only use for building not as host, it auto deletes after 6 months and there's a 1GB size limit. Each build is ~330MB for Windows.

pcercuei:
@fluffyfreak: this is with #4427 rebased on top of master?

fluffyfreak:
@pcercuei that's with my own branch which has some other hacking in it. So rebasing atop #4427 might be better

abbottjord94:
Genuine noob question: If I'm able to build from source on a given platform and pack it into a .7z file (and document the whole process), is that acceptable? 

I ask because I can build for Windows and Linux (without CMake though), and I'd be happy to provide both of them if that's what we're looking for.

impaktor:
[...]

PR #5698, opened 2023-12-27
Fix bugs in fixed division

### LINKED ISSUE #5697: Bugs in fixed.h division
Pioneer has its own fixed-precision arifmetic implementation. The code can be found in [src/fixed.h](https://github.com/pioneerspacesim/pioneer/blob/97d7be673f2124f1b42aee956c4b3a6fc2129558/src/fixed.h)
Usually we use 32 bits before the "decimal point" and 32 bits after, but StarSystemGenerator uses 16 bits before and 48 after.

The operator/ in fixed.h has multiple bugs.

## Using C++20 rules

1. This is undefined behaviour (signed integer overflow) if d is INT64_MIN

       if (d < 0) {
           d = -d;
           isneg = 1;
       }

2. This is undefined behaviour (signed integer overflow) if isneg is true and Sint64(quotient_lo) is INT64_MIN

       return (isneg ? -Sint64(quotient_lo) : quotient_lo);

   Here is what I find ironic:

   1. Assuming I correctly understand the rules for the ternary operator with mixed types (a very optimistic assumption),
      the code above is equivalent to

          return (isneg ? Uint64(-Sint64(quotient_lo)) : quotient_lo);

   2. Assuming I correctly understand what the unary minus does for unsigned types,
      the only difference between `Uint64(-Sint64(quotient_lo))` and `-quotient_lo` is that
      the former trigger undefined behaviour if quotient_lo is exactly `1 << 63`

3. `remainder` can become negative, causing incorrect results. Here is one example, but any value of b such that `abs(b.v) > (1 << 62)` can be problematic.
 
   1. Let FRAC be 32
   2. Let a be `0x4000'0000.0000'0000` and b be `0x4000'0000.0000'0001` (Note the "hexadecimal point")
   3. The expected result is `0x0000'0000.ffff'ffff`
   4. The actual result is 0

   [Here is a proof of concept on godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM1ykrgAyeAyYAHI%2BAEaYxCAA7ACcpAAOqAqETgwe3r7%2BgemZjgKh4VEssfHJtpj2JQxCBEzEBLk%2BfgG19dlNLQRlkTFxiSkKza3t%2BV3j/YMVVaMAlLaoXsTI7BzmAMxhyN5YANQmO27I4/iCp9gmGgCCu/uHmCdnTiyGeKk3d49mewYBy8x1Obic42ImFYvwefwIAE9UpgsFQjmECAA2SQAfQIRyEGOxpysD0RyNRRy8RNx%2BLkNJJfz%2BXkyRiOlxAKDWBEZDxZYWA7II6E5rlovMeDzQDHG6MERwAYgAle5uN4AESOOzMEr%2BADdUHh0Ed0D5UhBCYJsUdVEsTglSfcjkd6VbJDbVBqbRLnXg0RBPWCjho7SYHX9nc60F58WCg%2BYzEcALQJn2R1SBnaapMZtNhzV1JT2x2Ro7R2NneNmRMnat5hLqiNl7lvNxBjkgJQEADuEC4AA5Q5Wzh7dQ2mQ9LVj3fg9RaaUcmKQCQvoqHww9nQB6LdHLhmftJ6KEE14PVGlFHaIIo7Yo%2BEZdEI6fADWrwU61eCDwwAQV8ICiLlCRwMKg%2BK0BkPKbiubpHAAjl4YF4IwBA4t%2BXpMG8tw7NgRwQNaSaKiqbhLGmrrTvBiGOChOIQRhrZBsqqpplO1rGqcmrRGmGLogo4SChxwYsQuUKfGEWDEF6Gi6tBfp4exI4hsWTbOgp2boGmvp8ZgAlZnu9aNnC0HlgxI4JmpqY7CWJpmhA6CkVZTYmXGI5ijJTrNjGplqgmolMOJcQapZ1mmiw5p%2BQFxAOdZznDj51aUUhNHoRxwVNqF5oIUlghoXg0VOS2LnxYmWXUTldGpXWjnQRlEClch5WoPlxmFXFRxudVHn8JJEA8XgUkkuirZ7geg14NY1jrtZzrkdaCjHhWmoQLNkh9kObYjpiOyhmYmKJWVqHfppRwRQwEneYJXDHXJEDzYQpHQaWp3nWGG2aldnWlju7LflQ%2BL1ShRz0H9%2BmPcpYPOgDOUpXFl3Hb6/pQ6hFW7XhK1rd5t7bQ9HmlpDVENYd/WvXDn140jtGoBdekfdN9qGbj6L%2Bs9gW/Hp9ng4zzos5JKbs/D%2B2E5T9pvfpZPOvmE6M7FG3FTzQVVXTtU8810utbLtYlQTyXE1maVg7VFPfqrpYy1WWvZcjVOVTq4s2WFdXa41JuRmbrlneKZOS9BUIEOsDB4Xg2m6QqyasatFMQXaICCzRUdjgzfw8WJDAQFNBVeUVQoiiACCYJmGsdgoCCoN20RMEowmwVCH60Itp5zhoqiSBordt%2B3HfLk3Lcd73rdcC7bty5gtf1wmmM114deYx11m%2B/7QmdfmHArLQnAAKy8H4HBaKQqCcG2ljWOyawbK8uw8KQBCaCvKwviAkiYgAdEkXAaEkSSSAk2I7EkyQaOvfQnBJC8BYBIVupBt6733hwXgCgQAaCvjfFYcBYAwEQCAfOmBkAxhIOQSgLRgAKGUIYOoQgS7dm3pfNAYU6BMAaCQ8ItByGlygbwGhqQ6AjGAFwbUpAOFcOIBEVgWx2GoFofQYgAB5GMLDKE3yCKobB9xiBEM4LwLByAmhXHUTIQQIgxDsCkHo%2BQSg1AKN0IEAwRgUATUsPoPA0R4GQBWKgVIDR4EcCTJcDipgj6WDMNAvUcRiAXmcWnUgxAvCCDwGwAAKqgTw4SVgKFPpsPQlwwiMLIRQqhvACBQnYIg7sxAmCpE4DwVeG8t4KJgdgJROCiCSVUP2TESZrTAGQMgPcOwn6JggIfKw9iji4EICQWsOwB75OQXfEA69ekaH7EkVu68ATr0xBoQJSRAFrw4CA0gYDMT9ifuvVpiylm/3WZiBIgQ2F710fAxB18d4r1IKgjBAjJH4IgJ87hvCzB8DoAQOI8CIDRAUceZgxAEQVN4JCloCIpHRG0Ng551DxFsEEFIhgtAYUvNIFgaIXhgBuDELQTxl8sCfCMOIfF%2BAoQODPCPBRmiYxbEvhiOoCjaCONKdCjwWAFEFNibClYVADBEIAGrIW7FI5EeSTEGPEMY/gghFAqHUPiyx%2BhDDGDsTYHlTj4CuPcdkTx3jhS%2BP1YE3gqBgnEFCVgZJ3RUXZBcGdKYfhAghDCEMSoIxCgZCyAIT1egijBoYPMYY8RAh2FdQIPokxPAdAyXUeNjQJgDF9QsANthM2htjZmqN/qY0pLSUYqpHBN6QNqZwG0rT2nuk6d03hfS8KjKaRMqZSCXlLBWHnJgElKCVv2WAyZT9JC8Nfm/DZkgdjaiSJiGt%2BKYFwIQT2rQfbSD33mU/RZyyAFrI2VsnZnAdg1JXbo55m7K1mAvdAq9MzSD2syM4SQQA)

       d =  0x4000000000000001
       remainder =                   0 quotient_hi =          0x40000000 quotient_lo =                   0
       remainder =                   0 quotient_hi =          0x80000000 quotient_lo =                   0
       remainder =                   0 quotient_hi =         0x100000000 quotient_lo =                   0
       remainder =                   0 quotient_hi =         0x200000000 quotient_lo =                   0
       remainder =                   0 quotient_hi =         0x400000000 quotient_lo =                   0
       remainder =                   0 quotient_hi =         0x800000000 quotient_lo =                   0
       ...
       remainder =                   0 quotient_hi =  0x1000000000000000 quotient_lo =                   0
       remainder =                   0 quotient_hi =  0x2000000000000000 quotient_lo =                   0
       remainder =                   0 quotient_hi =  0x4000000000000000 quotient_lo =                   0
       remainder =                   0 quotient_hi = -0x8000000000000000 quotient_lo =                   0
       remainder =                 0x1 quotient_hi =                   0 quotient_lo =                   0
       remainder =                 0x2 quotient_hi =                   0 quotient_lo =                   0
       remainder =                 0x4 quotient_hi =                   0 quotient_lo =                   0
       remainder =                 0x8 quotient_hi =                   0 quotient_lo =                   0
       ...
       remainder =  0x1000000000000000 quotient_hi =                   0 quotient_lo =                   0
       remainder =  0x2000000000000000 quotient_hi =                   0 quotient_lo =                   0
       remainder =  0x4000000000000000 quotient_hi =                   0 quotient_lo =                   0
       remainder = -0x8000000000000000 quotient_hi =                   0 quotient_lo =                   0
       remainder =                   0 quotient_hi =                   0 quotient_lo =                   0
       ...
       remainder =                   0 quotient_hi =                   0 quotient_lo =                   0
       result = 0

4. Most combinarions of a negative `a` with any `b` produce incorrect results. It puzzles me greatly that no one noticed. Example:

   1. Let FRAC be 32
   2. Let a be -3.0 and b be 3.0
   3. The expected result is -1.0
   4. The actual result is `0x5555'5554.5555'5555`

   [Here is a proof of concept on godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEIAAcAMykAA6oCoRODB7evv6p6ZkCIWGRLDFxSbaY9o4CQgRMxAQ5Pn4BdpgOWQ1NBCUR0bEJyQqNza15HeP9oYPlw0kAlLaoXsTI7BzmiaHI3lgA1CaJbshj%2BIIn2CYaAII7eweYx6dOLIZ4Kde3D2a7DH2XiOJzcTjGxEwrB%2B91%2BBAAnilMFgqIdQgQAGySAD6BEOQnRWJOVnuCKRKMOXkJOLxcmpxN%2Bvy8GSMhwuIBQawIDPuzNCwDZBHQHNctB5D3uaAYYzRgkOADEAEp3NyvAAih0SZnFvwAbqg8OhDugfCkIATBFjDqolscAOwku6HZ2HAD0ACpfs66ZbJNbVOrreLnXhURAA6DDhpbSYHV6XWgvHjQZHzGZDgBaNPBl2qCOJDUZvM52MampKe2Ol2HRPJ06pszp46Nkt2tXx92u%2BO115uBvptO9yPskBKAgAdwgXHiMfrpyDiUdpcZ9wtmL9%2BF1hsw5uphyYpHxe6iMbj92drtdhy4ZniGaihGNeC3WCNUXhhyx98Ih6Ihw%2BADWLwKOsLwIHgwAIIcD4EAo%2B6QocDCoHitDpNy55Hr6hwAI5eMheCMAQ2LgYGTCvDciTYIcEBWhmCrKm4Sw5j6644XhjiEdiqGkUO85KiqOZrlaRonBqUQ5uiaIKGEAqiVGgl7pCHyhFgxCBhoOoYaG1EifO0aVvGzq6YW6A5iG0mYLJBbXq27awhhPYpvOabGdmi7xiaLBmugTHuQ5XK8aqoqaU6NYBU5qppkpTAqbE6puVWxqmhA0WxcQvmJY5c6RY2bH4ZxJGiQlHnJbh%2BWCMReAZd24XZc26ZlRxFXcUVLZ%2BaFnlmo1BHNag1X%2BUmgWHMF7XOvwakQJJeDqcSaK9tet6zXg1jWKeiXenuCgwYGEAsViU6zn284YokMZmBieVNUR4FmYcqUMKpgVyVwt3aRAW2EExGHVvdj2xkdGovaNLqXmy4FUHi3WEYc9AQzZ30GQjzpQxVhXZc9t0hmGKNES153UXtkgHUNJ1LF9oXVsj7E9dd03/RjwPVjjXGoE91lA%2Bt9p2RTaJhr9cU/NZPmIzzzr82pWZC5jl00yz9oAzZjNcyuPNZUdOVmOL8VtZznUpZgykPbE/Wq7V6v1TLBV0wWxUI3rzPgSb1Zq/2lu9dr2pK/b1OcahTsJmbkYjYly4YZCBDrAw1F4BZVnypmQlE8zfuHCAbu431OptirkmGxAa01YNEWCsKIAIJg%2BbmyOCgIKg45REwSgKVhkIgbQdYapu24QBmySav7LvOY2rdeO3gaDsXI9j8XwfxuHkfye1pYcCstCcAArLwfgcFopCoJwfaWNYbJrBsLw7DwpAEJoK8rABICSBiAB0ACcXAaC/L%2BSHaWKJC/dovw0OvfQnBJC8BYBIDQGhSDb13vvDgvAFAgGgdfHeK9SBwFgDARAIAK5dCTCQcglAmjAAUMoQwNQhC13HNvS%2BaAvJ0CYHUBg5Cwi0CoXXWBvB6EpDoMMYAXAtSkB4Xw4g4RWBbG4agBh9BiAAHkkwcJoTfQIqguh3GIKQzgvA8HIAaJcbRMhBAiDEOwKQRj5BKDUCo3QXB9CGGMCtSw%2Bg8BRCQZAFYqAUjMKQRwDMFxRKmCPpYMwcDdSxGINudx%2BdSDEC8IIPAbAAAqqBPDRJWAoU%2Bmw9AXFCKwyh1DaG8AIJCdg0DxzECYCkTgPBV4by3io%2BB2A1HIAIWpVQ8QMQZitMAZAyBryJCfumCAh8rDOMOLgQgJBmyJC4EsYpN8yakHvuvQZGh4iAKAf8deGINChJfsAteHAwGkAgRieIT916dPWRsv%2BOyMR2jsVwvehikEoMWRg7BEAkAiNkUQiAvz%2BGCICDQdusQkEQCiCoh8zBiDwhqbwGFTR4RyKiNoLoqC6HSLYIIORDBaDwrQaQLAUQvDADcGIWgvjL5YA%2BEYcQRL8CQm6OE3xu9dFJi2JfdENQVG0FcZUuFHgsAqJKYkhFKwqAGFIQANQIuOORSIikWJMeIcx/BBCKBUOoIltj7FGBQE4mw/K3HwE8d4rIvj/FCkCUa0JvBUDhOIJErA6Tqi1CyC4B6kw/B2OCHMMoFQ9BpAyMwn1wbCjMIGIG4YdjOjdHqDMcNcaagYp6DMaNQw4hxqTZ4NoOSM0BqzRIDJWSzF1I4JvGBjTODWk6d0v0vT%2BmCKGdRSZRA1IX3mVfRZKxy5MFUpQCtJyIGzKfpIQRb9367MkIkLUL8MTVqJfAxByCe1oKWSstZGyoHr22bs/ZhzOCJAacuwxqCtBLKOWYU9cDz29tIE6jIzhJBAA)

       d =                0x3
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffffffd00000000
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffffffa00000000
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffffff400000001
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffffffe800000002
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffffffd000000005
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffffffa00000000a
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffffff4000000015
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffffe800000002a
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffffd0000000055
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffffa00000000aa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffff40000000155
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffffe800000002aa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffffd00000000555
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffffa00000000aaa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffff400000001555
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffe800000002aaa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffd000000005555
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfffa00000000aaaa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfff4000000015555
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffe800000002aaaa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffd0000000055555
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xffa00000000aaaaa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xff40000000155555
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfe800000002aaaaa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfd00000000555555
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xfa00000000aaaaaa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xf400000001555555
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xe800000002aaaaaa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0xd000000005555555
       remainder =                0x1 quotient_hi = 0xffffffffffffffff quotient_lo = 0xa00000000aaaaaaa
       remainder =                  0 quotient_hi = 0xffffffffffffffff quotient_lo = 0x4000000015555555
       remainder =                0x1 quotient_hi = 0xfffffffffffffffe quotient_lo = 0x800000002aaaaaaa
       remainder =                  0 quotient_hi = 0xfffffffffffffffd quotient_lo =         0x55555555
       remainder =                0x1 quotient_hi = 0xfffffffffffffffa quotient_lo =         0xaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffffffffffff4 quotient_lo =        0x155555555
       remainder =                0x1 quotient_hi = 0xffffffffffffffe8 quotient_lo =        0x2aaaaaaaa
       remainder =                  0 quotient_hi = 0xffffffffffffffd0 quotient_lo =        0x555555555
       remainder =                0x1 quotient_hi = 0xffffffffffffffa0 quotient_lo =        0xaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffffffffffff40 quotient_lo =       0x1555555555
       remainder =                0x1 quotient_hi = 0xfffffffffffffe80 quotient_lo =       0x2aaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffffffffffd00 quotient_lo =       0x5555555555
       remainder =                0x1 quotient_hi = 0xfffffffffffffa00 quotient_lo =       0xaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffffffffff400 quotient_lo =      0x15555555555
       remainder =                0x1 quotient_hi = 0xffffffffffffe800 quotient_lo =      0x2aaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffffffffffd000 quotient_lo =      0x55555555555
       remainder =                0x1 quotient_hi = 0xffffffffffffa000 quotient_lo =      0xaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffffffffff4000 quotient_lo =     0x155555555555
       remainder =                0x1 quotient_hi = 0xfffffffffffe8000 quotient_lo =     0x2aaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffffffffd0000 quotient_lo =     0x555555555555
       remainder =                0x1 quotient_hi = 0xfffffffffffa0000 quotient_lo =     0xaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffffffff40000 quotient_lo =    0x1555555555555
       remainder =                0x1 quotient_hi = 0xffffffffffe80000 quotient_lo =    0x2aaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffffffffd00000 quotient_lo =    0x5555555555555
       remainder =                0x1 quotient_hi = 0xffffffffffa00000 quotient_lo =    0xaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffffffff400000 quotient_lo =   0x15555555555555
       remainder =                0x1 quotient_hi = 0xfffffffffe800000 quotient_lo =   0x2aaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffffffd000000 quotient_lo =   0x55555555555555
       remainder =                0x1 quotient_hi = 0xfffffffffa000000 quotient_lo =   0xaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffffff4000000 quotient_lo =  0x155555555555555
       remainder =                0x1 quotient_hi = 0xffffffffe8000000 quotient_lo =  0x2aaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffffffd0000000 quotient_lo =  0x555555555555555
       remainder =                0x1 quotient_hi = 0xffffffffa0000000 quotient_lo =  0xaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffffff40000000 quotient_lo = 0x1555555555555555
       remainder =                0x1 quotient_hi = 0xfffffffe80000000 quotient_lo = 0x2aaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffffd00000000 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xfffffffa00000000 quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffff400000001 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xffffffe800000002 quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffffd000000005 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xffffffa00000000a quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffff4000000015 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xfffffe800000002a quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffffd0000000055 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xfffffa00000000aa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffff40000000155 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xffffe800000002aa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffffd00000000555 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xffffa00000000aaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffff400000001555 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xfffe800000002aaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfffd000000005555 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xfffa00000000aaaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfff4000000015555 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xffe800000002aaaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xffd0000000055555 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xffa00000000aaaaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xff40000000155555 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xfe800000002aaaaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xfd00000000555555 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xfa00000000aaaaaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xf400000001555555 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xe800000002aaaaaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0xd000000005555555 quotient_lo = 0x5555555555555555
       remainder =                0x1 quotient_hi = 0xa00000000aaaaaaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                  0 quotient_hi = 0x4000000015555555 quotient_lo = 0x5555555555555555
       remainder =                  0 quotient_hi = 0x800000002aaaaaaa quotient_lo = 0xaaaaaaaaaaaaaaaa
       remainder =                0x1 quotient_hi =         0x55555555 quotient_lo = 0x5555555555555554
       remainder =                0x2 quotient_hi =         0xaaaaaaaa quotient_lo = 0xaaaaaaaaaaaaaaa8
       remainder =                0x1 quotient_hi =        0x155555555 quotient_lo = 0x5555555555555551
       remainder =                0x2 quotient_hi =        0x2aaaaaaaa quotient_lo = 0xaaaaaaaaaaaaaaa2
       remainder =                0x1 quotient_hi =        0x555555555 quotient_lo = 0x5555555555555545
       remainder =                0x2 quotient_hi =        0xaaaaaaaaa quotient_lo = 0xaaaaaaaaaaaaaa8a
       remainder =                0x1 quotient_hi =       0x1555555555 quotient_lo = 0x5555555555555515
       remainder =                0x2 quotient_hi =       0x2aaaaaaaaa quotient_lo = 0xaaaaaaaaaaaaaa2a
       remainder =                0x1 quotient_hi =       0x5555555555 quotient_lo = 0x5555555555555455
       remainder =                0x2 quotient_hi =       0xaaaaaaaaaa quotient_lo = 0xaaaaaaaaaaaaa8aa
       remainder =                0x1 quotient_hi =      0x15555555555 quotient_lo = 0x5555555555555155
       remainder =                0x2 quotient_hi =      0x2aaaaaaaaaa quotient_lo = 0xaaaaaaaaaaaaa2aa
       remainder =                0x1 quotient_hi =      0x55555555555 quotient_lo = 0x5555555555554555
       remainder =                0x2 quotient_hi =      0xaaaaaaaaaaa quotient_lo = 0xaaaaaaaaaaaa8aaa
       remainder =                0x1 quotient_hi =     0x155555555555 quotient_lo = 0x5555555555551555
       remainder =                0x2 quotient_hi =     0x2aaaaaaaaaaa quotient_lo = 0xaaaaaaaaaaaa2aaa
       remainder =                0x1 quotient_hi =     0x555555555555 quotient_lo = 0x5555555555545555
       remainder =                0x2 quotient_hi =     0xaaaaaaaaaaaa quotient_lo = 0xaaaaaaaaaaa8aaaa
       remainder =                0x1 quotient_hi =    0x1555555555555 quotient_lo = 0x5555555555515555
       remainder =                0x2 quotient_hi =    0x2aaaaaaaaaaaa quotient_lo = 0xaaaaaaaaaaa2aaaa
       remainder =                0x1 quotient_hi =    0x5555555555555 quotient_lo = 0x5555555555455555
       remainder =                0x2 quotient_hi =    0xaaaaaaaaaaaaa quotient_lo = 0xaaaaaaaaaa8aaaaa
       remainder =                0x1 quotient_hi =   0x15555555555555 quotient_lo = 0x5555555555155555
       remainder =                0x2 quotient_hi =   0x2aaaaaaaaaaaaa quotient_lo = 0xaaaaaaaaaa2aaaaa
       remainder =                0x1 quotient_hi =   0x55555555555555 quotient_lo = 0x5555555554555555
       remainder =                0x2 quotient_hi =   0xaaaaaaaaaaaaaa quotient_lo = 0xaaaaaaaaa8aaaaaa
       remainder =                0x1 quotient_hi =  0x155555555555555 quotient_lo = 0x5555555551555555
       remainder =                0x2 quotient_hi =  0x2aaaaaaaaaaaaaa quotient_lo = 0xaaaaaaaaa2aaaaaa
       remainder =                0x1 quotient_hi =  0x555555555555555 quotient_lo = 0x5555555545555555
       remainder =                0x2 quotient_hi =  0xaaaaaaaaaaaaaaa quotient_lo = 0xaaaaaaaa8aaaaaaa
       remainder =                0x1 quotient_hi = 0x1555555555555555 quotient_lo = 0x5555555515555555
       remainder =                0x2 quotient_hi = 0x2aaaaaaaaaaaaaaa quotient_lo = 0xaaaaaaaa2aaaaaaa
       remainder =                0x1 quotient_hi = 0x5555555555555555 quotient_lo = 0x5555555455555555
       result = 0x5555555455555555

## Style notes

1. `int isneg` should be `bool isneg`
2. To me this code:

       if (sbit) remainder |= 1;
       // shift quotient left 1
       {
           quotient_hi <<= 1;
           if (quotient_lo & (Uint64(1) << 63)) quotient_hi |= 1;
           quotient_lo <<= 1;
       }

   Looks as if it ment this:

       if (sbit)
       {
           quotient_hi <<= 1;
           if (quotient_lo & (Uint64(1) << 63)) quotient_hi |= 1;
           quotient_lo <<= 1;
       }

   But it actually means:

       if (sbit) {
           remainder |= 1;
       }

       // shift quotient left 1
       {
           quotient_hi <<= 1;
           if (quotient_lo & (Uint64(1) << 63)) quotient_hi |= 1;
           quotient_lo <<= 1;
       }

## Using C++17 rules

Pioneer currently targets the C++17 standard:

https://github.com/pioneerspacesim/pioneer/blob/97d7be673f2124f1b42aee956c4b3a6fc2129558/CMakeLists.txt#L5

C++17 is much less permissive than C++20 when it comes to `<<`, `>>` and unsigned-to-signed conversions:

1. `a.v >> (64 - FRAC)` is implementation-defined if `a.v < 0`
2. `a.v << FRAC` is undefined if `a.v < 0`
3. `quotient_hi <<= 1;` is implementation-defined if the mathematical value of `quotient_hi * 2` is greater than INT64_MAX (because the defenition of `<<` references unsigned-to-signed conversion)
4. `quotient_hi <<= 1;` is undefined if `quotient_hi < 0`
5. `Sint64(quotient_lo)` is implementation-defined if quotient_lo is greater than INT64_MAX 
6. The return statement converts Uint64 to fixedf, which requires an implicit conversion to Sint64. The conversion is implementation-defined if the returned value is greater than INT64_MAX

* 3 and 4 are triggered multiple times for every `a` except 0.
* [1, 3 and 5 appear to be OK according to the gcc documentation](https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html)
* [1 appears to be OK in Microsoft C++](https://learn.microsoft.com/en-us/cpp/cpp/left-shift-and-right-shift-operators-input-and-output?view=msvc-170#right-shifts)
* [3 and 5 appear to be OK in Microsoft C++](https://learn.microsoft.com/en-us/cpp/cpp/type-conversions-and-type-safety-modern-cpp?view=msvc-170#signed---unsigned-conversions)
* Because Clang **does not** document implementation-defined behaviour, we can only hope and pray it does what gcc does.

## Note

I think most of this issues can be fixed by using only Uint64 inside this function.

[Safely obtaining the modulus of a signed integer as an unsigned integer](https://stackoverflow.com/a/17313717/14516046)

Adapend for out needs:

```C++
Uint64 uabs(Sint64 input)
{
    if (input >= 0)
    {
        return static_cast<Uint64>(input);
    }
    else
    {
        return -static_cast<Uint64>(input);
    }
}
```

hexagonrecursion:
Fixes #5697

Note: the return statement at the end does convert Uint64 to Sint64, which is *implementation-defined* in C++17 if the value is > INT64_MAX. Here is why I think this is probably OK:

1. My biggest argument why I think this is OK is that C++20 guarantees wraparound modulo 2^64 in this case.
2. [gcc guarantees the right result](https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html)
3. [Microsoft C++ guarantees the right result](https://learn.microsoft.com/en-us/cpp/cpp/type-conversions-and-type-safety-modern-cpp?view=msvc-170#signed---unsigned-conversions)
4. [clang unfortunately does not document its implementation-defined behaviour](https://github.com/llvm/llvm-project/issues/11644)

### DIFF
--- a/src/fixed.h
+++ b/src/fixed.h
@@ -149,34 +149,39 @@ public:
 	}
 	friend fixedf operator/(const fixedf &a, const fixedf &b)
 	{
-		// 128-bit divided by 64-bit, to make sure high bits are not lost
-		Sint64 quotient_hi = a.v >> (64 - FRAC);
-		Uint64 quotient_lo = a.v << FRAC;
-		Sint64 d = b.v;
-		int isneg = 0;
-		Sint64 remainder = 0;
-
-		if (d < 0) {
-			d = -d;
-			isneg = 1;
+		Uint64 abs_a = a.v, abs_b = b.v;
+		bool is_neg = false;
+		if (a.v < 0) {
+			abs_a = -abs_a;
+			is_neg = !is_neg;
 		}
-
+		if (b.v < 0) {
+			abs_b = -abs_b;
+			is_neg = !is_neg;
+		}
+		// 128-bit divided by 64-bit, to make sure high bits are not lost
+		Uint64 quotient_hi = abs_a >> (64 - FRAC);
+		Uint64 quotient_lo = abs_a << FRAC;
+		Uint64 remainder = 0;
 		for (int i = 0; i < 128; i++) {
 			Uint64 sbit = (Uint64(1) << 63) & quotient_hi;
+			// The most significant bit of remainder is 0 because abs_b <= (Uint64(1) << 63)
 			remainder <<= 1;
-			if (sbit) remainder |= 1;
+			if (sbit) {
+				remainder |= 1;
+			}
 			// shift quotient left 1
 			{
 				quotient_hi <<= 1;
 				if (quotient_lo & (Uint64(1) << 63)) quotient_hi |= 1;
 				quotient_lo <<= 1;
 			}
-			if (remainder >= d) {
-				remainder -= d;
+			if (remainder >= abs_b) {
+				remainder -= abs_b;
 				quotient_lo |= 1;
 			}
 		}
-		return (isneg ? -Sint64(quotient_lo) : quotient_lo);
+		return is_neg ? -quotient_lo : quotient_lo;
 	}
 	friend bool operator==(const fixedf &a, const fixedf &b) { return a.v == b.v; }
 	friend bool operator!=(const fixedf &a, const fixedf &b) { return a.v != b.v; }
--- a/src/test/TestFixed.cpp
+++ b/src/test/TestFixed.cpp
@@ -1,0 +1,33 @@
+#include <limits>
+#include "doctest.h"
+#include "fixed.h"
+
+TEST_CASE("Fixed")
+{
+	SUBCASE("operator ==")
+	{
+		// 42.0 == 42.0
+		CHECK(fixed(42, 1) == fixed(42, 1));
+		// 123.0 != 321.0
+		CHECK(fixed(123, 1) != fixed(321, 1));
+		// 7.0 == 7.0
+		CHECK(fixed(0x7'00000000) == fixed(7, 1));
+	}
+
+	SUBCASE("operator /")
+	{
+		// 3.0 / 3.0 == 1.0
+		CHECK(fixed(3, 1) / fixed(3, 1) == fixed(1, 1));
+		// -3.0 / 3.0 == -1.0
+		// There used to be a bug causing this to produce an incorrect result
+		CHECK(fixed(-3'00000000) / fixed(3, 1) == fixed(-1'00000000));
+		// 0x40000000.00000000 / 0x40000000.00000001 == 0x00000000.ffffffff (because we round down)
+		// There used to be a bug causing this to produce an incorrect result
+		CHECK(fixed(0x40000000'00000000) / fixed(0x40000000'00000001) == fixed(0x00000000'ffffffff));
+		// fixed(INT_MIN) / fixed(INT_MIN) used to trigger a signed int overflow
+		Sint64 min = std::numeric_limits<Sint64>::min();
+		CHECK(fixed(min) / fixed(min) == fixed(1, 1));
+		// fixed(INT_MIN) / -1.0 used to trigger a signed int overflow
+		CHECK(fixed(min) / fixed(-1'00000000) == fixed(min) / fixed(-1'00000000));
+	}
+}


hexagonrecursion:
[I believe I have found a way to implement unsigned to signed conversion with wraparound in C++without relying on implementation defined behavour.](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyahUAUgBMAIUtXyKxqiIEh1ZpgDC6egFc2TEAtydwAZAiZsADk/ACNsUikANnIAB3QlYhcmL19/QNT0zKEwiOi2OISeZIdsJyyRIhZSIhy/AJ57bEdnIQamohKo2PiDJUbm1ryOsf7BsoqpAEp7dB9SVE4uSwBmcNRfHABqM22PVDHCYRPcMy0AQVu7ogBPFOwcGkPwokTpAH0iIcRN9ficbPcXm8PocfCD/oCFHCwY9HsDhL9DsB0FgIIj0dJDkw/ItjgB2cF3Q5Ur6fCBEtjHba4Q64uEkgCSkQAKr8/gBZO4ADRJZnJj2pEsOAHopYd2WwUowOMJDgB3UgsFJNTCrJiYcWSqn0w4AWhOABEWXiftIOdzeQLBcj7oajX5TRbDjxnZTXWibYdSNglD56IDPfSfa7A8HQ%2BHrJ7OTz/nzOVHXUGiGsmDGQ2H01TReaDVTM9nCX4fUWuMt6NwAKz8AJcHTkdDcDy2WyHJSrdbYY4WbZ8chEbQ15YAaxA222ADpZ4ul8vknWuNJ%2BGwQPWtORm6321x%2BEoQLuxy2a%2BQ4LAUBgcPhiGRKNQ6IxWBxuCPBMIxBJODI5GEZQ1E0C9yH0IIjBMNAu2sTpuiyNw9UmAIglCcIhnKEYAA4CgyHpsm8Np8jSfCsjmYYElwmo6l6cYWiIvIghogi%2BmaCisKo%2Bx6JQ/IZnYjD5hw5ZezWDZuB2PYDgHE4zgub5rhRCFXnebBPjhAEgSRbYKUhVTPlhfFNOtUEdKUu5/QxNgWGeOI/ixHETIJekRTFF1qTLUgc0jMz7mrWsGybcc2w7WCrB7PsNkHbYLH4c8dEWZYkGwFgcASCAAvXTdt13fd%2BEPY9T1HcdEvIacLHrOcAE5622LQtAsRIeC0KrEkSLR61XbhtiCsCCuKi9lmvBB4GGiBb3QBUGHiZ8IAwKbGASYAeBigQGCIeITwgGJgpicImmeT9%2BD21hSGeAB5GJdFqc8R3m5UiHOph6EOsCcBiHxgA8CR6BPXh%2BBwayTEkN6CCDOoADdg2C7B1FqHwNqOyhhC6YL6AIGINTOrwcGCohSAILd/uWGgjGAJQADUCGwVVzreZsv3kX9JAA79FBUDRgv0DooNMMLDAxk9IGWdAUgIv6TXOnhTQuC1zATaweFJfKodIAmcCFjL4JuxCIHcXiOnQ0pKIMUiikI3J2jw82OIWaYuh1uj%2BgN7XaKYNiBkEk3ph4xirf4z3jc4pYVjE/9MsbPdgsPQ51GwxITQxYBUFQL15wsFkHxIUhop4RY4pKqcZ3nRIqukLRpHrcvSUSYdGvrQxuA3cgtx3KO%2Bu4QqzxKq9EHGtBJpSaaKCoObB%2BHkBltW18NtILadrAk6DqRpezsu66nCR%2B7mEe57Xtbd7Pu%2B%2BhfqRwHjGAEGD7BnWob%2B1tYfhxH/uRja11bdHMYOnHNlbfHCaOiTMmlNqa03pkjNmzN/yyDZsBTmYF9AN15jBBWdhP6axFmLLIEspYyyIPqbY5p5Y2GsBYA8qt1bQ2Fq7AiSFPB%2BwMEbTCdtrYERdmbAitsRj2wQk7CYDCeGO3dvRLhlRuLOwEeIgSQc7YiUiuHRuXBI55RClwWO8dE4EmTqnFac4M4QCzmQXO%2BcBoJSSilNK1Ai6zjnIkHcVVsKziqhYKqWhsJVEUc3FR/UTzd0Gr3UaE0FozVHvNIei00C8yVloDoM9NrUAXq2Ver0RzJPXjdLek0HpPResFQ%2BX0fp/RHOfYGv8AY30htDMCj9UAI02COb4qMwKfyxs8H%2BeMCZEz4EAlg5MqY0zpswCBTNxAsxgfIOBoFWwQUMBfFBJC0GC3gJg8W3BJbSxNLLQhxDbBKxVvEShGCaG631pIphQkxEcKyOwwonCvbB0EW7D2LsWL1BEfclhAcXnvJkdwuRYcQ5rmUdHbg6iE4Yn2BfL0pI5xaDnNLAxhBs7GILoNcxqURha2nDFWxVcqqkh4DwKqZd2qdU8b1A8nd7BFXihORRsUW45XbpSo8pjLxjT7kE8JITqBhInlPIIcS54JN2vtM6K8xUXSuhkl%2B29hA5P3gDbAH0CknyKUqoGl8ynkEIODZwd8YZw1qc/BpKN378Bad/LA2r/5dMvKTXpICBngJfpA0Z0DALsxAlzGcszoI7Lgug5ZbYsFCBwRsrZRCwp7LbBQggGtg2vNcHrZCZy9SiNNrc65kirnFA%2Bdw45fCGKW1GA7J5PzmEFq%2BZIgOGa86h37IECOFL8qgrjuCrRKc056MzkioxOwLAmNpaVZKGL0rWPnNhCwpItC11JM4ux2EHHksZW3bxVLfFsqGpygewSR68vHhEgVa0wzxO2qK06KTjqSvSZvWVWSd4Krycqo%2BhSz7YE1VfcpeqCAGuqUaupSNGnmoFl/bG1qOkAOJgIYB/SwFDNdSMv8UhxlAQ5lMvQIBVxzIDYsmIRzRarK4OsvBBCo2oLIfstW8aqFayTQEFN9CS2G3Tfmy5WahA3LInm35Yj6PPJrWW1iFaLmjF9sxqRgdK2VH%2BY2wdijgUdzUe2zRmIu26P0YYnOA6h2F3IKOyxWLi7wqqnVbC5cOo8Hji1WQa4vEgtZZu4dASby7u5fusee7IkX2ibE9aZ7EnXsvRKy9t7br8DlbvXJb0X2qtPi/EpWq8kVP1VUh%2BAGTX8GA2jDGrT2lgVtYAmDjq4ODIZvwN1yHWYTPQz62zOH%2BZBuoYR7BazcGbPwXLaNytY0HJo0c%2BjdCXbnO9qw7NEnc1MDrYW4REiJP8ZE6N6t83FsPNk%2BJeTQKW2qLBapyFJhoWwvhb2x82mhyDtRWY/TFjMXWIsNVed2F6wWDIRYHgeKPF2e2z4mlem1wMtbrlBzl26XTlJNIWF0hS74u2EurQTjtiKJ6sy1trLh2ZQZeutHenVYZFcNIIAA%3D%3D) Would this be overkill?

```C++
#include <cstdint>

typedef int64_t Sint64;
typedef uint64_t Uint64;

Sint64 good(Uint64 num) {
    if (num > (Uint64) INT64_MAX) {
        // Implement wrapardound
        num -= (Uint64) INT64_MAX;
        num -= 1;
        Sint64 result = num;
        result += INT64_MIN;
        return result;
    }
    return num;
}
```

impaktor:
@hexagonrecursion Since you seem to have been playing around with our code base, I'm just informing you we'll soon feature freeze (-ish), for the annual [February 03](https://github.com/pioneerspacesim/pioneer/issues/4496#issuecomment-459761900) release.

hexagonrecursion:
Thanks. I have a lot on my plate though so I don't expect to make more pioneer pull requests any time soon.

src/fixed.h
```diff
@@ -149,34 +149,39 @@ class fixedf {
...
+		bool is_neg = false;
+		if (a.v < 0) {
+			abs_a = -abs_a;
+			is_neg = !is_neg;
 		}
-
+		if (b.v < 0) {
+			abs_b = -abs_b;
+			is_neg = !is_neg;
+		}
```
Web-eWorks, line 161: I would strongly recommend using bitwise ops and std::abs rather than explicit branches, as it will generally compile into optimized code under more scenarios. E.g. something like:
```c++
uint64_t abs_a = std::abs(a.v), abs_b = std::abs(b.v);
bool is_neg = int(a.v < 0) ^ int(b.v < 0);
```

(Note: this is well-defined, `true` is defined to promote to the integral value `1`.)
See this Godbolt for example of how this affects code that would be generated in debug mode: https://godbolt.org/z/6MsTPv8dz

Yes, under `-O3` or equivalent this code would produce the same generated assembly, but the above suggestion will both produce better performance in debug mode (wasting less programmer time) and more clearly expresses what is logically occurring in the if statements without the extra cognitive overhead of the order-dependent logic.

Web-eWorks:
> Would this be overkill?

The primary concern of the fixed-point function library is determinism between GCC-on-Linux and MSVC-on-Windows. Performance is a *very* close second, so if GCC and MSVC provide uniform outcomes of their implementation-defined behavior (and Clang can be tested to comply with the expected behavior) then I would strongly advise against introducing additional branches into math code that is expected to have extremely high throughput.

As a corollary, given this PR addresses very niche/specific behavior which can be implementation or platform dependent and thus remain "invisible" if broken, I'd *strongly* recommend adding a new test case validating the outcome of the specific division cases you're addressing to our unit testing suite. We define unit tests in `src/test` using DocTest, and it should be extremely simple to add new tests to that suite.

Web-eWorks (changes requested):
Overall, I'm quite glad someone is addressing this. I believe the fixed-point code to either predate or be written around the time C++11 was introduced, and certainly well before Pioneer began adopting the standard, much less C++17.

Thank you for taking the time to ensure it is numerically correct and stable, I'm sure it will save some time in the long run otherwise spent hunting down ghost bugs!

I've left one change request that I'd like to see addressed before merge - minimizing branches where possible in high-throughput code is strongly preferred, and the system generation code performance as a whole is primarily dependent on the fixed-point math library.

src/test/TestFixed.cpp
```diff
@@ -0,0 +1,33 @@
+#include <limits>
```
impaktor, line 1: Please add:

```c++
// Copyright © 2008-2024 Pioneer Developers. See AUTHORS.txt for details

```

impaktor (changes requested):

src/test/TestFixed.cpp
hexagonrecursion: Oops. I forgot

src/fixed.h
hexagonrecursion: > I would strongly recommend using bitwise ops and std::abs rather than explicit branches, as it will generally compile into optimized code under more scenarios. E.g. something like:
> 
> ```c++
> uint64_t abs_a = std::abs(a.v), abs_b = std::abs(b.v);
> bool is_neg = int(a.v < 0) ^ int(b.v < 0);
> ```
> 
> (Note: this is well-defined, `true` is defined to promote to the integral value `1`.) See this Godbolt for example of how this affects code that would be generated in debug mode: https://godbolt.org/z/6MsTPv8dz
> 
> Yes, under `-O3` or equivalent this code would produce the same generated assembly, but the above suggestion will both produce better performance in debug mode (wasting less programmer time) and more clearly expresses what is logically occurring in the if statements without the extra cognitive overhead of the order-dependent logic.

Done. Please note that with the requested changes dividing fixed(INT64_MIN) by anything or dividing anything by fixed(INT64_MIN) causes undefined behaviour because std::abs(INT64_MIN) is undefined behaviour.

Web-eWorks (approved):
I'd suggest a "canary" test be added to the test suite to check for the behavior of division tests involving INT64_MIN which rely on undefined behavior. Otherwise, this looks good to me - thanks for addressing the review feedback.

Because this has a non-zero chance of implicitly altering procedural generation determinism when merged, I'm going to defer merge of this PR until we've fully started the development cycle for the next major release.

hexagonrecursion:
> I'd suggest a "canary" test be added to the test suite to check for the behavior of division tests involving INT64_MIN which rely on undefined behavior

@Web-eWorks I do not understand. At the time of writing _none_ of the tests added by this pull request (to the best of my knowledge) rely on undefined behavior. Are you suggesting we add a test that deliberately triggets undefined behaviour? What would this test assert?

The use of std::abs() is to the best of my knowledge the only source of undefined behaviour in my current implementation of `fixedf operator/(fixedf,fixedf)`. This can only be triggered by passing fixed(INT64_MIN) as a numerator or demoninator to the / operator - all other values should be fine.

undefined behaviour-free version:

    int64_t a = INT64_MIN;
    uint64_t abs_a = a;
    if (a < 0 /* true */) {
        abs_a = -abs_a;  // Defined: -uint64_t(INT64_MIN) == uint64_t(INT64_MIN)
    }

undefined behaviour version:

    int64_t a = INT64_MIN;
    // Undefined behaviour because the return type of std::abs() is
    // int, long or long long (depending on overload resolution)
    // can't represent the absolute value of INT64_MIN
    uint64_t abs_a = std::abs(a);  

1. I thought you suggested std::abs() because you were confident fixed(INT64_MIN) will never occur in practice
1. It is impossible to "test" undefined behaviour - a passing test proves _nothing_.
1. A test that "tests" undefined behaviour is problematic because it will be flagged when someone runs the test suite with sanitizers.

hexagonrecursion:
#### What do you think I should do?

Brainstorming alternatives:
1. `std::abs(a.v)`
    * _undefined behaviour_ if the absolute value of a.v is outside the range of the return type of abs().
    * For C++20 there should be only one such value: INT64_MIN
    * For gcc (regardless of C++ standard) there should be only one such value: INT64_MIN because "[GCC supports only two’s complement integer types](https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html)"
    * For msvc  (regardless of C++ standard) there should be only one such value: INT64_MIN because "[Signed integers are represented in two's-complement form](https://learn.microsoft.com/en-us/cpp/c-language/range-of-integer-values?view=msvc-170)"
    * So don't divide by fixed(INT64_MIN) and don't divide fixed(INT64_MIN) by anything
3. `if (a.v < 0) abs_a = -abs_a;`
    * Free of UB
    * [Adds a branch under -Og](https://godbolt.org/z/835YxKdGd) - may or may not have a measurable performance impact because the same function contains a loop with 128 iterations - the cost of the loop may overshadow the cost of the branch. I could look into doing a performance benchmark.
4. [I could look into the possibility of changing optimization level for one function](https://stackoverflow.com/a/49353441)
5. I could look into the possibility of implementing the calculation of the absloute value without branches using inline assembly with a portable C++ fallback

sturnclaw (approved):
I consider the current implementation acceptable, as division of or by INT64_MIN is unlikely to intentionally occur without also triggering overflow.

The primary intent of a test was to ensure that the undefined behavior of the platform in question matches the expected result of the undefined behavior, i.e. to validate that:

```c++
fixed(INT64_MIN) / fixed(1, 1) == fixed(EXPECTED_VALUE);
```

If that test failed on a platform we intend to support, we would have a reason to replace the `std::abs()` implementation with the slower but fully-defined implementation, as the real-world effect would be a loss of determinism.

In short: undefined behavior isn't a bad thing as long as it's the *same* undefined behavior on all platforms we intend to support.

PR #5706, opened 2024-01-04
Add a function to recover from old versions of saves

Gliese852:
[recording.webm](https://github.com/pioneerspacesim/pioneer/assets/18342621/f38107d6-35ca-45ab-ad8d-8c5049211509)

If the saveload menu decides that it cannot load a save (most likely due to a version mismatch), the `Load` button will be replaced with `Recover` and you can try to restore the game.

In fact, this is the start of a new game, but we are trying to read the parameters from the save. After a recovery attempt, the same dialog for starting a new game opens, but without selecting a start option.

If an error occurs when reading a parameter from a save, or after restoring the parameter is in an invalid state, it is unlocked so that the player can manually specify it. The exception is the flight log - it is always locked.

Version 89 is currently being successfully restored (2023-02-03). For older versions, it is planned to add "reader" versions in the future. Added the ability to create a "versioned" function; when it is launched, dispatching occurs depending on the specified save version.

Also, now when you launch a new game window in debug mode (with ctrl), it immediately switches to "Custom start". In this case, there are no changes to the parameters after the previous launch of the new game window. That is, if you previously ran a recovery, you can open a new game window in debug mode and there will be the same options, but unlocked.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -359,6 +359,10 @@
     "description": "Time of departure, used in flight log",
     "message": "Departure date"
   },
+  "DESCRIPTION": {
+    "description": "Start game parameter, e.g. 'This is an easy start from Cydonia'",
+    "message": "Description"
+  },
   "DESTROY_ENEMY_SHIP": {
     "description": "",
     "message": "Destroy enemy ship"
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -2307,6 +2307,14 @@
     "description": "",
     "message": "Type"
   },
+  "UNKNOWN_FACTION": {
+    "description": "For entry in the flight log",
+    "message": "Unknown faction"
+  },
+  "UNKNOWN_LOCATION_IN_SECTOR_X": {
+    "description": "For entry in the flight log",
+    "message": "Unknown location in sector {sector}"
+  },
   "UNLAWFUL_WEAPONS_DISCHARGE": {
     "description": "",
     "message": "Unlawful weapons discharge"
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -19,6 +19,10 @@
     "description": "",
     "message": "Allegiance:"
   },
+  "ALL_PARAMETERS_WERE_RECOVERED_SUCCESSFULLY": {
+    "description": "Message when restoring savegame",
+    "message": "All parameters were recovered successfully."
+  },
   "ALL_UP_WEIGHT": {
     "description": "",
     "message": "All-up weight"
@@ -50,6 +54,10 @@
   "ATTEMPT_TO_REPAIR_HULL": {
     "description": "",
     "message": "Attempt to repair hull"
+  },
+  "ATTEMPTING_TO_RECOVER_GAME_PROGRESS_FROM_SAVE_X": {
+    "description": "Message when restoring savegame",
+    "message": "Attempting to recover game progress from save: {save}"
   },
   "AUTO_ROUTE": {
     "description": "Automatically find shortest jump route (measured in time) in star map",
@@ -98,6 +106,14 @@
   "CANNOT_SELL_ITEM": {
     "description": "",
     "message": "This item cannot be sold"
+  },
+  "CANT_FIND_PATH": {
+    "description": "Message when restoring savegame, file path",
+    "message": "Can't find path: %s"
+  },
+  "CANT_FIND_PLAYERS_SHIP_IN_SPACE": {
+    "description": "Message when restoring savegame",
+    "message": "Can't find player's ship in space"
   },
   "CAPACITY": {
     "description": "Capacity of ship to fit more equipment",
@@ -267,9 +283,17 @@
     "description": "",
     "message": "Could not delete save"
   },
+  "COULD_NOT_FIND_SYSTEM_BODY_FOR_FRAME": {
+    "description": "Message when restoring savegame",
+    "message": "Could not find system body for frame"
+  },
   "COULD_NOT_LOAD_GAME": {
     "description": "",
     "message": "Could not load game: "
+  },
+  "COULD_NOT_READ_THE_SAVE_VERSION": {
+    "description": "Message when restoring savegame",
+    "message": "Could not read the save version"
   },
   "CREDIBLE": {
     "description": "For player reputation",
@@ -302,6 +326,10 @@
   "CURRENT_FUEL": {
     "description": "For hyperjump planner",
     "message": "Current Fuel:"
+  },
+  "CURRENT_SAVEGAME_VERSION": {
+    "description": "Message when restoring savegame",
+    "message": "Current savegame version: %s"
   },
   "CURRENT_SHIP": {
     "description": "For tooltip comparing current ship specs to one in the shipyard",
@@ -411,6 +439,10 @@
     "description": "",
     "message": "Docked at"
   },
+  "DOCKED_TO_A_BODY_THAT_IS_NOT_A_STATION": {
+    "description": "Message when restoring savegame",
+    "message": "Docked to a body that is not a station"
+  },
   "DOCK_AT_CURRENT_TARGET": {
     "description": "",
     "message": "Dock at current target"
@@ -503,6 +535,10 @@
     "description": "",
     "message": "Error"
   },
+  "ERROR_APPLYING_PATCH": {
+    "description": "Message when restoring savegame",
+    "message": "Error applying patch"
+  },
   "ERROR_LANDING_GEAR_DOWN": {
     "description": "",
     "message": "Can not initiate hyperjump; landing gear is not retracted!"
@@ -582,6 +618,10 @@
   "FEMALE": {
     "description": "",
     "message": "Female"
+  },
+  "FILE_SAVEGAME_VERSION": {
+    "description": "Message when restoring savegame",
+    "message": "File savegame version: %s"
   },
   "FINAL_TARGET": {
     "description": "Hyperjump planner, referring to star system target",
@@ -1491,6 +1531,10 @@
     "description": "",
     "message": "More info..."
   },
+  "MORE_THAN_ONE_PLAYERS_SHIP_FOUND_IN_SPACE": {
+    "description": "Message when restoring savegame",
+    "message": "More than one player's ship found in space"
+  },
   "MOSTLY_HARMLESS": {
     "description": "",
     "message": "Mostly Harmless"
@@ -1727,6 +1771,10 @@
     "description": "Entry for ship info",
     "message": "Passenger cabin capacity"
   },
+  "PATH_FOUND_BUT_VERIFICATION_FAILED": {
+    "description": "Message when restoring savegame",
+    "message": "Path found but verification failed: %s"
+  },
   "PATTERN": {
     "description": "Ship painting pattern.",
     "message": "Pattern"
@@ -1863,6 +1911,14 @@
     "description": "",
     "message": "Rear weapon"
   },
+  "RECOVER": {
+    "description": "Saveload menu button",
+    "message": "Recover"
+  },
+  "RECOVER_SAVEGAME": {
+    "description": "",
+    "message": "Recover savegame"
+  },
   "REFUEL": {
     "description": "",
     "message": "Refuel"
@@ -1922,6 +1978,10 @@
   "RESET_PREVIEW": {
     "description": "Paintshop button",
     "message": "Reset Preview"
+  },
+  "RECOVERED_BUT_NOT_VALID": {
+    "description": "Message when restoring savegame",
+    "message": "Recovered but not valid"
   },
   "RETURN_TO_GAME": {
     "description": "",
@@ -2195,6 +2255,10 @@
     "description": "Tab item in the start game menu",
     "message": "Summary"
   },
+  "SYSTEM_BODY_INDEX_IS_OUT_OF_RANGE": {
+    "description": "Message when restoring savegame",
+    "message": "System body index is out of range"
+  },
   "TECH_CERTIFIED": {
     "description": "Lobby screen shows the tech level (an integer number) of the station",
     "message": "This station holds a level {tech_level} technology certificate."
@@ -2218,6 +2282,10 @@
   "THERE_IS_NOBODY_ELSE_ON_BOARD_ABLE_TO_FLY_THIS_SHIP": {
     "description": "",
     "message": "There is nobody else on board able to fly this ship."
+  },
+  "THERE_WERE_ERRORS_IN_RECOVERY": {
+    "description": "Message when restoring savegame",
+    "message": "There were errors in recovery:"
   },
   "THE_SHIP_IS_UNDER_STATION_CONTROL_COMMANDER": {
     "description": "",
@@ -2311,6 +2379,10 @@
     "description": "",
     "message": "Type"
   },
+  "UNKNOWN_CUSTOM_LOG_ENTRY_FORMAT": {
+    "description": "Message when restoring savegame",
+    "message": "Unknown custom log entry format"
+  },
   "UNKNOWN_FACTION": {
     "description": "For entry in the flight log",
     "message": "Unknown faction"
@@ -2318,6 +2390,14 @@
   "UNKNOWN_LOCATION_IN_SECTOR_X": {
     "description": "For entry in the flight log",
     "message": "Unknown location in sector {sector}"
+  },
+  "UNKNOWN_STATION_LOG_ENTRY_FORMAT": {
+    "description": "Message when restoring savegame",
+    "message": "Unknown station log entry format"
+  },
+  "UNKNOWN_SYSTEM_LOG_ENTRY_FORMAT": {
+    "description": "Message when restoring savegame",
+    "message": "Unknown system log entry format"
   },
   "UNLAWFUL_WEAPONS_DISCHARGE": {
     "description": "",
@@ -2334,6 +2414,10 @@
   "UNRELIABLE": {
     "description": "",
     "message": "Unreliable"
+  },
+  "UNSUCCESSFULLY_RECOVERED_PARAMETERS_ARE_NOW_UNLOCKED_PLEASE_SET_THEM_MANUALLY": {
+    "description": "Message when restoring savegame",
+    "message": "Unsuccessfully recovered parameters are now unlocked, please set them manually."
   },
   "UP_ACCEL": {
     "description": "",
--- a/data/pigui/libs/message-box.lua
+++ b/data/pigui/libs/message-box.lua
@@ -5,49 +5,58 @@ local ui = require 'pigui'
 local ModalWindow = require 'pigui.libs.modal-win'
 local Lang = require 'Lang'
 local lui = Lang.GetResource("ui-core")
-
-local OK_BUTTON_WIDTH = 100
+local font = ui.fonts.pionillium.body
+local msgButtonWidth = font.size * 7
 
 local msgbox = {}
 
-msgbox.OK = function(msg)
+local function createBoxModal(msg, footer, footerWidth)
 	ModalWindow.New('PopupMessageBox', function(self)
-		ui.text(msg)
+
+		local textSize = ui.calcTextSize(tostring(msg) .. "\n.")
+		textSize.x = math.min(textSize.x, ui.screenWidth * 0.6)
+		textSize.y = math.min(textSize.y, ui.screenHeight * 0.6)
+		ui.child('PopupMessageBoxText', textSize, { 'HorizontalScrollbar' }, function ()
+			ui.text(msg)
+		end)
+
+		-- centered footer
 		local width = ui.getContentRegion().x
-		if width > OK_BUTTON_WIDTH then
-			ui.dummy(Vector2((width - OK_BUTTON_WIDTH) / 2, 0))
-			ui.sameLine()
+		if width > footerWidth then
+			ui.dummy(Vector2((width - footerWidth) / 2, 0))
+			ui.sameLine(0, 0)
 		end
-		if ui.button(lui.OK, Vector2(OK_BUTTON_WIDTH, 0)) then
-			self:close()
-		end
+		footer(self)
 	end,
 	function (_, drawPopupFn)
 		ui.setNextWindowPosCenter('Always')
-		ui.withStyleColors({ PopupBg = ui.theme.colors.modalBackground }, drawPopupFn)
+		ui.withFont(font, function()
+			ui.withStyleColors({ PopupBg = ui.theme.colors.modalBackground }, drawPopupFn)
+		end)
 	end):open()
 end
 
+msgbox.OK = function(msg)
+	createBoxModal(msg, function(self)
+		if ui.button(lui.OK, Vector2(msgButtonWidth, 0)) then
+			self:close()
+		end
+	end, msgButtonWidth)
+end
+
 msgbox.OK_CANCEL = function(msg, callback)
-	ModalWindow.New('PopupMessageBox', function(self)
-		ui.text(msg)
-		local width = ui.getContentRegion().x
-		local okButton = ui.button(lui.OK, Vector2(OK_BUTTON_WIDTH, 0))
-		if okButton then
+	createBoxModal(msg, function(self)
+		if ui.button(lui.OK, Vector2(msgButtonWidth, 0)) then
 			if callback then
-				callback(okButton)
+				callback()
 			end
 			self:close()
 		end
-		ui.sameLine(width - OK_BUTTON_WIDTH, 0)
-		if ui.button(lui.CANCEL, Vector2(OK_BUTTON_WIDTH, 0)) then
+		ui.sameLine()
+		if ui.button(lui.CANCEL, Vector2(msgButtonWidth, 0)) then
 			self:close()
 		end
-	end,
-	function (_, drawPopupFn)
-		ui.setNextWindowPosCenter('Always')
-		ui.withStyleColors({ PopupBg = ui.theme.colors.modalBackground }, drawPopupFn)
-	end):open()
+	end, msgButtonWidth * 2 + ui.getItemSpacing().x)
 end
 
 return msgbox
--- a/data/pigui/modules/new-game-window/class.lua
+++ b/data/pigui/modules/new-game-window/class.lua
@@ -196,6 +196,7 @@ local function startGame(gameParams)
 			logWarning("Wrong entry for the station flight log")
 		end
 	end
+	FlightLog.OrganizeEntries()
 
 	if gameParams.autoExec then
 		gameParams.autoExec()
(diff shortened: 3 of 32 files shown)


impaktor:
It would be absolutely fantastic to have this in master before next "stable" release (in less than a month now!). 

My thoughts based on what I saw in the video:

(italics indicate suggested change)
1. could it be "_Minimum_ supported game save version 90" instead of "Actual"?
2. and continue the "unsuccessfully restored parameters are _now_ unlocked, _please set them manually_"

In what way will this collide with @JonBooth78's PR #5666? Just checking so he knows if there's something he should do before we merge that PR.

Great work!

Gliese852:
@impaktor Thanks! For questions:
> 1. could it be "_Minimum_ supported game save version 90" instead of "Actual"?

The fact is that in a sense, all saves are supported, there is no minimum version, it’s just that the older the save, the more errors there will be when restoring. And although the "readers" are written based on version 89, perhaps older versions will also be restored somehow. I have a save of version 86 (October 11, 2020), "only" cargo and equipment failed to be restored from it.
> 2. and continue the "unsuccessfully restored parameters are _now_ unlocked, _please set them manually_"

placed in TODO, there might be more similar comments

impaktor:
> The fact is that in a sense, all saves are supported, there is no minimum version

I guess this comes down to semantics in the end. Either way, I think calling version 90 "Current version" or "Minimum version" is preferable over "Actual version", as I don't think the use of the word "Actual" is optimal here.

JonBooth78:
I don't think this will make much difference to my change in #5666 - if it does I can account for that as I'll need to rebase it anyway now as things have moved on.

JonBooth78:
Hmm, OK, I'll need to look at that then; since the new version was written to read with backwards compatibility I'd hoped it would 'just work'...

Moving forwards then, after this change to give us backward compatibility on old save versions will we need to write multiple code paths?

Web-eWorks:
Note: we will be savebumping to v90 as part of this release. I'd ideally like to merge this PR "last" after all other major feature PRs to keep the amount of rewriting required to a minimum.

Gliese852:
@JonBooth78 Yes, I saw you provided backwards compatibility, but unfortunately, backwards compatibility won't work because as soon as the save version is increased, the save simply won't load, at least that's how it works now. 
It would probably be possible to change this by maintaining backwards compatibility at all times, ensuring that the previous version of a save will load in the next one, but it seems that maintaining compatibility through a recovery system should be less expensive since a significant amount of data is ignored.

Gliese852:
@JonBooth78 
> Moving forwards then, after this change to give us backward compatibility on old save versions will we need to write multiple code paths?

If I understood the question correctly, "inside the game code", you do not need to maintain any backward compatibility, and your code can be quite straightforward.

But it is required to maintain the correct operation of changed "starting parameter" in the start menu, including editing in the "Custom Game" mode and restoring from a save.

Here is a list of starting parameters to make it more clear what I’m talking about:
https://github.com/Gliese852/pioneer/blob/1ccd9a20e635924474b3831c5266bd49841ceef3/data/pigui/modules/new-game-window/layout.lua#L15

Gliese852:
I thought it would be useful to be able to also restore saves from the current version, for testing purposes or if an incompatible change occurs between releases. Therefore, I added this thing - if you hold Ctrl, the button forcibly changes to "Recover".

impaktor (commented):
Just went through the strings.

data/lang/ui-core/en.json
```diff
@@ -2131,6 +2195,10 @@
     "description": "Tab item in the start game menu",
     "message": "Summary"
   },
+  "SYSTEM_BODY_INDEX_IS_TOO_BIG": {
+    "description": "Message when restoring savegame",
+    "message": "System body index is too big"
```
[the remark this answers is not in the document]
Web-eWorks: I'd recommend `System body index is out of range` instead - because that's the "real" problem. It's an invalid index, not a too-large index.

Gliese852:
Rebased and removed conflicts.

JonBooth78:
I now see what you've done here - very nice.

I was playing around with this trying to load some old save games that were made unloadable during this dev cycle.

Turns out the flight log for the stations isn't getting loaded properly with graceful degradation as at least one of the stations appears to be on an invalid index planet.

(I was looking to see if I could help integrate this with the updated flightlog code...)

Attached:

[jump to pirates.zip](https://github.com/pioneerspacesim/pioneer/files/14000597/jump.to.pirates.zip)

Gliese852:
@JonBooth78 Thank you! I'll do something about it.

Gliese852:
Fixed the problem, but only in the recovery logic and the start menu - inside the game it will crash when going to the station log, since it does not know how to show "outdated" system paths. Haven’t touched the GUI in the game yet to avoid unnecessary conflicts with #5666.

data/pigui/modules/new-game-window/recovery.lua
```diff
@@ -0,0 +1,43 @@
...
+local Recovery = {}
+
+-- patch
+--    from - number (version)
+--    to - number (version)
+--    apply(gameDoc) -> errorString? - function that modifies whole gameDoc
+--
+--    it is assumed that the patches are arranged in the array in ascending
+--    order of versions, and do not intersect
+local Patches = {}
```
Web-eWorks, line 17: This should probably be made public, as I'd want to write savefile migration functions in their own file which `require()s` the `Recovery` module.

data/pigui/modules/new-game-window/ship.lua
```diff
@@ -735,14 +802,89 @@ function ShipEquip:fromStartVariant(variant)
...
+	end
+end
+
+ShipEquip.reader = Helpers.versioned {{
+	version = 89,
+	fnc = function(saveGame)
+		-- a table with all possible equipment is located directly in the saveGame
+		-- The ship's equipment set stores refs to this table items, and the
+		-- string IDs of the equipment are only in this table
+		local eqTable, errorString = Helpers.getByPath(saveGame, "lua_modules_json/@Equipment")
```
Web-eWorks, line 840: Note: with EquipSet v2 (hopefully this doesn't offend @laarmen - I believe there was only ever 1 EquipSet implementation...), equipment will be migrated to a new "persistent object" feature stored in `lua_modules_persistent`. (It will also need a new frontend for setting equipment when starting the game, but that's a different topic entirely).

data/pigui/modules/new-game-window/recovery.lua
```diff
@@ -0,0 +1,43 @@
...
+
+-- no patches yes
+
+-- the function returns the full game save document patched to the supported
+-- version or a string with an error
+function Recovery.loadDoc(saveName)
+	local gameDoc = Json.LoadSaveFile(saveName)
```
Web-eWorks, line 24: Note: sooner or later, this function will be required to post-process the save game and fix up all contained table references. The order and location a table value *definition* is serialized in is non-deterministic and any JSON value encoding a reference to a specific table value may potentially be the one in which that table is serialized to.

As a result, I'd recommend this function go through all `{ ref: number }` values serialized in the `lua_modules_json` (and eventually `lua_modules_persistent`) tables and perform a partial unserialization / unpickling, enough to ensure the savefile is traversable as a "normal" Lua table with the appropriate table identities preserved.

src/lua/LuaJson.cpp
```diff
@@ -34,10 +38,8 @@ static void _push_json_to_lua(lua_State *l, Json &obj)
 	case Json::value_t::array: {
 		lua_newtable(l);
 		size_t size = obj.size();
-		lua_pushinteger(l, size);
-		lua_setfield(l, -2, "n");
 		for (size_t idx = 0; idx < size; idx++) {
-			lua_pushinteger(l, idx);
+			lua_pushinteger(l, idx + 1);
```
Web-eWorks, line 42: Thanks for catching this - I hadn't actually exercised this codepath until writing the equipment rework.

src/lua/LuaSystemBody.cpp
```diff
@@ -148,15 +148,16 @@ static int l_sbody_attr_parent(lua_State *l)
...
-	// back to the starsystem proper to get what we need.
-	RefCountedPtr<StarSystem> s = Pi::game->GetGalaxy()->GetStarSystem(sbody->GetPath());
-	SystemBody *live_sbody = s->GetBodyByPath(sbody->GetPath());
-
-	if (!live_sbody->GetParent())
+	if (!sbody->GetStarSystem()) {
+		// orphan, its system has already been deleted, but SystemPath remains
+		// we'll make a new one just like it
+		RefCountedPtr<StarSystem> s = Pi::game->GetGalaxy()->GetStarSystem(sbody->GetPath());
+		sbody = s->GetBodyByPath(sbody->GetPath());
```
Web-eWorks, line 155: Note: `GetStarSystem` does not maintain a strong reference to the returned StarSystem (just a weak one so it can be removed from the master cache when all other references expire), so this system (AFAIK) will be deleted as soon as the `if` block scope is closed. This would orphan the just-created system body, leaving it still without parent.

If this is behaving differently than I've described, something's holding on to a reference to the star system that's been created and I retract my comment.

src/savegamedump.cpp
```diff
@@ -6,17 +6,31 @@
...
+
+	int indent = -1;
+	int shift = 0;
+
 	std::string filename = argv[1];
-	std::string outname = argc > 2 ? argv[2] : filename + ".json";
+	if (filename == "--pretty") {
+		indent = 2;
+		shift = 1;
+	}
```
Web-eWorks, line 29: Note: I prefer to run dumped savegames through https://jsonviewer.stack.hu/, but this is a good secondary option if your text editor is happy with highlighting 1MB+ JSON files.

Web-eWorks (changes requested):
I'm noticing a general pattern that `reader` functions do two things: they define a path in the savefile to unpickle data from, and then they convert the unpickled saved data into a "live" object.

I wonder if it'd be worthwhile to evolve this pattern to treat those as two separate processes, such that there is one function concerned only with loading unpickled data from the current save version, and there is another (versioned) function concerned with finding the data from the save game and unpickling it into the format currently expected by the game.

In a lot of cases, readers seem to boil down to a simple "passthrough" function which loads data from a (set of) specific key(s), unpickles it, and optionally transforms it before assigning to the parameter value. Some of this functionality could be made implicit, I think.

This is probably not a topic for this iteration of the save game restoration however, but I did want to record the idea for potential future improvements.

-----

For more on-topic discourse, I've left several review comments and change requests where there are things that I'd like to see addressed before merge. As this PR has a (somewhat) low bug surface and extremely high value for the upcoming release I'm willing to merge it even after we enter feature freeze if they can't be addressed before then, but "the sooner the better" is the motto of the remaining 8 days before release. :smile:

Apologies that this has taken so long to review - I've not had mental bandwidth available to deal with the review process for most of this week. In general, I have high confidence in the quality of code being written here, and since it's mostly touching code you've written yourself I don't have to worry about it not matching the existing style of (this particular corner of) the codebase. :smile:

Gliese852:
Made a rebase to resolve the conflict, minimally adapted to the new flightlog, it works, but there are 2 problems. Firstly, the entries in the game are grouped by type in violation of chronology. Secondly, the log inside the game cannot show a "non-existent" system/station path, an error occurs.

Gliese852:
Fixed the above 2 problems and it is now everything is working.

src/lua/LuaSystemBody.cpp
Gliese852: Yes, you're right! Probably nothing broke for me, because in fact I never managed to get into this branch when testing the game. Apparently, it is enough for the system to survive until the end of the function.

src/savegamedump.cpp
Gliese852: Fortunately, I don't experience any problems with large files (in particular json) in vim. And all my bindings with me!

data/pigui/libs/message-box.lua
```diff
@@ -5,49 +5,63 @@ local ui = require 'pigui'
...
+local font = ui.fonts.pionillium.medium
 
 local msgbox = {}
 
-msgbox.OK = function(msg)
+local function okButtonWidth()
+	local width = ui.withFont(font, function() return ui.calcTextSize("OK_BUTTON").x end)
+	okButtonWidth = function() return width end
+	return width
+end
```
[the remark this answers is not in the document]
Gliese852: Done!

data/pigui/modules/new-game-window/class.lua
```diff
@@ -279,6 +310,53 @@ end, function (_, drawPopupFn)
 	ui.withStyleColors({ PopupBg = ui.theme.colors.modalBackground }, drawPopupFn)
 end)
 
+function NewGameWindow.restoreSaveGame(selectedSave)
+	local saveGame = Recovery.loadDoc(selectedSave)
+	if type(saveGame) == 'table' then
+		Helpers.implicitVersion = saveGame.version
+		local paramErrors = {}
+		local report = string.format(lui.ACTUAL_SAVEGAME_VERSION, tostring(Game.CurrentSaveVersion()))
```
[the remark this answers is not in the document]
Gliese852: Done!

data/pigui/modules/new-game-window/crew.lua
```diff
@@ -220,25 +258,25 @@ local function crewTraits(lock, char, tableWidth)
 end
 
 function Crew:removeMember(member)
-	for k, v in pairs(self.value) do
-		if v == member.valueRef then
+	for k, testMember in pairs(self.value) do
```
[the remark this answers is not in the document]
Gliese852: Done!

data/pigui/modules/new-game-window/crew.lua
```diff
@@ -331,25 +368,76 @@ function Crew:drawMember(member)
...
+	self.list = Table.New("CrewListTable", false, {
+		columnCount = 1,
+		initTable = function(selfInternal)
+			ui.setColumnWidth(0, namesWidth)
+		end,
+		renderItem = function(selfInternal, memberEntry, key)
+			if memberEntry.value == 'ADD_NEW_CHARACTER' then ui.text("+") ui.nextColumn() return end
+
+			local spacing = ui.getItemSpacing()
+			ui.withStyleVars({ItemSpacing = spacing}, function()
```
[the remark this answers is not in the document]
Gliese852: Done!

data/pigui/modules/new-game-window/crew.lua
```diff
@@ -409,27 +468,36 @@ function Crew:draw()
...
+end
+
+Crew.reader = Helpers.versioned {{
+	version = 89,
+	fnc = function(saveGame)
+		local pickled, errorString = Helpers.getByPath(saveGame, "lua_modules_json/@ShipClass/$1")
+		if errorString then return nil, errorString end
+		local crew = Helpers.unpickle(pickled)
+		local value = {}
+		for i = 2, #crew do
```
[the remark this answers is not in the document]
Gliese852: Done!

data/pigui/modules/new-game-window/helpers.lua
```diff
@@ -0,0 +1,344 @@
...
+--  var - number, unused
+
+Helpers.iteratePickled = Helpers.versioned {{
+	version = 89,
+	fnc = function(pickled)
+		assert(pickled.table)
+		return function (a, _)
+			local k = a.tbl[a.i]
+			local v = a.tbl[a.i + 1]
+			if k or v then
```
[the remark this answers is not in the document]
Gliese852: Done!

data/pigui/modules/new-game-window/helpers.lua
```diff
@@ -0,0 +1,344 @@
...
+
+--
+-- Function: Helpers.getByPath
+--
+-- Looks for an element in the table, sequentially obtaining indices from the
+-- provided string and checking for their presence.
+--
+-- If the key starts with @, we look for it in the "pickled" table
+-- If the key looks like $N, we iterate through the "pickled" table N times and
+-- take the value (see Helpers.isPickled)
```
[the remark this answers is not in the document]
Gliese852: Done!

data/pigui/modules/new-game-window/location.lua
```diff
@@ -277,13 +289,77 @@ function Location:fromStartVariant(variant)
...
+	end
+	for _, childFrame in pairs(frame.child_frames) do
+		local sbi = getSystemBodyIndexForFrameID(id, childFrame)
+		if sbi then return sbi end
+	end
+end
+
+Location.reader = Helpers.versioned {{
+	version = 89,
+	fnc = function(saveGame)
```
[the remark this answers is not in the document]
Gliese852: Inserted blank lines, yes, that’s better.

data/pigui/modules/new-game-window/recovery.lua
Gliese852: Done!

data/pigui/modules/new-game-window/recovery.lua
Gliese852: I did the substitution of refs, but did not do the unpickling - all readers unpickle what they need themselves, is it worth redoing? It seems that the reader code will not be simplified by this.
Although, after resolving the refs, I can simplify the equipment reader - there was a search using refs, which is now not needed.

data/pigui/modules/new-game-window/location.lua
```diff
@@ -277,13 +289,84 @@ function Location:fromStartVariant(variant)
...
+		-- (Space::m_bodyIndex and Space::m_sbodyIndex) turn out to be the same
+		local dockedTo
+		dockedTo, errorString = Helpers.getPlayerShipParameter(saveGame, "ship/index_for_body_docked_with")
+		if errorString then return nil, errorString end
+
+		if dockedTo ~= 0 then
+			-- bodyIndex in system path is 1 less than index in Space::m_sbodyIndex,
+			dockedTo = dockedTo - 1
+			local starSystem = Location:getGalaxy():GetStarSystem(path)
+			local paths = starSystem:GetBodyPaths()
```
Web-eWorks, line 342: Unless you're using this `paths` variable somewhere else, I'd recommend using the attr `starSystem.numberOfBodies` instead - it doesn't push new SystemPath objects to Lua for every (potentially 70+) body in the system.

data/pigui/modules/new-game-window/location.lua
```diff
@@ -277,13 +289,84 @@ function Location:fromStartVariant(variant)
...
+
+		-- the indices of the bodies and the corresponding system bodies
+		-- (Space::m_bodyIndex and Space::m_sbodyIndex) turn out to be the same
+		local dockedTo
+		dockedTo, errorString = Helpers.getPlayerShipParameter(saveGame, "ship/index_for_body_docked_with")
+		if errorString then return nil, errorString end
+
+		if dockedTo ~= 0 then
+			-- bodyIndex in system path is 1 less than index in Space::m_sbodyIndex,
+			dockedTo = dockedTo - 1
```
Web-eWorks, line 340: This seems *very* prone to breakage and dependent on internal logic. At the least, I think this should have a longer documentation comment indicating the purpose of the `dockedTo` variable here and *why* it's needed to subtract it to match or not match the values contained in a private array member in C++.

Web-eWorks (changes requested):

data/lang/ui-core/en.json
```diff
@@ -1811,10 +1859,22 @@
     "description": "",
     "message": "Rating:"
   },
+  "READ_ANOTHER_SAVEGAME": {
+    "description": "Message when restoring savegame",
+    "message": "Read another savegame"
```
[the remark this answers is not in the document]
Gliese852: Oh, it turns out this line is no longer needed!

data/lang/ui-core/en.json
```diff
@@ -19,6 +23,10 @@
     "description": "",
     "message": "Allegiance:"
   },
+  "ALL_PARAMETERS_ARE_RESTORED_SUCCESSFULLY": {
```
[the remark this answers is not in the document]
Gliese852: Done!

data/lang/ui-core/en.json
```diff
@@ -7,6 +7,10 @@
     "description": "",
     "message": "Active"
   },
+  "ACTUAL_SAVEGAME_VERSION": {
+    "description": "Message when restoring savegame",
+    "message": "Actual savegame version: %s"
```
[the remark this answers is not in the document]
Gliese852: Named as 'Current'.

data/lang/ui-core/en.json
```diff
@@ -99,6 +107,14 @@
     "description": "",
     "message": "This item cannot be sold"
   },
+  "CANT_FIND_PATH": {
+    "description": "Message when restoring savegame",
```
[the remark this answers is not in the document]
Gliese852: Done!

data/lang/ui-core/en.json
```diff
@@ -263,10 +279,18 @@
     "description": "",
     "message": "Control Options"
   },
+  "COULDNT_FIND_SYSTEM_BODY_FOR_FRAME": {
```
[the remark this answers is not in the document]
Gliese852: Done!

data/lang/ui-core/en.json
```diff
@@ -347,6 +371,10 @@
     "description": "Time of departure, used in flight log",
     "message": "Departure date"
   },
+  "DESCRIPTION": {
+    "description": "Start game parameter",
```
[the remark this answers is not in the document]
Gliese852: Added e.g. in the description

data/lang/ui-core/en.json
```diff
@@ -487,6 +519,10 @@
     "description": "",
     "message": "Error"
   },
+  "ERROR_APPLY_PATCH": {
+    "description": "Message when restoring savegame",
+    "message": "Error apply patch"
```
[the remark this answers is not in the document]
Gliese852: Done!

data/lang/ui-core/en.json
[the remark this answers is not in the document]
Gliese852: Done!

data/pigui/modules/new-game-window/location.lua
Gliese852: Done!

data/pigui/modules/new-game-window/location.lua
Gliese852: Unfortunately, I didn't find another way. In my defense, it seems to me that the entire recovery system depends heavily on internal logic. And if something goes wrong, you will at least find yourself in the same system. Also added an explanatory comment.

Web-eWorks: That comment is quite satisfactory, thank you!

src/lua/LuaSystemBody.cpp
Web-eWorks: In general I think the code pattern of interacting with orphaned SystemBody objects should be strongly discouraged, due to the steep cost of re-generating the system on every access to the `parent` member of both the current orphaned body and the returned parent body. However, this is an acceptable solution to prevent total breakage.

data/pigui/modules/new-game-window/recovery.lua
Gliese852: Did a complete unpickling before calling the readers. Now only the fixupDoc function in the Recovery module is responsible for compatibility with the Lua serializer.

Gliese852:
I think I finished everything I wanted.

data/pigui/modules/new-game-window/recovery.lua
```diff
@@ -30,23 +30,37 @@ local fixupDoc = Helpers.versioned {{
...
-			end
-			if tbl.ref and not tbl.table then
-				tbl.table = refs[tbl.ref]
+			if cache[tbl] then return end
+			cache[tbl] = true
+			for k, v in pairs(tbl) do
+				if type(v) == 'table' then
+					if v.ref then
+						tbl[k] = refs[v.ref]
+					end
```
Web-eWorks, line 61: Note: serialized keys can also be table instances. The in-progress EquipSet v2 came close to using the pattern of serializing `{ [table] = true }` but eventually was refactored away. I'm unsure if we have other occurrences of that pattern.

data/pigui/modules/new-game-window/recovery.lua
```diff
@@ -30,23 +30,37 @@ local fixupDoc = Helpers.versioned {{
 
 		local refs = {}
 
+		local function unpickle(tbl)
+			local result = {}
+			tbl = tbl.table
+			for i = 1, #tbl, 2 do
+				result[tbl[i]] = tbl[i + 1]
```
Web-eWorks, line 37: This function could handle references as well:

```lua
local k, v = tbl[i], tbl[i + 1]
if type(k) == "table" and k.ref then k = cache[k.ref] end
if type(v) == "table" and v.ref then v = cache[v.ref] end
result[k] = v
```

It could even be handled another way to combine the `getRefs`, `putRefs`, and `unpickle` functions together:

```lua
local cache = {}

local function unpickle(v)
	if type(v) ~= 'table' or not v.ref then return v end
	if not v.table then return cache[v.ref] end

	local result = {}
	cache[v.ref] = result

	for i = 1, #v.table, 2 do
		local key = unpickle(v.table[i])
		local val = unpickle(v.table[i + 1])
		result[key] = val
	end

	return result
end
```

Web-eWorks (changes requested):

data/pigui/modules/new-game-window/recovery.lua
Web-eWorks: Creating readers is now less complicated and the pickling mechanism is now just an implementation detail - thanks for tackling that!

data/pigui/modules/new-game-window/recovery.lua
Gliese852: Indeed, there are such places, although not in the restored parameters.

data/pigui/modules/new-game-window/recovery.lua
Gliese852: The proposed version of the function will definitely not work, because the function will end on the first line - `gameDoc` does not have a `ref` field.
It is also assumed that the definition of a lua object will necessarily be encountered before other references to it, but this does not seem to be the case, at least because we are dealing with general Lua table at top level.

Web-eWorks: It was meant to illustrate the concept more than be tested/proven-working code. That being said, I see your point about iteration order; although pickled Lua tables internally require that the first encountered instance of a table value contains its definition when unserializing, that may not hold true when operating on recovered save data that's not going through the C++ serialization process.

data/pigui/modules/new-game-window/recovery.lua
Web-eWorks: Not in the recovered parameters *right now* at least. This may change at any time.

Gliese852: Now I take this into account!

data/pigui/modules/new-game-window/recovery.lua
Gliese852: I picked up some things from your code, but I still have to do two passes.

Gliese852:
This is a clean copy.

laarmen:
Finally, peace and quiet in my inbox!

Nice work folks, that's a nifty feature 🙂

PR #5707, opened 2024-01-06
Simplified travel times

robothauler:
This is an attempt to simplify delivery times so that they can be better balanced if necessary.

I have used the DeliverPackage module as a basis and fitted the other with one factor.
deliverpackage
cargorun
taxi
assassination

The combat module is now better for ships like the Bowfin.
combat

I have also added a "local part". This should calculate much better delivery times for large systems.
If no location is specified, a distance of 1AU is assumed.

**Now the local deliveries:**

I took the average flight time of a Xylophis from 'System Administration Resting' to 'OPLI Pax' as the basis for the local deliveries. approx. 2 days for 1AU.
local
scoop

Finally, the due time is rounded to 1h or 1/4h. In addition, the advert is deleted after half of the calculated travel time. Makes more sense in my opinion.

### DIFF
--- a/data/modules/Assassination/Assassination.lua
+++ b/data/modules/Assassination/Assassination.lua
@@ -9,6 +9,7 @@ local Comms = require 'Comms'
 local Timer = require 'Timer'
 local Event = require 'Event'
 local Mission = require 'Mission'
+local MissionUtils = require 'modules.MissionUtils'
 local NameGen = require 'NameGen'
 local Character = require 'Character'
 local Commodities = require 'Commodities'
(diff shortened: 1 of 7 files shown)


Web-eWorks (changes requested):
Please define the `travelTimeLocal` and `travelTime` functions in `data/modules/MissionUtils.lua` instead.

We already have some helpers for calculating mission parameters in that file, and the top-level `utils.lua` file is intended for general-purpose "language level" utilities.

Otherwise, I'm quite happy to see some effort be made to standardize balance across multiple mission modules and provide more tools to adjust and tweak the balance of the whole game "in one place"!

If you are willing to make additional changes, I'd like to see the mission duration and reward calculations used by each module unified into a common function (or set of functions) operating on some set of semantically well-defined variables similar to the current MissionUtils.Calculator helper.

impaktor:
Welcomed indeed. My gut feeling tells me this needs a bit of play testing to find edge cases, and might be better suitable for merge after pioneer-day, so only people building from source will be play-testing it, to find if there's some mission time limit that's unfair / doesn't make sense.

data/modules/Assassination/Assassination.lua
```diff
@@ -9,6 +9,7 @@ local Comms = require 'Comms'
 local Timer = require 'Timer'
 local Event = require 'Event'
 local Mission = require 'Mission'
+local MissionUtils = require "modules.MissionUtils"
```
Web-eWorks, line 12: As a matter of convention, we prefer to use single-quotes for module names, especially when omitting the parentheses to `require()`.

data/modules/MissionUtils.lua
```diff
@@ -173,5 +173,59 @@ function MissionUtils.GetNearbyStationPaths(system, range_ly, system_filter, sta
...
+--
+-- Parameters:
+--
+--   distance - the distance to the target in meters
+--
+-- Returns:
+--
+--   the travel time in seconds
+--
+function MissionUtils.travelTimeLocal(distance)
```
Web-eWorks, line 192: By convention, we use `PascalCase` for functions defined in a class or namespace like this, with the general idea that a leading capital letter in the module name requires a leading capital letter in the function name. `camelCase` names should be reserved for instance or class variables.

`utils` is a bit of a special case in that it's written in "Lua stdlib style" as it's meant to be a semi-portable extension to the Lua utility functions that isn't directly dependent on any game-specific code.

Web-eWorks (approved):
Looks good to me! I've left a few comments where the formatting of the code isn't uniform with our existing style conventions, but otherwise the code is approved.

Regarding balance, I've not yet played with it but the graphs look sensible. I'll make an effort to playtest this at some point, but I believe we can issue a post-Pioneer-day release if the balance is significantly negatively affected in a way that we don't catch.

PR #5708, opened 2024-01-08
improve planet rings shaders

Mc-Pain:
All rings are assumed to be icy

1) Each object in ring is darker depending of its phase
   (like Moon phases)
2) Each object shines brightly if sun is behind the ring
   (lot of low angle specular reflections)
3) Each object shines brightly if looking from the sun
   (simulating high geometric albedo > 1.0)

### DIFF
--- a/data/shaders/opengl/planetrings.frag
+++ b/data/shaders/opengl/planetrings.frag
@@ -28,11 +28,11 @@ void main(void)
 
 			// second term: diffuse mie (scattering through ring)
 			float g = 0.76f;
-			float phaseThrough = 3.f / (8.f * 3.141592) * ((1.f - g * g) * (1.f + mu * mu)) / ((2.f + g * g) * pow(1.f + g * g - 2.f * g * mu, 1.5f));
+			float phaseThrough = miePhaseFunction(g, mu);
 
 			// third term: reflect mie (imitate albedo >= 1.0)
 			float muRev = dot(-normalize(vec3(uLight[i].position)), eyenorm);
-			float phaseReflect = 3.f / (8.f * 3.141592) * ((1.f - g * g) * (1.f + muRev * muRev)) / ((2.f + g * g) * pow(1.f + g * g - 2.f * g * muRev, 1.5f));
+			float phaseReflect = miePhaseFunction(g, muRev);
 
 			col = col + texCol * (diffuse + phaseThrough + phaseReflect) * uLight[i].diffuse;
 		}
--- a/data/shaders/opengl/planetrings.frag
+++ b/data/shaders/opengl/planetrings.frag
@@ -7,6 +7,7 @@
 uniform sampler2D texture0;
 in vec2 texCoord0;
 in vec4 texCoord1;
+in vec4 varyingEyepos;
 
 out vec4 frag_color;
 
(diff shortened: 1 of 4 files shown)


Web-eWorks:
Snapshot_2024-01-08_20-44-24
Snapshot_2024-01-08_20-42-59
Snapshot_2024-01-08_20-42-01
Snapshot_2024-01-08_20-40-12

This looks *quite* good, excellent work! I have two concerns:
- First, there is a clear black "terminator" at the inner and outer edges of the overall ring structures. I assume this is probably caused by erroneous vertex colors being passed when generating the ring meshes. I'd like to see this resolved before merge.
- Secondly, there's a clear/sharp shadow terminator between shadowed portions of the ring and unshadowed portions. I'd like to see a softer shadow computation here, as the current terminator is not only a bit unnatural (ignores partially-shadowed light reaching the ring / gravitational lensing / edge-of-atmosphere light diffusion) but also suffers from a lack of anti-aliasing.
  This is not a concern that needs to be addressed before merge, but I would like to be improved post-merge if you are able.

data/shaders/opengl/planetrings.frag
```diff
@@ -16,10 +17,24 @@ void main(void)
...
+			float mu = dot(normalize(vec3(uLight[i].position)), eyenorm);
+			float diffuse = sqrt((1 - mu) / 2);
+
+			// second term: diffuse mie (scattering through ring)
+			float g = 0.76f;
+			float phaseThrough = 3.f / (8.f * 3.141592) * ((1.f - g * g) * (1.f + mu * mu)) / ((2.f + g * g) * pow(1.f + g * g - 2.f * g * mu, 1.5f));
+
+			// third term: reflect mie (imitate albedo >= 1.0)
+			float muRev = dot(-normalize(vec3(uLight[i].position)), eyenorm);
+			float phaseReflect = 3.f / (8.f * 3.141592) * ((1.f - g * g) * (1.f + muRev * muRev)) / ((2.f + g * g) * pow(1.f + g * g - 2.f * g * muRev, 1.5f));
```
Web-eWorks, line 35: I'd prefer `phaseReflect` and `phaseThrough` equations be consolidated to a single function parameterized on `mu` / `muRev`, with a comment indicating the equation's source and ideally with a link to the article or resource the Mie phase equation is derived from.

Web-eWorks (changes requested):

impaktor:
> Secondly, there's a clear/sharp shadow terminator between shadowed portions of the ring and unshadowed portions. I'd like to see a softer shadow computation here .... This is not a concern that needs to be addressed before merge

Indeed, this can be left to another day. But forgive me if I sound like an imbecile, but shadows in vacuum of space are very "hard", right?. So personally, I don't see that as an issue, (but I haven't tried the branch).

nozmajner:
@impaktor Shadows softness is more of a fiction of the size of the light source and the distance between the object and the surface where the shadow is cast. The bigger it is, the softer the shadows, and the further away the surface is, the softer as well. You can observ this even here on Earth: the shadow of a tall pole will be much sharper at the base compared to its top/furthest from the object. Similar to how they show the penumbra on those lunar eclipse explanations: 
Geometry_of_a_Lunar_Eclipse svg


The Sun is about 10 times larger than Saturn for example, and the shadow on the rings would look about like this: 
image
(Quickly put together in Blender, Saturn and Sun size and distance is to scale)

Web-eWorks:
Additionally, the sharpness of the shadow is partially because the shadowing condition is evaluated per screen pixel and produces a very "jagged" edge without any attempt at anti-aliasing.

Web-eWorks (approved):

Issue #5712, opened 2024-01-12
Reproducible crash shortly after jumping to next star, autopiloting to starbase


mikehgentry:
### Observed behaviour

Reproducible crash.

Let me know if crash reports from the Steam Deck aren't useful of course, but for now I'll assume it's helpful if they're reproducible.

### Expected behaviour

Gets to Red Cloud without crashing

### Steps to reproduce

From this saved game, fill up with conventional fuel at the station, launch, and hyperspace to Fomalhaut. It's right next to where you are.

When you arrive, set autopilot to dock at Red Cloud, and accelerate time to max. On my Deck it crashes every time.

It is possible to get to Red Cloud, by for example jumping somewhere else first, or by flying towards Fomalhaut for a bit before heading towards Red Cloud.

__My pioneer version (and OS):__

20231206 (from flatpak), SteamOS 3.5.7

__My output.txt (required) and game save (optional, but recommended)__
[_quicksave.txt](https://github.com/pioneerspacesim/pioneer/files/13924189/_quicksave.txt)
[output.txt](https://github.com/pioneerspacesim/pioneer/files/13924191/output.txt)
[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/13924194/opengl.txt)

mikehgentry:
[terminaloutput.txt](https://github.com/pioneerspacesim/pioneer/files/13924567/terminaloutput.txt)

Gliese852:
I think, in general, crashes from the steamdeck are useful, but the flatpak version is quite outdated (relative to the "unstable" branch), and this bug was fixed a long time ago, in #5564.

mikehgentry:
OK, thanks.

Not that I don't believe you, I'm just rather new to using flatpaks and confused: any idea why it says 20231206 on the title screen if it's an old build? 

The way I installed it was just to run "flatpak install net.pioneerspacesim.Pioneer" from a terminal

nozmajner:
Could you post a screenshot of the main menu? There were changes there in the last couple of months, so it whould help to see if it is a build of master or flathub just rebuilt it recently for some reason.

mikehgentry:
![372917_20240113154848_1](https://github.com/pioneerspacesim/pioneer/assets/1724168/e43e00a9-64e5-4f44-b7cd-29e8491fdaa7)

hexagonrecursion:
[...]

PR #5713, opened 2024-01-13
Hardcode the project version

Gliese852:
Since the build version is actually the date the build was run, if it is rebuilt on a different day it becomes erroneous. For example, since in any distribution the build into the image will most likely occur after the release, or even rebuild it regularly, the version will differ from the release one, which can be misleading.

There is also a second part of the version - the commit hash, but if the stable version is (re)built from the source archive (which is correct), there is no git repo at all.

Therefore, it is proposed to freeze this manually (with a script) in CMakeLists, so that it unambiguously indicates the version of the game.


impaktor:
Ping @pcercuei 
(this came up with people playing the new flatpak build)

pcercuei:
I disagree about hardcoding the values. This is the worst solution.

What I'd suggest - get the latest tag that was created, which is always the build date.
```
git describe --abbrev=0 --match 20*
```

Of course, make sure that this tag pattern is reserved for releases.

pcercuei:
Or, since you don't like to annotate your release tags,
```
git describe --tags --match 20*
```

Gliese852:
@pcercuei the problem is that, as already written, the original git repo is not always available. For example, in a flatpak, do you see this tag when building?

Gliese852:
I would like to add that, as far as I know, Fedora, OpenSUSE, and ALT Linux build Pioneer from the source archive and will not see this tag. Moreover, no one guarantees that any service tag in the internal package repository will not match the proposed glob.

Gliese852:
Hmm, the [screenshot](https://github.com/pioneerspacesim/pioneer/issues/5712#issuecomment-1890532884) from the guy who uses the flatpak version, you can clearly see that there is no git repo during the build (no hash).

Also I thought, it seems it would be better to hardcode only the "date-version", but get (or not) the commit hash during the build. This will give us more information about the origins of a particular version and may prevent some unpleasant surprises.

impaktor:
This change is of no use if package maintainers don't know about it. Therefore: might I suggest a section in README.txt, called "Package maintainers", or "Versioning packages", or similar, that explains where/how to get the appropriate version?

CMakeLists.txt
```diff
@@ -118,7 +118,7 @@ if (GIT_FOUND AND EXISTS ${PROJECT_SOURCE_DIR}/.git)
 	)
 endif()
 
-string(TIMESTAMP PROJECT_VERSION "%Y%m%d")
+set(PROJECT_VERSION 20240113)
```
Web-eWorks, line 121:
```suggestion
set(PROJECT_VERSION 20240113
	CACHE STRING "Version identifier for current compiled build")

if (NOT PROJECT_VERSION_GIT)
	string(TIMESTAMP PROJECT_VERSION_GIT "built %Y%m%d")
endif()
```

Web-eWorks (changes requested):
This is a good idea in theory - in practice there are a few additional considerations:

1. The version number needs to be correct "out of the box" for distribution packaging specifically. We cannot rely on every package maintainer knowing the correct invocation to ensure a consistent version number is displayed on the main screen. This is not a concern for "single user" builds, as presumably if a user is making a bug report from their own source build, they're technically competent enough to correctly report the commit or branch they're building from.
2. We need some way to clearly indicate non-release builds as such, especially when it comes to CI builds (where we may want to e.g. version a build as `devel-pr#5600 (f19c0a13)`). Any version string should be configurable via CMake command line parameter.
3. Ideally, in any situation where a git repo is available, the secondary build information should display the git commit hash and/or specific platform/build configuration set at build time. E.g. `20240203 (f129104912 x64+avx2)`. If not available, a fallback to the build timestamp may be appropriate.

I've left a suggested change that would allow specifying custom build versions for our CI workflows, as well as falling back to timestamp if there isn't an explicitly set version info string.

I don't think there's a need for an `update-version` script at all, given that we will only be modifying the value defined in CMakeLists.txt when tagging a new release (and most likely in the tagged commit which constitutes that release.)

I'd suggest it might be a good idea to move the definitions of `PROJECT_VERSION` and `PROJECT_VERSION_GIT` all the way to the top of CMakeLists.txt for visibility. This would make it easier to append to the `PROJECT_VERSION_GIT` variable (really, the build info string) with specific build options as they're defined.

Web-eWorks:
> Or maybe we can go a little further and generalize PIONEER_VERSION_GIT to PIONEER_VERSION_AUX or something like that? 

I'd recommend adding a new variable `PIONEER_VERSION_INFO` which has the git commit information appended to it if present (and also specific build types as needed).

Gliese852:
@Web-eWorks 

> 1. The version number needs to be correct "out of the box" for distribution packaging specifically. We cannot rely on every package maintainer knowing the correct invocation to ensure a consistent version number is displayed on the main screen. This is not a concern for "single user" builds, as presumably if a user is making a bug report from their own source build, they're technically competent enough to correctly report the commit or branch they're building from.

If the maintainer does not specify anything, it will be just a version (without additional information in brackets). The absence of an additional version will already tell us that this is just a release archive, and there was no git repository. If the maintainer also additionally indicates useful information in brackets, thank him, but this is not critical.

> 3\. Ideally, in any situation where a git repo is available, the secondary build information should display the git commit hash and/or specific platform/build configuration set at build time. E.g. `20240203 (f129104912 x64+avx2)`. If not available, a fallback to the build timestamp may be appropriate.

That is, by default, it will be like this: `20240203 (built 20241010)` but is it worth it to overload the version string like that? What does the build date tell us?

> I don't think there's a need for an `update-version` script at all, given that we will only be modifying the value defined in CMakeLists.txt when tagging a new release, and specifically setting that value to the tag name for the release (and most likely in the tagged commit which constitutes that release.)

Why not update this more often? For those who build unstable packages, for example. If they also have their own version history, there will be a timestamp of the latest release (e.g. half a year ago) and some hash unknown to us. We'll have to suffer again.

Web-eWorks:
> That is, by default, it will be like this: 20240203 (built 20241010) but is it worth it to overload the version string like that? What does the build date tell us?

On second thought, the build timestamp is probably not as useful as I was thinking when I originally proposed that idea. I'd like to keep the commit hash  + build flags however in any case where that information is available.

> Why not update this more often? For those who build unstable packages, for example. If they also have their own version history, there will be a timestamp of the latest release (e.g. half a year ago) and some hash unknown to us. We'll have to suffer again.

I don't see a need to do this, based on our current and past volume of bug reports.

All of our bug-reporting traffic on unstable versions of Pioneer in the last year that I'm aware of came from (in order of volume):

1. self-compiled checkouts of the game code, which have valid git hashes (unless they're actively making commits and developing the game);
2. users who've installed `pioneer-git` from the AUR, which has valid git hashes because it's a direct compilation of our sources without patches;
3. users downloading our CI build artifacts, which will have extremely detailed version information set directly from our CI workflow script.

To the best of my knowledge there are no distributions with an unstable build of Pioneer other than Arch's `pioneer-git` which are responsible for more than 1-3 bug reports _per year_. For the above reasons, I don't see any need to modify the "stable version number" on our end outside of officially-tagged pre-releases / release candidate builds.

 I think there's a danger here of trying to "be all things to all people" in a way that will consume far more of our developer time than it would save in terms of bug report clarity. For a hardcoded "unstable version number" to be effective, we would have to be changing the version number after every single PR merge or commit made to the master branch.

At most, I'd suggest authoring one commit per release immediately after the tagged commit which changes `PROJECT_VERSION` to something like `"20240203-develop"` - that would indicate it's an unstable version and provide some general timeframe for what commit the binary is built from.

Gliese852:
Yes, we should probably just release more often, beta releases or something like that.

CMakeLists.txt
Gliese852: Done, except for the default timestamp.

CMakeLists.txt
```diff
@@ -4,6 +4,12 @@ project(pioneer LANGUAGES CXX)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(CMAKE_CXX_STANDARD 17)
 
+set(PROJECT_VERSION 20240119
+	CACHE STRING "Version identifier for current compiled build")
```
Web-eWorks, line 8:
```suggestion
set(PROJECT_VERSION 20240119-develop
	CACHE STRING "Version identifier for current compiled build")
```

Web-eWorks (approved):
I'd recommend adding the `-develop` suffix for non-release builds, but this looks good. Once squashed and merged I'll look at modifying the CI workflows to annotate with build information.

CMakeLists.txt
Gliese852: Made a version as discussed later.

Gliese852:
Forced it all together into one clean commit.

impaktor:
@pcercuei We just published a new release, do you have time to fix a new flatpak for us, pretty please?

PR #5717, opened 2024-01-18
Fix  assert on startup when calling back on an empty string

### LINKED ISSUE #5716: "Assertion failed!" when the game is loading
### Observed behaviour
On current master when the game is loading, when it reaches 9.1 bil years, it will show this error shortly after
AssertionFailed
After checking out several times it seems like 682109718cbe3d98852bbb9c718ee0abb9669925 is causing this error, because when cheking out to a463999301d33c68f678e03ba76b0a7e2c3cc5bc just before 682109718cbe3d98852bbb9c718ee0abb9669925, the game loads successfully.

### Call stack
pioneer.exe!std::string::back() Line 3873	C++
pioneer.exe!path_to_module(std::string path) Line 149	C++
pioneer.exe!get_caller_module_name(lua_State * L, int depth) Line 456	C++
pioneer.exe!l_get_module_name(lua_State * L) Line 486	C++
pioneer.exe!luaD_precall(lua_State * L, lua_TValue * func, int nresults) Line 318	C
pioneer.exe!luaV_execute(lua_State * L) Line 709	C
pioneer.exe!luaD_call(lua_State * L, lua_TValue * func, int nResults, int allowyield) Line 396	C
pioneer.exe!lua_callk(lua_State * L, int nargs, int nresults, int ctx, int(*)(lua_State *) k) Line 906	C
pioneer.exe!LuaConsole::RegisterAutoexec() Line 175	C++
pioneer.exe!LuaConsole::LuaConsole() Line 58	C++
pioneer.exe!StartupScreen::Start::__l2::<lambda>() Line 591	C++
[External code]	
pioneer.exe!StartupScreen::RunNewLoader() Line 632	C++
pioneer.exe!StartupScreen::Update(float deltaTime) Line 610	C++
pioneer.exe!Application::Run() Line 195	C++
pioneer.exe!SDL_main(int argc, char * * argv) Line 183	C++
[External code]	

### __My output.txt
[output.txt](https://github.com/pioneerspacesim/pioneer/files/13931329/output.txt)

JonBooth78:
Fixes #5716

### DIFF
--- a/src/lua/core/Import.cpp
+++ b/src/lua/core/Import.cpp
@@ -146,7 +146,7 @@ static std::string path_to_module(std::string path)
 		end = path.find_first_of('/', start);
 	}
 
-	if (module_name.back() == '.')
+	if (!module_name.empty() && module_name.back() == '.')
 		module_name.pop_back();
 
 	return module_name;
--- a/src/lua/core/Import.cpp
+++ b/src/lua/core/Import.cpp
@@ -24,42 +24,42 @@
 #endif
 
 /**
-*	We use Lua's require() functionality, but with a twist.
-*
-*	When specifying a module name to `require()`, beginning the module name
-*	with a separator (`.`) is syntactic sugar for prepending the source
-*	directory of the calling file to the module name, forming a system of
-*	relative name lookup.
-*
-*	Relative name lookup cannot ascend levels in the filesystem, nor can it be
-*	used to exit the virtual filesystem. It's merely a shorthand to implicitly
-*	specify the module name of the calling file as a prefix to the module name
-*	passed to `require()`. This is done as follows:
-*
-*		-- when called from myname/modname/mod-load.lua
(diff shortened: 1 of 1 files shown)


Max5377:
The question is, why `module_name` is empty in the first place when it should normally return `lua` module name?

Web-eWorks:
The module name is empty because `LuaConsole.cpp` is directly calling `Event.Register(...)` and the calling module context requested from `Event.Register` is a C function which has no valid filepath and therefore cannot be translated into a module name.

This technically exposes a flaw in the new Event system where C++ can only register one Lua callback handler for each event type without overwriting the previously-set handler. That's not a problem in practice (as generally speaking a C++ function shouldn't be listening for a Lua event but rather use a sigc::signal instead), but something I'll have to keep in mind for the future.

Web-eWorks (approved):

Issue #5178, opened 2021-05-13
Saved game does not save, in Sector map, hyperjump steps and setting visualization choices


tropicalpioneer:
### Observed behaviour
The Saved game will not save, in the sector map:
a) the hyperjump planned steps 
b) the settings ticks, for example will always tick "Label uninhabited system" 

### Expected behaviour
planning hyperjump is some work that you don't want to re-do when loading your game

### Steps to reproduce

__My pioneer version (and OS):__
Pioneer downloaded and installed at the end of April 2021
Linux Mint 20.1 Cinnamon

tropicalpioneer:
Just noticed this 2018 thread in the dev forum about _hyperspace route planner_ https://pioneerspacesim.net/forum/viewtopic.php?f=3&t=546

Bodasey:
actual master observation: Hyperjumps are saved and loaded, but not recognized when you open the sector map for the first time:
[Master_Paralleltest_vor_Sprung.txt](https://github.com/pioneerspacesim/pioneer/files/9166917/Master_Paralleltest_vor_Sprung.txt)
Accelerate and wait until the saved hyperjump to Luyten is possible - or wait until the jump expires. (why?)

Issue #5187, opened 2021-05-16
Bulletin Board - Fast-forward generates similar missions on same destinations


tropicalpioneer:
### Observed behaviour
Fast forwarding has generated the second half of listed missions in the screenshot. 
Some of the missions are same type and destination, still different in payment and number of elements ( i.e. fuel needed or crew mobility).
![issue generated missions](https://user-images.githubusercontent.com/84130826/118395867-36b70d00-b67f-11eb-927b-eebb6c1a173d.png)
### Expected behaviour
It is of course convenient to earn multiple amounts of money going to the same place :) , but the pseudo-redundantly generated missions are probably unwanted behavior.

### Steps to reproduce
fast forward time in bulletin board

__My pioneer version (and OS):__
20210205
Mint Cinnemon 20.1

impaktor:
So is this specifically about the red icon missions? There can not be more than one emergency per system?

Not sure I see a bug here. 

@clausimu this works as you intended the Search And Rescue missions? If so, we can close this.

Web-eWorks:
I believe the unintended behavior is mostly the high concentration of search and rescue missions, although this might be a starport such that the other mission types don't have good targets to generate nearby. Ultimately this is the correct behavior, adverts are supposed to regenerate over time and there is no particular concern as to how (or how well) it covers the "surrounding space".

impaktor:
Right, I'll consider nerfing the spawn rate some, if it's generating too quickly relative to the others

impaktor:
OK, I've considered it, and I'm not going to spend any time on this -> closing.

PR #5721, opened 2024-01-21
Save and load the current planned hyperjump in the save file

JonBooth78:
The planned hyperspace route, with all stops is saved along with the save game, and restored on load.

### DIFF
--- a/data/pigui/views/map-sector-view.lua
+++ b/data/pigui/views/map-sector-view.lua
@@ -17,6 +17,8 @@ local Color = _G.Color
 local ui = require 'pigui'
 local layout = require 'pigui.libs.window-layout'
 
+local Serializer = require 'Serializer'
+
 local player = nil
 local colors = ui.theme.colors
 local icons = ui.theme.icons
@@ -39,10 +41,16 @@ local buttonState = {
 	[false]       = ui.theme.buttonColors.transparent
 }
 
-local draw_vertical_lines = false
-local draw_out_range_labels = false
-local draw_uninhabited_labels = true
-local automatic_system_selection = true
+local settings = 
+{
+	draw_vertical_lines=false,
+	draw_out_range_labels=false,
+	draw_uninhabited_labels=true,
+	automatic_system_selection=true
+}
+
+local loaded_data = nil
+
 
 local function textIcon(icon, tooltip)
 	ui.icon(icon, Vector2(ui.getTextLineHeight()), svColor.FONT, tooltip)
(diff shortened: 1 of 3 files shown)


impaktor:
This fixes #5178 ?

JonBooth78:
@impaktor , only half of it; I'll get onto the rest later - thanks for pointing it out

Web-eWorks (commented):
Especially given that you're planning on serializing more sector view state, I'd recommend that you marshal the serialized data through a `Serializer:Register()` call in `map-sector-view.lua` instead of registering the serialization callback in `hyperjump-planner.lua`.

The hyperjump planner / sector view split is slightly unconventional / unfinished in that the hyperjump planner is not quite a re-usable class that the sector map owns an instance of (yet), but it's close enough that I'd recommend applying that same mental model.

This is not strictly required, but it will make it easier to clean up / finish encapsulating the hyperjump planner widget in the future.

Web-eWorks:
Hmm... the more I look at this code the more I think the serialization should be entirely handled in `map-sector-view.lua`. 99% of this code is manipulating the C++ sector view object itself, and the only thing that involves the hyperjump planner widget is a one-off call in an onGameStart event to `hyperjumpPlanner.buildRouteList()`.

JonBooth78:
@Web-eWorks , I think your right; I also suspect that the call to `hyperjumpPlanner.buildRouteList()` is also redundant as I think it gets called on entry to the screen; I just wanted to be super-safe.

I felt that as this was all to do with the route-planner, it made sense to live in here, although I agree, it really has little to do with that element, which is a UX one.

I'll have a crack at moving it to the suggested place and add the other missing bits of #5187

JonBooth78:
@Web-eWorks , ready for re-review, all seems to be complete and working, with a bonus bug-fix thrown in at the end :-)

JonBooth78:
oh snap, take that back found a bug

JonBooth78:
OK, fixed that bug (doesn't look like I introduced it, oddly) and then clang formatted a file I touched on the way through. @Web-eWorks now ready to go

data/pigui/views/map-sector-view.lua
```diff
@@ -17,6 +17,8 @@ local Color = _G.Color
 local ui = require 'pigui'
 local layout = require 'pigui.libs.window-layout'
 
+local Serializer = require 'Serializer'
```
Web-eWorks, line 20: Teensy tiny nitpick but we generally prefer to require "core" modules like Serializer, Game, Event etc. at the top of the file before other requires. In this case, I'd move the `require 'Serializer'` statement to just after (or before, if you want to sort semi-alphabetically) `require 'SystemPath'`.

src/SectorView.cpp
```diff
@@ -24,9 +20,13 @@
 #include "galaxy/Galaxy.h"
 #include "galaxy/Sector.h"
 #include "galaxy/StarSystem.h"
+#include "lua/LuaObject.h"
 #include "lua/LuaRef.h"
 #include "lua/LuaTable.h"
 #include "matrix4x4.h"
+#include <assert.h>
+#include <cmath>
+#include <sstream>
```
Web-eWorks, line 29: Since clang-format is already nitpicking this, I'd move these includes down by one line so they're contiguous with the `#include <unordered_set>` statement and separated from `#include "matrix4x4.h".

Web-eWorks (approved):
Two very minor tiny nitpicks but the code looks good otherwise. Haven't had time to test this branch, so it's provisionally approved pending a playtest sometime in the next 24 hours.

Web-eWorks:
> OK, fixed that bug (doesn't look like I introduced it, oddly) and then clang formatted a file I touched on the way through.

This is quite the ancient bug oddly enough (4 years old!) but was completely benign up until yesterday when I merged #5724. I had the feeling something was odd with the `onEnterSystem` callback signature but rationalized it away due to lack of time. Thanks for catching that!

PR #5722, opened 2024-01-22
Add BUILD_DATE variable for windows builds

Gliese852:
This is needed for the FILEVERSION constant because it only accepts numeric values, and now our version can be not only numerical.
Please note that I did not change this line:
```
VALUE "ProductVersion",     "@PROJECT_VERSION@"
```

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -11,6 +11,7 @@ on:
     paths:
     - 'src/**.cpp'
     - 'src/**.h'
+    - '**/CMakeLists.txt'
   release:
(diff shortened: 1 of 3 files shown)


impaktor:
Did we want to put a new section in COMPILING.txt on how to get version, for packagers? 
(I'm afraid I can't do it, as I'm not grasping the details, only why it's might be needed)

Web-eWorks:
CC @JonBooth78 / @fluffyfreak given this is a Windows build specific issue with compiling the .RC file.

Web-eWorks (approved):

PR #5727, opened 2024-01-29
Add tracks by diduuz

WKFO:
New music tracks by diduuz on the dev forum. I'm merely handling the git part. Thanks are directed to him for the contributions.

Forum thread: https://forum.pioneerspacesim.net/viewtopic.php?f=9&t=637&start=40

- 'distant_home v3' for space music
- 'insignificance_v2' for unexplored space / hyperspace music
- 'floating' as discovery music

The categorization was my idea, but as always, is open for suggestions. I made it considering the feel of current tracks already in place in those categories. I didn't know what I could do with another track by the name of 'focus_calm', leaving it to be polished for its final form, as I said in the forum.


Web-eWorks:
While I had originally suggested that `insignificance_v2` could use some more polish, I'm fine to include it if it's understood to be receiving additional work / mastering effort post-release at some point.

`floating` on the other hand I don't think is a good fit for the discovery category; it's a very tense track with horror music undertones. I'd strongly prefer that track receive at least one more major revision before it's considered for inclusion. I'd actually rather include `focus_calm` over `floating` at this stage, as `focus_calm` in its current state is *far* more suited for discovery music in my eyes.

WKFO:
> `floating` on the other hand I don't think is a good fit for the discovery category; it's a very tense track with horror music undertones.

That's exactly why I thought it'd "feel at home" in 'discovery'.

IMHO `focus_calm` can be best added into unexplored/hyperspace, if it was to fit into any category.

impaktor:
@WKFO He has updated the tracks on the forum thread.

PR #5731, opened 2024-02-02
Have MSVC Builds show branch information from git in the menu

JonBooth78:
This should make the build information parsed from the git ref of the build show on the main menu

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -56,13 +56,14 @@ jobs:
 
     - name: Build MSVC
       shell: cmd
-      run: |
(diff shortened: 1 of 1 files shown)


Web-eWorks:
From the limited digging I did:
- cmd.exe doesn't parse single-quote characters, so everything has to be escaped double-quotes, including the escaped quotes in the awk expression. (seems that works already)
- The shell script specified in the workflow file is written to disk and invoked via `call <script-path>`, so the `%i` in `for` needs to be `%%i`. See https://ss64.com/nt/for_cmd.html.

PR #5737, opened 2024-02-03
Adding editor.txt

### LINKED ISSUE #5736: couldn't run a editor
### Observed behaviour
run a editor

### Expected behaviour
the editor closes automatically

### Steps to reproduce
1.open editor app

Windows 11
Pioneer version: 2024-02-03

nozmajner:
I've copied over the modelviewer help from the modelviewer.txt, to editor.txt for easier discoverability. 
Also added a short overview for the system editor.
I've kept the modelviewer.txt in place for those who are used to the old pioneer -mv way, with a short text about its new home.

Fixes #5736 (In a sense :P )


editor.txt
```diff
@@ -0,0 +1,107 @@
...
+Without any parameters, the editor will not start.
+	
+		System editor
+		~~~~~~~~~~~~
+		
+This is a gui tool for editing and authoring json formatted star systems.
+
+Usage:
+~~~~~~
+editor -system
```
Web-eWorks, line 13: `editor --system` - while it might technically function with a single hyphen here (for backwards compatibility with the modelviewer flag), POSIX style requires the use of two hyphens when the flag contains more than a single character and thus I make no guarantees this remains compatible.

Web-eWorks (changes requested):

impaktor:
I would suggest removing Modelviewer.txt completely.

nozmajner:
Okay, will do

editor.txt
```diff
@@ -0,0 +1,107 @@
...
+end-user use.
+
+Usage:
+~~~~~~
+editor -mv <model>
+
+<model>		- name of the model defined in lua
+
+	Example:
+	./editor -mv interdictor
```
impaktor, line 69: For symmetry, I'd suggest 
```
    Example:
    ./editor --modelviewer
	./editor -mv interdictor
```

Come to think of it, does passing model name as argument work anymore? (just tried with `Dome1`, and `cargo` to no avail).

impaktor (commented):
I think we should also mention that the editor binary logs to editor.txt instead of output.txt

editor.txt
impaktor: Easier to change documentation than the code :)

(Classic example: "map is wrong, the giant rock is on the wrong side of the road" - move rock to where map said)

Web-eWorks: You should be able to run `editor --modelviewer=<modelname>`, supporting the `editor -mv <modelname>` case is something that I can restore support for.

Web-eWorks (approved):

PR #5743, opened 2024-02-04
Fix path of pioneer executable

reinerh:
PIONEER_EXECUTABLE is used in the .desktop file, but with CMAKE_INSTALL_FULL_BINDIR, it always points to /usr/bin, even when a different binary path has been configured via InstallPioneer.cmake. The actual path where the binary will get installed is in PIONEER_INSTALL_BINDIR.


Web-eWorks:
Thanks!

PR #5748, opened 2024-02-07
Refactor and rework random system generator

Mc-Pain:
Current implementation allows up to *four* stars in the system with only layout of "two binary systems orbiting each other"

- [x] Clean up the code
- [x] Work out on star layouts
- [x] Remove upper limit of stars being generated (separate PR)

### DIFF
--- a/src/galaxy/StarSystemGenerator.cpp
+++ b/src/galaxy/StarSystemGenerator.cpp
@@ -1274,6 +1274,21 @@ SystemBody* StarSystemRandomGenerator::MakeSystemBody(RefCountedPtr<StarSystem::
 	return newBody;
 }
 
+SystemBody* StarSystemRandomGenerator::MakeGravFor(RefCountedPtr<StarSystem::GeneratorAPI> system, const std::string &systemName, SystemBody * const body1, SystemBody * const body2)
+{
+	const std::string suffix1 = body1->m_name.substr(systemName.size() + 1);
+	const std::string suffix2 = body2->m_name.substr(systemName.size() + 1);
+	const std::string suffixGrav = suffix1 + ',' + suffix2;
+
+	SystemBody *grav = MakeSystemBody(system, 0, systemName, suffixGrav.c_str());
+	grav->m_type = SystemBody::TYPE_GRAVPOINT;
+	grav->m_mass = body1->GetMassAsFixed() + body2->GetMassAsFixed();
+	grav->m_children.push_back(body1);
+	grav->m_children.push_back(body2);
+
+	return grav;
+}
+
 bool StarSystemRandomGenerator::Apply(Random &rng, RefCountedPtr<Galaxy> galaxy, RefCountedPtr<StarSystem::GeneratorAPI> system, GalaxyGenerator::StarSystemConfig *config)
 {
 	PROFILE_SCOPED()
@@ -1299,10 +1314,6 @@ bool StarSystemRandomGenerator::Apply(Random &rng, RefCountedPtr<Galaxy> galaxy,
 		system->SetRootBody(star[0]);
 		system->SetNumStars(1);
 	} else {
-		centGrav1 = MakeSystemBody(system, 0, systemName, "A,B");
-		centGrav1->m_type = SystemBody::TYPE_GRAVPOINT;
-		system->SetRootBody(centGrav1);
-
 		SystemBody::BodyType type = secSys.GetStarType(0);
 		star[0] = MakeSystemBody(system, centGrav1, systemName, "A");
 		MakeStarOfType(star[0], type, rng);
@@ -1310,9 +1321,9 @@ bool StarSystemRandomGenerator::Apply(Random &rng, RefCountedPtr<Galaxy> galaxy,
 		star[1] = MakeSystemBody(system, centGrav1, systemName, "B");
 		MakeStarOfTypeLighterThan(star[1], secSys.GetStarType(1), star[0]->GetMassAsFixed(), rng);
 
-		centGrav1->m_mass = star[0]->GetMassAsFixed() + star[1]->GetMassAsFixed();
-		centGrav1->m_children.push_back(star[0]);
-		centGrav1->m_children.push_back(star[1]);
+		centGrav1 = MakeGravFor(system, systemName, star[0], star[1]);
+		system->SetRootBody(centGrav1);
+
 		// Separate stars by 0.2 radii for each, so that their planets don't bump into the other star
 		const fixed minDist1 = (fixed(12, 10) * star[0]->GetRadiusAsFixed() + fixed(12, 10) * star[1]->GetRadiusAsFixed()) * AU_SOL_RADIUS;
 	try_that_again_guvnah:
@@ -1334,33 +1345,27 @@ bool StarSystemRandomGenerator::Apply(Random &rng, RefCountedPtr<Galaxy> galaxy,
 				centGrav2 = star[2];
 				system->SetNumStars(3);
 			} else {
-				centGrav2 = MakeSystemBody(system, 0, systemName, "C,D");
-				centGrav2->m_type = SystemBody::TYPE_GRAVPOINT;
-				centGrav2->m_orbMax = 0;
-
 				star[2] = MakeSystemBody(system, centGrav2, systemName, "C");
 				MakeStarOfTypeLighterThan(star[2], secSys.GetStarType(2), star[0]->GetMassAsFixed(), rng);
 
 				star[3] = MakeSystemBody(system, centGrav2, systemName, "D");
 				MakeStarOfTypeLighterThan(star[3], secSys.GetStarType(3), star[2]->GetMassAsFixed(), rng);
 
+				centGrav2 = MakeGravFor(system, systemName, star[2], star[3]);
+				centGrav2->m_orbMax = 0;
+
 				// Separate stars by 0.2 radii for each, so that their planets don't bump into the other star
 				const fixed minDist2 = (fixed(12, 10) * star[2]->GetRadiusAsFixed() + fixed(12, 10) * star[3]->GetRadiusAsFixed()) * AU_SOL_RADIUS;
 				MakeBinaryPair(star[2], star[3], minDist2, rng);
-				centGrav2->m_mass = star[2]->GetMassAsFixed() + star[3]->GetMassAsFixed();
-				centGrav2->m_children.push_back(star[2]);
-				centGrav2->m_children.push_back(star[3]);
 				system->SetNumStars(4);
 			}
-			SystemBody *superCentGrav = MakeSystemBody(system, 0, systemName, nullptr);
-			superCentGrav->m_type = SystemBody::TYPE_GRAVPOINT;
+			SystemBody *superCentGrav = MakeGravFor(system, systemName, centGrav1, centGrav2);
+			superCentGrav->m_name = systemName;
 			centGrav1->m_parent = superCentGrav;
 			centGrav2->m_parent = superCentGrav;
 			system->SetRootBody(superCentGrav);
 			const fixed minDistSuper = star[0]->m_orbMax + star[2]->m_orbMax;
 			MakeBinaryPair(centGrav1, centGrav2, 4 * minDistSuper, rng);
-			superCentGrav->m_children.push_back(centGrav1);
-			superCentGrav->m_children.push_back(centGrav2);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/galaxy/StarSystemGenerator.cpp
```diff
@@ -1259,6 +1259,37 @@ void StarSystemRandomGenerator::MakeBinaryPair(SystemBody *a, SystemBody *b, fix
...
+	newBody->m_name = systemName;
+	if (bodySuffix != nullptr) {
+		newBody->m_name += " ";
+		newBody->m_name += bodySuffix;
+	}
+
+	return newBody;
+}
+
+SystemBody* StarSystemRandomGenerator::MakeGravFor(RefCountedPtr<StarSystem::GeneratorAPI> system, const std::string &systemName, SystemBody * body1, SystemBody * body2)
```
fluffyfreak, line 1274: Whilst you're refactoring, could you rename this to `MakeGravPointForBodies` please?

Mc-Pain:
Looks messy
изображение
изображение

Mc-Pain:
изображение

Mc-Pain:
removed WIP: sign, third task would be completed through separate PR

fluffyfreak (approved):
I can't spot any coding issues, although the ULGY HACK part could be nice to cleanup

Mc-Pain:
> I can't spot any coding issues, although the ULGY HACK part could be nice to cleanup

I don't know how to make it pretty, it generates an i-th uppercase letter

fluffyfreak:
Does this change the universe generation at all?

Mc-Pain:
> Does this change the universe generation at all?

In what terms?

fluffyfreak:
> In what terms?

Naming of systems, numbers of stars/planets/etc, anything like that or is it purely a code reorganisation?

Mc-Pain:
> > In what terms?
> 
> Naming of systems, numbers of stars/planets/etc, anything like that or is it purely a code reorganisation?

Number of planets could vary. Number of stars for each system is done in sector generator.

fluffyfreak:
@Mc-Pain would you mind if I pushed a small change to your PR to cleanup the `UGLY HACK` ?

Mc-Pain:
> @Mc-Pain would you mind if I pushed a small change to your PR to cleanup the `UGLY HACK` ?

+, go ahead

impaktor:
@fluffyfreak Wait with asking until after you've done it. (looks like mcpain gave thumbs up on your post, but you're not notified by that)

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -1290,13 +1290,9 @@ SystemBody* StarSystemRandomGenerator::MakeGravPointForBodies(RefCountedPtr<Star
 	return grav;
 }
 
-std::string StarSystemRandomGenerator::GetStarNameFromId(const int id)
+char StarSystemRandomGenerator::GetStarSuffixFromOffset(const int offsetId) const
```
Mc-Pain, line 1293: Maybe this would help?
```C++
std::string StarSystemRandomGenerator::GetStarSuffixFromOffset(const int offsetId) const
{
    return std::string(1, 'A' + Clamp(offsetId, 0, 26));
}
```

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -1290,13 +1290,9 @@ SystemBody* StarSystemRandomGenerator::MakeGravPointForBodies(RefCountedPtr<Star
...
 {
-	// UGLY HACK
-	assert((id >= 0) && (id < 26));
```
Mc-Pain: I'd suggest not removing this assert.

Silently ignoring wrong values would break unique star names.
offsetId < 0 -> "SystemName A" for any negative value
offsetId >= 26 -> "SystemName Z" for any value

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -1290,13 +1290,9 @@ SystemBody* StarSystemRandomGenerator::MakeGravPointForBodies(RefCountedPtr<Star
...
-	char buf[16];
-	snprintf(buf, sizeof(buf), "%c", 'A' + id);
-	return buf;
+	return 'A' + Clamp(offsetId, 0, 26);
```
Mc-Pain, line 1295: is 26 included?

'A' + 26 = '['

src/galaxy/StarSystemGenerator.cpp
fluffyfreak: I'd like to avoid constructing a `std::string` then return it to a method taking a reference when all you want from this is the single character whcih you _might_ then use in a `std::string`

src/galaxy/StarSystemGenerator.cpp
fluffyfreak: hmm, should probably be 25, assuming we'll be having systems with >25 stars

fluffyfreak: damn, yeah it's inclusive clamp, i'll change it

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -1259,6 +1259,122 @@ void StarSystemRandomGenerator::MakeBinaryPair(SystemBody *a, SystemBody *b, fix
...
+{
+	const std::string systemName = secSys.GetName();
+	SystemBody *body;
+
+	if (numStars == 1) {
+		// We have only one star in the system, make it root
+
+		SystemBody::BodyType type = secSys.GetStarType(0);
+		const std::string suffix(1, GetStarSuffixFromOffset(offset));
+		body = MakeSystemBody(system, systemName, suffix);
```
fluffyfreak, line 1308: Something I did just spot, this shouldnt be taking any suffix at all. If there's a single star in a system then it's just the systemName

src/galaxy/StarSystemGenerator.cpp
Mc-Pain: > I'd like to avoid constructing a `std::string`

I see from patch you're constructing it anyway.
Better do it inside `GetStarSuffixFromOffset` to prevent typos since common code is used *twice*

When having 26+ stars, we'll just have to rewrite this to generate (...X, Y, Z, AA, AB, AC...)
And rewrite GravPoint names to make them joinable via comma instead of concatenating (not sure which is the case now)

**UPD**
suffixes are joined via comma, allowing AA, AB, AC... while E:D just concatenates them (but I've never seen more than 5 stars in that game - 11+ brown dwarfs around main sequence star counts as planets)

src/galaxy/StarSystemGenerator.cpp
Mc-Pain: > Something I did just spot, this shouldnt be taking any suffix at all. If there's a single star in a system then it's just the systemName

It is called recursively, empty root suffix is handled in `StarSystemRandomGenerator::Apply` after calling PlaceStars

src/galaxy/StarSystemGenerator.cpp
fluffyfreak: No I mean that I would like to avoid constructing the `std::string` as a return value, it's fine if you construct one outside but this is really a wrapper around `'A' + int` which gives you a single char that you don't have to use with a `std::string` and you could refactor to work with insert/replace on a string.

Refactoring is for a future that never comes, also it should only be used once.

src/galaxy/StarSystemGenerator.cpp
fluffyfreak: so name it, then always override the name later?

src/galaxy/StarSystemGenerator.cpp
fluffyfreak: restored it

src/galaxy/StarSystemGenerator.cpp
fluffyfreak: I think once you're talking more than 5 or 6 stars in a single system they have to be so distant from one another that defining them as a single star system gets challenging to the meaning of the word.

They'd have to be light years apart and in reality would be affected by other star systems as much as by the other stars they're supposed to be in a system with.

src/galaxy/StarSystemGenerator.cpp
Mc-Pain: > so name it, then always override the name later?

yes. It was easier than pre-naming checks.

src/galaxy/StarSystemGenerator.cpp
Mc-Pain: The record for E:D is: 11 stars
https://spansh.co.uk/system/5645877119803#system-bodies
https://spansh.co.uk/system/527009230954354#system-bodies
https://spansh.co.uk/system/40122780307#system-bodies

fluffyfreak: That'd only get us to `K`, I think we're safe with 14 remaining letters :D

fluffyfreak:
@Web-eWorks I'm happy with this as-is but as I've (tried to) contribute to the code I'm not going to judge whether it should be merged, also with the changes it may have changed the generated universe which requires a savegame bump?

impaktor:
I'd recommend holding of on merge until after 3.14, when we intend to release a bug-fix zuper-ztable Tauoneer (tau = 2 x pi, Tauoneer = Pi-Pioneer - yes, I'll keep proposing that). Players of the 3.14 release are Tauonauts.

Only bug-fixes and PRs that are unlikely to introduce bugs before 3.14, and no save-bumps, according to the Board of Directors of Pioneer.
Tauoneer - dammit!.

Web-eWorks:
Apologies on the delay officially responding here. I'll be reviewing this PR soon after the 3.14 hotfix release, as I'd prefer not to introduce a savegame bump in the upcoming release if it's not absolutely necessary.

impaktor:
2024-05-24-130233_1312x739_scrot

Web-eWorks:
Looking at the result of system generation here - I do like that systems are much more flexible with their arrangements of binary stars, but I suspect the order of bodies generated in this system does not match what they'd be named if the system was discovered by astronomers today. We have an informal pattern of naming objects in a system according to their distance to the system center, and as you can see Qutiio A breaks that pattern (in system Qutiio (-1, 5, 0)).

image

I suspect some of the arrangements here would be much better served by introducing the ability for stars to orbit other stars, rather than requiring they be children of a gravpoint. Additionally, I suspect it'd be quite trivial under the current model for Qutiio D to overlap with the orbits of Qutiio A or Qutiio B,C, compounding the existing issues with body orbits being generated "inside" other bodies.

Bodasey:
Please only read this when you are interested in getting systems who are in accordance with reality...!



Web-eWorks, you are wrong...

Stars and exoplanets get their names by the order they have been discovered.

So, in general, the brightest component gets the letter A, and the brightest star is the most massive one and therefore closest to the gravitational center of the system.

The same for exoplanets: if a gas giant is discovered first, it gets the letter b (not a = star itself) and keeps this letter even when there are more planets detected inside its orbit.

I don't know how pioneer generates systems, gravpoints are a good help if you have only two components, but with more, this won't work (unless you get a quasi-stable system by accident). Imagine all stars in this Quitio system move to one side - where is the gravpoint? Where the gravitational center?

For save system generation, you need
- gravitaional hierarchy
- hill spheres (see #5839)

Video by Harald Lesch (german physician and philosopher, auto-generated english subtitles available)

https://www.youtube.com/watch?v=tV1ZGZaYf1I

Bodasey:
> I'd argue that from a cartographic, navigation perspective, an inside-out naming order is better.

This is of course all right for systems completely generated by a computer or human imagination - and no real or fictional discovery history.

Web-eWorks:
> Web-eWorks, you are wrong...

Indeed, I should have specified that I was speaking in the context of the nomenclature that would be used for a system that had been discovered in its entirety in an instantaneous fashion, rather than a system in which bodies had been discovered piecemeal over tens or even several hundred years.

> Stars and exoplanets get their names by the order they have been discovered.

While this is true in real life, every single system you visit in Pioneer has planets and moons named and ordered by their distance to the primary. Pioneer does not currently attempt to generate any kind of history for random systems, partially because no one's implemented it yet, and partially because the runtime performance for system generation is a primary concern of the system.

> So, in general, the brightest component gets the letter A, and the brightest star is the most massive one and therefore closest to the gravitational center of the system.

...Which is what I was attempting to say originally. The first star generated in a system is *always* the most massive one. Thus it should be closest to the center. However, the code in question can - as evidenced by the screenshot - place the first, most massive star significantly farther from the combined gravitational center than the remaining successive bodies.

> I don't know how pioneer generates systems, gravpoints are a good help if you have only two components, but with more, this won't work (unless you get a quasi-stable system by accident).

This is a good point. I retract my comment regarding the pair-of-binary-pairs - it may produce implausible looking systems to the average player's eye but is a decent enough arrangement with regards to the orbits of the associated bodies.

Zireael07:
Why close?

PR #5756, opened 2024-02-10
Local missions should list distances in AU rather than Ly

### LINKED ISSUE #5755: Local missions have mission distance 0 Ly
### Observed behaviour
mjFvmjy

## Bug

I think the error is that it's measuring distance between player's current system, and mission system which are the same, i.e. it's not just a display bug rounding distance to 0 ly.
```lua
		local playerSystem = Game.system or Game.player:GetHyperspaceTarget()
		local dist = playerSystem:DistanceTo(mission.location)
```

__My pioneer version (and OS):__ Current master, 2024-02-03

impaktor:
When listing table of all player's accepted missions, local deliveries all have distance "0 ly". Switch to AU instead.

Fixes #5755

## Demonstration
2024-02-10-122202_1266x687_scrot

## Pondering
- do we have AU units defined elsewhere that I should use, or defining it in the local file is good?
- under which circumstances would `mission.location:GetSystemBody()` return nil, and what should one display in that circumstance?

### DIFF
--- a/data/pigui/modules/info-view/04-missions.lua
+++ b/data/pigui/modules/info-view/04-missions.lua
@@ -102,13 +102,27 @@ local function makeMissionRows()
 		end
 
 		local playerSystem = Game.system or Game.player:GetHyperspaceTarget()
-		local dist = playerSystem:DistanceTo(mission.location)
 		local days = math.max(0, (mission.due - Game.time) / (24*60*60))
+
+		-- Use AU for interplanetary, LY for interstellar distances
+		local dist, dist_display
+		if mission.location:IsSameSystem(playerSystem.path) then
+			if mission.location:IsBodyPath() then
+				local body = mission.location:GetSystemBody().body
+				dist = Game.player:GetPositionRelTo(body):length()
+				dist_display = "\n" .. ui.Format.Distance(dist)
+			else
+				dist_display = "\n-"
+			end
+		else
+			dist = playerSystem:DistanceTo(mission.location)
+			dist_display = string.format("\n%.2f %s", dist, l.LY)
+		end
 
 		local row = {
 			mission:GetTypeDescription(),
 			mission.client.name,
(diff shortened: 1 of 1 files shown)


Web-eWorks (changes requested):
- Before accessing the result of `SystemPath:GetSystemBody()` you should check to make sure `mission.location:IsBodyPath()` is true. The location might be a path to a system rather than a body in that system, in which case the system body will be `nil` and dereferencing `.body` will raise an error.
- Use `ui.Format.Distance` to display any in-system distances (passing the raw distance to the target). It will handle dividing by meters/km/AU as needed and format the result accordingly.
- Off the top of my head, we don't have a pre-built formatter for inter-system distances, so keeping the old `string.format` method for LY is fine.

data/pigui/modules/info-view/04-missions.lua
```diff
@@ -102,13 +102,27 @@ local function makeMissionRows()
...
 		local playerSystem = Game.system or Game.player:GetHyperspaceTarget()
-		local dist = playerSystem:DistanceTo(mission.location)
 		local days = math.max(0, (mission.due - Game.time) / (24*60*60))
 
+		-- Use AU for interplanetary, LY for interstellar distances
+		local dist, dist_display
+		if mission.location:IsSameSystem(playerSystem.path) then
+			if mission.location:IsBodyPath() then
+				local systemBody = mission.location:GetSystemBody().body
+				dist = Game.player:GetPositionRelTo(systemBody):length()
```
Web-eWorks, line 112: This body is not a SystemBody, it's just a regular Body. Would strongly recommend the variable not be named systemBody as that usually implies the type of the variable is actually the SystemBody returned from GetSystemBody().

(Variable names are important in dynamically-typed languages...)

Web-eWorks (approved):
Looks good overall, if it's been tested then it gets my thumbs up. There's one variable name that should be changed before merge, but otherwise looks solid.

PR #5758, opened 2024-02-10
Fix star background generation

### LINKED ISSUE #5414: Irregular display of stars in the sky
### Observed behaviour

https://user-images.githubusercontent.com/18342621/199968979-4440df9d-91c8-47fb-833f-244799db1317.mp4

On the oXY plane, a strip is clearly visible, there are more stars than usual, and two dark spots are also visible

master at https://github.com/pioneerspacesim/pioneer/commit/824adc84b4451e31c758aec36360b432cc8b7541

Gliese852:
- really do early out

  - rotate coordinate system from galaxy (Z-up) to scene (Y-up)

  - more fairly distribute the number of stars among tasks - these are spherical segments

  - do not overlap task ranges

  - calculate the average of the median brightnesses, and use the same value in all pieces

Fixes #5414

### DIFF
--- a/src/Background.cpp
+++ b/src/Background.cpp
@@ -30,6 +30,7 @@
 #include <SDL_stdinc.h>
 #include <iostream>
 #include <sstream>
+#include <numeric>
 
 using namespace Graphics;
 
@@ -250,7 +251,6 @@ namespace Background {
 
 	struct StarQueryInfo {
 		const SystemPath *systemPath;
-		uint32_t numStars;
 		int32_t sectorMin;
 		int32_t sectorMax;
 		int32_t visibleRadiusSqr;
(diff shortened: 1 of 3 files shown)


src/Background.cpp
```diff
@@ -326,27 +328,22 @@ namespace Background {
 							// use a logarithmic scala for brightness since this looks more natural to the human eye
 							float brightness = log(luminositySystemSum / (4 * M_PI * distance.Length() * distance.Length()));
 
-							stars.pos.push_back(distance.Normalized() * 1000.0f);
+							// galaxy is Z-up but the scene is Y-up
```
Web-eWorks, line 331: I think it'd be simpler to handle this in Z-up space and merely set the transform matrix to rotate Y-up to Z-up when rendering the starfield.

Web-eWorks (changes requested):
Thanks for working on this! The code overall looks good, hopefully the TaskGraph / TaskSet API wasn't too difficult to understand. I've left one comment where I think an operation should be moved to a different point in the lifetime of the Background class.

Will give this a test in the next few days.

src/Background.cpp
Gliese852: Done! Yes, it’s better this way, considering that in the possible future I would like each system to have its own transformation matrix. (It is strange that the axes of all our systems are parallel to the axis of the galaxy.)

Web-eWorks (approved):
Test looks good. If interested in doing further work in this area in another PR, now that starfield generation is multithreaded the SystemMapViewport could own its own Background instance and rebuild the background when the currently viewed system changes (currently stars are only drawn in the map background if viewing the current system in the game, and never in the editor because of this).

PR #5759, opened 2024-02-11
Update libfmt, fix compilation warnings and invalid JSON CustomSystem position

### LINKED ISSUE #5693: Should we update fmt?
### Observed behaviour

When I compile pioneer, stderr is flooded with warnings like below. This makes it likely that other warnings may be missed.

```
In file included from /home/cdda/git/pioneer/contrib/fmt/include/fmt/format.h:44,
                 from /home/cdda/git/pioneer/src/core/Log.h:7,
                 from /home/cdda/git/pioneer/src/utils.h:7,
                 from /home/cdda/git/pioneer/src/ship/ShipViewController.h:10,
                 from /home/cdda/git/pioneer/src/WorldView.h:10,
                 from /home/cdda/git/pioneer/src/Player.cpp:19:
/home/cdda/git/pioneer/contrib/fmt/include/fmt/core.h:443:7: warning: identifier ‘c
har8_t’ is a keyword in C++20 [-Wc++20-compat]
  443 | using char8_t FMT_DEPRECATED_ALIAS = internal::char8_type;
      |       ^~~~~~~
```

### Expected behaviour

No warnings

### Steps to reproduce

`cmake -S . -B build/ && make -j2 -C build/ `

__My pioneer version (and OS):__ 97d7be673f2124f1b42aee956c4b3a6fc2129558 (and fedora 29)

g++ (GCC) 13.2.1 20231205 (Red Hat 13.2.1-6)

cmake version 3.27.7

GNU Make 4.4.1

### Notes

It appears that fmt>=7.0.0  removed the code that was causing the warning https://github.com/fmtlib/fmt/commit/d6cea50d01d7779ecb5ae4da0d785d9bba47a2f7 (if I am looking at the right git repo)

Web-eWorks:
Been sitting on this PR for longer than I intended, life got busy.

Essentially does what it says on the tin, updates libfmt to v10.2.1 (fixes #5693) and fixes a few sets of compilation warnings while in the area.

I've found and fixed a bug that was causing JSON custom systems to lose their position values (more correctly, the position was initialized from the sector index instead, which could cause all kinds of bugs) - it's benign at the moment since we don't ship any JSON custom systems other than Sol with a `{0,0,0}` sector index and position.

This also bumps our CI build version to `20240203-dev` following the schema we've set up - this PR was intended to be opened immediately after the release / that commit pushed directly to master, but my attention has been unfortunately elsewhere.


Web-eWorks:
This PR also introduces a new test case addressing the `-Wsequence-point` warning GCC provides, validating that all compilers we build and test against properly implement the C++17 standard's sequencing rules for initializer clauses in a braced-init-list (which we depend on for `pi_lua_multiple_pull`).

PR #5763, opened 2024-02-13
Add a recovery reader for the v90 flight log

Gliese852:
Although the flight log is now a single list, it still has the same three types of entries. Therefore it can still be shown without changing the GUI in the start menu, where 3 sections (entry types) are shown separately, depending on the combobox.

It is now possible to determine the log entry type only by the object type, which becomes available after full deserialization. But I would not want to do this, including reusing the actual code from the game, since it may change over time. Information about the object type is contained in the `lua_class` field, which is lost when the save tree is unpickled. In order to save this information, and at the same time not change anything radically in the unpickled tree, it was decided to put all type names into one separate subtable in the save tree, `type_names`, the keys in which will be the tables corresponding to the unpickled objects.

### DIFF
--- a/data/pigui/modules/new-game-window/flight-log.lua
+++ b/data/pigui/modules/new-game-window/flight-log.lua
@@ -233,6 +233,44 @@ FlightLog.reader = Helpers.versioned {{
 
 		return value
 	end
+
+}, {
+
+	version = 90,
+	fnc = function(saveGame)
+
+		local function entryWithSafeSystemPath(entry)
+			local parsed = {}
+			for k,v in pairs(entry) do
+				if k == 'systemp' then
+					parsed.path = systemPathFromTable(entry.systemp.inner)
+				else
+					parsed[k] = v
+				end
+			end
+			return parsed
+		end
+
+		local value = { Custom = {}, System = {}, Station = {} }
+
+		local logData, errorString = Helpers.getByPath(saveGame, "lua_modules_json/FlightLog/Data")
+		if errorString then return nil, errorString end
+
+		for _, entry in ipairs(logData) do
+			local entryType = Helpers.getLuaClass(entry)
+
+			if entryType == 'FlightLogEntry.Custom' then
+				table.insert(value.Custom, entryWithSafeSystemPath(entry))
+			elseif entryType == 'FlightLogEntry.System' then
+				table.insert(value.System, entryWithSafeSystemPath(entry))
+			elseif entryType == 'FlightLogEntry.Station' then
+				table.insert(value.Station, entryWithSafeSystemPath(entry))
+			else
+				return nil, lui.UNKNOWN_STATION_LOG_ENTRY_FORMAT
+			end
+		end
+		return value
+	end
 }}
 
 FlightLog.updateLayout = false
--- a/data/pigui/modules/new-game-window/flight-log.lua
+++ b/data/pigui/modules/new-game-window/flight-log.lua
@@ -26,7 +26,7 @@ function FlightLog:fromStartVariant(variant)
 	self.value.System = {}
 	self.value.Station = {}
 	self.value.Custom = {
-		variant.logmsg and { text = variant.logmsg }
+		variant.logmsg and { entry = variant.logmsg }
 	}
 end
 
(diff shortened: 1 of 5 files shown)


impaktor:
Pinging in @JonBooth78

data/pigui/modules/new-game-window/flight-log.lua
```diff
@@ -233,6 +233,44 @@ FlightLog.reader = Helpers.versioned {{
 
 		return value
 	end
+
+}, {
+
+	version = 90,
```
impaktor, line 239: Will this hardcoding of the version need to be updated or changed at some point? Or it's only there while we want to have compatibility with versions prior to 90? 
(Sorry, I'm too lazy to read the code)

Gliese852: This number is here so that the following function is called for saves v90 and newer. Immediately before you can see the function for v89, it will be called when restoring saves v89 and older. This list can grow on both sides to support older and newer save versions.

JonBooth78:
I'm concerned about a couple of things here.  Firstly, I'm not a fan of renaming the `text` field of an `entry` to `entry`.  IMHO the entry encompasses all the data, including location, time etc.   The `text` is just one of those pieces of data - having an `entry` within an `entry` and so perhaps accessing it like `entry.entry` feels off.  Fairly minor thing though.

JonBooth78:
My second concern here is we're moving to a world where we maintain 2 versions of the flight log code - both serialization and UI, one for in-game and one for restoration.

This is very far from ideal as it essentially will make it twice as hard to evolve that part of the game and makes it brittle and hard to maintain.  So I'd actually like look at consolidating it all together.

I've not had time to dig into this far enough but I think this means that loading of flight-logs should always goes through the 'restore' path somehow, or that we have a common interface for deserialization that they both conform to in order to provide the POD that needs interpreting.

Similarly we should provide a 'flightlog' UI element that can be embedded elsewhere and is configurable.

All of this is possible and I think we should consider as the architectural end-goal.

None of this stops the usefulness of where we are now; I love the restore function and I'm happy for this kind of change until we (perhaps I) have time to address that architectural issue.

Gliese852:
@JonBooth78 regarding the `entry` field, I renamed it because that's what you called it, in the old flight log, the entries were arrays. https://github.com/pioneerspacesim/pioneer/blob/2619555d0dcf4f76c646e35e91d67c6d11e4ece5/data/modules/FlightLog/FlightLogEntries.lua#L100

Regarding the two versions of the UI and deserialization, I agree with you. I would like to note that the problem of GUI reuse most likely relates not to recovery, but to the new game start menu. There are examples of GUI reuse - `SystemOverviewWidget` is used in the 'Location' tab.

The problem of reusing deserialization seems to me on the one hand more complex - I think that in order to fully reuse this, we need to run the full process of loading the game, but we can hardly do this for the old save, but that's the essence of recovery.
On the other hand, deserialization within a game is usually trivial:
https://github.com/pioneerspacesim/pioneer/blob/2619555d0dcf4f76c646e35e91d67c6d11e4ece5/data/modules/FlightLog/FlightLog.lua#L440
The code you wrote to support previous version is not actually needed (just this could be transferred to the recovery area, but the recovery is already there, so you can just delete it).

But overall, this is a very interesting topic to think about, and there is definitely room for improvement.

Web-eWorks:
@Gliese852

> In order to save this information, and at the same time not change anything radically in the unpickled tree, it was decided to put all type names into one separate subtable in the save tree, type_names, the keys in which will be the tables corresponding to the unpickled objects.

Note: you could also create one "dummy" metatable per unique `lua_class` name and provide a `Helpers.GetLuaClass(obj)` which retrieves the object's lua_class by indexing into the table mapping metatables to strings (or just store the string name in the metatable directly).

> regarding the entry field, I renamed it because that's what you called it, in the old flight log, the entries were arrays. 

I would strongly recommend this field either remain `text` or be named something like `note`, as that's what it is - it's a string containing the player's notes for that entry in the logbook. You have the overall "form entry" in the "logbook" and then inside that form entry there's a text field for the pilot/captain to record any special observations.

Gliese852:
@Web-eWorks 
> I would strongly recommend this field either remain `text` or be named something like `note`

Perhaps I wasn't specific enough, the name `entry` is now in the 20240203, and in v90 saves. In v89 it wasn't called anything at all. In this sense, it cannot remain `text` since it was never called that.
Now, in order not to call this `entry`, we need to change the main FlightLog code, and this will be savegamebump.

Gliese852:
@Web-eWorks Implemented what was suggested in fixup commit (about metatables).

JonBooth78:
Darn, it seems like my change from `entry` -> `text` also got eaten by my poor rebase and so we went with `entry`, this wasn't what was intended :-( 

I think I should look at re-tidying this up.

Re: the versioning code; it's actually very handy to have in the flight log as then one can change this without having to recover or version bump the entire thing :-)

The whole `FlightLog.InsertCustomEntry` set of functions in FlightLog seem off, however, that's due to the dichotomy of the loading/saving code, which I think needs a good look over.

Web-eWorks (approved):
Looks sane to me, sorry about the delay in reviewing. I'd suggest this PR also remove the `Version == 1` migration code present in the FlightLog itself, since that should now be entirely handled by the savegame recovery code (as saves with the previous version of flight log data are no longer compatible).

Gliese852:
Forced a fixup commit with a small amendment and removed the in-game migration code.

PR #5766, opened 2024-02-16
Generate atmospheres for gas giants in legacy custom systems

Web-eWorks:
Prior to the merge of the System Editor and associated system generation refactor, all gas giants had a fixed 14.0 kg/m3 atmospheric surface density value which overrode the value of `m_volatileGas` and thus many Lua custom systems never set the atmospheric density parameter, which resulted gas giants in custom systems (e.g. Sol, Noril, etc.) missing scoopable atmospheres.

Sol was resolved by manually editing the gas giants to set density to real-world values; this PR follows up and adds an additional "fixup" step when loading legacy Lua custom systems that generates "missing" atmospheric density values for gas giants.

**Testing Steps:**
Visit Noril C and attempt to scoop it for fuel. It should have an atmospheric density > 0 before you crash :wink:

### DIFF
--- a/src/galaxy/CustomSystem.cpp
+++ b/src/galaxy/CustomSystem.cpp
@@ -1018,7 +1018,9 @@ void CustomSystemsDatabase::RunLuaSystemSanityChecks(CustomSystem *csys)
 
 		bool wantRings = body->ringStatus == CustomSystemBody::WANT_RANDOM_RINGS || body->ringStatus == CustomSystemBody::WANT_RINGS;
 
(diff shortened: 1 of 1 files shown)


JonBooth78:
[_exit.zip](https://github.com/pioneerspacesim/pioneer/files/14317798/_exit.zip)

Here is a save file for testing

JonBooth78:
Just gave this a quick test and it appears to work.  I haven't tried testing but the assert no longer appears when loading the above save, which means the density is > 0.

PR #5767, opened 2024-02-17
Fix projectiles being added a frame behind where they are fired and beam weapons not able to hit

### LINKED ISSUE #5695: Projectiles/laser appear offset from the ship depending on speed
### Observed behaviour
On master shots fired appear offset from the ship. It seems to depend on the speed/vector. The faster you fly the further they appear from the ship, like if they are lagging. Or released where you were a split second before.
screenshot-20231226-184726
screenshot-20231226-185301
screenshot-20231226-185312

At zero speed, it come from the right place:
screenshot-20231226-185349

Not sure if FPS affects it, but time acceleration does:
screenshot-20231226-190118

Edit:
Happens to enemy ships too:
screenshot-20231226-191300
And everything snaps back to its proper place when the game is paused:
screenshot-20231226-191304

### Expected behaviour
Shots appear at their proper place

### Steps to reproduce
Fly fast and shoot your weapons. 

20231222 (70d77a2) (I believe it was a translation update build)
Win10

### LINKED ISSUE #5741: Laser weapons do not damage ships
### Observed behaviour

Laser weapons do not damage ships. It only works at a very close distance (a few meters).

https://github.com/pioneerspacesim/pioneer/assets/99830004/eba5125e-de96-4c11-83ed-fe592e1fc2dd

### Expected behaviour

The laser should damage ships

### Steps to reproduce

1. Buy laser weapons
2. Call the ship through debugging tools
3. Shoot

ver 20240203 (e7f3122) on: Windows

[output.txt](https://github.com/pioneerspacesim/pioneer/files/14156517/output.txt)

JonBooth78:
This makes combat impossible if travelling at speed

Fixes #5695 and Fixes #5741

Revert initial change that fixed adding items to the vector whilst iterating it but introduced the frame lag and instead fix in a simple way.

### DIFF
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -1076,7 +1076,7 @@ void Space::TimeStep(float step)
 	// https://github.com/pioneerspacesim/pioneer/issues/5695
 	//
 	// THIS IS A HACK/WORKAROUND until a more proper solution can be found
-	for (size_t i = 0, e = m_bodies.size(); i != e; ++i) {
+	for (size_t i = 0; i < m_bodies.size(); ++i) {
 		auto b = m_bodies[i];
 		b->StaticUpdate(step);
 	}
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -18,9 +18,9 @@
 #include "SpaceStation.h"
 #include "Star.h"
 #include "SystemView.h"
-#include "core/Log.h"
 #include "collider/CollisionContact.h"
 #include "collider/CollisionSpace.h"
+#include "core/Log.h"
 #include "galaxy/Galaxy.h"
 #include "graphics/Graphics.h"
 #include "lua/LuaEvent.h"
@@ -263,7 +263,7 @@ Space::Space(Game *game, RefCountedPtr<Galaxy> galaxy, const Json &jsonObj, doub
 	try {
 		Json bodyArray = spaceObj["bodies"].get<Json::array_t>();
 		for (Uint32 i = 0; i < bodyArray.size(); i++) {
-			if(bodyArray[i].count("is_not_in_space") > 0)
+			if (bodyArray[i].count("is_not_in_space") > 0)
 				continue;
 			m_bodies.push_back(Body::FromJson(bodyArray[i], this));
 		}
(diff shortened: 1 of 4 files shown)


JonBooth78:
If anyone is interested, my overly clever fix is here: https://github.com/pioneerspacesim/pioneer/commit/563d305bd201b1c06f0acf5637e99e48d04c42ab

src/Space.cpp
```diff
@@ -1071,9 +1068,18 @@ void Space::TimeStep(float step)
...
-	for (Body *b : m_bodies)
+	// NOTE: The AI can add bodies here so we can't use an iterator
+	// this restores the previous version where only the initial list is
+	// updated unless the bodies vector reallocated where anything could
+	// have happened
+	// alternative fixes to delay addition caused
+	// https://github.com/pioneerspacesim/pioneer/issues/5695
+	//
+	// THIS IS A HACK/WORKAROUND until a more proper solution can be found
+	for (size_t i = 0, e = m_bodies.size(); i != e; ++i) {
```
Web-eWorks, line 1079: As far as I can tell, this will retain the old behavior where only the bodies present at the start of iteration are actually ticked here. To iterate over new-this-frame bodies, the condition needs to be changed to `i != m_bodies.size()` - `e` is evaluated once at the start of the loop and if new bodies are added to the end they will not be updated.

JonBooth78: So... the original behavior used an 

`for (auto b : m_bodies )`

Which expands to 

`for (auto i = m_bodies.begin(), auto e = m_bodies.end(), i != e; ++i ) { auto b = *i; `

So I went for the same logic, which means that their static update wouldn't get called as I wanted to make the code behave as close as possible to how it did before, without the risk of having invalid iterators.

After our IRC Chat, I agree, we should check `size()` each iteration rather than store it at the start as I think for projectiles (which are often the things added in this loop) we probably want to run their update too.

This will, I think make beam weapons more deadly as they'll hit in the first frame and not the second, and so have not degraded their power at all (if I read it all correctly) :-)

Web-eWorks (approved):
Looks good to me.

PR #5734, opened 2024-02-03
Ship Equipment Overhaul - EquipSet v2

Web-eWorks:
Creating as WIP PR for visibility and collaboration purposes. No guarantees this runs or is fully functional at this juncture, though I've tried to ensure I have a working test case during most of development.

This branch is the culmination of several years of sporadic design discussion about Pioneer's current equipment system and the problems it presents for certain important gameplay systems we'd like to enable.

We've been laying the groundwork for a replacement system on IRC and in the dev docs for quite some time, and enough ducks lined up in a row to be able to put fingers to keyboard and start writing the replacement for the existing system.

### The Problems

1. It's very difficult to support more than one weapon per ship, since there's no concept of where an equipment slot is, or which individual slot within a type (e.g. `laser_front`) an item is installed in
2. Equipment items aren't unique instances - there's no way to have two shield modules installed but only have one of them turned on, damaged, etc. without writing a separate system to handle each equipment state
3. Since equipment items are "all the same", you can't buy a repaired / slightly damaged equipment module off the secondhand market for cheap. You're buying the exact same module you buy from the ship dealer, down to the *table identity*. It also makes things more complicated to handle storing purchased equipment items at a station when unequipped but not sold (when we get to that point).

### The Solution

We've settled on a logical extension to the slot-based design of the prior equipment system, but allowed players to interact with the individual slots on their ship as first class features rather than hidden constraints.

Slots have type and size constraints (potentially extended to resource availability, e.g. power/ammo in the future) which limit the types of equipment that can be installed and allow some additional stratification of ship roles; an AC33 is probably not going to be happy with a shield generator designed for a Pumpkinseed, and a 20MW pulsecannon cannot physically be connected to a Coronatrix no matter how much you shuffle around the insides.

To complement this system, all equipment items that are installed on ships are now thin instances of equipment item "prototypes". This allows for very easy future support for individual equipment item state, like damage/wear, power on/off, and "tuning" equipment items to be better in specific capabilities (at the expense of others).

### Status

This branch is by no means feature complete. As of time of writing, the diff is 4k lines added, 1.8k lines removed, and quite a bit more is expected. I'm creating this PR now to collaborate with other contributors and to seek some feedback from interested parties on anything I might be forgetting.

The overall effort to move to the new equipment system is broken down into three stages:
1. (This PR) Convert existing equipment items and code to use the new system, and provide very basic slot configurations for all existing ships to get everything playable. 80% of the effort.
[...]

PR #5773, opened 2024-02-21
Misc cleanup, fixes, and Lua function library improvements

### LINKED ISSUE #5765: Hyperspace text color is unintentional
Color used to be green (since 2008). Many other green HUD elemets have been shifted to orange/amber (like nav tunnel).

Now the color has changed, to be that from the market, I think, for some reason, which I don't think was intentional. 

We just need to explicitly set some intentional color, I think we agreed on in IRC.

2024-02-14-180132_1536x960_scrot

Web-eWorks:
This PR is a bit of a catch-all arising from development of #5734 and contains most of the feature-agnostic commits from that branch.

In no particular order:
- Improved the Lua runtime library with several new utilities for operating on tables
- Added a new `utils.proto` helper function for working with a prototype-based object model
- Fixed an error when attempting to interpolate a non-string object into a string
- Fixed a long-running bug with `utils.print_r` which caused it to print incorrect information about recursive tables (tables that have already been printed)
- Fixed default forwarding constructor for `utils.inherits` - you can now receive arguments from the default `.New()` constructor of a subclass.
- Fixed #5765
- Made most PiGui methods available to Lua autocomplete through the `ui` object. Documenting forwarded methods from C++ is still a WIP and will be improved over time.
- Added two new utility UI functions - ui.getItemRect() and ui.Format.Volume
- Improved the resilience of the tab views against erroring code
- Some misc. preparatory changes for EquipSet v2.

### DIFF
--- a/data/lang/core/en.json
+++ b/data/lang/core/en.json
@@ -1835,6 +1835,10 @@
     "description": "Duration unit: one 7-day week",
     "message": "w"
   },
+  "UNIT_CUBIC_METERS": {
+	"description": "Volume unit: cubic meter",
+	"message": "m³"
+  },
   "UNKNOWN": {
     "description": "",
     "message": "<unknown>"
--- a/data/libs/SpaceStation.lua
+++ b/data/libs/SpaceStation.lua
@@ -2,6 +2,7 @@
 -- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
 ---@class SpaceStation : ModelBody
+---@field techLevel integer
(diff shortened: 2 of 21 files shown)


impaktor:
Is `onChangeEquipment` incompatible with the future equip-v2?

Web-eWorks:
`onChangeEquipment` was a legacy C++ signal for dispatching to C++ code that no longer exists (I'm guessing probably in the sector map / hyperjump planner when that was still C++). It's totally unused in current master, and will continue to be unused with the EquipSet v2 rewrite.

PR #5774, opened 2024-02-21
Add Lua "Persistent Object" Serialization

Web-eWorks:
This PR adds support for serializing "persistent objects", i.e. objects that will always be created at runtime. References to these persistent objects are transparently replaced with references to the "live" copy of the persistent object when loading a saved game, unless that persistent object is no longer present (i.e. incompatible saves) in which case the persistent object is "revivified" from the savefile.

This is a difficult concept to explain, but if you're familiar with how commodity types function currently (where saved copies of a commodity type instance are replaced with the commodity type instance that was created at startup) this is an identical concept but with less manual work required.

The primary goal of this system is to migrate equipment items to use this rather than overwriting all equipment item tables every time a save is loaded. As this is primarily accomplished in #5734, this PR only includes the (bugtested) persistent serialization mechanism itself to avoid necessitating a savebump.

TODO:
- [x] Convert CommodityType to use `Serializer:RegisterPersistent()`

### DIFF
--- a/data/meta/Serializer.lua
+++ b/data/meta/Serializer.lua
@@ -25,4 +25,11 @@ function Serializer:Register(key, serialize, unserialize) end
 ---@param class table
 function Serializer:RegisterClass(key, class) end
 
+-- Register a table as a "persistent" value. All references to the saved
+-- instance of that table will be transparently replaced across savegames to
+-- maintain table instance identity.
+---@param key string
+---@param value table
+function Serializer:RegisterPersistent(key, value) end
+
 return Serializer
--- a/src/lua/LuaSerializer.cpp
+++ b/src/lua/LuaSerializer.cpp
@@ -391,7 +391,8 @@ void LuaSerializer::SavePersistent(Json &json)
 		std::string id = LuaPull<std::string>(l, -2);
 
 		// Serialize all registered persistent objects as a "backup" version
-		// (This will populate the PiSerializerTableRefs cache)
+		// This will populate the table identity cache to avoid persistent
+		// objects being serialized twice.
 		entry["persistId"] = id;
 		pickle_json(l, -1, entry, id);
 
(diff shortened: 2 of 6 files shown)


src/lua/LuaSerializer.cpp
```diff
@@ -359,12 +367,85 @@ void LuaSerializer::UninitTableRefs()
...
+
+	for (const auto &item : persist) {
+		std::string id = item["persistId"].get<std::string>();
+
+		lua_pushinteger(l, item["ref"].get<lua_Integer>());
+		lua_getfield(l, -2, id.c_str()); // reftable, persist, ptr, pvalue
+
+		// We must unpickle the saved persistent object, because it may contain
+		// "trapped" serialized representations of non-persistent tables
+		unpickle_json(l, item);
```
Gliese852, line 430: Taking this opportunity, could you please document a little about the side effects and the non-obvious result of this function?

src/lua/LuaSerializer.cpp
```diff
@@ -359,12 +367,85 @@ void LuaSerializer::UninitTableRefs()
...
+		if (lua_isnil(l, -2)) {
+			Log::Warning("Restoring missing persistent object '{}' from savefile", id);
+			lua_replace(l, -2); // reftable, persistent, ptr, svalue
+		} else {
+			lua_pop(l, 1); // reftable, persistent, ptr, pvalue
+		}
+
+		// All references to the prior saved value are replaced with the persistent object
+		lua_settable(l, idx_reftable);
+	}
```
Gliese852, line 442: It seems that the missing persistent object is not put into the NS_PERSISTENT table, and therefore will not be saved to the next save? So the next time we load we will still get a dangling object referencing it?

Gliese852 (commented):
This is a very interesting functionality, since in fact such logic already exists in Pioneer, giving it a specific place seems like a great solution!

data/meta/Serializer.lua
```diff
@@ -25,4 +25,11 @@ function Serializer:Register(key, serialize, unserialize) end
 ---@param class table
 function Serializer:RegisterClass(key, class) end
 
+-- Register a table as a "persistent" value. All references to the saved
+-- instance of that table will be transparently replaced across savegames to
+-- maintain table instance identity.
+---@param key string
+---@param value table
+function Serializer:RegisterPersistent(key, value) end
```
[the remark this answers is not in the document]
Web-eWorks: The nomenclature is borrowed from the Eris / Pluto projects for Lua serialization - given that these are two entirely different API surfaces (`LuaObjectBase::RegisterPersistent` can never be called from Lua) I don't think it's a problem but I'm open to suggestions for better names.

src/lua/LuaSerializer.cpp
Web-eWorks: Heh, returning the favor I see. Sure, I can definitely do that :smile:

src/lua/LuaSerializer.cpp
```diff
@@ -13,6 +13,13 @@
 #include "core/Log.h"
 #include "profiler/Profiler.h"
 
+// Well-known names of various serialization-related caches stored in the
+// Lua Registry
+static const char *NS_REFTABLE = "PiSerializerTableRefs";
```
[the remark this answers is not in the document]
Web-eWorks: This is for consistency's sake - rather than manually writing or copy-pasting the string name every time we need to access the table, all references in code to this table use a single constant variable as an identifier. It reduces the possibility of mis-spelling the table name and allows it to participate in autocompletion. It can also be made `extern` if it needs to be referred to outside this file.

src/lua/LuaSerializer.cpp
```diff
@@ -359,12 +367,85 @@ void LuaSerializer::UninitTableRefs()
...
+	lua_pushnil(l);
+
+	while (lua_next(l, -2)) {
+		Json entry = Json::object();
+
+		// persistent, id, value
+		std::string id = LuaPull<std::string>(l, -2);
+
+		// Serialize all registered persistent objects as a "backup" version
+		// (This will populate the PiSerializerTableRefs cache)
```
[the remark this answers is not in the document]
Web-eWorks: This is just a comment, which I don't consider high priority to rewrite (if we mis-spell it as PiSierializerTablRefs the code will work correctly and not silently fail) but should probably be expanded to "This will populate the table identity cache for use in unserialization".

src/lua/LuaSerializer.cpp
Web-eWorks: *Good* catch, thank you! I had completely overlooked this. You are correct, this should probably be written back to the persistent object cache. However, modifying the persistent object cache in this way will affect other saves made after starting a new game during the same session, so it's not the cleanest solution.

The latter issue can be resolved by reloading the Lua environment between game starts as we are (eventually) planning on doing, so it's not a huge issue at the moment.

PR #5783, opened 2024-02-26
Fix the accumulation of zombie traders and some other unnecessary things

### LINKED ISSUE #5735: Game is becoming slow after long flights
Loading this savefile will bring the game into slowness.

[_quicksave.gz](https://github.com/pioneerspacesim/pioneer/files/14150868/_quicksave.gz)

I suspect it is being accululated through gameplay until next jump.

Will investigate further

Gliese852:
Collected the simplest, but quite effective fixes so that when you stay in one system for a long time, there is no noticeable drop in performance.

Fixes #5735
(taking into account that #5127 and #5785 are merged)

~~Although at least there remains the problem that occasionally the tradeship runs out of fuel; repairing it probably requires serious intervention in the autopilot, or the creation of evacuation ships, for example.~~ **just explode them**

### DIFF
--- a/src/lua/LuaShip.cpp
+++ b/src/lua/LuaShip.cpp
@@ -1325,6 +1325,28 @@ static int l_ship_get_thruster_state(lua_State *l)
 	return 1;
 }
 
+/* Method: GetThrusterAcceleration
+ *
+ * Returns maximum acceleration in one of 6 directions
+ *
+ * Parameters:
+ *
+ *   direction - number, the <ShipTypeThruster> enum value
+ *
+ * Returns:
+ *
+ *   acceleration - number, m/s/s
+ *
+ */
+static int l_ship_get_thruster_acceleration(lua_State *l)
+{
+	Ship *s = LuaObject<Ship>::CheckFromLua(1);
+	auto direction = Thruster(LuaPull<int>(l, 2));
+	double acc = s->GetComponent<Propulsion>()->GetAccel(direction);
+	LuaPush(l, acc);
+	return 1;
+}
+
 /*
  * Group: AI methods
  *
@@ -1706,6 +1728,7 @@ void LuaObject<Ship>::RegisterClass()
 		{ "GetVelocity", l_ship_get_velocity },
 		{ "GetPosition", l_ship_get_position },
 		{ "GetThrusterState", l_ship_get_thruster_state },
+		{ "GetThrusterAcceleration", l_ship_get_thruster_acceleration },
 
 		{ "IsDocked", l_ship_is_docked },
 		{ "IsLanded", l_ship_is_landed },
--- a/data/modules/LifeSupport.lua
+++ b/data/modules/LifeSupport.lua
@@ -10,10 +10,10 @@ local Timer       = require 'Timer'
 local lc = require 'Lang'.GetResource("core")
 
 local function LifeSupportCallback(self)
-	if not self:exists() then return false end
+	if not self:exists() then return true end
 
 	if self:GetDockedWith() then
-		return true
+		return false
 	end
 
 	---@type CargoManager
@@ -32,7 +32,7 @@ local function LifeSupportCallback(self)
 	-- 	Engine.RequestProfileFrame()
 	-- end
 
-	return true
+	return false
 end
 
 -- TODO: this can be massively improved as far as performance goes; it should
@@ -48,6 +48,6 @@ Event.Register('onShipCreated', function (ship)
 	-- so distribute timer callback load over the full 5s to avoid massive
 	-- single-frame spikes where every ship is processed at once
 	Timer:CallAt(Game.time + 5.0 * Engine.rand:Number(), function()
-		Timer:CallEvery(5.0, function() LifeSupportCallback(ship) end)
+		Timer:CallEvery(5.0, function() return LifeSupportCallback(ship) end)
 	end)
 end)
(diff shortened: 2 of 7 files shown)


src/lua/LuaShip.cpp
```diff
@@ -1325,6 +1325,28 @@ static int l_ship_get_thruster_state(lua_State *l)
...
+ *
+ * Returns maximum acceleration in one of 6 directions
+ *
+ * Parameters:
+ *
+ *   direction - number, the <ShipTypeThruster> enum value
+ *
+ * Returns:
+ *
+ *   acceleration - number, m/s/s
```
impaktor, line 1338: pointless nitpick from me: do we want `m/s/s` or `m/s^2` ?

impaktor (commented):
nice!

fluffyfreak:
>  occasionally the tradeship runs out of fuel; repairing it probably requires serious intervention
Or just delete them... the player won't notice, and log a warning

Web-eWorks (approved):
> occasionally the tradeship runs out of fuel

This sounds like a SAR mission opportunity to me.

@Gliese852 thank you for taking the time to dig into this issue! Very good catch with the life support callback remaining as a zombie, that was definitely my fault when I moved the callback from C++ into Lua. I think the callback requires further thought - it really only needs to run every 5 seconds for the player's ship - all others can have it run once every hour or so. We'd technically lose some "systemic gameplay" from making it player-only, but NPC tradeships probably shouldn't be created with missing life support equipment when carrying perishable cargo.

The GetThrusterAcceleration method is fine as-is, though I'd like to make the Propulsion component directly available via `ship:GetComponent("Propulsion")` and move most of the "passthrough" thruster-related APIs to that component.

claudiusmueller:
> This sounds like a SAR mission opportunity to me.

I wish we could start connecting modules like this! 

I guess within SAR I could check for any ships without fuel? Hmm.....

nozmajner:
SAR missions from out-of-fuel ships also will likely giv a different kind of mission, maybe even a more urgent one.
At least I suspect that most of them run out of fuel in the deceleration phase, so they likely still have a relatively high speed. Which is a quite different catch-up maneuver than for an orbiting ship.
Maybe there could even be a comms message about the distress, and they cold be highlighted on the orbital map? And then the player could either give them fuel, or evacuate, depending on capacities.

Gliese852:
@claudiusmueller 
> I guess within SAR I could check for any ships without fuel? Hmm.....

The `OutOfFuel` module monitors the status of ship tanks, and I think it can be taught to generate an `onShipOutOfFuel` event that your `SAR` module could catch.
https://github.com/pioneerspacesim/pioneer/blob/a7f2781dc9ddcfda6a21fad4239dbec13aff5621/data/modules/System/OutOfFuel.lua#L18

Gliese852:
@nozmajner 
> At least I suspect that most of them run out of fuel in the deceleration phase

In our case, these are almost always ships that have wasted their fuel during maneuvers while waiting for docking, so they are most likely hanging out in a low orbit.

impaktor:
In case the SAR discussion is serious, I think it would be better to just blow up the NPC ship when it runs out of fuel. Connecting SAR to it sounds like a lot of work, and to generate missions for something that's not really supposed to happen seems like difficult to debug (unless it's built into the Tradeship module that some fraction should run out of fuel and call on SAR).

Gliese852:
@fluffyfreak 
> Or just delete them... the player won't notice, and log a warning

I also thought the tradeship could be refueled a little.

fluffyfreak:
> @fluffyfreak
> 
> > Or just delete them... the player won't notice, and log a warning
> 
> I also thought the tradeship could be refueled a little.

🤔 yes, I guess that would be a less violent option 😆 NPC behaviour far from a player doesn't really need a lot of finesse, as long as things keep working they might as well have infinite fuel.

Does the player care of an NPC across the star system has fuel issues? WIll they even know?
My suggestion would be to just delete them, or refuel them, the player won't notice either way unless it happens right in front of them... so of the two, refuel seems safest unless you want to add a distance test to refuel those closer than 1 AU and delete any further away

fluffyfreak:
Always pick the "Keep It Simple Stupid" option with something like this, then make a note to fix WHY they're running out of fuel later, and even later on add a SAR mission if someone wants to do that.

claudiusmueller:
> Always pick the "Keep It Simple Stupid" option with something like this, then make a note to fix WHY they're running out of fuel later, and even later on add a SAR mission if someone wants to do that.

This gets my vote. I'll make a note of the idea to maybe hook up SAR to out of fuel ships irrespective of which module they come from. But that's outside of this scope.

Gliese852:
Sorry, I made a small force push, otherwise LifeSupport check would have been disabled when docking, and wouldn't turn on again.

fluffyfreak (approved):

Gliese852:
Tested this, apparently, after more than a year, there are approximately the estimated number of tradeships.

image

Although fps does drop over time - this is due to the fact that Mola Ramsayi accumulate over the pads and cannot dock - the model does not have a "landing spike". A very detailed collision test going on there and it consumes a lot of CPU. @nozmajner is already aware and promised to correct the model. And this will be the final nail in the coffin #5735

Gliese852:
I did a cherry-pick of #5785, and left the computer for ~6 hours at x10000, (7 years in the game), good to see performance hasn't changed, and the number of ships is approximately estimated. So we can say that this PR fixes #5735 completely.
image

Web-eWorks (approved):

PR #5789, opened 2024-03-03
Improve narrow-phase collision physics performance

Web-eWorks:
This PR makes the performance of the narrow-phase ("are these objects *really* colliding?") collision detection pass more betterer, with an average gain for nearly-colliding ship-station pairs on the order of 1.5x-5x. It's slightly slower for interpenetrating objects, as the old code saturated the maximum 8 collision contacts without traversing as many layers of the BVH, but that's not the scenario we should be optimizing for.

I've intentionally structured the commits so that intermediate commits can be built and tested if needed to check performance numbers across multiple systems. This PR strongly needs testing by as many people as possible, as it has a high possibility of completely broken physics if bugs slip through.

Please note when testing that this PR does *not* improve the **quality** of the physics - if you were able to fall through the landing pad before, you'll be able to fall through the pad now. It also only affects object-object collision, so ships vs. stations, buildings, and other ships. Collision with terrain is not affected by this PR.

I'm mainly looking for feedback about the performance and any regressions - if you phase entirely through an object that is solid in `master`, that's a problem. Review is welcome if anyone feels like it, but testing is more important, especially if you want this in the upcoming 2024-03-14 release.

CC @nozmajner @fluffyfreak @JonBooth78(?) for sanity-testing on Windows.


fluffyfreak:
Looks good, given it a test and not had any problems so far 👍

impaktor:
> given it a test and not had any problems so far

Perhaps not the first time you're back in Pioneer after hiatus, but I imagine it would be a mix of shock of all the new stuff intertwined with stuff being like they used to.

src/collider/BVHTree.cpp
```diff
@@ -322,91 +223,138 @@ uint32_t SingleBVHTree::Partition(SortKey *keys, uint32_t numKeys, const AABBd &
...
+
+	// It is ~10% faster to sort the indices than to sort the whole AABB array
+	// (cache hit rate vs. memory bandwidth)
+	// Sorting in general is extremely fast.
+	while (startIdx < endIdx) {
+		if (keys[startIdx].center[axis] < pivot) {
+			startIdx++;
+		} else {
+			endIdx--;
+			std::swap(keys[startIdx], keys[endIdx]);
```
Gliese852, line 301: This swap turns out to be useless if the element at the end is also greater.  Generally, if the array is already sorted, it seems to make a lot of unnecessary swaps. Maybe it's possible to check the element in `endIdx` before swap? (applies to all places where this algorithm is used)

src/collider/BVHTree.cpp
```diff
@@ -322,91 +223,138 @@ uint32_t SingleBVHTree::Partition(SortKey *keys, uint32_t numKeys, const AABBd &
...
-	// Push the root node
-	stack[stackLevel++] = 0;
+// ============================================================================
 
-	while (stackLevel > 0) {
-		const SingleBVHTree::Node *node = &m_nodes[stack[--stackLevel]];
 
-		if (!nodeAabb.Intersects(node->aabb))
-			continue;
+void BinnedAreaBVHTree::BuildNode(Node *node, SortKey *keys, uint32_t numKeys, const AABBd *objAabbs, uint32_t height)
```
Gliese852, line 230: Two identical functions, maybe it’s still possible to do something about it?

But all I came up with is a template static method in a [friend base](https://github.com/pioneerspacesim/pioneer/commit/d8ae46ba557ff1cdc723e57f0aee495edf85a0b7)

src/collider/BVHTree.cpp
```diff
@@ -322,91 +223,138 @@ uint32_t SingleBVHTree::Partition(SortKey *keys, uint32_t numKeys, const AABBd &
...
-	// The SAH metric doesn't include the cost of the root node
-	outSAH -= 1.0;
+	AABBd leftBox = AABBd::Invalid();
+	AABBd rightBox = AABBd::Invalid();
+	uint32_t leftSum = 0;
+	uint32_t rightSum = 0;
 
-	return outSAH;
+	// Calculate the left-side SAH cost of each splitting plane
+	for (int i = 0; i < MAX_BIN; i++) {
```
Gliese852, line 340: warning "comparison of integers of different signs"

src/collider/BVHTree.cpp
```diff
@@ -322,91 +223,138 @@ uint32_t SingleBVHTree::Partition(SortKey *keys, uint32_t numKeys, const AABBd &
...
+		leftSum += bins[i].objCount;
+		leftBox.Update(bins[i].bounds);
+		planeCost[i] = leftSum * leftBox.SurfaceArea();
+	}
+
+	// Calculate the right-side SAH cost of each splitting plane
+	for (int i = MAX_BIN - 1; i >= 0; i--) {
+		rightSum += bins[i].objCount;
```
Gliese852, line 348: `bins[MAX_BIN]` not considered anywhere.

src/collider/BVHTree.h
```diff
@@ -84,30 +30,84 @@ class SingleBVHTree {
...
+protected:
+	struct SortBin {
+		AABBd bounds = {};
+		uint32_t objCount = 0;
+	};
+
+	// 12 bins to sort triangles into provides a decent quality to performance
+	// tradeoff, especially as this is run once for each axis.
+	// This provides split lines at 1/2, 1/3, 1/4, and 1/6ths, which has been
+	// experimentally shown to provide a better SAH for human-authored models.
```
Gliese852, line 105: Any link to the experiments would be great.

src/collider/CollisionContact.h
```diff
@@ -14,9 +14,9 @@ struct CollisionContact {
 	double distance; // distance travelled to hit point
 	double timestep;
 	int triIdx;
-	void *userData1, *userData2;
 	int geomFlag;
-	
+	void *userData1, *userData2;
```
Gliese852, line 18: Now it gives me a warning: `field 'userData2' will be initialized after field 'geomFlag'`

src/collider/Geom.cpp
```diff
@@ -73,160 +68,161 @@ void Geom::CollideSphere(Sphere &sphere, void (*callback)(CollisionContact *)) c
...
+	contacts.reserve(max_contacts);
+
 	/* Collide this geom's edges against tri-mesh of geom b */
 	transTo = b->m_invOrient * m_orient;
-	this->CollideEdgesWithTrisOf(max_contacts, b, transTo, callback);
+	this->CollideEdgesWithTrisOf(contacts, max_contacts, b, transTo);
 
 	/* Collide b's edges against this geom's tri-mesh */
-	if (max_contacts > 0) {
+	if (contacts.size() < max_contacts) {
```
Gliese852, line 85: Warning about signs. (and in two more places)

src/collider/Geom.cpp
```diff
@@ -73,160 +68,161 @@ void Geom::CollideSphere(Sphere &sphere, void (*callback)(CollisionContact *)) c
...
-		// does the edgeNode (with its aabb described in 6 planes transformed and rotated to
-		// b's coordinates) intersect with one or other of b's child nodes?
-		if (triNode->triIndicesStart || edgeNode->triIndicesStart) {
-			// reached triangle leaf node or edge leaf node.
-			// Intersect all edges under edgeNode with this leaf
-			CollideEdgesTris(maxContacts, edgeNode, transTo, b, triNode, callback);
+	int stackpos = 0;
+
+	// Reserve a conservative worst-case amount of memory to store node pairs
+	stackobj *stack = new stackobj[std::max(edgeBvh->GetHeight(), triBvh->GetHeight()) * 2 + 1];
```
Gliese852, line 129: Why did you decide to make allocations here?

src/collider/Geom.cpp
```diff
@@ -73,160 +68,161 @@ void Geom::CollideSphere(Sphere &sphere, void (*callback)(CollisionContact *)) c
...
+			const bool left = rotAabb.Intersects(triBvh->GetNode(triNode->kids[0])->aabb);
+			const bool right = rotAabb.Intersects(triBvh->GetNode(triNode->kids[1])->aabb);
+
+			if (left & right) {
+				// Recurse into edge nodes until we find one that's smaller than the tri node
+				// (or hit a single edge leaf)
+				stack[++stackpos] = stackobj { edgeNode->kids[1], curr.triNode };
+				stack[++stackpos] = stackobj { edgeNode->kids[0], curr.triNode };
+			} else if (left) {
+				// Triangle BVH node is larger than the edge, split it and try again
```
Gliese852, line 167: Please explain why "larger"? It seems that it can be of any size, it just intersected not with two but with one edge node.

src/collider/Geom.cpp
```diff
@@ -73,160 +68,161 @@ void Geom::CollideSphere(Sphere &sphere, void (*callback)(CollisionContact *)) c
...
+	const vector3f _from(float(v1.x), float(v1.y), float(v1.z));
+
+	vector3d dir = transToB.ApplyRotationOnly(vector3d(edge.dir));
+	isect_t isect;
+
+	isect.dist = edge.len;
+	isect.triIdx = -1;
+
+	isect_buf.clear();
+	b->GetGeomTree()->GetTriTree()->TraceRay(v1, 1.0 / dir, edge.len, isect_buf, triNode);
```
Gliese852, line 200: `dir` components can be 0.

src/collider/Geom.cpp
```diff
@@ -73,160 +68,161 @@ void Geom::CollideSphere(Sphere &sphere, void (*callback)(CollisionContact *)) c
...
+
+	// TODO
+	// This is subtly dependent on overwriting intersections with the last triangle to be processed
+	// Ideally all contacts here would be reported...
+	vector3f traceDir = vector3f(dir);
+	for (uint32_t tri : isect_buf) {
```
Gliese852, line 206: Why go through all the items here then? You can just take the last one.

src/collider/Geom.cpp
```diff
@@ -73,160 +68,161 @@ void Geom::CollideSphere(Sphere &sphere, void (*callback)(CollisionContact *)) c
...
+	contact.distance = isect.dist;
+
+	contact.depth = edge.len - isect.dist;
+	contact.triIdx = isect.triIdx;
+	contact.userData1 = m_data;
+	contact.userData2 = b->m_data;
+	// contact geomFlag is bitwise OR of triangle's and edge's flags
+	contact.geomFlag = b->m_geomtree->GetTriFlag(isect.triIdx) | edge.triFlag;
+
+	contacts.push_back(contact);
```
Gliese852, line 227: `contact.timestep` is not defined.

src/collider/GeomTree.cpp
```diff
@@ -24,9 +24,8 @@ GeomTree::GeomTree(const int numVerts, const int numTris, const std::vector<vect
 	m_triFlags(triFlags)
 {
 	PROFILE_SCOPED()
+
 	assert(static_cast<int>(vertices.size()) == m_numVertices);
-	Profiler::Timer timer;
-	timer.Start();
 
 	m_aabb.min = vector3d(FLT_MAX, FLT_MAX, FLT_MAX);
```
Gliese852, line 30: `DBL_MAX` ?

src/collider/GeomTree.cpp
```diff
@@ -132,18 +112,40 @@ GeomTree::GeomTree(const int numVerts, const int numTris, const std::vector<vect
...
+		PROFILE_SCOPED_DESC("GeomTree::CreateEdgeTree");
+		m_edgeTree.reset(new BinnedAreaBVHTree());
+		m_edgeTree->Build(bounds, m_edgeAABBs.get(), m_numEdges);
+	}
+
+	{
+		PROFILE_SCOPED_DESC("GeomTree::CreateTriTree");
+
+		m_triAABBs.reset(new AABBd[m_numTris]);
+		for (size_t i = 0; i < m_numTris; i++) {
```
Gliese852, line 135: Warning about signs. (and in one more place)

src/collider/GeomTree.cpp
```diff
@@ -189,90 +203,34 @@ GeomTree::GeomTree(Serializer::Reader &rd)
...
-
-	return ((lmax >= 0.f) & (lmax >= lmin) & (lmin < isect->dist));
 }
 
 void GeomTree::TraceRay(const vector3f &start, const vector3f &dir, isect_t *isect) const
 {
-	TraceRay(m_triTree->GetRoot(), start, dir, isect);
-}
+	std::vector<uint32_t> tri_isect;
+	m_triTree->TraceRay(vector3d(start), vector3d(1.f / dir), FLT_MAX, tri_isect);
```
Gliese852, line 230: `dir` components can be 0.

src/collider/GeomTree.cpp
```diff
@@ -328,9 +286,14 @@ void GeomTree::Save(Serializer::Writer &wr) const
...
 	wr.Double(m_aabb.radius);
 
+	// TODO: the entire edge and triangle BVH should be written to disk and
+	// loaded in future SGM versions rather than being re-computed on each load
+
 	wr.Int32(m_numEdges);
 	for (Sint32 iAabb = 0; iAabb < m_numEdges; ++iAabb) {
-		wr << m_aabbs[iAabb];
+		AABBd &aabb = m_edgeAABBs.get()[iAabb];
+		// Write back an old-style min-max-radius Aabb for compatibility with old SGM versions
```
Gliese852, line 295: Why do we need compatibility with the old SGM format, we update everything with every sneeze.

Gliese852 (commented):
It was very interesting to dive into this topic, I hope my nitpicking is useful!

src/Aabb.h
```diff
@@ -119,6 +119,13 @@ struct AABB {
 
 	auto IsIn(Vec3 point) const { return point >= min && point <= max; }
 	auto Intersects(const AABB &rhs) const { return min < rhs.max && max > rhs.min; }
+
+	// Compute the surface area of this AABB for surface area heuristic computations
+	auto SurfaceArea() const
+	{
+		Vec3 size = max - min;
+		return Number(2.0) * size.x * size.y * size.z;
```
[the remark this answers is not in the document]
Web-eWorks: You are entirely correct. I don't remember where I got that equation for "area" from, but it's definitely not the right one.

src/collider/BVHTree.cpp
```diff
@@ -9,185 +9,129 @@
...
+	m_nodes.clear();
+	m_nodes.reserve(2 * numObjs + 1);
 
-	node->numTris = numTris;
-	node->triIndicesStart = &m_objPtrAlloc[m_objPtrAllocPos];
-	//if (objs.size()>3) Log::Debug("fat node {}\n", objs.size());
+	// compute a remapping term to express object positions with highest precision using single floating point
+	// use the average of the bounding volue to remap into [-1 .. 1] space
+	m_boundsCenter = (bounds.max + bounds.min) * 0.5;
+	vector3d inv_scale = 1.0 / (bounds.max - m_boundsCenter);
```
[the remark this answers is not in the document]
Web-eWorks: A BVH with a zero-width axis (any axis) is invalid in the general case and should never participate in collision resolution. I can early-out with a single invalid node if absolutely needed (aabb.min > aabb.max), but I'm not sure this is a problem in practice.

src/collider/BVHTree.cpp
Web-eWorks: Good catch, thank you! This was a refactoring/bugfix error late at night.

src/collider/BVHTree.h
Web-eWorks: This was experimentally determined by changing the bin count and using the model viewer's new collision information menu to check the surface area heuristic of the resulting trees. Even with the fixed SAH computation, changes to the bin count still show 12 is a good bin count for our available models.

src/collider/Geom.cpp
Web-eWorks: This is because the result from `GetTriTree()->TraceRay()` only returns intersections between the ray and the triangle's AABB. It does not guarantee an intersection between the ray and the triangle stored in the AABB leaf node. Thus, all potential contacts reported from the BVHTree need to be verified to determine if they're an actual contact against the underlying triangle data.

src/collider/Geom.cpp
Web-eWorks: Existing behavior. Resolved with explicit zero-initialization.

src/collider/GeomTree.cpp
Web-eWorks: Existing behavior.

src/collider/GeomTree.cpp
Web-eWorks: Existing code / behavior. Not going to address it here (see comment further down).

src/collider/BVHTree.cpp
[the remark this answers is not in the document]
Web-eWorks: Resolved with `vector3::ReciprocalSafe()` function.

src/collider/GeomTree.cpp
Web-eWorks: I want to address several issues with an SGM format bump (vertex formats of stored geometry, for one), so I prefer here to simply maintain compatibility with the existing wire format and then completely rewrite the serialization code for GeomTree in a separate PR.

Essentially, this PR is concerned with addressing runtime performance and I'd rather not muddy the water with an overhaul of the SGM format right now.

src/collider/BVHTree.cpp
Web-eWorks: This broadly breaks down into two categories. Either the array is "unknown" - i.e. we cannot assume the elements are sorted - or the array is already known to be sorted.

1. If the array is unknown, we cannot determine whether the element at the end that we are swapping with is greater until we read it. Either we read it before swapping and conditionally swap based on the value, or we always swap and then read it as part of the next iteration (at which point the value can be cached in a register).
2. If the array is known to be sorted, this function will perform unnecessary swaps for approximately 1/2 of the array, amortized.

This function is used almost exclusively over unsorted data, as the pivot axis and point change with every successive call to `Partition()`. Because this does not perform an in-order ascending/descending sort, but rather a simple partition over the dataset, any change to pivot axis or point invalidates the sorted state of the array.

The algorithm is intentionally designed to be as simple and unconditional as possible. There is a single less-than check - the minimum possible to implement the algorithm - and the algorithm is intended to perform as few operations as possible to make the best use of instruction and data caches.

src/collider/Geom.cpp
Web-eWorks: This is my best interpretation of an existing algorithm that was otherwise undocumented.

This seemed to be the rationale behind the decision to split of intersecting with only one edge node - it tries to ensure both the edge and triangle nodes are of approximately equal size to prevent one large parent BVH node on the left tree triggering an evaluation against all child BVH nodes on the right tree before splitting the left node.

src/collider/GeomTree.cpp
Web-eWorks: See https://tavianator.com/2011/ray_box.html - this is intentional and well defined on IEEE754-conforming implementations (i.e. every modern processor today). I've removed ReciprocalSafe() as a result, since all divide-by-zero occurrences are in fact intended.

Bodasey:
first quick thing to note: loading games from the actual master branch freezes the game. Maybe merging heals this?

`Thread 1 "pioneer" received signal SIGSEGV, Segmentation fault.
0x00007ffff72bdc82 in __dynamic_cast () from /lib64/libstdc++.so.6
(gdb)`

Bodasey:
#5875 unchanged

The ship also still enters rough surfaces while doing surface scans, but i would say it's not a big difference to the master branch. Also getting damages seems to be unchanged.

Are there more wishes for other situations to be tested?

Web-eWorks:
I think that's all that needed to be tested - if you can reliably collide with stations, starports, buildings on the ground, and other ships (i.e. spawn a ship set to kamikaze) then that's as tested as the PR can likely get.

Bodasey:
No ship to ship collision situation found - the other crashes work as intended. I'd say it's ready to merge.

src/collider/BVHTree.cpp
Web-eWorks: I've just tested this to confirm my gut is correct - changing `SingleBVHTree::Partition` (as it's actually called per-frame rather than only once at load time) to the following:
```c++
	while (startIdx < endIdx) {
		if (keys[startIdx].center[axis] < pivot) {
			startIdx++;
		} else {
			do { endIdx--; } while (endIdx > startIdx && keys[endIdx].center[axis] > pivot);
			std::swap(keys[startIdx], keys[endIdx]);
		}
	}
```
Results in 2-4us slower execution time (15us -> 17-19us in a test scene).

Web-eWorks:
Thank you again for the very detailed PR feedback @Gliese852!

PR #5795, opened 2024-03-13
Consider InputBindings::Axis active only if it's value is non-zero

Gliese852:
`StaticUpdate` is now called between `PollEvents` and `DispatchEvents`. In this interval in the first frame, the axis bindings are already "active", but the value has not yet been calculated.

Remove the direct dependence of axis "activeness" on bindings so that `IsActive` is always consistent with `GetValue`.

### DIFF
--- a/src/InputBindings.h
+++ b/src/InputBindings.h
@@ -189,7 +189,7 @@ namespace InputBindings {
 		// NOTE: sigc::signals cannot be copied, this function is for convenience to copy bindings only
 		Axis &operator=(const Axis &rhs);
 
-		bool IsActive() { return m_value != 0.0 || positive.IsActive() || negative.IsActive(); }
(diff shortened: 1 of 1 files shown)


impaktor:
This is to fix the un-documented bug related to some navigation issue that was introduced in the February release, if I've followed IRC conversation correctly?

Gliese852:
@impaktor
Fortunately, this problem is not present in release 20240203; it appeared later, when the editor was fixed. Yes, it often manifests itself as the inability of the ship to move in cruise mode if it is stuck at 0.

Web-eWorks:
Yes - this fixes a bug I introduced accidentally in the process of enabling editor and game window resizing. I've tested this fix locally and found no issues related to it, so it gets my thumbs-up.

Web-eWorks (approved):

PR #5796, opened 2024-03-14
Adding proper collision mesh

nozmajner:
Fixing the issue where the collision mesh was quite larger than the ship, and it overlapped the landing tag, causing collisions on takeoff. Now it has a more form-fitting collision mesh with the customary landing spike. (Did not go too overboard with the form-fitting part)
Ping @Gliese852 
image


impaktor:
Mergy mergy

PR #5801, opened 2024-03-16
Fix location recovery

Gliese852:
It turned out that the index of the system body in the path does not always coincide with its index in the save file, because when saving there is a separate indexing.

In addition, if there are gravpoints in the system, the numbering of physical bodies and system bodies also differs, what was not taken into account.

(This led to errors when recovering the location.)

Fortunately, the physical body (planet or station) has a reference to the serializable index of the system body (`index_for_system_body`), and this way the reliable connection can be found.

### DIFF
--- a/data/pigui/modules/new-game-window/location.lua
+++ b/data/pigui/modules/new-game-window/location.lua
@@ -316,6 +316,26 @@ local function getSystemBodyIndexForFrameID(id, frame)
 	end
 end
 
+-- traverse the system bodies tree depth-first, as happens when assigning
+-- indexes during serialization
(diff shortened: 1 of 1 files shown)


Web-eWorks (approved):
Approved pending testing. Technically, it's more efficient to calculate `systemBodies` only after checking for `dockedToBody ~= 0`, but that's not a big deal in my opinion.

Gliese852:
@Web-eWorks in fact, it is also needed in the opposite case, in the implicit else, if the ship is saved in space.

Web-eWorks (approved):

PR #5808, opened 2024-03-25
Do not install libfmt by default

### LINKED ISSUE #5805: cmake tries to install libfmt
I'm trying to fix the new archlinux pioneer package. All commands are taken from their build script here.
https://gitlab.archlinux.org/archlinux/packaging/packages/pioneer

### Observed behaviour
After build cmake --install tries to install libfmt in the system

### Expected behaviour
libfmt is not installed by cmake

### Steps to reproduce
Build with 
```bash
cmake -S . -B build -G Ninja \
  -D CMAKE_INSTALL_PREFIX:PATH=/usr/local \ # Changed the prefix path because I don't want to put non-packaged stuff in /usr
  -D PIONEER_DATA_DIR:PATH=/usr/local/share/pioneer/data \
  -D USE_SYSTEM_LIBGLEW:BOOL=ON \
  -D USE_SYSTEM_LIBLUA:BOOL=ON \
  -D CMAKE_EXPORT_COMPILE_COMMANDS=1 \
  -Wno-dev

cmake --build build -j 10 --target all build-data
```
Install with
```bash
cmake --install build --strip
```

__My pioneer version (and OS):__ Sources from release 20240314 on ArtixLinux (Arch based, the only difference is the init)

EDIT: This is the same on Archlinux official build servers.

orhun:
closes #5805

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,6 +258,8 @@ find_package(Threads REQUIRED)
 find_package(Freetype REQUIRED)
 find_package(OpenGL REQUIRED)
 
+set(FMT_INSTALL OFF CACHE BOOL "Enable install of libfmt" FORCE)
+
 add_subdirectory(contrib/lz4)
 add_subdirectory(contrib/fmt)
 


Web-eWorks (changes requested):
Modifying the CMakeLists.txt provided by libfmt means this change will be overwritten every time libfmt is updated. I'd strongly recommend you `set(FMT_INSTALL OFF CACHE)` the variable (possibly using FORCE) from the top-level CMakeLists.txt before the `add_subdirectory(contrib/fmt)` call.

Web-eWorks:
Thank you for taking the time to address this issue!

orhun:
Made the recommended change in c73781675cef2e2062f87a5760f6ea492569baa1

Not sure if we still need to keep the `option`. Let me know how it looks!

Web-eWorks:
I'll try to test this later today, but there should be no need to keep the modification to `contrib/fmt/CMakeLists.txt`.

pcercuei:
@orhun if you `git rebase -i origin/master`, it will open a text editor, in which you can remove the "do not install libfmt by default" commit, and the revert commit, and only keeping the "Set FMT_INSTALL..." commit. Once you have that, you can force-push your PR with `git push -f`.

orhun:
I'm not sure why the CI fails.

pcercuei:
@orhun well it fails locally too. This is not the right syntax, it's missing a description string. This works:
```cmake
set(FMT_INSTALL OFF CACHE BOOL "Enable install of libfmt" FORCE)
```

orhun:
Updated, thanks!

Web-eWorks (approved):

Web-eWorks:
Thanks for the patience! Tested locally and it all looks good.

Issue #5202, opened 2021-05-31
Police not intervening to protect Cargo run from pirates


tropicalpioneer:
### Observed behaviour
Upon arrival in the destination system, the pirate ship engages my ship and I received a Comm message about the fired shots and sending for help. 
No support is approaching, the police keeps at a 15km distance even if I try to get closer (looks like it keeps an escort distance).

I suspect there is something to be fixed, as well because when I managed to get the chasing pirate closer to the police, the police shot some fire at the pirate, but stopped when me and pirate got back at >6-7 km distance.

### Expected behaviour
Police should approach and engage the enemy until destroyed.

### Steps to reproduce
load the saved game, enjoy the police apathy while you try to aim with a bluenose at a fighter.

[issue police apathy.zip](https://github.com/pioneerspacesim/pioneer/files/6571397/issue.police.apathy.zip)
__My pioneer version (and OS):__
Mint Cinnamon 20.1
20210513 (0c84df0)

DamagedEngine:
I am still experiencing similar issues in ver 20250501 (https://github.com/pioneerspacesim/pioneer/commit/a2313f5be18e9935c6e24e171090778f9ff8021a) on Windows, but during VIP passenger missions instead of cargo runs. I unfortunately don't have a separate save file for such a situation, but the current police behavior is as follows:

1. I pick up a VIP passenger mission to transport someone wanted by the mafia rather than the police
2. I arrive at the destination system. There are a mafia/pirate ship and a police ship waiting for me at the hyperspace exit point.
3. The mafia ship sends their threat message and begins to intercept me. This triggers the police ship to respond with something along the lines of "Piracy will not be tolerated here. [Player ship ID], you are free to open fire."
4. The mafia ship proceeds to intercept me and shoot at me while the police ship does not seem to plot any intercept course and just remains near its starting point.

In this case the police ship was a Pumpkinseed while the mafia ship was a slower ship, so the police ship should have been easily capable of intercepting the mafia ship and destroying it.

I would expect the police ship to attack the mafia ship at least by the time it opens fire on me and begins to cause hull damage.

zonkmachine:
> Upon arrival in the destination system,

I haven't looked into the save game but I wonder if the police patrol at place may have been doing a scan of your ship. If you get scanned for contraband, etc., upon arriving in a system, the patrol seem to have no method to proceed after the job is done, so they will just stay there inactive until you leave. If you then are fired upon, you can't leave.

PR #5809, opened 2024-03-29
Police patrol improvement

### LINKED ISSUE #5603: Police Issues
### Observed behaviour
- Firing a weapon or torpedo at pirate or into empty space when a police ship is in your vicinity (not sure what the exact range is) makes you get the piracy charge.
- Police almost never come to your aid when you are attacked by pirates
- When police do come to your aid they don't shoot pirates (at least that I saw)

### Expected behaviour
- Hitting a police or civilian ship with a weapon or torpedo while police are in the vicinity should be the only way to trigger a piracy charge. 
- Police should fire on pirates in their vicinity

You should probably be able to appeal a piracy charge if there was a pirate in range when you received the charge.

### Steps to reproduce
Enter a system that has police at the hyperspace arrival area. Fire main weapons or torpedoes at pirate (or even into empty space).

robothauler:
Not entirely perfect. But an improvement.
Fixes #5603
It does not solve the police apathy problem described in #5202


impaktor:
If we aim to do one last "bug fix release" then this would be merged after that. So far, there's just been one bug fix merged since the 3.14 release.

data/modules/PolicePatrol/PolicePatrol.lua
```diff
@@ -169,13 +163,37 @@ local onEnterSystem = function (player)
 			end)
 		end
 	end
+
+	local police = ShipDef[system.faction.policeShip]
+	Timer:CallEvery(15, function ()
+		if not Game.system or #patrol == 0 then return true end
+		if target then return end
```
Web-eWorks, line 170: This ideally should be an explicit `return false` - the function is contractually bound to return a boolean value indicating whether the timer should be cancelled. While this does work currently, it's more confusing to future readers this way.

data/lang/module-policepatrol/en.json
```diff
@@ -51,6 +51,22 @@
...
+    "description": "",
+    "message": "The threat has been eliminated. The restrictions are active again!"
+  },
+  "RESTRICTIONS_ESTABLISHED_2": {
+    "description": "",
+    "message": "Cease fire, cease fire!"
+  },
+  "RESTRICTIONS_WITHDRAWN_1": {
+    "description": "",
+    "message": "All restrictions withdrawn!"
```
Web-eWorks, line 64: This is a bit confusing to the player (and to the in-universe pilot unless it's clearly established jargon/brevity code). I'd recommend something along the lines of "\<LX-1204>, you are authorized to open fire." or "You are cleared weapons free, \<LX-1204>!", replacing the brackets with the player's ship registration ID.

Similarly, "the restrictions are active again" might be better served as "Weapon authorization withdrawn. \<LX-1204>, hold fire."

Web-eWorks (changes requested):
Thank you for tackling this issue! I have a few concerns / changes regarding the comms messages presented to the player, but overall I think this is a very useful change.

data/lang/module-policepatrol/en.json
impaktor: As for "the restrictions are active again", I'm thinking "weapons fire restrictions are now in effect" sound better (assuming it's not "affect", I hope?). Plus include ship label, as suggested above

data/modules/PolicePatrol/PolicePatrol.lua
```diff
@@ -78,50 +85,37 @@ local onShipHit = function (ship, attacker)
...
-	if piracy then return end
+	if piracy or target then return end
 
 	local police = ShipDef[Game.system.faction.policeShip]
 	if ship.shipId ~= police.id then
 		for i = 1, #patrol do
-			if ship:DistanceTo(patrol[i]) <= 4000000 then
-				Comms.ImportantMessage(string.interp(l_ui_core.X_CANNOT_BE_TOLERATED_HERE, { crime = l_ui_core.PIRACY }, patrol[i].label))
+			if ship:DistanceTo(patrol[i]) <= lawEnforcedRange then
+				Comms.ImportantMessage(string.interp(l_ui_core.X_CANNOT_BE_TOLERATED_HERE, { crime = l_ui_core.UNLAWFUL_WEAPONS_DISCHARGE }, patrol[i].label))
```
Web-eWorks, line 100: Note for future reference that `string.interp(l.SOME_STRING, { crime = ... })` is also available via a custom operator as `l.SOME_STRING % { crime = ... }`. You don't need to amend this PR to prefer the use of either one however.

Web-eWorks (approved):
The updated code looks good to me! Provisionally approved pending playtesting from myself or another contributor.

data/modules/PolicePatrol/PolicePatrol.lua
```diff
@@ -169,13 +163,37 @@ local onEnterSystem = function (player)
...
+	Timer:CallEvery(15, function ()
+		if not Game.system or #patrol == 0 then return true end
+		if target then return false end
+
+		local ships = Space.GetBodiesNear(patrol[1], lawEnforcedRange, "Ship")
+		for _, enemy in ipairs(ships) do
+			if enemy.shipId ~= police.id and enemy:GetCurrentAICommand() == "CMD_KILL" then
+				if not piracy then
+					Comms.ImportantMessage(string.interp(l_ui_core.X_CANNOT_BE_TOLERATED_HERE, { crime = l_ui_core.PIRACY }, patrol[1].label))
+					Comms.ImportantMessage(string.interp(l["RESTRICTIONS_WITHDRAWN_" .. Engine.rand:Integer(1, getNumberOfFlavours("RESTRICTIONS_WITHDRAWN"))], { ship_label = player.label }, patrol[1].label))
```
Web-eWorks, line 177: This code is passing the police ship's label as a third parameter to `string.interp` rather than the `from` parameter to `Comms.ImportantMessage`.

data/modules/PolicePatrol/PolicePatrol.lua
```diff
@@ -169,13 +163,37 @@ local onEnterSystem = function (player)
...
+					piracy = true
+				end
+				attackShip(enemy)
+				break
+			end
+		end
+		if piracy and not target then
+			Comms.ImportantMessage(string.interp(l["RESTRICTIONS_ESTABLISHED_" .. Engine.rand:Integer(1, getNumberOfFlavours("RESTRICTIONS_ESTABLISHED"))], { ship_label = player.label }, patrol[1].label))
```
Web-eWorks, line 185: Same here.

Web-eWorks (changes requested):

Web-eWorks (approved):

PR #5812, opened 2024-04-03
Fix loud white nosie after mouse steering

### LINKED ISSUE #5804: Extremely LOUD white noise until game exit after mouse steering
### Observed behaviour

And extremely LOUD white noise can be triggered by just playing the game. When it triggers, steering with the keyboard is not possible anymore. The only way to stop the white noise, is to exit back to the main menu.

### Expected behaviour

Calm and quiet game session.

### Steps to reproduce

1. Get my savestate from https://app.box.com/s/6mg4zw13oqec83yg6533t5vf43aeg6oi
2. Load the savestate in the game
3. Press the "Take off" button on the bottom bar
4. When in the air, right-click on an empty space of the screen, then drag the mouse. Repeat a few times if needed, until the bug appears.

Pioneer master (7f1cb216e)

[output.txt](https://github.com/pioneerspacesim/pioneer/files/14736177/output.txt)

neumond:
Fix #5804

### DIFF
--- a/src/DynamicBody.cpp
+++ b/src/DynamicBody.cpp
@@ -68,6 +68,7 @@ DynamicBody::DynamicBody(const Json &jsonObj, Space *space) :
 	}
 
 	// fix saves with nans
+	// SAVEBUMP: This can be removed starting with save version 91
 	if (std::isnan(m_angVel.x) || std::isnan(m_angVel.y) || std::isnan(m_angVel.z)) {
 		m_angVel = vector3d(0.0);
 	}
--- a/src/DynamicBody.cpp
+++ b/src/DynamicBody.cpp
@@ -67,6 +67,11 @@ DynamicBody::DynamicBody(const Json &jsonObj, Space *space) :
 		throw SavedGameCorruptException();
 	}
 
+	// fix saves with nans
+	if (std::isnan(m_angVel.x) || std::isnan(m_angVel.y) || std::isnan(m_angVel.z)) {
+		m_angVel = vector3d(0.0);
+	}
+
 	m_aiMessage = AIError::AIERROR_NONE;
 	m_decelerating = false;
 }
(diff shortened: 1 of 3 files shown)


pcercuei:
I can confirm, it fixes the problem. Thank you!

impaktor:
@Web-eWorks Do you want to review code before merge?

Web-eWorks:
Yes, will give this a review in the next few days. I've been very busy recently and sadly haven't gotten a chance to review any of the new PRs yet.

src/DynamicBody.cpp
```diff
@@ -67,6 +67,11 @@ DynamicBody::DynamicBody(const Json &jsonObj, Space *space) :
 		throw SavedGameCorruptException();
 	}
 
+	// fix saves with nans
```
Web-eWorks, line 70: I'd recommend this comment be annotated with the current savefile version (see Game.cpp) so that the workaround can be removed when the next savebump occurs.

Web-eWorks (changes requested):

Web-eWorks:
Other than the requested change, the code looks good - thanks for the fix!

src/DynamicBody.cpp
impaktor: I think we used to add "todo" notes to SAVEBUMP.txt as well, things to fix for next release, so that might be a good idea.
I should add "checking SAVEBUMP.txt" to the release guide, if we think this is a good strategy?

Web-eWorks: That would be a good idea on both counts; this definitely deserves to be in the SAVEBUMP.txt file as well.

Web-eWorks (approved):

Web-eWorks:
I've pushed the required documentation changes to the source branch - thanks for the contribution!

PR #5817, opened 2024-04-09
Renderer: Support running under Wayland

pcercuei:
Ignore an error from GLEW in order to run natively under Wayland (instead of going through XWayland) on Linux.

Pioneer can be forced to run with Wayland like this: SDL_VIDEODRIVER=wayland ./pioneer

### DIFF
--- a/src/graphics/opengl/RendererGL.cpp
+++ b/src/graphics/opengl/RendererGL.cpp
@@ -173,8 +173,19 @@ namespace Graphics {
 	{
 		PROFILE_SCOPED()
 		glewExperimental = true;
-		GLenum glew_err;
-		if ((glew_err = glewInit()) != GLEW_OK)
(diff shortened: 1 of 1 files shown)


Web-eWorks (commented):
Given the slightly-hacky nature of this change, I'd appreciate it if a logging statement (e.g. `Output()`) was added to the if-branch to ensure log files indicate if this workaround is active for future issue triage purposes.

Otherwise, this looks sensible - thanks for testing on Wayland!

pcercuei:
@Web-eWorks Updated.

pcercuei:
If you don't like graphical operating systems, I slightly modified the code so that Pioneer now also runs on top of KMS/DRM directly.

Web-eWorks (approved):

Issue #4399, opened 2018-08-19
Lighting of ships inside stations


WKFO:
I'm kind of suprised no one opened this issue before.

### Observed behaviour
Inside of the space stations are pretty bright, but ships are not illuminated by the station so they become black after the station goes to the dark side of the planet, while the internal walls of the stations stay bright. I know stations don't have actual light sources and the lighting is just a texture trick or sth, but it just feels very wrong.

### Expected behaviour
Perhaps stations should illuminate the ships inside.

### Steps to reproduce
Dock with a station on the night side of a planet.

__My pioneer version (and OS):__
Latest github master

nozmajner:
Yes, it is a known issue. @Fluffyfreak thought about how you could solve it if I'm correct, but not sure how far he got.
The problem is, the station inside aren't lit at all, it is shadeless, and all lighting information is on the texture. Which won't look good on ships in my opinion.
Not that I think of it, maybe it could be done by placing  a (couple of) virtual sun inside the station with some range setting, and ramp up the ambient light level, as you go inside. No cast shadows and such nuissance. Would that work?

ConfedMajor:
I hate this a lot, one time, i entered a station in a system that has a red dwarf, and it was beneath the station...

The ship looked like half black, half red... it was pretty weird...

If the lightning has shadow casting, the station will go completely dark (or maybe a little bit with light if a star is in front of the exit or enter)

So, that is why stations should have inner lights, and also this is why pioneer should have shadow casting, but this is a very very very expensive thing for a OpenGL game

fluffyfreak:
My plan was just to turn off the lighting, or transition it from lit, to unlit and thus rely on the baked in diffuse material lighting as you got "inside" a station.

I mean I had other, more complex ideas, but i thought that one was achievable relatively easily

Keeper1st:
Maybe the easiest way for that is to have it stop lighting/shading on ships once they are set to docked state on a spacestation (but not a ground station). It would mean they stay dark until they dock, of course, but rarely would a player see another ship moving inside a station (other than the player's own -- but that's what you get for not flying "first person"!). I imagine that would be a kind of quick-fix without having to do any really complicated stuff.

PR #5823, opened 2024-04-15
Add interior lighting to stations

tatjam:
Makes ships that are inside stations (as defined on the model file using tags and a series of new parser commands) have increased ambient light and reduced direct lightning, fixing the issue of black ships on eclipsed stations (Issue #4399).

TO-DO:
- [x] Working interiors
- [x] Make all stations compatible with the code
- [ ] Adjustable lighting colors with some kind of station config?
- [x] Make it work with the ship cockpit and other bodies that may be found inside stations

Demo: https://www.youtube.com/watch?v=D56iUkvpOac

### DIFF
--- a/src/Camera.cpp
+++ b/src/Camera.cpp
@@ -15,6 +15,7 @@
 #include "graphics/TextureBuilder.h"
 #include "graphics/Types.h"
 #include "graphics/RenderState.h"
+#include "SpaceStation.h"
 
 using namespace Graphics;
 
@@ -150,11 +151,13 @@ void Camera::Update()
 
 	// evaluate each body and determine if/where/how to draw it
 	m_sortedBodies.clear();
+	m_spaceStations.clear();
 	for (Body *b : Pi::game->GetSpace()->GetBodies()) {
 		BodyAttrs attrs;
 		attrs.body = b;
 		attrs.billboard = false; // false by default
 		attrs.calcAtmosphereLighting = false; // false by default
+		attrs.calcInteriorLighting = false;
 
 		// If the body wishes to be excluded from the draw, skip it.
 		if (b->GetFlags() & Body::FLAG_DRAW_EXCLUDE)
@@ -222,7 +225,15 @@ void Camera::Update()
 				attrs.calcAtmosphereLighting = true;
 		}
 
+		if(b->IsType(ObjectType::SHIP)) {
+			attrs.calcInteriorLighting = true;
+		}
+
 		m_sortedBodies.push_back(attrs);
+
+		if(b->IsType(ObjectType::SPACESTATION)) {
+			m_spaceStations.push_back(b);
+		}
 	}
 
 	// depth sort
@@ -297,11 +308,12 @@ void Camera::Draw(const Body *excludeBody)
 		m_renderer->SetLights(rendererLights.size(), &rendererLights[0]);
 	}
 
-	std::vector<float> oldIntensities;
+	// Save lights for later restoring
+	oldLightIntensities.clear();
 	std::vector<float> lightIntensities;
 	for (size_t i = 0; i < m_lightSources.size(); i++) {
-		lightIntensities.push_back(1.0);
-		oldIntensities.push_back(m_renderer->GetLight(i).GetIntensity());
+		oldLightIntensities.push_back(m_renderer->GetLight(i).GetIntensity());
+		lightIntensities.push_back(1.0f);
 	}
 
 	Graphics::VertexArray billboards(Graphics::ATTRIB_POSITION | Graphics::ATTRIB_NORMAL);
@@ -319,23 +331,11 @@ void Camera::Draw(const Body *excludeBody)
 			continue;
 		}
 
-		double ambient = 0.05, direct = 1.0;
-		if (attrs->calcAtmosphereLighting)
-			CalcLighting(attrs->body, ambient, direct);
-
-		for (size_t i = 0; i < m_lightSources.size(); i++)
-			lightIntensities[i] = direct * ShadowedIntensity(i, attrs->body);
-
-		// Setup dynamic lighting parameters
-		m_renderer->SetAmbientColor(Color(ambient * 255, ambient * 255, ambient * 255));
-		m_renderer->SetLightIntensity(m_lightSources.size(), lightIntensities.data());
-
+		PrepareLighting(attrs->body, attrs->calcAtmosphereLighting, attrs->calcInteriorLighting);
 		attrs->body->Render(m_renderer, this, attrs->viewCoords, attrs->viewTransform);
 	}
 
-	// Restore default ambient color and direct light intensities
-	m_renderer->SetAmbientColor(Color(255, 255, 255));
-	m_renderer->SetLightIntensity(m_lightSources.size(), oldIntensities.data());
+	RestoreLighting();
 
 	if (!billboards.IsEmpty()) {
 		Graphics::Renderer::MatrixTicket mt(m_renderer, matrix4x4f::Identity());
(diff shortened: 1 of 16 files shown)


fluffyfreak:
That's an interesting approach, another might be to gather the nearest 3 or 4 lights from within the station and apply them as light sources to the ship. However I'm not sure if "lights" like that are defined in the model or if that's something that could be done later.

nozmajner:
There are no lights in the model. I can add them if needed. Not sure if Blender lights are exported to collada, with their settings. Or they could be tags.

tatjam:
Love the idea, that way it's more in-line with the new PBR renderer idea. Guess adjusting ambient lighting is useful even then, if no dynamical env-maps are used.

Even if actual lights are used I guess the interior checking is important, to prevent outside ships from being lit. (Otherwise shadow casting is needed, or really thick exterior walls).

nozmajner:
I'd suggest first doing the ambient light only, get it to merge, and then you can think about the next step, or if you want to do other features instead. No need to build a huge PR, especially when you are still getting your feet wet. This is a free-time passion project after all, avoid burnout and such!

tatjam:
If I'm not mistaken all stations use either hub A o B so that should be done

nozmajner:
Yes, all stations use those two hubs.

impaktor:
@tatjam For testing big vs. small ships, I assume you're aware of Ctrl+i to change your ship model.

tatjam:
Messed around a bit with various ships and a fixed fade value feels good. There's really not any ship that's so long its center is way outside the station before it visually is outside.

impaktor:
@tatjam If you're feeling done, please rebase commits if/as needed, and remove draft status on PR

tatjam:
Fine! The custom light feature could be implemented later on if needed.

impaktor:
Going to need some review on this @Web-eWorks  or/and @fluffyfreak

Web-eWorks:
Will do - not had a chance to get to this yet but it's been on my TODO list.

src/SpaceStation.cpp
```diff
@@ -903,3 +903,20 @@ double SpaceStation::GetUndockAnimStageDuration(int bay, DockStage stage) const
...
+	bool any_inside = false;
+	vector3d b_point = b->GetPositionRelTo(this);
+	float dist = GetModel()->DistanceFromPointToBound("interior", vector3f(b_point));
+
+	// This number can be tweaked to make fading more / less gradual,
+	// it could depend on ship size but this value feels good on all current ships
+	const float bRadius = 30.0f;
+	if(dist - bRadius < 0.0f) {
+		sIntense = std::max(std::min((bRadius - dist) / bRadius, 1.0f), 0.0f);
+		sLight = Color(255, 255, 255).Shade(0.5f);
```
fluffyfreak, line 917: I think this should be the constant `Color::GRAY`

src/Camera.cpp
```diff
@@ -555,3 +555,50 @@ void Camera::PrincipalShadows(const Body *b, const int n, std::vector<Shadow> &s
...
+	ambientMix.a = 255;
+
+	m_renderer->SetAmbientColor(ambientMix);
+	m_renderer->SetLightIntensity(m_lightSources.size(), lightIntensities.data());
+
+}
+
+void Camera::RestoreLighting() const
+{
+	m_renderer->SetAmbientColor(Color(255, 255, 255));
```
fluffyfreak, line 602: Replace with `Color::WHITE`

fluffyfreak (approved):
Aside from a couple of minor things which could be using the Color constants I think this is great work 👍

src/Camera.cpp
```diff
@@ -555,3 +555,50 @@ void Camera::PrincipalShadows(const Body *b, const int n, std::vector<Shadow> &s
...
 	}
 }
+void Camera::CalcInteriorLighting(const Body* b, Color4ub &sLight, double &sFac) const
+{
+	for(const auto& ss : m_spaceStations)
+	{
+		SpaceStation* as_ss = (SpaceStation*)ss;
+		// Only one of these calls will do anything
+		if(as_ss->CalcInteriorLighting(b, sLight, sFac))
+			return;
```
Web-eWorks, line 565: This function should perform early-out rejection if the distance between the body and the space station is less than `b->GetClipRadius() + ss->GetClipRadius()`.

src/scenegraph/Parser.cpp
```diff
@@ -180,6 +180,14 @@ namespace SceneGraph {
 					throw ParseError("Animation start/end frames seem wrong");
 				m_model->animDefs.push_back(AnimDefinition(animName, startFrame, endFrame, loopMode));
 				return true;
+			} else if (match(token, "bound_thick_line")) {
```
Web-eWorks, line 183: Personally, I think `bound_cylinder` is a better choice of name here, but I'll leave it to your discretion whether thick_line is the technically correct name.

src/scenegraph/LoaderDefinitions.h
```diff
@@ -69,12 +70,36 @@ namespace SceneGraph {
 		bool loop;
 	};
 
+	struct BoundDefinition {
```
Web-eWorks, line 73: In general, I see this structure as unnecessarily future-proofed. Do you have any further plans to use `tags[2]` or `params[1]` for other bound types that you plan to implement immediately? If not, I'd recommend reducing the struct to only what's needed to express the minimum set of bounds currently in use.

Web-eWorks (changes requested):
This is certainly better than no lighting at all, and the implementation is fairly well-contained and easy to upgrade to shadow maps and punctual lights later down the line when we add support for them.

However, there are a few issues. The most notable one is that the new model bounds information will not be present when a model is converted to .SGM format when we ship a compiled build of the game, which will manifest as a complete lack of the new station lighting. The bounds information will need to be round-tripped through `src/scenegraph/BinaryConverter.cpp` in the `Save` and `Load` methods.

The second is that there is a worrying number of string comparisons and operations involved in computing interior lighting for stations. `Model::DistanceFromPointToBound` is called `N*M` times per frame, where N=number of ships on screen and M=number of stations within render distance. If a model contains more than one named bound type, there's a linear loop over bound entries by name. There is quite a bit of room for improvement here, starting with ignoring station-ship pairs with no overlap of visible cull radius. I'll leave some inline review comments where the behavior is non-optimal and presents scaling issues on lower-end hardware.

On the whole I'm quite happy with the PR and I do apologize that it took this long to review only to ask you to make major changes. I'd be happy to discuss further on the individual change requests if needed.

src/scenegraph/LoaderDefinitions.h
tatjam: I've got a feeling as most stations are going to be relatively linear, the cylinder / thick line thing is good enough. I'm going to change it to only support that, and if in the future other types are needed an union could be used.

tatjam: Actually using an union doesn't seem like a great idea if strings are used to store the tags, so probably just having different meanings for the tags. For example, an AABB can also be defined by an start and end tag, and the radius could be used for rounding.

AABBs and the cylinders can get you some really complex geometries!

src/scenegraph/Parser.cpp
tatjam: I've chosen "capsule" as a name instead, as this matches the typical capsule collider used in most game engines (cylinder with rounded endcaps)

Web-eWorks: Definitely a better choice - I would not have been able to tell that was the underlying shape from the existing nomenclature.

Web-eWorks:
Just taking a first look over the code from my phone here - named struct field initializers are (AFAIK) a c++20 feature (which we don't currently target). Would recommend double-checking for that in the last few commits.

tatjam:
Dangit! I've been lately working on a C++20 codebase and have gotten used to its quirks! Gonna change it back to simply setting the fields, which should be equivalent under any good compiler.

tatjam:
How can I get the game to try loading the binary files instead of data?

Web-eWorks:
You'll want to run the CMake `build-data` target through your editor or terminal of choice. That invokes the `modelcompiler` binary to build all models into `.sgm` files and leaves them in the install tree - you'll need to re-run the command when you modify the SGM format to rebuild the compiled models.

Keep in mind that you'll need to bump the SGM format version to add the bounds information to the binary stream that makes up an SGM file.

tatjam:
Oh, but does the game automatically load the SGM files instead of the "raw" ones?

tatjam:
Nevermind, it indeed does!

tatjam:
I've left a "TODO", because the existing SGM code appeared to save the tags: https://github.com/pioneerspacesim/pioneer/blob/a7ce69ad3d87e1dee9a620a81b6124253217c1a9/src/scenegraph/BinaryConverter.cpp#L139
but they are not really used afterwards (or it appears so at a first glance, as they should be loaded from the model itself?)

Web-eWorks (approved):
With great apologies for the length of time this sat waiting for a review - this functions well and you've addressed my prior concerns very sufficiently.

If you're still present, I'd ask that you squash together the various "fix X" commits before merge; otherwise I can merge this PR as is.

Web-eWorks:
> I've left a "TODO", because the existing SGM code appeared to save the tags:
> 
> https://github.com/pioneerspacesim/pioneer/blob/a7ce69ad3d87e1dee9a620a81b6124253217c1a9/src/scenegraph/BinaryConverter.cpp#L139
> 
> but they are not really used afterwards (or it appears so at a first glance, as they should be loaded from the model itself?)

I need to do a cleanup / fixup pass on the serialization code now that we've (finally) broken compatibility with file version 6. I wouldn't worry about this - I have a larger planned rewrite of how SGM files are internally structured.

tatjam:
No problem! I've more or less cleaned up the commit history into somehow related blocks, although they could be further reduced.

Web-eWorks:
Thank you @tatjam! Please let us know the name (handle, real name, etc.) you'd like to be listed as in AUTHORS.txt - this definitely meets the threshold to be added as a contributor. :smile:

(CC @impaktor for AUTHORS.txt / Changelog update)

tatjam:
Btw I guess Issue #4399 is solved by this?

fluffyfreak:
@tatjam I just need to echo what others have said, thankyou so much for doing this! 
It was one of the very annoying issues that needed to be fixed and I think you've done a really good job here.

PR #5826, opened 2024-04-23
Set unique names for utility programs

AMDmi3:
Prefix modelcompiler, savegamedump, editor binary names with `pioneer-` to avoid too generic names which may conflict with other packages when installed into system.


AMDmi3:
Could not fix the build on msvc. No idea why it uses base binary name while generator expression should return real output name. Probably we can fix it by renaming targets if that's ok.

impaktor:
Yup, I agree.

Would also have to change editor.txt, and COMPILING.txt. (I could push a commit for that)
(and [wiki](https://wiki.pioneerspacesim.net/wiki/Pioneer_Wiki), and maybe [devdocs](https://dev.pioneerspacesim.net/), but can do that after merge)

impaktor:
I'm spending some time on the AUR [pioneer-git](https://aur.archlinux.org/packages/pioneer-git) package, and I'd like to bump this issue before I potentially push any update to it (if/when  #6188 gets merged the AUR will need updating for sure). 

> Could not fix the build on msvc.

@fluffyfreak ?

fluffyfreak:
I don't even know what the problems are 🤷

PR #5827, opened 2024-04-23
Include stdlib.h instead of alloca.h for alloca(3) on FreeBSD

AMDmi3:


AMDmi3:
For the record, it fixes build on FreeBSD which doesn't have `alloca.h` and `alloca(3)` is defined in `stdlib.h`. I've just checked and including `stdlib.h` instead of `alloca.h` works for me on linux (ubuntu 22.04) as well, so we may consider just switching to `stdlib.h` as a simpler solution, however I'm not aware if it's portable. The best thing would be to check for `alloca.h` in cmake, but it's a bit more complicated. Can redo this PR only any of mentioned alternative solutions if it's deemed more suitable.

Web-eWorks (approved):

PR #5828, opened 2024-04-23
Include <algorithm> for std::max

AMDmi3:


impaktor:
This silences a warning, or fixes the build on some OS?

AMDmi3:
This fixes build on FreeBSD (clang + libc++).

Web-eWorks (approved):

PR #5829, opened 2024-04-25
Fix Terrain Disappearing

### LINKED ISSUE #5806: Terrain textures disappear
### Observed behaviour
When approaching the planet at a certain distance some terrain textures disappear.
On a first screenshot textures are proper
screenshot-20240324-215601
But at some point they(textures) start bugging like this
screenshot-20240324-215625
screenshot-20240324-215536
screenshot-20240324-215524

### Expected behaviour
Terrain textures should be stable and shouldn't disappear

### Steps to reproduce
Fly up to critical distance to certain planets and moons (in this case it is Pallas, I also saw this bug approaching Thebe)

ver 20240314 (8b6ae92)
OS: Windows 10 Home Edition

[output.txt](https://github.com/pioneerspacesim/pioneer/files/14736373/output.txt)
[save.zip](https://github.com/pioneerspacesim/pioneer/files/14736377/save.zip)

fluffyfreak:
Fixes #5806 ... if we're lucky

Switched to using the recalculated `m_centroid` instead of `m_clipCentroid` which is surprisingly far below the surface on asteroids. I believe some terrains generate heights which are FAR above expected which buries `m_clipCentroid` much further below the surface than expected.

All of this will need some testing with things to look out for like:
- terrain disappearing
- patches appearing and disappearing as you look around
- terrain not being generated at higher detail

### DIFF
--- a/src/Color.cpp
+++ b/src/Color.cpp
@@ -15,6 +15,7 @@ const Color4f Color4f::YELLOW = Color4f(1.0f, 1.0f, 0.0f, 1.0f);
 const Color4f Color4f::GRAY = Color4f(0.5f, 0.5f, 0.5f, 1.f);
 const Color4f Color4f::STEELBLUE = Color4f(0.27f, 0.51f, 0.71f, 1.f);
 const Color4f Color4f::BLANK = Color4f(0.0f, 0.0f, 0.0f, 0.0f);
+const Color4f Color4f::CYAN = Color4f(0.0f, 1.0f, 1.0f, 1.0f);
 
 const Color4ub Color::BLACK = Color(0, 0, 0, 255);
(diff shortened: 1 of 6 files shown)


fluffyfreak:
@Web-eWorks dammit I did some testing last night and this doesn't 100% fix it, It's much improved but I still get some drop-out :(

impaktor:
@fluffyfreak I assume it means that the understanding of the bug was mostly correct?

fluffyfreak:
@impaktor it's on the right track, I just don't understand what exactly has changed that it's become a problem suddenly

fluffyfreak:
I'm going to close this for now whilst I rethink the problem

PR #5835, opened 2024-05-18
Removing stray triangles from Bluenose

nozmajner:
Removing some stray vertices which show up on the model as oddly placed triangles. Thanks for @Maks2103 form discord for catching it.
image
image

I've also fixed some normals while I was at it.


fluffyfreak:
Were the normals... abnormal? 😉

PR #5840, opened 2024-06-17
Separate shield meshes from ship models

Web-eWorks:
Per discussion on IRC from quite a while ago, this PR implements the technical changes necessary to support ship shields being loaded from a separate SGM file and rendered as a separate entity. This solves some long-standing technical debt related to the *very* invasive process needed to prepare shields for rendering under the old architecture and opens the door to a more efficient architecture for the scenegraph system as a whole.

I've also removed the old Lua-based ship loading code, as we have not compiled with that code enabled in at least five years and it has been thoroughly deprecated at this point.

This PR also fixes a bug present in master (and likely present in many past releases) where the player's shields would not be visible when the player's ship is hit by projectiles or collides with objects.

CC @bszlrd as this PR should include or be closely followed by a set of changes separating shield meshes into their own models for all ships currently present in the game. I'm leaving the fallback path in this PR until we're ready to migrate all shields, but I'd like the old method to be completely removed before merge.

As this PR requires a savegame bump, it will not be merged until after we have released the accumulated set of hotfixes from the last few months.


impaktor:
> As this PR requires a savegame bump, it will not be merged until after we have released the accumulated set of hotfixes from the last few months.

When should we do the final bug-fix release?

Web-eWorks:
> > As this PR requires a savegame bump, it will not be merged until after we have released the accumulated set of hotfixes from the last few months.
> 
> When should we do the final bug-fix release?

Soon. I'm going through the issue list today and tomorrow to see if there are any other quick fixes we need to get in before we cut a release.

bszlrd:
Okay,will do. I can't get to it for at least a couple of weeks unfortunately though.

Web-eWorks:
Merging this - have given it local testing through development and pushed documentation related to the new shield model workflow to the dev docs repository.

PR #5842, opened 2024-06-18
Fix lots of missing ship labels

### LINKED ISSUE #5838: Ship registration number disappearing after loading saved game
### Observed behaviour
Ship's registration number stops being displayed after loading a saved game
screenshot-20240605-192602

### Expected behaviour
Registration number being displayed properly
screenshot-20240605-192528

### Steps to reproduce
1) Start a new game
2) Save game
3) Exit to main menu
4) Load saved game

Pioneer version: public release from 14.03.2024
OS: Windows 10 x64

[output.txt](https://github.com/user-attachments/files/15593659/output.txt)

[RegNum_TestSave.zip](https://github.com/user-attachments/files/15593664/RegNum_TestSave.zip)

Also adding this, just in case, because render-related
[opengl.txt](https://github.com/user-attachments/files/15593680/opengl.txt)

Web-eWorks:
This PR fixes #5838, and in the process fixes a number of spots where ship labels either were not appearing or may have been incorrect prior.

- Ship labels now display and update in the new game start window. Modifying the ship label adjusts it on the craft as well, if the craft has a properly-named label node.
- Ships on sale in a space station now display their labels on the craft if the craft model supports labels.
- The paint shop technically appeared correct but would have resulted in the same missing label bug when saving and loading after applying the paint.
- SearchRescue spawned ships will continue to show labels (they technically were painting a label on the ship hull and then attempting-but-failing to clear it).

CC @Gliese852 - I've made minor changes to the new game window widget API and hooked a few things up to get labels to update in sync with the model spinner; just want to double check I haven't inadvertently introduced new bugs :smile:
CC @bszlrd - the Coronatrix at the very least has improperly named label nodes, causing them to be picked up as tags instead. Label nodes need to be named `label_*`; `tag_label_*` results in a tag instead. There is no need to address this immediately, but if you get five minutes it might be worth a quick fix.

### DIFF
--- a/data/libs/SpaceStation.lua
+++ b/data/libs/SpaceStation.lua
@@ -576,11 +576,12 @@ local function addRandomShipAdvert(station, num)
 		local def = avail[Engine.rand:Integer(1,#avail)]
 		local model = Engine.GetModel(def.modelName)
 		local pattern = model.numPatterns ~= 0 and Engine.rand:Integer(1,model.numPatterns) or nil
+		local label = Ship.MakeRandomLabel()
 		addShipOnSale(station, {
 			def     = def,
-			skin    = ModelSkin.New():SetRandomColors(Engine.rand):SetDecal(def.manufacturer),
+			skin    = ModelSkin.New():SetRandomColors(Engine.rand):SetDecal(def.manufacturer):SetLabel(label),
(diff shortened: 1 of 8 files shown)


Web-eWorks:
Picture!
image

PR #5850, opened 2024-07-01
Fix DLLs not being installed on Windows builds

Web-eWorks:
Potential fix for DLLs not being copied into the release archive for Windows builds. Not sure what broke this, as it was working at least a few months ago if not sooner.


Web-eWorks:
For posterity: the DLL installation step was moved to before `MSVC_ARCH` was defined in the build script.

PR #5856, opened 2024-07-05
Fix Military Drive 2 not for sale & tweaks

impaktor:
This PR contains multitides:

## 1. Fix a bug
Class 2 military drives require "MILITARY" tech level, which under the hood is interpreted as tech-level 11 (because that's how Frontier implemented it). Higher military drives are available at higher tech levels. Code had a bug where in stations of higher tech level than "MILITARY", you could only buy military drive of class 1, 3, and 4, but not military drive class 2.

I note there's equipment that has "tech level = 11" set, that is technically the same as tech level "Military". Seems to be the way I did it in #3505, so looks intentional, but not sure what the reasoning was.

I saw the bug when watching the "Pioneer - Frequently Failing" play through of the game (last screenshot from Poseidon in orbit around Neptune, level 12 station):
mpv-shot0001
2024-07-05-154225_812x396_scrot

## 2. Tweak suggestion
Don't generate "Unknown commodity" to scoop - Removing this since player might be tricked to think this has value, being some "discovery" of alien origin, sought after by someone.

### DIFF
--- a/data/modules/Debug/DebugRPG.lua
+++ b/data/modules/Debug/DebugRPG.lua
@@ -7,7 +7,6 @@ local ui = require 'pigui'
 local debugView = require 'pigui.views.debug'
 local Commodities = require 'Commodities'
 local amount = 1000
-local selected = 0
 
 local Legal = require "Legal"
 local utils = require "utils"
@@ -16,7 +15,7 @@ local l = Lang.GetResource("ui-core")
 
 -- build list of all crime types:
 local crime_types = {}
-for k, v in pairs(Legal.CrimeType) do
+for k, _ in pairs(Legal.CrimeType) do
 	table.insert(crime_types, k)
 end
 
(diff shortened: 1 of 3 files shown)


robothauler:
Regarding point 2, maybe rename it to "Empty" or something else.

impaktor:
## 3. FIx News Event demand
Now that we have and actual "demand" mechanic in the game (how many units the seller is allowed to sell / market is willing to buy), this `demand` should also be adjusted accordingly.

For the future: 
1. The demand should approach zero as deadline approaches
2.  perhaps total demand should be the equivalent of 1-3 player ship filled with the commodity?

Actually, I think I'll look into 1 above tomorrow.

impaktor:
OK, I've split off News event / commodity work on a separate branch, as that will be some work beyond "small commits". I consider this ready for review.

Web-eWorks (approved):
CommodityType already maintains save game compatibility even with removed commodity types, so this should be good to merge.

PR #5859, opened 2024-07-09
Fix demand at News Events not being used

impaktor:
Demand mechanics was never ported into News Event module, thus when there's a large demand for something at a station, the demand could still be very low, preventing player from selling. Also, we don't want to accumulate price/stock/demand at each "OnDock" event.

EDIT: I'll open this as WIP, for feedback, the first and last commits, that both attempt to fix the above bug should be squashed, as first does it wrongly, I believe.

Also I've added a debug tab, which allows generating (random) news. Usually news only spawn +6 months into the game, and only when jumping / in hyperspace.
2024-07-09-212655_1536x960_scrot

### DIFF
--- a/data/modules/NewsEventCommodity/NewsEventCommodity.lua
+++ b/data/modules/NewsEventCommodity/NewsEventCommodity.lua
@@ -23,7 +23,6 @@ local Event = require 'Event'
 local Format = require 'Format'
 local Serializer = require 'Serializer'
 local Commodities = require 'Commodities'
-local Equipment = require 'Equipment'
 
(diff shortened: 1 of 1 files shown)


impaktor:
I've tried it with both "sold out" and "surplus" event, and note price/stock/demand is modified when docking, and reset when undocking. 

I've also tested it with illegal goods. (I realize news event is a good at finding where to buy illegal goods. if the event is for e.g. battle weapons, it will cause a shortage in some system where those are a legal commodity. I had forgotten that was how it worked)

...then I did a rebase. That _should_ not have messed up anything.

PR #5861, opened 2024-07-11
Fix debug commodities tab

impaktor:
- Fix standard deviation lacking a `sqrt`
- Add filtering on `purchasable` goods on/off (sadly we can't buy the fuzzy little Quibbles on the commodity market)
- Add some collapsing headers UI stuff when selected a commodity, for bling bling
- Added some info on commodity at current station (if docked). I used this for #5859, and might need it again. It's not über-pretty, but oh, well `¯\_(ツ)_/¯`

## Collapsed
2024-07-11-163357_1536x960_scrot
## Expanded
2024-07-11-163416_1536x960_scrot
## Include non-purchasable
2024-07-11-163555_1536x960_scrot
## Some info of station levels
2024-07-11-163903_1536x960_scrot

### DIFF
--- a/data/modules/Debug/DebugCommodityPrices.lua
+++ b/data/modules/Debug/DebugCommodityPrices.lua
@@ -4,6 +4,7 @@
 
 local Game = require "Game"
 local ui = require 'pigui'
+local Economy = require 'Economy'
 local debugView = require 'pigui.views.debug'
 local Format = require 'Format'
 local Commodities = require 'Commodities'
@@ -13,16 +14,18 @@ local radius = 20
 local N = 0
 local commodities = nil
 local include_illegal = false
+local include_only_purchasable = true
 
 -- Create class to document our commodity
 local Commodity = {}
 Commodity.__index = Commodity
 
-function Commodity.new (o, name, baseprice)
+function Commodity.new (o, name, id_name, baseprice)
 	-- o = o or {}
 	local self = setmetatable({}, Commodity)
 	-- self.__index = self
-	self.name = name		    -- name of commodity
+	self.name = name			-- name of commodity
+	self.id_name = id_name		-- key of commodity
 	self.max = 0				-- max price for this commodity
 	self.min = 1e9				-- min price for this commodity
 	self.prices = {}			-- table of every price for this commodity
@@ -159,22 +162,21 @@ local scan_systems = function(dist)
 			-- include goods if legal, or if we've chosen to include it
 			local include = include_illegal or sys:IsCommodityLegal(comm.name)
 
-			-- TODO: also include negative products, right?
-			-- if comm.price > 0 then
+			if include_only_purchasable then
+				include = include and comm.purchasable
+			end
 
-				local name = comm:GetName()
-				local price = getprice(comm, sys)
+			local name = comm:GetName()
+			local price = getprice(comm, sys)
 
-				if not commodities[name] then
-					commodities[name] = Commodity:new(name, comm.price)
-					print(name, comm.price)
-				end
+			if include and not commodities[name] then
+				commodities[name] = Commodity:new(name, comm.name, comm.price)
+			end
 
-				-- if interesting, get name, and check price
-				if include then
-					commodities[name]:add_price(price, sys)
-				end
-			-- end
+			-- if interesting, get name, and check price
+			if include then
+				commodities[name]:add_price(price, sys)
+			end
 		end
 	end
 	return commodities, #nearby_systems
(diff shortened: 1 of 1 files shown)


data/modules/Debug/DebugCommodityPrices.lua
```diff
@@ -276,37 +262,90 @@ local function show_details_on_commodity(commodities, clicked)
...
+
+		ui.plotHistogram("##price distribution", data, #data, 0, "", 0, y_max, Vector2(400, 100))
+		ui.text("Min: " .. Format.Money(c.min))
+		ui.text("Max: " .. Format.Money(c.max))
+	end
+	return purchasable
 end
 
 
+local function station_economy(commodities, clicked, station)
```
impaktor, line 309: @Web-eWorks If you intend to review this PR, then this function is the one where I'm half stumbling around trying to understand the economy system. 

Basically, what info is meaningful to display, are the ones I have correctly named?

For completeness, I'll add this uncommitted comment I have on my local copy, adapted from IRC conversation the other day:
```
	-- GetCommodityStockFromFlow() returns "equilibrium" stock/demand,
	-- and price of the commodity is stored as a "pricemod", a
	-- percent-of-100 offset to the commodity's base price to compute
	-- market price for e.g. oxygen (substitute any commodity ID),
	-- you'd call
	--
	--    Economy.UpdateCommodityPriceMod(sBody, "oxygen", { stock, demand, pricemod }),
	--
	-- which stores the new pricemod in the table passed as the 3rd argument.
	-- You can then call
	--
	--    Economy.GetMarketPrice(commodity.price, pricemod)
	--
	-- to get the credit cost of the commodity
```

data/modules/Debug/DebugCommodityPrices.lua
```diff
@@ -175,24 +160,27 @@ local scan_systems = function(dist)
 			-- table.sort(comm, function (a,b) return a:GetName() > b:GetName() end)
 
 			-- include goods if legal, or if we've chosen to include it
-			local include = include_illegal or sys:IsCommodityLegal(comm.name)
+			local include = sys:IsCommodityLegal(comm.name)
+			if include_illegal then
+				include = true
+			end
```
Web-eWorks, line 166: Note: this does the same thing in more lines.

data/modules/Debug/DebugCommodityPrices.lua
```diff
@@ -175,24 +160,27 @@ local scan_systems = function(dist)
...
+			local name = comm:GetName()
+			local price = getprice(comm, sys)
 
-				if not commodities[name] then
-					commodities[name] = Commodity:new(name, comm.price)
-					print(name, comm.price)
-				end
+			if include and not commodities[name] then
+				commodities[name] = Commodity:new(name, comm.name, comm.price)
+				print("Adding:", name, comm.name, comm.price)
```
Web-eWorks, line 177: Should this print statement remain? If the debug UI is working I don't think it's necessary to validate the results.

data/modules/Debug/DebugCommodityPrices.lua
```diff
@@ -260,13 +248,11 @@ local function show_system_info(min, sys, str)
...
-	ui.text("") 	-- PolitEcon
 
 	-- Add from C++ side, E.g:
 	-- System type: Mining colony
 	-- Government type: Earth Federation Democrac
 	-- Economy type: capitalist
-	ui.text("")
 end
 
 local function show_details_on_commodity(commodities, clicked)
```
Web-eWorks, line 258: Nitpick: if I understand this code correctly, `clicked` should probably be something like `selected`. Typically we use "clicked" for an action happening to a specific thing (was-clicked) rather than a noun.

data/modules/Debug/DebugCommodityPrices.lua
Web-eWorks: This looks pretty good. Some of this information could be sourced from the commodity market datastore directly (`station:GetCommodityMarket()`) rather than through `GetCommodityStock()` et. al., though the only time that matters is in the case of a custom commodity price being set by a mission module.

Another thing to look at would potentially be to inspect the persistent market cache (`stationMarket` in Economy.lua), though you'd need an accessor to retrive the variable outside of the Economy module.

If you wanted to get fancier, you could provide a button that simulates the effect of restocking ticks on the commodity market - possibly make a copy of the market datastore returned from `GetCommodityMarket()` and call:

```lua
Economy.UpdateStationCommodityMarket(sBody, Engine.rand, marketCopy, comm.id_name, ticksToSimulate)
```

You could then display some sort of graph showing the commodity stock/demand levels over the simulate time - that could be a very useful tool when doing further work on the economy simulation.

data/modules/Debug/DebugCommodityPrices.lua
impaktor: Yes, I re-wrote it like that, (temporarily, I thought) when debugging. I'll revert it.

data/modules/Debug/DebugCommodityPrices.lua
impaktor: I'll remove it. Had some debugging that was driving me crazy.

data/modules/Debug/DebugCommodityPrices.lua
impaktor: thanks for suggestions. As my main driving motivation to work on this was when fixing  #5859, I'll "call it a day" as for features, but keep your suggestions in mind for next time I'll want to dive into commodity balance.

Web-eWorks:
Pushed a minor fix for layout and typographical issues.

PR #5862, opened 2024-07-12
News Event: debug can also remove adverts

impaktor:
I _think_ the new release will crash if player:
- loads a save that has more than 6 months in game
- AND save had a spawned a news event
- AND player docks at that system

So thought to ad a "Delete" button to the news event debug screen, if anyone need to use it (which they will have to if they also happen to have another mission to that system, highly unlikely that it may be):
2024-07-12-110329_837x463_scrot


bszlrd:
[For reference, here's the original report from reddit](https://www.reddit.com/r/pioneerspacesim/comments/1e0jzmp/midsommar_bugfix_release/)

bszlrd:
From reddit: 

MeanSolean: "Alright. I've replaced the old file with the new one. It will delete new events fine but the old crash when trying to open the event in the debug window with the following error:
```
[T] modules/NewsEventsCommodity/NewsEventsCommodity.lua:505: attempt to concatenate field 'multiplier' (a nil value) stack traceback:
    [T] modules/NewsEventsCommodity/NewsEventsCommodity.lua:505 in function 'f'
    [T] pigui/views/debug.lua:39: in function <[T] pigui/views/debug.lua:37>
```
Waiting the old events out seems to work but they do seem do last much longer then the new ones do. I think sending you my save is probably the best way to clear this up so
[SaveFile.json](https://github.com/user-attachments/files/16194088/SaveFile.json)
(not a json)

impaktor:
@bszlrd Fixed it.

PR #5867, opened 2024-07-17
Add "jettison all" buttons to cargo sidebar menu

### LINKED ISSUE #5860: Feature improvments Dumpin menu needs buttons for quikly geting rid of mass on hot entery.
### Observed behavior
When I'm coming in hot to a planet or want to dump all my 50 tons of radioactive waste I cannot do so quickly. 
This could be intended as it takes time to load and dump stuff in reality I guess.

Maybe a Timer could be used to delay the dump to 100ms/ton or something.

I have implemented this here but you will want to cherry pick this as my whole mod will not be ready for some time .. ever ;) 

[Commit msg](https://github.com/syonfox/pioneer/commit/6525896d321de5a9499c2a63605ef30c6e489454)

I need to verify that burn up also triggers the dumping crime as the idea is you dump the pallets and blast them with your main engine effectively turning them into dust.

Anyway ps if you are interested in the syon mod lmk in a bit there will be another commit with my clandestine mission update

impaktor:
Closes #5860 feature request by @syonfox (I was going to close it anyway, as we don't do feature requests on the issue tracker, but figured I might as well implement it, low hanging fruit). 

2024-07-17-165854_1536x960_scrot

Note: just appending "all" to the tooltip hardcoded. Dunno if we keep it, and put it in ui-core, or remove. Not obvious to me why the `tooltips` are passed as arguments to this function?
```lua
transfer_buttons(amount, min, max, tooltip_reduce, tooltip_increase)
```

### DIFF
--- a/data/pigui/modules/sidebar/cargo.lua
+++ b/data/pigui/modules/sidebar/cargo.lua
@@ -53,6 +53,10 @@ local function transfer_button(icon, tooltip, enabled)
 end
 
 local function transfer_buttons(amount, min, max, tooltip_reduce, tooltip_increase)
+	if transfer_button(icons.time_backward_10x, tooltip_reduce .. "##all", amount > min) then
+		amount = min
+	end
+	ui.sameLine(0, 2)
 	if transfer_button(icons.time_backward_1x, tooltip_reduce, amount > min) then
 		amount = amount - 1
 	end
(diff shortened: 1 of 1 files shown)


impaktor:
OK, forced pushed to use same tooltip as other buttons. 
Good to go.

syonfox:
> why the tooltips are passed as arguments to this function?

I am not sure why the original tooltips were passed in but makes sense a bit for re-usability. 

mt purpose of adding the new tooltips was so that people could disable these buttons.

ie 

`slider(0 100, "down", "up")` would be as  it is now 

`slider(0 100, "down", "up", "min", "max")` would show both and you could do 

`slider(0 100, "down", "up", nil, "max")`  this to only show the one button

Web-eWorks:
> Not obvious to me why the tooltips are passed as arguments to this function?

Because it leaves open the possibility for more specific tooltips like "increase scan frequency" or "shrink garbage compactor". Anything that's not passed as an argument is hardcoded to a specific value.
Another side-benefit is the `transfer_buttons` function can be copy-pasted into another file with less work.

PR #5871, opened 2024-07-19
Don't build against lzcnt by default and enable USE_SSE42 only on Intel architecture

Web-eWorks:
This PR addresses some slightly over-zealous feature enablement from the previous release cycle that we are not yet materially using. The build system now only defaults the USE_SSE42 flag to 'on' if we're compiling for a non-ARM64 target, and we no longer enable utilization of the `lzcnt` instruction for SSE4.2 builds - it was introduced in the Haswell microarchitecture alongside AVX2 and as such has been moved to that build option.

I've also added feature-test flags for the build script to use in the future to detect whether we're compiling for an ARM64 target or an Intel target (and whether we're compiling for 32 or 64-bit x86 architecture).

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -133,41 +133,6 @@ jobs:
         token: ${{ secrets.GITHUB_TOKEN }}
         upload_release: true
 
-  build-gcc-avx:
-    runs-on: ubuntu-20.04
-
-    steps:
-    # Checkout the repository as $GITHUB_WORKSPACE
-    - uses: actions/checkout@v4
-
-    - name: Install Dependencies
-      run: |
-        sudo apt-fast update
-        sudo apt-fast install -y ${{ env.packages }}
-
-    - name: Setup CMake
-      run: |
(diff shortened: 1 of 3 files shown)


Web-eWorks:
CC @pcercuei - are there any other changes that should be made to make the Flatpak build easier to maintain?

pcercuei:
Your TargetArchitecture.cmake looks overly complicated, all it needs is
```cmake
if (${CMAKE_SYSTEM_PROCESSOR} MATCHES x86|x64)
       set(PIONEER_TARGET_INTEL ON)
endif()
```

You *could* check for ARM in there too, but you're not using it anywhere, so just introduce it when you need it.

Note that for Flatpaks I'll have to force-disable those two options anyway.

Web-eWorks:
Thanks! I hadn't realized CMake was smart enough for regex matches, that makes this code much easier to express.

Regarding feature flags that need to be disabled - AVX2 is no longer enabled by default in any capacity, and SSE4.2 is default-enabled only on Intel targets for parity between user-compiled checkouts of Pioneer and the official builds we ship.

PR #5879, opened 2024-07-30
Make navigation lights blink

Mc-Pain:
thanks @bszlrd for gifs

Before:

After:

### DIFF
--- a/src/NavLights.cpp
+++ b/src/NavLights.cpp
@@ -44,9 +44,9 @@ static inline vector2f LoadLightColorUVoffset(const std::string &spec)
 	return vector2f(v[0], v[1]);
 }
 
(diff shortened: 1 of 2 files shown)


Web-eWorks (approved):

Web-eWorks (changes requested):
After testing, I've found that at low timewarp values, this works fine. However, once you go past 10x time acceleration, the timeslicing causes severe graphical issues where ship navlights turn on and off in seemingly irregular\* patterns. Ideally, the navlights would either be solidly on or solidly off under high time acceleration, mimicking the perception of such a light during a long-exposure timelapse shot.

\* I'm aware these are regular patterns caused by subsampling the original pattern, however they will not appear to be logical to the average person.

fluffyfreak:
It'd be a shame to close this, I think it's a good change.
Perhaps it needs to be adaptable so it uses different patterns at high time acceleration and keeps this new pattern at regular time accelerations?

Issue #5885, opened 2024-08-04
Weird system where a planet's reference frame includes it's star


lolo101:
### Observed behaviour

I just jumped into system **GJ1057** and my ship is immediately in the reference frame of **GJ1057 d**
Furthermore, planet **GJ1057 d** has a moon **GJ1057 d 3** that orbits at 0.75 AU from its planet, whereas the planet itself is at 0.71 AU from its star.

Also the autopilot behaves strangely

### Expected behaviour

In a system, when the ship is closer to the star than to any planet it should be in the frame of the star. A planet should not have moons that orbit further from the planet than the planet is from its star

### Steps to reproduce

Just load the saved game 
1. The ship is 0.04 AU from the star (0.41 solar masses)
2. The ship is 0.65 AU from the planet (10.29 earth masses)
3. The frame of reference is the one of the planet :shrug: 

__My pioneer version (and OS):__

20240314 (Linux)

__My output.txt (required) and game save (optional, but recommended)__
[output.txt](https://github.com/user-attachments/files/16488866/output.txt)
[lolo1.gz](https://github.com/user-attachments/files/16488865/lolo1.gz)

lolo101:
Actually it looks like the whole system is in the reference frame of **GJ1057 d**

lolo101:
Not sure that's related but when I ask the autopilot to send me to one of the stations on Boer's claim, it sends me right through the planet (and the ship doesn't crash, but the autopilot is unable to get out of the planet)

lolo101:
And it's impossible to land on the station. The ship simple go through the landing pad and into the planet (still without crahing)

bszlrd:
I think this is related to #5830

fluffyfreak:
I think that our system generation has gone a little awry, I added some code to export the system to a json file that can be loaded and edited by the system editor and this is a screenshot from it showing that `GJ 1057 d` has moons in orbits that overlap with the main star of the system!
![image](https://github.com/user-attachments/assets/b79739da-ae3c-49b9-8b2c-436421860b40)

fluffyfreak:
@impaktor @Web-eWorks Does the Frame sysytem work purely off of the distance of the main bodies distance from the centre of the system? It's not very easy to understand, but it seems like the `GJ 1057 d` bodies Frame has a vast radius that encompasses all of it's moons orbits, those moons shouldn't exist as they overlap with the orbits of other planets and indeed the main star.

fluffyfreak:
[...]

Issue #5830, opened 2024-04-26
NPC ships stuck at orbital starport


maxso-216:
### Observed behaviour

Multiple NPC ships stuck at the orbital starport and create huge lag. Also, it is not possible to land in the spaceport.

![Screenshot from 2024-04-26 23-52-29](https://github.com/pioneerspacesim/pioneer/assets/107261566/35bd1822-70d7-4c30-8195-39485e6b0b28)

![Screenshot from 2024-04-26 23-52-45](https://github.com/pioneerspacesim/pioneer/assets/107261566/377c74d5-391b-44ac-b59b-d57387178154)

The collision effect is also offset-ed, but I believe this is another bug that is fixed in newer version.

### Expected behaviour

No NPC ships should be stucked.

### Steps to reproduce

As it is happened randomly(?), I do not know exact steps to reproduce.

__My pioneer version (and OS):__

20240203 (via flatpak), Ubuntu 23.10

__My output.txt (required) and game save (optional, but recommended)__

[output.txt](https://github.com/pioneerspacesim/pioneer/files/15132782/output.txt)

save file : https://shota.nu/ss/ybf8tjpd.gz

impaktor:
I believe this was fixed in the 3.14 release, by multiple NPS ship fixes:
https://github.com/pioneerspacesim/pioneer/releases/tag/20240314

@Gliese852 I assume we can close this?

Gliese852:
@impaktor I'm not sure, I don't think I've encountered such a problem, need to check.

impaktor:
@maxso-216 Please use the latest release (3.14), to load your old saves, you might have to hold down Ctrl to force load them, the first time.

maxso-216:
@impaktor I tried with latest release and recover the save file. Seems like the situation still exists. 

![Screenshot from 2024-04-27 20-44-03](https://github.com/pioneerspacesim/pioneer/assets/107261566/7ac554d2-dc24-4822-a2c6-f11dbbf040a6)

Haven't tried going to other systems and see if same thing happen again, maybe report later.

maxso-216:
@impaktor I build from source code and play for 2 hours, and same thing happen again. Its very lag that I can't fly close to it.

![Screenshot from 2024-04-28 00-30-16](https://github.com/pioneerspacesim/pioneer/assets/107261566/5a10f7ca-e0ef-40ae-8be5-e98ba3f6a35d)

save file:https://shota.nu/r5akwxcl.gz


[...]

PR #5888, opened 2024-08-10
Added Perf-Info button to Dump a system to systemeditor compatible JSON

fluffyfreak:
Whilst debugging #5885 and #5830 today I wanted a quick way of getting a dumped system into the system editor so I've added a 2nd button to the Perf-Info screen which already allowed dumping the JSON to the output log.

Instead it now generates a JSON file which can be directly loaded into the system editor for viewing / modification and outputs it into the users pioneer directory where screenshots, savefiles, etc folders are located.

This should also make it easier for people to create custom systems based on generated ones they've found interesting to make that process a little easier.

### DIFF
--- a/src/pigui/PerfInfo.cpp
+++ b/src/pigui/PerfInfo.cpp
@@ -413,9 +413,7 @@ void PerfInfo::DrawWorldViewStats()
 			RefCountedPtr<StarSystem> system = Pi::game->GetGalaxy()->GetStarSystem(path);
 
 			if (system) {
-				char fileName[1024];
-				sprintf(fileName, "%s.json", system->GetName().c_str());
-				const std::string fname = FileSystem::JoinPathBelow(FileSystem::userFiles.GetRoot(), fileName);
+				const std::string fname = FileSystem::JoinPathBelow(FileSystem::userFiles.GetRoot(), fmt::format("{}.json", system->GetName()));
 				FILE *f = fopen(fname.c_str(), "w");
 
 				if (f) {
--- a/src/pigui/PerfInfo.cpp
+++ b/src/pigui/PerfInfo.cpp
@@ -17,6 +17,8 @@
 #include "lua/Lua.h"
 #include "lua/LuaManager.h"
 #include "scenegraph/Model.h"
+#include "JsonUtils.h"
+#include "FileSystem.h"
 
 #include <fmt/core.h>
 #include <imgui/imgui.h>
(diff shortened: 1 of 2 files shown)


fluffyfreak:
I apologise for the terrible grammar / phrasing above, it's the weekend, clearly my brains know that! 😆

impaktor:
Welcome back!

I assume it then deprecates the code snippet I have on my local copy for adding system export (to lua!) from debug screen (adds a "system" tab with an export button)
```lua
-- Copyright © 2008-2024 Pioneer Developers. See AUTHORS.txt for details

local Game = require 'Game'
local ui = require 'pigui'
local debugView = require 'pigui.views.debug'

local Lang = require 'Lang'
local l = Lang.GetResource("ui-core")


debugView.registerTab("System-debug", function()
	if Game.player == nil then return end
	if not ui.beginTabItem("System") then return end

	if ui.button("Export system", Vector2(100, 0)) then
		Game.system:ExportToLua()
	end
end)

```

fluffyfreak:
Different results but the same idea, one is to Lua the other to JSON 🤷

Web-eWorks:
I completely forgot I originally added the button to the perf info window, or I'd have updated it myself - thanks, this will be useful.

The original plan was for the system editor to have an embedded sector map and be able to visually pick systems to edit, but this way means you don't have to load the editor and remember the system you wanted to edit.

I'll take a look at the `comment` field - it "should" be entirely optional and if it's not that's my bad.

src/pigui/PerfInfo.cpp
```diff
@@ -405,6 +407,32 @@ void PerfInfo::DrawWorldViewStats()
...
 				system->Dump(Log::GetLog()->GetLogFileHandle());
 		}
+
+		if (ImGui::Button("Dump System-Editor JSON")) {
+			SystemPath path = Pi::game->GetSectorView()->GetSelected();
+			RefCountedPtr<StarSystem> system = Pi::game->GetGalaxy()->GetStarSystem(path);
+
+			if (system) {
+				char fileName[1024];
+				sprintf(fileName, "%s.json", system->GetName().c_str());
```
Web-eWorks, line 417: I'd recommend using `fmt::format("{}.json", system->GetName())` or `system->GetName() + ".json"` - this avoids stack-allocating a 1kb buffer to store a grand total of ~32 bytes of string data. It *also* avoids the improbable but possible chance of overflowing the buffer (given that you're not using `snprintf`). The heap allocation is completely negligible in this case, especially given that it's being further processed through `FileSystem::JoinPathBelow`.

fluffyfreak: Yeah you're right, that was me being extremely lazy :D

Issue #5886, opened 2024-08-06
After reloading a previous save, Bulletin Board is empty


lolo101:
### Observed behaviour

After landing at a station, I reloaded a save where I was in another system. Then I fly to the same station again and this time the Bulletin Board is empty
![image](https://github.com/user-attachments/assets/f80209a7-2850-436f-b7d7-ff18f3fc4e25)

### Expected behaviour

The Bulletin Board should show some missions, ads and so on

### Steps to reproduce

Load the attached savegame and open the station's BB

__My pioneer version (and OS):__

20240710 (e123a3f91) (Linux)

__My output.txt (required) and game save (optional, but recommended)__

[output.txt](https://github.com/user-attachments/files/16511417/output.txt)
[lolo1.gz](https://github.com/user-attachments/files/16511419/lolo1.gz)

impaktor:
Strange / new one. 
I'd be surprised if it has anything to do with an issue as old as 899, since populating the BBS adverts was re-written just 1-2 years ago, and the UI has been completely re-written since then as well - twice I think, since that must be OldUI (-> NewUI -> PiGui).

zonkmachine:
I've seen this before and I assumed it was related to https://github.com/pioneerspacesim/pioneer/issues/5305 which I was testing at the time . When I load the save in `lolo1.gz` and let it run for a while, the ship market is updated as it should but there are no new ads in the BBS ever so it's properly crashed.

Since the ship market is updated and `updateShipsOnSale(station)` and `updateAdverts(station)` are called at the same time we know that call is made.

I checked the Lua console and this logwarning is firing:
**"SpaceStation.lua: updateAdverts called for station that hasn't been visited"**

When I comment out all lines except `Event.Queue("onUpdateBB", station) ` in updateAdverts(), the BBS starts to repopulate itself.
https://github.com/pioneerspacesim/pioneer/blob/e123a3f91518ff4d0c84f90270803bf87dd2848c/data/libs/SpaceStation.lua#L901-L907

zonkmachine:
Maybe `updateAdverts()` should look something like:

```lua
local function updateAdverts (station)
	if not SpaceStation.adverts[station] then
		logWarning("SpaceStation.lua: updateAdverts called for station that hasn't been visited")
	end
	Event.Queue("onUpdateBB", station)
end
```

Web-eWorks:
[...]

PR #5894, opened 2024-08-20
Repopulate BBS if empty

zonkmachine:
There is an infrequent issue where the BBS will show up empty on a game reload as SpaceStation.adverts[station] is missing. This fix will cause the BBS to repopulate itself but only from those modules that impements onUpdateBB(). It will not solve the underlying cause.

Addresses #5886

### DIFF
--- a/data/libs/SpaceStation.lua
+++ b/data/libs/SpaceStation.lua
@@ -901,9 +901,8 @@ end
 local function updateAdverts (station)
 	if not SpaceStation.adverts[station] then
 		logWarning("SpaceStation.lua: updateAdverts called for station that hasn't been visited")
-	else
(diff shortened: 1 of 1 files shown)


impaktor:
In a way, this would mask the bug, possibly making it harder to find the root cause.

zonkmachine:
> In a way, this would mask the bug, possibly making it harder to find the root cause.

Maybe, but the station repairs itself this way. I think it's worth a try. There should probably be some tests in the code to track what's going on. Does the save or the reload fail?

Web-eWorks (approved):
This is an acceptable stopgap solution for now. I'd rather it be merged now and a more holistic look taken at the whole problem another time.

Issue #5659, opened 2023-11-03
Ship spinner always starts on same ship


impaktor:
On current master 579d5c6 every time you start the game it's the xylophis that is the first ship in the ship spinner in Main Menu. It used to be random, I believe, and which is what I expected.

Max5377:
Ship is random everytime when starting the game on current master a2170e982f1b830c6dbb57c43a8f483c8839183b

impaktor:
I just compiled a2170e982f1b830c6dbb57c43a8f483c8839183b and I'm still getting Xylophis 5 times out of 5. 

Must be something with my setup then? Mods? (I have bulkship-models (Vgr-battlecruiser & longnose) in my mod folder, together with Evarchart's facegen assets).

Max5377:
Can you give me a link to these? I can't seem to find them anywhere.
**Fun fact**: Pioneer has `moddb` page which has a very old [alpha 7](https://www.moddb.com/games/pioneer/downloads/pioneer-alpha-7) version.

impaktor:
> Fun fact: Pioneer has moddb page 

Yes, Geraldine used to upload every new release to there, but that died off, about the same time that SSC went dormant.

I'm attaching the mods to this post. Background on the mods:

- [Longnose](https://forum.pioneerspacesim.net/viewtopic.php?f=6&t=340) is lacking textures, 
- Vgr-battlecruiser has wrong scale, so I think it's 1 cm big or something (anyone feeling up to play with it in Blender?), thus often there's no bulkship outside the station, because it's too tiny. (Bulkship.lua has some probability for spawning bulkships)
- [Evarchart](https://forum.pioneerspacesim.net/viewtopic.php?f=8&t=382) (aka CaptainNow on [EAB](https://eab.abime.net/)) did the Facegen, but he's been [through a bit](https://eab.abime.net/showpost.php?p=1662161&postcount=801), and now the Amiga game he's done all art & assets for is nearing a physical release.

[facegen.zip](https://github.com/pioneerspacesim/pioneer/files/13987650/facegen.zip)
[longnose0.4.zip](https://github.com/pioneerspacesim/pioneer/files/13987651/longnose0.4.zip)
[Vgr_battlecruiser.zip](https://github.com/pioneerspacesim/pioneer/files/13987653/Vgr_battlecruiser.zip)

Max5377:
Good luck to Evarchart with the launch of his game.
[...]

PR #5896, opened 2024-08-21
Fix intro screen ship randomness

zonkmachine:
Fix PiRngWrapper to use the full range of the random object. std::shuffle expects the range provided by max().
See explanation [here](https://github.com/pioneerspacesim/pioneer/issues/5659#issuecomment-2300720284).

Fixes https://github.com/pioneerspacesim/pioneer/issues/5659

### DIFF
--- a/src/Intro.cpp
+++ b/src/Intro.cpp
@@ -16,18 +16,13 @@
 
 class PiRngWrapper {
 public:
-	PiRngWrapper(size_t maxValue) :
-		maxVal(maxValue) {}
 	typedef unsigned int result_type;
 	static constexpr unsigned int min() { return 0; }
 	static constexpr unsigned int max() { return std::numeric_limits<uint32_t>::max(); }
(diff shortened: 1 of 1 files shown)


impaktor:
Makes me nervous to see the fix is to _remove_ the `maxValue` variable, but if it works, `¯\_(ツ)_/¯`. I'll give this a try in a bit.

impaktor:
Yep, I confirm, this fixes the issue.

Web-eWorks (approved):

Web-eWorks:
`PiRngWrapper` in this file provides what the standard calls a *UniformRandomBitGenerator* - i.e. `std::shuffle` just wants a random integer in the inclusive range `[ PiRngWrapper::min(), PiRngWrapper::max() ]` and will take that integer and generate a random index into the shuffled range itself. This is the computationally simplest method of addressing the bug and removes a superfluous modulo operation.

PR #5899, opened 2024-08-23
Improve PerfInfo debug window

Web-eWorks:
I've converted the PerfInfo debug tabs to use icons rather than names, made the debug tabs themselves hot-reloadable, and improved the amount of data shown in the performance plots displayed in the performance section. Frame graphs now display the last 5 seconds of frame timings (up from 1s), and have both a recent "instantaneous average" and whole-graph average. Memory graphs (which have a sample rate of 1Hz instead of ~60Hz) now show the last five minutes of memory usage for a more complete picture.

The PerfInfo window is easier to read and use at smaller window sizes, and can be collapsed completely to only display the current FPS while flying around the world as a built-in FPS counter.

I've also made the counters slightly cheaper to update in most cases, which should entirely offset any performance loss from increasing the number of samples per counter by 5x.

image

### DIFF
--- a/data/modules/Debug/DebugCommodityPrices.lua
+++ b/data/modules/Debug/DebugCommodityPrices.lua
@@ -378,10 +378,9 @@ local function main()
 end
 
 
-debugView.registerTab("Commodities", function()
-  if Game.player == nil then return end
-  if ui.beginTabItem("Commodities") then
-	main()
-	ui.endTabItem()
-  end
-end)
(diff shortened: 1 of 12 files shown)


Web-eWorks:
This work was mostly done late last year - I had originally intended to add some additional functionality in the meantime and shelved the branch. Time did it's usual thing in the interim and I've completely forgotten what I intended to add further, so it's getting PR'd now.

PR #5901, opened 2024-08-23
Add ability to reroll random systems to System Editor

Web-eWorks:
Continuing on from the discussion in #5892, this PR cleans up some generation bugs in binary systems which caused the orbits of stars to change once edited (due to orbital parameters not being written back to the body), and adds a new "regenerate system from seed" functionality to quickly generate new variants of the edited system. It also addresses a unit conversion issue in body radius display when using Earth Radius as a unit.

Due to data plumbing issues, regenerating a system can only be done for a full-procedurally generated system or a "partially random" system sourced from the `local_stars` database of systems with randomly generated contents.

https://github.com/user-attachments/assets/cb859248-1401-43f9-8bb7-56ec3c35cb0b

### DIFF
--- a/src/editor/EditorWindow.cpp
+++ b/src/editor/EditorWindow.cpp
@@ -11,7 +11,8 @@
 using namespace Editor;
 
 EditorWindow::EditorWindow(EditorApp *app) :
-	m_app(app)
+	m_app(app),
+	m_canBeClosed(true)
 {}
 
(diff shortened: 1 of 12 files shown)


Web-eWorks:
I've tested this code to ensure it doesn't change the consistency of system layouts in bugged binary systems, so it should be compatible with old saves - not that it matters, since we're already savebumping.

PR #5910, opened 2024-08-30
Add Hawaiian names

zonkmachine:
Aloha!


data/culture/haw.lua
```diff
@@ -0,0 +1,274 @@
...
+	'Kimo',
+	'Kimokea',
+	'Koa',
+	'Koalani',
+	'Konane',
+	'Kulani',
+	'Ky',
+	'Lihau',
+	'Lono',
+	'makai',
```
Web-eWorks, line 68: Wrong case for a Proper Noun.

data/culture/haw.lua
```diff
@@ -0,0 +1,274 @@
...
+	'Kulani',
+	'Kuʻuipo',
+	'Ky',
+	'Lana',
+	'Lei',
+	'Leinani',
+	'Lihau',
+	'Lilo', -- ;)
+	'Lono',
+	'luana',
```
Web-eWorks, line 139: Wrong case.

data/culture/haw.lua
```diff
@@ -0,0 +1,274 @@
...
+	'Keliʻi',
+	'Keona',
+	'Konani',
+	'Lani',
+	'Lee',
+	'Leilani',
+	'Leimomi',
+	'Leo',
+	'Likeke',
+	'Lilo',
```
Web-eWorks, line 230: Note that this could create a name e.g. 'Lilo Lilo'. Is this well formed?

impaktor:
I assume you've tested it in game by setting weight to something very high so everyone will become Hawaiian?

data/culture/haw.lua
zonkmachine: Yes. I believe this is how it's done.

zonkmachine:
> I assume you've tested it in game by setting weight to something very high so everyone will become Hawaiian?

I sure have. A lot of testing to make sure the special signs are not shown.

zonkmachine:
Many Hawaiian names are compound names and it's tempting to make a generator for it but I pass for now. It looks simple enough but it's for another day.

It's tempting to add a 'phrases' section to the language files. So the character could go randomly Aloha! in game.

Thanks for the review! I think this is ready to merge.

PR #5912, opened 2024-08-30
Create BBS if empty

zonkmachine:
This is a fixup from #5894 / a005ad4
Instead of updating the BBS it should be created if SpaceStation.adverts[station] is missing.

Addresses #5894
Addresses #5886


zonkmachine:
I have tested that new and reloaded projects works.

Web-eWorks (approved):
Approved. My worktree is full, so I'll consider this sufficiently tested-by-author for a merge (and point any resulting bugs your way :stuck_out_tongue:).

PR #5915, opened 2024-09-07
Simplified random station selection for delivery missions

robothauler:
The already existing function GetNearbyStationPaths() in MissionUtils is used to find the stations in nearby systems.
This also increases the probability for systems with more stations. Not too bad IMHO.


Web-eWorks (approved):
I'm not in a position to test this at the moment, but the code looks good - I really appreciate when contributors take the time to convert old callsites to use newer APIs that get added over time. Thanks a lot!

@impaktor feel free to merge this once you think it's received sufficient testing (or live dangerously and merge it immediately!)

PR #5916, opened 2024-09-07
ECM crime - Don't use your ECM near a station

zonkmachine:
~~_I hesitated a bit to submit this because of the overly harsh reaction if you happen to press the ecm button while taking off, but this is the same for all crime, dumping waste or firing a weapon. You will be terminated._~~

Updated: The effective range of the ban is now the same as the effective range of the ECM, 4 km.

I think there could be some more around this like a couple of the lights going out on the platform or the BBS being out of order the next time you land there. ECM should be used responsibly.

ecmdischarge

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -727,15 +727,20 @@ Ship::ECMResult Ship::UseECM()
 		Space::BodyNearList nearby = Pi::game->GetSpace()->GetBodiesMaybeNear(this, ECM_RADIUS);
 		for (Body *body : nearby) {
 			if (body->GetFrame() != GetFrame()) continue;
-			if (!body->IsType(ObjectType::MISSILE)) continue;
 
 			double dist = (body->GetPosition() - GetPosition()).Length();
-			if (dist < ECM_RADIUS) {
-				// increasing chance of destroying it with proximity
-				if (Pi::rng.Double() > (dist / ECM_RADIUS)) {
-					static_cast<Missile *>(body)->ECMAttack(ecm_power_cap);
+			if (body->IsType(ObjectType::MISSILE)) {
+				if (dist < ECM_RADIUS) {
+					// increasing chance of destroying it with proximity
+					if (Pi::rng.Double() > (dist / ECM_RADIUS)) {
+						static_cast<Missile *>(body)->ECMAttack(ecm_power_cap);
+					}
 				}
 			}
+			else if (body->IsType(ObjectType::SPACESTATION) && dist < ECM_RADIUS) {
+				LuaEvent::Queue("unlawfulDischargeECM", this);
+			}
+			else { continue; }
 		}
 		return ECM_ACTIVATED;
 	} else
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -2583,6 +2583,10 @@
     "description": "",
     "message": "Unlawful weapons discharge"
   },
+  "UNLAWFUL_ECM_DISCHARGE": {
+    "description": "",
+    "message": "Unlawful ECM discharge"
+  },
   "UNOCCUPIED_PASSENGER_CABINS": {
     "description": "",
     "message": "Unoccupied Passenger Cabins"
--- a/data/libs/Legal.lua
+++ b/data/libs/Legal.lua
@@ -21,6 +21,7 @@ local l = Lang.GetResource("ui-core")
 -- PIRACY - fired on ship
 -- TRADING_ILLEGAL_GOODS - attempted to sell illegal goods
 -- WEAPONS_DISCHARGE - weapons discharged too close to station
+-- ECM_DISCHARGE - ECM discharged too close to station
 --
 -- Availability:
 --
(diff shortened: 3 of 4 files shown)


impaktor:
> I hesitated a bit to submit this because of the overly harsh reaction if you happen to press the ecm button while taking off,

Always wise to have some distance in situations of discharge....

I do believe the new equipment system aims to add a arm/disarm for weapons, but I could be wrong. Would be good to have for ECM as well.

src/Ship.cpp
```diff
@@ -800,15 +800,20 @@ Ship::ECMResult Ship::UseECM()
...
+			if (body->IsType(ObjectType::MISSILE)) {
+				double dist = (body->GetPosition() - GetPosition()).Length();
+				if (dist < ECM_RADIUS) {
+					// increasing chance of destroying it with proximity
+					if (Pi::rng.Double() > (dist / ECM_RADIUS)) {
+						static_cast<Missile *>(body)->ECMAttack(ecm_power_cap);
+					}
 				}
 			}
+			else if (body->IsType(ObjectType::SPACESTATION)) {
```
Web-eWorks, line 813: I'm a little hesitant about this check here and the associated event naming. This seems a bit brittle, and the `onUseECM` name doesn't really match what's happening here.

Web-eWorks (changes requested):
Given that you're hesitant about submitting it, I'd suggest that this PR should probably sit and receive a bit more thought before it's merged. The actual implementation of the ECM event is a bit misleading from the perspective of the LuaAPI - at minimum, I'd like to see the name of the event changed to more clearly indicate which situations it triggers in before I can sign off on the PR.

The ECM equipment / game system itself will also hopefully be receiving some design thought as we move towards Stage 2 / Stage 3 of the EquipSet rework. Feel free to contribute in that respect as well - the implementation of this PR may need to adjust to take into account some of the interactions of the new equipment system as we iteratively implement it.

fluffyfreak:
I don't mind the naming of `onUseECM` that seems to match the rest of our scheme 🤷 
Don't we have a crime system where we issue a fine and DON'T try to kill the player? 🤔

sturnclaw:
My specific problem was that the `onUseECM` event is only ever triggered if the ECM system has been clicked within a specific radius of a space station, which is absolutely not the behavior you'd expect from the event name.

In general, the ECM system really needs a re-think and I'd prefer to avoid hardcoding more behaviors into it until we've had a chance to do a design pass over it.

zonkmachine:
Thanks for the feedback!

> DON'T try to kill the player? 🤔

Memo: don't try to kill the player...

I had a vague idea that you could have things break down on the station if it's close enough when you fire the ECM. Such as some of the light's on the pad going out or the BBS being blank after an outage with a message of failure.

bszlrd:
Heh, handful of reputation decrease could be apt as well.

zonkmachine:
Rebased and gave it a quick test. I also tested it rebased on #6128 and it works intended with no overaggressive police.

zonkmachine:
> I do believe the new equipment system aims to add a arm/disarm for weapons, but I could be wrong. Would be good to have for ECM as well.

One for both i assume.

> Don't we have a crime system where we issue a fine and DON'T try to kill the player? 🤔

It's not really fine tuned. The level over which the police will 'open fire until dead' is a fine of 300$. The lowest base fine of any crime is 500$. It's addressed in https://github.com/pioneerspacesim/pioneer/pull/6128

> Heh, handful of reputation decrease could be apt as well.

That would be appropriate, yes. And a future BBS news flash module will bring it up.

> My specific problem was that the `onUseECM` event is only ever triggered if the ECM system has been clicked within a specific radius of a space station, which is absolutely not the behavior you'd expect from the event name.

OK. Looking into it.

zonkmachine:
> My specific problem was that the `onUseECM` event is only ever triggered if the ECM system has been clicked within a specific radius of a space station, which is absolutely not the behavior you'd expect from the event name.

Would **unlawfulUseECM** work?

zonkmachine:
* Rebased on latest master
* onUseECM changed to unlawfulDischargeECM
* Radius of ban effect same as effective radius (4 km).

zonkmachine:
> Given that you're hesitant about submitting it,

The issue I had with this PR is the sudden execution by the police if you use it in proximity of a space station. This is fixed in https://github.com/pioneerspacesim/pioneer/pull/6128

sturnclaw:
This is acceptable. I'd prefer to eventually move the responsibility for generating the `unlawfulDischargeECM` event into Lua, as the current implementation hardcodes part of the legal system in C++. That's not an immediate concern right now however, and can be tackled in the future when countermeasures gameplay is looked at in more depth.

zonkmachine:
> The fine being $500 is fine by me, though maybe we make it $750 since it's more difficult to "accidentally" click the ECM button?

I changed it to $750 to get some variation.

> A good "fudge" number for enforcement range would be around 15km from the city center, as that prevents any of the buildings for currently-generated cities from being within the effect radius.

I thought selecting the legal range would be trivial. It only is if it's within the ECM_RADIUS. 
We explicitly test for ECM_RADIUS here:

https://github.com/pioneerspacesim/pioneer/blob/9f0f20639afe91bc5c415713665d5cc8acc44205/src/Ship.cpp#L727 

If it's larger I think I would have to create an outer test for the legal range and that would kind of look bad. Maybe wait with this until you move it to the LUA side?

sturnclaw:
> If it's larger I think I would have to create an outer test for the legal range and that would kind of look bad. Maybe wait with this until you move it to the LUA side?

That's fine - this illustrates my point about hardcoding the legal system within C++ though... :wink:

zonkmachine:
> That's fine - this illustrates my point about hardcoding the legal system within C++ though... 😉

Right, got it!

I think this PR is ready.

Issue #5917, opened 2024-09-09
"savefiles" is not created automatically


Bodasey:
### Observed behaviour

The directory "savefiles" is not created automatically
The load/save game routine crashes the game

### Steps to reproduce

rename the directory "savefiles"

__My pioneer version (and OS):__

actual master on SuSE Tumbleweed

__My output.txt (required) and game save (optional, but recommended)__

When saving:
`Info: Error picking auto-save
Info: Created shader gassphere_base (address=0x29d82b30)
Info: Created shader geosphere_sky (address=0x26ad5100)
Info: Created shader planetrings (address=0x2867af10)
Info: Created shader geosphere_terrain (address=0x29e79df0)
Info: Error picking auto-save
Info: Error picking auto-save
Info: Error picking auto-save
Info: Error picking auto-save
Info: Error picking auto-save
Info: Error picking auto-save
Info: Error picking auto-save
Info: Error picking auto-save
Info: Error picking auto-save
Warning: Caught error in Lua UI code:
        [T] pigui/modules/saveloadgame.lua:281: bad argument #1 to 'sort' (table expected, got string)
libpng warning: iCCP: known incorrect sRGB profile
Info: WARNING: texture 'icons/object_star_m.png' is not power-of-two and may not display correctly
Info: WARNING: texture 'icons/object_planet_ice.png' is not power-of-two and may not display correctly
Info: WARNING: texture 'icons/object_orbital_starport.png' is not power-of-two and may not display correctly
Info: StarSystemCache: misses: 595, slave hits: 0, master hits: 235799
Info: SectorCache: misses: 100, slave hits: 5057, master hits: 161920
Info: Pi shutting down.
Info: StarSystemCache: misses: 0, slave hits: 0, master hits: 0
Info: SectorCache: misses: 0, slave hits: 0, master hits: 0
`

when loading
`Fatal: [T] pigui/modules/saveloadgame.lua:281: bad argument #1 to 'sort' (table expected, got string)
stack traceback:
        [C]: in function 'sort'
        [T] pigui/modules/saveloadgame.lua:281: in function 'makeFileList'
        [T] pigui/modules/saveloadgame.lua:355: in function '?'
        [T] pigui/libs/module.lua:43: in function 'update'
        [T] pigui/modules/saveloadgame.lua:336: in function 'update'
        [T] pigui/libs/modal-win.lua:72: in function 'drawModals'
        [T] pigui/libs/modal-win.lua:96: in function 'draw'
[...]

PR #5919, opened 2024-09-23
fix(savegame): fix savefiles directory not being created (#5917)

mwerle:
Fixes https://github.com/pioneerspacesim/pioneer/issues/5917

A refactor of the load/save UI in commit
7b6a79700682124d29351fb83562778eaf5a2da4 introduced a bug preventing the "savefiles" directory from getting created, and causing an error-popup whenever the directory was attempted to be accessed by the LUA UI.

The bug was resetting a variable to an unexpected value which then caused the "sort" algorithm to throw an error. This was resolved by ensuring that the variable always had a valid value for the sort algorithm.

### DIFF
--- a/data/pigui/modules/saveloadgame.lua
+++ b/data/pigui/modules/saveloadgame.lua
@@ -274,9 +274,9 @@ function SaveLoadWindow:makeFileList()
 	if not ok then
 		Notification.add(Notification.Type.Error, lui.COULD_NOT_LOAD_SAVE_FILES, files --[[@as string]])
 		self.files = {}
+	else
+		self.files = files
 	end
-
-	self.files = files
 
 	table.sort(self.files, function(a, b)
 		return a.mtime.timestamp > b.mtime.timestamp


impaktor:
> Issue https://github.com/pioneerspacesim/pioneer/issues/5917

Please use "fixes" or "closes", as it will trigger auto-close of the issue once this is merged.

mwerle:
> Please use "fixes" or "closes", as it will trigger auto-close of the issue once this is merged.

Fixed.

Bodasey:
fix confirmed

There still appears a pop-up error message "user://savefiles is not a directory" - without reason or function, i think.

mwerle:
> There still appears a pop-up error message "user://savefiles is not a directory" - without reason or function, i think.

Yes, that's because the current logic only creates the directory when saving, not when loading. This could be changed. In fact, I'm not sure why the directory is not simply created on program startup just like the "pioneer" directory itself and the "mods" subdirectory and instead delegated to the save routines.

Web-eWorks:
Good catch, thank you! Feel free to submit a follow-up PR creating the `savefiles` directory at program startup (i.e. in `Pi::App::OnStartup()` or as a task inside `StartupScreen:Start()` in `Pi.cpp`).

mwerle:
> Good catch, thank you! Feel free to submit a follow-up PR creating the `savefiles` directory at program startup (i.e. in `Pi::App::OnStartup()` or as a task inside `StartupScreen:Start()` in `Pi.cpp`).

Nitpick:  we should have done this first. Now the git history will contain several fixes for the same issue.  (I'm currently working in a team where one of the leads is rather anal about maintaining a "clean" git history..)

But thank you for your assistance with my first PR :)

Web-eWorks:
Speaking as something of a nitpick-y fellow about git history myself - I consider it no harm done to handle this in two separate PRs. :smile:

This PR fixes a program error in the UI arising from invalid handling of a correct-but-untested state, while the latter PR changes the game's behavior in a first-startup condition rather than resolving a concrete program error or assertion.

impaktor:
> Nitpick: we should have done this first. Now the git history will contain several fixes for the same issue

Brings a tear to my eye

> (I'm currently working in a team where one of the leads is rather anal about maintaining a "clean" git history..)

What? This is the only project I'm involved in, I swear!

PR #5922, opened 2024-09-27
Feat/reset view on hyperspace exit

### LINKED ISSUE #5691: System Map is not updated after hyperjump
1) Plot a route
2) Jump to a next system
3) Open system overview

Observed: either departure system or target system is "shown" (Unexplored system. System view unavailable)

Expected: current system is shown

mwerle:
Fixes #5691 

## Background
The current game has a concept of a "selected system" in the `Sector Map`. Whenever opening the `System Overview` or the `System Map`, this is the system whose details will be displayed.

The only way to change the "selected system" is by the user changing it in the `Sector Map`.

This is somewhat unintuitive from a players' perspective. When the player is interacting with the various maps, then yes, this is the expected behaviour, however, when the player is jumping into a new system, upon arrival, it is more intuitive for the `System Map` and `System Overview` to show the current system.

## Proposed changes
This PR proposes to fix this with 2 partially-related changes:

1. Add an option to reset the view to the WorldView following a hyperspace jump. 
2. Add a selection button to the Map Views allowing the player to choose to display either the current system, or the selected system.

When the option to reset the view to the WorldView is unchecked (default), and the Map Display is set to the selected system (default), there is no change to the game behaviour from the current behaviour.

### Reset View after hyperspace exit

This was originally added to avoid certain issues with the selected system changing, but is kept as some players may prefer this . **This could be split into a separate PR.**

A benefit of enabling this option is that players can immediately react to hostile actions following a hyperspace exit.

An improvement to this option may be to also reset the view to "internal" and "front" - the current implementation maintains the last camera setting.

### Add Map View mode to selected or current

When the Map View is set to the current system, the System Map and the System Overview will show the current system instead of the selected system. This can be especially useful if the player is exploring and wishes to quickly examine the current system while following a hyperspace route of multiple jumps.

One side-effect of this change is that if the player is currently viewing one of the maps while performing a hyperspace jump, then the view is updated as soon as the new system is reached. 

On the one hand this could be seen as a benefit as the player is immediately notified of arrival in a new system, on the other it could be disruptive if the player is currently examining some aspect of the previous system.

It would require additional work, potentially complex and fragile, to keep the view on the currently opened system.

## Issues with these changes

Apart from those already mentioned, there are no known issues or new bugs introduced by these changes.

## Original Draft-PR Description
<details>

## Background
The current game has a concept of a "selected system" in the `Sector Map`. Whenever opening the `System Overview` or the `System Map`, this is the system whose details will be displayed.

The only way to change the "selected system" is by the user changing it in the `Sector Map`.

This is somewhat unintuitive from a players' perspective. When the player is interacting with the various maps, then yes, this is the expected behaviour, however, when the player is jumping into a new system, upon arrival, it is more intuitive for the `System Map` and `System Overview` to show the current system.

## Proposed changes
This PR proposes to fix this by resetting the "selected system" to the "current system" when entering a new system.

## Issues with the fix
The current implementation has the downside that if the player is viewing the `System Map` or `System Overview` _while_ performing a hyperspace jump, the map is changed as soon as the jump is complete, even if the player is interacting with the map at the time.

## Discussion
Hence this PR is still a "draft" while ways to prevent the map from changing are investigated. While some players might like the current approach (ie, it's obvious when the jump is complete), others might prefer to keep viewing the old map.

One solution could be that the player is forcibly returned to the flight screen when the hyperspace jump is completed. This has the benefit that the player is immediately able to react to, for example, hostile actions by other ships.
This should be fairly easy to implement.

Another suggestion has been that the "selected system" is _not_ changed, but rather that the `System Map` and `System Overview` show the "selected system" when opened from the `Sector Map` (and possibly some other screens such as the BBS), but the "current system" when opened from the spaceflight screen. This might lead to confusing behaviour depending on the exact sequence of opening maps and the starting screen compared to what the player expects. This is also likely to be somewhat more complex to implement.

</details>

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -743,6 +743,10 @@
     "description": "",
     "message": "Full screen"
   },
+  "GAME_OPTIONS": {
+    "description": "GameOptions - tooltip for the Game Options tab in the Settings Window",
+    "message": "Game Options"
+  },
   "GAME_TIME": {
     "description": "",
     "message": "Game time"
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -967,6 +967,10 @@
     "description": "Tooltip: Sidereal camera view",
     "message": "Sidereal view"
   },
+  "HUD_BUTTON_SWITCH_TO_CURRENT_SYSTEM": {
+    "description": "Tooltip: Switch to current system",
+    "message": "Switch to current system"
+  },
   "HUD_BUTTON_SWITCH_TO_GALAXY_MAP": {
     "description": "Tooltip: Switch to galaxy map",
     "message": "Switch to galaxy map"
@@ -974,6 +978,10 @@
   "HUD_BUTTON_SWITCH_TO_SECTOR_MAP": {
     "description": "Tooltip: Switch to sector map",
     "message": "Switch to sector map"
+  },
+  "HUD_BUTTON_SWITCH_TO_SELECTED_SYSTEM": {
+    "description": "Tooltip: Switch to selected system",
+    "message": "Switch to selected system"
   },
   "HUD_BUTTON_SWITCH_TO_SYSTEM_MAP": {
     "description": "Tooltip: Switch to system map",
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -2031,6 +2031,14 @@
     "description": "Paintshop button",
     "message": "Reset Preview"
   },
+  "RESET_VIEW_ON_HYPERSPACE_EXIT": {
+    "description": "Reset View on Hyperspace Exit option",
+    "message": "Reset View on Hyperspace Exit"
+  },
+  "RESET_VIEW_ON_HYPERSPACE_EXIT_DESC": {
+    "description": "The tooltip of Reset View on Hyperspace Exit in the settings menu",
+    "message": "Reset view to main game view when exiting from hyperspace"
+  },
   "RETURN_TO_GAME": {
     "description": "",
     "message": "Return to game"
--- a/src/SystemView.h
+++ b/src/SystemView.h
@@ -168,6 +168,11 @@ public:
 		Atlas = 1
 	};
 
+	enum class SystemSelectionMode { // <enum name=SystemSelectionMode scope='SystemView::SystemSelectionMode' public>
+		CURRENT_SYSTEM = 0,
+		SELECTED_SYSTEM = 1,
+	};
+
 	SystemView(Game *game);
 	~SystemView() override;
 	void Update() override;
@@ -176,6 +181,9 @@ public:
 
 	Mode GetDisplayMode() { return m_displayMode; }
 	void SetDisplayMode(Mode displayMode) { m_displayMode = displayMode; }
+
+	SystemSelectionMode GetSystemSelectionMode() { return m_systemSelectionMode; }
+	void SetSystemSelectionMode(SystemSelectionMode systemMode);
 
 	TransferPlanner *GetTransferPlanner() const { return m_planner; }
 	double GetOrbitPlannerStartTime() const { return m_planner->GetStartTime(); }
@@ -199,6 +207,7 @@ private:
 	std::list<std::pair<Ship *, Orbit>> m_contacts;
 
 	Mode m_displayMode;
+	SystemSelectionMode m_systemSelectionMode;
 	bool m_viewingCurrentSystem;
 	bool m_unexplored;
 };
--- a/src/enum_table.cpp
+++ b/src/enum_table.cpp
@@ -171,6 +171,12 @@ const struct EnumItem ENUM_SystemViewMode[] = {
 	{ 0, 0 },
 };
 
+const struct EnumItem ENUM_SystemSelectionMode[] = {
+	{ "CURRENT_SYSTEM", int(SystemView::SystemSelectionMode::CURRENT_SYSTEM) },
+	{ "SELECTED_SYSTEM", int(SystemView::SystemSelectionMode::SELECTED_SYSTEM) },
+	{ 0, 0 },
+};
+
 const struct EnumItem ENUM_SystemViewColorIndex[] = {
 	{ "GRID", int(SystemMapViewport::GRID) },
 	{ "GRID_LEG", int(SystemMapViewport::GRID_LEG) },
@@ -353,6 +359,7 @@ const struct EnumTable ENUM_TABLES[] = {
 	{ "ProjectableTypes", ENUM_ProjectableTypes },
 	{ "ProjectableBases", ENUM_ProjectableBases },
 	{ "SystemViewMode", ENUM_SystemViewMode },
+	{ "SystemSelectionMode", ENUM_SystemSelectionMode },
 	{ "SystemViewColorIndex", ENUM_SystemViewColorIndex },
 	{ "PolitEcon", ENUM_PolitEcon },
 	{ "PolitGovType", ENUM_PolitGovType },
@@ -384,6 +391,7 @@ const struct EnumTable ENUM_TABLES_PUBLIC[] = {
 	{ "ProjectableTypes", ENUM_ProjectableTypes },
 	{ "ProjectableBases", ENUM_ProjectableBases },
 	{ "SystemViewMode", ENUM_SystemViewMode },
+	{ "SystemSelectionMode", ENUM_SystemSelectionMode },
 	{ "SystemViewColorIndex", ENUM_SystemViewColorIndex },
 	{ "PolitEcon", ENUM_PolitEcon },
 	{ "PolitGovType", ENUM_PolitGovType },
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -357,11 +357,14 @@ if (NATURALDOCS)
 	WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
 endif()
 
-find_program(PYTHON NAMES python)
-if (PYTHON)
+find_package(Python2 COMPONENTS Interpreter)
+if (Python2_Interpreter_FOUND)
 	add_custom_target(enums
-		COMMAND scripts/scan_enums.py -o src/enum_table.cpp --pattern='*.h' -r src
-		WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+		COMMAND "${Python2_EXECUTABLE}" scripts/scan_enums.py -o src/enum_table.cpp --pattern='*.h' -r src
+		WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+	)
+else()
+	message(WARNING, "Python 2 not found; enums will not be scanned.")
 endif()
 
 target_link_libraries(pioneer-lib PUBLIC lz4 fmt::fmt)
--- a/data/libs/Player.lua
+++ b/data/libs/Player.lua
@@ -11,6 +11,7 @@
 local Player = package.core["Player"]
 
 local Serializer = require 'Serializer'
+local Engine = require 'Engine'
 local Event = require 'Event'
 local Game = require 'Game'
 local utils = require 'utils'
@@ -430,8 +431,16 @@ local onGameEnd = function ()
 	-- clean up for next game:
 end
 
+local onEnterSystem = function ()
+	-- Return to game view when we exit hyperspace
+	if Engine.GetResetViewOnHyperspaceExit() and Game.CurrentView() ~= "world" then
+		Game.SetView("world")
+	end
+end
+
 Event.Register("onGameEnd", onGameEnd)
 Event.Register("onGameStart", onGameStart)
+Event.Register("onEnterSystem", onEnterSystem)
 Serializer:RegisterClass("CrimeRecord", CrimeRecord)
 Serializer:Register("Player", serialize, unserialize)
 
--- a/data/meta/Constants.lua
+++ b/data/meta/Constants.lua
@@ -221,6 +221,16 @@ local SystemViewMode = {
 ---@type SystemViewMode[]
 Constants.SystemViewMode = {}
 
+-- A <Constants.SystemSelectionMode> string
+---@enum (key) SystemSelectionMode
+local SystemSelectionMode = {
+	CURRENT_SYSTEM = 1,
+	SELECTED_SYSTEM = 2,
+}
+
+---@type SystemSelectionMode[]
+Constants.SystemSelectionMode = {}
+
 -- A <Constants.SystemViewColorIndex> string
 ---@enum (key) SystemViewColorIndex
 local SystemViewColorIndex = {
--- a/data/pigui/modules/settings-window.lua
+++ b/data/pigui/modules/settings-window.lua
@@ -170,7 +170,6 @@ local function showVideoOptions()
 	local displaySpeedLines = Engine.GetDisplaySpeedLines()
 	local displayHudTrails = Engine.GetDisplayHudTrails()
 	local enableCockpit = Engine.GetCockpitEnabled()
-	local enableAutoSave = Engine.GetAutosaveEnabled()
 
 	local c
 	ui.text(lui.VIDEO_CONFIGURATION_RESTART_GAME_TO_APPLY)
@@ -256,11 +255,6 @@ local function showVideoOptions()
 	c,enableCockpit = checkbox(lui.ENABLE_COCKPIT, enableCockpit, lui.ENABLE_COCKPIT_DESC)
 	if c then
 		Engine.SetCockpitEnabled(enableCockpit)
-	end
-
-	c,enableAutoSave = checkbox(lui.ENABLE_AUTOSAVE, enableAutoSave, lui.ENABLE_AUTOSAVE_DESC)
-	if c then
-		Engine.SetAutosaveEnabled(enableAutoSave)
 	end
 
 	c,starDensity = slider(lui.STAR_FIELD_DENSITY, starDensity, 0, 100)
@@ -637,11 +631,22 @@ local function showControlsOptions()
 	end
 end
 
+local function showGameOptions()
+	local enableAutoSave = Engine.GetAutosaveEnabled()
+	local c
+
+	c,enableAutoSave = checkbox(lui.ENABLE_AUTOSAVE, enableAutoSave, lui.ENABLE_AUTOSAVE_DESC)
+	if c then
+		Engine.SetAutosaveEnabled(enableAutoSave)
+	end
+end
+
 local optionsTabs = {
 	["video"]=showVideoOptions,
 	["sound"]=showSoundOptions,
 	["language"]=showLanguageOptions,
-	["controls"]=showControlsOptions
+	["controls"]=showControlsOptions,
+	["game"]=showGameOptions
 }
 
 ui.optionsWindow = ModalWindow.New("Options", function()
@@ -660,6 +665,9 @@ ui.optionsWindow = ModalWindow.New("Options", function()
 
 		mainButton(icons.controls, lui.CONTROLS, showTab=='controls', function()
 			showTab = 'controls'
+		end)
+		mainButton(icons.settings, lui.GAME_OPTIONS, showTab=='game', function()
+			showTab = 'game'
 		end)
 	end)
 
--- a/data/pigui/modules/settings-window.lua
+++ b/data/pigui/modules/settings-window.lua
@@ -633,11 +633,17 @@ end
 
 local function showGameOptions()
 	local enableAutoSave = Engine.GetAutosaveEnabled()
+	local resetViewOnHyperspaceExit = Engine.GetResetViewOnHyperspaceExit()
 	local c
 
 	c,enableAutoSave = checkbox(lui.ENABLE_AUTOSAVE, enableAutoSave, lui.ENABLE_AUTOSAVE_DESC)
 	if c then
 		Engine.SetAutosaveEnabled(enableAutoSave)
+	end
+
+	c,resetViewOnHyperspaceExit = checkbox(lui.RESET_VIEW_ON_HYPERSPACE_EXIT, resetViewOnHyperspaceExit, lui.RESET_VIEW_ON_HYPERSPACE_EXIT_DESC)
+	if c then
+		Engine.SetResetViewOnHyperspaceExit(resetViewOnHyperspaceExit)
 	end
 end
 
--- a/data/pigui/modules/system-view-ui.lua
+++ b/data/pigui/modules/system-view-ui.lua
@@ -270,6 +270,8 @@ end
 
 function Windows.edgeButtons.Show()
 	local isOrrery = systemView:GetDisplayMode() == "Orrery"
+	local isCurrent = systemView:GetSystemSelectionMode() == "CURRENT_SYSTEM"
+
 	if ui.mainMenuButton(icons.reset_view, luc.RESET_ORIENTATION_AND_ZOOM) then
 		systemView:SetVisibility("RESET_VIEW")
 	end
@@ -281,6 +283,16 @@ function Windows.edgeButtons.Show()
 	ui.newLine()
 
 	drawWindowControlButton(Windows.objectInfo, icons.info, lc.OBJECT_INFO)
+
+	-- view control buttons
+	if not isCurrent and ui.mainMenuButton(icons.planet_grid, luc.HUD_BUTTON_SWITCH_TO_CURRENT_SYSTEM) then
+		systemView:SetSystemSelectionMode("CURRENT_SYSTEM")
+	end
+
+	if isCurrent and ui.mainMenuButton(icons.galaxy_map, luc.HUD_BUTTON_SWITCH_TO_SELECTED_SYSTEM) then
+		systemView:SetSystemSelectionMode("SELECTED_SYSTEM")
+	end
+
 	-- visibility control buttons
 	if isOrrery then
 		if ui.mainMenuButton(buttonState[ship_drawing].icon, lc.SHIPS_DISPLAY_MODE_TOGGLE, buttonState[ship_drawing].state) then
@@ -403,7 +415,12 @@ local function getColor(obj)
 end
 
 function Windows.systemName.Show()
-	local path = Game.sectorView:GetSelectedSystemPath()
+	local path
+	if systemView:GetSystemSelectionMode() == "SELECTED_SYSTEM" then
+		path = Game.sectorView:GetSelectedSystemPath()
+	else
+		path = systemView:GetSystem().path
+	end
 	ui.text(ui.Format.SystemPath(path))
 end
 
--- a/data/pigui/modules/system-view-ui.lua
+++ b/data/pigui/modules/system-view-ui.lua
@@ -270,7 +270,6 @@ end
 
 function Windows.edgeButtons.Show()
 	local isOrrery = systemView:GetDisplayMode() == "Orrery"
-	-- view control buttons
 	if ui.mainMenuButton(icons.reset_view, luc.RESET_ORIENTATION_AND_ZOOM) then
 		systemView:SetVisibility("RESET_VIEW")
 	end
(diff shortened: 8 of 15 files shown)


impaktor:
> The current implementation has the downside that if the player is viewing the System Map or System Overview while performing a hyperspace jump, the map is changed as soon as the jump is complete

I suspect this is an edge use case. I think we had a crash bug when doing this, and took a while before it was discovered.

Web-eWorks (approved):
This looks good to me, thank you for tackling it! If no one else volunteers to test first, I'll give it a playtest early next week once my worktree is clean again and merge.

Web-eWorks:
This PR can also of course be extended to tackle some of the issues with the proposed change and I'll defer merge until that's been accomplished at your discretion.

I similarly had the idea that it might be good to have an explicit "view system" button in the Sector Map which opened the System View on that specific system and otherwise decouple the System View from the Sector Map selection.

mwerle:
So I did some testing and it seems that moving the code to LUA breaks it - calling `ResetHyperspaceTarget` in LUA breaks a route as it doesn't advance the hyperspace target to the next system in the route. No idea why it works when called from the C++ but breaks when called from LUA. Possibly because the LUA runs after the C++, and the order is now changed? In either case, _not_ explicitly calling `ResetHyperspaceTarget` at all seems to also work..  looking at the LUA, it appears that it is called anyway as part of `hyperJumpPlanner.onEnterSystem` from the `ship.onEnterSystem` callback.

I've also added code to force-change the view back to the game view when exiting Hyperspace, and I think that might not be a bad option to add to the game. Not sure how to add options though...  

---- 
(offtopic)
I find it a royal PITA to try to reverse-engineer the logic flow when half the logic is in LUA instead of C++..  C++ I can set breakpoints and trace through, LUA is a black box.  Personally I am finding contributing to this mish-mash of C++ and LUA to be quite a challenge. I know it was mentioned that moving logic to LUA is to make it easier to contribute, but.. well, definitely not for me :)

----
(offtopic 2)
How do you use the LUA console? It seems the wiki page is extremely outdated, and I can't figure out how to add some commodities (hydrogen) to the ship in the current version.. that would make testing this feature a lot easier..

I got as far as `require "CargoManager".AddCommodity()" but it wants an actual type for the commodity, not a string. I managed to retrieve a "CommodityType" for "hydrogen" as well, but passing it to `AddCommodity` just gives me errors; something about trying to index a value (number). At this point I gave up.

mwerle:
> Not sure how to add options though...

NVM, figured it out.

mwerle:
Ok, so I've added an option "Reset View on Hyperspace Exit". This option is on a new tab "Options", and I've also moved the "Enable Autosave" option to this tab as it's got nothing to do with the various graphics settings. This is done in 2 commits to logically separate the work.

If "Reset View on Hyperspace Exit" is unchecked, then the game stays on whichever view the player is currently in. If the current view is the "System Map" or the "System Overview", this will change to the current system when the hyperspace jump is completed.

If "Reset View on Hyperspace Exit" is checked, the player is ejected back to the Game View from whichever screen they are currently in.

Brief testing appears to show no issues.

If people like these changes, I will squash the various commits and convert the PR from draft to a real PR. Otherwise open to improvement suggestions.

mwerle:
> I similarly had the idea that it might be good to have an explicit "view system" button in the Sector Map which opened the System View on that specific system and otherwise decouple the System View from the Sector Map selection.

I'm not entirely sure I follow this thought.  Which system should be opened from this explicit "View System" button?  The only one that makes sense is the currently selected system.. which is what happens already. Unless you add a right-click menu for a star..

impaktor:
@mwerle I think you might want to try the debug tool we have in-game: Ctrl+i to give yourself hydrogen, ship, enemy, etc.
As for lua console, I believe it's bound to the key you've set in settings.

mwerle:
> I think you might want to try the debug tool we have in-game

Thank you; that is very useful!  I had actually stumbled on this before, but didn't see the section where you can buy commodities through it.

>  As for lua console, I believe it's bound to the key you've set in settings.

Well, yes, I can open it.. but the example given on the [wiki page](https://wiki.pioneerspacesim.net/wiki/Lua_Console) to add cargo doesn't work (I didn't try the others) and I couldn't figure out a way to make it work. No longer required due to the debug tool, but we might want to update the wiki..

mwerle:
Ok, big changes to the proposal.

No longer change the currently selected system on hyperspace exit.

Instead, add a new button to the System Map and System Overview screens to switch the display between "Current" system and "Selected" system (defaults to "Selected" to keep current mode of operation). If the display has been switched to "Current" system, then the map is updated on hyperspace exit to show the arrival system.

Also add a new game setting to "Reset View on Hyperspace Exit". When checked, the view is reset to the game view when exiting hyperspace. The view itself is not changed (ie, internal/external/direction). There's an argument to reset the view to internal-forward..

----

Also found and fixed an edge-case bug where the next hyperspace system was not being properly updated on hyperspace exit if the player did not enter the "Sector Map" following starting and loading the game, and immediately starting a hyperspace jump.

----

Also refactored the "edge buttons" of all map screens - they are now aligned to the top-right, and the first set of buttons is the same between all screens.

----

Feedback welcome; especially choice of icons. Have done some testing, but not comprehensive yet (will do so after dinner).

## Test plan:

### Game View:

- [ ] Jump while in game-view with "reset view" unchecked
       EXPECT: jump completes, no change to views or systems
       ACTUAL:
      
- [ ] Jump while in game-view with "reset view" checked
       EXPECT: jump completes, no change to views or systems
       ACTUAL:

### No-Game and Non-Map View (eg, Personal Details):
- [ ] Jump while in a non-map view with "reset view" unchecked
       EXPECT: jump completes, view doesn't change
       ACTUAL:

- [ ] Jump while in a non-map view with "reset view" checked
       EXPECT: jump completes, view returns to game-view; game view should be same as before
       ACTUAL:  

### Sector Map View:
- [ ] Jump while in Sector Map view with "reset view" unchecked
       EXPECT: jump completes, map is updated with Current System (and possibly Hyperspace Target) updated
       ACTUAL:

- [ ] Jump while in Sector Map view with "reset view" checked
       EXPECT: jump completes, view returns to game-view; game view should be same as before
       ACTUAL:

### System Map or System Overview:
- [ ] Jump while in view with "reset view" unchecked and "system selection" is set to "selected"
        EXPECT: jump completes, no change to view
       ACTUAL:

- [ ] Jump while in view with "reset view" unchecked and "system selection" is set to "current"
      EXPECT: jump completes, map changes to arrival system
      ACTUAL:

- [ ] Jump while in a Sector Map view with "reset view" checked
      EXPECT: jump completes, view returns to game-view; game view should be same as before
      ACTUAL:

mwerle:
## Test Run : Micha 2024-09-30

### Game View:

- [ ] Jump while in game-view with "reset view" unchecked
    - EXPECT: jump completes, no change to views or systems
    - ACTUAL:
       - Jump: completed successfully
       - View: unchanged
       - Selected system: unchanged
       - Hyperspace target: moved to next system in route
      
- [ ] Jump while in game-view with "reset view" checked
    - EXPECT: jump completes, no change to views or systems
    - ACTUAL:
       - Jump: completed successfully
       - View: unchanged
       - Selected system: unchanged
       - Hyperspace target: moved to next system in route

### No-Game and Non-Map View (eg, Personal Details):
- [ ] Jump while in a non-map view with "reset view" unchecked
    - EXPECT: jump completes, view doesn't change
    - ACTUAL:
       - Jump: completed successfully
       - View: unchanged (log view: new entry added on arrival)
       - Selected system: unchanged
       - Hyperspace target: moved to next system in route

- [ ] Jump while in a non-map view with "reset view" checked
    - EXPECT: jump completes, view returns to game-view; game view should be same as before
    - ACTUAL:
       - Jump: completed successfully
       - View: returned to game-view
       - Selected system: unchanged
       - Hyperspace target: moved to next system in route

### Sector Map View:
- [ ] Jump while in Sector Map view with "reset view" unchecked
    - EXPECT: jump completes, map is updated with Current System (and possibly Hyperspace Target) updated
    - ACTUAL:
       - Jump: completed successfully
       - View: remains on Sector Map; animated to show new current system and hyperspace target system
       - Selected system: unchanged
       - Hyperspace target: moved to next system in route

- [ ] Jump while in Sector Map view with "reset view" checked
    - EXPECT: jump completes, view returns to game-view; game view should be same as before
    - ACTUAL:
       - Jump: completed successfully
       - View: remains on Sector Map; animated to show new current system and hyperspace target system; then returns to Game View (same view as before; front, left, right, etc)
       - Selected system: unchanged
       - Hyperspace target: moved to next system in route

### System Map or System Overview:
- [ ] Jump while in view with "reset view" unchecked and "system selection" is set to "selected"
    - EXPECT: jump completes, no change to view
    - ACTUAL:
        - Jump: completed successfully
        - View: remains on System Map, displayed system unchanged
        - Selected system: unchanged
        - Hyperspace target: moved to next system in route

- [ ] Jump while in view with "reset view" unchecked and "system selection" is set to "current"
    - EXPECT: jump completes, map changes to arrival system
    - ACTUAL:
        - Jump: completed successfully
        - View: remains on System Map, displayed system changed to new system
        - Selected system: unchanged
        - Hyperspace target: moved to next system in route

- [ ] Jump while in a Sector Map view with "reset view" checked
    - EXPECT: jump completes, view returns to game-view; game view should be same as before
    - ACTUAL:
        - Jump: completed successfully
        - View: returned to Game View; in-flight view is same as before
        - Selected system: unchanged
        - Hyperspace target: moved to next system in route

impaktor:
regarding pain of lua/c++ development:
- there's ability for hot reloading the lua files, so you don't need to restart pioneer between each change
- there's ability to send lua code from your IDE to running pioneer instance.

mwerle:
> regarding pain of lua/c++ development:
> 
>     * there's ability for hot reloading the lua files, so you don't need to restart pioneer between each change
> 
>     * there's ability to send lua code from your IDE to running pioneer instance.

It's mostly not being able to trace through to see the logic in the Lua executing with variable views. It took me well over 3 hours to track down a rather trivial bug, which _should_ have taken less than 10 minutes.  And I only cracked it because I went back to the "poor man's debugger" and added `printf`'s everywhere :P  Oh well, it is what it is. (In Japanese: shoganai; wonderful word!)

But yes, the hot-reload is very nice when fiddling with UI layouts.  And, I suspect, tweaking of AI routines for the NPC ships. Personally I'd still leave the vast majority of the game logic in the C++ though.. just add relevant Lua hooks everywhere for modders.

Web-eWorks:
> I find it a royal PITA to try to reverse-engineer the logic flow when half the logic is in LUA instead of C++.. C++ I can set breakpoints and trace through, LUA is a black box. 

This is definitely a pain-point to be sure. We have a function `void pi_lua_stacktrace(lua_State *l)` that you can execute inside a C++ side breakpoint/assert to print the Lua callstack, and a *very* verbose set of context information is written to the `output.txt` log when Lua encounters an error, but a general-purpose Lua debugger is something we're definitely missing. It's not the easiest thing to implement either, though there might be an existing project we can leverage in this regard for IDE compatibility.

> How do you use the LUA console? It seems the wiki page is extremely outdated, and I can't figure out how to add some commodities (hydrogen) to the ship in the current version.. that would make testing this feature a lot easier..

The wiki is *quite* outdated and the developer-facing portions are in the process of being moved to https://dev.pioneerspacesim.net/. We have a debug tool for cargo as you've found, so updating that wiki has been historically low-priority. If you want to programmatically add cargo you can do so via:

```lua
player:GetComponent('CargoManager'):AddCommodity(require 'CommodityType'.GetCommodity("hydrogen"), 2)
```

Note that we also have a Lua [codedoc](https://codedoc.pioneerspacesim.net) providing documentation for the various Lua utilities and classes.

If you're using VSCode or another language-server-protocol compatible IDE, the [LuaLS](https://github.com/LuaLS/lua-language-server) tool is incredibly helpful and provides type information and autocomplete when working with type-annotated Lua files. We're incrementally working on expanding the type information / annotation coverage of the Lua codebase as time goes by.

Web-eWorks:
> Ok, so I've added an option "Reset View on Hyperspace Exit". This option is on a new tab "Options", and I've also moved the "Enable Autosave" option to this tab as it's got nothing to do with the various graphics settings. This is done in 2 commits to logically separate the work.

This sounds good. I'd name the tab "Game Options" specifically, as it has to do with in-game behaviors.

> If "Reset View on Hyperspace Exit" is checked, the player is ejected back to the Game View from whichever screen they are currently in.

I'd recommend renaming this to "Close Map on Hyperspace Exit" - "Reset View" is semantically overloaded and can refer to changing the camera mode, changing the camera orientation, or changing to the WorldView.

> If people like these changes, I will squash the various commits and convert the PR from draft to a real PR. Otherwise open to improvement suggestions.

I like this change, thank you!

---

> I'm not entirely sure I follow this thought. Which system should be opened from this explicit "View System" button? The only one that makes sense is the currently selected system.. which is what happens already. Unless you add a right-click menu for a star..

The idea here was to completely remove the automatic system selection of the system map, and instead add an explicit button in the sector map to open the system view for the selected system, which system would remain open in the system map even if the sector map cursor was moved.

Because the default behavior would be for the system view to not change the system viewed unless the user explicitly commanded the viewing of a different system, we'd most likely need a separate button accessible from the WorldView to open the current system in the system map.

This is a quite involved change however, so I'm merely proposing it for discussion / evaluation rather than implementation.

---

Note that you can collapse long lists on Github inside a `<details>` tag like so - I've gone ahead and done so for the the testing lists above to make the discussion easier to read.

```md
<details>

The empty lines are important
Some long list

</details>
```

Web-eWorks:
> Also refactored the "edge buttons" of all map screens - they are now aligned to the top-right, and the first set of buttons is the same between all screens.

Note that these edge buttons are loosely intended to be converted to use the `sidebar.lua` widget used on the main WorldView screen, similar to [this mockup](https://i.imgur.com/NxCWiYQ.png). The map views are pending a UI design/normalization pass which may involve some aggressive redesigning.

> Personally I'd still leave the vast majority of the game logic in the C++ though.. just add relevant Lua hooks everywhere for modders.

There is a lot of opinionated game logic in Pioneer that is neither a performance concern nor a "universal constant" but is still hard-coded in C++ and cannot be altered without significant work.

My personal approach is that logic in C++ should be "composable" - it's a set of building blocks that the "default mod" arranges to produce the intended balance and behavior of the game. Other mods can then override and alter the arrangement of those building blocks to produce a different result.

What we have now is the "sprinkle hooks everywhere" method, and it's produced a very brittle and inflexible codebase. For example, fuel scooping is "game logic" that is (and was even more so) completely hardcoded in C++. As a result, mods cannot e.g. allow scooping a star, change the minimum atmospheric density to scoop, or allow scooping any other material other than thruster fuel.

...that's a bit of an off-topic rant though. I do agree that a lot of game logic can (and often should) live in C++, but there has to be a great amount of attention paid to how that logic interconnects with other parts of the game.

---

All that out of the way, I'll try to give this branch a playtest soon and provide more actionable feedback about the concerns you raised above. My mental queue is currently full, so it may be a few days however.

mwerle:
>  I'd name the tab "Game Options"

**Question**

Do you mean the internal name of the tab, or the tooltip text (or both)? I simply chose "Options" for the tooltip to re-use an existing translation but trivial to update.

----

> I'd recommend renaming this to "Close Map on Hyperspace Exit" - "Reset View" is semantically overloaded and can refer to changing the camera mode, changing the camera orientation, or changing to the WorldView.

Well, it does the last - returns to the Game View (is that "WorldView"? I am meaning the in-flight view, either internal or external) irrespective of what the current view is. It doesn't just return to the Game View from the various map screens, but also from the various informations screens which can be accessed in-flight.

There's an argument for not only returning to the Game View but also resetting the view to "internal-front" so the player can instantly react to hostile actions etc.  For now I did not change the actual view configuration as the player might prefer to play in external-camera mode.

----

> The idea here was to completely remove the automatic system selection of the system map, and instead add an explicit button in the sector map to open the system view for the selected system, which system would remain open in the system map even if the sector map cursor was moved.

So the user can already toggle the automatic system selection in the Sector Map; if the user doesn't like this, they can always disable that, which means the System Map / System Overview remain on the currently selected system.

And with my addition of toggling between the currently selected system and the current system there's no need for the user to change the currently selected system unless they want to view a different system.

So while it doesn't quite work the way you envision with explicit buttons and a second "selected map system" state, the user-story that you envision can, I think, be achieved with the currently proposed changes.

----

> The map views are pending a UI design/normalization pass which may involve some aggressive redesigning.

I can revert the edge-button changes; it just made a little bit more sense to me to have them in a more fixed location and with a more consistent ordering.  

If no coding has actually been starting on the proposed UI update, then I don't think my changes would interfere with that work and, until that is done, it is a slightly improvement IMHO.  

**Question**

I can leave it as-is for now or revert, please let me know.

----

Thank you for your time in replying and evaluating this PR, as well as your additional information about the game design and debugging/investigating the Lua components while developing. I will try your suggestions.

impaktor:
> returns to the Game View (is that "WorldView"? I am meaning the in-flight view

Yes, that's the WorldView
Then we have SectorView and SystemView
(We used to have GalaxyView as well, that could be resurrected, if anyone is interested. Didn't serve a function, other than immersion)

I must say, it's really great to see some more contributors playing with our source code. Also @Web-eWorks is currently swamped with #5734 (feel free to help playtest and/or review, although the latter would be daunting for any mortal man), so replies & review of your PRs might be a while.

> As initial feedback, personally I feel that mockup is condensing too much information into a single side-view, which the user needs to manage by opening/collapsing various items in order to see all the information. This would be especially true on lower screen resolutions. Is there a reason the right-side-bar isn't used for some of this information? Most monitors these days are significantly wider than tall. Also it would require opening the side-view before being able to use the various action buttons. Some of these buttons only toggle additional information views, but some perform actual actions. I think it makes sense to condense the various information views into a sidebar, but not so sure about the action buttons.

I'll let @bszlrd answer :point_up:

Web-eWorks:
> Do you mean the internal name of the tab, or the tooltip text (or both)?

The tooltip text specifically shown to the user, as well as an appropriate icon. Internal name can be whatever is semantically connected to the text that is shown to the user - either is fine.

> Well, it does the last - returns to the Game View

I saw that it closes e.g. the Personal Info screen as well on exiting hyperspace after writing that comment. Leaving the name of the option as-is is fine in light of that information - a clearer name would be too verbose.

> And with my addition of toggling between the currently selected system and the current system there's no need for the user to change the currently selected system unless they want to view a different system.

This is good, thanks! I like what you're describing. I expect this to be ergonomic enough that there's no need for my proposed solution.

> I feel that mockup is condensing too much information into a single side-view [...] Is there a reason the right-side-bar isn't used for some of this information? Most monitors these days are significantly wider than tall..

Yes - it's a mockup to show general layout and hierarchy, not final layout and design. :smile:

My intention when implementing that mockup would be to split functionality between left and right sidebars as needed, moving commonly-used context-free buttons (i.e. rotate map, reset view) either to a central button dock at the top/bottom center of the screen, or simply as a "regular button" in the row of sidebar toggle buttons.

> I can leave the edge-button refactor as-is for now or revert, please let me know your preference.

I've not yet gotten a chance to play with it or review the code, but I'll say leave as-is. This was mostly to inform you about future plans for the screen so you don't feel like your work is accepted and then (eventually) immediately discarded in a redesign.

---

> Thank you for your time in replying and evaluating this PR

And thank you for your time developing it! I'm happy to spend the time I have available to accelerate other contributors - many hands make light work, after all. I don't know if I can promise a timeline for "final review" but as soon as I clear some other IRL responsibilities off my plate I'll try to give this a formal review + playtest.

mwerle:
> The tooltip text specifically shown to the user, as well as an appropriate icon. Internal name can be whatever is semantically connected to the text that is shown to the user - either is fine.

Fixed. The tooltip is not "Game Options".  I left the icon as the "cog" as that is a very commonly used icon for settings/options.  If there is a better icon to use, please suggest one.

----

Rebased/squashed commits.  Based on the initial feedback, I think this PR is now ready for proper review.

data/lang/ui-core/de.json
```diff
@@ -715,6 +715,10 @@
     "description": "",
     "message": "Vollbildmodus"
   },
+  "GAME_OPTIONS": {
```
impaktor, line 718: You only need to update the en.json file, it's the "master", this will then propagate new strings to all other language files (might be temporarily not working though, but that's on me).

mwerle: As I speak German, I added the German translation as well..

impaktor: Your edit will probably get over-written by transifex. I can never quite remember what happens when editing the non-en.json files.

If you want to maintain the german translation, that should be done through transifex.

data/lang/ui-core/en.json
```diff
@@ -2027,6 +2039,14 @@
     "description": "Paintshop button",
     "message": "Reset Preview"
   },
+  "RESET_VIEW_ON_HYPERSPACE_EXIT": {
+    "description": "Reset View on Hyperspace Exit option",
+    "message": "Reset View on Hyperspace Exit"
+  },
+  "RESET_VIEW_ON_HYPERSPACE_EXIT_DESC": {
+    "description": "tooltip description in settings menu",
+    "message": "reset view to main game view when exiting from hyperspace"
```
Web-eWorks, line 2048: Leading word should be capitalized.

src/enum_table.cpp
```diff
@@ -171,6 +171,12 @@ const struct EnumItem ENUM_SystemViewMode[] = {
 	{ 0, 0 },
 };
 
+const struct EnumItem ENUM_SystemSelectionMode[] = {
+	{ "SELECTED_SYSTEM", int(SystemView::SystemSelectionMode::SelectedSystem) },
+	{ "CURRENT_SYSTEM", int(SystemView::SystemSelectionMode::CurrentSystem) },
+	{ 0, 0 },
+};
+
```
Web-eWorks, line 179: Note that these are generated by running `scripts/scan_enums.py -o src/enum_table.cpp --pattern='*.h' -r src/`. The script reads appropriately-annotated C++ `enum` declarations and generates the appropriate entries. Manual changes to these files will be overridden.

src/SystemView.h
```diff
@@ -168,6 +168,11 @@ class SystemView : public PiGuiView, public DeleteEmitter {
 		Atlas = 1
 	};
 
+	enum class SystemSelectionMode {
+		CurrentSystem = 0,
+		SelectedSystem = 1,
+	};
```
Web-eWorks, line 174: This enum should be annotated for `scan_enums.py` to generate `enum_table` entries.

```suggestion
	enum class SystemSelectionMode { // <enum name=SystemSelectionMode scope='SystemView::SystemSelectionMode' public>
		CurrentSystem = 0,
		SelectedSystem = 1,
	};
```

Web-eWorks (changes requested):
I'm quite happy with this! Two minor nitpicks in terms of consistency and maintainability, but overall I like the effect of the changes.

Something to consider for a future PR - the player might want a "shortcut" to open the system view and switch to current system view mode from the WorldView (and similarly for selected system mode from the Sector Map) - it's extra cognitive load to check that the map is in the correct view mode after opening it.

That's not something I expect you to implement at current however; it's a better topic for later down the road when the Sector Map UI gets looked at for a rework.

data/lang/ui-core/en.json
mwerle: Sure. But i used the surrounding definitions as a template to stay with the existing style.

do you have a coding guideline/style for new code? perhaps add pre-commit hooks?

src/SystemView.h
mwerle: Fixed in 62aad68a0b87fcf1a7a9d57a8022668cf50b7a65

----

Generated code should (IMHO) never be checked into source control; instead, the relevant generator should be invoked during configuration and/or build time..(*)  However, requiring this would mean developers would need a version of Python2 installed.. which is currently not the case as long as a developer doesn't add a new enum which should be usable by Lua.

That said, if I'd read the top of the file I guess I would have figured it out too ;) I literally only searched for the other enum and inserted all the required bits for my new one.

(*)  Which, it appears, does happen on system which have a version of Python 2 in the path as `python` - on modern Debian systems, this is not the case.   I fixed this in 5bcbfd6da05003e9b9f5bca91bab517d0b060bd0

src/enum_table.cpp
mwerle: Fixed - please see https://github.com/pioneerspacesim/pioneer/pull/5922#discussion_r1789093383 for details.

data/lang/ui-core/en.json
mwerle: Fixed in fdb32c9e85790942aaa6a366b855a544751bb349

data/lang/ui-core/de.json
mwerle: Removed the german translations.. will add them again when they become available on transifex.

That said, I think it would be better if people could supply translations however they wished, with transifex being one option. transifex should populate itself from the existing data.. but no idea if that actually is possible. First time hearing about that system.

mwerle:
Rebased on `master` and added requested changes.  Will squash and force-push once the review is accepted.

data/lang/ui-core/en.json
Web-eWorks: In this area specifically, there's no enforceable style guide because the content of the strings are heavily context-based. In this case, it's just a matter of ensuring tooltip strings follow standard capitalization rules and are normalized with the rest of the interface (in this case, the string you copied from ideally also needs to be capitalized.)

We have/had a `clang-format` script to format C++ code; however, we've since disabled that as a pull-request check because `clang-format` has a very annoying habit of introducing opinionated formatting changes (or possibly bugs, who knows) that cannot be controlled via its own config.

Between that and the difficulty of keeping a sane config across 4+ simultaneous versions of clang-format (all of which throw errors if they see a config key from a later version of clang-format) it turned out to be significantly more trouble than it was worth.

src/SystemView.h
Web-eWorks: So there's one specific problem with this approach - using the Ninja generator, every time a rebuild or incremental compile is triggered every file which includes `enum_tables.h` is recompiled as the file has "changed" on disk. Worse yet, it looks like those changes to `enum_tables.h` aren't actually picked up the first time the project is built following a valid change to an annotated enum - it seems CMake (or in this case Ninja) scans dependencies before invoking the `scan_enums.py` script, so it's actually the *second* rebuild that produces a valid program state.

I'd ask that you remove the dependency between `pioneer-core` and the `scan_enums` invocation - instead, leave it defined as a CMake command and we can document it as something that should be run manually in the (very low-frequency) case of modifying an annotated enum value.

Web-eWorks: While I do agree with you in the general case that autogenerated code should likely not be checked into the build system at all, this is a sort of "gray area" where the file typically changes only once every few months, does not have easily-tracked dependencies to determine when it should be re-generated, and touches several distinct areas of the codebase (for example, the script also generates `data/meta/Constants.lua`). As a result it's (historically, in this codebase) "easier" for it to be checked into source control and manually re-generated when actually necessary.

Web-eWorks (approved):
Pending the change to the `scan_enums` CMake target, I give this PR my blessing. Thank you @mwerle for the idea and implementation!

src/SystemView.h
mwerle: I couldn't figure out a way of running it manually and assumed that it just ran at some point during the build process, hence adding the explicit dependency to ensure it ran early in the build.

`ninja enums` gave me an invalid target warning. I'll investigate why.

Good catch that it doesn't build properly the first time around. 

Perhaps instead of having it in the CMake just document that it needs to be invoked? Or run it as part of the build configuration phase?

mwerle: ...and trying again now it works as expected, with both `ninja` and `make` invoking the command..  so removing the target dependency, but leaving in the change to properly find Python 2.

mwerle:
Remove target dependency for the `enums` target and rebased on current `master`.

Thank you for reviewing and the various comments, suggestions, and assistance with implementing these changes.

src/SystemView.h
Web-eWorks: Thanks! Having it in CMake is mostly so it can be quickly invoked, e.g. via `cmake --build build/ --target enums` or `make -C build enums`, as well as serving as "living documentation" for the correct invocation of the script.

impaktor:
> Perhaps instead of having it in the CMake just document that it needs to be invoked? Or run it as part of the build configuration phase?

If there's anything missing from COMPILING.txt that can be helpful for package maintainers, or anyone wanting to compile & run pioneer, please add.

PR #5925, opened 2024-09-30
Bugfix/fix hyperspace route update bug 5924

### LINKED ISSUE #5924: Hyperspace route does not update following a jump if Sector Map was never opened.
### Observed behaviour
If performing a hyperspace jump on a route after loading the game, but without entering the Sector View, the route is not updated properly and it is not possible to jump to the next system in the route.

### Expected behaviour
Route is updated and able to continue jumping along the route.

### Steps to reproduce

1. Create a savegame which has a route of at least 2 jumps plotted (and a ship with at least 2t of Hydrogen)
2. Exit, restart the game, and load the savegame
3. Perform a hyperspace jump (F7)
4. Attempt to perform the next hyperspace jump -> fails

Upon inspecting the Sector View, it is seen that the hyperspace destination is "stuck" on the current system, with the "Route Jumps" starting on the current system instead of the next system in the route.

If the "Sector View" is opened _before_ performing the jump, the route works as intended.

"master", Debian GNU/Linux 12

- [output.txt](https://github.com/user-attachments/files/17191903/output.txt)
- [Hyperspace_test.zip](https://github.com/user-attachments/files/17191900/Hyperspace_test.zip)  (remove ".zip" extension)

mwerle:
Fixes #5924 

This bug manifested itself in the edge-case of loading a game with an active hyperspace route, and jumping to the next system in the route without entering the Sector View first. On arrival in the new system, the Hyperspace Target is not updated to the next system in the route.
    
The bug is caused by the implementation of the `hyperJumpPlanner` Lua  module tracking the size of the route in a local variable "route_jumps".  This variable is not initialized until the Sector View is opened, but is crucial for correctly updating the HyperSpace target system after a jump.
    
2 ways were identified to fix this bug:
    
1. update the variable in the "buildJumpRouteList" function
2. remove the variable and read the size of the "hyperjump_route" table directly wherever it is needed.
    
The second option was chosen for this fix as it removes an unnecessary local variable at the expense of slightly more verbose code. This is deemed to be more robust as future refactoring or feature additions will not introduce similar issues where the variable and the table size get out of sync.

(EDIT: fix wording to use present tense to describe the bug)

### DIFF
--- a/data/pigui/modules/hyperjump-planner.lua
+++ b/data/pigui/modules/hyperjump-planner.lua
@@ -22,7 +22,6 @@ local hyperJumpPlanner = {} -- for export
 
 -- hyperjump route stuff
 local hyperjump_route = {}
-local route_jumps = 0
 local current_system
 local current_path
 local map_selected_path
@@ -66,8 +65,8 @@ local function showInfo()
 
 		textIcon(icons.route_destination, lui.FINAL_TARGET)
 
-		if route_jumps > 0 then
-			local final_path = hyperjump_route[route_jumps].path
+		if #hyperjump_route > 0 then
+			local final_path = hyperjump_route[#hyperjump_route].path
 			ui.sameLine()
 			if ui.selectable(ui.Format.SystemPath(final_path), false, {}) then
 				sectorView:SwitchToPath(final_path)
@@ -314,7 +313,6 @@ function hyperJumpPlanner.display()
 	current_path = Game.system and current_system.path -- will be nil during the hyperjump
 	current_fuel = Game.player:GetComponent('CargoManager'):CountCommodity(fuel_type)
 	map_selected_path = sectorView:GetSelectedSystemPath()
-	route_jumps = sectorView:GetRouteSize()
 	showHyperJumpPlannerWindow()
 end -- hyperJumpPlanner.display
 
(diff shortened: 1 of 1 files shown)


Web-eWorks:
This is a good change, thanks for taking the second option. I wouldn't worry about the verboseness - as long as `hyperjump_route` is already synchronized with the SectorView properly then this code removes a redundant state variable.

As a convention, we prefer for the person merging the PR (or @impaktor recently) to push changelog commits. For a small bugfix like this I'm not going to ask you to re-do it, but we typically add changelog entries based on the date the change is available to the player, not the date the PR was opened. It's also more efficient to do a batch of changelog entries in a single update rather than a lot of mini-commits adding a single line each.

If @impaktor hasn't already asked you, what name would you prefer to be added to AUTHORS.txt with (if any)?

mwerle:
> I'm not going to ask you to re-do it,

My apologies, I wasn't aware. I was attempting to reduce other contributor effort. It's trivial to remove the second commit from this PR... please let me know what you prefer.

> what name would you prefer to be added to AUTHORS.txt 
"Michael Werle", thank you :)

impaktor:
I've added you to AUTHORS on my local copy. I'll include it in next changelog update.

mwerle:
> Could you please drop the changelog update

Dropped.

PR #5927, opened 2024-10-04
Feat/save game manager

### LINKED ISSUE #5921: savefiles directory not created on program start
Follow-up issue to #5917 to track an improvement to the fix. The current implementation delays creating the "savefiles" directory until a game is being saved.

Details in the #5919 PR discussion thread, the TL;DR is below:

----

After the fix, there is still a popup "user://savefiles is not a directory" when attempting to load a game if the "savefiles" directory has not yet been created.

Fix by creating the "savefiles" directory on program startup in `Pi::App::OnStartup()` or as a task inside `StartupScreen:Start()` in "Pi.cpp".

mwerle:
Fixes #5921

The [first commit](64d48a8cef8f5a636b0a78b5c087045793456866) fixes the issue. This prevents a spurious popup from being shown when the game is started for the first time and the user attempts to load a game, as the creation of the "savefiles" directory was deferred until the first savegame was created. It also moves the remaining savegame-related items from the "Pi" class into the "Game" class to clean up the codebase.

The [second commit](63d31a389766f40487b91d4f8eb66a0efa98eb42) refactors the implementation to break all the savegame handling out of the "Game" class and into a new "SaveGameManager" class. This refactor is purely to compartmentalise the existing implementation better and follows the pattern laid out by the "ModManager" class.

The [third commit])(b60d6ca958617273ff3266340ee2626172a686f7) adds a "ListSaves" API, which allows a refactor of the Lua to remove the direct reference to the "savefiles" string and instead delegate this to the "SaveGameManager".

----

Initial testing shows all functionality behaving as expected (Load/Save games, list games, Autosave games).

No tests were made as to loading older saves to ensure upgrades still work, but there is nothing in this work to suggest that this should break.

----

NOTE: Only the first commit is necessary to fix the actual issue; the others are optional and can be reverted.

### DIFF
--- a/src/SaveGameManager.h
+++ b/src/SaveGameManager.h
@@ -56,7 +56,9 @@ public:
 
 	/** Delete a savegame file. */
 	static bool DeleteSave(const std::string &name);
+
+	/** Return a  list of saved games. */
+	static std::vector<FileSystem::FileInfo> ListSaves();
 };
 
 #endif
-
--- a/src/SaveGameManager.h
+++ b/src/SaveGameManager.h
@@ -1,0 +1,62 @@
+// Copyright © 2008-2024 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+#ifndef _SAVEGAMEMANAGER_H
+#define _SAVEGAMEMANAGER_H
+
+#include "FileSystem.h"
+#include "JsonFwd.h"
+
+#include <string>
+#include <vector>
+
+class Game;
+
+class SaveGameManager {
+public:
+	static void Init();
+	static void Uninit();
+
+	/** Return currently supported version of the save game format. */
+	static int CurrentSaveVersion();
+
+	/** Return the savegame direcotry name.
+	 * NOTE: This should only be used for logging purposes. The game nor any
+	 * mods should ever directly access the savegame directory.
+	 */
+	static const std::string GetSaveGameDirectory();
+
+	/** Check whether the named save exists and can be opened.
+	 * NOTE: This does not check whether the named save actually contains a
+	 *       compatible savegame file.
+	 */
+	static bool CanLoadGame(const std::string &name);
+
+	/** Load a game and return a new Game object.
+	 * This function will throw an exception if there is an error during the
+	 * loading process.
+	*/
+	static Game *LoadGame(const std::string &name);
+
+	/** Loads a game and returns it as a Json object.
+	 * NOTE: this function will not perform any sanity checks that the loaded
+	 * file actually contains a Pioneer savegame. It will return a null Json
+	 * object if the contents of the file could not be decoded as a Json object.
+	 *
+	 * This is only required to support the LUA game upgrade/recovery
+	 * implementation.
+	 */
+	static Json LoadGameToJson(const std::string &name);
+
+	/** Save the game.
+	 * NOTE: This function will throw an exception if an error occurs while
+	 * saving the game.
+	*/
+	static void SaveGame(const std::string &name, Game *game);
+
+	/** Delete a savegame file. */
+	static bool DeleteSave(const std::string &name);
+};
+
+#endif
+
--- a/src/lua/LuaGame.cpp
+++ b/src/lua/LuaGame.cpp
@@ -322,6 +322,47 @@ static int l_game_delete_save(lua_State *l)
 }
 
 /*
+ * Function: ListSaves
+ *
+ * Retrieves a list of all game saves along with their last-modified time.
+ *
+ * > Game.ListSaves()
+ *
+ * Parameters:
+ *
+ * Return:
+ *
+ *   saves - a list of save-games as names and modification-date
+ *
+ * Availability:
+ *
+ *   September 2024
+ *
+ * Status:
+ *
+ *   experimental
+ */
+static int l_game_list_saves(lua_State *l)
+{
+	auto saves = SaveGameManager::ListSaves();
+
+	lua_newtable(l);
+	int savesTable = lua_gettop(l);
+	int saves_len = 0;
+
+	for (const auto &save : saves)
+	{
+		LuaTable saveTable(l);
+		saveTable.Set("name", save.GetName());
+		saveTable.Set("mtime", save.GetModificationTime());
+
+		lua_rawseti(l, savesTable, ++saves_len);
+	}
+
+	return 1;
+}
+
+/*
  * Function: EndGame
  *
  * End the current game and return to the main menu.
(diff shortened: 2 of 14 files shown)


src/lua/LuaUtils.cpp
```diff
@@ -303,6 +303,22 @@ static int l_readonly_table_ipairs(lua_State *l)
 	return 3;
 }
 
+void pi_lua_push_date_time(lua_State *l, const Time::DateTime &dt)
```
Web-eWorks, line 306: Note that this could also be exposed as a `pi_lua_generic_push` overload, allowing the use of `DateTime` objects via the `LuaPush<T>()` mechanism. See `LuaPushPull.h` for details on the specific API arguments.

The `LuaPush/Pull` API is used in quite a few places to provide C++ template "glue" for a nicer API when calling or moving data into Lua and vice versa.

src/lua/LuaGame.cpp
```diff
@@ -316,7 +317,49 @@ static int l_game_save_game(lua_State *l)
...
+	lua_newtable(l);
+	int savesTable = lua_gettop(l);
+	int saves_len = 0;
+
+	for (const auto &save : saves)
+	{
+		lua_newtable(l);
+		pi_lua_settable(l, "name", save.GetName().c_str());
+		pi_lua_push_date_time(l, save.GetModificationTime());
+		lua_setfield(l, -2, "mtime");
```
Web-eWorks, line 358: Note that if the related comment about `pi_lua_generic_push` is implemented, this can use the LuaTable wrapper for convenience:

```c++
LuaTable(l) save;
save.Set("name", save.GetName());
save.Set("mtime", save.GetModificationTime());
```

Web-eWorks (commented):
I wasn't entirely certain how the API would look when I started reading this PR, but I'm quite happy with what I'm seeing!

I definitely won't ask for a revert here - this makes several save-related tasks conceptually simpler and I'm happy with Game being responsible for only the currently running game state.

I've left a comment regarding the exposure of the DateTime push-to-lua function - when able, exposing types with "sane representations" to the LuaPush APIs allows us to have better interoperability between more C++ and Lua code.

While we technically have existing source code documentation using the Javadoc style and ostensibly consume it with Doxygen, it's not a hard requirement to adhere to any one documentation format or style.

Lua-facing API on the other hand is documented via Natural Docs 2, but it's quite flexible in terms of what syntax it will parse. The `Availability` and `Status` fields are of limited practical use (no one goes around promoting APIs to "stable" so they stay experimental for 10+ years), but if you're willing to fill them in when writing the documentation, feel free.

---

As mentioned on the other PR, review is pending my availability (will try to set aside time on Saturday) but from my read-through of the code I like this PR.

src/lua/LuaUtils.cpp
mwerle: I did the bare minimum as my knowledge of Lua/C++ interoperability is a work-in-progress, and my C++ skills also aren't the best (I mainly program in C).

I'll take a look and improve my PR. Thank you for the suggestion.

src/lua/LuaGame.cpp
mwerle: I'm away for the weekend; will pick this up next week. Thank you for the suggestions.

src/lua/LuaUtils.cpp
Web-eWorks: No worries - this comment was mostly aimed at providing you with additional knowledge regarding our specific flavor of C++/Lua interop. The `pi_lua_generic_push` overload can be implemented on an as-needed basis in the future, it's definitely not a requirement for merge.

src/SaveGameManager.h
```diff
@@ -0,0 +1,64 @@
+// Copyright © 2008-2024 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+#ifndef _SAVEGAMEMANAGER_H
+#define _SAVEGAMEMANAGER_H
+
+#include "FileSystem.h"
+#include "JsonUtils.h"
```
Web-eWorks, line 8: To keep compile times down we prefer to use forward declaration wherever possible, especially around the JSON library's very template-heavy code. Unless absolutely necessary, headers should include `JsonFwd.h` and their matching cpp files include `JsonUtils.h`.

Web-eWorks (changes requested):
I've found only one actual change that should be done before merge, and it's a minor nitpick-y compilation time related thing. Good work! I'm exceptionally happy to see an `#include "Pi.h"` be removed from the codebase :tada:

src/lua/LuaUtils.cpp
mwerle: No I'd very much like to use the overloaded "push pull" approach, just not sure how as the other implementations are primitive types. Is it ok to just add a new table to the lua state for compound types in the push-pull overloads?

src/SaveGameManager.h
mwerle: Fixed in 977cfc5aff4d61a352fbe153dbc42d8c4a619fa3

src/lua/LuaUtils.cpp
mwerle: Fixed (?) in e3fca64b76ea7b51d5f18b7e63a2c7d6cf486a55

src/lua/LuaGame.cpp
mwerle: Fixed (?) in e3fca64b76ea7b51d5f18b7e63a2c7d6cf486a55

mwerle:
Fixed all requested changes. If accepted, I will squash the fixup commits and force-push ready for merge.

Web-eWorks (approved):
Looks good to me - if you've given it a cursory test to ensure functionality isn't affected, I'd say this is ready for merge.

Something you might consider for a follow-on PR - `data/pigui/modules/saveloadgame.lua` calls `Game.SaveGameStats` repeatedly to load details about savefiles on disk. That function *should\** be able to be moved off-thread using the Job system (finalizing the job and delivering the result via LuaEvent on the main thread) so there isn't a perceptual hitch/stutter when enumerating details about saved games.

That's the equivalent of throwing you in the deep end to see if you can swim, so feel free to demur and I'll tackle it when I get time.
* = I've not rigorously analyzed it to ensure it's free of dependencies against the main thread - that would be part of the work involved in moving it to an asynchronous job.

src/lua/LuaUtils.cpp
Web-eWorks: Didn't get a chance to answer this last night - the API contract of `pi_lua_generic_push` expects that a single value is left on the top of the Lua stack after the function has finished. That value can be a primitive type, or a compound type like a table or userdata - the important thing is just to not leave any "extra values" on the stack or the calling code's assumptions will break.

mwerle: ... eh, no idea if what I did does that; as I said, I have literally zero idea (well, maybe 0.1% now) of how Lua and C(++) interact.  That said, initial testing appears to show things working as expected following this rework..

mwerle:
> Looks good to me - if you've given it a cursory test to ensure functionality isn't affected, I'd say this is ready for merge.

Squashed the fixup commits. Thanks for your time reviewing this.

----

> Something you might consider for a follow-on PR

Created #5929 to track this suggestion.

Web-eWorks:
> Thanks for your time reviewing this.

Thank *you* for the implementation!

Issue #5319, opened 2022-01-10
UI roadmap (2nd pass)


Web-eWorks:
This issue collects future roadmap goals for the UI beyond the initial implementation or port to PiGui. All developers are welcome to add roadmap goals based on mockups or planned work, but please discuss them on IRC first!

See [the 1st pass project](https://github.com/pioneerspacesim/pioneer/projects/1?fullscreen=true)
[Mockup album.](http://imgur.com/a/6fqrd)
[Old mockups.](https://goo.gl/photos/usbRXT9TECbCPeNi8)

## World View UI
- [ ] Make ship information available as subwindow?
- [ ] Add more information to planetary situation display (bottom right)
- [ ] Comms window for chatting with other ships

#### FlightUI (reticle)
- [ ] Hide option? - (a user asked for it, so it won't be on top of the ship in external view)
- [ ] Follow-target / set speed mode selector
- [ ] Planetary flight / pitch ladder mode?
- [ ] Simplified version for in-cockpit 3d display?
- [ ] Add button/options for autopilot orientation holds

#### Comms Window
- [ ] Display message received timestamps
- [ ] New unread messages count

## Info Screen
- [ ] Convert to tabbed windows in WorldView?
- [ ] Format for use as 3d UI panels in ship cockpits?

#### Mission Display
- [ ] Orderable, filterable list
- [ ] A way to cancel missions

## Station Screen
- [ ] Mass and TWR data available on all screens

#### Bulletin Board System
- [ ] Ordering and filtering
- [ ] Active missions list?
- [ ] Cabin data for passenger missions?
- [ ] Display chat interface in BBS UI instead of popup

#### Commodity Market
- [ ] Ordering and filtering
- [ ] Cargo hold list?
- [ ] Commodity mass-per-unit display

#### Ship Market
- [ ] Compare new ship to current ship

#### Equipment Market
- [ ] Equipment details view (power, heat, etc...)
- [ ] Store purchased / ship equipment at station
- [ ] Equipment loadout presets?
- [ ] Move repair functionality to equipment screen?

## Sector Map
- [ ] Unify layout with [mockups](https://imgur.com/a/YuU79)
- [ ] Alternate star / label coloring (trade price, system economy, etc.)

## System Map
- [ ] Add "local situation" display of current body with planet heightmap, surface bookmarks, range to target, etc
- [ ] Brachistocrone calculator? (deltaV calculator for point-to-point planetary transfer)
[...]

PR #5932, opened 2024-10-07
feat: fix completion display of scan manager

### LINKED ISSUE #5930: Orbital Scan missions always require 100% completion even if contract requires less
### Observed behaviour
* Accept orbital scan mission
* Fly to target body and start scan
* Scan continues past target percentage and won't complete until 100% is reached

Orbit Scan Completion

### Expected behaviour
* Scan completes once target percentage is reached

### Steps to reproduce
As per "Observed behaviour"

### Analysis
The display of the orbital scanner shows the completion rate of the current target, not the percentage of the body scanned.

This is very confusing for a player.

This patch fixes the issue (not yet tested for surface scans, which should either continue to use the 0-100% completion, or display the distance covered so far):

**Patch 1 - fix orbital scans to show scan percentage, not completion percentage**

<details>

```diff
diff --git a/data/modules/Scout/ScanDisplay.lua b/data/modules/Scout/ScanDisplay.lua
index fc40c29ff..d01096aa9 100644
--- a/data/modules/Scout/ScanDisplay.lua
+++ b/data/modules/Scout/ScanDisplay.lua
@@ -195,12 +195,17 @@ function scanDisplay:drawScanInfo(scan, isHighlighted)
                target = ui.Format.Distance(scan.targetCoverage * 1000.0, "%.1f")
        end
 
-       local completion = math.min(1.0, scan.coverage / scan.targetCoverage)
+       local completion
+       if scan.orbital then
+               completion = math.floor(scan.coverage * 1000) / 10
+       else
+               completion = math.min(1.0, scan.coverage / scan.targetCoverage) * 100.0
+       end
 
        local data = {
                title = sBody.name .. ", " .. scan.bodyPath:GetStarSystem().name,
                target = target,
-               completion = string.format("%2.1f%%", completion * 100.0),
+               completion = string.format("%2.1f%%", completion),
                isActive = self.scanMgr:GetActiveScan() == scan,
                scan = scan,
                icon = scan.orbital and icons.map or icons.scanner,
```

</details>

Orbit Scan Completion - fixed

**Patch 2 - fix surface scans to show kilometers scanned instead of completion percentage**

<details>

```diff
diff --git a/data/modules/Scout/ScanDisplay.lua b/data/modules/Scout/ScanDisplay.lua
index fc40c29ff..3b40d270e 100644
--- a/data/modules/Scout/ScanDisplay.lua
+++ b/data/modules/Scout/ScanDisplay.lua
@@ -183,6 +183,7 @@ function scanDisplay:drawScanInfo(scan, isHighlighted)
        -- displayed when the active scanner cannot carry out the scan
        local altitude = string.upper(ls.INVALID_SCANNER)
        local target = ""
+       local completion = ""
 
        local params = scanMgr:GetScanParameters(sBody, scan.minResolution, scan.orbital)
        if params and params.canScan then
@@ -191,16 +192,16 @@ function scanDisplay:drawScanInfo(scan, isHighlighted)
 
        if scan.orbital then
                target = string.format("%.1f%%", scan.targetCoverage * 100.0)
+               completion = string.format("%2.1f%%", math.floor(scan.coverage * 1000) / 10)
        else
                target = ui.Format.Distance(scan.targetCoverage * 1000.0, "%.1f")
+               completion = ui.Format.Distance(scan.coverage * 1000.0, "%.1f")
        end
 
-       local completion = math.min(1.0, scan.coverage / scan.targetCoverage)
-
        local data = {
                title = sBody.name .. ", " .. scan.bodyPath:GetStarSystem().name,
                target = target,
-               completion = string.format("%2.1f%%", completion * 100.0),
+               completion = completion,
                isActive = self.scanMgr:GetActiveScan() == scan,
                scan = scan,
                icon = scan.orbital and icons.map or icons.scanner,
```

</details>

Orbit Scan Completion - fixed 2

Pioneer version: "trunk"
OS : Debian GNU/Linux 12 "bookworm"

[Test_scan.zip](https://github.com/user-attachments/files/17280109/Test_scan.zip)

mwerle:
## Bug link(s)
Fixes #5930 

### Background

The scan manager shows the completion percentage of the current scan target while performing a scan. This is confusing for players when performing orbital scan missions which require less than 100% of the planet to be scanned.

For example, the mission might request that 42% of the target planet needs to be scanned, but the completion percentage always goes from 0..100% - confusing the player when it goes past 42%.

For surface scan missions it was ok since the scan target is in kilometres, so having a 0..100% completion is understandable.

### Changes this PR provides

This PR proposes to modify both displays to instead of showing a completion percentage, showing the actual amount of square kilometres scanned so far. In addition, a progress bar showing the completion percentage is shown for the currently active scan.

Additionally:

* Fix for Scout missions completing at the first station the player docks at.
* Improvements to the way Scout missions are displayed in the players' mission log.

----

## Images

<details>

### Previously:

Orbit Scan Completion

### Now:

image

</details>


mwerle:
As an aside, I'm not sure how the scan resolution is useful to the player. For my first scanning mission I assumed it was the target altitude - somewhat scary skimming the surface of Mercury at 73m while in orbit (3km/s)..

It would also be really useful if the orbit planner could show some details of the orbit, such as aphelion and perihelion distances. Or even full orbital parameters, which could then tie into advanced scanning missions requiring the player to get into a particular orbit similar to KSP satellite missions (although that might be a bit technical for most players.. as-is according to some comments it's already too difficult if the scanning isn't in one of the "low"/"medium"/"high" autopilot orbits...)  That said, personally I'd love to see the orbit planner to be closer to the KSP one, and once a bunch of manoeuvre burns are defined the autopilot would simply execute those.

Web-eWorks (approved):
I think your Previously and Now images are flipped.

This is a reasonable change, although the one impact this has is that it can be much more difficult to see if the scan progress is going up on very large bodies.

As much as I personally think a 0...100% progress indicator is "clear", apparently it's unclear to users so I won't object.

> As an aside, I'm not sure how the scan resolution is useful to the player.

I agree, though I was the one to originally implement it there's definitely some missing information for the player to be able to intuit "this affects how high I can fly". I don't have strong ideas for what to "do" with it at current though.

> It would also be really useful if the orbit planner could[...]

These are all loosely planned. I have a WIP branch implementing an orbital situation display for the WorldView so you can see your current Ap/Pe, and adding full orbital details to the System View is definitely desired. We're currently tracking UI mockups in #5319 - the Transfer/Orbit Planner heading (and "Orbital map ideas" under the Other Mockups section) contain some ideas for how to present that information to the player.

Someone willing to deal with the mess that is the autopilot code and implement an `AICmdExecuteManeuver()` method would be a huge boon towards making that a reality - I tend to stay busy with the systemic side of things and haven't gotten enough time to tackle the math required.

mwerle:
> I think your Previously and Now images are flipped.
> 
Eh, I just copied the image links from the bug report instead of uploading the images again, and forgot that the first two images were to demonstrate the issue..  removed the incorrect image.

> This is a reasonable change, although the one impact this has is that it can be much more difficult to see if the orbital scan progress is going up on very large bodies with an e.g. 20% coverage scan contract.
> 
It took forever to go up a percentage point before as well.. but unless we increase the number of decimals I don't think there's any way to improve on this.

Also, given this is all floats there's some rounding issues - even with these changes, it can look as if the scan has completed, but it still has a little way to go. 

For example:
```
Target: 0.43746  -> Display: 43.7%
Current: 0.43713 -> Display: 43.7%
```

One way to fix this would be to clamp the target distance/percentage to a single decimal point.  Ideally I'd prefer to rework all of this to just use integers, but Lua is untyped (or rather, you can't specify a type, values have types as determined by Lua) as far as I've been able to work out.

> As much as I personally think a 0...100% progress indicator is "clear", apparently it's unclear to users so I won't object.
> 
It's clear enough for the surface scan, but it was (IMHO) very confusing for the orbit scan, and I had found an online discussion _somewhere_ (despite lots of Googling and going back through browser history I was unable to find this again - there's a possibility I hallucinated it even though I am fairly certain I'm not an AI..) with a person raising exactly the same issue. I found it while trying to find an online guide on how to use the scanners..

The "MINIMUM ALTITUDE" was completely unclear to me.. I thought it meant I was still too high.

Perhaps instead of displaying the scanning status in the title, add  additional icons to the scanner display?  Altitude low/optimal/high/out-of-range? (With out-of-range used for both too low as well as too high).

Now that I've played with it more I've seen more information in-game such as the minimum altitude in the scanner description, so this may not be a big deal for most people, but I did struggle to intuit how to use the scanner.

> > As an aside, I'm not sure how the scan resolution is useful to the player.
> 
> I agree, though I was the one to originally implement it there's definitely some missing information for the player to be able to intuit "this affects how high I can fly". I don't have strong ideas for what to "do" with it at current though.
> 
I think it would make it too simple to simply list minimum and maximum altitudes; it would be nice if there was a way for players to "learn" the optimal altitude for a scanner and the desired resolution. Or perhaps add the current resolution to the display and have it change based on the altitude, that way a player can home in on the optimal altitude.

Does the speed of scanning change based on altitude?  Lower orbits are faster, but higher orbits can allow for a wider area to be scanned per orbit.. but the debug display seems to indicate that the number of required orbits is fixed for a given body/scan resolution/completion percentage.

> > It would also be really useful if the orbit planner could[...]
> 
> These are all loosely planned. 

Yayy :)

mwerle:
... re-reading (and a few coffees later), I now understand what you meant with it taking longer to show progress.  Yes, that is a valid point.

Instead of showing a completion percentage as a number, how about having a bar fill up to show the completion? (This would require a bit more work). 

**Mockup:**


Scan Progress Bar Mockup

impaktor:
> I found it while trying to find an online guide on how to use the scanners

Ideally, the wiki should be the place. E.g. there's some on the [FAQ](https://wiki.pioneerspacesim.net/wiki/FAQ#Map_survey_.2F_scan_mission_does_not_progress) or [mission types](https://wiki.pioneerspacesim.net/wiki/Mission_Types_and_BBS) page

> It took forever to go up a percentage point before as well.. but unless we increase the number of decimals I don't think there's any way to improve on this.

From a game perspective, player would expect to see their action (flying within correct parameters) to have a visible effect in some way so they don't have to wonder if they're doing it right or not, so wold be nice if this was improved.

mwerle:
Got the progress bar working :

image


Yayy/nayy?

impaktor:
I guess it comes down to what the percentage should represent: surface scanned, or mission completion? What makes the player think the percentage would be anything other than total mission progress? Does the mission contractor say "scan X % of the planet"? I can't remember that (I think I wrote/edited those mission texts). If they did, I'd opt to change in that end.

Why is the Mars, Sol mission at 1384%?

mwerle:
> I guess it comes down to what the percentage should represent: surface scanned, or mission completion? What makes the player think the percentage would be anything other than total mission progress? Does the mission contractor say "scan X % of the planet"? I can't remember that (I think I wrote/edited those mission texts). If they did, I'd opt to change in that end.
> 

Yes, the mission-giver states that the mission requires, for example, "13.8%" of the target planet to be scanned. Which is also what the requirements show (bottom-left of the scan card).

 Up to now, the completion indicator (top-right of the card) always went from 0..100% - which I found rather confusing when it went past the "13.8%" of the target. 


> Why is the Mars, Sol mission at 1384%?
Because I stuffed up creating a scan mission.. the debug screen was (I'm guessing) originally designed for surface scans, so it doesn't have a percentage slider, only a kilometers one, and 1km == 100%, not 1%..

mwerle:
Mission details screen:

Screenshot_20241007_210156

Zireael07:
Another idea:

> Yes, the mission-giver states that the mission requires, for example, "13.8%" of the target planet to be scanned. Which is also what the requirements show (bottom-left of the scan card).

> Up to now, the completion indicator (top-right of the card) always went from 0..100% - which I found rather confusing when it went past the "13.8%" of the target.

Have a thick line across the progress bar at 13.8% and/or change the color of the bar after 13.8%? Show a big tick in the top right after 13.8%?

mwerle:
> Another idea:
> 
> > Yes, the mission-giver states that the mission requires, for example, "13.8%" of the target planet to be scanned. Which is also what the requirements show (bottom-left of the scan card).
> 
> > Up to now, the completion indicator (top-right of the card) always went from 0..100% - which I found rather confusing when it went past the "13.8%" of the target.
> 

So there's the initial issue I had which was that the progress percentage (top-right) went from 0..100%, even though the mission requires only 13.8%. I found this very confusing.  Which is why I changed it so the percentage only goes from 0..13.8%.

@Web-eWorks then pointed out that this means that the percentage display potentially ticks up much slower than before, which is why I added the progress bar. The progress bar "displays" 0..100%, but does so graphically, so there is no confusion between percentages. So the progress bar will (should) tick up as quickly as the original display did.

Note that this all only affects the orbital scan missions.  

The surface scan missions (which I hadn't done and which, I think, came first) require a set distance (area?) to be covered, so in this case, the original percentage display 0..100% was fine - there was nothing to confuse anybody as the target progress was in km, not in %.

> Have a thick line across the progress bar at 13.8% and/or change the color of the bar after 13.8%? Show a big tick in the top right after 13.8%?

... why? the progress bar goes from 0..100% _of the_ 13.8%. Adding the 13.8% to the progress bar would reintroduce the original confusion. Why would the mission not complete once 13.8% is reached and instead require to continue scanning all the way to 100%?

mwerle:
Ok, we may be talking at cross-purposes (it's never easy with text).

My idea is to remove any sort of confusion caused by displaying multiple percentages, one being the target scan percentage, and the other the mission progress percentage.  Having the former display 13.8%, but the latter display 0..100% was (for me at least) very confusing the first time I tried an orbital scan mission. It was so confusing that I spent several hours digging into the code to try to understand it.

Now, having changed the progress percentage to count from 0..13.8 _instead_ of 0..100, to make it match the target scan percentage, worked fine.  The problem with this is that it counts up slower than before, and players might think nothing is actually happening and the mission isn't triggering or what have you.

So following on from my initial "fix" I now propose to add a mission progress bar WITHOUT display any sort of percentage to avoid the initial confusion in order to provide faster feedback to the player that something is happening and that the scan is progressing.

----

> Hence my idea to keep 0-100% of the planet surface

This has got nothing to do with **planet surface scans**. I am talking about **planet orbital scans**.

Surface scans don't require a percentage completion, they require a distance completion.. and for those the 0..100% display made sense and is intuitive and there are no arguments or quibbles with that.

----

If people think the original display is clearer and more intuitive I will close this PR. I now understand how the scanning missions work. I still think it is incredibly unintuitive for first timers which is why I proposed these changes.

mwerle:
> It's totally not clear to the player (who doesn't know the code and the backstory) that it's 0-100 of 13.8%. Hence my idea to keep 0-100% of the planet surface (which is simple and intuitively understandable) and add a line or a tick at the place the mission requires

If your proposal is for the progress bar to render 0..100, but to _complete_ at 13.8 (sorry, that's not how I read it - english is not my first language), then we have the same problem as simply displaying 0..13.8 as a percentage in the top-right : nothing actually happens for ages since it will potentially take a long time for the first bit of progress to start displaying.

----

I guess instead of "Scan Progress" the progress bar could instead say "Mission Progress".  Would that improve it?

data/modules/Scout/ScanDisplay.lua
```diff
@@ -224,7 +224,7 @@ function scanDisplay:drawScanProgress(scan)
 	-- suitable across any style changes in the future.
 	local progressBarColor = ui.getStyleColor("SliderGrabActive")
 	ui.withStyleColors({ ["PlotHistogram"] = progressBarColor }, function()
-		ui.progressBar(completion, Vector2(width, 0), "Scan Progress")
+		ui.progressBar(completion, Vector2(width, 0), ls.MISSION_PROGRESS)
```
Web-eWorks, line 227: I recommend displaying something like "Data Collected:" instead of "Mission Progress:" - it's a little bit clearer what's going on, and we don't have to change it later if a mission wants more than one body to be scanned at once (or e.g. an orbital scan and a surface scan of a specific area).

Web-eWorks (changes requested):

data/modules/Scout/ScanDisplay.lua
```diff
@@ -191,27 +193,41 @@ function scanDisplay:drawScanInfo(scan, isHighlighted)
...
 end
 
+---@param scan ScanData
+function scanDisplay:drawScanProgress(scan)
+	local completion = math.min(1.0, scan.coverage / scan.targetCoverage)
+	local width = ui.getContentRegion().x
+	-- The style progress bar colour is yellow which is very jarring for this
+	-- display. So instead lets use the slider grab colour which should remain
+	-- suitable across any style changes in the future.
+	local progressBarColor = ui.getStyleColor("SliderGrabActive")
```
Web-eWorks, line 225: Note: this should use `colors.SliderGrabActive` or (for contrast, given this is displaying a white label overtop) `colors.uiPrimary` or `colors.uiPrimaryLight`. `ui.getStyleColor()` is an old hack from before we had a consistent UI style, and explicitly referencing the style color allows easier/better refactoring.

Web-eWorks (changes requested):
I am happy with the result of the scan progress bar. It is significantly more intuitive, and for those who want to compare apples to apples the actual scan card provides that data in an easy-to-read manner.

I do have a suggestion, in that you could include the percentage to scan completion on the progress bar as "Data Collected: 43.7%", as 0.01% is likely a smaller metric than the width of a pixel on the progress bar.

However, feel free to ignore this specific suggestion if it winds up devolving back into "which percentage do I care about anyways?"

Please address the `getStyleColor()` usage and optionally change the string used for the progress bar, and I consider this PR to be approved.

data/modules/Scout/ScanDisplay.lua
mwerle: I think there'd need to be some rework required regardless if we end up with missions which require multiple scans.

I've played with the different percentages again in the progress bar - showing the "coverage" has the same issue as the original percentage display, although we can alleviate it a little bit by increasing the number of decimals. Showing the "0..100%" is, imho, reverting back to being confusing.

Finally, as the label, "Data Collected" has no other reference; how about "Coverage" for now - this would be directly related to the tooltip "Target Coverage". But it would likely need to be reworked once there are missions with multiple steps.

**Data Collected with coverage percentage**
image

**Data Collected with progressbar percentage**
image

**Coverage with coverage percentage**
image

**Coverage Progress with no percent**
image

**Coverage tooltip**
image

----

What would be nice is an animated progress bar which animates only while the scan is active and data is being collected. That would be real and immediate feedback to the user and is something they should be familiar with. That would require significant work though as we'd need to modify the `ImGUI::ProgressBar` implementation.

data/modules/Scout/ScanDisplay.lua
mwerle: Fixed in 0174ee9f587352af9af65266a7abdd1372db0d1d

I chose `colors.uiPrimaryLight` just to give it a tiny bit of difference from the standard background. I do prefer the even lighter `SliderGrabActive` but it does reduce contrast, and ImGui currently does not have support for rendering progress bar text with inverted colors when being covered by the progress.

mwerle: Incidentally, there's a new version of ImGui which fixes the alignment of the text in the progress bar.

The current version of ImGui places the text into the non-progress portion of the bar which makes the text move as the progress bar fills, which looks a bit weird. The new version has a fixed place for the text inside the progress bar.

What's the procedure for updating ImGui? (I'm guessing a fair bit of testing would be required..)

data/modules/Scout/ScanDisplay.lua
Web-eWorks: "Data" is referred to by almost all of the mission texts / BBS adverts. I think it meets the minimum threshold to not be confusing to the player (your scan generates data -> the mission giver wants scan data).

If you think "Data Collected: XX.X%" is confusing, then please just use the label "Data Collected" for the progress bar - personally I think it's clear enough that this is a separate element from the actual scan parameters which are displayed on the card. Adding an additional `ui.spacing()` before the scan progress bar might prevent confusion as well.

data/modules/Scout/ScanDisplay.lua
Web-eWorks: Hah... funny you ask. I've had "Update ImGui" on my TODO list for the last ~week. The short and snarky form of the procedure is "get @Web-eWorks to do it".

The actual procedure is a bit longer than that - we have several patches applied onto the base ImGui codebase, which we have stored in-tree as `contrib/imgui/patches/*.patch`. Those have to be evaluated, applied, and potentially rewritten. Then, after that's done, a pass on API breaking changes from the ImGui side has to be done, and the Lua API checked to see if it uses any of the now-deprecated functions from the ImGui API.

If Lua does use a now-removed API, all callsites have to be migrated to either a drop-in wrapper in `data/pigui/libs/wrappers.lua` or the replacement API.

Most of the testing required from an ImGui upgrade is handled by "does this compile" - ocornut is very good about documenting breaking changes and keeping APIs stable. The fun bit is exposing all of the new ImGui API functions to Lua in a way that makes sense - that is typically an iterative process that lasts until (and often past) the next ImGui upgrade.

data/modules/Scout/ScanDisplay.lua
mwerle: Ok; I was thinking in the context of the scan, not the overall mission/game.

Based on the mockups above, which percentage do you think should be used here - the percentage of the progress-bar (0..100), or the percentage of the "target coverage" (0..[target coverage])?

(And thank you for your feedback!)

data/modules/Scout/ScanDisplay.lua
mwerle: Oof; a bit more involved than I had anticipated.. them's the breaks when you maintain your own patches!

data/modules/Scout/ScanDisplay.lua
mwerle: Here's the latest version :

Scan - new detail display

If approved, I'll squash the various commits.

Web-eWorks: Progress bar should display progress bar completion, 0..100%. Top right of the active scan card should always display current coverage (in percent or km), reaching the target coverage value at Data Collected: 100%. (i.e. revert 975d7923db574e34fb94ff222247fea1720f319f).

Web-eWorks: Thank you for proposing multiple different implementations and sticking with the PR! My goal here is to converge on the solution that provides the most relevant information to the player. I understand everyone has their own opinion and don't want anyone to feel like their ideas are discarded, but as the reviewer I'm trying to avoid a "too many cooks" situation.

mwerle: > Progress bar should display progress bar completion, 0..100%. Top right of the active scan card should always display current coverage (in percent or km), reaching the target coverage value at Data Collected: 100%. (i.e. revert [975d792](https://github.com/pioneerspacesim/pioneer/commit/975d7923db574e34fb94ff222247fea1720f319f)).

:+1: 

The original idea for adding the progress bar was to _replace_ the top-right percentage to reduce clutter, I had just not gotten around to that in order to first see whether or not the progress bar was feasible and useful.

> Thank you for proposing multiple different implementations and sticking with the PR! My goal here is to converge on the solution that provides the most relevant information to the player. I understand everyone has their own opinion and don't want anyone to feel like their ideas are discarded, but as the reviewer I'm trying to avoid a "too many cooks" situation.

Not at all! Thank you so much for your feedback. I'm worried I'm taking up too much of you guys' time with reviewing my silly little PR's..

mwerle: image

Gliese852:
Forgive me for interfering with my 5 cents, maybe it’s possible to convert “surface percentage” into other units, into square kilometers, for example?

mwerle:
> Forgive me for interfering with my 5 cents, maybe it’s possible to convert “surface percentage” into other units, into square kilometers, for example?

There is no "surface percentage" anywhere.. the surface scans are in kilometers (I wasn't sure if they should be square kilometers..)

Gliese852:
> > Forgive me for interfering with my 5 cents, maybe it’s possible to convert “surface percentage” into other units, into square kilometers, for example?
> 
> There is no "surface percentage" anywhere.. the surface scans are in kilometers (I wasn't sure if they should be square kilometers..)

Hm, but what do you scan during an orbital scan?  Orbit?

data/modules/Scout/ScanDisplay.lua
Zireael07: Let's give this a spin and then maybe remove the top right % in a follow up?

Gliese852:
> I'm fairly sure this could be implemented.. area of a sphere is easy enough to calculate. This would remove all the percentages except for the completion percentage. But this assumes the game can access the size of the target body when generating the mission, potentially in a system far far away..

Oh well, this is the least of the problems, you are doing `GetSystemBody`, [it's all there](https://codedoc.pioneerspacesim.net/#CClass:SystemBody).

data/modules/Scout/ScanDisplay.lua
mwerle: > Let's give this a spin and then maybe remove the top right % in a follow up?

I had removed it (see screenshot above with all the red annotations) but @Web-eWorks requested it be added again...

mwerle:
> Oh well, this is the least of the problems, you are doing `GetSystemBody`, [it's all there](https://codedoc.pioneerspacesim.net/#CClass:SystemBody).

Ok no worries; based on some other code comments and tinkering with the System Map it appeared as if not all information was available when not in-system..

What do others think about this change?

impaktor:
> What do others think about this change?

My headspace is more in fixing the "upstream" issue of why give / show the player the percentage of the body to scan in the first place. I assume orbit scan does the same as surface, in that it measures distance traveled along surface (but from high altitude). The information player might be interested in is if it's a large or small "scan job", but I also assume the difference is pretty negligible for orbit missions? For surface missions you have to do actual flying and constantly keep an eye on the altitude. For orbit scanner, I assume it's just smooth gliding for some time, so might not matter much to the player how many percent of the planet body is to be scanned? 

Either way, I'm sure what ever solution you converge to will be an improvement over what we have now.

Zireael07:
Changing to area might be a good idea (you no longer get the two % that mean different things), but I think it would be better left to a follow-up PR

mwerle:
> For orbit scanner, I assume it's just smooth gliding for some time, so might not matter much to the player how many percent of the planet body is to be scanned?

It is, but it can take many in-game days which, given the low time acceleration near planets, can translate to hours of real-time.  Not sure if the 1000x time acceleration kicks in within range of the scanner - but reducing the amount of real-time would increase the amount of time in-game (higher orbits take longer) and the potential to miss deadlines.

So a higher percentage means more time, but the actual amount of time required also depends on the size of the body. So it is probably a bit difficult for someone to gauge.

Converting the display to area instead of percentage might provide players with a better chance to estimate the amount of time required..  I'm assuming that scanning a particular amount of area should take roughly the same amount of time regardless of the size of the body. So it would be similar to the surface scan which also is a simple distance measure.

Alternatively the debug window provides the number of orbits required to achieve the desired coverage.  This is currently not exposed to the player.  This would probably also be fairly straightforward for a player to estimate to complete the contract as long as they check the size of the body - larger planets will take longer to orbit.  

----

I'll see if I can come up with some implementations..  (lol, talk about feature-creep!)

mwerle:
So I've done some research and playing around..  changing the *surface* scans to km² is easily doable and probably should be done - doesn't make sense for it to be a distance. The calculations in the game are already in km², it's only the way they're being displayed that's distance rather than area for some reason. Adding Area units to the formatter was pretty trivial.

*Orbital* scans, on the other hand, are a different kettle of fish. Again, the progress is originally calculated in km², but then converted to a percentage of the area of the target body.. and that makes sense when you start realising the sizes involved. Earth's surface, for example, is 510 Mm² (5.1 * 10^14 m², or 510,000,000 km²)..  Titan is 83 Mm². Phobos is 5695 km², so the range is also huge.

In an orbit at the minimum altitude (650km) on Earth, the XKM-650 scanner sweeps out an area of about 20,000 km² per second (according to the in-game calculations). This may seem a lot, but it would still take 12½ years just to scan Australia! (Of course, at higher altitudes it would go a lot quicker..)

But it seems that our in-game scanners aren't that great compared to real Earth Observation satellites which are around 500km altitude and can cover the entire Earth every couple of days.

Although I may have made some serious errors in my maths..

Gliese852:
I think large areas should not scare us - we don’t know exactly what resolution there is. Obviously, the orbital scan is very comprehensive and small-scale.

mwerle:
> I think large areas should not scare us - we don’t know exactly what resolution there is. Obviously, the orbital scan is very survey and small-scale.

The Scout mission generator so far has mostly given me orbital survey missions which take a looong time to complete (surface scan missions are generally quick), requiring multiple orbits at least, and often many orbits. So the orbital scans are not so small-scale. But they are nifty because they force players to get into orbits.. (although I guess most will simply click on the autopilot buttons..)

The resolution of the scans are very much a thing and calculated internally, just not exposed to the Player. This is what effectively limits the maximum height above the terrain at which the scan is active.

----

Large areas don't scare me - but I don't know how to display them nicely in the rather limited amount of screen real estate available.

- 10,000 m² = 0.01 km²
- 1,000,000 m² = 1 km²
- 10,000 km² = 0.01 Mm²
- 1,000,000 km² = 1 Mm²

I guess we could switch between m² to km² and then to Mm² earlier?  But most people have no concept of Mm².. (I didn't before looking into it yesterday - [Wikipaedia](https://en.wikipedia.org/wiki/Orders_of_magnitude_(area)) gives a really good overview).

-> 0 - 100 m² display with decimals (0.12m², 99.37 m²)
-> 100 - 9,999 m² display as m² with no decimals (237 m², 8,312 m²)
-> 10,000m² to 100km²  : display km² with decimals (0.02 km², 12.94 km²)
-> etc.

mwerle:
Ok, I've tried this and I think it works..  might still require tweaking the mission generator though.  I already reduced it by a factor of 10, but it still generates pretty long missions. I don't think I really affected the generator apart from converting the percentage to area and didn't playtest the original implementation enough to see how it compares, but it did take a long time as well.

These range from 1.6Mm² to 8.1Mm2.

<details>

```
Info: Generated orbital scan mission, coverage=0.020064%, body_coverage=0.019422%, body_coverage_km2=1627868.236900km2
Info: Generated orbital scan mission, coverage=0.040726%, body_coverage=0.015963%, body_coverage_km2=8160559.381900km2
Info: Generated orbital scan mission, coverage=0.046700%, body_coverage=0.008052%, body_coverage_km2=44040562.392800km2
Info: Generated orbital scan mission, coverage=0.039873%, body_coverage=0.001236%, body_coverage_km2=101028920.272300km2
Info: Generated orbital scan mission, coverage=0.012820%, body_coverage=0.004096%, body_coverage_km2=4721460.098400km2
```

</details>

For reference, a 100% scan of Dione (3.97Mm²) in a "high orbit" (1.57Mm) took 8.5 days game-time and 12 minutes real-time at a time-acceleration of 1000x

I did not have enough time to perform the scan as well as fly to/from Mars to Dione to complete the mission in time, and that was with a full hold of fuel.


The display is nice though:

<details>

Below 100km²
Screenshot_20241010_110636

Below 10,000km²
Screenshot_20241010_110755

From 10,000km² onwards (0.01Mm²)
Screenshot_20241010_110821

</details>

mwerle:
Rebased on "master" and pushed my current WiP.

Also changing PR back to "Draft" since it's no longer "ready to merge".

mwerle:
... bug (feature)? Scan missions complete as soon as you dock (at least in the same system). But the mission description says the player needs to return to the mission-giver.

What is actually intended? In-system I guess it's plausible since the data can simply be transmitted.. but then why not also have an in-system BBS for missions instead of per-station?

Web-eWorks:
Alright, getting a chance to respond now.

> Ok, I've tried this and I think it works.. might still require tweaking the mission generator though.

I like that, it's quite clear. Note that this should probably use the `ui.Format.Number` utility, to handle thousands-separators.

(Addendum: `ui.Format.Number` could also be updated to use a thousands-separator character loaded from the `core` resource, rather than always using a comma. It should also likely be moved to C++ for performance reasons. Neither of these are things to tackle in this PR.)

> I don't think I really affected the generator apart from converting the percentage to area and didn't playtest the original implementation enough to see how it compares, but it did take a long time as well.

Mission difficulty and duration is a bit of an open question. When originally implementing the mission generators, their real-time cost was likely the least-balanced and least-tested aspect - I wouldn't worry too much about changing that other than to consider the reward per player time cost metric.

Note that the game's timestep limits can be overridden by ctrl-clicking the time acceleration buttons, although that's generally considered a "debug" thing like other hidden ctrl- or alt-click interactions.

> ... bug (feature)? Scan missions complete as soon as you dock (at least in the same system). But the mission description says the player needs to return to the mission-giver.

I think some missions are technically supposed to be able to complete if handed in at any station (i.e. you're speaking to a representative of an organization with offices on multiple planets). However, that definitely looks like a bug!

> EDIT: Earth, 15.29Mm² took only 4.5 hours in-game (time acceleration 100x, ~3min real time) - medium orbit 3.97Mm. Increasing the orbit to 6.5Mm reduced the time to 4h in-game. Scan generated about 1060km²/s

2-3 minutes real-time at optimal orbit height (and without overriding the game's timestep limits) is a good balance for completing a generated scan mission, I think.

Given that the pay stacks up somewhere between an assassination mission and a passenger taxi mission, is free of external threats, and only requires the player to autopilot into orbit, burn prograde until the (to-be-implemented) orbital status display reports the correct apoapsis / periapsis, and just wait, some amount of player time cost is to be expected.

Mission difficulty for scan missions should ideally be adjusted in a follow-on PR however.

On a tangent, I'd like to add some more player interactivity to scan missions - you'll note some TODO list items in `ScanManager.lua` involving scanning a specific area (requiring the player to manually enter an orbit, for example) or perhaps having some sort of analysis-related gameplay loop while waiting in orbit (e.g. you're scanning for a specific signal type and need to fine-tune a "filter chain" to get valid data).

> Rebased on "master" and pushed my current WiP.
> Also changing PR back to "Draft" since it's no longer "ready to merge".

This PR is beginning to get quite long on the discussion side of things. From an official standpoint, I'd like to ask that any "additional features" be deferred to another PR and the focus be on finishing up the proposed features already implemented here.

For future "feature proposals" with similar scope of potential discussion, I'd recommend floating the idea in IRC first to gauge interest/pushback and/or opening a draft PR (full implementation not required!) as an open forum for discussion. Typically we ask for feature *requests* to be directed to the [forum](https://forum.pioneerspacesim.net/), but as it would be a proposal made with an intention to implement it Github is fine.

My hope in this is to provide a clearer differentiation between users and contributors (including myself) providing feedback or suggestions regarding a potential feature, versus the duty of the maintainer/reviewer to accept or reject a specific concrete change or implementation.

Web-eWorks:
Case in point, that was quite a wall of text. I'll do a formal code review once you think the PR is ready (should not require significant additional changes, if I'm reading your comment correctly).

Thank you again for your efforts on this PR! It's quite a small change, but the effect is very positive.

mwerle:
> This PR is beginning to get quite long on the discussion side of things

Indeed.. 😅   it started out relatively quick and simple I thought..  will discuss in forums first in future.

----

> Note that this should probably use the ui.Format.Number utility, to handle thousands-separators.

I did extend the `ui.Format` class to add `Area` formatting. None of the other "dimensional" formatters (distance, mass, etc) which I used as a template implement thousands-separators either..  just saw `Volume` does.

This would be slightly complicated because for Area I variably use 0 or 2 decimal places depending on when the different units kick in. But solvable if you think it's important enough (it would aid readability at the cost of taking up more space)

----

> I think some missions are technically supposed to be able to complete if handed in at any station (i.e. you're speaking to a representative of an organization with offices on multiple planets). However, that definitely looks like a bug!

I don't think any of the "Scout" missions currently have that capability, but that would be a nice addition for a future PR. Ditto with my thought about simply being able to transmit the scan results back to the person if they are in the same system.

----

I did see the various TODO's but those are well out-of-scope for this PR.  Some might require some enhanced UI support first anyway (such as needing to perform the scan from a particular orbit would require being able to see the orbit parameters)

----

> I'll do a formal code review once you think the PR is ready (should not require significant additional changes, if I'm reading your comment correctly).

I've thought this before ;)   Will wait for the other participants to chime in to see what they think before cleaning up the code and marking it ready-to-review again.

Thank you as always for your comments!

Web-eWorks:
> will discuss in forums first in future.

Do note that the forums are ...quite asynchronous in nature. They're good for long-range discussion but discussion surrounding active implementation effort tends to happen here or on IRC (with the balance depending on the day). We normally consign feature requests there as an "idea repository".

data/modules/Scout/ScanManager.lua
```diff
@@ -100,7 +100,7 @@ end
 
 -- Update different scan types at different rates
 local SURFACE_SCAN_UPDATE_RATE = 1
-local ORBITAL_SCAN_UPDATE_RATE = 60
+local ORBITAL_SCAN_UPDATE_RATE = 1
```
Web-eWorks, line 103: Note that this is a game-time duration of a scan "tick". At a typical timewarp rate of 100x, this means the tick function is attempting to fire every 1/100th of a second (and being unable to do so, is instead executing every single frame).

mwerle:
As there were no further comments/suggestions/concerns, I have now cleaned up the PR, rebased on master, and marked it "Ready for Review". I've also updated the description to reflect the current solution.

Finally, this PR includes a fix to prevent Scout missions from completing at the first port a player docks at after finishing a scan, as well as improving (imho) the way missions are displayed after being accepted in the players' mission log. The full description is now always shown, as well as the return location (if known). It now also integrates with the "Set Return Location" navigation button from the mission manager.

data/modules/Scout/Scout.lua
```diff
@@ -303,24 +303,30 @@ local function calcOrbitalScanMission(sBody, difficulty, reward)
...
 		resolution = resolution * (1.0 + math.log(radiusKm / p.nominal_radius, 10))
 	end
 
-	coverage = math.min(coverage * realDifficulty, 1.0)
+	-- adjust coverage against difficulty and clamp to a maximum of 100%
+	body_coverage = math.min(body_coverage * realDifficulty, 1.0)
+
+	-- finally, convert the coverage into km2 and limit the number of decimals
+	local body_area = 4 * math.pi * radiusKm^2
+	local coverageKm2 = math.floor(body_area * body_coverage * 10000) / 10000
```
Web-eWorks, line 322: Note that we have `utils.round(val, unit_last_place)`, in this case `utils.round(body_area * body_coverage, 0.0001)`.

data/modules/Scout/Scout.lua
```diff
@@ -365,8 +371,11 @@ local function calcSurfaceScanMission(sBody, difficulty, reward)
 		+ p.reward_interesting
 		* bodyReward
 
+	-- finally, convert the coverage into km2 and limit the number of decimals
+	local coverageKm2 = math.floor(coverage * 10000) / 10000
```
Web-eWorks, line 375: Similarly, `utils.round(coverage, 0.0001)`.

data/modules/Scout/Scout.lua
```diff
@@ -658,7 +667,7 @@ local onShipDocked = function (player, station)
 			mission.status = "FAILED"
 		end
 
-		if station.path == mission.location and mission.status == "FAILED" or mission.status == "COMPLETED" then
+		if station.path == mission.location and (mission.status == "FAILED" or mission.status == "COMPLETED") then
```
Web-eWorks, line 670: To double-check - should this be `mission.location` or `mission.station` now? I see you've removed an overwrite of `mission.location` in the diff above.

data/pigui/views/game.lua
```diff
@@ -309,6 +309,8 @@ ui.registerHandler('game', function(delta_t)
 			debugReload(gameView.modules)
 			debugReload(gameView.hudModules)
 			debugReload(gameView.sidebarModules)
+
+			Event.Queue("onDebugReload");
```
Web-eWorks, line 313: Is this used somewhere? Note that our current model is meant to have a "selective" hot-reload where only the active module (for some definition of active module) is reloaded rather than "reloading the world".

For context: we're moving towards a model where individual files/modules are automatically hot-reloaded on change, and modules can opt-in to replace their copies of the module when notified of a submodule having been reloaded.

Web-eWorks (changes requested):
Just a few things to look at here - I'll give this a full playtest as soon as I'm available to do so.

data/modules/Scout/ScanManager.lua
mwerle: Leaving it at the original 60 meant that the display is updating very slowly - and part of the various changes was ensuring that the player would get more immediate feedback as to progress.

Is the timewarp rate available? If so we could scale the update rate according to timewarp..

data/modules/Scout/Scout.lua
mwerle: Fixed in 778e5a38e2cf9c784a046d38e6b55c47bfb3e5f6

data/modules/Scout/Scout.lua
mwerle: Ditto; fixed in 778e5a38e2cf9c784a046d38e6b55c47bfb3e5f6

data/modules/Scout/Scout.lua
mwerle: Good catch, thank you. I tested that the various UI displays worked before and after completing the scanning, but I didn't test handing in the mission (it was 2am, I'd just gotten back from a 3-day motorbiking weekend a few hours earlier..)

data/pigui/views/game.lua
mwerle: Ah, it's a leftover from before I added the loop to call the various gauges to reload themselves. It was used in the ScanGauge as it was the only gauge (that I know of) that was "missing" after a reload.  I didn't even know it existed for ages since it'd always been killed by me reloading the scripts - I only found it when searching through the code and then tried to figure out why it wasn't being displayed.

data/modules/Scout/Scout.lua
mwerle: Fixed (and tested) in 8d6092a78ec91a503274e0e6c5f12395a6a45cf1

Tested both landing at an incorrect station after completing a scan, as well as at the correct station, following setting a new destination on scan completion.

data/pigui/views/game.lua
mwerle: Removed in 61645c227c7814ce1634452cb69d0f545d97227a

data/modules/Scout/ScanManager.lua
mwerle: .. so to answer myself, yes, there is. However there's no event that's emitted when time acceleration is changed. Is this simply because there hasn't been a need, or is there a reason?

Web-eWorks: Hasn't been a need historically. I'm strongly thinking of extending the LuaTimer API to support queuing a "timewarp invariant" timer (which stops ticking when the game is paused, of course). This would solve this (and other!) issues with timers-under-timewarp.

Web-eWorks (changes requested):
I will potentially have more changes to suggest upon the morrow after a proper playtest - this is where I got to tonight.

data/modules/Scout/ScanManager.lua
mwerle: .. so what's the way forward here, should I revert this change (at the cost of significantly delayed UI updates at lower timwarps) or should we hold this PR until the "timewarp-invariant-timer" is implemented? Or leave it as-is and add a TODO comment to fix this?

impaktor: > or should we hold this PR until the "timewarp-invariant-timer" is implemented?

My general thought (not code-review): as long as this PR is a net improvement, we should be wary to stall it / feature creep it further, as it might end up bit-rotting for years.

Web-eWorks: Leave the change as is, potentially adding a TODO comment to it. I'll look at adding a timewarp-invariant timer API in a follow-on PR, but the performance cost of scanning is not a pressing concern worth delaying the PR over.

data/modules/Scout/ScanDisplay.lua
```diff
@@ -183,35 +183,47 @@ function scanDisplay:drawScanInfo(scan, isHighlighted)
...
 
-	if scan.orbital then
-		target = string.format("%.1f%%", scan.targetCoverage * 100.0)
-	else
-		target = ui.Format.Distance(scan.targetCoverage * 1000.0, "%.1f")
-	end
-
-	local completion = math.min(1.0, scan.coverage / scan.targetCoverage)
+	target = ui.Format.Area(scan.targetCoverage * 1e6)
+	completion = ui.Format.Area(scan.coverage * 1e6)
```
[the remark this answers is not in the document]
mwerle: Sure; I left it this way to reduce the diff. `completion` can be removed entirely, but `target` is used multiple times.

And just a habit of mine to have all variables declared at the top of a scope - makes it (IMHO) easier to see where they all are and what a function might be using instead of when they are interspersed throughout.

data/modules/Scout/ScanDisplay.lua
```diff
@@ -183,35 +183,47 @@ function scanDisplay:drawScanInfo(scan, isHighlighted)
...
+function scanDisplay:drawScanProgress(scan)
+	local completion = math.min(1.0, scan.coverage / scan.targetCoverage)
+	local width = ui.getContentRegion().x
+	-- The default progress bar colour style is yellow which is very jarring
+	-- for this display. So instead lets use a more suitable colour style.
+	local progressBarColor = colors.uiPrimaryLight
+	local progressBarText = ls.DATA_COLLECTED_PROGRESS % {
+		percent_completed = string.format("%.2f", completion * 100.0)
+	}
+	ui.withStyleColors({ ["PlotHistogram"] = progressBarColor }, function()
```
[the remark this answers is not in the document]
mwerle: Copy-pasted from `widgets.lua:143`.. 

In general I prefer to have code consistent, even if it's more verbose. Inexperienced programmers such as myself have no idea about such subtleties :)

I'd be tempted to leave this as-is but will change if the "cleaned up" version is / should be the coding style for this project.

data/modules/Scout/ScanGauge.lua
```diff
@@ -24,5 +27,6 @@ gauges.registerGauge(10, {
 	end,
 	unit = '%', format = '%.2f', min = 0, max = 100,
 	icon = icons.scanner, color = colors.gaugeScanner,
-	tooltip = ls.HUD_SCAN_PROGRESS
+	tooltip = ls.HUD_SCAN_PROGRESS,
+	debugReload = onDebugReload,
```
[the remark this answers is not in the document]
mwerle: Sure. My thinking was to implement the general case so no refactoring is needed if more functionality is added in the future. Finding the function is pretty trivial with a modern IDE (and I placed it right next to its use) but agreed, not as instant as having it as a lambda.

Fixed.  No commit ref as I'm squashing-and-rebasing to hopefully have this PR finalised :)

data/modules/Scout/ScanDisplay.lua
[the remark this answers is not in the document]
mwerle: Fixed.  No commit ref as I'm squashing-and-rebasing to hopefully have this PR finalised :)

data/modules/Scout/ScanManager.lua
mwerle: Fixed.  No commit ref as I'm squashing-and-rebasing to hopefully have this PR finalised :)

mwerle:
Rebased on main; squashed commits; addressed most comments.

Please continue review at your leisure.

Bodasey:
first test results:

- Getting the cash only when reached the station where the mission has been accepted now working.
- Surface scan works normally, percentage completion is shown correctly.
- New orbital scan missions sometimes have a very small area to cover and are completed in less than one second, a mission accepted in former master branch was also completed in two complete orbits, percentage completion did not work there.

Change suggestion: The symbols for scan missions from the bulletin board could/should also be used in the scan manager.

mwerle:
> Change suggestion: The symbols for scan missions from the bulletin board could/should also be used in the scan manager.

This is non-trivial - the BBS icons are loaded from individual icon files as a `PiImage`, and then drawn by invoking the `Draw` method. The scan display is inherited from `ItemCard` which assumes the icon is in the `ui.theme.icon`s and performs calculations to find the offset inside that file for the icon to be rendered before drawing it with `ui.icon()`. The `ui.theme.icon`s currently does not include the BBS icon images.

`ItemCard` is also used in `EquipCard` so that would also need to be modified if we pass in a `PiImage` instead of a `ui.theme.icons` defined icon.

We should probably have a unified approach for icons in the game to avoid such discrepancies - either they should all be `PiImage`s, or they should all be `ui.theme.icon`s. Having individual icon files is probably better for modders..

data/modules/Scout/ScanDisplay.lua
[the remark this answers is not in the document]
Web-eWorks: The cleaned-up version should be the default. We generally are trending towards its use everywhere (many `ui.withStyleVars` calls already use the style). The original style dates to circa 2016, well before I joined the project.

I do agree that I would like it to be consistent. Feel free to also amend the copy site in `widgets.lua` for this specific cleanup.

data/modules/Scout/ScanDisplay.lua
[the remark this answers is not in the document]
Web-eWorks: I definitely understand the desire to have clearly declared variables - in the interest of consistency however, we tend not to have separately-declared variables unless a function result needs to be assigned to a mix of local and non-local variables, or the variable is set inside of a subordinate block scope.

This is partially because the practice is a lot more resistant to inadvertently using a variable before it's been initialized, and partially just a stylistic convention of "declare variables where needed".

Web-eWorks:
I've not forgotten this, just been horribly busy over the weekend. If all goes well, will do final review in ~12hrs. Thank you again for your patience!

mwerle:
> I've not forgotten this, just been horribly busy over the weekend. If all goes well, will do final review in ~12hrs. Thank you again for your patience!

No rush..

data/modules/Scout/ScanDisplay.lua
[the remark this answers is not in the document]
mwerle: I think a clean-up commit should target all places where the old style is in use. A quick grep shows there's  about a dozen or so places using the old style.  Up to you whether we add it to this PR or leave it for a clean-up PR.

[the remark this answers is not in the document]
impaktor: > I can roll this set of commits in with this PR, or just the first one and make a new PR for the other two commits. 

My general thoughts: If it's expected we need separate discussion for something, a separate PR can be warranted. If it's just some stuff you happened to fix while working on some feature, you can include it in the same PR, as separate commits.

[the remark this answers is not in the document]
Web-eWorks: Thanks! The diff is definitely not as bad as I was expecting - it's fine to keep in this PR!

data/modules/Scout/Scout.lua
```diff
@@ -303,24 +303,30 @@ local function calcOrbitalScanMission(sBody, difficulty, reward)
...
 		resolution = resolution * (1.0 + math.log(radiusKm / p.nominal_radius, 10))
 	end
 
-	coverage = math.min(coverage * realDifficulty, 1.0)
+	-- adjust coverage against difficulty and clamp to a maximum of 100%
+	body_coverage = math.min(body_coverage * realDifficulty, 1.0)
+
+	-- finally, convert the coverage into km2 and limit the number of decimals
+	local body_area = 4 * math.pi * radiusKm^2
+	local coverageKm2 = utils.round(body_area, 0.0001)
```
Web-eWorks, line 322: This line is computing the mission's coverage as the entirety of the body's surface area, regardless of the `body_coverage` value above. Multiply `body_area` by `body_coverage` and we're good to go.

data/modules/Scout/Scout.lua
```diff
@@ -303,24 +303,30 @@ local function calcOrbitalScanMission(sBody, difficulty, reward)
...
 
 	local rewardAmount = reward
 		* p.reward_per_km * (math.pi * radiusKm * coverage)
```
Web-eWorks, line 325: Reward needs to be updated to use `math.sqrt(coverageKm2)` rather than `pi * radiusKm * coverage`.

Additionally, `reward_per_km` may need to be adjusted in the `orbital_params` definition to scale mission rewards back into line. Quick playtesting with the requested changes suggests it should be increased by up to 20% to make small moon scans viable and reward difficult low-orbit scans around large bodies.

Web-eWorks (changes requested):
I'm glad I playtested this, as I noticed that a mission to Saturn wants me to scan all 41,000 Mm² of Saturn's surface - and a mission to scan all 7.2Mm² of Oberon's surface pays a measly $620, barely enough to cover fuel for the very long trip into deep space.

Stuff like this tends to slip through the cracks on long PRs sadly - other than this I think the PR is good to go.

mwerle:
> I'm glad I playtested this, as I noticed that a mission to Saturn wants me to scan all 41,000 Mm² of Saturn's surface - and a mission to scan all 7.2Mm² of Oberon's surface pays a measly $620, barely enough to cover fuel for the very long trip into deep space.
> 
> Stuff like this tends to slip through the cracks on long PRs sadly - other than this I think the PR is good to go.

Hmm, this is really odd - I playtested this quite a bit and the only time I got full coverage was for very tiny moons.   I see the current code is wrong, as you point out, but due to all the squashing/rebasing, I've lost when it was working as intended.

.... ok, found it - I _was_ multiplying the `body_area` by `body_coverage`, and at the same time reducing the number of decimals. You then asked me to refactor to use `utils.round()` and in that update I somehow dropped the multiplication. I guess I only briefly  tested from that point onwards.

Well caught, thank you.

As for the reward amount, yes, I never adjusted it.

data/modules/Scout/Scout.lua
mwerle: Fixed in 0ba8f053b60e217f44971e4ee53d5565eff2776e

This was originally correct, but got lost in a refactor to use "utils.round()"..

data/modules/Scout/Scout.lua
mwerle: Fixed in c69f89755327d169f7256c9a47dba352263120d1

Anecdotally, the rewards seem to be significantly less now..

Web-eWorks: I'm noticing that as well. For now, let's set `reward_per_km = 0.11`, and `reward_resolution_min = 12.0` - in testing, this seems to produce adequately-lucrative results. A follow-on PR should adjust `MissionUtils.CalcReward` to take into account 1) this is a two-way trip rather than a single-hop to a station, and 2) properly calculate the distance between the closest hyperspace jump-in point and the target body in a distant system.

mwerle: The rewards for these missions (before this PR) seem to be fairly high, given there's no risk and quite often are even in the same system. As soon as a player can afford the scanners this seems to be the most lucrative mission type for starting-out players (based on my rather limited playtime so far..)

mwerle: Updated rewards in d6d1f690a0d76a1c17bae0b6496ce81c14e8e6c3

Web-eWorks: > A future PR

These are all great ideas - I would hate for them to get lost. Please feel free to add them as a comment header to `Scout.lua`, or document them on the forum at your discretion.

Web-eWorks (approved):
Making the call to accept this PR. Rewards can/will be rebalanced in a future PR, ideally post-equipment-rework once we've had some ability to look at how scan missions function and what the player's "buy in" is to accept more lucrative missions.

Thanks @mwerle for continuing to update the PR in a timely manner even when my own availability is quite spotty!

data/modules/Scout/Scout.lua
mwerle: > > A future PR
> 
> These are all great ideas - I would hate for them to get lost. Please feel free to add them as a comment header to `Scout.lua`, or document them on the forum at your discretion.

Added in b8d008bd3efb92b527a5ed552d95926996383300

mwerle:
> Thanks @mwerle for continuing to update the PR in a timely manner even when my own availability is quite spotty!

Not at all - I've been away myself. This is only a hobby after all :)

Rebased and squashed the final changes.

PR #5934, opened 2024-10-09
refactor: run SaveGameStats in a job

### LINKED ISSUE #5929: Improvement - move calls to Game.SaveGameStats into Jobs
Follow-up to #5927; details in [this discussion](https://github.com/pioneerspacesim/pioneer/pull/5927#pullrequestreview-2350528888).

----

Web-eWorks:

Something you might consider for a follow-on PR - data/pigui/modules/saveloadgame.lua calls Game.SaveGameStats repeatedly to load details about savefiles on disk. That function should* be able to be moved off-thread using the Job system (finalizing the job and delivering the result via LuaEvent on the main thread) so there isn't a perceptual hitch/stutter when enumerating details about saved games.

That's the equivalent of throwing you in the deep end to see if you can swim, so feel free to demur and I'll tackle it when I get time.

    = I've not rigorously analyzed it to ensure it's free of dependencies against the main thread - that would be part of the work involved in moving it to an asynchronous job.

mwerle:
##' Bugs / Feature requests

Attempts to fix #5929 

## Background

The idea is to run this expensive task in an asynchronous job and notify the Lua side once it is completed. This is to prevent the stutter from occurring every time the SaveGameStats API is called from Lua.

## Details

The code was modified to call the expensive part of the SaveGameStats API (loading and parsing the game file) into a Job. Once finished, the data is marshalled into a LuaTable and a LuaEvent is triggered to pass the data to the Lua side.

The Lua side registers an Event listener, which parses the data and updates the SaveGame cache entry for the data.

Practically, this is achieved by the following changes :

* add a job manager to the LuaManager class which can run any job on
  the Lua job queue.
* add a LoadGameToJsonJob implementation in SaveGameManager, and a new function, LoadGameToJsonAsync(), which schedules an instance of this job on the Lua job queue.
* call this new function from Lua, as well as registering a callback for when the job completes, in saveloadgame.lua.

Many thanks to @Web-eWorks for feedback throughout the development of this change (details of which have been squashed, but the implementation went through several iterations) as well as providing the Lua profiler and removing the no longer required `SaveLoadWindow::update()` callback.

### DIFF
--- a/data/pigui/modules/saveloadgame.lua
+++ b/data/pigui/modules/saveloadgame.lua
@@ -300,6 +300,13 @@ function SaveLoadWindow:makeFileList()
 		return a.mtime.timestamp > b.mtime.timestamp
 	end)
 
+	-- Cache details about each savefile
+	for _, file in ipairs(self.files) do
+		if not self.entryCache[file.name] or self.entryCache[file.name].timestamp ~= file.mtime.timestamp then
+			self.entryCache[file.name] = makeEntryForSave(file)
+		end
+	end
+
 	self:makeFilteredList()
 	-- profileEndScope()
 end
@@ -349,21 +356,6 @@ function SaveLoadWindow:deleteSelectedSave()
 		-- List of files changed on disk, need to update our copy of it
 		self:makeFileList()
 	end)
-end
-
-function SaveLoadWindow:update()
-	ModalWindow.update(self)
-
-	-- Incrementally update cache until all files are up to date
-	-- We don't need to manually clear the cache, as changes to the list of
-	-- files will trigger the cache to be updated
-	local uncached = utils.find_if(self.files, function(_, file)
-		return not self.entryCache[file.name] or self.entryCache[file.name].timestamp ~= file.mtime.timestamp
-	end)
-
-	if uncached then
-		self.entryCache[uncached.name] = makeEntryForSave(uncached)
-	end
 end
 
 --=============================================================================
--- a/data/pigui/modules/saveloadgame.lua
+++ b/data/pigui/modules/saveloadgame.lua
@@ -2,6 +2,7 @@
 -- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
 local Game       = require 'Game'
+local Event      = require 'Event'
 local FileSystem = require 'FileSystem'
 local Lang       = require 'Lang'
 local ShipDef    = require 'ShipDef'
@@ -197,42 +198,56 @@ end
 
 --=============================================================================
 
-local function makeEntryForSave(file)
-	local compatible, saveInfo = pcall(Game.SaveGameStats, file.name)
-	if not compatible then
-		saveInfo = {}
-	end
-
+-- Event callback once the savegame information has been loaded
+-- Updates the entryCache with the full details of the savegame.
+function SaveLoadWindow:onSaveGameStats(saveInfo)
+	-- local profileEndScope = utils.profile("SaveLoadWindow:onSaveGameStats()")
 	local location = saveInfo.system or lc.UNKNOWN
-
 	if saveInfo.docked_at then
 		location = location .. ", " .. saveInfo.docked_at
 	end
 
-	local saveEntry = SaveGameEntry:clone({
-		name = file.name,
-		compatible = compatible,
-		isAutosave = file.name:sub(1, 1) == "_",
-		character = saveInfo.character,
-		timestamp = file.mtime.timestamp,
-		gameTime = saveInfo.time,
-		duration = saveInfo.duration,
-		locationName = location,
-		credits = saveInfo.credits,
-		shipName = saveInfo.shipName,
-		shipHull = saveInfo.shipHull,
-	})
-
 	-- Old saves store only the name of the ship's *model* file for some dumb reason
 	-- Treat the model name as the ship id and otherwise ignore it if we have proper data
+	local shipHull
 	if not saveInfo.shipHull then
 		local shipDef = ShipDef[saveInfo.ship]
-
 		if shipDef then
-			saveEntry.shipHull = shipDef.name
+			shipHull = shipDef.name
 		end
+	else
+		shipHull = saveInfo.shipHull
 	end
 
+	local entry = self.entryCache[saveInfo.filename]
+	entry.character = saveInfo.character
+	entry.compatible = saveInfo.compatible
+	entry.credits = saveInfo.credits
+	entry.duration = saveInfo.duration
+	entry.gameTime = saveInfo.time
+	entry.locationName = location
+	entry.shipName = saveInfo.shipName
+	entry.shipHull = shipHull
+
+	-- profileEndScope()
+end
+
+local function onSaveGameStats(saveInfo)
+	ui.saveLoadWindow:onSaveGameStats(saveInfo)
+end
+
+-- Trigger load of savegame information and return bare-bones entry
+local function makeEntryForSave(file)
+	-- local profileEndScope = utils.profile("makeEntryForSave()")
+	Game.SaveGameStats(file.name)
+
+	local saveEntry = SaveGameEntry:clone({
+		name = file.name,
+		isAutosave = file.name:sub(1, 1) == "_",
+		timestamp = file.mtime.timestamp,
+	})
+
+	-- profileEndScope()
 	return saveEntry
 end
 
@@ -243,6 +258,7 @@ end
 --=============================================================================
 
 function SaveLoadWindow:makeFilteredList()
+	-- local profileEndScope = utils.profile("SaveLoadWindow::makeFilteredList()")
 	local shouldShow = function(f)
 		if not self.showAutosaves and f.name:sub(1, 1) == "_" then
 			return false
@@ -266,9 +282,11 @@ function SaveLoadWindow:makeFilteredList()
 	if not utils.contains_if(self.filteredFiles, isSelectedFile) then
 		self.selectedFile = nil
 	end
+	-- profileEndScope()
 end
 
 function SaveLoadWindow:makeFileList()
+	-- local profileEndScope = utils.profile("SaveLoadWindow::makeFileList()")
 	local ok, files = pcall(Game.ListSaves)
 
 	if not ok then
@@ -283,6 +301,7 @@ function SaveLoadWindow:makeFileList()
 	end)
 
 	self:makeFilteredList()
+	-- profileEndScope()
 end
 
 function SaveLoadWindow:loadSelectedSave()
@@ -551,6 +570,8 @@ function SaveLoadWindow:render()
 end
 
 --=============================================================================
+
+Event.Register("onSaveGameStats", onSaveGameStats)
 
 ui.saveLoadWindow = SaveLoadWindow
 
--- a/src/lua/LuaGame.cpp
+++ b/src/lua/LuaGame.cpp
@@ -9,6 +9,7 @@
 #include "GameSaveError.h"
 #include "Json.h"
 #include "Lang.h"
+#include "LuaEvent.h"
 #include "LuaObject.h"
 #include "LuaTable.h"
 #include "LuaUtils.h"
@@ -73,10 +74,63 @@ static int l_game_start_game(lua_State *l)
 	return 0;
 }
 
+/* Marshall the game info into a LuaTable and send it as an event to Lua.
+ * Callback from the savegame_stats job when the job is finished. This function
+ * will run in the main thread. */
+static void onSaveGameStatsJobFinished(std::string_view filename, const Json &rootNode)
+{
+	PROFILE_SCOPED()
+	auto ls = Lua::manager->GetLuaState();
+	ScopedTable t(ls);
+	t.Set("filename", filename);
+
+	try {
+		t.Set("time", rootNode["time"].get<double>());
+
+		// if this is a newer saved game, show the embedded info
+		if (rootNode["game_info"].is_object()) {
+			const Json &gameInfo = rootNode["game_info"];
+			t.Set("compatible", true);
+			t.Set("system", gameInfo["system"].get<std::string>());
+			t.Set("ship", gameInfo["ship"].get<std::string>());
+			t.Set("credits", gameInfo["credits"].get<float>());
+			t.Set("flight_state", gameInfo["flight_state"].get<std::string>());
+			if (gameInfo.count("docked_at")) {
+				t.Set("docked_at", gameInfo["docked_at"].get<std::string>());
+			}
+			if (gameInfo.count("shipHull")) {
+				t.Set("shipHull", gameInfo["shipHull"].get<std::string>());
+			}
+			if (gameInfo.count("shipName")) {
+				t.Set("shipName", gameInfo["shipName"].get<std::string>());
+			}
+			if (gameInfo.count("duration")) {
+				t.Set("duration", gameInfo["duration"].get<double>());
+			}
+			if (gameInfo.count("character")) {
+				t.Set("character", gameInfo["character"].get<std::string>());
+			}
+		} else {
+			// this is an older saved game...try to show something useful
+			const Json &shipNode = rootNode["space"]["bodies"][rootNode["player"].get<int>() - 1];
+			t.Set("frame", rootNode["space"]["bodies"][shipNode["body"]["index_for_frame"].get<int>() - 1]["body"]["label"].get<std::string>());
+			t.Set("ship", shipNode["model_body"]["model_name"].get<std::string>());
+			t.Set("compatible", false);
+		}
+	} catch (const Json::type_error &) {
+		t.Set("compatible", false);
+	} catch (const Json::out_of_range &) {
+		t.Set("compatible", false);
+	}
+
+	LuaEvent::Queue("onSaveGameStats", LuaTable(t));
+	LuaEvent::Emit();
+}
+
 /*
  * Function: SaveGameStats
  *
- * Return stats about a game.
+ * Start a Job to read the SaveGameStats for a particular save game
  *
  * > Game.SaveGameStats(filename)
  *
@@ -89,6 +143,12 @@ static int l_game_start_game(lua_State *l)
  *
  *   2018-02-10
  *
+ * Modified:
+ *
+ *   2024-10-11 - the function no longer returns the Stats directly, but starts
+ *                a Job. The Lua caller is responsible for registering an
+ *                "onSaveGameStats" event handler to process the data.
+ *
  * Status:
  *
  *   experimental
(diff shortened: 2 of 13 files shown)


Web-eWorks (commented):
Initial feedback - instead of adding the new LuaEvent class to piecewise construct an argument table, construct a LuaTable object inside of `l_game_savegame_stats_callback`, add values to it as normal, and pass that LuaTable handle as the args parameter to `LuaEvent::Queue`. (Don't forget to dispose of the underlying table from the stack - a LuaTable doesn't auto-pop its table.)

If that doesn't currently compile, let me know and I'll do the template metaprogrammery to get it to work.

A note - we tend to reserve the `l_<namespace>_<method>` naming convention for functions following the `Lua_Function` API prototype, i.e. `static int l_name(lua_State *l)`. This is just a "regular" callback and can have any particular name.

Regarding profiling, we have an in-process profiler for C++ which can be used via the `PROFILE_SCOPED()` macro (with the appropriate include of `profiler/Profiler.h`). Details on usage are in the [dev docs](https://dev.pioneerspacesim.net/contribute/profiling).

src/lua/LuaGame.cpp
```diff
@@ -93,10 +137,50 @@ static int l_game_start_game(lua_State *l)
  *
  *   experimental
  */
+static void l_game_savegame_stats_callback(std::string filename, Json rootNode)
+{
+	LuaEvent::LuaEvent e("onSaveGameStats");
+	e.addParameter("filename", filename);
+	try {
+		Json gameInfo = rootNode["game_info"];
```
Web-eWorks, line 145: Should be `const Json &` here - currently this is a copy-assign which recursively copies the whole node.

Web-eWorks:
Thanks for working on this! We unfortunately don't have good (or any) Lua-side profiling utilities at current, but I'll pull the branch and take a closer look to see what's responsible for the stutters.

Web-eWorks:
Hacked in a manual lua-side profiler to test performance. Note that printing to the console has an overhead of about `0.065ms` - I've disabled most of the print statements for the sake of measuring real performance.

<details>

With hot caches:

```
Info: PROFILE | SaveLoadWindow:makeFilteredList() took 0.0085ms
Info: PROFILE | SaveLoadWindow:makeFileList() took 0.1471ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0033ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0036ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0038ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0040ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0039ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0050ms
Info: error in JSON file 'savefiles/_autosave9': [json.exception.parse_error.110] parse error at 1: unexpected end of input
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0049ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0038ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0029ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0041ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0054ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0034ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0037ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0034ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0041ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0113ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0042ms
```

With fully cold cache (`echo 3 | sudo tee /proc/sys/vm/drop_caches`):
```
Info: PROFILE | SaveLoadWindow:makeFilteredList() took 0.0092ms
Info: PROFILE | SaveLoadWindow:makeFileList() took 5.8037ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0030ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0033ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0043ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0036ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0028ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0048ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0069ms
Info: error in JSON file 'savefiles/_autosave9': [json.exception.parse_error.110] parse error at 1: unexpected end of input
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0035ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0060ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0035ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0037ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0035ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0027ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0031ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0045ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0039ms
Info: PROFILE | SaveLoadWindow:onSaveGameStats() took 0.0042ms
```

Looking at the profile from that run (`Engine.RequestProfileFrame()`), we're almost entirely disk bound inside of `SaveGameManager::ListSaves`.
image

</details>

With the manual rate-limiting removed from `SaveLoadWindow:update()`, loading information about saves is almost instant as all saves are processed at once. I'm not seeing the major stutter you mentioned either - even with cold caches loading the list of saves from disk (with an N of 18) causes at most a single frame skip at 60Hz.

This does raise the follow-on question of also moving `Game.ListSaves()` to an off-thread operation - as I imagine a user with a decent number of savefiles (say, 150) in combination with a legacy HDD or eMMC boot drive could potentially have an observable stutter in that function lasting a significant fraction of a second.

Would you like me to push my profiler implementation to this branch?

src/lua/LuaGame.cpp
mwerle: Yes, my initial version is "dumb" because I don't know C++ very well. Or rather, I don't know how to predict object lifetimes very well, so didn't want things going out of scope and blowing up for now.

I'm old-skool, give me full control over object lifecycles. And yes, I know C++ does that, but it obfuscates it.. which is fine in a small or well-known codebase, but when coming in to a new codebase, until learning in detail all the various object behaviours, it's difficult to understand just when things get constructed and destroyed. I prefer explicit "malloc" and "free".

mwerle:
Thank you for having a look and providing feedback!

> Initial feedback - instead of adding the new LuaEvent class to piecewise construct an argument table, construct a LuaTable object inside of `l_game_savegame_stats_callback`, add values to it as normal, and pass that LuaTable handle as the args parameter to `LuaEvent::Queue`.
> 
I'm afraid I still have extremely limited knowledge of how the Lua stack works and what can and can't be constructed on it and passed around. It's not really made any clearer (to me) by the wrappers you guys have developed which have various restrictions such as the ScopedTable. If there is a book or other training material you can point me at so I can learn more about it that would be appreciated.

So I didn't think a LuaTable could be constructed on the stack and then passed around and modified etc before invoking a Lua function while other parts of the code also potentially modify the Lua stack (ie, another Job in another thread).

Inside `l_game_savegame_stats_callback` I don't have a `lua_State` which can be used to create a table.

In the "normal" functions the `lua_State` comes directly as a parameter, but I didn't know if that state would survive being passed into another thread and used/modified there while other Lua functions run in other threads. I kinda figured the Lua stuff would probably all need to be on the same thread without adding a lot of synchronisation.

This is why I made that LuaEvent class to construct the table and manage the `lua_State`. But looking at it now, it probably boils down to the same thing.. as I said, C++ isn't my forte, and the whole Lua thing is brand-new to me.

>  (Don't forget to dispose of the underlying table from the stack - a LuaTable doesn't auto-pop its table.)

???

> A note - we tend to reserve the `l_<namespace>_<method>` naming convention for functions following the `Lua_Function` API prototype, i.e. `static int l_name(lua_State *l)`. This is just a "regular" callback and can have any particular name.
> 
Noted.

> Regarding profiling, we have an in-process profiler for C++ which can be used via the `PROFILE_SCOPED()` macro (with the appropriate include of `profiler/Profiler.h`). Details on usage are in the [dev docs](https://dev.pioneerspacesim.net/contribute/profiling).

I saw the macro, but haven't had time to look into it. Thanks for the link, I'll have a read.

mwerle:
> Would you like me to push my profiler implementation to this branch?

Yes please. No need for me to reinvent the wheel..

src/lua/LuaGame.cpp
mwerle: ... and reading the rest of your comments, it seems this class should be dropped anyway.

I guess what you're saying is that I can just pass a `lua_State`, or obtain one from within the callback, and then use that to construct the table directly? That was my main sticking point.

Web-eWorks: Yes. The `LuaEvent::Queue()` method (if not passed a specific queue) implicitly operates on the "main" lua_State from `Lua::manager->GetLuaState()` - you can make a Lua table using that state, populate it with the data that needs to be passed back, and then send that table through LuaEvent.

Pushing a `LuaTable` object as a LuaEvent parameter essentially just copies the reference to the underlying Lua table object, which itself has automatically-managed lifetime by the Lua GC.

Web-eWorks:
> I'm afraid I still have extremely limited knowledge of how the Lua stack works and what can and can't be constructed on it and passed around.

This is partially why the Lua*** wrappers were developed, to simplify interacting with Lua from the programmer's perspective. I'm sorry they're not clear and easy to use either - we (at this point, I) probably should write a primer on interacting with Lua and add it to the dev docs.

> So I didn't think a LuaTable could be constructed on the stack and then passed around and modified etc before invoking a Lua function while other parts of the code also potentially modify the Lua stack (ie, another Job in another thread).

> but I didn't know if that state would survive being passed into another thread and used/modified there while other Lua functions run in other threads.

Lua is implicitly and explicitly single-threaded only. Only the main thread is allowed to call any Lua API functions on the main Lua state (`Lua::manager->GetLuaState()`). Inside of a Job / Task's main execution method, you generally will never see any Lua API being called.

I should have mentioned this when I was originally describing the problem to you, sorry. :sweat_smile:

The `OnFinish` method of the Job class specifically **always runs on the main thread** for this reason, to allow interaction with non-threaded resources like OpenGL or Lua.

Other Lua states may be constructed for use on separate threads, but they operate under a share-nothing model in relation to the main Lua state from both a C and Lua-side perspective.

As mentioned in another reply, from the "main" thread you can  `lua_State *l = Lua::manager->GetLuaState();` to interact with the game's Lua context at (almost*) any time.

\* Startup and shutdown excluded, for obvious reasons.

> This is why I made that LuaEvent class to construct the table and manage the lua_State. But looking at it now, it probably boils down to the same thing.. as I said, C++ isn't my forte, and the whole Lua thing is brand-new to me.

Don't worry about it at all - this is why I'm taking the time to review and provide feedback in such depth on your PRs. My hope is to help smooth along the learning process; feel free to mention me (as sturnclaw) in IRC with questions if that's a more expedient way to learn.

> > (Don't forget to dispose of the underlying table from the stack - a LuaTable doesn't auto-pop its table.)
>
> ???

Here's an annotated example:

```c++
// Create a new table on the Lua stack.
// This is a shorthand to call lua_newtable(l); LuaTable(l, -1).
// Passing a stack index to the constructor instead aliases an existing table.
LuaTable tab { l };

// Interact with the table as normal, don't need to bother with the underlying Lua API.
tab.Set("myKey", myVal);

// Pass this table to a Lua event. The LuaTable wrapper aliases the underlying table on
// the Lua stack when copied.
// LuaEvent::Queue pushes all parameters to Lua and then destroys its arguments. You
// generally don't have to worry about lifetime concerns as long as you're not passing
// a ScopedTable handle.
// LuaEvent::Queue is guaranteed to leave the stack in the same state as when you called it.
LuaEvent::Queue("myEvent", tab);

// Because LuaTable aliases a table on the stack, it does not remove that value from
// the stack when it is destroyed. Instead, you have to manually call lua_pop() to remove
// the table implicitly created via LuaTable(lua_State *l).
lua_pop(l, 1);
```

data/pigui/modules/saveloadgame.lua
```diff
@@ -197,42 +198,107 @@ end
...
 	end
 
+	self.entryCache[saveInfo.filename].character = saveInfo.character
+	self.entryCache[saveInfo.filename].compatible = saveInfo.compatible
+	self.entryCache[saveInfo.filename].credits = saveInfo.credits
+	self.entryCache[saveInfo.filename].duration = saveInfo.duration
+	self.entryCache[saveInfo.filename].gameTime = saveInfo.time
+	self.entryCache[saveInfo.filename].locationName = location
+	self.entryCache[saveInfo.filename].shipName = saveInfo.shipName
+	self.entryCache[saveInfo.filename].shipHull = shipHull
```
Web-eWorks, line 249: Note: Lua cannot optimize out the repeated table lookups here. Prefer storing `self.entryCache[saveInfo.filename]` in a local variable and directly assigning to it.

src/lua/LuaEvent.h
```diff
@@ -51,7 +51,6 @@ namespace LuaEvent {
 	{
 		Queue(GetEventQueue(), event);
 	}
-
```
Web-eWorks, line 54: Errant deleted whitespace here - preferably LuaEvent isn't touched at all by this PR.

data/pigui/modules/saveloadgame.lua
```diff
@@ -197,42 +198,107 @@ end
...
+end
+
+local function onSaveGameStats(saveInfo)
+	ui.saveLoadWindow:onSaveGameStats(saveInfo)
+end
+
+-- Trigger load of savegame information and return bare-bones entry
+local function makeEntryForSave(file)
+	print('makeEntryForSave : file=', file.name)
+	pcall(Game.SaveGameStats, file.name)
```
Web-eWorks, line 259: This can be converted to `Game.SaveGameStats(file.name)` since the function no longer throws an error.

Web-eWorks (changes requested):

data/pigui/modules/saveloadgame.lua
mwerle: Heh, ok, thanks for the heads-up.  Currently working with someone who very much dislikes "unnecessary" local variables, and C just optimises this sort of stuff out.

src/lua/LuaEvent.h
mwerle: PR isn't cleaned up yet.. but thanks for pointing it out.

mwerle:
> Finally, as a suggestion - you may want to move the JobSet into a more central API like `Pi::App()` or `Lua::manager` as a shared "Lua Jobs" interface. That would avoid the need to keep a static object at file scope, and allow object deinitialization to be controlled as well as initialization.

Sure; I first wanted to see if it would work at all so just hacked it into the `LuaGame.cpp`.


I moved the Job itself into the `SaveGameManager` class and simply exposed a function to create a new job.

I moved the jobset into `Lua::manager` as suggested.

The issue is that the `Lua::Manager` implementation now has a #include "Pi.h" which I seem to remember is undesired..?

Web-eWorks:
> I moved the Job itself into the `SaveGameManager` class and simply exposed a function to create a new job.

Looks good. Naming is a little stilted, I might personally try to name it something like `LoadGameToJsonAsync` and allow the return value type to do the talking on what it returns.

> I moved the jobset into `Lua::manager` as suggested.
> 
> The issue is that the `Lua::Manager` implementation now has a #include "Pi.h" which I seem to remember is undesired..?

this is fine

More seriously, this isn't ideal. I suspect the editor will likely crash on startup because of this, as `Pi::App` doesn't exist outside of "main game" contexts. Just one of those details about the codebase you'll trip over in the process of learning it.

Preferably, the `LuaManager` constructor takes a `JobQueue *asyncJobQueue` parameter instead. As a knock-on effect, this means that `Lua::Init()` now needs to either be passed a `LuaManager` for it to take ownership of on startup, or be passed a pointer to the running application's AsyncJobQueue to then pass to LuaManager. The latter is likely much easier to implement and thus what I'd suggest.

mwerle:
> Preferably, the LuaManager constructor takes a JobQueue *asyncJobQueue parameter instead. As a knock-on effect, this means that Lua::Init() now needs to either be passed a LuaManager for it to take ownership of on startup, or be passed a pointer to the running application's AsyncJobQueue to then pass to LuaManager. The latter is likely much easier to implement and thus what I'd suggest.

This is proving to be marginally challenging as the other Applications do not have any Job Queues..

Web-eWorks:
https://github.com/pioneerspacesim/pioneer/blob/2884328dd61ef3c1fb7dadcf59b39b6413abdd1f/src/core/Application.h#L89

mwerle:
> https://github.com/pioneerspacesim/pioneer/blob/2884328dd61ef3c1fb7dadcf59b39b6413abdd1f/src/core/Application.h#L89

No idea how or why I missed this.. I was grubbing around inside the implementation files and must have missed the base class.

Proposed fix in 63c95d1e0ed1850be46b52ec4dff14c69c3db96d

data/pigui/modules/saveloadgame.lua
mwerle: Fixed in c32dd13852548473bb5ce72aabf3c54ec7571008

data/pigui/modules/saveloadgame.lua
mwerle: Fixed in c32dd13852548473bb5ce72aabf3c54ec7571008

mwerle:
> Looks good. Naming is a little stilted, I might personally try to name it something like `LoadGameToJsonAsync` and allow the return value type to do the talking on what it returns.
> 

Fixed in 1a871c89891620b3c3ee4e6674b5d4b36562004d

Web-eWorks:
> No idea how or why I missed this.. I was grubbing around inside the implementation files and must have missed the base class.
> 
> Proposed fix in [63c95d1](https://github.com/pioneerspacesim/pioneer/commit/63c95d1e0ed1850be46b52ec4dff14c69c3db96d)

Looks good. Will give this a whirl tomorrow.

data/pigui/modules/saveloadgame.lua
```diff
@@ -243,6 +258,7 @@ end
 --=============================================================================
 
 function SaveLoadWindow:makeFilteredList()
+	local profileEndScope = utils.profile("SaveLoadWindow::makeFilteredList()")
```
Web-eWorks, line 261: Leftover profiling call here.

Web-eWorks (approved):
So far this looks good! I've left one comment where a bit more cleanup is needed, and I've pushed a commit to the branch removing the manual rate-limiting code in `SaveLoadWindow:update()` (and that overridden function entirely). Feel free to rebase/squash/etc. that commit - I'm not worried about the author/committer information.

data/pigui/modules/saveloadgame.lua
mwerle: The change is impressive, even Debug builds now load the list of saves almost instantly, down from about 14 seconds, on my PC with about 40 saves.

I'll squash/merge/etc and mark the PR as ready then. Thank you for all of your input!

mwerle:
Squashed commits, rebased on main, ready for review, thank you.

Web-eWorks:
My prior approval stands. Good work, and thank you for implementing this! Hopefully it's given you a bit better visibility into a "vertical slice" of the codebase - I'll chalk that up as a win for shamelessly asking other contributors to do my TODO items for me. :smile:

In all seriousness, this is a good implementation and I'm glad you proposed the idea of the SaveGameManager - I likely wouldn't have come up with that on my own.

mwerle:
Thank you, yes it was a great learning exercise and I think I gained some insight into how the codebase, especially interactions with Lua, work.

PR #5936, opened 2024-10-09
fix: only reset view to WorldView when Player enters system

### LINKED ISSUE #5935: View is sometimes reset to WorldView if "Reset View On Hyperspace Exit" is enabled
### Observed behaviour
Enter SystemView (or any other view) at high time acceleration. Randomly the view is reset to the WorldView.

### Expected behaviour
View stays in the selected View unless hyperspacing.

### Steps to reproduce

* Enter a view (eg, SystemView)
* Turn on high time acceleration factor
* Wait
* View is reset to WorldView after a random amount of time

### Analysis / Patch

Bug was introduced in #5922

The intended switch back to the WorldView in "Player.lua" in the `onEnterSystem` event callback performs the switch without checking that the ship is the player.

This is due to the developer (me) not understanding that the event is triggered for _all_ ships, not just the player ship..

Patch:
```diff
diff --git a/data/libs/Player.lua b/data/libs/Player.lua
index 52a591bb4..ad5e8be0a 100644
--- a/data/libs/Player.lua
+++ b/data/libs/Player.lua
@@ -431,7 +431,8 @@ local onGameEnd = function ()
        -- clean up for next game:
 end
 
-local onEnterSystem = function ()
+local onEnterSystem = function (ship)
+       if not ship.IsPlayer() then return end
        -- Return to game view when we exit hyperspace
        if Engine.GetResetViewOnHyperspaceExit() and Game.CurrentView() ~= "world" then
                Game.SetView("world")
```

Pioner "master"; Debian GNU/Linux 12

n/a

mwerle:
Fixes #5935

Bug introduced in 99b0acdb9843d65fbb6f19e720c3e59a5abec86e, the 'onEnterSystem' event needs to check whether the ship which entered the system is the Player.

### DIFF
--- a/data/libs/Player.lua
+++ b/data/libs/Player.lua
@@ -431,7 +431,8 @@ local onGameEnd = function ()
 	-- clean up for next game:
 end
 
-local onEnterSystem = function ()
+local onEnterSystem = function (ship)
+	if not ship.IsPlayer() then return end
 	-- Return to game view when we exit hyperspace
 	if Engine.GetResetViewOnHyperspaceExit() and Game.CurrentView() ~= "world" then
 		Game.SetView("world")


mwerle:
> If you _want_ you can open an issue + PR pair, but if you already have the fix ready, we usually skip opening the issue. In case you want to save yourself the trouble & time of being neat.

I saw all (most?) PR's had an issue attached so I kinda figured it was required for this project.

PR #5945, opened 2024-11-01
Add Buy and Sell prices to the Commodity Market

### LINKED ISSUE #5903: Selling prices are not displayed until you actually try to sell a commodity
### Observed behaviour

In the picture below, Hydrogen's price displayed is $3.95
However, when I sell 1 unit, the price is $3.87

Though I understand that the selling price is lower than the buying price, I would like to have it displayed somewhere in order to
- know the selling price of a commodity even when I don't own any
- develop a feeling of how much I can expect to make with a certain quantity of a commodity
- improve immersion

image

### Expected behaviour

I have access to the selling price along with the buying price

### Steps to reproduce

N/A

20240710 Linux

mwerle:
Fixes #5903 

Replace the "Price" column with a "Buy Price" and "Sell Price" column. This reflects what the user of the commodity market actually needs to know, instead of delegating this information until after the user has started buying or selling a particular commodity.

Also right-aligns all numbers in the table, and middle-aligns the column headings for a much more readable table.

**Before:**

image

**After:**

image

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -85,7 +85,7 @@
   },
   "BUY_PRICE": {
     "description": "Title for buy-price column in commodity market",
-    "message": "Buy Price"
+    "message": "Buy"
   },
   "BUY_SHIP": {
     "description": "",
@@ -234,6 +234,10 @@
   "COMMANDER": {
     "description": "",
     "message": "Commander"
+  },
+  "COMMODITY": {
+    "description": "Title for commodity name column in commodity market",
+    "message": "Commodity"
   },
   "COMMODITY_MARKET": {
     "description": "",
@@ -1299,9 +1303,13 @@
     "description": "",
     "message": "In cargo hold"
   },
+ "IN_HOLD": {
+    "description": "Column title of the Commodity Market screen",
+    "message": "In hold"
+  },
   "IN_STOCK": {
-    "description": "",
-    "message": "In stock"
+    "description": "Column title of Commodity and Ship Equipment markets showing current station stock level.",
+    "message": "Stock"
   },
   "IN_SYSTEM": {
     "description": "Used in flight log, show which star system we are/were in",
@@ -2157,7 +2165,7 @@
   },
   "SELL_PRICE": {
     "description": "Title for sell-price column in commodity market",
-    "message": "Sell Price"
+    "message": "Sell"
   },
   "SENSORS": {
     "description": "Crew skill",
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -83,6 +83,10 @@
     "description": "Commodities and equipment prices",
     "message": "Buy"
   },
+  "BUY_PRICE": {
+    "description": "Title for buy-price column in commodity market",
+    "message": "Buy Price"
+  },
   "BUY_SHIP": {
     "description": "",
     "message": "Buy Ship"
@@ -2150,6 +2154,10 @@
   "SELL_EQUIPPED": {
     "description": "Button text to sell selected ship equipment",
     "message": "Sell Equipped"
+  },
+  "SELL_PRICE": {
+    "description": "Title for sell-price column in commodity market",
+    "message": "Sell Price"
   },
   "SENSORS": {
     "description": "Crew skill",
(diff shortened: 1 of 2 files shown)


impaktor:
From what I can see, I'm fine with not squashing the commits. 

As for alignment, ideally they should be aligned on the decimal point, or which ever is the closest to it based on left/right, I assume?

While you're here/"there" [in the code], I'd like to open for yet another feature creep: I'm not a fan of player not being able to sell beyond demand. Could it be possible to have the "demand" be just at that price, once it's reached 0, price is adjusted accordingly? Or just reduce price by half or 20 % or similar if demand has reached constant 0?

Granted I haven't play tested the current market on master, but seems harsh for a trader to arrive and not be able to sell their inventory, but rather to have to jettison it (or fly to another station?).

mwerle:
I think adjusting price after demand has been met is out-of-scope for this PR and liable to generate some discussion. So I'd prefer to leave that for another PR. I can create a new issue in order to track it. But is it realistic that a market would buy a commodity, even discounted, that they don't want? The risk is on the trader.

----

Aligning on decimal point would require a fair bit more work. In this case, the only decimal values are "Money", which always have 2 decimal places, so it's a moot point.

----

Looking at the new display, I'm now thinking instead of "Buy Price" and "Sell Price" to simply head the columns as "Buy" and "Sell". It should be implicitly obvious it's the price given the monetary values. Thoughts?

bszlrd:
Ommiting the Price from the header sounds good. Less clutter. Maybe even "In stock" could be just "Stock" and also "Cargo" could be changed to either "Hold" on In hold".

mwerle:
image

Web-eWorks:
This generally looks good. The screen needs to be fully rewritten to use the new ImGui Tables feature, as well as move away from the hardcoded magic-constants design, but for an interim change this works well enough.

A few thoughts:
- The commodity name field is a bit excessively wide. Is this to accommodate e.g. the German translation?
- Since the dividing line between the two fields is no longer in the center, it should be at either 62% (golden ratio) or at 2/3rds across the screen.
- For examples regarding how "modern" UI screens should be structured, look at `saveloadwindow.lua` or the equipment outfitter UI on the `equip-v2` branch. Those are the most up-to-date complete rewrites of a UI screen, and demonstrate the programming patterns I would like to see adopted across the entire UI.

mwerle:
I can take a look at rewriting it, thanks for the examples of how it should be written, but no idea when I'll get around to a larger-scale rework. I thought I was picking some low-hanging fruit ;)

Changing the ratio to 62%, let alone 2/3, reduces the right-side size even further, which causes the button-bar to look very squashed. Given this will be rewritten "properly" "soon" anyway, can we leave this PR as-is for now?

Web-eWorks:
> I can take a look at rewriting it [...] I thought I was picking some low-hanging fruit ;)

For clarity, I was mentioning it in the abstract rather than suggesting you should do it yourself - feel free to do so if you're interested though!

Displaying buy/sell price separately is definitely a low-hanging fruit and I'm glad you've tackled it.

> Given this will be rewritten "properly" "soon" anyway

You're speaking my language now! :smile:

> can we leave this PR as-is for now?

That's fine. I don't know when exactly we'll want to fully redesign the screen, but likely not before the next release.

mwerle:
TODO: the BBS (black) markets need a similar update.. this will be a bit more challenging as they open in a smaller popup window.

Ideally the code also should be refactored so all commodity market displays use the same code.. ie, make a "commodity market" card whose input is the list of commodities and their prices to display, which can then render itself within any parent window.

mwerle:
Ok I thought I had introduced the issue on the BBS markets that resizes the window to full-height as soon as a commodity item is selected and the buy/sell part of the interface is populated, but the issue already exists in the current code.

I have tried to fix this but cannot find the reason why the window expands as soon as a commodity item is clicked on. This may need deferring to a future PR when the entire screen is re-written using the new UI paradigm.

----

Before clicking on Commodity - window takes up partial screen:
image


After clicking on Commodity - window is expanded to take up entire vertical space and "Hang Up" button is not visible (must be scrolled):
image

Web-eWorks:
I think the commodity market resizing out of control in a BBS ad is a known issue - I recall trying to fix it once before, but that either never made it to a PR or I never got as far as an actual fix.

Go ahead and defer it to another PR - I'll review this code as-is soon.

PR #5946, opened 2024-11-03
Fix builds which use "-DDEBUG"

mwerle:
Thanks to @WickedSmoke (IRC) for pointing out that the debug builds were broken when "-DDEBUG" is specified as part of the build.

Specifically, "DEBUG" enables the "LUA_DEBUG" symbols, and these no longer compile in the current codebase. This patch fixes the compilation errors, but, notably, now cause runtime problems such as:

```
Pioneer/src/lua/LuaVector.cpp:240: lua stack difference is 1, expected 0
```

This patch also replaces all instances of checking "DEBUG" with checks against "NDEBUG" instead, as the latter is part of the C/C++ language standards and thus independant of any particular build system.

### DIFF
--- a/src/GeoSphere.h
+++ b/src/GeoSphere.h
@@ -38,7 +38,7 @@ public:
 	virtual double GetHeight(const vector3d &p) const override final
 	{
 		const double h = m_terrain->GetHeight(p);
-#ifdef DEBUG
+#ifndef NDEBUG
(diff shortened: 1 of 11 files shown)


Web-eWorks (approved):
This looks good, and I think we've caught most if not all of the runtime errors associated with failed checks.

Thanks for tackling this @mwerle and thanks WickedSmoke for reporting it in the first place!

PR #5948, opened 2024-11-03
Improvement: Radar auto-zoom on cargo cannisters

mwerle:
Similar to auto-zooming on ships, add the capability for the radar to auto-zoom on cargo containers. This is very useful when scooping multiple cargo containers in an area, as otherwise the radar is liable to zoom out to max-zoom-range in-between scooping containers. This makes it very difficult to lock on to the next container.

TODO: 

The radar should be switchable between auto and manual modes, with the manual mode being controllable by the keyboard/joystick. 

A radar mode and range indicator might also be useful to add.

In future, different radar equipment could provide different ranges and/or capabilities.

### DIFF
--- a/data/pigui/modules/radar.lua
+++ b/data/pigui/modules/radar.lua
@@ -122,6 +122,7 @@ local function display3DRadar(center, size)
 	local navTarget = player:GetNavTarget()
 	local maxBodyDist = 0.0
 	local maxShipDist = 0.0
+	local maxCargoDist = 0.0
 
 	radar.size = size
 	radar.zoom = currentZoomDist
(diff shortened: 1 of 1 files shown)


bszlrd:
The ability to switch radar zoom modes would be most welcome. I think you could do that with the old radar on the old dashboard UI, but not sure if it ever worked. 
I'd imagine that there could be a target mode, where your target would define the zoom, and the ship would stay at the edge of the disk until it is close to being in weapon range.

Web-eWorks (approved):
Would you like to track the TODO items somewhere else or address them before merge?

I give this branch my blessing as a useful feature.

mwerle:
> Would you like to track the TODO items somewhere else or address them before merge?
> 
> I give this branch my blessing as a useful feature.

I think the TODO is going to be non-trivial (and I mostly wrote it as a braindump rather than something specifically for this PR) but i think the change as-is is already quite valuable and could be merged as an immediate QoL improvement.

Captured the TODO here (please let me know if there's a better place) :

https://github.com/mwerle/pioneer/issues/3

Web-eWorks:
That is a good enough spot - do note that you (should) have the ability to push changes to the dev docs now, to document "officially accepted" ~~pipe dreams~~ future development plans.

PR #5949, opened 2024-11-03
feat: degrade scooped cargo requiring cargo life support

mwerle:
If there is no cargo bay life-support module installed, scooping certain cargo cannisters should degrade their contents.

This PR randomly turns live animals into animal meat, fertilizer, or even rubbish. Slaves are turned into fertilizer or rubbish.

### DIFF
--- a/data/modules/LifeSupport.lua
+++ b/data/modules/LifeSupport.lua
@@ -23,6 +23,23 @@ local function LifeSupportCallback(self)
 	if commodityName then
 		cargoMgr:RemoveCommodity(Commodities[commodityName], 1)
 
+		-- randomly convert the commodity into a suitable alternative
+		-- TODO: have a more generic check instead of hard-coding the only two commodities which current require life-support.
+		local dice = 0
+		if commodityName == "live_animals" then
+			Engine.rand:Integer(0,3)
+		elseif commodityName == "slaves" then
+			Engine.rand:Integer(0,2)
+		end
+		if dice == 3 then
+			commodityName = "animal_meat"
+		elseif dice == 2 then
+			commodityName = "fertilizer"
+		else
+			commodityName = "rubbish"
+		end
+		cargoMgr:AddCommodity(Commodities[commodityName], 1)
+
 		if self == Game.player then
 			Game.AddCommsLogLine(lc.CARGO_BAY_LIFE_SUPPORT_LOST)
 		end
--- a/data/modules/LifeSupport.lua
+++ b/data/modules/LifeSupport.lua
@@ -27,9 +27,9 @@ local function LifeSupportCallback(self)
 		-- TODO: have a more generic check instead of hard-coding the only two commodities which current require life-support.
 		local dice = 0
 		if commodityName == "live_animals" then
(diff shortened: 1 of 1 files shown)


mwerle:
Quick way to test: add a bunch of "live animals" and/or "slaves" to your ship (as long as it doesn't have cargo life-support) using the debug menu and watch them slowly get composted..

Web-eWorks:
I know this is the way the original Elite/Frontier did it, but long-term I don't think we should keep this change. IMO it might be more appropriate to have a "dead animals" or "dead slaves" commodity item which they turn into, with the presentation to the user that it's a distinct state of the original cargo item, akin to something like a cargo container of `Fine Artistry [Ruined]`.

This also provides additional hooks for gameplay, perhaps the legal punishment for carrying or dumping dead animals is worse than the punishment for carrying live animals in jurisdictions which prohibit them, and the only way to get rid of the cargo container is to jettison it outside of sensor range or to deal with the station's black market.

From a technical perspective, this is fine enough in the short term, though it is quite dark. I'd question specifically if scooped "person containers" (i.e. escape pods, slaves) should degrade over time, or if they would have life support built in - in my mind, cargo bay life support is needed for transporting perishable goods like fruits, vegetables, and livestock; people are only ever stored in a cargo hold if they're in some sort of cryostasis / life-support providing conveyance.

impaktor:
I think this is a net improvement.
 No telling how many years it will be until we get "the real way to do it, with added game play".

mwerle:
The original Elite had "Slaves" carried in cryogenic pods; ie, 1 "unit" of Slaves was the Slave, plus the cryogenic machinery needed to keep them alive.

In Frontier/FFE, Slaves were presumably carried live in some sort of in-hold containment cell, since in those games, they did degrade to Fertilizer if the cargo bay life-support failed. But that also begs the question of needing to have food and water on-board to feed them - in these games a trip can take many weeks, after all. The latter also applies to live animals.
(EDIT: Having just re-read the FFE manual, the Cargo Bay Life-Support module includes food and drink dispensers.. although would have been an interesting game-mechanic to have to ensure that you keep grain/food/water on-board sufficient to keep your cargo alive for the trip..)

The fact we still have Slaves in the first place in the 33rd century is what bothers me, if anything. This is just some dark humour (which, admittedly, may no longer be socially acceptable).

----

I'm ambivalent as to how we fix this; but just removing the cargo as per the current implementation is, IMHO, the worst solution.

Web-eWorks:
With @impaktor's sign-off, I'll merge it as is, given that it is a net-improvement to the mass mysteriously vanishing from the hold.

mwerle:
Ayaah...  thank you for the merge, but there was a `fixup` commit which should have been squashed first. Now my mistake will live on in eternity! Lol.

Web-eWorks:
Oops. My bad for not looking at the commit list more closely - I normally just review the diff :smile:

mwerle:
No worries; was just a joke.  But as I (think) I mentioned before, I'm currently working with someone who is absolutely anal about "clean" commit histories..

PR #5956, opened 2024-11-05
Build Cleanup

WickedSmoke:
These commits eliminate all the compiler warnings on my box except for the following (which I have no clue how to fix):

```
In file included from /home/karl/osrc/games/pioneer/src/lua/LuaObject.h:9,
                 from /home/karl/osrc/games/pioneer/src/lua/LuaEvent.h:8,
                 from /home/karl/osrc/games/pioneer/src/Ship.cpp:28:
/home/karl/osrc/games/pioneer/src/lua/LuaPushPull.h: In instantiation of ‘std::tuple<typename std::remove_reference<Args>::type ...> pi_lua_multiple_pull(lua_State*, int) [with Types = {double, double}; lua_State = lua_State]’:
/home/karl/osrc/games/pioneer/src/lua/LuaObject.h:528:60:   required from ‘static std::tuple<Ret1, Ret2, Ret ...> LuaObject<T>::CallMethod(T*, const Key&, const Args& ...) [with Ret1 = double; Ret2 = double; Ret = {}; Key = char [19]; Args = {}; T = Ship]’
/home/karl/osrc/games/pioneer/src/Ship.cpp:660:47:   required from here
/home/karl/osrc/games/pioneer/src/lua/LuaPushPull.h:168:91: warning: operation on ‘beg’ may be undefined [-Wsequence-point]
  168 |         return std::tuple<std::remove_reference_t<Types>...>{ LuaPull<Types>(l, beg++)... };
      |                                                                                           ^
```


src/Background.cpp
```diff
@@ -340,7 +340,7 @@ namespace Background {
 						}
 
 						// Don't process any more sectors if we've generated our quota of stars.
-						if (stars.pos.size() >= starsLimit)
+						if (stars.pos.size() >= (size_t) starsLimit)
```
mwerle, line 343: I had a brief look at this a while back; is there actually a good reason that `starsLimit` is not a `size_t`? 

In general, casting indicates "code smell".

src/collider/GeomTree.cpp
```diff
@@ -133,7 +133,7 @@ GeomTree::GeomTree(const int numVerts, const int numTris, const std::vector<vect
 		PROFILE_SCOPED_DESC("GeomTree::CreateTriTree");
 
 		m_triAABBs.reset(new AABBd[m_numTris]);
-		for (size_t i = 0; i < m_numTris; i++) {
+		for (size_t i = 0; i < (size_t) m_numTris; i++) {
```
mwerle, line 136: The other `for` loops in this function use `int i` when iterating over `m_numTris` - should probably follow that precedent instead of a cast here.

In general, why is `m_numTris` (and the other counters in this class) an `int` and not a `size_t`? A really brief look indicates that none of these values can actually be negative..

src/collider/GeomTree.cpp
```diff
@@ -209,7 +209,7 @@ GeomTree::GeomTree(Serializer::Reader &rd)
 
 		// TODO: triangle AABBs should be written to the SGM file similarly to edge AABBs
 		m_triAABBs.reset(new AABBd[m_numTris]);
-		for (size_t i = 0; i < m_numTris; i++) {
+		for (size_t i = 0; i < (size_t) m_numTris; i++) {
```
mwerle, line 212: Ditto..

src/collider/GeomTree.cpp
WickedSmoke: Commit 0551cd44b1 shows what changes when the n_num* members become size_t.

mwerle: LGTM, but I don't know the codebase well-enough to see if this impacts anything else.

Thank you for your effort in helping clean this up!

mwerle: (What I mean is, it requires additional review from a member who is more familiar with the codebase..)

src/collider/GeomTree.h
```diff
@@ -70,9 +70,9 @@ class GeomTree {
 	void RayTriIntersect(int numRays, const vector3f &origin, const vector3f *dirs, int triIdx, isect_t *isects) const;
 private:
 
-	int m_numVertices;
-	int m_numEdges;
-	int m_numTris;
+	size_t m_numVertices;
+	size_t m_numEdges;
+	size_t m_numTris;
```
Web-eWorks, line 76: This should be `uint32_t` rather than `size_t` - Geom is not intended to operate on mesh data containing more than 4 billion vertices, triangles, or edges. Additionally, its upstream dependency, SingleBVHTreeBase, has a hard limit of < 2^32 triangles or edges.

Web-eWorks (changes requested):
Thanks for addressing this! We have been moving towards the use of standard sized-integer typedefs for a while, so this is a welcome change - especially moving away from signed integers in array indices.

Web-eWorks:
Note that the `-Wsequence-point` warning is an over-zealous false-positive; expressions in an initializer-list are strongly ordered and the standard guarantees that they will be executed in a defined and consistent order. This is a GCC problem rather than an "us" problem.

src/collider/GeomTree.h
WickedSmoke: The m_num* members are now uint32_t.   If starsLimit is good as size_t then I'll squash the commits.

Web-eWorks: size_t is appropriate for starsLimit - go ahead and squash.

src/collider/GeomTree.cpp
Web-eWorks: Thanks for your review @mwerle!

Web-eWorks (approved):

PR #5957, opened 2024-11-05
fix(Passengers): fix crash bugs in Passengers introduced by equip-v2

mwerle:
This PR contains the following fixes:

- fix(Passengers): fix boolean expression in Passengers:EmbarkPassenger
  The precedence of a boolean expression is invalid, with the unary 'not' operator being evaluated before the arithmetic expression. This results in a comparison between a boolean and a number, which gives a runtime error.

  This is fixed by refactoring the overall expression to remove the 'not' operator.

- avoid a nil dereference if hull can't fit any cabins in Passengers:GetMaxPassengersForHull
  When generating certain types of missions, the logic attempts to fill a ship with passenger cabins. Under the new system it's possible that a hull can't fit any. Add a sanity check to ensure we never dereference a nil cabin.

### DIFF
--- a/data/libs/Passengers.lua
+++ b/data/libs/Passengers.lua
@@ -106,7 +106,7 @@ function Passengers.EmbarkPassenger(ship, passenger, cabin)
 		return false
 	end
 
-	if not cabin:GetFreeBerths() > 0 then
+	if cabin:GetFreeBerths() == 0 then
 		return false
 	end
 
--- a/data/libs/Passengers.lua
+++ b/data/libs/Passengers.lua
@@ -177,9 +177,10 @@ function Passengers.GetMaxPassengersForHull(hull, maxMass)
 					and (availMass - equip.mass > 0)
 					and equip.capabilities.cabin or nil
 			end)
-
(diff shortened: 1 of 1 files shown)


data/libs/Passengers.lua
```diff
@@ -179,7 +179,7 @@ function Passengers.GetMaxPassengersForHull(hull, maxMass)
 			end)
 
 			numPassengers = numPassengers + passengers
-			availMass = availMass - cabin.mass
+			availMass = availMass - cabin and cabin.mass or 0
```
Web-eWorks, line 182: I'd prefer this and the previous line be wrapped in an `if cabin then ...` block.

Web-eWorks (approved):
Thanks for addressing this! Other than a minor nitpick, this looks good to me and catches a very obvious bug I should have seen earlier.

data/libs/Passengers.lua
mwerle: Fixed in c1e4556d7a1f82be98800599406cef86386ee8e9

PR #5958, opened 2024-11-06
feat(radar): add manual zoom mode to scanner

mwerle:
Fixes: https://github.com/mwerle/pioneer/issues/3

----

The radar is zoomed manually, but the scanner has automatic zoom which can lead to some issues while trying to get an overview of the current area around the ship as it is usually zoomed out too far.

This commit adds key bindings to switch between the radar and the scanner, as well as bindings to zoom the scanner and reset the mode back to automatic.

The zoom bar of the scanner is moved up to make room for the new manual/automatic zoom mode indicator on the left as well as the current range indicator which is in the same position on the right as the radar display.

To make the display less cluttered, the left and right button bars were moved away from the center slightly, they now end and start, respectively, on the edges of the radar area.

----

**TODO:**

- [x] add hover-mode for the scanner to show ship information
- [x] use keybindings instead of hardcoded keys
- [x] use icons for the automatic/manual mode indicator
  - [x] get final icon glyphs - THANK YOU @bszlrd 
- [x] get buttons actually working
- [x] (possibly?) use the new Input system to register and use keybindings
- [x] code review
- [x] clean up git history before merge

**Auto-zoom:**

image

**Manual Zoom:**
image


impaktor:
This works for 2D azimutal radar as well? (Right click the radar)

mwerle:
> This works for 2D azimutal radar as well? (Right click the radar)

image

Yes - in fact, the radar is only manual zoom. It already could be zoomed using the mouse scroll-wheel, and, now, using the zoom keys I defined. I haven't added an auto-zoom mode to it (yet.. is it desired?)

The mouse scroll-wheel has the issue that it isn't captured - the event is passed to the rest of the game engine and causes the view to zoom as well.

bszlrd:
How about a button like this? Not yet sure about the icons though, the A-M on them is kind of a cop-out, but I have to figure out a good iconography for it.
radar-01

bszlrd:
How about icons like this?

mwerle:
> How about icons like this?

Might have to see them "in the flesh", but I suspect that they will be too "busy" at the size we need.

I was thinking [AF] / [MF]  or just [A]/[M] as per common camera icons:

https://thenounproject.com/icon/manual-3549434/

https://thenounproject.com/icon/automatic-mode-3555025/

bszlrd:
I'll put them in flesh tomorrow evening, and we can see how busy they would be.

bszlrd:
Been able to sneak in some time while waiting. How about this?
radar-02

Or this? I prefer this one, with the radar mode switcher, and the AM button next to it, when it is on planar. Helps with the discoverability of the azimuthal radar. I think a lot of people don't know about it.
radar-03

mwerle:
> Been able to sneak in some time while waiting. How about this?

I quite like the first A/M icons! But they'd be a bit big I guess in combination with the switcher icon unless we distribute the icons around the scanner instead of just having them at the bottom.  Eg, have the switcher at the top-left?

Personally I think the switcher-icon is a little bit abstract (I know it's a representation of the different scanner visuals) and making the A/M icon that small .. I'd have to see it in-game I guess.

That said, I'm not a UI person; technically I'm getting close to having it work and I'll leave the iconography up to you / everybody else to vote on.

Thank you for the mockups!!

mwerle:
The mouse-wheel _already_ is used for zoom (although, as I said, it clashes with the main-viewport zoom functionality so is kindof broken at present).

I'm not a fan of adding a button for every possible function to the main display; it's already cluttered.. but ultimately, I'll leave the UI design up to other people as I have limited understanding of UI design... .  I'm more interested in the actual functionality.

So whatever you think looks and works best  :)

mwerle:
Changed the "radar" to match the "scanner" colours. (Sorry, "azimuthal" is just not something which rolls off the tongue or which the average gamer can be expected to understand, imho)

image

mwerle:
I've been tweaking the implementation a little bit and managed to figure out how to use bound keys in the Lua. Still need to decide on a good set of defaults (or even none).

One final issue I can't crack - actually getting the buttons to work.
For some reason, the buttons don't react to the mouse at all, nor do the tooltips show up. If I put the entire radar into a ui.window() (as per other display elements), then the 3d radar doesn't get rendered nor do the buttons (although I may have stuffed up some coordinate calculations).

But experimenting with creating a separate small window with a button inside the radar's draw routine had a properly-working button with tooltip and reacting to the mouse.

Once I get the buttons working (help?), and some button glyphs from @bszlrd I think this PR is getting close to being ready for review.

bszlrd:
I'll get to it, I think I will be able to tonight

mwerle:
> I'll get to it, I think I will be able to tonight

Thank you! But no rush :)   I still can't get the bleeding buttons to actually work...

bszlrd:
Mabye the scanner overlaps and blocks them?

mwerle:
Here's the latest WiP screenshot:

image

mwerle:
> Mabye the scanner overlaps and blocks them?

The text-label for the zoom works properly with a tooltip.. and it (was) way more overlapped than the buttons.  I have no idea..

mwerle:
> Here are the icons:
Thanks! You should add the commit to the PR directly in order to be accredited for your work :)

Actually the A/M icons would be more useful right now - I'm thinking that unless someone can help me fix the toggle button, it's best not to display that at all. It's kindof pointless unless it works, whereas the A/M indicator is important feedback and, as it's just an indicator, does not need functionality (although it would be nice to have a tooltip at least..).

I can also revert to just using text directly if you think it's better than using icons.

mwerle:
Hmm.. they seem a bit small and difficult to make out given the button size. And I have a pretty big monitor.

image

image



What do you think?

bszlrd:
Enlarged the icons to 39px (icon box is 40px in the svg)
[icons.zip](https://github.com/user-attachments/files/17714273/icons.zip)

bszlrd:
Enlarged and thickened the mode icons, and added zoom icons (added a new row for those)
[icons.zip](https://github.com/user-attachments/files/17714381/icons.zip)
kép

mwerle:
image

image


I replaced the other button with text using the "orbiteer" font.

mwerle:
image

mwerle:
Ok, I think this is close to being ready for code review..  Git history needs cleaning up before merge.

impaktor:
> Hmm.. they seem a bit small and difficult to make out given the button size.

Is there tooltip / mouse over? Then we could be satisfied with this as first iteration.

bszlrd:
If the icons aren't readable, I can easily play with them after it is merged, and open a PR for them later on.

data/lang/input-core/en.json
```diff
@@ -131,6 +131,22 @@
     "description": "Descriptive name for the PrimaryFire action.",
     "message": "Primary Fire"
   },
+  "BIND_RADAR_TOGGLE_MODE": {
+    "description": "Descriptive name for the RadarToggle action.",
+    "message": "Toggle Radar between 2D and 3D modes"
```
Web-eWorks, line 136:
```suggestion
    "message": "Toggle Radar Display Mode"
```

data/lang/input-core/en.json
```diff
@@ -131,6 +131,22 @@
...
+  },
+  "BIND_RADAR_ZOOM_IN": {
+    "description": "Descriptive name for the RadarZoomIn action.",
+    "message": "Manually zoom radar in"
```
Web-eWorks, line 140:
```suggestion
    "message": "Manual Radar Zoom In"
```

data/lang/input-core/en.json
```diff
@@ -131,6 +131,22 @@
...
+  },
+  "BIND_RADAR_ZOOM_OUT": {
+    "description": "Descriptive name for the RadarZoomOut action.",
+    "message": "Manually zoom radar out"
```
Web-eWorks, line 144:
```suggestion
    "message": "Manual Radar Zoom Out"
```

data/lang/input-core/en.json
```diff
@@ -131,6 +131,22 @@
...
+  },
+  "BIND_RADAR_ZOOM_RESET": {
+    "description": "Descriptive name for the RadarZoomReset action.",
+    "message": "Reset 3D radar zoom to automatic mode"
```
Web-eWorks, line 148:
```suggestion
    "message": "Reset Radar Zoom Mode"
```

src/ship/PlayerShipController.cpp
```diff
@@ -166,6 +166,13 @@ REGISTER_INPUT_BINDING(PlayerShipController)
 	auto landingGroup = controlsPage->GetBindingGroup("LandingControl");
 	input->AddActionBinding("BindToggleLandingGear", landingGroup, Action({ SDLK_n }));
 	input->AddAxisBinding("BindControlLandingGear", landingGroup, Axis());
+
+	auto radarGroup = controlsPage->GetBindingGroup("RadarControl");
```
Web-eWorks, line 170: This should likely be in a new "Flight HUD" page rather than "Ship - Controls" - the latter is intended for bindings affecting the physical control state of the ship, while the new keybindings are related only to the radar widget and don't affect anything about the "in-universe" ship.

data/pigui/modules/radar.lua
```diff
@@ -95,144 +164,212 @@ local function display2DRadar(cntr, size)
...
-			display3DRadar(cntr, Vector2(ui.reticuleCircleRadius * 1.8, size * 2))
+
+		if ui.isMouseClicked(1) then
+			click_on_radar = true
+		end
+		if not toggle_radar and click_on_radar and ui.isMouseReleased(1) then
+			ui.openPopup("radarselector")
+		end
+		-- TODO: figure out how to "capture" the mouse wheel to prevent
+		-- the game engine from using it to also zoom the viewport
```
Web-eWorks, line 318: For this, you'll want to push an InputFrame (`Input.CreateInputFrame("RadarScroll")`) when the player starts hovering the "scrollable area" and pop it when the mouse leaves. That InputFrame should contain an axis binding with mouse wheel up/down assigned to its positive/negative bindings.

The Input system will then allow that InputFrame to "consume" the mouse wheel motion, and the lower-priority binding to zoom the main view will not be activated.

data/pigui/modules/radar.lua
```diff
@@ -95,144 +164,212 @@ local function display2DRadar(cntr, size)
...
+		ui.addStyledText(textpos, ui.anchor.right, ui.anchor.bottom, distance, colors.frame, pionillium.small, lui.HUD_RADAR_DISTANCE, colors.lightBlackBackground)
+	end) -- window
+end -- function displayRadar()
 
 -- reset radar to default at game end
-Event.Register("onGameEnd", function() shouldDisplay2DRadar = false end)
+Event.Register("onGameEnd", function()
+	shouldDisplay2DRadar = false,
+	radar2d:resetZoom(),
+	radar3d:resetZoom()
```
Web-eWorks, line 371: Left-over commas here turn this into a three-value assignment to a single variable rather than three statements.

data/pigui/modules/radar.lua
```diff
@@ -95,144 +164,212 @@ local function display2DRadar(cntr, size)
...
+			ui.addCursorPos(Vector2(0, window_height - button_size))
+			icon = instrument:isAutoZoom() and icons.radar_automatic or icons.radar_manual
+			ui.mainMenuButton(icon, tt, ui.theme.buttonColors.disabled, Vector2(button_size))
+		end
+
+		-- Draw radar range
+		local distance = ui.Format.Distance(instrument:getZoom())
+		local textpos = ui.getWindowPos() + Vector2(window_width, window_height)
+		ui.addStyledText(textpos, ui.anchor.right, ui.anchor.bottom, distance, colors.frame, pionillium.small, lui.HUD_RADAR_DISTANCE, colors.lightBlackBackground)
+	end) -- window
```
Web-eWorks, line 364: Stylistic nitpick - with large enough function bodies, I tend to prefer surrounding both the external `ui.window(... end)` call and the internal function body with newlines, i.e.

```lua
-- code

ui.window(..., function()

    -- code

end)

-- code
```

I picked this up from another Lua codebase, but I find it helps *significantly* with readability. It applies to generally any scope - large `if` blocks or `for` loops benefit, as well as any other place where inline functions are used.

src/ship/PlayerShipController.cpp
```diff
@@ -166,6 +166,13 @@ REGISTER_INPUT_BINDING(PlayerShipController)
...
+	input->AddActionBinding("BindRadarZoomReset", radarGroup, Action({ SDLK_UP }));
+	input->AddActionBinding("BindRadarToggleMode", radarGroup, Action({ SDLK_DOWN }));
```
Web-eWorks, line 172: Nitpick: I personally expect the "down" arrow key to be semantically associated with "reset" - it has both the "go back" semantic association and is in-line with the other manual zoom mode keys by default.

src/ship/PlayerShipController.cpp
```diff
@@ -243,6 +250,12 @@ void PlayerShipController::InputBinding::RegisterBindings()
 
 	toggleLandingGear = AddAction("BindToggleLandingGear");
 	controlLandingGear = AddAxis("BindControlLandingGear");
+
+	AddAction("BindRadarToggleMode");
+	AddAction("BindRadarZoomReset");
+	// TODO: Convert to axis?
```
Web-eWorks, line 256: As mentioned above, I'd recommend adding an additional axis which is only activated when hovering the radar widget. Future code could also make the axis "active" when the radar widget has received some sort of input focus, for use with e.g. controllers.

data/pigui/modules/radar.lua
```diff
@@ -44,38 +52,99 @@ local function getColorFor(item)
...
-	--	ui.addCircle(cntr, halfsize, colors.reticuleCircle, ui.circleSegments(halfsize), ui.reticuleCircleThickness)
-	ui.addCircle(cntr, thirdsize, colors.reticuleCircle, ui.circleSegments(thirdsize), ui.reticuleCircleThickness)
-	ui.addCircle(cntr, twothirdsize, colors.reticuleCircle, ui.circleSegments(twothirdsize), ui.reticuleCircleThickness)
+
+	-- radar background and border
+	ui.addCircleFilled(center, self.size, bgColor, ui.circleSegments(self.size), 1)
+	ui.addCircle(center, self.size, fgColor, ui.circleSegments(self.size), lineThickness * 2)
+
+	-- inner circles
+	--	ui.addCircle(center, halfsize, fgColor, ui.circleSegments(halfsize), lineThickness)
```
Web-eWorks, line 126: Cleanup.

data/pigui/modules/radar.lua
```diff
@@ -44,38 +52,99 @@ local function getColorFor(item)
...
+	local thirdsize = self.size * 0.3
+	local twothirdsize = self.size * 0.7
+	local fgColor = colors.uiPrimary
+	local bgColor = colors.uiBackground:opacity(0.54)
+	local lineThickness = 1.5
 
 	local function line(x,y)
-		-- ui.addLine(cntr + Vector2(x, y) * halfsize, cntr + Vector2(x,y) * size, colors.reticuleCircle, ui.reticuleCircleThickness)
-		ui.addLine(cntr + Vector2(x, y) * thirdsize, cntr + Vector2(x,y) * twothirdsize, colors.reticuleCircle, ui.reticuleCircleThickness)
+		-- ui.addLine(center + Vector2(x, y) * halfsize, center + Vector2(x,y) * size, colors.uiPrimaryDark, lineThickness)
```
Web-eWorks, line 117: Cleanup.

Web-eWorks (changes requested):
Gave this a quick read-through and left comments/suggestions where appropriate. The keybinding names should a) be somewhat short, and b) should be nouns rather than sentences if possible. We can look at adding description strings to the keybindings if needed for clarity.

Overall, the new buttons for the radar mode are implemented in a very sane manner (by the standards of this codebase) and they function reasonably well at smaller screen sizes - except for the A/M icons.

There is a bit of overlap at 1280x720 however, as the HUD window draws on top of the rest of the Flight UI. I'll consider that to be a task for another PR, as that's a bit out-of-scope of this work.

image

---

I leave it to your discretion whether you want to merge this PR as is (after addressing review feedback) and move the Input bindings to Lua in a later PR, or wait until #5962 and the associated hooks for pushing InputFrames have been merged.

bszlrd:
@Web-eWorks noticed an issue with some other icons in the previous file (the warning !!!s). I've fixed them in this one:
[icons.zip](https://github.com/user-attachments/files/17739430/icons.zip)

data/pigui/modules/radar.lua
mwerle: These were in the original code and add additional lines to the 2D radar. I left them in as they may want to be enabled again in a future UI revision?

data/pigui/modules/radar.lua
mwerle: As per https://github.com/pioneerspacesim/pioneer/pull/5958#discussion_r1841241466

mwerle:
> @Web-eWorks noticed an issue with some other icons in the previous file (the warning !!!s). I've fixed them in this one: [icons.zip](https://github.com/user-attachments/files/17739430/icons.zip)

Thanks! Committed in acd89d8044f313154e5c951d9cc7510effb6e0f8

data/lang/input-core/en.json
mwerle: Fixed in 556f767fed6741ec1621a0cd96527f9f321b9d32

data/lang/input-core/en.json
mwerle: Fixed in Fixed in 556f767fed6741ec1621a0cd96527f9f321b9d32

data/lang/input-core/en.json
mwerle: Fixed in Fixed in 556f767fed6741ec1621a0cd96527f9f321b9d32

data/lang/input-core/en.json
mwerle: Fixed in Fixed in 556f767fed6741ec1621a0cd96527f9f321b9d32

src/ship/PlayerShipController.cpp
mwerle: Ok; this may take a few days (have a conference the next 2 days, then hopefully away on a motorbike trip)

mwerle:
> Overall, the new buttons for the radar mode are implemented in a very sane manner (by the standards of this codebase) and they function reasonably well at smaller screen sizes - except for the A/M icons.

Hmm; will need to take a look to see what can be done here, but it'll take a few days as I'm away for the next few.


> I leave it to your discretion whether you want to merge this PR as is (after addressing review feedback) and move the Input bindings to Lua in a later PR, or wait until #5962 and the associated hooks for pushing InputFrames have been merged.

There's a couple of other fixes which I likely won't get around to for a few days, so I can rewrite the input to use #5962 if that has been merged by then. Probably not a bad idea to showcase the new system.

src/ship/PlayerShipController.cpp
mwerle: Fixed in d0e849130043b3ccf6a84d15cc0c026a8e5e542c

src/ship/PlayerShipController.cpp
mwerle: Axis partially added in d0e849130043b3ccf6a84d15cc0c026a8e5e542c ; will need to investigate how this can be activated on mouse-over.

data/pigui/modules/radar.lua
mwerle: Fixed in 4f2069f05c56d9475e530db9feb17fb57059d159

data/pigui/modules/radar.lua
mwerle: Fixed in 4f2069f05c56d9475e530db9feb17fb57059d159

data/pigui/modules/radar.lua
Web-eWorks: Go ahead and leave them in then, but leave a comment inline with them mentioning the above please.

bszlrd:
Of course I forgot to increase the sizes of AM :D

data/pigui/modules/radar.lua
mwerle: There is no code anywhere which binds an axis to the mousewheel.. there are simply APIs called "getWheel" (or similar). Both C++ and Lua side.  There are axis definitions for joysticks and keyboard pairs. I suspect some additional plumbing is required somewhere? Or I'm misunderstanding how the mouse wheel can be turned into an axis..

src/ship/PlayerShipController.cpp
mwerle: Fixed in e2b015d534b99b2a9776d7fa669a565bc2b782ae

mwerle:
> I leave it to your discretion whether you want to merge this PR as is (after addressing review feedback) and move the Input bindings to Lua in a later PR, or wait until #5962 and the associated hooks for pushing InputFrames have been merged.

Fixed in e2b015d534b99b2a9776d7fa669a565bc2b782ae

mwerle:
> Fixed the AM icons too: [icons.zip](https://github.com/user-attachments/files/17751588/icons.zip)

Merged into 3d54882d61a5366d0bb6b69ee9cb9b94f3b8cf15 -> fixup commit replacing the original commit.

data/pigui/modules/radar.lua
mwerle: Merged into commit 81a7203a24bd82bf69b485cd638779eac2950ead

src/ship/PlayerShipController.cpp
mwerle: Axis removed again until I can fully implement it.

data/pigui/modules/radar.lua
mwerle: Ok, fixing the mouse zoom is out-of-scope for this PR.

Web-eWorks (approved):
Provisionally approved - I'll give it a playtest soon :tm:.

data/pigui/modules/radar.lua
```diff
@@ -95,144 +172,222 @@ local function display2DRadar(cntr, size)
...
+		local textpos = ui.getWindowPos() + Vector2(window_width, window_height)
+		ui.addStyledText(textpos, ui.anchor.right, ui.anchor.bottom, distance, colors.frame, pionillium.small, lui.HUD_RADAR_DISTANCE, colors.lightBlackBackground)
+
+	end) -- window
+
+end -- function displayRadar()
+
+-- enable radar input frame on game start
+Event.Register("onGameStart", function()
+	input_frame:AddToStack()
```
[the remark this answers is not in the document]
mwerle: Let me take a look at how to do this..

data/pigui/modules/radar.lua
```diff
@@ -17,13 +15,29 @@ local pionillium = ui.fonts.pionillium
...
 
-local shouldDisplay2DRadar
-local current_radar_size = 10000
+local shouldDisplay2DRadar = false
 local blobSize = 6.0
 
+local input_group = 'ShipHUD.RadarControl'
+local keys = {
+	radar_reset = Input.RegisterActionBinding('BindRadarZoomReset', input_group, { activator = { key = Input.keys['slash'] } }),
+	radar_toggle_mode = Input.RegisterActionBinding('BindRadarToggleMode', input_group, { activator = { key = Input.keys['semicolon'] } }),
```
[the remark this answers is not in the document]
mwerle: Ah, I kinda thought it odd that we're using strings instead of symbols. Another reason to have used UPPERCASE ;)

Alternatively, could do `key_return` or `KEY_RETURN`.  Often a good idea to prefix constant symbols..

Feedback: it would be good to provide the intended use-case as a sample, not the exception, when adding a new feature.

Fixed in 1b370e33f984658232158bddafb4e4840b8a9df1

data/pigui/modules/radar.lua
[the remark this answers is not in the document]
mwerle: Fixed in 1b370e33f984658232158bddafb4e4840b8a9df1

data/pigui/modules/radar.lua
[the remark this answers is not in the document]
Web-eWorks: Given the overall convention of the codebase is `ui.keys.xxx` I originally thought it was close enough to self-obvious that `Input.keys[...]` would be the outlier - especially given that I didn't intend TestInput to be anything other than a quick validation for the new code.

That being said, I can see where people might want to copy the code as-is - I'll see about addressing that later.

data/pigui/modules/radar.lua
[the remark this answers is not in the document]
Web-eWorks: Note: `InputFrame:RemoveFromStack()` "should" be idempotent - you can instead do `if currentView ~= "world" then frame:RemoveFromStack() end`.

If it's not idempotent, bug me about it and I'll fix that.

src/lua/LuaGame.cpp
```diff
@@ -776,6 +816,7 @@ void LuaGame::Register()
 
 		{ "SwitchView", l_game_switch_view },
 		{ "CurrentView", l_game_current_view },
+		{ "PreviousView", l_game_previous_view },
```
Web-eWorks, line 819: I would generally prefer to add additional events fired by a View when it is made (in)active than introduce this new method - though if pushing/popping the InputFrame is implemented according to https://github.com/pioneerspacesim/pioneer/pull/5958#discussion_r1844271088 neither should be necessary.

Web-eWorks (changes requested):

data/pigui/modules/radar.lua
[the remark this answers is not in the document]
mwerle: That's less performant though calling into `RemoveFromStack()` on every view switch instead of only when necessary..  `RemoveFromStack` is not only a call into the C++, but it then also to look through the entire vector.

data/pigui/modules/radar.lua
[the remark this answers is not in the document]
mwerle: Forgive me; I'm still new to the codebase and when I search for examples of how to do something in this case I only found the TestInput section.

src/lua/LuaGame.cpp
mwerle: Hmm.. ok, getting very mixed signals (pun not intended) here..  in another PR you wanted me to remove events and use function calls. Here you're saying I should be using events.

What paradigm is this codebase supposed to use?!

data/pigui/modules/radar.lua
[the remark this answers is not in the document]
Web-eWorks: No forgiveness necessary!

data/pigui/modules/radar.lua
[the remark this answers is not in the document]
Web-eWorks: While I do agree, I consider this to be only a stopgap solution, and a few microseconds to walk a vector of pointers (with N < 30 on average) once every 5-300 seconds is not a particular performance issue in my mind. I'm much more concerned about introducing new API which will likely wind up completely unneeded, especially if we add support for opening a View "inside" another View as a floating window.

src/lua/LuaGame.cpp
Web-eWorks: Sorry about the confusion - there are three (now four) possible states to implement the functionality of popping the input frame:
1. The simplest way would be to statically call `RemoveFromStack` every time `onViewChanged` is triggered, regardless of if the input frame is active. (I can add an `Input.HasInputFrame(frame)` method if you're concerned about the performance here.)
2. Pass the name of the previous frame as a parameter to the `onViewChanged` event.
3. If option 1 or 2 is not chosen, I'd prefer PiGuiView queue an `onViewActivated` / `onViewDeactivated` event.
4. The currently-implemented method of `Game.PreviousView`, which adds an additional API call and duplicates the view name lookup code.

I had originally suggested option 1 in https://github.com/pioneerspacesim/pioneer/pull/5958#discussion_r1844271088, with this comment expressing my desire to use option 3 instead of 4 if option 1 was not chosen. Upon reflection, I think option 2 would be the simplest possible way to achieve this, while still being performant.

Web-eWorks: At this point - just pick whatever you want to implement; I've nitpicked the PR to death and you're more than welcome to say "not going to bother" to nitpicks :)

mwerle: Lol, I actual thought of "2" while I was reverting..   I'll leave it as-is for now, but adding "2" at some point would be a good addition I think. That'd be the most performant and least invasive.

Web-eWorks (approved):
As before, provisionally approved - I should get a chance to playtest this tomorrow morning.

PR #5959, opened 2024-11-09
Sound Cleanup

WickedSmoke:
These commits cleanup the Sound and SoundMusic interfaces to allow plugging in an alternative backend (e.g. OpenAL) simply by switching out sound/Sound.cpp.

This pull request does not address the make file changes required to actually do this.  However, a working proof of concept can be found on [my repository](https://codeberg.org/wickedsmoke/pioneer/commits/branch/faun).

Future work could include:
- Adding a music crossfade method (to allow reducing the five mutex locks used in the SDL implementation down to one).
- Moving common backend code to it's own file.

### DIFF
--- a/src/sound/Sound.cpp
+++ b/src/sound/Sound.cpp
@@ -794,6 +794,14 @@ namespace Sound {
 		return status;
 	}
 
+	bool Event::FadeOut(float dv_dt, Op op)
+	{
+		bool found = VolumeAnimate(0.0f, 0.0f, dv_dt, dv_dt);
+		if (found)
+			SetOp(op | Sound::OP_STOP_AT_TARGET_VOLUME);
+		return found;
+	}
+
 	const std::vector<std::string> GetMusicFiles()
 	{
 		std::vector<std::string> songs;
--- a/src/sound/Sound.cpp
+++ b/src/sound/Sound.cpp
@@ -183,10 +183,10 @@ namespace Sound {
 		const Sample *sample;
 		OggVorbis_File *oggv; // if sample->buf = 0 then stream this
 		OggFileDataStream ogg_data_stream;
-		Uint32 buf_pos;
+		uint32_t buf_pos;
 		float volume[2]; // left and right channels
 		eventid identifier;
-		Uint32 op;
+		uint32_t op;
 
 		float targetVolume[2];
 		float rateOfChange[2]; // per sample
@@ -244,12 +244,12 @@ namespace Sound {
 	/*
  * Volume should be 0-65535
  */
-	static Uint32 identifier = 1;
+	static uint32_t identifier = 1;
 	eventid PlaySfx(const char *fx, const float volume_left, const float volume_right, const Op op)
 	{
 		SDL_LockAudioDevice(m_audioDevice);
 		unsigned int idx;
-		Uint32 age;
+		uint32_t age;
 		/* find free wavstream (first two reserved for music) */
 		for (idx = 2; idx < MAX_WAVSTREAMS; idx++) {
 			if (!wavstream[idx].sample) break;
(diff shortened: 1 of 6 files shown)


src/sound/Sound.cpp
```diff
@@ -690,7 +696,19 @@ namespace Sound {
...
+		}
+	}
+
+	void Event::PlayMusic(const char *fx, const float volume_left, const float volume_right, Op op)
+	{
+		Stop();
+		Sample* sample = GetSample(fx);
+		if (sample) {
+			eid = PlayMusicSample(sample, volume_left, volume_right, op);
+		}
```
mwerle, line 711: This will stop the music even if the new music fails to load.

```suggestion
		Sample* sample = GetSample(fx);
		if (sample) {
			Stop();
			eid = PlayMusicSample(sample, volume_left, volume_right, op);
		}
```

WickedSmoke: I kept the previous behavior of MusicEvent::Play().  I'm thinking this is desirable since having no sample is an error and it will let the user know something is wrong if no music is playing.

mwerle: Yeah, sorry I realized later when i saw another function with similar behavior but forgot to update this comment. 

Is the missing sound file logged? That would be useful for the devs.

WickedSmoke: GetSample() has a commented out SilentWarning() call.

WickedSmoke:
I have my sound back end working reasonably well, so I went ahead and did the crossfade consolidation.

Web-eWorks:
Will review this soon - been busy with some other concerns. What I'm reading so far seems to be moving in the right direction; I definitely appreciate moving the "fiddly details" of audio mixing entirely behind a higher-level API.

Do note that any proposed audio backend replacement would necessarily need to be able to handle 3d directional audio, so I have my eyes on something along the lines of soloud or an equivalent middleware - my understanding of OpenAL is that it is a somewhat heavyweight dependency.

src/sound/Sound.cpp
mwerle: Wonder why it's commented out.. anyway, out-of-scope of this PR I guess.

src/sound/SoundMusic.cpp
```diff
@@ -109,18 +89,7 @@ namespace Sound {
...
-		std::vector<std::string> songs;
-		songs.reserve(samples.size());
-		for (std::map<string, Sample>::const_iterator it = samples.begin();
-			 it != samples.end(); ++it) {
-			if (it->second.isMusic)
-				songs.emplace_back(it->first.c_str());
-		}
-
-		return songs;
+		return GetMusicFiles();
```
Web-eWorks, line 92: This is conceptually messy, but in the interest of not increasing the scope of this PR I'll consider this "not worse than before"; the compiler is very likely to perform RVO here, so it's not a significant performance issue.

Web-eWorks (changes requested):
One minor issue I'd like you to address before merge. This is a general improvement of the Sound API by hiding the messy low-level details behind the API surface, which I'm very much a fan of.

src/sound/Sound.h
```diff
@@ -1,10 +1,10 @@
 // Copyright © 2008-2024 Pioneer Developers. See AUTHORS.txt for details
 // Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
-#ifndef __OGGMIX_H
-#define __OGGMIX_H
+#ifndef __SOUND_H
+#define __SOUND_H
 
-#include <SDL_stdinc.h>
+#include <stdint.h>
```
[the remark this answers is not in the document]
WickedSmoke: The include has been changed.

PR #5962, opened 2024-11-12
feat(ui): add keybinding for target radial

mwerle:
Add a keybinding to open the target radial menu. The bind is empty by default as per the other target radials.

This allows keyboard controls to fly to, orbit, or dock with the current frame object or nav target, as appropriate.

### DIFF
--- a/data/pigui/modules/flight-ui/reticule.lua
+++ b/data/pigui/modules/flight-ui/reticule.lua
@@ -43,7 +43,8 @@ local showNavigationalNumbers = true
 -- to interact with actions and axes in the input system
 local bindings = {
 	assistRadial = bindManager.registerAction('BindFlightAssistRadial'),
-	fixheadingRadial = bindManager.registerAction('BindFixheadingRadial')
+	fixheadingRadial = bindManager.registerAction('BindFixheadingRadial'),
+	targetRadial = bindManager.registerAction('BindTargetRadial'),
 }
 
 -- display the pitch indicator on the right inside of the reticule circle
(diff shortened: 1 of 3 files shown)


data/pigui/modules/flight-ui/reticule.lua
```diff
@@ -321,8 +322,10 @@ local function displayDetailData(target, radius, colorLight, colorDark, tooltip,
 	local uiPos = ui.pointOnClock(center, radius, 2.46)
 	-- label of target
 	local nameSize = ui.addStyledText(uiPos, ui.anchor.left, ui.anchor.baseline, target.label, colorDark, pionillium.medium, tooltip, colors.lightBlackBackground)
-	if ui.isMouseHoveringRect(uiPos - Vector2(0, pionillium.medium.size), uiPos + nameSize - Vector2(0, pionillium.medium.size)) and ui.isMouseClicked(1) and ui.noModifierHeld() then
-		ui.openDefaultRadialMenu("game", target)
+	if ui.isMouseHoveringRect(uiPos - Vector2(0, pionillium.medium.size), uiPos + nameSize - Vector2(0, pionillium.medium.size)) and ui.isMouseClicked(1) and ui.noModifierHeld()
```
Web-eWorks, line 325: I would strongly prefer that the `ui.isMouseHoveringRect()` part of the condition be split into another temporary variable named `isHovered`, as with the introduction of the `or` case it's quite lengthy and bug-prone.

Web-eWorks (changes requested):
Looks simple enough - I have only one spot where I'd prefer some cleanup be done to increase the future maintainability of the code.

I cannot promise a timeline, but I'll be working on allowing Lua to register inputs without the need for the C++-side `AddActionBinding` calls.

data/pigui/modules/flight-ui/reticule.lua
mwerle: Fixed in 894b7ad45be1c7494907132406299a3ed3a431fc

mwerle:
> I cannot promise a timeline, but I'll be working on allowing Lua to register inputs without the need for the C++-side `AddActionBinding` calls.

It appears the timeline was <12hrs...  should I update this PR to make use of the new API?

Web-eWorks:
> It appears the timeline was <12hrs... should I update this PR to make use of the new API?

The only thing that's currently missing to do so is a nice place to hang a Lua-side InputFrame off of - we don't currently have a good way to track when the WorldView is made active and push/pop an InputFrame in response. Some work would likely be needed to amend `PiGuiView.cpp` to trigger `LuaEvent::Queue("onViewActive", GetViewName());` etc. in the `OnSwitchTo` and `OnSwitchFrom` methods.

I leave it at your discretion whether you want to merge this first, then convert all of the radial menu keybinds as described, or expand the scope of this PR and tackle all of the above in this PR.

Web-eWorks (approved):

mwerle:
> I leave it at your discretion whether you want to merge this first, then convert all of the radial menu keybinds as described, or expand the scope of this PR and tackle all of the above in this PR.

Let's keep the scope to the actual feature for now; please merge if acceptable and I'll create a TODO ticket to put in the relevant plumbing and upgrade all radial menus / Lua keybinds in a clean-up ticket.

Web-eWorks:
Merged, thanks!

Issue #5566, opened 2023-03-05
AI Natrix: very nonstandard takeoff procedure


craigo-:
savegame (2023-02-03 level): [Vark Sijyn (The Breadmaker) KILL NATRIX.zip](https://github.com/pioneerspacesim/pioneer/files/10890651/Vark.Sijyn.The.Breadmaker.KILL.NATRIX.zip)

### Observed behaviour

So I'm on Leclercq in the Tsuhotakko system [-3, -5, -2 for you geospatial nerds]. It's sort of Earth-sized, but without the useful atmosphere. Let's just say I'm not buying a condo here any time soon.

My current mission is to kill Cornelius Huber, in Natrix TS-3270. This is a high-value target and I don't ask questions. I'm parked right beside him in my innocent-looking Mola Mola, looking out the left side of the cockpit as he's about to take off less than a minute from now (at 10:31:13).

Then I see Cornelius attempt to takeoff, and my jaw drops. The ship leans right and starts pitching up, as if the landing pad is buckling under an earthquake. I check my sensors, but my Mola Mola is straight and level. Then I notice the Natrix's RCS thrusters firing, and I slowly grok that this is what Huber *wants* the ship to do.

Tortured alloy sparks shoot upwards as Cornelius' Natrix slowly but surely wrenches itself into a vertical position, surely tearing ragged holes into itself and the pad. Is he gonna get billed for that?

Finally, the ship (or what remains of it) is somehow vertical, Cornelius engages the Natrix's main thrusters and the ship wrenches itself clear of the pad and gains altitude.

I love the Natrix. Hey, I started out in its little brother the Xylophis space van, so I have a soft spot for anything OPLI-Barnard. I've read the Natrix's sales brochure many times, but it never talked about needing a flippin' _launch tower_ so it could take off vertically. They should have said, "10% hull damage free with every takeoff!" Suffice to say, I'm never buying one.

I let Cornelius Huber go. He has enough problems.

### Expected behaviour

The Natrix is able to take off in 0.94G

### Steps to reproduce

Unzip and load attached savegame

__Screenshot__

![image](https://user-images.githubusercontent.com/4467963/222939860-c34856c2-b1fe-474e-b036-1aa163ef9158.png)

__My pioneer version (and OS):__

Master (f79db1d1a1e435fa3647ec75a89412573900d988)
Windows 10 Enterprise 22H2

NB: possibly related to issue #5243

Web-eWorks:
[...]

PR #5963, opened 2024-11-12
Natrix overhaul

bszlrd:
Finally been able to finish up the Natrix overhaul.
kép

I've turned it into a heavy courier, the largest one in the courier line of OPLI. The proper cargo replacement for the old Natrix is coming soon™, named Fowlea. The scale is pretty much the same as the original, a very tiny bit thicker and longer, but with a skinnier rear section:
kép

I pulled the stats from Notion, so it uses the new equipment setup (and crashes the game, as all other ships except the Coro I believe. It won't crash the midsommar release, but won't start a new game with it either). If anybody wants to try it out with midsommar, then this [natrix.json](https://github.com/user-attachments/files/17716068/natrix.json) should work.

It has 4k\*2k textures, and 4k\*4k for the cockpit, all DDS except the normal maps. All have mipmaps. I still need to do the patterns (wink wink @Keeper1st :) )
Tri counts:
- High Lod: 90k
- Medium: 26k
- Low: 13k
- Collision: 1135
- Shield: 576
- Cockpit: 228k

I've opted not to shrink the main engines as we discussed. I want to get it out the door without needing to rebake the AO. And I'm kinda leaning into this scale being the S3 engine, as it turned out that the old S3 size is quite small. And I think it looks better this way, with (if ever so slightly) heftier engines.

Mostly shares cockpit with the Xylophis, so I've put its model and textures in that directory. I've added two back seats, but for now I kept them without any instruments or so (and it is really only one additional seat, because the other won't be visible). The cockpit itself is a bit barebones compared to the other two couriers, but I do want to iterate on that later on. 

[**Sketchfab model**](https://sketchfab.com/3d-models/natrix-courier-redesign-ca4647ab60a44a1990e50a309cde1b6d)

kép
kép
kép
natrix_overhaul_preview_031
natrix_overhaul_preview_032

S5pzTnV
et4TsdK
xPO87xT


fluffyfreak:
That looks SO good! 🎉

Web-eWorks:
Very much a fan of this - it's good to finally have a "proper" Natrix that can take off from a 1G world! Don't worry about the engine sizing - I think the Natrix engine will likely become the S3 engine and the "generic" S2 template will be a bit bigger than the OPLI S2 engine in response.

Web-eWorks (approved):
With a playtest, it feels quite well.

I'm a big fan of how ponderous the ship feels, especially rotationally, but I think it could stand to have maybe 10% more angular acceleration as it feels like a much larger ship and not an OPLI courier at present. Additionally, `0.1` lift coefficient is pretty bad for the ship's size and shape - I bumped it to 0.3 and it felt pretty good.

DeltaV is... fairly bad with full equipment and a full cargo load; a trip to Neptune from Mars took 5-10 minutes of real time at maximum timewarp. Not sure if that's something to address at this point however, as it's very much a downstream issue of being able to set autopilot fuel reserve and the player's decision to take a full load of cargo and halve their dV.

We can import those changes later with the next batch of ship.json updates however, so I'm going to merge this as is.

Web-eWorks:
Squash-merged to not have duplicate copies of the dds files in the git history.

craigo-:
> Very much a fan of this - it's good to finally have a "proper" Natrix that can take off from a 1G world!

Huzzah! Perhaps [my mark](https://github.com/pioneerspacesim/pioneer/issues/5566) has a fighting chance in a future engagement.

PR #5964, opened 2024-11-12
Removing normal map from default cockpit materials

### LINKED ISSUE #5960: default_cockpit_norm.png missing from models/cockpits/default_cockpit/
It looks like the default_cockpit does not have this normal map committed but it is referenced within the `default_cockpit.model` 

### Expected behaviour
Either the file needs adding or the reference needs removing

### Steps to reproduce
Load the game, enable cockpit in options, start a new game and look at the output

__My pioneer version (and OS):__ Latest [Git version](https://github.com/pioneerspacesim/pioneer/commit/5e06ba6473b66ee5dfc2fcffde1df323bc62867a) on Windows and Raspberry Pi

__My output.txt (required)__
[output.txt](https://github.com/user-attachments/files/17687849/output.txt)

bszlrd:
Fixes #5960 
I do have the normal map, I forgot to include it, it seems. But its effect is negligible, and the file is 6mb, so I'd rather remove it. (Or I could add it if you peeps think it should be there, could also reduce its 4k size).

### DIFF
--- a/data/models/cockpits/default_cockpit/default_cockpit.model
+++ b/data/models/cockpits/default_cockpit/default_cockpit.model
@@ -5,7 +5,6 @@ shininess 43
 tex_diff default_cockpit_diff.dds
 tex_spec default_cockpit_spec.dds
 tex_glow default_cockpit_glow.dds
-tex_norm default_cockpit_norm.png
 
 material default_cockpit_transp
 diffuse 0.9 0.9 0.9
@@ -15,7 +14,6 @@ opacity 30
 tex_diff default_cockpit_diff.dds
 tex_spec default_cockpit_spec.dds
 tex_glow default_cockpit_glow.dds
-tex_norm default_cockpit_norm.png
 
 material default_cockpit_screen
 diffuse 0.333 0.324 0.329


mwerle:
"The best part is no part"

fluffyfreak:
Before making a decision would you mind adding a with/without comparison?
Also as it's a normal map perhaps we can get away with a lower 1kx1k, or even 2kx2k resolution?

bszlrd:
With:
screenshot-20241112-151724
Without:
screenshot-20241112-152048

Here's the normal map, if you want to try it out:
default_cockpit_norm

And here it is with cranked up contrast. Mosty the details are on the seat:
kép

Web-eWorks:
If the normal map was never committed to the repo, then I'd say go ahead and just remove it from the materials entirely. We can revisit it when we implement "proper" lighting, though if we move to tiling materials rather than baked maps to complement our this-decade lighting approach, we probably won't need the normal map anyways.

bszlrd:
It wasn't. The commit removes the references to it

Web-eWorks:
Thanks!

PR #5965, opened 2024-11-13
Fully expose InputBindings to Lua

Web-eWorks:
Since the Input subsystem was written, we've had partial support for interacting with it from Lua code. Over time, those capabilities have been expanded, but have remained strictly read-only.

This PR changes that, providing a proper way for Lua code to register input bindings, make those bindings active and able to receive input, and listen for input events coming from those bindings.

Now, Lua code can register Action/AxisBindings at startup, create an InputFrame to make those bindings active, push that frame when appropriate, register callbacks to be called from C++ when the input binding is activated, and then pop the input frame when the bindings should no longer be receiving input.

This PR does *not*:
- Provide support for Lua to add an Action/Axis binding to an InputFrame owned by C++
- Add mechanisms to automatically push/pop InputFrames, e.g. on view change
- Automatically clean up registered callbacks on hot-reload of the file that defined the callback
- Migrate any existing code to use the new API

Further work by other developers would be very much appreciated to move old code using the `ui.isKeyPressed()` API to the new Input API, and to provide keybindings to otherwise un-bindable actions that are currently only accessible through the UI.

### DIFF
--- a/src/lua/LuaInput.cpp
+++ b/src/lua/LuaInput.cpp
@@ -12,10 +12,152 @@
 #include "LuaWrappable.h"
 #include "Pi.h"
 
-#include "SDL_keyboard.h"
+#include <SDL_keyboard.h>
+#include <SDL_keycode.h>
 #include <sstream>
 
 using namespace InputBindings;
+
+static std::vector<std::pair<const char *, int>> s_input_keys = {
+	// Symbol keys
+
+	{ "exclaim", SDLK_EXCLAIM },
+	{ "quotedbl", SDLK_QUOTEDBL },
+	{ "hash", SDLK_HASH },
+	{ "percent", SDLK_PERCENT },
+	{ "dollar", SDLK_DOLLAR },
+	{ "ampersand", SDLK_AMPERSAND },
+	{ "quote", SDLK_QUOTE },
+	{ "leftparen", SDLK_LEFTPAREN },
+	{ "rightparen", SDLK_RIGHTPAREN },
+	{ "asterisk", SDLK_ASTERISK },
+	{ "plus", SDLK_PLUS },
+	{ "comma", SDLK_COMMA },
+	{ "minus", SDLK_MINUS },
+	{ "period", SDLK_PERIOD },
+	{ "slash", SDLK_SLASH },
+
+	{ "colon", SDLK_COLON },
+	{ "semicolon", SDLK_SEMICOLON },
+	{ "less", SDLK_LESS },
+	{ "equals", SDLK_EQUALS },
+	{ "greater", SDLK_GREATER },
+	{ "question", SDLK_QUESTION },
+	{ "at", SDLK_AT },
+
+	{ "leftbracket", SDLK_LEFTBRACKET },
+	{ "backslash", SDLK_BACKSLASH },
+	{ "rightbracket", SDLK_RIGHTBRACKET },
+	{ "caret", SDLK_CARET },
+	{ "underscore", SDLK_UNDERSCORE },
+	{ "backquote", SDLK_BACKQUOTE },
+
+	// Misc. keys
+	{ "return", SDLK_RETURN },
+	{ "escape", SDLK_ESCAPE },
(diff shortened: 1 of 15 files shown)


src/lua/LuaInput.cpp
```diff
@@ -2,21 +2,245 @@
...
+#include <SDL_keyboard.h>
+#include <SDL_keycode.h>
 #include <sstream>
 
 using namespace InputBindings;
 
+static std::vector<std::pair<const char *, int>> s_input_keys = {
+	// Symbol keys
+
+	{ "exclaim", SDLK_EXCLAIM },
```
mwerle, line 27: Q: Why have the LUA symbols lowercase instead of uppercase? 

It makes searching for constants harder (I tend to use case-sensitive searches when exploring the codebase to reduce the number of false positives)

Also: aren't strings significantly less performant than integer indeces (No idea if LUA has a concept of enums)?

src/lua/LuaInput.cpp
```diff
@@ -2,21 +2,245 @@
...
+
+	{ "colon", SDLK_COLON },
+	{ "semicolon", SDLK_SEMICOLON },
+	{ "less", SDLK_LESS },
+	{ "equals", SDLK_EQUALS },
+	{ "greater", SDLK_GREATER },
+	{ "question", SDLK_QUESTION },
+	{ "at", SDLK_AT },
+
+    { "leftbracket", SDLK_LEFTBRACKET },
```
mwerle, line 51: Nitpick: indentation is inconsistent..  spacve vs tab perchance?

src/lua/LuaPiGui.cpp
```diff
@@ -3740,7 +3734,6 @@ void LuaObject<PiGui::Instance>::RegisterClass()
 		{ "key_none", l_attr_key_none },
 		{ "key_shift", l_attr_key_shift },
 		{ "key_alt", l_attr_key_alt },
-		{ "keys", l_attr_keys },
```
mwerle, line 3743: Q: why have the keys been moved to LuaInput.cpp, but the modifier keys not?

mwerle (commented):
Don't pretend to understand half of this, and haven't actually tried it, but looking good. 

Otsukaresamadeshita!

src/lua/LuaInput.cpp
Web-eWorks: > Why have the LUA symbols lowercase instead of uppercase?

1. For compatibility with existing code - this matches the existing `ui.keys.xxxx` defines and directly replaces them.
2. I happen to like the look of `Input.keys.delete` more than I like `Input.keys.DELETE`, and it's more conformant to the closest thing we have to a [style guide](https://forum.pioneerspacesim.net/viewtopic.php?p=6293#p6293).

> Also: aren't strings significantly less performant than integer indeces (No idea if LUA has a concept of enums)?

These are named aliases to integer values. They will typically be looked up once at the start of the program when registering key bindings, so performance isn't really an issue. This is effectively the Lua concept of an enum.

Note that in Lua, all strings are immutable and interned, so comparing strings is a direct pointer comparison and equal strings have a shared "pointer identity" used to look up the string in a hashmap (i.e. as a table key).

src/lua/LuaInput.cpp
Web-eWorks: Yep, the SDL2 headers use space indentation and my IDE didn't automatically convert when pasting. Will fix.

src/lua/LuaPiGui.cpp
Web-eWorks: The `pigui.keys` value is a static enumeration of SDLK_XXXX integer values. `pigui.key_alt`, `pigui.key_shift`, `pigui.key_ctrl` and `pigui.key_none` are instantaneous accessors which query the current ImGui keyboard state and return whether those specific keys are held down.

This is actually a separation of concerns as the previous `pigui.keys` had absolutely no relation to the `pigui.key_*` accessors.

PR #5968, opened 2024-11-14
Small fixes and improvements

Web-eWorks:
Another batch of "cleanup" work - this adds autocomplete suggestions for `Input.RegisterAxis/ActionBinding`, adds a CMake option to enable an address-sanitizer build, and cleans up some very unneeded header includes in `core/StringName.h`.

I think that was the last `#include "profiler/Profiler.h"` in a widely-used header, so I had to address quite a few files in the process.


CMakeLists.txt
```diff
@@ -99,6 +99,16 @@ if (USE_TIME_TRACE)
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ftime-trace")
 endif()
 
+option(USE_ASAN "Use -fsanitize=asan when compiling (requires Clang)" OFF)
```
Gliese852, line 102: -fsanitize=address?

Web-eWorks: Yep, I overlooked that - this originally also had `-fsanitize=thread` as an option until I realized our codebase uses too many non-pthread synchronization primitives.

PR #5970, opened 2024-11-16
Allocate light data UBO from per-frame allocator, misc. fixes and cleanup

### LINKED ISSUE #5940: Negative delta-v displayed for certain ships in shipyard
**Initially discovered by Cat'O'Wisdom on Discord**

### Observed behaviour
Some ships (currently confirmed: Pumpkinseed, Varada, Xylophis and Kanara) have negative delta-v displayed in stats viewed in shipyard

### Expected behaviour
Having correct value displayed

### Steps to reproduce
Load or start a game, find any of the listed ships in the station's shipyard

Pioneer version: release from 10.07.2024
OS: Windows 10 x64

__Screenshots__
screenshot-20241026-110941
screenshot-20241026-111757
screenshot-20241026-111823
screenshot-20241026-111849

[output.txt](https://github.com/user-attachments/files/17529599/output.txt)

Web-eWorks:
Another general collection of "fix" commits.

This PR adds additional profiling information to the renderer, and resolves an "early serialization" point which required the previous frame's rendering to be ~mostly complete before starting to record commands for the current frame.

I don't expect any significant performance increase from this PR, except in the case of extremely GPU-bound devices where you might see an improvement of up to a few milliseconds in frame-time, as more CPU work is able to be overlapped with the previous frame's GPU execution time.

This change does fully enforce the existing invariant that you must call `Renderer::SetLights()` *before* attempting to record a draw command using a lit material. An assertion will be thrown in debug mode if this invariant is broken.

It also fixes #5940 - turned out to be quite a simple one-liner once I realized what the problem was.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ if (USE_TIME_TRACE)
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ftime-trace")
 endif()
 
-option(USE_ASAN "Use -fsanitize=asan when compiling (requires Clang)" OFF)
+option(USE_ASAN "Use -fsanitize=address when compiling (requires Clang)" OFF)
 if (USE_ASAN)
 	add_compile_options(
 		-g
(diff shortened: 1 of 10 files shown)


fluffyfreak (approved):
Looks good, took me a while to figure out the UniformBuffer only ever allocating but I got there.

fluffyfreak:
Significant performance improvement on the RPi5, at least as good as #5972 so I'll close that one 👍

PR #5971, opened 2024-11-17
Fix the ability to start from all start positions

### LINKED ISSUE #5969: Crash/assert when starting new game other than Mars
### Observed behaviour
Launch Pioneer, choose "New Game" at the menu, select "Start at New Hope" from the profile drop-down and you get the following assert. 
PioneerLuaNewGameCrash

If you accept these you'll receive another and then a warning and an error before Pioneer quits out.

It looks like an issue with an EquipSet slot

### Expected behaviour
Not expected to assert and crash 😁

Windows 10, 2668aa861
Also seen on Raspberry Pi.

[output.txt](https://github.com/user-attachments/files/17787435/output.txt)

Gliese852:
Although setting up and even loading the correct equipment from the start variant does not yet work, there is no crash when choosing a different starting position.

Thanks to @fluffyfreak
Fixes #5969

### DIFF
--- a/data/pigui/modules/new-game-window/class.lua
+++ b/data/pigui/modules/new-game-window/class.lua
@@ -185,16 +185,16 @@ local function startGame(gameParams)
 		-- print("Installing equipment {} (proto: {}) into slot {}" % { item, proto, slot })
 		if type(slot) == "string" then
 			local slotHandle = equipSet:GetSlotHandle(slot)
-			assert(slotHandle)
+			if slotHandle then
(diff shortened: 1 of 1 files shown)


fluffyfreak (commented):
Obviously I think that it looks good 😁👍 I'll let someone else code review and merge

Web-eWorks:
If it's been tested, it has my approval. Won't get a chance to do anything myself for another 4 hours or so.

PR #5974, opened 2024-11-18
Feat/view event and names

mwerle:
This PR is a proposal to be discussed prior to merge.

The first couple of commits fix a few outstanding compiler warnings.

The first "real" commit sanitized the various view names, both between all the views, as well as between the C++ side and the Lua side. Perhaps there is a good reason to have a different name between the Lua and the C++ side, but if there is, it escapes me. Personally I find the proposed naming to be much more self-documenting as well as easier to find and it reduces overloading of some of the strings which were used in multiple places with different meanings.

This is the only change which I think needs discussing as I'm not sure if this change is valid/wanted. (I think it makes sense, but I may not have all the relevant background information..)

Following this renaming, I added "previous" and "new" view names to the "onViewChanged" event.

Finally, the new "onViewChanged" parameters are used in the "radar" module to improve the way the input frame is added and removed, as previously discussed here : https://github.com/pioneerspacesim/pioneer/pull/5958#discussion_r1844722173


mwerle:
> No complaints here - this is a generally good change, my only nitpick is that we don't actually _have_ a SettingsView and that name isn't used anywhere.
> 

Hmm, literally replaced what was there to match the new naming - perhaps it was a leftover from previous refactors..  I should've checked a bit closer.


Ok, killed `PiGuiView` and cleaned up the commits a little bit.  "Works for me" but YMMV..

Web-eWorks (approved):
Good work, thanks for tackling this! I'm glad to reduce the bug surface when it comes to identifying views between C++ and Lua. Overall it looks good and I don't see any fixup commits in the commit list, so this is getting merged!

PR #5975, opened 2024-11-18
Natrix patterns

bszlrd:
Made some patterns during the weekend.
They are half-sized, because at fill size they were almost 12mb each, so I decided to halve the dimensions.
(First is no-pattern. That file is 1*1px)
pattern01
pattern02
pattern03
pattern04
pattern05
pattern06
pattern07
pattern08
pattern09


mwerle:
Nice!!

..but why are all the ships "worn" looking? Presumably I can buy a new ship and/or give a ship a quick paintjob just before selling it?

Web-eWorks:
Just FYI - the pattern shader only samples from the red channel to determine which color to use. Unfortunately, we can't really get any smaller textures by using grayscale patterns as the DXT1/DXT5 formats specifically store three-component color data.

bszlrd:
I tried making them from grayscale pngs, but it did not work

PR #5978, opened 2024-11-19
feat(radar): capture mouse wheel for zooming radar

mwerle:
By placing all of the radar drawing routines into a window, the mouse wheel is captured and is not passed back to the world view.

The work to add manual zoom to the 3D radar (#5958) failed to get this to function properly as the wrong API was used for setting the center of the 3D radar. Using 'setCursorScreenPops()' instead of 'setCursorPos()' has the 3D radar background being drawn properly  when placed inside a window.

~~TODO?: When zooming the 3D radar with the mouse, the mode is switched from automatic zoom to manual zoom, but there is not way (using the mouse) to switch the radar back to automatic zoom. Should there be? One option could be that zooming past the maximum zoom automatically switches back to auto-zoom, but testing shows that this is too easy to do accidentally.~~

->  Implemented by making the "Automatic/Manual" mode indicator button switch between manual and automatic mode.

### DIFF
--- a/data/pigui/modules/radar.lua
+++ b/data/pigui/modules/radar.lua
@@ -15,7 +15,7 @@ local pionillium = ui.fonts.pionillium
 local colors = ui.theme.colors
 local icons = ui.theme.icons
 
-local SCREEN_BORDER = 4
+local SCREEN_BORDER = 6
 
 local MAX_RADAR_SIZE = 1000000000
 local MIN_RADAR_SIZE = 1000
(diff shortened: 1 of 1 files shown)


bszlrd:
How about double-clikcing on the radar to switch between auto and manual? Would need to be documented, likely on the tooltip as well, but could be handy.

mwerle:
> How about double-clikcing on the radar to switch between auto and manual? Would need to be documented, likely on the tooltip as well, but could be handy.

Not sure I like double-clicking (can add if people think it's good).

But completely forgot about the zoom mode indicator button - I've enabled clicking on it to toggle zoom modes instead of it simply being an indicator.

bszlrd:
Oh, I thought you want to add another way on top of the mode indicator. I assumed that it is a button

mwerle:
> Oh, I thought you want to add another way on top of the mode indicator. I assumed that it is a button

In the original PR, it was just an indicator; didn't seem to make sense to have a specific button to change modes if there's no buttons for changing the zoom manually.

In this PR I made it a button to go back to Automatic mode after changing to Manual mode with the mouse wheel.

Web-eWorks (approved):
Functions well, good work!

A thought for a follow-up PR - when clicking the auto/manual toggle button, the manual zoom level should be set so that the tracked target is still visible on the radar scope.

...Additionally, it'd be nice to have intermediary steps between log10 increments for manual zoom, e.g. 1Mm, 2Mm, 5Mm, 10Mm.

mwerle:
Rebased on master ready for merge.

mwerle:
> Functions well, good work!
> 
> A thought for a follow-up PR - when clicking the auto/manual toggle button, the manual zoom level should be set so that the tracked target is still visible on the radar scope.
> 
> ...Additionally, it'd be nice to have intermediary steps between log10 increments for manual zoom, e.g. 1Mm, 2Mm, 5Mm, 10Mm.

#5984 :)

PR #5981, opened 2024-11-22
Feat/modelspinnermousebutton

mwerle:
Took over #5594 as @zonkmachine indicated they no longer have the time to complete their PR.

Changes from the last commit:

- Unify almost all spinning by creating Input::Manager::IsMouseRotatePressed()
  - almost a single place to implement the logic instead of in multiple places throughout the code.
  - the current button combos for rotation are the MMB and Ctrl+LMB
  - The ModelSpinner has its own implementation as it uses `ImGui`, not `Input::Manager`
  - TODO: remove the now-obsolete "No middle mouse button" option
  - TODO: add user configuration options for the buttons/modifiers used
- Add Input::Manager::GetRotateSpeedShiftModifier() to also unify rotation speed modifiers
  - change using the Right-Shift key to the Left-Alt key for the secondary speed and rotation modifiers
- Fix SystemView deselecting objects when rotation is started
- Fix ShipView rotating when rotating the ship (via RMB) itself.

**Discussion:**
Given that the RMB moves the ship itself, would this be the more natural default to spin everything instead of the MMB? That could free up the MMB for panning/moving in the SectorView and SystemView.

### DIFF
--- a/src/Input.cpp
+++ b/src/Input.cpp
@@ -251,11 +251,13 @@ Manager::Manager(IniConfig *config, SDL_Window *window) :
 	m_capturingMouse(false),
 	joystickEnabled(true),
 	mouseYInvert(false),
+	noMiddleMouseButton(false), // Which means a middle mouse button is expected by default
 	m_enableBindings(true),
 	m_frameListChanged(false)
 {
 	joystickEnabled = (m_config->Int("EnableJoystick")) ? true : false;
 	mouseYInvert = (m_config->Int("InvertMouseY")) ? true : false;
+	noMiddleMouseButton = (m_config->Int("noMiddleMouseButton")) ? true : false;
 
 	Input::InitJoysticks(m_config);
 }
@@ -513,6 +515,15 @@ void Manager::SetMouseYInvert(bool state)
 	mouseYInvert = state;
 	if (m_enableConfigSaving) {
 		m_config->SetInt("InvertMouseY", mouseYInvert);
+		m_config->Save();
+	}
+}
+
+void Manager::SetMiddleMouseButton(bool state)
+{
+	noMiddleMouseButton = state;
+	if (m_enableConfigSaving) {
+		m_config->SetInt("noMiddleMouseButton", noMiddleMouseButton);
 		m_config->Save();
 	}
 }
--- a/src/Input.cpp
+++ b/src/Input.cpp
@@ -857,10 +857,16 @@ void Manager::DispatchEvents()
 
 */
 
+// Mouse rotation is enabled if:
+// MMB is pressed, or Ctrl+LMB is pressed, AND no other mouse buttons are pressed
 bool Manager::IsMouseRotatePressed()
 {
-	return MouseButtonState(SDL_BUTTON_MIDDLE) ||
-	       ((keyModState & KMOD_LCTRL) && MouseButtonState(SDL_BUTTON_LEFT));
+	int state = keyModState & KMOD_LCTRL ? 0x01 : 0x00;
+	state |= MouseButtonState(SDL_BUTTON_LEFT) ? 0x02 : 0x00;
+	state |= MouseButtonState(SDL_BUTTON_MIDDLE) ? 0x04 : 0x00;
+	state |= MouseButtonState(SDL_BUTTON_RIGHT) ? 0x08 : 0x00;
+
+	return state == 0x03 || state == 0x04;
 }
 
 void Manager::SetCapturingMouse(bool grabbed)
@@ -884,6 +890,13 @@ float Manager::GetMoveSpeedShiftModifier()
 {
 	// Suggestion: make x1000 speed on pressing both keys?
 	if (KeyState(SDLK_LSHIFT)) return 100.f;
-	if (KeyState(SDLK_RSHIFT)) return 10.f;
+	if (KeyState(SDLK_LALT)) return 10.f;
+	return 1;
+}
+
+float Manager::GetRotateSpeedShiftModifier()
+{
+	if (KeyState(SDLK_LSHIFT)) return 10.f;
+	if (KeyState(SDLK_LALT)) return 0.1f;
 	return 1;
 }
--- a/src/Input.cpp
+++ b/src/Input.cpp
@@ -857,6 +857,12 @@ void Manager::DispatchEvents()
 
 */
 
+bool Manager::IsMouseRotatePressed()
+{
+	return MouseButtonState(SDL_BUTTON_MIDDLE) ||
+	       ((keyModState & KMOD_LCTRL) && MouseButtonState(SDL_BUTTON_LEFT));
+}
+
 void Manager::SetCapturingMouse(bool grabbed)
 {
 	// early-out to avoid changing (possibly) expensive WM state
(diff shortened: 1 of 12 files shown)


bszlrd:
Having pan on MMB would be nice

mwerle:
> **Discussion:** Given that the RMB moves the ship itself, would this be the more natural default to spin everything instead of the MMB? That could free up the MMB for panning/moving in the SectorView and SystemView.

... although that still leaves the question of which combo to use for rotating the camera (not the ship) in the ShipView.. I guess it could stay on MMB..?

<- not a UX designer, I can do the technical implementation but open to suggestion on which actual buttons/modifiers to use for the various tasks  (which were summarised in a "Details" pop-up here : https://github.com/pioneerspacesim/pioneer/pull/5594#issuecomment-2491710476) ..

bszlrd:
I'd say that can stay on MMB. In Blender at least MMB is always the primary viewport navigation: in 3D view it rotates the view, in 2D windows it pans, and I don't think I ever got confused over that. 
Although admittedly, when switching to other software, especially to Krita sometime I do get confused sometimes, and try to use shift+MMB for pan, which spins the view.

Web-eWorks:
> Given that the RMB moves the ship itself, would this be the more natural default to spin everything instead of the MMB? That could free up the MMB for panning/moving in the SectorView and SystemView.

MMB should "generally" be rotate view, with the WASD keys (and/or left stick on controller, equivalent on joystick) serving to move the "map cursors" in Sector/System map view. I find it more helpful to think about the maps as having a discrete cursor (at the center of the view) which is moved rather than moving by panning the view.

Specifically this is because the map cursors in both maps should move on (some sort) of plane, whether that's the equatorial / ecliptic plane or a custom plane for the specific view (i.e. close zoom on planets wraps around the surface via lat/lon).

src/Input.cpp
```diff
@@ -846,6 +857,18 @@ void Manager::DispatchEvents()
 
 */
 
+// Mouse rotation is enabled if:
+// MMB is pressed, or Ctrl+LMB is pressed, AND no other mouse buttons are pressed
+bool Manager::IsMouseRotatePressed()
```
Web-eWorks, line 862: I am hesitant to accept this. I do not think (philosophically or practically) that it is the role of the `Input::Manager` to define the conditions for rotating the view across the entire application. I see `Input::Manager` as being *significantly* lower-level and this being the responsibility of a higher layer of the application.

Additionally, note that implementing it at this level provides no ability for the user to rebind which modifier key is used when emulating the middle mouse button (perhaps they'd like to use Alt or the meta keys?)

I wish I had an easy solution I could suggest to accomplish this. The mouse rotation code is one of those things that is common to a number of views, but specific enough in scope it's not really a good fit to throw in `Pi`, `Input::Manager`, or `View`. It's game-specific code, but there's a little bit of overlap with the editor as the editor also uses "game" widgets.

mwerle: > I am hesitant to accept this. I do not think (philosophically or practically) that it is the role of the `Input::Manager` to define the conditions for rotating the view across the entire application. I see `Input::Manager` as being _significantly_ lower-level and this being the responsibility of a higher layer of the application.
> 
Ok, just seemed like a natural/easy place for this functionality. It can be placed somewhere else instead. I'm very loathe to re-implement it in every place where it's needed though. I very much dislike code duplication and it already chafes me that I have to duplicate between these views and the Ship Spinner..

> Additionally, note that implementing it at this level provides no ability for the user to rebind which modifier key is used when emulating the middle mouse button (perhaps they'd like to use Alt or the meta keys?)
> 
Why not? I haven't gotten around to adding options for rebinding yet but I wasn't aware of there being a technical issue with adding such.

Web-eWorks: > I'm very loathe to re-implement it in every place where it's needed though.

Same. I do think it should be centralized somewhere (perhaps a static method on some sort of helper namespace?) but just not in the low-level input subsystem.

> Why not? I haven't gotten around to adding options for rebinding yet but I wasn't aware of there being a technical issue with adding such.

The "no ability" referred to the as-implemented state. In terms of potential implementations, it's not so much that there is a technical problem with it, but rather that the API becomes significantly messier as the `Input::Manager` now has to export a specific action binding that all code which wants to use the function needs to ensure is present in the InputFrame stack before calling the method.

fluffyfreak: Could this be moved to a higher level like the `Pi::` namespace where we already use things like `Pi::renderer` throughout the code?

mwerle: > I am hesitant to accept this. I do not think (philosophically or practically) that it is the role of the Input::Manager to define the conditions for rotating the view across the entire application. 

~~Re-opening this conversation, the (probable) reason I chose to put the code here initially was that `Input::Manager` already has very similar code, such as `GetRotateSpeedShiftModifier` and `GetMoveSpeedShiftModifier`.~~

~~These do a very similar thing - set the speed shift modifiers for the entire application! So if these are centralised here, why not the mouse rotation? And if these shouldn´t be here either, then where?~~

LOL - turns out I put those in there as well, so never mind... the below question stands though.

---- 

Instead of me spending time on an alternate implementation which may have similar objections, I´d like some guidance on how to go ahead with this.

sturnclaw: Had a think about this, as I don't want to leave it spinning forever. The best design I could come up with that didn't cram a bunch of high-level functionality in the low-level input interface is this:

- Move `GetXXXShiftSpeedModifier` and `IsMouseRotatePressed` to a new `Input::Helper` namespace.
- That namespace has an `Init()` function which takes an `Input::Manager *`.
- That namespace provides the function `InputFrame *GetModifierFrame()` which returns an InputFrame containing all of the bindings necessary for rebindable modifier support.
- It is the responsibility of external code to ensure the modifier frame is present on the input frame stack before calling the above getter functions.

Both MMB and Ctrl+LMB should be implemented as InputActions and added to the above (likely static storage duration) InputFrame. The input system was not written with the idea of actions being shared between InputFrames in mind (there are some deinitialization semantics which preclude this at current), so it's best to have a single InputFrame for the common helper bindings.

sturnclaw: Addenda: ideally the Input::Helper namespace should be in a new file and not `Input.h`. It is the intent that this maintains a clear semantic separation between engine-level input handling (`Input::Manager` et. al.) and application specific utility/helper code. It is intended that `Input.h` will be moved into the `core` library at some point and thus should remain pure of non-engine code as much as possible.

mwerle: Since I have no idea what an InputFrame is or the architecture of how this all hangs together, is this something specific to Pioneer or is it something from SDL / elsewhere which I can read up about to get a better overall understanding?

sturnclaw: This is specific to Pioneer - it's one part of the input subsystem and is (partially) documented in Input.h.

TL;DR: InputFrames are the organizational method for input actions/axes, and serve as a simple grouping/priority system allowing for contextual input. For example, InputFrames are the mechanism which allows the player to bind the same key to a gameplay input as well as a generic radial menu interaction and have the key trigger two different, exclusive actions depending on whether the radial menu is open (read: its InputFrame is on the top of the stack).

PR #5982, opened 2024-11-23
Fixing icons.svg after the radar updates

bszlrd:
Fixing the icons.svg I provided for the radar PR, where I put the new icons for that onto an older version of the file. The nice thing is that the free space isthe same in both, so there's no need to do anything in the theme, if I'm reading it correct.
The icon file with the error:
image
Previous good version:
image
This PR:
icons


Web-eWorks:
Yep, the new icons are in the same slots in both versions so this is a simple fix. Thanks!

PR #5983, opened 2024-11-25
Fix circle rendering in HyperspaceCloud

balovess:
- Corrected the calculation of the last vertex in the circle rendering function to ensure it lies on the circle's circumference.
- Added a default parameter 'stepAngle' to the 'make_circle_thing' function to allow for customizable circle granularity.
- Ensured that the circle is closed by adding the first point again at the end of the vertex array.

### DIFF
--- a/src/HyperspaceCloud.cpp
+++ b/src/HyperspaceCloud.cpp
@@ -138,13 +138,21 @@ Ship *HyperspaceCloud::EvictShip()
 	return s;
 }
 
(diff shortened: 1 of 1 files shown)


mwerle (commented):
LGTM, and I, for one, am appreciative of the additional comments.

fluffyfreak (approved):
Code looks good and comments are very welcome, thankyou

Web-eWorks (changes requested):

PR #5984, opened 2024-11-25
Feat/radar improvements

mwerle:
**NOTE:** The code could probably do with some tidying up...

----

1. ensure targeted object remains visible when switching from automatic to manual zoom (https://github.com/pioneerspacesim/pioneer/pull/5978#pullrequestreview-2456782749)
2. add intermediate zoom steps for manual zoom (https://github.com/pioneerspacesim/pioneer/pull/5978#pullrequestreview-2456782749)
3. fix popup disappearing when mouse moves outside of radar area but still over popup
4. improve checking when mouse is over radar area (especially for the 3D scanner mode display)
5. add left-click to set navigation target when clicking on a target "pip". This will show a popup if there are multiple targets on a "pip".
6. add double-click to clear the navigation target

Additionally fix an issue introduced in 678d46e4a1efec765322076090fa54349a8af6ea where a couple of parameters are swapped when calling "PlayMusic" (found due to compiler warning).

**TODO:**

- [ ] code review
- [ ] clean up commits
- [ ] merge

### DIFF
--- a/data/pigui/modules/radar.lua
+++ b/data/pigui/modules/radar.lua
@@ -371,7 +371,7 @@ local function displayRadar()
 			local clicked = ui.mainMenuButton(icon, tt, theme, Vector2(button_size))
 			if clicked then
 				if instrument:isAutoZoom() then
-					instrument:zoomIn()
+					instrument:zoomOut()
 				else
 					instrument:resetZoom()
 				end
--- a/data/pigui/modules/radar.lua
+++ b/data/pigui/modules/radar.lua
@@ -317,6 +317,7 @@ end
 -- between frames. We are trying to ensure that a mouse right-click started and
 -- finished inside the radar area in order to trigger the popup.
 local click_on_radar = false
+local radar_popup_displayed = false
 
 -- display either the 3D or the 2D radar, show a popup on right click to select
 local function displayRadar()
@@ -360,21 +361,29 @@ local function displayRadar()
 			Vector2(center.x + ui.reticuleCircleRadius * 0.9,
 			        center.y + ui.reticuleCircleRadius * 0.7))
 	end
-	if isMouseOverRadar() then
+	if isMouseOverRadar() or radar_popup_displayed then
 		ui.popup("radarselector", function()
 			if ui.selectable(lui.HUD_2D_RADAR, shouldDisplay2DRadar, {}) then
-				toggle_radar = true
+				if not shouldDisplay2DRadar then
+					toggle_radar = true
+				end
 			end
 			if ui.selectable(lui.HUD_3D_RADAR, not shouldDisplay2DRadar, {}) then
-				toggle_radar = true
+				if shouldDisplay2DRadar then
+					toggle_radar = true
+				end
 			end
 		end)
 
 		if ui.isMouseClicked(1) then
 			click_on_radar = true
 		end
+		if click_on_radar and ui.isMouseReleased(0) then
+			radar_popup_displayed = false
+		end
 		if not toggle_radar and click_on_radar and ui.isMouseReleased(1) then
 			ui.openPopup("radarselector")
+			radar_popup_displayed = true
 		end
 		-- TODO: figure out how to "capture" the mouse wheel to prevent
 		-- the game engine from using it to also zoom the viewport
--- a/data/pigui/modules/radar.lua
+++ b/data/pigui/modules/radar.lua
@@ -367,8 +367,14 @@ local function displayRadar()
 	local zoom = 0
 	local toggle_radar = false
 
-	-- Handle keyboard
-	-- TODO: Convert to axis?
+	local window_width = ui.reticuleCircleRadius * 1.8
+	local window_height = radar2d.size * 2
+	local window_pos = Vector2(center.x - window_width / 2, center.y - window_height / 2)
+	local windowFlags = ui.WindowFlags {"NoTitleBar", "NoResize", "NoFocusOnAppearing", "NoBringToFrontOnFocus", "NoSavedSettings"}
+
+	--
+	-- keyboard handling
+	--
 	if keys.radar_zoom_in:IsJustActive() then
 		zoom = 1
 	elseif keys.radar_zoom_out:IsJustActive() then
@@ -382,67 +388,52 @@ local function displayRadar()
 		toggle_radar = true
 	end
 
-	-- Handle mouse if it is in the radar area
+	--
+	-- popup functions
+	--
+	ui.popup("radartargetselector", function()
+		for k,v in pairs(popup_targets) do
+			if ui.selectable(v.label) then
+				onTargetClicked(v)
+				radar_popup_displayed = false
+			end
+		end
+	end)
+
+	--
+	-- mouse handling
+	--
 	local isMouseOverRadar = function()
 		if shouldDisplay2DRadar then
 			local mp = ui.getMousePos()
 			return (mp - center):length() < radar2d.getRadius(radar2d)
 		end
 		return ui.isMouseHoveringRect(
-			Vector2(center.x - ui.reticuleCircleRadius * 0.9,
-			        center.y - ui.reticuleCircleRadius * 0.7),
-			Vector2(center.x + ui.reticuleCircleRadius * 0.9,
-			        center.y + ui.reticuleCircleRadius * 0.7))
+			Vector2(center.x - window_width/2,
+			        center.y - window_height/2),
+			Vector2(center.x + window_width/2,
+			        center.y + window_height/2))
 	end
-	if radar_popup_displayed or isMouseOverRadar() then
-		ui.popup("radarselector", function()
-			if ui.selectable(lui.HUD_2D_RADAR, shouldDisplay2DRadar, {}) then
-				if not shouldDisplay2DRadar then
-					toggle_radar = true
-				end
-				radar_popup_displayed = false
-			end
-			if ui.selectable(lui.HUD_3D_RADAR, not shouldDisplay2DRadar, {}) then
-				if shouldDisplay2DRadar then
-					toggle_radar = true
-				end
-				radar_popup_displayed = false
-			end
-		end)
-		ui.popup("radartargetselector", function()
-			for k,v in pairs(popup_targets) do
-				if ui.selectable(v.label) then
-					onTargetClicked(v)
-					radar_popup_displayed = false
-				end
-			end
-		end)
 
-		if not click_on_radar and not radar_popup_displayed and ui.isMouseClicked(0) or ui.isMouseClicked(1) then
-			click_on_radar = true
-		end
-		if radar_popup_displayed and (ui.isMouseReleased(0) or ui.isMouseReleased(1)) then
+	if radar_popup_displayed then
+		if ui.isMouseReleased(0) then
 			radar_popup_displayed = false
 			click_on_radar = false
 		end
-		if click_on_radar then
-			if ui.isMouseReleased(0) then
-				popup_targets = mouseTargets
-				-- check if we are on a pip and if so, set it as the target
-				if #popup_targets > 1 then
-					ui.openPopup("radartargetselector")
-					radar_popup_displayed = true
-				elseif #popup_targets > 0 then
-					onTargetClicked(popup_targets[1])
-				end
-				click_on_radar = false
-			elseif ui.isMouseReleased(1) then
-				if not toggle_radar then
-					ui.openPopup("radarselector")
-					radar_popup_displayed = true
-				end
-				click_on_radar = false
+	elseif isMouseOverRadar() then
+		if not click_on_radar and ui.isMouseClicked(0) then
+			click_on_radar = true
+		end
+		if click_on_radar and ui.isMouseReleased(0) then
+			-- check if we are on a pip and if so, set it as the target
+			popup_targets = mouseTargets
+			if #popup_targets > 1 then
+				ui.openPopup("radartargetselector")
+				radar_popup_displayed = true
+			elseif #popup_targets > 0 then
+				onTargetClicked(popup_targets[1])
 			end
+			click_on_radar = false
 		end
 		if zoom == 0 then
 			zoom = ui.getMouseWheel()
@@ -456,12 +447,10 @@ local function displayRadar()
 		instrument:zoomOut()
 	end
 
-	-- Draw the radar, radar buttons and info
+	--
+	-- drawing functionality
+	--
 	-- This is in a window so the buttons work and the mouse-wheel is captured
-	local window_width = ui.reticuleCircleRadius * 1.8
-	local window_height = radar2d.size * 2
-	local window_pos = Vector2(center.x - window_width / 2, center.y - window_height / 2)
-	local windowFlags = ui.WindowFlags {"NoTitleBar", "NoResize", "NoFocusOnAppearing", "NoBringToFrontOnFocus", "NoSavedSettings"}
 	ui.setNextWindowPos(window_pos, "Always")
 	ui.setNextWindowPadding(Vector2(0))
 	ui.setNextWindowSize(Vector2(window_width, window_height), "Always")
--- a/data/pigui/modules/radar.lua
+++ b/data/pigui/modules/radar.lua
@@ -58,16 +58,78 @@ local function getColorFor(item)
 	return colors.radarUnknown
 end
 
+-- Return digits from a number
+-- If the number is invalid, it returns 1
+local function getDigits(number, digitStart, digitEnd)
+	if digitStart == nil then
+		digitStart = 1
+	end
+	if digitEnd == nil then
+		digitEnd = digitStart
+	end
+	return tonumber(string.sub(tostring(number),
+	                math.min(digitStart, #tostring(number)),
+	                math.min(digitEnd, #tostring(number))))
+end
+
+-- Generate the next in a sequence of 1 2 5 10 20 ...
+local function nextZoomSeq(currZoom, maxZoom)
+	local newZoom = 0
+	if getDigits(currZoom) == 2 then
+		newZoom = currZoom * 2.5
+	else
+		newZoom = currZoom * 2
+	end
+	return math.min(newZoom, maxZoom)
+end
+
+-- Generate the previous in a sequence of 1 2 5 10 20 ...
+local function prevZoomSeq(currZoom, minZoom)
+	local newZoom = 0
+	if getDigits(currZoom) == 5 then
+		newZoom = currZoom / 2.5
+	else
+		newZoom = currZoom / 2
+	end
+	return math.max(newZoom, minZoom)
+end
+
+-- Normalize a number into the zoom sequence (1, 2, 5, 10, 20, ...), rounding
+-- it up or down to the next item in the sequence (default: down).
+local function normalizeToZoomSeq(number, up)
+	if up == nil then
+		up = false
+	end
+
+	local firstDigits = getDigits(number, 1, 2)
+	if firstDigits > 50 then
+		firstDigits = up and 10 or 5
+	elseif firstDigits == 50 then
+		firstDigits = 5
+	elseif firstDigits > 20 then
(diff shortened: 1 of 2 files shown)


data/pigui/modules/radar.lua
```diff
@@ -305,26 +390,67 @@ local function displayRadar()
...
+				radar_popup_displayed = false
 			end
 			if ui.selectable(lui.HUD_3D_RADAR, not shouldDisplay2DRadar, {}) then
-				toggle_radar = true
+				if shouldDisplay2DRadar then
+					toggle_radar = true
+				end
+				radar_popup_displayed = false
+			end
+		end)
```
mwerle, line 418: Q: Now that we have the icon button to switch the radar between 2d and 3d modes, do we still want to support the right-click popup?  The code is a bit messy and I don't really see the utility.

data/pigui/modules/radar.lua
```diff
@@ -58,16 +69,78 @@ local function getColorFor(item)
...
+	elseif firstDigits > 10 then
+		firstDigits = up and 2 or 1
+	else
+		firstDigits = 1
+	end
+
+	local numDigits = #tostring(number)
+	return firstDigits * 10^(numDigits-1)
+end
+
```
mwerle, line 133: These functions are possibly of general use and could be placed into a utility library?

data/pigui/modules/radar.lua
```diff
@@ -304,27 +390,78 @@ local function displayRadar()
...
+	local isMouseOverRadar = function()
+		if shouldDisplay2DRadar then
+			local mp = ui.getMousePos()
+			return (mp - center):length() < radar2d.getRadius(radar2d)
+		end
+		return ui.isMouseHoveringRect(
+			Vector2(center.x - ui.reticuleCircleRadius * 0.9,
+			        center.y - ui.reticuleCircleRadius * 0.7),
+			Vector2(center.x + ui.reticuleCircleRadius * 0.9,
+			        center.y + ui.reticuleCircleRadius * 0.7))
```
mwerle, line 431: Should probably move the window size definitions (below) to the top, and re-use the sizes here.

data/pigui/modules/radar.lua
```diff
@@ -269,10 +354,10 @@ radar3d.draw = function(self, center)
...
--- between frames. We are trying to ensure that a mouse right-click started and
--- finished inside the radar area in order to trigger the popup.
-local click_on_radar = false
+local function onTargetClicked(target)
+	-- TODO: Should ships be set as nav or combat targets?
+	Game.player:SetNavTarget(target.body)
+end
 
 -- display either the 3D or the 2D radar, show a popup on right click to select
 local function displayRadar()
```
mwerle, line 363: Ideally I'd like to split this function into multiple functions, but that would require moving a LOT of variables into the file scope. Is there another way?

What is the best practise in Lua for doing this?

data/pigui/modules/radar.lua
Web-eWorks: Note that the `normalizeToZoomSeq` function can be generalized by taking the log10 of the zoom and snapping the fractional part to one of 0, ~0.3, or ~0.7 (for 1, 2, 5 respectively). The particular values should likely be precomputed as `math.log10(2)` etc. as they have a significant number of digits.

I'll leave the particulars of the implementation to you, but that's how I'd approach the problem.

data/pigui/modules/radar.lua
Web-eWorks: Depends - what are you trying to split out of the function? You have a few options:
1. Pass read-only parameters via function arguments, granularize function actions such that they can return "changed values" as needed. This allows one "master function" to own the local functions and pass them around. If needed, the master function can pass a table value wrapping multiple variables into sub-functions, but that's basically option 2:
2. Convert the radar display into a module (see `data/pigui/libs/module.lua`, and how it's used in e.g. the save/load window) and implement most of your drawing functions as methods on the "radar module". If you do this route, I'd strongly recommend attempting to make your "display" functions non-mutating and use the built-in `message()` facilities to handle state changes.

data/pigui/modules/radar.lua
```diff
@@ -269,10 +354,10 @@ radar3d.draw = function(self, center)
 	end
 end
 
--- This variable needs to be outside the function in order to capture state
--- between frames. We are trying to ensure that a mouse right-click started and
--- finished inside the radar area in order to trigger the popup.
-local click_on_radar = false
+local function onTargetClicked(target)
+	-- TODO: Should ships be set as nav or combat targets?
+	Game.player:SetNavTarget(target.body)
```
Web-eWorks, line 359: Ships are always combat targets.

data/pigui/modules/radar.lua
```diff
@@ -304,27 +390,78 @@ local function displayRadar()
...
+			Vector2(center.x + ui.reticuleCircleRadius * 0.9,
+			        center.y + ui.reticuleCircleRadius * 0.7))
+	end
+
+	if radar_popup_displayed then
+		if ui.isMouseReleased(0) or ui.isMouseReleased(1) then
+			radar_popup_displayed = false
+			click_on_radar = false
+		end
+	elseif isMouseOverRadar() then
```
Web-eWorks, line 439: Note that instead of calling `isMouseOverRadar()`, you can instead potentially submit a `ui.invisibleButton` with the appropriate button flags (might need to expose a few more in `src/lua/LuaPiGui.cpp`) before drawing the radars and test `ui.isItemClicked()` / `ui.isItemHovered()` etc.

This would allow for each radar to define its own "click area" and potentially unify some of the "clicked on radar" handling.

src/sound/SoundMusic.cpp
```diff
@@ -50,7 +50,7 @@ namespace Sound {
 			current = &m_eventTwo;
 			next = &m_eventOne;
 		}
-		next->PlayMusic(name.c_str(), m_volume, repeat, fadeDelta, current);
+		next->PlayMusic(name.c_str(), m_volume, fadeDelta, repeat, current);
```
Web-eWorks, line 53: Good catch, thank you!

Web-eWorks (commented):
> add double-click to clear the navigation target

General thought - it might be better to have a "reject/unlock target" keybinding which unsets combat then navigation targets in that order. Nav/combat target is generally a state which is managed (by the player) at a higher level than the radar widget.

data/pigui/modules/radar.lua
mwerle: Muscle-memory for right-click popups is pretty poor anyway..

Thanks, so unless someone else is going to step in with a strong argument "for" I'll remove it.

data/pigui/modules/radar.lua
mwerle: I'll have a look, although I daresay floating-point arithmetic is going to result in rather imprecise results. I guess performance is moot since it's not going to be calculated every frame..

mwerle: That said, the core question here was whether these functions should be part of a utility library, and if so where, instead of being in "radar.lua"..

data/pigui/modules/radar.lua
mwerle: I was thinking input handling and actually drawing the radar are two rather different logical tasks which should be split.. so yeah, I guess I can put the entire state of the radar module into a table and pass that around (I forgot about tables..).

It's a shame the 3d radar is partially in C++ otherwise it would be much neater to refactor this thing.  If/when we upgrade to the next version of ImGUi we could move this into Lua as well since the next version of ImGui has native ellipse drawing primitives.

I'm thinking the next step will be to actually define multiple pieces of radar equipment and then have functionality depending on what equipment is installed - range, target identification, etc.  So any refactoring here should take this into account. Or possibly require further refactoring, in which case, might as well leave this as-is for now and leave the refactoring to when the next step is done.

Thoughts?

src/sound/SoundMusic.cpp
mwerle: > Good catch, thank you!

It helps to compile with BOTH clang and gcc... :)

We -should- set "-Wall, -Weverything, -Werror", at least for release builds, except the "contrib" folder.

No excuse to have a warning in our code. If the code can really not be modified to fix a warning (compilers aren't infallible either), can always set a #pragma with documentation explaining why.

data/pigui/modules/radar.lua
mwerle: Can you draw circular buttons? And how will this invisible button affect the other buttons because I don't think I'll be able to make an elliptical button!

impaktor:
I suspect most players don't even know about the 2D vs 3D radar.

bszlrd:
Yeah, that button will definitely help with the discoverability

data/pigui/modules/radar.lua
mwerle: > Ships are always combat targets.

Not really - rescue missions, for example.

mwerle:
> I suspect most players don't even know about the 2D vs 3D radar.

I didn't.. (but then again I'm newly back to Pioneer..)

data/pigui/modules/radar.lua
sturnclaw: I generally think they should stay in Radar. We can move them out to a utility library if needed, but I don't see a concrete usecase for them anywhere else at this time.

My general philosophy is we only promote things to a utility library once we've identified one or more other areas of the code which have a current or closely-planned future need for that functionality.

data/pigui/modules/radar.lua
sturnclaw: > Or possibly require further refactoring, in which case, might as well leave this as-is for now and leave the refactoring to when the next step is done.

I'd prefer to leave this as-is - future expansions to Sensor gameplay are going to need significant buy-in from both C++ code and game design, so it's better to leave it "easy to upgrade" rather than try to anticipate our future needs before we've even figured out what those are yet.

data/pigui/modules/radar.lua
sturnclaw: All interaction primitives are based on rectangles, including the window background behind the radar which is blocking mouse events from reaching the rest of the game code - so no elliptical buttons.

The overlap case definitely complicates things. Handling it the way you're currently doing it is "probably" simpler - while I could potentially add an invisible elliptical button primitive, it's simpler to just fake it at this level.

data/pigui/modules/radar.lua
mwerle: So I don't have enough maths background to figure out your suggestion. I'm not particularly proud of my implementation, but it was trivial to write and is easy to understand for anybody.

data/pigui/modules/radar.lua
mwerle: Removed in e717c283c9012a11f037e919848e2ea590ea6242

data/pigui/modules/radar.lua
mwerle: Fixed in 88cc619fbcde8c5b82620aece290ce7b8e3139d6

data/pigui/modules/radar.lua
mwerle: Leaving as-is for now..

data/pigui/modules/radar.lua
mwerle: Fixed in 0665cc57934c34ee990dd7c0eedabe5691bce352

mwerle:
> > > add double-click to clear the navigation target
> > 
> > 
> > General thought - it might be better to have a "reject/unlock target" keybinding which unsets combat then navigation targets in that order. Nav/combat target is generally a state which is managed (by the player) at a higher level than the radar widget.

Reverted deselecting nav-target by double-click in 77e96f9179b5e8e294e0d3e6d56813ee2f8f0ada

data/pigui/modules/radar.lua
mwerle: So let's leave it for next version of ImGui and/or refactoring when/if radar equipment comes along?

data/pigui/modules/radar.lua
mwerle: FWIW, I tried ChatGPT (after the discussion yesterday) to come up with a solution, and it was absolutely horrendously hopeless, constantly recommending the same "solution" which wasn't even close to being correct. As per my expectations really.

Pattern prediction systems are not "AI" nor do they have even the closest "understanding" of anything asked of them. Rather interesting toys, and for sure have some application in teasing out patterns in large data sets, but I would never rely on their correctness..

mwerle: > I'll leave the particulars of the implementation to you, but that's how I'd approach the problem.

I had a (brief) look, and I'm afraid I really have no idea how to use logs to generate this sequence.  My approach is perhaps not the best, but it works, and everybody can trivially understand it by reading the code. 

FWIW, I tried asking ChatGPT and it kept providing the same broken solution no matter how much I tried to tell it it was wrong. It eventually realised it was wrong, but couldn't come up with anything else either.
Pattern prediction systems are not "AI", nor do they have any "understanding" of a problem space. They "simply" extend based on previous patterns. Cute toys, and somewhat useful in teasing out patterns out of large data sets, but definitely not something I would consider using for real work. At least not without very carefully reviewing whatever it spits out.  As expected, really.

Addendum: just had a look at installing copilot into VSCode, and I am _PISSED_.  They use open-source public repo's for training their models, but don't offer a free version for open-source contributors? I wonder if we can/should amend our open source licenses to include restrictions for tools based off language models trained on the sources..

data/pigui/modules/radar.lua
sturnclaw: Sounds good! I mentioned this originally just to ensure you were aware of the alternate method :smile:

sturnclaw:
> EDIT: Some additional testing - if both Combat and Nav targets are set, 'Y' will disable the Combat target first, and then the Nav target. So this already (mostly) works as you suggest - but if there is a target over the reticule, it is set as the target instead (combat if ship, nav otherwise).

I'd like the deselect part of this to be duplicated with an additional keybind for "Unlock Target" (in addition to tapping the "Lock Target" keybind over empty space) - is that something you'd like to tackle, or should I make an issue for later work?

mwerle:
> > EDIT: Some additional testing - if both Combat and Nav targets are set, 'Y' will disable the Combat target first, and then the Nav target. So this already (mostly) works as you suggest - but if there is a target over the reticule, it is set as the target instead (combat if ship, nav otherwise).
> 
> I'd like the deselect part of this to be duplicated with an additional keybind for "Unlock Target" (in addition to tapping the "Lock Target" keybind over empty space) - is that something you'd like to tackle, or should I make an issue for later work?

I can do it. Do you think it belongs in the radar module and this PR though?
I also want to add keybinds to toggle through all radar targets...  at the moment there's only a keybind to toggle through "hostiles". Could be a tad tricky because toggling through ships ("potential hostiles") is different than nav targets - and should nav targets be restricted to "dockables"?  It probably requires a bit more thought and is a bit beyond the scope of what I was aiming for here.

data/pigui/modules/radar.lua
sturnclaw: I've pushed a commit that implements this - feel free to amend or leave as is.

I definitely agree regarding Copilot, it's something I'm not happy with at all.

sturnclaw:
> I can do it. Do you think it belongs in the radar module and this PR though?

Probably not. I had only brought it up because I thought you might want to ensure the functionality to deselect an explicitly-selected target existed when the PR was merged, though if you want to tackle it in a follow-up PR please feel free.

data/pigui/modules/radar.lua
mwerle: Does the ship you've targeted as a "Combat Target" instead of as a "Nav Target" "know" how you targeted it?

Ie, does it get an alert "weapons lock" and go into combat-mode itself? Or is this purely a distinction in the player ships' HUD/systems?

mwerle:
> > I can do it. Do you think it belongs in the radar module and this PR though?
> 
> Probably not. I had only brought it up because I thought you might want to ensure the functionality to deselect an explicitly-selected target existed when the PR was merged, though if you want to tackle it in a follow-up PR please feel free.

It already exists.. pressing 'Y' (default) deselects whatever has been selected.   I personally would prefer to have a mouse-option on the radar given there's now a mouse option to select a target on the radar as well, but I understand that this is not the core functionality of the radar.

<- not a UI/UX engineer..


Anyway, let's punt this into a future PR..  should really do a major review of the UI/UX as a whole, define what functionality we want, and then come up with a unified control scheme.. _ideally_ everything should be usable from keyboard/joystick with mouse optional.. (future VR support..?) But I think currently we are pretty far away from that goal.

data/pigui/modules/radar.lua
sturnclaw: Not yet. This is something I want to dive into more deeply in a sensors rework - adding different types of passive information your own sensors can pick up (radar emissions, transponder signals, ship drive plumes, IR/EM flares from missile launches, etc), and giving NPC ships an understanding of them.

With that rework, nav targets would likely be something your computer maintains passively from sensor inputs, with a separate class of targets that your active sensors are maintaining a "trackfile" on, and finally what we currently call the combat target being the primary trackfile on which your sensors are providing good position, velocity, and aspect estimations for firing solutions.

I would also like it to tie into a rework of missiles and countermeasures, introducing different guidance methods and corresponding countermeasure methods to attempt to defeat missile guidance and enemy sensors.

...but that's a topic for another day. Feel free to add this to the dev docs (with your own thoughts on the matter) if you so desire.

data/pigui/modules/radar.lua
mwerle: Thanks. It works and is more concise, but, at least for me, it's not intuitive or understandable. I guess I should re-learn logs sometime.

sturnclaw: I will say, I don't consider myself as someone who has a strong background in math (or any background, for that matter) - I wish I could eloquently explain *how* the above code does what it does, but I only have an intuitive grasp and not an expository grasp of the math involved. Here's an attempt at a long-form explanation:

Fundamentally, the code works via the definition of a logarithm: if $x = b^y$ then the equation can be rewritten as $y = log_b x$.

Since our "primary" zoom steps (1, 10, 100, 1000, etc.) are powers of ten (i.e. $10^0, 10^1, 10^2$), taking the logarithm of them gives us an integer value corresponding to the order of magnitude; for example, $log_{10}10^3 = 3.0$.

Happily, this means that taking the base-10 logarithm of all numbers between 10,000 and 100,000 gives a decimal result in the range `[3..4)`. Similarly, 100,000 to 1,000,000 gives a decimal result in the range `[4..5)`, and so on.

Each secondary zoom step is always the same regardless of primary zoom step; the only difference is how many zeros follow them (2, 5, 20, 50, etc.). This means that when taking the base-10 logarithm of each secondary zoom step, it will always have the same fractional part regardless of the primary zoom step; only the integer part will change based on which primary zoom step we're in.

The rest of the math is simple at this point - we separate the fractional and integer parts of the logarithm, pick the right fractional part which gives us the proper zoom snap, recombine, and raise 10 to the power of our new exponent.

I've found [this resource](https://tutorial.math.lamar.edu/Classes/Alg/ExpAndLog.aspx) to be an extremely helpful source for understanding exponents/logarithms - it's quite comprehensive and not dumbed-down like most "Let's Learn Math"-style sites.

Note that all of the above math I outlined can be used in *any* base, not just base-10 - base-10 just so happens to be the most common for display-to-human applications.

sturnclaw: More concisely, this is exploiting the property of logarithms that $log_b(xy) = log_bx + log_by$. Since we functionally have three zoom steps (1, 2, 5) which are multiplied by a power-of-ten distance, we can separate those two components by taking the base-ten logarithm of the final zoom number.

sturnclaw (approved):
Good work! Thank you for putting up with my pedantic review as always, and for fixing that issue in Sound!

sturnclaw:
Per discussion on IRC, the PR needs some method of deselecting the selected target via the radar interface - the easiest to discover method is to select the current target again (c.f. the behavior of labels in the WorldView when you click the targeted object again). As the diff is small, no re-approval is necessary before merge.

PR #5985, opened 2024-11-26
Noise detail texture sampling for gas giant rings

fluffyfreak:
This is a port of Luomo's ring detail shader to latest codebase. I think he last worked on it 11 years ago and I stumbled across it during a trip down memory lane earlier today.

This uses a noise texture to multiply the colour of the ring and fade it in with distance from the eye to the ring.

Pictures are better than words so.
SaturnsRingsNoisey
SaturnsRingsNoiseyObjectViewer


fluffyfreak:
The 2 commits for this can definitely be squashed btw

bszlrd:
I think that was introduced with @Mc-Pain 's new ring shader.

fluffyfreak:
That was tricky to work out. It's because the light position is no longer normalised in the C++ code, so we have to do it in the shader before multiplying it with the `uViewMatrixInverse`. This ends up also being normalised and you need both the input too it, and the output from that multiplication to be normalised.

So shadows cast on the rings are back 👍

sturnclaw:
General thought: use a second octave sample from the noise texture to break up the obvious "noise" look of the fine details. Either a separate channel in the noise texture with e.g. perlin noise, or just sample the noise texture again at a larger scale and multiply with a low intensity.

fluffyfreak:
Not quite sure what you mean, but that's probably because I'm tired tonight. I'm happy with how it looks as is since it's just some extra detail for as you whizz past the rings. Happy to accept commits to the branch though 👍

sturnclaw:
I'll take a look at it then - my main problem is just the obvious look of "this is a noise texture" when you're close. There should be a few ways to break that up a bit and make it look a bit more natural.

sturnclaw:
Here is a perhaps slightly more improved version of the ring shader...

image

Comparison video (note - uses a very slightly darker version of the "new" shader than the final screenshot above): 

https://youtu.be/Lyx64gOF8Jk

Technical notes: this uses multiple octaves from the provided white-noise texture, with parameters tweaked to provide a fairly nice-looking general noise pattern. I've expanded the visibility range significantly (from 2 Mm to 15 Mm) and used exponential falloff to make the "noisy" section blend in better with the overall ring, as well as increased the brightness of the low density sections of the noise pattern.

I also added the density term to the Mie reflectance function to slightly reduce the amount of backscattered light - overall, the ring appeared brighter when looking away from the sun rather than towards the sun; this simply limits that brightness to only high-density areas of the ring.

sturnclaw (approved):

fluffyfreak:
That is an amazing improvement!!!

PR #5991, opened 2024-12-05
Fully support installing multiple weapons per ship

sturnclaw:
Newly-necromanced from the grave of its 2021 implementation returns the GunManager - a complete replacement of the legacy FixedGuns code. This new implementation:

- Supports more than a single weapon per ship per direction
- Allows you to separate weapons into groups<sup>1</sup>
- Handles per-hardpoint gimbal limits
- Can track each individual weapon's aiming point and temperature separately<sup>1</sup>
- Supports rear-facing, side-facing, and any-other-facing weapons<sup>1</sup>
- Is ready to support visible weapons on ship models<sup>2</sup>
- And can be extended to support additional weapon types (i.e. ballistics / rocket launchers) and even turrets in the future.

I've been talking about this for quite a long time, so I don't think I need to provide a wall of text for the reason this PR was written. This PR represents one of two major changes needed to release the equipment rewrite on the upcoming Pioneertag - the other being the combined fuel-and-hyperdrive rework.

This implementation brings us to feature-parity with the existing FixedGuns system, but feature-enablement work will have to be done in the future to fully harness all of the features of the implementation.

TODO:
- [x] Implement serialization code for the GunManager component
- [x] Thoroughly test hyperspace travel and save/load code paths

Tagging fluffy/mwerle for code review on what's completed while I'm finishing up the serialization code.

---
1: The GunManager supports this - the UI and the rest of the code still does not.
2: Since we don't have any weapon models other than a single placeholder, I've stubbed out the necessary placeholders and will fully implement weapon rendering in 2025.


src/Quaternion.h
```diff
@@ -30,21 +30,42 @@ class Quaternion {
 		y = axis.y * sinHalfAng;
 		z = axis.z * sinHalfAng;
 	}
-	// from axis, assume angle == PI
-	// optimized fast path using sin(PI/2) = 1
-	Quaternion(vector3<T> axis)
+	// Create quaternion from normalized direction vectors.
+	// This creates a quaternion representing the rotation from the second
+	// unit direction vector to the first unit direction vector.
+	Quaternion(vector3<T> to, vector3<T> from)
```
mwerle, line 36: Any particular reason we're passing by-value instead of by-const-ref here?

sturnclaw: This particular constructor is circa 2021... that'd be why. I'll fix that.

src/ship/GunManager.h
```diff
@@ -0,0 +1,222 @@
...
+private:
+
+	// Handle checking and firing a given gun.
+	// Note that this currently does not nicely handle spawning multiple projectiles per timestep - i.e. timewarp or a weapon RPM higher than 3600
+	// Projectile spawns are also "snapped" to the start of a timestep if they are not direct multiples of the timestep duration
+	void Fire(WeaponState *weapon, GroupState *group);
+
+	// Calculate the position a given gun should aim at to hit the current target body
+	// This is effectively the position of the target at T+n
+	void CalcWeaponLead(WeaponState *state, vector3d position, vector3d relativeVelocity);
```
mwerle: The `vector3d` parameters should be const refs, not value.

src/ship/GunManager.h
```diff
@@ -0,0 +1,222 @@
...
+		//TODO: integrate this with a whole-ship cooling system
+		float coolingPerSecond = 14.2; // nominal amount of thermal energy removed per second (kW)
+		float overheatThreshold = 280; // total amount of thermal energy(kJ) the gun can store while functioning
+
+		std::string modelPath; // model to render this weapon with
+
+		ProjectileData projectile; // deprecated, to replace with RefCountedPtr<ProjectileDef>
+		ProjectileType projectileType = PROJECTILE_BALLISTIC;
+		uint8_t numBarrels = 1;		 // total number of barrels on the model
+		bool staggerBarrels = false; // should we fire one barrel after another, or both at the same time?
```
mwerle, line 86: What are all these numbers for (as in, the comments say what the variable is, but these numbers look like the concrete implementation of a specific weapon...?)

src/ship/GunManager.h
```diff
@@ -0,0 +1,222 @@
...
+	void LoadFromJson(const Json &jsonObj, Space *space);
+
+	void StaticUpdate(float deltaTime);
+
+	// ==========================================
+
+	// Add a weapon mount to this gun manager
+	void AddWeaponMount(StringName id, StringName tagName, vector2f gimbalLimitDegrees);
+	// Remove a weapon mount from this gun manager. This will fail if a weapon is still mounted.
+	void RemoveWeaponMount(StringName id);
```
mwerle, line 140: How can the caller know if this failed? Should it return a `bool`?

src/ship/GunManager.h
```diff
@@ -0,0 +1,222 @@
...
+
+	// Attach a weapon to a specific mount
+	bool MountWeapon(StringName hardpoint, const WeaponData &data);
```
mwerle, line 143: What's the bool return value for?

(Reading the code, it fails if either the hardpoint doesn't exist, or if a weapon is already mounted..)

src/ship/GunManager.cpp
```diff
@@ -0,0 +1,519 @@
...
+void GunManager::StaticUpdate(float deltaTime)
+{
+	bool isAnyFiring = false;
+
+	static ImGuiOnceUponAFrame thisFrame;
+	bool shouldDrawDebug = m_parent && m_parent->IsType(ObjectType::PLAYER) && thisFrame;
+	if (shouldDrawDebug)
+		ImGui::Begin("GunMangerDebug", nullptr, ImGuiWindowFlags_NoFocusOnAppearing);
+
+	if (shouldDrawDebug) {
```
mwerle: Duplication of check..
1. Combine checks
2. Style suggestion (safety) _always_ use braces for compound statements

src/ship/GunManager.cpp
```diff
@@ -0,0 +1,519 @@
...
+}
+
+void GunManager::AssignWeaponToGroup(uint32_t numWeapon, uint32_t group)
+{
+	if (numWeapon > m_weapons.size())
+		return;
+
+	if (group >= m_groups.size()) {
+		m_groups.resize(group + 1);
+	}
```
mwerle, line 172: Are we really this constrained by memory? Seems horribly inefficient to increase/decrease the weapon groups by 1, and, realistically, how many groups are we going to need? At most as many as there are weapons on the ship, or could a weapon belong to multiple groups?

I could imagine a player wanting weapons good against shields in one group, good against hull in another, or "all", for example.

In either case, it might make sense to pre-allocate the array. Conversely I guess this is not going to be called in a performance-critical codepath?

src/ship/GunManager.h
```diff
@@ -0,0 +1,222 @@
...
+	void SetTrackingTarget(const Body *target) { return SetGroupTarget(0, target); }
+	void SetAllGroupsFiring(bool firing = false);
+
+	// For player weapon management
+
+	// Assign a specific weapon to a firing group
+	// Weapons are by default all assigned to group 0
+	void AssignWeaponToGroup(uint32_t numWeapon, uint32_t group);
+	// Remove a weapon group in its entirety
+	void RemoveGroup(uint32_t group);
```
mwerle, line 168: Are groups just integers? Seems a player might want a more descriptive name..
(Haven't looked at the later commits yet, apologies if this is handled in LUA)

src/ship/GunManager.cpp
```diff
@@ -0,0 +1,519 @@
...
+
+void GunManager::SetGroupTarget(uint32_t group, const Body *target)
+{
+	if (group >= m_groups.size())
+		return;
+
+	GroupState &gs = m_groups[group];
+	// No weapons at all, ignore
+	if (!gs.weapons.any())
+		return;
```
mwerle, line 230: Given groups appear to get pruned when the last weapon is removed from them, is this check required in the current implementation?

(I'll discuss separately about design decisions regarding empty groups etc)

mwerle (commented):
Going to stop the (nitpick) review for now.  (and it also teaches me not to do per-commit reviews.. something that I never really did before, but a colleague of mine swears by! Apologies for some completely unwarranted comments..)

Please check all functions:
* whether they are in the correct class
* parameter signatures (in general, use references / const refs in preference to pointers and pass-by-value for all non-base types) - I'm guessing a lot of this code is going to be running per-frame and even if not it's just good modern C++.
* wrap all compound statements in braces (safe coding practice) - the current PR uses a mix of styles sometimes wrapping single-line compounds in braces, and sometimes not. But let's at least be consistent even if my suggested best-practise is not consistent with the codebase :)

Functionally I guess it works (sorry haven't tried it myself yet); great job!  Sounds like space combat is about to get a heck of a lot more deadly!!

---
Weapon Groups (Design Questions):
1. Should groups be nameable at the API level, or is this something for a future UI?
2. Groups appear to be very dynamic at present; a player may like a fixed set of groups, in a fixed order, even if for a particular ship/loadout that doesn't make sense. So the player should be able to keep a group even if it is empty, but have it be disabled. Very useful for certain joysticks which have sliders selecting between multiple set points. Might be more relevant once players can own multiple ships and/or rapidly switch between different loadouts.

src/ship/GunManager.cpp
mwerle: Actually removed in a later commit.

Note to self: don't do per-commit reviews :P

bszlrd:
Do you need UI mockups for it? I'd imagine there will be a gun group management interface somewhere. F3 infoscreen perhaps?

src/Quaternion.h
mwerle: ...?  this has been part of C++ since .. forever. 

Since C++03 (more so since C++11 with move semantics) there are some performance benefits passing a complex type by value if you're going to make a copy of it anyway, but apart from that you would always use (const) references for parameters.

C++ is absolutely horrendous in its complicated minutely different details; there's a reason I generally prefer working in C. For example, [there's a book](https://www.amazon.com/Initialization-Story-Through-Options-Related/dp/B0BW38DDBK) on the different ways to initialise variables in C++; it runs to 295 pages or so (no joke). The equivalent book for C would be 1 page. 2 at most.

fluffyfreak: Yeah but not all of it was written by sturnclaw, it's just being resurrected. Please don't be so harsh in review, some of our contributors are learning the language(s) used.

src/Quaternion.h
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2008-2024 Pioneer Developers. See AUTHORS.txt for details
+// Copyright © 2008-2023 Pioneer Developers. See AUTHORS.txt for details
```
fluffyfreak, line 1: Just a little backslide

src/Quaternion.h
mwerle: My apologies, I asked whether there was a reason and the answer was just "2021".. that didn't make sense to me. The rest of my comment was basically acknowledging that C++ is an incredibly complex beast which few can hope to be reasonable at, let alone master (at best I myself fall into the former category; certainly not into the latter).

mwerle: ..and, not a real excuse but background, I've recently been doing a lot of code reviews at work which is _extremely_ strict and I'm obviously not managing to tone it down enough for hobby-level here. Again, my apologies. No offence or overt criticality meant.  There should probably have been a smiley somewhere in my comment to make it more obvious that it wasn't meant that harshly.

fluffyfreak: No worries, just didn't want critique to distract from a good solid review 👍

Gliese852: I would like to add that in the disassembler (x86) I did not see `Quaternion` constructors at all, apparently they are all inlined and optimized.

sturnclaw: > I would like to add that in the disassembler (x86) I did not see `Quaternion` constructors at all, apparently they are all inlined and optimized.

Quaternion was used in an intermediate stage in the GunManager, but is not currently used (as it's more efficient to avoid a round-trip through a Quaternion for the code's specific needs).

src/ship/GunManager.h
sturnclaw: There is some design intent that's missing here.

`WeaponState` is intended to be as close to a POD object as possible, rather than a traditional OO "data-with-methods" class. To that end, I don't want to add any methods to it, and rely upon field initializers rather than a constructor. `CalcWeaponLead` and `GetMountTransform` are internal implementation details of the GunManager, rather than public API of the "WeaponState" object.

I'd push back slightly on the `vector3d` parameters - the final version of this function has a 122-character function signature and is only ever passed temporary variables, so I don't see a significant win here - in fact, the codegen might be worse with pass-by-reference parameters, as accesses to those parameters have to pointer-chase rather than being directly accessed from the stack. This is admittedly post-facto justification for keeping the function signature the same.

I agree with you that it is better to pass the state object by reference than by pointer however, as the pointer is not intended to be nullable.

src/ship/GunManager.h
sturnclaw: These are "sane defaults" that were added as an example during prototyping - I can amend these to be initialized with valid but nonspecific values instead. It is intended that all fields of a `WeaponData` object are filled by external code before it is used by the GunManager, so these defaults are not particularly useful.

src/ship/GunManager.h
sturnclaw: The caller should as a matter of convention always validate that the mount is empty (via `IsWeaponMounted`) before calling `RemoveWeaponMount` - as a consequence, I don't think returning an explicit indication of failure here is necessary. This comment is meant to indicate the above precondition rather than indicate that the function is meant to be called in an "exploratory" manner.

src/ship/GunManager.cpp
sturnclaw: Note that this is meant to handle the case where there is one extant group, and the caller decides that weapon 3 should be assigned to group 6.

> realistically, how many groups are we going to need?
> ...
> I could imagine a player wanting weapons good against shields in one group, good against hull in another, or "all", for example.

Group management is purposefully meant to be handled at a higher level than the GunManager. This API provides a low-level interface to assign weapons to numeric groups and control each group's target and firing state separately. There is a built-in convenience "shim" for operating on group 0 for ease of porting existing AI code, but the GunManager otherwise does not make any assumptions about how groups are going to be used or handled.

> At most as many as there are weapons on the ship, or could a weapon belong to multiple groups?

Each weapon can belong to at most one group - this is the main constraint on the system. I anticipate that higher-level code may want create a new group for a weapon on installation, but only for the player's ship.

> In either case, it might make sense to pre-allocate the array. Conversely I guess this is not going to be called in a performance-critical codepath?

Since the GunManager can't know how many weapons or groups will be used, it's not particularly helpful (or possible) to pre-allocate an array. Groups are also intended to be set once then forgotten; at most, one group reassignment is likely to happen per frame. Also note that the cost of `resize` is likely to be amortized as `std::vector` grows its internal storage based on a heuristic.

src/ship/GunManager.h
sturnclaw: Per my comment above, groups are a low-level construct. Any names, etc. will be handled at a much higher level (as will sorting of weapons). The intended design is something like Elite Dangerous or Starsector; see this example from E:D (note that at current I plan to only support a single fire button):

src/ship/GunManager.cpp
sturnclaw: This is technically unneeded, though I've not fully decided whether I want automatic group removal to remain a thing or if it should be an explicit action on the user's part.

sturnclaw:
@mwerle
> Weapon Groups (Design Questions):
>
>    Should groups be nameable at the API level, or is this something for a future UI?
>    Groups appear to be very dynamic at present; a player may like a fixed set of groups, in a fixed order, even if for a particular ship/loadout that doesn't make sense. So the player should be able to keep a group even if it is empty, but have it be disabled. Very useful for certain joysticks which have sliders selecting between multiple set points. Might be more relevant once players can own multiple ships and/or rapidly switch between different loadouts.

- I'm unsure if groups are something that will ever receive names, but certainly not at the GunManager level. Per my rationale in another response, the GunManager's group implementation is intended to be very low level and dynamic; what the player interacts with will definitely be augmented at a higher level with additional data to display to the user.
- Empty / disabled groups are definitely something I'm open to supporting - that's part of why I kept a number of checks in the code to see if a group is empty, even though default behavior right now is to prune empty groups.
- Setting groups is intended to be something the player does after changing weapons, and it's meant to support most configurations of weapons the player would like.

These are very useful concerns that I don't intend to tackle at this juncture yet. The focus of this PR is to get to feature parity with the existing FixedGuns implementation, and I intend to defer explicit support for player-assignable fire control groups to after the upcoming release (there's not enough time to do everything on the TODO list!)

@bszlrd 
> Do you need UI mockups for it? I'd imagine there will be a gun group management interface somewhere. F3 infoscreen perhaps?

I'd always love UI mockups, you know this. :smile: A UI for fire control groups is something I don't plan to implement until after Pioneertag, but I was likewise thinking about sticking it on the ship info screen. Feel free to make a mockup if you have inspiration - if the improbable happens and I have extra free time I might be able to squeeze it in before mid-January feature freeze.

src/ship/GunManager.h
sturnclaw: Added a clearer comment as to the meaning of the return value.

src/Quaternion.h
sturnclaw: Fixed!

sturnclaw:
Serialization is implemented (ugly thing that it is), and I've addressed most or all of the review feedback received so far. Please continue to review (and test!) - this branch should be feature-complete at this point and just needs a more strenuous test regimen than I've given it so far.

src/ship/GunManager.h
mwerle: Are the WeaponState and other structures expected to be used outside of GunManager?  Looks to me like they should be internal state only.. in which case moving some of the functionality into the PODs makes sense to me to reduce the ultra-tight coupling.  (I see it's currently directly accessed in Ship.cpp, but marked as "abominable" and furthermore, used for introspection; it might be better to move the weapon sound playback into the GunManager, or expose relevant accessors to the GunManager to provide the required high-level information back to the Ship class).

But maybe there are future plans for more functionality around this data and it's not desired or practical to wrap all of this data into accessors; that's something I can't know.

And yes, if `CalcWeaponLead` is only ever called with temporaries, calling-by-value is actually better.. but this is very difficult to know as a reviewer reviewing a function in isolation without a comprehensive design document outlining these decisions..

src/ship/GunManager.h
mwerle: Ok, so this API is somewhat inconcsistent. Why does MountWeapon() return a bool instead of not also requiring a prior call to IsWeaponMounted()?

And if these really are preconditions and a convention which must be followed, they should be documented, and the convention enforced by asserting the preconditions to prevent misuse by other developers instead of just silently failing.

src/ship/GunManager.cpp
mwerle: I see; most ships will be AI so only have a single group, and the player will only setup groups rarely.

> Each weapon can belong to at most one group - this is the main constraint on the system. I anticipate that higher-level code may want create a new group for a weapon on installation, but only for the player's ship.

I can see this being a potential future limit; like I said, a player may want a "all" group for a full volley, for example. But possibly this can be expanded on in future if we decide we want it?

src/ship/GunManager.h
mwerle: Ok, that was my secondary question whether the naming would be handled in a higher layer, thank you.

A single fire button? We currently have Primary and Secondary already..?

src/ship/GunManager.cpp
sturnclaw: > I can see this being a potential future limit; like I said, a player may want a "all" group for a full volley, for example. But possibly this can be expanded on in future if we decide we want it?

I should perhaps explain a bit more what I mean with "low level" - the player's "active weapons" state and trigger inputs are not meant to map 1:1 with the firing state of weapon groups in the GunManager. It's entirely possible to have a "volley fire" option with multiple weapon groups (internally it'd dispatch a fire command to all groups) or to have separate sets of weapons bound to different fire buttons, or have autonomous / auto-fire weapon groups with separate targeting.

The primary reason for the single-group constraint is so there's no need for a "priority system" to determine which group's inputs (firing, not firing) should be used to determine whether a weapon is firing or not.

TL;DR: Responsibility for dispatching input to individual fire-control groups is intentionally left to a higher-level layer than the GunManager, specifically so that we can do whatever we want.

src/ship/GunManager.h
sturnclaw: I'll put a boolean return value on this then, as it is a bit inconsistent. The convention is documented in the (updated) function comment, and the only reason the function returns rather than asserting is because a failure to follow the conventions is a recoverable error and asserts are strongly non-recoverable.

src/ship/GunManager.h
sturnclaw: We have a "trigger" input and a "weapon release" input (Gun / Missile) - while the system can certainly be expanded to support multiple gun triggers (or trigger stages), I'd much prefer to handle that at a higher level than the GunManager. It's easier to have four sets of groups for four fire inputs than it is to have four bitfields for every single group.

src/ship/GunManager.h
sturnclaw: Note that the design intent regarding these specific data structs is something akin to C-with-namespaces, rather than "modern C++" - state objects do not typically mutate themselves, and have no "agency" of their own. They are instead acted upon by other code external to them, and at most might have helpers to ease setting multiple variables or accessors as needed to compute derived variables which are not stored directly.

mwerle: In general, mixing development styles in the same project should be highly discouraged.

I've nothing against a C-with-namespaces approach, but then that should be universally applied. IMHO.

Do we have an overarching style guide which future code should attempt to adhere to? This would make it easier for both contributors as well as reviewers.

src/ship/GunManager.h
mwerle: It becomes a question of whether you want to enforce a convention or not.  If you do, then an assert is the preferred approach as that forces code to adopt the convention instead of relying on reviewers to discover any transgressions. If you don't, then having the capability for client code to gracefully recover is essential. 

And, playing the devil's advocate, if these 2 functions return bool, then they all should if they can fail to complete their intended action.

The worst (IMHO) are silent failures; they'll take forever to discover and debug when some player reports that "something" isn't working.

src/ship/GunManager.cpp
mwerle: ...okay.  Then I'm not entirely sure what the point of these low-level weapon groups are if the higher level code is expected to implement it's own grouping mechanism for the purposes of firing and button assignments etc. Sorry; colour me confused.

src/ship/GunManager.h
sturnclaw: > The worst (IMHO) are silent failures; they'll take forever to discover and debug when some player reports that "something" isn't working.

Agreed. I've modified the return value and added some extra validation in `Ship.lua` when adding weapon mounts.

src/ship/GunManager.cpp
sturnclaw: These low-level groups are the building blocks of future higher-level "weapon behavior". If we (for example) wanted to have a system such that button A fires all long-range laser weapons, and button B fires all close-range ballistic weapons, that would be implemented by dispatching fire/stop-firing commands to separate groups. If we want a simple single-active-group system controlled by a single button, it'd be a thin shim over the existing functionality. If we decide that the player should manage each weapon separately, then we just assign weapons unique group indices and pretend like a weapon+group is the same "thing".

The system is implemented in this way at this time to defer any need to commit to a specific design for handling player fire-control to the future, once we've put some thought into the best way to handle it.

...my explanations are perhaps more confusing than they should be at this point - in short, it's best to consider it as something that will be revisited once we have a much more concrete understanding of our design requirements.

src/ship/GunManager.h
```diff
@@ -0,0 +1,228 @@
...
+	void SaveToJson(Json &jsonObj, Space *space);
+	void LoadFromJson(const Json &jsonObj, Space *space);
+
+	void StaticUpdate(float deltaTime);
+
+	// ==========================================
+
+	// Add a weapon mount to this gun manager.
+	// Returns false if a hardpoint already exists on this GunManager with the specified name.
+	bool AddWeaponMount(StringName id, StringName tagName, vector2f gimbalLimitDegrees);
```
mwerle, line 142: Nitpick: parameters should be const-ref's?  (ditto all other functions taking StringName parameters?)

(Going by JsonUtils.cpp using const-ref's for StringName types)

src/ship/GunManager.h
mwerle: I was going by the in-game configuration screen for weapons; didn't look into the code to see how those were implemented/handled. Sorry.

I'm confused and perhaps misunderstood or talking at cross-purposes..  but what does having the number of fire buttons have to do with the implementation in GunManager? I thought you mentioned the single fire-button restriction as a direct comment on the E:D interface..

src/ship/GunManager.h
sturnclaw: > In general, mixing development styles in the same project should be highly discouraged.
> I've nothing against a C-with-namespaces approach, but then that should be universally applied. IMHO.

I agree with this. In fact, my intent is not to use completely different development styles in different parts of the code, but rather to use different facets of the same development style where they are needed and where they make sense.

Perhaps the "C-with-namespaces" is a red herring - the important part I was attempting to get at is the distinction between data being acted upon (mutated) by an external authority, and an operation intrinsic to the data itself. An operation in the former category is extrinsic and IMO a better fit to be located in an external actor (i.e. class), while an operation in the latter category (e.g. taking the dot product of a vector) is better located in the data itself.

> Do we have an overarching style guide which future code should attempt to adhere to? This would make it easier for both contributors as well as reviewers.

Unfortunately no. We have a bit of a franken-style both at the high level and at the individual code syntax level. The closest I can point at is the general philosophy of data-oriented design, with some of the "only use as much C++ as you need" mentality of Orthodox C++.

src/ship/GunManager.h
sturnclaw: In this case, the const-refs are to satisfy the function prototype of `nlohman::to_json`, used for implicit object<->JSON conversion. `StringName` itself is technically intended as a pass-by-value object, as it has small-string optimization to avoid memory allocation and copies only increment an internal refcount even for large strings.

...with that being said, I can (and might) convert these to const ref parameters as the refcount-copy operation does technically have a small fixed cost.

src/ship/GunManager.h
sturnclaw: > I'm confused and perhaps misunderstood or talking at cross-purposes.. but what does having the number of fire buttons have to do with the implementation in GunManager? I thought you mentioned the single fire-button restriction as a direct comment on the E:D interface..

I did mention that as a direct comment on the screenshot posted - there's a bit of topic-wandering on my end.

The mention of the number of fire buttons related to the implementation was meant to elucidate my thoughts on how we could implement (potentially multiple) simultaneous weapon-fire buttons (i.e. Primary, Secondary, Tertiary, Missile Release; or Fire Group 1, Fire Group 2, etc.) without having to make the GunManager responsible for the number of weapon inputs the player will use.

src/ShipAICmd.cpp
```diff
@@ -434,12 +447,12 @@ bool AICmdKill::TimeStepUpdate()
...
 		vissize += (0.05 + 0.5 * leaddiff) * Pi::rng.Double() * skillShoot;
 		if (vissize > headdiff)
-			m_fguns->SetGunFiringState(0, 1);
+			m_guns->SetAllGroupsFiring(true);
 		else
-			m_fguns->SetGunFiringState(0, 0);
-		float max_fire_dist = m_fguns->GetGunRange(0);
+			m_guns->SetAllGroupsFiring(false);
+		float max_fire_dist = stats.range;
 		if (max_fire_dist > 4000) max_fire_dist = 4000;
```
mwerle, line 454:
```suggestion
		float max_fire_dist = std::min(stats.range, 4000.0f);
```
(It took me a few seconds to figure out why the local variable was needed in the first place; this makes it marginally more obvious IMHO)

I know this is not part of this PR, but why are we clamping a weapon range to some arbitrary maximum range here instead of just trusting the weapon stats?

src/ship/GunManager.h
mwerle: Thank you for your patience answering my variously rambling and discombobulated thoughts.

src/ship/GunManager.cpp
mwerle: I see; my assumption was that it was a given that weapon assignment to groups would be fully player-controlled. I didn't realise that wasn't decided yet.

Your last explanation clears this up and I (think I) understand the aim of this level implementation now.

src/ship/GunManager.h
mwerle: The WeaponManager has a simple interface : `Set(All)GroupFire()`.  Up to the UI to manage which button triggers this..?  

Anyway; this comment is getting extremely derailed here!!

The original question was answered long ago, thank you.

src/ShipAICmd.cpp
mwerle: Ah, fair, it's just for AI ships? (I should probably look at the filename lol)  So player-ship-weapons are not affected by this clamping; gotcha.

Interesting, so flying 4km's from a target and taking pot-shots at distance guarantees a safe fight..  this requires further testing!  (Perhaps it was deemed too difficult/unfair for AI ships to have perfect aim from so far away..) Then again, AI should have a skill modifier, which adds a skill-scaled amount of jitter to the aiming. (No idea whether or not it does, just saying)

PR #5992, opened 2024-12-06
Feat/equipment descriptions

### LINKED ISSUE #5951: Equip-v2 : have flavour text for all equipment
From discussion in #5734  : https://github.com/pioneerspacesim/pioneer/pull/5734#discussion_r1827333285

----

TODO:

GENERAL: All equipment should contain a small blurb about itself, what it does, and some flavour text. Some equipment already has "what it does"

-     Missile racks should display the number/type/size of missiles it can store.
-     Missiles should display their damage, max-range, and manoeuvrability
-     Hyperdrives should display the max jump-distance and fuel-usage (given the current ship)
-     Shields should display protection stats for the given ship
-

mwerle:
**Anybody can contribute to this PR; feel free to add commits**

Even though this PR is a draft, as long as people have no objections with the proposed text, it could be merged at any time.  Specifically the first commit (add missing translations) is more a bug fix than a feature.

This aims to fix #5951 

----

Feature creep:

I've started adding additional missile racks to this PR..  this work -probably- belongs in a new PR but a very quick attempt to cherry-pick the commits into a new PR turned out to be a pain due to the interleaved descriptions. Yes, I can pick it apart, but would there be a huge issue with turning this PR simply into a "missiles and missile racks" PR instead?

----

Some equipment has additional functionality or slots which are not immediately obvious to the player, so it is important to provide an in-game description to allow them to choose the correct equipment as buying and selling loses money (there should probably be a "return" option where a purchased piece of equipment can be sold back at the same price as long as the player has not undocked in the meantime - this could be considered as "browsing").

Then there's also the question of simply adding more flavour to the game by adding some, potentially humorous, descriptions to equipment which might include extravagant advertising from the manufacturer.

- Add missing translations for left- and right- missile racks and pylons
- Add descriptions for all currently defined missile racks and rails
- Add descriptions for all missiles
- Add missile racks up to Slot 5
  -  Design : https://www.notion.so/bszlrd/Databases-1018c82e283f80d29d30e09a1f72f095
- ...?

### DIFF
--- a/data/lang/equipment-core/en.json
+++ b/data/lang/equipment-core/en.json
@@ -363,14 +363,6 @@
     "description": "Equipment description",
     "message": "External missile rack able to carry up to 2 large missiles."
   },
-  "MISSILE_RACK_481": {
-    "description": "Name for a multiple-missile external missile rack. 'Hydri' is a brand name.",
-    "message": "LH-180 Hydri Missile Rack"
-  },
-  "MISSILE_RACK_481_DESCRIPTION": {
-    "description": "Equipment description",
-    "message": "External missile rack able to carry up to 8 small missiles. More fireworks!"
-  },
   "MISSILE_RACK_524": {
     "description": "Name for a multiple-missile external missile rack. 'Hydri' is a brand name.",
     "message": "LH-420 Hydri Missile Rack"
@@ -386,14 +378,6 @@
   "MISSILE_RACK_543_DESCRIPTION": {
     "description": "Equipment description",
     "message": "External missile rack able to carry up to 4 large missiles."
-  },
-  "MISSILE_RACK_582": {
-    "description": "Name for a multiple-missile external missile rack. 'Hydri' is a brand name.",
-    "message": "LH-280 Hydri Missile Rack"
-  },
-  "MISSILE_RACK_582_DESCRIPTION": {
-    "description": "Equipment description",
-    "message": "External missile rack able to carry up to 8 medium missiles."
   },
   "MISSILE_RAIL_S1": {
     "description": "Name for a single-missile external launcher. 'Cnida' is a brand name.",
--- a/data/lang/equipment-core/en.json
+++ b/data/lang/equipment-core/en.json
@@ -325,7 +325,7 @@
   },
   "MISSILE_RACK_322": {
     "description": "Name for a multiple-missile external missile rack. 'Hydri' is a brand name.",
-    "message": "LH-230 Hydri Missile Rack"
+    "message": "LH-220 Hydri Missile Rack"
   },
   "MISSILE_RACK_322_DESCRIPTION": {
     "description": "Equipment description",
--- a/data/lang/equipment-core/en.json
+++ b/data/lang/equipment-core/en.json
@@ -307,6 +307,10 @@
     "description": "",
     "message": "Guided Missile"
   },
+  "MISSILE_GUIDED_DESCRIPTION": {
+    "description": "Equipment description",
+    "message": "A guided missile able to home in on your target using advanced LIDAR."
+  },
   "MISSILE_NAVAL": {
     "description": "",
     "message": "Naval Missile"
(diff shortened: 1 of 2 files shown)


data/lang/equipment-core/en.json
```diff
@@ -307,6 +307,10 @@
     "description": "",
     "message": "Guided Missile"
   },
+  "MISSILE_GUIDED_DESCRIPTION": {
+    "description": "Equipment description",
+    "message": "A guided missile able to home in on your target using advanced LIDAR."
```
impaktor, line 312: "home" -> "hone"

data/lang/equipment-core/en.json
```diff
@@ -315,34 +319,66 @@
...
+    "description": "Equipment description",
+    "message": "External missile rack able to carry up to 2 small missiles."
+  },
   "MISSILE_RACK_322": {
     "description": "Name for a multiple-missile external missile rack. 'Hydri' is a brand name.",
     "message": "LH-230 Hydri Missile Rack"
   },
+  "MISSILE_RACK_322_DESCRIPTION": {
+    "description": "Equipment description",
+    "message": "External missile rack able to carry up to 2 medium-sized missiles. Bigger bangs for your bucks!"
```
impaktor, line 332: I'd like the sillyness to be in the sales-pitch. I assume this string is for tooltip?
I'm all for silliness, but we've usually (always?) kept it out from the tooltip.

data/lang/equipment-core/en.json
```diff
@@ -315,34 +319,66 @@
...
+    "description": "Equipment description",
+    "message": "A smart rocket capable of tracking down your target using an advanced AI module."
+  },
   "MISSILE_UNGUIDED": {
     "description": "",
     "message": "R40 Unguided Rocket"
   },
+  "MISSILE_UNGUIDED_DESCRIPTION": {
+    "description": "Equipment description",
+    "message": "A generic dumb-fire rocket mass-produced on most worlds. Simply aim and fire to solve your problems."
```
impaktor, line 380: Same. I'm not saying to remove it, I'm just opening it up for discussion.

data/lang/equipment-core/en.json
mwerle: "home in" -> target or approach a destination
"hone" -> sharpen, eg, a knife..

data/lang/equipment-core/en.json
mwerle: This is the description on the equipment-purchase page. It's also the tooltip when hovering over the equipment on the same page. (I discovered that this is used for a tooltip after seeing it used for the description..)


image

data/lang/equipment-core/en.json
mwerle: *sigh*  .. although it appears that, as usual, there is some language-mangling going on across the Pond and "hone in" is starting to see accepted usage over there.  (and yes, I will brutally mock Americans for their corruption of the English language; not that the British can lay any claims to their sophistication of spelling their language in the first place; I mean, Worcestershiresauce (woostersheersoz) anyone???)

mwerle: Please tell me that you do do not not subscribe to the Oxford Comma? There is absolutely no way for a computer programmer to make logical sense of a list which does not use it.. apart from that I have no idea what the difference between the two is.

fluffyfreak: Obviously I prefer British English 😁 go for "home in" 👍

mwerle:
I've merged the commits from this PR into #5993 ; seems a bit silly to manage separate PR's, and it was getting a bit confusing trying to keep in mind what was changed here and what was changed there since really, it's pretty closely-related work.

I can easily split it apart again if preferred.

data/lang/equipment-core/en.json
sturnclaw: "Home" / "Homing" is the correct verb to use. In this specific technical context, "hone" is entirely incorrect, as the process refers to guiding the missile to the source of some emission - i.e. Home-On-Jam.

data/lang/equipment-core/en.json
sturnclaw: IMO the tooltip / short description should ideally be a single sentence - at most two, if they're short.

"External missile rack able to carry up to 2 medium-sized missiles." <- Impartial description of what the item is/does; should be used as tooltip / item description.

"Bigger bangs for your bucks!" <- Colorful flavor text, product reviews, information about the manufacturer or item's history. Generally should not be displayed as part of the tooltip.

mwerle: Fixed in #5993 

The "_DESCRIPTION" is used for equipment tooltips and to start off the description blurb in the side-panel now. The "_FLAVOURTEXT", if present, is displayed underneath that and can add anything additional.

mwerle: FWIW, I'm not a huge fan of the tooltips hovering over everything.. I think I'd prefer the hovering to auto-update the right-hand panel instead..

sturnclaw: @bszlrd thoughts? IMO we'd need to either change the background color or in some way indicate that the stats shown are from the hovered item and not the selected item, or it could get confusing quickly.

bszlrd: I wouldn't change anything on the right panel on hover. I think that is not good UX, makes things a bit unpredictable, jumpy. And now you would have to keep the  mouse on the item to see its data.
Do we even need a tooltip in this case?

mwerle: Personally I find the tooltips a bit annoying as they tend to hover over the top of the information on the card which I _do_ want to see.  But I also see the point of the tooltips in being able to quickly browse through a list and get some basic information without having to click on each item.

But at the end of the day, the player will have to click on the item to either purchase it, or to look at the detailed info which isn't in the tooltip anyway.

I've seen quite a few UIs which use the detail-pane-on-hover-over-list-item approach which is why I mentioned it.  If you want to look at an item in more detail, you can just click on it to actually select it and "pin" it to the detail pane - my thinking was that the selected item is the one displayed unless the player hovers over something else.

It looks like it could be implemented pretty easily. But I see the point on it being potentially a "jumpy" and disorienting experience for people not used to this kind of interface.

bszlrd: I agree, tooltips hiding the stuff under them is not good either. I never found that detail pane approach you mention good. They feel like an artifact of console UX, where you don't have a proper mouse, so you have to do these odd gymnastics. 
But on the other hand, it could very well work, I just can't imagine it. How complex it would be to prototype it for testing?
Or how about having a hover portion of that screen that shows it, instead of a tooltip?

Another approach I could imagine, but not fully sure about:
Expanding the equipment card downrads to show the description:
equip-hover-1
equip-hover-2
It would still be jumpy in a way, but I not excessively I think. The expansion should only be one line at most, so the height change upon going over a new card wouldn't make it too unpredictable. With keyboard navigation it wouldn't make a difference. And maybe there vould be a very quick animation to it to make it easier on the eyes. I think there are games having a similar UI solution, but I can't recall any on top of my head.
Not sure how complex it would be from the implementation side though.

PR #5994, opened 2024-12-16
Display how many instanced objects are drawn

fluffyfreak:
Extend the information shown about Instanced Draw Calls to include the total number of instances actually drawn by them.
image

I'm not sure about the formatting of the display `80 | 813 Instanced Calls` so would welcome feedback.
Maybe it should be `80 (813 total) Instanced` or something? There's not a lot of space available.

### DIFF
--- a/src/graphics/Stats.cpp
+++ b/src/graphics/Stats.cpp
@@ -16,6 +16,8 @@ namespace Graphics {
 
 		m_counterRefs = {
 			GetOrCreateCounter("DrawMesh Calls"),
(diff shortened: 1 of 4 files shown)


sturnclaw:
Ideally something like `813 Instances (80 Draw calls)`. The stat is somewhat meaningless, and just there to ballpark how much work is being done in the scene when diagnosing performance issues.

fluffyfreak:
Done, and yeah I think it looks clearer now 👍

PR #5997, opened 2024-12-24
Fix billboard atlas animation playing order and frame count

### LINKED ISSUE #5881: Atlases in Sfx.ini are played backwards
### Observed behaviour

Image atlases are played from bottom right to up left.

### Expected behaviour

Image atlases should (at least, the default explosion.dds seems to be made to) be played from up left to bottom right.

### Steps to reproduce

- Create a custom image atlas with a clearly different beginning and end, then edit Sfx.ini to match it (here are both of my files for testing: [explosion_custom.zip](https://github.com/user-attachments/files/16472570/explosion_custom.zip)).
- Start Pioneer
- Blow up yourself / another ship

__My pioneer version (and OS):__ Pioneer 2024-03-14 (didn't test the newest version yet, but didn't find anything about this here so i guess it's the same) on Garuda Linux, system last updated today.

fluffyfreak:
Fixes #5881 

It looks like we had the texture atlas selection inverted so it was playing from the "end" of the atlases animation.

I say "inverted" and "end" but these are slightly arbitrary because we define what they should be. However starting at the top-left is a reasonable convention and not an arbitrary point within the texture atlas, progressing to the top-left which is how it was behaving.

This is a handy test atlas with numbers in each slot so you can change `Sfx.ini` to point to it, and then instead of the explosion you can see the numbers counting up.
[test_atlas.zip](https://github.com/user-attachments/files/18241664/test_atlas.zip)

Taking the above test-atlas you could see it counting down from 30 to 0, which I think most people would find odd.
image

There was also an issue with it playing only 31 out of 32 frames of animation from the animation.

Finally I added the ability to press Ctrl+F8 if you're running a dev build with devkeys enables and it will spawn an explosion billboard right atop your ship. It's purely visual but very useful for testing, for example if someone wanted to replace the existing explosion etc, and could be extended to show other effects easily.

I've tested all of this quite a bit, sadly the texture supplied in the original bug by @Miner34dev isn't suitable for use as is because it fades to the black instead of alpha, and then terminates suddenly.

### DIFF
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -783,6 +783,12 @@ void Pi::HandleKeyDown(SDL_Keysym *key)
 	case SDLK_F11: // Reload shaders
 		renderer->ReloadShaders();
 		break;
+
+	case SDLK_F8: // EXPLOSION!
+	{
+		SfxManager::AddExplosion(Pi::game->GetPlayer());
+		break;
+	}
 #endif /* DEVKEYS */
 
 #if WITH_OBJECTVIEWER
--- a/src/Sfx.cpp
+++ b/src/Sfx.cpp
@@ -194,6 +194,8 @@ void SfxManager::Add(const Body *b, SFX_TYPE t)
 
 void SfxManager::AddExplosion(Body *b)
 {
+	if (!b) return;
+
 	SfxManager *sfxman = AllocSfxInFrame(b->GetFrame());
 	if (!sfxman) return;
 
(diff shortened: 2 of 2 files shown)


impaktor:
Cool

> Finally I added the ability to press Ctrl+F8 if you're running a dev build with devkeys enables and it will spawn an explosion billboard right atop your ship

Should we document this somewhere on the wiki, or could we have it bound to a button in the debug window?

sturnclaw:
> However starting at the top-left is a reasonable convention

An expected convention, even. This is a good change, thank you. One minor nitpick - the comment in code mentions counting "down", but the PR appears to be changing to counting "up"; am I misreading that?

> Should we document this somewhere on the wiki, or could we have it bound to a button in the debug window?

At your discretion - I'm really not worried about documenting it however, as it's strongly unlikely to be useful outside of extremely niche development/testing cases.

fluffyfreak:
> > However starting at the top-left is a reasonable convention
> 
> An expected convention, even. This is a good change, thank you. One minor nitpick - the comment in code mentions counting "down", but the PR appears to be changing to counting "up"; am I misreading that?
> 
I think it's just me using English badly, as all Englishmen do :D 
In my head I was counting DOWN as being through the images as though they were on a film reel, but really it's UP through the indexes 🤔 

I'll change it to up and make it clear that it is UP through the indexes, thanks

sturnclaw:
In that light, I'll merge as-is, thanks!

PR #6005, opened 2025-01-06
Ship pattern fix

robothauler:
A small fix for the Search & Rescue mission.
I changed also the display for the remaining mission time. In my opinion this looks better than "0 Days" for short missions.


sturnclaw (approved):
Thanks!

PR #6008, opened 2025-01-07
Add four new cultures

zonkmachine:
Add Filipinos, Basques, Albanians and the North Sami to the cultures. Flavorful stuff...

According to cultures/common.lua the ISO code is supposed to be ISO 639-1, which is two letters.
`	code = "xx",   -- ISO ISO 639-1 language code`
Filipino is a sociolect of Tagalog, which has an ISO 639-1 code `tl`, but filipino seem to be more broad, contemporary and fit better to the game and I've used the ISO 639-2. This is also true for Hawaiian from an earlier PR which also doesn't have an ISO 639-1 code.

This was done heavily assisted by AI. The Albanian file was done in one run and I essentially only fixed up the comments and tested it. The cultures have been tested against the latest master separately with a high value of culture weight and the resulting names tested on the net to find real persons with the corresponding culture.


sturnclaw:
I'll leave this one to @impaktor to review. We do (now) have a project-wide stance on AI-generated *code*, though the contents of this PR are strictly speaking data rather than code, and therefore only need to be compliant with CC-BY-SA-4.0 rather than GPLv3.

Considering that cultural names are in the broadest sense public domain, this PR falls into a gray area with regards to that stance. If the PR materially contains enough of **your own work** such that you are able to legally grant the Pioneer project a license to use the contents thereof under the terms of the Creative Commons license, then it is mergeable.

zonkmachine:
> If the PR materially contains enough of **your own work** such that you are able to legally grant the Pioneer project a license to use the contents thereof under the terms of the Creative Commons license, then it is mergeable.

I believe it does but I'll look into it one more time.

zonkmachine:
Made it a draft. The licenses were not as clear as I assumed. It's a bit tricky but I'm on the case.

zonkmachine:
I found this name data project that releases a subset of it's database in GPL-3.0:
https://github.com/Ninble/name-census-top-100

It consists of the top 100 names of male, female and surnames in 139 countries.

bszlrd:
I would appreciate if you used that instead of the genai output. I would really like to wait after things settle around genai before even considering using any output.

zonkmachine:
> I would appreciate if you used that instead of the genai output.

Fixed! That was annoying but all licenses are now clear. Added Armenian.

data/culture/fil.lua
```diff
@@ -0,0 +1,366 @@
...
+    "Michelle",
+    "Monica",
+    "Nicole",
+    "Nikki",
+    "Nina",
+    "Patricia",
+    "Paula",
+    "Pauline",
+    "Princess",
+    "Queen",
```
sturnclaw, line 201: Are these titles or proper Filipino first names? I imagine to an english-speaking audience (or one not familiar with Filipino culture) these might seem a bit weird as a character is known only as "Princess Sarmiento" or "Queen Dimaculangan" (for example).

sturnclaw (approved):
Much better now that the AI authorship issue has been resolved, thank you! The list of names looks pretty good - pending testing and addressing the one review comment, this has my approval.

data/culture/fil.lua
zonkmachine: I just went with the most common names that popped up and didn't censor anything that would sound funny. Not my culture, I know how funny Swedish sounds if you're not from here. Cj is the most far out for me.

> Princess Sarmiento"

https://www.google.com/search?client=ubuntu-sn&channel=fs&q=Princess+Sarmiento

Queen is on the list but doesn't show up as easily as Princess. I can remove Princess or change it for Reyna.
My main method was on testing it in the game and searching the net to see if real persons would pop up.

sturnclaw: Seems they are indeed actual first names - that clears it up, thanks!

sturnclaw:
Tested and merged as impaktor is not available to review at current.

PR #6026, opened 2025-01-17
New hyperdrive algorithms and internal hyperfuel tanks

### LINKED ISSUE #5914: Hyperfuel ussage can be avoided.
### Observed behaviour
Hyperfuel not used.

### Expected behaviour
Hyperfuel is used.

### Steps to reproduce
1. Start a hyperjump.
2. Before the countdown finishes put hyperfuel into propellant tank.
3. Wait for hyperjump to complete and see that the fuel put into the propellant tank is still there (the fuel meant to be used for the jump)

Version: 2024/07/10
OS: Windows 11 23H2

[output.txt](https://github.com/user-attachments/files/16861231/output.txt)
[_exit.zip](https://github.com/user-attachments/files/16861263/_exit.zip)

### LINKED ISSUE #6025: Hyperdrives don't match ship hyperdrive slot scaling
Since we're in the middle of the equipment rewrite, ship hyperdrive slots only go up to size 5 now. The hyperdrive algorithms expect up to 9 classes of ship hyperdrives however, which causes a bit of a problem where big ships can't go anywhere.

The solution is to implement the proposed [replacement hyperdrive algorithms](https://dev.pioneerspacesim.net/design-document/hyperdrive-redesign) which were authored with the equipment refactor in mind. Some tweaking to constants may be required for gameplay balance.

sturnclaw:
Redesigns the hyperdrive fuel use and duration-per-distance-travelled equations to implement the [redesign](https://dev.pioneerspacesim.net/design-document/hyperdrive-redesign) we've discussed and accepted.

This PR also converts hyperdrives to use an internal fuel reservoir rather than pulling directly from the cargo bay, and allows transferring fuel between the main tank and the hyperdrive reservoir. This enables fractional fuel usage in jumps - no longer will jumps of differing lengths use the same amount of fuel!

image

As part of this effort, about 50% of the capabilities of the planned fuel management UI were implemented; the fuel management will be left at this level until we get further time (after the upcoming release) to implement the full management interface.

image

This PR fully implements all five sizes of hyperdrives in both military and civilian variants. Values have been assigned that "should" give decent jump ranges across the ship masses of that size, but a rigorous balance pass has **not** been done in interest of time.

Military drives have not been tested yet, but should function fine - I'd welcome any and all testers willing to give this branch a spin!

TODO:

- [x] ShipBuilder should fill hyperdrive tanks on spawn.
- [x] SAR mission needs to be able to interact with new hyperdrive tanks.
- [x] Military fuel needs to be handled correctly.

Fixes #5914.
Fixes #6025.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -2690,5 +2690,9 @@
   "ZOOM": {
     "description": "Label for a zoom (magnification) control bar.",
     "message": "Zoom"
+  },
+  "TRANSFER_FUEL": {
+    "description": "Label indicating the ability to transfer fuel between equipment tanks",
+    "message": "Transfer Fuel"
   }
 }
--- a/data/libs/EquipSet.lua
+++ b/data/libs/EquipSet.lua
@@ -16,7 +16,7 @@ local utils = require 'utils'
 ---@field New fun(ship: Ship): EquipSet
 local EquipSet = utils.class("EquipSet")
 
----@alias EquipSet.Listener fun(op: 'install'|'remove', equip: EquipType, slot: HullConfig.Slot?)
+---@alias EquipSet.Listener fun(op: 'install'|'remove'|'modify', equip: EquipType, slot: HullConfig.Slot?)
 
 -- Function: SlotTypeMatches
 --
@@ -417,6 +417,18 @@ end
 -- Remove a previously-registered event listener function.
 function EquipSet:RemoveListener(fun)
 	utils.remove_elem(self.listeners, fun)
+end
+
+-- Method: NotifyListeners
+--
+-- Send a notification event to listeners.
+---@param op string The type of event to send.
+---@param equipment EquipType The equipment item generating the event
+---@param slot HullConfig.Slot? The slot handle on this EquipSet that is relevant to the event
+function EquipSet:NotifyListeners(op, equipment, slot)
+	for _, listener in ipairs(self.listeners) do
+		listener(op, equipment, slot)
+	end
 end
 
 -- Method: Install
--- a/data/libs/EquipType.lua
+++ b/data/libs/EquipType.lua
@@ -350,23 +350,48 @@ end
 ---@field byproduct CommodityType?
 local HyperdriveType = utils.inherits(EquipType, "Equipment.HyperdriveType")
 
-function HyperdriveType:GetMaximumRange(ship)
-	return 625.0*(self.capabilities.hyperclass ^ 2) / (ship.staticMass + ship.fuelMassLeft)
+-- Static factors
+HyperdriveType.factor_eff = 55
+HyperdriveType.factor_time = 1.0
+HyperdriveType.fuel_resv_size = 1
+
+-- Travel time exponents
+local range_exp = 1.6
+local mass_exp = 1.2
+-- Fuel use exponents
+local fuel_use_exp = 0.9
+local fuel_use_exp_inv = 1.0 / 0.9
+
+---@param ship Ship
+function HyperdriveType:GetEfficiencyTerm(ship)
+	return self.factor_eff / (ship.staticMass + ship.fuelMassLeft)^(3/5) * (1 + self.capabilities.hyperclass) / self.capabilities.hyperclass
 end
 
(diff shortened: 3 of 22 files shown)


syonfox:
It appears you can still jump with no fuel in the hyper drive tank the button is not disabled from the flight screen.  and when you jump it may make some number negative internally or something

image
image

And the button works but the drive acts funky after sometimes.

Bodasey:
found here, may be also broken in master? Or do i something wrong here? I am not familiar with rescue missions...

Screenshot_20250119_131836
Screenshot_20250119_131857

more reports to come...

Bodasey:
As far, the additional hyperdrive fuel tank works as intended, but i think it is a rather complicated way to fix #5914.

As you can move fuel as you want to, you get, as a player, de facto one single, but now bigger fuel tank than in the original ship design.

Thoughts:

- The hyperdrive machine should get the fuel directly from the existing fuel tank.
- Hydrogen in the cargo space is bottled, while it is not in the fuel tank. (or tank = one big bottle)
- Scooped fuel should also go into the tank (and then may be bottled in the cargo space)
- There are no tanks for military fuel and its radioactive waste (thrusters stop when waste tank is full)
- Rockets also need oxygen as fuel, why don't we?
- Interplanetary flights by autopilot always end with 1/4 fuel left in the tank, can't this be controlled?

Bodasey:
I have completed a need hydrogen rescue mission in the master branch - works there.

Miner34dev:
>     Rockets also need oxygen as fuel, why don't we?

Ion/Fusion thrusters don't need oxygen.

sturnclaw:
> As far, the additional hyperdrive fuel tank works as intended, but i think it is a rather complicated way to fix https://github.com/pioneerspacesim/pioneer/issues/5914.

It is indeed a complicated way to fix 5914, however it was never intended primarily as a fix for that issue. Hyperdrive-internal fuel reservoir and fractional fuel usage has been a planned feature for multiple months, and the fix for the hyperdrive exploit is actually completely unrelated to this new feature.

> There are no tanks for military fuel and its radioactive waste

This is an oversight in the current implementation, and the inability to fuel the drive with military fuel will be fixed before merge. Thanks for catching this one!

As for the rest of the thoughts on hyperdrive implementation, most of those are either orthogonal to this specific PR or something we've already addressed during the design phase for this specific feature (i.e. pulling hyperfuel from the ship's main fuel tank).

syonfox:
https://youtu.be/YrW7xvnbYjo

Looks like you fixed it.  There may be a small secondary bug of the fuel calculations not being rerun when you switch to the galaxy map. 

And I'm curious if that solves the "dispersing button" but will test further and see if I run into it again.

syonfox:
Ok with new code i still got this bug. 

https://youtu.be/jPoh2t4RuEY

Looks like sometimes with no fuel in cargo bay the button is disabled.

sturnclaw:
I'm unable to reproduce either of those bugs with the most recent code. When testing, always make sure to start a new save - there's no guarantee that old saves from the same branch are still compatible (and/or don't contain subtle bugs due to changes in the code).

sturnclaw:
> found here, may be also broken in master? Or do i something wrong here? I am not familiar with rescue missions...

This bug is not endemic to this particular branch. The SAR mission was written when all ships had physically improbable cargo holds that were a significant fraction of the fuel tank size; when that was changed with the equipment refactor, some stranded ships don't have enough cargo space to transfer all fuel to the target's cargo hold before refueling the fuel tank. I've fixed the issue in a commit on this branch.

With the latest batch of commits, this branch *should* be ready for merge. I'll give it a last round of testing, rebase what I can, and if nothing crops up I'm pulling the trigger.

syonfox:
Nice work! 

I had an idea to implement a fuel computer just made a branch but might need some work on the language.  Will make a pull quickly before bed.

syonfox:
> I'm unable to reproduce either of those bugs with the most recent code. When testing, always make sure to start a new save - there's no guarantee that old saves from the same branch are still compatible (and/or don't contain subtle bugs due to changes in the code).

I haven't seen it again in a new save file so might be fixed.

PR #6027, opened 2025-01-18
Improve Naval and Smart missiles

### LINKED ISSUE #4159: Unexpected behaviour from UI buttons: Missiles.
### Observed behaviour

Clicking the missile buttons to fire a missile causes unexpected results, for example if there are 3 types of missiles shown, each having 4 units, firing a missile of one type sometimes causes the displayed number of units of the other types to change, sometimes the buttons disappear. Pressing the m key to fire a missile also results in the buttons disappearing. 

Buttons showing incorrect values does not have any affect on the number of missiles carried by the ship, which can still be fired using the m key.

### Steps to reproduce

1) Load this savegame  
[_quicksave.zip](https://github.com/pioneerspacesim/pioneer/files/1312439/_quicksave.zip)
2) Fire a naval missile at the selected target.
3) The values of the smart and guided missiles change.
4) Fire a missile using the keyboard.

a614c6f 

Linux Mint 18 64-bit

opengl.txt
[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/1312448/opengl.txt)

### LINKED ISSUE #5926: Missile firing during paused game
### Observed behaviour
Missiles can be fired (spawned) when the game is paused.
Instead of seeing a swarm of missiles flying toward the target, there is very high risk of blowing up the player's own ship.

My version of Pioneer:
pioneer-20240710-win

This probably is just an unhappy behavior, rather than a legit bug.
However, I recommend adding pause check like this, in \data\libs\Ship.lua

```
function Ship:FireMissileAt(which_missile, target)
	local missile_object = false
	local game_paused = (Game.GetTimeAcceleration() == "paused") and true or false
	if type(which_missile) == "number" and (not game_paused) then
		local missile_equip = self:GetEquip("missile", which_missile)
		if missile_equip then
			missile_object = self:SpawnMissile(missile_equip.missile_type)
			if missile_object ~= nil then
				self:SetEquip("missile", which_missile)
			end
		end
	elseif (not game_paused) then
		for i,m in pairs(self:GetEquip("missile")) do
			if which_missile == "any" then
				missile_object = self:SpawnMissile(m.missile_type)
				if missile_object ~= nil then
					self:SetEquip("missile", i)
					break
				end
			end
		end
	end
...
end
```

### LINKED ISSUE #6014: Missiles should have thrust/fuel/damage/guidance information set by equipment
Missiles are currently defined by one of four "named missiles", that being `unguided`, `guided`, `smart`, and `naval`. This includes all information about the missile, including health, damage, and guidance.

Obviously, this is suboptimal and provides very little ability for modders or developers to introduce interesting gameplay.

sturnclaw:
Alternative title: finally fixed a bit of the mess that is missile handling in this game.

This PR rewrites how missiles get their "data", i.e. all parameters used in-flight. Missiles still reference a special ShipType file (one of the `missile_*.json` shipdefs), but we're no longer constrained to only four missile types any more!

What's more, Naval and Smart missiles now actually do more damage than regular guided missiles (as one would expect), and missiles in general got a bit better about detonating close to the target rather than at a 100m+ standoff.

As a quick perusal of `Missile.cpp` and `data/modules/Equipment/Weapons.lua` will show, most meaningful parameters for missiles are now defined in "data" rather than code, allowing mods to introduce new missile equipment and making it much easier to tweak missile gameplay balance.

Fixes #5926.
Fixes #6014.
Fixes #4159.

### DIFF
--- a/data/libs/EquipType.lua
+++ b/data/libs/EquipType.lua
@@ -548,6 +548,12 @@ local ThrusterType = utils.inherits(EquipType, "Equipment.ThrusterType")
 
 --==============================================================================
 
+---@class Equipment.MissileType : EquipType
+---@field missile_stats table
+local MissileType = utils.inherits(EquipType, "Equipment.MissileType")
+
+--==============================================================================
+
 Serializer:RegisterClass("EquipType", EquipType)
 Serializer:RegisterClass("Equipment.LaserType", LaserType)
 Serializer:RegisterClass("Equipment.HyperdriveType", HyperdriveType)
@@ -555,6 +561,7 @@ Serializer:RegisterClass("Equipment.SensorType", SensorType)
 Serializer:RegisterClass("Equipment.BodyScannerType", BodyScannerType)
 Serializer:RegisterClass("Equipment.CabinType", CabinType)
 Serializer:RegisterClass("Equipment.ThrusterType", ThrusterType)
+Serializer:RegisterClass("Equipment.MissileType", MissileType)
 
 EquipType:SetupPrototype()
 LaserType:SetupPrototype()
@@ -563,6 +570,7 @@ SensorType:SetupPrototype()
 BodyScannerType:SetupPrototype()
 CabinType:SetupPrototype()
 ThrusterType:SetupPrototype()
+MissileType:SetupPrototype()
 
 return {
 	laser			= laser,
@@ -575,4 +583,5 @@ return {
 	BodyScannerType	= BodyScannerType,
 	CabinType       = CabinType,
 	ThrusterType    = ThrusterType,
+	MissileType     = MissileType,
 }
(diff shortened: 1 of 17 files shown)


impaktor:
> What's more, Naval and Smart missiles now actually do more damage than regular guided missiles (as one would expect),

Is this distinction visible to the player when buying?

fluffyfreak:
This is fantastic work! 👍🎉 It's a very exciting improvement to an ancient system that needed reworking, good job

sturnclaw:
> > What's more, Naval and Smart missiles now actually do more damage than regular guided missiles (as one would expect),
> 
> Is this distinction visible to the player when buying?

Not yet. That's tracked in #6014 for implementation before we release.

syonfox:
Just testing this with both the hyperdrive code and this and when I install a Naval missile rack I get the fallowing error: 

image


[output.txt](https://github.com/user-attachments/files/18467885/output.txt)

image

This happens when i but a rack supporting huge missiles. 

Can reproduce on ac33 as well

sturnclaw:
Completely unrelated to either branch, it's caused by attempting to draw details for a MILITARY tech-level item. Will fix in a separate PR.

PR #6028, opened 2025-01-19
Updating weapon tags to match the ship DB on Notion

bszlrd:
- Updated and added missing gun tags
- Added missile related tags too, based on Notion
- Fixed the Bluenose model being off-center on all lods
- No change to the Xylophis, it does not have a gun

Checked all of them in the model viewer, but please double-check.


sturnclaw:
One problem: the Kanara tags on the model are spelled `lr`, rather than the `tr` of the slot. 

In balance terms, 4x S3 pylons are *very* overkill for a Kanara, 4x S2 pylons or even just plain missile mounts would be fine and it would be quite effective in combat still. I'd recommend changing that on both Notion and in the model if possible (while fixing the misspellings).

Otherwise, you've fixed most of the validation issues - I need to add two new translation strings for the unused turret slots on the Natrix, but we're good to go from there once the Kanara and Varada configs are imported from Notion.

sturnclaw:
I'd recommend rebasing the PR on latest master by the way, you'll have access to the ShipDef validation tool (in Ctrl+I) I wrote that way.

sturnclaw (approved):

sturnclaw:
Everything looks good! I'll finish a final stat pass on the Kanara/Varada/Xylophis and import the shipdefs in a separate PR - thanks for the work (and on short notice too!)

PR #6029, opened 2025-01-22
Add support for new equipment to the game start menu.

Gliese852:
изображение
изображение

The available equipment is shown in a table, as seen in the first pic. If this parameter is unlocked, a "pencil" button appears, which opens a dialog in which you can edit the equipment (second pic).

Also added a "value patch" for the v89 equipment reader to bring it to the new data structure. It doesn't bother with finding the correct names for the slots, as this can be fixed later.  There is a small simple heuristic that corrects some equipment IDs.

When updating data, the menu itself tries to move equipment from invalid slots to valid ones, which improves the results both when recovering the game and when changing the ship through the menu.

Also added a recovery reader for the current version (v91).


sturnclaw:
Excellent news! I'll get this tested and reviewed in the next few days - this is the last major missing piece of the puzzle!

data/pigui/modules/new-game-window/class.lua
```diff
@@ -158,48 +152,55 @@ local function startGame(gameParams)
...
+
+			if prefix then
+				slot = prefix .. slot
+			end
+
+			local proto = Equipment.Get(item)
+			if type(slot) == "string" then
+				local slotHandle = equipSet:GetSlotHandle(slot)
+				if slotHandle then
+					local inst = proto:Instance()
```
sturnclaw, line 186: Need an `if inst.SpecializeForShip then inst:SpecializeForShip(equipSet.config) end` here.

sturnclaw (changes requested):

data/libs/autoload.lua
```diff
@@ -98,6 +98,27 @@ table.merge = function(a, b, transformer)
 	return a
 end
 
+-- Copy table recursively using pairs()
+--
+-- Does not copy metatable
+---@generic T
+---@param t T
+---@return T
+table.deepCopy = function(t)
```
sturnclaw, line 107: For consistency with the Lua standard library conventions, this should be `table.deepcopy`.

sturnclaw (approved):
Overall, I'm happy with the end result - I don't have time to nitpick the implementation, other than two points:

1. As indicated below(?), extensions to the Lua standard libraries should follow Lua's naming conventions, i.e. `squash.wordstogether`.
2. The edit dialog for ship equipment is missing the ability to delete an item from a slot without overwriting the contents of the clipboard. Adding a 'Delete' button next to 'Install' would solve that.

Otherwise, this looks good - thanks for tackling the new game window part of the refactor!

Gliese852:
Squashed fixup commits and ready to merge.

sturnclaw:
Excellent work, thank you!

PR #6031, opened 2025-01-28
Fuel Computer + Some Reusable Fuel Code

syonfox:
Hi its 6am now but i think this is almost working. Fallow up on #6026

What do you guys think of adding a Fuel computer like a Trade computer. 

Especially with small class one hyperdirves it gets tedious to click back and forth to the menu each time to manually transfer fuel. 

Enter the fancy fuel computer that automatically does it when you enter a system for only 450$. 

But it takes the slot of your trade computer ;) 

Now you just set your reserved fuel and warp to your hearts content. 

image

image

I guess there are types and stuff and i need to test mil drives again but should be working. 

anyway though I would get your thoughts on the idea before I sink too much time into it.

### DIFF
--- a/data/modules/FuelComputer/FuelComputer.lua
+++ b/data/modules/FuelComputer/FuelComputer.lua
@@ -1,0 +1,64 @@
+
+
+
+-- Copyright © 2008-2025 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local Event = require 'Event'
+local Game = require 'Game'
+local Engine = require 'Engine'
+local Timer = require 'Timer'
+local Serializer = require 'Serializer'
+local Comms = require 'Comms'
+
+local fuel = require 'pigui.libs.fuel-transfer'
+
+-- store which station sent them out
+local policeDispatched = false
+
+local loaded_data
+
+
+local onGameStart = function ()
+	if (loaded_data) then
+	    fuel.setComputerReserve(loaded_data.mainReserve, loaded_data.cargoReserve)
+	end
+	loaded_data = nil
+end
+
+
+local serialize = function ()
+	local data = fuel.getComputerReserve()
+	return data
+end
+
+
+local unserialize = function (data)
+	loaded_data = data
+end
+
+
+
+local onEnterSystem = function (player)
+    if not player:IsPlayer() then return end
+
+	if (loaded_data) then
+		    fuel.setComputerReserve(loaded_data.mainReserve, loaded_data.cargoReserve)
+	end
+
+	loaded_data = nil
+
+	if(fuel.hasFuelComputerCapability()) then
+	    fuel.computerTransfer()
+	end
+end
+
+
+
+
+Event.Register("onGameStart", onGameStart)
+Event.Register("onEnterSystem", onEnterSystem)
+Event.Register("", onEnterSystem)
+
+
+Serializer:Register("FuelComputer", serialize, unserialize)
--- a/data/pigui/libs/fuel-transfer.lua
+++ b/data/pigui/libs/fuel-transfer.lua
@@ -10,6 +10,8 @@ local Passengers = require 'Passengers'
 local Commodities = require 'Commodities'
 local CommodityType = require 'CommodityType'
 
+local Comms = require 'Comms'
+
 local l = Lang.GetResource("ui-core")
 
 local colors = ui.theme.colors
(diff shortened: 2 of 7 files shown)


sturnclaw:
I can't give this an in-depth review until after the release. I will say however, that we have a full interface planned for managing fuel and also intend to introduce the concept of fuel reservoir equipment (allowing you to extend your fuel tanks/storage) during the next development cycle, so something like this would need to take into account those plans.

I like the concept of automatically transferring fuel into the hyperdrive after a jump, but I think it might not be something we'd introduce a separate equipment item for - rather making it a toggleable option in the fuel management interface we plan to implement.

syonfox:
Sounds good In theory This doesn't have to depend on the equipment item can can be just a saved setting.  Im still testing they military hyperdrive 

It seems we need a solution for when the cargo bay is full there is nowhere to put the wast. and when fractional wast is put in the cargo it shows as 0. Will try and track it down.

syonfox:
Ok so it seam when a fractional amount if mil fule is used it produces one radioactive waste

And if there is no soace then 0t of wast is forced into invintory and you canoot dump it. 

One solution could be to automatically dump one 1 of wast into space (it would be interesting if the polliese were there they caught you for dumping in this occasion.)

Alternitive solition is to add a waste tank for mil-drives the same as the main tank and the user has to pump it out one ton at a time. 

The problem I'm still running into is that for a class 1 drive you cannot refuel from your cargo bay if you have some fractional amount of fuel left.  

An appropriate middle way might be to accept the refuel but "dump the overflow into waste tank"

This would waste fuel but allow you to make a chain of jumps without stopping at a station. 

I think for the current implementation that might be a elagent solution 


```
Add waste tank of same size as hyperdrive tank. 

ON jump pump fuel to waste. 

IF no space in waste cant jump?? (or auto pump down waste)

On refuel overflow excess into waste. 

ie. 
1t in tank and 2t in storage
jump rout cost 0.3, 0.3, 0.5
jump 1 and 2 move 0.6t to waste. 
need to refuel 1t but it wont fit so 0.4t is added to waste conserving mass.
waste tank is full so now dump it to cargo as 1t waste.
perform last jump. 
buy some fuel darn now its not in balance any more. 

Maye the simpaler option is better for a quick fix.


jump 1 and 2 put 2 tones of waste in storage.  user can dump them as needed
refuel tops up tank to 100% or by 1t loseing the extra. 

Im not sure if this is the most realistic but it does provide balance of mil-hyperdrive needing better planing of routes.  and its not so unrealistic if they are standard single use canisters or something.   

```

Anyway will think about it a bit.

syonfox:
Ok I have tested with a class2 mill-drive and things work pretty good.  at least not worse then without the change.  the only TODO is to figure out how to refuel a 1t millitary drive. 

One lazy solution is to make the class1 drive have a 2 ton fuel tank. but this might make the range crazy. maybe this is ok balance.

impaktor:
What I can see, this PR overlaps part of work that is about to be done by the development team, and also you seem to currently still be iterating design decisions in your mind, indicating it's work in progress. 

It would be nice to accept all offered code contributions, especially from new contributors, but I'm assuming this covers aspect of the game that is about to be re-written, as such it might not be viable to be merged any time soon - if at all.

I do like the idea of auto-fueling as upgrade, but would have to take very little space, lest no one would use it. Actually, perhaps this feature is better done by assigning it to the crew?

syonfox:
With that last test this is currently finished if you wish to work off of this I should have made a new issues for the mil-drive fuel usage bug I guess. It is a working refactor of the same logic previosly and the fuel messages need translations but that's about it. The mil stuff seems to still be broken in other areas so I wont worry about it too much.

To summarize. 

Makes fuel clamp function reusable.

Adds fuel computer to enable/ disable

Fuel computer has configurable reserve for cargo and main drive

Adds a reusable fuelStats that standardizes the different random methods internally for a interface. 

I think it would be a good starting place and you should consider this as a simple solution and avoid over engineering it to much. 

To clarify this makes the game more fun without adding a lot of complexity. 

You should test it out.

syonfox:
> but would have to take very little space, lest no one would use it. Actually, perhaps this feature is better done by assigning it to the crew?

Right now its the same as trade computer but lighter.  but could make it smaller. I was thinking the actual pumping system might take some space.

syonfox:
Here are some updates screenshots feel free to take the branch for a spin it currently merges with main no problem. and let me know what you think when you have time 

image


image


image

sturnclaw:
> Right now its the same as trade computer but lighter. but could make it smaller. I was thinking the actual pumping system might take some space.

I would highly recommend perusing the [Fuel Management](https://dev.pioneerspacesim.net/design-document/fuel_management) design doc if you've not already.

data/pigui/libs/fuel-transfer.lua
```diff
@@ -0,0 +1,406 @@
+local ui = require 'pigui'
```
impaktor, line 1: copyright header missing

data/modules/FuelComputer/FuelComputer.lua
```diff
@@ -0,0 +1,64 @@
+
```
impaktor, line 1: 3 empty new lines? + 2026 now

PR #6032, opened 2025-01-30
Fix bugs, expand equipment coverage, add detailed stats for equipment

### LINKED ISSUE #5893: Assassination stack trace and crash when hyperspacing into the target system too late
### Observed behaviour

As I was hyperspacing into the system where the assassination target was exected, a stack trace poped up and the game crashed.
Note that I was late and the target had probably jumped out of the system
image

### Expected behaviour

The game does not raise exception and/or crash

### Steps to reproduce

Just press "Initiate hyperjump" button in the attached saved game

20240710 (Linux)

[output.txt](https://github.com/user-attachments/files/16681586/output.txt)
[assassination.gz](https://github.com/user-attachments/files/16681587/assassination.gz)

### LINKED ISSUE #6015: Add equipment preview stats
Add equipment preview stats to the detailed statistics pane for multiple categories of equipment items:

- Hyperdrives
- Thrusters
- Missiles (pending #6014)

### LINKED ISSUE #6016: Hull autorepair equipment is only available to very large ships
After the equipment rewrite, the existing hull autorepair equipment was assigned to size-4 hull slot. This is rather suboptimal, as it creates a tradeoff between atmospheric shielding for the hull or the ability to attempt to repair combat damage.

At minimum, multiple size classes of hull autorepair should be created to restore the ability of medium and larger ships to access this feature.

sturnclaw:
Another catch-all fix/feature PR. Here's the short list:

- Fixed #5893.
- Fixed #5209 (more specifically, added those launch settings to the example/default `.code-workspace` file we have in source control).
- Converted the cargo scoop to a slot-less equipment item that scales its costs based on the size of the ship hull. This allows the player to be able to install it without having to sacrifice a hull slot (i.e. for atmospheric shielding). This is a compromise based on discussion from IRC - it may be converted back to a slot-based item in the future based on future design decisions.
- Atmospheric shielding and hull autorepair now have appropriately-sized variants available for all or most (in the case of Hull Autorepair) ships. Fixes #6016.
- Fixed #6015.

As part of this PR, I've changed how new "types" of equipment are defined; they now use `EquipType:NewType("Equipment.MyTypeName")` and I've migrated the definitions for individual EquipTypes to live in the `modules/Equipment/Types.lua` file rather than in `libs/EquipType.lua`.


impaktor:
> Converted the cargo scoop to a slot-less equipment item that scales its costs based on the size of the ship hull.

We have (had?) a combined fuel + cargo scoop that saved that precious 1t. Is fuel scoop still taking a slot? No multiscoop anymore?

> As part of this PR, I've changed how new "types" of equipment are defined

@zonkmachine might want to check wiki lua tutorial?

zonkmachine:
> > As part of this PR, I've changed how new "types" of equipment are defined
> 
> @zonkmachine might want to check wiki lua tutorial?

The tutorials I've touched doesn't cover this. Closest tutorial I can think of is this one and it's already outdated:
https://wiki.pioneerspacesim.net/wiki/Lua-based_equipment

I think it's better if someone who's into that code looks into this though.

sturnclaw:
> We have (had?) a combined fuel + cargo scoop that saved that precious 1t. Is fuel scoop still taking a slot? No multiscoop anymore?

Correct - we (@bszlrd and I) decided to remove the multiscoop as a fuel scoop and cargo scoop are two completely different things both in-universe and outside of universe. The eventual goal is to expand fuel scooping gameplay to provide some tradeoffs in the kind of equipment you can install in the fuel scoop slot.

Equipment weight and volume is something that will need some balance sooner or later, but for the time being the extra 1m3 volume isn't a particularly big concern now that equipment volume and cargo space are separate concerns.

> Closest tutorial I can think of is this one and it's already outdated [...]
> I think it's better if someone who's into that code looks into this though.

This tutorial is so hilariously outdated it's not really even worth bothering with - it's about 12 years and two major rewrites of the equipment system out of date.

At some point, I will attempt to make an effort to write a tutorial on the new equipment system for the dev docs, but I cannot promise a timeframe. It will definitely not be before this upcoming release :smile:

impaktor:
> At some point, I will attempt to make an effort to write a tutorial on the new equipment system

My thought is any wiki tutorial will be bound to be updated, and weigh in the back of the minds of the guilty developer, and it is better to just designate a few lua files as "tutorial" with bountiful comments, and link to them from the wiki. Sure, comments will need to be updated when code is, but I deem that much more likely, than someone maintaining the wiki.

E.g. DonateToCranks is very simple interaction with BBS, and DeliverPackage for some deeper mission. I think I have an ancient Donate module with comments, for this purpose.

Those were my two kronor.
(Although _öre_ would be more analogous with _cents_)

PR #6042, opened 2025-02-04
Reverting reticule text background opacity back to 0.4

bszlrd:
Reported on discord by Admetus, the text background around the reticule is quite dark:
image
Makes sense on light backgrounds, but is distracting over dark:
image

This PR reverts back the 0.8 opacity to 0.4 in the theme:
screenshot-20250204-115458
screenshot-20250204-115453
screenshot-20250204-115652

In the long run I think it would be a better solution, if the blending mode of them would be Multiply. Then it would only darken, but not brighten. Not sure if that's easily and cheaply doable though.


sturnclaw:
So - what needs to happen here is that the main flight UI elements need to be moved to a separate color name, as lightBlackBackground (horrible name) is used in a lot of places that should not be made less opaque. I'll look at it when I'm back from my break.

bszlrd:
Damn, I should have thought of that.

PR #6044, opened 2025-02-04
Resurrect ServerAgent for external tool access

fluffyfreak:
This is not a new feature but after a discussion on the Discord channel I did promise to take a look at it and to stop me forgetting (_again_) I'm going to track this here as a WIP 😁 

Todo:
- this code is NOT equivalent
`		Json::FastWriter writer;
		req.buffer = writer.write(req.data);
		req.buffer = req.data;`
- responding to success is commented out due to me not understanding the JSON changes yet
- once the above is fixed it will need testing

### DIFF
--- a/src/ServerAgent.cpp
+++ b/src/ServerAgent.cpp
@@ -11,9 +11,12 @@
 #pragma comment(lib, "libcurl.lib")
 #endif
 
+#include "buildopts.h"
 #include "ServerAgent.h"
 #include "StringF.h"
 #include <curl/curl.h>
+#include <SDL.h>
+#include <SDL_Thread.h>
 
 void NullServerAgent::Call(const std::string &method, const Json &data, SuccessCallback onSuccess, FailCallback onFail, void *userdata)
 {
@@ -153,8 +156,7 @@ void HTTPServerAgent::ThreadMain()
 		// done with the queue
 		SDL_UnlockMutex(m_requestQueueLock);
 
-		Json::FastWriter writer;
-		req.buffer = writer.write(req.data);
+		req.buffer = req.data;
 
 		Response resp(req.onSuccess, req.onFail, req.userdata);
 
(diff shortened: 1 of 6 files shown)


src/ServerAgent.cpp
```diff
@@ -153,8 +156,7 @@ void HTTPServerAgent::ThreadMain()
 		// done with the queue
 		SDL_UnlockMutex(m_requestQueueLock);
 
-		Json::FastWriter writer;
-		req.buffer = writer.write(req.data);
+		req.buffer = req.data;
```
fluffyfreak, line 159: @sturnclaw this is one of the two problems I could do with some advice about, this line is not an equivalent replacement for the `writer.write(req.data)` which used to be there which I believe serialised the data.

src/ServerAgent.cpp
```diff
@@ -178,10 +180,10 @@ void HTTPServerAgent::ThreadMain()
 		}
 
 		if (resp.success) {
-			Json::Reader reader;
-			resp.success = reader.parse(resp.buffer, resp.data, false);
-			if (!resp.success)
-				resp.buffer = std::string("JSON parse error: " + reader.getFormattedErrorMessages());
+			//Json::Reader reader;
```
fluffyfreak, line 183: @sturnclaw second is this section which I think deserialised a repsonse

sturnclaw: This will be `resp.data = Json::parse(resp.buffer); resp.success = !resp.data.is_null();`. Substitute the `!is_null()` check for an `is_object()` or `is_array()` depending on the wire format of your message type.

sturnclaw: You'll likely also need to wrap it in a try block or use a non-throwing method; still getting my development environment back up so that's an exercise left to the reader.

PR #6045, opened 2025-02-04
Replace `texture2D()` with `texture()`

fluffyfreak:
Urgh, `texture2D` is deprecated, it's replacement `texture` is the one to use.

Bug made possible by nVidia drivers just working despite that 🤦 always catches me out.

image
A big thankyou to `Natrix BQ-3566 pilot` from the Discord channel for pointing it out.


fluffyfreak:
@impaktor I'll merge this since it's tested and a well known type of change, we've already made it elsewhere in the code

sturnclaw:
Whoops, that's also my bad. Thanks for catching this!

PR #6046, opened 2025-02-05
equipment-outfitter.lua: Interpret MILITARY as tech level 11.

### LINKED ISSUE #6039: GUI module crash while trying to install hull auto-repair
### Observed behaviour
Equipment market tab throws out an error when trying to install the hull auto-repair module

Image

### Expected behaviour
UI tab working properly and without issues

### Steps to reproduce
Open station comms, open equipment market, click on "Hull Auto-Repair System" in the module list

Pioneer version: 03.02.2025
OS: Windows 10 x64

[output.txt](https://github.com/user-attachments/files/18650341/output.txt)

pghvlaans:
The function `Outfitter:renderCompareRow()` in `equipment-outfitter.lua` assumes a numeric value. This causes a GUI crash in the equipment outfitter when selecting an item with the tech level `MILITARY` due to attempting either number formatting (if the slot is empty) or a numeric comparison (if the slot is full).

Give this the same handling as `MILITARY` level stations get by changing the level to `11` for comparison and display.

fix #6039

### DIFF
--- a/data/pigui/libs/equipment-outfitter.lua
+++ b/data/pigui/libs/equipment-outfitter.lua
@@ -491,13 +491,6 @@ function Outfitter:renderCompareRow(label, stat_a, stat_b)
 	ui.tableNextColumn()
 	ui.text(label)
 
-	if stat_a then
-		stat_a[3] = stat_a[3] == "MILITARY" and 11 or stat_a[3]
-	end
-	if stat_b then
-		stat_b[3] = stat_b[3] == "MILITARY" and 11 or stat_b[3]
-	end
-
 	local icon_size = Vector2(ui.getTextLineHeight())
 	local color = stat_a and stat_b
 		and compare(stat_a[3], stat_b[3], stat_a[5])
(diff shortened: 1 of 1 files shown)


sturnclaw:
I'd strongly recommend you change `modifyEquipmentStats()` to transform 'MILITARY'->11 there instead - it's significantly less intrusive to fix it at the "source" as it were. Otherwise this is a good change, thank you for proposing it!

pghvlaans:
Makes sense, thanks! It should be OK now.

impaktor:
I've not looked at the code in context of other code, but just be aware that military tech level could be construed as "level 11", except it has equipment unique to it, that is _not_ available at tech level 12. 

All other tech levels are "bigger = better", i.e. all other tech levels have everything offered at the lower (except this virtual tech level 11 = military level")

pghvlaans:
> that is *not* available at tech level 12.

I just checked Torvalds Platform (level 12) and it had military stuff for sale, so the latest commit won't work. I'll have another look tomorrow.

pghvlaans:
There's a bit more going on here:

* `stationHasTech` makes "MILITARY" equipment available in level 12 stations
* Two of the military drives have tech level 11 and 12, respectively, which makes them available in level 12 stations anyway
* 11 is a distinct tech level from "MILITARY", so the output probably shouldn't say 11

I'll close and open a new PR.

PR #6048, opened 2025-02-06
SetFuel after hyperjumps

pghvlaans:
In `Types.lua`, the function `OnLeaveHyperspace()` currently adjusts the level of stored fuel without  calling `SetFuel`. The result is that the mass of the hyperdrive's fuel reserve does not decrease when fuel is used. To verify this, start a new game and jump anywhere; the reported mass of the hyperdrive will remain at 8.0 for a Mars start. The hyperdrive can gain a considerable amount of mass over time as the player refuels.

The fix is to call `SetFuel` in `OnLeaveHyperspace()` rather than adjusting `storedFuel` directly.

### DIFF
--- a/data/modules/Equipment/Types.lua
+++ b/data/modules/Equipment/Types.lua
@@ -267,7 +267,7 @@ function HyperdriveType:OnLeaveHyperspace(ship)
 		local amount = ship['nextJumpFuelUse']
 		ship:unsetprop('nextJumpFuelUse')
 
-		self.storedFuel = math.max(0, self.storedFuel - amount)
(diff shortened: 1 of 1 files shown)


impaktor:
@pghvlaans Lest I forget, if/when this is merged, do you want to be in [AUTHORS.txt](https://github.com/pioneerspacesim/pioneer/blob/master/AUTHORS.txt), and by what name?

pghvlaans:
@impaktor Sure, thank you. "K. Eugene Carlson," please.

sturnclaw:
Good catch, thank you! I very much appreciate getting this quality of "bug reporting" :smile:

PR #6050, opened 2025-02-06
Generalize to any OpenGL binding

ManDay:
Due to an unfortunate design of the FindOpenGL.cmake module, not specifying `COMPONENTS OpenGL` would fail for systems without GLX, and OPENGL_LIBRARIES would not be set, either.

This should hopefully work for both now (until cmake modifies the module).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,7 +267,7 @@ endif (MSVC)
 
 find_package(Threads REQUIRED)
 find_package(Freetype REQUIRED)
-find_package(OpenGL REQUIRED)
+find_package(OpenGL REQUIRED COMPONENTS OpenGL)
 
 set(FMT_INSTALL OFF CACHE BOOL "Enable install of libfmt" FORCE)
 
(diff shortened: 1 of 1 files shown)


ManDay:
FTR: The latest release of GLEW also doesn't compile against EGL and needs some minor Makefile modifications. So this will only be useful when GLEW releases a new version for EGL support.

impaktor:
No feedback from devs on this. I'm pinging in @mwerle and @pcercuei who might know CMake, and @sturnclaw (when you have time).

pcercuei:
Looks good.

mwerle:
Haven't tested but looks ok. 

Curious what systems exhibit the issue since it seems to have worked fine up to now.

fluffyfreak:
This change appears to break Visual Studio, I'll investigate it later
image

fluffyfreak:
Reverting just the `OpenGL::OpenGL` part fixes it for MSVC 2022

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 46b81d76f..08f43a863 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -385,7 +385,7 @@ list(APPEND pioneerLibs
 	pioneer-lua
 	${ASSIMP_LIBRARIES}
 	${FREETYPE_LIBRARIES}
-	OpenGL::OpenGL
+	${OPENGL_LIBRARIES}
 	${SDL2_LIBRARIES}
 	${SDL2_IMAGE_LIBRARIES}
 	${SIGCPP_LIBRARIES}
```

pcercuei:
@fluffyfreak try with `OpenGL::GL`.

fluffyfreak:
@pcercuei that seems to work for me 👍 I'll raise a PR for it

PR #6054, opened 2025-02-07
Lodos normal re-fix

bszlrd:
The Lodos had its normals inverted on the RCS thrusters. @Maks2103 provided a fix, but when I updated the tags to the new equipment sys, I accidentally used the old version. This commit re-fixes the model, but the tags are the new ones.
I'm also adding him to the authors.txt (should have happened with #5782 )
lodos-hole
lodos-fix


sturnclaw:
Clearly this is a divine indication that you need to redesign the Lodos next... :wink:

In all seriousness, thanks for fixing this again!

PR #6058, opened 2025-02-08
Formatting fixes

zonkmachine:
**System overview**
* Introducing a space before the unit. This was in response to the formatting of objects in the system overview but **Format** is called elsewhere. I have looked for possible issues in other modules but so far have found none. (Images)
* Add unit for density.

**System Editor**
* Add Mean Density to the System Editor. It's good to be able to quickly assess if the mass/volume ratio is within a reasonable
range.

**Ship Market**
* Space before weight (t) unit.

**Cargo Run**
* Space before weight unit.

**New game window - Ship tab**
* Format fixes - more space

Before
exampleearth

Fixed
format


zonkmachine:
Fixed up gravitational acceleration. I'm not quite sure what changes would be needed in **Format** for a more permanent solution.

fluffyfreak (approved):
I think these are good, I don't see any issues with the code 👍

zonkmachine:
> I think these are good, I don't see any issues with the code 👍

Right, I agree. Still, parts of this should probably be fixed in **Format** but I haven't wrapped my head around it yet. It's not a new issue though and can be fixed up later.

zonkmachine:
I squashed it down to two commits.

sturnclaw:
Tons per cubic meter is (AFAIK) directly equivalent to grams per cubic centimeter; i.e. the number is the same regardless of the suffix (cm3 -> m3 is a 10^6 increase, as is grams -> tons). While the SI unit is technically more correct to use, I have no preference either way.

zonkmachine:
> Tons per cubic meter is (AFAIK) directly equivalent to grams per cubic centimeter; i.e. the number is the same regardless of the suffix (cm3 -> m3 is a 10^6 increase, as is grams -> tons). While the SI unit is technically more correct to use, I have no preference either way.

I also have no preference. Right now it's **t/m³** in the System Overview and **g/cm³** in the System Editor. Maybe should settle for one of them? My guess is that t/m³ feels a bit more intuitive when you're dealing with something larger like a planet. At least it does to me.

data/pigui/modules/system-view-ui.lua
```diff
@@ -814,11 +814,11 @@ function Windows.objectInfo:Show()
 				{ name = lc.RADIUS, icon = icons.body_radius,
 					value = (not starport) and ui.Format.Distance(body.radius) or nil },
 				{ name = lc.SURFACE_GRAVITY, icon = icons.body_radius,
-					value = (not starport) and ui.Format.Speed(body.gravity, true).." ("..ui.Format.Gravity(body.gravity / 9.80665)..")" or nil },
+					value = (not starport) and ui.Format.Speed(body.gravity, true).."²".." ("..ui.Format.Gravity(body.gravity / 9.80665)..")" or nil },
 				{ name = lc.ESCAPE_VELOCITY, icon = icons.body_radius,
 					value = (not starport) and ui.Format.Speed(body.escapeVelocity , true) or nil },
 				{ name = lc.MEAN_DENSITY, icon = icons.body_radius,
-					value = (not starport) and ui.Format.Mass(body.meanDensity).."/m3" or nil },
+					value = (not starport) and ui.Format.Mass(body.meanDensity).."/m³" or nil },
```
zonkmachine, line 821: @sturnclaw I looked into changing the mean density to **g/cm** in this PR but I think this would need some work. Right now we're (**I** really as I'm the one responsible for the original addition) just using **Format.Mass** which is in tons and stick a **/m³** to the end. To change this to **g/cm** we would need something like a new **Format.Density** to call. Also, instead of using Format.Mass, one could just use **body.meanDensity** and fix the format manually. Maybe I missed a way but I think adding a new **Format.Density** is the best way to do it.

sturnclaw: Adding Format.Density would be fine, as would changing the line to be `ui.Format.Number(body.meanDensity, 3) .. lc.UNIT_DENSITY` with `UNIT_DENSITY` being defined as the appropriate `g/m^3`. Regardless of the tack taken, the unit suffix should ideally be a localized string for consistency (as unlikely as it is for SI units to be written differently in other languages...)

zonkmachine: OK. The SI unit for density (Wikipedia) is kg/m3 and that's what you get from calcMeanDensity. You need to divide with 1000 (like I did in the editor) so you get **ui.Format.Number(body.meanDensity / 1000.0, 3)**
I'll poke around some more with it.

zonkmachine: I switched to SI and kg/m3 instead. It made the code cleaner and it feels more like a unit that miners and fixers in space a thousand years from now would use.

zonkmachine:
Right. Some testing needed but otherwise I think this is ready.

zonkmachine:
Found some last issues in this run. Are these changes wanted in the new game menus ship tab
Should there be space around the `" / "` sign?

**Before**
formatnewgamemenubefore

**After**
formatnewgameship

zonkmachine:
@fluffyfreak @sturnclaw I went astray and fixed up things I didn't intend to and that no-one asked me to. Please accept my invitation for a re-review. I've now also touched the new game window's ship tab, the ship market, and the CargoRun module. Some direct and some via the Format function (data/pigui/libs/text.lua). I think I'm done for this run.

zonkmachine:
Squashed and done here!

The only thing I would like to see answered is the question I bring up above: https://github.com/pioneerspacesim/pioneer/pull/6058#issuecomment-2755399589

zonkmachine:
Please excuse the noise. I did some final changes on the following pages.
```
modified:   data/pigui/modules/info-view/01-ship-info.lua
modified:   data/pigui/modules/info-view/03-econ-trade.lua
modified:   data/pigui/modules/station-view/01-lobby.lua
```

I believe this is ready to merge.

impaktor:
Looks fine to me. 

I assume @sturnclaw is fine with it, as per previous comment from him.

zonkmachine:
> I assume @sturnclaw is fine with it, as per previous comment from him.

That was two weeks ago. I've done unspeakable things to the strings since.

zonkmachine:
Does it need a proper commit message? Maybe it should mention the added info in the System Editor.

```
* Fixes to formatting of presented data (mostly SI).
* Adds mean density to Body Parameters in the System Editor.
```

sturnclaw:
Alright, I'll take a look at this for a final review as soon as possible (if only so @zonkmachine doesn't have to feel like he has to change ten more files :wink:).

sturnclaw:
With apologies for losing track of this PR - thank you very much @zonkmachine!

PR #6059, opened 2025-02-08
Store AtmosphereParameters in BaseSphere, calculate with material setup

fluffyfreak:
This is a bit of tidying up, dealing with the old comment `//XXX no need to calculate AP every frame` by calculating the atmosphere parameters only once.

It does mean bumping the BaseSphere size from 48 bytes to 128 bytes but that can't really be avoided.

No visual or gameplay differences

### DIFF
--- a/src/BaseSphere.h
+++ b/src/BaseSphere.h
@@ -48,21 +48,22 @@ public:
 	RefCountedPtr<Graphics::Material> GetSurfaceMaterial() const { return m_surfaceMaterial; }
 
 protected:
+	virtual void SetUpMaterials() = 0;
+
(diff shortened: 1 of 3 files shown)


impaktor:
So lighter on CPU / more FPS heavier on RAM?

fluffyfreak:
Yeah but it's drop in the coean on both sides. This will be barely noticable on the CPU, or the RAM. In a system of 100 planets it's going from 48 * 100 = 4800 bytes or just under 5KB, to 128 * 100 = 12800 bytes or just under 12KB 🤷 

It's why I labelled it as cleanup

PR #6063, opened 2025-02-09
Change modelcompiler single model filepathing, change Editor location for MSVC

fluffyfreak:
I noticed a bug/issue when trying to convert a single model to SGM format. It would only output the SGM to the full filepath if I specified `inplace` or some other 3rd parameter.

I've changed this to always find the full filepath regardless of having extra parameters.

The `Editor.exe` was also being output to a folder which meant I couldn't debug it from MSVC so for MSVC builds it now output to the CMAKE_SOURCE_DIR like the game and other tools.

### DIFF
--- a/src/modelcompiler.cpp
+++ b/src/modelcompiler.cpp
@@ -220,25 +220,26 @@ start:
 			if (argc > 3) {
 				std::string arg3 = argv[3];
 				isInPlace = (arg3 == "inplace" || arg3 == "true");
+			}
 
-				// find all of the models
-				FileSystem::FileSource &fileSource = FileSystem::gameDataFiles;
-				for (FileSystem::FileEnumerator files(fileSource, "models", FileSystem::FileEnumerator::Recurse); !files.Finished(); files.Next()) {
-					const FileSystem::FileInfo &info = files.Current();
-					const std::string &fpath = info.GetPath();
+			// find all of the models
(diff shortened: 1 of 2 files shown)


src/modelcompiler.cpp
```diff
@@ -220,25 +220,26 @@ extern "C" int main(int argc, char **argv)
...
-					if (info.IsFile()) {
-						if (ends_with_ci(fpath, ".model")) { // store the path for ".model" files
-							const std::string shortname(info.GetName().substr(0, info.GetName().size() - 6));
-							if (shortname == modelName) {
-								filePath = fpath;
-								break;
-							}
+			// find all of the models
+			FileSystem::FileSource &fileSource = FileSystem::gameDataFiles;
+			for (FileSystem::FileEnumerator files(fileSource, "models", FileSystem::FileEnumerator::Recurse); !files.Finished(); files.Next()) {
```
sturnclaw, line 227: Post-facto review: there is a significantly nicer range-for iterator available for FileEnumerator now, if you happen to revisit this code.

```c++
// Full typename included for informational purposes, feel free to abbreviate to auto
for (const FileSystem::FileInfo &info : fileSource.Enumerate("models", FileSystem::FileEnumerator::Recurse)) {
    // ...
}

// Alternatively:
for (const FileSystem::FileInfo &info : fileSource.Recurse("models" /*, additional flags */)) // ...
```

PR #6067, opened 2025-02-11
Make Sphere3D/Icosphere reserve space better

fluffyfreak:
Whilst doing something else I found that the Icosphere::Generate method wasn't reserving space for the indices it generates, and it was also miscalculating the number of vertices it would generate.

I spent some time trying to figure out a nifty formula to calculate those values and then realised I could just build a table of 11 entries for each of the valid subdivision we allow instead.

I have also increased the BaseSphere use of subdivision to 5 (from 4) for smoother atmospheres.

EDIT:
Made the `BaseSphere::m_atmos` into a static so it's only created once and used throughout the games lifetime instead of creating a new one for every single planet with an atmosphere. Making the icosphere generation faster is good, but avoiding doing it altogether is better still!

### DIFF
--- a/src/BaseSphere.cpp
+++ b/src/BaseSphere.cpp
@@ -80,8 +80,12 @@ void BaseSphere::DrawAtmosphereSurface(Graphics::Renderer *renderer,
 
 	renderer->SetTransform(matrix4x4f(modelView * matrix4x4d::ScaleMatrix(rad) * invrot));
 
-	if (!m_atmos)
-		m_atmos.reset(new Drawables::Sphere3D(renderer, 4, 1.0f, ATTRIB_POSITION));
+	if (!m_atmos) {
+		// 5 subdivision = 20472 verts, 61440 indices NB: this should be configurable!
+		static constexpr int subdivisions = 5;
+		m_atmos.reset(new Drawables::Sphere3D(renderer, subdivisions, 1.0f, ATTRIB_POSITION));
+	}
+	
 	m_atmos->Draw(renderer, mat.Get());
 
 	renderer->GetStats().AddToStatCount(Graphics::Stats::STAT_ATMOSPHERES, 1);
--- a/src/BaseSphere.cpp
+++ b/src/BaseSphere.cpp
@@ -6,6 +6,8 @@
 #include "GasGiant.h"
 #include "GeoSphere.h"
 
+#include "Pi.h"
+
 #include "galaxy/AtmosphereParameters.h"
 #include "galaxy/SystemBody.h"
 #include "graphics/Drawables.h"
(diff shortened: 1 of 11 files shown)


src/graphics/Drawables.cpp
```diff
@@ -482,16 +482,36 @@ namespace Graphics {
...
+		// NB: { num vertices, num indices }
+		// num indices == (num_vertices + 8) * 3
+		// The original formula used to reserve space was WILDLY wrong "(subdivs * subdivs) * 20 * 3"
+		// I bet there's a really clever way to calculate the vertices and indices, a smart person could do it
+		// ...but this is me, and here's a table that I generated through brute force - AndyC
+		static const size_t icosahedron_counts[11][2] = {
+			{ 12, 60 }, // subdivs 0
+			{ 72, 240 },
+			{ 312, 960 },
+			{ 1272, 3840 }, // subdivs 4 (planets atmopsheres as of 2025/02/11)
```
sturnclaw, line 494: Comment should be updated ;)

src/BaseSphere.cpp
```diff
@@ -80,8 +80,12 @@ void BaseSphere::DrawAtmosphereSurface(Graphics::Renderer *renderer,
 
 	renderer->SetTransform(matrix4x4f(modelView * matrix4x4d::ScaleMatrix(rad) * invrot));
 
-	if (!m_atmos)
-		m_atmos.reset(new Drawables::Sphere3D(renderer, 4, 1.0f, ATTRIB_POSITION));
+	if (!m_atmos) {
+		// 5 subdivision = 20472 verts, 61440 indices NB: this should be configurable!
+		static constexpr int subdivisions = 5;
+		m_atmos.reset(new Drawables::Sphere3D(renderer, subdivisions, 1.0f, ATTRIB_POSITION));
```
sturnclaw, line 88: Note: I would strongly prefer the number of subdivisions used for planets be controlled in part by the planet detail setting. We probably don't want to be wasting bandwidth on ~5,000 triangles of atmosphere mesh (15k indices) when significantly fewer would suffice on Raspberry Pi, for example.

sturnclaw (approved):
Overall I like this change, just one "side quest" thing I'd like to see tackled if you get time.

src/graphics/Drawables.cpp
fluffyfreak: oops :D

src/BaseSphere.cpp
fluffyfreak: phew, ok that was an adventure but I've done it, High and Very High get subdivs 5, everything else gets 4

fluffyfreak:
Changing it based on the planets detail level ended up being quite involved so this now hits 11 files 🤷

PR #6069, opened 2025-02-12
Paint Shop - Randomize colors

zonkmachine:
paintshoprandom1
paintshoprandom2

Add a button to randomize the colors in the Paint Shop
It randomizes:
* All colors to different values (top image)
* Two colors the same value
* All colors the same value (bottom image)

### DIFF
--- a/data/pigui/modules/station-view/06-shipRepairs.lua
+++ b/data/pigui/modules/station-view/06-shipRepairs.lua
@@ -5,6 +5,7 @@ local ui = require 'pigui'
 local StationView = require 'pigui.views.station-view'
 local ShipDef = require "ShipDef"
 local Game = require "Game"
+local Engine = require 'Engine'
 local Rand = require "Rand"
 local PiGuiFace = require 'pigui.libs.face'
 local Format = require "Format"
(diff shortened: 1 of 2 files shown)


bszlrd:
Nice idea!

data/pigui/modules/station-view/06-shipRepairs.lua
```diff
@@ -320,11 +316,42 @@ local function drawPaintshop()
...
 
 				ui.sameLine()
-
-				if ui.button(l.RESET_PREVIEW, Vector2(200, 36)) then
+				if ui.button(l.RESET_PREVIEW, Vector2(150, 36)) then
 					resetPreview()
 				end
 
+				ui.sameLine()
+				if ui.button(l.RANDOM_COLORS, Vector2(150, 36)) then
```
zonkmachine, line 324: I'm thinking I can declare **randomColor()** here instead. Or maybe skip it and copy  that single string for each of the **priviewColors[]** below.

zonkmachine: **randomColor()** moved

zonkmachine:
Squashed and done.

fluffyfreak (approved):
Good idea 👍

zonkmachine:
Found an issue. The Buttons don't scale well on lower resolution. Looking into.

fluffyfreak:
I think that's a known issue with the UI. How low a resolution are you looking at because I think it was discussed that anything below 800x600 there are scaling issues throughout the UI

zonkmachine:
> I think it was discussed that anything below 800x600 there are scaling issues throughout the UI

OK. I fixed the width on the buttons I worked with here. I haven't looked at the rest of Pioneer though. On screen sizes below  800x600 you see the right side, color selection, being pushed down a bit and given a scroll bar as the top left button, **Visit Repair Services**, reserves space along the whole top of the screen. That could be fixed to make it reserve space just above the model spinner. I failed to find a quick solution for it here though.

sturnclaw:
I'll merge this for the upcoming hotfix if you squash the branch. Thanks for the contribution!

PR #6070, opened 2025-02-12
Sort lua components during serialization

### LINKED ISSUE #6064: Most savegames can not be loaded and crash the game
With the latest release (from source) this applies to a good 80% of the saves I have made. To reproduce:
 - Start the game
 - Select "Load Game"
 - Select one of the provided examples
 - Click "Load"

Observed:
After a few moments, a message appears in the lower right corner "This saved game cannot be loaded because it contains errors" for a split second, then the game closes with a message "Info: Pi shutting down"

Additionally (maybe a separate bug?), to reproduce:
 - Start the game
 - Select "Options"
 - Click "Debug info"
 - Select "Load Game"
 - Select one of the provided examples
 - Click "Load"

Observed:
The game segfaults.

Additional information: There is a clock skew in the system time (CMOS battery dead), I don't expect this to be related.

[savefiles.zip](https://github.com/user-attachments/files/18734229/savefiles.zip)

Gliese852:
In the serialization of the Lua objects there is a system of "one definition", when all subsequent mention of the object only refer to it. It is vital that the full definition meets before the rest of the refs to the object, otherwise an error occurs.

The same objects can now meet in different lua components, and this means that we must ensure the order of processing procedure during saving and loading.

fixes #6064

### DIFF
--- a/src/lua/LuaObject.cpp
+++ b/src/lua/LuaObject.cpp
@@ -759,6 +759,10 @@ bool LuaObjectBase::SerializeComponents(LuaWrappable *object, Json &out)
 	}
 
 	lua_pushnil(l);
+
+	// sort the components before serialization
+	std::vector<std::string> componentNames;
+
 	while (lua_next(l, -2) != 0) {
 		if (!lua_isstring(l, -2)) {
 			lua_pop(l, 1);
@@ -772,6 +776,14 @@ bool LuaObjectBase::SerializeComponents(LuaWrappable *object, Json &out)
 			continue;
 		}
 
+		componentNames.emplace_back(key);
+		lua_pop(l, 1);
+	}
+	std::sort(componentNames.begin(), componentNames.end(), std::less{});
+
+	for (const auto &key : componentNames) {
+		lua_pushstring(l, key.data());
+		lua_rawget(l, -2);
 		// Pickle the table to json
 		LuaSerializer::pickle_json(l, lua_gettop(l), out[key.data()], "BodyComponent");
 		lua_pop(l, 1);
@@ -809,7 +821,7 @@ bool LuaObjectBase::DeserializeComponents(LuaWrappable *object, const Json &obj)
 		return false;
 	}
 
-	// Deserialize all components
+	// Deserialize all components, in alphabetical order
 	for (const auto &pair : obj.items()) {
 		lua_pushstring(l, pair.key().c_str());
 		LuaSerializer::unpickle_json(l, pair.value());


Gliese852:
So I looked again and now I saw some promise that the values inside the json object are sorted:
https://github.com/pioneerspacesim/pioneer/blob/89d503adacfd8a777f6fa755f8a09dbba5078a4a/contrib/json/json.hpp#L11543-L11546

I wonder, can we trust this and not even use `std::set`?

sturnclaw:
> So I looked again and now I saw some promise that the values inside the json object are sorted:
> 
> https://github.com/pioneerspacesim/pioneer/blob/89d503adacfd8a777f6fa755f8a09dbba5078a4a/contrib/json/json.hpp#L11543-L11546
> 
> I wonder, can we trust this and not even use `std::set`? On the one hand, this is put into the docs, on the other hand, there is the word "Internally".

There is a strong guarantee that the contents of a JSON object are lexicographically sorted when written to and read from disk, and that sort order is based on the container type used in the `basic_json` type definition. By default (which we use) all sorting of JSON keys uses `std::less<std::string>`.

The problem at hand is that we are potentially deserializing keys in a different order than we are serializing them, due to the instability of Lua table traversal. This requires a sorted order for keys when serializing, but as long as that sort is performed using `std::less` we can deserialize directly without an additional sort step.

Note that the data referred to by `std::string_view key` *must* be assumed to go out-of-scope as soon as the underlying string key is popped from the Lua stack. Lua [does not provide any guarantees](https://www.lua.org/manual/5.2/manual.html#lua_tolstring) that string data is valid unless the string value is present on the stack. Thus, the `std::set<...> componentNames` should be `std::set<std::string>`.

Gliese852:
@sturnclaw 
> Note that the data referred to by `std::string_view key` _must_ be assumed to go out-of-scope as soon as the underlying string key is popped from the Lua stack. Lua [does not provide any guarantees](https://www.lua.org/manual/5.2/manual.html#lua_tolstring) that string data is valid unless the string value is present on the stack. Thus, the `std::set<...> componentNames` should be `std::set<std::string>`.

 Indeed, very specifically written, thanks! But still, maybe we can count on the fact that these strings in a sense are still on the stack, because the table containing them is on the stack? Not that I am against `std::string`, I just want to know your opinion on this.

sturnclaw:
> But still, maybe we can count on the fact that these strings in a sense are still on the stack, because the table containing them is on the stack?

That would be relying on undocumented behavior, and so I'd somewhat prefer to avoid doing so in the general case.

In this particular case however, because strings are uniquely interned in the global Lua string table (not actually a `LUA_TTABLE`) and IIRC have pointer identity, as long as the string value in question is considered reachable by the Lua GC it's *probably* safe to do so.

Regardless of `std::string` or `std::string_view`, I'd recommend pushing all component strings into a single `std::vector` and then performing a sort using the `std::less` comparator.

Gliese852:
@fluffyfreak 
> I've just tried this PR with a savegame reported on Discord https://discord.com/channels/705212721666326618/705542778716684368/1342498734855946334 but without success, i.e. it does not load a savegame that has already been "corrupted".


Yes, because while saving, the order of serialization of Lua objects was randomly violated, and now when loading it is difficult to determine what order was there.

To load the corrupted savegames, it would be possible to completely remove the requirement to maintain the order for loading the lua objects (as is the case when recovering), but this is a more serious intervention and more chances of any regression. I decided to follow the simple path, and just return the maintenance of order while saving. Given that the game can be recovered (if hold Ctrl in the load menu), it seems to be a sufficient fix.

> In their case user `pebblegarden` has deduced that the problem started after buying the surface scanner.

Yes, because the scanner is inside the two components `ScanManager` and `EquipSet`, the order for saving of which is selected randomly (via lua-function `next`), but loading always occurs in alphabetical order.

sturnclaw:
I've thought about changing the way savefiles are serialized a few times; it would definitely reduce some of the implicit ordering constraints (by converting them to explicit-ordering and serializing Lua tables into a flat array) but would significantly reduce our ability to easily read and debug savefiles. There are also still issues with circular dependencies regardless of how we try to serialize data.

Regardless, that thought isn't really relevant to the issue at hand; @Gliese852 has done a good job here of ensuring serialization is consistent with deserialization order.

sturnclaw (approved):
Approved pending testing; I'm unsure I'll have the time available to test in the next day or so.

Testing procedure should involve multiple runs of spawning in, purchasing (or using the debug tools to add) a planetary scanner, saving the game, and attempting to load the save. At least three runs should be performed.

fluffyfreak (approved):

sturnclaw:
Thanks @Gliese852!

PR #6072, opened 2025-02-13
Update ubuntu version 20.04 -> 22.04 for build

impaktor:
We need to update these, as github will deprecate them shortly (April?).
option is: 
- ubuntu 22.04
- ubuntu 24.04
- ubuntu-latest

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -107,7 +107,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   build-gcc:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
     # Checkout the repository as $GITHUB_WORKSPACE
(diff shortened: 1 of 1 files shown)


sturnclaw:
Thanks for updating these! 22.04 is definitely the right runner to select, but I'd prefer to defer this merge until after we cut a hotfix just to prevent any implicit binary dependency issues from arising due to the build environment change.

impaktor:
I'm in no rush. For future reference, this is github's schedule:

> The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025. To raise awareness of the upcoming removal, we will temporarily fail jobs using Ubuntu 20.04.
> Builds that are scheduled to run during the brownout periods will fail. The brownouts are scheduled for the following dates and times:      
> *March 4 14:00 UTC – 22:00 UTC 
> *March 11 13:00 UTC – 21:00 UTC
> *March 18 13:00 UTC – 21:00 UTC
> *March 25 13:00 UTC – 21:00 UTC

sturnclaw:
We're close enough to the cutoff date I'll merge this anyways. Thanks for the work @impaktor!

PR #6073, opened 2025-02-14
Fix  lighting on models that use normal maps

jeffmd:
The lighting on the models that use a normal map were not looking correct and the normal map had no effect on lighting.
I setup a test shader that did a dump of a_tangent and it was always (0, 0, 0) which explained why the normal calc in the fragment shader was incorrect.  I then did a dump of a_uv1 and it had the tangent data.  I swapped the locations of a_tangent and a_uv1 in attributes.glsl and the lighting on normal mapped models now looks correct , to me :).

After some digging I found that attributes.glsl and Program.cpp have tangent at location 5 but the vertex buffer layout in VertexBufferGL.cpp (line 23) has tangent at index 4.

So by putting a_tangent in location 4 the lighting is now correct for models using a normal map

### DIFF
--- a/src/graphics/opengl/VertexBufferGL.cpp
+++ b/src/graphics/opengl/VertexBufferGL.cpp
@@ -20,7 +20,7 @@ namespace Graphics {
 			case ATTRIB_NORMAL: return 1;
 			case ATTRIB_DIFFUSE: return 2;
 			case ATTRIB_UV0: return 3;
-			case ATTRIB_TANGENT: return 4;
+			case ATTRIB_TANGENT: return 5;
 			default:
 				assert(false);
 				return 0;


jeffmd:
marsLanding

sturnclaw (changes requested):
Thank you for catching this issue!

Because the meaning of `ATTRIB_TANGENT` is intended to be the numeric value `5`, I'd prefer you change VertexBufferGL.cpp line 23 to return the proper attribute index instead. It's slightly simpler that way and shares the same value across both shaders and C++ source.

jeffmd:
Yep, makes sense. Done.

sturnclaw:
As a general git-etiquette thing, we prefer to avoid mentioning issue numbers and @-usernames in commit messages; every time the commit in question is pushed (including to other users' forks, etc.) Github triggers a notification of the mentioned issue/user.

Otherwise the code looks quite simple now! :wink:

fluffyfreak (approved):
Approved 👍 one character bug fixes are the best 😁

Issue #6079, opened 2025-02-16
Combat Mission status do not update until payment


Helioth:
### Observed behaviour
When a mission is completed, its status does not change anymore and stays “Active” in “Missions” tab (tested on combat missions only).

![Image](https://github.com/user-attachments/assets/4688eb1e-c2db-4c5b-90bc-1f1ec1008345)

![Image](https://github.com/user-attachments/assets/9187dea5-7930-492a-8ff5-657cdfd49f73)

### Expected behaviour
Prior to 20250203 (v586b518 master), the status was changing to something else which made it easier to spot remaining open missions.

### Steps to reproduce
Open the saved file and complete the mission.

__My pioneer version (and OS):__
20250203 (v586b518 master), W11

__My output.txt (required) and game save (optional, but recommended)__
[Helioth_Bug.zip](https://github.com/user-attachments/files/18814009/Helioth_Bug.zip)

impaktor:
@robothauler why were the 2 "COMPLETE" lines removed? 
8fc6d104544024de185a9bc4fa722dc88e62be01

robothauler:
@impaktor yes. This was removed because of #5890.

impaktor:
@robothauler Indeed. I think it would be good to bring back those two lines, but with some other state for "target eliminated - not yet back to base". 

That would be a 2 line fix, but need to wrack our brains for what such a state could be called? PENDING might not be too bad, actually.

But if I'm to think of a bigger change, I imagine what matters to the client is that the target is eliminated before some date, not that they also make it back to base in time. ...unless information flow is slow, and pilot making it back in time is the only way the client will learn that the target has been eliminated. Not sure how much would have to be changed in code and text strings for that, which is why I suggest the above simpler change, if you & other's agree.

Helioth:
btw "Target eliminated" would not be an accurate wording because sometimes such a mission completes without fights.

robothauler:
Correct. A change to the mission status sounds reasonable.

sturnclaw:
@robothauler is this issue addressed by the merge of #6082?

robothauler:
Yes, in my opinion this has been solved.

Issue #5517, opened 2023-02-01
Set destination as navigation target


antonspace:
### Observed behaviour

Some missions, once completed, clicking the "Set destination as navigation target" button in the "Mission", then "more info" page will set the destination to where the mission reward is to be collected, in recent combat missions, it seems this button remains set to the location of the mission, while it is no big deal to search the map and find the final destination, the inconsistency is what is the issue here

Here is the savegame.
[skipjack9.gz](https://github.com/pioneerspacesim/pioneer/files/10557515/skipjack9.gz)

__My pioneer version (and OS):__

Linux.
20230131
![Screenshot_2023-02-01_13-25-10](https://user-images.githubusercontent.com/32020353/216058729-d93c826c-6d99-4390-a5e0-b1477c9069c5.png)

impaktor:
You mean, it should switch back to the station where you're supposed to collect the payment?

antonspace:
Yes, In this case payment was collected upon entering the system,

Another thing, it may be using flavour9, that is if I implemented it correctly.

PR #6082, opened 2025-02-16
Show status "pending" if the mission has not yet been fully completed

robothauler:
Related to #6079. The area for the status is quite small. So too much text does not look good.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -1915,9 +1915,9 @@
     "description": "",
     "message": "Pay fine of {amount}"
   },
-  "PENDING": {
+  "PENDING_RETURN": {
     "description": "Status if the mission has not yet been fully completed",
-    "message": "Pending\nReturn"
+    "message": "Pending Return"
   },
   "PERSEUS_ARM": {
     "description": "Arm of the Milky Way galaxy",
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -1915,6 +1915,10 @@
     "description": "",
     "message": "Pay fine of {amount}"
   },
+  "PENDING": {
+    "description": "Status if the mission has not yet been fully completed",
+    "message": "Pending\nReturn"
+  },
   "PERSEUS_ARM": {
     "description": "Arm of the Milky Way galaxy",
     "message": "Perseus arm"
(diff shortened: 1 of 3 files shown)


zonkmachine:
LGTM (haven't tested). You could let the location button in the mission description swap to the return location on 'Pending'.
https://github.com/pioneerspacesim/pioneer/blob/ddb008daeebdc23716d0acdb516a7f38a33c98b8/data/modules/Combat/Combat.lua#L526

There's a similar combat mission issue on that topic here: https://github.com/pioneerspacesim/pioneer/issues/5517

data/modules/Combat/Combat.lua
```diff
@@ -523,7 +525,12 @@ local buildMissionDescription = function(mission)
 	})
 
 	desc.client = mission.client
-	desc.location = mission.location
+
+	if mission.status == "PENDING" and mission.rendezvous then
+		desc.location = mission.rendezvous
```
zonkmachine, line 530:
```diff
-desc.location = mission.rendezvous
+desc.returnLocation = mission.rendezvous
```
I'm not sure if the only difference there is the message on the button. With **returnLocation** it's set to
`Set return route` instead of `Set destination as navigation target`

zonkmachine: Sorry, this is correct. There is no one clear return station.

data/lang/ui-core/en.json
```diff
@@ -1915,6 +1915,10 @@
     "description": "",
     "message": "Pay fine of {amount}"
   },
+  "PENDING": {
```
sturnclaw, line 1918: The name of the string should be `PENDING_RETURN`.

data/lang/ui-core/en.json
```diff
@@ -1915,6 +1915,10 @@
...
+    "description": "Status if the mission has not yet been fully completed",
+    "message": "Pending\nReturn"
```
sturnclaw, line 1920:
```suggestion
    "message": "Pending Return"
```

Prefer to avoid including manual line-breaks in the translation string. Display code can wrap the text as needed to fit the area.

sturnclaw (changes requested):
Overall I like the change, thanks for proposing it! Just a few minor fixes required; if you're able, I'd like to merge this for the upcoming hotfix.

sturnclaw (approved):

PR #6083, opened 2025-02-21
Read/write save games from/to the savefiles folder

fluffyfreak:
I was trying to use the `savegamedump.exe` to extract the JSON for a few save game file when debugging a crash issue reported on Discord. It refused to load the file and I realised that it was looking in the root of the userfiles location. However the savegames are actually saved in the sub-directory `savefiles`

~~This is a `draft` PR because I wonder if there's a better way to do the pathing. Otherwise I've tested it and it all works~~
I've tested it and it all works

### DIFF
--- a/src/savegamedump.cpp
+++ b/src/savegamedump.cpp
@@ -6,6 +6,9 @@
 #include "core/GZipFormat.h"
 #include <SDL.h>
 
+// HACK, this is copied from /src/SaveGameManager.cpp where it is defined
+static const char s_saveDirName[] = "savefiles";
+
 int info()
 {
 	printf(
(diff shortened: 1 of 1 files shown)


sturnclaw:
SaveGameManager should likely provide a method to retrieve the savefile path. The existing method to retrieve the directory name is meant to be used for logging purposes only (hence why it returns an absolute filesystem path), but for the use of tooling I can definitely endorse providing methods to query the savefile directory within the FileSystem API.

fluffyfreak:
> SaveGameManager should likely provide a method to retrieve the savefile path. The existing method to retrieve the directory name is meant to be used for logging purposes only (hence why it returns an absolute filesystem path), but for the use of tooling I can definitely endorse providing methods to query the savefile directory within the FileSystem API.

Small problem, this requires me to pull in `SaveGameManager.cpp` which brings with it Jobs, Game, etc and an expanding cloud of dependencies and problems to get `savegamedump.exe` to build.

Whereas this PR is currently quite a self contained change for a utility tool.

It's far more hassle than it's worth just so I can type `filename` intead of typing `savefiles/filename`

sturnclaw:
Hmm, yeah, good point. As-is, this is fine. I had originally required the use of the `savegames/` prefix to support loading savefiles from anywhere within the user directory, but that's probably not a needed feature at all (in fact I've never used it once myself).

PR #6084, opened 2025-02-28
System name generator upgrade

ARGHouse:
pioneer_ni2I63KuGN

Update to pioneer including:
-more languages for system name generation (Chinese, Korean, French, Spanish, Kurmanji, Persian, Xhosa, Quechuan, Maori, Aboriginee and a mix of western european that I call 'Westernees'.), this comes with a chance weight means of selecting languages.
-numStars dice improved to give a more realistic distribution of multi-star systems
->singular stars have an ~80% chance of spawning
->trinary stars have an ~19% chance of spawning
->binary stars have a -9% chance of spawning
->quadruple stars have a 0.05% chance of spawning
-Addition of a Random64 header file for larger 64 bit dice, linked to functions in GalaxyGenerator and SectorGenerator so the name generator works (it uses a 64 bit dice).
-Addition of a simple 'Using' file containing keyword shortcuts for variables such as 'const int' etc (Const Uint32 for example gets shortened down to just cui3), handy for time saving during coding.


impaktor:
Ah, your language addition were in C++ side. I did not catch that when you wrote about it [on the forum](https://forum.pioneerspacesim.net/viewtopic.php?p=7295). I know @sturnclaw will have of feedback on this PR.

Btw, please update the "Copyright © 2008-2024" to 2025.

sturnclaw (changes requested):
Hi @ARGHouse, thanks for the PR! Due to the fact that this PR changes the generated galaxy, this will be unable to be considered for merge until after the upcoming hotfix.

Consequently, I will be deferring a full in-depth review until that time. However, I will address some immediate stand-out items in the diff that will need to be addressed before merge:

- By convention in this codebase, the pointer specifier `*` and reference specifier `&` are right-aligned to be adjacent to the variable or function name, not the type. Please make the appropriate configuration to your automatic formatting tool, as I see a number of diff sites just changing the alignment of those characters.
- Your `Random64` class still provides a 32-bit random hash as `pcg32` writes a 32-bit output value; the upper 32 bits of the generated value are empty.
    Instead of copy-pasting the entire file, please just add an `Int64()` method to the existing `Random` class in a separate PR.
- Pioneer is not interested in adopting the list of shortcuts proposed in `Using.h`. By convention the codebase prefers to use standard fixed-size types only (e.g. uint32_t), and I would strongly prefer to avoid introducing additional alternate names for fundamental types, as it introduces significant cognitive load when reading the codebase and makes it much more difficult for a new contributor / new programmer to work with.
    If you are interested in saving yourself typing load, I recommend adding these shortcuts to your editor's autocomplete/snippet function instead.

sturnclaw (changes requested):
Please note that we are unable to accept copyrighted material or material original to another fictional universe. For that reason, the Sebacean and Aure(k)-Besh phonetic lists will need to be removed from this PR.

I'd additionally like to know what sources (if any) you used for each name list. Per discussion on #6008, any AI-generated content will need further scrutiny to ensure the generated material is compliant with the GPLv3 license our source code falls under.

sturnclaw:
I am considering moving name lists to be driven by JSON configuration files rather than hard-coded into `NameGenerator.cpp`. Most of the name lists you've provided here would be relatively easy to translate; if you are interested in waiting for that code to be written, the Aurek-Besh and Sebacean lists could live on as a personal mod if you so desire.

sturnclaw:
Converting to draft due to inactivity.

src/Random64.h
```diff
@@ -0,0 +1,274 @@
+// Copyright © 2008-2024 Pioneer Developers. See AUTHORS.txt for details
```
impaktor, line 1: You're living in the past, man...

src/Using.h
```diff
@@ -0,0 +1,63 @@
+#pragma once
```
impaktor, line 1: No copyright header?

PR #6086, opened 2025-03-03
Resurrect region based terrain flattening

### LINKED ISSUE #5154: The Pads goes into the ground

fluffyfreak:
This was originally done by @Ae-2222 many (13/14) years ago in [this branch](https://github.com/Ae-2222/pioneer/tree/RegionBasedTerrainControl) and I've resurrected it here because issue #5154 is still a problem.

Fixes #5154

This works by querying for starports on a planet, adding them to a list, then checking that list whenever a height sample is requested... this has some performance implications but on my high end machine it runs very well. I've not yet tried it on a Raspberry Pi...

Putting this up as a draft to started discussion.

I have:
- reworked how it gets setup and initialised
- changed what was stored a little

Things I think it needs:
- optimising
- the blend method is strange to me but it works with a small enough region/city size set

**Issues:**
**- inconsistent results across machines!**
**- I don't think that it currently covers all starports on a planet!**
**- somtimes flattens terrain away from a starport whilst not flattening the starport**

Image taken at (0, -2, -2) `Gliese 1`
PioneerGliese1

### DIFF
--- a/src/GeoSphere.cpp
+++ b/src/GeoSphere.cpp
@@ -217,6 +217,8 @@ GeoSphere::GeoSphere(const SystemBody *body) :
 	}
 
 	//SetUpMaterials is not called until first Render since light count is zero :)
+
+	m_terrain->InitCityRegions(body);
 }
 
 GeoSphere::~GeoSphere()
--- a/src/terrain/Terrain.cpp
+++ b/src/terrain/Terrain.cpp
@@ -633,7 +633,7 @@ Terrain::~Terrain()
 {
 }
 
-static constexpr double TARGET_CITY_RADIUS = 5000.0;
+static constexpr double TARGET_CITY_RADIUS = 600.0;
 
 // Set up region data for each of the system body's child surface starports
 void Terrain::InitCityRegions(const SystemBody *sb)
--- a/src/terrain/Terrain.cpp
+++ b/src/terrain/Terrain.cpp
@@ -634,36 +634,35 @@ Terrain::~Terrain()
 }
 
 static constexpr double TARGET_CITY_RADIUS = 5000.0;
-#pragma optimize(off, "")
+
 // Set up region data for each of the system body's child surface starports
-void Terrain::InitCityRegions(const SystemBody *body)
+void Terrain::InitCityRegions(const SystemBody *sb)
 {
 	m_regionTypes.clear();
 	m_positions.clear();
 
-	const SystemBody *&sb = body;
 	if (!sb->HasChildren()) {
 		return;
 	}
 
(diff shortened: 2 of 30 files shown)


sturnclaw:
`O(N)` dot-product-and-branches is not great when we're running that for every single terrain point generated on the entire planet with potentially 20+ flattening regions. That's not a condemnation of you or the original author, to be clear, but something like this IMO would be a much better fit for a GPU-based terrain generation pipeline.

However, for the purpose of getting *something* in, I'd personally envision the API as passing a span of valid flatten-region structs into the `GetHeight()` call (maybe even a `GetHeightWithOverride()` method wrapping `GetHeight()` so you don't have to `ApplySimpleHeightRegions()` in every terrain generator?), and making it the responsibility of GeoPatch to compute the list of applicable flattening regions to pass into the call for the patch in question.

The logic for that can be made pretty simple; each patch maintains a `std::vector<SimpleHeightRegion> *` until there's only one height region in the patch and then it's just a direct `SimpleHeightRegion *` passed down to all applicable children. In the view of optimizing for terrain generation time, it's probably better to maintain contiguous lists of height regions per patch than it is to try to store indices and pay the performance cost of doing non-contiguous pointer-chasing for each terrain sample generated.

Regardless of how it's implemented, this does potentially run the risk of making terrain height values inconsistent based on the source that's querying terrain height (e.g. collision, rendering, etc.). I'd want to make sure we're presenting a consistent model of terrain, as future graphical features (e.g. surface-scattered objects like grass, etc.) depend on queried terrain height matching the graphical model of terrain.

I'm not sure there's an elegant solution to that problem though, outside of stuffing significantly more logic into the physics engine to generate collision representation meshes "on demand" based on nearby potential-colliders rather than doing arbitrary terrain generation queries every single collision check.

fluffyfreak:
Great feedback! I'll have a think about resolving both of those issues

fluffyfreak:
Tracking further development of this in https://github.com/fluffyfreak/pioneer/tree/regions-branch

> Regardless of how it's implemented, this does potentially run the risk of making terrain height values inconsistent based on the source that's querying terrain height 

When something queries the height via GetHeight that is done on the BaseSphere -> GeoSphere which holds the terrain, so it can all be unified through that 👍 should be consistent. 

Hard part right now is making sure that it's consistent with the GeoPatchJobs system. Finding the appropriate region(s) to pass through to those jobs because at high levels of the quadtree there could be many regions involved.

Issue #6047, opened 2025-02-06
Often crash when loading seemingly without reason


DynV:
### Observed behaviour
The game keep crashing when loading seemingly without reason. I tried to pinpoint the reason multiple times. I want to set my game as it was in the previous official release, so I set my character appearance, kill count, reputation and ship to what I had then save the same, I then grab missions while going between starports hoping to get all my ship items back. I've done that at least twice, wasting > 1/2 hr each time. I then started to save every few minutes then loaded the save to see if the load would crash; this seems like a perfect recipe to corrupt my gamesave but I don't see any other way to not waste many minutes.

I've tried to pinpoint a reason by starting the game again multiple times. As well as upgrade my ship a few items at a time to see if one caused a problem; I was convinced one caused the loading crash but then it turned out not to be the case. What I'm thinking is that some items are indeed buggy but not consistently, that they cause a bug the majority of the time but not always.

Using my paranoid save every few minutes then reload it right after if the loading crash the game, there was no crash for some time but then there was
[Pioneer-250203-gamesaves-Bradbury_to_Olympus.zip](https://github.com/user-attachments/files/18682268/Pioneer-250203-gamesaves-Bradbury_to_Olympus.zip)
This is just one of multiple gamesave that caused a crash when loading, not to point to something specific about it. The difference between those 2 gamesaves is that I picked mission at Bradbury Landing then saved as soon as I got to Olympus Mons.

**Update:**
The gamesave that crashes was basically remade successfully in #6049 

### Expected behaviour
Both gamesaves to load properly.

### Steps to reproduce
Load the latest of the 2 gamesaves, with highest #, which makes the latest official release crash. Load the previous gamesave, with lowest #, which doesn't make the game crash.

__My pioneer version (and OS):__
250203 Windows

__My output.txt (required) and game save (optional, but recommended)__
[output.txt](https://github.com/user-attachments/files/18682412/output.txt)
Gamesave in section Observed behaviour.

sturnclaw:
This should be fixed in #6070.

PR #6087, opened 2025-03-05
Minor optimisations and tweaks

fluffyfreak:
Optimisations parts:
- Inlined methods and added a Normalise-in-place to vector3d
- Batched generating heights, avoid calling GetHeight 1225 times and instead request all heights at once
- Forced override of GetHeight in BaseSphere since it always is anyway
- `m_entropy[12]` is only used in 2 places and only ever used the first value but refreshed it hundreds of times every Terrain Ctor

Tweak:
- Display the Height and Color fractal names used for a terrain in ObjectViewerView (Ctrl+F10)

These changes probably do affect the terrain generation though I haven't noticed massive changes in testing. Unsure if we'll need a savegame bump

### DIFF
--- a/src/GeoPatchJobs.cpp
+++ b/src/GeoPatchJobs.cpp
@@ -30,23 +30,31 @@ void SSingleSplitRequest::GenerateMesh() const
 {
 	PROFILE_SCOPED()
 	const int borderedEdgeLen = edgeLen + (BORDER_SIZE * 2);
-#ifndef NDEBUG
 	const int numBorderedVerts = borderedEdgeLen * borderedEdgeLen;
-#endif
 
 	// generate heights plus a 1 unit border
-	double *bhts = borderHeights.get();
-	vector3d *vrts = borderVertexs.get();
+	std::vector<vector3d> positions;
+	positions.reserve(numBorderedVerts);
+	std::vector<double> heightsOut;
+	heightsOut.resize(numBorderedVerts);
 	for (int y = -BORDER_SIZE; y < borderedEdgeLen - BORDER_SIZE; y++) {
 		const double yfrac = double(y) * fracStep;
 		for (int x = -BORDER_SIZE; x < borderedEdgeLen - BORDER_SIZE; x++) {
 			const double xfrac = double(x) * fracStep;
 			const vector3d p = GetSpherePoint(v0, v1, v2, v3, xfrac, yfrac);
-			const double height = pTerrain->GetHeight(p);
-			assert(height >= 0.0f && height <= 1.0f);
-			*(bhts++) = height;
-			*(vrts++) = p * (height + 1.0);
+			positions.emplace_back(p);
 		}
+	}
+	pTerrain->GetHeights(positions, heightsOut);
+
+	double *bhts = borderHeights.get();
+	vector3d *vrts = borderVertexs.get();
+	for (int iHeights = 0; iHeights < numBorderedVerts; iHeights++)
+	{
+		const vector3d &p = positions[iHeights];
+		const double height = heightsOut[iHeights];
+		*(bhts++) = height;
+		*(vrts++) = p * (height + 1.0);
 	}
 	assert(bhts == &borderHeights.get()[numBorderedVerts]);
 
@@ -190,23 +198,30 @@ void SQuadSplitRequest::GenerateBorderedData() const
 {
 	PROFILE_SCOPED()
 	const int borderedEdgeLen = (edgeLen * 2) + (BORDER_SIZE * 2) - 1;
-#ifndef NDEBUG
 	const int numBorderedVerts = borderedEdgeLen * borderedEdgeLen;
-#endif
 
 	// generate heights plus a N=BORDER_SIZE unit border
-	double *bhts = borderHeights.get();
-	vector3d *vrts = borderVertexs.get();
+	std::vector<vector3d> positions;
+	positions.reserve(numBorderedVerts);
+	std::vector<double> heightsOut;
+	heightsOut.resize(numBorderedVerts);
 	for (int y = -BORDER_SIZE; y < (borderedEdgeLen - BORDER_SIZE); y++) {
 		const double yfrac = double(y) * (fracStep * 0.5);
 		for (int x = -BORDER_SIZE; x < (borderedEdgeLen - BORDER_SIZE); x++) {
 			const double xfrac = double(x) * (fracStep * 0.5);
 			const vector3d p = GetSpherePoint(v0, v1, v2, v3, xfrac, yfrac);
-			const double height = pTerrain->GetHeight(p);
-			assert(height >= 0.0f && height <= 1.0f);
-			*(bhts++) = height;
-			*(vrts++) = p * (height + 1.0);
+			positions.emplace_back(p);
 		}
+	}
+	pTerrain->GetHeights(positions, heightsOut);
+
+	double *bhts = borderHeights.get();
+	vector3d *vrts = borderVertexs.get();
+	for (int iHeights = 0; iHeights < numBorderedVerts; iHeights++) {
+		const vector3d &p = positions[iHeights];
+		const double height = heightsOut[iHeights];
+		*(bhts++) = height;
+		*(vrts++) = p * (height + 1.0);
 	}
 	assert(bhts == &borderHeights[numBorderedVerts]);
 }
--- a/src/GeoSphere.h
+++ b/src/GeoSphere.h
@@ -35,15 +35,18 @@ public:
 	void Update() override;
 	void Render(Graphics::Renderer *renderer, const matrix4x4d &modelView, vector3d campos, const float radius, const std::vector<Camera::Shadow> &shadows) override;
 
-	double GetHeight(const vector3d &p) const final
+	double GetTerrainHeight(const vector3d &p) const final
 	{
-		const double h = m_terrain->GetHeight(p);
+		std::vector<vector3d> pv{p};
+		std::vector<double> hv(size_t(1));
+		m_terrain->GetHeights(pv, hv);
+		const double h = hv[0];
 #ifndef NDEBUG
 		// XXX don't remove this. Fix your fractals instead
 		// Fractals absolutely MUST return heights >= 0.0 (one planet radius)
 		// otherwise atmosphere and other things break.
 		if (h < 0.0) {
-			Output("GetHeight({ %f, %f, %f }) returned %f\n", p.x, p.y, p.z, h);
+			Output("GetTerrainHeight({ %f, %f, %f }) returned %f\n", p.x, p.y, p.z, h);
 			m_terrain->DebugDump();
 			assert(h >= 0.0);
 		}
--- a/src/TerrainBody.cpp
+++ b/src/TerrainBody.cpp
@@ -143,7 +143,7 @@ double TerrainBody::GetTerrainHeight(const vector3d &pos_) const
 {
 	double radius = m_sbody->GetRadius();
 	if (m_baseSphere) {
-		return radius * (1.0 + m_baseSphere->GetHeight(pos_));
+		return radius * (1.0 + m_baseSphere->GetTerrainHeight(pos_));
 	} else {
 		assert(0);
 		return radius;
--- a/src/TerrainBody.cpp
+++ b/src/TerrainBody.cpp
@@ -155,3 +155,20 @@ void TerrainBody::OnChangeDetailLevel(Graphics::Renderer *r)
 {
 	BaseSphere::OnChangeDetailLevel(r);
 }
+
+const char *TerrainBody::GetHeightFractalName() const
+{
+	if (Terrain* terrain = m_baseSphere->GetTerrain())
+	{
+		return terrain->GetHeightFractalName();
+	}
+	return nullptr;
+}
+
+const char* TerrainBody::GetColorFractalName() const
+{
+	if (Terrain *terrain = m_baseSphere->GetTerrain()) {
+		return terrain->GetColorFractalName();
+	}
+	return nullptr;
+}
--- a/src/terrain/Terrain.h
+++ b/src/terrain/Terrain.h
@@ -91,7 +91,7 @@ protected:
 	double m_invPlanetRadius;
 	double m_planetEarthRadii;
 
-	double m_entropy[12];
+	double m_entropy;
 
 	vector3d m_rockColor[8];
 	vector3d m_darkrockColor[8];
--- a/src/terrain/Terrain.h
+++ b/src/terrain/Terrain.h
@@ -43,8 +43,7 @@ public:
 		return m_fracdef[index];
 	}
 
-	virtual double GetHeight(const vector3d &p) const = 0;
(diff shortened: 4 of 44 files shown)


fluffyfreak:
I haven't seen the issues with save game loading you're seeing bodasey. Can you give me some reproducation steps, like where you started, and what you did?

It does seem to change generation enough that there would need to be a savegame bump, but I've had no issues with saving a loading save generated by this branch.

Bodasey:
This happened during game play, just after the program start - whatever has been saved with this branch could not be loaded again and terminated pioneer. I'll upload my saves here when my pioneer special computer is on again, do you upload yours, please?

Bodasey:
Maybe related to #6047 and #6064

Bodasey:
[2025-Test-surface-Makenzie.txt](https://github.com/user-attachments/files/19315323/2025-Test-surface-Makenzie.txt)
[2025-Test-surface.txt](https://github.com/user-attachments/files/19315324/2025-Test-surface.txt)
[558641.txt](https://github.com/user-attachments/files/19315326/558641.txt)

fluffyfreak:
Nah this is just due review really, but thanks @Bodasey

sturnclaw:
Right - I'll review this soon, been putting it off until we kick a hotfix due to the savegame bump.

src/GeoPatchJobs.cpp
```diff
@@ -30,24 +30,32 @@ void SSingleSplitRequest::GenerateMesh() const
...
+
+	double *bhts = borderHeights.get();
+	vector3d *vrts = borderVertexs.get();
+	for (int iHeights = 0; iHeights < numBorderedVerts; iHeights++)
+	{
+		const vector3d &p = positions[iHeights];
+		const double height = heightsOut[iHeights];
+		*(bhts++) = height;
+		*(vrts++) = p * (height + 1.0);
+	}
```
sturnclaw: Note that instead of allocating an intermediate buffer for positions, you can instead fill the `borderVertexs` and `borderHeights` arrays in the initial pass, and in this pass just modify the `borderVertexs` array like so:
```c++
vrts[i] *= (bhts[i] + 1.0);
```

This results in ~1/6th - ~1/8th less bandwidth in this pass, as you're not copying the heights array from an intermediate buffer.

src/GeoPatchJobs.cpp
```diff
@@ -190,24 +198,31 @@ void SQuadSplitRequest::GenerateBorderedData() const
...
 	}
+	pTerrain->GetHeights(positions, heightsOut);
+
+	double *bhts = borderHeights.get();
+	vector3d *vrts = borderVertexs.get();
+	for (int iHeights = 0; iHeights < numBorderedVerts; iHeights++) {
+		const vector3d &p = positions[iHeights];
+		const double height = heightsOut[iHeights];
+		*(bhts++) = height;
+		*(vrts++) = p * (height + 1.0);
```
sturnclaw: Ditto.

src/terrain/Terrain.h
```diff
@@ -43,7 +43,7 @@ class Terrain : public RefCounted {
 		return m_fracdef[index];
 	}
 
-	virtual double GetHeight(const vector3d &p) const = 0;
+	virtual void GetHeights(const std::vector<vector3d> &vP, std::vector<double> &heightsOut) const = 0;
```
sturnclaw: Note: in lieu of a proper span type, I'd recommend this function signature instead be:
```suggestion
	virtual void GetHeights(const vector3d *vP, double *heightsOut, size_t count) const = 0;
```

This prevents the need to construct a single-element vector, properly expresses that the number of vertices must equal the number of heights generated, and allows for more flexibility in the underlying storage.

Regardless of function signature, bounds checking *must* be performed in the calling function - this just makes that more explicit.

ALSO: you *really* don't want to be calling the bounds-checking `std::vector::at` method for every single iteration of the noise function, nor do you want to be resizing the vector mid-fill. Both have significant performance implications at the quantity of vertices we process.

src/terrain/TerrainHeightAsteroid.cpp
```diff
@@ -23,9 +23,11 @@ TerrainHeightFractal<TerrainHeightAsteroid>::TerrainHeightFractal(const SystemBo
...
-double TerrainHeightFractal<TerrainHeightAsteroid>::GetHeight(const vector3d &p) const
+void TerrainHeightFractal<TerrainHeightAsteroid>::GetHeights(const std::vector<vector3d> &vP, std::vector<double> &heightsOut) const
 {
-	const double n = octavenoise(GetFracDef(0), 0.4, p) * dunes_octavenoise(GetFracDef(1), 0.5, p);
-
-	return (n > 0.0 ? m_maxHeight * n : 0.0);
+	for (size_t i = 0; i < vP.size(); i++) {
+		const vector3d &p = vP[i];
+		const double n = octavenoise(GetFracDef(0), 0.4, p) * dunes_octavenoise(GetFracDef(1), 0.5, p) * m_maxHeight;
+		heightsOut.at(i) = (n > 0.0 ? n : 0.0);
```
sturnclaw: `std::vector::at()` must die - repeat for following callsites.

src/terrain/TerrainHeightHillsCraters.cpp
```diff
@@ -24,21 +24,27 @@ TerrainHeightFractal<TerrainHeightHillsCraters>::TerrainHeightFractal(const Syst
...
+	for (size_t i = 0; i < vP.size(); i++) {
+		const vector3d &p = vP[i];
+		const double continents = octavenoise(GetFracDef(0), 0.5, p) - m_sealevel;
+		if (continents < 0.0)
+			heightsOut.at(i) = 0.0;
+
+		// == TERRAIN_HILLS_NORMAL except river_octavenoise
+		double n = 0.3 * continents;
+		double distrib = river_octavenoise(GetFracDef(2), 0.5, p);
+		double m = GetFracDef(1).amplitude * river_octavenoise(GetFracDef(1), 0.5 * distrib, p);
```
sturnclaw: While the compiler can indeed (somewhat) trivially see that the underlying `m_fracdef` is const during the loop and move this load outside of the loop in release mode, I'd ideally prefer `GetFracDef(n)` be replaced with `m_fracdef[n]` for debug-mode performance, given that we call it up to 8 times per loop iteration.

src/terrain/TerrainHeightMapped2.cpp
```diff
@@ -17,18 +17,20 @@ TerrainHeightFractal<TerrainHeightMapped2>::TerrainHeightFractal(const SystemBod
...
+		v *= m_invPlanetRadius;
 
-	v += 0.1;
-	double h = 1.5 * v * v * v * ridged_octavenoise(16, 4.0 * v, 4.0, p);
-	h += 30000.0 * v * v * v * v * v * v * v * ridged_octavenoise(16, 5.0 * v, 20.0 * v, p);
-	h += v;
-	h -= 0.09;
+		v += 0.1;
+		double h = 1.5 * v * v * v * ridged_octavenoise(16, 4.0 * v, 4.0, p);
+		h += 30000.0 * v * v * v * v * v * v * v * ridged_octavenoise(16, 5.0 * v, 20.0 * v, p);
```
sturnclaw, line 30: If you happen to remember the reason for this string of `v`s I'd love a comment as to the underlying mathematical function being expressed here.

src/vector3.h
```diff
@@ -32,116 +32,123 @@ class alignas(sizeof(T)) vector3 {
...
+	inline T LengthSqr() const { return x * x + y * y + z * z; }
+	inline vector3 Lerp(const vector3 &b, const double percent) const
 	{
 		return *this + percent * (b - *this);
 	}
-	vector3 Normalized() const
+	inline void Normalize()
 	{
-		const T l = 1.0f / sqrt(x * x + y * y + z * z);
+		const T l = 1.0f / Length();
```
sturnclaw: On the topic of debug-mode performance where nothing gets inlined - I'd prefer we retain the `sqrt(...)` call here rather than dispatching to another one-liner member function.

src/vector3.h
```diff
@@ -32,116 +32,123 @@ class alignas(sizeof(T)) vector3 {
...
+	inline vector3 Normalized() const
+	{
+		const T l = 1.0f / Length();
 		return vector3(x * l, y * l, z * l);
 	}
-	vector3 NormalizedSafe() const
+	inline vector3 NormalizedSafe() const
 	{
-		const T lenSqr = x * x + y * y + z * z;
+		const T lenSqr = LengthSqr();
```
sturnclaw: Ditto - IMO `x*x + y*y + z*z` is not sufficiently complicated that having a single-source-of-truth is better than having optimized code in debug mode.

src/vector3.h
```diff
@@ -150,7 +157,7 @@ class alignas(sizeof(T)) vector3 {
 		}
 	}
 
-	void Print() const { printf("v(%f,%f,%f)\n", x, y, z); }
+	inline void Print() const { printf("v(%f,%f,%f)\n", x, y, z); }
```
sturnclaw: Do we... need this? I'd much rather drop the header-level dependency on stdio.h.

sturnclaw (changes requested):
At long last, I have feedback. Apologies in advance for being nothing but critical in the comments - on the whole I quite like this PR and it will be incredibly instrumental for future advances in terrain generation performance.

I think I have a stash somewhere which reimplemented terrain noise using a SIMD library and the big hangup was that noise values were generated one-at-a-time for a single terrain point - hopefully this PR will make that stash a bit more of a performance win.

fluffyfreak:
> At long last, I have feedback. Apologies in advance for being nothing but critical in the comments - on the whole I quite like this PR and it will be incredibly instrumental for future advances in terrain generation performance.
> 
> I think I have a stash somewhere which reimplemented terrain noise using a SIMD library and the big hangup was that noise values were generated one-at-a-time for a single terrain point - hopefully this PR will make that stash a bit more of a performance win.

Getting to a SIMD terrain generator is exactly how I started down this mad plan.

ALso critique is fine this was a bit experimental, I'll have to remember everything I did now 😁

src/terrain/TerrainHeightAsteroid.cpp
fluffyfreak: I really do not know or remember why I used `.at()` 🤷 will fix asap.
Maybe I was trying to catch something whilst developing/debugging 🤔

src/GeoSphere.h
```diff
@@ -35,15 +35,18 @@ class GeoSphere : public BaseSphere {
...
 	void Render(Graphics::Renderer *renderer, const matrix4x4d &modelView, vector3d campos, const float radius, const std::vector<Camera::Shadow> &shadows) override;
 
-	double GetHeight(const vector3d &p) const final
+	double GetTerrainHeight(const vector3d &p) const final
 	{
-		const double h = m_terrain->GetHeight(p);
+		std::vector<vector3d> pv{p};
+		std::vector<double> hv(size_t(1));
+		m_terrain->GetHeights(pv, hv);
+		const double h = hv[0];
```
sturnclaw: Note that if `Terrain::GetHeights(...)` takes two pointers and a size, you save the overhead of two single-element heap allocations and consume about 40 bytes less stack space here.
```suggestion
		double h = 0;
		m_terrain->GetHeights(&p, &h, 1);
```

src/terrain/TerrainHeightMapped2.cpp
fluffyfreak: Absolutely no idea, there's been more than a decade of discussion over just wtf all of this handcrafted (_artisinal_) noise generation actually does. It's why I've got branches languishing that change it to json etc etc

src/vector3.h
fluffyfreak: I checked and it's not used, so I'll remove it and the ones from vector2, matrix3x3, and matrix4x4 as well!

src/terrain/TerrainHeightHillsCraters.cpp
fluffyfreak: If it's ok I'm not going to touch this one for this PR. It's just used everywhere in the terrain noise and always has been. So I'd rather do that as a separate PR

src/GeoPatchJobs.cpp
fluffyfreak: Not going to lie I had to really think about this one for a while! 🤣
I've been looking at and maintaining this code for so many years that this just just never occurred to me! Excellent suggestion!

src/terrain/TerrainHeightHillsCraters.cpp
sturnclaw: That's fine - I was mentioning it as it'd be a simple find-replace of `GetFracDef` -> `m_fracDef`. A separate PR would be nice though!

src/GeoPatchJobs.cpp
```diff
@@ -33,28 +33,23 @@ void SSingleSplitRequest::GenerateMesh() const
 	const int numBorderedVerts = borderedEdgeLen * borderedEdgeLen;
 
 	// generate heights plus a 1 unit border
-	std::vector<vector3d> positions;
-	positions.reserve(numBorderedVerts);
-	std::vector<double> heightsOut;
-	heightsOut.resize(numBorderedVerts);
+	vector3d *vrts = borderVertexs.get();
```
sturnclaw: Nitpick: as this pointer is mutated during the fill loop, I'd prefer it not be reused later in a separate loop, instead naming it something like `fill_verts`.

src/TerrainBody.cpp
```diff
@@ -155,3 +155,20 @@ void TerrainBody::OnChangeDetailLevel(Graphics::Renderer *r)
 {
 	BaseSphere::OnChangeDetailLevel(r);
 }
+
+const char *TerrainBody::GetHeightFractalName() const
+{
+	if (Terrain* terrain = m_baseSphere->GetTerrain())
+	{
```
sturnclaw: Nitpick: the stylistic convention is for if-braces to be on the line of the if-condition. (I personally like this style too, but the former is the convention the code is written in.)

sturnclaw (approved):
A few minor nitpicks but this looks good! I'm going to beg off merging this immediately as I don't want to break save compatibility quite yet, but it will get pulled in this release cycle.

fluffyfreak:
Tidied up the commits a little

fluffyfreak:
Actually replacing the calls to `GetFracDef` with `m_fracdef` ended up being trivial so it's gone into this PR as well 👍

fluffyfreak:
I've rebased and this appears to just be failing on data building which this PR doesn't affect.
Probably just some CI error

fluffyfreak:
@sturnclaw am I ok to merge/rebase-merge this?

fluffyfreak:
@sturnclaw one of the things which could be great to add as an optimisation is (_as well as a SIMD library_) [Simplex Derivatives](https://catlikecoding.com/unity/tutorials/pseudorandom-surfaces/simplex-derivatives/) to generate the normals directly from the derivatives. You can also get things like the slope I think. 
[Some libraries](https://github.com/simongeilfus/SimplexNoise/tree/master) already provide versions which supply derivatives from basic noise.

Sadly I am failing to understand the maths due to being dumb 😁

Issue #5760, opened 2024-02-11
Bulletin Board All Items Shown


Digit:
### Observed behaviour
Bulletin Board is populated with many dozens of missions etc.

### Expected behaviour
Typically only maybe a couple scan missions if even, maybe a dozen or two missions/items max total.

### Steps to reproduce
* Yet to determine.  
** Half notion may relate to save reloading, but that's a tenuously formed notion, may be a red-herring
*** (Marginally less tenuous a notion, when encountering this before, _[(while saves were not retaining shop state from a prior bug(... I thought I knew which to link, but failed to find ... maybe mandela effect, heh.)]_, an exit out and load back in and this could be reverted to a normal bulletin board listing... ~ I think.  not 100% sure about that.  Vague recollection suggests.   But now, I caught a save with that overpopulated bulletinboard, that reloads with the bulletin board still overloaded.)

__My pioneer version (and OS):__
* Pioneer version: Master branch, 48efc4667, 
* OS: Linux 

__My output.txt (required) and game save (optional, but recommended)__
* output: http://ks392457.kimsufi.com/stuff/pioneer/output.txt
* Save with example overpopulated bulletin board at Poseidon station : http://ks392457.kimsufi.com/stuff/pioneer/M20240130%20b%20mola%20hd4%20100

__Screenshots:__
http://ks392457.kimsufi.com/stuff/pioneer/scrots/
I made these with a scrot ; sleep 1 ; scrot ; sleep 1 type of command chain while I scrolled through.  These screenshots give some idea what it's like, though by far not all that's there.  For example, look how small the scrollbar handle is.
![1](http://ks392457.kimsufi.com/stuff/pioneer/scrots/2024-02-11-094944_1920x1080_scrot.png)
![2](http://ks392457.kimsufi.com/stuff/pioneer/scrots/2024-02-11-094948_1920x1080_scrot.png)
![3](http://ks392457.kimsufi.com/stuff/pioneer/scrots/2024-02-11-094952_1920x1080_scrot.png)
![4](http://ks392457.kimsufi.com/stuff/pioneer/scrots/2024-02-11-094956_1920x1080_scrot.png)
![5](http://ks392457.kimsufi.com/stuff/pioneer/scrots/2024-02-11-095000_1920x1080_scrot.png)
![6](http://ks392457.kimsufi.com/stuff/pioneer/scrots/2024-02-11-095003_1920x1080_scrot.png)
[...]

PR #6095, opened 2025-03-22
Balancing the Scout mission ads

zonkmachine:
In response to https://github.com/pioneerspacesim/pioneer/issues/5760 where the BBS over time lose balance as some modules take over. Especially the Scout module keeps adding missions in excess of the demand. This is a quick fix to soften the issue and with this PR the missions balance out at around ~30 missions per station if none is take by the player.

The fix works by creating fewer adverts for Scout missions. One sixth of previous rate.

I haven't tested this other than sitting at Cydonia for 9 months and watching the mission list build up.
Here is what it looks like on master when the missions build up while not doing anything.
```
Number of ads per module after...
                     ~2 months            ~4 months
------------------------------------------------------
Assassination               10                  5
Cargo run                   21                 32
Scoop                       ~1                  1
Scout                      119 !!!            174 ???
Combat                       6                  4
Deliver Package             24                 24
Search and rescue            3                  5
Second hand                  2                  1
Taxi                        34                 38
```

### DIFF
--- a/data/modules/Scout/Scout.lua
+++ b/data/modules/Scout/Scout.lua
@@ -568,14 +568,14 @@ end
 local onUpdateBB = function (station)
 	for ref,ad in pairs(ads) do
 		if not flavours[ad.flavour].localscout
-			and ad.due < Game.time + 432000 then -- 5 days
+			and ad.due < Game.time + 5*24*60*60 then
 			ad.station:RemoveAdvert(ref)
 		elseif flavours[ad.flavour].localscout
-			and ad.due < Game.time + 172800 then -- 2 days
+			and ad.due < Game.time + 2*24*60*60 then
 			ad.station:RemoveAdvert(ref)
 		end
 	end
-	if Engine.rand:Integer(43200) < 3600 then    -- 12 h < 1 h
+	if Engine.rand:Integer(3*24*60*60) < 60*60 then
 		makeAdvert(station)
 	end
 end


data/modules/Scout/Scout.lua
```diff
@@ -568,14 +568,14 @@ end
 local onUpdateBB = function (station)
 	for ref,ad in pairs(ads) do
 		if not flavours[ad.flavour].localscout
-			and ad.due < Game.time + 432000 then -- 5 days
+			and ad.due < Game.time then -- 5 days
 			ad.station:RemoveAdvert(ref)
 		elseif flavours[ad.flavour].localscout
-			and ad.due < Game.time + 172800 then -- 2 days
+			and ad.due < Game.time then -- 2 days
```
impaktor, line 574: isn't the comment wrong now?  172800 was the 2 days? Same on L571 with 5 days

impaktor (commented):
I still like my way of doing it: #3243
Perhaps over-complicated.
To each his own.

data/modules/Scout/Scout.lua
zonkmachine: Thanks, fixed!

zonkmachine:
> I still like my way of doing it: #3243 Perhaps over-complicated. To each his own.

I love your way of doing it. This is just me getting stressed up when it's time for release and I see a random bug. It looked easy and safe enough to make it into the next release. It's quite an annoying bug. As for the method, I think your way of doing it is super nice and I think this should be revisited for a proper makeover.

zonkmachine:
Marked as a draft as I found some issues. Will look into this in the evening.

zonkmachine:
> Marked as a draft as I found some issues. Will look into this in the evening.

I misread the code for ad removal in onUpdateBB and have reverted that part of the PR but instead made some changes for improved clarity.

The also softened the change in adverts created just slightly.

zonkmachine:
Did a last push to make the code more clear. Squashed and done.
Merge?

zonkmachine:
Can I merge this?

The only change I would do personally is do an even more drastic cut in Scout missions but I think maybe it's best to gradually dial in the right level. There is also the prospect of having a more advanced algorithm that takes population and other system data into consideration.

impaktor:
> Can I merge this?

Fine by me

PR #6098, opened 2025-03-24
Refresh the icons if the source changed

reinerh:
This change allows rebuilding the icons via CMake and will rebuild them when the svg files are updated.


sturnclaw:
As you'll notice by the failing checks, this PR has the knock-on effect of making `optipng`, `inkscape` and potentially `icotool` required dependencies to build Pioneer. This is not a desired outcome; I'd much prefer that you instead change the scope of the CMake command to re-generate the `application-icon/` files in the source tree, and only when manually run.

More to the point, the application icon is something that we do not expect to change very often (if at all; at most we intend to change it once in the foreseeable future). I'm not sure why you felt the need to wire it up to the build system so that it is automatically regenerated on (potentially) every build? The existing shell script is more than adequate for the anticipated once-a-decade usage.

reinerh:
You're right. It's sufficient to regenerate it when it really changes.

PR #6102, opened 2025-03-30
Combat mission - Strings

zonkmachine:
Fixed wonky strings in the Combat mission.

 * Wrong case
 * a/an
 * Change name of variable 'type', which is a Lua built-in function.

**Before**
combat

**After**
combat3


zonkmachine:
It's a bug fix but if you restart an old project with it there may be some minor issues. The one I've found is the article being used twice. As in the first example above. **"a an armed reconnaissance"**

combat2

data/lang/module-combat/en.json
```diff
@@ -271,6 +271,18 @@
     "description": "Name of mission type. Derived from Fighter Sweep. An aggressive mission to hunt down and destroy every enemy fighter in the allotted area.",
     "message": "Area Sweep"
   },
+  "MISSION_TYPE_CONVO_1": {
+    "description": "Name of mission type, used in conversation",
+    "message": "a reconnaissance"
+  },
+  "MISSION_TYPE_CONVO_2": {
+    "description": "Name of mission type, used in conversation",
+    "message": "an armed reconnaissance"
```
impaktor, line 280: I wonder if this makes it hard to translate, when it's a fraction of a sentence. Perhaps description could mention which other translation variable makes use of it, and/or provide example sentence with surrounding context?

impaktor:
git grep -i " a {" data/lang/*en.json 

search and rescue might have similar sentence construction?

sturnclaw (approved):
Eventually we may want a more powerful translation system but for now this is a good solution - thanks!

PR #5852, opened 2024-07-02
Fix bad placeholders in translated strings

impaktor:
This mostly fixes #5849 (translators getting the placeholders in strings wrong), BUT:
- I assume we don't want to merge this PR, but push branch to transifex
- Ignoring lang/core/ for now, due to some parsing issue with `%` and it getting close to my bed time (I'll fix tomorrow), that will add < 10 more files with corrections 
- What's the deal with the end of file character?

Please have a scroll of the diff. (Ping @zonkmachine, @JeromeChignoli

PR #6103, opened 2025-03-30
Assassination - article quick fix

zonkmachine:
In the Assassination module the answer to chat question "Where do I find {target}" is always the same and the ending looks a bit off as the string doesn't cater for the ship model, a/an.

`... The ship is Wave and has registration ID QA-7856.`

The quick solution is to assume article 'a' as by chance it fits in with all ship models except AC-33. Now the string is off only one out of 27 times.

### DIFF
--- a/data/lang/module-assassination/en.json
+++ b/data/lang/module-assassination/en.json
@@ -349,6 +349,6 @@
   },
   "X_WILL_BE_LEAVING": {
     "description": "",
-    "message": "{target} will be leaving {spaceport} in the {system} system ({sectorX}, {sectorY}, {sectorZ}), distance {dist} ly, at {date}. The ship is {shipname} and has registration id {shipregid}."
+    "message": "{target} will be leaving {spaceport} in the {system} system ({sectorX}, {sectorY}, {sectorZ}), distance {dist} ly, at {date}. The ship is a {shipname} and has registration id {shipregid}."
   }
 }


impaktor:
Would "of model" fix it?

"of model AC..."
"of model Wave"

or "of make/", "type"?

bszlrd:
Out of scope, but how about having a "a" or "an" defined in the ship.jsons, so missions could see which to use? Or I'm overthinking again?

zonkmachine:
> Would "of model" fix it?
> 
> "of model AC..." "of model Wave"
> 
> or "of make/", "type"?

You would make it grammatically correct but it's not how people speak. It would still be annoying.


> Out of scope, but how about having a "a" or "an" defined in the ship.jsons, so missions could see which to use? Or I'm overthinking again?

I think you would need to make the ships translatable somehow but I don't know if that would work in a json file. Maybe have ship terms defined in a lua file, make, synonyms, etc. and that could be easily translatable.

zonkmachine:
OK, I'll try a simple "ships" json file.

zonkmachine:
This now works but only with new games so it wouldn't be for the next release. I think maybe, especially if the AC-33 isn't even ever selected for assassination targets, that the first one-liner still has the merits of just bringing good and you can add it to the next release. The new fuller version is better in the long run and can be applied to other modules as well.

zonkmachine:
**"a Pumpkinseed Police"**

I guess this wouldn't look good in the game.

**The ship is a Pumpkinseed Police and has registration id {shipregid}.**

Maybe

**"a Pumpkinseed police ship"** ?
**The ship is a Pumpkinseed police ship and has registration id {shipregid}.**

zonkmachine:
assassino

zonkmachine:
> I'm not convinced the best and simplest work-around is to re-word the context in which it is used, something along the lines of my initial suggestion

Unfortunately that creates a language that isn't flowing very well and that's how it's done all over the code. I've tried rephrasing stuff like that and it doesn't work as well as the real deal. On the other hand, I've poked around ways to address it and in precisely every case it runs away in complexity super fast. My wet dream right now is to just pipe the strings, at least some of them, to an external app that can fix them up on the fly with at least the most basic stuff, like a/an or genitive s or whatever is most worth it in other languages.

I think this way of doing it here, by making the ships translatable, is a reasonable way of doing it. Lot's of strings but a simple to understand idea so it should be easy on the translators. It will take some consideration though. You could add more fields for the ships so some of them could have alternative names or nicknames. The AC-33 could be either AC-33 or Dropstar or both.

I think it is also reasonable to merge the original one-line fix for the up coming stable bug fix. It's a small change but that string is on ever single assassination mission and it looks bad.

I approve of this message.

zonkmachine:
> ```json
>     "ac33": {
>         "description": "",
>         "message": "an AC-33 Dropstar"
>     },
> ```

As I understand it, as these are new translations, the strings here would be pushed to all other languages and come out with English articles. That wouldn't be good. Is there a way to do a first push to Transifex with ships.json without articles so the translations will look the same as today and then change the strings in a later commit?

zonkmachine:
> push to Transifex with ships.json without articles

Do all languages at the same time and with the same English translation without article or with a rough automatic translation? This is where we could use Grok right?

impaktor:
> Is there a way to do a first push to Transifex with ships.json without articles so the translations will look the same as today and then change the strings in a later commit?

No, not that I know of.

> Do all languages at the same time and with the same English translation without article or with a rough automatic translation? This is where we could use Grok right?

Hm? When you push they get the default English until translated. However, the cron-job that pulls the strings back to master doesn't like to run, so at the moment nothing gets back to master until I kick the machine (which we also do when releasing new version)

zonkmachine:
> When you push they get the default English until translated.

Yes, but the strings are still in the English original. If we merge this, all languages will show ups with the English article. I looked at some language files and it looks like Swedish, Danish, Italian, and French, all could use their ship file with just the current name added. You don't need to be a translator to manage that. If we pushed these file pretranslated, would Transifex accept that?

zonkmachine:
OK. Definitely not a quick fix no longer.

Added strings for Search and Rescue too. The variable names are not unison for the modules. Will fix.

impaktor:
Would like thoughts from @sturnclaw here, on if adding a translation module ship is the best solution (and if it is to be added, would the name be "ships-core" or "ships"?)
 
(Besides, this will need further hooking up into transifex upon merge, by me)

zonkmachine:
> (Besides, this will need further hooking up into transifex upon merge, by me)

It's a bit like #5852. Did you have to do anything special on merge?

impaktor:
@zonkmachine no, difference is this adds a _new_ module, that needs some configuring on transifex side.

Mc-Pain:
> Where do I find {target}
< [...] The target is using Wave with registration ID QA-7856.

zonkmachine:
> Where do I find {target}
> < [...] The target is using Wave with registration ID QA-7856.

Please be more verbose. What am I looking at here?

Mc-Pain:
> > Where do I find {target}
> > < [...] The target is using Wave with registration ID QA-7856.
> 
> Please be more verbose. What am I looking at here?

Oops. It was not supposed to be a quote.
It was a suggestion to change answer in order to resolve the issue.

zonkmachine:
> Oops. It was not supposed to be a quote.
> It was a suggestion to change answer in order to resolve the issue.

Sorry, English is not my first language. ; )
A - What is that issue?
B - What is the solution?

zonkmachine:
> < [...] The target is using Wave with registration ID QA-7856.

I think you still need an article in there for this sentence to be complete.
 **The target is using a Wave with registration ID QA-7856.**

zonkmachine:
Rebased on master and squashed. I don't know how to deal with other languages. I can add in all of them with just the base name like in the Swedish ship name file or just leave it to the translators to solve. It may be easier to add in the files via this PR or the untranslated ship names will pop up with English articles where they should not.

data/lang/ships/sv.json
mwerle, line 1: There are no prefixes to nouns in Swedish? And no translation of "police"?

data/lang/ships/en.json
```diff
@@ -0,0 +1,110 @@
+{
+    "ac33": {
+        "description": "",
```
mwerle: For this (and all others) I think it is very important to add the context of where this translation will be used and that it requires a valid article in front of  (or, I guess, behind depending on the language?) the ship name. Otherwise translators will not know how to correctly translate these fields.

data/lang/module-searchrescue/en.json
```diff
@@ -396,8 +396,8 @@
     "message": "The replacement crew has to reach the ship by {due}."
   },
   "FLAVOUR_7_INTROTEXT": {
-    "description": "",
-    "message": "My name is {name} of {entity}. Our only ship, {ship} {shiplabel}, is drifting with a crew of {crew} after {problem}. I am willing to pay {cash} for immediate help."
+    "description": "{shipsing} is translated in /data/lang/ships/",
```
mwerle: What is the difference between `shipname` and `shipsing`? The "data/lang/ships/en.json" has neither and only contains a single entry for each ship type.

data/modules/Assassination/Assassination.lua
```diff
@@ -216,7 +217,7 @@ local makeAdvert = function (station)
 		reward = reward,
 		threat = threat,
 		shipid = hullConfig.id,
-		shipname = ShipDef[hullConfig.id].name,
+		shipname = ShipDef[hullConfig.id].modelName,
```
mwerle: It might be better to rename the variable to indicate to other developers that this is no longer the actual shipname, but the modelname which must then be used as a key to the translation.

Alternatively, why not perform the translation here which means you don't have to call the translation function everywhere else in this module?

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -727,7 +728,8 @@ local onChat = function (form, ref, option)
 			entity       = ad.entity,
 			problem      = ad.problem,
 			cash         = Format.Money(ad.reward),
-			ship         = shipdef.name,
+			shipname     = shipdef.name,				-- Ship name, base form ex. 'Malabar'
```
mwerle: This name is not translated.

For example, german:
`shipname` == "Sinonatrix (Police)"
`shipsing` == "ein Sinonatrix Polizeischiff"

Even in english there will be a difference, so somewhat confusing and disjointed player experience I think.

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -2120,7 +2122,7 @@ local buildMissionDescription = function(mission)
 			l.LON.." "..decToDegMinSec(math.rad2deg(mission.long))
 	end
 
-	local shipname = ShipDef[mission.shipid].name
+	local shipname = ShipDef[mission.shipid].name -- local shipname = ls[ShipDef[mission.shipid].modelName]
```
mwerle: ... what is the meaning of this line?

mwerle (changes requested):
Thank you for your work, but after looking at this PR I think it is only the start of translating ship names.

Firstly, it only adds translations for "singular" ship names.

Secondly it only updates the Assassination and SearchRescue missions. There are many other areas in the game which display the ship name to the user. While most of the ship names are literally just names, a few contain a descriptor as well, which needs translating.

Furthermore some of the ship names are real nouns/descriptors so it might actually make sense for different languages to provide a different name. For example "Wave"(en) -> "Welle"(de).

data/lang/module-searchrescue/en.json
mwerle: NVM, figured it out. Not good self-documenting code though. I would strongly suggest to use `shipname_singular` - `shipsing` literally gave me zero knowledge of what it was supposed to be.

data/lang/ships/sv.json
zonkmachine: The Swedish translation file is part of a discussion on how to deal with the other languages in this PR. If the file above is duplicated for all languages the ship names will show up like they do today instead of suddenly showing up anglified.

mikehgentry:
"The ship is class Wave and has registration..."

or

"The ship, classed Wave and bearing registration..."

might work, sounds sufficiently 'naval'. No idea how it translates though.

Zireael07:
I feel like it should be "this ship is Wave class", see Wookiepedia talking about Executor class ships

zonkmachine:
@Zireael07 @mikehgentry Thanks! Also 'model' is suggested earlier. This kind of works, but in my ears it doesn't flow super natural. I'm going about the assignment a bit differently. If this works I'm pretty sure you can use the base name of the ship as you suggest together with model/class. Should be no problem. I'll get back to this one in a couple of weeks.

data/modules/SearchRescue/SearchRescue.lua
zonkmachine: Looks like commented out experimental code.

zonkmachine:
> Thank you for your work, but after looking at this PR I think it is only the start of translating ship names.
> 
> Firstly, it only adds translations for "singular" ship names.
> 
> Secondly it only updates the Assassination and SearchRescue missions. There are many other areas in the game which display the ship name to the user. While most of the ship names are literally just names, a few contain a descriptor as well, which needs translating.

Yes, it's an early experiment. I don't know if it's a good idea to accommodate 'plural' in the code but I'm sure going to poke into it. If needed it could perhaps be handled by dedicated strings where a known model is used.

> Furthermore some of the ship names are real nouns/descriptors so it might actually make sense for different languages to provide a different name. For example "Wave"(en) -> "Welle"(de).

Good point. Thanks for the feedback. I'll look into the rest of your comments in a, hopefully, short while.
The branch has been rebased on master.

zonkmachine:
Something like this. Base form, indefinite form (a,an), definite form (the).
```json
  "CORONATRIX_BASE": {
    "description": "",
    "message": "Coronatrix"
  },
  "CORONATRIX_INDEF": {
    "description": "",
    "message": "a Coronatrix"
  },
  "CORONATRIX_DEF": {
    "description": "",
    "message": "the Coronatrix"
  },

```

> "this ship is Wave class"

Would look something like this (pseudocode)
ship = l.WAVE_BASE
"this ship is {ship} class"

data/lang/module-searchrescue/en.json
zonkmachine: Now it follows basic grammar NAME_BASE, NAME_INDEF and NAME_DEF. Currently the tokens are in upper case to follow the rest of the lang code.

data/modules/SearchRescue/SearchRescue.lua
zonkmachine: No. It's a work in progress. Now more strings are translated.

zonkmachine:
```json
  "CORONATRIX_BASE": {
    "description": "",
    "message": "Coronatrix"
  },
  "CORONATRIX_INDEF": {
    "description": "",
    "message": "a Coronatrix"
  },
  "CORONATRIX_DEF": {
    "description": "",
    "message": "the Coronatrix"
  },

```

I think this would work with inheritance. You could have a function to pull in the general thing it represents or a role it fills. A ship, in this case. "A ship" / "A Coronatrix" / "A police ship".  "A Coronatrix police ship" is a bit heavy in a sentence but "A police ship" isn't. Better flow. So SHIP_BASE, SHIP_INDEF, SHIP_DEF, as well as POLICE_SHIP_BASE, etc., could go into this file as well and be handled by a function to retrieve also the words an object inherits from.

```json
  "POLICE_SHIP_BASE": {
    "description": "",
    "message": "police ship"
  },
  "POLICE_SHIP_INDEF": {
    "description": "",
    "message": "a police ship"
  },
  "POLICE_SHIP_DEF": {
    "description": "",
    "message": "the police ship"
  },
  "SHIP_BASE": {
    "description": "",
    "message": "ship"
  },
  "SHIP_INDEF": {
    "description": "",
    "message": "a ship"
  },
  "SHIP_DEF": {
    "description": "",
    "message": "the ship"
  },

```

He is riding a Coronatrix.
She flies a police ship.
The ship has been seen near Aldebaran just the other week.

I think this will work well. Some languages may not work and they may just copy in the base form, or whatever works best, on all places.

zonkmachine:
Q - I changed the ship tokens to upper case to harmonize with the rest of the translation files. It has some issues as now I need to go string.upper on the ship names. You could instead change the ship names to be an upper case token? I'm inclined to go back to the original lower case and not touch the ship json files at this point. Any ideas on this? Should I go all lower case? Is there a convention for this?

sturnclaw:
> Q - I changed the ship tokens to upper case to harmonize with the rest of the translation files. It has some issues as now I need to go string.upper on the ship names. You could instead change the ship names to be an upper case token? I'm inclined to go back to the original lower case and not touch the ship json files at this point. Any ideas on this? Should I go all lower case? Is there a convention for this?

Our existing convention is that translation string identifiers are all uppercase, using underscores for word separation. This arises from a C convention in which macros and constants use the UPPER_CASE naming convention. See, for example, `LangStrings.inc.h`. The use of `string.upper` in this case is entirely reasonable; the only *better* way to handle it would be to introduce a new field to the ship.json file containing the `i18n_key` which is composed with the particular suffixes.

If you desire to take the latter route, I can make the necessary changes to the Notion database and trigger a reimport of ship.json files once i18n keys have been added to all ship entries.

> I feel like it should be "this ship is Wave class", see Wookiepedia talking about Executor class ships

For the record, in (American) English this is grammatically malformed. "\<name>[-]class" is treated as a single modifying noun. You would render the above in English as "the Wave-class ship Mark Twain", "this is a Wave-class ship", or "this ship is a Wave-class fighter". Due to its nature as a modifying noun, it cannot stand on its own. The use of "class" indicates that the ship is similar in construction to the first ship in the class, the name of which typically names the class.

You might also (depending on flavor) wish to introduce the ship as a "Wave-class hull". I do not know of a way to phrase the above sentence using "model" that does not sound stilted... "the ship is a Wave model" does not necessarily convey much of anything. You'd be better off treating it like a model of car, and rendering it something like "the target flies a heavily-modified Auronox Wave".

sturnclaw:
For the record, since I've not chimed in on this yet - I very much appreciate the initiative taken here! We've definitely needed better ways to handle string interpolation of various names, and I think the definite, indefinite, generic approach works well. You might want an additional form for plural, though I do not know if it would be helpful or necessary in most languages; e.g. in English we'd say "the targets fly three Wave-class ships" or (informal) "they're flying three Waves".

sturnclaw:
A third thing to consider: in English, modifying adjectives and nouns can be inserted between the article ('a', 'the') and the actual noun ('Coronatrix') without changing the spelling of either. Does this also keep with Swedish, etc. or would mission modules need to provide custom translated ship names if they wanted to include adjectives (i.e. "the target flies a militarized AC33")? If the former works well, it may suffice to have the definite/indefinite forms omit the article, and have the article in the language string the ship name is to be interpolated into.

zonkmachine:
> in English, modifying adjectives and nouns can be inserted between the article ('a', 'the') and the actual noun ('Coronatrix') without changing the spelling of either

Give me some examples. I thought of modifying nouns just an hour ago and my understanding is that it will exponentially increase the difficulty of translation. In English (this is my understanding), a/an is decided by the pronunciation of the following word.

A Coronatrix
An old Coronatrix

In Swedish the article depends on the noun and the adjective is changed too. Sometimes...
En Coronatrix
En gammal Coronatrix

But for the word ship:
Ett skepp
Ett gammalt skepp 

So, old in english didn't change, but in swedish it did. And this is only two languages. Spontaneously I think this will be hard. I think rather quickly we would need to use something like a dedicated translation software (Appertium or something similar) where you can add grammatical rules between two languages to shuffle the order between them and handle suffixes etc. It would probably be better to look into some dedicated software for this.
I haven't even thought much about plural.

I think this would need a bigger discussion involving more than two languages. Maybe it's easier than I think. Like how we only could have renamed AC-33 to BC-33 and then all ships in English would have been "a {ship}".

sturnclaw:
> I think this would need a bigger discussion involving more than two languages. Maybe it's easier than I think. Like how we only could have renamed AC-33 to BC-33 and then all ships in English would have been "a {ship}".

If only life were that simple...

> I thought of modifying nouns just an hour ago and my understanding is that it will exponentially increase the difficulty of translation. In English (this is my understanding), a/an is decided by the pronunciation of the following word [...]

I figured that would be the case for Swedish, and good catch with the form of the article changing in English! I generally agree, handling modifying nouns/adjectives is something that probably should ship with custom lang strings in the module that actually uses them, rather than attempting to make them part of a "generic" system.

zonkmachine:
> If only life were that simple...

If only...

What I do think would be worth looking into, is the cargo section, but this is another topic.

mwerle:
What we need here is a simple A"I" model to generate the correct sentences. This is one of the few areas in which A"I" might actually be useful! I have no background in A"I" models though so no idea how much work it'd be to train an A"I" model (models? one per language?) for this task, nor how large it would be to add to the codebase or its runtime impacts. But this would more or less be the "outsourcing to an external application" idea which was mooted a while back.

Zireael07:
@mwerle: An AI (or, more properly, an LLM) model might be overkill. I think what's presented here is doable with a Markov chain

zonkmachine:
> ... the only _better_ way to handle it would be to introduce a new field to the ship.json file containing the `i18n_key` which is composed with the particular suffixes.
> 
> If you desire to take the latter route, I can make the necessary changes to the Notion database and trigger a reimport of ship.json files once i18n keys have been added to all ship entries.

Do you mean like this?
```JSON
{
	"model": "ac33",
	"name": "AC-33 Dropstar",
	"i18n_key": "AC33",
	"cockpit": "",
```

zonkmachine:
@mwerle @Zireael07 : @Shadmar made an AI powered chatbot mod to Pioneer already but it's not out in the wild yet.
https://www.youtube.com/watch?v=wJnSdDNDIe4

data/ships/ac33.json
```diff
@@ -1,6 +1,7 @@
 {
 	"model": "ac33",
 	"name": "AC-33 Dropstar",
+	"i18n_key": "AC33",
```
zonkmachine, line 3: * @sturnclaw Is this what you had in mind?
* What do I do with the missiles. Should they also have the i18n_key tags?

zonkmachine: > @sturnclaw Is this what you had in mind?

Looks like I need a common component like SHIP.

```json
"i18n_key": "SHIP_AC33",
```

zonkmachine: ~~Fixed. I've looked at how it's done with the hardpoints and took it from there. Just the keys are added for now.~~ _Reverted_

zonkmachine:
> the only _better_ way to handle it would be to introduce a new field to the ship.json file containing the `i18n_key` which is composed with the particular suffixes.

What do you mean with _'particular suffixes'_?  So it's `SHIP_ + suffix? SHIP_WAVE,.. etc.?
`

> If you desire to take the latter route, I can make the necessary changes to the Notion database and trigger a reimport of ship.json files once i18n keys have been added to all ship entries.

Well. I've added some kind of key to the ships.

zonkmachine:
@sturnclaw Sorry about the noise. I managed to get the i18n_key to work. I did it from the c++ side though. Is this correct?

The code, at least the transifex stuff, needs commenting.

zonkmachine:
> ```json
> "an AC-33 Dropstar"
> ```

It is an AC-33 Dropstar. But you would probably not address a vehicle with a line like this. You would say "it's and AC-33" or "it's a Dropstar". There could be optional pseudonyms to fix this. Not in this PR though.

sturnclaw:
> > the only _better_ way to handle it would be to introduce a new field to the ship.json file containing the `i18n_key` which is composed with the particular suffixes.
> 
> What do you mean with _'particular suffixes'_? So it's `SHIP_ + suffix? SHIP_WAVE,.. etc.? `

The I18N key should be composed with the definite/indefinite identifiers to form the actual key looked up. So AC33 + _DEF, AC33 + _INDEF, etc.

> > If you desire to take the latter route, I can make the necessary changes to the Notion database and trigger a reimport of ship.json files once i18n keys have been added to all ship entries.
> 
> Well. I've added some kind of key to the ships.

Rather than manually adding a key to each shipdef JSON, we should modify the Notion database that is the canonical source of the shipdefs. We sync the .json files with that database through a semi-automatic process using GitHub Actions (which opens a PR with the imported changes).

I'll have to amend the export data format in Notion to export the new table field, but after that it will be just a matter of modifying the data there.

zonkmachine:
> I'll have to amend the export data format in Notion to export the new table field, but after that it will be just a matter of modifying the data there.

OK. Standing by for now. :saluting_face:

data/ships/ac33.json
sturnclaw: AFAIK missiles do not need I18N data. Missiles in general "should not" be defined as actual ship hulls; they should have a completely different (much *much* lighter-weight) definition and propulsion system. It's on the eternal WIP pile though.

zonkmachine: oki

zonkmachine:
OK, We're rebased and squashed to two commits for easier review. It now must be tested together with #6241

data/meta/ShipDef.lua
```diff
@@ -11,6 +11,7 @@
 ---@field id string
 ---@field path string
 ---@field name string
+---@field i18n_key string
```
zonkmachine, line 14: Does this belong to notion too?
_Edit: it doesn't look like it..._

zonkmachine:
Added translatable ship names in the Views.
* Person Info - Ship Information (model spinner)
* Ship Market - Ship list and model spinner
* Equipment Market (model spinner)

zonkmachine:
I prematurely removed the ship i18n tags ahead of #6241. Added back for the time being or the PR will break when testing. _Fixed: #6241 is merged, temporary hack removed, and PR rebased._

zonkmachine:
> Secondly it only updates the Assassination and SearchRescue missions. There are many other areas in the game which display the ship name to the user. While most of the ship names are literally just names, a few contain a descriptor as well, which needs translating.

I've added translated ship names to the comms/info views now. What other places needs fixing?

> Firstly, it only adds translations for "singular" ship names.

I think I will only do 'singular' in this round. If it's needed we can add this later. Are ship models used in plural anywhere in the game apart from maybe the lore?

@sturnclaw I kept the esoteric stuff separate (https://github.com/pioneerspacesim/pioneer/pull/6103/commits/77bf4de60f905cebb9d5732b33675766d07b6adc). Esoteric meaning C++. The meaty stuff, (lua) I think I have some kind of grasp on but src/lua/*stuff*, not so much. I'll let this one rest now, no more force pushes.

zonkmachine:
Original post updated and now contains a todo list.

data/modules/Assassination/Assassination.lua
```diff
@@ -490,7 +490,7 @@ local function buildMissionDescription(mission)
 		{ l.SYSTEM, ui.Format.SystemPath(mission.location) },
 		{ l.DISTANCE, dist.." "..lc.UNIT_LY },
 		false,
-		{ l.SHIP, mission.shipname },
+		{ l.SHIP, ls[ShipDef[mission.shipid].i18n_key .. "_BASE"] },
```
zonkmachine: Maybe it's enough with the bare i18n_key as **_BASE**? Just loose that suffix?

sturnclaw: Yes, that would be fine.

Additionally, due to some quirks in Lua error reporting, it's probably best to split `ShipDef[mission.shipid]` into its own statement; if `mission.shipid` is nil, it's difficult to differentiate that error from a `mission.shipid` that doesn't exist in the ShipDef table or a shipdef that doesn't have an i18n key (e.g. from a mod). Any kind of nested table access like this where the inner key is not guaranteed to be valid can get a bit dicey to diagnose.

zonkmachine: You mean like this? And similar cases? Looking into.

```diff
-    local shipname = ls[ShipDef[mission.shipid].i18n_key]
+    local shipdef = ShipDef[mission.shipid]
+    local shipname = ls[shipdef.i18n_key]
```

> if `mission.shipid` is nil

Should there be a fallback string in this case?
"Unidentified ship"

zonkmachine: > it's probably best to split `ShipDef[mission.shipid]` into its own statement

2b731d4d29e394e999e5eb0b852a4895ce6a8446

zonkmachine:
Squashed to 1 commit. There are still two todo's that I don't think I can fix. It may not be necessary for a merge however. If you change the name of a ship in translation and not just adjust it grammatically, then the ship name will show the original name in the New Game Window and in game when you select a ship and look at it's name under 'info' in the   System Map. It could be two easy fixes but I didn't manage to wrap my mind around it. Other than those two fixes this is ready for a review/play test.

zonkmachine:
Rebased on master and commit message improved

zonkmachine:
Fixed ship translations in New Game window.

zonkmachine:
The last instances I can find that needs ship names translated are these:

https://github.com/pioneerspacesim/pioneer/blob/5a7eb6ae1a6b182be7b258b994455cb606cbd0c4/data/pigui/modules/system-view-ui.lua#L428

https://github.com/pioneerspacesim/pioneer/blob/5a7eb6ae1a6b182be7b258b994455cb606cbd0c4/data/pigui/modules/system-view-ui.lua#L445

They use **GetShipType()** which handles back the 'name' and not the modelName. I'm not sure what's the best way to approach this.

zonkmachine:
> They use **GetShipType()** which handles back the 'name' and not the modelName. I'm not sure what's the best way to approach this.

~~Poked it a bit on the C++ side. Now ready for review/test.~~ _found a better way_

src/lua/LuaShipDef.cpp
```diff
@@ -21,6 +21,29 @@
...
+ * SHIP			- ship, base form.
+ * SHIP_DEF		- the ship, definitive form.
+ * SHIP_INDEF	- a ship, indefinite form.
+ *
+ * Example:
+ *
+ * > local ship = ShipDef[ad.shipid].i18n_key            -- 'NATRIX'
+ * >
+ * > local ship_def = ls[ship .. "_DEF"],                -- 'NATRIX_DEF' - 'the Natrix'
+ * > print("We're counting on " .. ship_def .. " to give us some resistance!)
```
sturnclaw: Nitpick-iest of nitpicks: missing closing quotation mark here. I only mention it because someone new to programming may try to copy-paste from the docs while writing a module.

sturnclaw (approved):
> @sturnclaw I kept the esoteric stuff separate ([77bf4de](https://github.com/pioneerspacesim/pioneer/commit/77bf4de60f905cebb9d5732b33675766d07b6adc)). Esoteric meaning C++. The meaty stuff, (lua) I think I have some kind of grasp on but src/lua/_stuff_, not so much. I'll let this one rest now, no more force pushes.

Looks good to me, adding a single field to LuaShipDef is very straightforward and doesn't need much rocket surgery.

I like what I'm seeing here, though I don't see any (current) consumers of the `SHIP_DEF` form? On the whole, this PR rectifies a long-running deficiency in the localization capability of the game, and you get two big thumbs up for improving that!

zonkmachine:
> I like what I'm seeing here, though I don't see any (current) consumers of the `SHIP_DEF` form?

I thought it was in use actually. I do remember having a case of 'the kanara' in there but it must have been dropped along the way. Yes. I think I had a problem with it because I wanted it in the beginning of a sentence but ended rewriting it because of the problem with changing case, which now also is fixed.

zonkmachine:
> I like what I'm seeing here, though I don't see any (current) consumers of the `SHIP_DEF` form?

It's not as easy to work into a text as the indefinite form. The usual use case would be more easily solved by using the more general 'the ship'. It doesn't hurt to keep and maybe it would make more sense in some other language. I almost expect there to be a ship name in there like in: "{SHIP_DEF} the Black Pearl ... " or "{SHIP_DEF} Boaty McBoatface MCXXIV".

Issue #6107, opened 2025-04-05
Search and Rescue - Issue with name generation


zonkmachine:
The issue is maybe rather in Character name generation but the Search and Rescue module shows them off well. The name is generated as one string which means you can't directly access the given name/surname after generation. This leads to methods like **splitName()** below for the cases where you want a character and use the surname for something, like in the case below, as a company name.

https://github.com/pioneerspacesim/pioneer/blob/dd7d40dfc29c15a09647798913b59fdbf2fa570e/data/modules/SearchRescue/SearchRescue.lua#L284-L290

In the case of **splitName()** above we see two problems:
1 - It doesn't handle more than one given names well. It assumes just two name components in total so the second given name is listed as surname.
2 - You don't seem to be able to access the substitute characters in the culture files as the logic breaks down when names have non ASCII signs in them.

It looks like the name should best be generated as a list where the given names and the surname/s are kept separate. Then you could have a **Character.surname** method to access the surname.

### Observed behaviour

In the examples below a family company is in trouble and the client presents themselves with full name followed by the company name. The clients surname is then reused as a component in the company name. As we see this doesn't always work.

**Examples of oddities with non ascii signs**

![Image](https://github.com/user-attachments/assets/c428090f-a0c6-416f-b6d5-53d291a1a240)

![Image](https://github.com/user-attachments/assets/ab42526a-d101-49d7-88c5-ffbed5561a2e)

![Image](https://github.com/user-attachments/assets/e0bde4df-c925-4ace-a34b-46b159d951fd)

**Example of name with more than two components**

![Image](https://github.com/user-attachments/assets/713168c5-9ab5-420e-b811-2446279bbd25)

### Steps to reproduce

Generate a lot of Search and Rescue missions and read the mission intros. There aren't that many double surnames in the game so I manually edited a culture file with only multiple surnames and/or non ASCII signs.

__My pioneer version (and OS):__
master: dd7d40dfc29c15a09647798913b59fdbf2fa570e as of April 4, 2025

impaktor:
> It looks like the name should best be generated as a list where the given names and the surname/s are kept separate

As long as they're generated together, so they match the culture file.
I imagine you could use some struct / OO approach?

zonkmachine:
I think I have a potential fix for this. Meanwhile I'll drop some observations and data here. Printing out the resulting names of splitName you quickly see how non ascii cultures ruins the day for everyone else.

**This is what two names looks like**
```
[...]

PR #6112, opened 2025-04-10
Fix name generation in Search and Rescue

zonkmachine:
### Fix for issues with Character names in the Search and Rescue module.
Addresses #6107

The failing method is only used in the Search and Rescue module and only in one of the flavours. The one that pops up on top of the BBS named "**EMERGENCY CREW TRANSPORT**".

### Issue one
The **splitName()** function assumes a name with only two components and it looks at the second one to find the surname. This doesn't work if the characters name is 'Sophie van der Linden'. It could accidentally work out to look cool but it's an uncertain.

### Issue two
The function doesn't handle non-ASCII characters well and these cause the name to be split in more parts leading to a similar kind of problem as issue 1 but often more odd looking.

### Sorting out names containing non-ASCII characters
The name is now first tested for 'suspicious' content and if it isn't only clean printable ASCII, the string will be sorted on space, " ", as delimiter and the last item returned. This has so far never failed to do a correct split but will fail 100% with multiple surnames. The result will still be a proper name and as such useful. If the algorithm doesn't find a space in the name either something is very wrong or the character is a mononym. Pioneer doesn't have mononyms as far as I know. In any case, should this happen, the full name is returned. (this may look odd but a. not as odd as the worst cases today and b. it hasn't failed so far and I've tested this a lot.

#### 1 - Function now returns a string, not a table.
**splitName()** splits the name into it's components and stores them one by one in a table which it returns on completed mission. It seems the function was intended to have more general use to allow access to first name and surname but this is hard to make in reality as names are quite complex. A proper fix would mean a more fine grained access to the names via the character. `customer.name` or `target.surname`. We can't assume that the surname is number 2 in the table so splitName now returns the surname and as a string.

#### 2 - Count hyphen and apostrophe in the word
_From here on we describe the actual split/sorting stages the name goes through_

Some names use apostrophe in the middle of it. They are also not used to split names between first name and surname so by doing this we make both those groups statistically larger and the final split is pushed more toward the middle of the name. Even if we get the final split wrong the result will be more of a proper name.

#### 3 - Return string two
If, after sorting into the list, there are only two elements, return the second one. This will obviously always be correct.

#### 4 - Use known indicators of surnames

Van Damme, de Lavall, etc. Instantly return anything after, and including, this marker. This is a small list and also make a small difference.

#### 5 - Return the last element
Any remaining names are in the final station decided to have the final element as it's surname. Slightly less than one name in a hundred reach are undecided at this point and they all seam to be split correctly as they are Latino double first names with one surname.

**Test if there are non-ASCII stuff in the name. Use space `" "` as delimiter and return the last element.**
image

**When separating the name components into the table, count hyphen and apostrophe as part of the names.**
image

**Look for common multi surname indicators, (van, de), and capitalize the resulting string.**
image


zonkmachine:
**Look for common multi surname indicators, (van, de), and capitalize the resulting string.**
image

**Test if there are non-ASCII stuff in the name. If there are any, maybe the last name isn't one of them?
Step it from the end and if there aren't any bad data before the first " ", we've got a surname.**
image

**When separating the name components into the table, count hyphen and apostrophe as part of the names.**
image

zonkmachine:
Some thoughts.

* After refactoring, **splitName()** isn't a good name for the function. **extractSurname()** is more what it does.
* I want this, or a similar utility, accessible from other scripts. I'm currently running it tucked at the end of utils.lua and it works really well. You can easily improve messages without touching the strings.

Assassination victim
image

zonkmachine:
This got overly complex. I'm working on solving this at the source in Culture, NameGen, and Character instead. Closing.

PR #6114, opened 2025-04-14
Character - Separate first name and surname

### LINKED ISSUE #6107: Search and Rescue - Issue with name generation

zonkmachine:
Generate first name and surname separately.

We need to access the names separately but they were previously generated together and saved as one in **Character.name**, with the result of having to go through the hoops of separating the names which is really hard if the name is composed of more than two names.
    
Now they are generated as two separate strings, **Character.firstname** and **Character.surname**. **Character.name** is still present. **Culture.Names()** / **NameGen.Names()** is introduced to generate a full name as two separate strings.

Example NameGen
```
name1, name2 = NameGen.Names()
```

Example Character
```
someperson = Character.New()
print("Hi " .. someperson.firstname)
```

The issue in Search and Rescue is fixed and also, the Assassination module gets a touch-up. No translated strings touched.

Supersedes #6112
Fixes #6107


data/modules/Assassination/Assassination.lua
```diff
@@ -99,8 +99,8 @@ local onChat = function (form, ref, option)
 		local sbody = ad.location:GetSystemBody()
 
 		form:SetMessage(string.interp(l.X_WILL_BE_LEAVING, {
-		  target    = ad.target,
-		  spaceport = sbody.name,
+		  target    = true and ad.targetsurname or ad.target,	-- If targetsurname is missing, then older version.
```
sturnclaw, line 102: Note that you do not need a trinary expression here - `ad.targetsurname or ad.target` would suffice.

sturnclaw (approved):
Looks good to me!

impaktor:
Is there some filter somewhere that makes all names ascii compatible? Or this is not needed?

Just thinking if someone is searching "Örjan's system" in star chart, and not endowed with umlaut keyboard. 

(I've forgotten the exact code path how this stuff works, but that's the one thought I had).

zonkmachine:
> Is there some filter somewhere that makes all names ascii compatible? Or this is not needed?

No. It looks like the ascii filter is only used in body naming. If I flip that test off and change body names to FirstName the body names will show non-ascii signs just well. Not needed unless we want **Emma V** and **Bob** to turn up. The motivation for only ascii signs in body names is given in the code as just _"-- want only ascii compatible characers in name"_, they still work though. Also, first names used to name places isn't uncommon. I can look into an ascii test for first names too.

> Just thinking if someone is searching "Örjan's system" i

Are the culture names used in system names at all? I've only seen them on planets, asteroids and settlements/stations.

data/modules/Assassination/Assassination.lua
zonkmachine: Fixed!

zonkmachine:
FYI! I haven't tested the documentation as NaturalDocs wouldn't work for me for some reason. I won't have time to look into that.

zonkmachine:
Last fix to capitalize surnames when used as company names "de Jong" becomes "De Jong". Not heavily into grammar here but that seems to be the way it's done.

data/libs/NameGen.lua
```diff
@@ -56,6 +56,79 @@ NameGen = {
...
+--
+--
+-- Availability:
+--
+--   2025
+--
+-- Status:
+--
+--   experimental
+--
```
impaktor, line 90: You don't need to remove or edit anything, but I think robn said - way back - that the "availability" and "experimental" tags are deprecated, as they don't really serve any function, I think, but I'm not willing to swear by it.

data/modules/Assassination/Assassination.lua
```diff
@@ -99,8 +99,8 @@ local onChat = function (form, ref, option)
 		local sbody = ad.location:GetSystemBody()
 
 		form:SetMessage(string.interp(l.X_WILL_BE_LEAVING, {
-		  target    = ad.target,
-		  spaceport = sbody.name,
+		  target    = ad.targetsurname or ad.target,	-- If targetsurname is missing, then older version.
+		  spaceport = sbody.name,						-- Instead hand over previously used full name + title.
```
impaktor, line 103: Is that comment indented with space or tabs? github editor gives me the impression it's tabs?

impaktor (approved):
I've not tested it, but I assume it's been self-tested/reviewed
I suggest merge after the zuper-ztable release is out.

Also, I note an asymetry in using "first name" & "surname" vs. "first and last name", in documentation (and variables?). I dunno which is best, I'm just pointing it out.

data/libs/NameGen.lua
zonkmachine: Then I'll probably remove it. It's unnecessary noise.

zonkmachine:
> I've not tested it, but I assume it's been self-tested/reviewed I suggest merge after the zuper-ztable release is out.

:+1: I've tested it extensively. Both Assassination and Search and Rescue upgrades by opening up already created missions as they were earlier and newer missions as corrected. Works fine.

> Also, I note an asymetry in using "first name" & "surname" vs. "first and last name", in documentation (and variables?). I dunno which is best, I'm just pointing it out.

`grep -R [lL]ast[nN]ame data/` 

shows that it's only **Search and Rescue** and some of the culture files that use lastname. I didn't change it from it's earlier form but it's probably a good idea to go surname at least in that file.

data/modules/Assassination/Assassination.lua
zonkmachine: Fixed!

data/libs/NameGen.lua
zonkmachine: Fixed!

zonkmachine:
https://github.com/pioneerspacesim/pioneer/labels/Savegame%20bump

I don't mind the label but the PR shouldn't break ongoing games.

sturnclaw:
Right then - I'll merge this for the upcoming hotfix. I don't think full names are used in system generation, so there should be no impact on breaking saves in that regard.

PR #6117, opened 2025-04-21
Fix pressure limit value if no atmo shield is installed

robothauler:
The ship info screen shows always 0 atm if no atmospheric shield is installed.

### DIFF
--- a/data/pigui/modules/info-view/01-ship-info.lua
+++ b/data/pigui/modules/info-view/01-ship-info.lua
@@ -41,7 +41,7 @@ local function shipStats()
 	local up_acc = player:GetAcceleration("up")
 
 	local atmo_shield = equipSet:GetInstalledOfType("hull.atmo_shield")[1]
-	local atmo_shield_cap = player["atmo_shield_cap"] or 1
+	local atmo_shield_cap = math.max(player["atmo_shield_cap"], 1)
 
 	textTable.draw({
 		{ l.REGISTRATION_NUMBER..":",	shipLabel},


data/pigui/modules/info-view/01-ship-info.lua
```diff
@@ -76,7 +76,7 @@ local function shipStats()
 		{ l.PASSENGER_CABIN_CAPACITY..":",    cabinMaximum },
 		false,
 		{ l.ATMOSPHERIC_SHIELDING..":", atmo_shield and l.YES or l.NO },
-		{ l.ATMO_PRESS_LIMIT..":",      string.format("%d atm", shipDef.atmosphericPressureLimit * atmo_shield_cap) },
+		{ l.ATMO_PRESS_LIMIT..":",      string.format("%d atm", atmo_shield_cap == 0 and shipDef.atmosphericPressureLimit or shipDef.atmosphericPressureLimit * atmo_shield_cap) },
```
sturnclaw, line 79: This can be removed if you change the first line to `math.max(player["atmo_shield_cap"], 1)`.

zonkmachine:
I've tested the PR quite thoroughly and it works. The whole atmospheric shield is a bit unclear to me though. All ships in the ship market are flagged as"Atmospheric capable: Yes" and upon purchase, in the "Ship information" they all say "Atmospheric shielding: No". Are these two not describing the same thing? Are all ships supposed to be capable on landing on planets? Including DSMiner?

fluffyfreak:
@zonkmachine Not all planets have atmospheres. 
So all ships can land on planets, but if it has an atmosphere you might need atmospheric shielding

zonkmachine:
> So all ships can land on planets, but if it has an atmosphere you might need atmospheric shielding

If I back some 500 commits on master (f0478a57c2aefa0942d86f789a2ab60b8167c908), and start the game just to check out the ship markets, the ship models Nerodia and DSMiner only shows up on the space station on the Barnard's star start. They are the only ones that has the element atmo_shield explicitly defined and they are both set to 0.

```
$ grep -R atmo_shield data/ships/
data/ships/dsminer.json:		"atmo_shield": 0,
data/ships/nerodia.json:		"atmo_shield": 0,
```

Also, on this commit, both the ship market and the ship information page has the same term "Atmospheric shielding" and it seem to mean the same thing. If all ships are potentially ""Atmospheric capable", and always answered 'yes', then that's not very interesting info on a ship prospect.

sturnclaw:
`Atmospheric Shielding` on the ship information page indicates whether an atmospheric shield equipment item is installed. I do not see any mention of `Atmosphere Capable` on the ship market. The `Atmospheric Shielding` on the ship market should indicate whether an atmospheric shielding equipment item can be installed on the ship in question.

Atmospheric shielding and atmospheric pressure tolerance does need some work, but it's a better topic for after the hotfix is released I think.

There may have been a regression where the DSMiner, etc. now have the ability to receive atmospheric shielding where they did not before, and that's a topic we can tackle during the current development cycle after this hotfix is released.

Issue #6127, opened 2025-05-05
Infinite hyperspace fuel exploit


ElonSatoshi:
### Observed behaviour
There is a loophole to not lose any fuel after a hyperspace jump

### Expected behaviour
There is not a loophole to not lose any fuel after a hyperspace jump

### Steps to reproduce
1. Start with full fuel tank, empty hyperspace tank
2. Transfer fuel to hyperspace tank
3. Start a hyperspace jump
4. During the hyperspace countdown, go back to the fuel transfer display
5. When the hyperspace jump starts, transfer all hyperspace fuel to the fuel tank
6. When you arrive to the next system, transfer from the fuel tank to the hyperspace tank
7. Repeat the process

__My pioneer version (and OS):__
20250501 flatpak
Linux

__My output.txt (required) and game save (optional, but recommended)__

PR #6129, opened 2025-05-05
Fix hyperspace fuel cheat

zonkmachine:
Fixes https://github.com/pioneerspacesim/pioneer/issues/6127, where you can pump back fuel when a hyper jump has been initiated.

### DIFF
--- a/data/pigui/modules/info-view/03-econ-trade.lua
+++ b/data/pigui/modules/info-view/03-econ-trade.lua
@@ -217,16 +217,26 @@ local function transfer_hyperfuel_mil(hyperdrive, amt)
 end
 
 local function fuelTransferButton(drive, amt)
+	local driveEnabled = false
+
+	-- Don't allow transferring fuel while the hyperdrive is doing its thing
+	if Game.player:IsHyperspaceActive() or
+			Game.player.flightState == "JUMPING" or
+			Game.player.flightState == "HYPERSPACE" then
+		driveEnabled = true
+	end
+
+	local color = driveEnabled and ui.theme.buttonColors.disabled or ui.theme.buttonColors.default
 	local icon = amt < 0 and icons.chevron_down or icons.chevron_up
 
-	if ui.button(ui.get_icon_glyph(icon) .. tostring(math.abs(amt))) then
-
-		if drive.fuel == Commodities.hydrogen then
-			transfer_hyperfuel_hydrogen(drive, amt)
-		elseif drive.fuel == Commodities.military_fuel then
-			transfer_hyperfuel_mil(drive, amt)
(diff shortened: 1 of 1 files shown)


impaktor:
So don't draw the transfer button if jumping?
Just casually wondering if it's possible to gray them out instead, and have them inactive?

zonkmachine:
> Just casually wondering if it's possible to gray them out instead, and have them inactive?

OK. I think I got it to work. More testing needed.

zonkmachine:
Tested, works. @sturnclaw touch this code last so requesting a review. Merge at will.

data/pigui/modules/info-view/03-econ-trade.lua
```diff
@@ -217,16 +217,26 @@ local function transfer_hyperfuel_mil(hyperdrive, amt)
...
 
 local function fuelTransferButton(drive, amt)
-	local icon = amt < 0 and icons.chevron_down or icons.chevron_up
+	local color = ui.theme.buttonColors.default
+	local enabled = true
 
-	if ui.button(ui.get_icon_glyph(icon) .. tostring(math.abs(amt))) then
+	-- Don't allow transferring fuel while the hyperdrive is doing its thing
+	if Game.player:IsHyperspaceActive() or Game.player.flightState == "JUMPING"
+										or Game.player.flightState == "HYPERSPACE" then
```
sturnclaw, line 225: General nitpick - we prefer to avoid aligning operands like this; instead use a single tab of indentation on a following line (or break the condition out into a separate local variable).

data/pigui/modules/info-view/03-econ-trade.lua
```diff
@@ -217,16 +217,26 @@ local function transfer_hyperfuel_mil(hyperdrive, amt)
...
+		enabled = false
+		color = ui.theme.buttonColors.disabled
```
sturnclaw, line 227: Note that instead of defining the color variable at the top, you can instead just define it following `local icon = ...` as:

```lua
local color = enabled and ui.theme.buttonColors.default or ui.theme.buttonColors.disabled
```

sturnclaw (approved):
Two minor stylistic/code quality nitpicks, otherwise this looks good to me.

data/pigui/modules/info-view/03-econ-trade.lua
zonkmachine: fixed

data/pigui/modules/info-view/03-econ-trade.lua
zonkmachine: Fixed, I believe.

zonkmachine:
OK. Fixed up issues from review and approved. Merging. No cheating...

Issue #6052, opened 2025-02-06
Atmosphere of some planets appears as transparent sphere when seen from outside


MirceaKitsune:
When approaching some planets, the atmopshere has a hard edge and appears as a transparent sphere when seen from the outside, it will appear normal once the ship enters it and it's seen from below. Attached two screenshots of the issue: First is the atmosphere seen before the ship is close, second is the correct result once the ship is inside the atmosphere.

![Image](https://github.com/user-attachments/assets/3a2c034b-dcf5-4627-bcf8-2eb8b4275347)
![Image](https://github.com/user-attachments/assets/b62e2496-9fd4-4408-946d-bce3fd64cf81)

Scattering is set to Rayleigh / MIE (accurate per-pixel), haven't checked if it also happens with Legacy. OS is Linux (Manjaro, KDE Plasma, Wayland) stock amdgpu driver, Pioneer compiled from latest Git master.

bszlrd:
![Image](https://github.com/user-attachments/assets/c578dad1-70b3-4308-a07a-87b01923d0e1)
(Sorry, couldn't resist :P)

fluffyfreak:
Not that it doesn't maybe need tweaking of course. The actual function responsible for rendering the atmospheres is `BaseSphere::DrawAtmosphereSurface`, which is both hackier and uglier than I remember it being, I must be getting old.

MirceaKitsune:
The geometry shell makes sense: Virtually the volume is rendered inside a fixed sphere. The only issue seems to be that it doesn't fade toward the outer edge when seen from outside.

Mc-Pain:
Does it look good enough?

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/e351d102-0b96-40df-80d2-9a54209f5193" />
<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/7a9458f8-1f7b-44d4-8cfb-4e8b3b3eeb80" />
<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/1da7e425-e8c6-4ea1-b8c0-266445245520" />

impaktor:
@Mc-Pain Looks amazing! Just hope it doesn't melt my old hardware

Mc-Pain:
The one from Mars:

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/3bac030b-1adb-47c5-96d2-8d2cba31d322" />

PR #6135, opened 2025-05-21
rayleigh scattering: use accurate model

Mc-Pain:
Faster approach lead to some unpleasant render artefacts, use slower but accurate approach.

This also makes water reflect rays from skylight, bringing realistic Earth colors.


data/shaders/opengl/geosphere_terrain.frag
```diff
@@ -49,21 +54,31 @@ void main(void)
 	vec4 detailMul = mix(vec4(1.0), detailVal, detailMix);
 
 #ifdef TERRAIN_WITH_WATER
-	float specularReflection=0.0;
+	vec4 waterSpecular = vec4(0.0);
 #endif
 
 #if (NUM_LIGHTS > 0)
 	vec3 V = normalize(eyeposScaled - geosphereCenter);
+	float AU = 149598000000.0;
```
fluffyfreak, line 62: This can be a constant

data/shaders/opengl/geosphere_terrain.frag
```diff
@@ -49,21 +54,31 @@ void main(void)
...
 		float uneclipsed = clamp(calcUneclipsed(eclipse, NumShadows, V, L), 0.0, 1.0);
 		CalcPlanetDiffuse(diff, uLight[i].diffuse, L, tnorm, uneclipsed);
 
 #ifdef TERRAIN_WITH_WATER
 		//water only for specular
-	    if (vertexColor.b > 0.05 && vertexColor.r < 0.05) {
-			CalcPlanetSpec(specularReflection, uLight[i], L, tnorm, eyenorm, WATER_SHINE);
+		if (vertexColor.b > 0.05 && vertexColor.r < 0.05) {
+			// Convert from radius-relative to real coordinates
+			vec3 lightPosAU = uLight[i].position.xyz / AU;
```
fluffyfreak, line 78: if AU is const, then you can have a constant called `INV_AU = 1.0f / AU`. Then change this line to something like `uLight[i].position.xyz * INV_AU;` as mul is faster than div

Mc-Pain:
> Would you mind giving some explanation of the `above tropopause` work in here just so that I can understand it.

Model uses two different formulas for calculating pressure/density
One below tropopause (where temperature decreases with height)
One above tropopause (where temperature is constant)

The one above tropopause is not implemented yet.

Mc-Pain:
> Any pictures?
https://github.com/pioneerspacesim/pioneer/pull/5617#issuecomment-2792595616

> What would "slow" vs "fast" mean here, in performance? Would this put a burden on older machines that can still run pioneer?
slow - O(n^2) complexity: 8 samples for calculating density for light ray
fast - O(n) complexity: 1 sample approximating density for light ray but it looks wrong near terminator

Mc-Pain:
This fixes https://github.com/pioneerspacesim/pioneer/issues/6052

src/editor/ModelViewer.cpp
```diff
@@ -402,7 +402,7 @@ void ModelViewer::OnModelLoaded()
 	ResetThrusters();
 
 	SceneGraph::DumpVisitor d(model);
-	model->GetRoot()->Accept(d);
```
fluffyfreak: Is there a reason this is in this review?

src/modelcompiler.cpp
```diff
@@ -2,32 +2,23 @@
 // Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
 #include "buildopts.h"
-#include "core/Log.h"
```
fluffyfreak: Did these changes get included in this review by mistake? I cannot see how they're related to the atmosphere rendering

fluffyfreak:
There seem to be some commits that don't belong in this change history now

Mc-Pain:
> There seem to be some commits that don't belong in this change history now

fixed

src/editor/ModelViewer.cpp
Mc-Pain: Oops

data/shaders/opengl/basesphere_uniforms.glsl
```diff
@@ -30,6 +30,7 @@ layout(std140) uniform BaseSphereData {
 
 	// Eclipse data
 	Eclipse eclipse;
+	vec2 tropoHeight;			// height for (R, M) in km, pressure is 0.1 atm
```
sturnclaw, line 33: `tropoHeight` does not appear to actually be used anywhere.

data/shaders/opengl/geosphere_terrain.frag
```diff
@@ -49,21 +54,30 @@ void main(void)
...
-	float specularReflection=0.0;
+	vec4 waterSpecular = vec4(0.0);
 #endif
 
 #if (NUM_LIGHTS > 0)
 	vec3 V = normalize(eyeposScaled - geosphereCenter);
 
+	vec3 I = normalize(eyeposScaled - planetIntersect);
+	vec3 center = I * geosphereRadius;
+	frag_color = vec4(0.f);
```
sturnclaw, line 65: This is an early write to frag_color that has no effect.

data/shaders/opengl/geosphere_terrain.frag
```diff
@@ -40,6 +41,10 @@ void main(void)
 	vec3 tnorm = normalize(varyingNormal);
 	vec4 diff = vec4(0.0);
 
+	vec2 atmosDist  = raySphereIntersect(geosphereCenter, eyenorm, geosphereAtmosTopRad);
+	vec2 planetDist = raySphereIntersect(geosphereCenter, eyenorm, geosphereInvRadius);
```
sturnclaw, line 45: Is there a reason you're using the inverse of the planet's view-space radius here to compute the planetary intersection? In testing, this actually does nothing, as you're attempting to compute an intersection in planet-radii space with an incredibly small sphere (smaller than 1/1000th the visible size of the planet).

data/shaders/opengl/geosphere_terrain.frag
```diff
@@ -40,6 +41,10 @@ void main(void)
...
+	vec3 planetIntersect = geosphereCenter + (eyenorm * planetDist.x);
```
sturnclaw, line 46: Under all cases, the value of planetIntersect is equal to the center of the geosphere or well behind the planet in the incredibly tiny area (no more than a fraction of a pixel) where the geosphere's center lies exactly on the camera ray.

data/shaders/opengl/geosphere_terrain.frag
```diff
@@ -49,21 +54,30 @@ void main(void)
...
 #ifdef TERRAIN_WITH_WATER
-	float specularReflection=0.0;
+	vec4 waterSpecular = vec4(0.0);
 #endif
 
 #if (NUM_LIGHTS > 0)
 	vec3 V = normalize(eyeposScaled - geosphereCenter);
 
+	vec3 I = normalize(eyeposScaled - planetIntersect);
+	vec3 center = I * geosphereRadius;
```
sturnclaw, line 64: As a consequence, I=V under all cases (and produces graphical artifacts when it is not). Thus, the entire ray-sphere intersect test can be dropped entirely, and `center` be redefined as `V * geosphereRadius`.

data/shaders/opengl/rayleigh.glsl
```diff
@@ -147,6 +151,11 @@ void processRay(inout vec3 sumR, inout vec3 sumM, inout vec2 opticalDepth, const
 	      atmosphereHeight = atmosphereRadius - geosphereRadius;
 
 	float tCurrent = boundaries.x;
+	float maxSegment = atmosphereHeight / numSamples;
+
+	if (height(tCurrent * dir, center) > 0 && numSamples * maxSegment < (boundaries.y - boundaries.x)) {
+		numSamples = min(int(ceil((boundaries.y - boundaries.x) / maxSegment)), 256); // max 256 segments
```
sturnclaw, line 157: 256 segments is quite excessive. The perceptual difference is small, and the performance cost is quite stark.

In testing with a 2070 Super (already a very hi-spec card by Pioneer's standards), just adding this branch with a maximum of 16 samples is a baseline cost of 0.2ms/frame, and with a scene set on Earth, a 64 sample maximum represents an additional cost of 2ms/frame.

At 256 samples, the additional cost of 6ms per frame(!) on high-spec hardware is decidedly well out of our performance budget, especially for the *incredibly* tiny perceptual difference from a 64 sample maximum.

These tests were performed at 1920x1080, paused, with all other details set to low so as not to introduce any other competing factors.

src/galaxy/SystemBody.cpp
```diff
@@ -217,6 +226,10 @@ double SystemBody::GetAtmAverageTemp(double altitude) const
 {
 	// temperature at height
 	const double lapseRate_L = CalcSurfaceGravity() / GetSpecificHeat(GetSuperType()); // deg/m
+	if (altitude > m_tropopause) {
+		return double(GetAverageTemp()) - lapseRate_L * m_tropopause;
+	}
```
sturnclaw, line 231: Note that this appears to set a constant atmospheric density above tropopause regardless of altitude. Is this actually physically correct?

If this is intended, this if-block isn't needed, as you can simply use `std::min(altitude, m_tropopause)` to replace the `altitude` scalar in the following line.

src/galaxy/SystemBody.cpp
```diff
@@ -241,7 +254,18 @@ void SystemBody::SetAtmFromParameters()
...
+		// get tropopause: height for pressure 0.1 atm
+		m_tropopause = (1.0 - exp(RLdivgM * log(0.1 / m_atmosPressure))) * surfaceTemperature_T0 / lapseRate_L;
+
+		// if tropopause is below surface (surface pressure < 0.1 atm)
+		if (m_tropopause < 0.0) {
+			m_tropopause = 0.0;
+		}
+
+		double tropopause_temperature = surfaceTemperature_T0 - lapseRate_L * m_tropopause;
+		m_atmosRadius = log(0.001 / 0.1) * GAS_CONSTANT_R * tropopause_temperature / (-surfaceGravity_g * GetMolarMass(GetSuperType())) + m_tropopause;
```
sturnclaw, line 268: I'd appreciate a comment here explaining the resulting value of `m_atmosRadius`. At what pressure or altitude is the radius set?

sturnclaw (changes requested):
Overall, I'm pretty happy with the perceptual quality on display here, though it does come at a potentially staggering performance cost.

In my performance testing, the baseline frametime with `#define FAST 1` and 16 samples in the outer loop is ~9ms. I believe this to be the same effect as currently in master:

Screenshot From 2025-09-09 17-45-52

I'm using that as the baseline for the following measurements with `#define FAST 0`. For measurements which say "no variable sample count", I've removed the if-block which computes `numSamples` based on the distance through the atmosphere.

`numSamples=16`, no variable sample count (+0.1ms):

Screenshot From 2025-09-09 17-36-33

`numSamples=16`, max 256 samples; this PR as currently written (+6ms):

Screenshot From 2025-09-09 17-40-06

`numSamples=8`, max 64 samples (+1ms):

Screenshot From 2025-09-09 17-36-58

Finally for an idea of the cost and appearance of the cheapest possible version of the effect, I performed a test with `numSamples=8`, no variable sample count (-0.5ms):

Screenshot From 2025-09-09 17-36-50

Given the extremely small perceptual difference between the `n=16, max=256`, and `n=8, max=64` cases, I cannot justify the performance cost of an additional 5ms on a high-end GPU. I'd like to see this PR adopt those settings before merge.

Eventually I'd prefer being able to specialize those numbers via a quality setting, for example a low-quality preset could support `n=8, max=32`, and a high-quality preset could go up to `n=16, max=128`, but that is somewhat out-of-scope for this PR.

data/shaders/opengl/geosphere_terrain.frag
```diff
@@ -49,21 +54,30 @@ void main(void)
...
 		vec3 L = normalize(uLight[i].position.xyz);
+
 		float uneclipsed = clamp(calcUneclipsed(eclipse, NumShadows, V, L), 0.0, 1.0);
 		CalcPlanetDiffuse(diff, uLight[i].diffuse, L, tnorm, uneclipsed);
 
 #ifdef TERRAIN_WITH_WATER
 		//water only for specular
-	    if (vertexColor.b > 0.05 && vertexColor.r < 0.05) {
-			CalcPlanetSpec(specularReflection, uLight[i], L, tnorm, eyenorm, WATER_SHINE);
+		if (vertexColor.b > 0.05 && vertexColor.r < 0.05) {
```
sturnclaw, line 75: This test causes noticeable "speckling" on regular Mars terrain (i.e. not water) when looking in the direction of the sun. In testing, the `vertexColor.b > 0.05` test needs to be increased to at least `0.15` to eliminate errant pixels from being classified as water.

Mc-Pain:
@sturnclaw you just tested the scattering effects for the most common case possible: direct daylight
Please try again with twilight light on ground or at some altitude as near as possible to light/shadow border

UPD there are much more corner cases like atmosphere on Venus

Mc-Pain:
Replaced by https://github.com/pioneerspacesim/pioneer/pull/6285

PR #6139, opened 2025-05-23
Fix game slowdown after jettisoning A LOT of cargo

### LINKED ISSUE #6131: Game lags after jettison A LOT of cargo
### Observed behaviour
Gameplay is lagging after you jettison in one time all of your cargo (for example 400t)

### Expected behaviour
Feel free to dump all cargo without a lag.

### Steps to reproduce
Make sure you have a big ship (in my case, a Mola Ramsayi) buy A LOT of cargo (rubbish/radiactives), perform an hyperjump and then dump all of them from jettison mode in one time; you'll notice that the game is sooo slow and continues even if you are far from abandoned cargo.
If you hyper to a new system, game is restored at the moment of hyperjump; if you come back to the system where you dump the cargo, game still working.

Pioneer 20250501 - Windows 10

[output.txt](https://github.com/user-attachments/files/20109679/output.txt)

I have a video recording too but is 1 GB; maybe should I have to upload in my youtube account?

Thank you! Have a good work ;)

fluffyfreak:
Fixes #6131 

Spawning a LOT of cargo at once caused a massive spike in frametimes due to, in the case of the original post, spawning 400 new cargo bodies at once which all overlapped and needed simulating.
All of the cargo bodies were instantly, and permanently, in collision due to being spawned at exactly the same time, position, and velocity. They also lived forever.

Highlights are:
- Store a quantity per-cargo-body so each one can contain upto 10 of the cargo type
- Pass the transform by reference, and use the raw min/max directly when possible
- Jitter position and velocity of spawned cargo so move them away from each other
- Pass through lifetime and quantity from lua to C++
- kill them after 5 minutes
- cache each CargoBody to be spawned and spawn them over time
- cargo scooping has been tested and works!

Things not yet fully tested:
- ~cargo scooping, making sure that if you spawn a cargo body with 10 items inside, that you get 10 items back when it's collected~

Things I don't plan to change/fix:
- the cargo spawn queue is not serialised, so if you jettison, then quickly pause, save, & quit, then reload you might not spawn any of the cargo and it will be lost.


data/pigui/modules/sidebar/cargo.lua
```diff
@@ -90,8 +90,15 @@ table.insert(module.transferModes, {
...
 			local commodity = Commodities[k]
-			for i = 1, v do
-				ship:Jettison(commodity)
+			-- fill cargo bodies with N items
+			local deci = math.floor(v / 10)
+			for i = 1, deci do
+				ship:Jettison(commodity, 10, 600)
+			end
+			-- fill a cargo body with any remaining entries
+			local remaining = v - (deci * 10)
```
sturnclaw, line 99:
```suggestion
			local remaining = v % 10
```

src/Ship.cpp
```diff
@@ -1453,15 +1460,48 @@ void Ship::Render(Graphics::Renderer *renderer, const Camera *camera, const vect
...
+}
+
+void Ship::ProcessSpawnQueue()
+{
+	// check we're due to spawn something
+	const double latestTime = Pi::game->GetTime();
+	if (m_latestSpawnTime > latestTime)
+		m_latestSpawnTime = 0.0; // don't allow time to be in the future, reset it
+
+	if (m_latestSpawnTime + 0.2 <= latestTime) {
```
sturnclaw, line 1483: Hardcoded spawn interval - this entire method should ideally be moved to a Lua timer IMO with a single API `bool Ship::SpawnCargo(...)` which returns false if the area to stick the cargo in isn't clear.

src/collider/Geom.cpp
```diff
@@ -90,17 +90,17 @@ std::vector<CollisionContact> Geom::Collide(Geom *b) const
 	return contacts;
 }
 
-static AABBd rotateAaabb(const AABBd &a, const matrix4x4d transA)
+static AABBd rotateAaabb(const AABBd &a, const matrix4x4d &transA)
```
sturnclaw, line 93: Little bit of an extra 'a' in 'Aaabb' here - probably my bug.

sturnclaw: Also if you're interested in implementing it, [this trick](https://zeux.io/2010/10/17/aabb-from-obb-with-component-wise-abs/) should give a bit of a speedup in constructing the AABB bounds for the oriented-bounding-box `transA * a`.

src/Ship.h
```diff
@@ -334,6 +336,9 @@ class Ship : public DynamicBody {
 
 	double m_hydrogenScoopedAccumulator = 0;
 
+	double m_latestSpawnTime = 0.0;
+	std::deque<CargoBody*> m_cargoSpawnQueue;
```
sturnclaw, line 340: Stylistic nitpick: pointer/reference are right-aligned, so there should be a space after `CargoBody`.

src/collider/Geom.cpp
fluffyfreak: I fixed the name typo, and implemented that neat little trick

src/Ship.cpp
fluffyfreak: I could not figure out what to do with the Lua, and it kinda requires the whole interface be different. You'll never be able to spawn them, there'd have be collision checking etc. Just way beyond anything I have time to do or the Lua ability right now

src/collider/Geom.cpp
```diff
@@ -90,6 +90,17 @@ std::vector<CollisionContact> Geom::Collide(Geom *b) const
 	return contacts;
 }
 
+static AABBd rotateAabbFast(const AABBd &a, const matrix4x4d &transA)
+{
+	const vector3d center((a.min + a.max) * 0.5f);
+	const vector3d extent((a.max - a.min) * 0.5f);
```
sturnclaw, line 96: Note: to prevent loss of precision, I prefer to calculate center as `(min + extent)` rather than `(min + max) * 0.5`.

src/Ship.cpp
sturnclaw: That's fine, I was mentioning it as the "ideal scenario" when all of the pieces of the puzzle are in place. This is a good enough compromise for now I think.

src/matrix4x4.h
```diff
@@ -72,6 +73,23 @@ class matrix4x4 {
 		m[8] = cell[10];
 		return m;
 	}
+
+	// used by Geom.cpp rotateAabbFast function, where we need a 3x3 orientation matrix but with abs values
+	matrix3x3<T> GetOrientAbs() const
```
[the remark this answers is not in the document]
fluffyfreak: @sturnclaw just to follow-up on this https://github.com/pioneerspacesim/pioneer/blob/master/src/matrix4x4.h#L75
It looks like it transposes for interactions with matrix3x3 🤯
I'm not gonna touch that with a barge-pole for this PR but I think it will need dealing with in the future.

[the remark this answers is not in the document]
sturnclaw: I definitely agree - please leave a reminder comment in the code so this doesn't get lost!

fluffyfreak:
When this review is complete I think everyhing can just be squashed into a single commit. There's not a lot of stuff here.

bszlrd:
> In general, I think a 30 minute lifetime (1800s) should be more than plenty for gameplay purposes, until such a time as we have a better framework for handling non-motive bodies during the update phase. I suspect performance will be dramatically improved already now that the bodies aren't self-intersecting.

Hehe, it is a tangent, but Kitten Space Agency, a wip KSP successor says in the debug menu, which bloke is in control of any given object: Kepler or Newton.
(Also, dear-imgui ;) )
image

src/CargoBody.cpp
```diff
@@ -49,7 +45,7 @@ CargoBody::CargoBody(const Json &jsonObj, Space *space) :
 		Init();
 		m_hitpoints = cargoBodyObj["hit_points"];
 		m_selfdestructTimer = cargoBodyObj["self_destruct_timer"];
-		m_hasSelfdestruct = cargoBodyObj["has_self_destruct"];
+		m_quantity = cargoBodyObj["quantity"];
```
sturnclaw: This will require a savebump. Either use `cargoBodyObj.value("quantity", 1);` with an appropriate `SAVEBUMP` comment indicating when the fallback can be removed, or add a savebump label...

src/lua/LuaShip.cpp
```diff
@@ -475,10 +475,16 @@ static int l_ship_spawn_cargo(lua_State *l)
 
 	if (lua_gettop(l) >= 3) {
 		float lifeTime = lua_tonumber(l, 3);
-		c_body = new CargoBody(model, LuaRef(l, 2), lifeTime);
+		if (lua_gettop(l) >= 4) {
+			float quantity = lua_tonumber(l, 4);
+			c_body = new CargoBody(model, LuaRef(l, 2), lifeTime, quantity);
+		} else {
+			c_body = new CargoBody(model, LuaRef(l, 2), lifeTime);
+		}
```
sturnclaw: This can be expressed as:
```suggestion
		uint32_t quantity = LuaPull<uint32_t>(l, 4, 1);
		c_body = new CargoBody(model, LuaRef(l, 2), lifeTime, quantity);
```

Similarly with the `lifeTime` parameter, you can pull with a default lifetime value and remove the branch entirely.

Regardless, `quantity` should be expressed as an integer rather than a float here. :smile:

src/matrix3x3.h
```diff
@@ -5,13 +5,16 @@
...
+#include "core/todo.h"
 #include <math.h>
 #include <stdio.h>
 #include <type_traits>
 
 template <typename T>
 class matrix3x3 {
 private:
+	// row-major ordering
+	PI_TODO("row-major is at odds with matrix4x4 which is column-major, this needs fixing")
```
sturnclaw: Rather than adding a macro, I'd strongly prefer that we use `// TODO` or `// FIXME` as it's compatible with a number of editor plugins for generating a list of TODOs from the codebase.

Also, this is going to spam the console with a message pragma for *every single file* which imports `matrix3x3.h` on every single debug compile between now and when someone gets a chance to look at it. I'd prefer we keep compilation messages limited to specifically the things which need to be fixed immediately, as otherwise it becomes a lot harder to catch useful warning messages in the sea of spam.

sturnclaw:
Apologies for the delay - had a busy few weeks and forgot this one was waiting on a review instead of waiting on more changes. :facepalm:

src/CargoBody.cpp
fluffyfreak: > when the fallback can be removed

Not sure what you mean by this 🤔 but I'll add a `// SAVEBUMP` comment?

fluffyfreak:
Thanks, good feedback! I've made most of the changes with just one question about the CargoBody SAVEBUMP comment

src/CargoBody.cpp
sturnclaw: Technically speaking the savebump comment isn't even necessary, I just wanted to be sure it was clear to future maintainers (i.e. me) that the `quantity` key is required to always be present in save version 92+. As is, just use `cargoBodyObj.value<uint32_t>("quantity", 1);`

sturnclaw:
With that, this branch should be save-compatible with v91 and no longer need that savegame bump label anymore.

sturnclaw (approved):
Pending a git history cleanup, this looks good to me. I'd prefer this get squashed down into about 3-4 commits (CargoBody API changes / Ship API changes, Geom changes, matrix documentation) before merge so I don't have to squash the whole thing into a single megacommit.

fluffyfreak:
Squashed, not very elegantly 🤕 but better than 20 commits of varying chaos and experimentation 👍

sturnclaw:
I'd prefer if we drop the `todo.h` file - it appears to have revived itself during the squash.

data/libs/Ship.lua
```diff
@@ -372,18 +372,22 @@ end
...
 --
 -- Parameters:
 --
---   item - a commodity type object (e.g., Commodity.radioactives)
+--   cargoType - a commodity type object (e.g., Commodity.radioactives)
 --          specifying the type of item to jettison.
 --
+--   quantity - how many of the item to jettison
+--
+--   lifetime - how long should the jettisoned item life before exploding
```
impaktor, line 389: This is in seconds or game ticks?

src/matrix4x4.h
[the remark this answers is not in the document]
impaktor: > please leave a reminder comment in the code so this doesn't get lost!

Has this been done?

impaktor (commented):
I've not run the code (not on hardware nor on wetware), but just left two comments in the code. 

Also, the last commit should be 'fixup'-ed into the commit that added it before merge.

Merge-person can do it (e.g. I or @sturnclaw) if it spares fluffy pain. :)

src/matrix4x4.h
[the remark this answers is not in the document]
sturnclaw: Yes - it's been documented to my satisfaction. Thanks for checking!

data/libs/Ship.lua
sturnclaw: Seconds. No duration values ever (to my knowledge) use game ticks since we have variable timestep.

sturnclaw:
With apologies for the long delay - thanks for the work @fluffyfreak!

PR #6140, opened 2025-05-24
Add long description to system view

### LINKED ISSUE #6138: Star system Long Description not seen in game
Hello. Inside the game installation folder we go to data/systems/custom and we see about 20 files named after core systems. Opening them with system editor there is quite a lot of text at the "Long description"  for each one. This text however is not shown in game, neither in sector view nor in system view, just the system name and a (very) small description.

### Observed behaviour
Text contained in /data/systems/custom doesn't  appear in game

### Expected behaviour
Text contained in /data/systems/custom should appear in game

### Steps to reproduce
Open system in system editor, check text. Run game, open sector view and choose the same system, no text description

2025-05-01, xubuntu 22.04

impaktor:
Fixes #6138 

Placement and font is WIP
output


impaktor:
I change my mind, this is better, just need to figure out how to make the height scale to the length of the text:

2025-05-24-113559_1276x707_scrot

impaktor:
OK, fixed. Also, I couldn't resist cleaning out some dead code I found in various places (separate commit).
- Dunno about tooltip, I'm re-using a string "More info"
- issue: first frame when clicking sector view the left hand search/econ panel is maximized before it scales down to the search tool's height.
- 
2025-05-24-120323_1920x1080_scrot

bszlrd:
I was going to suggest that it could go to the left panel. 
(Unrelated but the system body tree could also be shown there eventually for the sector map)

impaktor:
> (Unrelated but the system body tree could also be shown there eventually for the sector map)

Then we wouldn't have to hide the info-tab, as there would always be something to display, so that's a good suggestion. Not sure I'm motivated to add that though, as it sounds like a lot more work than my mind signed up for when I made this PR.

Would the description be on top and then body tree, or have collapsing headers?

For reference, this is what Sirius looks like:
2025-05-25-091803_1268x710_scrot

bszlrd:
I though that as another tab there

impaktor:
OK, made the info tab depend on existence of a long description. Not sure I love the code, but so it goes.
Remaining issue: 

> first frame when clicking sector view the left hand search/econ panel is maximized before it scales down to the search tool's height.

output

zonkmachine (changes requested):
Just a quick comment on a typo.

Using the space to the left works really well. The long comment texts now need a bit of a fix-up. The earth long comment should probably be removed.

sturnclaw:
My general thought so far is that adding things to the list is done with the '+' buttons on the left, and then the right only contains what's needed to manipulate the current planned route. I might rotate the search widget to the right (and put a right sidebar in there).

PR #6141, opened 2025-05-24
Migrate renderer to use explicit vertex attribute format descriptors

sturnclaw:
As is usual, this PR is the fruit of a few weeks of frustrated coding spread out over about two years.

The headlining feature here is that this PR upgrades the renderer API to deal in fully-specified vertex formats rather than the limited semantic-based format descriptor we had previously. This means that rendering with multiple vertex buffers is now fully supported, as well as fully supporting arbitrary formats for per-instance data (no more hard-coded matrix4x4f InstanceBuffers!)

As part of this work, creating a material now requires the programmer to pass in the vertex format it will be used to render. This is a major step forward to a graphics API like Vulkan, as we now have (most of) the data needed to construct a Pipeline State Object (PSO) specified at material creation time.

I've also changed the way the SceneGraph Model system handles materials; it now stores one material instance per-vertex-format rather than depending on the OpenGL driver to recompile the shader in the background when the same material is used to draw meshes with different vertex formats.

Along the way, I've made some minor fixes and improvements, as well as some not-so-minor changes:
- Loading of SGM models should be a bit faster, as we directly stream vertex data to and from disk rather than trying to rearrange each attribute into the same spot it's already in. This has the (minor) downside of further enforcing the codebase's little-endian-only requirement.
- I've removed the most expensive copy-operation involved in generation of collision meshes; we no longer try to readback from GPU memory when creating collision meshes, and instead just use the source Assimp data directly to generate collision data.

CC @fluffyfreak - you work on renderers in your "real job", so I appreciate any feedback you have regarding whether I'm shooting myself in the foot here with regards to adoption of DX12/Vulkan-class APIs in the future.

I'm in no great hurry to get this branch merged at current, but I'd prefer not to leave it open for more than a few weeks if no issues arise.

### Testing Steps:

- Compile in debug mode, set the `CHECK_VERTEX_FORMAT_MATCH` #define in `src/graphics/opengl/RendererGL.cpp` to 1.
- Launch the game, play normally; look for assertion failures. 
- Test with precompiled SGM models from the `master` branch.
- ???
- Profit!


sturnclaw:
@fluffyfreak you may be interested in checking out [this branch](https://github.com/sturnclaw/pioneer/tree/model-delod) - as a consequence of the changes here, I've been able to fix a lot of the waste in `Model::RenderInstanced()`.

I suspect the linked branch is likely to yield small but noticeable performance gains on the Raspberry Pi test branch; I'm seeing as much as a 40% decrease in CPU time taken to render cities.

fluffyfreak:
Ok that is a big review, it'll take me a few days to get through this but good work 👍

fluffyfreak:
One initial request, can we keep a fallback `#ifdef` path for the SGM loading that can be compiled in on platforms like the Raspberry Pi and other Arm machines? Just in case the little endian direct loading of SGM fails on some of them.

sturnclaw:
> One initial request, can we keep a fallback `#ifdef` path for the SGM loading that can be compiled in on platforms like the Raspberry Pi and other Arm machines? Just in case the little endian direct loading of SGM fails on some of them.

Per my comment above, we've already been constrained to little-endian only for 6 years by the `Serializer.h` implementation and how it serializes floating-point values. In that regard, I'd rather not maintain a (constantly bitrotting) trivial fallback path when significantly more complicated work would be required to make that fallback path relevant.

However, the point is entirely moot as there are no commercial big-endian architectures left in the desktop/gaming segment (as far as I know). ARM64 is entirely little-endian with an optional compatibility mode for big-endian load/store operations; PowerPC is almost nonexistent but still has a little-endian mode (PPC64LE), and... that's it. Even Apple Silicon is a derivative of ARM64 and is little-endian itself.

Vertex data format is specified independently of platform (as long as MSB/LSB order is correct), so the Pi and other ARM64 machines will be able to consume SGM data just fine without the fallback path.

fluffyfreak:
> However, the point is entirely moot as there are no commercial big-endian architectures left in the desktop/gaming segment (as far as I know).

My dream of running it on an Amiga 68k is dead! 🤣

sturnclaw:
Additionally, to my understanding since we're directly blitting the data from disk into an untyped GPU transfer buffer at byte granularity - it's not the endianness of the host that matters, but rather of the *GPU itself* as the only time the data will be read at word granularity is during vertex fetch.

Now, the GL driver could do some very arcane black magic after we upload the buffer (when we bind it to a Vertex Array Object) to reinterpret each vertex attribute component from host endianness to device endianness, but that seems... very unperformant and a violation of Occam's Razor.

fluffyfreak (approved):
Sorry for the late reply to this. It's a _lot_ of code to review.

So far I'm liking the change the vertex format description system, it really cleans up all of the buffer management and then hand written format stuff!
Ok so finished a first pass after a couple of hours. Some of it's still bubbling away in my brain to understand stuff but I like it 👍 
I've run it a bunch of times and not noticed any issues everything seems solid

Yeah I can't think of anything bad about this, it cleans up and removes a lot of stuff which I'm really glad of.

I'll have to go over it again to make sure I understand all of the changes to materials and shaders but I think tying the material and vertex format are good and a definite win.

sturnclaw:
> Sorry for the late reply to this. It's a lot of code to review.

No worries - I was *just* about to ping you and make sure this was still in-progress; I'd like to get it merged soon so I can get the important branch (the one that optimizes instanced rendering) up as a PR.

> I'll have to go over it again to make sure I understand all of the changes to materials and shaders but I think tying the material and vertex format are good and a definite win.

This is mostly informed by Vulkan and (to a lesser extent) D3D12's API - when creating a graphics pipeline on those architectures, the vertex format descriptor is a required part of the pipeline description.

We technically could have gone the way of Godot et. al. and hash vertex formats + maintain a PSO cache in the back-end for each shader to remain vertex-format independent... but that's a lot of work and very inefficient.

sturnclaw:
@fluffyfreak is there any further code review you'd like to do before this branch is merged?

Bodasey:
Savegame incompatible?

Load a game you have used until now (I have got two of them), then this happens:

Click on the "show comms" button -> The screen is black, no further menu is accessible

`Warning: Caught error in Lua UI code:
        [T] pigui/views/station-view.lua:43: bad argument #1 to 'Money' (number expected, got nil)`


Then undock your ship, a little moment later:

Screenshot_20250711_152004
Screenshot_20250711_152353

[output_2025-1-8.txt](https://github.com/user-attachments/files/21185022/output_2025-1-8.txt)

The game crashes.

No such things are observed when a new game has been started (yet).

sturnclaw:
> Savegame incompatible?

You are attempting to "time travel" your save backwards in time. This PR was branched off of master several months ago, and multiple PRs have been merged recently that modify the save format and provide backwards compatibility *only*. Saves from recent master builds are not compatible with older builds of the game.

This branch does not affect save compatibility in any way and was tested with existing saves during development.

fluffyfreak:
No I'm happy with the code as-is 👍 looking forward to the city rendering refactor too!

sturnclaw:
Merging then - thanks for the review!

PR #6143, opened 2025-05-30
Upgrade Imgui to 1.91.9b

sturnclaw:
Has it been two years already? How time flies...

This PR upgrades the ImGui version, does some maintenance, exposes the very lovely `ui.tableAngledHeadersRow()` function, and generally migrates two years worth of deprecated callsites to their more modern equivalents.

It turns out this upgrade was quite a bit more difficult than "routine maintenance", as a number of breaking changes were made in ImGui since 1.89.8. Hopefully it'll be a bit easier to keep on top of changes in the future.

I do not expect significant review here, as I've verified the code builds and runs and looks pretty much exactly the same. I'll leave the PR open for ~18 hours for @impaktor to nominate any new APIs he'd like to have exposed to Lua before I merge this branch and rebase #6140 on it.


impaktor:
I've compiled and played with it a (tiny) bit. 
Great work!

PR #6147, opened 2025-05-30
Cleaning up unused icons

bszlrd:
Removed some duplicate and unused icons, [based on this overview](https://github.com/pioneerspacesim/pioneer/issues/5319#issuecomment-2280422441).
I've kept the first skull. They weren't the same, but could be considered a duplicate.
I did not touch the theme file, please do that for me. As far as I gathered, they are all commented out already.
image


sturnclaw:
Yes - I'll handle that.

bszlrd:
I've added the reset and crowd icon, and fixed the no takeoff exclamation mark, which had the default font. Now it is an Orbiteer one, converted to path.
image

impaktor:
Actually, I have complaints. (Yes, I'm the one who just circled icons that are the same and suggested removing them in the firs t place, see link in OP). but these are different, first three are filter-icons, the second ship is for target ship. If we don't like having them be the same, one could draw cross hairs on the second ship:
icons-master-combat

Also:
- some/many of the unused icons on row 12 are unused due to being duplicates of icons/bbs/*png, I don't know if the plan was to migrate BBS to use icons.svg?
- wrench in pos 190 (middle of second to last column) is also unused.
- second to last column, fourth row, "horizontal line of planets", is unused.

sturnclaw:
> Actually, I have complaints. (Yes, I'm the one who just circled icons that are the same and suggested removing them in the firs t place, see link in OP). but these are different, first three are filter-icons, the second ship is for target ship. If we don't like having them be the same, one could draw cross hairs on the second ship:
icons-master-combat

In my opinion, it is counterproductive to have duplicate icons in the sheet regardless of our plans. If the intent is for currently-duplicate icons to be semantically separate, the duplicate icon should be removed from the SVG file and the semantic definition moved to the end of the list in `default.lua` until such a time as a separate icon is designed for the usage. Any duplicates you find that you'd like to have a different/separate icon should be treated in this way so that @bszlrd has an easy list of icons to make next time he is feeling productive :smile:

> some/many of the unused icons on row 12 are unused due to being duplicates of icons/bbs/*png, I don't know if the plan was to migrate BBS to use icons.svg?

That's still the plan. My general idea was to do the conversion around the same time as we redesign the station interface (we're already going to have to rewrite all of the mission module chat forms anyways...)

Regardless of when we do the conversion, those icons should remain on the sheet as they are planned-to-be-used, not unused. Any duplicates in that list of icons should be removed obviously, and the commented-out definition aliased to the original copy of the icon.

sturnclaw:
Thanks @bszlrd and thank you @impaktor for reminding me that we need a list of icons to de-duplicate! I've tracked that information in https://github.com/pioneerspacesim/pioneer/issues/5319#issuecomment-2234108478 for future work.

PR #6148, opened 2025-05-30
Redesign Sector Map interface, allow jumping from Sector Map

sturnclaw:
They say a picture's worth a thousand words... this has been discussed quite a bit on IRC, so I'll leave the comparison to do the talking.

Before:
Before

After:
After

This PR pretty much completely rewrites the SectorMap interface to use layout and display primitives that we've built in the years since it was initially migrated to PiGui. This finally implements @bszlrd's [mockups](https://imgur.com/a/sector-map-YuU79) for the SectorMap interface (don't ask how many years ago they were made...), meaning he is now free to redesign the station interfaces after these many years.

I don't expect much review, but would like some testers to play the game for a bit and let me know any assertions or major pain points you find. Particular focus should be on playing multiple games per session (i.e. start, jump places, save+quit, load a different save, test jumping).

I'll leave this one open for a little bit but don't expect it to languish too long in the "awaiting review category".


sturnclaw:
TODO:
- [x] Properly display player's current fuel state

PR #6150, opened 2025-06-06
Add persistent "player state" datastore, player versions of dock/undock/landing events

sturnclaw:
This PR adds a PlayerState API which is responsible for storing the player's finances and other persistent "meta" state which isn't/shouldn't be tied to a specific ship object.

The PlayerState is responsible for handling:
- A bookmark implementation (done, UI PR coming soon) allowing the player to remember systems and bodies they deem important.
- All of the player's current cash (implemented in this PR.)
- Legal state and criminal record (now implemented in this PR!)
- Loans and other financial obligations (not implemented in this PR.)
- Reputation with major factions and system authorities (not implemented in this PR.)

Serialization for the player's cash value is handled correctly and this PR is save-compatible with v91.

Additionally, this PR contains new `onPlayerDocked`/`onPlayerUndocked`/`onPlayerLanded`/`onPlayerTakeOff` events. The rationale behind this is the large quantity of event handlers which start with the preamble:
```lua
if not ship:IsPlayer() then return end
```

Due to time and personal bandwidth constraints, I've not migrated mission modules to these new events. I strongly recommend doing so, as this will reduce the performance impact of running these event handlers for every NPC ship in the system.


data/modules/Rondel/Rondel.lua
```diff
@@ -11,14 +11,15 @@ local ShipDef = require 'ShipDef'
...
+local PlayerState = require 'PlayerState'
 
 local ShipBuilder = require 'modules.MissionUtils.ShipBuilder'
 local OutfitRules = ShipBuilder.OutfitRules
 
 --local Character = require 'Character'
 
 local l_rondel = Lang.GetResource("module-rondel")
-local l_ui_core = Lang.GetResource("ui-core")
+local lui = Lang.GetResource("ui-core")
```
impaktor, line 22: Isn't this line unused?

src/Player.h
```diff
@@ -20,7 +20,9 @@ class Player : public Ship {
 	Player(const Json &jsonObj, Space *space);
 	Player(const ShipType::Id &shipId);
 
-	virtual void SetDockedWith(SpaceStation *, int port) override;
+	void OnDocked(SpaceStation *, int) override;
+	void OnUndocked(SpaceStation *, int) override;
+
```
impaktor, line 25: Curious on why `SetDockedWith` is split up?

impaktor (approved):
just did a scroll through, not play-tested it.

data/modules/Rondel/Rondel.lua
sturnclaw: Yes. I could have removed it, but chose to leave it there (in canonical format) for future improvements to the module.

In this particular case, the rationale is that it makes it easier for new/junior developers to follow our format/conventions when they're referencing Rondel as an example gameplay module. I'm not too concerned about the performance cost.

src/Player.h
sturnclaw: It's not split up; this just removes an unneeded override of the base class. `OnDocked`/`OnUndocked` are event handlers called well after `SetDockedWith` has already returned.

sturnclaw:
Going to leave this here for now; I've migrated the crime APIs to PlayerState as well (leaving Player.lua almost empty!). Testing would be much appreciated as this last batch was coded while quite tired.

zonkmachine:
@sturnclaw  681684850b97e25dda53b067067b7b7fe80fc474 partially breaks https://github.com/pioneerspacesim/pioneer/commit/9f0f20639afe91bc5c415713665d5cc8acc44205
Now when you start a game in an unexplored system it will get explored but the first message `GETTING_SENSOR_DATA` (**"System data gathering and analysis in progress...")** is skipped.

_Edit: Fixed in 7a1c9944456fc240a5343e02220acd9944f7b925_

PR #6152, opened 2025-06-10
Added long description to the sector map

zackwhalen1:
Created a new window for the long description to be seen on the middle left-side of the sector map. The window needs to have a maximum height implemented to prevent the text from covering the date information in the lower right side of the screen. 

pioneer_sector_map

### DIFF
--- a/CppProperties.json
+++ b/CppProperties.json
@@ -1,0 +1,15 @@
+{
+  "configurations": [
+    {
+      "inheritEnvironments": [
+        "mingw_64"
+      ],
+      "name": "Mingw64",
+      "includePath": [
+        "${env.INCLUDE}",
+        "${workspaceRoot}\\**"
+      ],
+      "intelliSenseMode": "linux-gcc-x64"
+    }
+  ]
+}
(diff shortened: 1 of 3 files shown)


impaktor:
Are you not up-to sync on latest master? That entire interface has been re-made, #6148

sturnclaw:
Hi - I appreciate your initiative in contributing to Pioneer! As @impaktor mentioned, in an unfortunate coincidence the long description was already added in a PR merged just a few days ago.

On the bright side, I can still give you a few pointers regarding the implementation you've proposed here, and I hope that you'll consider joining the #pioneer IRC channel and contributing further. :smile:

- We try to ensure that all strings displayed to the user are translated. This involves adding a new entry to the appropriate `data/lang/<module>/en.json` file and referencing that module via `Lang.GetResource("<module>")`.
  In the case of this PR, you'd add (if needed) a `DESCRIPTION` string to `data/lang/ui-core/en.json` and use it as `ui.text(lui.DESCRIPTION)`.
- Typically we don't sign contributions in source files; that's what the git history is for. At most we'll use something like `-- TODO(sturnclaw)` in rare cases to indicate who should be pinged about a specific area of code when an improvement becomes required for other work to proceed.

impaktor:
Thanks for the effort, we appreciate it, but as pointed out above, this has already been implemented in master.

We hope you'll find other code contributions to make.

PR #6151, opened 2025-06-08
Add a pirate system

impaktor:
## Introduction
I figured we need a pirate system, so I made this branch....
...ten years ago.

I figured it might be time to actually open a PR on it now. The reason I've stalled was I'd ideally like the station model to be a broken orbital, or a drifting bulkship/large ship, or a crashed large ship, but perhaps we'll get there in another 10 years.

## What this does
This adds a custom system, with some "piraty" named bodies, as the buccaneers have named after their role models from the 17th century (and 20th century fiction).

I did some reading, first on [Pirates in fiction](https://en.wikipedia.org/wiki/Pirates_in_the_arts_and_popular_culture) which naturally lead to [TailSpin](https://en.wikipedia.org/wiki/TaleSpin) and fondly remembering [Tales of the Black Freighter](https://en.wikipedia.org/wiki/Watchmen#Tales_of_the_Black_Freighter), or just [Air](https://en.wikipedia.org/wiki/Air_pirate) and [Space](https://en.wikipedia.org/wiki/Space_pirate) pirates in general, and learned of famous opera [Seeräuber-Jenny](https://en.wikipedia.org/wiki/Pirate_Jenny).

2025-06-08-202828_1322x946_scrot

### Description
Short:
> A system infested by men of moral depravity, and the base of unscrupulous privateers, unconscionable buccaneers and home of infamous pirates

Long: 
>  The orbiting outpost of Tortuga was once known as Windward Forward Outpost, established in 2305 to act as the final stepping stone into new frontiers. However, as exploration started to shift elsewhere, the station became economically unfeasible and was eventually abandoned and faded from memories. These days Tortuga it a place of legends, and tales about space buccaneers and ships vanishing mysteriously.

### Wikipedia trivia
Also learned that Pastafarians think (anti)correlation between reduction in Piracy and increase in global temperature is causal:

> "Pirates also play a central role in the parody religion of Pastafarianism. Established in 2005,[7] Pastafarians (members of The Church of the Flying Spaghetti Monster) claim to believe that global warming is a result of the severe decrease in pirates since the 18th century, explaining the coldness associated with winter months that follow Halloween as a direct effect of the number of pirates that make their presence known in celebration."

## Actual features
This adds:
- A Custom pirate fuel club "Gas Guzzler's Gathering"
  guzzler

- System has `lawlessness=1` meaning:
  - All commodities are legal, Nerve Gas, Battle Weapons and all
    commodity

  - Police in police tab has gone home
    police-tab

  - No police ships in the station
[...]

PR #6153, opened 2025-06-10
Support "custom only" station types

sturnclaw:
These station types will not be selected by random generation and will only be used when explicitly specified by name in a custom system definition.

This is controlled with the `"custom_only"` key in the station.json file.


sturnclaw:
Considering this is part of the support work for #6151, I'll go ahead and merge this then - thanks for the quick review @fluffyfreak!

fluffyfreak:
Small review quick, big review slow 😁

PR #6157, opened 2025-06-13
Fix breakages introduced by the updated version of Dear ImGui (PR #6143)

mwerle:
~~**DRAFT / WIP - DO NOT MERGE YET**~~

I haven't triggered any more asserts (that's not to say there aren't more), but I guess we can always add fixes as and when we come across them. For now, merging this into `main` will at least let the game start in debug mode.

----

PR #6143 updated the version of Dear ImGui used by Pioneer, but failed to update all code to the new Dear ImGui specifications.

Compiling and running Pioneer with the `IM_ASSERT` macro enabled (default yes in debug builds) resulted in `assert()` failures during game startup and in various places while simply navigating the UI.

This PR aims to address this by updating the code to be compatible with the new Dear ImGui specifications.

Specifically:
* fix disabling Ctrl-Tab window switching
* fix use of the AlwaysAutoResize flag for child windows
* fix use of round corner magic numbers when creating filled rectangles


src/lua/LuaPiGui.cpp
```diff
@@ -1779,19 +1795,20 @@ static int l_pigui_add_rect_faded(lua_State *l)
...
 		vertices[0].col = color_b;
-	if (round_corners & ImDrawFlags_RoundCornersTopRight)
+	if (draw_flags & ImDrawFlags_RoundCornersTopRight)
 		vertices[1].col = color_b;
-	if (round_corners & ImDrawFlags_RoundCornersBottomRight)
+	if (draw_flags & ImDrawFlags_RoundCornersBottomRight)
 		vertices[2].col = color_b;
-	if (round_corners & ImDrawFlags_RoundCornersBottomLeft)
+	if (draw_flags & ImDrawFlags_RoundCornersBottomLeft)
 		vertices[3].col = color_b;
```
sturnclaw: This fallback path was specifically chosen to avoid the overhead of constructing a table and parsing a string for every `addRectXXX` call in the UI. I would prefer that both of these changes be made:
1. Where heavy use of corner rounding is desired, the flags value be cached at file-scope via `local flagsXYZ = ui.DrawFlags {...}` rather than an inline table be constructed during the call (spurious allocations).
2. A fallback path be retained such that passing an integer value 0..15 as the `round_corners` parameter to addRect/addRectFilled is interpreted as the legacy RoundCorners values (i.e. left-shift by 4 if no bits are set above bit 4).

Both can be achieved while still taking an `ImDrawFlags` value, as the default implementation for pulling an ImGui flag from Lua passes a raw integer through unmodified (as that is the fast path for pre-computing the flags value via `ui.XXXFlags {...}`).

mwerle:
> Thanks for the initiative on this! I have a commit which also fixes the shortcut handling asserts in the editor (didn't find that one yet, eh?) if you want me to push that to this branch as well.
> 
Feel free to push to this PR. And no, I had not found that yet.

> I have one concern here regarding the `addRect`/`addRectFilled` corner flags parameter that I've elucidated in more detail below; the bit-shift path present in master was indeed an intentional choice and not just laziness.

I see. I understand that this code can get called in more performance-critical game loops so I'll see about improving the solution. For me it was primarily about fixing the asserts which prevented me from running the game in debug mode. Not that that really helps much since I still haven't figured out an effective way of debugging the LUA yet and most issues seem to be in the LUA these days..

src/lua/LuaPiGui.cpp
mwerle: Understood. But I'll create a set of definitions for the legacy values. Having magic-values sprinkled throughout the code is, IMHO, pretty poor form. There wasn't even a comment anywhere to tell me what they were supposed to mean and the Dear ImGui API docs leave quite a bit to be desired as well..

mwerle: Changed in 19551115b45bf39c4c9c0a874fca06fb9e5ef8e1 (to be squashed)

I've completely reverted the use of the flags table and added the magic conversion to the other 2 functions which need it in "LuaPiGui.cpp". I did replace all the magic numbers with symbols though.

sturnclaw:
Thanks - I'll push that commit to this branch soon; currently have the guts of an economy rewrite clogging my working tree (and working memory).

mwerle:
> Looks good, passes my testing in debug mode - thanks! I'll merge once squashed.

Appears to have already been squashed, but I reworded the commit tagged "fixup" and rebased on top of current master.

PR #6159, opened 2025-06-14
Fix: SecondHand equipment market : add clearer error messages (#6130)

### LINKED ISSUE #6130: SecondHand module needs clearer error messages
The SecondHand module has never been clear at the best of times, but it should have separate, better quality error messages to more clearly inform the player of these four error conditions:

1. The ship has no slots (filled or empty) that can support the item on sale;
2. No empty slots for the item are currently available;
3. Not enough equipment capacity is available to install the item;
4. The player doesn't have enough money to buy the item.

Bonus points for informing the player of this information in the description *before* they click the buy button!

mwerle:
Fixes #6130 

Introduce additional error messages which detail why the equipment can not be installed in the ship. The second-hand equipment market now distinguishes between the ship not having a compatible equipment slot, not having a free slot, and not having enough free space.
    
It now also disables ads which the player cannot purchase as well as giving the reason when the player clicks on a disabled ad. Previously the player had to interact with the seller and click on buy before the error was shown.

### **Reviewer Notes**

The first commit fixes the spacing in the `SecondHand.lua` file, the actual changes are in the second commit.

### **Testing**

Use the attached save file (not a zip, just renamed to enable uploading).

[6130 - BBS Secondhand Equipment Market test - Big Ship.zip](https://github.com/user-attachments/files/20737750/6130.-.BBS.Secondhand.Equipment.Market.test.-.Big.Ship.zip)

The BBS has the following equipment for sale:
- Atmospheric Shielding (S4) - incompatible slot equipment
  - can be used to test the "no compatible slot" scenario
- Extra Passenger Cabin (S3) - compatible slot equipment
  - can be used to test various scenarios
- Reinforced Structure (S5) - compatible slot equipment
  - can be used to test various scenarios

The ship is a Malabar and has almost been filled with equipment. 
- There is a 20.4m³ free
- A S3 Passenger cabin slot is free
- A S5 Reinforced Structure slot is free

The player has enough money to purchase the S3 Cabin.

#### Test scenarios:

- [x] Not enough money
- [ ] Not enough space (non-slot equipment)
- [x] Not enough space (slot equipment)
- [x] No compatible slot (slot equipment)
- [x] No free slot (slot equipment)
- [x] Can buy (slot equipment)
- [ ] Can buy (non-slot equipment)

### **Out-of-scope notes**

Some equipment does not reduce the available volume. For example, Missiles. This makes sense as the missile rack presumably takes the space.

For other equipment it doesn't seem to make sense, for example, passenger cabins. If the ship already pre-reserves the space for passenger cabins, then what is actually being installed, just the interior fit-out? In which case an "empty" passenger cabin slot _could_ potentially be used for cargo instead.

Yet other equipment requires additional space even though logically there shouldn't be space in the hull for it - such as Weapons and Thrusters. Unless the idea is that the additional space is used up by supporting equipment which is not necessarily colocated within the actual hull slots for these equipment items.

### DIFF
--- a/data/lang/module-secondhand/en.json
+++ b/data/lang/module-secondhand/en.json
@@ -3,6 +3,22 @@
     "description": "Buy a piece of ship equipment",
     "message": "Buy"
   },
+  "EQUIPMENT_NO_COMPATIBLE_SLOT": {
+    "description": "Message given when the second-hand equipment is not compatible with the ship slots.",
+    "message": "This equipment item is not compatible with your ship."
+  },
+  "EQUIPMENT_NO_FREE_SLOT": {
+    "description": "Message shown by second-hand market if the ship does not have a free slot for the equipment.",
+    "message": "Your ship does not have a free slot for this equipment item."
+  },
+  "EQUIPMENT_NO_SPACE": {
+    "description": "Message shown by second-hand market if the equipment requires a free slot or free capacity to fit",
+    "message": "Your ship does not have enough free space for this equipment item."
+  },
+  "EQUIPMENT_SLOT_NO_SPACE":{
+    "description": "Message given when the ship has a compatible slot, but not enough space left to fit the equipment item.",
+    "message": "Your ship has a compatible slot, but not enough space to install this equipment item."
+  },
   "FITTING_IS_INCLUDED_IN_THE_PRICE": {
     "description": "Regarding installing a component on the ship",
     "message": "Fitting is included in the price."
@@ -98,6 +114,22 @@
   "NO_LONGER_AVAILABLE_2": {
     "description": "If the sentence is hard to translate, just make up an excuse why the item is no longer available.",
     "message": "I've changed my mind, it's not for sale anymore."
+  },
+  "NOT_ENOUGH_MONEY_0": {
+    "description": "When the player doesn't have enough money to afford the item.",
+    "message": "You don't have enough money to afford this."
+  },
+  "NOT_ENOUGH_MONEY_1": {
+    "description": "When the player doesn't have enough money to afford the item.",
+    "message": "Come back when you have enough money."
+  },
+  "NOT_ENOUGH_MONEY_2": {
+    "description": "When the player doesn't have enough money to afford the item.",
+    "message": "I'm not a charity, come back when you can afford this."
+  },
+  "NOT_ENOUGH_MONEY_3": {
+    "description": "When the player doesn't have enough money to afford the item.",
+    "message": "It may be used, but the price is not negotiable."
   },
   "REPEAT_OFFER": {
     "description": "",
--- a/data/modules/SecondHand/SecondHand.lua
+++ b/data/modules/SecondHand/SecondHand.lua
@@ -14,7 +14,6 @@ local PlayerState = require 'PlayerState'
 local utils = require 'utils'
 
 local l = Lang.GetResource("module-secondhand")
-local l2 = Lang.GetResource("ui-core")
 
 -- average number of adverts per BBS and billion population
 local N_equil = 0.1 -- [ads/(BBS*unit_population)]
@@ -22,9 +21,11 @@ local N_equil = 0.1 -- [ads/(BBS*unit_population)]
 -- inverse half life of an advert in (approximate) hours
 local inv_tau = 1.0 / (4 * 24)
 
+-- Maximum indeces for randomised messages; see data/lang/module-secondhand/en.json
 -- Note: including the 0
 local max_flavour_index = 5
 local max_surprise_index = 2
+local max_money_index = 3
 
 local flavours = {}
 for i = 0, max_flavour_index do
@@ -42,20 +43,50 @@ local onDelete = function(ref)
 	ads[ref] = nil
 end
 
--- check it fits on the ship, both the slot for that equipment and the
--- weight.
+-- Check if a piece of equipment can fit into the ship. Both the available space
+-- and, for equipemnt which requires a slot, a free and compatible slot are
+-- checked.
+--
+-- Returns a triple of:
+--
+--   ok      - whether the equipment can be installed
+--
+--   slot?   - if ok and the equipment requires a slot, the free slot into which
+--             it can be installed
+--
+--   message - a message containing the reason why the equipment can't be installed
+--
 ---@param e EquipType
+---@return boolean ok
+---@return Hullconfig.Slot? slot
+---@return string error
 local canFit = function(e)
 	local equipSet = Game.player:GetComponent("EquipSet")
+	local hasEnoughFreeVolume = equipSet:GetFreeVolume() >= e.volume
 
-	if e.slot then
-		local slot = equipSet:GetFreeSlotForEquip(e)
-		return slot ~= nil, slot, l2.SHIP_IS_FULLY_EQUIPPED
-	else
-		return equipSet:CanInstallLoose(e), nil, l2.SHIP_IS_FULLY_EQUIPPED
+	if not e.slot then
+		return hasEnoughFreeVolume, nil, l.EQUIPMENT_NO_SPACE
 	end
+
+	if not equipSet:HasCompatibleSlotForEquipment(e) then
+		return false, nil, l.EQUIPMENT_NO_COMPATIBLE_SLOT
+	end
+	if not hasEnoughFreeVolume then
+		return false, nil, l.EQUIPMENT_SLOT_NO_SPACE
+	end
+	local slot = equipSet:GetFreeSlotForEquip(e)
+	return slot ~= nil, slot, l.EQUIPMENT_NO_FREE_SLOT
 end
 
+-- Check if the player can afford the equipment item for sale
+local canAfford = function(ad)
+	return PlayerState.GetMoney() >= ad.price
+end
+
+-- Return true if the ad should be enabled in the BBS
+local isEnabled = function(ref)
+	return ads[ref] ~= nil and canAfford(ads[ref]) and canFit(ads[ref].equipment)
+end
 
 local onChat = function(form, ref, option)
 	local ad = ads[ref]
@@ -68,6 +99,19 @@ local onChat = function(form, ref, option)
 	end
 
 	form:SetFace(ad.character)
+
+	local ok, slot, message_str = canFit(ad.equipment)
+	if not ok then
+		form:SetMessage(message_str)
+		return
+	end
+
+	-- It is more performant to check the money first, but it is a better player
+	-- experience to check for equipment compatibility first.
+	if not canAfford(ad) then
+		form:SetMessage(l["NOT_ENOUGH_MONEY_" .. Engine.rand:Integer(0, max_money_index)])
+		return
+	end
 
 	if option == 0 then -- state offer
 		local adbody = string.interp(flavours[ad.flavour].adbody, {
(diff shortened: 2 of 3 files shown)


data/modules/SecondHand/SecondHand.lua
```diff
@@ -45,17 +47,25 @@ end
...
 local canFit = function(e)
 	local equipSet = Game.player:GetComponent("EquipSet")
 
-	if e.slot then
-		local slot = equipSet:GetFreeSlotForEquip(e)
-		return slot ~= nil, slot, l2.SHIP_IS_FULLY_EQUIPPED
-	else
-		return equipSet:CanInstallLoose(e), nil, l2.SHIP_IS_FULLY_EQUIPPED
+	if not e.slot then
+		return equipSet:CanInstallLoose(e), nil, l.EQUIPMENT_NO_SPACE_IN_SHIP
```
sturnclaw: This check should have an `else` branch which checks for free equipment volume for slot-based equipment (otherwise you'll get a `NO_FREE_SLOT_IN_SHIP` with a free slot but not enough spare volume).

sturnclaw:
Thank you! One minor nitpick here, but otherwise this is a very good change and I appreciate the attention to flavor with the multiple responses for being poor.

data/modules/SecondHand/SecondHand.lua
mwerle: Well picked up! I did actually pick up on this, but didn't want to add even more checks for what I figured was very much an edge-case. As-is there's quite a lot of extra logic happening now. Then again, this isn't exactly performance-critical code.

Will add the extra check..

mwerle: Fixed in 6c6e3c1ffe6571f2fe5a7a5c22601cbe073f7835  (to be squashed)

Did some minor cleanup/tweaking as well.

I was unable to test the scenarios where the equipment does not require a slot, but code inspection shows no reason why it would not work.

TODO: add a debug module to create specific BBS adverts.. Or at least the ability to re-roll the BBS.

mwerle:
> Thank you! One minor nitpick here, but otherwise this is a very good change and I appreciate the attention to flavor with the multiple responses for being poor.

Thank you. Comments addressed; please re-review at your leisure.

Bodasey:
play test passed

sturnclaw:
> For other equipment it doesn't seem to make sense, for example, passenger cabins. If the ship already pre-reserves the space for passenger cabins, then what is actually being installed, just the interior fit-out? In which case an "empty" passenger cabin slot could potentially be used for cargo instead.

Correct - it is planned that empty cabin slots must be filled with either a blank "plug", a module that adds additional crew capacity (and perhaps improves crew morale), a module that adds additional regular cargo space, or a module that adds slightly less additional "shielded" cargo space for smuggling.

Because cabin slots have somewhat esoteric constraints on size and positioning within the functional model of the ship's internals, this model was chosen over the fully-flexible "install anywhere within equipment volume" model as the volume dedicated to cabin slots is intrinsically separate from regular equipment-installable volume.

> Yet other equipment requires additional space even though logically there shouldn't be space in the hull for it - such as Weapons and Thrusters. Unless the idea is that the additional space is used up by supporting equipment which is not necessarily colocated within the actual hull slots for these equipment items.

Correct. The equipment volume for weapons, thrusters, and missiles represents additional "supporting hardware" volume needed to mount the item, rather than the space of the item itself, hence why (currently all) missiles have 0 volume requirements to mount as the missile is entirely self-contained.

data/modules/SecondHand/SecondHand.lua
```diff
@@ -44,29 +43,54 @@ local onDelete = function(ref)
 	ads[ref] = nil
 end
 
--- check it fits on the ship, both the slot for that equipment and the
--- weight.
+-- Function: canFit
```
sturnclaw: Note that this style of documentation comment is unnecessary as this is an internal function of a specific module. The NaturalDocs style is used to document public API functions and documentation using its keywords winds up in the [codedoc](https://codedoc.pioneerspacesim.net) - this function is completely inaccessible outside of this Lua module and thus may actually generate misleading documentation for this function representing it as an API function.

data/modules/SecondHand/SecondHand.lua
sturnclaw: Thanks! One minor nitpick: the established convention so far is that simple volume-related checks are performed inline rather than delegating to an opaque function on the EquipSet (i.e. `equipSet:GetFreeVolume() >= e.volume`). I find that significantly easier to read and understand than trying to remember the specific semantics of `HasVolumeForEquipment` (does the function check the ship's total volume? the current free volume?) and would prefer to follow that convention.

BBS functionality is definitely something I'd like to see improved at some point - on a related tangent, I noticed looking at the diff that we have a convention of each module reproducing effectively the same code to "hydrate" an advert with function handlers that are uniform across the entire module, and effectively the same code to mirror a copy of the station's advert storage in each module (the `ads` table).

On that note, it might be possible to split ads into "data" and "handler" portions, and register the latter once as an `AdvertType`, with the former containing the type ID of the ad in question. That might allow eliminating the per-module `ads` table entirely (with a few bits of standardized/genericized functionality) as serialization could then be implemented as a pair of transformer functions in the AdvertType called by the shared station-advert datastore.

data/modules/SecondHand/SecondHand.lua
mwerle: An automated documentation system should be able to distinguish between public and private functions and the output should be configurable to output just the public API or all APIs. At least that's how `Doxygen` works.

Apart from the rather obvious "local" declaration of the function, is there a way to tell codedoc to not publish a documentation block to the API docs?

I'm a firm believer of good documentation as it helps people (like me) who have a poor grasp of a language.

----

TL;DR: having just read some of the Natural Docs (https://www.naturaldocs.org/) documentation, I guess just removing the "Function:" line will fix this?

data/modules/SecondHand/SecondHand.lua
mwerle: > Thanks! One minor nitpick: the established convention so far is that simple volume-related checks are performed inline rather than delegating to an opaque function on the EquipSet (i.e. `equipSet:GetFreeVolume() >= e.volume`). I find that significantly easier to read and understand than trying to remember the specific semantics of `HasVolumeForEquipment` (does the function check the ship's total volume? the current free volume?) and would prefer to follow that convention.

Fixed in 6c6e3c1ffe6571f2fe5a7a5c22601cbe073f7835

data/modules/SecondHand/SecondHand.lua
mwerle: Fixed (?) in 6c6e3c1ffe6571f2fe5a7a5c22601cbe073f7835

sturnclaw: Yep - NaturalDocs is intentionally not smart enough to extract semantic meaning from the source file; it merely regexes the function signature you're documenting.

That fix should work; I was mentioning it more to say that you don't need to go through the formalism of reproducing the codedoc structure since the comment will only ever be useful to someone looking at the code directly - and thus the focus should be on clearly communicating the *intent* of the function rather than repeating obvious facts (like the parameter list) for the benefit of an audience that doesn't/can't read the implementation.

data/lang/module-secondhand/en.json
```diff
@@ -3,6 +3,22 @@
...
+    "message": "Your ship does not have a free slot for this equipment."
+  },
+  "EQUIPMENT_NO_SPACE": {
+    "description": "Message shown by second-hand market if the equipment requires a free slot or free capacity to fit",
+    "message": "Your ship does not have enough free space for this equipment item."
+  },
+  "EQUIPMENT_SLOT_NO_SPACE":{
+    "description": "Message given when the ship has a compatible slot, but not enough space left to fit the equipment item.",
+    "message": "Your ship has a compatible slot, but not enough space to install this equipment."
+  },
```
sturnclaw: These strings should be normalized to use only one of "equipment", "equipment item" or "item". Currently it's a 50/50 split between "equipment" and "equipment item" (of which I'd prefer "equipment item" or just "item").

data/modules/SecondHand/SecondHand.lua
```diff
@@ -14,116 +14,166 @@ local PlayerState = require 'PlayerState'
...
+--
+--   slot?   - if ok and the equipment requires a slot, the free slot into which
+--             it can be installed
+--
+--   message - a message containing the reason why the equipment can't be installed
+--
 ---@param e EquipType
-local canFit = function (e)
-
+---@return boolean,Hullconfig.Slot?,string
```
sturnclaw: Nitpick: prefer documenting multiple returns as:
```suggestion
---@return boolean ok
---@return Hullconfig.Slot? slot
---@return string error
```

data/modules/SecondHand/SecondHand.lua
```diff
@@ -14,116 +14,166 @@ local PlayerState = require 'PlayerState'
...
+end
 
-local onChat = function (form, ref, option)
-    local ad = ads[ref]
+-- Return true if the ad should be enabled in the BBS
+local isEnabled = function(ref)
+	if ads[ref] == nil then return false end
+	if not canAfford(ads[ref]) then return false end
+	if not canFit(ads[ref].equipment) then return false end
+	return true
```
sturnclaw: Nitpick: rather than multiple early-returns, prefer to use short-circuit evaluation for trivial conditions like this:

```suggestion
	return ads[ref] and canAfford(ads[ref]) and canFit(ads[ref].equipment)
```

sturnclaw (approved):
Looks good - I give this my thumbs-up pending a minor tweak to the language strings to normalize how they refer to the equipment item in question.

data/modules/SecondHand/SecondHand.lua
mwerle: My focus was more on good documentation for people like me who are not very good with LUA rather than people who can't access the implementation. Especially providing type information.

data/lang/module-secondhand/en.json
mwerle: Fixed; I've already squashed and force-pushed to prepare for merging into `main`.

data/modules/SecondHand/SecondHand.lua
mwerle: Fixed (and squashed)

data/modules/SecondHand/SecondHand.lua
mwerle: In general I also prefer a single return, but in this case I made it multi-return because:

- this is effectively "template" code and having it like this makes it more generic (each condition is by itself; some can be reused by other modules, others deleted/replaced)
- it makes it easy to add/remove conditions
- IMHO it makes it easier to see what each individual condition is rather than having a huge long line (easier to read multiple independant conditions top-to-bottom than left-to-right)

.. but changed to fit your preference. (squashed)

---

PS: it would be easier to have all this feedback in the original review so it can all be addressed once instead of having to iterate over this for days..

sturnclaw: I agree wholeheartedly regarding continued iteration - I apologize for the delay in getting this feedback in, but the initial batch was a 4am review and there were some things I didn't catch.

Regarding the template-like strategy - I also prefer to do it that way once the number of conditions scales past 3-4; at the current number of parameters it's not a significant hassle to refactor into a multi-return form-factor and so I prefer to use the shorter variant where possible (less keyword noise means it's a lot easier to read and understand).

sturnclaw (approved):
Looks good - many thanks for the iterative improvements!

data/modules/SecondHand/SecondHand.lua
mwerle: > I agree wholeheartedly regarding continued iteration - I apologize

No worries; sorry for mentioning it..

mwerle:
> Looks good - many thanks for the iterative improvements!

Thank you for taking the time to review and merge!

impaktor:
> it is planned that empty cabin slots must be filled with either a blank "plug", a module that adds additional crew capacity

Ping-pong table!

PR #6163, opened 2025-06-15
Replace onShip*() with onPlayer*() events where functional

robothauler:
See also #6150


sturnclaw:
**Thank you!** I'll give this a look over in the morning. While you're on the topic, do you have a hot-list of other event handlers which would see a significant improvement from having player-specific versions?

robothauler:
No hot-list, but player-specific versions for onEnterSystem, onLeaveSystem and onFrameChanged would be great, I think.

PR #6173, opened 2025-06-16
Feat/UI system view work

### LINKED ISSUE #6170: System overview and map are missing surface temperature and pressure values
### Observed behaviour
The system overview and system map do not show the surface temperature and atmospheric pressure of bodies. These values are critical for safe navigation, since high atmospheric pressures and temperatures (heat damage does not yet seem to be implemented) can destroy a ship. The atmospheric pressure is communicated by surface ports when sending a docking request, but at that point it can already be too late if the atmosphere is very dense and your ship has a weak vacuum-rated hull. Fortunately, the equally important surface gravity is already shown on the system map to prevent you from crashing into the surface during approach.

Image

### Expected behaviour
I would expect the system overview and system map to show the surface temperature and atmospheric pressure of bodies, at least for those which are not stars or gas giants with no defined surface.

In old versions of the game, the surface temperature was shown:

Image

### Steps to reproduce
1. Find a planet with an atmosphere according to its qualitative description in the system map and overview
2. Observe the lack of quantitative atmospheric pressure value and surface temperature values.

mwerle:
**DRAFT PR**

The code still needs some cleaning up, but preparing a PR for initial review/commenting on the implementation and direction.

Specifically:
- keep view-setting buttons on the bottom or move them into the left sidebar (the latter makes them more consistent with the Sector View)
- if moved into sidebar, should they be converted to checkboxes or drop-downs, again, for consistency?

----

Fixes #6170  (see issue for screenshots; before, during, and final)

After the trivial work of adding the missing values, some of the values were shown using icons instead of text labels due to the Object Info window being too narrow (constrained by the time control buttons).

So the Object Info window was converted to a Sidebar Module and added to the left sidebar, which displayed nicely, but prevented being able to click through system objects using the overview window while seeing the object information.

So the entire view was rewritten to be more in-line with the sector map:
- right-side button bar moved to bottom-center
- right-hand sidebar created
- sidebar Object View moved to right sidebar
- Orbit Planner converted to sidebar module and added to right sidebar
- a new settings sidebar module was created and the view settings buttons copied into it (Proof-of-Concept)
- some other minor tweaks, such as renaming the translate-button when in the System Overview mode

### DIFF
--- a/data/lang/core/en.json
+++ b/data/lang/core/en.json
@@ -1603,6 +1603,10 @@
     "description": "",
     "message": "Surface gravity"
   },
+  "SURFACE_PRESSURE": {
+    "description": "",
+    "message": "Surface pressure"
+  },
   "SURFACE_TEMPERATURE": {
     "description": "",
     "message": "Surface temperature"
@@ -1850,6 +1854,18 @@
   "UNIT_SQUARE_METERS": {
     "description": "Area unit: square meter (m²)",
     "message": "m²"
+  },
+  "UNIT_TEMPERATURE_CELSIUS": {
+    "description": "Temperature unit: degrees Celsius (°C)",
+    "message": "°C"
+  },
+  "UNIT_TEMPERATURE_FAHRENHEIT": {
+    "description": "Temperature unit: Fahrenheit (F)",
+    "message": "F"
+  },
+  "UNIT_TEMPERATURE_KELVIN": {
+    "description": "Temperature unit: degrees Kelvin (°K)",
+    "message": "°K"
   },
   "UNIT_TERATONNES": {
     "description": "Mass unit: one teratonne",
--- a/data/pigui/libs/text.lua
+++ b/data/pigui/libs/text.lua
@@ -286,8 +286,12 @@ ui.Format = {
 			return string.format(fmt, mass / 1e18), lc.UNIT_PETATONNES
 		elseif m < EARTH_MASS * 1e3 then
 			return oldFmt(lc.N_EARTH_MASSES, { mass = mass / EARTH_MASS }), ""
+			-- fmt = "%0." .. (digits or 3) .. "f"
+			--return string.format(fmt, mass / EARTH_MASS), "EM" -- Should be: "M🜨"
 		end
 		return oldFmt(lc.N_SOLAR_MASSES, { mass = mass / SOL_MASS }), ""
+		-- fmt = "%0." .. (digits or 3) .. "f"
+		--return string.format(fmt, mass / SOL_MASS), "SM" -- Should be : "M☉"
 	end,
 	Mass = function(mass, digits)
 		local m, u = ui.Format.MassUnit(mass, digits)
@@ -303,7 +307,7 @@ ui.Format = {
 		return string.format("%0.2f", grav) .. " " .. lc.UNIT_EARTH_GRAVITY
 	end,
 	Pressure = function(pres)
-		return string.format("%0.2f", pres) .. lc.UNIT_PRESSURE_ATMOSPHERES
+		return string.format("%0.2f ", pres) .. lc.UNIT_PRESSURE_ATMOSPHERES
 	end,
 	TemperatureCelsius = function(tempInKelvin)
 		return string.format("%0.2f", tempInKelvin - 273.15) .. lc.UNIT_TEMPERATURE_CELSIUS
--- a/data/pigui/libs/text.lua
+++ b/data/pigui/libs/text.lua
@@ -305,6 +305,19 @@ ui.Format = {
 	Pressure = function(pres)
 		return string.format("%0.2f", pres) .. lc.UNIT_PRESSURE_ATMOSPHERES
 	end,
+	TemperatureCelsius = function(tempInKelvin)
+		return string.format("%0.2f", tempInKelvin - 273.15) .. lc.UNIT_TEMPERATURE_CELSIUS
+	end,
+	TemperatureFahrenheit = function(tempInKelvin)
+		return string.format("%0.2f", (tempInKelvin - 273.15) * 9/5 + 32) .. lc.UNIT_TEMPERATURE_FAHRENHEIT
+	end,
+	TemperatureKelvin = function(tempInKelvin)
+		return string.format("%0.2f ", tempInKelvin) .. lc.UNIT_TEMPERATURE_KELVIN
+	end,
+	Temperature = function(tempInKelvin)
+		-- TODO: use user preference
+		return ui.Format.TemperatureCelsius(tempInKelvin)
+	end,
 	-- produce number..denominator format
 	NumberAbbv = function(number, places)
 		local s = number < 0.0 and "-" or ""
--- a/data/pigui/modules/system-view-ui.lua
+++ b/data/pigui/modules/system-view-ui.lua
@@ -16,6 +16,8 @@ local luc = Lang.GetResource("ui-core")
 local layout = require 'pigui.libs.window-layout'
 local Sidebar = require 'pigui.libs.sidebar'
 
+local systemEconView = require 'pigui.modules.system-econ-view'.New()
+
 local player = nil
 local colors = ui.theme.colors
 local icons = ui.theme.icons
@@ -109,6 +111,37 @@ local onEnterSystem = function (ship)
 	end
 end
 
+-- all windows in this view
+local Windows = {
+	systemName = layout.NewWindow("SystemMapSystemName"),
+	edgeButtons = layout.NewWindow("SystemMapEdgeButtons"),
+	timeButtons = layout.NewWindow("SystemMapTimeButtons"),
+	unexplored = layout.NewWindow("SystemMapUnexplored")
+}
+
+Windows.systemName.style_colors["WindowBg"] = colors.transparent
+
+local systemViewLayout = layout.New(Windows)
+systemViewLayout.mainFont = winfont
+
+local leftSidebar = Sidebar.New("##SystemMapLeft", "left")
+local rightSidebar = Sidebar.New("##SystemMapRight", "right")
+
+local systemOverviewWidget = require 'pigui.modules.system-overview-window'.New()
+systemOverviewWidget.visible = true
+
+function systemOverviewWidget:onBodySelected(sBody)
+	systemView:SetSelectedObject(Projectable.OBJECT, Projectable.SYSTEMBODY, sBody)
+end
+
+function systemOverviewWidget:onBodyDoubleClicked(sBody)
+	systemView:ViewSelectedObject()
+end
+
+--
+-- Helper functions
+--
+
 local function textIcon(icon, tooltip)
 	ui.icon(icon, Vector2(ui.getTextLineHeight()), svColor.FONT, tooltip)
 	ui.sameLine()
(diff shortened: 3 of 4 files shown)


sturnclaw:
I like the orbit planner as a sidebar widget; I think that makes a lot more sense and removes a barrier to modernizing its functionality.

The "settings" buttons are tristate and could perhaps benefit from being made combo widgets (i.e. dropdowns) in the settings tab, but I'm unsure if that would hurt discoverability and ease of access as toggling ship display specifically is a somewhat high-frequency operation.

For consistency with the sector map, the object info and system overview widgets should swap sidebar sides. That more closely matches both the WorldView's placement of the system overview widget, as well as the general pattern from the SectorMap with object info being on the left and search/selection controls on the right. Additionally, it allows using the economy display widget with the overview open as well.

mwerle:
> The "settings" buttons are tristate and could perhaps benefit from being made combo widgets (i.e. dropdowns) in the settings tab, but I'm unsure if that would hurt discoverability and ease of access as toggling ship display specifically is a somewhat high-frequency operation.
> 
That's why I left this open as a question and copied the buttons for now instead of moving them / converting them to drop-downs until I got some feedback. I just felt it was more consistent with the Sector Map view to have them in the sidebar..

I didn't realise people would toggle the ship display constantly, I thought people would choose one and leave it. Perhaps it's more likely people would toggle ship lines on/off rather than ships themselves?

Also, these settings should probably be saved; it's a little bit annoying having to enable the grid every time I open the game, for example.

> For consistency with the sector map, the object info and system overview widgets should swap sidebar sides. 

Done (db24d60d1655ea2f987d269955f18548ab4e8058):

image

mwerle:
One issue is that object icons and labels are not being dimmed behind the sidebars. The same happens in every view..

Is there a way to raise the z-order of the sidebars to be on top?

impaktor:
I don't think we write out the full name of the setting, usually. Basically because then Germans come with their impossible long words, like: _über-den-shipen-waffen-getrinken-displayung-jawhol-bitte-sehr_

456245250-3114b9dd-abb9-4cff-95ea-899eb49dec46

mwerle:
> I don't think we write out the full name of the setting, usually. Basically because then Germans come with their impossible long words, like: _über-den-shipen-waffen-getrinken-displayung-jawhol-bitte-sehr_
> 
:rofl: 

As a german, I resemble this remark! And while the sentiment is not entirely inaccurate, Pioneer does have labels for settings everywhere else in the UI. Just open the main settings dialog, for example. Or, closer to home as it were, the Settings tab in the Sector Map.

Anyway, this is currently only a placeholder to determine how / where to have these settings. For consistency with the Sector Map, they should be in a sidebar panel.

sturnclaw:
> That's why I left this open as a question and copied the buttons for now instead of moving them / converting them to drop-downs until I got some feedback. I just felt it was more consistent with the Sector Map view to have them in the sidebar..
> 
> I didn't realise people would toggle the ship display constantly, I thought people would choose one and leave it. Perhaps it's more likely people would toggle ship lines on/off rather than ships themselves?

Ship display is a common toggle (usually for mission purposes), and having it enabled it has performance implications. Full ship display with orbit lines is very inefficient right now and can cost multiple milliseconds of frame time even on high-end computers. Grids and L-points on the other hand have almost no gameplay purpose, and are rarely toggled in my experience.

Continuing discussion from IRC, my feedback/design decision is that we should retain a two-state button in the action bar (with a keyboard shortcut... eventually) which toggles display of ships (and hyperspace clouds) globally on/off.

Control of whether to render ship lines should be set with a combo or checkbox in the sidebar, defaulting to off (for performance and visual clutter reasons).

Control of whether to render arrival only, departure only, or all hyperspace clouds should be set in the sidebar as well. This ideally should be a dropdown (`ui.combo()`, uses zero-based indices) with three distinct options.

Each of these settings will only have an effect when the master toggles for ships and hyperspace clouds (respectively) are enabled.

Display of grids and lagrange points should also be combo boxes in the sidebar.

There is a premade UI style for themed combo boxes present in master; it's used in the sector map and can be accessed like so:

```lua
local comboStyle = require 'pigui.styles'.combo

comboStyle:withStyle(function()
    ui.combo(...) -- or ui.comboBox if you want to manually submit contents without an array
end)

**Comment by mwerle** [comment]
Finally got around to continuing this work.

- Rebased on current `master`
- Buttons are now toggle-buttons (on/off) for the various display items.
- Drop-downs in the settings set the specific modes
- Added hyperspace clouds
- Added hyperspace cloud details (only if hyperspace cloud analyzer is installed)

TODO:
- save/load the display settings


image

**Comment by mwerle** [comment]
... and with updated comboStyle:

image

**Comment by sturnclaw** [comment]
> Buttons are now toggle-buttons (on/off) for the various display items.

I perhaps didn't specify it in my earlier comment: I'd prefer if the L-point and Grid display visibility is *only* controlled through the sidebar, i.e. drop the master on/off buttons in the action bar and add an "off" option to their respective combo boxes. Until such a time as either have a gameplay purpose, they're purely cosmetic options and are toggled at extremely low frequency.

Otherwise, I'm really liking the look of this, especially with that combo box style!

**Comment by bszlrd** [comment]
May I suggest a button/toggle row with icons instead of dropdowns, if it is not too late? Sorry, haven't been able to follow this PR. Dropdowns are a clunky solution here in my opinion, with three or so options at most.
I can do a mockup after I get home today.

**Comment by mwerle** [comment]
> Otherwise, I'm really liking the look of this, especially with that combo box style!

Removed the buttons for Lagrange Points and the Grid, but left the Ship and Hyperspace Clouds buttons.

Code needs some refactoring/cleaning up if this is good?

image

**Comment by mwerle** [comment]
> May I suggest a button/toggle row with icons instead of dropdowns, if it is not too late? Sorry, haven't been able to follow this PR. Dropdowns are a clunky solution here in my opinion, with three or so options at most. I can do a mockup after I get home today.

Sorry.. I think the time has passed for buttons in the side-panel. Especially buttons which require even more states. The drop-downs are more flexible for multiple options, and also look more unified with other similar side-bar panels which have options. Having multi-state buttons here would (IMHO) look a bit odd.

Feel free to try to convince me (us?) otherwise.

**Comment by bszlrd** [comment]
I think it is doable without multi-state buttons:
settings-buttons
Each one is either on (blue) or off (gray), so all of them are toggles. Less cluttered and more compact this way, no need for translations (apart from tooltips). 
First row is for ships and hyperclouds, second row is for map display features.
And it could even be expanded later on if so inclinded:
image

Additionally a sidenote: if we want to differentiate between toggles and buttons on the UI later on, toggles could have rounded corners, or a border.

Alternatively it could be done with text toggles, but that looks worse and icons are self-explanatory already.
settings-buttons-labels

And in my opinion drop-downs are more clunky: one more click to do a thing, especially when it is already hidden behind a settings panel. True, drop-downs are more flexible when it comes to lots of options, but in this case there are three at most, which are already combinations of two or so options.
Admittedly the hypecloud icons need some designing, for now I just cropped the wormhole and added an arrow triangle.

**Comment by sturnclaw** [comment]
@bszlrd 
> And in my opinion drop-downs are more clunky: one more click to do a thing, especially when it is already hidden behind a settings panel. True, drop-downs are more flexible when it comes to lots of options, but in this case there are three at most, which are already combinations of two or so options.

True, dropdowns can be more clunky, but they excel at clearly communicating what the options actually do. The icons are not very clear, and lean very heavily on the use of tooltips to provide explanation to the player. For new to intermediate users, a sea of icon buttons is *very* confusing, and provides extremely low discoverability.

Adding to the additional confusion, some of the icons in that matrix are independent state (hypercloud visibility) and some are dependent state (grid legs).

For the purposes of getting this PR merged, I'd much prefer the use of the dropdowns when combined with a globally-serialized state variable (i.e. not a per-save setting) to remember your preferred setting at startup. We can definitely revisit the idea later, as I'm very much not thrilled with the current system map UI as a whole.

**Comment by mwerle** [comment]
> For the purposes of getting this PR merged, I'd much prefer the use of the dropdowns when combined with a globally-serialized state variable (i.e. not a per-save setting) to remember your preferred setting at startup. We can definitely revisit the idea later, as I'm very much not thrilled with the current system map UI as a whole.

@sturnclaw :

The Display Mode configuration is now loaded/saved from the global game configuration. It can possibly be improved, such as using a Section. I have also saved the "selected" and "displayOn" settings separately instead of just saving  the display mode string as the latter would make loading and applying the configuration much more complicated. Again, it can be done, I just chose the easier way out for now.

----

Also refactored the code. Apart from cleaning up the commit history, I think this PR is ready to be merged, pending any review comment fixes to the actual implementation.

**Comment by mwerle** [comment]
.. game settings for the System View are now saved into a separate section.

FWIW, the "GalaxyConfig()" constructor calls "Save()" on its settings many times when the main menu is entered.  This is likely not good for SSD's.. 

This was added in 61ceb71b4172f3879c7289983c7de814a472e776

**Comment by mwerle** [comment]
Fixed the unnecessary saves in GalaxyConfig - this is strictly not part of this PR so I can make it a separate PR if desired..

File: data/meta/CoreObject/Game.meta.lua [hunk 1]
```diff
@@ -104,32 +104,64 @@ function Game.GetWorldCamType() end
...
+--------------------------------------------------------------------------------
+--- Setters and Getters for game configuration options.
+---
+--- Game configuration options are applied to the entire game, not to each save.
+---
+--- Options are saved into sections and identified by unique keys within that
+--- section. The global section is identified by the empty string ("").
+---
+--- Options are only saved to disk when the Engine.SaveSettings() call is
+--- explicitly called, so there is no significant performance overhead or
```
sturnclaw:
```suggestion
--- Options are only saved to disk when Engine.SaveSettings() is
--- explicitly called, so there is no significant performance overhead or
```

data/pigui/libs/text.lua
```diff
@@ -286,8 +286,12 @@ ui.Format = {
 			return string.format(fmt, mass / 1e18), lc.UNIT_PETATONNES
 		elseif m < EARTH_MASS * 1e3 then
 			return oldFmt(lc.N_EARTH_MASSES, { mass = mass / EARTH_MASS }), ""
+			-- fmt = "%0." .. (digits or 3) .. "f"
+			--return string.format(fmt, mass / EARTH_MASS), "EM" -- Should be: "M🜨"
 		end
 		return oldFmt(lc.N_SOLAR_MASSES, { mass = mass / SOL_MASS }), ""
+		-- fmt = "%0." .. (digits or 3) .. "f"
+		--return string.format(fmt, mass / SOL_MASS), "SM" -- Should be : "M☉"
```
sturnclaw: Cleanup required.

data/pigui/modules/system-view-ui.lua
```diff
@@ -68,27 +72,180 @@ local svColor = {
...
+		Game.SetConfigInt("SystemView", self.configKeySelectedItem, self.selectedItem)
+		print("Saved " .. self.configKeySelectedItem .. ", value=" .. self.selectedItem)
+	end
+}
+
+--- A base class extending the functionality of SystemViewComboState to provide
+--- for a separate on/off toggle.
+--- The options are only to select the display state, a separate state is
+--- defined for when the display mode should be Off.
+local SystemViewComboStateOnOff = SystemViewComboState:new {
```
sturnclaw: Typically I find data-polymorphism to be of more utility than code-polymorphism. In this particular case, this class represents a significant amount of extra code and screenspace to handle the common-case of an additional implicit "off" setting.

Instead of making a separate class to override methods, I'd recommend combining the two ComboState implementations into a single one that checks for `self.displayModeOff ~= ""` (or simply use a nil default value) and executes the extra code if the state has been configured with that toggle.

src/SystemView.h
```diff
@@ -363,6 +376,7 @@ class SystemMapViewport {
 	bool m_showGravpoints;
 	ShowLagrange m_showL4L5;
 	ShipDrawing m_shipDrawing;
+	CloudDrawing m_cloudDrawing;
```
sturnclaw, line 379: Note: I'm not really happy with this being in the Viewport class, as the Viewport is intended to be the low-level display code with no (runtime or compile-time) dependencies on the game context. I left the ship drawing state in this class during the last refactor to avoid the pain of rewriting `SetVisibility` into a proper API, but I'd like the "game" state handling to be moved into `SystemView` at some point.

This is not a concern for this PR at current, but is something I'd like to have done sooner or later.

src/galaxy/GalaxyConfig.cpp
```diff
@@ -6,13 +6,14 @@
...
-	map["GalaxyExploredMin"] = "65";
-	map["GalaxyExploredMix"] = "40";
-
 	Read(FileSystem::userFiles, "galaxy.ini");
 
-	Save();
+	// set defaults
+	SetInt("GalaxyExploredMax", 90);
+	SetInt("GalaxyExploredMin", 65);
+	SetInt("GalaxyExploredMix", 40);
```
sturnclaw: Since this is after the `Read()` call, these are no longer defaults but overrides, preventing the user from tweaking galaxy generation settings. Please restore the prior direct assignment as it is significantly more efficient (removes 2x unneeded lookups for the `""` section) and instead directly set `m_unsaved` if no `galaxy.ini` file exists.

Feel free to set the unsaved flag inside of `Read()` if the specified configuration file cannot be read for any reason.

src/lua/LuaGame.cpp
```diff
@@ -738,6 +739,92 @@ static int l_game_get_parts_from_date_time(lua_State *l)
 	return 6;
 }
 
+// Helper functions for the generic l_game_get_config and l_game_set_config
+// template functions since C++ does not have a "static_if" yet.
```
sturnclaw: That's called `if constexpr`. `std::decay` is helpful here as it strips reference and cv-qualification from the variable type.

```c++
if constexpr (std::is_same_v<std::decay_t<T>, bool>) {
/// ...
} else if constexpr (...) {
// ...
}
```

Note that you cannot `static_assert(false)` in the `else` branch [until C++23](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2593r0.html) (because Reasons).

sturnclaw (changes requested):
The diff for `system-view-ui.lua` is Sufficiently Complex that I'm reserving the right to have additional review comments once I pull the branch locally and test it / read the code in non-diff format (likely in 1-3 days, brain+working tree currently busy with mission module boilerplate reduction). Otherwise, there are some minor nitpicks and one major area where I'd suggest a cleanup/refactor of the helper state object.

On the whole this looks good to me though, and I'm looking forward to getting it merged!

src/galaxy/GalaxyConfig.cpp
mwerle: This code is called during setup, so efficiency is largely immaterial here. Furthermore this all happens in memory, so I challenge you to find any significantly measurable performance impact...

----

That said, you're right, these are now overrides instead of defaults. I'll apply your suggested fix and test it. Thank you for pointing my mistake out.

sturnclaw: > This code is called during setup, so efficiency is largely immaterial here. Furthermore this all happens in memory, so I challenge you to find any significantly measurable performance impact...

Fair, there is in fact no performance impact - I was merely disputing the necessity of rewriting of code that already had the same effect into a version that added extra integer-to-string conversions and additional lookups. Other than setting the `m_unsaved` flag, I don't really see the purpose of this change?

data/pigui/modules/system-view-ui.lua
mwerle: fixed in 2c28b4525591d2be86613c2c0d9f01909c82b256

data/pigui/libs/text.lua
mwerle: fixed in 7f6b734e2b419e4bbf1d6d68f569ee79dbf197b2

src/SystemView.h
mwerle: I added it here since all the other ones were here as well.

I created an issue for keeping track of this: https://github.com/pioneerspacesim/pioneer/issues/6212

src/galaxy/GalaxyConfig.cpp
mwerle: Reworked in af5d2f37af3136fd9387596c1407c6dd161eec1c

mwerle: >  I don't really see the purpose of this change?

IMHO code should use existing functions/methods, not directly massage the internals of an implementation. It makes it MUCH more difficult for future refactoring/maintenance work.

src/lua/LuaGame.cpp
mwerle: .. is there a change request here, or just commentary? Sorry, I'm confused..

sturnclaw: This is a change request - rather than creating a bunch of function overloads, you can instead use `if constexpr` to perform compile-time branching (only the chosen branch of the if block is evaluated) inside l_game_get_config.

mwerle: Fixed in 96cc679f82379f008c854f177f2deead894483fc

Thank you for sharing the "if constexpr" approach; learned something new :)

I did quite a bit of research but could not find that approach mentioned anywhere. The closest was a "static_if" which somebody made but required a LOT of supporting magic foo to make happen.

TBH, in retrospect I actually prefer the overloaded method approach - it's cleaner and more readable. But yes, without other code making use of them it's a lot of unnecessary code.

mwerle: > .. is there a change request here, or just commentary? Sorry, I'm confused..

Sorry, in retrospect it was clear you wanted this changed; I was very tired last night after a pretty hefty day at my dayjob..

sturnclaw: > Sorry, in retrospect it was clear you wanted this changed; I was very tired last night after a pretty hefty day at my dayjob..

No worries, I'm happy to elucidate! That the comment was a "change request" wasn't clearly stated to start with, so I understand the confusion.

mwerle: Are there more changes required here or is there a reason this conversation was not resolved?

src/galaxy/GalaxyConfig.cpp
mwerle: Are more changes required or can this conversation be resolved?

mwerle:
> On the whole this looks good to me though, and I'm looking forward to getting it merged!

I'm very sorry, but is this PR still of interest?  I appreciate this is a hobby project for everybody, and that real life happens, but it's been a little while since the last update.

impaktor:
> I'm very sorry, but is this PR still of interest? I appreciate this is a hobby project for everybody, and that real life happens, but it's been a little while since the last update.

I think its very interesting, and hope to see it in master, alas, @sturnclaw has been very busy with real life, and I'm not comfortable reviewing it, (I'd rather you self-review).

As for who click's the "resolved" button in review conversation I'm not sure: is it reviewer or author?

mwerle:
> As for who click's the "resolved" button in review conversation I'm not sure: is it reviewer or author?

Personally I think the person who opened a conversation should be the one to accept that it has been resolved. Someone can provide an explanation/fix, but it is up to the originator to accept that explanation/fix.

data/pigui/modules/system-view-ui.lua
```diff
@@ -98,15 +234,58 @@ local onGameStart = function ()
 		systemView:SetColor(key, svColor[key])
 	end
 	-- update visibility states
-	systemView:SetVisibility(ship_drawing)
-	systemView:SetVisibility(show_lagrange)
-	systemView:SetVisibility(show_grid)
+	print("Loading SystemView display mode configuration")
```
sturnclaw: Leftover debug print statement.

data/pigui/modules/system-view-ui.lua
```diff
@@ -253,91 +578,169 @@ table.insert(leftSidebar.modules, {
...
+	--exclusive = true,
 
-	drawWindowControlButton(Windows.objectInfo, icons.info, lc.OBJECT_INFO)
+	drawTitle = function()
+		local obj = systemView:GetSelectedObject()
+		if obj ~= nil and obj.ref ~= nil then
+			local isSystemBody = obj.base == Projectable.SYSTEMBODY
+			local body = obj.ref
+			textIcon(getBodyIcon(obj))
+			ui.text(isSystemBody and body.name or body.label)
```
sturnclaw: The body icon is vertically off-center because legacy image-based icons do not inherit the text baseline offset used to center the sidebar panel title. The workaround to this is to draw the icon using the new glyph-based approach (which does not yet have its own convenience function).

```suggestion
			ui.text(ui.get_icon_glyph(getBodyIcon(obj)) .. " " .. (isSystemBody and body.name or body.label))
```

src/lua/LuaGame.cpp
```diff
@@ -738,6 +739,75 @@ static int l_game_get_parts_from_date_time(lua_State *l)
...
 
+template<typename T>
+static int l_game_get_config(lua_State *l)
+{
+	// Restrict the template function to only a few types
+	static_assert(
+		std::is_same<T, bool>::value ||
+		std::is_same<T, int>::value ||
+		std::is_same<T, float>::value ||
+		std::is_same<T, std::string>::value
```
sturnclaw: For consistency, these should use `std::is_same_v<T, type>`.
```suggestion
		std::is_same_v<T, bool> ||
		std::is_same_v<T, int> ||
		std::is_same_v<T, float> ||
		std::is_same_v<T, std::string>
```

src/lua/LuaGame.cpp
```diff
@@ -738,6 +739,75 @@ static int l_game_get_parts_from_date_time(lua_State *l)
...
+}
+template<typename T>
+static int l_game_set_config(lua_State *l)
+{
+	// Restrict the template function to only a few types
+	static_assert(
+		std::is_same<T, bool>::value ||
+		std::is_same<T, int>::value ||
+		std::is_same<T, float>::value ||
+		std::is_same<T, std::string>::value
```
sturnclaw:
```suggestion
		std::is_same_v<T, bool> ||
		std::is_same_v<T, int> ||
		std::is_same_v<T, float> ||
		std::is_same_v<T, std::string>
```

sturnclaw (approved):
I'm very happy with these changes and greatly anticipate its inclusion in master! I do apologize for how long it's taken to re-review this PR; I am indeed the "bus factor" right now and I appreciate @impaktor and others taking over the PR review and merging duties while I've been unavailable.

@mwerle please feel free to merge your own PRs if you feel they've been unattended long enough (and do not have highly-adversarial feedback expressed against them by other members of the committers group). I'd rather maintain momentum and enable the contributors who have time and energy to work on the game, instead of having every little PR get stuck waiting on me. You've done an excellent job with this PR (notwithstanding one minor icon alignment issue, the cause of which is quite arcane), and I don't mind fixing minor issues that get merged with a follow-up PR or commit to master.

I've left a few nitpick changes which would be worthwhile to fix; please feel free to address those now or in a latter commit to master.

data/pigui/modules/system-view-ui.lua
mwerle: Fixed in 752009f1529279d28203dcc0cdf3d44c50bd6deb

mwerle:
> I've left a few nitpick changes which would be worthwhile to fix; please feel free to address those now or in a latter commit to master.

Thank you; I fixed the remaining nitpicks.

I will rebase on `main` and squash commits before merging, but it may take a few days as I'm about to pick my mum up from the airport..

sturnclaw:
Totally understood! Feel free to hit the merge button once rebased/squashed - thank you again for tackling this!

mwerle:
> Totally understood! Feel free to hit the merge button once rebased/squashed - thank you again for tackling this!

Thank you, merged. I hope people like the new layout and additional functionality.

Thanks again for all the help and comments on this. Apologies if I sounded a bit snarky; just find it very difficult to pick things up again after several months. To be honest, I'd moved on to work on some KSP mods instead - only got reminded because I was reviewing some other stuff in GitHub. Would be nice to get back into Pioneer though.

data/lang/core/en.json
```diff
@@ -1851,6 +1919,18 @@
     "description": "Area unit: square meter (m²)",
     "message": "m²"
   },
+  "UNIT_TEMPERATURE_CELSIUS": {
+    "description": "Temperature unit: degrees Celsius (°C)",
+    "message": "°C"
+  },
+  "UNIT_TEMPERATURE_FAHRENHEIT": {
+    "description": "Temperature unit: Fahrenheit (F)",
+    "message": "F"
```
zonkmachine, line 1928: F -> °F

data/lang/core/en.json
```diff
@@ -1851,6 +1919,18 @@
...
+  },
+  "UNIT_TEMPERATURE_KELVIN": {
+    "description": "Temperature unit: degrees Kelvin (°K)",
+    "message": "°K"
```
zonkmachine, line 1932: °K -> K

data/lang/core/en.json
mwerle: Lol, reviews should come _before_ a merge, not after.. and it's not like this PR wasn't around for long enough either ;)

That said, well spotted and thank you.

mwerle: Fixed in 74f390aaeacdb1a5be8e34ca3d343e415f39b82d

@sturnclaw not sure this warrants a PR merge - just a cherry-pick on `master`? Or add it to a separate "tweaks and fixes" PR and merge once there's enough of them?

Anyway, I'll be afk for a few days so feel free to merge or whatever you think best.

And apologies to bother you about such a triviality.

data/lang/core/en.json
mwerle: Fixed in 74f390aaeacdb1a5be8e34ca3d343e415f39b82d

data/lang/core/en.json
mwerle: ...hmm, the new commit isn't showing up in this PR despite me having restored the branch and pushing to it. I guess closed PR's can't be amended and re-merged.

zonkmachine:
No, close this one and just fix it in a new pr. I don't know what ~~resurrecting~~ restoring a closed pr means/does.

PR #6175, opened 2025-06-17
Fix/minor fixes

mwerle:
A collection of semi-unrelated minor tweaks, fixes, and changes.

Pick-and-choose which should be added to master:

- [ ] fix muting/unmuting of effects and music. This "fix" only restores the original volume of the Effects and Music when unmuting. It does not actually fix muting/unmuting as such. Muting/unmuting really only works for the Master volume, it's mostly broken for Music and Effects.
- [ ]  fix sliders in the settings dialog from constantly generating "new" values, leading to constant calls to update the settings file and invoke potentially expensive additional logic.
- [ ] fix variable name for realisticScattering in the settings dialog. Trivial copy-paste error.
- [ ] add "onFocusGained" and "onFocusLost" UI events. (Thanks Sturnclaw)
- [ ] expose generic getters/setters to LUA for the settings file, as suggested by Sturnclaw

### DIFF
--- a/src/lua/LuaEngine.cpp
+++ b/src/lua/LuaEngine.cpp
@@ -417,6 +417,75 @@ static int l_engine_get_enum_value(lua_State *l)
 	return 1;
 }
 
+static int l_engine_settings_get_bool(lua_State *l)
+{
+	if (lua_isnone(l, 1))
+		return luaL_error(l, "SettingsGetBool takes one string argument");
+	const char *name = lua_tostring(l, 1);
+	lua_pushboolean(l, Pi::config->Int(name) != 0);
+	return 1;
+}
+static int l_engine_settings_get_int(lua_State *l)
+{
+	if (lua_isnone(l, 1))
+		return luaL_error(l, "SettingsGetInt takes one string argument");
+	const char *name = lua_tostring(l, 1);
+	lua_pushinteger(l, Pi::config->Int(name));
+	return 1;
+}
+static int l_engine_settings_get_number(lua_State *l)
+{
+	if (lua_isnone(l, 1))
+		return luaL_error(l, "SettingsGetNumber takes one string argument");
+	const char *name = lua_tostring(l, 1);
+	lua_pushnumber(l, Pi::config->Float(name));
+	return 1;
+}
+static int l_engine_settings_get_string(lua_State *l)
+{
+	if (lua_isnone(l, 1))
+		return luaL_error(l, "SettingsGetString takes one string argument");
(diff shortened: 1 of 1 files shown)


sturnclaw (approved):
I'll take "All of the Above" for 1800, Alex.

mwerle:
> I'll take "All of the Above" for 1800, Alex.

A single squashed commit or individual commits as-is?

sturnclaw:
Commits as-is - I'll merge once I've tested this branch, but still tied up with economy rewrites right now.

mwerle:
> Ideally, to save disk wear it'd be best to call that function when the settings menu is closed, but it could also be called at the end of each frame.

Thank you for picking that up!

----

In general the settings system should get an overhaul. Settings should be saved in memory, a dirty flag set, and only actually saved after some time has elapsed and/or when the settings window has been closed and/or when the game is exited. Apart from slowdowns, it doesn't matter quite so much on spinning disks, but I daresay most people are on SSD these days where it does matter.

Some settings (eg, changing volume) save the settings every frame while dragging the slider, for example (and prior to this set of fixes, every frame while the settings window was open).

-> Not sure fixing this should be part of this PR though as that goes beyond "quick".

sturnclaw:
If you want to tackle it in a separate PR, I'll merge this; the main concern I had was that the new API functions are not actually persistent as-implemented.

sturnclaw (approved):

mwerle:
> If you want to tackle it in a separate PR, I'll merge this; the main concern I had was that the new API functions are not actually persistent as-implemented.

I'll add the "Save" API. Just a minute.

I just meant refactoring the entire settings system to only save every-now-and-again is beyond the scope of this PR.

sturnclaw:
...I misread what you meant, my bad. Don't worry about it; I'll tackle the Save() and save-on-close part in a separate PR.

src/Pi.cpp
```diff
@@ -865,6 +865,14 @@ void Pi::App::HandleRequests()
 	internalRequests.clear();
 }
 
+void Pi::App::OnWindowKeyboardFocusChanged(bool newFocus)
+{
+    if (!PiGui::GetEventQueue().IsValid()) {
```
Gliese852, line 870: Spaces are inserted instead of tabs in several places.

impaktor: @Gliese852 If this is something affecting your current work, please keep indentation space/tab changes as a separate commit to help review (and future commit-archeologist), but I assume you're already on top of this

PR #6177, opened 2025-06-18
Redefine meaning of onEnterSystem/onLeaveSystem events

sturnclaw:
Per discussion on #6163, this PR adds player-specific versions of the `onEnterSystem`/`onLeaveSystem` events... by making those events player-specific.

This neatly maps to the semantic meaning of those names, as these events (when trigged for the player) change the actively-loaded system that the game is played in. Generic versions have been introduced as `onShipEnterSystem`/`onShipLeaveSystem` and the few mission modules interested in receiving events for all NPC ships have been migrated to use those events.

CC @robothauler for a quick review.

### DIFF
--- a/data/libs/Event.lua
+++ b/data/libs/Event.lua
@@ -318,19 +318,20 @@ end
 --
 -- Event: onEnterSystem
 --
--- Triggered when a ship enters a system after hyperspace.
+-- Triggered when a new system is loaded after hyperspace.
+-- For the event that is called when any ship enters the current system, see <onShipEnterSystem>.
 --
--- > local onEnterSystem = function (ship) ... end
+-- > local onEnterSystem = function (player) ... end
 -- > Event.Register("onEnterSystem", onEnterSystem)
 --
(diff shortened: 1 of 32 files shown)


impaktor:
We don't want to call them : `onPlayerEnterSystem`/`onPlayerLeaveSystem`?

sturnclaw:
No - if the events had the same semantic meaning as events for other ships jumping into/leaving the system, I'd name them that. These events encapsulate the semantic meaning of "The entire game area is about to be unloaded" / "A new game area has just been loaded", which is a higher-level semantic than just "the player has entered a new system".

To reinforce the point regarding semantics, all handlers (after this PR) which listen for the raw `onEnterSystem` et. al. events are specifically doing so to wipe caches they maintain for the actively loaded system. The remainder of handlers (which now use `onShipEnterSystem` et. al.) have internal codepaths for both the player and other ships, and treat them effectively the same.

sturnclaw:
Thanks for the thumbs-up @robothauler!

PR #6179, opened 2025-06-19
Save settings to disk when closing the settings window

sturnclaw:
This PR does two (mostly) related things:

- It prevents the config file from being written to disk on every toggle of a setting, only actually writing to disk when the settings menu is closed and there was a change in state.
- It moves the input bindings settings into their own `[Input]` section within the config file, preventing them from stomping (or being stomped by) actual user settings.

As we now have a generic settings interface for mods to read/write their own arbitrary settings, I consider it important to move the input bindings into their own section for the reason outlined above. It also has the nice side-effect of making config files neater and easier to read and debug.

Existing user-configured input bindings can be migrated with a simple cut-and-paste to the new `[Input]` section of the config file. It is unclear at this time whether the code has the information necessary to perform automatic migration of config files. The old bindings are not lost, but simply not read from anymore.

### DIFF
--- a/src/core/IniConfig.cpp
+++ b/src/core/IniConfig.cpp
@@ -24,8 +24,12 @@ void IniConfig::SetFloat(const std::string &section, const std::string &key, flo
 
 void IniConfig::SetString(const std::string &section, const std::string &key, const std::string &val)
 {
-	m_map[section][key] = val;
-	m_unsaved = true;
+	std::string &store = m_map[section][key];
+
+	if (store != val) {
+		store = val;
+		m_unsaved = true;
+	}
 }
 
 int IniConfig::Int(const std::string &section, const std::string &key, int defval) const
@@ -163,6 +167,10 @@ bool IniConfig::Save()
 		return false;
 	}
 
+	if (!Write(*m_fs, m_path)) {
+		return false;
+	}
+
 	m_unsaved = false;
-	return Write(*m_fs, m_path);
+	return true;
 }
--- a/src/core/IniConfig.cpp
+++ b/src/core/IniConfig.cpp
@@ -25,6 +25,7 @@ void IniConfig::SetFloat(const std::string &section, const std::string &key, flo
 void IniConfig::SetString(const std::string &section, const std::string &key, const std::string &val)
 {
 	m_map[section][key] = val;
+	m_unsaved = true;
 }
 
 int IniConfig::Int(const std::string &section, const std::string &key, int defval) const
@@ -162,5 +163,6 @@ bool IniConfig::Save()
 		return false;
 	}
 
+	m_unsaved = false;
 	return Write(*m_fs, m_path);
 }
--- a/src/core/IniConfig.h
+++ b/src/core/IniConfig.h
@@ -60,6 +60,8 @@ public:
 	bool HasEntry(const std::string &key) const { return HasEntry("", key); }
 	SectionMapType &GetSections() { return m_map; }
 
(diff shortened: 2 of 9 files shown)


src/core/IniConfig.cpp
```diff
@@ -25,6 +25,7 @@ void IniConfig::SetFloat(const std::string &section, const std::string &key, flo
 void IniConfig::SetString(const std::string &section, const std::string &key, const std::string &val)
 {
 	m_map[section][key] = val;
+	m_unsaved = true;
```
mwerle: Suggest to avoid unnecessary writes:

```suggestion
	if (m_map[section][key] != val) {
		m_map[section][key] != val
		m_unsaved = true;
	}
```

src/core/IniConfig.cpp
```diff
@@ -162,5 +163,6 @@ bool IniConfig::Save()
 		return false;
 	}
 
+	m_unsaved = false;
 	return Write(*m_fs, m_path);
```
mwerle:
```suggestion
	if (Write(*m_fs, m_path)) {
		m_unsaved = false;
		return true;
	}
	return false;
```

src/core/IniConfig.h
```diff
@@ -60,13 +60,16 @@ class IniConfig {
 	bool HasEntry(const std::string &key) const { return HasEntry("", key); }
 	SectionMapType &GetSections() { return m_map; }
 
+	bool HasUnsavedChanges() const { return m_unsaved; }
```
mwerle, line 63: Nitpick: more idiomatic is "IsDirty()" (with similar changes to the member variable etc), but this is clear enough.

src/lua/LuaEngine.cpp
```diff
@@ -486,6 +484,13 @@ static int l_engine_settings_set_string(lua_State *l)
 	return 0;
 }
 
+static int l_engine_save_settings(lua_State *l)
+{
+	if (Pi::config->HasUnsavedChanges())
+		Pi::config->Save();
```
mwerle, line 490: Should "Save()" internalise the call to "HasUnsavedChanges()"? That way it would always do the check even when called from the C++ side.

ie:
```
bool Save() {
  if (HasUnsavedChanges()) {
    // do the saving
  }
```

sturnclaw: IMO no - we (for example) may want to force a `Save()` to write to disk even if there aren't any changes; for example, when creating the file for the first time.

I'd prefer not to make the Save method "smarter" but rather have the calling code just smart enough to know when it's appropriate to call it.

src/core/IniConfig.cpp
sturnclaw: Good catch, thank you! I don't think the return value of `Save()` is ever checked to see if a write fails...

src/core/IniConfig.cpp
sturnclaw: Sure, I can make that change; it wouldn't be a problem.

src/lua/LuaEngine.cpp
mwerle: When creating the file for the first time, by definition, there are changes (nil -> some value).

I'd prefer not to have to duplicate each location which calls `Save()` to also have to call `HasChanges` (too easy to forget and imho unnecessary boilerplate), but your call :)

sturnclaw:
All review feedback has been addressed, merging this branch once CI finishes.

PR #6180, opened 2025-06-20
feat(ui): target radial support hyperspace clouds

mwerle:
Implement a long-running TODO item in the target radial popup menu.

When the target is a departure hyperspace cloud and the player has a hyperspace cloud analyzer fitted, add a target option to set the players' hyperspace target to the destination system of the cloud.

TODO Items:
- [x] update the route planner to reflect setting the hyperspace destination.

----
Update route planner:

Trying to `require` the hyperjumpPlanner directly results in an error "too many C levels".

Possibly we can expose the hyperjump-planner via a "gettter" on the system map object or otherwise have a small utility class for managing hyperspace jumps. Basically it should be possible to do everything which the player can do manually using the interface via an API. This could be useful for various missions.

There should also be a helper function to convert a SystemPath object into one which can be passed as a hyperjump destination; currently there are several different ways of how to do this in the game code, such as:

```lua
local target = dest.GetSystemPathOnly()
```
or
```lua
local target = dest:IsBodyPath() and dest:GetSystemBody().nearestJumpable.path or dest:GetStarSystem().path
```

The latter seems to achieve what is actually desired, but this is IMHO too complex as boilerplate and also relies on internal knowledge of the various classes which is a code smell to me.

### DIFF
--- a/data/pigui/libs/radial-menu.lua
+++ b/data/pigui/libs/radial-menu.lua
@@ -72,7 +72,34 @@ local hasAutopilotLevel = function(level)
 	return (Game.player["autopilot_cap"] or 0) >= level
 end
 
--- TODO: add cloud Lang::SET_HYPERSPACE_TARGET_TO_FOLLOW_THIS_DEPARTURE
+local radial_menu_actions_hyperspace_cloud = {
+	{
+		icon = ui.theme.icons.hyperspace,
+		tooltip = lc.SET_HYPERSPACE_TARGET_TO_FOLLOW_THIS_DEPARTURE,
+		action = function(target)
+			local hypercloud_level = (Game.player["hypercloud_analyzer_cap"] or 0)
+			local ship = not target:IsArrival() and target:GetShip()
+			if hypercloud_level > 0 then
+				-- implicitly drops second return value of GetHyperspaceDestination() when using this "and" construct
+				local destPath = ship and ship:GetHyperspaceDestination()
+				if destPath then
+					local target = destPath:IsBodyPath() and destPath:GetSystemBody().nearestJumpable.path or destPath:GetStarSystem().path
+					Game.player:SetHyperspaceTarget(target)
+					Game.sectorView:ClearRoute()
+					Game.sectorView:AddToRoute(target)
+					-- FIXME: lazy-require here due to insufficient event handling around hyperjump route management.
+					require 'pigui.modules.hyperjump-planner'.updateRouteList()
+					Game.sectorView:SwitchToPath(target)
+					ui.playSfx("OK")
+				end
+			else
+				Game.AddCommsLogLine(lc.NO_HYPERCLOUD_SCANNER_INSTALLED)
+			end
+		end
+
+	}
+}
+
 local radial_menu_actions_station = {
 	{
 		icon=ui.theme.icons.comms, tooltip=lc.REQUEST_DOCKING_CLEARANCE,
@@ -164,6 +191,10 @@ function ui.openDefaultRadialMenu(id, body, pos, action_binding)
 		end
 		if body:IsStation() then
 			for _,v in pairs(radial_menu_actions_station) do
+				table.insert(actions, v)
+			end
+		elseif body:IsHyperspaceCloud() then
+			for _,v in pairs(radial_menu_actions_hyperspace_cloud) do
 				table.insert(actions, v)
 			end
 		elseif body:GetSystemBody() then
--- a/data/pigui/libs/radial-menu.lua
+++ b/data/pigui/libs/radial-menu.lua
@@ -186,21 +186,13 @@ local radial_menu_actions_systembody = {
 function ui.openDefaultRadialMenu(id, body, pos, action_binding)
 	if body then
 		local actions = {}
-		for _,v in pairs(radial_menu_actions_all_bodies) do
-			table.insert(actions, v)
-		end
+		table.append(actions, radial_menu_actions_all_bodies)
 		if body:IsStation() then
-			for _,v in pairs(radial_menu_actions_station) do
-				table.insert(actions, v)
-			end
+			table.append(actions, radial_menu_actions_station)
 		elseif body:IsHyperspaceCloud() then
-			for _,v in pairs(radial_menu_actions_hyperspace_cloud) do
-				table.insert(actions, v)
-			end
+			table.append(actions, radial_menu_actions_hyperspace_cloud)
 		elseif body:GetSystemBody() then
-			for _,v in pairs(radial_menu_actions_systembody) do
-				table.insert(actions, v)
-			end
+			table.append(actions, radial_menu_actions_systembody)
 		end
 		ui.openRadialMenu(id, body, 1, defaultRadialMenuIconSize, actions, defaultRadialMenuPadding, pos, action_binding)
 	end


mwerle:
image

data/pigui/libs/radial-menu.lua
```diff
@@ -72,7 +72,34 @@ local hasAutopilotLevel = function(level)
 	return (Game.player["autopilot_cap"] or 0) >= level
 end
 
--- TODO: add cloud Lang::SET_HYPERSPACE_TARGET_TO_FOLLOW_THIS_DEPARTURE
+local radial_menu_actions_hyperspace_cloud = {
+	{
+		icon = ui.theme.icons.hyperspace,
+		tooltip=lc.SET_HYPERSPACE_TARGET_TO_FOLLOW_THIS_DEPARTURE,
```
sturnclaw: Nitpick: spacing around the equals-sign.

data/pigui/libs/radial-menu.lua
```diff
@@ -166,6 +193,10 @@ function ui.openDefaultRadialMenu(id, body, pos, action_binding)
 			for _,v in pairs(radial_menu_actions_station) do
 				table.insert(actions, v)
 			end
+		elseif body:IsHyperspaceCloud() then
+			for _,v in pairs(radial_menu_actions_hyperspace_cloud) do
+				table.insert(actions, v)
+			end
```
sturnclaw: Out-of-scope nitpick: this (and other branches) can be migrated to:
```suggestion
			table.append(actions, radial_menu_actions_hyperspace_cloud)
```

The existing for loop is technically "invalid" anyways because it's iterating over array-tables with the map-table iterator (pairs) which has undefined iteration order (though implementation-defined behavior is to loop over the array-part first in numeric order).

data/pigui/libs/radial-menu.lua
```diff
@@ -72,7 +72,34 @@ local hasAutopilotLevel = function(level)
...
+				local destPath,_ = ship and ship:GetHyperspaceDestination()
+				if destPath then
+					--Game.player:SetHyperspaceTarget(destPath:SystemOnly())
+					local target = destPath:IsBodyPath() and destPath:GetSystemBody().nearestJumpable.path or destPath:GetStarSystem().path
+					Game.player:SetHyperspaceTarget(target)
+					--TODO: figure out how to set the destination so that the
+					-- route planner also contains the jump, otherwise the jump
+					-- is cleared if the player enters the sector map
+					Game.sectorView:ClearRoute()
+					Game.sectorView:SwitchToPath(target)
```
sturnclaw: You'll want to add after `ClearRoute()`:

```lua
Game.sectorView:AddToRoute(target)
-- FIXME: lazy-require here due to insufficient event handling around hyperjump route management.
require 'pigui.modules.hyperjump-planner'.updateRouteList()
```

sturnclaw (changes requested):
With apologies for the delay, I've finally gotten a chance to review this.

Overall I appreciate the micro-feature (and the compactness of the diff!), and there are only a few minor changes to be made to finish out the PR.

sturnclaw:
> Trying to require the hyperjumpPlanner directly results in an error "too many C levels".

This is stack recursion due to circular imports; `require 'ui'` loads the radial menu definitions, which (with this change) loads the hyperjump planner which attempts to `require 'ui'` which isn't done loading yet so it starts a new load step, which recurses all the way down.

sturnclaw:
> There should also be a helper function to convert a SystemPath object into one which can be passed as a hyperjump destination; currently there are several different ways of how to do this in the game code, such as:

Feel free to hang this off of SectorMap or (maybe) Game; I don't think it's a good fit for the SystemPath API directly, and SystemBody already provides the `nearestJumpable` accessor.

data/pigui/libs/radial-menu.lua
mwerle: Fixed in c738b32b6f4d1072294035a5b5a5c89ecd9f64d4

data/pigui/libs/radial-menu.lua
mwerle: Fixed in 96a64c8fb96c1e864f104616102fb92ce97de284

Thank you for the suggestion!

data/pigui/libs/radial-menu.lua
mwerle: ... the `require` goes inside the function, or at the top with all the other `require` statements?

sturnclaw: Inside the function. `require` is not actually a keyword, it's just a function like any other. Putting it in the function defers its execution until after the current module has finished loading; if the module we're requiring is already loaded, it will be returned from the `package.loaded` table rather than loaded again. [See the manual](https://www.lua.org/manual/5.2/manual.html#pdf-require) for the semantics that we (loosely) model.

mwerle: Fixed in a5f667fd18ed8175f8f81292756d678bfefbb95a

data/pigui/libs/radial-menu.lua
```diff
@@ -72,7 +72,33 @@ local hasAutopilotLevel = function(level)
...
--- TODO: add cloud Lang::SET_HYPERSPACE_TARGET_TO_FOLLOW_THIS_DEPARTURE
+local radial_menu_actions_hyperspace_cloud = {
+	{
+		icon = ui.theme.icons.hyperspace,
+		tooltip = lc.SET_HYPERSPACE_TARGET_TO_FOLLOW_THIS_DEPARTURE,
+		action = function(target)
+			local hypercloud_level = (Game.player["hypercloud_analyzer_cap"] or 0)
+			local ship = not target:IsArrival() and target:GetShip()
+			if hypercloud_level > 0 then
+				local destPath,_ = ship and ship:GetHyperspaceDestination()
```
sturnclaw: `GetHyperspaceDestination` returns a single value, so only the `destPath` variable needs to be declared and assigned to here.

mwerle: This is the documentation (in `Ship.meta.lua`):
```
-- Return the current hyperspace destination.
---@return SystemPath path the destination system and body
---@return string name the name of the destination
function Ship:GetHyperspaceDestination() end
```
and the implementation and all usages (except 1) confirm this.

The non-conforming usage is in `hyperjump-planner.lua` on Line 44 - while it uses `player` instead of `ship` I don't see the `player` class overriding this function.

mwerle:
> Looks good to me, no issues that I've found in testing. Please squash together any fixup commits and I'll merge once you're satisfied with the commit history.

Thank you very much for your time and effort reviewing and testing. We need to clone you! ;)

Git history squashed and pushed.  I see the PR has been approved, but the various comments have not been resolved?

data/pigui/libs/radial-menu.lua
sturnclaw: Whoops, I skipped right past that first return annotation when diagnosing why the tooling was reporting a missing value for the second variable. That is a very valid catch.

However... there is still only a single value here because of the `and` expression. It's a single-value boolean operator, which means all return values after the first from `ship:GetHyperspaceDestination()` are implicitly discarded.

sturnclaw:
> Thank you very much for your time and effort reviewing and testing. We need to clone you! ;)

I agree! I apologize for the very long time it's taken to get around to finishing the review; IRL has been quite hectic recently. Please feel free to review and merge smaller PRs yourself if you feel comfortable with doing so.

sturnclaw:
> I see the PR has been approved, but the various comments have not been resolved?

Thanks, I've now resolved everything except the very last comment. If you intend to address that last nitpick, I'll hold off on merging.

data/pigui/libs/radial-menu.lua
mwerle: ... so if I actually _wanted_ the second parameter, I can't use this `and` construct?

So the "proper" way to invoke this function and have this check is really:

```
local destPath
if ship then
  destPath,_ = ship:GetHyperspaceDestination()
end
```

?

mwerle:
> > I see the PR has been approved, but the various comments have not been resolved?
> 
> Thanks, I've now resolved everything except the very last comment. If you intend to address that last nitpick, I'll hold off on merging.

I'm happy to address the nitpick in order to have good reference code :)

Sorry, have to run to a medical checkup; can finish this in a couple of hours.

data/pigui/libs/radial-menu.lua
sturnclaw: > ... so if I actually _wanted_ the second parameter, I can't use this `and` construct?
> 
> So the "proper" way to invoke this function and have this check is really:
> 
> ```
> local destPath
> if ship then
>   destPath,_ = ship:GetHyperspaceDestination()
> end
> ```
> 
> ?

Correct, if you want both parameters you would want to declare both variables and execute the assignment inside an if block. However, the current code never uses the second variable, so instead this should just simplify to:

```suggestion
				local destPath = ship and ship:GetHyperspaceDestination()
```

mwerle: I take it this is a Well Known Construct among LUA programmers, or is a comment warranted?

sturnclaw: Eh... a comment is probably warranted, given that most contributors are beginner Lua programmers to start with. This is a scenario I'd consider somewhere between journeyman and professional levels of Lua knowledge.

mwerle: I've updated (and force-pushed) the suggested change. Please either resolve and merge, or request addition of a comment as you think is best. Thank you.

mwerle: SNAP.
Ok, added a comment as well :)

mwerle:
> > Thank you very much for your time and effort reviewing and testing. We need to clone you! ;)
> 
> I agree! I apologize for the very long time it's taken to get around to finishing the review; IRL has been quite hectic recently. Please feel free to review and merge smaller PRs yourself if you feel comfortable with doing so.

..and my apologies for having been pushy.

For tiny fixes, sure, I think I'd be comfortable doing so. But I still have a lot to learn about LUA, and the codebase and its best practices in general so I'm a bit wary of merging non-trivial changes without someone giving them a quick once-over..

sturnclaw:
At long last... merged!

PR #6181, opened 2025-06-20
feat(ui): rework target scanner display

### LINKED ISSUE #6162: UI: Interface icons show up on top of the side panels
### Observed behaviour

Some icons on the side of the screen show up on top of expanded side-bar panels.

### Expected behaviour

Side-bar panels cover the icons

### Steps to reproduce

1. Add some missiles to a ship -> missile icon shows up in the in-flight view
2. Open the comms window
3. Missile icon shows up on top of comms, interfering with the text

- Pioneer: Trunk (2025-06-15)
- OS: Debian GNU/Linux 12

N/A, instead attaching screenshot.

Image

mwerle:
Partially fixes #6162

Convert it to a HudModule - this means it will be automatically hidden when a sidebar module is opened.

The way the data for both ship details and hyperspace clouds are rendered has been cleaned up and modernised.

### Old:
image

Overlapping with right-hand sidebar module:
image

### New:
image

No more overlapping (auto-hide):
image


bszlrd:
Nice! 
I think it would be useful if there would be a show-hide icon for it. There's a planned orbit display that would occupy the same area in its opened form. (I can't pull out the mockups right now)

mwerle:
if the orbit display is also a hud module, they'll automatically rearrange with each other according to priority.

mwerle:
WiP : as part of this work, I've decided to also refactor the hyperspace cloud a little bit and moved the hyperspace origin into it, instead of overloading the ships' hyperspace destination.

I've pushed this to a separate branch as I'm not sure it's entirely warranted here, and I don't want to hold this PR up unnecessarily, but if people are interested in taking a quick look, it's here : https://github.com/pioneerspacesim/pioneer/commit/64ab51e818a49924a67b075b97b793751f8b845b

data/pigui/modules/flight-ui/target-scanner.lua
```diff
@@ -5,116 +5,187 @@ local ui = require 'pigui'
 local Vector2 = _G.Vector2
 
 -- cache ui
-local pionillium = ui.fonts.pionillium
+local font_heading = ui.fonts.pionillium.medlarge
+local font_content = ui.fonts.pionicons.medium
```
sturnclaw: Note: these font definitions are retained for legacy purposes only. New and updated code should migrate to one of the semantic breakpoints (e.g. `body`, `heading`, `details`) for visual conformity with the rest of the UI.

data/pigui/modules/flight-ui/target-scanner.lua
```diff
@@ -5,116 +5,187 @@ local ui = require 'pigui'
...
+-- draw a 2-column table of name/value data
+local function drawTable(data, maxWidth)
+	-- data is a table containing:
+	-- - name of item
+	-- - value of item
+	if not data or #data == 0 then return end
+	maxWidth = maxWidth or ui.gauge_width
+	local nameWidth = 0
+	local valueWidth = 0
+	ui.columns(2, "Attributes", false)
```
sturnclaw: Note that you can (and likely should) use the `ui.beginTable()` / `ui.endTable()` API here. It doesn't yet have a convenient wrapper, but will automatically size columns as needed; the `ui.columns()` API is considered legacy. See the Table documentation in `imgui.h` and other uses of `ui.beginTable()` in the codebase for details.

data/pigui/modules/flight-ui/target-scanner.lua
```diff
@@ -5,116 +5,187 @@ local ui = require 'pigui'
...
+	local spacing = ui.gauge_height * 1.4
+	local yOff = Vector2(0, ui.gauge_height * 0.5)
+	local uiPos = ui.getCursorScreenPos()
+
 	if shield then
-		ui.gauge(uiPos - Vector2(ui.gauge_width, 0), shield, nil, nil, 0, 100, icons.shield, colors.gaugeShield, lui.HUD_SHIELD_STRENGTH)
+		ui.gauge(uiPos + yOff, shield, nil, nil, 0, 100, icons.shield,
+			colors.gaugeShield, lui.HUD_SHIELD_STRENGTH, maxWidth)
+		yOff.y = yOff.y + spacing
 	end
```
sturnclaw, line 78: Note: since this widget has been moved to be bottom-aligned, I would suggest rendering the gauges below the detail text instead of above it.

data/pigui/modules/flight-ui/target-scanner.lua
```diff
@@ -5,116 +5,187 @@ local ui = require 'pigui'
...
-	local cargo = target.usedCargo
-	if engine then
-		engine = engine:GetName()
-	else
-		engine = 'No Hyperdrive'
-	end
-	local uiPos = Vector2(ui.screenWidth - 30, 1 * ui.gauge_height)
+
+	local data = {
+		{ name = target.label, value = target:GetShipType() },
```
sturnclaw: I would suggest rendering the target label before the details list in a larger font (e.g. `heading`) as as `Target: {label}`; this would create a visual indicator that the block of details applies to the currently selected target.

data/pigui/modules/flight-ui/target-scanner.lua
```diff
@@ -5,116 +5,187 @@ local ui = require 'pigui'
...
+	-- instead somehow get it from the displayTargetScannerFor() function
+	local height = (ui.gauge_height * 1.4) * 2 + textHeight * 4 + ui.gauge_height * 0.5
+
+	local pos, size = ui.rectcut(min, max, height, ui.sides.bottom)
+
+	ui.setNextWindowPos(pos, "Always")
+	ui.setNextWindowSize(size, "Always")
+	ui.setNextWindowPadding(Vector2(0, 0))
+	ui.window("TargetScanner", scannerWindowFlags, function()
+		displayTargetScannerFor(target, size.x)
```
sturnclaw, line 112: Note that `size.x` for `maxWidth` here is somewhat visually inconsistent with the left sidebar, which renders gauges at the width of the time button window. It's not a significant concern by any means, but I'd suggest it might be possible to format the data in such a way that a narrower width is acceptable.

data/pigui/modules/flight-ui/target-scanner.lua
```diff
@@ -5,116 +5,187 @@ local ui = require 'pigui'
...
+		ui.withFont(font_heading, function()
+			ui.text(target:GetLabel())
+		end)
+		if ship then
+			-- NOTE: ships in arrival clouds have their destination set to the
+			-- source system. This should probably get refactored some time in
+			-- the future. While clever reuse of existing state, it makes for
+			-- challenging maintenance and understanding of the code.
+			local _,systemName = ship:GetHyperspaceDestination()
+			local systemLabel = arrival and lui.HUD_HYPERSPACE_ORIGIN or lui.HYPERSPACE_DESTINATION
```
sturnclaw: Invalid string ID here:

```suggestion
			local systemLabel = arrival and lui.HUD_HYPERSPACE_ORIGIN or lui.HUD_HYPERSPACE_DESTINATION
```

data/pigui/modules/flight-ui/target-scanner.lua
```diff
@@ -5,116 +5,187 @@ local ui = require 'pigui'
...
+
+			local data = {
+				{ name = lui.HUD_MASS, value = formatMass(ship.staticMass) },
+				{ name = systemLabel, value = systemName },
+				{ name = lui.DATE, value = ui.Format.Datetime(target:GetDueDate()) }
```
sturnclaw: It is unclear to the user whether this is meant to be the arrival or destination date. A new string is ideally required.

data/pigui/modules/flight-ui/target-scanner.lua
mwerle: Fixed in bc191f1d8c028ec03afb01b43e571e2f06a0c7a1

data/pigui/modules/flight-ui/target-scanner.lua
mwerle: Fixed in 1eb26f6777771ff05113a3580f9d4b01da78785a

data/pigui/modules/flight-ui/target-scanner.lua
mwerle: Fixed in bc191f1d8c028ec03afb01b43e571e2f06a0c7a1

I used `{label} {type}` instead of an explicit "Target" heading to avoid needing an extra data line.

data/pigui/modules/flight-ui/target-scanner.lua
mwerle: Fixed in c157c5e32758ddda2fc1b3442909f89c90f88d1a

data/pigui/modules/flight-ui/target-scanner.lua
mwerle: Fixed in 5c839a2b6135888482e34dbf4da1b0780c6d4c02

data/pigui/modules/flight-ui/target-scanner.lua
mwerle: Fixed in a5ac13dfc023012f7ccaf627a121a49f0a02056d

mwerle:
> Overall I'm quite happy with this PR, thank you for taking the initiative with it! I have a few nitpicks and places where I'd like to see some improvements, most notable of which is better differentiating where the hypercloud analyzer display ends and the target scanner display begins:

...

> Another thing I'd suggest is to at least display the type of ship "in" the departure/arrival cloud, since we're already displaying its mass. I think we can make the assumption that the analyzer has hyperspace "wake signatures" to compare against known models of ships.

How's this : 

image


EDIT: There's a newer screenshot which restricts the width here : https://github.com/pioneerspacesim/pioneer/pull/6181#discussion_r2239402421

data/pigui/modules/flight-ui/target-scanner.lua
mwerle: Fixed in 7dec03b0b1978ed60a50a0f8ca87bbb407e20f20

I manually adjusted the position and size of the windows. I initially tried to use `rectcut` but it was giving very strange results; I guess I don't fully understand how it works and how `min` and `max` tie in to the HUD modules.

Please let me know if there's a better way (I realise my way is a bit of a hack).

image

mwerle:
I think I addressed all comments. Please re-review at your leisure.

(Will squash the various fixup commits once accepted prior to merge)

mwerle:
> Another thing I'd suggest is to at least display the type of ship "in" the departure/arrival cloud, since we're already displaying its mass. I think we can make the assumption that the analyzer has hyperspace "wake signatures" to compare against known models of ships.

I've done this for now, but in future perhaps we can have different versions of the Hyperspace Cloud Analyzer, giving different amounts of information. Then again, could just lead to unnecessary equipment creep.

bszlrd:
I think that kind of equipment creep would be a good thing. Especially when the syístems become more complex. Like one sensor could have lower range and less data, but cheaper and available at lower tech levels. Another medium quality would be at higher tech levels, would have more range, and could guess the ship size 50% of the time, and the best version would guess 90% of the time at higher ranges, but is a very sofisticated instrument. Power and maintenance needs could also differ.

data/pigui/modules/flight-ui/target-scanner.lua
sturnclaw: I'd recommend doing the conversion in the `draw` function of the respective hud modules, with one of two ways:

```lua
-- Update the min coordinate to be no greater than timeWindowSize.x units from the right margin
min.x = math.max(min.x, max.x - ui.timeWindowSize.x)

-- or, making new min/max vectors
local pos, size = ui.rectcut(min, max, ui.timeWindowSize.x, ui.sides.right)
drawTargetScanner(pos, pos + size)
```

sturnclaw:
> I think that kind of equipment creep would be a good thing. Especially when the syístems become more complex. Like one sensor could have lower range and less data, but cheaper and available at lower tech levels. Another medium quality would be at higher tech levels, would have more range, and could guess the ship size 50% of the time, and the best version would guess 90% of the time at higher ranges, but is a very sofisticated instrument. Power and maintenance needs could also differ.

I generally agree that having additional equipment options is a good thing, as long as those options represent meaningful choices for the player. If the "best option" has no cost other than money, players will only ever fit something else as a temporary stopgap.

data/pigui/themes/default.lua
```diff
@@ -206,6 +207,7 @@ theme.colors = {
 	notificationBackground	= styleColors.panel_800,
 	modalBackground			= styleColors.panel_900,
 	tableBackground			= styleColors.primary_900,
+	tableHeaderBg			= styleColors.gray_800,
```
sturnclaw: The `panel` set of semantic colors is intended to be used as background colors, while the `gray` set are foreground colors (for e.g. text elements). In this particular case, I'd strongly recommend using `panel_800` for consistency with the rest of the UI instead.

The inability to directly theme the `TableHeaderBg` color is something that can be addressed by adding the enum definition to `src/lua/LuaPiGui.cpp` in the `imgui_col_enums` definition.

sturnclaw (approved):
One minor change that I'd like to see addressed regarding the addition of the `gray_800` color breakpoint, but otherwise I'm exceptionally happy with this PR!

data/pigui/themes/default.lua
mwerle: The problem is that no currently defined color matches the color used in the table header row. This appears to only be implemented in the C++ at the moment, which is why I created a new color in the LUA to mimic it.

What, if anything, do the numbers in the color definitions mean anyway?

data/pigui/modules/flight-ui/target-scanner.lua
mwerle: Changed in 92f66e816e28e6946f4eaf8861de7c630ea1b263

I moved the adjustment into the `draw` functions.

Using additional calls to `rectcut` messes up the display when multiple HUD modules are active at the same time as each call to `rectcut` affects subsequent calls.

data/pigui/themes/default.lua
mwerle: image

This is using `panel_800` - as you can see, it's a different grey to the one used in the actual table header.

For context, the header for the "hyperspace cloud analyzer" is drawn as a filled background rectangle and then a text line as I can't figure out a way to have a table row span multiple columns.

The "target scanner" header is a table header row, rendered using the `ImGuiCol_Header` colour. This is nominally exposed to LUA as the `Header` colour, but the current theme has it as a purple colour.

So this is probably a bug that the theme affects the LUA side, but not the C++ side..

When using `colors.Header`:
image

mwerle: @sturnclaw ? (or anybody else with knowledge of how this works..)

data/pigui/modules/flight-ui/target-scanner.lua
mwerle: Sorry, the fixup commit no longer exists - I squashed and rebased as the history was getting a bit unwieldy - please see the "draw" functions of the HUD modules at the bottom of `target-scanner.lua` for the revised implementation.

data/pigui/themes/default.lua
sturnclaw: Numbers in color definitions are the color "value", loosely based on a [semantic color system](https://dev.to/ynab/a-semantic-color-system-the-theory-hk7). They provide a categorization system to ensure that colors with the same value number appear relatively consistent across the entire UI.

The inability to theme the table header color is because the table header row doesn't use the `Header` color at all, but rather the `TableHeaderBg` color (see imgui_tables.cpp in the `Imgui::TableBeginRow()` function). That color is not yet exposed to Lua; to do so, you'd need to add a new entry to the `imgui_col_enums` table in `src/lua/LuaPiGui.cpp`, starting on line 254.

Once that color has been added, it should be normalized with our theme by modifying `data/pigui/themes/default.lua` to add `TableHeaderBg = styleColors.panel_800,` somewhere around line 240.

mwerle: So... it seems that the LUA side doesn't directly use the ImGui theme colours (defined in `contrib/imgui/imgui_draw.cpp`) and instead defines its own (which don't always match the ImGui ones) in `data/pigui/themes/default.lua`. It's not helped that ImGui uses float values, PiGui uses decimals, and LUA uses hexadecimals to define all the colours.

The LUA side can override the C++ side using the "withStyleXxx" APIs. I've added some missing colour values (such as `TableHeaderBg`) to the LUA interface.

I've also added the exact grey used by ImGui to render table header row backgrounds (see `imgui_tables.cpp:TableBeginRow()`) to the LUA theme (as `panel_800a`). As nobody has answered I still have no clue what the numbers in these theme color basenames are supposed to mean. There's also talk of "breakpoints" in the comments which, to me, has something to do with debugging, not colours or UI themes... anyway.

----

The TL;DR is that I've exposed the ImGuiCol_TableHeaderBg to LUA, and am using that to render the background of the title for the Cloud Analyzer in order to match the header row used by the Target Analyzer.

The alternative would be to render the Target Analyzer table using `withStyleColors` and override the TableHeaderBg with something else, eg, `panel_800`, and use the same colour for the Cloud Analyzer title background.

data/pigui/modules/flight-ui/target-scanner.lua
sturnclaw: Yep, I forgot to take into account that these are separate modules and thus mutating the `max` value through the rectcut function would be a Bad Idea. The change looks good; thanks!

data/pigui/themes/default.lua
mwerle: > Once that color has been added, it should be normalized with our theme by modifying `data/pigui/themes/default.lua` to add `TableHeaderBg = styleColors.panel_800,` somewhere around line 240.

... oh, so simply defining the colours here override the C++ colours? That would be a useful comment to add :)

mwerle: ... yes they do.

Ok, fixed up - the `TableHeaderBg` (newly added and exposed to LUA) is now defined as `panel_800`.

sturnclaw: > As nobody has answered I still have no clue what the numbers in these theme color basenames are supposed to mean.

No one (that I've been able to find) really knows *what* the numbers mean precisely, they're just a rough categorical scale from 000=white to 999=black. You can kind of think of the value as a Z-index; UI elements with a smaller color value number go on top / inside of elements with a greater number (until we start talking about light themes, in which case it inverts).

The linked article goes into the philosophy of the color system we've (mostly) adopted, but the convention of using a three-digit suffix to indicate relative color value is a web design thing going back at least a decade to my knowledge.

mwerle: I had a quick read through that article (thank you). IMHO it might have been useful to link that at the top of the `themes/default.lua` file..

mwerle: > The inability to directly theme the TableHeaderBg color is something that can be addressed by adding the enum definition to src/lua/LuaPiGui.cpp in the imgui_col_enums definition.

And my apologies, I missed / didn't understand this part of your original comment..

sturnclaw (approved):
Looks good to me!

We historically don't maintain our own "copies" of deprecated ImGui functions or enum values, and instead perform a migration to the new value once ImGui fully removes the deprecated value in an upstream update. That being said, there's no rule against maintaining a "deprecated list" in LuaPiGui, and I don't consider it to have any bearing on merging this PR.

data/pigui/themes/default.lua
sturnclaw: > I had a quick read through that article (thank you). IMHO it might have been useful to link that at the top of the `themes/default.lua` file..

I keep forgetting to do so every time I do a major theme update. If you don't intend to add such to this PR, I'll push it as a commit to master after merge.

mwerle: I can quickly add it.. Give me a minute.

mwerle:
> Looks good to me!
> 
> We historically don't maintain our own "copies" of deprecated ImGui functions or enum values, and instead perform a migration to the new value once ImGui fully removes the deprecated value in an upstream update. That being said, there's no rule against maintaining a "deprecated list" in LuaPiGui, and I don't consider it to have any bearing on merging this PR.

The deprecated colour value was already in the table and I was too lazy to see if it was being used anywhere, so didn't want to remove it and potentially break something...  and ImGui hasn't fully removed it yet. At least, depending on whether or not `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` is defined.

I did a `grep` through the codebase and it appears unused, so I can remove it prior to merge.

sturnclaw:
> I did a grep through the codebase and it appears unused, so I can remove it prior to merge.

The perfectionist in me would appreciate it - but it's not required.

PR #6187, opened 2025-06-25
Minor Terrain files cleanup

fluffyfreak:
There was an unused terrain colour `TerrainColorDeadWithWater`, added some comments about terrain type picked,and made some `const` into `constexpr`

nothing that should affect anything, just cleanup


impaktor:
> nothing that should affect anything, just cleanup

Then I expect fire and brimstone to rain down from the heavens upon merge.

fluffyfreak:
Fire and brimstone incoming then I guess :D

PR #6188, opened 2025-06-29
OpenAL audio backend

mfep:
- Defined `AudioBackend` pure virtual class that abstracts base audio functionality
- Extracted current audio system to `SDLAudioBackend`
- Implemented `AlAudioBackend` that uses OpenAL Soft
- Updated audio menu and INI config to be able to switch between audio backends
- Updated CMakeLists so that it is possible to build either with or without the OpenAL audio backend
  - On Windows, dependencies are provided by pioneer-thirdparty. Added the required binaries in https://github.com/pioneerspacesim/pioneer-thirdparty/pull/34

Note: this is what I think the most minimal changeset that brings OpenAL audio to Pioneer. It is out of scope for this PR to change how the rest of the program interacts with the audio system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,8 +200,6 @@ else (WIN32)
 	)
 endif (WIN32)
 
-configure_file(buildopts.h.cmakein buildopts.h @ONLY)
-
 LIST(APPEND PIONEER_CXX_FILES ${FILESYSTEM_CXX_FILES})
 
 option(USE_PIONEER_THIRDPARTY "Use pioneer's thirdparty library repository." OFF)
@@ -273,6 +271,8 @@ elseif(NOT DEFINED CACHE{BUILD_WITH_OPENAL} AND OpenAL_FOUND)
 endif()
 
 set(FMT_INSTALL OFF CACHE BOOL "Enable install of libfmt" FORCE)
+
+configure_file(buildopts.h.cmakein buildopts.h @ONLY)
 
 add_subdirectory(contrib/lz4)
 add_subdirectory(contrib/fmt)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -381,7 +381,6 @@ else()
 endif()
 
 target_link_libraries(pioneer-lib PUBLIC lz4 fmt::fmt $<$<BOOL:${BUILD_WITH_OPENAL}>:OpenAL::OpenAL>)
-target_compile_definitions(pioneer-lib PUBLIC $<$<BOOL:${BUILD_WITH_OPENAL}>:PI_BUILD_WITH_OPENAL>)
 
 # Fallback available on most legacy systems
 set(PIONEER_GL_LINK_TARGET OpenGL::GL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,7 +262,15 @@ endif (MSVC)
 find_package(Threads REQUIRED)
 find_package(Freetype REQUIRED)
 find_package(OpenGL REQUIRED COMPONENTS OpenGL)
-find_package(OpenAL CONFIG REQUIRED)
+if(NOT TARGET OpenAL::OpenAL)
+	find_package(OpenAL)
+endif()
+if (CACHE{BUILD_WITH_OPENAL} AND NOT OpenAL_FOUND)
+	message(FATAL_ERROR "Requested BUILD_WITH_OPENAL=ON, but the library is not found")
+elseif(NOT DEFINED CACHE{BUILD_WITH_OPENAL} AND OpenAL_FOUND)
+	message(STATUS "Found OpenAL, building with support")
+	set(BUILD_WITH_OPENAL ON CACHE BOOL "Enable or disable support for OpenAL")
+endif()
 
 set(FMT_INSTALL OFF CACHE BOOL "Enable install of libfmt" FORCE)
 
@@ -372,7 +380,8 @@ else()
 	message(WARNING, "Python 3 not found; enums will not be scanned.")
 endif()
 
-target_link_libraries(pioneer-lib PUBLIC lz4 fmt::fmt OpenAL::OpenAL)
+target_link_libraries(pioneer-lib PUBLIC lz4 fmt::fmt $<$<BOOL:${BUILD_WITH_OPENAL}>:OpenAL::OpenAL>)
+target_compile_definitions(pioneer-lib PUBLIC $<$<BOOL:${BUILD_WITH_OPENAL}>:PI_BUILD_WITH_OPENAL>)
 
 # Fallback available on most legacy systems
 set(PIONEER_GL_LINK_TARGET OpenGL::GL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,7 @@ endif (MSVC)
 find_package(Threads REQUIRED)
 find_package(Freetype REQUIRED)
 find_package(OpenGL REQUIRED COMPONENTS OpenGL)
+find_package(OpenAL CONFIG REQUIRED)
 
 set(FMT_INSTALL OFF CACHE BOOL "Enable install of libfmt" FORCE)
 
@@ -371,7 +372,7 @@ else()
 	message(WARNING, "Python 3 not found; enums will not be scanned.")
 endif()
 
-target_link_libraries(pioneer-lib PUBLIC lz4 fmt::fmt)
+target_link_libraries(pioneer-lib PUBLIC lz4 fmt::fmt OpenAL::OpenAL)
 
 # Fallback available on most legacy systems
 set(PIONEER_GL_LINK_TARGET OpenGL::GL)
--- a/COMPILING.txt
+++ b/COMPILING.txt
@@ -63,7 +63,7 @@ questions are welcome anytime.
     libassimp-dev >= 5.0.1
     libsdl2-dev
     libsdl2-image-dev
-    libopenal-dev
+    libopenal-dev (optional, used for the OpenAL sound backend)
 
    If your platform doesn't have assimp 5, you'll need to build it from
    source. It's available in pioneer-thirdparty, see below.
--- a/COMPILING.txt
+++ b/COMPILING.txt
@@ -63,6 +63,7 @@ questions are welcome anytime.
     libassimp-dev >= 5.0.1
     libsdl2-dev
     libsdl2-image-dev
+    libopenal-dev
 
    If your platform doesn't have assimp 5, you'll need to build it from
    source. It's available in pioneer-thirdparty, see below.
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -67,6 +67,10 @@
     "description": "",
     "message": "Attempt to repair hull"
   },
+  "AUDIO_BACKEND": {
+    "description": "Audio backend label in settings menu",
+    "message": "Audio Backend"
+  },
   "AUTOMATED_MESSAGE": {
     "description": "For comm messages from bots",
     "message": "AUTOMATED MESSAGE"
@@ -90,6 +94,14 @@
   "BACKWARD_ACCEL": {
     "description": "",
     "message": "Backward acceleration"
+  },
+  "BINAURAL_RENDERING": {
+    "description": "Label of the binaural rendering checkbox in the audio settings menu",
+    "message": "Binaural Rendering"
+  },
+  "BINAURAL_RENDERING_DESC": {
+    "description": "Tooltip of the binaural rendering checkbox in the audio settings menu",
+    "message": "When enabled, HRTF filters are applied to the audio output. Works best with headphones."
   },
   "BODY_IN_ROUTE": {
     "description": "Label indicating a star is part of a planned hyperjump route",
--- a/data/pigui/modules/settings-window.lua
+++ b/data/pigui/modules/settings-window.lua
@@ -334,6 +334,8 @@ end, function (_, drawPopupFn)
 end)
 
 local function showSoundOptions()
+	local available_backends = Engine.GetAvailableSoundBackends()
+	local current_backend_id = Engine.GetSoundBackendId()
 	local masterMuted = Engine.GetMasterMuted()
 	local masterLevel = Engine.GetMasterVolume()*100
 	local musicMuted = Engine.GetMusicMuted()
@@ -342,6 +344,9 @@ local function showSoundOptions()
 	local effectsLevel = Engine.GetEffectsVolume()*100
 
 	local c
+
+	c,new_backend_id = combo("Audio Backend", current_backend_id - 1, available_backends)
+	if c then Engine.SetSoundBackend(new_backend_id + 1) end
 
 	c,masterMuted = checkbox(lui.MUTE.."##master", masterMuted)
 	if c then Engine.SetMasterMuted(masterMuted) end
--- a/data/pigui/modules/settings-window.lua
+++ b/data/pigui/modules/settings-window.lua
@@ -335,7 +335,7 @@ end)
 
 local function showSoundOptions()
 	local available_backends = Engine.GetAvailableSoundBackends()
-	local current_backend_id = Engine.GetSoundBackendId()
+	local current_backend = Engine.GetSoundBackend()
 	local masterMuted = Engine.GetMasterMuted()
 	local masterLevel = Engine.GetMasterVolume()*100
 	local musicMuted = Engine.GetMusicMuted()
@@ -346,9 +346,17 @@ local function showSoundOptions()
 	local binaural_enabled = Engine.GetBinauralRendering()
 
 	local c
+	local new_backend_idx
+	local current_backend_idx
 
-	c,new_backend_id = combo(lui.AUDIO_BACKEND, current_backend_id - 1, available_backends)
-	if c then Engine.SetSoundBackend(new_backend_id + 1) end
+	for i, v in ipairs(available_backends) do
+		if v == current_backend then
+			current_backend_idx = i
+		end
+	end
+
+	c, new_backend_idx = combo(lui.AUDIO_BACKEND, current_backend_idx - 1, available_backends)
+	if c then Engine.SetSoundBackend(available_backends[new_backend_idx + 1]) end
 
 	c,masterMuted = checkbox(lui.MUTE.."##master", masterMuted)
 	if c then Engine.SetMasterMuted(masterMuted) end
--- a/data/pigui/modules/settings-window.lua
+++ b/data/pigui/modules/settings-window.lua
@@ -342,10 +342,12 @@ local function showSoundOptions()
 	local musicLevel = Engine.GetMusicVolume()*100
 	local effectsMuted = Engine.GetEffectsMuted()
 	local effectsLevel = Engine.GetEffectsVolume()*100
+	local binaural_supported = Engine.IsBinauralRenderingSupported()
+	local binaural_enabled = Engine.GetBinauralRendering()
 
 	local c
 
-	c,new_backend_id = combo("Audio Backend", current_backend_id - 1, available_backends)
+	c,new_backend_id = combo(lui.AUDIO_BACKEND, current_backend_id - 1, available_backends)
 	if c then Engine.SetSoundBackend(new_backend_id + 1) end
 
 	c,masterMuted = checkbox(lui.MUTE.."##master", masterMuted)
@@ -365,6 +367,11 @@ local function showSoundOptions()
 	ui.sameLine()
 	c,effectsLevel = slider(lui.EFFECTS, effectsLevel, 0, 100)
 	if c then Engine.SetEffectsVolume(effectsLevel/100) end
+
+	if binaural_supported then
+		c,binaural_enabled = checkbox(lui.BINAURAL_RENDERING, binaural_enabled, lui.BINAURAL_RENDERING_DESC)
+		if c then Engine.SetBinauralRendering(binaural_enabled) end
+	end
 end
 
 local function showLanguageOptions()
--- a/msvc-defaults.cmake
+++ b/msvc-defaults.cmake
@@ -21,3 +21,11 @@ set(VORBISFILE_LIBRARIES ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${M
 
 set(SIGCPP_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/include/sigc++-2.0)
 set(SIGCPP_LIBRARIES optimized ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${MSVC_ARCH}/vs2019/sigc-vc140-2_0.lib debug ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${MSVC_ARCH}/vs2019/sigc-vc140-d-2_0.lib)
+
+add_library(OpenAL SHARED IMPORTED)
+add_library(OpenAL::OpenAL ALIAS OpenAL)
+set_target_properties(OpenAL PROPERTIES
+    IMPORTED_LOCATION ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/bin/${MSVC_ARCH}/vs2019/OpenAL32.dll
+    IMPORTED_IMPLIB ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${MSVC_ARCH}/vs2019/libOpenAL32.dll.a
+)
+set(OpenAL_FOUND ON)
--- a/msvc-defaults.cmake
+++ b/msvc-defaults.cmake
@@ -28,4 +28,5 @@ set_target_properties(OpenAL PROPERTIES
     IMPORTED_LOCATION ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/bin/${MSVC_ARCH}/vs2019/OpenAL32.dll
     IMPORTED_IMPLIB ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${MSVC_ARCH}/vs2019/libOpenAL32.dll.a
 )
+target_include_directories(OpenAL INTERFACE ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/include)
 set(OpenAL_FOUND ON)
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -538,6 +538,7 @@ void StartupScreen::Start()
 		Sound::SetMasterVolume(Pi::config->Float("MasterVolume"));
 		Sound::SetSfxVolume(Pi::config->Float("SfxVolume"));
 		Pi::GetMusicPlayer().SetVolume(Pi::config->Float("MusicVolume"));
+		Sound::EnableBinaural(Pi::config->Int("BinauralRendering"));
 
 		Sound::Pause(0);
 		if (Pi::config->Int("MasterMuted")) Sound::Pause(1);
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -534,7 +534,7 @@ void StartupScreen::Start()
 		if (Pi::GetApp()->HeadlessMode() || Pi::config->Int("DisableSound"))
 			return;
 
-		Sound::Init(Pi::config->Int("AudioBackendId", Sound::AudioBackend_Default));
+		Sound::Init(Pi::config->String("AudioBackend"));
 		Pi::GetMusicPlayer().SetVolume(Pi::config->Float("MusicVolume"));
 		if (Pi::config->Int("MusicMuted")) Pi::GetMusicPlayer().SetEnabled(false);
 	});
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -534,7 +534,7 @@ void StartupScreen::Start()
 		if (Pi::GetApp()->HeadlessMode() || Pi::config->Int("DisableSound"))
 			return;
 
-		Sound::Init(Pi::config->Int("AudioBackendId", Sound::AudioBackend_OpenAL));
+		Sound::Init(Pi::config->Int("AudioBackendId", Sound::AudioBackend_Default));
 		Sound::SetMasterVolume(Pi::config->Float("MasterVolume"));
 		Sound::SetSfxVolume(Pi::config->Float("SfxVolume"));
 		Pi::GetMusicPlayer().SetVolume(Pi::config->Float("MusicVolume"));
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -534,7 +534,7 @@ void StartupScreen::Start()
 		if (Pi::GetApp()->HeadlessMode() || Pi::config->Int("DisableSound"))
 			return;
 
-		Sound::Init();
+		Sound::Init(Pi::config->Int("AudioBackendId", Sound::AudioBackend_OpenAL));
 		Sound::SetMasterVolume(Pi::config->Float("MasterVolume"));
 		Sound::SetSfxVolume(Pi::config->Float("SfxVolume"));
 		Pi::GetMusicPlayer().SetVolume(Pi::config->Float("MusicVolume"));
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -535,14 +535,7 @@ void StartupScreen::Start()
 			return;
 
 		Sound::Init(Pi::config->Int("AudioBackendId", Sound::AudioBackend_Default));
-		Sound::SetMasterVolume(Pi::config->Float("MasterVolume"));
-		Sound::SetSfxVolume(Pi::config->Float("SfxVolume"));
 		Pi::GetMusicPlayer().SetVolume(Pi::config->Float("MusicVolume"));
-		Sound::EnableBinaural(Pi::config->Int("BinauralRendering"));
-
-		Sound::Pause(0);
-		if (Pi::config->Int("MasterMuted")) Sound::Pause(1);
-		if (Pi::config->Int("SfxMuted")) Sound::SetSfxVolume(0.f);
 		if (Pi::config->Int("MusicMuted")) Pi::GetMusicPlayer().SetEnabled(false);
 	});
 
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -715,6 +715,7 @@ void MainMenu::Update(float deltaTime)
 
 	Pi::renderer->ClearDepthBuffer();
 	Pi::pigui->Render();
+	Sound::Update(deltaTime);
 
 	if (Pi::game) {
 		RequestEndLifecycle();
@@ -1091,6 +1092,7 @@ void GameLoop::Update(float deltaTime)
 	}
 
 	Pi::GetMusicPlayer().Update();
+	Sound::Update(deltaTime);
 
 	perfInfoDisplay->Update(deltaTime);
 	perfInfoDisplay->UpdateCounter(PiGui::PerfInfo::COUNTER_PHYS, phys_time);
--- a/src/lua/LuaEngine.cpp
+++ b/src/lua/LuaEngine.cpp
@@ -611,13 +611,11 @@ static int l_engine_get_available_sound_backends(lua_State *l)
 {
 	const Sound::BackendFlags backends = Sound::GetAvailableBackends();
 	lua_newtable(l);
-	if (backends & Sound::AudioBackend_SDL)
-	{
+	if (backends & Sound::AudioBackend_SDL) {
 		lua_pushstring(l, "SDL");
 		lua_rawseti(l, -2, 1);
 	}
-	if (backends & Sound::AudioBackend_OpenAL)
-	{
+	if (backends & Sound::AudioBackend_OpenAL) {
 		lua_pushstring(l, "OpenAL");
 		lua_rawseti(l, -2, 2);
 	}
@@ -742,6 +740,28 @@ static int l_engine_set_music_volume(lua_State *l)
 {
 	const float volume = Clamp(luaL_checknumber(l, 1), 0.0, 1.0);
 	set_music_volume(Pi::config->Int("MusicMuted") != 0, volume);
+	return 0;
+}
+
+static int l_engine_is_binaural_supported(lua_State *l)
+{
+	lua_pushboolean(l, Sound::IsBinauralSupported());
+	return 1;
+}
+
+static int l_engine_get_binaural_rendering(lua_State *l)
+{
+	lua_pushboolean(l, Pi::config->Int("BinauralRendering"));
+	return 1;
+}
+
+static int l_engine_set_binaural_rendering(lua_State *l)
+{
+	if (lua_isnone(l, 1))
+		return luaL_error(l, "SetBinauralRendering takes one boolean argument");
+	const bool enabled = lua_toboolean(l, 1);
+	Sound::EnableBinaural(enabled);
+	Pi::config->SetInt("BinauralRendering", enabled ? 1 : 0);
 	return 0;
 }
 
@@ -1094,6 +1114,9 @@ void LuaEngine::Register()
 		{ "SetMusicMuted", l_engine_set_music_muted },
 		{ "GetMusicVolume", l_engine_get_music_volume },
 		{ "SetMusicVolume", l_engine_set_music_volume },
+		{ "IsBinauralRenderingSupported", l_engine_is_binaural_supported },
+		{ "GetBinauralRendering", l_engine_get_binaural_rendering },
+		{ "SetBinauralRendering", l_engine_set_binaural_rendering },
 
 		{ "CanBrowseUserFolder", l_get_can_browse_user_folders },
 		{ "OpenBrowseUserFolder", l_browse_user_folders },
--- a/src/lua/LuaEngine.cpp
+++ b/src/lua/LuaEngine.cpp
@@ -628,10 +628,14 @@ static int l_engine_get_sound_backend_id(lua_State *l)
 	return 1;
 }
 
+static void set_music_volume(const bool muted, const float volume);
+
 static int l_engine_set_sound_backend(lua_State *l)
 {
 	const Sound::BackendId id = luaL_checkinteger(l, 1);
 	Sound::Init(id);
+	Pi::GetMusicPlayer().PlayAgain();
+	set_music_volume(Pi::config->Int("MusicMuted") != 0, Pi::config->Float("MusicVolume"));
 	Pi::config->SetInt("AudioBackendId", Sound::GetBackendId());
 	return 0;
 }
--- a/src/lua/LuaEngine.cpp
+++ b/src/lua/LuaEngine.cpp
@@ -609,22 +609,21 @@ static int l_engine_get_star_field_star_size_factor(lua_State *l)
 
 static int l_engine_get_available_sound_backends(lua_State *l)
 {
-	const Sound::BackendFlags backends = Sound::GetAvailableBackends();
+	const std::vector<std::string_view> backends = Sound::GetAvailableBackends();
 	lua_newtable(l);
-	if (backends & Sound::AudioBackend_SDL) {
-		lua_pushstring(l, "SDL");
-		lua_rawseti(l, -2, 1);
-	}
-	if (backends & Sound::AudioBackend_OpenAL) {
-		lua_pushstring(l, "OpenAL");
-		lua_rawseti(l, -2, 2);
+	int idx = 1;
+	for (std::string_view backend : backends)
+	{
+		lua_pushlstring(l, backend.data(), backend.size());
+		lua_rawseti(l, -2, idx++);
 	}
 	return 1;
 }
 
-static int l_engine_get_sound_backend_id(lua_State *l)
+static int l_engine_get_sound_backend(lua_State *l)
 {
-	lua_pushnumber(l, Sound::GetBackendId());
+	std::string_view backend = Sound::GetBackend();
+	lua_pushlstring(l, backend.data(), backend.size());
 	return 1;
 }
 
@@ -632,11 +631,10 @@ static void set_music_volume(const bool muted, const float volume);
 
 static int l_engine_set_sound_backend(lua_State *l)
 {
-	const Sound::BackendId id = luaL_checkinteger(l, 1);
-	Sound::Init(id);
+	const std::string backend = luaL_checkstring(l, 1);
+	Sound::Init(backend);
 	Pi::GetMusicPlayer().PlayAgain();
-	set_music_volume(Pi::config->Int("MusicMuted") != 0, Pi::config->Float("MusicVolume"));
-	Pi::config->SetInt("AudioBackendId", Sound::GetBackendId());
+	Pi::config->SetString("AudioBackend", backend);
 	return 0;
 }
 
@@ -1103,9 +1101,8 @@ void LuaEngine::Register()
 		{ "GetStarFieldStarSizeFactor", l_engine_get_star_field_star_size_factor },
 
 		{ "GetAvailableSoundBackends", l_engine_get_available_sound_backends },
-		{ "GetSoundBackendId", l_engine_get_sound_backend_id },
+		{ "GetSoundBackend", l_engine_get_sound_backend },
 		{ "SetSoundBackend", l_engine_set_sound_backend },
-		{ "GetAvailableSoundBackends", l_engine_get_available_sound_backends },
 		{ "GetMasterMuted", l_engine_get_master_muted },
 		{ "SetMasterMuted", l_engine_set_master_muted },
 		{ "GetMasterVolume", l_engine_get_master_volume },
--- a/src/lua/LuaEngine.cpp
+++ b/src/lua/LuaEngine.cpp
@@ -607,6 +607,37 @@ static int l_engine_get_star_field_star_size_factor(lua_State *l)
 	return 1;
 }
 
+static int l_engine_get_available_sound_backends(lua_State *l)
+{
+	const Sound::BackendFlags backends = Sound::GetAvailableBackends();
+	lua_newtable(l);
+	if (backends & Sound::AudioBackend_SDL)
+	{
+		lua_pushstring(l, "SDL");
+		lua_rawseti(l, -2, 1);
+	}
+	if (backends & Sound::AudioBackend_OpenAL)
+	{
+		lua_pushstring(l, "OpenAL");
+		lua_rawseti(l, -2, 2);
+	}
+	return 1;
+}
+
+static int l_engine_get_sound_backend_id(lua_State *l)
+{
+	lua_pushnumber(l, Sound::GetBackendId());
+	return 1;
+}
+
+static int l_engine_set_sound_backend(lua_State *l)
+{
+	const Sound::BackendId id = luaL_checkinteger(l, 1);
+	Sound::Init(id);
+	Pi::config->SetInt("AudioBackendId", Sound::GetBackendId());
+	return 0;
+}
+
 static void set_master_volume(const bool muted, const float volume)
 {
 	Sound::Pause(muted || is_zero_exact(volume));
@@ -1047,6 +1078,10 @@ void LuaEngine::Register()
 		{ "SetStarFieldStarSizeFactor", l_engine_set_star_field_star_size_factor },
 		{ "GetStarFieldStarSizeFactor", l_engine_get_star_field_star_size_factor },
 
+		{ "GetAvailableSoundBackends", l_engine_get_available_sound_backends },
+		{ "GetSoundBackendId", l_engine_get_sound_backend_id },
+		{ "SetSoundBackend", l_engine_set_sound_backend },
+		{ "GetAvailableSoundBackends", l_engine_get_available_sound_backends },
 		{ "GetMasterMuted", l_engine_get_master_muted },
 		{ "SetMasterMuted", l_engine_set_master_muted },
 		{ "GetMasterVolume", l_engine_get_master_volume },
--- a/src/lua/LuaEngine.cpp
+++ b/src/lua/LuaEngine.cpp
@@ -627,8 +627,6 @@ static int l_engine_get_sound_backend(lua_State *l)
 	return 1;
 }
 
-static void set_music_volume(const bool muted, const float volume);
-
 static int l_engine_set_sound_backend(lua_State *l)
 {
 	const std::string backend = luaL_checkstring(l, 1);
--- a/src/sound/AlAudioBackend.cpp
+++ b/src/sound/AlAudioBackend.cpp
@@ -135,13 +135,14 @@ bool Sound::AlAudioBackend::EventSetVolume(eventid eid, const float vol_left, co
 
 void Sound::AlAudioBackend::Pause(int on)
 {
+	m_paused = on;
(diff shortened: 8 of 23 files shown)


mfep:
Errata:
- [x] Audio is muted after switching between backends. Fixed in https://github.com/pioneerspacesim/pioneer/pull/6188/commits/68154e4dcb0d9dd3b794c0044953d8989ab583d3
- [x] Music playback in the menu does not resume after switching backends. Fixed in https://github.com/pioneerspacesim/pioneer/pull/6188/commits/eff501a96f3f2f76d1ea38d6e806f457b1e5a17b
  - [x] This is still not working properly. Fixed in https://github.com/pioneerspacesim/pioneer/pull/6188/commits/454763d166c93dd747e82b162743e4e76175cf5e
- [x] It has occurred, that due to an unknown reason, OpenAL returned an error after calling `alGenSources`. This currently causes the program to crash. Do we need more forgiving error handling?
  - This was due to too many sources generated by far away events. Fixed in https://github.com/pioneerspacesim/pioneer/pull/6188/commits/1da68659781974337c71e7adb58909b7bb800adf
- [x] Directional thruster sound is not correct using the OpenAL backend: Fixed in https://github.com/pioneerspacesim/pioneer/pull/6188/commits/65e6136494b61ab0765e78a6ba3bafbf31574c9e
- [x] Verify HRTF deployment on Windows
  - Fixed in https://github.com/pioneerspacesim/pioneer/pull/6188/commits/ba51c048716c6603a0f74420678032d2c850c316 and verified
- [x] No i18n in Sound menu. Fixed in https://github.com/pioneerspacesim/pioneer/pull/6188/commits/8518d052b3db5d0413dc02b6b11c72c0da9a64c3

impaktor:
I'm glad to see you're still improving upon this.

I don't know what @sturnclaw would prefer, but I suspect if he hasn't started review, that all the fixes be squashed into its suitable commit, so reviewer doesn't review code that has been bug-fixed in a later commit.

mfep:
> I'm glad to see you're still improving upon this.
> 
> I don't know what @sturnclaw would prefer, but I suspect if he hasn't started review, that all the fixes be squashed into its suitable commit, so reviewer doesn't review code that has been bug-fixed in a later commit.

Do you mean to squash the whole PR into a single commit? Or to squash just the fixes into a single commit that improves upon the initial implementation?

Anyway, I fixed every issue I was aware of, so unless the reviewers and testers find outstanding problems, I don't intend to push any more changes. Therefore I can suggest just to review the current tip of the feature branch.

impaktor:
> Do you mean to squash the whole PR into a single commit? 

No, probably not. 

I mean: if commit B is a bugfix to commit A, B could be squashed into A, such that A is bug-free upon review. I haven't looked at your code, so I don't know if it's applicable in your case, but based on the activity I've seen it makes me suspect it enough to post the above.

It should not be squashed if review has already started though. Well, that's just my five nickles on the topic.

mfep:
> I mean: if commit B is a bugfix to commit A, B could be squashed into A, such that A is bug-free upon review. I haven't looked at your code, so I don't know if it's applicable in your case, but based on the activity I've seen it makes me suspect it enough to post the above.

Thanks, I get it now. You are right, the current branch history developed organically, and it is not the best way to review it commit-by-commit. I can turn it to something more sensible, just let me know.

sturnclaw:
> I can turn it to something more sensible, just let me know.

That would be appreciated, thank you! I'm planning on starting the review tomorrow (in 10-12hr).

mfep:
> That would be appreciated, thank you! I'm planning on starting the review tomorrow (in 10-12hr).

OK, I've done what I could, and now the commit chain is more sensible. Each commit should build successfully. They are not separated perfectly, but hopefully enough to aid the review process and to be suitable for inclusion in the history.

data/pigui/modules/settings-window.lua
```diff
@@ -334,15 +334,22 @@ end, function (_, drawPopupFn)
...
 	local musicMuted = Engine.GetMusicMuted()
 	local musicLevel = Engine.GetMusicVolume()*100
 	local effectsMuted = Engine.GetEffectsMuted()
 	local effectsLevel = Engine.GetEffectsVolume()*100
+	local binaural_supported = Engine.IsBinauralRenderingSupported()
+	local binaural_enabled = Engine.GetBinauralRendering()
 
 	local c
 
+	c,new_backend_id = combo(lui.AUDIO_BACKEND, current_backend_id - 1, available_backends)
```
sturnclaw: `new_backend_id` is not a declared local here and will be silently written to the "global scope" of the execution environment. It should be declared as a local variable similar to `c`.

Nitpick: always use a space after a variable separator (`,`) and around mathematical operators; the existing style of this function is wrong.

```suggestion
	local c, new_backend_id = combo(lui.AUDIO_BACKEND, current_backend_id - 1, available_backends)
```

msvc-defaults.cmake
```diff
@@ -21,3 +21,11 @@ set(VORBISFILE_LIBRARIES ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${M
...
 set(SIGCPP_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/include/sigc++-2.0)
 set(SIGCPP_LIBRARIES optimized ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${MSVC_ARCH}/vs2019/sigc-vc140-2_0.lib debug ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${MSVC_ARCH}/vs2019/sigc-vc140-d-2_0.lib)
+
+add_library(OpenAL SHARED IMPORTED)
+add_library(OpenAL::OpenAL ALIAS OpenAL)
+set_target_properties(OpenAL PROPERTIES
+    IMPORTED_LOCATION ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/bin/${MSVC_ARCH}/vs2019/OpenAL32.dll
+    IMPORTED_IMPLIB ${CMAKE_SOURCE_DIR}/../pioneer-thirdparty/win32/lib/${MSVC_ARCH}/vs2019/libOpenAL32.dll.a
+)
+set(OpenAL_FOUND ON)
```
sturnclaw, line 32: Question: no need to define INCLUDE_DIRS for OpenAL headers on MSVC?

src/Pi.cpp
```diff
@@ -534,14 +534,8 @@ void StartupScreen::Start()
 		if (Pi::GetApp()->HeadlessMode() || Pi::config->Int("DisableSound"))
 			return;
 
-		Sound::Init();
-		Sound::SetMasterVolume(Pi::config->Float("MasterVolume"));
-		Sound::SetSfxVolume(Pi::config->Float("SfxVolume"));
+		Sound::Init(Pi::config->Int("AudioBackendId", Sound::AudioBackend_Default));
```
sturnclaw: I would strongly prefer the config value to be a string key rather than an opaque integer. Most of the string parsing code can be delegated to `EnumStrings` by decorating the `Sound::AudioBackend` enumeration with an `// <enum ...>` comment and re-running `scan_enums.py`.

src/Pi.cpp
```diff
@@ -715,6 +709,7 @@ void MainMenu::Update(float deltaTime)
 
 	Pi::renderer->ClearDepthBuffer();
 	Pi::pigui->Render();
+	Sound::Update(deltaTime);
```
sturnclaw, line 712: Question: is it better for the main-thread tasks to be run at the absolute end of the frame (before swap/presentation) or at the start of the frame? I'm unsure of OpenAL's semantics on the matter.

src/lua/LuaEngine.cpp
```diff
@@ -724,6 +724,39 @@ static int l_engine_get_star_field_star_size_factor(lua_State *l)
...
+	const Sound::BackendFlags backends = Sound::GetAvailableBackends();
+	lua_newtable(l);
+	if (backends & Sound::AudioBackend_SDL) {
+		lua_pushstring(l, "SDL");
+		lua_rawseti(l, -2, 1);
+	}
+	if (backends & Sound::AudioBackend_OpenAL) {
+		lua_pushstring(l, "OpenAL");
+		lua_rawseti(l, -2, 2);
+	}
```
sturnclaw: Similarly, I'd prefer to use the EnumStrings facilities here rather than hardcode an enum->string mapping here.

src/sound/AlAudioBackend.cpp
```diff
@@ -0,0 +1,466 @@
...
+		Error("Could not make OpenAL context current");
+		throw std::exception();
+	}
+
+	CHECK_OPENAL_ERROR(alDistanceModel, AL_INVERSE_DISTANCE_CLAMPED);
+	CHECK_OPENAL_ERROR(alSpeedOfSound, 1000.f);
+
+	std::default_random_engine prng(std::random_device{}());
+	std::uniform_int_distribution<decltype(m_next_event_id)> dist;
+	m_next_event_id = dist(prng);
```
sturnclaw: Note: we have the `Random.h` facilities; I'd prefer to use them rather than the stdlib `<random>`.

src/sound/AlAudioBackend.cpp
```diff
@@ -0,0 +1,466 @@
...
+	it->second.SetGain(volume_left * m_sfxVolume, volume_right * m_sfxVolume);
+	CHECK_OPENAL_ERROR(alSourcePlay, it->second.GetSource());
+	return it->first;
+}
+
+void Sound::AlAudioBackend::BodyMakeNoise(const Body *b, std::string_view key, float vol)
+{
+	constexpr double distance_threshold = 3000;
+	auto pos = b->GetPositionRelTo(Pi::player);
+	pos = pos * Pi::player->GetOrient();
```
sturnclaw, line 177: Unrelated thought: the intrusion of gameplay code into the sound backend is less than optimal.

I'd prefer to have a specific "World" for sound objects containing only lightweight "Listeners" and "Emitters" (and obstacles/attenuators as needed) that the Sound code operates on, and move the responsibility for syncing the position/velocity of these sound "shadow" objects with the primary Space into a separate system that runs as part of the update step.

src/sound/AlAudioBackend.cpp
```diff
@@ -0,0 +1,466 @@
...
+
+void Sound::AlAudioBackend::SetMasterVolume(float vol)
+{
+	CHECK_OPENAL_ERROR(alListenerf, AL_GAIN, vol);
+}
+
+float Sound::AlAudioBackend::GetMasterVolume()
+{
+	float vol;
+	CHECK_OPENAL_ERROR(alGetListenerf, AL_GAIN, &vol);
```
sturnclaw: If this involves a sync point with the sound thread (i.e. mutex), I'd strongly prefer to retain the master volume value as a cached variable in the sound backend and return it directly rather than querying the OpenAL API.

Given we already follow that pattern for the SfxVolume API, I think it's worth implementing here as well.

src/sound/AlAudioBackend.cpp
```diff
@@ -0,0 +1,466 @@
...
+	ALint state;
+	CHECK_OPENAL_ERROR(alGetSourcei, source, AL_SOURCE_STATE, &state);
+	return state == AL_STOPPED;
+}
+
+void Sound::AlAudioBackend::SoundEvent::FillBuffer(int idx)
+{
+	constexpr int buffer_bytes = 44100 * 2;
+	char data[buffer_bytes]{};
+	int bitstream{};
```
sturnclaw: Nitpick: when using C++11 brace-initializers, always separate them with a space from the variable declaration. (Apply to other occurrences of this pattern).

```suggestion
	char data[buffer_bytes] {};
	int bitstream {};
```

src/sound/AlAudioBackend.cpp
```diff
@@ -0,0 +1,466 @@
...
+	}
+}
+
+void Sound::AlAudioBackend::SoundEvent::SetOp(Op new_op)
+{
+	op = new_op;
+	if (oggv) // streamed shouldn't use AL_LOOPING
+	{
+		return;
+	}
```
sturnclaw: Nitpick: this should follow the stylistic convention of the rest of the function, i.e.
```suggestion
	// streamed shouldn't use AL_LOOPING
	if (oggv) {
		return;
	}
```

src/sound/OggFileDataStream.cpp
```diff
@@ -0,0 +1,82 @@
+#include "OggFileDataStream.h"
+
+Sound::OggFileDataStream::OggFileDataStream() :
+	m_cursor(nullptr) {}
+
+Sound::OggFileDataStream::OggFileDataStream(const RefCountedPtr<FileSystem::FileData> &data) :
+	m_data(data),
+	m_cursor(data->GetData()) { assert(data); }
```
sturnclaw, line 13: Non-empty constructor bodies should always be split (incl. braces) onto a separate line.

src/sound/Sound.h
```diff
@@ -18,14 +17,26 @@ namespace Sound {
...
 	typedef uint32_t Op;
+	enum {
+		// Bitflags!
+		AudioBackend_SDL = 1 << 0,
+		AudioBackend_OpenAL = 1 << 1,
+#ifdef PI_BUILD_WITH_OPENAL
+		AudioBackend_Default = AudioBackend_OpenAL,
+#else
+		AudioBackend_Default = AudioBackend_SDL,
+#endif
```
sturnclaw: I would recommend `AudioBackend_Default` be a separate enumeration value and the value be resolved into a "real" backend inside of Sound.cpp.

sturnclaw (changes requested):
Initial batch of comments and change requests from the diff - thank you for your patience! I'll give this a "live test" and go over the final versions of Sound.cpp etc. for the next batch of review comments before the end of the week.

data/pigui/modules/settings-window.lua
mfep: Done as requested.

msvc-defaults.cmake
mfep: This include directory was added by the previous lines, so it worked, but yes, and fixed.

src/Pi.cpp
mfep: I came to the conclusion that representing the backends as an enum brings unnecessary complications, so I changed this to use strings only, which eliminates the need of this kind of conversion. See https://github.com/pioneerspacesim/pioneer/pull/6188/commits/a3d53b5cb0cab5515b11436ae7431c2a18ab2276

src/Pi.cpp
mfep: To my knowledge, it does not matter. The main point is that the object positions and velocities are updated already at this point. The audio processing is performed asynchronously by the OpenAL thread.

src/lua/LuaEngine.cpp
mfep: This is no longer needed, see other reply: https://github.com/pioneerspacesim/pioneer/pull/6188#discussion_r2232958347

src/sound/AlAudioBackend.cpp
mfep: Done as requested.

src/sound/AlAudioBackend.cpp
mfep: Done as requested.

src/sound/AlAudioBackend.cpp
mfep: Fixed.

src/sound/AlAudioBackend.cpp
mfep: Fixed.

src/sound/OggFileDataStream.cpp
mfep: Fixed.

src/sound/Sound.h
mfep: This is no longer needed, see other reply: https://github.com/pioneerspacesim/pioneer/pull/6188#discussion_r2232958347
Now an empty string represents automatic backend selection.

impaktor:
@mwerle do you have any dog to throw into this review? (Thinking you both have PRs waiting for review)

mwerle:
> @mwerle do you have any dog to throw into this review? (Thinking you both have PRs waiting for review)

I can take a look, but I'm not much of an audio person. I'll be back home in 2 days so will check it then if it hasn't been done.

COMPILING.txt
```diff
@@ -63,6 +63,7 @@ questions are welcome anytime.
     libassimp-dev >= 5.0.1
     libsdl2-dev
     libsdl2-image-dev
+    libopenal-dev
```
mwerle: My understanding is that this is optional (only used if `BUILD_WITH_OPENAL` is defined).

Should we separate this list into "required" and "optional"?

data/lang/ui-core/en.json
```diff
@@ -83,6 +87,14 @@
     "description": "",
     "message": "Backward acceleration"
   },
+  "BINAURAL_RENDERING": {
+    "description": "Label of the binaural rendering checkbox in the audio settings menu",
+    "message": "Binaural Rendering"
+  },
+  "BINAURAL_RENDERING_DESC": {
+    "description": "Tooltip of the binaural rendering checkbox in the audio settings menu",
+    "message": "When enabled, HRTF filters are applied to the audio output. Works best with headphones."
```
mwerle, line 104: What is "HRTF" (yes, I can google it, but someone playing the game won't necessarily want to do that when reading a tooltip)

- Head Related Transfer Function

So other games might call it "3D positional audio, works best with headphones." (unless those are separate technologies?) which might be easier to understand for non-audio people? Maybe I'm just out of touch and "Binaural" and/or "HRTF" are common enough terms for gamers these days.

data/pigui/modules/settings-window.lua
```diff
@@ -360,6 +375,11 @@ local function showSoundOptions()
 	ui.sameLine()
 	c,effectsLevel = slider(lui.EFFECTS, effectsLevel, 0, 100)
 	if c then Engine.SetEffectsVolume(effectsLevel/100) end
+
+	if binaural_supported then
+		c,binaural_enabled = checkbox(lui.BINAURAL_RENDERING, binaural_enabled, lui.BINAURAL_RENDERING_DESC)
+		if c then Engine.SetBinauralRendering(binaural_enabled) end
+	end
```
mwerle, line 382: Suggestion: it might be better to always show this checkbox, but have it disabled if binaural is not supported.

Otherwise it might cause weird display/layout issues if/when more options are added to the dialog?

mwerle (commented):
I started a review and left a couple of quick comments/questions. Overall LGTM, but I haven't looked at the code in detail yet.

A quick playtest:
- after recompiling, "OpenAL" was the default. Is this intended or should SDL remain the default until the player changes it? Also, is it possible/desired to compile without SDL audio support?
- I had set Master to "mute" and this setting was not recognised on first startup - I had to toggle it off and on again.
- Effects mute is still bugged - it does not mute currently playing effects. Specifically the station effect is a very long-running effect which is where this is noticeable. (This might be out-of-scope for this PR)


Re mute: actually, this is never honoured when switching from SDL to OpenAL. Switching back to SDL it seems to be honoured.

Edit: Re mute (again): OpenAL backend does not honour the "mute" flag on game restart either.

src/Pi.cpp
```diff
@@ -534,14 +534,8 @@ void StartupScreen::Start()
...
-		Sound::Init();
-		Sound::SetMasterVolume(Pi::config->Float("MasterVolume"));
-		Sound::SetSfxVolume(Pi::config->Float("SfxVolume"));
+		Sound::Init(Pi::config->String("AudioBackend"));
 		Pi::GetMusicPlayer().SetVolume(Pi::config->Float("MusicVolume"));
-
-		Sound::Pause(0);
-		if (Pi::config->Int("MasterMuted")) Sound::Pause(1);
-		if (Pi::config->Int("SfxMuted")) Sound::SetSfxVolume(0.f);
 		if (Pi::config->Int("MusicMuted")) Pi::GetMusicPlayer().SetEnabled(false);
```
mwerle, line 539: With all the volume settings having been removed from the game startup, where are they applied now?

The reason I'm asking is that at least the "mute" (master) setting is not applied with the OpenAL backend - either when the game is restarted, or when the audio backend is switched from SDL to OpenAL.

I haven't tested the various volume settings.

src/sound/Sound.cpp
```diff
@@ -6,147 +6,51 @@
...
+#ifdef PI_BUILD_WITH_OPENAL
+#include "AlAudioBackend.h"
+#endif
+#include "AudioBackend.h"
 #include "Body.h"
 #include "FileSystem.h"
 #include "JobQueue.h"
 #include "Pi.h"
 #include "Player.h"
+#include "SdlAudioBackend.h"
```
mwerle, line 21: Should there be a compile-time option to NOT build with SDL audio as well?

src/sound/Sound.h
```diff
@@ -46,11 +45,11 @@ namespace Sound {
 		uint32_t eid;
 	};
 
-	bool Init(bool automaticallyOpenDevice = true);
-	bool InitDevice(std::string &name);
+	std::vector<std::string_view> GetAvailableBackends();
+	std::string_view GetBackend();
+
+	bool Init(std::string_view backend);
```
mwerle, line 51: Is it really the best thing to be passing strings around and comparing them instead of enums/instance pointers? Depending on the backend the string might need localisation before display, and you have them explicitly defined inside "Sound.cpp" so the current implementation doesn't allow for a runtime plugin style architecture anyway.

src/sound/Sound.h
```diff
@@ -46,11 +45,11 @@ namespace Sound {
...
 	void Uninit();
-	std::vector<std::string> &GetAudioDevices();
-	void UpdateAudioDevices();
```
mwerle, line 53: So it looks like these were implemented, but, at least in recent versions of Pioneer, not exposed. At least I can't find any other references to these functions in current `master`.

But is this a regression? As in, do we want people to be able to choose the audio device to which Pioneer outputs its audio? This could be useful for streaming or simply where people want the game audio on a different device to the default one. My gaming PC has several audio output devices.. the speakers, headphones, VR headset, and a HDMI monitor. I've sometimes found it useful to default certain applications to certain output devices.

src/sound/Sound.cpp
```diff
@@ -588,125 +195,107 @@ namespace Sound {
...
-
-		// If we're going to manually pick a device later, don't open a default one now.
-		if (!automaticallyOpenDevice) {
-			return true;
+		if (backend.empty()) {
+#ifdef PI_BUILD_WITH_OPENAL
+			backend = OpenALBackendName;
+#else
+			backend = SDLBackendName;
+#endif
```
mwerle: Do we want to default to OpenAL or SDL?

Usually you don't want to change a players' experience automatically..

src/sound/Sound.cpp
```diff
@@ -588,125 +195,107 @@ namespace Sound {
...
-			Output("Could not open audio device: %s\n", SDL_GetError());
+		try {
+			if (backend == SDLBackendName) {
+				m_backend = new SdlAudioBackend();
+			}
+#ifdef PI_BUILD_WITH_OPENAL
+			else if (backend == OpenALBackendName) {
+				m_backend = new AlAudioBackend();
+			}
+#endif
```
mwerle, line 252: If backends are `enum` instead of strings we could do a `switch` here..

Even better if they would be instance/proxy pointers, then you can just directly invoke the function and shold be able to keep the `Sound` class generic even if we were to add new backends in the future.

mwerle (commented):
Added a few more comments (not sure they all make sense).

src/sound/Sound.h
sturnclaw: Chiming in here - we absolutely do want to be able to control the output audio device... some of the time. Defaulting to the OS's selection is a good idea when the OS can remember which audio device the user wants for desktop audio, but not all OS's natively allow the user to assign different applications to different devices easily.

TL;DR: we need both a "Default" audio device and the ability to explicitly bind the game to a specific audio device.

src/sound/Sound.cpp
sturnclaw: If this PR makes it through the review process (which is likely), I would strongly prefer to ditch the raw SDL audio backend entirely. We do not do anything nearly sophisticated enough to *need* our own hand-written audio mixer.

COMPILING.txt
mfep: Ok, adding a comment indicating that this is an optional dependency.

Note: Logic around `BUILD_WITH_OPENAL` is a bit more involved: it defaults to `ON`, if the dev library is found on the system. So one way of enabling it is just installing the OpenAL Soft development libraries on the system.

data/lang/ui-core/en.json
mfep: > "3D positional audio, works best with headphones." 

This is a bit misleading, we have 3D positional audio without binaural. While HRTF may not be common knowledge among players, perhaps it makes sense to drop it, in case someone was interested what exactly is going on? The tooltip already gives a hint when this option should be enabled (when using headphones).

data/pigui/modules/settings-window.lua
mfep: I agree, but that functionality is not yet exposed to Lua (I searched for `ImGui::BeginDisabled`), so I find this out of scope. If deemed necessary, I can open a follow-up PR exposing this functionality and applying it here.

src/Pi.cpp
mfep: Now the values are loaded in `Sound::Init`. This is required because every time the sound backend is instantiated, the appropriate settings must be loaded again. This can happen during any time of the gameplay, not only at initialization.

I'll debug the reported master mute problem.

src/sound/Sound.h
mfep: I might misunderstand what you mean by "instance pointers", but if you mean to pass around `AudioBackend*`, I have quite some arguments against that. Maybe the strongest is that in this solution, the `Sound.h` facade is responsible for managing the audio backend instances, and exposing that to the outer word would defeat the purpose of this facade. Moreover, it would be impractical to keep more than one backends living at the same time, I never considered that due to resource (not only memory, but also audio system handles) management reasons.

It is arguable that an enum can be used to identify the audio backend types, in fact, a previous version of this PR went that way. However, that was much more complicated than using strings, and I don't think that the current backend names require i18n.

src/sound/Sound.cpp
mfep: This method is not called often, so I don't think string comparison is hurtful here.

About instance pointers, please read  https://github.com/pioneerspacesim/pioneer/pull/6188#discussion_r2274362482

About proxy pointers: that might work, but does it really worth the effort and the additional complexity?

src/sound/Sound.cpp
mfep: This also occurred to me, but I concluded that it makes sense to switch to the new backend by default. After all bugs are fixed (that is the ambition of this PR), it will be technically superior (e.g. the number of concurrent sources is much higher, no clicking when running out of voices).
Also, as @sturnclaw mentioned in the other comment, it would make sense to retire the old backend after we are sure that the new one works, and setting it as a default is one way of getting feedback (or the lack of it, which is even better).

data/lang/ui-core/en.json
mwerle: Sure, I just never heard of it before and, as a player, I am extremely unlikely to look up what HRTF means. So I think it's better to describe what this option _does_ rather than what it _is_.

But I don't have a big issue with this either way. We can always tweak this if people start asking what this means.

data/pigui/modules/settings-window.lua
mwerle: Hmm.. looks like an omission in ImGui : https://github.com/ocornut/imgui/issues/901 ; my apologies, I didn't know this wasn't possible.

So yeah, dropping this request. A bit out of scope for us to implement this ourselves.

src/sound/Sound.h
mwerle: Sure, the rest of the game engine seems to love passing strings around as well. The OCD in me very much dislikes that due to performance and handling reasons so I was wondering if there's a better way to implement this but in practicality it probably makes very little difference.

Sorry, I wasn't meaning to keep multiple fully instantiated and initialized audio backends around and pass references to those. I was thinking along the lines of a proxy/factory instance for each backend which would load/initialise the real backend when it is selected.

But sturnclaw mentioned that we'll likely remove the SDL audio backend once the OpenAL backend is fully up and running anyway. So there's probably very little reason ever to have multiple audio backends compiled into the binary and having the ability for the user select between them at runtime, it's more likely to be a compile-time option depending on platform.

src/sound/Sound.cpp
mwerle: Sure; I wasn't aware that this was replacement work so I agree with your reasoning here now. Closing.

src/sound/Sound.cpp
mwerle: Answered in the other comment.

Ultimately it's more likely that the codebase will only have a single backend compiled into it so all of this runtime choosing code might go away in the future anyway?

src/Pi.cpp
mwerle: Ok, thanks.

COMPILING.txt
mfep: Fixed.

mfep:
@mwerle Thanks for the report, I found the problem behind the master mute not working in the case of the OpenAL backend. This should be fixed now, as well as the SFX volume and mute should now have instantaneous effect in the case of the OpenAL backend (The SDL backend still exhibits the old behavior as it did previously). 

Edit: also rebased to the current master, to fix the Projectile crash.

mwerle:
> @mwerle Thanks for the report, I found the problem behind the master mute not working in the case of the OpenAL backend. This should be fixed now, as well as the SFX volume and mute should now have instantaneous effect in the case of the OpenAL backend (The SDL backend still exhibits the old behavior as it did previously).
> 
Awesome; confirmed fixed.

zonkmachine:
@mfep Sorry for the delay, I'm testing this now. I estimate it will take a week to do a proper job. I accidentally assigned this PR to me instead of just adding me as a reviewer. That was a bit rude, again, I'm sorry. 

At a quick test this works. The settings seem to operate a bit smoother than SDL. No glitches there. The only question I have after testing it for an hour is that the effect is added on all sounds. Should it really apply to the music as well?

PS. What documentation should I use? The one from [OpenAL](https://openal.org/documentation/)? Is that the same as OpenAL-Soft?
_Edit: Think I found it here: https://github.com/kcat/openal-soft/wiki/Programmer's-Guide_

zonkmachine:
@mfep I hear an issue that only seem to affect the thruster sound. On OpenAL they switch from being panned 100% one side. I haven't seen a more specific pattern to it. Just take off from Cydonia to another station with music off and compare thruster sound between SDL and OpenAL.
_Edit: Set 'Binaural Rendering' to off when testing_

mfep:
> PS. What documentation should I use? The one from [OpenAL](https://openal.org/documentation/)? Is that the same as OpenAL-Soft?

I was mostly using the official documentation and programming guide that you linked, and the somewhat hidden [OpenAL Soft extension docs](https://openal-soft.org/openal-extensions).

> The only question I have after testing it for an hour is that the effect is added on all sounds. Should it really apply to the music as well?

No actually. While the spatialization effect is not avoidable using the base OpenAL API, on a closer look I found an extension that makes direct output channels possible: [`AL_SOFT_direct_channels`](https://openal-soft.org/openal-extensions/SOFT_direct_channels.txt). As I understand, this allows the music audio file to be read and passed to the audio driver sample-by-sample without additional processing. This is added in de870e42de3c673ab2f243af6af9dd7c19808b35

PS: technically the space and station ambience falls in this category too, but since the effect is subtle AND fixing it requires sound API changes, I think that is for a future PR.

> I hear an issue that only seem to affect the thruster sound. On OpenAL they switch from being panned 100% one side.

You either have excellent ears or mine are particularly bad, since I had to record and then to resort to Audacity to visually confirm this. :sweat_smile: 

Indeed, since OpenAL has no concept of "stereo pan", the original API that makes it possible to pass a left and a right gain is simulated by computing the sound source's location on a circle with unit radius around the listener's horizontal plane. So in your case, the thruster is either 1m left from the listener or 1m right from the listener. In both cases, in OpenAL's simulation, sound is received by the opposite ear too, although on a much lower amplitude. I think that in the case of the ship thrusters this is closer to physical realism, so I'd like to keep this as it is currently.

zonkmachine:
> > PS. What documentation should I use? The one from [OpenAL](https://openal.org/documentation/)? Is that the same as OpenAL-Soft?
> 
> I was mostly using the official documentation and programming guide that you linked, and the somewhat hidden [OpenAL Soft extension docs](https://openal-soft.org/openal-extensions).

:+1: 

> > The only question I have after testing it for an hour is that the effect is added on all sounds. Should it really apply to the music as well?
> 
> No actually. While the spatialization effect is not avoidable using the base OpenAL API, on a closer look I found an extension that makes direct output channels possible: [`AL_SOFT_direct_channels`](https://openal-soft.org/openal-extensions/SOFT_direct_channels.txt). As I understand, this allows the music audio file to be read and passed to the audio driver sample-by-sample without additional processing. This is added in [de870e4](https://github.com/pioneerspacesim/pioneer/commit/de870e42de3c673ab2f243af6af9dd7c19808b35)

Fix works!

> PS: technically the space and station ambience falls in this category too, but since the effect is subtle AND fixing it requires sound API changes, I think that is for a future PR.

I agree.

> > I hear an issue that only seem to affect the thruster sound. On OpenAL they switch from being panned 100% one side.
> 
> You either have excellent ears or mine are particularly bad, since I had to record and then to resort to Audacity to visually confirm this. 😅

:wink: I do not have excellent ears but I'm used to troubleshoot sound generators. Both hardware and software. Noise is the way to do it so I'm used to listening to different sorts of noise for days at a time.

> Indeed, since OpenAL has no concept of "stereo pan", the original API that makes it possible to pass a left and a right gain is simulated by computing the sound source's location on a circle with unit radius around the listener's horizontal plane. So in your case, the thruster is either 1m left from the listener or 1m right from the listener. In both cases, in OpenAL's simulation, sound is received by the opposite ear too, although on a much lower amplitude. I think that in the case of the ship thrusters this is closer to physical realism, so I'd like to keep this as it is currently.

The '1m left .. 1m right' sounds like an extreme case of panning. I would like to poke these settings a bit. Can you point me to where in the code this is set?

Together with the left/right panning, when you reach orbit there is a sound effect playing. In the case of reaching Mars/Earth I believe it's the **/data/sounds/Space/G_Star.ogg**. Now that there is a thruster burning on one side it will appear to only play on the other side. This may not be incorrect at all but I haven't quite gotten used to it.
What decides to which side the thruster sample is going to be panned?

The thrusters/engines seem to be different depending on what camera you use. When you're outside of the ship the volume is bumped up noticeably. This is not specific to this PR but is already there in master. I'm not sure if this is a good thing. In the tests done previously in this PR I believe I was listening from outside of the ship (External and Sidereal views).

mfep:
@zonkmachine thank you for the patience! I've just returned from travel, and can be more responsive from now on.
 
> The '1m left .. 1m right' sounds like an extreme case of panning. I would like to poke these settings a bit. Can you point me to where in the code this is set?

Sure, the thruster sound is controlled in function `Ship::DoThrusterSounds` (Ship.cpp:964).

This extreme panning also predates the changes in my PR. If the thrust vector has no other component than X, and there is no angular thrust, the SDL audio also pans this extremely. If something, using the method with OpenAL I describe above softens this a bit, because in OpenAL's simulation, even in this case, some audio is played back in the opposite channel, even in the extreme case. Whereas in the SDL backend, it is truly just one side playing.

> Together with the left/right panning, when you reach orbit there is a sound effect playing. In the case of reaching Mars/Earth I believe it's the **/data/sounds/Space/G_Star.ogg**. Now that there is a thruster burning on one side it will appear to only play on the other side. This may not be incorrect at all but I haven't quite gotten used to it. What decides to which side the thruster sample is going to be panned?

The panning is controlled in the `Ship::DoThrusterSounds` function I mention above. The model is very simple: a single instance of the `Thruster_large` sampled is played and panned. The "mid" (common between L and R) component's level is controlled by the magnitude of the thrust vector, while the "side" (L or R) component's level depends only on the X coordinate of the thrust vector.

The ambience sound is not changed, so I assume that you only hear a difference, because it is quiet compared to the thruster sample.

> The thrusters/engines seem to be different depending on what camera you use. When you're outside of the ship the volume is bumped up noticeably. This is not specific to this PR but is already there in master. I'm not sure if this is a good thing. In the tests done previously in this PR I believe I was listening from outside of the ship (External and Sidereal views).

Yes, this is also controlled in the `DoThrusterSounds` function and as you mentioned, predates the changes in this PR.

PS, not in scope for this PR:

I think there is a lot of room for improvement in the thruster sound. It would add more realism, if the sound was simulated on a thruster engine basis (a sound source for each engine), instead of being based on the thrust vector. This will not be practical with the SDL backend due to the very limited sound concurrency, but nothing prevents it using OpenAL. This would also make it possible to use a different sample for different engines, which would help with defining a character for various ships.

impaktor:
What's the status of this PR? @mfep are you just waiting for final review / merge?

mfep:
> What's the status of this PR? @mfep are you just waiting for final review / merge?

@impaktor , that is correct, I have nothing left on my to-do list for this PR.

sturnclaw:
Will make reviewing this a priority then. I definitely agree - our thruster and general "ship" audio needs a major overhaul.

zonkmachine:
> Sure, the thruster sound is controlled in function `Ship::DoThrusterSounds` (Ship.cpp:964).

Thanks! Have played around with it some but haven't come up with something interesting.

Since I am the only one who seem to see any real difference here it shouldn't hold the PR back. I can switch back to SDL if I need to. I think it's better to merge and get the players feedback.

src/sound/Sound.cpp
```diff
@@ -588,125 +195,107 @@ namespace Sound {
...
-	bool Init(bool automaticallyOpenDevice)
+	std::string_view GetBackend()
 	{
-		PROFILE_SCOPED()
-		if (m_audioDevice) {
-			DestroyAllEvents();
-			return true;
+		if (dynamic_cast<SdlAudioBackend *>(m_backend) != nullptr) {
+			return SDLBackendName;
+		} else if (dynamic_cast<AlAudioBackend *>(m_backend) != nullptr) {
```
sturnclaw, line 217: Compilation error if BUILD_WITH_OPENAL is not set.

CMakeLists.txt
```diff
@@ -377,7 +387,8 @@ else()
 	message(WARNING, "Python 2 not found; enums will not be scanned.")
 endif()
 
-target_link_libraries(pioneer-lib PUBLIC lz4 fmt::fmt)
+target_link_libraries(pioneer-lib PUBLIC lz4 fmt::fmt $<$<BOOL:${BUILD_WITH_OPENAL}>:OpenAL::OpenAL>)
+target_compile_definitions(pioneer-lib PUBLIC $<$<BOOL:${BUILD_WITH_OPENAL}>:PI_BUILD_WITH_OPENAL>)
```
sturnclaw: Because this is applied as a public compilation definition, changing the option `BUILD_WITH_OPENAL` triggers a full rebuild of the codebase. Instead, I'd recommend you use the file `buildopts.h` (generated via CMAKE) and `#include` that in only the appropriate source files.

sturnclaw (changes requested):
I've given this branch a quick playtest now that my work tree is somewhat clear and my IRL duties are a little lighter now. I'll echo the feedback of other contributors that the thrust audio panning with the OpenAL backend is extremely intense.

With binaural rendering disabled, changing the primary direction of thrust from left to right (while other thrusters are active) under OpenAL completely mutes the audio from playing on the other channel (tested with a pair of headphones with good noise isolation), whereas SDL does a much better job of simulating the noise of thrust bouncing around "inside the cockpit" and plays it on the other channel at a reduced intensity. Even with binaural rendering enabled, OpenAL gives only a very slight amount of audio pickup on the opposing channel.

I know that's not necessarily the "fault" of OpenAL, but it is a very disconcerting effect to listen to, and is a manifest regression from the SDL backend. I'd appreciate it if that effect could at least be mitigated before merge, as I'm hesitant to bless OpenAL as the new default backend when the most listened-to sound effect in the game has a major regression.

Additionally (though this is likely also not OpenAL's fault), the binaural rendering is very undercut by the fact that the camera direction is not taken into account when calculating sound directionality; looking far to the left of a noise source in the internal cockpit view can play the sound in the "wrong" ear, when the sound source is between the centerline of the ship and the view direction. It appears that the player's ship is used to determine the "facing" orientation and not the view direction.

src/sound/AlAudioBackend.cpp
```diff
@@ -0,0 +1,475 @@
...
+	it->second.SetGain(volume_left * m_sfxVolume, volume_right * m_sfxVolume);
+	if (!m_paused) {
+		CHECK_OPENAL_ERROR(alSourcePlay, it->second.GetSource());
+	}
+	return it->first;
+}
+
+void Sound::AlAudioBackend::BodyMakeNoise(const Body *b, std::string_view key, float vol)
+{
+	constexpr double distance_threshold = 3000;
```
sturnclaw, line 174: I'd appreciate a comment here indicating that this is an area for improvement. Ideally loud sounds should not be pruned just because they originate 3km away, and we need a way to differentiate between "virtual" sources (those not currently assigned to an OpenAL source due to resource exhaustion) and "pruned" sources (those that are not played because they have zero contribution to the soundscape).

mfep:
@sturnclaw Thanks for the review again. About the remarks about the weird thruster sound: I have a growing feeling that we're watching a different movie - I hear no major difference between the two backends in terms of thruster sound. Would you mind creating a recording that shows where you find the difference disconcerting?

Here is a sample captured on my system:

https://github.com/user-attachments/assets/19eed991-8627-4eac-a195-8e54a1a629c7

src/sound/AlAudioBackend.cpp
mfep: Added comment.

CMakeLists.txt
mfep: Nice catch, fixed.

src/sound/Sound.cpp
mfep: Fixed.

impaktor:
@sturnclaw I can not hear any difference between the two in the video above. I suspect @mfep setup some how doesn't reproduce the differences you notice?

mfep:
I tried out the Windows build, and didn't register notable differences from the video I posted above. I had to fix some problems though, my previous commit regarding buildopts.h had some problems. Also I had to revert one change in https://github.com/pioneerspacesim/pioneer-thirdparty/pull/34.

For the video, I used a binary built natively on my Fedora Linux system, which ships with OpenAL-Soft version 1.24.2. @sturnclaw , @zonkmachine could you please share how exactly you tested this branch? Was it the Windows build? In that case, it should have used the binaries checked in as part of https://github.com/pioneerspacesim/pioneer-thirdparty/pull/34. Or was it Linux? Where did you get OpenAL from?

impaktor:
@mfep For what it's worth, sturnclaw is on Linux, we're all on Linux, except fluffyfreak and bszlrd, as far as I know.

mfep:
Just tested the CI build artifact "Linux 64-bit" on an laptop that runs MX Linux 23, which  has all dependencies to run the binary natively. This has OpenAL installed from the system repos, and it is on version 1.19.1.

This also could not reproduce anything weird about the OpenAL backend. Consider the following recording:

https://github.com/user-attachments/assets/7c09be74-6497-4efa-a231-4c3b51a46965

Without further assistance, I will not be able to continue with this.

src/sound/AlAudioBackend.h
```diff
@@ -0,0 +1,109 @@
+// Copyright © 2008-2025 Pioneer Developers. See AUTHORS.txt for details
```
impaktor: 2026 (needs doing in all newly added files)

sturnclaw:
@mfep I will record a video tomorrow or Tuesday, and attempt to dig into the root cause as well myself. Thank you for your great patience; I've been trying to make use of my limited Pioneering time to finish up outstanding feature PRs for the upcoming release.

sturnclaw:
> This also could not reproduce anything weird about the OpenAL backend. Consider the following recording:
> Video_2025-12-08_21-03-24.mp4
> 
> Without further assistance, I will not be able to continue with this.

Update: no recording on my end necessary, the issue is present in your recording.

@mfep if you listen to your recording around 0:13 with only the **left ear** of a pair of headphones with neutral left/right balance, you should hear the issue in question. The thruster audio is only present in the right channel in that recording.

Around 0:54, when you switch to SDL, the thruster audio for up thrust resumes playing in both ears with the SDL audio driver.

src/sound/AlAudioBackend.h
mfep: Done as requested.

mfep:
> Update: no recording on my end necessary, the issue is present in your recording.

Thank you for clarifying. I see now, that the issue is something I thought is expected behavior, and is due to the cruise mode being set.

There was indeed an issue, which roots in the fact that OpenAL has no concept of stereo panning (because it cannot know if the final audio render is going to be stereo or else). Therefore I needed to simulate the panning API interface used for the thrusters by moving the virtual speaker on a horizontal line around the listener. Due to the inaccuracy of the computation, a theoretical zero pan was not exactly zero numerically, and that caused a large difference in the result. **This should be fixed by the latest change**, which moves the virtual speaker on a unit circle around the listener instead.

Also, the branch has been rebased on master, and copyright dates are fixed. Let me know is something else comes up.

zonkmachine:
> **This should be fixed by the latest change**, which moves the virtual speaker on a unit circle around the listener instead.

Great, fixed!

src/sound/Sound.cpp
WickedSmoke: If you only want to support one backend at run-time then I'd prefer not having a virtual base class.  I was able to swap out the audio system with my [Faun library](https://github.com/WickedSmoke/pioneer/commit/3f41f1970d) by changing only a single file (Sound.cpp -> Sound_faun.cpp).

mfep: @WickedSmoke  In my view, the goal here is to give an easy way back to the SDL backend if something goes wrong and easy A/B testing between the backends - all this without recompiling the binary. Once we'd retire the original audio backend, the virtual functions can also be removed quite easily.

sturnclaw:
Due to a lack of time, I'm going to make the call to defer this PR to a mid-cycle release 1-2 months after our usual Feb. 3 "yearly" release.

mfep:
Rebased to current master.

mfep:
Rebased to current master.

src/lua/LuaEngine.cpp
```diff
@@ -607,6 +607,37 @@ static int l_engine_get_star_field_star_size_factor(lua_State *l)
...
+}
+
+static int l_engine_get_sound_backend(lua_State *l)
+{
+	std::string_view backend = Sound::GetBackend();
+	lua_pushlstring(l, backend.data(), backend.size());
+	return 1;
+}
+
+static void set_music_volume(const bool muted, const float volume);
```
sturnclaw: Errant function prototype here, leftover from refactoring?

src/sound/AlAudioBackend.cpp
sturnclaw: This has become unfixed in the latest diff.

sturnclaw (changes requested):
Only minor errors on a second read-through, I consider this PR pretty much ready for merge.

src/lua/LuaEngine.cpp
mfep: Fixed.

src/sound/AlAudioBackend.cpp
mfep: Just now I noticed that clang-format updates the file not to include the space. The setting that you might want to add to `.clang-format` is `SpaceBeforeCpp11BracedList: true`. I can fix this, but rather recommend to update the .clang-format spec and apply the formatter in a separate commit.

mfep: Anyway, I just applied the formatter to the sources in `src/sound`, because I see that the formatting is not enforced generally, and perhaps the reformatting will open up some discussions and will take a while.

impaktor:
Either we squash it all down to single commit at merge time, or there needs to be some fixup/squashing of the commits, like all review fixes, copyright, and development history of the PR, to what makes sense when someone views it in master.

WickedSmoke:
I'm going to reiterate my complaint from January about this implementation.

These changes introduce an extra layer of abstraction with the AudioBackend class.  Previously, the abstraction layer was Sound & Sound::Event.  When I reworked that the Sample struct was made private behind the Sound interface since the Faun backend completely hides how data is stored (which happens to be float arrays rather than uint16_t).

I can currently replace the audio backend by swapping out a single file (Sound.cpp).   If this is pulled I would have to redo work previously done in hiding Sample but on the AudioBackend level to get my stuff working again.

mfep:
@impaktor I suggest to just squash merge it. I doubt that the development history of this is going to be useful for the future maintainers.

@WickedSmoke I'm not sure if I understand the problem. After this PR, you will still be able to override Sound.cpp and entirely bypass the backend interface. The "public interface", that is defined in Sound.h changed only in minimal and additive ways in this PR, and you will be able to very quickly adapt your existing patch to it.

impaktor:
@WickedSmoke what's the purpose of your Faun library? What does it do that we currently lack?

WickedSmoke:
@impaktor I've used Pioneer to exercise my own audio library.

@mfep Yes, it looks like Sound.h hasn't changed much, so I will be able to avoid AudioBackend.

To facilitate that it would be good to support more backends on the project level by replacing the CMake BUILD_WITH_OPENAL with something like an `AUDIO=<backend>` variable and not pulling all `src/sound` files into the build.  Those would be conditional based upon AUDIO.

It looks like there's an `#include <map>` in Sound.h which may be unnecessary.

mfep:
> Yes, it looks like Sound.h hasn't changed much, so I will be able to avoid AudioBackend.

@WickedSmoke It's good to hear that this works for you!

> replacing the CMake BUILD_WITH_OPENAL with something like an AUDIO=<backend>

The point here is to be able to dynamically switch between the backends at runtime. The reason for this is to be able to quickly compare the different backends as a developer and/or resort to a working backend from an erroneous one as a player. So we need to compile multiple backends, therefore the switch that you suggest would not work because it operates at build time (CMake configuration time, to be precise). Instead, each backend (except the default one) should be switchable from CMake independently -- e.g. if Faust was added, it would require a CMake option `BUILD_WITH_FAUST`.

> not pulling all src/sound files into the build

Yeah, I'd prefer that too, but since the source files are collected via `GLOB`, it's difficult and not in scope for this PR.

> It looks like there's an #include <map> in Sound.h which may be unnecessary.

This changeset removes the `#include <map>` from Sound.h. Did I miss something?

WickedSmoke:
I was mistaken about the `<map>` - must have been looking at the wrong file when diffing against master.

There's no reason it should be difficult to support swapping out the Sound.cpp backend.  CMakeLists is already changing with this request so it's an opportunity to make it more future proof. 

Currently there are only three .cpp files in sound so replacing the globbing with individual files is a small change.  If Sound.cpp changes to `Sound_<backend>.cpp`  then a `SOUND=<backend>` variable can be used to pull in the correct one.  New backends which don't require all the AudioBackend stuff can then just work without changing CMake files.

To support the AudioBackend changes all the new files could be put into a "dynamic" subdirectory and sound/Sound.cpp becomes sound/Sound_dynamic.cpp.  The new .cpp files could be globbed inside a `if (SOUND=dynamic)` conditional.

impaktor:
> Currently there are only three .cpp files in sound so replacing the globbing with individual files is a small change

The PR that @sturnclaw approved was as it currently stands, before renaming files and moving things around. I'm not familiar with the code or "best practice" but I wouldn't want to see a reduction in code sanity to meet a very niche demand of one user who use their own audio library.

sturnclaw:
> I can currently replace the audio backend by swapping out a single file (Sound.cpp). If this is pulled I would have to redo work previously done in hiding Sample but on the AudioBackend level to get my stuff working again.

I will be somewhat explicit here - supporting a singular user's personal research fork is not a priority for us at all. While I certainly have no problems with you replacing the game's audio backend for your own use in your own copy of the repo, that fact will not be influencing our software design decisions.

The design of this PR adequately meets the design constraints it is intended to meet. We do not have any intention of replacing the sound backend after this PR is merged, unless it is with a significantly more professional and full-featured solution than OpenAL Soft.

> Currently there are only three .cpp files in sound so replacing the globbing with individual files is a small change. If Sound.cpp changes to Sound_<backend>.cpp then a SOUND=<backend> variable can be used to pull in the correct one. New backends which don't require all the AudioBackend stuff can then just work without changing CMake files.

This change will not be made; there is no need to future-proof the implementation against changes that we do not intend to make in the future. Adding build-time audio implementation selection is out-of-scope and only introduces additional untested and unused build permutations.

You are welcome to re-implement your Faun audio library backend using the AudioBackend interface. That interface will likely persist as the only nod towards multiple audio backend support. If you have suggestions towards making the sound sample data management code more generic and more efficient, I'd welcome them as an additional PR after this is merged.

WickedSmoke:
> You are welcome to re-implement your Faun audio library backend using the AudioBackend interface.

That option ignores my main point.  The Sound interface was where the implementation was isolated.  AudioBackend re-introduces backend specific stuff which would require re-hiding what I had already hidden behind Sound.  From all the replies it sounds like getting my changes re-done at that level will be an uphill battle, so it's probably best if I just steer clear of this project going forward.

> If you have suggestions towards making the sound sample data management code more generic and more efficient, I'd welcome them as an additional PR after this is merged.

One inefficiency is transforming the stream samples.  Ogg Vorbis decodes to floats, which gets converted to short via ov_read() since that's what all your backends are using, which gets re-converted to floats by the operating system.

The best approach to swapping out backends and keeping them isolated may be for each one to have it's own directory with a CMakeLists that defines the files, link libraries, and any configuration variables.  I'll just end with that suggestion should it ever be reconsidered and bid you all adieu.

sturnclaw (approved):

sturnclaw:
@WickedSmoke 

> That option ignores my main point. The Sound interface was where the implementation was isolated. AudioBackend re-introduces backend specific stuff which would require re-hiding what I had already hidden behind Sound. From all the replies it sounds like getting my changes re-done at that level will be an uphill battle, so it's probably best if I just steer clear of this project going forward.

You are of course always welcome to choose not to contribute to the Pioneer project. If you do not wish to maintain a patch to the CMake build files in your personal fork to support your personal audio implementation, I wish you the best in your other endeavors.

-----

@mfep I've given this a final review and test. Everything looks reasonably good in the code, and many eyes have passed over it. During playtesting, I noticed the SDL backend has slightly better audio fidelity with thruster sounds; especially with Binaural Rendering enabled thruster audio becomes somewhat muffled and washed out. It is not a concern for this PR, as I suspect it's a side effect of moving from stereo panning to proper 3d audio.

### After slightly more than a year, I'm happy to properly merge this PR and fully endorse it to all of our users!

I know you've had multiple ideas you've been working on while waiting for this to be properly reviewed and merged, and I'd like to suggest a few more areas of improvement:

- 3d audio feels very weird when looking around in internal ship view as the audio direction using headphones doesn't match the view direction.
- External cameras suffer the same problem; it may be best to add a toggle setting between using the camera as the audio listener or the ship as the audio listener.
- For both of those to work, thrusters would have to have proper positional audio. To avoid allocating e.g. 300 sources, I'd recommend we add specific tags during the ship import process to specify discrete audio emitter locations which groups of thrusters are associated with.

Issue #6136, opened 2025-05-22
Ship's autopilot sometimes does not look smooth


Mc-Pain:
### Steps to reproduce
1. Start a new game
2. Select "Mars High"
3. Set autopilot to "Fly to vicinity"
4. Accelerate to 1 km/s
**Observed:** ship speed jumps between `999 m/s` and `1.00 km/s` each frame
**Expected:** ship speed does not change

5. Wait until decelerating near starport
**Observed:** ship thrust fluctuates each frame
**Expected:** ship thrust is smooth and constant

_My pioneer version (and OS):_ current master `b66733d49`

wonderinglostsoul44:
<html>
<body>
<h3 data-start="531" data-end="561">Proposed comment for #6136</h3>
<blockquote data-start="563" data-end="734">
<p data-start="565" data-end="734">Thanks for the super‑clear repro steps – I could trigger exactly the same “999 ↔ 1.00 k” oscillation on today’s <code data-start="677" data-end="685">master</code> (b66733d). <span class="" data-state="closed"><span class="ms-1 inline-flex max-w-full items-center relative top-[-0.094rem] animate-[show_150ms_ease-in]"><a href="https://github.com/pioneerspacesim/pioneer/issues/6136" target="_blank" rel="noopener" alt="https://github.com/pioneerspacesim/pioneer/issues/6136" class="flex h-4.5 overflow-hidden rounded-xl px-2 text-[0.5625em] font-medium text-token-text-secondary! bg-[#F4F4F4]! dark:bg-[#303030]! transition-colors duration-150 ease-in-out"><span class="relative start-0 bottom-0 flex h-full w-full items-center"><span class="flex h-4 w-full items-center justify-between overflow-hidden"><span class="max-w-full grow truncate overflow-hidden text-center">github.com</span></span></span></a></span></span></p>
</blockquote>
<hr data-start="736" data-end="739">
<h4 data-start="741" data-end="798">1&nbsp;&nbsp;What’s really happening – two independent effects</h4>
<div class="_tableContainer_16hzy_1"><div tabindex="-1" class="_tableWrapper_16hzy_14 group flex w-fit flex-col-reverse">
Symptom on screen | True underlying state | Why it happens
-- | -- | --
[...]

PR #6189, opened 2025-07-02
make autopilot smooth

Mc-Pain:
This aims to close https://github.com/pioneerspacesim/pioneer/issues/6136

### DIFF
--- a/src/ship/Propulsion.cpp
+++ b/src/ship/Propulsion.cpp
@@ -149,11 +149,11 @@ vector3d Propulsion::ClampLinThrusterState(const vector3d &levels) const
 
 vector3d Propulsion::ClampLinThrust(const vector3d &levels) const
 {
-	vector3d clamped = levels;
+	vector3d clamped;
 
-	clamped.x = Clamp(clamped.x, -1.0, 1.0);
-	clamped.y = Clamp(clamped.y, -1.0, 1.0);
-	clamped.z = Clamp(clamped.z, -1.0, 1.0);
+	clamped.x = Clamp(levels.x, -1.0, 1.0);
+	clamped.y = Clamp(levels.y, -1.0, 1.0);
+	clamped.z = Clamp(levels.z, -1.0, 1.0);
 
 	return clamped;
 }


src/ship/Propulsion.cpp
```diff
@@ -327,16 +320,29 @@ bool Propulsion::AIMatchVel(const vector3d &vel, const vector3d &powerLimit)
...
+
+	// required internal velocity change (should extdv be the opposite direction?)
+	vector3d intdvGoal = diffvel - extdv;
+
+	// internal force and acceleration
+	vector3d intf = GetThrust(intdvGoal);
+	vector3d inta = intf / m_dBody->GetMass();
+	vector3d intdv = inta * frameTimestep;
+
+	vector3d intdvSmooth = intdv * 60.0;
```
sturnclaw: What is the purpose of the magic number `60` here? Is it related to the framerate or just an arbitrary scalar?

Mc-Pain: Just an arbitrary scalar, it smoothens the calculated force by 1 second of gameplay (but could be anything)

src/ship/Propulsion.cpp
```diff
@@ -147,6 +147,17 @@ vector3d Propulsion::ClampLinThrusterState(const vector3d &levels) const
 	return clamped;
 }
 
+vector3d Propulsion::ClampLinThrust(const vector3d &levels) const
+{
+	vector3d clamped = levels;
```
mwerle: Nitpick: This assignment is unnecessary; the below assignments can use "levels.x" etc instead.

The compiler should optimise this unnecessary assignment out.

src/ship/Propulsion.cpp
```diff
@@ -281,38 +301,31 @@ void Propulsion::AIAccelToModelRelativeVelocity(const vector3d &v)
 // sometimes endvel is too low to catch moving objects
 // worked around with half-accel hack in dynamicbody & pi.cpp
 
-double calc_ivel(double dist, double vel, double acc)
+// version for all-positive values
```
mwerle, line 304: Nitpick: should there be an assert/check to enforce that this function is only ever called with positive values for `dist`? I'm thinking future developers/refactoring.

src/ship/Propulsion.cpp
```diff
@@ -327,17 +340,29 @@ bool Propulsion::AIMatchVel(const vector3d &vel, const vector3d &powerLimit)
...
-	vector3d maxThrust = GetThrust(diffvel2);
-	vector3d maxFrameAccel = maxThrust * (Pi::game->GetTimeStep() / m_dBody->GetMass());
 	vector3d thrust(
-		Clamp(diffvel2.x / maxFrameAccel.x, -powerLimit.x, powerLimit.x),
-		Clamp(diffvel2.y / maxFrameAccel.y, -powerLimit.y, powerLimit.y),
-		Clamp(diffvel2.z / maxFrameAccel.z, -powerLimit.z, powerLimit.z));
-	SetLinThrusterState(thrust); // use clamping
+		Clamp(intdvGoal.x / intdv.x, -powerLimit.x, powerLimit.x),
+		Clamp(intdvGoal.y / intdv.y, -powerLimit.y, powerLimit.y),
+		Clamp(intdvGoal.z / intdv.z, -powerLimit.z, powerLimit.z));
```
mwerle, line 362: Nitpick: inconsistent indentation.

mwerle (commented):
Didn't dive into all the maths, just left a couple of minor nitpicks about the code.

Playtesting indicates it resolves the speed flickering at 1km/s after takeoff.

But I could not replicate the inconsistent engine firing during final approach to Mars High using `master`, so I don't see any difference there. Did you observe the inconsistent thruster firing while slowing down using the forward thrusters, or after the ship flips and uses its main thruster?

src/ship/Propulsion.cpp
Mc-Pain: clang-format does not see any

mwerle: My apologies; I just opened the file in an editor and checked and these lines are inside a constructor call, so yes, they are indented correctly.

The GitHup diff view was being gimpy.

zonkmachine:
Testing. There seem to be an issue when the numbers are around -1, 0, 1. When, for instance, approaching speed (top left 'The speed of approach towards the frame of reference') is around 0. You will see it as a 0 and a 1 on top of each other. You will see this when approaching a station and the ship does the last flip. I haven't seen this on master so this may be new (or reinforced) in this PR.

Mc-Pain:
> Testing. There seem to be an issue when the numbers are around -1, 0, 1. When, for instance, approaching speed (top left 'The speed of approach towards the frame of reference') is around 0. You will see it as a 0 and a 1 on top of each other. You will see this when approaching a station and the ship does the last flip. I haven't seen this on master so this may be new (or reinforced) in this PR.

I've seen this on 20250501, this PR is based on it

src/Orbit.cpp
```diff
@@ -173,6 +183,19 @@ vector3d Orbit::OrbitalPosAtTime(double t) const
 	return m_orient * vector3d(-cos_v * r, sin_v * r, 0);
 }
 
+vector3d Orbit::OrbitalDisplacementAtTime(double t, double dt) const
+{
+	if (is_zero_general(m_semiMajorAxis)) return m_positionForStaticBody;
+	double cos_v_t1, sin_v_t1, r1, M1 = MeanAnomalyAtTime(t);
+	double cos_v_t2, sin_v_t2, r2, M2 = MeanAnomalyDiff(M1, dt);
+	calc_position_from_mean_anomaly(M1, m_eccentricity, m_semiMajorAxis, cos_v_t1, sin_v_t1, &r1);
+	calc_position_from_mean_anomaly(M2, m_eccentricity, m_semiMajorAxis, cos_v_t2, sin_v_t2, &r2);
```
sturnclaw, line 192: This is doing a lot of extra work. You compute the delta between a previously-known positition (calculated at Frame.cpp line 490) by re-deriving the position, then computing the future position to take the delta relative to the current position. This function should ideally receive the previously-computed orbital position-at-current-time value, or be amended to return both the current position and the delta for the next position.

src/SpaceStationType.cpp
```diff
@@ -247,7 +247,7 @@ bool SpaceStationType::GetShipApproachWaypoints(const unsigned int port, DockSta
 	if (pPort) {
 		TMapBayIDMat::const_iterator stageDataIt = pPort->m_approach.find(stage);
 		if (stageDataIt != pPort->m_approach.end()) {
-			const matrix4x4f &mt = pPort->m_approach.at(stage);
+			const matrix4x4d &mt = pPort->m_approach.at(stage);
```
sturnclaw, line 250: If you're concerned about post-transform normality of the vector, this should instead be:

```c++
matrix4x4d mt = matrix4x4d(pPort->m_approach.at(stage));
mt.Renormalize();
...
```

src/SpaceStationType.cpp
```diff
@@ -119,7 +119,7 @@ void SpaceStationType::OnSetupComplete()
 		int bay, portId;
 		int minSize, maxSize;
 		char padname[8];
-		matrix4x4f locTransform = locIter->GetGlobalTransform();
+		matrix4x4d locTransform = matrix4x4d(locIter->GetGlobalTransform());
```
sturnclaw, line 122: Did you confirm this had any measurable impact on the accuracy or numerical stability of flight? Promotion to double of all of these variables has no impact on precision, given that the underlying values are stored and expressed in single-precision floating point. I'd ideally like to see this addressed by simply renormalizing the output matrix used to set the position and rotation of ships on the docking pad.

src/ship/Propulsion.cpp
```diff
@@ -341,7 +361,8 @@ bool Propulsion::AIChangeVelBy(const vector3d &diffvel, const vector3d &powerLim
 		Clamp(intdvGoal.y / intdv.y, -powerLimit.y, powerLimit.y),
 		Clamp(intdvGoal.z / intdv.z, -powerLimit.z, powerLimit.z));
 
-	SetLinThrusterState(thrust); // use clamping
+	SetLinThrust(thrust); // no clamping
```
sturnclaw, line 364: This change removes the per-axis-side acceleration limiter. Additionally, what is the purpose of this commit? It does not appear to resolve anything or improve numerical stability (and no test data is provided to prove any hypothesis of such).

src/DynamicBody.cpp
```diff
@@ -239,7 +239,7 @@ void DynamicBody::TimeStepUpdate(const float timeStep)
 		if (len > 1e-16) {
 			vector3d axis = m_angVel * (1.0 / len);
 			matrix3x3d r = matrix3x3d::Rotate(len * timeStep, axis);
-			SetOrient(r * GetOrient());
+			SetOrient(r * GetOrient().Normalized());
```
sturnclaw, line 243: This still accumulates error into all following calculations that depend on `Body::GetOrient()`, as the value set to `m_orient` is not normalized. This should be `SetOrient((r * GetOrient()).Normalized());`.

src/vector3.h
```diff
@@ -136,12 +132,12 @@ class alignas(sizeof(T)) vector3 {
 	}
 	vector3 Normalized() const
 	{
-		const T l = 1.0f / sqrt(x * x + y * y + z * z);
-		return vector3(x * l, y * l, z * l);
+		const T l = Length();
```
sturnclaw, line 135: On the basis that `1.0 / sqrt(...)` may compile into an `rsqrt` instruction with significantly lower precision than `sqrt`, I'm willing to accept the conversion to division rather than multiplication by the reciprocal.

However, please do not replace manually-inlined operations like this with calls to a separate function. These inlined operations are written such to avoid function call overhead when compiling in debug mode for math operations that are called millions of times per second.

src/Orbit.cpp
Mc-Pain: I can change this to
```
calc_position_from_mean_anomaly(M1, M2, m_eccentricity, m_semiMajorAxis, cos_v_t1, sin_v_t1, &r1, &r2);
```

src/SpaceStationType.cpp
Mc-Pain: As far as I remember, renormalizing 4x4f will be much less precise than renormalizing 4x4d.

It might not impact on the accuracy of flight but it *still* looks wrong normalizing in `float` space.

src/ship/Propulsion.cpp
Mc-Pain: I'll try to rebase and re-apply this with more clear explanation if still required.

src/vector3.h
Mc-Pain: This was pretty-not obvious. I'll state this in the code above.

src/SpaceStationType.cpp
sturnclaw: Please revert the change then. Without any measurable impact on numerical precision in a "real world" application, it is not worth the extra runtime cost to take values stored and manipulated in single-precision floating point (coming from the model system) and convert them to double-precision just because we don't like the thought of doing renormalization on single-precision datatypes.

src/ship/Propulsion.cpp
sturnclaw: I appreciate that. As currently implemented, this removes a feature that was intentionally added to the flight control system.

src/vector3.h
sturnclaw: Thank you! I agree, the reasoning behind the duplication is not obvious without documentation.

src/Orbit.cpp
sturnclaw: The problem isn't that the function is being called twice, it's that `pos1` is already known and computed beforehand in `Frame.cpp`. This function should *not* compute `pos1` at all, and should re-use the value from where it is originally computed.

sam52796:
@Mc-Pain - I believe I've addressed most of the issues. You can use _git apply_ to put this diff [patch.txt](https://github.com/user-attachments/files/32630285/patch.txt) onto your dev system and try the changes out.

Highlights are:

- Changed the HUD speed readout so that it rounds instead of floors, otherwise a target speed of 1.0 km/s can stay forever at 999 m/s, which looks a bit lame.
- Get orbital velocity directly from `OrbitalVelocityAtTime()` instead of calculating from the positions in `UpdateOrbitRails()`.
- Made changes as per sturnclaw's review requests.
- Removed now-unused code.

I did not look into the discontinuity in `calc_position_from_mean_anomaly()` - not sure if that's a blocker for merge or not.

Issue #6081, opened 2025-02-16
Follow orientation cannot be turned off if far from target (spin control problem, unknown reason)


Bodasey:
### Observed behaviour

The ship gets a spin who can not be controlled anymore

### Expected behaviour

The ship's spin can always be controlled

### Steps to reproduce

Load this game
[fixed_spin.txt](https://github.com/user-attachments/files/18815928/fixed_spin.txt)
or accept a rescue mission with a floating target.

Switch rotation dampening on/off has no influence, maybe the autopilot can not be switched off completely or the target is not properly set and confuses the autopilot?

__My pioneer version (and OS):__
actual master ddb008d on SuSE Tumbleweed

__My output.txt (required) and game save (optional, but recommended)__
no output generated for this

bszlrd:
Seems like there's no way to turn off Follow Orientation flight assist if you are too far from what it is set to. So it will try to match its rotation still. The two quarter circles on the Flight Assist icon means that it is engaged.

![Image](https://github.com/user-attachments/assets/9594ba35-9f6a-453a-9cd8-6c819190a4d4)

![Image](https://github.com/user-attachments/assets/97b4c1fc-585d-41ca-9330-edaae73771e1)

As a workaround, you could autopilot to any target, and immedietely turn off the autopilot. That should clear the flight assist.

Ideally the Follow Orientation should clear itself when you leave the frame, or you are too far from it to even be turned on.
Also the off button on the flight assist radial should be available when one is too far to engage, but it is already engaged.
Ping @Gliese852 , since you know the guts of the system the most.

I'll edit the issue title to be more specific.

Gliese852:
I'll try to fix this.

sturnclaw:
Partially fixed in my microfixes branch by enabling the ability to disable follow modes even if the player is unable to turn them *on* because they are too far away.

sturnclaw:
Fixed in #6116. There is currently no autodisable of follow modes over distance; that should be tackled in a separate issue/PR.

PR #6190, opened 2025-07-03
Improve the follow modes for too far ranges

Gliese852:
Probably a slightly more thorough solution to the problem of #6081. 
If earlier this problem was solved through the opportunity to manually turn off the follow mode, even when it is too far for it, now it is turned off automatically. Also, a different maximum range is assigned for different modes. For follow position mode, this is 1000 km, since it can be convenient for intercepting the ship. For the follow orientation mode, it is 10 km, since it makes little sense over long distances, and the ship most likely cannot correctly follow the rotating target.

Also along the way fixed some warnings in `PlayerShipController`.

### DIFF
--- a/src/ship/PlayerShipController.cpp
+++ b/src/ship/PlayerShipController.cpp
@@ -638,8 +638,8 @@ void PlayerShipController::PollControls(const float timeStep, int *mouseMotion, 
 	// only linear thrust ^ is allowed in the system map
 	if (Pi::GetView() == Pi::game->GetSystemView()) return;
 
-	m_ship->SetGunState(0, 0);
-	m_ship->SetGunState(1, 0);
+	auto gunManager = m_ship->GetComponent<GunManager>();
+	if (gunManager) gunManager->SetAllGroupsFiring(false);
 
 	if (Pi::input->MouseButtonState(SDL_BUTTON_RIGHT)) {
 		// use ship rotation relative to system, unchanged by frame transitions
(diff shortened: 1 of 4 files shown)


src/ship/PlayerShipController.cpp
```diff
@@ -687,9 +696,9 @@ void PlayerShipController::PollControls(const float timeStep, int *mouseMotion,
 		m_mouseActive = false;
 	}
 
-	if (InputBindings.primaryFire->IsActive() || (Pi::input->MouseButtonState(SDL_BUTTON_LEFT) && Pi::input->MouseButtonState(SDL_BUTTON_RIGHT))) {
-		//XXX worldview? madness, ask from ship instead
-		m_ship->SetGunState(Pi::game->GetWorldView()->GetActiveWeapon(), 1);
+	if (gunManager &&
+		(InputBindings.primaryFire->IsActive() || (Pi::input->MouseButtonState(SDL_BUTTON_LEFT) && Pi::input->MouseButtonState(SDL_BUTTON_RIGHT)))) {
+		gunManager->SetAllGroupsFiring(true);
```
sturnclaw: Note that this change will cause weapons in group 0 (by default, forward-facing weapons) to fire when the player is looking backwards and trying to fire weapons in group 1 (by default, rear-facing weapons).

I'm not sure we currently have ships with rear-facing weapons at this time, but I'd appreciate a comment here indicating that support for per-view weapon groups needs to be restored in the future.

Bodasey:
At least in my case, it has worked well.

src/ship/PlayerShipController.cpp
Gliese852: @sturnclaw Maybe then to do it right away?
```cpp
gunManager->SetGroupFiring(Pi::game->GetWorldView()->GetActiveWeapon(), true);
```

sturnclaw: That works, thank you. In the future I'd like to assign mounts to individual views or "seats" (e.g. crew-manned turrets) but for now we can use the WorldView.

Gliese852: Force-pushed this change.

sturnclaw:
Merged, thank you!

Issue #6161, opened 2025-06-14
Equipment rewrite: illogical categorizing of the equipment


Bodasey:
### Expected behaviour

Orbital, surface and target scanner should be found in the sensor group, whether they are mounted outside
or not

Cargo scoop should be found in hull mounts, same as fuel scoop

Miscellaneous should be used not so often?

### Observed behaviour

disorder?

__My pioneer version (and OS):__
current master

DamagedEngine:
I agree that the "Sensor" group should be renamed to "Radar". This is because this group can only fit the ship-to-ship radar and other things which could be called sensors (such as orbital, surface and target scanners) cannot be mounted here.

PR #6191, opened 2025-07-07
Feat/ship market show details

### LINKED ISSUE #6183: Equipment rewrite: The ship market relays on outdated data structures
### Observed behaviour

All ships have e.g. no scoop slots or missile mounts when presented in the ship market.

### Expected behaviour

A way to see the available slots for a ship from the ship market

Suggestion: add a "product description" button who opens an overview to the slots available.

current master

mwerle:
**DRAFT PR - discuss details**

----

**TODO:** 

- display a list of equipment installed on the ship.
- introspect the ships and display data accordingly instead of hard-coding the various slots.
- display the slot sizes?

**DISCUSS:**

- refactor equipment and slots? For example, scanners are in "utility" slots instead of "sensor" slots. The different scoops are in different slots. There are 3 (!) slot types which take missiles or missile equipment. etc.

----

Fixes #6183

The ShipMarket does not display all the statistic about the ships, specifically, the different slots available on each ship.

This PR adds (most of?) the different slots to the ship comparison display.

image

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -307,6 +307,10 @@
     "description": "graphics option in game settings menu ",
     "message": "Compress Textures"
   },
+  "COMPUTER_MOUNTS": {
+    "description": "How many slots or mounts exist on the ship for fitting COMPUTER equipment (e.g. autopilot, trade computer, etc)",
+    "message": "Computer mounts"
+  },
   "CONFIRM_PURCHASE": {
     "description": "",
     "message": "Confirm purchase"
@@ -1307,6 +1311,10 @@
     "description": "",
     "message": "Hull integrity"
   },
+  "HULL_MOUNTS": {
+    "description": "How many slots or mounts exist on the ship for fitting Hull mounted equipment (eg. Atmospheric Shielding)",
+    "message": "Hull mounts"
+  },
   "HULL_REPAIRED_BY_NAME_NOW_AT_N_PERCENT": {
     "description": "",
     "message": "Hull repaired by {name}, now at {repairPercent}%"
@@ -1675,9 +1683,17 @@
     "description": "For displaying commodity information",
     "message": "Minor Import"
   },
+  "MISSILE_BAYS": {
+    "description": "How many slots or mounts exist on the ship for Missile Bays.",
+    "message": "Missile bays"
+  },
   "MISSILE_MOUNTS": {
     "description": "",
     "message": "Missile mounts"
+  },
+  "MISSILE_PYLONS": {
+    "description": "How many slots or mounts exist on the ship for Missile Pylons.",
+    "message": "Missile pylons"
   },
   "MISSIONS": {
     "description": "",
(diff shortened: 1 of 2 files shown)


Bodasey:
A little tested and it looks fine so far, thank you, mwerle!

What i have not mentioned: The Capacity shown in the left half of the screen is still the capacity for the equipment, expected is the Cargo space. I Recommend to split the Ship market window in a 1:2 ratio and to show on the left side only symbol, name and list price of the ship, so there is more space for details on the right side, as cargo capacity may not be of interest for fighters. (maybe 3 columns for details? for wider/16:9 allover game resolution?)

For refactor equipment, see #6161

sturnclaw:
Regarding slots, I'd suggest it might be better to display the list of slots directly (using e.g. the display from the info view / outfitter), rather than trying to categorize them at such granularity.

> For example, scanners are in "utility" slots instead of "sensor" slots. The different scoops are in different slots. There are 3 (!) slot types which take missiles or missile equipment. etc.

Correct - this diversity is *entirely* intended. The sensor slot is reserved for the primary flight+fire-control radar/lidar sensors (which will be customizable by the player when sensor gameplay is implemented), while target/hypercloud scanners are a fusion of a standalone sensor aperture and a separate computer system.

The use of utility hardpoints is intended to be the primary "meaningful choice" with such equipment; you *can* mount a target scanner, you *can* mount a hypercloud analyzer, but if you do both you might not have enough hardpoints left to mount another piece of utility equipment you actually want to use.

mwerle:
> I'd suggest it might be better to display the list of slots directly (using e.g. the display from the info view / outfitter),

Sorry, I had a look, and I don't see what you mean. The info view shows practically nothing slot-related. The equipment outfitter view has various slot types grouped together in categories and is even more granular within those categories than what I've done so far.

If you could give a slightly more detailed example of what you envision here that would be much appreciated.

mwerle:
> What i have not mentioned: The Capacity shown in the left half of the screen is still the capacity for the equipment, expected is the Cargo space.

I added the cargo and kept the equipment capacity in this mockup. It depends on what people want to quickly see to distinguish ships.. I guess the ship-type can be inferred from the icon, but what are the other most important selection criteria?

image

sturnclaw:
> If you could give a slightly more detailed example of what you envision here that would be much appreciated.

I am proposing to use this widget (the slot list on the left) via a separate page/mode in the ship market to allow the player to inspect the slots of the ship as though they were attempting to outfit it. No attempt to reduce a ship's slots into a fixed set of named numbers will be successful in the long run; if we do not add a ship that breaks the mold, mods surely will.

Screenshot From 2025-09-15 02-23-50

It's not critical for this PR (the current approach is serviceable until we fully redesign the market screen), but if you want to comprehensively present the available equipment slots to the player then actually rendering the list of available slots is pretty much the only option to avoid omitting certain slot types.

Note that the slot list is currently quite entangled with the outfitter screen code, so it would take a bit of surgery to extract it into its own module.

sturnclaw:
> I guess the ship-type can be inferred from the icon, but what are the other most important selection criteria?

From a combat perspective, scanning the default missile bay equipment (I believe most missile bay mounts are marked up with a default equipment item) for individual missile mounts is more helpful/useful than displaying the number of missile bays (e.g. display 20 missile mounts rather than/in addition to 4 bays). Neither fully capture the nuance of missile *sizing*, but the layout of the screen is archaic enough that there's no way to cram that kind of detail into this list.

PR #6193, opened 2025-07-16
Add tooltip for pump down buttons on info view

impaktor:
Do we want this?
I've added it because I can never remember how these buttons work, which might be an indication there's a deeper UI issue. 

Anyway, if we want it I can clean this up and add it to localization. Looks like so now:

https://github.com/user-attachments/assets/3e8a63ea-ddde-41e8-968b-cf22ea6e6064

ping @bszlrd & @sturnclaw

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -2051,6 +2051,22 @@
     "description": "Drain the fuel tank",
     "message": "Pump down"
   },
+  "PUMP_DOWN_FROM_CARGO_TOOLTIP": {
+    "description": "Tooltip for Info screen",
+    "message": "Pump %i t from cargo"
+  },
+  "PUMP_DOWN_FROM_HYPERDRIVE_TOOLTIP": {
+    "description": "Tooltip for Info screen",
+    "message": "Pump %i t from hyperdrive to fuel tanks"
+  },
+  "PUMP_DOWN_TO_CARGO_TOOLTIP": {
+    "description": "Tooltip for Info screen",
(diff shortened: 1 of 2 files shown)


fluffyfreak:
More tooltips is generally more betterer 😉I think it's helpful

bszlrd:
I agree with Fluffy, the more tooltips, the better.

sturnclaw:
Grug say "many tooltip, more knowledge".

(The tooltip should probably say "Pump X t of fuel to/from cargo".)

impaktor:
OK, done.
Ready for review.

https://github.com/user-attachments/assets/eddfb28b-39c4-4bf8-b26f-72d7c8d88e19

Bodasey:
Any other open source user here who can not watch these videos? (file corrupt)

> I've added it because I can never remember how these buttons work, which might be an indication there's a deeper UI issue.

Yes, these arrows are contra-intuitive...some suggestions how it might be better? These arrows always indicate the direction of fuel flow:

Screenshot_20250719_164720
fuelflow2

I have added the signs +- as suggestion in the second draft to have the same look as in the commodity market, but then, the text was better "from Cargo bay".

craigo-:
I like the numbers on the buttons, I hope they remain.

Suggestion:

- Place a single "<" (or another existing arrow equivalent) before the word "Refuel" to indicate direction
- Place a single ">" (or another existing arrow equivalent) before the phrase "Pump down" to indicate direction
- Keep the numbers on the buttons, but remove the negative sign for the "Pump down" options

PR #6197, opened 2025-08-02
Do not suggest a donation if the donation must be more than 100k credits

### LINKED ISSUE #6074: Donate to Cranks - Please donate $0
Image

In the module **'Donate to Cranks'** When some level of notoriety has been reached, the suggested amount to donate is 0. This is by design but it's not a very helpful message and appears to the user as a bug.

### Observed behaviour

Message in campaign
 
> Your reputation would benefit by donating $0 to our cause, but any donation would help.

### Expected behaviour

Maybe just suggest a gradually higher value or a friendly message?

### Steps to reproduce

* Open the Debug window and bump up your reputation.
* In the BBS. Open up the **'Donate to Cranks'** campaign post.

As seen in Pioneer-2025-02-03

This part of the code looks a wee bit relevant

https://github.com/pioneerspacesim/pioneer/blob/ddb008daeebdc23716d0acdb516a7f38a33c98b8/data/modules/DonateToCranks/DonateToCranks.lua#L56-L58

robothauler:
If the player has reached the reputation “CREDIBLE”, he would have to donate more than 100000 credits to get to the next level.

Closes #6074

### DIFF
--- a/data/modules/DonateToCranks/DonateToCranks.lua
+++ b/data/modules/DonateToCranks/DonateToCranks.lua
@@ -27,14 +27,7 @@ local ads = {}
 local addReputation = function (money, character)
 	local char = character or Character.persistent.player
 	local curRep = char.reputation
-	local newRep
-
-	if curRep >= 1 then
-		local exp = math.log(money)/math.log(10) - (math.log(curRep)/math.log(2) - 1)
-		newRep = curRep + 2^exp
-	else
-		newRep = curRep + 2^(math.log(money)/math.log(10))
-	end
+	local newRep = curRep + math.sqrt(money * 0.004) + 0.00236 * money
 	char.reputation = newRep
 	return newRep
 end
@@ -44,7 +37,7 @@ local computeReputation = function (ad)
 	-- compute money to get to next level of reputation
 
 	local current_reputation = 	Character.persistent.player:GetReputationRating()
-	print("\n\nCURRENT:", current_reputation)
+	print("CURRENT REPUTATION:", current_reputation)
 
 	for i=0,5,1 do
 		local donate = math.floor(10^i)
--- a/data/modules/DonateToCranks/DonateToCranks.lua
+++ b/data/modules/DonateToCranks/DonateToCranks.lua
@@ -73,8 +73,13 @@ local onChat = function (form, ref, option)
 	end
 
 	if option == 0 then
-		form:SetMessage(string.interp(flavours[ad.n].message, ad.stringVariables) .. "\n\n" .. string.interp(
-			l.SALES_PITCH, {cash = Format.Money(computeReputation(ad), false)}))
+		local cash = computeReputation(ad)
+		if cash > 0 then
+			form:SetMessage(string.interp(flavours[ad.n].message, ad.stringVariables) .. "\n\n" .. string.interp(
+				l.SALES_PITCH, {cash = Format.Money(cash, false)}))
+		else
+			form:SetMessage(string.interp(flavours[ad.n].message, ad.stringVariables))
+		end
 
 	elseif PlayerState.GetMoney() < option then
 		form:SetMessage(l.YOU_DO_NOT_HAVE_ENOUGH_MONEY)


zonkmachine:
OK. This skips the sales pitch if it comes out as 0$. This is reasonable.
If I throw a whole lot of money in donations my reputation goes up. It seem to level out at 122/TRUSTWORTHY, when it should go up to 512/MASTER. After 122, I have thrown some 50 000 000$ at the charities and nothing happens. Maybe I misunderstand reputation, and you're not supposed to be able to buy yourself to the top like that? I think there is a round off error happening in addReputation causing it to level out at 122. If I create a 1000000$ donation button I can again increase the reputation slightly (I haven't tried to find if it levels out somewhere else with this button).
@impaktor 

https://github.com/pioneerspacesim/pioneer/blob/98d295fbc7867558de5cda12ed7c3c1421e8e3ed/data/modules/DonateToCranks/DonateToCranks.lua#L27-L40

I think this is a reasonable change but it doesn't fix the underlying issue, and the github issues shouldn't be closed.

impaktor:
There is nothing inherently wrong with being able to buy yourself reputation all the way to the top I think. 
If a new mechanic is suggested, or current equations tweaked, I'd like to see a graph of donation $ vs. reputation.

I think the idea with having the log is to reputation gain flatten out / cost more and more to buy a reputation score.

zonkmachine:
> I think the idea with having the log is to reputation gain flatten out / cost more and more to buy a reputation score.

And currently it seem to flatten out completely at too early in the career, at TRUSTWORTHY.

impaktor:
1. possible we want to change / tweak the `f_reputation(x_money)` function
2. what should be the behavior when any further donation has 0 impact on reputation?

zonkmachine:
> 1. possible we want to change / tweak the `f_reputation(x_money)` function

Probably. I tested to step up the reputation in the debug tool and see what the suggested donation would be per reputation. Here is the result for reputation 0 - 26 (Max 512).

|Reputation|Suggested donation|
|----------------|-----------------------------|
|0|100|
|1| 10|
|2| 10|
|3| 10|
|4| 10000|
|5| 1000|
|6| 1000|
|7| 100|
|8| 100000|
|9| 100000|
|10| 100000|
|11 |100000|
|12| 100000|
|13| 100000|
|14| 10000|
|15| 1000|
|16| 0|
|...|0|
|26| 0|


>     2. what should be the behavior when any further donation has 0 impact on reputation?

Not suggesting 0$ as proposed in this PR is a good start. I think maybe the reputation should still be impacted by donations at all time but just lesser with increasing reputation and the suggested donations increasing. 

_Edit: table, fixed value at reputation 4_

robothauler:
This is the current status when the reputation is <1:
reputation_lt1
I added 3 curves for reputation -3, -5 and -10.
In my opinion, that's fine.

This happens when the reputation is >= 1.
reputation_ge1
Both axes in the graph are logarithmic!
The graphic is a bit confusing, but I think you can see the most important points.

sturnclaw:
The graphs of the function as-is look like a very poor-quality gameplay experience. It seems when your rep is *exactly* -8, no donation can affect it... but a donation of 100 credits can take a -10 rep to -6? When you have 0 reputation or any power-of-two value, your reputation is apparently completely unable to be helped via donation.

In general, the absolute value of the player's current reputation should have a linear or slightly-exponential effect on reputation gain via donations (i.e. a $10 donation at rep 0 is enough for 1 reputation, while increasing one reputation level at rep 10 is $100). This should allow the total reputation gain for an arbitrary donation to be computed, while avoiding the pitfall that is some reputation levels causing your donation to simply go up in smoke.

robothauler:
It now takes 1,000 credits to go from Level NOBODY (0) to INEXPERIENCED (4) at the start of the game. 100 credits seemed too cheap to me. It now takes 100,000 credits to go from EXPERT (256) to MASTER (512).

robothauler:
Just for the sake of completeness.
donate1
donate2

PR #5474, opened 2023-01-16
Dynamic Station Economies

Web-eWorks:
...Feature freeze can't catch me if I type fast enough!

In all seriousness, this is a down-to-the-wire PR delayed by a round of sickness and a PC rebuild, but one that may prove to be one of the bigger features of the patch. This PR completely rebalances the economic system of Pioneer, providing a (perhaps too generous) 40-50% profit margin for savvy pilots who can sniff out the perfect pair of systems and stations to trade between - but punishes traders who get too complacent with seriously diminishing returns for running the same route too often.

The headlining feature of this PR is of course the "dynamic" economy for stations - we now have a proper supply-and-demand economic system where each station has its own set of commodities produced and consumed (and associated quantities for each), though the system is not quite as well-rounded as I'd like due to the paucity of good input data - I've shoehorned this all in to the existing three major economic archetypes, when I'd really like to split our commodities between 6-8 economic subtypes with more sane production and consumption rates... something for another PR.

Of note: the economic simulation is not "dynamic" in that prices change over time naturally. The existing behavior of station stocking is preserved in that stations will restock / increase demand back to an "equilibrium" value but once reaching that will remain in steady-state until the player takes an action. This is also a subject for another PR.

A secondary feature is an improved trade calculator tool (still WIP as of this writing) that provides extremely "cheaty" information on exactly which system/station has the best trade prices for a commodity (available for developers via alt-click) as well as providing more reasonable information about a station's economic flow for all commodities, similar to the existing system trade tool.

https://user-images.githubusercontent.com/4218491/212623201-1d02e669-af98-48d2-902d-ca1de5aa7eb2.mp4

The commodity market (the video shows an outdated version) now displays the station's local import/export disposition of a commodity, and if the player has a trade computer mounted additionally displays the interstellar import/export nature of a commodity.

I've also included a slightly-opinionated change to increase the easy/medium start's amount of available credits - as in-system missions don't pay much more than fuel cost and immediately running to an interstellar mission to make money isn't the greatest new-player introduction. On a meta level, this also balances removing the Sinonatrix from the easy start with a bit of extra cash instead :smile:

## TESTING REQUIRED

This branch was written hurriedly and not thoroughly tested, so I'd really appreciate everyone reading this PR description downloading a build from the Checks section and testing the economic behavior specifically - you can cheat in money with the Ctrl+I menu (as well as give yourself a better ship for bulk trading) and get setup for trading that way.

[...]

PR #6199, opened 2025-08-04
Add context-based station economic industries

### LINKED ISSUE #5889: Traders' stock/demand values don't get updated on sell
When I trade with traders from bulletin board, their "in stock" and "demand" values don't get updated when I buy or sell items. This happens both with fuel and illegal traders. As a result, I can sell more than their demand is, as opposed to the normal trade where you can't sell more than they need. Using 20240710 version for Windows

sturnclaw:
In short, this PR finishes the placeholder economic simulation that was introduced in #5474. It replaces the three-factor random roll which determined whether a commodity was imported or exported at a station with a set of context-dependent starport industries which consume and produce goods, adds a local "trade neighborhood" modelling short-hop NPC trading between stations around the same primary body, and models long-term oversaturation of commodity trades at markets.

**Please consider this PR an open call for content!**

I've provided a basic set of starter industries, but this is by no means a complete set and I would welcome contributions! New industries do not have to have a wholly unique set of commodities; additional variants of existing industries tailored for unique situations (e.g. high-temperature volcanic worlds might have a special metal smelting industry?) are strongly encouraged.

At current, the PR is "feature complete" however there are some aspects I'd like to explore adding before final merge:
- [ ] ~~Add a neighborhood context for condition tags (e.g. propagate the gas giant tag to starports on its moons).~~ Not used at this time. No current use, introduces hierarchical constraints.
- [ ] ~~Add govtype condition tags to system-wide context. (This would allow for government or economy-specific industries, e.g. Collective Industry for communist govtypes instead of Civil Industry.)~~ Not used at this time. Trivial to add.
- [x] Add a more intelligent system for selecting starport industries. ~~Current thought is a set of "upgrade trees" which are predicated on condition tags (e.g. build ore mine, then build refinery, then build heavy industry). Starport industries would be selected from the valid set of upgrade trees.~~ Implemented as per-industry build-next list, with condition tag and random-chance support.
- [x] Balancing existing commodity prices and quantities.
- [x] Adding additional commodities to round out production chains and create more rational supply/demand networks. (e.g. Rare Ores, Luxury Goods, *medical precursors that aren't food*, etc...)
- [ ] ~~Displaying industries present at a station in the map screens and on the station's comms page.~~ Not relevant at this time.
- [x] Write a full system description for developers in codedoc.
- [ ] Write a full system description for modders / content authors in devdocs.

The condition tags system should be somewhat self-documenting; it works by comparing named fields of the `SystemBody` object to a literal number or string. A few additional condition tags are directly injected into the context evaluated by industries so as to reduce the number of custom-authored condition definitions. An additional documentation pass is probably required on the conditions system, as Conditions.lua is somewhat lacking in documentation.


data/economy/conditions/basic.json
```diff
@@ -0,0 +1,78 @@
...
+		"conditions": [ "gravity <= 3.3" ]
+	},
+	"low_gravity": {
+		"context": "planet",
+		"conditions": [ "gravity > 3.3" ]
+	},
+	"norm_gravity": {
+		"context": "planet",
+		"conditions": [ "gravity > 6.0" ]
+	},
```
mwerle, line 17: Without reading code, how do these conditions prevent a planet with gravity "7.5" from being both "low_gravity" and "norm_gravity"?  

While this might be more concise, I think it would be more readable to make the conditions explicit, like was done for the atmosphere conditions below,  ie:

```
	"low_gravity": {
		"context": "planet",
		"conditions": [ "gravity > 3.3", "gravity <= 6.0" ]
	},
```

---

Secondly, what is the unit here? I would assume `g`, but given the definitions I'm guessing `m/s²`.  It would be good to document things like this.

---

Thirdly, is there a reason a definition for "high_gravity" and/or "extreme_gravity" are missing? Might have some interesting industries in those places.. 

---

Also, is there a way to `OR` conditions?

data/economy/conditions/basic.json
```diff
@@ -0,0 +1,78 @@
...
+		"conditions": [ "atmosDensity >= 0.1", "atmosDensity < 1.0" ]
+	},
+	"atmos_unbreathable": {
+		"context": "planet",
+		"conditions": [ "atmosDensity >= 1.0", "atmosOxidizing <= 0.7" ]
+	},
+	"atmos_breathable": {
+		"context": "planet",
+		"conditions": [ "atmosDensity >= 1.0", "atmosOxidizing > 0.7" ]
+	},
```
mwerle, line 33: Is the "atmosDensity" in units  of bar? If so, it is definitely possible to breathe atmospheres significantly less than a bar. The limiting factor for humans is a PPO2 of about 18%, which is a combination of oxygen content and pressure. There are, of course, other factors at play (at some point lack of pressure leads to pressure injuries), but there are human settlements at altitudes around 5km which have an air pressure of only around 1/2 bar (based on a quick internet search) - this actually results in a PPO2 of only around 11%, so people living at those altitudes need lengthy adaptations and/or genetic predisposition. 

Conversely Gemini astronauts breathed pure oxygen at only about 1/3 bar pressure with no ill effects - this was a PPO2 of around 35%.

A PPO2 above 160% can lead to oxygen toxicity, so a super-dense atmosphere even with a low oxygen content may not be breathable either.

(This is probably TMI for the context of Pioneer..)

data/economy/industries/agricultural.json
```diff
@@ -0,0 +1,62 @@
...
+			"fertilizer": 2,
+			"air_processors": 1,
+			"water": 4
+		},
+		"outputs": {
+			"fruit_and_veg": 3,
+			"grain": 3
+		},
+		"modifiers": {
+			"rare_microbial_life": [ "i:fertilizer-1", "o:*+1" ]
```
mwerle, line 16: While easy enough to figure out, is this all documented somewhere?

data/libs/Economy/Conditions.lua
```diff
@@ -0,0 +1,152 @@
...
+		elseif cond.op == Op.GE then
+			pass = val >= cond.value
+		elseif cond.op == Op.LE then
+			pass = val <= cond.value
+		end
+
+		-- Implicit AND between each condition
+		if not pass then
+			return false
+		end
```
mwerle, line 115: There is no scenario in which `OR` conditions might be useful?

data/libs/Economy/Industry.lua
```diff
@@ -0,0 +1,262 @@
...
+---@type table<string, integer>
+IndustryDef.outputs = {}
+
+-- Map of optional conditions to demand/production modifiers
+---@type table<string, Economy.IndustryModifier>
+IndustryDef.modifiers = {}
+
+-- Amount of workforce required to operate this industry
+IndustryDef.workforce = 1
+
```
mwerle: I didn't comment before, but what about using Slaves as workforce? Especially for industries such as remote (illegal?) mining or farming operations. Basically anything not hi-tech requiring lots of labour.

data/libs/Economy/Industry.lua
```diff
@@ -0,0 +1,262 @@
...
+local function parseModifierString(i, o, str)
+	local io, comm, effect, val = string.match(str, "([io]):(%g+)([%+%-])(%d+)")
+	local amt = effect == "-" and -tonumber(val) or tonumber(val)
+
+	if io == "i" then
+		i[comm] = (i[comm] or 0) + amt
+	else
+		o[comm] = (o[comm] or 0) + amt
+	end
+end
```
mwerle: I'd really love to see more descriptive variable names. It doesn't really cost anything these days.

`i` == `inputs`, `o` == `outputs`, `comm` == `commodity` etc.

Think about future maintainers reading this code in isolation instead of in a single PR.

data/libs/SpaceStation.lua
```diff
@@ -55,9 +55,9 @@ end
 local visited = {}
 local equipmentStock = {}
 
--- transientMarket is a cache for commodity markets initialized in unvisited stations
--- it's used to reduce the impact of looking up goods in stations that haven't been generated yet
-local transientMarket = utils.automagic()
+-- transientMarket is a cache for commodity markets associated with the curren stations
```
mwerle: Nitpick: spelling

```suggestion
-- transientMarket is a cache for commodity markets associated with the current stations
```

src/galaxy/SystemBody.h
```diff
@@ -118,7 +118,14 @@ class SystemBodyData {
 	fixed m_volatileLiquid; // 1.0 = 100% ocean cover (earth = 70%)
 	fixed m_volatileIces;	// 1.0 = 100% ice cover (earth = 3%)
 	fixed m_volatileGas;	// 1.225 = earth atmosphere density, kg/m^3
+	// TODO: atmosphere hydrogen-nitrogen ratio from mass, carbon-oxygen ratio from atmosCarbon
+	// (approximate the distribution of nitrogen vs. hydrogen by using the mass of the body; < 5 Me predominantly nitrogen, > 7 Me predominantly hydrogen)
+	// this differentiates nitrogen-rich and hydrogen rich atmospheres, as well as (hydro)carbon-rich from oxygen-rich atmospheres
+	// thus, an earth-like atmosphere would be atmosOxidizing = 0.3, atmosCarbon <= 0.05
```
mwerle, line 125: This does not match the "atmosBreathable" condition which requires "atmosOxidizing" to be ">0.7".

Line 32: 

https://github.com/pioneerspacesim/pioneer/pull/6199/files#diff-00965a8502cc3697fc7b3b329b8cf3df17cdd540c19daf13a8cff3c0bb3e0f47

data/libs/Economy.lua
```diff
@@ -3,24 +3,76 @@
...
+-- It models economies at the levels of individual starports, representing them
+-- as a collection of <Industries> which inform the supply and demand values
+-- of the local economy.
+--
+-- Any commodities which are not produced at a station are first sourced from
+-- the "local neighborhood"; commodities not produced locally are assumed to
+-- be imported from distant planets or via interstellar trade.
+--
+
+local AU = 149598000000
```
mwerle, line 34: Nitpick:

Common constants should really be somewhere centralised.. (utils?)

mwerle (commented):
Initial quick look, no playtesting yet.

Some questions, a few nitpicks. Overall LGTM (as far as I understand what's going on, which isn't very). Would appreciate some more detailed docs of how this is all supposed to work so reviews can compare against the design.

Bodasey:
Hard to contribute something without visualisation...but let's try some thoughts, hoping they are what you are looking for.

> * Adding additional commodities to round out production chains

We can make a commodity of every chemical element and base chemical compound (CH4, NH3, H2O, CO2, SiO2, FeO2, ...) depending on how detailed we want to model space mining and trade. E.g. "fertilizer" yet is a tailor made mixture of mainly Nitrogen, Phosphor, Kalium and other elements for different plants and environments.


Condition tags? Ok, let's list some natural resources on the bodies for the surface, crust, hydrosphere and atmosphere. 

> * Add a neighborhood context for condition tags (e.g. propagate the gas giant tag to starports on its moons).

Can be made, but even Hydrogen needs to be scooped and transported, and what about Venus and her CO2? => extended scoop equipment and even scoop missions?

Condition tags have to include type and intensity of energy sources (electric current from magnetic fields, wind, geothermal energy, solar radiation).

> * Add a more intelligent system for selecting starport industries.

Ok, I appreciate your model as intelligent enough if it generates a starport in the orbit around Venus who processes CO2 from her atmosphere with H2 from the outer solar system to Carbon ore and plastics with O2 as byproduct, or even better imports H2O from ice moons to get the H2 from fuel cells because you can directly use electric current from more than 2000W/m2 solar radiation which is far more efficient than the 1340W/m2 we get in Earth's orbit. O2 may then be shipped back to the gas giants to be used in fuel cells, which could be another additional commodity. And oh yes, fuel cells are a source of energy to be tagged if no other source is available, meaning a demand for H2 and O2.

(Airships would be a better solution to scoop gas for permanent industrial supply, shall we model them as orbiting objects?)

data/economy/conditions/basic.json
sturnclaw: 1. Some condition tags are intended as a "at least X" semantic. In this case, the `low_gravity` condition could/should be expressed as `min_low_gravity`, as the condition is intended to be used to select only bodies with at minimum 3.3m/s of surface gravity.
2. The unit is / should be specified in `SystemBody.h` and `data/meta/SystemBody.lua`. The explicit context that a condition definition is evaluated against is a SystemBody object, and the names here are the named fields of a SystemBody.
3. Condition tags were added during development to support industries added during development. The system is designed to be extremely extensible, so no attempt was made at defining a completely rigorous set of conditions "ahead of time".
4. There is no way to OR condition *requirements* together; condition tags represent "named facts" about a body, so all parts of a "fact" must be true for that "fact" to be true. The OR operator is instead expressed via the `conditions_any` field of an industry.

I'm not entirely opposed to adding an OR operator in there somewhere, but I'd need to see a convincing case that isn't expressible via the existing system to justify the added complexity of condition definitions, parsing, and evaluation.

data/economy/conditions/basic.json
sturnclaw: > Is the "atmosDensity" in units of bar?

No, Bar/Atm is a unit of pressure. Atmospheric density is measured in kg/m3, with earth having a 1.225kg/m3 density.

...yes, this is probably the wrong field to use to determine a breathable atmosphere. Suggestions for a minimum and maximum `surfacePressure` value for breathable atmosphere thresholds would be appreciated!

data/economy/industries/agricultural.json
sturnclaw: Not at the moment. During development, this was judged self-documenting (and not yet set in stone), so I have not yet written up documentation for modders.

data/libs/Economy/Conditions.lua
sturnclaw: See comment above regarding the `conditions_any` field.

data/libs/Economy/Industry.lua
sturnclaw: Workforce is currently completely unimplemented, as station populations are drawn completely randomly "out of a hat" at system generation time. I have nebulous plans for using the workforce value (aggregated across industries) to determine things like station immigration and goods consumption, or to limit industries from being present on smaller stations, but nothing fully developed at this time.

data/libs/Economy/Industry.lua
sturnclaw: I agree, both the conditions and industries files need to be cleaned up and documented.

src/galaxy/SystemBody.h
sturnclaw: This is correct. You'll notice that the current semantic model for atmospheres has Earth's atmosphere as `atmosOxidizing = 1.0`... which if taken literally, would mean the atmosphere is 100% O2. The current model has absolutely no way to differentiate between a carbon-dioxide atmosphere and a pure oxygen atmosphere, vs. a nitrogen, methane, or hydrogen atmosphere, other than magic number constants which are used to generate a flavor text string.

This comment describes a proposed physically-based system to be implemented at a later time, which expresses a sample earthlike atmosphere as ~30% O2, < 5% CO2, and the remainder N2 and trace gasses.

sturnclaw:
> We can make a commodity of every chemical element and base chemical compound

I appreciate the enthusiasm, but this is definitely not the direction we want to go. New commodities should be added because they fill a logical "gap" in a production chain or in commodity pricing, i.e. adding a "Chemicals" commodity produced at a refinery and consumed at a laboratory, or adding a high-value commodity to fill the gap between Robots and Precious Metals.

> Can be made, but even Hydrogen needs to be scooped and transported, and what about Venus and her CO2? => extended scoop equipment and even scoop missions?

Adding additional mission types, equipment, etc. is decidedly outside of the scope of this PR. It is the underlying assumption of this economic simulation that this sort of NPC trade is happening "in the background", and making these equilibrium activities visible to the player is a different topic.

> if it generates a starport in the orbit around Venus who processes CO2 from her atmosphere with H2 from the outer solar system to Carbon ore and plastics with O2 as byproduct, or even better imports H2O from ice moons to get the H2 from fuel cells because you can directly use electric current from more than 2000W/m2 solar radiation which is far more efficient than the 1340W/m2 we get in Earth's orbit. O2 may then be shipped back to the gas giants to be used in fuel cells, which could be another additional commodity. And oh yes, fuel cells are a source of energy to be tagged if no other source is available, meaning a demand for H2 and O2.

Condensing solid commodities from atmospheric gas is unlikely to be efficient enough to make a material difference to trade. Energy is not addressed by this economic simulation, as it is assumed that solar power and advanced microfusion using H2 provides more than enough power for industrial purposes.

For the most part, the complexity of simulation you describe above is not the focus of Pioneer. It can certainly be implied that all of this is happening in the background, carried out by the millions of people who inhabit star systems, but Pioneer is not a "tycoon" or economic-resource-management game and the player owning and managing a space station, etc. is a stated anti-goal of the project.

data/economy/conditions/basic.json
mwerle: 1. How are developers supposed to know whether a condition is an "at least" or a fully-specified condition without comments? Also, doesn't this mean that "min_low_gravity" means all bodies which have a gravity of greater than 3.3m/s²?  Sorry, but I am having difficulty understanding this one. If there are low-gravity industries, how can this condition be used to select bodies for them if even a 15m/s² body will match this condition?
2. Needing to cross-reference multiple files without context to understand something is not fantastic (imho). A comment would go a long way - even if the comment only refers to the source-of-truth in the other files. Please remember that not all contributors have dozens of years of experience with this codebase.
3. Ok, no worries. As long as it's understood that future industries may require adding condition tags..
4. Understood.

data/economy/conditions/basic.json
mwerle: As I said, for humans, "breathable" is a combination of surface pressure _and_ oxygen content (and absence of other toxic gases; inert gases don't matter). So might be a bit more complex..

Can the condition system generate derived units / have formulae as part of it? And/or call functions do so for it?

data/economy/industries/agricultural.json
mwerle: Anywhere to list  "todo" items so things like this won't be forgotten/pushed down the track ad infinitum? (Happy to help writing docs once I understand it sufficiently..)

data/libs/Economy/Industry.lua
mwerle: Ok. Hmm. Not sure I like having unimplemented stuff in the code base. We'll end up with half the codebase being placeholders which may or may not end up getting implemented (ok, being somewhat facetious here, obviously..)

Possibly add a comment saying so?

src/galaxy/SystemBody.h
mwerle: Ok, sorry, I don't understand the current mechanism so I'll close this comment.

data/economy/conditions/basic.json
sturnclaw: 1. Yes, `(min_)low_gravity` (intentionally) includes all bodies with at least 3.3m/s of surface gravity. If an industry should be limited to *only* 3.3-6.0m/s of surface gravity, the industry's `conditions` field could be something like `[ "low_gravity", "!norm_gravity" ]`. Alternatively, if the intent is to express a constraint that an industry can only exist in a low gravitational field or no gravity at all, the `conditions` field of an industry definition could instead require `[ "!norm_gravity" ]`.

    The philosophy here is that **conditions exist to serve industries**. If a constraint is not correctly expressing the semantic requirements of the industry definitions we have, it should be rewritten to either have different numeric values, a different name, or an entirely different condition added.

    As to developers understanding conditions - the list of conditions is two screens long, written in the most self-documenting way possible. Between *reading the list of conditions* and the documentation page I intend to author explaining the entire system for the use of modders/developers, I do not think there will be any confusion at all. Comments can certainly be added (just add a `"//": "..."` field in the JSON) but are intended to be unneeded as much as possible.

2. A comment where? I'm planning on a full documentation pass *most likely* in the [codedoc](https://codedoc.pioneerspacesim.net) but possibly also in the dev docs. Said documentation would fully explain the system, with links to the relevant documentation pages for allowed fields, etc.

data/economy/conditions/basic.json
sturnclaw: > Can the condition system generate derived units / have formulae as part of it? And/or call functions do so for it?

No, no, and no. The condition tags system is meant to be as simple as possible (JSON only, no dependencies, etc.) so that it can be moved into C++ and run at system generation time for microsecond-level performance impact. Any/all of the above is something that should be implemented as part of SystemBody instead, and the derived value stored/exposed as an attribute.

> As I said, for humans, "breathable" is a combination of surface pressure and oxygen content (and absence of other toxic gases; inert gases don't matter). So might be a bit more complex..

Simplifying the complex problem is fine. For the most part, condition tags will be invisible to the player, and there are a multitude of "hidden factors" which may influence *why* an atmosphere is breathable (terraforming efforts, generational genetic adaptations, cybernetic assistance, etc.) from a scientific/lore perspective.

Keep in mind that at current, a "breathable atmosphere" as the code considers it is an atmosphere with the magic number "atmosOxidizing" greater than a completely unspecified (and unscientific!) amount. Even when the model is improved, the player is unlikely to ever see the precise gas composition of an atmosphere.

If a "breathable atmosphere" is sufficiently multivariate or nonlinear such that it cannot be approximated as a two or three variable "box" on a graph, then I'd suggest it should be implemented as a computed quantity of a SystemBody and a new boolean (or scalar!) field `atmosBreathable` added for it instead.

data/economy/conditions/basic.json
mwerle: 1. Ok; I'm really sorry, but I find this set of conditions for gravity really confusing, especially when the conditions for atmospheres just below have specific bands. And needing to combine gravity conditions like that in an industry seems to me to also be unintuitive when the same isn't needed for atmospheres. I most certainly did not find it intuitive or self-documenting. But you can put it down to me being a numpty.
2. I was meaning a comment in the source. That's what people mostly read. But I'm sorry, I forgot this is `JSON` and doesn't natively support comments. One reason I somewhat dislike `JSON`. Having good docs (and reminding people to read them) is the next best thing. Closing.

data/economy/industries/agricultural.json
sturnclaw: Feel free to add a comment in this PR with a list of things to document - both you and I will be able to edit that list. I plan on merging documentation with this PR, so as to prevent it from being pushed down the track :)

data/libs/Economy/Industry.lua
sturnclaw: It will be present in the documentation pass, yes. I may strike it completely if I cannot find a compelling reason to use the workforce value; it may be simply out-of-scope for the depth of simulation that is appropriate for Pioneer.

data/economy/conditions/basic.json
mwerle: Ok, well out of scope for this PR :)

That said, I think it might be interesting to have fully-modelled (or at least better approximated) atmospheres at some point. I remember a discussion a while back talking about light scattering and realistic colours that would require/benefit from this as well.

And it could open exploration missions looking for new planets with specific atmospheric compositions - either for colonisation or specific industrial/mining applications.

sturnclaw:
Here's the ~~very short~~ skeleton of the high-level design of this system for review purposes:

1. Conditions are a data-driven way for mods (and the base game!) to express named semantics about individual bodies. They are meant to eventually be determined at system generation time in C++. Because they have access to a random number generator, they can be used by more than just the economic system; for example, conditions can be used as a way to designate planets with valuable flora/fauna (for scanning missions), rare ores (for @impaktor's mining machinery) or any other kind of "planetary feature" developers/modders can think of.

2.  Industries express the economic endeavors of a (logical) subregion of a starport. For surface ports, this is a number of buildings, for orbitals it's a chunk of the ring. Starports have a certain number of industry "slots" based on their population. Workforce is an abstract logarithmic quantity meant to express the amount of population required to operate an industry, but currently has no interaction with any of the systems in this PR.

3. Each industry expresses its production and consumption in a dimensionless logarithmic rate quantity called "flow". The rules for flow are simple, but intentionally inconsistent. An industry which produces a commodity with flow N can supply one other industry at that station with demand flow N, and two or more industries each with demand flow N-1. Two or more industries with supply flow N can supply a single industry with demand flow N+1, but can *never* satisfy an industry with demand flow N+2. The precise base of the logarithm used in flow is intentionally not defined, but varies from 2 to 10+.

    This "rule quirk" regarding industries satisfying N+1 demand is meant to provide a meaningful (if mostly invisible) effect to randomly generating multiple industries which produce or consume the same commodity at the same rate. Its concrete effect is that the rare station which consumes 6 flow of ore *twice* will have a higher sell price for ore than a station which only has one industry which consumes 6 flow of ore.

6. Flow is converted to a physical quantity via the supply/demand rates. The natural logarithm is used to determine how many discrete units of a commodity the station produces and consumes over an intentionally undefined period of time. Supply/demand rates determine the trade price of a commodity at a station, and are affected by the market's `history` value which records the player's interactions with the market.

7. All stations are assumed (currently) to be at an implicit equilibrium. This means that all commodity demand will be satisfied, regardless of how much of that demand is actually produced at the station. This is modeled via two mechanics: local and interstellar trade. Local trade allows any surplus flow (flow that meets or exceeds indigenous demand) from a nearby station to be consumed by the station with a deficit, and is imported at lower prices. Interstellar trade is the assumed source of the remaining deficit not satisfied by local trade, and is imported at significantly higher prices. Both local and interstellar trade are currently assumed to be happening in the background, and the responsibility for "visualizing" this process by instantiating NPCs is left to other modules like TradeShips.

Bodasey:
> Pioneer is not a "tycoon" or economic-resource-management game and the player owning and managing a space station, etc. is a stated anti-goal of the project.

As long as there are tradeable goods, you can play Pioneer in a "Transport Tycoon" or "Simutrans" mode...


> Condensing solid commodities from atmospheric gas is unlikely to be efficient enough to make a material difference to trade.

Maybe it is if you don't have an Earth nearby full of accumulated fossil carbon - or maybe it is if you have an Earth nearby with a mankind who finally has learnt the lesson how to handle them wisely...


Demanding Efficiency would, in the end, mean that spaceports must be generated by an economy simulation code based on natural resources. Even the starports on earth are mostly at the wrong place.


To make suggestions for what you want: What is the wished and upper limit target quantity of industries and tradeable commodities? How are byproducts handled, means can industries have more than one commodity as output, more than two commodities as output, as rubbish as commodity yet exists?

sturnclaw:
> To make suggestions for what you want: What is the wished and upper limit target quantity of industries and tradeable commodities? How are byproducts handled, means can industries have more than one commodity as output, more than two commodities as output, as rubbish as commodity yet exists?

The number of industry definitions has no hard limit. Avoiding too much granularity at this stage is desired (i.e. don't add a Mining Machinery Factory industry) as the industry system and number of discrete industries at a station are currently balanced around somewhat generic aggregations of multiple factories/companies/buildings into a logical "industry type".

As mentioned above, the number of commodities is something I'd like to avoid bloating, to keep each commodity feeling like it has an "identity" rather than disappearing into a sea of incredibly specific commodities. Additions for commodities should have a clear and defensible purpose.

Industries have any number of input and output commodities. Condition tags can be associated with modifiers to an industry's input and output commodities, which can increase and decrease commodity quantity and add new I/O commodities entirely.

data/libs/Economy.lua
```diff
@@ -38,395 +90,735 @@ Economy.TotalTradeFees = 2 * Economy.TradeFeeSplit
...
----@param commMarket table the current state of the commmodity market
-function Economy.UpdateCommodityPriceMod(sBody, key, commMarket)
-	local comm = Commodities[key]
-	if not comm then return 0 end
-
-	local maxStock = Economy.GetMaxStockForPrice(math.abs(comm.price))
-	local systemMod = sBody.system:GetCommodityBasePriceAlterations(key)
+--    sbody - SystemBody, the station for which to compute the equilibrium
+--    commodityId - string, the id of the commodity in question
+--
```
Gliese852, line 452: Need to add "supply" and "demand" arguments.

sturnclaw: Thanks!

data/libs/Economy/Industry.lua
sturnclaw: Workforce value has been struck completely; it makes sense in StarSector as a limiting mechanic for the player to build, but not in Pioneer where station population is determined before industries.

data/economy/industries/agricultural.json
sturnclaw: Modifiers are documented for developer use in `Industry.lua`. I plan on writing a documentation page on the JSON data format for modders in devdocs, but will not defer merge until doing so.

sturnclaw:
I am happy with this branch as-is. I've addressed all of the TODOs I had, improved the selection process for starport industries, tweaked balance on commodity prices and flows, and added a **new commodity**, Chemicals.

Chemicals are used as an intermediate step between grain/carbon ore and medicines, robots, and more. They occupy the price point Industrial Machinery had prior, with both industrial and farm machinery receiving a large boost in price to match their relatively small supply and demand quantities.

This branch attempts to remain a balance between commodity price and availability, with more widely produced commodities receiving lower prices. Additionally, the changes to commodity restock and long-term effects on supply and demand mean that higher-priced commodities are not necessarily going to be widely tradeable in large quantities (though the effectively infinite nature of inhabited space does limit the amount of impact any one station's market will have on trade viability).

I will not expect a rigorous re-review. I intend to document the data formats introduced in this PR more thoroughly in a content-creation page on the devdocs, but I will not be waiting on that for merge.

**! I intend to merge this PR no later than Friday, Jan 09 to allow adequate time for bugs to be found !**

@impaktor this will likely be a headlining feature for the release notes with multiple points related to how the economy now works. Please ping me on IRC once you start drafting the notes and I'll try to write a player-comprehensible introduction to the economy sim.

PR #6200, opened 2025-08-04
Show the return location in the mission list, if avaiable

### LINKED ISSUE #6185: Mission List: The Location column does not display the next/final destination
### Observed behaviour

The mission list always shows the first destination of any mission.

### Expected behaviour

The Location column should display the next destination of a mission, means
-the spaceport where mapping data have to be delivered when the status is completed;
-the spaceport where cargo has to be returned when status is active;
-the spaceport to return for fight missions (if such exist, I prefer peaceful play)

For cargo return, use status "to pick up" in the first step and "active" when loaded (same as one-way-delivery, just forget the history once the cargo is loaded?)

current master

next destination: write the code to be prepared/generalized for missions with n>2 steps?

robothauler:
I adopted the solution from the assassination module by simply overwriting mission.location.
In order not to change the mission description, I had to rename some variables.
If no defined return location is specified, the organization and, if necessary, the faction will be displayed.

Closes #6185

### DIFF
--- a/data/modules/FindPerson/FindPerson.lua
+++ b/data/modules/FindPerson/FindPerson.lua
@@ -128,7 +128,7 @@ local onChat = function (form, ref, option)
 			wanted      = ad.wanted,
 			company     = ad.company,
 			location    = ad.location,
-			destination = ad.location:GetStarSystem().path,
+			destination = ad.location:SystemOnly(),
 			visited     = {},
 			shipid      = ad.shipid,
 			ship        = nil,
@@ -478,7 +478,7 @@ end
 local buildMissionDescription = function (mission)
 	local ui = require 'pigui'
 	local desc = {}
-	local dist = Game.system and string.format("%.2f", Game.system:DistanceTo(mission.location:GetStarSystem().path)) or "???"
+	local dist = Game.system and string.format("%.2f", Game.system:DistanceTo(mission.location:SystemOnly())) or "???"
 	local domicileDist = Game.system and string.format("%.2f", Game.system:DistanceTo(mission.domicile)) or "???"
 	local danger = getRiskMsg(mission)
 
@@ -496,7 +496,7 @@ local buildMissionDescription = function (mission)
 		dist = dist
 	})
 
-	desc.location = mission.location:GetStarSystem().path
+	desc.location = mission.location:SystemOnly()
 	desc.client = mission.client
 	desc.returnLocation = mission.domicile
 
--- a/data/modules/FindPerson/FindPerson.lua
+++ b/data/modules/FindPerson/FindPerson.lua
@@ -127,8 +127,8 @@ local onChat = function (form, ref, option)
 			client      = ad.client,
 			wanted      = ad.wanted,
 			company     = ad.company,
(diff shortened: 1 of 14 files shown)


Bodasey:
HEEEYYYY!

MANY THANKS!

This is very good so!

But one thing is still wrong: Mapping missions are still marked as "Completed" instead of "Failed" if the due time is not met, and the money is transferred nevertheless.
Screenshot_20250804_202730
Screenshot_20250804_203424

sturnclaw:
> But one thing is still wrong: Mapping missions are still marked as "Completed" instead of "Failed" if the due time is not met, and the money is transferred nevertheless.

That is outside of the scope of this PR. You are welcome to make a separate issue for it! :smile:

----

@robothauler looking at the diff, could it be simpler to change the variable name used by the mission list to `destination` and keep the `location` variable as a mission-specific value instead? IMO 'destination' better expresses the 'next stop' semantics of the value we want to display in the mission list.

robothauler:
> @robothauler looking at the diff, could it be simpler to change the variable name used by the mission list to `destination` and keep the `location` variable as a mission-specific value instead? IMO 'destination' better expresses the 'next stop' semantics of the value we want to display in the mission list.

Yes, indeed. That makes sense.

robothauler:
I think it's ready for review now.

data/modules/FindPerson/FindPerson.lua
```diff
@@ -127,8 +127,8 @@ local onChat = function (form, ref, option)
 			client      = ad.client,
 			wanted      = ad.wanted,
 			company     = ad.company,
-			location    = ad.location:GetStarSystem().path,
-			destination = ad.location,
+			location    = ad.location,
+			destination = ad.location:GetStarSystem().path,
```
sturnclaw: Long after-the-fact correction: `SystemPath:GetStarSystem()` followed by `.path` is an incorrect form to use here. It potentially requires generating the entire StarSystem object again to fill the cache, which has performance implications.

Instead, code should always use `SystemPath:SystemOnly()` to retrieve the path to the containing system from an existing body path.

sturnclaw (approved):
Looks good - I'll try to find some time to playtest this (and I'd encourage others to do so as well!). One minor nitpick regarding existing code if you'd like to address it with this PR.

data/modules/FindPerson/FindPerson.lua
impaktor: I do wonder if there are places in the master branch that could do with this optimization? I can't recall myself being aware of this distinction...

sturnclaw (approved):
Thanks!

Bodasey:
Is it intended that the mission list of older games cannot be read any longer?
(has not been the case when i have tested this branch)

sturnclaw:
> Is it intended that the mission list of older games cannot be read any longer? (has not been the case when i have tested this branch)

It's not *intended*, but we've been meaning to do a savegame bump this release regardless. I'm considering this a "happy accident" and working on a larger project that is also going to cause us to break compatibility with saved mission state.

Bodasey:
Thank you for the information and the "savegame bump" label.

PR #6204, opened 2025-08-15
Animate thruster flame

fluffyfreak:
This is a resurrection of #6056 by possibly "jeffmd", it seemed a shame to lose the work with a nice visual effect

>The thrust flame is now animated by randomly expanding/contracting the flame and by making small changes in the flames diffuse color and alpha each frame. Its simplistic but for now adds some life to the ship when viewed externally.

Gifs by @bszlrd from that original PR
image
image

### DIFF
--- a/data/shaders/opengl/thruster.frag
+++ b/data/shaders/opengl/thruster.frag
@@ -1,0 +1,27 @@
+// Copyright © 2008-2025 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+#include "attributes.glsl"
+#include "lib.glsl"
+
+uniform sampler2D texture0; //diffuse
+
+in vec2 texCoord0;
+in vec4 flame_color;
+in float ramp;
+
+out vec4 frag_color;
+
+void main(void)
+{
+	vec4 color = texture(texture0, texCoord0);
+	vec3 old_color = color.rgb;
+	color *= flame_color;
+	// generate a light blue flickering glow around the flame
+	float ramp2 = ramp * ramp;
+	color.g += ramp2 * ramp2 * old_color.g;
+	color.r += ramp2 * ramp * old_color.r;
+	color.b += ramp2 * old_color.b;
+
+	frag_color = color;
+}
--- a/data/shaders/opengl/thruster.shaderdef
+++ b/data/shaders/opengl/thruster.shaderdef
@@ -1,0 +1,12 @@
+## Copyright © 2008-2025 Pioneer Developers. See AUTHORS.txt for details
+## Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+# Basic shader for drawing unlit, possibly textured models that are animated
+
+Shader thruster
+
+Texture sampler2D texture0 name=diffuse
+Buffer DrawData binding=1
+
+Vertex "thruster.vert"
+Fragment "thruster.frag"
--- a/data/shaders/opengl/thruster.vert
+++ b/data/shaders/opengl/thruster.vert
@@ -1,0 +1,46 @@
+// Copyright © 2008-2025 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+// #extension GL_ARB_gpu_shader5 : enable
+
+#include "attributes.glsl"
+#include "lib.glsl"
+
+out vec2 texCoord0;
+out vec4 flame_color;
+out float ramp;
+
+float map_percentage( float val, float base_percent)
+{
+   return base_percent + val * (1.0 - base_percent);
+}
+
+void main(void)
+{
+	// emission.r is the thruster power setting
+	float thruster_power = material.emission.r;
+	// emission.a is the random flicker value
+	float random = material.emission.a;
+	
+	// power setting effects the brightness and opacity of the flame
+	float brightness = map_percentage(thruster_power, 0.25);
+	// flicker should only change the brightness a small amount each frame
+	float flicker = map_percentage(random, 0.95);
+	
+	flame_color = material.diffuse * flicker * brightness;
+	// a very light blue flickering glow around flame
+	flame_color.rg *= vec2(flicker);
+
+	// the power setting also controls the length of the flame
+	// flame length can range from 50% to 100%
+	float flame_length = map_percentage(thruster_power, 0.5);
+	// modulate the flame shape using flicker
+	// flame_length only effects the z axis which is the thrust direction
+	vec4 scale = vec4(flicker, flicker, flicker * flame_length, 1.0);
+	
+	gl_Position = uViewProjectionMatrix * (a_vertex * scale);
+
+	texCoord0 = a_uv0.xy;
+	// set the blend ramp
+	ramp = a_uv0.y * 0.5;
+}
--- a/src/scenegraph/Thruster.cpp
+++ b/src/scenegraph/Thruster.cpp
@@ -29,7 +29,8 @@ namespace SceneGraph {
 		linearOnly(_linear),
 		dir(_dir),
 		pos(_pos),
-		currentColor(baseColor)
+		currentColor(baseColor),
+		m_time(0.f)
 	{
 		//set up materials
 		Graphics::MaterialDescriptor desc;
@@ -43,12 +44,12 @@ namespace SceneGraph {
 
 		auto vtxFormat = Graphics::VertexFormatDesc::FromAttribSet(Graphics::ATTRIB_POSITION | Graphics::ATTRIB_UV0);
 
-		m_tMat.Reset(r->CreateMaterial("unlit", desc, rsd, vtxFormat));
+		m_tMat.Reset(r->CreateMaterial("thruster", desc, rsd, vtxFormat));
 		m_tMat->SetTexture("texture0"_hash,
 			Graphics::TextureBuilder::Billboard(thrusterTextureFilename).GetOrCreateTexture(r, "billboard"));
 		m_tMat->diffuse = baseColor;
 
-		m_glowMat.Reset(r->CreateMaterial("unlit", desc, rsd, vtxFormat));
+		m_glowMat.Reset(r->CreateMaterial("thruster", desc, rsd, vtxFormat));
 		m_glowMat->SetTexture("texture0"_hash,
 			Graphics::TextureBuilder::Billboard(thrusterGlowTextureFilename).GetOrCreateTexture(r, "billboard"));
 		m_glowMat->diffuse = baseColor;
@@ -61,7 +62,8 @@ namespace SceneGraph {
 		linearOnly(thruster.linearOnly),
 		dir(thruster.dir),
 		pos(thruster.pos),
-		currentColor(thruster.currentColor)
+		currentColor(thruster.currentColor),
+		m_time(thruster.m_time)
 	{
 	}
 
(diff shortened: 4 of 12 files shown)


impaktor:
Good eye-candy to have in the release notes next year.

ghost:
I didn't resubmit this because I had used AI to generate most of the code back in Jan/Feb 2025 and in March noticed the AI clause in the README.md : https://github.com/pioneerspacesim/pioneer/blob/master/README.md#ai-policy.

src/scenegraph/Thruster.cpp
```diff
@@ -102,8 +104,24 @@ namespace SceneGraph {
...
 		}
 		if (power < 0.001f) return;
-
-		m_tMat->diffuse = m_glowMat->diffuse = currentColor * power;
+		
+		// animate the thrust flame using the thruster shader
+		// update animation time on each render
+		// a unique time stamp is needed for each thruster flame to have a unique flicker
+		// use thruster position to make each thruster time different
+		m_time += abs(trans[12] + trans[13] + trans[14]) * rd->timeStep * 1000.0f;
```
sturnclaw: Time should not be mutated per-thruster (nor should it be stored per-thruster). Instead two things should happen here:

1. The global "game time" should be used as the time source (this PR already pauses time accumulation when timestep = 0), multiplied by the desired thruster flicker frequency scaled by thruster size.
2. The thruster should compute a hash value from some pseudorandom source (sum of local positions is a good idea), *map that hash value to 0..1* and use it to mutate the frequency of the flicker. In testing, changing the phase while keeping the frequency the same across thrusters creates an uncomfortable visual effect, whereas changing the frequency (with a high-quality hash source) creates a feeling of "natural chaos".

Also, having started messing with the code to figure out why this thing works when it clearly shouldn't... `trans` is **view-space** position and will be different every frame, meaning thrusters will flicker during pause... only when the view is rotated.

src/scenegraph/Thruster.cpp
```diff
@@ -102,8 +104,24 @@ namespace SceneGraph {
...
+		// generate a psuedo random flicker
+		// this could be done in the vertex shader but the value only needs to be 
+		// generated once per frame, not for every vertex
+		const float flicker = sin(m_time) * 43758.5453f;
```
sturnclaw: Where the random constant came from, I don't know (possibly just AI-generated artifact).

This is an evil undefined behavior here (specifically the assignment to an 8-bit integer); looking at the disassembly the behavior *on my machine* is to convert to signed int32 then take the low byte of the integer representation and assign to the (unsigned) alpha channel. Specifically, the (AT&T-style) disassembly turns out to be `cvttss2si [stack], %eax` followed by `movb %al, [stack]`.

Obviously, this constant should be removed entirely and the function changed to `abs(sin())`.

fluffyfreak:
Ok I've rebased, fixed up the commented changes, adopted @sturnclaw hash system, passed through the game time, including guarding when there is no `game` running, and added a linear fade in/out effect like in the original gifs when using the keyboard.

This reduces the changed files to 5: shader + thruster files only.

fluffyfreak:
ah dammit, this doesn't like me dragging in those headers it seems

src/TimeUtils.h
```diff
@@ -0,0 +1,12 @@
+// Copyright © 2008-2025 Pioneer Developers. See AUTHORS.txt for details
+// Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+#ifndef _TIMEUTILS_H
+#define _TIMEUTILS_H
+
+namespace TimeUtils {
```
fluffyfreak: This is very hacky and I'm not keen on it, but I also don't like polluting the `Node.h` and every other class just to give it access to a `timeStep` that is either `(1.0 / 60.0)` or calculated only when the `Game` class is instantiated and in use.

src/Game.cpp
```diff
@@ -358,6 +359,10 @@ void Game::TimeStep(float step)
 
 	SfxManager::TimeStepAll(step, m_space->GetRootFrame());
 
+	// post these where they can be accessed without dependencies anywhere
```
fluffyfreak: It would be nice to have some kind of globally visible blackboard where these could be posted each update and queried at will. What's the better way of doing this?

src/TimeUtils.h
sturnclaw: Long-term, this kind of state should be stored in a `RenderWorld` object that is passed as a dependency to various render functions. My vision is for thrusters, StaticGeometry nodes, etc. to simply have a lightweight handle (32-bit ID+version) to a `RenderMesh` object associated with the RenderWorld, and the work of `Thruster::Render` to be setting the transform and material parameters of that RenderMesh. This proposed design would separate graphics resources from gameplay resources (e.g. tags, collision, etc).

Eventually, this kind of information should just be uploaded to the GPU in a global render-pass data buffer and all this math moved to the shader instead.

In the short term, this is an acceptable hack, though very very rough. I'd prefer passing this through the RenderData mechanism instead, and sprucing up that class as necessary to make it more than just an internal implementation detail.

src/scenegraph/Thruster.cpp
```diff
@@ -24,12 +25,24 @@ namespace SceneGraph {
...
+	}
+
 	Thruster::Thruster(Graphics::Renderer *r, bool _linear, const vector3f &_pos, const vector3f &_dir) :
 		Node(r, NODE_TRANSPARENT),
 		linearOnly(_linear),
 		dir(_dir),
 		pos(_pos),
-		currentColor(baseColor)
+		currentColor(baseColor),
+		m_time(0.f),
```
sturnclaw: `m_time` variable should be removed entirely.

src/scenegraph/Thruster.cpp
```diff
@@ -101,9 +116,35 @@ namespace SceneGraph {
...
+		// animate the thrust flame using the thruster shader
+		// update animation time on each render
+		// a unique time stamp is needed for each thruster flame to have a unique flicker
+		// use thruster position to make each thruster time different
+		// generate a psuedo random flicker
+		// this could be done in the vertex shader but the value only needs to be
+		// generated once per frame, not for every vertex
+		float hash = pos.x + pos.y + pos.z;
+		hash = (uint16_t(hash32(*reinterpret_cast<uint32_t *>(&hash)) & 0xFFFF)) / 65535.f;
+		m_time += timeStep * 55.f * (0.75f + hash * 0.5f);
```
sturnclaw: Rather than storing a per-thruster time value, strongly prefer accessing the global (application or game) time instead. This should use `TimeUtils::s_renderTime` (renamed from `s_gameTime`) in the following `abs(sin(time))` equation. Most of the above works regardless of whether you're applying the multiplication piecewise or to the time value as a whole.

src/Game.cpp
sturnclaw: Please see the prior response :smile:

sturnclaw (changes requested):

src/TimeUtils.h
fluffyfreak: I've had to revert this, it didn't work, and I don't quite understand why but suspect that it was defining multiple statics across multiple TUs. SO you'd set it from one place and the value wouldn't change when read elsewhere.

I can now totally understand why JeffMD did it via the RenderData :(

fluffyfreak:
I'm going to leave this for a couple of days now. Quite annoyed that I couldn't avoid modifying Node etc

data/shaders/opengl/thruster.frag
```diff
@@ -0,0 +1,27 @@
+// Copyright © 2008-2025 Pioneer Developers. See AUTHORS.txt for details
```
impaktor: 2026

data/shaders/opengl/thruster.shaderdef
```diff
@@ -0,0 +1,12 @@
+## Copyright © 2008-2025 Pioneer Developers. See AUTHORS.txt for details
```
impaktor: 2026

data/shaders/opengl/thruster.vert
```diff
@@ -0,0 +1,46 @@
+// Copyright © 2008-2025 Pioneer Developers. See AUTHORS.txt for details
```
impaktor: 2026

src/scenegraph/Thruster.cpp
```diff
@@ -184,18 +219,18 @@ namespace SceneGraph {
...
 
 			//uv coords
-			const vector2f topLeft(0.f, 1.f);
-			const vector2f topRight(1.f, 1.f);
-			const vector2f botLeft(0.f, 0.f);
-			const vector2f botRight(1.f, 0.f);
+			static const vector2f topLeft(0.f, 1.f);
+			static const vector2f topRight(1.f, 1.f);
+			static const vector2f botLeft(0.f, 0.f);
+			static const vector2f botRight(1.f, 0.f);
```
sturnclaw: Note: these can be `constexpr` as well; the `vector2f` constructor may need to be marked constexpr to enable this.

src/scenegraph/Thruster.cpp
```diff
@@ -101,9 +113,32 @@ namespace SceneGraph {
...
+		// animate the thrust flame using the thruster shader
+		// update animation time on each render
+		// a unique time stamp is needed for each thruster flame to have a unique flicker
+		// use thruster position to make each thruster time different
+		// generate a psuedo random flicker
+		// this could be done in the vertex shader but the value only needs to be
+		// generated once per frame, not for every vertex
+		float hash = pos.x + pos.y + pos.z;
+		hash = (uint16_t(hash32(*reinterpret_cast<uint32_t *>(&hash)) & 0xFFFF)) / 65535.f;
+		const float flicker = abs(sin(rd->renderTime * 55. * (0.75 + hash * 0.5)));
```
sturnclaw: Note: these are all double-precision constants and the math is being implicitly promoted to double-precision. Ideally prefer the use of single-precision (incl. `sinf` if necessary) for speed (some models *coughAC33cough* have 100s of thruster nodes).

src/scenegraph/Thruster.cpp
```diff
@@ -101,9 +113,32 @@ namespace SceneGraph {
...
+
+		// pass the power setting and flicker value using the material emissive
+		// emissive.a is the flicker value for the flame
+		m_tMat->emissive.a = m_glowMat->emissive.a = flicker * 255.f;
+		
+		// emissive.r is the thruster power setting which effects flame length and brightness
+		m_tMat->emissive.r = m_glowMat->emissive.r = 255.0f * displayedPower;
```
sturnclaw, line 139: I'm not a huge fan of retaining the `displayedPower` state variable, but if the smooth fade makes a perceptual difference then I will not protest it.

src/scenegraph/Thruster.cpp
fluffyfreak: had to `static_cast` the `renderTime` to float but everything else changed to float.

src/scenegraph/Thruster.cpp
fluffyfreak: Did indeed have to mark the ctor as `constexpr`

src/scenegraph/Thruster.cpp
fluffyfreak: I think it does, but feel free to hack at it and see if there's anything better!

sturnclaw: Will do any further hacking-about in master, thanks!

impaktor:
yay!

This will be nice bling-bling for the release announcement.

fluffyfreak:
It'd be nice to give jeffmd some credit for this but I've no idea how to contact them

bszlrd:
Their user is still present on the discord server. Not sure if active.

PR #6206, opened 2025-09-08
Scenegraph .model Format v2

sturnclaw:
Per comment in `scenegraph/Parser.h`:

> Newmodel .model config file parser.
> It's pretty bad, someone please redesign.

I am someone. I came, I saw, I redesigned.

The new format uses curly braces to delineate materials and LODs, uses a vastly improved parsing engine that can handle line comments, and **most importantly** allows specifying the shader, render state, and texture bindings used in materials. This means we're no longer locked to only using the `multi` shader with four specific textures (though there is still some cruft that wants model shaders to support heat coloring and patterns even if they're not the `multi` shader), and we can support fancy transparent blends like premultiplied alpha in model materials.

This is still not quite perfect, because the { shader, render-state } pair is really something that should be defined ahead of time in a `.shaderdef` file (so that the graphics state can be compiled at startup), but it's a vast improvement over the hacky hard-coded mess that we had prior.

I've also migrated the ShaderParser to use the new parsing engine, giving it much better support for end-of-line comments.

As a consequence of this, I've had to bump the SGM version; the material definition stored in SGM files was incapable of expressing the new flexibility and had to be expanded.

CC @fluffyfreak as I recall SceneGraph code being an area of your interest.

## Testing Instructions

I've tested this fairly extensively locally and am satisfied it works under most conditions. I'd like testers to focus on ensuring there are no visible bugs with our complete library of models (i.e. models rendering black, or ships/buildings completely without textures, etc.) and that things like space stations still render correctly in-game. The modelviewer is your friend here!

## Further Work

I've converted the Coronatrix to the new model format, but the remainder of the .model files has yet to be converted. I'd appreciate it if someone would tackle this; the format is fully exercised in `coronatrix.model`.

The dev docs related to .model file authoring need to be updated to provide information regarding the new syntax. Anyone is welcome to tackle this; if not, I'll do it sometime before merge.

### DIFF
--- a/src/scenegraph/Parser.h
+++ b/src/scenegraph/Parser.h
@@ -8,6 +8,7 @@
  * It's pretty bad, someone please redesign.
  *
  * sturnclaw (05-2025): redesign in progress
+ * sturnclaw (09-2025): redesign complete (kept old version for backwards compat)
  */
 #include "FileSystem.h"
 


fluffyfreak:
I did look at this, forgot to comment though. The code looks good, I need to find time for some testing.

fluffyfreak:
@sturnclaw I cannot push directly to your branch, but here are the shields and cockpits converted to version 2 https://github.com/pioneerspacesim/pioneer/commit/eb2b0ee5aec78d8eb8361a8ee2b1f47dff37d917

EDIT: I'm updating all of the model files, good braindead work for today https://github.com/fluffyfreak/pioneer/tree/model-parser-v2-model-files

EDIT 2: All ".model" files converted to version 2 🎉 I've checked them all and they all seem fine.
I also ran into an issue with there being 2 `sinonatrix_cockpit.model` and this causing an issue with picking them in the model viewer, so added a fix for that. @bszlrd do you know why there are two? I think it might have exposed a bug/issue with the patterns as well but I'll look into that seperately.

src/modelcompiler.cpp
```diff
@@ -143,7 +143,7 @@ void RunCompiler(const std::string &modelName, const std::string &filepath, cons
 	}
 
 	timer.Stop();
-	Output("Compiling \"%s\" took: %lf\n", modelName.c_str(), timer.millicycles());
+	Output("Compiling \"%s\" took: %0.2fms\n", modelName.c_str(), timer.milliseconds());
```
fluffyfreak, line 146: This is a MUCH better metric, no idea why I didn't do this originally, thankyou

bszlrd:
Hmm, now that's a good question. I can't recall why I did that. Likely forgot to delete the one in the sino directory when I moved it to its own cockpit dir.
The meshes and textures are duplicated too. I might want to fix that.

src/modelcompiler.cpp
sturnclaw: To be fair, I didn't write `Profiler::Clock` until well after I got here. At the time the modelcompiler was written, the only wall-clock timing code in the codebase was a homespun abomination based on counting `SDL_GetTicks()`.

sturnclaw:
> It would have been nice to have a `.model` format exporter as well as parser so that in the future the model viewer could be used to edit/create models... also might have made updating the files automatic but that's probably outside the scope.

That's... a good bit outside of scope for now. Ideally I'd like that to be part of the model editor, but that requires the model editor to be able to present the filesystem to the user, and would need a general purpose material / model edition tool to be written.

Hand-written text files is good enough at this point, but I'll consider it if/when we extend the format with more complicated stuffs.

sturnclaw:
> EDIT 2: All ".model" files converted to version 2 🎉 I've checked them all and they all seem fine.

@fluffyfreak if you don't mind, I'll ask you to just open that as a PR once this branch is merged. I have some documentation to update, so I'm targeting 2-3 days until merge.

sturnclaw:
Documentation updated: https://dev.pioneerspacesim.net/workflow/3d_assets/model_system. Going to merge this; thanks for the reviews!

Issue #5907, opened 2024-08-29
Asteroids of unusual size


zonkmachine:
Found this while looking up lots of systems for the sake of testing planet and base names. Rarely an A.O.U.S. (asteroid of unusual size) would pop up. I haven't seen this with any of the other system bodies.  Could be related to density calculations and https://github.com/pioneerspacesim/pioneer/pull/5592 .

master 2c962df832c69c5cd9f4b83e215d7b4916cb2b8f
Iumir (-14, 11, -9)
![asteroid](https://github.com/user-attachments/assets/f7f212a5-8237-4408-90af-fe279b2b85e2)

### Observed behaviour
Asteroid icon can take on a very large size. It seem to also show unrealistic physical values, density for instance.

### Steps to reproduce
Very odd to find a really big one but either manually go through 100's of systems or load the save game below and search for 'iumir).

Ubuntu 22.04
master (2c962df832c69c5cd9f4b83e215d7b4916cb2b8f)

[output.txt](https://github.com/user-attachments/files/16802119/output.txt)
[opengl.txt](https://github.com/user-attachments/files/16802168/opengl.txt)
[asteroidfun.zip](https://github.com/user-attachments/files/16802143/asteroidfun.zip) (not zipped, just renamed)

Bodasey:
Bodies of this mass usually have radii of around 10 km, not twice or more the radius of Jupiter.

Better set the mass, set the type (icy moon, dry moon, dry material asteroid), add an according density and calculate the size?

The real models of this big image, (101955) Bennu and (162173) Ryugu, destination of OSIRIS-REx and Hayabusa 2 who have brought material from there to Earth, are even smaller (250m and 500m radius) and lighter (7*10^10 and 5*10^12 kg) - I don't know which role such bodies shall have in our game.

zonkmachine:
This doesn't seem to be a new problem but maybe the new system body icon set, the one with asteroid **101955 Bennu** that is in current use, made the issue visible. That and the inclusion of density info that makes the issue more visible.

Here's an asteroid approaching the size of Mars from the commit before #5592 was merged
![asteroidsizeold](https://github.com/user-attachments/assets/946fe856-20da-4738-8bf4-4078b64fc395)

zonkmachine:
Duplicate of https://github.com/pioneerspacesim/pioneer/issues/5854

Issue #5854, opened 2024-07-05
Some celestial bodies: Massive Tiny Rock Moon with 0.00 kg/m3 density


impaktor:
Discovered in video 10, of [Pioneer - Frequently Faililing](https://www.youtube.com/playlist?list=PLRNUEKviDIzHqRQUQfYHnZquHT3_cpXG6)

### Observed behaviour
![2024-07-05-062826_1269x692_scrot](https://github.com/pioneerspacesim/pioneer/assets/619390/3dd61c1b-d5be-4367-ad6c-e53b6fe01446)

__My pioneer version (and OS):__ Current master, ae53202e087e2d660e630a2e9223a8a7c9882482

impaktor:
```
<sturnclaw> Pretty sure that's because the moon's mass is the smallest
            expressible number in 32.32 fixed-point [08:33:16]
<sturnclaw> that's an unfortunate issue but more a side-effect of how we
            generate bodies [08:33:37]
<sturnclaw> Nothing I can really do about it at the moment - I think the
            massive icon is a completely different issue from the moon being
            generated with a low mass [08:34:09]
<sturnclaw> ...or rather it's a display issue rather than something directly
            stemming from the moon's lack of mass [08:34:27]
```

sturnclaw:
Possibly fixed now, low-end precision has been significantly improved when generating moons and we should no longer have zero-mass moons.

PR #6209, opened 2025-09-14
Prevent spiky asteroids in random systems

### LINKED ISSUE #5999: Asteroid procedural generator issue
### Observed behaviour
asteroid seems to be procedural generator issue resulting asteroid is larger than planet (Iumir h for example) and even weird looking asteroid like spikes, resulting larger than individual asteroids or planet, and speed warp has reduced. similar to #5907
Iumir (-14,11,-9)

system map:
Screenshot (78)

world view:
Screenshot (80)
Screenshot (81)

### Expected behaviour
same as #5907

### Steps to reproduce
similar as #5907, the asteroid to be regenerated for all system to real size not very larger ones. And search systems for potential larger asteroids (Capella for Example) and use the debug mode for hyperspace much easier especially further system ones

20240710 (windows)

fluffyfreak:
The goal of this is to _try_ and fix #5999 and other issues like it, I suspect that it is at best a partial fix.

~I don't think that the conversion between fixedf<32> (_the default fixed format_) and fixedf<48> works, it may just copy the value raw which was producing planetary radius's millions of times greater than their mass.~

In Sol there are still bodies being created which have quite insane ratios, for example:
	Eros:
	height fractal: MountainsCraters2
	colour fractal: Rock
	seed: 842785371
	mass: 0.0000000009313225746154785
	radius: 0.0014999997802078724
	mass:radius ratio 1610612.5

That is a radius 1,610,612 times greater than it's mass, at which point some of our maths around terrain height scaling calculations goes absolutely insane.

However that's coming through our custom data, and this is about randomly generated bodies. This seems to fix the problem in my testing, but it will cause a savegame bump because it changes the galaxy.

Also this will need some testing by people to confirm, easiest way is to use the navigation menu and Ctrl+F10 to look at all of the bodies in a system.

### DIFF
--- a/src/galaxy/StarSystemGenerator.cpp
+++ b/src/galaxy/StarSystemGenerator.cpp
@@ -33,7 +33,7 @@ static const fixed TWOHUNDREDEUMASSES = fixed::FromDouble(200.0);
 
 // Estimate "prototype" body radius from sphere-density formula:
 // Convert 1 Earth Mass to a volume in cubic megameters (10^18 * m^3) then premultiply by 3/4pi
-static const fixedf<48> EARTH_MASS_TO_VOL_MM3 = fixedf<48>(2599, 10); // 259.901 = (3 / EARTH_DENSITY / 4*PI) * 5.9742
+static const fixedf<48> EARTH_MASS_TO_VOL_MM3 = fixedf<48>(25946, 100); // 259.46 = (3 / EARTH_DENSITY / 4*PI) * 5.9742
 // Convert a distance in megameters to earth radii
 static const fixedf<48> MM_TO_EARTH_RAD = fixedf<48>(15678, 100000);
 


mwerle:
Iumir B h 2 still looks pretty funky:

image

mwerle:
.. so after calculating the radius from the mass, there's an adjustment (line 686) which forces the body to be a minimum of 10km radius. Possibly this adjustment should also increase the mass of the body otherwise you end up with these extremely low-density objects which, if I understand correctly, is the cause of the "spiky" asteroids.

Finally, for Asteroids (minor bodies), there's no additional information added to the body. Perhaps we could specify whether it's metallic, rocky, or icy based on its density? (I appreciate that this goes well beyond the scope of this PR... but if we're already changing the galaxy and save-game bumping, we might as well?)

sturnclaw:
> I don't think that the conversion between fixedf<32> (the default fixed format) and fixedf<48> works, it may just copy the value raw which was producing planetary radius's millions of times greater than their mass.

Easy way to test that, I stuck it in a debugger and inspected the value on both sides of the conversion. The conversion from f32 to f48 is definitely happening correctly; testing `Lalande 21185 a` (for no reason other than it's the first planet to trigger the breakpoint on startup) with a mass value of 0.007 earths, the f32 value is `31556113` and the f48 value is `2068061421568`. Dividing these by the respective 2^N, it yields the same value.

> That is a radius 1,610,612 times greater than it's mass, at which point some of our maths around terrain height scaling calculations goes absolutely insane.

Remember that mass is expressed in earth masses, i.e. the unit is 5.9742e24 kg, while the radius unit is expressed in earth radii, i.e. a unit of 6.3781e6 m. The mass is still incredibly large.

> This seems to fix the problem in my testing, but it will cause a savegame bump because it changes the galaxy.

Unfortunately, all the proposed fix does is reduces the precision of the calculation such that you're getting a zero radius for bodies with a low enough mass.

Digging into the problem deeper, the root cause of bugged terrain happens to be an overflow triggered in `CubeRootOf`, where `x * x` is such a small value that it is causing a fixed-point overflow during division when dividing the initial mass value `a`. Fortunately, there's a well-defined way to solve the problem - we just do the cube root on a *much* larger number. 259.9 times larger, to be precise.

Rather than (wrongly) attempting to estimate radius from the cube root of the mass, we should instead estimate the volume of a sphere of earth-like material, then compute the radius of that sphere.

```c++
// Estimate "prototype" body radius from sphere-density formula:
// Convert 1 Earth Mass to a volume in cubic megameters (10^18 * m^3) then premultiply by 3/4pi
static const fixedf<48> EARTH_MASS_TO_VOL_MM3 = fixedf<48>(2599, 10); // 259.901 = (3 / EARTH_DENSITY / 4*PI) * 5.9742
// Convert a distance in megameters to earth radii
static const fixedf<48> MM_TO_EARTH_RAD = fixedf<48>(15678, 100000);

// (in StarSystemRandomGenerator::PickPlanetType)
if (sbody->GetMassAsFixed() <= fixed(1, 1)) {
	// We know the upper bound on this value is +259.9, so 16 integer bits is fine.
	fixedf<48> vol = fixedf<48>(sbody->GetMassAsFixed()) * EARTH_MASS_TO_VOL_MM3;
	// If the cube-root continues to underflow, pre-shift vol left by 6 bits, then post-shift the result by 2 bits.
	// sbody->m_radius = (fixedf<48>::CubeRootOf(vol << 6) >> 2) * MM_TO_EARTH_RAD;
	sbody->m_radius = fixedf<48>::CubeRootOf(vol) * MM_TO_EARTH_RAD;
}

```

Before:
image

After (note the correctly-calculated radius of 84.16km, yielding a mean density of 3720kg/m^3 after random density offset):
image
image

fluffyfreak:
> Iumir B h 2 still looks pretty funky:
> 
> image

That's just the funky terrain generator used, the asteroid ones are quite shit

fluffyfreak:
@sturnclaw that is a much nicer fix, shall I replace my fix with yours or do you want to PR it separately?

Also, why exactly `259.9 times larger`?

sturnclaw:
> @sturnclaw that is a much nicer fix, shall I replace my fix with yours or do you want to PR it separately?
> 
> Also, why exactly `259.9 times larger`?

Replacement is fine, I don't have the focus right now to spin up a new PR.

As to why that number, I'll expand the comment a little. We start with the end goal - we want to compute the radius of a body from the mass of the body. Starting with the formula for the radius of a sphere (remembering that $\sqrt[3] x = x^\frac 1 3$):

$$
r = \lparen \frac {3 V} { 4 \pi } \rparen ^ {\frac 1 3}
$$

We know the mass $M$ of the body, and have a reference mean density $\rho$ (in kg/m³) so we can estimate the volume of an ideal sphere with the mean density of the Earth. 

$$ V = \frac M \rho $$

I used the density value 5487.6kg/m³, which is technically incorrect as the real-world reference is 5513kg/m³, and the values in `gameconsts.h` imply a mean density of 5496.8kg/m³. I do not at this time recall exactly how I arrived at that value; I'll provide an updated multiplication factor at the end.

With the computation from mass, the equation winds up being:

$$
r = \lparen \frac { 3 \frac M {5487.6} } { 4 \pi } \rparen ^ \frac 1 3
$$

Now if we tried to do the calculation as-is on kilograms and meters we'd immediately exceed the range of a 32.32 fixed-point number, which can only represent a value about 9 orders of magnitude large. Instead, we need to represent $M$ as a value in a much larger unit. For this, I settled on a unit of petatons, or $10^{18}\text{kg}$.

Following the rule for exponentiation $(ab)^m = a^m b^m$, to multiply $M$ by $10^{18}$, we have to multiply both sides by $10^{18 \frac 1 3}$, or $10^6$.

$$
r\times10^6 = \lparen \frac { 3 \frac {M \times 10^{18} } {5487.6} } { 4 \pi } \rparen ^ \frac 1 3
$$

This gives us a radius in megameters, which is perfect because 1 earth radius $R_🜨$ happens to be 6.37 Mm.

Since we're doing all of this in fixed-point, we want to combine as many factors as possible to preserve precision (and reduce runtime cost). We start with the scaling factor $5.9742 \times 10^6$ to convert from a unit of earth masses $M_🜨$ into petatons, and then combine the remaining factors from the sphere-radius equation:

$$
s = \frac { 3 \times 5.9742 \times 10^6 } { 5487.6 \times 4 \pi  } = 259.9
$$

Assigning to `sbody->m_radius` simply involves a division (in our case, multiplication by the reciprocal) to scale the unit from Mm to earth radii $R_🜨$.

$$
R_🜨 = \frac { \sqrt[3] {M_🜨 s} } {6.378135}
$$

Because we've combined all of our factors to produce a relatively small scalar, we can represent any mass value < $1M_🜨$ in a 16.48 fixed-point number without worrying about an overflow, as the result of the cube-root will never be larger than the input (or 1 if the input is <1).

With the more accurate density value, the scalar `EARTH_MASS_TO_VOL_MM3` should actually be 259.46. `fixedf<48>(25946, 100)` should be computable without overflow, I think.

Footnote: the ULP for a 32.32 fixed-point representation in units of Earth Mass is approximately $2.33 \times 10^{14} \text{kg}$, or 0.23 teratons.

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -667,8 +667,20 @@ void StarSystemRandomGenerator::PickPlanetType(SystemBody *sbody, Random &rand)
 	// AndyC - Updated to use the empirically gathered data from this site:
 	// https://phl.upr.edu/library/labnotes/standard-mass-radius-relation-for-exoplanets
 	// but we still limit at the lowest end
+	// 
+	// Estimate "prototype" body radius from sphere-density formula:
+	// Convert 1 Earth Mass to a volume in cubic megameters (10^18 * m^3) then premultiply by 3/4pi
+	static const fixedf<48> EARTH_MASS_TO_VOL_MM3 = fixedf<48>(2599, 10); // 259.901 = (3 / EARTH_DENSITY / 4*PI) * 5.9742
+	// Convert a distance in megameters to earth radii
+	static const fixedf<48> MM_TO_EARTH_RAD = fixedf<48>(15678, 100000);
```
sturnclaw: Note that these are intended to be defined at file-scope with other static consts, rather than in the function.

fluffyfreak: Cool, I fixed up some comments, then moved those constants and some others to the file-scope. Also got rid of an unneeded temporary variable that was in there for some reason.

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -28,6 +28,14 @@ static const fixed AU_SOL_RADIUS = fixed(305, 65536);
 static const fixed AU_EARTH_RADIUS = fixed(3, 65536); // XXX Duplication from StarSystem.cpp
 static const fixed FIXED_PI = fixed(103993, 33102);	  // XXX Duplication from StarSystem.cpp
 static const double CELSIUS = 273.15;
+static const fixed ONEEUMASS = fixed::FromDouble(1);
+static const fixed TWOHUNDREDEUMASSES = fixed::FromDouble(200.0);
+
+// Estimate "prototype" body radius from sphere-density formula:
+// Convert 1 Earth Mass to a volume in cubic megameters (10^18 * m^3) then premultiply by 3/4pi
+static const fixedf<48> EARTH_MASS_TO_VOL_MM3 = fixedf<48>(2599, 10); // 259.901 = (3 / EARTH_DENSITY / 4*PI) * 5.9742
```
sturnclaw: Per comment above, the scalar that matches our values for `EARTH_RADIUS` and `EARTH_MASS` is actually 259.46 rather than 259.9.

```suggestion
static const fixedf<48> EARTH_MASS_TO_VOL_MM3 = fixedf<48>(25946, 100); // 259.46 = (3 / EARTH_DENSITY / 4*PI) * 5.9742
```

fluffyfreak:
I've adopted the suggested change. Since this is _probably_ a savegame bump shall I leave this for you to merge when that's more acceptable?

sturnclaw:
Works for me - it's a very minor impact savegame bump (should make no difference unless the save is on a planet affected by the bug), but might as well merge it with all the other terrain PRs.

I'm hoping to bump quite soon; I have a branch I'm currently writing that pulls a bunch of mission module boilerplate code into a single shared utility, but no ETA on when it's ready to go yet.

zonkmachine:
Tested, works fine. It looks like it takes care of https://github.com/pioneerspacesim/pioneer/issues/5907 and https://github.com/pioneerspacesim/pioneer/issues/5854 too.

I found another wonky asteroid in system Ackildr (2,5,1) which is also fixed in this PR.


> ... for example:
> Eros:

Right. Eros uses the real data but it's mean diameter instead of the mean radius and the mass in kg is used but we use tons. All wrong. I'll go over the solar system right now. I have a PR with fixes for the custom systems.

PR #6210, opened 2025-09-15
calculate star colors using Planck's law

Mc-Pain:
Small improvement to derive star colors from temperature

### DIFF
--- a/src/galaxy/SystemBody.cpp
+++ b/src/galaxy/SystemBody.cpp
@@ -23,6 +23,7 @@ SystemBodyData::SystemBodyData() :
 	m_averageTemp(0),
 	m_heightMapFractal(0)
 {
+	m_starColor = Color(255, 0, 255, 255);
 }
 
 void SystemBodyData::SaveToJson(Json &out)
--- a/src/galaxy/CustomSystem.cpp
+++ b/src/galaxy/CustomSystem.cpp
@@ -135,7 +135,6 @@ static double *getDoubleOrFixed(lua_State *L, int which)
 
 CSB_FIELD_SETTER_FIXED(radius, bodyData.m_radius)
 CSB_FIELD_SETTER_FIXED(mass, bodyData.m_mass)
-CSB_FIELD_SETTER_INT(temp, bodyData.m_averageTemp)
 CSB_FIELD_SETTER_FIXED(semi_major_axis, bodyData.m_semiMajorAxis)
 CSB_FIELD_SETTER_FIXED(eccentricity, bodyData.m_eccentricity)
 CSB_FIELD_SETTER_FIXED(rotation_period, bodyData.m_rotationPeriod)
@@ -152,6 +151,16 @@ CSB_FIELD_SETTER_STRING(space_station_type, bodyData.m_spaceStationType)
 #undef CSB_FIELD_SETTER_FIXED
 #undef CSB_FIELD_SETTER_FLOAT
 #undef CSB_FIELD_SETTER_INT
+
+static int l_csb_temp(lua_State *L)
+{
+	CustomSystemBody *csb = l_csb_check(L, 1);
+	int value = luaL_checkinteger(L, 2);
+	csb->bodyData.m_averageTemp = value;
+	csb->bodyData.GenerateStarColor();
+	lua_settop(L, 1);
+	return 1;
+}
 
 static int l_csb_radius_km(lua_State *L)
 {


src/galaxy/SystemBody.cpp
```diff
@@ -250,6 +250,100 @@ bool SystemBody::HasAtmosphere() const
...
+    vector3d( 0.000065f, 0.000025f, 0.000000f ),    // 770 nm
+    vector3d( 0.000033f, 0.000013f, 0.000000f ),    // 780 nm
+    vector3d( 0.000018f, 0.000007f, 0.000000f ),    // 790 nm
+    vector3d( 0.000009f, 0.000004f, 0.000000f ),    // 800 nm
+    vector3d( 0.000005f, 0.000002f, 0.000000f ),    // 810 nm
+    vector3d( 0.000003f, 0.000001f, 0.000000f ),    // 820 nm
+    vector3d( 0.000002f, 0.000001f, 0.000000f )     // 830 nm
+};
+
+Color SystemBody::GetColor() const
```
sturnclaw: This should be computed and cached at body creation time, rather than recomputed (potentially) multiple times every single frame. 48 exp()s and 96 pow()s per call is not cheap enough for the current call frequency.

Mc-Pain:
> Two questions:
> 
> 1. Is the table comprised of RGB color values or CIE values? I'm seeing a comment in the source shader indicating these values are actually in the CIE colorspace.

I just copy-pasted it and it seems to be actually RGB: note the red channel between 370 and 500 nm - it has red because this part of spectrum is bluish-purple

> 2. Have you considered precomputing a table of color values over the expected temperature range for stars (using perhaps an exponential mapping to preserve detail) rather than fully computing the Planck brightness every time the color of the star is requested?

Yes, I'll do it soon after fixing step 4 in https://github.com/pioneerspacesim/pioneer/issues/6136

src/galaxy/SystemBody.h
```diff
@@ -271,6 +271,8 @@ class SystemBody : public RefCounted, public SystemBodyType, protected SystemBod
 		// which is rendered when the body has a small screen size
 		return Color(200, 200, 200, 255);
 	}
+	double GetPlanckBrightness(const double wavelength_nm, const int temperature) const;
+	Color GetColor() const;
```
fluffyfreak: I'd prefer if this was more explicitly named as it's only for stars, e.g. `GetStarColor()`.

src/galaxy/SystemBody.h
```diff
@@ -271,6 +271,8 @@ class SystemBody : public RefCounted, public SystemBodyType, protected SystemBod
 		// which is rendered when the body has a small screen size
 		return Color(200, 200, 200, 255);
 	}
+	double GetPlanckBrightness(const double wavelength_nm, const int temperature) const;
```
fluffyfreak: If this is only used by the method `GetColor` then can it be moved into the private methods or does it need to be public for some later use?

src/galaxy/SystemBody.cpp
```diff
@@ -250,6 +250,100 @@ bool SystemBody::HasAtmosphere() const
 	return (m_volatileGas > fixed(1, 100));
 }
 
+double SystemBody::GetPlanckBrightness(const double wavelength_nm, const int temperature) const
+{
+	// Planck's constant
+	double h = 6.62607015e-34; // kg*m^2*s^-1 (J*s)
```
fluffyfreak: These single letter constants can all be made `static const` or probably better `constexpr`

fluffyfreak:
Nice idea I think and I'm interested to see results 👍 
If this is too much hassle then don't worry I'll dig into it myself but it would be good to see screenshots of before & after to see how this differs.

src/galaxy/SystemBody.cpp
fluffyfreak: Actually going further than that, since `GetPlanckBrightness` is a const method it could leave the class entirely and sit in the cpp just above the `GetColor` implemention, or in a namespace at the top of the file. Unless you want it accessible elsewhere.

Mc-Pain:
> Nice idea I think and I'm interested to see results 👍 If this is too much hassle then don't worry I'll dig into it myself but it would be good to see screenshots of before & after to see how this differs.

I'd expect to see something like this for O-types:
image

Bodasey:
A little visible test done and I think class M stars look a bit too much yellow?

Please remember that there are no stars in the sky who look really green for the human eye even when the radiation maximum is at green wavelengths.

Mc-Pain:
> there are no stars in the sky who look really green for the human eye even when the radiation maximum is at green wavelengths.

Where did you see green stars?

src/galaxy/SystemBody.cpp
Mc-Pain: I don't know where all bodies are created

Calling this method each time after initializing body temperature looks ugly for me

impaktor:
@Mc-Pain Status?
Also, perhaps rebase to master, so we don't have to re-compile every time we switch to this branch.
And squash fix-commits.

Mc-Pain:
> @Mc-Pain Status? Also, perhaps rebase to master, so we don't have to re-compile every time we switch to this branch. And squash fix-commits.

Rebased onto master [9de1f0a28]

Tonight I'll collect the color data from different temperatures by aligning Sol's temperature.
In given shadertoy there's a certain post-processing matrix, I'll test it too.

Mc-Pain:
Made a plot of different temperatures:

X axis - temperature in K
Y axis - log(luminosity)
Color - color of star, normalized to get maximum of 1.0

thin series: original calculation from shadertoy
thick series: color multiplied by xyz2rgb matrix

image

Mc-Pain:
Mars, Cydonia, Sunrise:

<details><summary>Without matrix</summary>
<p>

image

</p>
</details>

<details><summary>With matrix</summary>
<p>

image

</p>
</details>

impaktor:
@sturnclaw Do you have this in your "ready to review PRs"-queue?

sturnclaw:
I do now!

PR #6217, opened 2025-10-04
CMakeList and scan_enums update to python3

fluffyfreak:
When trying to build Pioneer on the latest Raspberry Pi OS "Trixie" it ran into issues with the `scan_enums.py` requiring Python2 which now appears to be so deprecated it cannot be recovered at all.

So I've tried updating it to Python3 in both `CMakeLists.txt` and the `scan_enums.py` script. This seems to work with only minimal changes and resolves the issues around Python2 deprecation. It successfully generates:

>	modified:   data/meta/Constants.lua
	modified:   src/enum_table.cpp
	modified:   src/enum_table.h

I have not included those modified files here.

### DIFF
--- a/scripts/scan_enums.py
+++ b/scripts/scan_enums.py
@@ -513,7 +513,7 @@ def main():
             if path == '-':
                 es = list(extract_enums(sys.stdin))
             else:
-                with open(path, 'rU') as fl:
+                with open(path, 'rt') as fl:
                     # skip an optional UTF-8 Byte Order Mark
                     if sys.version_info[0] >= 3:
                         hasbom = (fl.read(1) == '\uFEFF')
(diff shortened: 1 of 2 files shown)


scripts/scan_enums.py
```diff
@@ -513,7 +513,7 @@ def main():
             if path == '-':
                 es = list(extract_enums(sys.stdin))
             else:
-                with open(path, 'rU') as fl:
+                with open(path, 'rt') as fl:
```
mwerle: Nitpick: the "t" is superfluous - "r" implies text mode.

For other reviewers: it seems "U" (universal newlines) is now the default if the "newlines=" keyword is not specified as part of the open call, so there is no loss of functionality by changing the arguments to open.

fluffyfreak: Yeah I had to remove the `U` as it's not supported anymore that way and gives an error under Python3

impaktor:
Looks like running it adds Geoshpere & Gunmanager related constants.

impaktor:
Squashed, force-pushed & rebased to master. 
Thanks for PR

PR #6218, opened 2025-10-06
Assassination - refactor and string improvements

zonkmachine:
Fixing the dialogs of the assassination mission and refactoring the code. A first attempt to improve it. It could still use some variety in the missions.

From the commit messages:
     * Use target gender, FEMALE/MALE
     * Ship article. 'A' works for all ships present in the Assassination modules English strings.
     * Use different versions of the name, with or without title.
     * Remove unused seed
     * Remove old upgrade routine
     * Refactor - make target a character


data/lang/module-assassination/en.json
```diff
@@ -199,9 +199,13 @@
     "description": "",
     "message": "How will I be paid?"
   },
-  "IT_MUST_BE_DONE_AFTER": {
+  "IT_MUST_BE_DONE_AFTER_MALE": {
     "description": "",
-    "message": "It must be done after {target} leaves {spaceport}. Do not miss this opportunity."
```
mwerle, line 204: You could keep all 3 for more variety.

Ie, male, female, neutral. Have a certain percentage chance that neutral will be chosen instead of the gender-specific option.

data/modules/Assassination/Assassination.lua
```diff
@@ -64,6 +64,8 @@ end
 
 local onChat = function (form, ref, option)
 	local ad = ads[ref]
+	local target = ad.target
+	local female = ad.target.female and "FEMALE" or "MALE"
```
mwerle: Suggested change

```suggestion
	local gender = ad.target.female and "FEMALE" or "MALE"
```    
  

"female" as the variable name is completely misleading as to what the variable actually represents.
For some juvenile fun and snickering and less characters to type, use "sex" as the variable name instead.

As a somewhat out-of-scope comment, the same change should be made to the Character class. Or at the very least, the member should be renamed to `isFemale`.  Just `female` is .. weird and incorrect and not very self-documenting.

zonkmachine: > "female" as the variable name is completely misleading as to what the variable actually represents.
> For some juvenile fun and snickering and less characters to type, use "sex" as the variable name instead.
>
> As a somewhat out-of-scope comment, the same change should be made to the Character class. Or at the very least, the member should be renamed to isFemale. Just female is .. weird and incorrect and not very self-documenting.

I'll look into it but as you've noticed it's following the codebase which in this regard is matriarchal in nature.
ref: https://codedoc.pioneerspacesim.net/#LuaClass:Character:female

data/lang/module-assassination/en.json
zonkmachine: > Ie, male, female, neutral. Have a certain percentage chance that neutral will be chosen instead of the gender-specific option.

Yes. Pushed that in the last two commits. It works but isn't saved so if you currently push the button again it may show the other version. I don't think this is a problem though.

data/modules/Assassination/Assassination.lua
mwerle: Yes, I realise. While I accept fixing the rest of the codebase is out-of-scope here, I don't think we should continue the abomination.

I have nothing against matriarchy, but I am strongly opposed to obscure and non-self-describing code.

That said, this is all my personal opinion; I'm not in charge of this codebase :) I'm just being a coder-nazi and variable naming is somewhat subjective.

zonkmachine:
@mwerle Thank you for the review, I really appreciate it!

> These days people identify with a lot more genders than male or female...

People can identify precisely how they feel fits their body/spirit. Go forth and express yourself. This doesn't mean that people who don't identify themselves in those modern ways shouldn't have their part in the language. Right? Also, do you know the biological sex of the he/she in the strings suggested? No, you don't. You should know that one of the number one goals in the trans community is to 'pass' as the other sex. In this regard Pioneer is really good because it's impossible to see who is trans or not of the characters you come across. I know people who identify in all sorts of ways, partly because I have actually worked in projects with gay rights, have been on stage with transvestites and have lived in communities with gays, lesbian, transsexuals. etc. The people who want to erase he/she from language is a minority and they are free to express themselves in those ways as I am free to express myself as suggested in this PR.

> in some cultures/languages (eg, Japanese) gender is completely absent, at least in professional communication, where one would always refer to another person by their (sur)name and the "-san" suffix (which is a genderless version of the Master/Mr/Mrs/Ms/Miss/Madam/Sir/etc prefix in english).

And this will be reflected in the Japanese translation.

> My point is, currently the game is genderless.

Currently the game is not genderless. It's just harder to deal with in the strings to make it translatable. I tried to hack a bit around grammar but it was only when I looked into how it was done by @robothauler in the FindPerson module that it became really doable in my mind.

> This makes it is less offensive and more approachable to current cultural norms and, IMHO also more "futuristic".

It doesn't make the game less offensive. It makes the strings a bit odd. The offensive part should be that you can accept a contract to murder someone against payment or trade in slaves. Do you think that should be changed too? It's not more "futuristic" to use gender neutral language. It rather cements the game in a very isolated ideological discourse.

I just want to make the strings flow more naturally, that's all.

data/lang/module-assassination/en.json
mwerle: Thanks; I'll comment on the new lines separately. Closing this comment.

data/modules/Assassination/Assassination.lua
```diff
@@ -64,6 +64,8 @@ end
 
 local onChat = function (form, ref, option)
 	local ad = ads[ref]
+	local target = ad.target
+	local female = ad.target.female and "_FEMALE" or "_MALE"
```
mwerle: Suggestion: it might be better to generate the suffix here once and then use that instead of moving the randomizer into each chat?

Having the randomizer in each chat means the text may change when the user clicks repeatedly. This may be desired, or not.

ie:

```suggestion
	local genderSuffix = (Engine.rand:Integer(0,2) < 2) and ad.target.female or ""
```

data/modules/Assassination/Assassination.lua
```diff
@@ -116,8 +119,8 @@ local onChat = function (form, ref, option)
 	elseif option == 2 then
 		local sbody = ad.location:GetSystemBody()
 
-		form:SetMessage(string.interp(l.IT_MUST_BE_DONE_AFTER, {
-		  target    = ad.target,
+		form:SetMessage(string.interp(l["IT_MUST_BE_DONE_AFTER" .. female], {
```
mwerle: This string did not get the randomizer to have a gender-neutral wording. Hence my suggestion for line 68. Then this line would become:


```suggestion
		form:SetMessage(string.interp(l["IT_MUST_BE_DONE_AFTER" .. genderSuffix], {
```

mwerle:
> Currently the game is not genderless. 

Sorry; my bad; I should have a) done more research and/or b) said "in this part of the game"..

> I just want to make the strings flow more naturally, that's all.

Partially my point was that what is "more natural" today may not be in the future. Which is, of course, impossible to predict. But having some grammatical or other differences does create a different tone from the here-and-now.

Anyway; no worries. I just gave my 2p.. it's not my call in either case, I'll concentrate on reviewing just the technicalities.

data/modules/Assassination/Assassination.lua
mwerle: .. actually re-looking at the code, it won't make a difference since `onChat` is called for every interaction. But the suffix-randomizer should get moved here anyway so that it doesn't have to be duplicated in every section of the chat.

data/modules/Assassination/Assassination.lua
zonkmachine: Yes, **target.gender** is more clear than **target.female**. I'll look into it.

mwerle: > Yes, **target.gender** is more clear than **target.female**. I'll look into it.

I was thinking about this more (while trying to fall asleep) and, as it is a boolean value, `gender` doesn't actually work here. You'd have to change it to an enum or similar, and that would be a significantly more intrusive code change.

But changing it to `target.isFemale` would improve code readability and understanding a lot and would keep code changes to a minimum - just a simple variable name refactor.

zonkmachine:
OK. I have addressed your suggestions in the latest fixup.

data/modules/Assassination/Assassination.lua
zonkmachine: Fixed. I went with just 'suffix'.

data/modules/Assassination/Assassination.lua
zonkmachine: Fixed!

ElonSatoshi:
In defense of using pronouns in assassination missions:

I think it would humanize the victim, and give the player a subtle reminder that they're about to go on a mission to murder a person. This would help with immersion/giving the player second thoughts about committing the act.

It would be good to make it more than just a boolean in the coding, though, because it wouldn't hurt to throw in they/them or any other pronoun (or lack of pronoun) that we feel we could acknowledge.

zonkmachine:
> I think it would humanize the victim, and give the player a subtle reminder that they're about to go on a mission to murder a person. This would help with immersion/giving the player second thoughts about committing the act.

I agree. I was thinking also adding the more stereotypical 'the object' in some strings and maybe that could be used by clients who are more professional killers. That would probably be a separate set of flavors.

> It would be good to make it more than just a boolean in the coding, though, because it wouldn't hurt to throw in they/them or any other pronoun (or lack of pronoun) that we feel we could acknowledge.

It would be a lot of work with resources we don't have. I think it's better to keep it simple and neutral. Just plain ordinary language. I would like to stay clear off anything that is religion and ideology of today. Maybe we could have revolutionary characters who use the phrase 'comrade', but I wouldn't want to go any further than that personally. It's just a game.

impaktor:
> It would be good to make it more than just a boolean in the coding, though, because it wouldn't hurt to throw in they/them or any other pronoun (or lack of pronoun) that we feel we could acknowledge.

I'd advice against it (read: 'no'). Far too hot button political.

zonkmachine:
Currently the strings glitch some. Example;

```json
  "FLAVOUR_3_INTROTEXT": {
    "description": "",
    "message": "The {target} must be reduced to space dust. I'll award you {cash} to do this."
  },
```

I made the tile and name separate so this will now be The "Anna Johanssone must be...". It will take some more work. Setting as draft for now.

zonkmachine:
> "**The Anna Johansson must be...**".

"**The General must be...**".

That would work.

zonkmachine:
> "**The General must be...**".

OK, that was an easy fix. I added the title in there for some variation and it works well. In English that is. I don't know how well this is going to work in other languages. An option is to use the name (full name, no title) and just remove "The " from the beginning. That would probably be a bit more easy on the translators. I think the first will work too though.

I also changed the button text of 'Where can I find {target}." to read "Where can I find the target". It's shorter (unless your name is "Li Li", I've met someone with that name), it looks better.

zonkmachine:
> > "**The General must be...**".
> 
> OK, that was an easy fix. I added the title in there for some variation and it works well. In English that is. I don't know how well this is going to work in other languages. An option is to use the name (full name, no title) and just remove "The " from the beginning. That would probably be a bit more easy on the translators. I think the first will work to though.

I went with the safe choice and used the name instead. Using the title in a text like that would take some more consideration. 

Fixed some missed strings. Squashed, tested and done.

Issue #6219, opened 2025-10-09
StationTrafficControl.lua - 'Leaving controlled space' message when starting in space


zonkmachine:
<img width="836" height="374" alt="Image" src="https://github.com/user-attachments/assets/87f876da-1fde-4b1e-85a0-e767b5542064" />

StationTrafficControl.lua assumes that we start docked to a station. This makes a 'Leaving controlled space' message pop up when starting a game set manually to, for instance, (sectorx, sectory, sectorz, 0, 0), which is unexpected.

https://github.com/pioneerspacesim/pioneer/blob/1dd9c03fdbc90356c301977fa5317697894c2e06/data/modules/StationTrafficControl/StationTrafficControl.lua#L16-L18

If you would change  the snippet above to:
```lua
local playerInControlledSpace = false
```

, the message would be gone in this case but would now appear if you undock from a station that you're docked with from the start.

### Observed behaviour
When manually setting a start point at (sectorx, sectory, sectorz, 0, 0), the 'Leaving controlled space' from **data/modules/StationTrafficControl/StationTrafficControl.lua** is triggered.

### Expected behaviour
No message from traffic control on starting in outside of  a controlled space.

### Steps to reproduce
* Manually change **/data/pigui/modules/new-game-window/class.lua** as follows and start from Mars (which now is a starting point in space in a close orbit around the Sun).
* Wait and see a message pop up after five seconds.
```lua
diff --git a/data/pigui/modules/new-game-window/class.lua b/data/pigui/modules/new-game-window/class.lua
index fb559618f..e95981ac2 100644
--- a/data/pigui/modules/new-game-window/class.lua
+++ b/data/pigui/modules/new-game-window/class.lua
@@ -27,7 +27,7 @@ local profileCombo = { items = {}, selected = 0 }
 StartVariants.register({
     name       = lui.START_AT_MARS,
     desc       = lui.START_AT_MARS_DESC,
-    location   = SystemPath.New(0,0,0,0,18),
+    location   = SystemPath.New(0,0,0,0,0),
     logmsg     = lui.START_LOG_ENTRY_1,
     shipType   = 'coronatrix',
     money      = 600,
```

__My pioneer version (and OS):__
master (1dd9c03fdbc90356c301977fa5317697894c2e06)

__My output.txt (required) and game save (optional, but recommended)__
[output.txt](https://github.com/user-attachments/files/22801920/output.txt)

PR #6220, opened 2025-10-09
Fix traffic control message when starting in space

zonkmachine:
Fixes https://github.com/pioneerspacesim/pioneer/issues/6219 by making sure playerInControlledSpace == false when starting in space.

I don't know if there are possible edge cases where this new algorithm wouldn't work. If for instance we start in an orbit around a star but we're still in a controlled space via a space station.

### DIFF
--- a/data/modules/StationTrafficControl/StationTrafficControl.lua
+++ b/data/modules/StationTrafficControl/StationTrafficControl.lua
@@ -13,9 +13,12 @@ local Lang = require 'Lang'
 local l = Lang.GetResource("module-stationtrafficcontrol")
 local ui = require 'pigui'
 
--- game starts with player docked with a station
--- otherwise, serialized data already has the value
-local playerInControlledSpace = true
+-- The game usually start at a station but may start in space by providing a system
+-- path, either via the command line or in the game menu via a custom start option.
+-- In the most common case in a new game, where we start at a station, we are in a
+-- controlled space and this is set in the onGameStart() function.
+-- In the case of a saved game, serialized data already has the value.
+local playerInControlledSpace = false
 
 local getNumberOfFlavours = function (str)
 	-- Returns the number of flavours of the given string (assuming first flavour has suffix '_1').
@@ -106,8 +109,12 @@ end
 local loaded_data
 
 local onGameStart = function ()
+	local station = Game.player:FindNearestTo("SPACESTATION")
 	if (loaded_data) then
 		playerInControlledSpace = loaded_data.playerInControlledSpace
+	elseif Game.player.flightState == "DOCKED" or
+			station and station:DistanceTo(Game.player) < station.lawEnforcedRange then
+		playerInControlledSpace = true
 	end
 	loaded_data = nil
 	Timer:CallEvery(5, notifyControlledSpace)


zonkmachine:
I've tested this some more and it seem to work fine.

zonkmachine:
Merge?

impaktor:
What happens if the player is spawned outside a station? They're not docked, but in controlled space? Or do we not do that?

zonkmachine:
> What happens if the player is spawned outside a station? They're not docked, but in controlled space? Or do we not do that?

Let's find out. Starting in orbit around Phobos:
`location   = SystemPath.New(0,0,0,0,21),`

phobostrafficcontroll

In these cases, what is the expected result? I'm kind of cool with an edge case like this but it could be solved with an extra test for proximity to a station.

zonkmachine:
I guess in the situation above 27 km from Phobos Base, you wouldn't expect to start with a message like that. I'll fix it.

zonkmachine:
> What happens if the player is spawned outside a station? They're not docked, but in controlled space? Or do we not do that?

Fixed
```lua
	elseif Game.player.flightState == "DOCKED" or
			station:DistanceTo(Game.player) < station.lawEnforcedRange then
		playerInControlledSpace = true
```

zonkmachine:
I've tested it some more for possible edge cases but I'm low on ideas. It now works.

_(Edit: Both on master and this PR)_ It doesn't seem to be especially defined what the behavior should be when two stations are really close to each other. You get the message from one when you enter the area and then a goodby from the other when you exit after having transited from the first one. You will not get any traffic messages when moving in between the however. This is not a great problem. You could perhaps make control zones of stations smaller when they overlap but that may be just unnecessary noise. Maybe a different message could be used. "Now leaving 'x', we hand you over to 'y'. Something like that.

zonkmachine:
Merge?

data/modules/StationTrafficControl/StationTrafficControl.lua
```diff
@@ -106,8 +109,12 @@ end
 local loaded_data
 
 local onGameStart = function ()
+	local station = Game.player:FindNearestTo("SPACESTATION")
```
impaktor: I assume this will never be nil, as can happen if the player is in hyperspace, but since we can't start the game in hyperspace, this should be fine, I assume.

impaktor (approved):
I've not tested it, but looks fine when I ran the proposed changes in the software on my wetware (brain), it did not explode.

data/modules/StationTrafficControl/StationTrafficControl.lua
zonkmachine: Well spotted. No, you just need to start in a system with no station and this code will crash. Sorry about that.

zonkmachine: Fix pushed

data/modules/StationTrafficControl/StationTrafficControl.lua
```diff
@@ -106,8 +109,12 @@ end
...
 	if (loaded_data) then
 		playerInControlledSpace = loaded_data.playerInControlledSpace
+	elseif Game.player.flightState == "DOCKED" or
+			station and station:DistanceTo(Game.player) < station.lawEnforcedRange then
```
zonkmachine, line 116: @impaktor I added **'station and'** here in response to your earlier comment on station being nil. I use no parenthesis here and trust in short-circuit evaluation. This is OK and clean right?

data/modules/StationTrafficControl/StationTrafficControl.lua
impaktor: well, I anti-spotted it, thinking it was not an issue. Två kloka huvuden, or whatever the saying is.

zonkmachine: Merge at will...

PR #6234, opened 2025-10-28
Advice module - Three new advice

zonkmachine:
**Advice module - Some new advice** 

**49 - "A clear way out"**
"When stringing multiple jumps together, queuing them before departure will save you both a headache and perhaps your life, depending on how welcome you will be during your route."

**194 - "Running on fumes"**
"Half of a ships mass when the tanks are full consists of fuel. If you're doing shorter runs, or if you know for certain that you will have enough margin, you may fill up the tanks to a lower degree. Lower mass means less fuel consumption and a faster ship. Just remember to keep an eye on the fuel level."

**203 - "The Billie Bounce detour"**
"When you find yourself in the outskirts of a system and need to return to the inner planets, a detour to a sufficiently close neighboring system might be the fastest way."

Forum thread here: https://forum.pioneerspacesim.net/viewtopic.php?f=3&t=771

### DIFF
--- a/data/lang/module-advice/en.json
+++ b/data/lang/module-advice/en.json
@@ -15,6 +15,30 @@
     "description": "A reference to the song 'The Harder They Come', by Jimmy Cliff",
     "message": "The harder the g, the harder the fall"
   },
+  "ADVICE_12_BODYTEXT": {
+    "description": "",
+    "message": "When you find yourself in the outskirts of a system and need to return to the inner planets, a detour to a sufficiently close neighbouring system might be the fastest way."
+  },
+  "ADVICE_12_HEADLINE": {
+    "description": "",
+    "message": "The Billie Bounce detour"
+  },
+  "ADVICE_13_BODYTEXT": {
+    "description": "",
+    "message": "Lower mass means less fuel consumption and a faster ship. Fuel has mass. For some routes, running light with less fuel in the tanks can save both time and money."
+  },
+  "ADVICE_13_HEADLINE": {
+    "description": "",
+    "message": "Running on fumes"
+  },
+  "ADVICE_14_BODYTEXT": {
+    "description": "",
+    "message": "When stringing multiple jumps together, queuing them before departure will save you both a headache and perhaps your life, depending on how welcome you will be during your route."
+  },
+  "ADVICE_14_HEADLINE": {
+    "description": "",
+    "message": "A clear way out"
+  },
   "ADVICE_1_BODYTEXT": {
     "description": "Point made is: the real illegal goods traders are fixed to each station, but police's fake illegal goods traders will be random every time.",
     "message": "As you return to a station, be wary of those alternative market places you do not recognize, and you will stay out of trouble with the law."
--- a/data/modules/Advice/Advice.lua
+++ b/data/modules/Advice/Advice.lua
@@ -33,6 +33,9 @@ local travellers_advice_indices = {481, -- tame black market
 								   52,  -- service ship
 								   248, -- change faction
 								   171, -- the harder the g
+								   203, -- the billie bounce
+								   194, -- running on fumes
+								   49,	-- a clear way out
 }
 
 -- Hold all different types of advice/rumours available:


data/lang/module-advice/en.json
```diff
@@ -15,6 +15,30 @@
...
+    "message": "When you find yourself in the outskirts of a system and need to return to the inner planets, a detour to a sufficiently close neighbouring system might be the fastest way."
+  },
+  "ADVICE_12_HEADLINE": {
+    "description": "",
+    "message": "The Billie Bounce detour"
+  },
+  "ADVICE_13_BODYTEXT": {
+    "description": "",
+    "message": "Half of a ships mass when the tanks are full consists of fuel. If you're doing shorter runs, or if you know for certain that you will have enough margin, you may fill up the tanks to a lower degree. Lower mass means less fuel consumption and a faster ship. Just remember to keep an eye on the fuel level."
+  },
```
impaktor, line 29: This was the one [I found](https://forum.pioneerspacesim.net/viewtopic.php?p=7378&sid=c8bdbb5ca978e55485a7a145f3f61416#p7378) "too wordy" and detailed.

You suggested this change to my suggestion:

> Mass in movement is time and money. Fuel is mass. For some routes running with less fuel in the tanks can save you both.

Another option:

> Lower mass means less fuel consumption and a faster ship. Fuel has mass. For some routes running with less fuel in the tanks can save both time and money.

impaktor (commented):
Sorry to harass you on that string...

data/lang/module-advice/en.json
zonkmachine: What, the original string is still there? I'll fix.

zonkmachine: Check again.

zonkmachine: Wait. It's a bit too condensed now. I'm looking into your last suggestion.

zonkmachine: Check again. Fixed.

zonkmachine:
@impaktor Went with your last suggestion. I'm happy with this. Merge at will!

impaktor:
I think we have a Swedish Chef-situation on our hands, can we please get some native speaker to double check the strings? @mwerle ? e.g.
- Is it "welcome" or "welcomed"? 
- is "running" the best word? (see strings)

mwerle:
> I think we have a Swedish Chef-situation on our hands, can we please get some native speaker to double check the strings? @mwerle ? e.g.
> 

Lol, I'm hardly a native speaker :)


>     * Is it "welcome" or "welcomed"?
> 

"welcome" is correct, although the entire sentence sounds somewhat odd.

"... depending on how much trouble you can expect during your route." flows better but might lose some flavour.

>     * is "running" the best word? (see strings)
>

"flying", "travelling"?  Also not sure I'd have "For some routes" because it applies to _all_ routes as long as you have sufficient fuel for the route.

"Burning" might also be appropriate since Pioneer ships are similar to Expanse; ie, assuming you have the fuel, the fastest way from A to B is to burn the entire way, flipping half-way.

"Running" might also be ok depending on what tone you're trying to go for. These strings are all somewhat comedic (at least to some people); eg, "Kessel Run" from Star Wars also uses "running". But in that case possibly flipping the sentence might make it sound more natural:
"Having less fuel in the tanks can save both time and money on certain runs."?

zonkmachine:
> is "running" the best word?

Yes. Yes it is. Maybe there should be a comma in there?
"For some routes, running with less fuel in the tanks can save both time and money."

mwerle:
> > is "running" the best word?
> 
> Yes. Yes it is. 

Any sources backing up that claim? I mean, I don't have a doctorate in linguistics, nor am I a native English speaker, but after over 40 years of primary experience in various dialects of English, including British, it doesn't sound "natural" as currently proposed. This applies to both the word under discussion in particular, as well as the entire sentence more generally.

But perhaps there is a joke or backstory to this which I am missing..? I did have to read up the reference to the "Billie(sp? - Billy) Bounce detour" which, to me, made zero sense whatsoever and the backstory to it is somewhat, shall we say, tenuous? But the game does need to build up its own lore so who am I to argue. Personally I'd have leant on the existing background (Elite et al) and perhaps come up with the "Jameson Jump" or something similar but the game does appear to have been distancing itself somewhat from its origins.

zonkmachine:
> Any sources backing up that claim?

There was a tongue in cheek emoji in there, I should perhaps have typed it out. I just like 'running' in there. It hints at 'trucking'. Also, it's called a Cargo Run. I tried to pit Grok against you but it refused to take sides. It came up with changing it to 'Running Light' as that hints to the aviation term for precisely what this advice is about. That makes sense actually.

> "Jameson Jump"

Not a bad idea at all but I like Billie. And it's alliteration too. Perhaps there are other maneuvers in there that needs naming? I never got to properly sit down and fly the original space trader games so we don't have the same references. I did have Elite on the Amiga at one time but didn't get too involved in it.

zonkmachine:
Would 'hauling' be a better option?

mwerle:
> Would 'hauling' be a better option?

It sounds like you're looking at it from a cargo trucking perspective - what about miners, bounty-hunters, passenger transports, etc? They also all fly ships, sometimes with deadlines..

Anyway, I was asked to give my 2p, which I did.. take it or leave it, I'm not here to argue .. :)

zonkmachine:
> It sounds like you're looking at it from a cargo trucking perspective

Well, you may be right. I'm once more back to "running". I really like it but I usually don't fight very hard for things. I may very well let myself be convinced to drop it. "Running an errand" is maybe more general than hauling and that could be shorted to 'running'. I like it so much I really thought I came up with it myself but when I read through the forum post it turned out I actually didn't. :truck:

impaktor:
Everyone happy, or is there some additional tweak someone wants to see?

zonkmachine:
Happy.

zonkmachine:
> But perhaps there is a joke or backstory to this which I am missing..? I did have to read up the reference to the "Billie(sp? - Billy) Bounce detour" which, to me, made zero sense whatsoever and the backstory to it is somewhat, shall we say, tenuous? But the game does need to build up its own lore so who am I to argue. Personally I'd have leant on the existing background (Elite et al) and perhaps come up with the "Jameson Jump" or something similar but the game does appear to have been distancing itself somewhat from its origins.

I should have listened to you more closely. Today I feel pretty much the same. It's also pretty impossible to translate. Maybe a change here wouldn't be a bad idea? I like the hyper space jump mechanics of it though. I think that part adds value.

impaktor:
> I should have listened to you more closely. Today I feel pretty much the same. 

Feel free to tweak. No shame in that. Through iteration we push the game to be better bit by bit (hopefully)

zonkmachine:
OK. Currently it reads: "The Billie Bounce detour"

Suggestions from previous forum thread:
**"The faster detour"**
**"Bounce in"** - To me, this is the most descriptive. The net sum of the maneuver is that you end up in the inner parts of it. You bounce in.
**"Bouncing off a system"** - Also descriptive.

I still like the term **bounce**. Anyone else has a preference?

sturnclaw:
> I still like the term **bounce**. Anyone else has a preference?

I'd like to propose "Slingshot Maneuver". You're technically "slingshotting" around/past/through another star's gravity well to avoid an in-system hyperspace jump. IMO "bounce" is a little too vague or casual sounding.

zonkmachine:
> I'd like to propose "Slingshot Maneuver".

Strong contender. It sounds good but slingshot is used already. It's another word for gravity assist.

mikehgentry:
Ricochet? (or 'ricochet maneuvre')

Or some kind of compound term maybe

Hyper-bounce
Bounce-jump

zonkmachine:
> Ricochet? (or 'ricochet maneuvre')

I like Ricochet. It should be used for something but here it suffers from the same problems as bounce. It's a bit on the enigmatic side.

Issue #6225, opened 2025-10-19
Station staff - Issues with faces


zonkmachine:
Various issues with faces of the station staff. Sometimes disappearing and the faces are random on each dock/landing, not saved. 

<img width="821" height="461" alt="Image" src="https://github.com/user-attachments/assets/6e3f9806-ef05-43a9-9b05-f5b8ecf3d58e" />

<img width="821" height="462" alt="Image" src="https://github.com/user-attachments/assets/e1e23120-15ae-412b-a00c-f3a3a55a0f05" />

 - [x] Torvalds station - Seeded with 0. This makes the Lobby skip the face generation. Fixed by setting stationSeed to 'false' instead of 0. Torvalds station seem to be the only station in the known universe that is seeded with 0.
1 - Fixed in https://github.com/pioneerspacesim/pioneer/pull/6224
2 - Torvalds seed fixed in https://github.com/pioneerspacesim/pioneer/pull/6222
 - [ ] Other cases of staff disappearing. Both Lobby and Mechanic recorded on unscheduled leave on other stations. Cause uncertain. (see images).
 - [ ] The faces of the stations aren't fixed. Are they supposed to be? Seeding with a fixed value still gives different output when a station is created.

### Observed behaviour
As illustrated in the images above, faces sometimes disappear. Going away and coming back can sometimes make the staff come back. This has been seen in the lobby but also at the mechanics. 

### Expected behaviour
No breaks for station personnel

### Steps to reproduce
No steps to reproduce. Intermittent. Just travel the universe and make sure to go through the staff when visiting a station.

__My pioneer version (and OS):__
Pioneer versions (20250203 - mechanic image)
master (1dd9c03fdbc90356c301977fa5317697894c2e06) (missing station manager image)

__My output.txt (required) and game save (optional, but recommended)__

Output and game save from the missing mechanic on the 20250203 release:
[output.txt](https://github.com/user-attachments/files/22993135/output.txt)
[missingmechanic.txt](https://github.com/user-attachments/files/22993138/missingmechanic.txt)

sturnclaw:
@zonkmachine have you been able to reproduce this after #6236 was merged?

zonkmachine:
> [@zonkmachine](https://github.com/zonkmachine) have you been able to reproduce this after [#6236](https://github.com/pioneerspacesim/pioneer/pull/6236) was merged?

[...]

PR #6236, opened 2025-10-29
only set stationSeed from one place

zonkmachine:
In response to #6225
I'm not sure it's a fix yet but 'stationSeed' shouldn't be set on two different places. If it's set via **determinePaintshopAvailability()** then, when either of **refresh()**, **drawPaintshop()** or **drawShipRepair()** is executed and there are no face, the face will not be generated. It's not obvious to me yet when **refresh()** is called. In any case, this is only a proplem on bodies with smaller populations.
#6225 states that the missing faces are on bot the lobby and the mechanic and this would not be explained by this find/fix. However, I'm unsure if I have tested this correctly as my local test branch I used to test this bug with debug statements accidentally had a fix built in, so it essentially sent me to intermittent hell with the issue suddenly gone. We must assume the possibility that the rest of my trouble shooting lived up to the same quality.

A draft for now.

### DIFF
--- a/data/pigui/modules/station-view/06-shipRepairs.lua
+++ b/data/pigui/modules/station-view/06-shipRepairs.lua
@@ -105,7 +105,6 @@ local function determinePaintshopAvailability()
 	-- high population stations often have them
 	local pop = station:GetSystemBody().population
 	if pop > 0.00005 then -- Mars is about 0.0002
-		stationSeed = station.seed
 		local rand = Rand.New(station.seed .. '-paintshop')
 		if rand:Number(0,1) < 0.75 then
 			return true


zonkmachine:
I have found no way to faithfully reproduce the issues but one way to make the mechanic to disappear with some success rate seem to be:

* Go to Phobos
* Visit the Repair Shop
* Enter the Paint Shop
* Close Pioneer (Alt + F4)
* Resume game
* Visit Repair Shop which now most likely don't have the staff present

sturnclaw:
Ehhh... You might be better served refactoring the module to remove the dependence on the file-scope global stationSeed variable and instead pass it as a direct dependency to the various functions. That kind of global state is a huge bug vector.

Haven't dug much further into the module, but you should be safe to assume that station.seed is valid (though maybe not always nonzero) as I recall improving the consistency of seed generation in previous system-gen PRs.

zonkmachine:
>... refactoring the module to remove the dependence on the file-scope global stationSeed variable and instead pass it as a direct dependency to the various functions.

I'm just starting to get confident with the mission modules and the views have a different 'feel' to them. I'm pretty lost here right now. Spontaneously, I could maybe touch up the Issue instead and focus on some other part of the code that I know better. A tactical retreat...

sturnclaw:
Having taken a closer look at the module, this is a good change on the face of it.

Refresh is called when switching to the specific view tab within the station view - it should also be called when the station view is opened with that tab already active (though don't quote me on that). The view table specifies a draw function which renders the module (probably drawPaintshop) and the refresh function is meant to be used to refresh state that may have changed while the player is on other views.

zonkmachine:
@sturnclaw I don't know what to do here. The way I understand the code, the stationSeed shouldn't be set in determinePaintshopAvailability() and this PR fixes that. It tests fine on my side. I can reproduce the issue with high confidence and the fix works for me 100%. Anything more advanced, well, probably someone else should take over the 

Merge/Close/Someone else takes over?

sturnclaw:
You are correct, setting `stationSeed` here is invalid regardless of the context of any other issues with the module. I'll merge this and we can address the general module design at a later date (read: probably never).

PR #6237, opened 2025-10-30
Fix CMake OpenGL binding for MSVC

fluffyfreak:
The recent #6050 PR to CMakeLists.txt broke the MSVC version but thanks to @pcercuei this works.


impaktor:
What's the review process here, since I think it's mainly you who have MSVC? Or might this have a negative effect on other build-systems?
EDIT: Point being, I could leave it to you to merge (preferably use the rebase to master option in the GUI here below)

fluffyfreak:
There's a few who use MSVC right? And it's one of the `build-msvc` checks.
But yeah I can just merge it since it seems to fix the problem

PR #6238, opened 2025-10-31
Allow custom systems to override randomly-generated ones

sturnclaw:
As the headline says, custom systems can now elect to override randomly-generated systems. This option is turned on by default when picking a randomly-generated system with the Edit Selected button in the System Editor.

I've also fixed the bug that prevented opening Rondel.lua; now when opening a system with fully random contents, we side-trip through the galaxy generator and load the fully-generated version of that system.

I've added buttons to the system editor timeline controls to preview the system at epoch (important when setting the various phase-at-epoch parameters) as well as at the game-start time if you were to start a new game in that system.

Exported systems from the System Editor are also now always terminated with a trailing newline character.

A few misc fixes came along for the ride:

- ~~Fix a read-past-end crash bug in ConfigParser when file data is followed by a digit character in uninitialized memory.~~ Split into separate PR.
- Cleanup galaxy generation code and fix minor bugs (cherry-picked from another branch).

CC @zonkmachine for testing.

### DIFF
--- a/src/galaxy/CustomSystem.cpp
+++ b/src/galaxy/CustomSystem.cpp
@@ -780,22 +780,6 @@ void CustomSystemsDatabase::Load()
 
 	LoadAllLuaSystems();
 
-	// Load Json array files containing random-fill system definitions
-	std::string partialPath = FileSystem::JoinPathBelow(m_customSysDirectory, "partial");
-	for (auto &file : FileSystem::gameDataFiles.Recurse(partialPath)) {
-		if (!ends_with_ci(file.GetPath(), ".json"))
-			continue;
-
-		PROFILE_SCOPED_DESC("Load Partial System List")
-		const Json fileData = JsonUtils::LoadJsonDataFile(file.GetPath());
-		for (const Json &sysdef : fileData) {
-			if (!sysdef.is_object())
-				continue;
-
-			LoadSystemFromJSON(file.GetPath(), sysdef);
-		}
-	}
-
 	// Load top-level custom system defines
 	for (auto &file : FileSystem::gameDataFiles.Enumerate(m_customSysDirectory, 0)) {
 		if (!ends_with_ci(file.GetPath(), ".json"))
@@ -811,6 +795,22 @@ void CustomSystemsDatabase::Load()
 			continue;
 
 		LoadSystemFromJSON(file.GetPath(), JsonUtils::LoadJsonDataFile(file.GetPath()));
+	}
+
+	// Load Json array files containing random-fill system definitions
+	std::string partialPath = FileSystem::JoinPathBelow(m_customSysDirectory, "partial");
+	for (auto &file : FileSystem::gameDataFiles.Recurse(partialPath)) {
+		if (!ends_with_ci(file.GetPath(), ".json"))
+			continue;
+
+		PROFILE_SCOPED_DESC("Load Partial System List")
+		const Json fileData = JsonUtils::LoadJsonDataFile(file.GetPath());
+		for (const Json &sysdef : fileData) {
+			if (!sysdef.is_object())
+				continue;
+
+			LoadSystemFromJSON(file.GetPath(), sysdef);
+		}
 	}
 }
 


zonkmachine:
> As the headline says, custom systems can now elect to override randomly-generated systems. This option is turned on by default when picking a randomly-generated system with the Edit Selected button in the System Editor.

I can't test this as I get a crash when trying to opening the file 02_local_stars.json

```sh
Info: OpenFile: /home/zonkmachine/builds/pioneer/data/systems/partial/02_local_stars.json
terminate called after throwing an instance of 'nlohmann::detail::type_error'
  what():  [json.exception.type_error.305] cannot use operator[] with array
Aborted (core dumped)
```

_Edit: No output.txt for the editors it seems._


```gdb
Info: OpenFile: /home/zonkmachine/builds/pioneer/data/systems/partial/02_local_stars.json
terminate called after throwing an instance of 'nlohmann::detail::type_error'
  what():  [json.exception.type_error.305] cannot use operator[] with array

Thread 1 "editor" received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
warning: 44	./nptl/pthread_kill.c: No such file or directory
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff6e4527e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff6e288ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff72a5ff5 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff72bb0da in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff72a5a55 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007ffff72bb391 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#9  0x00005555556a0456 in nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer>::operator[]<char const> (
    this=this@entry=0x7fffffffd5e0, key=key@entry=0x555555a95937 "name")
    at /home/zonkmachine/builds/pioneer/contrib/json/json.hpp:14416
#10 0x000055555571e247 in CustomSystem::LoadFromJson (this=this@entry=0x555556d2c9b0, systemdef=...)
    at /home/zonkmachine/builds/pioneer/src/galaxy/CustomSystem.cpp:638
#11 0x000055555571f721 in CustomSystemsDatabase::LoadSystemFromJSON (this=this@entry=0x555556a8ae60, filename=..., systemdef=..., 
    mergeWithGalaxy=mergeWithGalaxy@entry=true) at /home/zonkmachine/builds/pioneer/src/galaxy/CustomSystem.cpp:863
#12 0x00005555556077be in Editor::SystemEditor::LoadSystemFromFile (this=this@entry=0x555556daacd0, file=...)
    at /usr/include/c++/13/string_view:146
#13 0x00005555556090a2 in Editor::SystemEditor::LoadSystemFromDisk (this=this@entry=0x555556daacd0, 
    absolutePath="/home/zonkmachine/builds/pioneer/data/systems/partial/02_local_stars.json")
    at /home/zonkmachine/builds/pioneer/src/editor/system/SystemEditor.cpp:188
#14 0x0000555555609561 in Editor::SystemEditor::Update (this=0x555556daacd0, deltaTime=<optimized out>)
    at /home/zonkmachine/builds/pioneer/src/editor/system/SystemEditor.cpp:700
#15 0x0000555555636f4f in Application::Run (this=this@entry=0x555555bcc5a0 <Editor::EditorApp::Get()::app>)
    at /home/zonkmachine/builds/pioneer/src/core/Application.cpp:193
#16 0x00005555555e236b in main (argc=<optimized out>, argv=<optimized out>)
    at /home/zonkmachine/builds/pioneer/src/editor/editormain.cpp:19

```

zonkmachine:
> I've also fixed the bug that prevented opening Rondel.lua; now when opening a system with fully random contents, we side-trip through the galaxy generator and load the fully-generated version of that system.

Fix confirmed



> I've added buttons to the system editor timeline controls to preview the system at epoch (important when setting the various phase-at-epoch parameters) as well as at the game-start time if you were to start a new game in that system.

Nice. Tested, works like a charm!


> Exported systems from the System Editor are also now always terminated with a trailing newline character.

Tested. Opened 01_werribee.lua and saved it. No bad messages. :+1:

zonkmachine:
I got a crash log for you but with no symbols I'm afraid. (edited first comment) Hope it helps.

impaktor:
Where do we stand on renaming the editor executable to avoid nameclash?

sturnclaw:
@zonkmachine Partials files cannot be edited by the system editor directly, as they contain a list of stars rather than a single full custom system definition. You should use the "New System" button in the file menu, navigate to the appropriate sector, select the star system from the list and click Edit Selected.

@impaktor: not tackling that in this PR yet. I'll address it along with separate icons in the next PR hopefully.

sturnclaw:
Note that systems defined via 02 local stars and 03 bright stars are overriden by name resolution; the index-based approach only works for overriding systems created by the random generator. (And thus can rename those systems.)

zonkmachine:
I'm unsure what the override is supposed to do. What does the number stand for? I've gotten quite a few, 'does not exist'. Is this supposed to be used in the custom system upgrade work?

sturnclaw:
> I'm unsure what the override is supposed to do. What does the number stand for? I've gotten quite a few, 'does not exist'. Is this supposed to be used in the custom system upgrade work?

Explanatory scenario:

1. You have a sector with two randomly generated systems in it, Random A (index 0) and Random B (index 1).
2. You create a new custom system as normal before this PR and place it in that sector. It will now contain three systems, Custom A (index 0), Random A (index 1), and Random B (index 2).
3. You create a new custom system, enable override, and set the override index to 2. You place it in that sector. That sector still contains three systems, but they're now Custom A (0), Random A (1), and Custom B (2).

An overriding custom system that refers to a non-random or non-existent index will fall back to being treated as though it overrode nothing, i.e. it is appended to the list of systems in the sector before random systems are generated.

zonkmachine:
> An overriding custom system that refers to a non-random or non-existent index will fall back to being treated as though it overrode nothing, i.e. it is appended to the list of systems in the sector before random systems are generated.

So if a later version of the system, when we change something to the randomness and the universe changes, the system will still pop into place somewhere?

Q - There are two types of custom systems. The ones that are described entirely and the ones that only contains a star and is presented on a list. Maybe three types? Rondel.lua is somewhere in between). When you talk about 'custom system', are we always talking of both of these equally? The ones that pop up as 'Is custom: Yes'?

src/core/ConfigParser.cpp
```diff
@@ -82,9 +83,8 @@ size_t Tokenizer::default_classifier(Tokenizer *t, std::string_view remaining, T
 	// a Number token type means there's an optional '-' and at least one leading digit;
 	// parse its value as an optionally-floating point number
 	if (out.type == Token::Number) {
-		char *end = nullptr;
-		out.value = std::strtod(remaining.data(), &end);
-		length = std::distance<const char *>(remaining.data(), end);
+		const char *end = fast_float::from_chars(remaining.data(), remaining.data() + remaining.size(), out.value).ptr;
```
fluffyfreak: I'd have preferred this as a separate PR for the fix instead of bundled into this one

AUTHORS.txt
```diff
@@ -225,6 +225,11 @@ Pioneer includes the following third-party software:
   Copyright 2014 Melissa O'Neill <oneill@pcg-random.org>
   Licensed under the Apache License, Version 2.0
 
+  fast_float
```
fluffyfreak: No way to do this without bringing in a another thirdparty library?

sturnclaw: No. To my knowledge, no standard library method (even including istringstream) allows reading from a size-delimited rather than null-terminated string - and especially without copying the file data to an owned buffer! Additionally, fast_float is faster than the standard library methods and cross-platform deterministic; i.e. the same input string will give the same bit-exact value on any hardware with IEEE754 semantics.

I could not have used the JSON parser's string-to-double parsing implementation, as that's buried in its lexer and just calls std::strtod. I admittedly only confirmed this just now.

I don't necessarily consider this a bad dependency to add, though I do lament that we have to pull in a dependency just to get a non-null string parsed.

src/core/ConfigParser.cpp
sturnclaw: I'll split it out tomorrow into a new PR. I wanted to get this branch uploaded before I ran out of steam, but it's definitely not on-topic.

PR #6228, opened 2025-10-25
Codedoc cleanup

zonkmachine:
Some clean-up of codedoc. Touches only the files that had the Availability and Status tags.
* Remove **Availability** and **Status**
* Spelling
* Add missing crimes to Legal.lua codedoc comment

Mentioned [here](https://github.com/pioneerspacesim/pioneer/pull/6114#discussion_r2047053907) that the Availability/Status tags are deprecated-ish. Sort of. I was tired and managed to get NaturalDocs working so I had a go and removed thins and I think It looks better. I also fixed up mostly spelling in those files touched by the first commit.

PR #6242, opened 2025-11-04
Codedoc cleanup 2

zonkmachine:
Continuation of #6228. Remove **Availability** and **Status** from the codedocs on the C++ side.


sturnclaw (approved):
Woot! Next thing you know, we'll be dropping releases entirely and only offering nightly builds of the game! (joking)

Thank you for tackling this! Those fields are pretty much just noise; at best we might want to maintain two "mirrors" of the codedoc site, one for `master` and one for `release` if there's a major concern about being able to differentiate between documentation for the release build vs. not-yet-released features.

PR #6244, opened 2025-11-05
Only add systems once in GetNearbyStarSystemsByName

### LINKED ISSUE #6240: System search shows repeted matches for same system. And does not show matched other name.
### Observed behaviour

Image

You can see 2 examples of the search matching the system multipule times.

Image

Maybe the name that is matched should be shown.

Image

Or only the main name of the system and it should be match at most once.

Image

### Expected behavior

The search should match each system only once. 

The matched name should be shown as the result.

### Steps to reproduce

Load a system with multiple other names. (each with a space ' '... or any other common substring.)

navigate to nearby in the map and search for ' ' 

the system will be listed multiple times

Debian12

https://github.com/pioneerspacesim/pioneer/commit/0f2d4dea92a9acd6e110226906b5ac2d7acb6e25

fluffyfreak:
This partially fixes #6240 by eliminating the adding of duplicates.

It was adding, incorrectly I feel, the SystemPath per matching name or `otherName` entry.
This was then looked up in the UI and always displaying the main system name for each entry resulting in the bug.

This could be easily repeaed by starting a new game and just searching for a blank space ` `

### DIFF
--- a/src/SectorMap.cpp
+++ b/src/SectorMap.cpp
@@ -1276,15 +1276,18 @@ std::vector<SystemPath> SectorMap::GetNearbyStarSystemsByName(std::string patter
 				SystemPath match((*i).first);
 				match.systemIndex = systemIndex;
 				result.push_back(match);
-			}
-			// now also check other names of this system, if there are any
-			for (const std::string &other_name : ss->GetOtherNames()) {
-				if (strncasecmp(pattern.c_str(), other_name.c_str(), pattern.size()) == 0
-					// look for the pattern term somewhere within the current system
-					|| pi_strcasestr(other_name.c_str(), pattern.c_str())) {
-					SystemPath match((*i).first);
-					match.systemIndex = systemIndex;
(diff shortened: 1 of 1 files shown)


fluffyfreak:
I can't really help with the displaying of the `otherName` entries as that's in Lua but it looks like that code should be in `map-sector-view.lua` function `drawSearchResults`. Perhaps if there are `otherName` entries it could loop through and display them like `(aka: xxx, yyy, zzz)` or something. Although space might be an issue.

Maybe best as hover text 🤷‍♂️

zonkmachine:
I gave the issue of alternative names of stars a thought but I have no solution for it. One idea was that you could add a colored asterisk after the search entry if it has more than one alternative name. That would at least work today when there are a limited number of stars/systems that has more than one name.

zonkmachine (approved):

fluffyfreak:
@zonkmachine I think the `*` option would be good for now 👍

PR #6248, opened 2025-11-10
Increase the probability of being intercepted by pirates

robothauler:
In my opinion, being intercepted by pirates happens far too rarely. I experimented with different distances from the star, but the small ships often ran out of fuel. It worked best when the pirates orbited the populated planets and approached the player head-on at close range (I choose 0,1 AU).

I tested this in 36 Ophiuchi. Local transport is now also more risky.

### DIFF
--- a/data/modules/Pirates.lua
+++ b/data/modules/Pirates.lua
@@ -6,42 +6,162 @@ local Game = require 'Game'
 local Event = require 'Event'
 local ShipDef = require 'ShipDef'
 local utils = require 'utils'
+local Timer = require 'Timer'
+local Serializer = require 'Serializer'
 
 local MissionUtils = require 'modules.MissionUtils'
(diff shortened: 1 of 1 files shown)


impaktor:
This has been given thumbs up for merge (IRC), conditioned on if we can get someone to do some play testing of it.
Perhaps interesting to: @zonkmachine, @DynV, @Bodasey

sturnclaw:
Did a quick interplanetary cargo mission in Sol, didn't see any pirates (working as intended). This gets my thumbs-up, and I'll hack on it further once it's in master.

As an aside, the concept of pirates orbiting *destinations* in ambush is a really nice one and something that I think would arise naturally - unless you have a higher acceleration than the target, physics dictates that you will never catch a target in straight pursuit... but when the target is decelerating in your direction they're both a very "bright" signal source (from the drive plume) and already fighting their velocity and in the perfect phase for an ambush.

Issue #6201, opened 2025-08-05
Mapping missions don't get the status "failed"


Bodasey:
### Observed behaviour

Mapping missions don't get the status "failed" when the due time is missed and the money is paid nevertheless.

<img width="1282" height="749" alt="Image" src="https://github.com/user-attachments/assets/8e0acb16-5ca9-4f26-a5e3-527827605e9f" />
<img width="1282" height="749" alt="Image" src="https://github.com/user-attachments/assets/24604e3b-8110-4a32-a27d-9a8782c27383" />

zonkmachine:
<img width="1282" height="749" alt="Image" src="https://github.com/user-attachments/assets/8e0acb16-5ca9-4f26-a5e3-527827605e9f" />
There are a number of issues described here. In picture 1 (top register) a mission is registered as 'COMPLETED' although the timer is over due. The mission status is updated when onEnterSystem or onPlayerDocked is run. You would need to implement some further tests along the way in order for the status to update when you're just cruising. Maybe there could be a new optional function you could add in a module to allow for code to be run when the mission tab is opened?

In the case of **Scout.lua**, it would look something like this:

```lua
local onMissionUpdate = function (mission)
  if mission.status == "ACTIVE" and Game.time > mission.due then
    mission.status = "FAILED"
  end
end
```

And the module would be registered with:

```lua
Mission.RegisterType('Scout', l.MAPPING, buildMissionDescription, onMissionUpdate)
```
Then **onMissionUpdate** can be called, like buildMissionDescription, when the Mission list is opened.

As for picture 2, this is more of a bug. I haven't tested this but if I read the code correctly there are three possible outcomes from a Scout mission.
1. fail
2. success
3. mission completed but a bit late.

I think number 3 would apply to the example of the late but completed mission. There seem to be no dedicated strings for when the mission is completed but just a bit late. The outcome would have been a chastising message but still payment albeit a bit lower. There should probably be separate strings in this case to avoid confusion.

Here is the code that acts on a completed but slightly late scan:
https://github.com/pioneerspacesim/pioneer/blob/e6d74489091b26f791e90b4f6e999a80dce2fba5/data/modules/Scout/Scout.lua#L696-L700

What should have happen is the following
1. You get some money
2. flavour.failmsg (nag)
3. Reputation goes down by one

This doesn't seem to have happened by judging from the grateful message in picture 2 alone.

zonkmachine:
[...]

Issue #6249, opened 2025-11-11
Find Person module - Issues and updates


zonkmachine:
The Find Person module was added in #4746. It works really well but may need some nip/tuck. The following are suggestions most of which have been proposed already in the original PR thread (#4746). I think the five first suggestions below should be considered ahead of the next release 2026-02-03.

 - [ ] Translations are being done but aren't pulled back from Transifex with the rest of the strings. @impaktor 

 - [ ] Icons. Find Person uses the default icon at the moment. Idea to reuse other missions icons has been suggested. 

 - [ ] Repetitious, too many ads. On a new game start in Cydonia you can see as much as eight missions, and two of them being an incident about a malfunctioning delivery drone destroying furniture. There are currently 12 different story lines available. I suggest half the number of adverts would be a reasonable change.

 - [ ] To few missions. Write more strings. Will also help alleviate the issue above.

 - [ ] When a wanted person has been found and has boarded the ship for transfer home, there is nothing in the mission description that indicates that the mission has a passenger on board (apart from there being an occupied cabin).

---

 - [ ] The character name now is split in first name/surname, with new methods to use them, to allow some variation in how you address the characters involved. Instead of "He/She's a nut" you could go "Johnson is a nut". If it's about a friend, maybe there is a use case for the firstname. See
#6114

 - [ ] The Police station could have a list of wanted persons or could give info on people who have done crime and the wanted person could be on this list, literally wanted.

 - [ ] Maybe a search object has been killed in action or maybe they are just gone and can't be found? Maybe they have moved to another system?

 - [ ] Variation to existing missions. Instead of fixed roles, (stepsister), make some random character based on family/organisation structure. Maybe extend Character/Names with a function to just give you a random family member (cousin, uncle,...) or corporate member (colleague, supervisor,...).

@robothauler

zonkmachine:
https://github.com/pioneerspacesim/pioneer/blob/86a10a8be1b3289a8ab2ed91421a63716559b7fc/data/lang/module-findperson/en.json#L250-L257

The title could be changed to 'EXTRACTION' and the icon to 'combat': <img width="16" height="16" alt="Image" src="https://github.com/user-attachments/assets/16b05bd2-2c7e-42c4-97b3-94f5cd7328e0" />

zonkmachine:
[...]

PR #6250, opened 2025-11-16
Small improvements and a fix

robothauler:
Belongs to #6249 and #6201 
FindPerson adverts should now between 0 and 3.


impaktor:
Looks good to me.

Please prefix the issue tag in your post above with 'fixes #6201' if it fixes it, which it looks like it does, though I've not read the full issue report. That way we get an auto-close on merge.

robothauler:
> Please prefix the issue tag in your post above with 'fixes #6201' if it fixes it, ...

This fixes only the reduced payment for late missions. The problem with the status not being updated still persists.

zonkmachine:
Looks fine. I don't have time to test the scan fix but it looks good. I suggest just merge and maybe @Bodasey can have a go at testing it?

PR #6252, opened 2025-11-19
Correctly capitalize names of goods

zonkmachine:
cargocapitalized

* Make cargo types and commodities lower case by default.
* Extend string with function **capitalizeSentences()**. It finds the first letter in every sentence in a string and makes it upper case.
* Use **capitalizeSentences** in **CargoRun/Commodity Market**.

---

**Q/A - string.capitalizeSentences:**
* This part is almost verbatim the work of grok.
```lua
    s = s:gsub("([.!?]%s*)(%l)", function(punct, letter)
        return punct .. letter:upper()
    end)
```

* I placed the function in the top with string.trim. Maybe the string functions should be grouped together?

### DIFF
--- a/data/lang/commodity/en.json
+++ b/data/lang/commodity/en.json
@@ -1,7 +1,7 @@
 {
   "AIR_PROCESSORS": {
     "description": "",
-    "message": "Air processors"
+    "message": "air processors"
   },
   "AIR_PROCESSORS_DESCRIPTION": {
     "description": "This is an artistic description of the commodity. Translate as you deem appropriate or come up with your own description, long or short, its up to you.",
@@ -9,7 +9,7 @@
   },
   "ANIMAL_MEAT": {
     "description": "",
-    "message": "Animal meat"
+    "message": "animal meat"
   },
   "ANIMAL_MEAT_DESCRIPTION": {
     "description": "This is an artistic description of the commodity. Translate as you deem appropriate or come up with your own description, long or short, its up to you.",
@@ -17,7 +17,7 @@
   },
   "BATTLE_WEAPONS": {
     "description": "",
-    "message": "Battle weapons"
+    "message": "battle weapons"
   },
   "BATTLE_WEAPONS_DESCRIPTION": {
     "description": "This is an artistic description of the commodity. Translate as you deem appropriate or come up with your own description, long or short, its up to you.",
@@ -25,7 +25,7 @@
   },
   "CARBON_ORE": {
     "description": "",
-    "message": "Carbon ore"
+    "message": "carbon ore"
   },
   "CARBON_ORE_DESCRIPTION": {
     "description": "This is an artistic description of the commodity. Translate as you deem appropriate or come up with your own description, long or short, its up to you.",
@@ -33,7 +33,7 @@
   },
   "COMPUTERS": {
     "description": "",
-    "message": "Computers"
+    "message": "computers"
   },
   "COMPUTERS_DESCRIPTION": {
     "description": "This is an artistic description of the commodity. Translate as you deem appropriate or come up with your own description, long or short, its up to you.",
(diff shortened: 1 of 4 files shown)


zonkmachine:
Also fixed up the Commodity Market, Goods Trader, and News Event.

zonkmachine:
You could achieve the same by using lower case on the commodity strings instead and not laying it on the translators to fix a lot of strings. These are easy changes though and I think it reflects language best. You start with lower case words and upper case it when needed.

data/libs/autoload.lua
```diff
@@ -8,6 +8,16 @@ string.trim = function(s)
 	return string.gsub(s or "", "^%s*(.-)%s*$", "%1")
 end
 
+string.capitalizeSentences = function (s)
```
sturnclaw: Note that Lua APIs are invariably one or two words with no capitalization or underscores. Thus this should be `string.capitalize` or `string.sentencecase`. I'd prefer the latter, as the former should be reserved for a function which capitalizes the leading letters of all words in the string.

(If you can find a clear name that is even shorter, it would be preferred!)

zonkmachine: OK. string.sentencecase then. Maybe shorten it to string.sentcase?

sturnclaw: I don't think that `sentcase` has any semantic meaning. If we're going for abbreviation, I'd prefer `scase` or `scaps`, as that at least indicates to the user that the full meaning is not meant to be conveyed by the name, similar to e.g. `gsub` for `global-substitution`.

zonkmachine:
> how are commodity names that are Proper Nouns handled?

Can you give an example? If we sell 'Sturnclaws', then they will be capitalized where ever you put it. iPhones we can't do with this method.

data/libs/autoload.lua
zonkmachine: string.scase works for me.

sturnclaw:
> > how are commodity names that are Proper Nouns handled?
> 
> Can you give an example? If we sell 'Sturnclaws', then they will be capitalized where ever you put it. iPhones we can't do with this method.

If we were, for example, to sell "Rondellian Rondels", or perhaps "Bathlight Trinkets" akin to Elite's Rare Commodities.

zonkmachine:
> If we were, for example, to sell "Rondellian Rondels", or perhaps "Bathlight Trinkets" akin to Elite's Rare Commodities.

They would work, yes. I had to look up 'idempotent', but they would work fine.

data/libs/autoload.lua
```diff
@@ -8,6 +8,16 @@ string.trim = function(s)
 	return string.gsub(s or "", "^%s*(.-)%s*$", "%1")
 end
 
+string.scase = function (s)
```
sturnclaw, line 13: This function needs a short documentation string indicating its function, since the name is no longer descriptive.

```suggestion
-- Convert the input string to s(entence) case, capitalizing the
-- first letter of the string and any first letters of consecutive
-- sentences in the string.
string.scase = function (s)
```

(EDIT: manual word-wrapping in the Github interface is very difficult to determine...)

zonkmachine:
Is it rebase and merge? Will that put the three commits on top of master? I could just squash it to one commit as I usually do but I kept them separate along the way. I'm merge agnostic.

impaktor:
> Is` it rebase and merge? Will that put the three commits on top of master?

Yes.

I'd keep the commits separate as they are.
Merge commit could be fine (with me at least) here, as the three commits belong to the same theme.
Either way works, doesn't matter much to me.

zonkmachine:
@impaktor It looks like the strings from module-cargorun didn't find their way to Transifex. Could it be because of the separate commits? Maybe it just takes longer time?

impaktor:
@zonkmachine Sync at 3 AM, both ways.

PR #6253, opened 2025-11-21
Upgrade nlohmann::json, fix #include ciso646 warnings

sturnclaw:
As the title says, this PR performs a long-overdue upgrade of our JSON serialization library (from 3.2.0 to 3.12.0) to remove references to the now-deprecated `<ciso646>` header that was spamming the log with warnings when compiling.

It also gives us access to all of the nice new features that the library provides, including being able to serialize raw binary data into and out of CBOR, additional serialization targets to serialize JSON into, and better support for retrieving strings from JSON data with zero copies involved using `std::string_view`.

I've also upgraded libfmt from 10.2.1 -> 12.1.0, which brings some minor performance benefits, as well as upgrading doctest to the most recent version (and patching it to silence the warning it generates).

Assuming this doesn't cause build failures, I'll be merging it soon.

### DIFF
--- a/src/ship/PlayerShipController.cpp
+++ b/src/ship/PlayerShipController.cpp
@@ -205,7 +205,7 @@ PlayerShipController::PlayerShipController() :
 
 	m_toggleSpeedLimiter = InputBindings.toggleSpeedLimiter->onPressed.connect(
 		[this]() {
-			this->SetSpeedLimiterActive(not this->IsSpeedLimiterActive());
+			this->SetSpeedLimiterActive(!this->IsSpeedLimiterActive());
 		});
 
 	m_selectTarget = InputBindings.targetObject->onPressed.connect (


sturnclaw:
There was an errand "and" in StarSystemGenerator.cpp#1665 that had been there for... two years. This PR broke whatever was getting included to define "and" as `&&`, so we just now found it thanks to MSVC.

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -1662,7 +1662,7 @@ void PopulateStarSystemGenerator::PopulateAddStations(SystemBody *sbody, StarSys
 		}
 
 		// Always generate a station around a populated high-gravity world
-		if ((NumToMake == 0) and (sbody->CalcSurfaceGravity() > 10.5)) {  // 10.5 m/s2 = 1,07 g
+		if ((NumToMake == 0) && (sbody->CalcSurfaceGravity() > 10.5)) {  // 10.5 m/s2 = 1,07 g
```
fluffyfreak, line 1665: Wierd that it compiles for me, it's apparently defined in the `iso646.h` header in VS2022

fluffyfreak (approved):
Looks good to me, good to fix those `not`, `and` instances too

src/galaxy/StarSystemGenerator.cpp
sturnclaw: As I discovered (and mentioned on IRC), `and` and `not` are actually standards-compliant alternative keywords for C++ operators. The C++ standard requires that they be globally available in all translation units, which means MSVC is non-conforming in this regard by requiring iso646.

fluffyfreak:
Unfortunately the upgrade to libfmt v12.1.0 seems to break the PerfInfo screen. Specifically the use of `fmt::arg("unit", counter.unit)` causes it to crash now.
I have a crap fix for it but haven't been able to work out a better one. I'll post a PR asap with the crap fix so we can discuss a better one!

PR #6256, opened 2025-11-27
fmt::arg("unit", counter.unit) causes crash, hard code counter.unit

fluffyfreak:
Unfortunately the upgrade to libfmt v12.1.0 seems to break the PerfInfo screen.
Specifically the use of fmt::arg("unit", counter.unit) causes it to crash now.
To see this happen just open the Debug Tools with `ctrl+i` and then click the second tab for performance info.

This is something of a rubbish fix, but haven't been able to work out a better one.

Suggestions for how to _really_ fix the problem are welcome.


fluffyfreak:
Merged this for now, it can be fixed / tweaked easily enough as it's just one line

PR #6257, opened 2025-11-30
Tailsitters: Make ships land & launch like rockets / The Expanse / etc

### LINKED ISSUE #1178: The Case for Tail Landing/Launching Rockets
In case it needs to be made...

I have tried out Gernot's rocket - aside from the (hopefully soon to be corrected) sideways thrust vectoring and counter thrust control issues.
I really enjoyed it and even managed to dock it at Gates Spaceport :O

Using the Mains for lift-off:
This type of rocket could hold a unique roll in the Pioneer universe.
The fact that it would rely on the main engine for planet fall all the 

way to the surface opens up some interesting possibilities.
Lateral thrusters would still be essential to positioning your ship over a 

target landing pad or location.

Ships like this would make planets with very high gravity & thin atmospheres accessable to landing and take-off ONLY by such rockets.
These types of planets could only be visited by ships with extremely powerfull main engines.

 Perhaps for some super high grav planets only, AI robot ships need only apply - since the grav is far too great to support any type of biological lifeforms. 
Might make for a great AI controlled ore transport convoy system for heavy metals and other elements from such worlds to a nearby station.

It would be a ripe target for interception by an adventurous pirate :)

fluffyfreak:
Well this ones been a while coming...

### Description:
Tail-Sitting ships like modern day rockets, or sci-fi buck-rogers, TinTin, or The Expanse. Where the ship sits on it's rear where the largest engine(s) are based.

This means that the most powerful engines can be used for landing and takeoff on planets which is safer on high-G worlds.
Previously we landed using only the belly thrusters which are much weaker than the main engine(s).

Code is very work-in-progress, there's edge cases and situations everywhere it seems.

Screenshot speaks a thousand words.
screenshot-20251130-230505

fixes #1178 

~~I'll write up more tomorrow but I've got to sleep now now 😴~~

### Problems:
- [x] Undocking from ground station doesn't launch you off the pad
- [x] docking with orbital station the landing orientation is horizontal, then snaps to vertical
- [x] rough landing is still always belly orientation
- [ ] `Ship::SetLandedOn` will need fixing up

Modded Tail-Sitting MolaMolaTail by @bszlrd, place this into your mods folder, or extract it into the `data` directory for testing
[molatail.zip](https://github.com/user-attachments/files/23867130/molatail.zip)

### DIFF
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -902,10 +902,13 @@ void Ship::TestLanded()
 				const double planetRadius = static_cast<Planet *>(f->GetBody())->GetTerrainHeight(up);
 				SetPosition(up * (planetRadius - GetAabb().min.y));
 
-				// position facing in roughly the same direction
-				const vector3d right = up.Cross(GetDockingOrientation()).Normalized();
-				SetOrient(matrix3x3d::FromVectors(right, up));
-
+				// if we're tail sitting then stuck with however we cam in?
+				if (!CanTailSit()) {
+					// position facing in roughly the same direction
+					const vector3d right = up.Cross(GetOrient().VectorZ()).Normalized();
+					SetOrient(matrix3x3d::FromVectors(right, up));
+				}
+				
 				SetVelocity(vector3d::Zero);
 				SetAngVelocity(vector3d::Zero);
 				ClearThrusterState();


sturnclaw:
From a quick look on my phone, this looks like an acceptable way to handle it. Ideally, we'd split this into two different concepts: the "preferred reentry orientation" and "preferred landing orientation" vectors (and pull those from tags) rather than hard coding a selection between VectorY and VectorZ. In general, it would be a lot more flexible to define primary/alternate landing orientation tags in the model, so e.g. crazy modded ships could land at a 45 degree angle or ships could land on their noses etc.

Additionally, using tags for orientation opens the remote possibility to allow dynamically changing the landing mode based on ship animation - obviously not a concern right now but it would allow that flexibility.

fluffyfreak:
Well there's still a rough landing issue, but that appears to be a collision problem 🤔 
screenshot-20251203-222628

fluffyfreak:
@bszlrd @impaktor @sturnclaw this is ready for most testing, although I fully expect that it could break some re-fuel mission ships that scripts create as landed etc. 
Docking with orbitals, ground stations, and rough landings though need to be tested.

The slightly broken MolaMolaTail ship is attached as a mod to the top post.

I think the code could be much better but it's really tricky due to how difficult some of the landing code is to work with. Suggestions, or code commits, are welcome

bszlrd:
Great! I'll try to test it this weekend.

bszlrd:
Just tested it. So a ship is either tail or belly sitter for ground stations? Because I was unable to land the tail Mola on its belly, neither on the pad or rough.
Edit: I'm seeing the landing orientation stated as "either" in shipdef. So it is a bug? Or I need to select it somehow?
That will need some UI feedback then, communicating to the player, which orientation will be used.  Even better if one can choose. I thought I did some mockup for that, but it turns out I haven't, yet.
Another edge case is that if you start your game with a tail Mola, then it will belly-sit, but the launch assumes tail-sittings when it sets the speed.
Autopilot seems to handle it properly.
Haven't yet tested it any more than that.

fluffyfreak:
My heads all over the place with this but will try to get back into it asap.

Yeah, I need to rework some concepts around whether a ship is a tailsitter or not. I started this looking at planetary landings only, rough / station ones. It wasn't until I'd got them working that I considered space stations and realised that they'd probably _have_ to be bellysitters up there 🤦‍♂️ which was really annoying.

Personally I'd prefer ships to be one or the other in all situations so I wasn't thinking of UI, you'd buy a ship and know if it was a tail/belly sitter... 

The tail Mola is a bit hacky but I'm bet we always spawn them on their belly 🤔 needs much more work

bszlrd:
Ships being exclusively either belly or tail sitter could work. Not sure if the current orbitals have enough room for taller ones, but I think it should be fine in general. Might look a bit odd on an orbital though.
(I'm already thinking converting the Nerodia into a tailsitter cargo shuttle ;) )

fluffyfreak:
I was thinking that we may want to change the stations to have an inner area like the current station docking for belly sitters, and an external area like in The Expanse where they dock and the nose is pointed inwards and technically it's belly sitting but against a vertical wall like in the image

image

bszlrd:
Yeah, that could be nice. And a bit of a flight/navigation challenge as well for those who fly manually. But then either it is a sideways belly-sitting position if it is like on Tycho Station, or they tail-sit on an additional ring.

PR #6258, opened 2025-12-01
Update mission status if overdue

### LINKED ISSUE #6201: Mapping missions don't get the status "failed"

robothauler:
Closes #6201 
I added all the modules where I think it is safe to do so.

### DIFF
--- a/data/modules/CargoRun/CargoRun.lua
+++ b/data/modules/CargoRun/CargoRun.lua
@@ -253,7 +253,10 @@ onChat = function (form, ref, option)
 			cargotype       = ad.cargotype,
 			status          = cargo_picked_up and "ACTIVE" or "TO_PICK_UP",
 		}
-		table.insert(missions,Mission.New(mission))
+
+		mission = Mission.New(mission)
+		table.insert(missions, mission)
+		MissionUtils.SetupOverdueTimer(mission)
 
 		if ad.amount ~= ad.negotiated_amount then
 			-- recreate advert with the rest of the cargo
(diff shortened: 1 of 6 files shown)


sturnclaw (changes requested):
If the `MissionUtils` portion depends on the `due` key in the mission table, I'd recommend moving the `if Game.time < mission.due` portion into `FailedWhenOverdue`. Additionally, is there any validation to determine the mission status is a non-complete state, i.e. checking for the current state to be `ACTIVE`?

Separately, I'd recommend a more descriptive name of what's being accomplished inside the `FailedWhenOverdue` state. A verb would be appropriate here, such as `TrackMissionDueState` or `SetupOverdueTimer`. My main concern is that it's very difficult to tell what's actually happening inside the function, as well as whether the function is appropriate to call multiple times (or that it has to be called post-load). Both of the proposed name changes are directive, and the latter indicates that it's registering a timer (which is itself transient state and therefore has to be restored when loading a saved game).

robothauler:
During testing I noticed that missions[#missions] sometimes return nil. I do not understand this, but it should now be more robust for this edge case.

robothauler:
> Additionally, is there any validation to determine the mission status is a non-complete state, i.e. checking for the current state to be `ACTIVE`?

In my opinion, the modules are too different. The module must decide for itself what is still active.

PR #6259, opened 2025-12-05
Vector3 ZERO

fluffyfreak:
Added a static const member variable for ~~`ZERO`~~ `Zero` values to vector3 and replaced all instances of using `vector3d(0, 0, 0)` in the code I could find.

Did not do the same for cases of `vector3f(0.f, 0.f, 0.f)` as this often made sense in the situatiuns that it is used.

EDIT:
Changed the usage of `matrix3x3::Identity()` and `matrix4x4::Identity()` static functions to be a static const member named `Identity` and the `Identity()` renamed to `IdentityFunc()`. Also deleted the lightly used `matrix4x4fIdentity` and other values in favour of `matrix4x4f::Identity`

### DIFF
--- a/src/Beam.cpp
+++ b/src/Beam.cpp
@@ -314,7 +314,7 @@ void Beam::Render(Graphics::Renderer *renderer, const Camera *camera, const vect
 	const vector3f dir = vector3f(_dir).Normalized();
 
 	vector3f v1, v2;
-	matrix4x4f m = matrix4x4f::Identity();
+	matrix4x4f m = matrix4x4f::Identity;
 	v1.x = dir.y;
 	v1.y = dir.z;
 	v1.z = dir.x;
--- a/src/Body.cpp
+++ b/src/Body.cpp
@@ -25,9 +25,9 @@
 Body::Body() :
 	PropertiedObject(),
 	m_interpPos(0.0),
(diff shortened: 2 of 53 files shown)


sturnclaw:
Devil's advocate: does this make a difference in codegen outside of non-optimizing builds (i.e. debug without `-Og`)? The vector3 constructor is trivial and a candidate for constexpr. Additionally, the zero-argument default constructor initializes a vector3 to zero automatically (to prevent bugs with uninitialized vectors).

On the whole, this is probably better for consistency. It's easier to miss a non-zero constant in the explicit initializer form. One thing I'd like to see is for the `matrix4x4fIdentity` variables be moved to `matrix4x4f::IDENTITY` instead.

Final thought: the all-uppercase naming is appropriate for enumeration constants and macros, but as this is neither a macro nor an enum I think `vector3f::Zero` is more appropriate?

fluffyfreak:
This was initially less about the quality of code gen and was more of a quality of life coding change for me. It was really annoying me seeing all of these calls to `vector3f(0.f, 0.f, 0.f)` everywhere! I initially changed them to use the default ctor but in some places the code became less readable, so I added the constant ZERO to make them clearer.

I've now done the matrix classes `Identity` functions making them into a static const and removed the other static const `matrix4x4fIdentity` like values

PR #6269, opened 2026-01-05
Convert FOV to Radians before using with Frustum

### LINKED ISSUE #6121: Planet layer flicker on zoom
Image
Image

### Observed behaviour

On internal view, planet flickers when zooming (touchpad) in flight. Sometimes it gets stuck 'off'.
Time acceleration or scattering does not seem to affect this.

### Steps to reproduce

Start at Cydonia/Mars and take off for Bradbury. After the trajectory stabilizes and Mars is seen in the lower quarter, zoom in/out with the touchpad and observe the flicker. Eventually you can make the planet layer go away even though icons of planet features like cities are visible (see images above where Bradbury landing is floating over a starry sky). I think the planet layer may have skipped to a position much lower, below the screen.

Pioneer: master (003d8c0491b6919712e2f7df57845a88c9dfd14e)
Ubuntu 22.04 and 24.04

[output.txt](https://github.com/user-attachments/files/19933013/output.txt)

fluffyfreak:
This seems to fix #6121 for ships, as there was a small mistake when setting up the new Frustum. In the old you could pass it a FOV in degrees and it would internally convert to build the perspective matrix. This now needs to be done by the calling function which hadn't done that but did for the following usage a line below

PS: I'll merge this tomorrow unless anyone has comment or negative testing experiences

### DIFF
--- a/src/Camera.cpp
+++ b/src/Camera.cpp
@@ -53,8 +53,9 @@ CameraContext::~CameraContext()
 void CameraContext::SetFovAng(float newAng)
 {
 	m_fovAng = newAng;
-	m_frustum = Frustum(m_width, m_height, m_fovAng, m_zNear, m_zFar);
-	m_projMatrix = matrix4x4f::InfinitePerspectiveMatrix(DEG2RAD(m_fovAng), m_width / m_height, m_zNear);
+	const float fovAngRadians = DEG2RAD(m_fovAng);
+	m_frustum = Frustum(m_width, m_height, fovAngRadians, m_zNear, m_zFar);
+	m_projMatrix = matrix4x4f::InfinitePerspectiveMatrix(fovAngRadians, m_width / m_height, m_zNear);
 }
 
 void CameraContext::BeginFrame()


fluffyfreak:
Also seems to fix the issue with planets 👍

fluffyfreak:
@zonkmachine are you able to test this?

zonkmachine:
> @zonkmachine are you able to test this?

Nice! I tested it and the issue seem fixed. I did spot something and I don't know if it's related. It's not introduced in this PR however. I'll drop a saved game here. Start this game at slowest speed and watch pixels around expecially Olympus Mons. Pixels flashing. [olympusmons.txt](https://github.com/user-attachments/files/24436703/olympusmons.txt)

fluffyfreak:
Excellent, thankyou!

fluffyfreak:
Also there's other issues there around Olympus Mons!
image

Best to open a new issue for all of that please @zonkmachine

zonkmachine:
> Best to open a new issue for all of that please @zonkmachine

I don't see those other issues. Maybe it's better if you open the issue instead?

sturnclaw:
Excellent work, thank you for tackling this!

PR #6271, opened 2026-01-05
Dump all current faction home systems to JSON

sturnclaw:
This PR takes all current faction homeworlds (regardless of quality, e.g. some do not contain starports) and dumps them to custom system .json files. This is to prevent crashes when the galaxy generation state changes and some factions do not have randomly generated systems in their chosen home sector.

Per the commit description, this should not be seen as canonicalizing the faction homeworlds. These are terrible-quality randomly generated systems, and should be replaced with custom systems by contributors over time. The sole purpose of this PR is to give a starting-off point for contributors who like the current faction homeworlds, and a file to replace for contributors who do not.

Along the way, this PR fixes several bugs in the custom system authoring process:
- Invalid orbits were generated when loading a system that contained a gravpoint which was the parent of another gravpoint.
- Gravpoints implicitly required their masses to be the sum of their children's masses. The system editor now automatically calculates this and prevents the user from editing the mass, radius, or temperature of a gravpoint.
- Clicking the +/- buttons on inputs in the system editor did not record the change in the undo state.
- Some randomly generated systems with 3 or more stars had a non-zero root body index (e.g. for a 3-star system, the root body is actually body 4). I've added an explicit root body index for now, but this should be fixed in system generation eventually.


impaktor:
Nice.
Now I'm in the camp of slashing the +100 factions down to < 10, but that's for another time.

Also looking at the faction names they're very "procedural", in being just combinatorics of provided word list, e.g.
```
independent_network
independent_union
independent_league
```
this could also be changed (after this PR). Just throwing out some inspiration for @bszlrd 

No further comments.

bszlrd:
In my working build I slashed them to 4 for now

zonkmachine:
@sturnclaw I see slaves and nerve gas in the commodity market of Cydonia on master since this was merged.

sturnclaw:
Yep - I managed to forget during the creation of this PR that it introduces a bug where 102_Outer_Republic controls the Sol system despite that faction specifying a homeworld some 70ly away from Sol.

I'm digging into it now, and it's being somewhat finicky to debug; part of the problem seems to be that factions 102 and 045 share the same homeworld coordinates.

sturnclaw:
Fixed in #6274.

PR #6273, opened 2026-01-10
Fix infinite galaxy, remove visible density "blocks" in sector generation

sturnclaw:
This PR does two very simple things.

1. It fixes a read-past-end where the galaxy would "repeat" infinitely in every direction, which was actually just the density map texture reading into uninitialized memory. The galaxy now ends properly after ~650ly vertically.
2. It fixes a somewhat prevalent issue where the density of stars was uniform over a certain number of sectors, then visibly instantaneously changed in a rectilinear fashion. This was caused by a lack of filtering on the input galaxy map texture; you could quite literally see the pixels in the texture as they controlled the density of stars generated. To resolve this issue, the density map value is now sampled using bilinear filtering, averaging together the nearest 4 pixels based on the sector's position within the density map.

This PR also provides a new density map texture, generated from the currently unused galaxy color texture, with 4x the resolution (1024px vs 256px).

Additionally, the game will now complain at you if the density map texture is not a single-channel BMP image, rather than silently generating a completely wrong galaxy.


impaktor:
Nice!
I assume this is very much save-bumping.

PR #6284, opened 2026-02-02
Preparation for 20260203 release

sturnclaw:
General cleanup and fixes for 20260203. The headlining fix here is that thruster flicker works in-game now, and I've added a smooth fade so the individual thruster glow planes are not quite as visible anymore.

No, I will not apologize for using MM-DD-YY in source-code comments. :stuck_out_tongue:


fluffyfreak:
The ONLY correct date format is YY-MM-DD so it can be followed by HH-MM-SS 😛

bszlrd:
How about a MH-YS-DD-M date and time format? :P

PR #6287, opened 2026-02-05
.desktop: Prefer Wayland over X11

pcercuei:
Not specifying SDL_VIDEODRIVER will cause SDL to only use X11 and not even try Wayland, which is a problem on systems that only have Wayland.

By specifying SDL_VIDEODRIVER=wayland,x11 the game will work on both Wayland and X11 sessions.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -32,6 +32,10 @@ extern "C" int main(int argc, char **argv)
 	Profiler::detect(argc, argv);
 #endif
 
+	/* SDL2 does not enable Wayland by default, but we want to prefer it
+	 * when available, over X11 */
+	SDL_setenv("SDL_VIDEODRIVER", "windows,wayland,x11", 0);
+
 	OS::SetDPIAware();
 
 	RunMode mode = MODE_GAME;


pcercuei:
Related: https://github.com/flathub/net.pioneerspacesim.Pioneer/issues/16

We might need a different fix so I'll switch this to draft.

sturnclaw:
On quite literally almost any SDL2 runtime version (2.0.22+?), SDL_VIDEODRIVER can be set via the SDL_Hint mechanism before SDL_Init is called, rather than having to go through an environment variable. That might be a cleaner solution.

pcercuei:
Yes, I didn't go with that because I don't know how setting the hint to "wayland,x11"would  affect windows/OSX builds.

pcercuei:
Working properly now.

pcercuei:
> On quite literally almost any SDL2 runtime version (2.0.22+?), SDL_VIDEODRIVER can be set via the SDL_Hint mechanism before SDL_Init is called, rather than having to go through an environment variable. That might be a cleaner solution.

I just tried that and SDL just ignored the hint. So I used SDL_setenv() instead and now it works fine.

fluffyfreak:
Does this fix the flathub issues then and they'd take the build?

pcercuei:
> Does this fix the flathub issues then and they'd take the build?

It does, and they already accepted it:
https://flathub.org/fr/apps/net.pioneerspacesim.Pioneer

(I pack this commit as a patch to the release tarball in the flatpak recipe)

sturnclaw:
Note that this fix will also have to be applied to `editormain.cpp`.

PR #6290, opened 2026-02-06
Fix error shown in the Language selection list

fluffyfreak:
The languages selection list contains multiple double entries which causes errors in ImGui.

This fixes it by appending a hidden bit of text using the `##` formatting to add a number onto the end of the displayed string thus making it unique

Ekrano_kopija_20260206_190709


fluffyfreak:
I plan to merge this later today, or tomorrow (_Sunday_), if no-one responds as it's quite a simple fix

PR #6292, opened 2026-02-08
Fix overlapping craters, and early out of `GetHeights` that I broke previously

fluffyfreak:
Old distribution of craters:
screenshot-20260208-215719

New  distribution of craters:
screenshot-20260208-215712

I got so annoyed at these terrains whilst fixing the early-out tests, that I broke in a earlier commit 🤦‍♂️😢, that I finally had to do something about them.

Fixed the early-out of the `GetHeights` that I broke previously and only just realised.

It's hard to see how it's fixed in those screenshots.

For testing run the game on the master branch, start at Sol, and Ctrl+F10 on Vesta, Ceres, and Pallas. Then repeat whilst on this branch.

Previously any planet using those terrains would always have identically placed craters, and features would not change with the size of the planet. Now that craters have much more random distribution due to multiplying the vector `p` by a randomly generated quaternion, and the feature sizes are scaled based on the radius of the planet.

### DIFF
--- a/src/terrain/TerrainHeightMountainsCraters2.cpp
+++ b/src/terrain/TerrainHeightMountainsCraters2.cpp
@@ -8,6 +8,23 @@
 using namespace TerrainNoise;
 using namespace TerrainFeature;
 
+namespace {
+	constexpr double _featureWidthCentre[] = {
+		1000000.0,
+		 900000.0,
+		 800000.0,
+		1100000.0,
+		1200000.0
+	};
+	constexpr double _maxHeightMultiplier[] = {
+		0.05,
+		0.04,
+		0.05,
+		0.04,
+		0.07
+	};
+}
+
 template <>
 const char *TerrainHeightFractal<TerrainHeightMountainsCraters2>::GetHeightFractalName() const { return "MountainsCraters2"; }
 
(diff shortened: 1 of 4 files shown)


fluffyfreak:
@sturnclaw I'm going to leave this here until I get your approval. I know there's usually a few bug fix patches that sometimes have to be done after Pi-day 👍 

Feedback still welcome of course

src/terrain/TerrainHeightMountainsCraters2.cpp
```diff
@@ -15,19 +15,25 @@ template <>
...
-	SetFracDef(5, m_maxHeightInMeters * 0.05, 1e6, 10000.0);
-	SetFracDef(6, m_maxHeightInMeters * 0.04, 9e5, 10000.0);
-	SetFracDef(7, m_maxHeightInMeters * 0.05, 8e5, 10000.0);
-	SetFracDef(8, m_maxHeightInMeters * 0.04, 11e5, 10000.0);
-	SetFracDef(9, m_maxHeightInMeters * 0.07, 12e5, 10000.0);
+	SetFracDef(5, m_maxHeightInMeters * 0.05, GetRandWithinRangeAndMul(m_rand, 1000000, 0.05, invRadiusMul), GetRandWithinRangeAndMul(m_rand, 10000.0, 0.05, invRadiusMul));
+	SetFracDef(6, m_maxHeightInMeters * 0.04, GetRandWithinRangeAndMul(m_rand,  900000, 0.05, invRadiusMul), GetRandWithinRangeAndMul(m_rand, 10000.0, 0.05, invRadiusMul));
+	SetFracDef(7, m_maxHeightInMeters * 0.05, GetRandWithinRangeAndMul(m_rand,  800000, 0.05, invRadiusMul), GetRandWithinRangeAndMul(m_rand, 10000.0, 0.05, invRadiusMul));
+	SetFracDef(8, m_maxHeightInMeters * 0.04, GetRandWithinRangeAndMul(m_rand, 1100000, 0.05, invRadiusMul), GetRandWithinRangeAndMul(m_rand, 10000.0, 0.05, invRadiusMul));
+	SetFracDef(9, m_maxHeightInMeters * 0.07, GetRandWithinRangeAndMul(m_rand, 1200000, 0.05, invRadiusMul), GetRandWithinRangeAndMul(m_rand, 10000.0, 0.05, invRadiusMul));
```
sturnclaw: Nope nope nope nope. There's no sequence point in function arguments, so this is a non-deterministic mutation of hidden state. The standard does not provide any guarantee which `GetRandWithinRangeAndMul` will be called first.

src/terrain/TerrainNoise.h
```diff
@@ -176,6 +177,27 @@ namespace TerrainNoise {
...
+		vOut.z = rand.Double(-1.0, 1.0);
+	}
+
+	inline vector3d GetVector3Random(Random &rand)
+	{
+		return vector3d(rand.Double(-1.0, 1.0),
+			rand.Double(-1.0, 1.0),
+			rand.Double(-1.0, 1.0)).Normalized();
+	}
+
```
sturnclaw, line 201: Similarly, this is non-deterministic. Either `vector3d` must be [list-initialized](https://en.cppreference.com/w/cpp/language/list_initialization.html) or these values must be cached in separate full-expressions.

sturnclaw (changes requested):
The C++ standard would like to shoot you in the foot.

fluffyfreak:
This terrain issue, and there will be more, is also fixed by this PR.

The Early-out was preventing negative values making it through to the rest of the code.
Pioneer_15-Feb-26_13_56_04
Pioneer_15-Feb-26_13_56_23
screenshot-20260216-070011

src/terrain/TerrainHeightMountainsCraters2.cpp
fluffyfreak: This has been resolved

src/terrain/TerrainNoise.h
fluffyfreak: This has been resolved

impaktor:
Status?
Awaiting re-review?

fluffyfreak:
Awaiting re-review

sturnclaw:
Will review quite soon. I am intending to release a hotfix containing most of the non-savebump affecting PRs within the next few weeks; I'll leave it to your discretion whether this PR is of enough importance to potentially require a saved-game recovery in a small percentage of user cases.

fluffyfreak:
Sadly I think it is important enough, without this fix it leaves a lot of the planets with messed up giant mountains and other oddball terrain.

sturnclaw:
I'll put it on the shortlist then.

fluffyfreak:
I'll squash and merge this later this week if there's no objections by... Wednesday-ish 👍

fluffyfreak:
Merged

PR #6296, opened 2026-02-11
Improve calculations in AIMatchVel (AIChangeVelBy)

Gliese852:
The problem with the current version is that if the powerful thruster is turned away from the direction of the required speed, it pushes the ship to the side when gaining speed. This happens because the velocity vector is decomposed along the directions of the thrusters, and each thruster independently tries to maximize its component.

In the new version it is done so that the direction of the thrust vector necessarily coincides with the direction of the required speed.

Also wrote tests that allow to see the problem with the previous implementation and test the new one.

### DIFF
--- a/src/ship/Propulsion.cpp
+++ b/src/ship/Propulsion.cpp
@@ -334,15 +334,35 @@ bool Propulsion::AIChangeVelBy(const vector3d &diffvel, const vector3d &powerLim
 	vector3d extf = m_dBody->GetExternalForce() * (Pi::game->GetTimeStep() / m_dBody->GetMass());
 	vector3d diffvel2 = diffvel - extf * m_dBody->GetOrient();
 
-	vector3d maxThrust = GetThrust(diffvel2);
-	vector3d maxFrameAccel = maxThrust * (Pi::game->GetTimeStep() / m_dBody->GetMass());
-	vector3d thrust(
-		Clamp(diffvel2.x / maxFrameAccel.x, -powerLimit.x, powerLimit.x),
-		Clamp(diffvel2.y / maxFrameAccel.y, -powerLimit.y, powerLimit.y),
-		Clamp(diffvel2.z / maxFrameAccel.z, -powerLimit.z, powerLimit.z));
-	SetLinThrusterState(thrust); // use clamping
-	if (thrust.x * thrust.x > 1.0 || thrust.y * thrust.y > 1.0 || thrust.z * thrust.z > 1.0) return false;
-	return true;
+	double diffSpeedSqr = diffvel2.LengthSqr();
+	if (diffSpeedSqr == 0.0) {
+		SetLinThrusterState({ 0.0, 0.0, 0.0 });
+		return true;
+	}
+
+	vector3d thrustDir = diffvel2 / sqrt(diffSpeedSqr);
+	vector3d allThrust = GetThrust(diffvel2);
+	vector3d ltdThrust{ allThrust.x * powerLimit.x, allThrust.y * powerLimit.y, allThrust.z * powerLimit.z };
+
+	assert(ltdThrust.x != 0.0 && ltdThrust.y != 0.0 && ltdThrust.z != 0.0);
+
+	vector3d invScales{ thrustDir.x / ltdThrust.x, thrustDir.y / ltdThrust.y, thrustDir.z / ltdThrust.z };
+	double invScale = std::max(abs(invScales.x), std::max(abs(invScales.y), abs(invScales.z)));
+
+	// maximum thrust levels for acceleration in exactly the right direction
+	vector3d maxDirLevels = invScales / invScale;
+
+	vector3d maxDirThrust{ maxDirLevels.x * ltdThrust.x, maxDirLevels.y * ltdThrust.y, maxDirLevels.z * ltdThrust.z };
+	double maxDirFrameAccelSqr = (maxDirThrust * (Pi::game->GetTimeStep() / m_dBody->GetMass())).LengthSqr();
+
+	if (diffSpeedSqr >= maxDirFrameAccelSqr) {
+		SetLinThrusterState(maxDirLevels);
+		return false;
+	} else {
+		double amount = sqrt(diffSpeedSqr / maxDirFrameAccelSqr);
+		SetLinThrusterState(maxDirLevels * amount);
+		return true;
+	}
 }
 
 // Change object-space velocity in direction of param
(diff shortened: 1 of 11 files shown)


src/ship/Propulsion.cpp
```diff
@@ -334,15 +334,35 @@ bool Propulsion::AIChangeVelBy(const vector3d &diffvel, const vector3d &powerLim
...
+	if (diffSpeedSqr == 0.0) {
+		SetLinThrusterState({ 0.0, 0.0, 0.0 });
+		return true;
+	}
+
+	vector3d thrustDir = diffvel2 / sqrt(diffSpeedSqr);
+	vector3d allThrust = GetThrust(diffvel2);
+	vector3d ltdThrust{ allThrust.x * powerLimit.x, allThrust.y * powerLimit.y, allThrust.z * powerLimit.z };
+
+	assert(ltdThrust.x != 0.0 && ltdThrust.y != 0.0 && ltdThrust.z != 0.0);
```
sturnclaw, line 347: Note: this will assert for e.g. missiles which only have a single rear linear thruster. I'd recommend using `std::min` to filter NaN values to 0 after computing `invScales`.

src/ship/Propulsion.cpp
```diff
@@ -334,15 +334,35 @@ bool Propulsion::AIChangeVelBy(const vector3d &diffvel, const vector3d &powerLim
...
+
+	vector3d invScales{ thrustDir.x / ltdThrust.x, thrustDir.y / ltdThrust.y, thrustDir.z / ltdThrust.z };
+	double invScale = std::max(abs(invScales.x), std::max(abs(invScales.y), abs(invScales.z)));
+
+	// maximum thrust levels for acceleration in exactly the right direction
+	vector3d maxDirLevels = invScales / invScale;
```
sturnclaw, line 353: It is safe to assert `invScale > 0.0` here as a Propulsion object with no thrusters is a useless construct.

Gliese852:
Hurray, CI is working!

sturnclaw:
@Gliese852 - with apologies for the late review, did you test Pioneer's "Dummy" renderer backend instead of using SDL's "offscreen" render implementation? I don't believe any gameplay code requires the use of the GPU or fully-correct Renderer behavior...

sturnclaw (commented):
I have (apparently) some pending review commits from months ago that I thought I clicked submit on. I'm not certain how well the new propulsion code will handle missiles or other objects with linear thrust in only a single axis.

Gliese852:
> @Gliese852 - with apologies for the late review, did you test Pioneer's "Dummy" renderer backend instead of using SDL's "offscreen" render implementation? I don't believe any gameplay code requires the use of the GPU or fully-correct Renderer behavior...

I made another attempt to run the game with the Dummy renderer and this time it was successful, although I had to edit the code in some places. You can see the scope of changes [here](https://github.com/Gliese852/pioneer/commit/802d25a52ba0a5bf9b224eebf0a896bb37f41289). Here I simply corrected it to run the test, but in the final version will need to insert some branches in order to select the correct renderer.

Which path do you think is better to take? Should I correct the code to run with the Dummy renderer, or just use SDL offscreen backend?

sturnclaw:
> Which path do you think is better to take? Should I correct the code to run with the Dummy renderer, or just use SDL offscreen backend?

I would suggest using the Dummy Renderer instead. SDL's offscreen backend (AFAIK) does still attempt to produce an image via hardware/software rendering, which is going to involve extra overhead compared to what we're actually interested in for unit tests. In contrast, the dummy render backend intentionally discards uploaded data, which results in less computational overhead and memory pressure especially in a CI context.

src/ship/Propulsion.cpp
Gliese852: I tested it and it doesn't assert with missiles, because missiles don't use this function. Also only unguided missiles have zero side thrusters, but they do not need an autopilot at all - it flies forward as much as it can. Although the problem of using this function in the future by any limited ship remains. And here it’s worth thinking carefully, if any of the thrusters in the desired direction is not capable of pushing, then, according to the proposed approach, the ship cannot begin to accelerate in this direction at all, because in this orientation it will fly in the wrong direction anyway, and perhaps it’s worth turning around first? Simply zeroing the invScale component on the side of a non-functioning thruster can cause serious trajectory deviation. If you still insist on removing these assertions, maybe just replace them with a check and silently return instead of an assertion, in the hope that the autopilot will eventually turn the ship in the right direction?

On the other hand, from the missile's perspective, it seems ineffective. (It's worth checking though.) For a missile, it's probably better if it starts accelerating roughly in the direction of the target, in the hope that it will subsequently correct its direction. Perhaps that is why it is right that missiles use their own autopilot, and ships use their own.

src/ship/Propulsion.cpp
Gliese852: We are checking 3 sides, but the thrusters are located on 6 sides and the remaining three may be non-zero. Since we're even thinking about allowing zero thrusters in this place.

sturnclaw:
@Gliese852 I accept your argumentation regarding missile autopilots and the necessity of the asserts.

If you're able to update this PR to use the dummy renderer (and revert the runner version bump as that changes the glibc version we build against for release binaries) plus whatever hacks are needed to make it run, I'd like to get this PR merged quite soon.

Gliese852:
@sturnclaw I made a clean version with a dummy renderer, without touching the code in Propulsion.

sturnclaw (approved):

sturnclaw:
Hmm... it seems this branch doesn't interact favorably with something merged since it was authored. As of commit d62a20965d646a2a1d69d70838d328bd192c8970 (the merge commit for this PR), manually thrusting upwards and sideways under upwards cruise mode (hold R+A/D) causes the upwards thrust to be mostly disregarded and as a consequence the player's ship ascends extremely slowly or not at all. This is a behavior that is not present in the PR's branch, so the interaction is related to a change between commit 9754d3bf4eaaf9f71d409993dca52db1f4cca806 and the merge commit.

If you're unable to address this issue before the next release, I'll have to revert the merge commit. We may be looking at an August (late)Midsommar release, though if we don't release this month you may have some time.

Gliese852:
I'll get to this as soon as possible.

Mc-Pain:
This function does not account for local gravity

For example, ship holds position around body with local gravity of 2.0g
When AIMatchVel is called, per-direction thrust is divided proportionally per-direction speeds.
If vertical targetSpeed is much lower than horizontal targetSpeed, we would have near-zero vertical thrust and crash.

It's better to implement a skew term to compensate local gravity and ensure we're accelerating the way we intended.

sturnclaw:
This PR has been reverted temporarily in preparation for a release on the 7th. I'm quite happy to re-merge it once the unexpected interaction has been resolved.

PR #6298, opened 2026-02-14
Show the status of the wanted person

robothauler:
Related to #6249 point 5


data/lang/module-findperson/en.json
```diff
@@ -343,6 +343,10 @@
     "description": "",
     "message": "OK, agreed."
   },
+  "ON_BOARD": {
+    "description": "",
```
impaktor: Could you please add description field, this is pertaining whether a passenger is on board.

PR #6299, opened 2026-02-17
Start-At fix and log conditional

fluffyfreak:
I ahd trouble using the Start-At (`-sa=X,Y,X,#,#`) loading option and found that it was asserting. So wanted to avoid the assert and just log, but only when it failed to find any systems within a sector.

Thus the conditional logging works as a log-if-false.


zonkmachine:
Issues with the startat argument is mentioned here: https://github.com/pioneerspacesim/pioneer/issues/5285
I did some quick tests and it looks like this PR fixes it.

PR #6309, opened 2026-03-29
Approximate ring shadow penumbra

sturnclaw:
The ring shader has always had a minor problem where shadow edges are extremely sharp and alias horribly at long view distances.

Screenshot From 2026-03-28 21-49-31

They say a picture's worth a thousand words, so here's a video instead so I don't have to type out all thousand words.

https://github.com/user-attachments/assets/99ebc69e-ff93-46fc-a97d-974f69f3165b

This PR uses a completely non-physically based value for the shadow penumbra size (specifically performs a `smoothstep` over a distance corresponding to the light ray passing through ~0.5% of the planet's radius) to approximate atmosphere and light-source-size penumbra. I do not have the desire (or the exposed variables) at this time to make the value more physically based, and at the scales involved in Pioneer there would not be a significant amount of gain if I did so.

### DIFF
--- a/data/shaders/opengl/planetrings.frag
+++ b/data/shaders/opengl/planetrings.frag
@@ -14,11 +14,6 @@ out vec4 frag_color;
 
 #define NUM_OCTAVES 3
 
-float findSphereEyeRayEntryDistance(in vec3 sphereCenter, in vec3 eyeTo, in float radius)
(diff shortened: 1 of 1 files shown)


fluffyfreak (approved):
Looks good, nice change

PR #6310, opened 2026-03-29
Balance number of combat and scout missions

robothauler:
In my opinion, there are too many combat missions in low populated peaceful systems.

To put it a little exaggeratedly, the expensive scanner equipment is completely useless outside of Sol. There should at least be a chance that one mission will be generated.

### DIFF
--- a/data/modules/Assassination/Assassination.lua
+++ b/data/modules/Assassination/Assassination.lua
@@ -229,8 +229,8 @@ local makeAdvert = function (station)
 end
 
 local onCreateBB = function (station)
-	local num = Engine.rand:Integer(0, math.ceil(Game.system.population) / 2 + 1)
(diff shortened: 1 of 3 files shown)


impaktor:
From a review perspective I suspect it's difficult to assess how big the change is, I'm always thinking in graphs or similar data visualizations. Perhaps just reading the code is good enough for @sturnclaw ? Otherwise, some examples of number of adverts could be illustrative.

robothauler:
Just threw this together quickly.
assassination
combat
scout
Some points to consider:
Sol population 16.69 bil. lawlessness 0.01
36 Ophiuchi population 6.08 bil. lawlessness 0.98
Most systems have a population of less than 1 billion!

sturnclaw (approved):
Thanks!

PR #6313, opened 2026-03-30
Scoop mission fix

robothauler:
Game.player.frameBody is nil if the player is orbiting the system barycenter. This is a correction for #6279.


impaktor:
This works with both orbiting and surface station?
Was the issue that you could be in a frame different from the body of the station?

robothauler:
> This works with both orbiting and surface station?

Yes, I think so.

> Was the issue that you could be in a frame different from the body of the station?

Yes. If the ship moves from one star to another within the system and the ship is not in the frame of a star and the placeAdvert() function is triggered, the game crashes.

PR #6315, opened 2026-04-09
Fix for issue #5947. Draw HUD trails relative to the player.

### LINKED ISSUE #5947: Police (enemies in general?) shoot at velocity vector, not at ship
### Observed behaviour
Intercepted by police while carrying illegal goods.  Police opened fire, but never hit me - they are aiming at my velocity vector, not at me.

image

### Expected behaviour
Enemy ships target my ship.

### Steps to reproduce

* carry illegal goods
* approach station on autopilot (with an offset velocity vector)
* get intercepted by police
* watch police miss

Pioneer "master"; Debian GNU/Linux 12

sam52796:
Fixes #5947. Draw HUD trails to indicate motion relative to the player, rather than relative to a distant frame of reference, so that people don't mistake the trails for beam lasers. Also fix a minor issue with motion interpolation during a reference frame switch.

### DIFF
--- a/src/Body.cpp
+++ b/src/Body.cpp
@@ -307,8 +307,13 @@ void Body::SwitchToFrame(FrameId newFrameId)
 	const vector3d vel = GetVelocityRelTo(newFrameId); // do this first because it uses position
 	const vector3d fpos = frame->GetPositionRelTo(newFrameId);
 	const matrix3x3d forient = frame->GetOrientRelTo(newFrameId);
+	const vector3d ifpos = frame->GetInterpPositionRelTo(newFrameId);
+	const matrix3x3d iforient = frame->GetInterpOrientRelTo(newFrameId);
+
 	SetPosition(forient * GetPosition() + fpos);
 	SetOrient(forient * GetOrient());
+	m_interpPos = iforient * GetInterpPosition() + ifpos;
+	m_interpOrient = iforient * GetInterpOrient();
 	SetVelocity(vel + newFrame->GetStasisVelocity(GetPosition()));
 	SetFrame(newFrameId);
 
--- a/src/HudTrail.cpp
+++ b/src/HudTrail.cpp
@@ -6,6 +6,7 @@
 #include "Body.h"
 #include "Frame.h"
 #include "Pi.h"
+#include "Player.h"
 #include "graphics/Material.h"
 #include "graphics/RenderState.h"
 #include "graphics/Renderer.h"
@@ -18,7 +19,6 @@ const Uint16 MAX_POINTS = 100;
 
 HudTrail::HudTrail(Body *b, const Color &c) :
 	m_body(b),
-	m_currentFrame(b->GetFrame()),
 	m_updateTime(0.f),
 	m_color(c)
 {
@@ -34,19 +34,13 @@ HudTrail::HudTrail(Body *b, const Color &c) :
 void HudTrail::Update(float time)
 {
 	PROFILE_SCOPED();
-	//record position
+	//record the position relative to the player, the HUD trail then displays the history of this object's trajectory around the player
 	m_updateTime += time;
 	if (m_updateTime > UPDATE_INTERVAL) {
 		m_updateTime = 0.f;
-		FrameId bodyFrameId = m_body->GetFrame();
-
-		if (!m_currentFrame) {
-			m_currentFrame = bodyFrameId;
-			m_trailPoints.clear();
-		}
-
-		if (bodyFrameId == m_currentFrame)
-			m_trailPoints.emplace_back(m_body->GetInterpPosition());
+		// Always use a non-rotating frame, so that the relative positions are un-rotated
+		FrameId nrf = Frame::GetFrame(Pi::player->GetFrame())->GetNonRotFrame();
+		m_trailPoints.emplace_back(m_body->GetInterpPositionRelTo(nrf) - Pi::player->GetInterpPositionRelTo(nrf));
 	}
 
 	while (m_trailPoints.size() > MAX_POINTS)
@@ -56,9 +50,12 @@ void HudTrail::Update(float time)
 void HudTrail::Render(Graphics::Renderer *r)
 {
 	PROFILE_SCOPED();
-	//render trail
+	//render trail relative to the player's current position, this ensures that relative motion is displayed
 	if (m_trailPoints.size() > 1) {
-		const vector3d vpos = m_transform * m_body->GetInterpPosition();
+		// Set up a transform that includes the player's current position, so that everything we draw is relative to that
+		// Always use a non-rotating frame, so that the relative positions are un-rotated
+		FrameId nrf = Frame::GetFrame(Pi::player->GetFrame())->GetNonRotFrame();
+		const vector3d vpos = m_transform * Pi::player->GetInterpPositionRelTo(nrf);
 		m_transform[12] = vpos.x;
 		m_transform[13] = vpos.y;
 		m_transform[14] = vpos.z;
@@ -68,17 +65,18 @@ void HudTrail::Render(Graphics::Renderer *r)
 		static std::vector<Color> colors;
 		tvts.clear();
 		colors.clear();
-		const vector3d curpos = m_body->GetInterpPosition();
 		tvts.reserve(MAX_POINTS);
 		colors.reserve(MAX_POINTS);
 
-		tvts.emplace_back(vector3f(0.f));
-		colors.emplace_back(Color::BLANK);
+		// Start with the object's current position relative to the player (so when we apply the above transform this
+		// works out to the object's current position). Then draw a line along the trail history, gradually fading out.
+		tvts.emplace_back(m_body->GetInterpPositionRelTo(nrf) - Pi::player->GetInterpPositionRelTo(nrf));
+		colors.emplace_back(m_color);
 		float alpha = 1.f;
 		const float decrement = 1.f / m_trailPoints.size();
 		const Color tcolor = m_color;
 		for (size_t i = m_trailPoints.size() - 1; i > 0; i--) {
-			tvts.emplace_back(-vector3f(curpos - m_trailPoints[i]));
+			tvts.emplace_back(vector3f(m_trailPoints[i]));
 			alpha -= decrement;
 			colors.emplace_back(tcolor);
 			colors.back().a = Uint8(alpha * 255);
@@ -92,6 +90,5 @@ void HudTrail::Render(Graphics::Renderer *r)
 
 void HudTrail::Reset(FrameId newFrame)
 {
-	m_currentFrame = newFrame;
 	m_trailPoints.clear();
 }
--- a/src/HudTrail.h
+++ b/src/HudTrail.h
@@ -31,7 +31,6 @@ public:
 
 private:
 	Body *m_body;
-	FrameId m_currentFrame;
 	float m_updateTime;
 	Color m_color;
 	matrix4x4d m_transform;
--- a/src/Ship.cpp
+++ b/src/Ship.cpp
@@ -929,7 +929,6 @@ void Ship::SetLandedOn(Planet *p, float latitude, float longitude)
 void Ship::SetFrame(FrameId fId)
 {
 	DynamicBody::SetFrame(fId);
-	m_sensors->ResetTrails();
 }
 
 void Ship::TimeStepUpdate(const float timeStep)


src/GameConfig.cpp
```diff
@@ -42,7 +42,7 @@ GameConfig::GameConfig(const map_string &override_)
 	map["WorkerThreads"] = "0";
 	map["SpeedLines"] = "0";
 	map["EnableCockpit"] = "0";
-	map["HudTrails"] = "0";
+	map["HudTrails"] = "1";
```
fluffyfreak: this should still be 0 so they are off by default

fluffyfreak:
I like it, code looks good, I'll try to build and test it asap

sturnclaw (changes requested):
A general problem I'm seeing here is that the new player-relative encoding method does not include any mechanism to update the history buffer with the player's movement delta.

This would lead to the hud trail "displacing" from the target as the player moves perpendicularly to the target at high relative velocity, as the trail points keep their original player-relative position even though their location relative to the player should now be different.

Additionally, the frame-change handling needs to be restored, as `Body::GetInterpPositionRelTo` returns points in the orientation of the frame of the `RelTo` object, which can potentially change for e.g. axial tilt of rotating frames. When transiting into a frame with a different orientation, the stored hud trail points will be discontinuous with new trail points.

sam52796:
Relative motion in space is difficult to think about. Actually I originally tried a far more complicated technique where I tried to show the NPC ship's changes in velocity rather than just straightforward relative position, but the result was underwhelming. On the other hand, all the scenarios I tried with the method implemented here would track the NPC movement across the screen perfectly. This eventually made sense when I remembered that "the player moving perpendicularly to the target" is exactly the same as "the target moving perpendicularly to the player" since all movement in space is relative anyway.

So, are your objections just theoretical, or have you seen examples where the new HUD trails look wrong? If it's the latter, then can you provide a save where an NPC ship is moving across the player's viewscreen and the trail doesn't follow its motion? I wasn't able to come up with such a scenario during my own testing.

I'm also not sure how adding frames of reference back in would help. The new trails already look correct when an NPC ship changes frame, precisely because the trail is now only relative to the player. If the NPC ship has position X on my screen at tick 0, and position Y at tick 1, then the HUD trail drawn at tick 1 should be a line from Y to X. It doesn't matter which frames of reference I or the NPC ship are in at either X or Y, or which one of us was "really" doing the moving.

sam52796:
Thinking about this a bit more, I think that you would only see the trail drift if the player was actively accelerating. But since acceleration is fairly mild compared to relative speeds, and the trail only lasts for 10 seconds, and there would have to be a clear static background to see the trail against, I don't think this would be noticeable except in very specific occasional circumstances.

sturnclaw:
> Relative motion in space is difficult to think about. Actually I originally tried a far more complicated technique where I tried to show the NPC ship's changes in velocity rather than just straightforward relative position, but the result was underwhelming. On the other hand, all the scenarios I tried with the method implemented here would track the NPC movement across the screen perfectly. This eventually made sense when I remembered that "the player moving perpendicularly to the target" is exactly the same as "the target moving perpendicularly to the player" since all movement in space is relative anyway.
> 
> So, are your objections just theoretical, or have you seen examples where the new HUD trails look wrong? If it's the latter, then can you provide a save where an NPC ship is moving across the player's viewscreen and the trail doesn't follow its motion? I wasn't able to come up with such a scenario during my own testing.

The problem is conceptual, not just theoretical. The HUD trails display a ship's position at time $$t_{0-n}$$, which is a first-order value. The player's eye/brain then interprets the trail to intuit the velocity of the ship, a second-order value.

However, when the player is moving, the stored position of the target *moves with the player* because it is stored as a player-relative value with no compensation. This means that stored data points appear to maintain the same distance and direction relative to the player, regardless of the player's motion. A "trivial" (though far from easy to arrange) test case is to increase the lifetime of a straight-line HUD trail to 30s or so, and then attempt to fly perpendicular "through" the rendered HUD trail. The trail will appear to "curve" from a fixed point in front of the player towards the current position of the target ship.

What needs to be applied to the history buffer is the sum of the player's change in position (aka the delta of the player's position between when the sample was taken and now). Otherwise, the player's velocity is incorrectly summed to the displayed second-order velocity of the ship.

$$
{P'}_{t_{-n}} = P_{t_{-n}} - \Delta P_{player}
$$


> I'm also not sure how adding frames of reference back in would help. The new trails already look correct when an NPC ship changes frame, precisely because the trail is now only relative to the player. If the NPC ship has position X on my screen at tick 0, and position Y at tick 1, then the HUD trail drawn at tick 1 should be a line from Y to X. It doesn't matter which frames of reference I or the NPC ship are in at either X or Y, or which one of us was "really" doing the moving.

I'm not mentioning adding frames of reference, merely handling the case where the *player* changes reference frame. Because frames have their own orientation and stored points implicitly use the orientation of the player's frame when they were recorded, this means the player-relative stored positions in the history buffer are oriented incorrectly if the new frame has a different orientation than the old frame.

The result of `Body::Get[Interp]PositionRelTo()` is not oriented such that -Z forward is the front of the player's ship. It is instead in the orientation of the frame of the specified body, such that the *magnitude* of the vector is relative to the player's position, but the *direction* of the vector is relative to the orientation of the current frame.

sam52796:
I do feel a bit awkward waltzing into an open-source space sim project and starting an argument with one of the main contributors about how movement in space works, but here we are...

You say "when the player is moving", but this is the crux of the matter. On earth, we think of things as moving or being stationary, but in space there is no distinction. Everything is moving, all the time, relative to something. Hence why you have frames of reference in the Pioneer code. I'm sure you understand this, probably better than me.

So, when it comes to drawing a HUD trail, it's not enough just to think about "how things moved", the question is always "how things moved relative to something else". The HUD trail will show the recent trajectory of an NPC ship relative to... something. So now the question is - relative to what?

There are basically two choices here. We can either show movement relative to an existing Pioneer frame of reference, or relative to the player. The existing HUD trail code does the former, I'm suggesting that the latter is more useful in practice. Hopefully some examples will help to clear this up:

Imagine you are in the International Space Station, holding an apple out in front of you. You give it a gentle upward push, and watch it float away. If you had a HUD, you would expect the HUD trail to be a line from your hand up to the apple's current position.

Now imagine that the International Space Station disappears. You are still in the same position, holding an apple. You give it a gentle upward push, and watch it float away. What should the HUD display? My expectation is still a line from your hand up to the apple's current position.

The existing HUD trail code instead draws a long line extending out to one side, because in the absence of the station it's now tracking the apple's position relative to the earth, and both you and the apple are in orbit travelling at 8km/sec. Is this what you expected? Maybe it is what _you_ expected, but even if so, I would certainly argue that this kind of HUD trail is not _useful_.

The problem is clear if you watch an NPC ship leave a space station. The existing HUD trail code draws a sensible-looking trail for a few seconds, until the ship reaches the station's frame boundary, then suddenly there is a long line extending off into the distance to represent the massive orbital speed relative to the new frame of reference. People think that these lines are beam lasers, or random graphical glitches. Arguably, yes they are glitches. Even if you will say that, no, these are correct representations of the NPC's position relative to whatever frame of reference they were in when the position was recorded, I still don't think this is useful to the player. My new HUD trail code initially draws the same line as the existing code, and then continues that same line even after it switches frames. This seems more useful, particularly when you are in deep space, and looking at a ship that is moving slowly relative to you, but you are both moving at >1000km/s relative to the reference frame. My new HUD trail will show NPC movement as perceived by the player. The existing one shows a line shooting off in the direction of the X marker.

Your example is an NPC ship leaving a long trail in deep space, and then the player flying "through" the trail. I understand what you are getting at, but I think the trail would only appear to curve if you are accelerating towards it while it is being created. Otherwise it will still be a straight line. If you were travelling towards it at constant velocity while it was being created, then it would be a line angled towards you - but this is correct, because in that case the NPC ship's relative motion is a line angled towards you. I agree that the acceleration issue specifically leads to the trail not being 100% correct, but relative speeds are often over 100m/s, while acceleration is a few m/s at most. It just isn't noticeable as an effect when flying around NPC ships. Honestly, I was expecting to have to do some kind of adjustments for changes in velocity, but to my surprise the new simple trails look basically correct in all situations that I've tried.

A third example, and the one I expect to be the most controversial, is - what if you are skimming past a space station at a high speed, while an NPC ship is exiting from it. What will you see? If HUD trails are recorded as positions relative to the player, the origin of the trail will appear to move with you, so it looks like the start point of the trail, which you expect to stay at the door of the space station, moves away from the door. Aha! Surely this is wrong!

Actually, no, we're falling into the frame of reference trap again. Both the station and the NPC ship are moving relative to the player. This would be clear if the space station door also had a HUD trail - you would see that the trails met at the point where the ship left the station, and then diverged as both the ship and the station door moved relative to you. This only looks wrong because subconsciously you are treating the station like a fixed point.

I guess the truth of the matter is - there is no way to implement HUD trails in a way that looks intuitive in all possible scenarios. My suggestion is that we switch to player-relative trails because it a closer match to intuition in more scenarios, and hence an improvement over the existing way. I invite everyone to try the new HUD trail implementation and see if they agree. At the very least - can they find any instances where the new code does something obviously wrong? I tried moving my own ship over a frame boundary while watching an NPC ship, and the new HUD trail was a clean unbroken line, so I still don't know how the player's orientation frame is supposed to make it go wrong. I did realise that there is code to reset the trail when this happens, which is now no longer necessary, so I've removed that.

Alternatively, we give up on HUD trails and hide the UI option. Because leaving them how they currently are, with the "beam laser" effect, is definitely not the best course of action, in my opinion.

fluffyfreak:
> I do feel a bit awkward waltzing into an open-source space sim project and starting an argument with one of the main contributors about how movement in space works, but here we are...

Hah, don't worry, we've had a lot worse :D and at least you're thinking through and arguing the points.

I haven't had chance yet to test this code versus the old code but it's on my todo list for the weekend.

sam52796:
Thanks. Actually I've been thinking about this more, and I think that straightforwardly drawing the trail relative to the player works for acceleration too. The resulting line _should_ be curved, because acceleration looks like a curve.

If you stand on Earth and throw an apple horizontally away from you, it curves down towards the ground, because it's accelerating in that direction. A HUD trail for the apple would therefore show a curve too. Alternatively, if you're standing in a spaceship that is accelerating upwards at 1g, and you throw an apple in the same way, then actually the apple would stop accelerating (at least until it hits the floor) while you continue to accelerate. Yet the trajectory (and hence the HUD trail) would look the same in both cases.

Basically, a line drawn along the history of relative positions from the player will always look correct, regardless of who is moving or accelerating. 

Anyway, I look forward to some feedback from people who have tried it out. If you want to ensure you see the HUD trails easily then you can increase the distance that they are drawn at - it's this number in line 124 of Sensors.cpp:

`	Space::BodyNearList nearby = Pi::game->GetSpace()->GetBodiesMaybeNear(m_owner, 100000.0f);
`

fluffyfreak:
I've done some testing and this seems to handle transition between Mars and Mars high frames ok. It's actually pretty tough to get taht to happen without using the ingame `Ctrl+i` debug menu I think.

fluffyfreak:
@sturnclaw I'm not quite sure I understand the discussion between you two about the relative motion of the trails so I did a functional test. Everything seemed to move as I'd expect it, trails were attached to ships, faded out, didn't follow me strangley, and I moved relative to those trails in a way that seemed correct.

I'll hold off approving/merging until you've given some more feedback :+1:

sturnclaw:
> I do feel a bit awkward waltzing into an open-source space sim project and starting an argument with one of the main contributors about how movement in space works, but here we are...

On the contrary, thank you for elucidating your point! It was not clear to me from your initial posts that you intended to completely change the semantic of what the HUD trails display (in fact, it appeared to me that you had given up and intended only to improve the display of the current semantic), and I apologize for the confusion - it is not my intent to come across as adversarial, merely to clearly communicate the apparent issue with the code.

> You say "when the player is moving", but this is the crux of the matter. On earth, we think of things as moving or being stationary, but in space there is no distinction. Everything is moving, all the time, relative to something. Hence why you have frames of reference in the Pioneer code. I'm sure you understand this, probably better than me.
> 
> So, when it comes to drawing a HUD trail, it's not enough just to think about "how things moved", the question is always "how things moved relative to something else". The HUD trail will show the recent trajectory of an NPC ship relative to... something. So now the question is - relative to what?
> 
> There are basically two choices here. We can either show movement relative to an existing Pioneer frame of reference, or relative to the player. The existing HUD trail code does the former, I'm suggesting that the latter is more useful in practice.
>
> [...]
>
> Your example is an NPC ship leaving a long trail in deep space, and then the player flying "through" the trail. I understand what you are getting at, but I think the trail would only appear to curve if you are accelerating towards it while it is being created. Otherwise it will still be a straight line. If you were travelling towards it at constant velocity while it was being created, then it would be a line angled towards you - but this is correct, because in that case the NPC ship's relative motion is a line angled towards you. I agree that the acceleration issue specifically leads to the trail not being 100% correct, but relative speeds are often over 100m/s, while acceleration is a few m/s at most. It just isn't noticeable as an effect when flying around NPC ships. Honestly, I was expecting to have to do some kind of adjustments for changes in velocity, but to my surprise the new simple trails look basically correct in all situations that I've tried.

Now that I understand the intent behind the change, I agree that redefining HUD trails to effectively display a history window of player-relative velocity rather than a history of the target's position is likely a better option, as the current semantic does have issues handling an orbital reference frame with inherent velocity (such as an asteroid belt).

> I guess the truth of the matter is - there is no way to implement HUD trails in a way that looks intuitive in all possible scenarios. My suggestion is that we switch to player-relative trails because it a closer match to intuition in more scenarios, and hence an improvement over the existing way. I invite everyone to try the new HUD trail implementation and see if they agree. At the very least - can they find any instances where the new code does something obviously wrong? I tried moving my own ship over a frame boundary while watching an NPC ship, and the new HUD trail was a clean unbroken line, so I still don't know how the player's orientation frame is supposed to make it go wrong. I did realise that there is code to reset the trail when this happens, which is now no longer necessary, so I've removed that.

The main scenario I can think of that would be noticeably "wrong" to a player is if they are orbiting around a planet in tandem with another NPC ship and using the "follow position" autopilot option to remain at exactly the same relative position. In this case, the ship would not have any HUD trail at all (as its relative position is the same frame-to-frame), even though both ships are obviously moving at high speeds with regards to the orbited body.

If the intent is for HUD trails to semantically display only a target's velocity with respect to the player, this becomes less visibly wrong, though the tradeoff is that the HUD trails can no longer display the flight path of e.g. a scissors maneuver (image below), and would instead render the HUD trail as effectively a single 2-dimensional flat circle perpendicular to the flight path of both ships.

image

sturnclaw:
> I've done some testing and this seems to handle transition between Mars and Mars high frames ok. It's actually pretty tough to get taht to happen without using the ingame `Ctrl+i` debug menu I think.

Looking at `Space.cpp`, the only oriented frames are planets' rotation frames - they are oriented by current rotation + axial tilt of the body, while the orbital frame and space station frames use the identity matrix for their orientation. I suspect switching between the surface and orbital frame of Mars with a ship following you at +20km or so would show a very visible discontinuity.

sturnclaw:
image

As suspected, as soon as you cross the boundary between a rotating and non-rotating frame the old points are drawn in a completely incorrect orientation. It appears (have not rigorously tested it) that HUD trails are always rendered with `m_transform` being the model-view transform between the transform space of the player's current frame and the camera view space, so when the orientation of the player's frame changes due to a frame transition, the position of the previously-recorded points are incorrect.

(For reference: the trail *should* display a line "under" the player pointing towards the ship, as it flew exactly opposite to its current velocity shortly before the screenshot.)

sam52796:
Thanks. I was able to reproduce those circumstances, and saw the same thing. My latest commit fixes it by ensuring that the relative position is always calculated using a non-rotating frame of reference. I've tested this in three scenarios - deep space, switching from station to planet orbit, and switching from planet surface to planet orbit. It looked correct in all three.

If you are happy with this approach then I can squash the commits into one. (I left them now so you can see what I just changed.)

sturnclaw (approved):
I'm happy with the changes, thanks! I'd suggest an "extra mile" implementation would be to test `frame->IsRotFrame()` and only conditionally do the extra subtraction as >40% of the player's playtime will be in a non-rotating frame. However, the code is sufficiently performant as to not require that optimization for merge.

sturnclaw:
This PR is ready for merge once it has been squashed.

sam52796:
Thanks @sturnclaw, the commits should all be squashed now.

PR #6318, opened 2026-04-12
Fix for Barnard's Star display issues

### LINKED ISSUE #6293: Barnard's Star Bugs
### Observed behaviour
The surface pressure of Barnard's star is not rounded correctly and the system security shows "[NO_JSON]SYSTEM_SECURITY_0". I haven't observed this in the other systems so far.

### Expected behaviour
This bugs should not appear.

Image

Image

Pioneer 20260203 Flatpak on Debian 13 Trixie.

[output.txt](https://github.com/user-attachments/files/25195577/output.txt)

sam52796:
Fixes #6293. The security rating display issue was that lawlessness was specified as 0/100 in the system definition, but the code expects the lowest possible value to be 1/100. Changed the system definition to 1/100 and also added a "min 1" check to the code to prevent recurrence.

For the surface pressure stat, actually all stars in the game were showing wrong values. I've added a default pressure calculation for stars that don't specify any atmosphere parameters, based on their surface gravity, to fix this.

### DIFF
--- a/src/galaxy/SystemBody.cpp
+++ b/src/galaxy/SystemBody.cpp
@@ -292,6 +292,8 @@ double GetMolarMass(SystemBody::BodySuperType superType)
 {
 	if (superType == SystemBody::SUPERTYPE_GAS_GIANT)
 		return 0.0023139903; // molar mass, for a combination of hydrogen and helium
+	else if (superType == SystemBody::SUPERTYPE_STAR)
+		return 0.00125;	// approximation of the sun's molar mass (in kg/mol) - mostly hydrogen
 	else
 		// XXX using earth's molar mass of air...
 		return 0.02897;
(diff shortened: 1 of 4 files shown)


src/galaxy/SystemBody.cpp
```diff
@@ -328,6 +328,24 @@ void SystemBody::SetAtmFromParameters()
...
+		// Calculate one now based on the star's data so we can display something sensible.
+
+		double temperature = m_averageTemp;
+		if (temperature < 250) temperature = 250;
+
+		// We don't have a good way of determining a star's surface pressure, but the surface
+		// pressure for our Sun is about 0.01 atm, so divide that by its surface gravity value
+		// (274) to get a number that we can use to scale target pressure by star's gravity.
+		// This gives us the result that more dense or bigger stars have higher surface pressure.
+		double target_pressure = (0.01 / 274.0) * this->CalcSurfaceGravity();
```
sturnclaw: Is there a reason for the redundant use of `this->` in the expression?

src/galaxy/SystemBody.cpp
```diff
@@ -328,6 +328,24 @@ void SystemBody::SetAtmFromParameters()
...
+
+		double density = target_pressure * 101325.0 * gasMolarMass / (GAS_CONSTANT_R * temperature);
```
sturnclaw: Because this PR introduces an approximation for stellar gases, I'd recommend adding an additional branch in `GetMolarMass` that handles `SUPERTYPE_STAR` and returns an approximate molar mass for the gases expected in an average star's "atmosphere".

sturnclaw (changes requested):

sturnclaw (approved):
I would ideally like if you separated your original commit into the Barnard's Star changes as one commit and the SetAtmFromParameters changes in a second commit (optionally squashing your latest commit into that one). Other than that, this looks good - thanks for working on it!

sam52796:
My git-fu is weak, but I believe the commits are now the way you requested.

impaktor:
> My git-fu is weak, but I believe the commits are now the way you requested.

Looks good to me.
But if we're to be nitpicky, can you please reword the commit with very long title, to shorter title < 70-80 character? and then put additional comments in the body of the commit message?

That way I can read the commit messages better (on my Commodore 64 which is limited to 80 characters)

could e.g. be:
```
Add default stellar atmospheric pressure calculation

Add a default atmospheric pressure calculation for stars that don't specify any atmosphere parameters. 
Also add a default molar mass for their coronas.
```

sam52796:
My git-fu gains in strength. Soon your puny Commodore 64 will be no match for me.

impaktor:
First line is OK, but now second line overflows my monitor (granted, here it's in 40 column mode but still doesn't fit on not even two lines):
overflow


I'd advice:
Do 
```git log```
and see how it usually looks.
Though, I see both @sturnclaw  and @fluffyfreak have  some commits with "screen overflowing long lines", so I guess no change is needed for this PR, but for future reference, most of the commits have hard line breaks somewhere in the 74-80 character zone.

sturnclaw:
Traditional git commit message first-lines are 50 characters long, with message bodies manually wrapped at 80 characters. Because I use modern software to view commit messages which are capable of word-wrapping based on available space, I am not particularly picky about commit message descriptions.

However, I strongly prefer that the first line of a commit message be kept to <80 characters for the simple reason that it is intended to be a concise summary of the changes in the commit - brevity is the watchword of the day here.

A long commit name typically has a much poorer signal-to-noise ratio when reading a list of commits (to e.g. bisect a bug or get a feel for the changes in a branch).

sturnclaw (approved):
Code looks clean to me, thanks! I don't think this will break saved games, so it's good to merge.

Issue #6207, opened 2025-09-10
Segmentation fault while auto-piloting to Orbizzi Refinery, Shao, Kahariel


boomlinde:
### Observed behaviour
The game suddenly crashes when getting close to the body in the title.

### Expected behaviour
No crash

### Steps to reproduce
1. Load the provided save file (uncompressing it first)
2. Take the PASSAGE WANTED mission at the current starport
3. Plan a direct route to Kahariel
4. Leave the spaceport
5. Engage hyperspace travel
6. Select Dock with Orbizzi Refinery in the overview window
7. Select 10000x time compression
8. Wait for the ship to get close and observe the crash (hopefully)

__My pioneer version (and OS):__

 pioneer 20250501, Void Linux, x64, glibc 2.41

__My output.txt (required) and game save (optional, but recommended)__

* [output.txt](https://github.com/user-attachments/files/22261492/output.txt)
* [savefile.zip](https://github.com/user-attachments/files/22261513/savefile.zip)

sturnclaw:
Closing as likely fixed by #6319.

Issue #5831, opened 2024-05-01
When *10000 speed button pressed, game crashes


hermonochy:
When the button for *10000 speed is clicked after leaving hyperspace, the entire game crashes, giving this error message:`Segmentation fault (core dumped)
`

impaktor:
Is it reproducible?
Are you on 2024-03-14 version?

hermonochy:
Yes, I am on pioneer 2024-03-14.This error also comes up when trying to dock at a spaceport, but space stations work.

impaktor:
I don't think this will go anywhere unless we get a [backtrace](https://wiki.pioneerspacesim.net/wiki/FAQ#Debug_backtrace), or it can be reproduced, (and backtraced).

Clicking time accel after leaving hyperspace is a pretty standard thing to do, that seems to work 99% (100%?) of the time for others.

PR #6319, opened 2026-04-12
This seems to reduce the occurrence of random crashes.

### LINKED ISSUE #6126: Crash after killing second enemy in combat mission
### Observed behaviour
Game crashes after ship splodey

### Expected behaviour
Game doesn't crash after ship splodey, I go somewhere else and collect payment

### Steps to reproduce
Not sure if reproducible, will try to reproduce and upload the save

1. Take a combat mission
2. Destroy someone in self defense
3. Destroy someone else in self defense

20250501 Flatpak
Fedora Secureblue

[output.txt](https://github.com/user-attachments/files/20029850/output.txt)

sam52796:
Fixes #6126, maybe. And possibly #6207. And perhaps #5831.

When running in debug mode I often see crashes when Space::TimeStep() is iterating through m_bodies in order to run TimeStepUpdate() on them. Since this function can add entries to m_bodies, it makes sense that this would cause occasional problems, as per the comments elsewhere in Space::TimeStep(). So I applied the same logic to the place where I see the crash, and one other place too, just in case.

However, this is hard to test, since the crashes are random. On the other hand, I don't see how this fix could make things worse.

### DIFF
--- a/src/Space.cpp
+++ b/src/Space.cpp
@@ -1068,8 +1068,10 @@ void Space::TimeStep(float step)
 
 	Frame::CollideFrames(&hitCallback);
 
-	for (Body *b : m_bodies)
+	for (size_t i = 0; i < m_bodies.size(); ++i) {
+		auto b = m_bodies[i];
 		CollideWithTerrain(b, step);
+	}
 
 	// update frames of reference
 	for (Body *b : m_bodies)
@@ -1090,7 +1092,8 @@ void Space::TimeStep(float step)
 	}
 	Frame::UpdateOrbitRails(m_game->GetTime(), m_game->GetTimeStep());
 
-	for (Body *b : m_bodies) {
+	for (size_t i = 0; i < m_bodies.size(); ++i) {
+		auto b = m_bodies[i];
 		b->TimeStepUpdate(step);
 	}
 


sturnclaw (approved):
Good catch, thank you!

---

@fluffyfreak I dug into callsites that trigger `Space::AddBody` outside of `StaticUpdate` - this fixes potential crashes that could arise from the new cargo spawning queue, HyperspaceCloud's transition of a ship to normal space, and any spawning (missile or ship) that occurred during a callback directly executed during the `TimeStepUpdate` loop. That last category doesn't seem to be one we encountered at any point due to the deferral of Lua event callbacks.

The underlying reason is that under some circumstances, spawning a ship or cargo body could trigger resize of the `m_bodies` array, invalidating all iterators, including the implicit "next" iterator as a result of incrementing the current range-for iterator.

impaktor:
@sturnclaw  Do we also close #6207 ?

@sam52796 I'm too late, and you probably know by now, but first line of commit message should be "Commodore C64"-friendly. or joking aside: not too long, I dunno what the agreed standard is but below 80 characters is good.

sturnclaw:
@impaktor Yes, this is almost certainly the cause of almost all of the random/unreproducible "crash after X time" bugs we get, as `std::vector` resize thresholds are per-platform implementation details.

PR #6321, opened 2026-04-13
Fix for a bug where TraceRay could end up using stale static geometry.

### LINKED ISSUE #6320: Crash after delivering fuel to empty ship
The game crashes if I do a _ship out of fuel_ type mission. I have done these twice and both crashed immediately afterward. I can land beside the target ship, there is a moment for the 'fuel transfer' sequence, then it crashes. Both of those missions were on a planet's surface, not in orbit. I have not attempted one of these missions in orbit.

The target ships of the two missions were different types. The first was a skipjack and this time it was a wave.

### Steps to reproduce - Take the mission, get the hydrogen, land next to the target ship and wait.

I am playing on Linux Mint with version 20260203, the one from the flatpak repo available on the KDE application catalogue. 

[output.txt](https://github.com/user-attachments/files/26665480/output.txt)
[_quicksave.zip](https://github.com/user-attachments/files/26665539/_quicksave.zip)

Note I have added .zip to the save file because it seems github won't accept it without an extension. That save file should be my ship landed at the station where I took the mission, but the mission is not yet taken. I had a feeling I wouldn't be able to complete it without a crash (like my previous experience) so I saved and then took the mission.

sam52796:
Fixes #6320. The crash is observed just after a successful refuelling mission on a planet's surface, because the target ship is initially "static geometry", it is immobile. After the player transfers some hydrogen, the target ship is then able to take off, transitioning into dynamic geometry. There is therefore a one-tick window when the static geometry array is stale, but still being looped by CollisionSpace::TraceRay(). My fix is to add a check for staleness before starting that loop, similar to the check for dynamic geometry staleness immediately afterwards in the same function.

### DIFF
--- a/src/collider/CollisionSpace.cpp
+++ b/src/collider/CollisionSpace.cpp
@@ -121,11 +121,11 @@ void CollisionSpace::TraceRay(const vector3d &start, const vector3d &dir, double
 	std::vector<uint32_t> isect_result;
 	isect_result.reserve(8);
 
-	if (m_enabledStaticGeoms > 0) {
+	if (m_enabledStaticGeoms > 0 && !m_needStaticGeomRebuild) {
 		m_staticObjectTree->TraceRay(start, invDir, len, isect_result);
 
 		for (uint32_t &idx : isect_result) {
-			assert(idx <= m_staticGeoms.size());
+			assert(idx < m_staticGeoms.size());
 			Geom *g = m_staticGeoms[idx];
 
 			if (g != ignore)


sturnclaw (approved):
Good catch, thanks! A similar optimization to `CollisionSpace::RemoveGeom` can be applied to `CollisionSpace::RemoveStaticGeom`, checking whether the static geom is enabled and avoiding a rebuild if necessary.

In general, there seems to be a minor bug where a Geom's enabled/disabled state is not correctly propagated for static Geoms - the dynamic BVH tree is rebuilt every frame but no rebuild is triggered for static Geoms.

However, it looks like static bodies do not make use of the disabled state at this time.

impaktor:
@sam52796 very nice to see your activity here in PR-space! Please let us know if you want your name in AUTHORS.txt, and if so, what it should be.

PR #6323, opened 2026-04-14
Clear the next lifecycle variable after using it, so that it doesn't then repeat when it shouldn't

### LINKED ISSUE #6174: Quitting the game sometimes causes the death animation to play
### Observed behaviour
Sometimes, when quitting the game, the death animation plays before returning to the main menu despite the player being still alive.

### Expected behaviour
The game directly returns to the main menu without playing a death animation, since the player is still alive and has merely quit the game.

### Steps to reproduce
I have not yet found the conditions under which this occurs, but so far it only seems to occur when my game sessions are long (>1 hour). I will report back if I find more information and I manage to capture get an output.txt after this occurs.

sam52796:
Fixes #6174. Clear the next lifecycle variable after using it, so that it doesn't then repeat the next time the active lifecycle finishes. This fixes an issue whereby the spinning tombstone animation would be seen if the player previously died, and then starts a new game and ends it without dying.

I've been running the game on and off with this fix for a few days, and it doesn't seem to have broken anything.

### DIFF
--- a/src/core/Application.cpp
+++ b/src/core/Application.cpp
@@ -138,6 +138,7 @@ void Application::EndLifecycle()
 	// wait until we've finished the control flow for the lifecycle;
 	// the lifecycle may decide to set the next lifecycle in End()
 	m_priorityLifecycle = m_activeLifecycle->m_nextLifecycle;
+	m_activeLifecycle->m_nextLifecycle.Reset();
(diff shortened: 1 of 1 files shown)


sturnclaw (approved):
`m_nextLifecycle` is indeed intended to be a transient cache, thanks!

PR #6325, opened 2026-04-15
Fix docking clearance timeout and docking position issue with stale clearance

### LINKED ISSUE #6324: Docking moved the ship to the center of the planet.
I docked on a surface starport, and the ship moved to the center of the planet.

Image

This was my first time interacting with this planet at its surface. I reloaded and did it a second time and the same thing happened. The station radio feature works normally but trying to launch destroys the ship.

[cant_dock.zip](https://github.com/user-attachments/files/26746163/cant_dock.zip)
Note- remove .zip extension, added only for uploading.

Steps to reproduce: Dock at the starport in front of the ship.

The space station at about 65mm was bugged also in that I got jittery prograde marker when I approached it. Every time that happened in the past, I was unable to dock with the affected station so I didn't even attempt to this time. This bug report isn't about the station in orbit, I just wanted to point out there is more bugginess related to this planet, perhaps the things are related. I'll get a screenshot next time I find another one of those bugged orbital stations.

Pioneer version 20260203 from flathub running on Linux Mint.

[output.txt](https://github.com/user-attachments/files/26746216/output.txt)

sam52796:
Fixes #6324. 

If you request docking clearance from a station, and then don't dock, but fly to a different station and dock there, the code would try to use the docking pad position from the first station. This is particularly noticeable when the first station is orbital and the second station is on a planet, because it would put you in the centre of the planet (offset by one docking pad thickness).

We can fix this by ensuring that ships docking positions only get updated after DockStage::DOCKED is available to set the correct pad location. 

An ancillary problem is that docking clearances never expire. There is already code to expire docking clearance and show a comms message, but due to missing the timestep increment it never triggers. So I added a small increment to make docking clearance expire after 10 minutes. Now the rest of the code works as expected.

### DIFF
--- a/src/SpaceStation.cpp
+++ b/src/SpaceStation.cpp
@@ -592,6 +592,7 @@ void SpaceStation::DockingUpdate(const double timeStep)
 				m_doorAnimationStep = -0.3; // close door
 				SwitchToStage(i, DockStage::NONE);
 			}
+			dt.stagePos += timeStep / DOCKING_CLEARANCE_TIMEOUT;
 			continue;
 
 		default:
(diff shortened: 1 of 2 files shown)


sturnclaw (approved):
Thanks for tracking this bug down!

Looking at the code, `dt.stagePos` is reset when switching to the `DOCKING` stage, which is the next stage that will use the `stagePos` variable, so there's no worry about its use here causing the first animation stage to be partially or fully skipped.

Ideally we'd like the clearance timeout to be configurable from script, but this is an adequate solution and someone else can migrate it to a class-level variable if they have a gameplay need for it to have a different timer.

The autopilot *should* be able to re-request clearance if it times out on approach; if this is not the case the fix should be in a separate PR.

PR #6327, opened 2026-04-16
Don't mess with velocity if the ship just docked

### LINKED ISSUE #6215: Target/Frame of reference - Relative speed when landed/docked
You would expect the speed relative to the target or frame of reference to be 0 when you are docked/landed. This doesn't seem to be the case sometimes. I believe the issue may be only present when time acceleration is set to more than 1.

### Observed behaviour
With time acceleration
Image

### Expected behaviour
No time acceleration
Image

### Steps to reproduce
Start a game at Cydonia and go to Bradbury landing with time acceleration on max.
On landing, note the speeds in the reticule. If in the last second before landing you switch the time acceleration to x1, you'll be fine.

Pioneer master (1dd9c03fdbc90356c301977fa5317697894c2e06) - The issue is present on at least the last 150 commits.
Ubuntu Mate 24.04

[output.txt](https://github.com/user-attachments/files/22617226/output.txt)

sam52796:
Fixes #6215. 

Just after a successful docking, the time acceleration gets set to 1x, but the code that runs during time acceleration changes can apply one last burst of movement to the player's ship, the values of which then get stuck in the HUD display (and potentially trigger ground proximity warnings). Adding a filter to this time acceleration change code, so that it only applies to ships in flight, fixes this issue.

### DIFF
--- a/src/Game.cpp
+++ b/src/Game.cpp
@@ -721,11 +721,14 @@ void Game::SetTimeAccel(TimeAccel t)
 		m_player->SetAngThrusterState(vector3d(0.0));
 	}
 
-	// Give all ships a half-step acceleration to stop autopilot overshoot
+	// Give all currently flying ships a half-step acceleration to stop autopilot overshoot
 	if (t < m_timeAccel)
 		for (Body *b : m_space->GetBodies())
-			if (b->IsType(ObjectType::SHIP))
-				(static_cast<Ship *>(b))->TimeAccelAdjust(0.5f * GetTimeStep());
+			if (b->IsType(ObjectType::SHIP)) {
+				Ship *ship = static_cast<Ship *>(b);
+				if (ship->GetFlightState() == Ship::FLYING)
+					ship->TimeAccelAdjust(0.5f * GetTimeStep());
+			}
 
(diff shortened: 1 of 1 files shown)


zonkmachine:
I can't review the code but I retested the issue and can confirm this PR fixes it. Thanks!

fluffyfreak (approved):

sturnclaw (approved):
Code looks good - my only semantic nitpick is that if the flight control state has already changed, the velocity should have ideally been zeroed out by the function that changed state from `FLYING`->`DOCKED`. This is an acceptable solution however, as the station would have flight state `DOCKING` ships on rails already and not need the bump.

sam52796:
FYI - the function that "docks" the ship _does_ zero out the velocity and sets the position to the landing pad. That already all works correctly. The issue is caused when the player has time acceleration active, and it gets toggled to 1x speed just after the docking code has all run, and some code in the SetTimeAccel() path applies an additional half timestep based on m_lastForce, which hasn't been zeroed out. Having the docking code zero out m_lastForce doesn't feel right to me though, because it holds a valid value for the last force. Admittedly I didn't look into how that is used too deeply though. So the fix instead avoids applying the additional half timestep if the ship isn't flying.

PR #6328, opened 2026-04-18
Add string for missing hypercloud analyzer message

sam52796:
First commit is to add a missing string. When you click on a hyperspace cloud without an analyzer installed, you see [NO_JSON]NO_HYPERCLOUD_SCANNER_INSTALLED in the comms messages, because the code thinks there is a string with that name, but there isn't. I took the opportunity to rename the string from "scanner" to "analyzer" because everywhere else it's called an analyzer.

Second commit is to make the equipment name of the analyzer consistent. Most of the game refers to "hyperspace clouds" but the equipment name itself is "Hypercloud Analyzer". Therefore, the UI tooltip that refers to that piece of equipment should call it that too. I updated it in all language files that still have the English string. Curiously, some non-English languages are already consistent with themselves - German, Spanish, Romanian, and Russian, while Chinese and Italian are different only in a minor "choice of translation" way. All the other language files have faithfully preserved the difference between "hyperspace cloud" and "hypercloud". So basically, it's already a bit of a mix, but I think this change nudges things in the right direction.

Alternatively, we should rename the Hypercloud Analyzer to be called the Hyperspace Cloud Analyzer. Then it will be consistent with all the strings that say things like "Hyperspace departure cloud".

Let the debate begin!

### DIFF
--- a/data/lang/core/en.json
+++ b/data/lang/core/en.json
@@ -1405,7 +1405,7 @@
   },
   "SET_HYPERSPACE_TARGET_TO_FOLLOW_THIS_DEPARTURE": {
     "description": "",
-    "message": "Hyperspace cloud analyzer: Set hyperspace target to follow this departure"
+    "message": "Hypercloud analyzer: Set hyperspace target to follow this departure"
   },
   "SET_LOW_THRUST_POWER_LEVEL_TO_X_PERCENT": {
     "description": "",
--- a/data/lang/core/en.json
+++ b/data/lang/core/en.json
@@ -1035,6 +1035,10 @@
     "description": "",
     "message": "No established order"
   },
+  "NO_HYPERCLOUD_ANALYZER_INSTALLED": {
+    "description": "",
+    "message": "No hypercloud analyzer installed"
+  },
   "NO_HYPERDRIVE": {
     "description": "",
     "message": "No hyperdrive"
--- a/data/pigui/libs/radial-menu.lua
+++ b/data/pigui/libs/radial-menu.lua
@@ -93,7 +93,7 @@ local radial_menu_actions_hyperspace_cloud = {
 					ui.playSfx("OK")
 				end
 			else
-				Game.AddCommsLogLine(lc.NO_HYPERCLOUD_SCANNER_INSTALLED)
+				Game.AddCommsLogLine(lc.NO_HYPERCLOUD_ANALYZER_INSTALLED)
 			end
 		end
 


mwerle:
At the risk of being more verbose, I prefer "Hyperspace Cloud" rather than "Hypercloud". But, being very much dormant at present I will leave the final decision up to the active contributorship.


Apart from that, good PR, thank you for standardising all the strings. No idea how the missing string got committed.. I thought I had tested everything thoroughly at the time but it seems that it did get missed - offending commit 745b47950f74e6cf42114a5538fa3dc572615c4b

data/lang/core/af_ZA.json
```diff
@@ -1401,7 +1401,7 @@
   },
   "SET_HYPERSPACE_TARGET_TO_FOLLOW_THIS_DEPARTURE": {
     "description": "",
-    "message": "Hyperspace cloud analyzer: Set hyperspace target to follow this departure"
+    "message": "Hypercloud analyzer: Set hyperspace target to follow this departure"
```
mwerle, line 1404: Personally would prefer standardising on "Hyperspace cloud" but up to the more active contributorship.

mwerle (approved):
LGTM apart from the naming but I leave the final decision to others.

Please wait for at least one other reviewer's comment before merging.

sam52796:
Thanks. I note that "hypercloud analyzer" is already referenced by name on the Pioneer wiki, in release notes, and so on. Any change to the equipment name would have to include updating the wiki pages, at least.

Therefore, it would be easier to leave the equipment name as "Hypercloud Analyzer", to maintain consistency with the wiki. However, if the wiki maintainers are up for changing all the references there, I think "Hyperspace Cloud Analyzer" would be fine too. I guess I don't feel strongly about that one.

What I do feel more strongly about is that, particularly on the solar system map, the string "Hyperspace departure cloud" is quite long and clutters up the view. Planets are labelled just by their name. Ships are labelled something like "QQ-1234". So something shorter would fit well. Each hyperspace cloud already has a wormhole icon, so it's clear what they are. Therefore, I think it would be preferable if all the hyperclouds were labelled simply "Departure", "Arrival", or "Arrival (remnant)".

Happy to make that change if a consensus is reached.

sturnclaw:
While I personally feel `Hyperspace Cloud Analyzer` is clearer, it is also significantly more verbose (and I shudder to think how many characters it will be in e.g. German), which can be a problem for the low resolutions a lot of users play at.

I consider it better to accede to inertia on this matter and standardize as `Hypercloud Analyzer`, with the equipment description (which most equipment are lacking at present) written to indicate that it is an equipment item which analyzes hyperspace entry and departure clouds. Essentially, this would define `Hypercloud` as an in-universe trade name for hyperspace transit signatures.

From a technical standpoint, please only update the `en.json` files when modifying language strings. Transifex automates (and desires exclusive control over) the other language `.json` files; modifying a string in `en.json` will mark that string "dirty" and it will show as needing to be re-translated in the Transifex interface for all languages we support. Manually modifying the other language files can cause issues with Transifex's per-string tracking.

> Therefore, I think it would be preferable if all the hyperclouds were labelled simply "Departure", "Arrival", or "Arrival (remnant)".

This is a great topic for a new PR! It would be preferable if hyperspace clouds were indicated with a specific icon in the system map, at which point Departure/Entry would be acceptable labels.

data/lang/core/en.json
```diff
@@ -1401,7 +1405,7 @@
   },
   "SET_HYPERSPACE_TARGET_TO_FOLLOW_THIS_DEPARTURE": {
     "description": "",
-    "message": "Hyperspace cloud analyzer: Set hyperspace target to follow this departure"
+    "message": "Hypercloud analyzer: Set hyperspace target to follow this departure"
```
sturnclaw, line 1408: Tiniest of nitpicks: if "Hypercloud Analyzer" is a trade name, it should be capitalized as a proper noun in this sentence.

sturnclaw (approved):
I would love to revisit this decision once we have a chance to work on proper at-runtime UI scaling and automatic UI string abbreviation/fit-to-width scaling. However, I think standardizing existing practice is a reasonable solution in the interim.

Thank you for finding this and taking initiative on it @sam52796!

bszlrd:
As a sidenote to @sam52796 , I'd invite you to our IRC channel, if you would like to. :) You've did quite some work the past couple of weeks, don't hide! Sometimes the channel can feel deserted, but then we ramp up and can't stop talking, and made a lot of huge design leaps and such, so we can be a fun bunch!

Issue #6302, opened 2026-03-08
Crash on the assassination mission


SilentSKeptic:
### Observed behaviour
I came into system "Gliese 747" at Alamai planet. Planet seems buggy. Its impossible to make orbit on it. Like planet hasn't mass. I tryed to land in space port manually, but fell through landing platform. I can't land into station on orbit also. Autopilot goes crazy. After all I skipped time in the nearly system and jump here to catch mission event. I came to the planet at the time, when my target should leave space port. When time was closer to event, the game shows more and more statters. And then was crash with error.

### Steps to reproduce

__My pioneer version (and OS):__
20260203 (4b000c8)
My OS. Uh. Proton10.0-ARM64x-2
So I play through emulator gameshub 5.3.5 from android.
But the error was about the mission.
"modules/Assassination/Assassination.lua:519: Can't undock if not already docked"
And I happily live in the game 1.5 year by game time. So I don't think that problem is in my not ordinary gaming device.

__My output.txt (required) and game save (optional, but recommended)__

[opengl.txt](https://github.com/user-attachments/files/25823704/opengl.txt)
[output.txt](https://github.com/user-attachments/files/25823703/output.txt)

![Image](https://github.com/user-attachments/assets/2f093cb7-8ae0-4843-a823-8149188b54ca)
![Image](https://github.com/user-attachments/assets/93e1c24a-bc82-4a77-8b33-3c04700c11a6)
![Image](https://github.com/user-attachments/assets/201a7122-f34f-43d0-8ea3-db80d5e9e2e0)

SilentSKeptic:
[save.zip](https://github.com/user-attachments/files/25824367/save.zip)

sam52796:
I'm pretty sure that this is the same problem as #5885, since one of the nearby planets has a large number of moons with implausible orbits that are so big that they overlap with the star. This would confuse the frame determining logic, resulting in not being able to enter reference frames or collide with things. 

This is confirmed when I visit the system, and I see the same observed behaviour as OP - unable to enter the frame of the planet or land on the starport. Applying the proposed fix for #5885 makes the system map look much more reasonable, and then I can land on the planet surface starport as normal.

impaktor:
Closing since duplicate, and PR fix inbound as noted above.

PR #6329, opened 2026-04-20
Fix frames in binary star systems

### LINKED ISSUE #6312: No frame of reference in binary systems, can't enter planet's frame in some cases
### Observed behaviour
In binary (or so) systems you can exit the star's frame of reference. I've noticed it in Cancri 55 - Bathlight (3,3,2 - custom system), and at a pretty low distance of 2.2 AU from the star. 

Image
Image

The frame does not switch to the common gravpoint, and then you can not fly into the frame of the gas giant Eternal 4 AU from the star. You can not collide with the planet either.
You can fly back to the stars frame though, or into the frame of the companion star.
The primary star has about the same gravity as the Sun, so its frame should not end at 2 AU.
Happens with Procyon (1,0,0) too, which is a custom system, but only the coordinates and the star types are set in local_stars.json.
Happens in fully generated systems as well. The second attached save is from such system. 
It does not seem to happen in single systems.

#5885 #5830 #5839 #4988 #4164 might be related

### Expected behaviour
Game to switch into grav point frame, or at least switch to the planet's frame if that's easier to solve.

### Steps to reproduce
Do a custom start at Procyon (1,0,0) or Cancri 55 (3,3,2), in-orbit, of any planet. Fly outwards to leave the frame. Or to Eternal in Cancri 55's case, for the not entering its frame part. 

Win10, 20260203 release build (4b000c8). 
I have the pebblegarden skybox and honk mods activate. Neither should give issues I think, but a recovered save or custom start should also work Might mean a bit more flying.

In this I saved just before exiting the frame
[cancri 55 no frame error.txt](https://github.com/user-attachments/files/26355476/cancri.55.no.frame.error.txt)
[generated - no frame error.txt](https://github.com/user-attachments/files/26355473/generated.-.no.frame.error.txt)
[output.txt](https://github.com/user-attachments/files/26355815/output.txt)

sam52796:
Fixes #6312.

There were two problems with binary star systems.  

Firstly, the check to ensure that star frames in binary star systems don't overlap with each other was too conservative, resulting in some child bodies being outside their parent's frame. This confused the code that expects a nice consistent frame hierarchy. The first commit rearranges the radius checks to ensure that frames always encompass their children. The problem seems to only occur in custom binary systems, where the designer might have put planets further out than the procedurally generated code would place them. But now you have a better chance of entering frames correctly in custom binary systems.

The second issue isn't really an issue, but when the player is in a grav point frame, the HUD just doesn't show any frame of reference at all, which can be confusing, even when the frames _are_ working correctly. So the second commit displays the frame label in these cases, to reassure the player that everything is still okay. Actually having a full velocity and position readout would require deeper changes to the code, which assumes that there is no _Body_ object for gravpoints, while all the existing HUD display properties refer to _Body_ attributes. So this is kind of a patch, but a relatively safe one.

Note that for #6312 specifically, since the system body tree is stored in the save file, if you load that game after this fix is applied you will still see the problem. You have to jump out of the system and then back in before you can crash into Eternal.

### DIFF
--- a/data/pigui/modules/flight-ui/reticule.lua
+++ b/data/pigui/modules/flight-ui/reticule.lua
@@ -279,7 +279,7 @@ local function updateReticuleTarget(frame, navTarget, combatTarget)
 	end
 
 	if reticuleTarget == "frame" then
-		if not frame then
+		if not frame and not player.frameLabel then
 			reticuleTarget = nil
 		end
 	end
@@ -389,8 +389,24 @@ local function displayDetailData(target, radius, colorLight, colorDark, tooltip,
 
 end
 
+-- display only frame name in the right-side "detail" HUD section
+local function displayDetailFrameLabel(label, radius, colorLight, colorDark, tooltip)
+	local uiPos = ui.pointOnClock(center, radius, 2.46)
+	-- frame label
+	ui.addStyledText(uiPos, ui.anchor.left, ui.anchor.baseline, label, colorDark, pionillium.medium, tooltip, colors.lightBlackBackground)
+end
+
 -- display data relative to frame left of the reticule circle
-local function displayFrameData(frame, radius)
+local function displayFrameData(frame, frameLabel, radius)
+	local uiPos = ui.pointOnClock(center, radius, -2.46)
+	-- label of frame
+	local label = frameLabel
+	if frame then label = frame.label end
+	if not label then return end
+	-- if we're not in a framebody then just display the label
+	ui.addStyledText(uiPos, ui.anchor.right, ui.anchor.baseline, label, colors.frame, pionillium.medium, lui.HUD_CURRENT_FRAME, colors.lightBlackBackground)
+	if not frame then return end
+
 	local velocity = player:GetVelocityRelTo(frame)
 	local position = player:GetPositionRelTo(frame)
 	local altitude = player:GetAltitudeRelTo(frame)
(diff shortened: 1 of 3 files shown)


fluffyfreak:
The reason for the extremely conservative check on frame size is that we can have an opposite problem whereby another body with it's own frame is inside the frame of it's parent, or grandparents, frame which prevents ships from correctly switching to that bodies frame.

I think it would be worth testing for this, I think #5885 is an example of it though I believe there are others

bszlrd:
If generated systems are fine, then how about adding the ability to hand-define the frame sizes in costum systems? With visual feedback in the editor.

sam52796:
For physically sensible Keplerian body trees, I think that the frame calculations should always turn out logically correct. So any situation where a custom frame radius is required indicates that the system has a layout issue. I think that allowing a custom frame radius is more likely to increase the occurrence of bugged systems. So I don't see the need for that, unless the game plans to include high-Kardashev civilizations who have edited solar systems to otherwise impossible gravitational configurations. 

However, if this is a common problem then maybe the editor could include a toggle to show frames, so that designers can see in advance if something will go wrong with their current layout.

bszlrd:
I was saying that mostly as a simpler solution, if the other is too complicated. I'd prefer an astronomically correct solution. Showing the frames in the editor would be nice anyhow. 
Unrelated, but it would be nice to have that in the in-game system map as well eventually. At least the point of frame change and predicted orbit form there, like in KSP

sam52796:
Just to clarify, since this is potentially confusing - in the current master branch there are two issues with frames:

One is that in certain circumstances in sensibly-laid-out binary star systems, the code that calculates the frames is slightly too conservative, resulting in some outer planets not being within their parent star's frame. This pull request addresses that.

The second problem is that sometimes systems are not sensibly laid out, for example in #5885 and #6302, where some planets have moons in orbits that span the solar system and overlap with the star itself. In those cases, the frame calculation code has no chance, because there is no logical frame tree possible. The solution in those cases is to fix the system layout, and that's what my other pull request #6330 does.

sturnclaw (approved):

PR #6330, opened 2026-04-20
Fix moon orbit generation

### LINKED ISSUE #5885: Weird system where a planet's reference frame includes it's star

sam52796:
Fixes #5885. 

The procedural generation was going wrong when adding moons to planets whose masses were small compared to their star's mass, because calculating the Hill Radius involves a cube root of the mass ratio, and the fixed point algorithm couldn't handle it. The system generator therefore calculated a far too large orbital slice, and filled it with moons at huge orbits. Doing a bit of parameter scaling fixes this issue, but results in different numbers of moons being generated, and hence probably qualifies for a save bump.

The second commit is to fix binary star orbits in custom systems. The procedural generation appears to get binary star systems correct, but if they are added in the system designer, or by hand, parameters can be left to random chance, and then the stars don't line up in their orbits correctly. Binary stars should always be on opposite sides of the barycentre, i.e. at the furthest possible distance allowed by their orbits, such that a line drawn between the two stars passes through the barycentre. Without this fix, 55 Cancri (for example) has both stars on the same side of the barycentre, which is gravitationally impossible.

While the second commit isn't save-breaking in a technical sense, it could lead to players being stranded if the star they are heading towards gets re-oriented to a different point in the orbit, so it seems like a good idea to lump it in with other save-breaking changes.

### DIFF
--- a/src/fixed.h
+++ b/src/fixed.h
@@ -13,6 +13,7 @@ class fixedf {
 public:
 	static const int FRAC = FRAC_BITS;
 	static const Uint64 MASK = (Uint64(1UL) << FRAC_BITS) - 1;
+	static const fixedf PI;
 
 	fixedf() :
 		v(0) {}
@@ -250,5 +251,11 @@ public:
 };
 
 typedef fixedf<32> fixed;
+
+template<>
+constexpr fixed fixed::PI = fixed(13493037705LL);
+
+template<>
+constexpr fixedf<48> fixedf<48>::PI = fixedf<48>(884279719003555LL);
 
 #endif /* _FIXED_H */
--- a/src/fixed.h
+++ b/src/fixed.h
@@ -224,14 +224,26 @@ public:
 		return (fixedf(root));
 	}
 
-	static fixedf CubeRootOf(const fixedf &a)
+	static fixedf CubeRootOf(const fixedf &a_orig)
 	{
+		fixedf a = a_orig;
+		if (a == fixedf(0, 1)) return fixedf(0, 1);
+
+		// In order to prevent over or underflow in the (x * x) calculation
+		// below, we first scale the parameter into a reasonable range
+		int k = 0;
+		while (a.v > (Sint64(1) << 60)) { a >>= 3; k++; }
+		while (a < fixedf(1,1024)) { a <<= 3; k--; }
+
 		/* NR method. XXX very bad initial estimate (we get there in
 		 * the end... XXX */
 		fixedf x = a;
 		for (int i = 0; i < 48; i++)
 			x = fixedf(1, 3) * ((a / (x * x)) + 2 * x);
-		return x;
+
+		// We've got the cube root of the normalized parameter, now
+		// un-scale it back to get the real answer
+		return k >= 0 ? x << k : x >> -k;
 	}
 
 	Sint64 v;
--- a/src/galaxy/CustomSystem.cpp
+++ b/src/galaxy/CustomSystem.cpp
@@ -1106,7 +1106,30 @@ void CustomSystemsDatabase::RunLuaSystemSanityChecks(CustomSystem *csys)
 			body->bodyData.m_volatileGas = rand.NormFixed(fixed(1050, 1000), fixed(8000, 1000)).Abs();
 			body->bodyData.m_atmosOxidizing = rand.NormFixed(fixed(0, 1), fixed(300, 1000)).Abs();
 		}
+	}
 
+	// This loop runs after the one above, to ensure that system bodies are all set up first.
+	// Now we ensure that any binary star systems have orbits set correctly.
+	for (CustomSystemBody *body : csys->bodies) {
+
+		if ((body->bodyData.m_type == SystemBody::TYPE_GRAVPOINT) && (body->children.size() == 2)) {
+			SystemBodyData *star_A = &body->children[0]->bodyData;
+			SystemBodyData *star_B = &body->children[1]->bodyData;
+
+			if ((star_A->m_type >= SystemBodyType::TYPE_STAR_MIN) && (star_A->m_type <= SystemBodyType::TYPE_STAR_MAX) &&
+				(star_B->m_type >= SystemBodyType::TYPE_STAR_MIN) && (star_B->m_type <= SystemBodyType::TYPE_STAR_MAX)) {
+
+				// This is a binary star system. Ensure that the stars are opposite each other in their orbits
+
+				star_B->m_orbitalOffset = star_A->m_orbitalOffset;
+				star_B->m_inclination = star_A->m_inclination;
+				star_B->m_argOfPeriapsis = star_A->m_argOfPeriapsis;
+
+				star_B->m_orbitalPhaseAtStart = star_A->m_orbitalPhaseAtStart + FIXED_PI;
+				if (star_B->m_orbitalPhaseAtStart > fixed(2, 1) * FIXED_PI)
+					star_B->m_orbitalPhaseAtStart -= fixed(2, 1) * FIXED_PI;
+			}
+		}
 	}
 }
 
--- a/src/galaxy/CustomSystem.cpp
+++ b/src/galaxy/CustomSystem.cpp
@@ -1125,9 +1125,9 @@ void CustomSystemsDatabase::RunLuaSystemSanityChecks(CustomSystem *csys)
 				star_B->m_inclination = star_A->m_inclination;
 				star_B->m_argOfPeriapsis = star_A->m_argOfPeriapsis;
 
-				star_B->m_orbitalPhaseAtStart = star_A->m_orbitalPhaseAtStart + FIXED_PI;
-				if (star_B->m_orbitalPhaseAtStart > fixed(2, 1) * FIXED_PI)
-					star_B->m_orbitalPhaseAtStart -= fixed(2, 1) * FIXED_PI;
+				star_B->m_orbitalPhaseAtStart = star_A->m_orbitalPhaseAtStart + fixed::PI;
+				if (star_B->m_orbitalPhaseAtStart > fixed(2, 1) * fixed::PI)
+					star_B->m_orbitalPhaseAtStart -= fixed(2, 1) * fixed::PI;
 			}
 		}
 	}
--- a/src/galaxy/StarSystemGenerator.cpp
+++ b/src/galaxy/StarSystemGenerator.cpp
@@ -26,7 +26,6 @@ static const fixed SAFE_DIST_FROM_BINARY = fixed(5, 1);
 // very crudely
 static const fixed AU_SOL_RADIUS = fixed(305, 65536);
 static const fixed AU_EARTH_RADIUS = fixed(3, 65536); // XXX Duplication from StarSystem.cpp
-static const fixed FIXED_PI = fixed(103993, 33102);	  // XXX Duplication from StarSystem.cpp
 static const double CELSIUS = 273.15;
 static const fixed ONEEUMASS = fixed::FromDouble(1);
 static const fixed TWOHUNDREDEUMASSES = fixed::FromDouble(200.0);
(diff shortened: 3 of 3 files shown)


sam52796:
I realised that a better fix for the cube root issue was to do the scaling in the cube root function itself, so I've edited the first commit so that it should produce a better result for any caller, instead of relying on the caller to scale where appropriate.

While testing I found another issue which is that moons can be created with 0 kg mass, e.g. the first moon in Gliese 1, so I added a third commit to abort creating a planet/moon if it's going to have a mass of 0. This is probably also save breaking, since now some moons will no longer appear. Now that I think about it, this commit may belong more with my previous "Fix moon populations" pull request, so I can move it there if that's preferred.

src/galaxy/CustomSystem.cpp
```diff
@@ -27,6 +27,8 @@
 
 #include <map>
 
+static const fixed FIXED_PI = fixed(103993, 33102); // XXX Duplication from StarSystemGenerator.cpp
```
fluffyfreak: this might be better moved to a common header such as `StarSystemGenerator.h` which is visible to both, instead of duplicating

src/fixed.h
```diff
@@ -224,13 +224,28 @@ class fixedf {
 		return (fixedf(root));
 	}
 
-	static fixedf CubeRootOf(const fixedf &a)
+	static fixedf CubeRootOf(const fixedf &a_orig)
```
sturnclaw, line 228: I do not have time at current to fully dig into the codepath and analyse the numerical inaccuracy at play here (and the effects of this change on its stability) - however, I'd strongly prefer if the rescaling version of this function be moved into a separate function, e.g. `fixedf::CubeRootOfRescale`. The automatic rescaling may be undesired behavior in codepaths that operate on values with known input bounds, and I'd prefer it to be opt-in.

That being said, I appreciate your tackling of this issue and I'm looking forward to reviewing it in-depth!

sturnclaw (changes requested):

src/fixed.h
sam52796: Are you aware that the cube root function is only used in one other place, and right next to that place there is a comment implying that underflow is a persistent problem for the function?  In fact, that comment is what gave me the idea to use the bit shifting approach in the first place, since it explains how to use bit shifting to fix the problem.

I say this because I think it's better not to clutter up the codebase with duplicate library functions where the caller is expected to know which version to call. I realise that sometimes it's unavoidable when you really don't want to risk breaking things, but this doesn't seem to be one of those times.

src/galaxy/CustomSystem.cpp
sam52796: Since there is nothing star system generation specific about the value of pi, I thought it would be better in a more generic include file. It took all of my self-control not to move the definition of pi to the file called Pi.h, but I settled eventually on fixed.h.

src/fixed.h
sturnclaw: > Are you aware that the cube root function is only used in one other place, and right next to that place there is a comment implying that underflow is a persistent problem for the function?

Yes - I wrote that comment.

> I say this because I think it's better not to clutter up the codebase with duplicate library functions where the caller is expected to know which version to call.

I do agree that duplicating functionality is to be avoided. If this function has noticeable loss-of-precision relative to the original, then there is a material difference between the two functions and thus a reason to separate it as a separate entry point.

After having built a test harness for both versions of the function, the proposed implementation has a worst-case error magnitude increase of 20% when the `while (a > fixedf(1024, 1))` branch is taken, due to discarding __up to 24 bits__ of fractional precision.

The good news is that replacing that branch with `if (a > fixedf(1024, 1)) x = a * fixedf(2, 3);` as the initial estimate experimentally fixes a noticed issue where the `x * x` term can catastrophically overflow (with an input value of `0x4299E58D00000000L`), while retaining the same error magnitude as the original version of the function.

When that is resolved, I agree with your intent that the implementation is a direct upgrade in terms of numerical accuracy and stability and retract my initial recommendation.

src/galaxy/CustomSystem.cpp
sturnclaw: As a future-proofing method, I'd request the constant be defined as `fixedf::PI` instead, as that allows it to have a different definition based on the precision of the fixed-point template specialization instead.

```c++
template<...>
class fixedf {
    static const fixedf PI;
};

template<>
constexpr fixed fixed::PI = fixed(..., ...);

template<>
constexpr fixed fixedf<48>::PI = fixedf<48>(...);
```

src/fixed.h
sturnclaw: For clarity, the `while (a > fixedf(1024, 1))` branch did not present any reduction in computation error (with regards to a double-precision floating-point cube root using `std::pow`) compared to computing the N-R method cube root of the original value directly. Additionally, this increase in error metric was observed in both 16.48 fixed-point as well as 32.32 fixed-point representations.

The only noticed outliers as mentioned above were specific values triggering an overflow that caused the first iteration of the method to wildly overshoot beyond the ability of 47 more iterations to converge. Changing the initial estimate of `x` to a rough estimate of the first iteration result allows large input values to nominally remain within the allowed limits of the fixed-point representation.

src/fixed.h
```diff
@@ -224,13 +224,28 @@ class fixedf {
...
 	{
+		fixedf a = a_orig;
+		if (a == fixedf(0, 1)) return fixedf(0, 1);
+
+		// In order to prevent over or underflow in the (x * x) calculation
+		// below, we first scale the parameter into a reasonable range
+		int k = 0;
+		while (a > fixedf(1024,1)) { a >>= 3; k++; }
```
sturnclaw: This shift is only needed on extremely large values - the N-R method is quite precise as long as the initial value does not overflow significantly. I'd recommend shifting the value out of the high 3 bits (regardless of precision) and then stopping.

Note that this does require the input be a positive number, though I believe taking the cube root of a negative number is already undefined.

```suggestion
		while (a.v > (1L << 60)) { a >>= 3; k++; }
```

src/fixed.h
```diff
@@ -224,13 +224,28 @@ class fixedf {
...
+		while (a < fixedf(1,1024)) { a <<= 3; k--; }
+
 		/* NR method. XXX very bad initial estimate (we get there in
 		 * the end... XXX */
 		fixedf x = a;
 		for (int i = 0; i < 48; i++)
 			x = fixedf(1, 3) * ((a / (x * x)) + 2 * x);
```
sturnclaw, line 243: Per my experimental results, precision can be improved to around `±1e-10` across all results by replacing the multiplication-by-reciprocal with a direct division.

It displays some decrease in precision on scattered values within the input value range 0.02-0.0001 where the previous approximation happened to land exactly on the correct bit pattern by accident, but across the entire range of input values accuracy is much improved (by an order of magnitude at high input values).

Reducing the number of iterations to 42 allows the performance cost of the function ~~to be approximately the same while not noticeably increasing error.~~

EDIT: my performance measurement code was measuring the wrong version of the function. At 42 iterations, the code is 60% slower, but the increase in accuracy is almost worth it.

I am retracting this change as the precision loss without it is entirely manageable.

```suggestion
			x = (a / (x * x) + 2 * x) / 3;
```

src/fixed.h
```diff
@@ -224,13 +224,28 @@ class fixedf {
...
+
+		// We've got the cube root of the normalized parameter, now
+		// un-scale it back to get the real answer
+		if (k > 0) x <<= k;
+		if (k < 0) x >>= -k;
+
 		return x;
```
sturnclaw: This can be made more compact.

```suggestion
		return k > 0 ? x << k : x >> -k;
```

sturnclaw (changes requested):
After having dug into the behavior of the function across a wide range of input values, I'm happy to find the precision of the result is much improved as long as only a scant few fractional bits of precision are discarded.

I've left suggested changes based on my experimental findings - with all changes applied, the algorithm has effectively uniform accuracy across the entire input range in both 32.32 and 16.48 precision modes (until the value becomes too small for the existing code to convert to double-precision floating-point, at which case I cannot compute a reference value).

This typically represents an order of magnitude (or more) increase in accuracy at high input values (1,000+), and at extremely low values the shift trick is successfully gaining significant amounts of precision.

My original request to split the function was based on my intuition that the shift at high values would discard significant amounts of the fractional bits, and I'm happy to retract that feedback now that a way to avoid the loss of precision has been found.

sam52796:
I've made all of @sturnclaw's requested changes, I think. C++ templates and fixed point math are both things that I prefer to think about as little as possible, so the suggestions were mainly copied verbatim. It compiles though! And also seems to work in all the cases I tested.

Note that I moved the PI constant template change into its own commit because it felt like its own thang.

Also note that this pull request also fixes #6302.

src/galaxy/StarSystemGenerator.cpp
```diff
@@ -26,7 +26,6 @@ static const fixed SAFE_DIST_FROM_BINARY = fixed(5, 1);
 // very crudely
 static const fixed AU_SOL_RADIUS = fixed(305, 65536);
 static const fixed AU_EARTH_RADIUS = fixed(3, 65536); // XXX Duplication from StarSystem.cpp
-static const fixed FIXED_PI = fixed(103993, 33102);	  // XXX Duplication from StarSystem.cpp
```
sturnclaw, line 29: Nitpick: if this is duplicated, the file it is duplicated from should be updated to also use the new `fixed::PI` values.

src/galaxy/CustomSystem.cpp
```diff
@@ -1106,7 +1106,30 @@ void CustomSystemsDatabase::RunLuaSystemSanityChecks(CustomSystem *csys)
...
+	// This loop runs after the one above, to ensure that system bodies are all set up first.
+	// Now we ensure that any binary star systems have orbits set correctly.
+	for (CustomSystemBody *body : csys->bodies) {
+
+		if ((body->bodyData.m_type == SystemBody::TYPE_GRAVPOINT) && (body->children.size() == 2)) {
+			SystemBodyData *star_A = &body->children[0]->bodyData;
+			SystemBodyData *star_B = &body->children[1]->bodyData;
+
+			if ((star_A->m_type >= SystemBodyType::TYPE_STAR_MIN) && (star_A->m_type <= SystemBodyType::TYPE_STAR_MAX) &&
+				(star_B->m_type >= SystemBodyType::TYPE_STAR_MIN) && (star_B->m_type <= SystemBodyType::TYPE_STAR_MAX)) {
```
sturnclaw, line 1120: I believe this can be more succinctly expressed as `star_A->GetSuperType() == SystemBodyType::SUPERTYPE_STAR` etc.

In general the fixup-at-load approach is good. Ideally an extension to the editor for creating binary (and trinary) orbits from multiple bodies would be nice, though I suspect that would involve non-trivial engineering to track and express that a set of bodies are intended to be in a binary orbit.

src/fixed.h
```diff
@@ -224,19 +225,37 @@ class fixedf {
 		return (fixedf(root));
 	}
 
-	static fixedf CubeRootOf(const fixedf &a)
+	static fixedf CubeRootOf(const fixedf &a_orig)
 	{
+		fixedf a = a_orig;
+		if (a == fixedf(0, 1)) return fixedf(0, 1);
```
sturnclaw, line 231: Sanity-check: should this be `a <= fixedf(0, 1)` instead? I believe computing the N-th root of a negative number is invalid by definition until you start involving complex numbers...

sturnclaw (approved):
This looks good! I've left a few nitpick/sanity-check comments, but I consider this ready for squash. I'll merge it along with the next batch of save-bump PRs once we enter the full feature-development cycle for the year; I'd like to cut an interim save-compatible release with our current accumulated bugfixes + the OpenAL audio backend sometime around the end of the month.

PR #6331, opened 2026-04-21
Make scanning missions survive hyperspace jumps

### LINKED ISSUE #6304: Scanning mission disappeared after warping out from the orbit
[output.txt](https://github.com/user-attachments/files/25978195/output.txt)

### Observed behaviour
Cartography mission (with orbital scanner) 
Approached target planet, started scanning. Didn't notice that autopilot didn't establish orbit, and I was free falling on a planet, without possibility to recover using main engine, so jumped out to nearest system at the last moment. Refueled and returned, but scanning won't start and pressing scan button at the top left corner of UI does nothing

### Expected behaviour

Expected to keep scanning

### Steps to reproduce

1. Start scanning
2. Jump out of the system
3. Return to scan target and attempt scanning again

20260203; win11

both included

sam52796:
Fixes #6304. 

The scan manager code was not expecting the player to ever jump out mid-scan, so doing so triggered an assert() which then caused the module manager to disable that module. Adding a hook for the _onLeaveSystem_ event, and then clearing any active scan in progress, prevents the assert from firing and thereby saves the ScanManager module from certain death.

### DIFF
--- a/data/modules/Scout/ScanManager.lua
+++ b/data/modules/Scout/ScanManager.lua
@@ -669,4 +669,13 @@ Event.Register("onShipTypeChanged", function(ship)
 	end
 end)
 
+---@param ship Ship
+Event.Register("onLeaveSystem", function(ship)
+	local scanMgr = ship:GetComponent("ScanManager")
+
+	if scanMgr and scanMgr:GetActiveScan() then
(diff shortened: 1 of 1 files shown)


sturnclaw (approved):
Thanks, good catch!

PR #6333, opened 2026-04-23
Hyperdrive UI and sound

sam52796:
There are three commits (so far) in this pull request, all fairly minor tweaks:

First is to shorten the UI and HUD text for hyperspace cloud labels. Since all the clouds already have a "wormhole" icon, it's clear what they are, so instead of labelling them as "Hyperspace departure cloud" it can just say "Departure", and so on. With this change, the system map screen looks like this:

image
.

Second change is to pause all sound when the game is paused. It annoyed me that if you had the game paused and clicked on the hyperspace button, the hyperspace charge-up sound would start playing immediately. Then by the time you unpaused the game to start the countdown going, the sound had already finished. So I started adding flags to track the queuing of the sound while paused. Then I realised that the landing gear up/down sounds have the same issue. Rather than adding a whole complicated system to track which sounds should play at what times, just pausing the sound when the game is paused works flawlessly. It has the bonus of pausing all the ambient starport and ship sounds too, as well as the music if you have that active. I prefer it this way because previously it also annoyed me that if I paused the game to take a phone call, all the sounds would keep playing in the background. 

Third change is a direct result of an issue I personally experienced when trying the game for the first time, whereby I couldn't locate the "add jump" button on the sector map screen. I must have closed the info box, or clicked on the economy info which closes it, because it wasn't there. I found the routing window on the right, which _looks_ like it has buttons to add a jump, but actually doesn't. Rather than move the UI around, I've just added some code to always open the info box, if possible, when the sector map screen is opened. This could have saved me during my initial confusion, since one of the things I tried doing was selecting a mission and clicking on "Set destination as navigation target" and then going to the sector map to look for a "jump to here" button. Long term we might want to consider moving the "HyperJump Routing" section of the info box over to the Route Info box on the right, since that seems like a more logical place for it to live. Happy to do that myself if there is consensus agreement. Then I would remove this info box auto opening code too, since it would no longer be necessary.

### DIFF
--- a/data/lang/core/en.json
+++ b/data/lang/core/en.json
@@ -601,11 +601,11 @@
   },
   "HYPERSPACE_ARRIVAL_CLOUD": {
     "description": "",
-    "message": "Hyperspace arrival cloud"
+    "message": "Arrival"
   },
   "HYPERSPACE_ARRIVAL_CLOUD_REMNANT": {
     "description": "",
-    "message": "Hyperspace arrival cloud remnant"
+    "message": "Arrival (Remnant)"
   },
   "HYPERSPACE_CLOUDS_DISPLAY_MODE": {
     "description": "Label for hyperspace cloud display modes.",
@@ -629,7 +629,7 @@
   },
   "HYPERSPACE_DEPARTURE_CLOUD": {
     "description": "",
-    "message": "Hyperspace departure cloud"
+    "message": "Departure"
   },
   "HYPERSPACE_JUMP_ABORT": {
     "description": "Mouse over text for hyperdrive button on ship control panel",
--- a/data/pigui/views/map-sector-view.lua
+++ b/data/pigui/views/map-sector-view.lua
@@ -829,12 +829,29 @@ local function drawEdgeButtons()
 
 end
 
+local function hasActiveExclusiveLeftModule()
+	for _, module in ipairs(leftSidebar.modules) do
+		if module.exclusive and module.active and not module.closing then
+			return true
+		end
+	end
+	return false
+end
+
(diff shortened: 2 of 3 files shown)


impaktor:
> It has the bonus of pausing all the ambient starport and ship sounds too, as well as the music if you have that active.

I'm not sure pausing music is desired

sam52796:
It's desired by me. 😁

impaktor:
> Second change is to pause all sound when the game is paused

So from IRC:
- Better to have everything silent at pause than all sounds still running
- only silencing F/X and keeping music playing is a bigger coding effort, out side the scope of this PR

I interpret @bszlrd has signed off on this, just waiting if @sturnclaw agrees.

sturnclaw (approved):
This is an acceptable solution for now, given that this is the second PR to elect not to try to change the music API to allow it to continue while paused. I'd like there to be a comment in `Game::EmitPauseState` indicating a desire for music to be handled separately before merge, but it's not a requirement.

nikolas:
This is a great change - I paused the game and was surprised to have e.g. engine sound still playing. I was about to open a ticket on this issue. I have music turned off anyways, so I have no opinion on that one.

sturnclaw:
Thanks! I've pushed a small tweak to prevent re-opening the info view if it was already open.

I'm generally in favor of moving the routing information block over to the right sidebar, as I agree that it is more useful over there. I'd like to still retain some information about the physical layout of the selected system on the left info pane, but I'm not exactly sure how I'd express that yet.

In either case, automatically re-opening the selected system info pane is a good thing to have.

PR #6335, opened 2026-04-27
Fix placeAdvert error in Scoop.lua

nikolas:
This should fix an error I just got which crashed the game:
```
Pioneer Warning
[T] modules/Scoop/Scoop.lua:505: attempt to index local 'p' (a nil value) stack traceback:
[T] modules/Scoop/Scoop.lua:505: in function 'placeAdvert'
[T] modules/Scoop/Scoop.lua:561: in function 'makeAdvert [T] modules/Scoop/Scoop.lua:579: in function 'func'
[T] libs/Event.lua:31: in function 'executor'
[T] libs/Event.lua:79: in function '_Emit'
[T] libs/Event.lua:209: in function <[T] libs/Event.lua:208>
OK
```
Screenshot_2026-04-26_19-44-46

### DIFF
--- a/data/modules/Scoop/Scoop.lua
+++ b/data/modules/Scoop/Scoop.lua
@@ -496,13 +496,14 @@ local placeAdvert = function (station, ad)
 				isEnabled   = isEnabled
 			})
 		else
-			local p = station.path:GetSystemBody().parent.path
+			local p = station:GetSystemBody().parent.body
+
 			ref = station:AddAdvert({
 				title       = l["ADTITLE_" .. ad.id],
 				description = ad.desc,
 				icon        = ad.id == "RESCUE" and "searchrescue" or "haul",
 				due         = ad.due,
-				dist        = p.path == ad.planet and p:GetPhysicalRadius() * 3 or nil,
+				dist        = (p and p.path == ad.planet) and p:GetPhysicalRadius() * 3 or nil,
 				reward      = ad.reward,
 				location    = ad.location,
 				onChat      = onChat,


zonkmachine:
Cool! Looks like the issue may have been introduced in https://github.com/pioneerspacesim/pioneer/pull/6313

impaktor:
Not sure I understand what is triggering the issue? It's trying to place an advert on a station that doesn't exist? Under what condition did the bug trigger?

nikolas:
@impaktor I was flying from one station to another. I was in one of the Gliese systems. I am not familiar with the advert functionality.

nikolas:
For context, this crash happened alongside this error message popping up. May or may not be related, but the info may be helpful: https://bugzilla.redhat.com/show_bug.cgi?id=2463136

sagitter:
Hello.

Can this patch be ported to current stable release 20260203?

sam52796:
As per [this comment](https://github.com/pioneerspacesim/pioneer/pull/6330#pullrequestreview-4157338285) I believe a release with recent bugfixes is planned to be released soon.

sturnclaw:
Hi - when addressing a Lua error, you should make an effort to upload the `output.txt` generated when the game crashes (or if possible a save that can reproduce the crash!). It will have significantly more context about the error that will allow us to diagnose it accurately.

This patch does catch the nil dereference, but does not resolve the semantic issue in the code - the line `local p = ...` is invalid and will not generate correct results for the function, leading to invalid data being presented to the user.

Instead, the line `local p = station.path:GetSystemBody().parent.path` should be replaced with this:

```lua
local p = station:GetSystemBody().parent.body
```

You may then check whether `p` is nil, as in the incredibly rare case of a station that does not orbit a body (only the Wolf 359 system will trigger this case) the parent Body will be nil.

nikolas:
Thank you, and yes I apologize I am unfamiliar with the semantics of this code. If you decide to make your own patch for this that would make sense as well. Sorry I don't have logs or a way to reproduce this - if it happens again I will dig deeper.

sturnclaw (approved):
> Thank you, and yes I apologize I am unfamiliar with the semantics of this code. If you decide to make your own patch for this that would make sense as well. Sorry I don't have logs or a way to reproduce this - if it happens again I will dig deeper.

There is no need to apologize! I appreciate the initiative in submitting the original patch, and I certainly don't expect you (as a new contributor) to know the context involved in this particular bit of code that would be required to catch the semantic error.

The new patch looks good, thanks for your time and effort on the matter. Don't worry about the log in this case if you're unable to reproduce it reliably; in this case the error location is enough to diagnose the obvious issue and fix it.

PR #6337, opened 2026-04-27
Fix typo in MISSILE_NAVAL_FLAVOURTEXT: personel -> personnel

nikolas:


impaktor:
FYI , I believe this will trigger a re-translation of this string in all languages (unless transifex auto-ai-bot has gained super powers lately)

nikolas:
Ah okay, yes I was unclear if changes to the source language are made in transifex or here in git. I suppose there's no way for transifex to differentiate whether a source language change is a typo fix or an actual change that requires re-translation.

sturnclaw:
Transifex provides a per-language option to accept the previous translation when a string is updated, so it's not a big burden for translators to verify the translated version had correct spelling/semantics and re-approve the previously-translated version.

zonkmachine:
In the Transifex web editor (and through the api), you can edit the source language to fix up typos like this and then save the changes with the **Save & Keep Translations** option. This would have been perfect in this case.
[edit-source-strings-online](https://help.transifex.com/en/articles/6236820-edit-source-strings-online#h_7e60a9b609)

sturnclaw:
> In the Transifex web editor (and through the api), you can edit the source language to fix up typos like this and then save the changes with the **Save & Keep Translations** option. This would have been perfect in this case. [edit-source-strings-online](https://help.transifex.com/en/articles/6236820-edit-source-strings-online#h_7e60a9b609)

Thanks! I don't recall this being in Transifex's docs the last time I looked at it. This still has a general problem (as far as I can tell) of a lack of oversight from project maintainers that Github PRs provide - i.e. any random Transifex user could modify source strings without approval - and the History tab in Transifex allows reinstating the previous translation when source strings are changed in the repo.

I don't believe our Transifex integration to sync translations supports treating Transifex as the canonical source for the `en` resource, though my memory could be faulty.

impaktor:
> I don't believe our Transifex integration to sync translations supports treating Transifex as the canonical source for the en resource, though my memory could be faulty

Indeed, the transifex sync script (that runs 3 AM every morning) doesn't pull the `en` files from transifex, only pushes. This is (a copy of) the script that runs on the Pioneer box, for the inquiring mind:
https://github.com/pioneerspacesim/pioneer/blob/master/scripts/update-translations.sh

PR #6338, opened 2026-04-28
Fix typo in DebugShip.lua: Foward -> Forward

nikolas:


impaktor:
You could just push all fixes to the same branch, if that would make things easier.

PR #6340, opened 2026-05-03
Add custom "Load cargo" button to cargo pickup missions

### LINKED ISSUE #6168: Picking up cargo for missions requires re-docking if cargo bay is full of trade goods which are sold upon arrival
### Observed behaviour
If a cargo pickup and return mission is currently running and the ship's cargo bay is already full, then docking at the pickup location results in a "come back later" prompt. This is too literal because once the cargo bay has been emptied at the station, the only way to get the cargo pickup is to literally come back later via undocking and re-docking at the station.

### Expected behaviour
I would expect the mission tab to contain a button for picking up the cargo after the ship once again has enough space for cargo. This is important because in the early game, profits can be maximized by carrying trade goods to the pickup location and then selling them before picking up the mission cargo. Being forced to re-dock does not make sense in-universe and is annoying for the player.

### Steps to reproduce

1. Take a cargo pickup mission which requires picking up cargo at a different station before carrying it to another station.
2. Fill your ship's cargo bay with goods.
3. Dock at the pickup location.

sam52796:
Fixes #6168.

I added some code so that missions can now add custom action buttons, which appear in the mission screen. 

Then I added a "Load cargo" button to cargo-run pickup missions, so that if you land with a full cargo hold and sell your cargo you don't have to take off and land again in order to trigger the loading of the mission cargo - you can just click the "Load cargo" button in the mission screen. See example below:

image

### DIFF
--- a/data/lang/module-cargorun/en.json
+++ b/data/lang/module-cargorun/en.json
@@ -403,6 +403,10 @@
     "description": "",
     "message": "I'm taking damage. I need help."
   },
+  "LOAD_CARGO": {
+    "description": "UI button label",
+    "message": "Load cargo"
+  },
   "MACHINE_TOOLS": {
     "description": "custom cargo",
     "message": "machine tools"
@@ -661,7 +665,7 @@
   },
   "WE_HAVE_LOADED_UP_THE_CARGO_ON_YOUR_SHIP": {
     "description": "",
-    "message": "We have loaded up the cargo on your ship."
+    "message": "We have loaded the cargo onto your ship."
   },
   "WHAT_IS_THIS": {
     "description": "",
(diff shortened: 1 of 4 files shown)


impaktor:
It still auto-loads the cargo otherwise, as the seasoned player expects?

sam52796:
Correct, no existing functionality regarding cargo missions was changed.

sturnclaw:
Nitpick: the message `YOU_DO_NOT_HAVE_ENOUGH_EMPTY_CARGO_SPACE` should be changed to e.g. "Please contact me when you have sufficient cargo space.", hinting to the player that they need to use the mission dialog rather than undocking and redocking.

In general this is a very lovely change, thank you! We have been intending to split cargo loading out into an over-time process (per multiple design discussions on the subject), and this is a good step towards that reality.

sam52796:
Apologies for the delay. I've now updated that string.

sturnclaw (approved):

PR #6341, opened 2026-05-04
Make giant stars circular in system overview

sam52796:
When I first saw this, I thought it was a bug:

image
<br />
The star image appears to be offset from the circular mask that contains it, creating a strange "half moon" effect on the star. 

When I investigated, it turns out that the icons for giant stars are not the whole star - they are just a corner of it:

image
<br />
This could be a cool effect if the star image was always at the edge of the screen, to give the impression that the star is so huge it won't even fit in the view. And maybe this was the case back when these icons were last updated 8 years ago. But currently the system overview lets you zoom in and out, and you can have binary star systems too:
<br /><br />
image
<br />
The result is, I'm afraid, that this concept doesn't work and looks confusing. 

If I were more graphically inclined, I would submit a new set of giant star icons that were the whole star, but I am not, so instead I've modified the code to use the "regular" star icons in place of giant ones. The above system views now look like this:
<br />
image

image
<br />
The code to switch on all the different star types is still there, commented out, so if in future someone else creates a set of suitable giant star icons then it would be super easy to incorporate them, but in the meantime I think this is an improvement.

### DIFF
--- a/src/galaxy/SystemBody.cpp
+++ b/src/galaxy/SystemBody.cpp
@@ -761,13 +761,38 @@ const char *SystemBody::GetIcon() const
 	switch (m_type) {
 	case TYPE_BROWN_DWARF: return "icons/object_brown_dwarf.png";
 	case TYPE_WHITE_DWARF: return "icons/object_white_dwarf.png";
+	case TYPE_STAR_M_GIANT:
+	case TYPE_STAR_M_SUPER_GIANT:
+	case TYPE_STAR_M_HYPER_GIANT:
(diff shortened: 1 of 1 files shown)


sturnclaw:
@bszlrd thoughts? I'm okay with this as an interim fix...

bszlrd:
Works for me, absolutelly

impaktor:
> This could be a cool effect if the star image was always at the edge of the screen, to give the impression that the star is so huge it won't even fit in the view. And maybe this was the case back when these icons were last updated 8 years ago.

It was.

PR #6342, opened 2026-05-04
Make thruster visual more consistent

### LINKED ISSUE #6245: Thrusters - Uneven visual effect on thruster pairs
Image

Image

### Observed behaviour
The visual thruster 'flames' are uneven on some thruster pairs. This is most visible when comparing the forward thrusters of the Coronatrix on a landing/descent burn. Incidentally it looks like this could be an issue with only the forward/right thruster(s). It looks like it may be missing one of the planes in rendering. _Edit: Something similar seem to be going on with the Bowfin but it's on both sides._

### Expected behaviour
Visually even output from thruster pairs.

### Steps to reproduce
* Start at mars in a Coronatrix
* Go to Bradbury Landing
* When the landing burn starts, hit pause
* With the external camera, from the front, zoom around the forward thruster images from different angles and distances and against different backgrounds (land, night sky,.. etc.).

On Pioneer master (9de1f0a)
Ubuntu Mate 24.04.03

Coronatrix landing burn:
[output.txt](https://github.com/user-attachments/files/23421930/output.txt)
[opengl.txt](https://github.com/user-attachments/files/23421933/opengl.txt)
Save. Landing at Olympus Mons:[olympus.txt](https://github.com/user-attachments/files/23421946/olympus.txt)

sam52796:
Fixes #6245.

The math to calculate thruster flame intensity extracted a vector from the transformation matrix to represent its direction. This worked most of the time, except that certain thruster orientations could come out with different results. This was most noticeable on pairs of thrusters created by mirroring or rotating an existing thruster to a duplicate on the opposite side of the ship.

I've changed the code to use the actual thruster-to-camera vector instead of matrix-axis extraction. This eliminates any rotation-dependent brightness mismatch on mirrored thrusters, and the Coronatrix forward thrusters now look the same for me.

I also spent 5 minutes staring at the opening scene of the game that shows all the ships one by one rotating with all thrusters firing, and they all now look correct too.

### DIFF
--- a/src/ModelBody.cpp
+++ b/src/ModelBody.cpp
@@ -5,6 +5,7 @@
 
 #include "Camera.h"
 #include "Frame.h"
+#include "Game.h"
 #include "GameSaveError.h"
 #include "JsonUtils.h"
 #include "Pi.h"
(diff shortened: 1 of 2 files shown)


sam52796:
I've added a second commit which pauses the thruster flame flicker when the game is paused. In current master branch the thruster flicker is independent of game time, which looks a bit strange when all other movement is paused.

zonkmachine:
> I've added a second commit which pauses the thruster flame flicker when the game is paused.

Yes, this is a better behavior.

I've tested the Coronatrix (both game and model viewer) as per https://github.com/pioneerspacesim/pioneer/issues/6245 and the issue is fixed in this PR. Thanks!

sturnclaw (approved):
Good work, thanks!

PR #6345, opened 2026-05-10
[WIP] Autopilot v2

Mc-Pain:
This is an attempt to make a new generation autopilot with following fixes:
- If gravity is stronger than THRUSTER_UP, ship will fall down after "taking off" phase (accelerate upwards and keep 1km/s speed)
- If orbiting something heavy, autopilot will try to match target velocity, which leads to another instance of falling ship

Features planned:
- "Enter and maintain orbit around" command, where "orbiting around" may be performed in zero-gravity using only thrusters
- Intercepting body detection: just orbit around it instead of trying to evade
- Improved impact recovery - it's like driving a car, sometimes flying around is better then trying to stop at all costs and fail.

https://github.com/pioneerspacesim/pioneer/pull/6296 is included intentionally, new autopilot relies on it.

For testing purposes, new autopilot is applied to player only.
NPC's are not affected

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -41,6 +41,7 @@ env:
     libsdl2-image-dev
     libopenal-dev
     ninja-build
+    libfreetype-dev
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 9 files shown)


sturnclaw:
Marking as draft to match WIP state; please feel free to mark ready for review once you're satisfied with the state of the PR.

PR #6346, opened 2026-05-10
Thruster exhaust plumes

sam52796:
My previous post here may have ruffled some feathers, so here is a peace offering: Thruster exhaust plumes.

Hopefully this doesn't contradict any established lore to do with the way ship thrusters work.

The plumes are only visible within planet atmospheres, and the spread depends on the atmospheric density. But they kick up dust on the planet's surface whether there is an atmosphere or not. Matching the dust colour to the planet's actual render colour currently isn't really possible, so it kind of guesses. Also, if you're very near a landing pad which is up in the air, you can sometimes see the plumes pass through it a short distance, which isn't ideal, but that wouldn't be very easy to fix.

Let me know what you think.

### DIFF
--- a/src/Sfx.cpp
+++ b/src/Sfx.cpp
@@ -12,6 +12,7 @@
 #include "JsonUtils.h"
 #include "ModelBody.h"
 #include "Pi.h"
+#include "Space.h"
 #include "matrix4x4.h"
 
 #include "core/IniConfig.h"
@@ -26,9 +27,30 @@
 
 #include "profiler/Profiler.h"
 
+#include <algorithm>
+#include <vector>
+
 using namespace Graphics;
 
 namespace {
+	constexpr uint64_t EXHAUST_EMITTER_PTR_SORT_TAG = UINT64_C(0x8000000000000000);
+
+	inline uint64_t ExhaustStreamEmitterSortKey(const Sfx &s)
+	{
+		if (s.m_exhaustEmitter)
+			return EXHAUST_EMITTER_PTR_SORT_TAG | static_cast<uint64_t>(reinterpret_cast<uintptr_t>(s.m_exhaustEmitter));
+		return static_cast<uint64_t>(s.m_exhaustSavedEmitterBodyIdx);
+	}
+
+	inline bool SameExhaustJetStream(const Sfx &a, const Sfx &b)
+	{
+		if (a.m_exhaustJetIndex != b.m_exhaustJetIndex)
+			return false;
+		if (a.m_exhaustEmitter || b.m_exhaustEmitter)
+			return a.m_exhaustEmitter == b.m_exhaustEmitter;
+		return a.m_exhaustSavedEmitterBodyIdx == b.m_exhaustSavedEmitterBodyIdx;
+	}
+
 	float SizeToPixels(int screenHeight, const vector3f &trans, const float size)
 	{
 		//some hand-tweaked scaling, to make the lights seem larger from distance (final size is in pixels)
@@ -49,7 +71,7 @@ namespace {
 		case TYPE_SMOKE:
 			return Clamp(SizeToPixels(screenHeight, pos, (inst.m_speed * inst.m_age)), 0.1f, 50.0f);
 		case TYPE_EXHAUST:
-			// Calculated by the particle code
+			return 0.f; // billboards use explicit vertex quads in RenderAll
 		default:
 			return 0.f;
 		}
@@ -78,7 +100,12 @@ Sfx::Sfx(const vector3d &pos, const vector3d &vel, const float speed, const SFX_
 	m_dragScale(1.0f),
 	m_opacityScale(1.0f),
 	m_windVel(vector3d::Zero),
-	m_type(type)
+	m_type(type),
+	m_exhaustEmitter(nullptr),
+	m_exhaustSavedEmitterBodyIdx(Sfx::INVALID_EXHAUST_SAVED_BODY_IDX),
+	m_exhaustJetIndex(0),
+	m_exhaustBirthSeq(0),
+	m_exhaustSuppressStreakElongation(false)
 {
 	if (type == TYPE_EXHAUST)
 		m_lifetime = SfxParams::EXHAUST_LIFETIME * float(Pi::rng.Double(0.6, 1.0));
@@ -103,12 +130,17 @@ Sfx::Sfx(const Json &jsonObj)
 		m_opacityScale = sfxObj.value("opacityScale", 1.0f);
 		m_windVel = sfxObj.value("windVel", vector3d::Zero);
 		m_type = sfxObj["type"];
+		m_exhaustEmitter = nullptr;
+		m_exhaustSavedEmitterBodyIdx = sfxObj.value("exhaustEmitterBodyIdx", Sfx::INVALID_EXHAUST_SAVED_BODY_IDX);
+		m_exhaustJetIndex = Uint16(sfxObj.value("exhaustJetIndex", 0));
+		m_exhaustBirthSeq = sfxObj.value("exhaustBirthSeq", Uint32(0));
+		m_exhaustSuppressStreakElongation = sfxObj.value("exhaustSuppressStreakElongation", false);
 	} catch (Json::type_error &) {
 		throw SavedGameCorruptException();
 	}
 }
 
-void Sfx::SaveToJson(Json &jsonObj)
+void Sfx::SaveToJson(Json &jsonObj, const Space *space)
 {
 	Json sfxObj({}); // Create JSON object to contain sfx data.
 
@@ -124,6 +156,13 @@ void Sfx::SaveToJson(Json &jsonObj)
 		sfxObj["dragScale"] = m_dragScale;
 		sfxObj["opacityScale"] = m_opacityScale;
 		sfxObj["windVel"] = m_windVel;
+		Uint32 bodyIdxForSave = m_exhaustSavedEmitterBodyIdx;
+		if (m_exhaustEmitter && space && space->IsBodyIndexValid())
+			bodyIdxForSave = space->GetIndexForBody(m_exhaustEmitter);
+		sfxObj["exhaustEmitterBodyIdx"] = bodyIdxForSave;
+		sfxObj["exhaustJetIndex"] = m_exhaustJetIndex;
+		sfxObj["exhaustBirthSeq"] = m_exhaustBirthSeq;
+		sfxObj["exhaustSuppressStreakElongation"] = m_exhaustSuppressStreakElongation;
 	}
 	sfxObj["age"] = m_age;
 	sfxObj["seed"] = m_seed;
@@ -186,14 +225,15 @@ float Sfx::AgeBlend() const
 	return 0.0f;
 }
 
-SfxManager::SfxManager()
+SfxManager::SfxManager() :
+	m_nextExhaustBirthSeq(1)
 {
 	for (size_t t = 0; t < TYPE_NONE; t++) {
 		m_instances[t].clear();
 	}
 }
 
-void SfxManager::ToJson(Json &jsonObj, const FrameId fId)
+void SfxManager::ToJson(Json &jsonObj, const FrameId fId, const Space *space)
 {
 	Frame *f = Frame::GetFrame(fId);
 	Json sfxArray = Json::array(); // Create JSON array to contain sfx data.
@@ -204,7 +244,7 @@ void SfxManager::ToJson(Json &jsonObj, const FrameId fId)
 				Sfx &inst(f->m_sfx->GetInstanceByIndex(SFX_TYPE(t), i));
 				if (inst.m_type != TYPE_NONE) {
 					Json sfxArrayEl({}); // Create JSON object to contain sfx element.
-					inst.SaveToJson(sfxArrayEl);
+					inst.SaveToJson(sfxArrayEl, space);
 					sfxArray.push_back(sfxArrayEl); // Append sfx object to array.
 				}
 			}
@@ -273,7 +313,7 @@ void SfxManager::AddThrustSmoke(const Body *b, const float speed, const vector3d
 	sfxman->AddInstance(sfx);
 }
 
-void SfxManager::AddExhaust(const Body *b, const vector3d &startPos, const vector3d &backboneVel, const vector3d &plumeOffset, const vector3d &plumeOffsetVel, const float intensity, const float dragScale, const float opacityScale, const vector3d &windVel)
+void SfxManager::AddExhaust(const Body *b, const Uint16 exhaustJetIndex, const bool exhaustSuppressStreakElongation, const vector3d &startPos, const vector3d &backboneVel, const vector3d &plumeOffset, const vector3d &plumeOffsetVel, const float intensity, const float dragScale, const float opacityScale, const vector3d &windVel)
 {
 	SfxManager *sfxman = AllocSfxInFrame(b->GetFrame());
 	if (!sfxman) return;
@@ -289,6 +329,11 @@ void SfxManager::AddExhaust(const Body *b, const vector3d &startPos, const vecto
 	sfx.m_opacityScale = opacityScale;
 	sfx.m_windVel = windVel;
 	sfx.m_backboneAtStepStart = sfx.m_backbonePos;
+	sfx.m_exhaustJetIndex = exhaustJetIndex;
+	sfx.m_exhaustEmitter = b;
+	sfx.m_exhaustSavedEmitterBodyIdx = Sfx::INVALID_EXHAUST_SAVED_BODY_IDX;
+	sfx.m_exhaustSuppressStreakElongation = exhaustSuppressStreakElongation;
+	sfx.m_exhaustBirthSeq = sfxman->m_nextExhaustBirthSeq++;
 	sfxman->AddInstance(sfx);
 }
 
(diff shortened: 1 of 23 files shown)


bszlrd:
Sounds good! Could you show some screengrabs, perhaps even a gif? For posterity. I'll check it out as soon as I can.
Edit, got home, so here are some shots:
screenshot-20260511-195419
screenshot-20260511-195607
screenshot-20260511-195610
screenshot-20260511-200333
screenshot-20260511-200427
screenshot-20260511-200441
screenshot-20260511-200503

zonkmachine:
plume1

bszlrd:
I think it is quite nice. A bit too excessive for the RCS thrusters though, borderline hindering visibility.
Maybe make it configurable via the .model files?

impaktor:
Cool.
I imagine the _how much_ can be tweaked post merge, but good to get the basic amount right first.

sam52796:
I've made some further tweaks.  The exhaust and dust now scales with the actual thruster power, as well as the size of the thrusters on the 3D model itself. However, all of this hasn't made much difference to the "forward-facing exhaust obscures your view" problem, because most ships have quite powerful forward-facing thrusters. So much so that the standard planet landing procedure involves a powered descent balanced on the nose of the ship facing downwards.

If we're moving to a more realistic gravity well and TWR simulation then this problem could end up fixing itself, but in the meantime I guess... don't pick fights in atmospheres...

sturnclaw:
This looks very, very nice! I'll give this branch a thorough testing as soon as I'm able; I quite appreciate your initiative in adding QoL and player-experience features.

Off the top of my head, fading particle opacity based on distance to camera and potentially the dot product between the camera vector and the particle stream velocity may help to solve the vision-obscuring-cloud problem; close particles should fade out regardless to avoid unsightly popping issues with the near clip plane.

@bszlrd would you have time in your schedule to find/whip up an appropriately licensed (set of) dust particle alpha textures? At current I believe this effect is simply using a shader-based-fade ellipsoid particle with radial symmetry; the visual fidelity may be able to be improved even further with an appearance of non-uniform density.

bszlrd:
A bit of handwaving might be good for the non-main directions. Maybe more transparent?
To me, the transparency and its seeming strength at about 20-25% thrust seems about the sweet spot. More subtle, but still noticeable
screenshot-20260512-193221
screenshot-20260512-193234
screenshot-20260512-193346

Or maybe about 15%, if we check it in Earth's atmo:
screenshot-20260512-193531
screenshot-20260512-193546


~~And playing with it for a while, I think it might work better, if gravity and drag won't affect it, or at least much less. Especially gravity. Not sure if my minds eye models it correctly, but I think it would benefit it if gravity would not pull it down. If we are having airplane analogies, contrails aren't dropping either for example, at least not so soon.~~

sam52796:
I've added some procedural noise to the exhaust fragment shader. It's fairly large blobs - if the noise has smaller blobs then the exhaust looks too grainy. For this reason I think that actual textures would be overdoing it.

There aren't any cropping issues with the Z plane because the particles are "billboard" style so always face the camera - they're either drawn or not, and by the time you get close enough for them to not be drawn, you're in a cloud of exhaust or dust anyway and there are plenty of other particles around to make up for it.

If the first person view is really an issue, the simplest way would be to not draw the exhaust when in that view. I certainly think that the streaks coming from the corners of the screen look weird, though this is partly because the camera isn't where the pilot actually sits on the model. If the streaks were coming from below you, as they look like they should according to the thruster positions relative to the cockpit, then it might be fine. But adjusting the camera position according to the 3D model's cockpit seems out of scope for this change.

bszlrd:
The camera  should be in the correct position for all ships that have custom cockpits. The plumes did not seem out of place for the coro. They do obscure, especially as they dissipate, but at about 10℅ thrust it did not felt that excessive I think. Or maybe 5℅?

Not drawing them when you are in internal view could work. Although it enhances the feeling of presence for the down view a lot, I think. Perhaps a heavy handed muting for the other directions when in internal views?

sam52796:
Turns out the diagonal exhaust lines seen when in the first person view were due to a bug. I've now fixed that, and the thruster exhaust appears to come from where you'd expect, given the cockpit position.

zonkmachine:
Here is the launch in slow motion:
https://drive.google.com/file/d/1eWC3U8qjWiBbh21paH3CViTy5nZW1yKK/view?usp=sharing

The exhaust has a pretty cool anime kind of look at 1/10th speed.

sam52796:
Sometimes it could look like the wrong thrusters were firing. This wasn't because of any issue with the exhaust or thruster flame code - it's correctly reflecting the directions that the autopilot wants to push the ship. But the autopilot calculates everything at once, which sometimes looks wrong to a human being.

For example, if the ship is flying nose up away from the planet's surface, and you decide you want to travel parallel to the planet's surface instead, in your mind you will first rotate the nose downwards 90 degrees, and then thrust forward. But the autopilot will try to thrust in the new direction immediately. Since the ship is facing up, that means all the thrusters on the topside of the ship will fire, to start pushing it parallel to the surface, even while some other thrusters are working to rotate the ship. This isn't what you expect, even though it's perfectly valid.

In order to address this, I've added another commit which makes autopilot rotational burns block each other as well as blocking linear thrust burns. The ship will now first try to pitch towards the direction it wants to go, then yaw, then roll, then linear thrust. This has the nice effect when taking off under autopilot that the belly thrusters cut out, the RCTs fire to rotate the ship, then the main thrusters at the back engage to push up at full blast.

Since autopilot code is not really in scope for this change, and is a crucial area of the game, I've tried to be maximally conservative in how it is applied, specifically:

- The rotational blocking only occurs for "major turns". If the correction angle is less than 8 degrees then rotation doesn't block anything.
- The rotational blocking is entirely disabled when time acceleration is active.

Hopefully this means that in almost all situations the original autopilot behaviour is preserved, and so there won't be any issues introduced. Even so it could probably do with a bit of testing. I have checked taking off and landing on different stations on Mars, as well as heading back and forth to Mars High, but haven't tried it anywhere else.

impaktor:
So ready for review?

sam52796:
Yes, if everyone is happy with the behaviour of the plumes, then it's ready for review.

bszlrd:
Thinking about #6352 I came back to the thought that this should be data driven via the ship.json.
Maybe just a multiplier per direction or an amount and an alpha multiplier for more control, but a way to control it should be present.
Edit: with good defaults of course. Say 1x for main, 0.5x for retro and up, and 0.25x for other thrusters.

impaktor:
> I came back to the thought that this should be data driven via the ship.json

Could be a second PR, to not delay merging this too much, unless @sam52796 feels motivated to include it in this PR

sturnclaw (changes requested):
Image

The part circled in green is great. I love the aero cloud effect, even if it is a bit bright and overstated in Mars' thin atmosphere. The parts circled in red are less great.

## Visuals

The initial jet stream coming from the thrusters:
1. Doesn't visually match the thruster flame shape, and feels very disconnected as it only starts being visible as an extremely narrow "needle tip" tens of meters away from the ship.
2. Suffers significant frequency aliasing due to having UV-space noise on an extremely-squashed billboard mesh.
3. Would be better replaced with a separate particle effect (or procedural mesh effect) showing the thruster exhaust expansion in atmosphere that links the particle cloud to the actual thruster itself.

I believe the implementation here appears to be attempting to visually model the contrail effect, with a side of dust and particulate kicked up by the thruster wash? Pioneer's engines don't (traditionally) leave a chemical rocket exhaust trail.

If that is the case, I'd point to some reference images:

contrail-1
contrail-2

That second image in particular shows a bit of what I'm pointing towards as a "connecting link" between the visible white contrail and the thruster plume itself, with an atmospheric "sheath" of hot/refractive/slightly tinted air surrounding the thruster plume and joining the contrail cloud.

What I'm noticing as consistent among atmospheric contrails is that they become visible with a significantly wider initial shape than the current implementation, but have a smooth opacity fade from transparent to opaque during the first 10-20% of the trail.

## Performance

Now we come to the other parts circled in red. This implementation is *significantly* performance heavy, with a common scene (accelerating in forwards-cruise flight above Cydonia) consuming 10ms of CPU time in a debug build.

"That's a red herring," you may say, "that's debug build timing!" While true, we maintain a policy of keeping the debug build as optimized and usable as possible. Our players may spend most of their time using a release build, but we as developers spend most of our time in a debug build. The worse debug performance gets, the less useful it is, (and the more time wasted when trying to track down bugs) until eventually the debug build is so slow it is completely unusable.

In a release build, the following scene takes 1.1ms to perform thruster cloud simulation and GPU submission, more than twice the time needed to evaluate and submit the entire planet and city to the GPU for rendering.

Image

I've not slapped a profiler on the GPU side of things, but I doubt the smoke particles themselves are going to wind up a particularly major concern during rendering; the particle shader is fairly simple and should be ROP-unit -bound for transparent blending on most modern GPUs.

I hope the numbers above have already made this statement rhetorical: the performance budget of this PR is not acceptable for merge.

There is likely significant opportunity for optimization within the implementation; while I don't have the free time at the time of writing to dig deeply into the PR and point out areas of improvement, I'd encourage you to familiarize yourself with the [profiling tools](https://dev.pioneerspacesim.net/contribute/profiling) we support.

I will leave you with a few areas of potential optimization to think about:
- Many of our ships have large quantities of co-located thruster nodes that fire in the same direction or as a group. The contrail/dust-cloud visual effect does not require differentiating between individual thrusters in that group; it may help to reduce per-frame CPU time to support artist-placed "thruster trail" nodes that are associated at model import time with individual thruster nodes.

  I can provide some guidance with adding additional node types and handling the import step in `scenegraph/Loader.cpp`.
- It may (or may not) be faster to do a single collision line trace below the ship against the physics representation rather than attempt an analytic solution to the plume-pad vector in `TryGetLandingPadDustOverride`, especially for ground stations with many pads. (It also might produce more consistent visual results...) Profile it first to see if the function is contributing significantly to frame-time.

  Kudos for not doing a per-thruster-per-frame line trace out-of-the-box, I appreciate the amount of optimization that has already gone into the implementation!
- It may be significantly faster to move individual-particle mesh transforms into the vertex shader (passing the required parameters via a uniform block or per-instance data), as most of the required information should be already present.

  Constructing each individual particle vertex on the CPU is somewhat wasteful if the same effect can be built from GPU-transformed instanced quads with <10MB/s of GPU upload bandwidth. At current, 10MB/s bandwith would allow you to upload 240 instances of the entire Sfx struct to the GPU per frame at 165 FPS, (which is a very achievable performance target on semi-modern GPUs.)
- The `Sfx` struct is quite fat, clocking in at 264 bytes per particle. Long-term, it would be good to split data into per-emitter vs. per-particle information, but that's a significant refactor/implementation effort that may not be appropriate for this PR. For now, I'd recommend you identify areas of the struct that are duplicated across multiple particles coming from the same logical "source" and mark them as such with a comment for future work.

I appreciate the amount of work that has gone into this PR. I'd love to have it merged as an on-by-default feature, but if the performance issues cannot be addressed then it will need a default-off feature flag in the settings.

sam52796:
This is great feedback, thanks.

I've made some performance improvements in a new commit. I figure there is no point going into the cosmetic changes if the PR would never be accepted anyway. So let's get it to an acceptable performance level first. To address your specific points:

- I'm generally reluctant to make model artists add information that the code ought to be able to figure out itself. In theory, the grouping and intensity of thruster exhaust should be calculable from the positions and sizes of the existing thruster nodes. In practice, it is a bit of a pain to do that, but the code already approximates the same effect because each ship has an exhaust particle budget which gets divided among the nodes and scaled by thruster flame size. So when a Coronatrix takes off, there are 12 small streams each with 1/12 the number of particles shooting out from underneath it. Since the particles expand over time, they end up merging into basically the same exhaust cloud as you would get if there was a single big belly thruster.
- I profiled TryGetLandingPadDustOverride() and it was a tiny fraction of the render time (0%) so I don't think there is much of a performance gain to be had there. However, I agree that doing a raycast is better in principle, so I think that this would be a good improvement at some point in the future.
- Moving some of the work to the GPU did improve performance by about 40%. The code now sends one instance record per particle instead of six billboard vertices, and the normalization and cross-product math is done in a shader instead of in SfxManager::RenderAll().
- Agreed about the Sfx struct being too large. I've partitioned the fields that are per particle vs per jet/emitter, so that a future PR could do something intelligent about it, but since I don't think it would make a huge performance difference I have left it as is for now.

The other main change I made was removing the jet stream particle sort. Since each particle in a stream is elongated towards the previous one, they need to be drawn in sequence. I was previously sorting the stream using a multiple-comparison function (emitter + jet id + sequence id) but I realised that this is completely unnecessary. The particles are already in the correct order! They are created in sequence as the ship flies, and might be removed semi-randomly but removal doesn't change the order of surviving particles. The only complication is that the per-jet particles are interleaved. But all I have to do to solve that is walk the list separating the particles into buckets per-jet, and the end result is what I want.  This also means I can remove the m_exhaustBirthSeq field from the Sfx struct, since it was only added to enable the sorting.

Removing the sort improved performance by another 40% or so, and SfxManager::RenderAll() now takes significantly less time than it used to.

I've done a commit of these changes, and I would be grateful if you could give the code another test, to see if the performance is now at an acceptable level.

data/shaders/opengl/exhaust_instanced.vert
```diff
@@ -0,0 +1,58 @@
...
+	float lenSq = dot(v, v);
+	if (lenSq < 1e-18)
+		return vec3(1.0, 0.0, 0.0);
+	return v * inversesqrt(lenSq);
+}
+
+void main(void)
+{
+	float ux = mix(-1.0, 1.0, a_uv0.x);
+	float vy = mix(-1.0, 1.0, a_uv0.y);
```
sturnclaw: Note that there is no need to pass either position or UV values to the shader; you can derive the 0..1 UV coordinate from bits 1 and 2 of `gl_vertexID`. This creates a CCW-wound 2-tri quad starting in the lower-left corner.

src/Sfx.h
```diff
@@ -35,13 +60,55 @@ struct Sfx {
...
 
 	void TimeStepUpdate(const float timeStep);
-	void SaveToJson(Json &jsonObj);
+	void SaveToJson(Json &jsonObj, const Space *space);
 
+	// --- Shared by all SFX types ---
+	// For TYPE_EXHAUST m_pos and m_vel aren't directly used, but are updated to show the particle's
+	// current pos + vel so that any secondary particle analysis keeps working.
 	vector3d m_pos;
 	vector3d m_vel;
```
sturnclaw, line 68: These fields are unused outside of particle-type-specific codepaths. It is safe to use `m_pos` and `m_vel` as the storage for backbone position and velocity.

src/Sfx.h
```diff
@@ -35,13 +60,55 @@ struct Sfx {
...
+	// --- TYPE_EXHAUST stream kinematics (per particle) ---
+	// Thruster exhaust jets have a "backbone" which is the centre of the stream. This is used
+	// to track how stretched out each particle should be, and in which direction.
+	vector3d m_backbonePos;
+	vector3d m_backboneVel;
+	vector3d m_backboneAtStepStart;
+	// Each particle then has a perpendicular offset from the backbone, which grows over time
+	// and gives the exhaust plume its conical shape
+	vector3d m_plumeOffset;
+	vector3d m_plumeOffsetVel;
```
sturnclaw: Note that as the direction of the particle ideal "backbone" velocity rotates, the offset is no longer perpendicular. The current effect could likely be achieved by combining `m_plumeOffset` into `m_backbonePos` and tracking the particle velocity separately from the offset velocity.

Alternatively, I'd strongly suggest storing `m_plumeOffset` and its velocity value as a single-precision vector and compositing it with the camera-space particle position vector very late in the pipeline using a single-precision matrix3x3 transform to rotate its orientation.

src/Sfx.h
```diff
@@ -35,13 +60,55 @@ struct Sfx {
...
+	// current pos + vel so that any secondary particle analysis keeps working.
 	vector3d m_pos;
 	vector3d m_vel;
 	float m_age;
-	float m_speed;
+	float m_speed; // For TYPE_EXHAUST this is used for thruster intensity at spawn, which is per jet - could be moved to an auxilliary table
 	enum SFX_TYPE m_type;
+
+	float m_seed;
+	float m_lifetime;
```
sturnclaw, line 76: There are 4 bytes of "free" padding after `m_lifetime` before the next value in this structure; `m_exhaustSuppressStreakElongation` and `m_exhaustDustKick` should be moved into this space.

src/Sfx.h
```diff
@@ -35,13 +60,55 @@ struct Sfx {
...
+
+	// Emitter identity (for stream grouping)
+	const Body *m_exhaustEmitter;
+	Uint32 m_exhaustSavedEmitterBodyIdx;
+
+	Uint16 m_exhaustJetIndex; // The jet index to the emitter that created this particle
+
+	float m_opacityScale; // How visible the exhaust should be - thinner atmosphere and weaker thrust levels reduce this
+	double m_exhaustGroundRadius; // The height of the ground below this exhaust, calculated from the ground below the ship at spawn time
+	Color m_exhaustDustTint; // Dust tint and mix amount, calculated from the ground below the ship at spawn time.
```
sturnclaw, line 100: `m_exhaustDustTint` should be moved into the free padding bytes before `m_exhaustGroundRadius`.

src/Sfx.cpp
```diff
@@ -281,32 +544,166 @@ void SfxManager::RenderAll(Renderer *renderer, FrameId fId, FrameId camFrameId)
...
+						});
+					}
+				}
+
+				if (!exhaustInstances.empty()) {
+					const uint32_t numDrawn = uint32_t(exhaustInstances.size());
+					EnsureExhaustInstanceBuffer(renderer, numDrawn);
+
+					VertexBuffer *instanceVB = s_exhaustInstanceVB.Get();
+					ExhaustInstanceData *instanceData = instanceVB->Map<ExhaustInstanceData>(BUFFER_MAP_WRITE);
```
sturnclaw: Note that as this is per-frame transient data, you should make use of `Renderer::CreateTempVertexBuffer` (available in more recent versions of the master branch) rather than attempting to manage your own instance buffer. Among other benefits, this batches vertex/instance data upload into a single transaction, reducing the amount of OpenGL driver overhead incurred by this function.

sturnclaw (changes requested):
> I've done a commit of these changes, and I would be grateful if you could give the code another test, to see if the performance is now at an acceptable level.

Total overhead for update+render is ~6ms in debug mode (down from ~11ms for just rendering!), which is still not great but within acceptable tolerance. CPU render cost in a release build is about 0.5ms on a Ryzen 7 5800X, which is vastly improved from before. I've left several comments indicating where you can eek out some more performance; I'd prefer not to leave any performance on the table especially with such a high cost (~40% of the frame) in debug mode.

With these changes addressed, I would consider the PR within acceptable performance limits to be a merge candidate.

That still leaves the visual issues; in addition to the issues I raised in the previous review, the visual appearance of exhaust trails from lateral/maneuvering engines is significantly straining to suspension of disbelief; particle velocity and lifetime is uniform regardless of thruster power, and the lazily-floating-particle effect undercuts the fact that the ship is travelling at supersonic or near-supersonic speeds.

sam52796:
I've made all those changes, and the end result is that the Sfx struct should now be 168 bytes, according to an analysis tool that I used (I didn't count the packing by hand). I don't know if it's made any difference to performance in general, but it can't hurt.

Visually I've tried to tweak it to look more like a contrail, and made the jets more forceful overall, and made the non-main thrusters fainter in comparison. Not sure if that addresses all of your concerns, but any more drastic changes would require adding a new effect, plus possibly modifying the current flame effect to end up with a blending between the two, so I think this is as far as we can go with this iteration.

In terms of realism, I think it may be better not to dwell too much on what it ought to look like. As far as I know, the operation of Pioneer's thrusters is not really defined anywhere. I'd say that this is one of those things that benefits from being left as fairly vague. We know that hydrogen goes in, and "thrust" comes out. If this ends up looking like a "flame" followed by a "contrail" then rule of cool says that's fine. It also leaves open the possibility for someone to create better-looking exhaust plumes in future.

src/Ship.cpp
```diff
@@ -1535,14 +1537,14 @@ void Ship::SpawnThrusterExhaustParticles(float timeStep)
 	vector3d windDir = vector3d(0.0, 1.0, 0.0).Cross(localUp);
 	if (windDir.LengthSqr() < 1e-8) windDir = vector3d(1.0, 0.0, 0.0).Cross(localUp);
 	windDir = windDir.NormalizedSafe();
-	const vector3d windVel = windDir * double(SfxParams::EXHAUST_WIND_SPEED) * Clamp(density, 0.0, 1.0);
+	const vector3f windVel = vector3f(windDir) * (SfxParams::EXHAUST_WIND_SPEED * float(Clamp(density, 0.0, 1.0)));
```
sturnclaw: Note: single-precision float has enough effective precision to store a normalized direction vector. `localUp` should be converted to `vector3f` and these computations done using single-precision float ops.

src/Ship.cpp
```diff
@@ -1624,11 +1629,11 @@ void Ship::SpawnThrusterExhaustParticles(float timeStep)
 				const vector3d vAxis = exhaustDirWorld.Cross(uAxis).Normalized();
 
 				const vector3d jitterDirW = (uAxis * std::cos(jitterTheta) + vAxis * std::sin(jitterTheta)).Normalized();
-				const vector3d plumeOffset = jitterDirW * (jitterRadius * th->GetVisualSize() * SfxParams::EXHAUST_INITIAL_SPREAD);
+				const vector3f plumeOffset = vector3f(jitterDirW * jitterRadius * SfxParams::EXHAUST_INITIAL_SPREAD);
 
 				const double jitterVel = (SfxParams::EXHAUST_MAX_SPREAD * jitterRadius) / SfxParams::EXHAUST_LIFETIME;
 				const double spreadAmp = Clamp(1.0 / std::max(density, 1e-5), 0.05, 8.0);
-				const vector3d plumeOffsetVel = jitterDirW * (jitterVel * spreadAmp);
+				const vector3f plumeOffsetVel = vector3f(jitterDirW * (jitterVel * spreadAmp));
```
sturnclaw: Similarly here, if `uAxis`, `vAxis`, `jitterRadius`, and `jitterTheta` are only used to produce single-precision float values, they can be immediately converted to single-precision and all following calculations done in single-precision.

sturnclaw (commented):
There is a little bit of CPU performance left-over here. When handling unit-vector directions that are eventually being stored in a single-precision float target, single-precision float has perfectly acceptable precision for the whole computation chain. SSE2+ has equivalent throughput for non-packed single and double-precision ops, but this reduces the stack and register footprint by a modest amount and allows the compiler's autovectorization pass a little bit more flexibility.

I'll give the improved visuals a try soon and provide final feedback on that topic in a separate set of comments.

sam52796:
Made those changes too.

sturnclaw:
I've playtested the branch under several conditions.

- At night, the exhaust plume is unrealistically bright, especially for a contrail-style effect. Raising `exhaustIllumMul` to the power of 1.5 causes the exhaust contrail to match low ambient lighting at night while still looking similarly bright in direct light.
- Thrusters on different sides of the ship are still not visually differentiated from each other particularly well. For example, the upwards thrust on the Coronatrix does not appear as though each individual thruster is putting out a total of 27% the thrust of a singular rear thruster. A lower log scale (tested as far down as 1; not pictured) does produce a slightly better effect and I'd recommend a change be made there.
  
image

- Exhaust velocity should scale slightly with thruster power. A multiplier somewhere in the range `(0.6+0.4 * finalThrusterPower)` produced interesting results, and I suspect slightly reducing the effect of thruster scale on drag would naturally compliment the change.

Other than the faint feeling of physically-incorrect behavior of the thruster plumes, I do quite like the effect in total and I think with some tweaking it will be a very nice net positive for Pioneer.

sam52796:
I've made the exhaust darker at night.

I've made the manouvering thruster exhaust appear fainter in comparison to the main thruster exhaust.

I have not made the exhaust velocity scale with thruster power. When I was initially imagining what the thruster exhaust might look like, I reasoned that a futuristic spaceship thruster would always spew exhaust as fast as possible, attaining throttling by reducing the volume of output rather than the speed, in order to conserve propellant. So the smaller thruster exhaust would initially be just as fast as the big thruster exhaust, there would just be less of it. Hence I made the smaller thrusters fainter and with fewer particles, but otherwise the same as the big ones.

Turns out this was a very convenient decision for the modelling.

When you said that you found it too suspension-of-disbelief-y I tried applying a velocity scale factor. However this produced occasional streaks in the exhaust when the autopilot was used. After much head scratching I eventually figured out that while player input to thrusters is either on or off, the autopilot input can spend a frame at an intermediate level, if it calculates that a course adjustment needs a fraction of a frame's worth of thrust. With a scalable exhaust velocity this then leaves one particle travelling much slower, creating a discontinuity in the jet stream, and the elongation code makes the following particle elongate inappropriately, resulting in seemingly random streaks.

Rather than try to catch this edge case and deal with it properly somehow, I went back to the constant exhaust speed and instead increased the drag for smaller thrusters - I figure this is vaguely similar to what would happen in real life, since a smaller stream would lose momentum faster than a larger one, regardless of starting speed.

In summary - if you really think it will make the difference to whether this is merged or not, then I can go back and look at the scalable velocity again, but I personally I don't think the visual difference would be all that noticeable, and the likelihood of streaking appearing again under some circumstance that I didn't think of is high, so I don't think it's worth it.

bszlrd:
How about merging it as it is now? 
If we want to do an august release, then it will certainly add to the coolness factor, and we'll get eyes, playtime and feedback on it from players. And then it can be adjusted based on that and our more extensive testing. 
I'd still prefer being able to control the plume parameters from the ship.json per direction eventually, but that can be a next step. (In general, these things would be better off being data-driven instead of hard-coded for more flexibility, and for modders to play with)

sturnclaw:
> I have not made the exhaust velocity scale with thruster power.

> When you said that you found it too suspension-of-disbelief-y I tried applying a velocity scale factor. However this produced occasional streaks in the exhaust when the autopilot was used. After much head scratching I eventually figured out that while player input to thrusters is either on or off, the autopilot input can spend a frame at an intermediate level, if it calculates that a course adjustment needs a fraction of a frame's worth of thrust.

This one's on me - I was insufficiently specific in my original comment. I was referring to scaling exhaust particle velocity by the ratio between the thruster's computed maximum-throttle power and the maximum-throttle power of the most powerful thruster on the craft. This is meant as an intermediate approach to scaling particle exhaust velocity via the thruster's effective exhaust velocity value (eEV); we currently do not differentiate betwen maneuvering/primary thruster eEV, but in both fiction and design intent this is something we want to model. Hence my suggestion of the smaller thrusters having a visually less-energetic particle effect.

sturnclaw:
> How about merging it as it is now?

I'm in favor of this. I still have reservations about the appearance of the thrust particles, especially in e.g. a dogfight or when flying in cruise mode, but I'd be happy to get this out to users as part of a Latesommar release and see if my concerns are shared by our userbase.

sturnclaw (approved):
> I've made the exhaust darker at night.

> I've made the manouvering thruster exhaust appear fainter in comparison to the main thruster exhaust.

These both look much improved, thanks for addressing them! I'm going to go ahead and call this PR good as-is and merge it.

I'd like to CC @Mc-Pain and @Gliese852 as the resident autopilot experts to double-check if any changes need to be made to the autopilot post-merge to accommodate the rotational decoupling part of this PR. From my look at the code, the autopilot changes appear sufficiently limited in scope to not impact NPC flight consistency etc. and are not a blocker for merge.

Mc-Pain:
> > I've made the exhaust darker at night.
> 
> > I've made the manouvering thruster exhaust appear fainter in comparison to the main thruster exhaust.
> 
> These both look much improved, thanks for addressing them! I'm going to go ahead and call this PR good as-is and merge it.
> 
> I'd like to CC @Mc-Pain and @Gliese852 as the resident autopilot experts to double-check if any changes need to be made to the autopilot post-merge to accommodate the rotational decoupling part of this PR. From my look at the code, the autopilot changes appear sufficiently limited in scope to not impact NPC flight consistency etc. and are not a blocker for merge.

Has not read this thoroughly. Why progressive and rotational motions are blocking each other?

sam52796:
When the autopilot is making big orientation adjustments, it looks cooler if the thrusters adjust each axis one after the other, instead of all at the same time.

This behaviour should only be triggered when the total desired adjustment angle is more than 8 degrees, so minor in-flight heading adjustments on approach are not affected.

PR #6348, opened 2026-05-14
Render instanced models using per-frame scratch buffer

### LINKED ISSUE #5807: CTD when looking at settlements on a moon
### Observed behaviour
Crashes to desktop on a moon with 3 settlements close by after loading a save. Prior to that, getting there and navigating around all of the settlements to dock multiple times was crash-free.

### Steps to reproduce
Attached is a save. Reliably reproduced by: 

1. flying with nose straight up from docking bay
2. switching to back camera when at about 300~400m/s

This happens when just flying around with the settlements in sight as well. Occurs with "planet detail distance" and "cities detail level" from Very Low to Very High.

ver 20240314 (8b6ae92)
OS: Windows 11 Home

[output.txt](https://github.com/pioneerspacesim/pioneer/files/14748523/output.txt)
[opengl.txt](https://github.com/pioneerspacesim/pioneer/files/14748571/opengl.txt)
[save.zip](https://github.com/pioneerspacesim/pioneer/files/14748555/save.zip)

### LINKED ISSUE #6171: City buildings on small asteroids rendered incorrectly at a distance
### Observed behaviour
On a small asteroid, the buildings of a distant city are rendered incorrectly when the angle between the camera and the plane of the city is too large, i.e. when the origin point of the camera is not directly above the city. This can be seen for Cooper Station on Nofri Colony in the Piizess (-6, -4, 2) system:

Incorrect rendering when camera is to the side:
Image

Correct rendering when camera is above:
Image

### Expected behaviour
The buildings should always be rendered conformal to the surface.

### Steps to reproduce

1. Find a very small lumpy low-gravity body with a city surface port.
2. Approach close to the body or land on it 90° away from the city along the equator.
3. Activate the sidereal view, zoom out and look at the city.

__My output.txt and game save__
[output.txt](https://github.com/user-attachments/files/20747002/output.txt)

[opengl.txt](https://github.com/user-attachments/files/20747003/opengl.txt)

[Broken Asteroid City.txt](https://github.com/user-attachments/files/20747004/Broken.Asteroid.City.txt)

### LINKED ISSUE #6326: City rendering on densely citied small planets shows buildings in the sky
### Observed behavior
It looks like the city building rendering has a bug on small planets with many cities.

When you fly in close they properly snap to the ground but when in space near by 20-km ish they are rendering on a plane (see pic)

### Steps to reproduce

go to Deanus (7, 3, 1) 

Klein's Rock 

check out the city rendering

Image

Image

as i zoom in it seems the building get moved to the ground. 

Image

debian 11 

https://github.com/pioneerspacesim/pioneer/commit/6114ccb82ea87ba73cde2817b0523b344b328603#diff-45521a7c36fc1cc7d274cedbca45979c0162f9e55ca8b8e6954e59c8fef693e1L754

maybe Im stupid and this is drawing the unrendered buffer... but i think its just drawing the border around the city

... i recompiled without this and still showing the same 

Image

other then that this branch is up to date with master but has some added stars for save file compatibility

this is the exact current version. 

https://github.com/syonfox/pioneer/commit/d5525c5caade02668e67f7e670afa62e2d8bf4e0

(separate issue maybe this 	//SDL_setenv("SDL_VIDEODRIVER", "windows,wayland,x11", 0); ) is breaking some peoples builds are we sure it handles all cases/enviroments as best possible.

[output.txt](https://github.com/user-attachments/files/26781682/output.txt)

<rename zip>
[muiti_city_render.txt](https://github.com/user-attachments/files/26781715/muiti_city_render.txt)

[mods.7z.txt](https://github.com/user-attachments/files/26782845/mods.7z.txt)

may be related to https://github.com/pioneerspacesim/pioneer/issues/6171 but i dont know.

sturnclaw:
Fixes #6326, fixes #6171 (same root cause), fixes #5807.

This PR addresses an issue in the SceneGraph rendering code which was improperly re-using an instance buffer when drawing the same model multiple times per frame. The fix wound up being ~200LOC to stop using the instance buffer entirely; since we're overwriting it each frame, it's more efficient to just scratch-allocate the instance data into a single large temporary buffer that is uploaded to the GPU once rather than 100s of times per frame.

@fluffyfreak It is possible this PR will show a slight improvement in rendering efficiency on Raspberry Pi, as there is now less graphics API overhead. I think the Pi is still fill-rate bound however, so I don't expect much improvement.

In the process, the improved `Draw()` API superseded the old `DrawBufferDynamic` API, allowing for a significant reduction of code in the renderer.

In closing, have this screenshot proving city rendering still works...

Screenshot From 2026-05-14 16-25-08


src/graphics/opengl/RendererGL.cpp
```diff
@@ -42,6 +42,9 @@
 
 using RenderPassCmd = Graphics::OGL::CommandList::RenderPassCmd;
 
+// 1MB vertex draw buffer should be enough for most cases, right?
```
fluffyfreak, line 45: 640k ought to be enough for anyone ;)

fluffyfreak (approved):
This looks good, I've given it a test in a few situations and not run into any issues. As always the code looks good.

PR #6351, opened 2026-06-04
Advice fixup

zonkmachine:
Rename 'Billie bounce' to 'bounce in'. With no lore to back it up it becomes a bit too enigmatic and hard to translate.

Following a brief [discussion here](https://github.com/pioneerspacesim/pioneer/pull/6234#issuecomment-4537220919), I suggest changing the title of one of the advices added in #6234.

### DIFF
--- a/data/lang/module-advice/en.json
+++ b/data/lang/module-advice/en.json
@@ -21,7 +21,7 @@
   },
   "ADVICE_12_HEADLINE": {
     "description": "",
-    "message": "Bounce in"
+    "message": "Bounce out and back in"
   },
   "ADVICE_13_BODYTEXT": {
     "description": "",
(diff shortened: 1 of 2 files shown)


zonkmachine:
@mwerle @impaktor

data/lang/module-advice/en.json
```diff
@@ -21,7 +21,7 @@
   },
   "ADVICE_12_HEADLINE": {
     "description": "",
-    "message": "The Billie Bounce detour"
+    "message": "Bounce in"
```
sturnclaw:
```suggestion
    "message": "Bounce out and back in"
```

IMO this flows a little better, and encapsulates what's actually being advised.

zonkmachine: That works better. Or maybe "Jump out and back in". Too dry? I still like bounce but maybe it's not clear enough?

sturnclaw: I like the term "bounce", and I think it's fine enough for a non-serious advice like this.

sturnclaw (approved):

PR #6352, opened 2026-06-04
Dynamic thrust colors

Mc-Pain:
Make thrust color non-static. Only white/blue at full, smolder to yellow/red when barely fired.

Temperature clamp'ed between 0 and 30000K

image

image

image

### DIFF
--- a/src/scenegraph/Thruster.cpp
+++ b/src/scenegraph/Thruster.cpp
@@ -140,6 +140,11 @@ namespace SceneGraph {
 		// emissive.r is the thruster power setting which effects flame length and brightness
 		m_tMat->emissive.r = m_glowMat->emissive.r = 255.0f * displayedPower;
 
+		// emulate thruster temperature
+		double temperature = MathUtil::Lerp(0.0, 30000.0, power);
+		vector3d thrustColor = PhysicsUtil::GenerateTemperatureColor(temperature);
+		SetColor(Color(thrustColor.x, thrustColor.y, thrustColor.z, 255));
+
 		m_tMat->diffuse = m_glowMat->diffuse = currentColor;
 
 		//directional fade
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -408,7 +408,7 @@ add_subdirectory(src/editor)
 
 target_link_libraries(${PROJECT_NAME} LINK_PRIVATE ${pioneerLibs} ${winLibs})
 target_link_libraries(unittest LINK_PRIVATE ${pioneerLibs} ${winLibs})
-target_link_libraries(modelcompiler LINK_PRIVATE ${pioneerLibs} ${winLibs})
+target_link_libraries(modelcompiler LINK_PRIVATE pioneer-lib ${pioneerLibs} ${winLibs})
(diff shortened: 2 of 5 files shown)


impaktor:
What determines the thruster color? Engine model? Engine health? Ship class?

Mc-Pain:
> What determines the thruster color? Engine model? Engine health? Ship class?

Throttle amount

src/scenegraph/Thruster.cpp
```diff
@@ -140,16 +140,18 @@ namespace SceneGraph {
 		// emissive.r is the thruster power setting which effects flame length and brightness
 		m_tMat->emissive.r = m_glowMat->emissive.r = 255.0f * displayedPower;
 
+		// emulate thruster temperature
+		double temperature = MathUtil::Lerp(0.0, 30000.0, power);
+		vector3d thrustColor = PhysicsUtil::GenerateTemperatureColor(temperature);
```
sturnclaw, line 145: Once again, this is an excessive consumption of CPU cycles for something that should be a 30 or 60-entry precomputed lookup table. Feel free to store an interpolation curve with each entry in the LUT for maximum accuracy, but computing 48 wavelength bands *per thruster, every frame* is simply wasteful.

sturnclaw (changes requested):
General thoughts about the visual effect: deep red is generally immersion-breaking. Maneuvering thrusters are typically playing on the visual identity of a plasma jet or cold gas thruster (white/colorless spray), so a color temperature of 3000-5000K at the low end.

In general, varying the color of the entire thruster flame rather than having it fall off at the edges of the thrust plume is going to look slightly unrealistic, and regardless of the "real" Planck temperature of the thruster, the center will be an out-of-range white when viewed through a camera or human eye anyway.

I've left a comment below about the technical implementation that will need to be resolved before this is considered for merge.

impaktor:
@bszlrd along which dimension do you think thruster color is best variable? Throttle amount? Engine type? Engine class, ship model? Engine health?

I'm thinking throttle amount is already keyed to sound and size of flame?

bszlrd:
I'm not sure it is good to vary the colors on the fly. Thust power ( (thrust*EV)/2) could be a good candindate. Or EV directly.
Later on engine wear and tear could govern it.
But I think for now hand-setting their color would be better. Which did not work as it was documented in #3948 the last time I tried.

Anyhow, I think such things should be data driven. As in defined in the ship.json, say as a color for 0% thrust and 100%.

impaktor:
My main thought is: a player shouldn't see all the thruster colors from the start, but rather, the color should signal something (engine getting old/damaged or changed ship or changed engine or now on military fuel, etc). But that can be worked out later I guess.

zonkmachine:
> , but rather, the color should signal something (engine getting old/damaged or changed ship or changed engine or now on military fuel, etc).

The flight control could bump the player about a possible issue with their engine. On arrival or take off.

Mc-Pain:
> > , but rather, the color should signal something (engine getting old/damaged or changed ship or changed engine or now on military fuel, etc).
> 
> The flight control could bump the player about a possible issue with their engine. On arrival or take off.

I'd suggest changing thruster plumes quantity if engine is not working efficiently, similar to car's exhaust:
white - it's cold
bold white and sweet - coolant is leaking into combustion chamber
blue-ish - oil is leaking
black - incomplete combustion, typical for diesel engines

PR #6353, opened 2026-06-04
Avast! Be ye knowin' of the Smuggler's Hold?

sturnclaw:
This PR adds the long-awaited Smuggler's Hold equipment, known in-game as a "Concealed Cargo Rack". Available only in systems with a lawlessness > 0.5 (halfway through "Dangerous" or worse security rating), this equipment provides total immunity from police scans for the cargo stored inside.

It is designed as a habitat-slot equipment to both limit the amount of illegal cargo that can be smuggled without risk from a Doylist perspective, as well as to respect the fact that most ships' equipment bays are smaller than the dimensions of a 1u cargo container on at least one axis from a Watsonian perspective.

To complement this, police cargo scans have been partially rewritten. They now have a percentage chance to find your illegal goods, allowing you to smuggle small quantities of illegal items in a hold mostly filled with legitimate cargo. The detection chance starts at 45% for one item in a sea of cargo, and tops out at 100% when 55% of your cargo is illegal. Typically you will want to smuggle in the 5-10% range, though if you get caught...

...the police will no longer execute you on the spot! They'll instead assess you with a fine roughly equal to the intrinsic value of your illegal goods, and only proceed to immediate deresolution if that fine exceeds the usual threshold for immediate hostilities.

Note that smuggler's holds will *not* conceal their cargo from the police if you do get caught, whether that's by a random scan of the cargo in your hold or by buying or selling from a police honeypot at a station.

This PR also makes some minor QoL changes to the equipment stat comparison view, as well as changing the "Cabins" slot category to the more generic "Habitat". The intent is to expand this a bit to allow ships to have more modular habitat space, i.e. with crew amenities as equipment modules. Perhaps you'd like to change extra crew bunks out for a lavish zero-G kitchen for those longer travels out in the black?

I've not yet made more major changes to equipment slot names or types, seeking to save that for a later effort where we intentionally break save compatibility. Ideally that would go hand-in-hand with the crew happiness PR, as well as a "habitat space" metric that plays into crew dissatisfaction over longer trips.

### As usual, playtesting is extremely helpful!

Please make sure to test an existing save from before this PR to double-check that it is fully save-compatible.

### DIFF
--- a/data/lang/equipment-core/en.json
+++ b/data/lang/equipment-core/en.json
@@ -63,6 +63,10 @@
     "description": "Category header for pressurized cabin slots",
     "message": "Cabins"
   },
+  "HABITAT": {
+    "description": "Category header for pressurized habitat slots",
+    "message": "Habitat"
+  },
   "CARGO_LIFE_SUPPORT": {
     "description": "",
     "message": "Cargo Bay Life Support"
--- a/data/lang/equipment-core/en.json
+++ b/data/lang/equipment-core/en.json
@@ -938,5 +938,37 @@
   "WEAPONS": {
     "description": "Category name of weapon-related equipment",
     "message": "Weapons"
+  },
+  "AUXILIARY_CARGO_RACK": {
+    "description": "Equipment name for a cargo-carrying rack installed as a cabin module",
+    "message": "Auxiliary Cargo Rack"
+  },
+  "AUXILIARY_CARGO_RACK_DESCRIPTION": {
+    "description": "Equipment description for a cargo-carrying rack installed as a cabin module",
+    "message": "Additional cargo space mounted inside the ship's pressurized habitat."
+  },
+  "AUXILIARY_CARGO_RACK_FLAVOURTEXT": {
+    "description": "Equipment description for a cargo-carrying rack installed as a cabin module",
+    "message": "Put your unneeded habitat space to work for you! This conversion kit fits a wide variety of ship hulls. Manufactured by Seversk a.k."
+  },
+  "CONCEALED_CARGO_RACK": {
+    "description": "Equipment name for a cargo-carrying smuggler's hold installed as a cabin module",
+    "message": "Concealed Cargo Rack"
+  },
+  "CONCEALED_CARGO_RACK_DESCRIPTION": {
+    "description": "Equipment description for a cargo-carrying smuggler's hold installed as a cabin module",
+    "message": "Shielded cargo space disguised as an innocuous habitat wall. Hides the contents from police scans."
+  },
+  "CONCEALED_CARGO_RACK_FLAVOURTEXT": {
+    "description": "Equipment description for a cargo-carrying smuggler's hold installed as a cabin module",
+    "message": "This cargo rack from Seversk a.k. has been crudely concealed by welding interior hull panels to it. A handwritten sticker bears the text \"Guaranteed to pass cursory inspection, or your money back.\""
+  },
+  "CARGO_CAPACITY": {
+    "description": "Statistic name for an equipment that provides cargo capacity",
+    "message": "Additional Cargo Capacity"
+  },
+  "SHIELDED_CARGO_CAPACITY": {
+    "description": "Statistic name for an equipment that provides cargo scan resistance",
+    "message": "Shielded Cargo Capacity"
   }
 }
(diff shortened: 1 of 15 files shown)


sturnclaw:
CC @impaktor / @zonkmachine - might be a good idea to get an advice string mentioning that you can buy concealed cargo racks in more lawless systems?

impaktor:
> without risk from a Doylist perspective

[Watsonian vs Doylist](https://tvtropes.org/pmwiki/pmwiki.php/Main/WatsonianVersusDoylist) (for others needing to look it up)

> They'll instead assess you with a fine roughly equal to the intrinsic value of your illegal goods, and only proceed to immediate deresolution if that fine exceeds the usual threshold for immediate hostilities.

I assume the fine is some integer multiplier of the intrinsic value, that is scaled relative to the detection rate? Btw, it makes me think of the devious danes, who charged all ships passing through the strait based on the stated value of their cargo, with the option to buy it at the stated price if it seemed too low.

> this equipment provides total immunity from police scans for the cargo stored inside.

OK, makes sense, as long as player can't fill their entire ship with it, but as you mentioned on IRC, there are physical limitations to how many of these the player can fit to their ship

> Note that smuggler's holds will not conceal their cargo from the police if you do get caught, whether that's by a random scan of the cargo in your hold

By "smuggler's hold" you mean the cargo outside the concealed cargo rack? Or I'm confused as to how the two quoted statements are simultaneously true? Would this be if the player has illegal goods outside the concealed cargo rack, so they get caught, then the gestapo turns the inside of the ship upside-down and finds the concealed cargo rack?

impaktor:
Also, this makes me think of #6151 which classifies a system as a "pirate system" (with various pirate flavour stuff") if `lawlessness >= 1`. I'm thinking they should use the same threshold, and be rare / only custom systems so we can place them in neat "piraty places" (asteroid belt, or some exotic star system).

> might be a good idea to get an advice string mentioning that you can buy concealed cargo racks in more lawless systems?

Indeed. Possibly modify existing advice, and/or add new where needed, for informing:
1. sweet spot for smuggling 5%-10% to avoid detection _if_ police scan your ship
2. inform player of possibility to hide cargo / smuggler's hold in lawless systems (does not have to mention the actual name "concealed cargo rack"(TM), the advice guy is not a salesman for "Concealed Cargo Racks R Us").

Note, I have new advice flavour in the pirate PR #6151

sturnclaw:
> I assume the fine is some integer multiplier of the intrinsic value, that is scaled relative to the detection rate?

The integer is 1. It's not scaled to the detection rate, but merely a fine approximately equal (some modification by system lawlessness) to the value of the cargo transported. I consider it a half-way measure to the police boarding your ship and seizing the illegal cargo which is the expected course of action.

> Btw, it makes me think of the devious danes, who charged all ships passing through the strait based on the stated value of their cargo, with the option to buy it at the stated price if it seemed too low.

This is a good idea for a customs/border-patrol module, where you can negotiate with officials to import your goods into high-security or isolationist systems. Decidedly out of scope for this PR.

> By "smuggler's hold" you mean the cargo outside the concealed cargo rack? Or I'm confused as to how the two quoted statements are simultaneously true? Would this be if the player has illegal goods outside the concealed cargo rack, so they get caught, then the gestapo turns the inside of the ship upside-down and finds the concealed cargo rack?

If you are carrying illegal cargo that's not covered by a smuggler's hold and the police find it, they will fine you for the total value of all illegal cargo on the ship regardless of whether it's in a smuggler's hold or not. The logic becomes needlessly complex otherwise, and it has the Watsonian justification that the police board your ship / perform more invasive scans after finding one illegal container and find the rest.

sturnclaw:
Seeing no feedback, enacting the usual "playtest in master" protocol.

zonkmachine:
@sturnclaw 

I get a crash. Messages are from ImGui.

```gdb
	[T] modules/Equipment/Stats.lua:61: bad argument #2 to 'format' (not a number in proper range)
```

To replicate:
- New game (mars)
- Comms -> Equipment Market
- Press to select the pulse cannon -> Crash!

I believe this was introduced in cc4e64be0264251f21bd02fd0de2718427a4dcae (bisect)

sturnclaw:
@zonkmachine to be That Guy, that's technically a Lua exception, not a crash. Output.txt is strongly recommended in this case, as it has the smoking gun (in this case `v == 0` so we're attempting to format the result of divide-by-zero), but I've reproduced it already and will push the fix to master.

EDIT: thanks for letting me know, I'm glad this one didn't sneak through!

PR #6354, opened 2026-06-05
reticule: draw scale gauges for current speed and distance

Mc-Pain:
Trying to get a better reticle according to my local notes

image


Mc-Pain:
current progress:
image

sturnclaw:
If this is information already present on the frame info (as it appears in the screenshot), it might be better to render a small arrow on the inside of the scales, and allow the user's brain to intuitively understand their speed rather than have a direct readout.

bszlrd:
I'll post my related mockups and notes I posted on IRC here as well, so they are at hand.

The landing aid could be easily adapted to it, I think:
landing-aid-polish-009-largereticule

My mockup:
hud-large-reticule

I think it still needs some design thinking and polish, both from a functional and aesthetics point of view.
Based on the last screenshot, it would add a bit too much clutter, especially with all those large running numbers on both sides. And on a large monitor, the numbers are rather large and don't look good at with those proportions:
screenshot-20260703-142348

And them numbers sticking to the extremes for a while is also not a good way to do it in my opinion.

Admittedly my mockups on the other hand make that aspect too mute, I think we would need to find a good middle ground where it is informative and useful, but not overloaded. Maybe make the ladder numbers smaller, and ommitt and only show the unit at the actual readout?

I think our current reticule is really nice and elegant both from aesthetic and usablity aspects. Especially for transfers, where you'd want all the relevant information close to the center, lest you miss your flipower, or half-deltaV point. On the other hand, there were multiple requests for a cleaner HUD, where there are not so many elements obscuring things. And this HUD could be useful for that.
Also with a pitch ladder when flying close to a planet, this HUD could make more sense for atmo fligth, maybe even for orbital flight.

I think it would be nice, if there were multiple HUD modes, including the current reticule, this, and a minimal one (see bellow), and a button and shortcut to switch between them. And that hotkey should be made rebindable. 
Also in general,  the _hide HUD_ and _hide label_ functionality should be separated, and given rebindable hotkeys as well, and the same for cycling between frame, nav and combat target ideally.
image

Also, I'm not entirely sure, how the brake gauge works. At first glance it seems like the lower halfway point would be the flipover point, which is nice, but then both the retro and main thruster brake point would benefit from a tick, so you not only see that point after you went past it. 
And to me, it seems, the overshoot part is not proportional to the acceleration half, based on the central reticule readouts. The braking distance is almost the double of the current distance on this screenshot, but it does not look like that. The red part would need to be about double that size I think.
screenshot-20260703-142720

Importantly, this way it actually looses some precisity compared to the original, because the acceleration part is about half the size of the original. From that standpoint, I think it would be good if the gauge would work the same as before, growing the red from the bottom when overshooting.
image

I do like that the brake gauge also double as a climb/descent/closure rate indicator. 



And if I understand correctly, the deltaV gauge does not take your current speed into account at all. Which then makes it making sense only for a trip started with a full tank. Like in this example bellow, you wouldn't have a visual indication of the half-point of your available deltaV if you wanted fly to some other place from that, only if you memorize your current deltaV and watch that go to it's half. That's what the climbing gauge part on the original is doing. Showing your current speed relative to your deltaV.
On the other hand, the larger arc gives more precisity to the gauge, which is rather useful specifically for those low-tank diversions.
screenshot-20260703-143821


And I think the brake gauge and the deltaV gauge as well would benefit from a bit of separation from the reticule line to make it more readable. Possible even a bit of thickness, but I'm not sure about that.

bszlrd:
Posting here where thinking about the hull etc gauges took me:
large-reticule-ideation-001
I've ommitted the numerical readouts, because they clutter up the bottom a bit, and not elegant at all. See example bellow, where I also indicate, how the current speed component of the deltaV gauge could look (dark orange overlay on the deltaV gauge):
large-reticule-ideation-002
I've also moved the compass up a bit, and made it similar to the pitch ladder, and adding a chevron for current heading. Also the cardinals show their letters.
Additionally, I've moved the readout of the two main gauges to the outside, where they will obscure the running numbers when they are bellow (removing a bit of redundancy. And freeing up the inside for the gauges.
But I'm not really convinved about those gauges living there.

Weapon gauges could go onto the top half of the reticule, but that won't work with missiles.

I've also got the idea of a more compact thruster feedback, instead of the relatively large box we currently have. And a circular acceleration gauge next to it:
large-reticule-ideation-003

It does remove the little current vector line though. But on the other hand, it is rather more compact.
And the acceleration gauge could double as an acceleration limiter if clicked (or mousewheeled)

sturnclaw:
> I've also got the idea of a more compact thruster feedback, instead of the relatively large box we currently have. And a circular acceleration gauge next to it.

Dibs on implementing this. I'll do it in a separate branch.

> Posting here where thinking about the hull etc gauges took me

This is a *really* nice design. I still love the flight reticle, but this feels extremely polished with the little circle widgets at the bottom of the gauges. I'm very tempted to develop both as manufacturer or ship-specific designs (swappable via equipment) to provide some differentiation between ship manufacturers in the way dV Rings of Saturn does.

sturnclaw:
...while I'm in the neighborhood, would you like text shadows?

bszlrd:
Text shadows (or HUD shadows in general) would be useful. Might need some finessing to look good though

Mc-Pain:
Current progress:
image

PR #6355, opened 2026-06-06
Fix Kanara translation tokens

zonkmachine:
`Warning: Missing translation token 'KANARA_POLICE' in resource 'ships/en.json'.`

All ships with a police model follow the paradigm name/name_police apart from the Kanara which use kanara_civ/kanara where just 'kanara' is the police model. This caused me some confusion in #6103 where ships where made translatable. The translation tokens in the json files do follow the same pattern as the rest of the civ/police ships. I think just correcting the tokens in /data/lang/ships/en.json is enough for now but perhaps someone should look into the kanara json files too.

data/ships/kanara.json
```json
	"model": "kanara",
	"i18n_key": "KANARA_POLICE",
	"name": "Kanara (Police)",

```

data/ships/kanara_civ.json
```json
	"model": "kanara_civ",
	"i18n_key": "KANARA",
	"name": "Kanara (Civilian)",

```

### DIFF
--- a/data/lang/ships/en.json
+++ b/data/lang/ships/en.json
@@ -84,28 +84,28 @@
     "message": "a DS Miner"
   },
   "KANARA": {
-    "description": "ship name - base form",
+    "description": "ship name - base form, Kanara civilian",
     "message": "Kanara"
   },
-  "KANARA_CIV": {
-    "description": "ship name - base form",
-    "message": "Kanara Civ"
-  },
-  "KANARA_CIV_DEF": {
-    "description": "ship - definite form",
-    "message": "the Kanara Civ"
-  },
-  "KANARA_CIV_INDEF": {
-    "description": "ship name - indefinite form",
-    "message": "a Kanara Civ"
-  },
   "KANARA_DEF": {
-    "description": "ship - definite form",
+    "description": "ship - definite form, Kanara civilian",
     "message": "the Kanara"
   },
   "KANARA_INDEF": {
-    "description": "ship name - indefinite form",
+    "description": "ship name - indefinite form, Kanara civilian",
     "message": "a Kanara"
+  },
+  "KANARA_POLICE": {
+    "description": "ship name - base form",
+    "message": "Kanara police ship"
+  },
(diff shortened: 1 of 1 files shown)


zonkmachine:
@impaktor Perhaps this is better fixed manually on Transifex? So you can control precisely which tokens are renamed. It's not obvious from the diff.

data/lang/ships/en.json
```diff
@@ -84,28 +84,28 @@
...
+  "KANARA_POLICE_DEF": {
     "description": "ship - definite form",
-    "message": "the Kanara"
+    "message": "the Kanara police ship"
   },
-  "KANARA_INDEF": {
+  "KANARA_POLICE_INDEF": {
     "description": "ship name - indefinite form",
-    "message": "a Kanara"
+    "message": "a Kanara police ship"
```
sturnclaw: Typically this would be expressed as "a police model Kanara" in English, whereas in the form "X police ship", X typically refers to the owner, not the make and model. E.g. a "SolFed Police ship". Compare e.g. a "State Police interceptor" vs. "police model Dodge Charger".

zonkmachine: OK. Do you want me to fix the other police ship too?

sturnclaw: At your discretion. This is something specific to the indefinite form in English, you don't typically say "the police model Kanara" unless you're talking about a singular indefinite, e.g. "the police model of (the) Kanara" or a possessive e.g. "his police model Kanara". Not sure how you'd address the definite form (or the regular form for that matter); it would be quite context-dependent.

zonkmachine: OK, thanks! The Assassination string would then be:

**Her ship is a police model Kanara and has registration id XX-4521**

zonkmachine: Fixed!

sturnclaw: > OK, thanks! The Assassination string would then be:
> 
> **Her ship is a police model Kanara and has registration id XX-4521**

This reads quite naturally, thanks!

PR #6356, opened 2026-06-07
Upgrade Dear Imgui to 1.92.8

sturnclaw:
I've not (yet) gone through all of the intervening releases and dealt with all of the major/breaking changes from each, but this PR has the headlining feature of an improved font rendering stack.

We've entirely gotten rid of our patch-based intrusive glyph tracking system, and have adopted ImGui's new default scalable font rasterization. As a consequence, most font glyphs (especially outside the Latin-1 range, for non-English languages) should seamlessly load on the same frame they are used, with no need for a 1-frame delay to rebuild the font atlas anymore.

We have a new workaround for handling SVG font loading, as rasterizing the icon grid at large scales is still prohibitively expensive. We defer uploading missing icon glyph data until it becomes available, which will result in a 1-2 frame "pop in" as icons are rendered at new sizes.

I've had to do some rocket surgery to the Sector Map label rendering (cc @Gliese852), and would like to formally adopt a maximum size limit to labels in the map - they don't need to cover 1/4 of the viewport when very close to the camera. There is already a hacky maximum size limit I added to prevent out-of-memory conditions, but I'd appreciate a hand with changing the label system to use de-projected coordinates rather than the NDC Z coordinate directly...

Consider this PR a WIP that I intend to finish in the next few days - it's already taken ~12 hours to get to this state.

### DIFF
--- a/contrib/imgui/LICENSE.txt
+++ b/contrib/imgui/LICENSE.txt
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014-2025 Omar Cornut
+Copyright (c) 2014-2026 Omar Cornut
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/contrib/imgui/backends/imgui_impl_sdl2.cpp
+++ b/contrib/imgui/backends/imgui_impl_sdl2.cpp
@@ -7,13 +7,13 @@
 //  [X] Platform: Clipboard support.
 //  [X] Platform: Mouse support. Can discriminate Mouse/TouchScreen.
 //  [X] Platform: Keyboard support. Since 1.87 we are using the io.AddKeyEvent() function. Pass ImGuiKey values to all key functions e.g. ImGui::IsKeyPressed(ImGuiKey_Space). [Legacy SDL_SCANCODE_* values are obsolete since 1.87 and not supported since 1.91.5]
-//  [X] Platform: Gamepad support. Enabled with 'io.ConfigFlags |= ImGuiConfigFlags_NavEnableGamepad'.
+//  [X] Platform: Gamepad support.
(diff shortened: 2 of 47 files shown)


zonkmachine:
On a fresh build
```cpp
/pioneer/src/pigui/PiGui.cpp:243:27: error: qualified name does not name a class before ‘{’ token
  243 | struct PiGui::PiSVGLoader {
      |                           ^
/home/zonkmachine/builds/pioneer/src/pigui/PiGui.cpp: In member function ‘void PiGui::Instance::LoadFontDefinitionFromFile(const std::string&)’:
/home/zonkmachine/builds/pioneer/src/pigui/PiGui.cpp:427:42: error: ‘PiSVGLoader’ has not been declared
  427 |                         cfg.FontLoader = PiSVGLoader::GetFontLoader();
      |                                          ^~~~~~~~~~~

```

sturnclaw:
Good catch, thank you! My GCC version is a bit newer it would seem, as it has no problems with whatever C++ witchery I managed to accidentally commit.

Gliese852:
> I've had to do some rocket surgery to the Sector Map label rendering (cc @Gliese852), and would like to formally adopt a maximum size limit to labels in the map - they don't need to cover 1/4 of the viewport when very close to the camera. There is already a hacky maximum size limit I added to prevent out-of-memory conditions, but I'd appreciate a hand with changing the label system to use de-projected coordinates rather than the NDC Z coordinate directly...

@sturnclaw I'll try to dig into this if I can keep up with you. Although I personally like large letters that fill half the screen, they become more transparent and do not interfere (in my opinion) with view.

sturnclaw:
@Gliese852 

> I'll try to dig into this if I can keep up with you. Although I personally like large letters that fill half the screen, they become more transparent and do not interfere (in my opinion) with view.

Worst-case performance right now with the limited-maximum-size hack is a 2048x4096 font atlas, so it's not a critical bug. However, long-term the sector map labels need to be moved to a custom SDF-based text rendering system rather than using ImGui; ImGui's font atlas solution is not suited for smoothly/infinitely scaling labels. We can still use ImGui for hit-testing etc., but the label visuals themselves need to be rendered via a different method.

Using our own SDF font rendering path will also let us drop the `PrimDepth` patch to ImGui in favor of a more reasonable solution.

PR #6357, opened 2026-06-07
Cargo - upper case

zonkmachine:
Some cases of cargo being lower case fixed.

- World view cargo window
- Economy and Trade, cargo list

In the latter view, cargo is preceded by a weight amount, but I take it as being a different column and the cargo name looks better in upper case.

cargoucase

### DIFF
--- a/data/pigui/modules/info-view/03-econ-trade.lua
+++ b/data/pigui/modules/info-view/03-econ-trade.lua
@@ -79,7 +79,7 @@ local function cargolist ()
 		-- name
 		ui.tableNextColumn()
 		ui.alignTextToButtonPadding()
-		ui.text(entry.commodity:GetName())
+		ui.text(entry.commodity:GetProperName())
 
 		-- jettison button
 		ui.tableNextColumn()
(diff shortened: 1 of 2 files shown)


sturnclaw (changes requested):
Please prefer `CommodityType:GetProperName()` to `CommodityType:GetName():scase()`. We cache the proper-noun version as a pre-emptive optimization to expanding `scase` to a more complete (and more expensive) Unicode-aware version.

zonkmachine:
@sturnclaw Do you mean `CommodityType:GetProperName()` is preferred when you do string operations on the string object/string or just in general?

sturnclaw:
Any time you need to use uppercase names for commodity items, use `:GetProperName()` rather than performing an immediate `scase()`. There's no need to re-create the proper name as it's already cached.

sturnclaw:
In this particular case, the `:scase()` is redundant as `GetProperName()` is already correctly capitalized.

zonkmachine:
OK. Had a peek into `GetProperName()` and it's all clear now. :+1:

sturnclaw (approved):

PR #6358, opened 2026-06-07
Font padding

### LINKED ISSUE #5476: Top of letters with diacritical marks are cut off in some cases.
Letters on the first BBS post have their top part cut off if they stick up above the rest such as with diacritical marks. Example below is the Swedish letter **'Ä'** being decapitated. This seem to be general of the text on any first row on a page. Another example being the name of the station in the lobby. Swap out Cydonia in data/systems/00_sol.lua to 'Nuevo Östersund'.

### Observed behaviour
BBS-top-margin

```
git show
commit e3a86ef82c44d707fdc04ea3581991642d905779 (HEAD -> master, upstream/master, upstream/HEAD)
Author: Pioneer Transifex <pioneer-transifex@pioneerspacesim.net>
Date:   Tue Jan 17 03:01:20 2023 +0100

    auto-commit: translation updates
```

Release 22.04.1 LTS (Jammy Jellyfish) 64-bit
Kernel Linux 5.15.0-58-generic x86_64

zonkmachine:
Fixes #5476

Adds padding to the Lobby and the BBS, to allow for letters not to be cut off on the top. This is the case with some letters with umlaut, ÅÄÖÜ. The solution for the Lobby works well. The padding in the BBS also works, but it consumes a bit to much vertical space. On my screen this means 8 bulletin posts instead of 9.

Edit: Examples below in Swedish.

Now. Letters cut off on the first post 'AFFÄRSRESA'
bbsnow

Fixed. Less space on the BBS
bbsfixed

### DIFF
--- a/data/pigui/modules/station-view/02-bulletinBoard.lua
+++ b/data/pigui/modules/station-view/02-bulletinBoard.lua
@@ -77,6 +77,9 @@ bulletinBoard = Table.New("BulletinBoardTable", false, {
 		ui.setColumnWidth(0, self.style.size.x)
 	end,
 	renderItem = function(self, item, key)
+		if key == 1 then	-- Insert padding or Orbiteer gets cut off on the first bbs item.
+			ui.dummy(Vector2(0, 0))
+		end
 		local icon = item.icon or "default"
 		local region = ui.getContentRegion()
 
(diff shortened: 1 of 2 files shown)


data/pigui/modules/station-view/02-bulletinBoard.lua
```diff
@@ -77,6 +77,7 @@ bulletinBoard = Table.New("BulletinBoardTable", false, {
 		ui.setColumnWidth(0, self.style.size.x)
 	end,
 	renderItem = function(self, item, key)
+		--ui.dummy(Vector2(0, 0)) -- Insert padding or Orbiteer gets cut off on the first bbs item.
```
sturnclaw: Is this supposed to be commented out? 😛

sturnclaw:
In practice, what should be needed is just a spacing before submitting the first item in the BBS. It may already have an appropriate hook, or you may need to modify the container which submits the BBS list entries.

data/pigui/modules/station-view/02-bulletinBoard.lua
zonkmachine: Not really. No... 😢

zonkmachine:
> or you may need to modify the container which submits the BBS list entries.

I added a test for `key == 1` and I think this fixed it.

sturnclaw (approved):

PR #6359, opened 2026-06-07
Update CMake and CI build files

sturnclaw:
This PR should fix #6307, while not breaking MSVC builds linking to `OpenGL::GL`. CC @fluffyfreak for testing.

I've also addressed a bunch of pending deprecation notices from Github regarding the Node 20 runner used in a lot of our CI actions. They have been updated and changed to take advantage of new features in the action where it makes sense.


fluffyfreak:
Builds and run in MVSC on Windows 👍

fluffyfreak (approved):

PR #6360, opened 2026-06-08
Commodities in cargo missions

zonkmachine:
As mentioned on IRC. Adds some of the commodities to the cargo runs. Only the strings are used so they should not interfere with each other. The new cargo is registered again under a new dummy name cargo_<name of commodity>, and then added to branches with the other cargo.

medicines

### DIFF
--- a/data/lang/module-cargorun/en.json
+++ b/data/lang/module-cargorun/en.json
@@ -403,6 +403,10 @@
     "description": "",
     "message": "I'm taking damage. I need help."
   },
+  "LIQUOR_RISK_MESSAGE_1": {
+    "description": "",
+    "message": "Keep your hands off the booze."
+  },
   "LOAD_CARGO": {
     "description": "UI button label",
     "message": "Load cargo"
--- a/data/modules/CargoRun/CargoTypes.lua
+++ b/data/modules/CargoRun/CargoTypes.lua
@@ -110,18 +110,107 @@ local stem_bolts = CommodityType.RegisterCommodity("stem_bolts", {
 	l10n_resource="module-cargorun"
 })
 
+
+-- Adapted from the commodity market. They only share
+-- the strings and price from the commodities.
+local cargo_chemicals = CommodityType.RegisterCommodity("cargo_chemicals", {
+	l10n_key = 'CHEMICALS', price=41,			-- 124 (price in commodities)
+	mass=1, purchasable=false,
+	l10n_resource="commodity"
+})
+local cargo_computers = CommodityType.RegisterCommodity("cargo_computers", {
+	l10n_key = 'COMPUTERS', price=153,			-- 461
+	mass=1, purchasable=false,
+	l10n_resource="commodity"
+})
+local cargo_consumer_goods = CommodityType.RegisterCommodity("cargo_consumer_goods", {
+	l10n_key = 'CONSUMER_GOODS', price=82,		-- 246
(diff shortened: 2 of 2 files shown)


impaktor:
Remind me: 
- What would the profit for the player if they sold the medicine (in the screen shot) over the market instead? 
- If it would be more profitable to sell over the market, and the player does that, what happens with the mission?

I'm slightly skeptical, as the cargo run missions were a way to give the player a chance to transport something other than what's on the commodity market, plus some 'fun' falvours, like wedding dresses in urgent need. Adding the normal cargo might "water down" the mission, or at least be unnecessary?

EDIT: I imagine if the client needs medicine, she can buy it over the market at the station where it's needed. If medicine is sold out, then we're in the NewsEventCommodity module territory, or SoldOut module.

zonkmachine:
> I'm slightly skeptical, as the cargo run missions were a way to give the player a chance to transport something other than what's on the commodity market, plus some 'fun' falvours, like wedding dresses in urgent need. Adding the normal cargo might "water down" the mission, or at least be unnecessary?

To me it's an uncanny valley kind of thing. The two groups are kept apart but I see now reason for it. Why could you not have a shipment of medicine? It's not specified what kind of medicine in either case. It could be a sample of something and I've sketched on that kind of scenario. This is more akin to how it works in the real world. You buy and sell stuff or you get contracted to take similar stuff from point A to point B.


>* What would the profit for the player if they sold the medicine (in the screen shot) over the market instead?
>* If it would be more profitable to sell over the market, and the player does that, what happens with the mission?

In this iteration you can't sell the goods. Most of the strings of the new goods are either already translated or deliberately untranslated (.ja, .zh, etc.). I've used it for a while now and it doesn't feel strange at all. I could perhaps limit this cargo to just one branch each, to make it appear less frequently if that would somehow sweeten the deal? ;)

zonkmachine:
Issue: Custom risk messages get wonky in this PR. Looking into.

impaktor:
My words are not necessarily actionable, I'm just thinking out loud. @sturnclaw was positive to this, so.

zonkmachine:
> Issue: Custom risk messages get wonky in this PR. Looking into.

Fixed! The prices needed to be adapted to the prices of the original cargo with a max_price of 300, which is used to calculate the risk of the mission. I added one new custom risk message and they go into module-cargorun and not commodities as getNumberOfFlavours() is hardcoded for one module.

sturnclaw:
Chiming in here with some details from IRC that lead to this PR - the intent is that I will tackle the CargoManager during the upcoming development cycle and add support for carrying both equipment items and "tagged" cargo (e.g. unsaleable mission cargo, scooped stolen cargo, etc.). A planned goal is to be able to associate metadata such as purchase price with (groups of) cargo, so the flight log can record cargo trading profit in-game.

As a consequence, the current approach of mission-specific duplicated commodities is not one I expect to make it to merge; instead cargo missions that give the player "regular" commodities will also provide the necessary metadata to tag those commodities as "unowned" (with similar consequences to carrying stolen cargo if the player defaults on the mission).

zonkmachine:
> the current approach of mission-specific duplicated commodities is not one I expect to make it to merge;

OK. I drop it down to draft. I believe the general idea to have more variety in cargo/commodities to be valid though. I'll keep this on a simmer until the future CargoManager drops.

PR #6361, opened 2026-06-09
Save and recover Comms log

zonkmachine:
I looked into recovering the Comms log on game start and turned out to be pretty easy. So easy in fact it made me think that there might be a reason it isn't there in the first place but I checked our local Comms log and couldn't find any.

### DIFF
--- a/data/libs/Game.lua
+++ b/data/libs/Game.lua
@@ -9,7 +9,7 @@ Game.AddCommsLogLine = function(text, sender, priority)
 	table.insert(Game.comms_log_lines, { text=text, time=Game.time, sender=sender, priority = priority or 'normal'})
 	if type(priority) == "number" and priority > 0 then
 		Game.SetTimeAcceleration("1x")
-	end 
+	end
 end
 
 Game.GetCommsLines = function()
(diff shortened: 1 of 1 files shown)


data/libs/Game.lua
```diff
@@ -28,7 +28,6 @@ function Game.GetStartTime()
 end
 
 Event.Register('onGameStart', function()
-	Game.comms_log_lines = {}
```
Gliese852, line 31: Is this line definitely not needed when starting a new game?

zonkmachine: Thanks! It's definitely needed. If you exit a game and start a new one without shutting off the program in between, the previous games log is still there. `_deserialise()` seem to run before onGameStart() so something needs to change. I move it to onGameEnd() instead next to where gameStartTime is reset.

Gliese852 (approved):

data/libs/Game.lua
sturnclaw: As you go through the codebase, you'll notice a lot of deferred-reset patterns clearing game data in `onGameStart`. This is a legacy pattern because it was previously the case that `onGameEnd` handlers could run before serialization of autosaves occurred. It is generally safe to migrate cleanup tasks like this to `onGameEnd` now.

sturnclaw (approved):

PR #6362, opened 2026-06-12
clang-format: Add SpaceBeforeCpp11BracedList: true

mfep:
See https://github.com/pioneerspacesim/pioneer/pull/6188#discussion_r2226549096
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#spacebeforecpp11bracedlist


mfep:
Note: a comment in clang-format.yml:

>     # Disable clang-format workflow until it will obey its own config and
>     # stop aligning function parameters with the opening parenthesis

however I didn't notice such behavior using clang-format 21.1.8

PR #6365, opened 2026-06-15
Round off new commodity stock on news event

zonkmachine:
newseventprice

The rest of the commodities comes neatly rounded to a whole number but not the News event ones.


data/modules/NewsEventCommodity/NewsEventCommodity.lua
```diff
@@ -377,7 +377,7 @@ local onPlayerDocked = function (ship, station)
 
 			-- Blend from absolutely no stock to normal stock levels for this demand as the event goes on
 			if newDemand > 0 then
-				newStock = newStock * (1.0 - progress)^2
+				newStock = math.ceil(newStock * (1.0 - progress)^2)
```
zonkmachine, line 380: I wonder if the `math.ceil(...)` should be on it's own step after the if?

impaktor:
looks good to me

PR #6367, opened 2026-06-18
Remove debian/ directory

nikolas:
It's best practice to handle debian packaging separately, and not in this upstream repo for Pioneer.

Pioneer's Debian packaging can be seen here:

  https://salsa.debian.org/games-team/pioneer


sturnclaw:
Considering that the debian build harness we maintain hasn't been materially modified in ~15 years, I have no objection to it being maintained out-of-tree from our perspective. Thanks!

PR #6372, opened 2026-07-02
Ship landing improvements

### LINKED ISSUE #5208: Inconsistent rescue landing in mountainous/canyon environment
### Observed behaviour
In the attached rescue landing mission, when you approach the ship at some meters altitude, sometimes you are suddenly thrown back 200m-300m altitude high.

### Expected behaviour
Consistent navigation and landing in such situation, without sudden 200m-300m altitude jump.
Better visibility/lighting: a _feature request_ to facilitate approaching ships in the darkness would be _lights mounted on the ship_ to switch on/off, or at least a kind of aura to lighten the next 50-100m 

### Steps to reproduce
Load saved game attached, try to land by the ship to rescue
[issue landing dark mountain.zip](https://github.com/pioneerspacesim/pioneer/files/6610645/issue.landing.dark.mountain.zip)

Mint Cinnamon 20.1
20210513 (0c84df0)

### LINKED ISSUE #6420: cannot refuel landed ship in refuel mission
### Observed behaviour
I've landed with 2 tons of hydrogen in cargo but nothing happens

Image

also note that local gravity is 0.00G, maybe that's important.

Also unlike other reports terrain isn't a cliff and is quite flat

### Expected behaviour

Another ship is refueled and I get paid.

### Steps to reproduce

I don't know.
Maybe try looking for a refuel mission on a tiny asteroid.

20260203
Linux Mint 22.3 - Cinnamon 64-bit

[output.txt](https://github.com/user-attachments/files/31837899/output.txt)

[save23.zip](https://github.com/user-attachments/files/31837982/zapis23.zip)

sam52796:
I was playing Pioneer and trying to land a ship on a planet surface, and after a little while I thought "This should be easier".

If you've ever thought the same thing, then I have good news.

One issue preventing me from landing is that the ground was at an angle, and it took ages trying to rotate the ship in different directions to get it to "snap" to the landing state. Turns out the "correct" angle was strictly vertical. Even if the ground is at an angle, you were supposed to land your ship so that one edge is clipping into the ground and the other is in mid-air.

So the first commit here is for making landing on a planet easier by taking into account the terrain angle. The correct landing angle is now the same as the terrain angle - i.e. it's what you're looking at. I also made the tolerance a bit more generous. So you can align your ship with the ground and land a bit more easily.

The second commit is to add shadows cast by ships onto the ground and onto landing pads. It's a pretty crude render. If the shadow overlaps the edge of the landing pad then all bets are off. Just try not to pay attention to the shadows in that case. But landing in general is now slightly easier because the shadow tells you when you're approaching the ground in a more visually intuitive way than the altitude meter.

Or at least that's the idea. Feedback welcome.

(impaktor-EDIT: closes #5208, closes #6420)


impaktor:
@bszlrd did this mockup 3 years ago, for added feature to the radar/scanner, which might be inspiring:

https://github.com/user-attachments/assets/1a73bdb0-223d-44d5-afa9-9a8ce80fab6a

bszlrd:
That one is more useful for pad landings, but it could be adapted. Should I do some mockups?

zonkmachine:
I believe this closes https://github.com/pioneerspacesim/pioneer/issues/5208

sturnclaw:
I love this solution! The conceptual approach is hilariously janky in the best of ways, but I absolutely applaud your ingenuity (and audacity) in implementing ship shadows as a full-screen pseudo- deferred-rendering composite pass. Pending a closer code-and-performance review, I'm quite happy to merge this if for no other reason than as motivation to get working on a proper shadow-mapping implementation.

src/scenegraph/StaticGeometry.cpp
```diff
@@ -53,10 +53,9 @@ namespace SceneGraph {
 		PROFILE_SCOPED()
 		Graphics::Renderer *r = GetRenderer();
 		r->SetTransform(trans);
+		Graphics::Material *overrideMat = rd ? rd->overrideMaterial : nullptr;
```
fluffyfreak, line 56: I'd personally change this to be like this to avoid checking per-item
```		Graphics::Material *overrideMat = rd ? rd->overrideMaterial : nullptr;
		if (overrideMat) {
			for (auto &it : m_meshes) {
				r->DrawMesh(it.meshObject.Get(), overrideMat);
			}
		} else {
			for (auto &it : m_meshes) {
				r->DrawMesh(it.meshObject.Get(), it.material.Get());
			}
		}
```

bszlrd:
This looks so good:

screenshot-20260703-181330
screenshot-20260703-181303
screenshot-20260703-181227
screenshot-20260703-183049

Quite some parking brakes on the ships :D
screenshot-20260703-183632

bszlrd:
Here's a save which shows a potential issue with the shadow (not an actual zip):
Edit, a savefile without mods: 
[shadow-issue.zip](https://github.com/user-attachments/files/29644460/shadow-issue.zip)
The shadow floats above the ground:

screenshot-20260703-185313
screenshot-20260703-185311

sam52796:
I've improved the terrain angle algorithm so that it copes better with ridges and valleys. Floating or strangely positioned shadows should be less common now. It's still possible to end up with something odd if you really try, especially on the crinkly bits of Mars. Still, I believe this now meets the bar for "good enough". 

I've also made @fluffyfreak's requested change, so it's ready for further review and merge.

Mc-Pain:
Does not cast any shadow here:
[landing-test.gz](https://github.com/user-attachments/files/29693748/landing-test.gz)

```
[11:25:25] <sam52796> Yes, it has to choose which plane to project onto. If it ends up choosing a plane that is too shallow, so the shadow would be very long and drawn out (often happens close to sunset) then the shadow code just aborts.
[11:25:47] <sam52796> I think it's better not to draw a shadow than to draw one that looks wrong.
```

sam52796:
From IRC:

The situation in your save file is that it's close to sunset, so the shadow angle is very low, and the code decided not to draw any shadows. If you go back in time a bit (or more realistically go forward in time by a bit less than a day) you'll see the shadow is drawn most of the time, and then fades out as sunset approaches. Your save is just after it finished fading out.

If you're playing with the code, you can set SHADOW_MAX_ANGLE and SHADOW_FADE_ANGLE to, say, 89 and 85, and then you may see why I decided not to draw the shadows at those shallow angles.

PR #6373, opened 2026-07-03
Add docking aid to flight reticle

sturnclaw:
@bszlrd whipped up a simple mockup for a landing navigation aid this afternoon, and I saw it and said "hold my beer". Turns out all you need to draw a line across a circle is the 2d form of a ray-circle intersection test, and the rest was history. They say a video's worth a thousand words, so [here's one (Youtube)](https://youtu.be/YiGDzYPJYN8).

image

I may still add a few quality-of-life improvements to this; I'm not certain the heading error tick flipping sides is helpful, and it may help to have a tick that simply points to the center of the landing pad, for example. (And the 3d radar could display an icon for the landing pad position in addition to the center of the starport...)

Feel free to give it a try; I've not added any distance checks, so the navaid can behave somewhat weirdly if you e.g. request docking clearance 25km away from an orbital. Feedback is welcome on that aspect!


impaktor:
Very cool! 
Small question regarding color theme: I thought the green (e.g. for hyper jump count down) was a leftover from our Frontier-clone days, and we were moving to orange / blue theme, or something colorblind-friendly?

~Is rotating space station also supported?~

EDIT: from IRC:

https://github.com/user-attachments/assets/afe4ef6a-2a6f-4225-93c8-446ac980dc01

bszlrd:
Cool!
I was thinking a bit more about it (wanted to make a video mockup, but your quick implementation beats that).
How about this? The line is not all through the circle, but has some gaps, and a dot at the center for heading, or sideways deviation display.
landing-aid-polish-001
landing-aid-polish-004
landing-aid-polish-002
landing-aid-polish-007

The center part-circles indicate longitudinal deviation from the pad. Would be the size of the central dot at about 200m, and the size of the reticule when the ship is over the pad. And larger when one overshoots. The distances might need to be tweaked, but 200m or a 1000m as a first guess seems to be a good value for the furthest it would show.
And there's a tick on the left side of the reticule to show altitude over pad, here 1000m for the top value also seems good. (Here I assumed 200 for both)
landing-aid-polish-003


I was thinking about the center dot showing sideways deviation instead of angle, because then it won't go outside fast when you are on top of the pad. But it might not be intuitive in movement, so if it is an angle, then the longitudinal distance semi-circles could always originate from the crosshair instead of the dot. But then you'd need a sideways indicator as well, I guess.
landing-aid-polish-006

For rough landings, the longitudinal distance indicator could be ommitted, and the dot could point to north (or even all four cardinal directions?). And the line rotation and elevation could indicate the vector of the ground under the ship. And it could be colored white, to show that it is not for a nav target. (For the altitude indicator I assumed 1000m for top value on this mockup)
landing-aid-polish-008

Zireael07:
Small visibility request: can someone make a mockup with the line being magenta? From my colors research, saturated magenta, saturated green and saturated orange are the best for flight HUDs, and magenta works best on sky background

bszlrd:
Additionally, these could be adapted to #6354 too:
landing-aid-polish-009-largereticule

sturnclaw:
@bszlrd I like those mockups from a visual aesthetic approach, though clipping the circle primitives to the reticle circle is going to be painful. I'll evaluate them more closely when I have more brainpower available.

My intent was to keep all lateral positioning information on the 3d radar plane; the spoke-and-icon approach does a very good job of indicating 3d positioning relative to the pad (if it's showing the pad icon in the correct place). I may implement your previous X/Y error line mockup onto the radar display as well, but I wanted to keep the inside of the reticle very clean, and keep as much translation information out of the reticle as possible. It's already quite crowded with the selected-target-direction icon, the target-relative-velocity icon, and the landing attitude assist line.

sturnclaw:
image

I believe I'm calling this branch done here. I've looked at the mockups you provided @bszlrd and while I like them, I think they're somewhat out of scope to implement at this time, and I'm not certain they provide significant added clarity to what's already implemented.

The center dot in your mockup seems to provide the visual semantic of pointing towards the landing pad, which would make it invisible under 99% of ship orientations, and generally has more complications in attempting to render it when at a pitch approaching 90 degrees, where the line crosses an extremely small portion of the circle. I do like the altitude tick, but I think it's somewhat redundant when the vertical speed / altitude readouts are already in text form on the left side of the flight reticle, and I'm not going to expend the time to implement it at the moment.

Since I originally opened the branch, I've added a faint tint to the "ground half" of the area under the flight reticle circle to function as an artificial horizon indicator, and added x/z horizontal error lines to the planar radar (azimuthal radar has not received these lines and can be tackled by someone else in a separate PR). This should provide sufficient clarity to allow instrument-only landings on ground stations and orbitals without needing to use the down-facing camera for anything other than altitude-to-pad.

I'd definitely recommend adding the mockups to the UI rework issue/album and they can be tackled by another contributor (or myself) at a later date.

bszlrd:
I'll try it out tomorrow

bszlrd (approved):
Lookin' good. I don't have much to add. Tested it on orbitals and ground stations as well, and it makes landings much easier.

sturnclaw:
Merging; the test failures are caused by the OpenAL audio backend being initialized in an environment without audio hardware.

PR #6375, opened 2026-07-04
Facegen rework

bszlrd:
This is a WIP PR for the facegen rework. 
**It won't work as is**, I would like to request code to accomodate the changes in image name, size and layering. It might even crash the game.
Putting up the PR to have an asset basis for those code changes. (Which could also be pushed to this PR?)
image
faces_01
faces_02

[Here's a video showing 500 randomly generated ones](https://www.youtube.com/watch?v=qOEopHwDCZU) (My random hair color generator does not seem to be that random)
There is a bit of position and scale variation to the features, but I'm not seeing any uncanny combination in that video. So I think this stylized approach will work nicely. And it fits the game much better than our current very generic (and very cursed) facegen.

# Layer setup
Each image has fixed 1394x1346 dimensions, a bit over 4* the size of the original facegen. This will make them work nicely even on a 4k screen. Not sure about the compression rate though. We might want to consider halving the size, if they prove too large.
Ideally, the system would be scale-agnostic.

More importantly, they now don't need hardcoded offsets. Since they are png-s anyway, file size wouldn't be that much smaller if they were cropped. On the other hand this allows for more flexibility for shading and proportions.

There are 13 layers for now, in the following order:
- **Hair**
   - **02_ear** - tied to specific hair images. Needs to be separate layer, so hair can be independently colored
   - **03_hair_front** - lineart and gray silhouette, with shading painted onto it, to allow arbitrary colors
      - For coloring, I've used a solid color with Overlay blending mode on this layer. First I color the layer, then apply the alpha channel
   - _These, and 13_hair_back should be picked together_
- **Lineart of the head**
   - **04_eyes_line**
   - **05_nose_line**
   - **06_mouth_line**
   - **07_head_line**
   - _Made them separate from coloring, to avoid lines being obscured by shading_
   - _They should be picked together with their color counterparts_
- **Colors and shading of the head**
   - **08_nose_color** - on top of eye coloring, to avoid eye shading abscuring parts of it
   - **Eyes** - these two should be picked together
      - **09_eye_shading** - the whites of the eye, and highlights
      - **09_eye_color** - only the alpha channel is needed, to drive the transparency of an arbitrary solid color
   - **10_mouth_color** 
   - **11_head_color**
   - _It would be useful if the layers of the head coloring would clip to the alpha channel of 11_head_color to avoid brushstrokes appearing outside the head_ 
- **12_body** color and lineart combined. 
   - Combined with clothing, so uniforms will be additional body types picked for specific characters.
- **13_hair_black** the part of the hair that can be obscured by the head or the body. (and yes, there's a typo in the name)
   - _Should be picked with hair front and ears_ 
- There are only two generic flat color background for now. Eventually there could be specific background sets for ATC, ships, generic calls, mechanics, etc.

There are several features that are multi-layer, (ie: lineart+shading), and these should always be picked together. In other words: the 1st ear always goes with the first hair_front and hair_back. This allows for arbitrary eye and hair coloring for example, and lineart and coloring can be separated, where that's useful.
facegen_docs

Apart from the eye and hair coloring, all layers go on top of each other with simple alpha blending, no shenanigans needed.
For coloring those two, I could provide a list of "legal" colors for the procedural generator, but it would be nice, if the choice of arbitrary color would be available for the player to hand-pick. 

Eventually I want to include helmets, which would practically be a type of hair picked for specific characters.
Not yet sure about accessories and face details like freckles and scars and tattoos. Freckels for now are part of the nose lineart, but that could be changed. More problematic are headgear and such, since they need to conform to at least the ears and eyes, making them a bit unpredictable. 
Same goes for facial hair. For now I think I'll ommit them, or maybe including them on some head and nose layers. 

There's also a bit of noise on each layer, but it might be better to separate that to an additional top overlay for more flexibility.

# Directory structure
The assets in all layers are named frame####.png. This is how Krita does one-click batch exporting, so I'd like to keep it that way to avoid constant renaming.
This is why I've rearranged the directory structure as well. Now the male and female sets are in separate directories:
- species_0
   - race_0_0 (male)
   - race_0_1 (female)
   - race_1_0
   - race_1_1
   - race_2_0
   - race_2_1

I've kept the numbering convention (0=male, 1=female) to stay in line with the name generator.
I've included two genders of three races, so there's a bit of parity with the old facegen, but apart from caucasian female, all are placeholders. Males look left, females mirrored look right, and the other races are smurfs and ogres (same sets tinted blue and green).

Currently there are 8 set of features for caucasian females, including a bald option. If we consider the combinations of heads+mouths+noses+ears as unique faces, then it nets 4096 face for one gender of one race. When I reach parity of 20 each, that means 120k face per gender per race.


zonkmachine:
Edit: Missed this part...

> It won't work as is, 

...

I get a crash on starting a new game. I built facegen_rework on a clean repo because I suspected it was on my side but I get the same crash after selecting 'New Game' from the start menu.

```cpp
Thread 1 "pioneer" received signal SIGFPE, Arithmetic exception.
FaceParts::PickFaceParts (inout_face=..., seed=<optimized out>)
    at /home/zonkmachine/builds/pioneer-facegen/pioneer/src/FaceParts.cpp:395
395		_pick(rand, inout_face.head, _count_parts(s_partdb->heads, selector));
(gdb) bt
#0  FaceParts::PickFaceParts (inout_face=..., seed=<optimized out>)
    at /home/zonkmachine/builds/pioneer-facegen/pioneer/src/FaceParts.cpp:395
#1  0x000055555589c711 in PiGui::Face::Face (this=this@entry=0x55555b3ee310, face=..., seed=<optimized out>, seed@entry=3949037595)
    at /home/zonkmachine/builds/pioneer-facegen/pioneer/src/pigui/Face.cpp:21
#2  0x0000555555816126 in PiGui::LuaPiguiFace::l_new (l=<optimized out>)
    at /home/zonkmachine/builds/pioneer-facegen/pioneer/src/pigui/LuaFace.cpp:55
```

Ubuntu 24.04.04
[opengl.txt](https://github.com/user-attachments/files/29680911/opengl.txt)
[output.txt](https://github.com/user-attachments/files/29680912/output.txt)

PR #6376, opened 2026-07-04
Ship landing improvements 2: The Lightbringer

sam52796:
I was playing Pioneer and trying to land a ship at night, and after a little while I thought "This should be easier".

If you've ever thought the same thing, then I have good news.

With this change, ship nav lights and thrusters now illuminate the terrain at night. This addresses the other complaint mentioned in #5208.

As a little bonus, starport city lights also illuminate the terrain at night.

By the way, if you find the blinking red and green illumination annoying then don't blame me, it's the ship designer's fault! My code just shows the logical result of the lights that they put on the ships!

Seriously though, if you think any of this should be tweaked then let me know. Or play with the constants at the top of TerrainLocalLights.cpp yourself.

### DIFF
--- a/data/shaders/opengl/geosphere_terrain.frag
+++ b/data/shaders/opengl/geosphere_terrain.frag
@@ -4,6 +4,7 @@
 #include "attributes.glsl"
 #include "lib.glsl"
 #include "basesphere_uniforms.glsl"
+#include "terrain_local_lights.glsl"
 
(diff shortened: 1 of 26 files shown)


sam52796:
With thruster exhaust plumes in master, this PR can have a second commit to make ship nav-lights and thruster glow illuminate the exhaust.

Now there really is no escape from the blinking lights.

PR #6377, opened 2026-07-04
Search & Rescue mission checks cargo hold for additional fuel

sam52796:
If you do a Search & Rescue mission to deliver fuel to a stranded ship, and you don't have the fuel in your cargo hold when you reach them, the mission code gives up. You have to take off and land again in order to re-trigger the fuel check.

With this change, if you pump down fuel from your fuel tank to the cargo hold, the mission picks that up and resumes the transfer, without you needing to take off and land again.

I also made the mission show progress messages, e.g. "1 of 3 tons transferred", "2 of 3 tons transferred" as the transfer progresses. This makes things a bit clearer when you have some of the fuel in your cargo hold but not all of it - you'll see that some was transferred before getting the "Cannot complete transfer" message. I think this was the original intended implementation, given that the message was already written that way. So no new translations are needed for this.

### DIFF
--- a/data/modules/SearchRescue/SearchRescue.lua
+++ b/data/modules/SearchRescue/SearchRescue.lua
@@ -1378,6 +1378,7 @@ local missionStatusReset = function (mission)
 	for commodity,_ in pairs(mission.deliver_comm_check) do
 		if mission.deliver_comm_check[commodity] == "PARTIAL" then mission.deliver_comm_check[commodity] = "NOT" end
 	end
+	if mission.deliver_comm_msg_shown then mission.deliver_comm_msg_shown = nil end
 end
 
 local closeMission = function (mission)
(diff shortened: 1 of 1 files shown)


zonkmachine:
I gave this a quick spin and it works just fine.

I started to trouble shoot the Search and Rescue, fuel transfer mission, about two weeks ago and I believe I spent a full day on preparing an issue for it. Now it seem you've fixed all the issues I found except one. The wonky last approach when flying to the target on auto pilot.

sam52796:
It was your savegame that you added to #5208 that clued me in to this fuel transfer issue, so thanks for creating that and uploading it.

I had noticed before that the auto pilot setting "Fly to vicinity of X" doesn't work great when the target is a ship on the ground, but presumably that's a generic problem and not specific to this mission type. (Even if this mission type is the only time that it would come up.)

data/modules/SearchRescue/SearchRescue.lua
```diff
@@ -1691,14 +1700,16 @@ local deliverCommodity = function (mission, commodity)
...
-		-- show result message if done delivering this commodity
 		mission.deliver_comm[commodity] = mission.deliver_comm[commodity] - 1
 		mission.deliver_comm_check[commodity] = nil -- clear PARTIAL flag if present
+		if mission.deliver_comm_msg_shown then
+			mission.deliver_comm_msg_shown[commodity] = nil
+		end
 
 		local done = mission.deliver_comm_orig[commodity] - mission.deliver_comm[commodity]
+		local resulttxt = string.interp(l.RESULT_DELIVERY_COMM, {done = done, todo = todo, cargotype = commodity_name})
+		Comms.ImportantMessage(resulttxt)
```
sturnclaw: Note: this can spam the CommsLog with several hundred lines. I'd strongly recommend that this send only every n-th commodity where the total number of messages sent to transfer all commodities is 10 or fewer.

sturnclaw (changes requested):

sam52796:
I've restricted the number of "X of Y amount transferred" messages to a maximum of 10, and downgraded their importance so the player can time-accel past them.

Also removed some duplicate code that was causing the "Interaction distance reached" message to show up out of sequence.

sturnclaw (approved):

PR #6380, opened 2026-07-05
Fix ImGui error message of extending parent boundaries with `SetCursorPos`

mfep:
Since #6356, an ImGui warning message appeared on the Equipment screen, related to a changed API behavior of `SetCursorPos`. See https://github.com/ocornut/imgui/blob/776bf2ab0d61e719e58f2c6d27d109ab5dcf2af1/imgui.cpp#L11003-L11028 for reference.

This patch fixes this by adding a `Dummy`, as the documentation suggests.

Screenshot of the warning:

Screenshot From 2026-07-05 19-03-17

### DIFF
--- a/data/pigui/libs/ship-equipment.lua
+++ b/data/pigui/libs/ship-equipment.lua
@@ -440,6 +440,10 @@ function EquipmentWidget:drawSectionHeader(section, fun)
 
 	ui.setCursorPos(contentsPos)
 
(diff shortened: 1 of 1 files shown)


zonkmachine:
Same ImGui message appears in the system map and system overview windows if I press '**i**' (Object info) without anything selected.

_Edit: New issue #6395_

PR #6382, opened 2026-07-06
TWR

sam52796:
First commit attempts to clean up some UI and gameplay issues:

- The UI gauges along the bottom of the station view (comms) screens now have "Cargo" as well as "Equipment", where before they just showed "Capacity". I have tried to ban the word "capacity" from everywhere in the UI, to remove any ambiguity about what any given number refers to.
- If you have no passenger cabins fitted, the UI gauge now shows "None" instead of "100% of 0 cabins used".
- In most places in the UI that refer to "cabins" I have replaced this with "berths" since that number is generally what the player wants to know.
- In the equipment screen the mass of a fitted hyperdrive was shown including the fuel, which made the comparison to alternative (or even the same) hyperdrive types come out wrong. Hyperdrive mass is now shown as the bare mass. Fuel reservoir size was already shown separately.
- An existing bug was that if the player doesn't have room to add a new piece of equipment, the message shown was "This equipment is not compatible with your ship", even if it is compatible. I added an additional message which says "Not enough space" in those circumstances instead.
- I made some minor cosmetic tweaks to the way that gauges are drawn, and made the capitalization of some UI strings consistent with other UI strings.
- The Pumpkinseed and Xylophis cargo stats seemed to be the wrong way around. Xylophis looks smaller and is the "hard start" ship, while Pumpkinseed is "medium start", and yet Pumpkinseed has 4 and Xylophis has 8. It makes more sense for the Pumpkinseed to have a higher cargo capacity than Xylophis, so I have swapped them.

Second commit adds some TWR features:

- A readout for the bottom right of most UI screens showing the current ship weight, the TWR for where you're currently docked, and the TWR for whatever nav target you have selected. The TWR numbers turn yellow and then red as the TWR approaches 1.0 or below.
- The TWR for your current ship weight is shown as part of the planet info in the system overview screen.
- If the TWR is over 99 then it shows infinity instead. Although the infinity symbol in the Pioneer font is kind of small, so actually I show "oo" instead. Which sounds strange, but honestly looks better than the infinity symbol as it currently is.
- In normal flight, a HUD warning shows if you select a starport or ship on a planet surface where the surface gravity results in a TWR of 1.1 or less. 
- In normal flight, if the ship detects that its effective TWR in a gravity well at its current location is 1.1 or less then it shows another HUD warning, similar to the existing descent warning.

All the TWR calculations are based on the "Up" thruster power, not the "Forward" thruster power, because that's what matters for taking off (for now). This means that since forward thrusters are usually stronger, the HUD warning situations are mostly escapable if you don't let the situation deteriorate too much. I consider this to be a feature rather than a bug, because calculating the TWR differently depending on the ship's orientation would be more confusing and would probably have worse bugs too.

Also, because the ships are all rather overpowered and most planets have lower gravity than Earth, it's quite difficult to trigger any of these low TWR warnings in normal gameplay. You have to fly to a gas giant to get them. But you can see the TWR readout numbers change quite easily by starting on Mars and going to the system overview and clicking on Jupiter or Sol.


impaktor:
First commit sounds good to me. Each bullet point could have been a separate commit, for future code archeologists that are looking at the code & commit history, but never mind. Second commit (TWR) I'll leave to others.

> The Pumpkinseed and Xylophis cargo stats seemed to be the wrong way around.

I believe @bszlrd has done measurements of the 3D model in blender to derive the value.

Also perhaps rename PR to something more descriptive, took a long while for me to understand it was thrust to weight ratio (I assume?).  

> All the TWR calculations are based on the "Up" thruster power, not the "Forward" thruster power

There are plans for tailsitters, I assume you'll have to leave a comment/documentation or something to help avoid future bug when that is in the game.

sam52796:
Hmm... if TWR isn't even understood by the Pioneer contributors, then players might find it a bit challenging. I've added some tootlips to the TWR readout to try to explain it. But if anyone can think up a better term then we can use that instead. Maybe "Lift ratio"? It's short enough not to need abbreviation.

In terms of the ship abilities and features, I think that gameplay considerations should be driving their appearance, not vice versa.

bszlrd:
Great!
I think the readouts could live near the launch button in the lobby, and would benefit from progress bars:
TWR
They could live in the corner still for the other screens (although that is not the most ideal placement in general, but that's a wider station screen design question. That black area is not used at all, which is not ideal. But that's for another day)
TWR-2
The bar could go up to TWR 2, differentiating between bellow and above 1.1 with different colors. And bellow 1.1 it would use the warning colors from the theme:
image
Above 2 is more than enough, that's why it just goes to 2. But alternatively it could scale the different color proportions above 2, showing actually how much reserve the ship has:
image

I think I'd ommit the blinking of the warning on the HUD. It is distracting (not eye-chatching, but actually distracting, as one tries to read it and it disappears, to relatively slowly appear again). I think it could just flash between two warning colors instead of disappearing. (We already do that for the Cruise control setting, when it does not match current speed)
Also, it overlaps witht he hyperspace message:
screenshot-20260708-132619

And its scale is rather large, but that's true for the hyperspace message too. This could be an alternative scale (and way to show the hyperspace message, since I was at it already:
TWR-HUD
I've also included a TWR and acceleration gauge the the lower-left of the HUD, near the thrust readout. It fits, to that small space above the icons nicely, and is directly related to tho thrust readout anyway. I think they both provide useful information.

impaktor:
> I think the readouts could live near the launch button in the lobby, and would benefit from progress bars

In a way it's a bit strange that the station lobby is showing info that is specific to my ship. Ideally the player would have a monitor that is in their cockpit, with info on the ship, and another monitor that interfaces with the station where the player can request launch clearance, and some amount of fuel, etc. Not suggesting any changes, just throwing it out there to plant a seed that might lead others to somewhere if they agree.

bszlrd:
True, but there's already a lot of information on the lobby about the ship. Capacities, fuel, cabins/berths.

sturnclaw:
> The Pumpkinseed and Xylophis cargo stats seemed to be the wrong way around. Xylophis looks smaller and is the "hard start" ship, while Pumpkinseed is "medium start", and yet Pumpkinseed has 4 and Xylophis has 8. It makes more sense for the Pumpkinseed to have a higher cargo capacity than Xylophis, so I have swapped them.

This is not the correct arena for those changes. Ship stats are defined in a [Notion database](https://bszlrd.notion.site/Ships-designer-1018c82e283f80b1a4f5cde6a8c9df56); changes go through a round of discussion and are imported to this repo via Github CI.

The cargo capacity values for Pumpkinseed and Xylophis are both intended and physically correct; the "hard" start comes from the lack of fitted hyperdrive in the Xylo and its low dV requiring it to take very il-lucrative in-system missions to build starting capital. The Pumpkinseed by comparison is a starter combat ship, with a small amount of cargo capacity but high thrust and a comprehensive weapons loadout.

Many other things besides cargo capacity could be changed to make the Xylophis start more challenging, or the Pumpkinseed start less challenging. For example, starting with greater reputation in the Pumpkinseed (to make combat missions more readily available), or with a more comprehensive weapon loadout.

-----

In general, I'm liking what I see here. Thanks for tackling a bunch of papercuts related to splitting equipment and cargo capacity! I'll give this a playtest and formal review as soon as my schedule allows.

I'd generally echo @bszlrd's feedback regarding simple gauges for TWR data below the portrait on the station lobby, as well as cleanup for the master alarm positioning and font size. The TWR warning is probably not needed unless the player has a surface starport set as their nav target; generally TWR is irrelevant in orbit. For testing purposes I'm glad it displays in an easy-to-reach situation however!

sam52796:
I've added coloured bars to the TWR bottom-right widget, and added a similar TWR readout to the HUD, as per @bszlrd's suggestions. The HUD readout only shows when the TWR is less than 100, so if you're in deep space or at an orbital station then it is hidden.

Now that there is a dedicated readout in the HUD, the flashing warning for being in a high gravity well is no longer needed, so I removed that one. The other warning still shows if you select a nav target where you wouldn't be able to take off again if you landed there, but I made the flashing less obnoxious.

The HUD warning overlapping with the hyperspace countdown is a result of the general warning positioning and happens with all the others too (e.g. descent rate), so I fixed it for all of them by moving the hyperspace countdown message up a few pixels.

I left the station lobby TWR display in the bottom right because it has to go there on the trading and ship equipment and system overview screens, so I think having it in a different place just for the station lobby would be confusing.

sam52796:
I've converted the TWR HUD indicator to use the new circular style. It now looks like the below.

We may not need both the TWR and the gravity indicator, but I'll leave that up to others to decide.

image

src/pigui/Widgets.cpp
```diff
@@ -258,7 +258,13 @@ void Draw::CircleIndicator(ImGuiID id, float diameter, const char *label, const
...
 	if (value >= 0.01) {
 		// Draw the "active" dial
-		dl->PathArcTo(center, outer_radius, phase - value * M_TAU, phase, 32);
+		if (symmetric) {
+			// Symmetric arc centred on phase, value=1 fills the full circle.
+			const float spread = value * M_PI;
+			dl->PathArcTo(center, outer_radius, phase - spread, phase + spread, 32);
+		} else {
+			dl->PathArcTo(center, outer_radius, phase - value * M_TAU, phase, 32);
+		}
```
sturnclaw: If you're not planning to finish the parameter implementation by changing the `value_inv` dial to work properly in `symmetric` mode, please revert this commit and simply pass `phase = TwrGauge.HUD_ARC_PHASE + math.pi * value` from Lua, which does the exact same thing.

data/libs/Gravity.lua
```diff
@@ -0,0 +1,137 @@
+-- Copyright © 2008-2026 Pioneer Developers. See AUTHORS.txt for details
+-- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
+
+local Gravity = {}
+
+Gravity.INFINITY = "oo"
```
sturnclaw: Prefer using the single unicode character LATIN CAPITAL LETTER OO (U+A74E) (Ꝏ) which produces this display:

Image

data/pigui/libs/twr-gauge.lua
```diff
@@ -0,0 +1,95 @@
...
+
+local ui = require 'pigui'
+local Vector2 = _G.Vector2
+
+local colors = ui.theme.colors
+
+local TwrGauge = {}
+
+-- TWR at the top of the bar (values above this max the bar out)
+TwrGauge.MAX_LEVEL = 2.0
```
sturnclaw, line 16: Instead of maxing out at 2.0, it may produce a clearer gauge to use the TWR of the ship without cargo (or cargo+fuel) so that it is more intuitive how heavily-laden the ship currently is. This would mean that the bar would stay relatively static as the player flies around (similar to current) but would display the achievable proportion of maximum acceleration rather than a static full gauge.

This isn't necessarily a change request, but rather a suggestion for presenting clearer semantic information to the player.

sturnclaw (changes requested):
I like the visual effect of the new gauges. I have some feedback about the implementation:

Gravity.lua does not meet my personal threshold for sufficient gameplay-related utility to add to the `libs` namespace. Its contents are almost entirely UI related, which means it should at the very minimum be moved to `data/pigui/libs/`. Additionally, most of its functionality is a few functions that wrap a walk up the (system)body tree to find the first ancestor Planet or Star to solve the gravity equation with.

This walk is the useful part of the module; it should be implemented as a separate function, and possibly implemented in C++ as a function or accessor on either LuaBody or LuaSystemBody for speed. The majority of the rest of the module is abstractions over a few lines of formatting or a property access plus a division.

The code that computes current gravitational state should not produce an invalid/zero value if the current frame is an orbital starport, but should continue to walk up the body tree until it finds a planet/asteroid/gas giant/star. `GetGravityAtBody` should only ever return nil if the body doesn't exist or the player is currently in hyperspace.

`Gravity.GetSurfaceGravity` and `Gravity.GetGravityAtBody` are also unnecessarily defensively programmed:
- A `frameBody` will not be dynamic, because frames must be created around a body that has an associated SystemBody. Even if it were to be dynamic, the test is irrelevant because you're interested in whether the associated SystemBody is a gravitational body.
- Body has an accessor `GetSystemBody()`. There is no need to side-trip through the `SystemPath` to retrieve the system body associated with a body. You can research Lua methods on an object either by reading the `LuaXXX.cpp` file which registers functions and accessors, or reading the LuaLS typeinfo files in `data/meta/*.lua`.
- `Ship.IsLanded` will only ever be true if the ship is rough-landed on a terrain surface.
- Why does the code test for negative gravity values?

`FormatTWR` and `FormatTWRHud` differ only in the number of decimal places they format and their rounding behavior when a value is too small to format. This can be unified by specifying a `places` parameter, and driving that parameter from the calling code.

Following convention, the resulting UI formatter for TWR should be moved to `data/pigui/libs/text.lua` into the `ui.Format` table; a single function there should provide the prototype:

```lua
function ui.Format.TWR(twr, places, infinity_threshold = 99.9)
```

Under no circumstance other than a nil parameter should the function `CalcTWR` return nil. If necessary, it should return `math.huge` instead. It should be the decision of the `ui.Format.TWR` whether to format large values as infinity, rather than the function that computes thrust-to-weight-ratio.

Of note, moving the responsibility for *formatting* out of `CalcTWR` reduces the function to a single line `upAccel / gravity`; if there is not a significant number of callsites (my rule of thumb: 6+) then that singular calculation should just be interned into the calling code.

-----

Other than the particulars of the Gravity module implementation, I'm generally in favor of this PR. Having the new values available on the station screen is useful to the player, and the layout is acceptable given the age and design of the station screens.

sturnclaw:
> We may not need both the TWR and the gravity indicator, but I'll leave that up to others to decide.

I'm fine to get rid of the gravity indicator if no one suggests a reason to keep it; it was implemented as an example regarding how to use the new circular indicator.

sam52796:
I've made those changes:

The gravity Lua lib is now in pigui/libs instead, and gets the gravity value by calling the C++ GetGravityForce() function, so the TWR readouts should match exactly whatever the physics sim is doing.

The TWR formatting is done in ui.Format as per convention.

The addition of a "symmetric" flag was indeed unnecessary on the circular indicator, so I'm now calculating the arc directly in Lua instead. Kind of annoyed with myself for not realising that the first time around.

The infinity symbol is now Ꝏ instead of oo. TIL.

<hr/>

I like the concept of using the maximum theoretical value of TWR for the empty ship as the "max" value on the TWR gauges, and I think there is potential there, but we would need to refine the concept a bit. I tried a naive implementation and the result was underwhelming for two reasons:

The first issue is that the empty ship is quite light compared to all the things you load it up with, and the player is usually going to have some equipment and some cargo and a full fuel tank, so the gauge then spends all its time at or below the lower third mark. I realise that in normal gameplay, fixing the max at 2.0 has the opposite problem, since most of the time the TWR will be above that, but I think that in the end the "always full" reading is more representative of "TWR is high enough, don't worry about it" than a third full but still normal colour gauge.

The second issue is that because the max TWR depends on the current gravity, then when the ship is flying around, the gauge arc doesn't change, since the ratio of max to current is always the same, barring small amounts of fuel burn. So when flying into a high gravity well the player sees the number change, and the gauge changes colour, but the arc itself doesn't budge, and it looks like it's not working.

So I have not made that change in this PR. Happy to discuss further.

bszlrd:
How it would look if the gauge scaling was not linear, but some flavour of logarithmic? Not sure how intuitive that is, I'd have to see it in action.

sam52796:
That might help a bit with the first issue, but I don't think it will make any difference to the second one.

data/pigui/libs/text.lua
```diff
@@ -388,6 +392,23 @@ ui.Format = {
...
+		places = places or 2
+		infinity_threshold = infinity_threshold or 99.9
+		if twr == nil or twr > infinity_threshold then
+			return "Ꝏ"
+		end
+		local minDisplay = 10 ^ -places
+		if twr < minDisplay then
+			return string.format("%0." .. places .. "f", 0)
+		end
+		return string.format("%0." .. places .. "f", twr)
```
sturnclaw:
```suggestion
		return string.format("%0." .. places .. "f", twr >= minDisplay and twr or 0)
```

data/pigui/modules/info-view/01-ship-info.lua
```diff
@@ -55,10 +58,10 @@ local function shipStats()
 		},
 		false,
 		{ l.WEIGHT_EMPTY..":",     string.format("%d t", player.staticMass - player.loadedMass) },
-		{ l.CAPACITY_USED..":",    string.format("%s (%s "..l.MAX..")", ui.Format.Volume(player.equipVolume), ui.Format.Volume(player.totalVolume) ) },
-		{ l.CARGO_SPACE_USED..":", string.format("%d cu (%d cu "..l.MAX..")", player.usedCargo, player.cargo_cap) },
 		{ l.FUEL_WEIGHT..":",      string.format("%.1f t (%.1f t "..l.MAX..")", player.fuelMassLeft, shipDef.fuelTankMass ) },
 		{ l.ALL_UP_WEIGHT..":",    string.format("%d t", mass_with_fuel ) },
+		{ l.EQUIPMENT_USED..":",   string.format("%s (%s "..l.MAX..")", ui.Format.Volume(player.equipVolume), ui.Format.Volume(player.totalVolume) ) },
+		{ l.CARGO_SPACE_USED..":", string.format("%d cu (%d cu "..l.MAX..")", player.usedCargo, player.cargo_cap) },
```
sturnclaw: Apologies that I didn't catch this on the first pass; the optimal way to express this would be:
```suggestion
		{ l.CARGO_SPACE_USED..":", string.format("%s (%s %s)", ui.Format.Cargo(player.usedCargo), ui.Format.Cargo(player.cargo_cap), l.MAX) },
```

data/pigui/libs/text.lua
```diff
@@ -352,6 +352,10 @@ ui.Format = {
 	Volume = function(number, places)
 		return ui.Format.Number(number, places or 1) .. " " .. lc.UNIT_CUBIC_METERS
 	end,
+	-- Format a cargo quantity in cu
+	Cargo = function(number)
+		return ui.Format.Number(number, 0) .. " cu"
```
sturnclaw, line 357: If interested in going above and beyond, the string `"cu"` should be defined in `core` or `ui-core` as `UNIT_CARGO` and used here rather than hardcoded as a string. Considering the original code didn't go to that length, I won't require it for merge.

sturnclaw (changes requested):
> I've made those changes:
> 
> The gravity Lua lib is now in pigui/libs instead, and gets the gravity value by calling the C++ GetGravityForce() function, so the TWR readouts should match exactly whatever the physics sim is doing.
> 
> The TWR formatting is done in ui.Format as per convention.
> 
> The addition of a "symmetric" flag was indeed unnecessary on the circular indicator, so I'm now calculating the arc directly in Lua instead. Kind of annoyed with myself for not realising that the first time around.
> 
> The infinity symbol is now Ꝏ instead of oo. TIL.

Looks good! There are actually 4 different unicode symbols that can represent infinity, and we have glyphs for three of them in the Pionillium font. This was the best character for the given font size.

> I like the concept of using the maximum theoretical value of TWR for the empty ship as the "max" value on the TWR gauges, and I think there is potential there, but we would need to refine the concept a bit. I tried a naive implementation and the result was underwhelming for two reasons:
> 
> The first issue is that [TWR is usually very far above 1.0]
> 
> The second issue is that because the max TWR depends on the current gravity [there's little visual feedback to the player]
> 
> So I have not made that change in this PR. Happy to discuss further.

I suspected that would be the outcome, which was why I intended to phrase the suggestion as an idea rather than a change request. Thanks for looking into it - I agree that while it might eventually be able to be worked into an intuitive display, the current implementation has a perfectly satisfactory developer-time-to-clarity ratio.

-----

Please note that this PR will not be merged until the changes to the Xylophis and Pumpkinseed ship.json files have been reverted. You are more than welcome to bring that up as a topic of discussion in a separate issue, on IRC, or on the forum (or link to the location you've already done it) and changes to ship balance can be discussed there.

sam52796:
I've made those code tweaks, though not the "cu as lang string" one, but including the ship cargo reversion. I believe the PR is ready for merge.

sturnclaw (approved):

PR #6383, opened 2026-07-07
Saving face

sam52796:
People seem quite hung up on the facegen. Personally I don't think it's that bad, but I can see that they're not quite worthy of the huge image sizes that appear in the station lobby, ship repair, and police screens. So this PR makes the face images on those three screens smaller, putting them more in line with the smaller headshots that appear in other places.

I've also added noise to all the NPC face images. The idea is that this will make them look more like remote video images, and also maybe distract from the faces themselves. The noise is taken from PNG files in data/facegen/noise, so other people can add new ones, or remove existing ones if any are particularly unpopular.


impaktor:
> will make them look more like remote video images

We used to have that, before my time, I think. Maybe even a "connecting..." flicker animation? @Brianetta seems to still be around and hearkens back to those days.

Without seeing the result, I'll just say: Needs more scan lines! ;)

bszlrd:
I think that size works, but the placement is a bit haphazzard:
face-misalignment

It would benefit from aligning the button and the portrait to the fuel bars. Sizing the portrait to the width of the launch button:
face-aligned

That scale plays nicely with my facegen WIP too.
And I think that new whitespace could be utilized for your TWR readouts. Especially since it is quite relevant information for launching. (which could benefit from a progress bar, I'll jot that down to that PR in more detail).
face-aligned-newfacegen

The noise overlay approach won't play nicely with some of the examples provided, on the small BBS scale (although that one is likely too much even on the larger scales:
image

Also, in general, if you are doing visual changes, I would appreciate screenshots from the get-go, less opaque that way.

sturnclaw:
Due to the position in the compositing pipeline that noise is applied (i.e. before downscaling), high-frequency noise textures seem unlikely to produce good results. Otherwise, I like the idea at play here - thanks!

I definitely agree with @bszlrd regarding the sizing and positioning of the face in the station lobby screen.

sam52796:
I've aligned the lobby Request Launch button as requested.

I also made some of the pixel effects larger scale. But in terms of it not looking good scaled down, in a way that's the point - the jankiness of the filter distracts from the faces themselves. Still, anyone else can submit an alternative set of PNGs if they want - there's nothing special about the noise images themselves.

PR #6385, opened 2026-07-09
BulkShips fix for running out of parking slots

sam52796:
When the game tried to allocate parked "bulk" ships outside of stations, there was a tiny chance of seeing a crash when the code ran out of parking slots. 

Today, @impaktor was our winner! Congratulations! 

Unfortunately, with the adoption of this PR, we shouldn't be seeing any more winners in this area. Please try other game features regularly in order to get additional chances of winning.


sturnclaw (approved):
Congratulations, your winnings are... a more stable game!

PR #6387, opened 2026-07-10
Implement Circular Thrust Indicator

sturnclaw:
Per discussion on https://github.com/pioneerspacesim/pioneer/pull/6354#issuecomment-4928647819, this PR implements @bszlrd's mockup for a replacement ship thrust indicator. While this indicator does not show the direction of ship velocity, that information is far better conveyed using the prograde/flight-path marker on the main hud.

image

Sizing and positioning is currently placeholder; ideally it will be integrated into the central flight reticle at a slightly smaller size. The indicator scales smoothly across a range of screen sizes; it has been tested at 1280x720 and 2560x1440 in addition to 1920x1080.

### DIFF
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -1320,8 +1320,8 @@
     "message": "The speed relative to the target"
   },
   "HUD_THRUST_INDICATOR": {
-    "description": "Tooltip: describe what the thrust/velocity indicator shows",
-    "message": "Show current velocity for each direction (thin lines) and thrust (thick background)"
+    "description": "Tooltip: describe what the thrust indicator shows",
+    "message": "Show current thrust levels for maneuvering (outer ring) and primary or retro engines (center circle)"
   },
   "HUD_WARNING_DESCENT_RATE": {
     "description": "Shown on HUD as a warning flasher",
--- a/data/lang/ui-core/en.json
+++ b/data/lang/ui-core/en.json
@@ -1319,6 +1319,10 @@
     "description": "Tooltip: The speed relative to the target.",
     "message": "The speed relative to the target"
   },
+  "HUD_GRAVITY_INDICATOR": {
+    "description": "Tooltip: describe what the gravity indicator shows",
+    "message": "Local gravity acceleration (scale: ship maximum acceleration)"
+  },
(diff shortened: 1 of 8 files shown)


impaktor:
Old one is removed?

sturnclaw:
> Old one is removed?

Yes. It took up significantly more layout space and the new circular gauge style is something that @bszlrd's mockups are trending towards. It can be resurrected from the git history if needed for future use (e.g. for in-cockpit screens).

sturnclaw:
image

I've also added a generic circular widget for displaying things like TWR, acceleration, temperature, and more. A reference implementation has been provided which displays the current gravitational acceleration relative to the ship's maximum acceleration.

impaktor:
Is this one as clear as the old one? Looking at the vid uploaded on irc, I see numbers change and bars grow, but I'm not sure I understand what they mean, and my mind starts again to think about the need for some in-game manual / tutorial... / pop-up window with explanations

https://github.com/user-attachments/assets/91aeb987-b005-4e74-a9ac-1fe2d2414b9d

bszlrd:
I think it is clear. But some kind of tutorial would be important anyway.

sturnclaw:
The local gravity indicator (right circle with the 'x.x G' text) is a placeholder demonstration of the widget; it's meant to be a thrust-acceleration limit controller. The current thrust indicator is fairly self-evident when under thrust, but I may extend it to render a mostly-transparent thruster icon overtop to make it extremely clear.

I'm mostly looking for feedback on the thrust indicator widget (left circle); I have another branch in progress (see screenshots on IRC) that is cleaning up the central reticle so that these widgets can move up there to more closely match the mockup they come from.

I'll push a few commits to this branch that fix a bug I found in the data feed for the gravity indicator widget; if you're in orbit of a space station I inadvertently introduced an infinite loop...

sturnclaw:
> my mind starts again to think about the need for some in-game manual / tutorial... / pop-up window with explanations

All of the widgets I've implemented have mouseover tooltips explaining the type and meaning of the data they display.

sturnclaw:
Going to merge this branch as I've not heard any desire to keep the old thrust indicator. I have another PR in the pipeline that will provide final positioning for these widgets.

PR #6388, opened 2026-07-10
Add text shadow capability

sturnclaw:
This PR implements `ui.textShadowed()` and `ui.addTextShadowed()` which uses a simple duplicate-and-modify approach to render pairs of text-and-shadow glyphs for each character in a minimally-intrusive fashion. It only works for text that is rendered via the above two methods, so e.g. you cannot currently use text shadows inside a button.

As it is late, I've not converted *anything* in the UI to use this yet. I'll leave it up to other bold contributors to migrate parts of the flight UI to use shadowed text rendering.


impaktor:
shadow is intended to make white text on HUD when rendered on white star visible?

bszlrd:
Yes, it can be helpful over bright backgrounds

fluffyfreak:
This is how I realised that Imgui still doesn't use (M/)[SDF fonts](https://www.redblobgames.com/articles/sdf-fonts/), or [Slug](https://github.com/EricLengyel/Slug) for rendering text. I see there are some attempts to add SDK rendering to it but none have been adopted. Apparently 1.92 will be getting font scaling.

sturnclaw:
@fluffyfreak if you're interested in improving font rendering, I'd love it if you'd do the legwork to get libfreetype in pioneer-thirdparty for Windows builds. ImGUI can use either stb_truetype (our current backend) or FreeType to do font rasterization, and FreeType does a much better job of rasterizing font glyphs.

Slug is definitely an option I'd like to look at supporting, though more for 3d text (e.g. cockpits) than 2d text. Most of our current UI font rendering is adequately served by traditional glyph-rect font rendering.

PR #6390, opened 2026-07-10
Make probability of lava depend on volcanicity

### LINKED ISSUE #6371: Broken lava generation
After commit d49b6b5504ed399237c73bcaab8ed674490b661d planetary lava gone missing

### Observed behaviour
Volcanic planets does not have any lava

### Expected behaviour
Volcanic planets are rendered with lava
Image

### Steps to reproduce
Start near Terminus at Sagittarius A* [-3125, 0, 0]

20260203 is affected

fluffyfreak:
This attempts to improve the planets with surface lava a little bit.
It makes the probabilty of there being surface lava higher on worlds with higher levels of volcanic activity.
Add randomness to the `octavenoise` used in locating the lava so that it varies for each world and stops it clumping up in one big area as seemed common before.

Fixes #6371 I think by making the chance of surface lava dependant on the volcanicity of the planet. It might increase the chance of surface lava lakes a bit too much but this was very hard to test.

### DIFF
--- a/src/terrain/TerrainColorVolcanic.cpp
+++ b/src/terrain/TerrainColorVolcanic.cpp
@@ -13,9 +13,12 @@ template <>
 TerrainColorFractal<TerrainColorVolcanic>::TerrainColorFractal(const SystemBody *body) :
 	Terrain(body)
 {
-	// 50 percent chance of there being exposed lava
-	if (m_rand.Int32(100) > 50)
+	// percent chance of there being exposed lava based on planets volcanicity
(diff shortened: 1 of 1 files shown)


fluffyfreak:
Also this _might_ require a savegame bump so I've added that tag just-in-case

sturnclaw:
While this could technically invalidate positioning of the player for saves landed on lava worlds, it's not going to cause outright crashes like anything that changes SystemBody numbering.

PR #6397, opened 2026-07-15
Improve flight reticle display, add orbit apoapsis/periapsis information

sturnclaw:
Screenshot From 2026-07-14 21-27-05

This PR follows on to #6388 by converting the flight reticle and planetary info displays to use shadowed text rendering, and slightly changes the information displayed there.

@bszlrd has improved icons in the pipeline, but I'm not going to wait on them to merge; they can be updated in master.

This PR patches ImGUI with a new window flag, `NoCaptureMouse`. This flag should only be used with transparent full-screen windows; it prevents hovering and clicking the window background from gaining exclusive control of the mouse (which would prevent the game from reading those inputs).


bszlrd:
Neat! I will make the icons soon. (Any other that might be needed in the near future?)

I've noticed that the orbit planner and body list or any other panel in the system map and system overview are not working. There's an empty debug window, and the panels cannot be clicked.
screenshot-20260715-152313
screenshot-20260715-152348
screenshot-20260715-153016

bszlrd:
Also noticed that the apses readouts are giving negative values when you are on an escape trajectory. I'd suggest the apoases to say ESC in that case, and the periapsis then acts like a closest approach distance. Which is not, it seems. It is a little bit off (see screenshot).
image

sturnclaw:
> Also noticed that the apses readouts are giving negative values when you are on an escape trajectory. I'd suggest the apoases to say ESC in that case, and the periapsis then acts like a closest approach distance. Which is not, it seems. It is a little bit off (see screenshot).

The existing code that derives an orbit from the current flight state is... questionable. I'll see what I can do in a reasonable timeframe to clean up the presentation when on an escape trajectory, but I've spent more than enough time on this PR and computing a more accurate orbital state is very far outside of scope for this PR.

> I've noticed that the orbit planner and body list or any other panel in the system map and system overview are not working. There's an empty debug window, and the panels cannot be clicked.

Thanks! I thought I fully fixed this one, but it seems a corner case of some kind slipped through. Will get the map screen back to working.

sturnclaw:
> Which is not, it seems. It is a little bit off (see screenshot).

As far as I know this is actually accurate; the bottom-right periapsis value is using altitude-sea-level (using body radius), while the frame gauge displays terrain/radar altimetry.

sturnclaw:
> I've noticed that the orbit planner and body list or any other panel in the system map and system overview are not working. There's an empty debug window, and the panels cannot be clicked.

This should now be fixed. I had to make some sweeping changes to how the View system worked in C++; it now submits its own "root" pass-through window to act as a host for labels / other cosmetic UI elements. The lack of this root window was causing certain primitive elements (mostly the horizontal separators between sidebar buttons and the sidebars themselves) to be submitted to the implicit Debug window.

This change also fully fixes mouse input routing to all screens of the game; though a lot has changed under the hood the behavior should be identical to the user. Feedback and bug reports would be welcome!

robothauler:
In some cases, the mission screen does not update.
E.g. start a new Game.
Switch to the mission screen.
Go to the BB.
Accept a mission.
Switch back to the mission screen.

sturnclaw:
> In some cases, the mission screen does not update.

This is now fixed. I moved tracking of refresh-on-switch state into C++, which simplified the code surrounding that feature. Testing is appreciated!

robothauler:
> > In some cases, the mission screen does not update.
> 
> This is now fixed. I moved tracking of refresh-on-switch state into C++, which simplified the code surrounding that feature. Testing is appreciated!

Looks good! That also seems to solve the problem with the right mouse button in the other screens that I found a few minutes ago.

PR #6398, opened 2026-07-15
Scale down the skybox in SystemMap

### LINKED ISSUE #6369: System map - Skybox background glitch
Image

The skybox glitches in the system map. It's not really visible with our only default skybox, but if you try others out there, like pebblegardens skyboxes.zip, there will be the occasional corners acting up.

### Observed behaviour
In the system map, when you have a lighter background, such as you may see with a nebula in the background, the corners of the skybox background will glitch out when you rotate the view. Essentially, where the corners meet, they will display various black triangles.

### Steps to reproduce
- Use a lighter skybox such as the one zipped below (used in the image above). Just drop it in as a replacement and enjoy the new sky.
- Select the System map and either click the middle mouse button to rotate, or use the icon controller at the bottom. Move it around to rotate the system left/right and up/down, until the fine triangles show themselves.

Ubuntu Mate 24.04.04
master (3dea523) 

[output.txt](https://github.com/user-attachments/files/29163207/output.txt)
Grey skybox for testing: [skybox.dds.zip](https://github.com/user-attachments/files/29163043/skybox.dds.zip)

fluffyfreak:
Scale down the skybox so that it no longer clips in the SystemMap Orrery view

Fixes #6369

### DIFF
--- a/src/SystemView.cpp
+++ b/src/SystemView.cpp
@@ -636,6 +636,8 @@ void SystemMapViewport::DrawOrreryView()
 	trans2bg.RotateX(DEG2RAD(-m_rot_x));
 	trans2bg.RotateY(DEG2RAD(-m_rot_y));
 
+	trans2bg.Scale(.5);
+
 	m_background->SetIntensity(0.6);
 	m_background->Draw(trans2bg);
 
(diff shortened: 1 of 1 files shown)


fluffyfreak:
Merged as just a tiny change that fixes the targetted issue

PR #6399, opened 2026-07-19
Language options menu - add url to translation contrib wiki page

impaktor:
As the commits suggests:
1. update broken SSC links
2. font size for language selection tab in options was larger than the other tabs, this unifies it to be equally small (I note 'About' menu has larger font... mostly)
3. @bszlrd suggested adding a "breadcrumb" link to translation wiki from the language selection page. Alas, I had imagined having it beneath the language selection table, rather than a part of it but my pigui skills are running low at the moment. As I was about to localize the string I realized it might actually be better to have it hard coded in English, so not translatable for now.

share opinions!
2026-07-19-101426_687x642_scrot

### DIFF
--- a/data/pigui/modules/settings-window.lua
+++ b/data/pigui/modules/settings-window.lua
@@ -411,19 +411,16 @@ local function showLanguageOptions()
 		return string.format("%s (%d%%)", name, pct)
 	end
 
-	ui.withFont(pionillium.heading, function()
-		ui.text(lui.LANGUAGE_RESTART_GAME_TO_APPLY .. ":")
+	ui.text(lui.LANGUAGE_RESTART_GAME_TO_APPLY .. ":")
+
+	ui.child("##LanguageList", Vector2(0, 0), function()
+		for _, lang in ipairs(langs) do
+			if ui.selectable(languageLabel(lang) .. "##" .. lang, Lang.currentLanguage==lang, {}) then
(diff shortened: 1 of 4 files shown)


data/pigui/modules/settings-window.lua
```diff
@@ -411,18 +411,19 @@ local function showLanguageOptions()
...
-	end)
+	ui.text(lui.LANGUAGE_RESTART_GAME_TO_APPLY .. ":")
 
-	ui.withFont(pionillium.body, function()
-		ui.child("##LanguageList", Vector2(0, 0), function()
-			for _, lang in ipairs(langs) do
-				if ui.selectable(languageLabel(lang) .. "##" .. lang, Lang.currentLanguage==lang, {}) then
-					Lang.SetCurrentLanguage(lang)
-				end
+	ui.child("##LanguageList", Vector2(0, 0), function()
```
sturnclaw: To place the link outside the list, you need only reserve space for it and move the `ui.text` calls outside the child window.

```suggestion
	ui.child("##LanguageList", Vector2(0, -ui.getTextLineHeightWithSpacing()), function()
```

impaktor: (I've removed my late night post of only getting it to half work, after sleeping on it, it now works, thanks)

impaktor:
Now looks like so:
2026-07-20-105519_672x634_scrot

PR #6401, opened 2026-07-20
Persist NPC names and faces

### LINKED ISSUE #6400: Character faces are not persistent
### Observed behaviour
Faces that are supposed to be persistent (police, lobby, goodstrader) are not persistent. 

### Expected behaviour
When landing on station X, I want to feel "home", see the regular ground crew, and rest my eyes on lovely Cathy in the lobby - she's always helpful granting every launch request and looking out for my fuel level. 

In the case of the lobby and police they have different names and faces each time.

The (real) Goodstrader has no character name by warehouse name and that is persistent as it should be, but face is different every time. It helps avoiding the police if you recognize "old pirate-Bob" with his out of date mustache and eye-patch operating the real black market.

Image
Image

sam52796:
Fixes #6400. 

The existing code set everything up to use a per-station and per-character fixed seed to ensure that NPCs stayed the same each time, but missed the last step where you have to pass that fixed seed into the name and face generation functions. So it ended up generating new names and faces from the global game engine randomizer each time instead.

This PR makes things behave the way I think they were intended, so now the station manager at Mars High doesn't change every time you land there, and Duva Holdings always has the same cool guy fronting it.

### DIFF
--- a/data/culture/culture.lua
+++ b/data/culture/culture.lua
@@ -92,7 +92,7 @@ end
 --
 function Culture:FirstName (isFemale, rand, culture)
 	local c = self.lookup[culture] or utils.chooseNormalized(self.weights, rand).lang
-	return c:FirstName(isFemale)
+	return c:FirstName(isFemale, rand)
 end
 
 --
@@ -147,7 +147,7 @@ function Culture:FullName (isFemale, rand, culture)
 
 	-- local debug_code = "(".. c.code .. ") "
 	-- return debug_code .. c:FullName(isFemale, rand)
-	return c:FullName(isFemale)
+	return c:FullName(isFemale, rand)
 end
 
 --
@@ -182,7 +182,7 @@ end
 function Culture:Names (isFemale, rand, culture)
 	-- if 'culture' given as a string, e.g. "Russian" use that
 	local c = self.lookup[culture] or utils.chooseNormalized(self.weights, rand).lang
-	return c:FirstName(isFemale), c:Surname(isFemale), c.name
+	return c:FirstName(isFemale, rand), c:Surname(isFemale, rand), c.name
 end
 
 return Culture
--- a/data/pigui/libs/face.lua
+++ b/data/pigui/libs/face.lua
@@ -2,6 +2,7 @@
 -- Licensed under the terms of the GPL v3. See licenses/GPL-3.txt
 
 local Engine = require 'Engine'
+local Rand = require 'Rand'
 local FaceTextureGenerator = require 'PiGui.Modules.Face'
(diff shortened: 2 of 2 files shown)


zonkmachine:
Thanks for fixing this! Maybe we now need a number of staff members (size relative to station/base size) to man station, workshop, the police and other roles? So we don't let people work literally 100% of their life.

impaktor:
> Maybe we now need a number of staff members

Feels like that would defeat the purpose of recognizing the NPC, since then you'd have to have at least two faces for each lobby, police, etc. and station to familiarize yourself with. I'd consider it "over thinking". I wouldn't want to have "shift workers" unless there's far more mechanics in that, e.g. night lights at the station (when walking on base), or somehow the station feels "night" when interacting with it, and each station has its own uniforms and insignia, then you would still feel the station was "Torvald's Platform" and not "Job's pad".

Night crew would have 5 O'clock shade, and look sleep deprived and hold a cup of coffee, and have some emblem "night shift", and background would look "night".

sam52796:
I agree. I think there are shift workers there, but by enormous coincidence the same person always happens to be on duty whenever you are at the station.

Land at a different time of day? Ah, Cathy had to come in to cover someone else's shift today. Hello Cathy.

Spend 24 hours staring at Cathy's face with time acceleration? Oh, there was a glitch in the system and it didn't update the picture for the two other shifts that happened in the meantime, and now it's Cathy's shift again. Good ol' Cathy.

impaktor:
I'm more thinking that "spacers" work (close to) 24/7, to pay the bills and save up to their own ship to "make it big" in the universe. In the mean time, they're running on coffee (or "coffee") to stay awake yet another long shift.

zonkmachine:
> > 
> 
> Feels like that would defeat the purpose of recognizing the NPC, since then you'd have to have at least two faces for each lobby, police, etc. and station to familiarize yourself with.

It would not, as you would start to see the NPC pattern of a station if you stopped by 5+ times or so. If you just drop by once, none of this matters. In real life, you wouldn't expect to recognize individual staff members on a large airport unless you traffic it weekly for a longer time. You would be surprised though if you didn't recognize the staff at a small local airfield after some four landings in a month.

> I'd consider it "over thinking".

I agree on that label.

impaktor:
Looks good to me. 

When testing I was surprised staring station has 2 legit goods traders, and 1 fake/police. I imagine the opposite, police setting up multiple "honey pots", and the true black market one will have "retired" any legit competition, again, you would only need to remember one face / goods trader name / station. But that's outside the scope of this PR - just thinking out loud. 

Thanks!

PR #6403, opened 2026-07-26
Adding missing DSMiner engines, adjusting plume sizes

### LINKED ISSUE #6344: DSMiner engines
Image

### Observed behaviour

The Deep space miner has two engine plumes with no corresponding engine bell. These are the outmost rear engines.

Ubunt Mate 24.04.4 LTS (Noble Numbat) 64-bit

__My output.txt (required)__
[output.txt](https://github.com/user-attachments/files/27495395/output.txt)

bszlrd:
Fixes #6344 
This PR adds two engines to already existing thruster tags at the back of the DSMiner. Figured it has a ridiculous amount of engines already, so to [quote a classic.](https://www.youtube.com/watch?v=QHZ48AE3TOI)
I've also adjusted the scales of the plumes a bit, to match the engine and thruster sizes, and the scale of the ship better. Arguably the largest engine could even have it a bit larger, but that would imply a faster ship.
image


zonkmachine (approved):
NIce! Based on a quick in game comparison it looks good.

zonkmachine:
For a future remake of the DS Miner, I like the container structure (octagonal profile). It cold be repurposed for something like a large bulk freighter.

bszlrd:
I do have some sketches from back then, even blockouts

PR #6408, opened 2026-08-07
Atmospheric shield interactions

sam52796:
Wouldn't it be cool if your ship's shield had a re-entry glow effect when you entered an atmosphere at high speed?

With this PR, it does.

In Sfx.h there are now two constants:
```
	inline constexpr float REENTRY_GLOW_MIN_SPEED = 100.0f;
	inline constexpr float REENTRY_GLOW_MAX_SPEED = 200.0f;
```
With these settings, in a 1.0 atm environment, the shield starts glowing at 100m/s. The speed is inversely correlated to the pressure, so this means on Mars, with a weaker atmosphere, the shield will start glowing at around 160m/s. On a dense 2.0 atm planet you'd see it at 50m/s. 

For the purposes of testing I've set the above values really low, so you can all see the effect easily just by taking off from Mars and firing the thrusters for a few seconds.  For release I would expect to set the thresholds much higher, maybe 400 min and 600 max, but that can be debated.

image

image


sam52796:
For some reason, this PR caused a bunch of unit tests to throw exceptions in the audio backend. Not sure why this one would see the problem when other recent PRs didn't.

So I've added a second commit to add a null audio backend that can be loaded for unit tests and other headless modes. Then the rest of the code can continue to make sound-related API calls without issue. 

Unit tests now complete without error.

Mc-Pain:
Maybe make shields glow on taking damage and start damaging shields based on drag?

zonkmachine:
> For some reason, this PR caused a bunch of unit tests to throw exceptions in the audio backend. Not sure why this one would see the problem when other recent PRs didn't.

I haven't used the tests a lot but they seem to be all clear on my machine.
_Edit: Ubuntu 24.04.4 LTS (Noble Numbat) 64-bit)._

```
$ ./build/unittest 
...
[doctest] test cases:  10 |  10 passed | 0 failed | 0 skipped
[doctest] assertions: 966 | 966 passed | 0 failed |
[doctest] Status: SUCCESS!
```

sam52796:
The weird unit test thing could be because I started this several months ago, before the OpenAL merge. That could also be why the same commit shows up twice, and Git is just getting confused and compiling some strange hybrid of pre- and post- OpenAL changes. I'm certainly a bit confused myself as to what's going on with that.

If you just take the second commit, [Re-entry glow on shields at high velocities within atmospheres](https://github.com/pioneerspacesim/pioneer/pull/6408/changes/9f6cf71bab0de5e6f55d876c2be4ebf12a1bdd03), then you might not see any unit test failures, in which case the null audio backend one wouldn't be necessary.

zonkmachine:
> If you just take the second commit, [Re-entry glow on shields at high velocities within atmospheres](https://github.com/pioneerspacesim/pioneer/pull/6408/changes/9f6cf71bab0de5e6f55d876c2be4ebf12a1bdd03), then you might not see any unit test failures, in which case the null audio backend one wouldn't be necessary.

Right. I missed the part of a workaraound. If I back one commit the tests fails indeed.

sam52796:
I've moved the unit test fix to its own PR #6413, so that it can potentially be merged in separately if appropriate.

This PR now has a single commit which is the shield re-entry glow effect.

sturnclaw:
I don't think this PR will make the cutoff of Monday for the Midsommar release if we still need to debate the mechanism by which the shield starts glowing.

That being said, I'd like to begin the debate by suggesting the [Sutton-Graves equation](https://tfaws.nasa.gov/TFAWS12/Proceedings/Aerothermodynamics%20Course.pdf) for surface thermal flux $q_s$ over a blunt body of effective radius $R_n$ during re-entry:

$$q_s = k \left( \frac {\rho} {R_n} \right) ^ {\frac 1 2} V ^ 3$$

In the reference material, $k$ is suggested as a constant `1.7415e-4` for earth-like atmospheres and `1.9027e-4` for Martian atmospheres. This constant appears to be determined by the gaseous composition of the atmosphere, and we could likely approximate a value based on the current atmospheric parameters of the system body.

With that adopted, the atmospheric heating effect would be widely applicable to all bodies in the game by specifying a minimum and maximum thermal flux value to scale the visibility of the effect.

sam52796:
I think we can avoid overcomplicating the shield glow effect by saying that it's not heating per se - it's interactions between particles in the atmosphere and an unknown futuristic shield technology.

Given that the effect isn't core to the game in any sense - it's only a little bit of extra visual flair that might make a player think "nice" for a couple of seconds when they run into the circumstances that trigger the shield glow, I don't think it's worth spending much time making it accurate to imaginary science. 

(To be clear - I don't think the Sutton-Graves equation is imaginary, but I think the shields are.)

So - alternative proposal - we release as is, and then if a contributor feels that the effect is encountered too easily, they can try implementing whatever modelling they like for a more accurate simulation of the imaginary science. 

Or just increase those two constants a bit.

sturnclaw (approved):
350 min and 600 max seem like good starter values to me for a 1atm reference. I can implement Sutton-Graves later to make the effect more reactive to gas giants and varying ship sizes and shield orientations.

PR #6411, opened 2026-08-08
Missile exhaust improvements

sam52796:
The existing missile smoke sprites are underwhelming. But now we have a more whelming thruster exhaust system available, so this PR switches missiles to use that instead.

image

As a result, missiles have visible exhaust, even in space (let's assume they use chemical rockets) which makes them easier to see and also hopefully looks cooler, especially when the missile first detaches from the ship and the exhaust curves in a big arc while orientating towards the target.

image

The first commit is a slight refactor of the thruster exhaust code to put shared particle spawner code into its own file, and then modifications to the missile code to call that instead of the old billboards. The exhaust effect now has a configurable noise parameter, so that thruster plumes on take-off still look cloud-like, but space missile exhaust looks more like a constant stream. Because, as we all know, there is no noise in space. 

The second commit is because the new missile exhaust revealed that a missile's orientation goes haywire as it approaches a target, as the velocity difference gets smaller and normalizing tiny vectors results in huge directional over-corrections. Using a simpler "point at" technique has better visual results once the missile is more or less on target, and the missiles seem just as effective at hitting things.

### DIFF
--- a/data/shaders/opengl/exhaust.frag
+++ b/data/shaders/opengl/exhaust.frag
@@ -6,6 +6,7 @@
 
 in vec2 v_uv;
 in vec4 v_color;
+in float v_noiseStrength;
 
 out vec4 frag_color;
 
@@ -39,7 +40,7 @@ void main(void)
 	// Increase noise_cells for smaller blobs, decrease for larger ones.
 	const float noise_cells = 3.0;
 	vec2 pos = v_uv * noise_cells;
-	float noise_val = noise(pos) * 2.0 + 0.5;
+	float noise_val = noise(pos) * v_noiseStrength + 0.5;
 
 	float alpha = v_color.a * edge * noise_val;
 	frag_color = vec4(v_color.rgb, alpha);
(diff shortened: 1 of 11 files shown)


sam52796:
Looks like this PR also shows some failing unit tests due to an issue with the sound API. 

I have no idea what's going on. Either my repository is seriously borked in a way that makes Github include the wrong sound code when compiling unrelated changes upstream, or else the unit tests in master have been broken since the OpenAL merge and nobody else has noticed. Neither of these seem particularly likely.

Either way my guess is that the changes in [Null audio backend for running headless unit tests](https://github.com/pioneerspacesim/pioneer/pull/6408/changes/6450e45da5b2574c337fb2444883ed6ec2cc71dd) would make the tests pass.

sturnclaw (approved):

sturnclaw:
Very happy with this, proper exhaust particles for missiles will make it much easier to debug and improve their trajectory management. Thanks!

PR #6413, opened 2026-08-12
Null audio backend for running headless unit tests

sam52796:
This PR adds a null audio backend that can be loaded for unit tests and other headless modes. So now code can continue to make sound-related API calls in those contexts without issue.

This should fix the recent spate of integration test failures.

### DIFF
--- a/src/Pi.cpp
+++ b/src/Pi.cpp
@@ -406,6 +406,10 @@ void Pi::App::OnStartup()
 
 	QueueLifecycle(m_loader);
 
+	// Headless mode skips the normal startup screen, so initialize sound here.
+	if (m_noGui)
+		Sound::Init(Pi::config->String("AudioBackend"));
+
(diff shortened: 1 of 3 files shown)


sturnclaw (approved):

src/Pi.cpp
```diff
@@ -537,12 +541,11 @@ void StartupScreen::Start()
 	Pi::pigui->EndFrame();
 
 	AddStep("Sound::Init", []() {
-		if (Pi::GetApp()->HeadlessMode() || Pi::config->Int("DisableSound"))
-			return;
-
 		Sound::Init(Pi::config->String("AudioBackend"));
```
sturnclaw, line 544: In the long run, I'd prefer the choice of `Null` AudioBackend be made here (and at other callsites of Sound::Init) instead of automatically overriding the requested backend inside `Sound::Init`. This is an acceptable solution for the moment however.

PR #6423, opened 2026-09-06
sound: Sound mixer in settings window have dB scaling

mfep:
With this change, the human ear's logarithmic reception is respected, that makes the volume sliders more pleasant to use.

This change applies decibel scaling to the master, SFX and music volume setting in the settings menu. This is achieved by applying the transformation in the `Sound` and `SoundMusic` wrapper, and the audio backends receive the scaled value already. Therefore the configuration and UI logic did not have to change.

The current hard-coded setting is a scaling between -40 and 0 dB which I found pretty pleasant.

### DIFF
--- a/src/sound/AlAudioBackend.cpp
+++ b/src/sound/AlAudioBackend.cpp
@@ -207,11 +207,6 @@ void Sound::AlAudioBackend::SetMasterVolume(float vol)
 	CHECK_OPENAL_ERROR(alListenerf, AL_GAIN, m_masterVolume);
 }
 
-float Sound::AlAudioBackend::GetMasterVolume()
-{
-	return m_masterVolume;
-}
-
 void Sound::AlAudioBackend::SetSfxVolume(float vol)
 {
 	m_sfxVolume = vol;
@@ -220,11 +215,6 @@ void Sound::AlAudioBackend::SetSfxVolume(float vol)
 			ev.SetVolume(m_sfxVolume);
 		}
 	}
-}
-
-float Sound::AlAudioBackend::GetSfxVolume()
-{
-	return m_sfxVolume;
 }
 
 void Sound::AlAudioBackend::AddSample(std::string_view key, Sample &&sample)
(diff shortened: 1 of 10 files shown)


sturnclaw (changes requested):
I would encourage you to test this change with an actual software audio decibel meter (on e.g. the system audio bus).

This PR actually has the opposite effect of what you describe, resulting in sound being exponentially dampened rather than being adjusted according to the perceptual loudness. I've tested two situations at 50% and 100% volume under both OpenAL and SDL backends.

1. Music immediately after starting a new game. Normal 100% volume is approximately -10 dB. Under this PR, 50% music volume is -30 dB, while in master 50% volume is -14 dB.
2. Firing a pulse laser in space. At normal 100% volume the pulse laser effect is 0 dB. Under this PR at 50% master volume, the pulse laser effect is -20 dB (growing to -17 dB when master is 100% and effect is 50% due to the background ambience track). On master, it is -6 dB with either slider set to 50%.

I believe my findings show that both audio implementations already properly apply logarithmic scaling to incoming volume controls, resulting in a 50% volume control properly corresponding to a 50% perceptual volume reduction. If your experimental results differ on a platform other than Linux we could look at adding a per-platform workaround for that specific bug.

mfep:
@sturnclaw It is a good idea to measure the output signal amplitude from outside the application. Your above results actually prove that the existing audio backends apply the gain values linearly, and also are in line with what this PR should achieve.

E.g. in current `master`, when the master volume is set to 50%, that translates to a 0.5 multiplier on the output signal's amplitude. That is `20*log10(0.5) ~= -6 dB`, exactly as you measured. If you set it to 75%, it is -2.5 dB and if you set it to 25% then the gain is around -12 dB. 10% is exactly -20 dB (based on the formula, didn't measure). The lower you go, the more dB each step inflicts. Now perceived loudness is closer to being linear on the dB scale, so you want each "tick" on the UI slider correspond to the same step value expressed in dB. The current setup makes it difficult to tune in quieter values, while the top of the range barely changes the perceived loudness.

This PR scales the sliders linearly on the dB scale between 0 dB (linear 1.0) and -40 dB (linear 0.01). The -40 dB is arbitrary, we could go lower at the loss of precision. The 50% setting stands for -20 dB, exactly as you measured.

sam52796:
I just tested the latest release on Windows. The music slider at least seemed to work the way I expected - the 50% level sounded about half as loud as the 100% level. Assuming that the other sliders use the same scale, I'm not sure what this change would improve.

PR #6427, opened 2026-09-08
Revert atmosphere radius to the pre-Rayleigh values

### LINKED ISSUE #6426: loading a game failed, lua error
### Observed behaviour

When loading my game, it ends as

Fatal: [T] modules/DeliverPackage/DeliverPackage.lua:239: attempt to index a nil value
stack traceback:
        [T] modules/DeliverPackage/DeliverPackage.lua:239: in function 'placeAdvert'
        [T] modules/DeliverPackage/DeliverPackage.lua:444: in function 'func'
        [T] libs/Event.lua:31: in function 'executor'
        [T] libs/Event.lua:79: in function '_Emit'
        [T] libs/Event.lua:209: in function <[T] libs/Event.lua:208>
Fatal: Frame instance deletion outside 'DeleteFrame' [0]
romaschw@localhost:~/pioneer-master> 

Other observation: The game freezes

### Steps to reproduce

Load this game

[2026-3-9.txt](https://github.com/user-attachments/files/31962502/2026-3-9.txt)

c9300a8 master branch

sam52796:
Fixes #6426.

When the system generation code is creating orbital starports, one of the parameters it looks at is the atmosphere height, to see if there is enough room between the planet and its moons for an orbital.

The Rayleigh code changed the height of planet atmospheres, from a threshold of ~0.001 atm to one of 1e-15 atm instead (plus a tropopause layer), this means the atmosphere is now much taller.

As a result, the code decides to skip one or more orbital starports that would have been created before. This then results in further changes to anything that is generated afterwards, since the RNG has been used differently. As the SystemBody list is now different, any save game that uses an index into a previous system body list will end up with the wrong body, or potentially nothing if there aren't enough bodies anymore.

This PR restores the old atmosphere height calculation when generating orbital stations, so then old save games will load again.

### DIFF
--- a/src/galaxy/StarSystemGenerator.cpp
+++ b/src/galaxy/StarSystemGenerator.cpp
@@ -393,6 +393,26 @@ fixedf<48> StarSystemLegacyGeneratorBase::CalcHillRadius(SystemBody *sbody) cons
 	}
 }
 
+// Normalized atmosphere top used for starport orbit clearance (planet radii).
+// This must stay as the pre-Rayleigh 10-scale-height formula so randomly-generated
+// body counts/indices remain stable for old saves. 
+static float LegacyStarportAtmosRadius(const SystemBody *sbody)
+{
+	const double radiusPlanet_in_m = sbody->GetRadiusAsFixed().ToDouble() * EARTH_RADIUS;
+	const double massPlanet_in_kg = sbody->GetMassAsFixed().ToDouble() * EARTH_MASS;
+	const double g = G * massPlanet_in_kg / (radiusPlanet_in_m * radiusPlanet_in_m);
+
+	double T = static_cast<double>(sbody->GetAverageTemp());
+	// XXX hack to avoid issues with sysgen giving 0 temps
+	if (T < 1)
+		T = 165;
+
+	const double M = sbody->GetType() == SystemBody::TYPE_PLANET_GAS_GIANT ? 0.0023139903 : 0.02897;
+	const float atmosScaleHeight = static_cast<float>(GAS_CONSTANT_R * T / (M * g));
+
+	return static_cast<float>(1.0f + static_cast<float>(10.0f * atmosScaleHeight) / sbody->GetRadius());
+}
+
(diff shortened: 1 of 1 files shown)


sam52796:
I realised that merging this would be a bad idea, since then any game saved with the latest release would potentially have issues for the opposite reason when loaded with this code active - there would be too many SystemBodies as additional orbital stations would be created.

I'm not sure there is any easy way to fix this, but we could potentially keep this PR around for a bit so that in theory someone could use it to load a previous saved game with the latest release if they really want to.

Mc-Pain:
With space stations, I think we're doing it wrong.

Discard space stations *silently* because otherwise they are being spawned inside atmosphere and then run into an unexpected problem is the least favorable thing to me.

Instead, I'd suggest taking periapsis, check for pressure, discard (not silently!) if pressure is more than a given threshold (either constant value = 0.001 atm or drag given v = first space velocity)

We should not alter system generation or modify manually-plotted systems silently.

sam52796:
Not sure how we would discard space stations non-silently. Have them burning up in the atmosphere? 😅

Either way, using a more physically precise mechanism to determine where stations can fit is a good idea, but however we do it, the number of system bodies will likely change, requiring a save-bump.

Mc-Pain:
> Not sure how we would discard space stations non-silently. Have them burning up in the atmosphere? 😅

The same way we alter surface station positions if they are ended buried up - write logs.

sturnclaw:
We're going to bump the savegame version in the upcoming February release. There's no point in maintaining backwards compatibility with saves from *this* version much less the previous one. The proper place to address this is in StarSystemGenerator.cpp, adjusting the logic that computes the starting altitude for (station and moon) orbital shells.

sam52796:
Agreed. This code addresses that we released a supposed non-save-bump version with a change that breaks some saves from the previous version. It's an adjustment to StarSystemGenerator.cpp to use the legacy atmosphere height for computing the starting altitude for orbital shells, so that those older saves will load.

Given that the next February release will (hopefully) include things like #6317 and #6330, so the systems will be generated differently anyway, this PR is only useful for a player who wants to load a previous version save with the latest features now, and who is comfortable with using Git, etc. So we should only keep the PR around until February at most.

PR #6434, opened 2026-09-11
Fix for radar not working with non-station target

### LINKED ISSUE #6431: Radar not displaying after loading some saved games
### Observed behaviour

When loading some saved games, the radar is not displayed in the HUD anymore.

Afterwards, loading any other saved game (even those that do not trigger the bug) keeps the radar hidden.

Restarting the application and loading a saved game not affected by the bug shows the radar normally.

I've noticed this since version 20260907. Some saved games are affected, others not (i.e., some games saved  when the bug was triggered and the radar was not being shown were saved properly and do not trigger the bug when loaded after an application reload).

### Expected behaviour

The radar is shown in the HUD (obviously when a radar is installed in the ship).

### Steps to reproduce

Load the attached saved game (renamed as "78.gz" to upload it here, just rename it to barely "78").

20260907 under Ubuntu Linux 26.04.

[78.gz](https://github.com/user-attachments/files/32103849/78.gz)

[output.txt](https://github.com/user-attachments/files/32103709/output.txt)

Image

sam52796:
Fixes #6431.

The new landing navigation aid for the radar crashes if you click on anything that isn't a station. So then the module manager disables the radar and you can't get it back.

Adding a one-line object type check fixes this.

### DIFF
--- a/data/pigui/modules/radar.lua
+++ b/data/pigui/modules/radar.lua
@@ -323,7 +323,7 @@ radar3d.draw = function(self, center)
 	end)
 
 	-- Display landing navaid
-	if navTarget and navTarget:GetAssignedBayNumber(Game.player) >= 0 then
+	if navTarget and navTarget:IsStation() and navTarget:GetAssignedBayNumber(Game.player) >= 0 then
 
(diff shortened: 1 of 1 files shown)


Bodasey:
error not observed any more

PR #6436, opened 2026-09-12
Add a probability that the person being searched for has already pass…

robothauler:
related to #6249

### DIFF
--- a/data/modules/FindPerson/FindPerson.lua
+++ b/data/modules/FindPerson/FindPerson.lua
@@ -224,6 +224,7 @@ local makeAdvert = function (station)
 
 	local female = Engine.rand:Integer(1) == 1
 	local gender = female and "_FEMALE" or "_MALE"
+	local dead = Engine.rand:Number(1) < 0.125
 
 	local flavour_number = Engine.rand:Integer(1, #flavours)
 	local flavour = flavours[flavour_number]
@@ -246,7 +247,7 @@ local makeAdvert = function (station)
 		introtext = l[introtext .. "_" .. intro_number],
 		flavour   = flavour,
 		client    = Character.New(),
-		wanted    = Character.New({ female = female }),
+		wanted    = Character.New({ female = female, dead = dead, }),
 		employee  = employee,
 		friend    = friend,
 		relative  = relative,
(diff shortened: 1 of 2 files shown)


RacerBG:
I don't know the context of the strings "Alive" and "Deceased" but if they are related to the gender of the person in question you should map them for male and female strings accordingly.

zonkmachine (commented):
I think this works well. Maybe one in an eighth chance of a dead target is a bit too much though. But then again, space wants is a rough place.

The lines may not work as well for the cases where the target is a relative of the customer. You may want to test for that and do a line like **"What are you saying? My {relative} is dead?"**. Something along those lines.

You could lighten up the dialogue a bit by not using the full name all the time. This line **"What do you mean? {wanted} has passed away?..."** currently gives the full name for {wanted} but you could alternate with just the first name for **"What do you mean? Jake is dead!"**

zonkmachine (commented):
When I am informed of the fate of the person I'm looking for, I also get access to information that isn't explained in the game. This may not be a problem but if the information could be handed over from an official source like a police station, that would be better. It feels especially lacking when I'm looking for a relative.
So, an informer directs you to the right planet and the police can provide you with the proper information. Maybe that's unnecessary or it could be a later fix. I think it works quite well as it is.

Here I'm tipped off.
Image

Now I have information about the individual living in Jericho.
Image

data/modules/FindPerson/FindPerson.lua
```diff
@@ -591,7 +591,7 @@ local buildMissionDescription = function (mission)
 		{ l.WANTED, mission.wanted.name },
 		{ l.SYSTEM, ui.Format.SystemPath(mission.location) },
 		{ l.SPACEPORT, mission.tipster and not mission.wanted.dead and mission.location:GetSystemBody().name or l.UNKNOWN },
-		{ l.STATUS, l[status] },
+		{ l.STATUS, l[status .. gender] },
```
zonkmachine, line 594: The default 'UNKNOWN' isn't gendered.

`Warning: Missing translation token 'UNKNOWN_MALE' in resource 'module-findperson/en.json'.`

PR #6438, opened 2026-09-15
Switch from float to double for eclipse occlusion calcs

sam52796:
If you fly to Cennet (1-1-0)/Ellasar/Mendel and land at the starport, then fast forward time until the sunrise on 23 Sep, you will see flickering illumination on the buildings and your ship just before the sun peeks over the horizon.

This is caused by float arithmetic not having enough precision to deal with planet radius level calculations of tiny differences in sun angles.

This PR switches from float to double in some key places, so the result is a nice smooth illumination effect instead of any flickering.


zonkmachine:
Yes, I've noticed the occasional flicker. Issue and fix tested and confirmed.

fluffyfreak (approved):
Yeah this looks like the right approach. Good job 👍

bszlrd:
I suspect this also fixes the flickering one might experience flying out of the planet's shadow?

PR #6440, opened 2026-09-16
Limit TWR mass display to 2 DP

### LINKED ISSUE #6439: TWR related data flicker during flight
Image
Image

### Observed behaviour
Under the Info tab, when the ship is in motion, the TWR, mass related data on the bottom right, flicker as trailing zeroes are truncated and the layout is updated in real time.

### Expected behaviour
The number formatting either retains the zeroes or is centered around the decimal point.

### Steps to reproduce
* Launch into orbit
* Select "Show personal info"
* Watch in the bottom right how the values change and sometimes flicker.

Pioneer master (9d14b42)
Ubuntu 24.04.4 LTS

[output.txt](https://github.com/user-attachments/files/32266032/output.txt)

@sam52796

sam52796:
Fixes #6439.

Limit the mass display for the TWR widget in the lower right of the screen to 2 DP, so it doesn't jump around as you fly and expend fuel.


zonkmachine:
Nice! This works for me.

